PR #245, opened 2020-06-28
Peer resultification and defect only

sinkingsugar:


libp2p/peer.nim
```diff
@@ -143,37 +148,51 @@ proc init*(pid: var PeerID, data: string): bool =
...
       result = true
 
-proc init*(t: typedesc[PeerID], data: openarray[byte]): PeerID {.inline.} =
+proc init*(t: typedesc[PeerID], data: openarray[byte]): Result[PeerID, cstring] {.inline.} =
   ## Create new peer id from raw binary representation ``data``.
-  if not init(result, data):
-    raise newException(PeerIDError, "Incorrect PeerID binary form")
+  var res: PeerID
+  if not init(res, data):
+    err("Incorrect PeerID binary form")
```
arnetheduck, line 155:
```suggestion
    err("peerid: incorrect binary form")
```

libp2p/protocols/identify.nim
```diff
@@ -140,16 +140,18 @@ proc identify*(p: Identify,
...
-    # because that is the only thing we
-    # have in most cases
-    if peer != remotePeerInfo.peerId:
-      trace "Peer ids don't match",
-            remote = peer.pretty(),
-            local = remotePeerInfo.id
-
-      raise newException(IdentityNoMatchError, "Peer ids don't match")
+    if peer.isErr:
+      raise newException(IdentityInvalidMsgError, "Could not derived a PeerID")
```
arnetheduck, line 144: can use `$peer.error` as msg here

libp2p/peer.nim
```diff
@@ -8,10 +8,15 @@
 ## those terms.
 
 ## This module implementes API for libp2p peer.
+
```
arnetheduck, line 11: this module looks like it should be called `peerid` - wonder if it's possible to move the peerid type in here as well since it should be standalone

libp2p/peer.nim
sinkingsugar: oops :)

libp2p/peer.nim
sinkingsugar: you mean `PeerInfo`? cos `PeerID` is here

arnetheduck: the `type` section is not in here

arnetheduck:
also needs conflict resolution

libp2p/peer.nim
sinkingsugar: hmm? this one https://github.com/status-im/nim-libp2p/blob/c788a6a3c0ca614be4c67cc566aee6877da0250b/libp2p/peer.nim#L22 ?

arnetheduck: oops, I don't know how I missed it - clearly it's right where it should be :)

PR #249, opened 2020-07-04
reuse single RNG instance for all crypto key generation

arnetheduck:


sinkingsugar:
I would have gone for a static global (or thread local even) rather then passing around rng.. anyway it would have been really a single instance without touching any other global from within and so harmless.

But nevertheless since it's done this is fine too.

arnetheduck:
> I would have gone for a static global 

then it can't be shared with other libraries - ie `nim-eth` needs one etc - also, there's the question of seeding it - it should be reseeded which can typically be done by the owner

arnetheduck:
also, globals and threadlocals introduce interop issues with other languages and have poor platform support and bugs

PR #244, opened 2020-06-27
Publish backporting

sinkingsugar:
It started as a backport but it's much bigger now

replenish and fanout management is done during heartbeat,
solves fanout/gossip peers being removed earlier then expected..
publish now has no maintenance tasks and returns an int for how many peers it reached.

p.s.
not sure about spec conformity yet

PR #250, opened 2020-07-04
fix publishing

dryajov:
This is a WIP to fix some publishing issues, it duplicates some of the work in #244 so we might consider cherry picking either into this branch or into the #244 branch. I'm leaving it here for now as a reference ( @sinkingsugar  )


sinkingsugar:
I like it.. I wanted to do this refactor but it would have taken me much more to figure it out.. don't worry about my further changes.. I just added some locks to heartbeat.. which worked 99% of the times.. still triggered nil peer tho

tests/pubsub/testfloodsub.nim
```diff
@@ -60,7 +60,7 @@ suite "FloodSub":
       await nodes[1].subscribe("foobar", handler)
       await waitSub(nodes[0], nodes[1], "foobar")
 
-      await nodes[0].publish("foobar", "Hello!".toBytes())
+      discard await nodes[0].publish("foobar", "Hello!".toBytes())
```
arnetheduck, line 63: `check: await... == 1` (or whatever it should be?)

arnetheduck: `check: 1 == await ... ` maybe easier to read

sinkingsugar: that's from my branch btw 😄 , the thing is that this does not apply to flood, it will always be 0.

sinkingsugar: btw I find it useful (nbc side for example I'm using it) to have this int with gossip. but we can remove it once things stabilize if you guys don't think it's necessary.. it's a debug tool tho

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -181,36 +206,46 @@ proc dropFanoutPeers(g: GossipSub) {.async.} =
...
-              result[id] = ControlMessage()
-            result[id].ihave.add(ihave)
+    for id in allPeers:
+      if result.len >= GossipSubD:
+        trace "got gossip peers", peers = result.len
+        break
 
-    libp2p_gossipsub_peers_per_topic_gossipsub
-      .set(g.gossipsub.getOrDefault(topic).len.int64, labelValues = [topic])
+      if allPeers.len == 0:
```
arnetheduck, line 238: allpeers can't be zero in here

PR #253, opened 2020-07-08
Fix gossip messages seqno according to spec

sinkingsugar:
Quoting @oskarth 

> Am I reading it right that we currently don't have linearly increasing seqno in PubSub? https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/rpc/message.nim#L65-L66 It seems like it is picking a random set of bytes here. This seems different from https://github.com/libp2p/specs/blob/master/pubsub/README.md#the-message

Fixes

### DIFF
--- a/libp2p/peerinfo.nim
+++ b/libp2p/peerinfo.nim
@@ -39,7 +39,6 @@ type
       privateKey*: PrivateKey
     of HasPublic:
       key: Option[PublicKey]
-    seqno*: uint64
 
 proc id*(p: PeerInfo): string =
   if not(isNil(p)):
--- a/libp2p/peerinfo.nim
+++ b/libp2p/peerinfo.nim
@@ -39,6 +39,7 @@ type
       privateKey*: PrivateKey
     of HasPublic:
       key: Option[PublicKey]
+    seqno*: uint64
 
 proc id*(p: PeerInfo): string =
   if not(isNil(p)):
--- a/libp2p/protocols/pubsub/rpc/message.nim
+++ b/libp2p/protocols/pubsub/rpc/message.nim
@@ -57,22 +57,17 @@ proc verify*(m: Message, p: PeerInfo): bool =
   else:
     libp2p_pubsub_sig_verify_failure.inc()
 
-# cannot explicit init but:
-# Thread locals are initialized to binary zero (usually default(T)). (Araq)
-var tSeqno {.threadvar.}: uint64 
-
 proc init*(
     T: type Message,
     p: PeerInfo,
     data: seq[byte],
     topic: string,
+    seqno: uint64,
     sign: bool = true): Message {.gcsafe, raises: [CatchableError, Defect].} =
-  # peer is a ref obj
-  inc tSeqno
   result = Message(
     fromPeer: p.peerId,
     data: data,
-    seqno: @(tSeqno.toBytesBE), # unefficient, fine for now
+    seqno: @(seqno.toBytesBE), # unefficient, fine for now
     topicIDs: @[topic])
 
   if sign and p.publicKey.isSome:
--- a/libp2p/protocols/pubsub/rpc/message.nim
+++ b/libp2p/protocols/pubsub/rpc/message.nim
@@ -19,6 +19,7 @@ import messages, protobuf,
        ../../../peerinfo,
        ../../../crypto/crypto,
        ../../../protobuf/minprotobuf
+import stew/endians2
 
 logScope:
   topics = "pubsubmessage"
@@ -63,13 +64,12 @@ proc init*(
     topic: string,
     sign: bool = true): Message {.gcsafe, raises: [CatchableError, Defect].} =
   var seqno: seq[byte] = newSeq[byte](8)
-  if randomBytes(addr seqno[0], 8) <= 0:
-    raise (ref CatchableError)(msg: "Cannot get randomness for message")
-
+  # peer is a ref obj
+  inc p.seqno
   result = Message(
     fromPeer: p.peerId,
     data: data,
-    seqno: seqno,
+    seqno: @(p.seqno.toBytesBE), # unefficient, fine for now
     topicIDs: @[topic])
 
   if sign and p.publicKey.isSome:
--- a/libp2p/protocols/pubsub/rpc/message.nim
+++ b/libp2p/protocols/pubsub/rpc/message.nim
@@ -57,19 +57,22 @@ proc verify*(m: Message, p: PeerInfo): bool =
   else:
     libp2p_pubsub_sig_verify_failure.inc()
 
+# cannot explicit init but:
+# Thread locals are initialized to binary zero (usually default(T)). (Araq)
+var tSeqno {.threadvar.}: uint64 
+
 proc init*(
     T: type Message,
     p: PeerInfo,
     data: seq[byte],
     topic: string,
     sign: bool = true): Message {.gcsafe, raises: [CatchableError, Defect].} =
-  var seqno: seq[byte] = newSeq[byte](8)
   # peer is a ref obj
-  inc p.seqno
+  inc tSeqno
   result = Message(
     fromPeer: p.peerId,
     data: data,
-    seqno: @(p.seqno.toBytesBE), # unefficient, fine for now
+    seqno: @(tSeqno.toBytesBE), # unefficient, fine for now
     topicIDs: @[topic])
 
   if sign and p.publicKey.isSome:
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -133,7 +133,8 @@ method publish*(f: FloodSub,
     return
 
   trace "publishing on topic", name = topic
-  let msg = Message.init(f.peerInfo, data, topic, f.sign)
+  inc f.msgSeqno
+  let msg = Message.init(f.peerInfo, data, topic, f.msgSeqno, f.sign)
   # start the future but do not wait yet
   let (published, failed) = await f.sendHelper(f.floodsub.getOrDefault(topic), @[msg])
   for p in failed:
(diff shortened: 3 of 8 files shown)


oskarth:
Closes https://github.com/status-im/nim-libp2p/issues/254

libp2p/peerinfo.nim
```diff
@@ -39,6 +39,7 @@ type
       privateKey*: PrivateKey
     of HasPublic:
       key: Option[PublicKey]
+    seqno*: uint64
```
arnetheduck, line 42: storing this in the peerinfo looks odd - it means we'll keep a copy for every peer out there

dryajov: it should definitely not be here, `PeerInfo` is a generic object that doesn't know anything about `seqno`, in addition to what @arnetheduck it is also a leaking abstraction.

sinkingsugar: Should/Could be for each peer tho, 
I'm fine with adding a static global tho.. or thread local if fine for you (i'd prefer even)

sinkingsugar:
Btw I added a commit for gossip table fixes on this PR.. .will fix that later

libp2p/protocols/pubsub/rpc/message.nim
```diff
@@ -57,19 +57,22 @@ proc verify*(m: Message, p: PeerInfo): bool =
   else:
     libp2p_pubsub_sig_verify_failure.inc()
 
+# cannot explicit init but:
+# Thread locals are initialized to binary zero (usually default(T)). (Araq)
+var tSeqno {.threadvar.}: uint64
```
arnetheduck, line 62: this should be a field in `PubSub` - the more threadvars a library uses, the less flexible it is to use - it's also harder to test, has worse platform support, generates poor code and a number of other reasons

dryajov: Agree, this is a pretty arbitrary number either way, in fact it should have probably been left unspecified in the spec, because `linearly increasing` doesn't add anything at all, since the number only makes sense for the peer emitting the message all other peers shouldn't and cannot make any assumptions about it. So, just having a counter in `PubSub` should be enough.

sinkingsugar: I guess this week is full of critics for me...

I honestly don't understand your adversities towards TLS.
It's an amazing tool that works very well in many scenarios. And it's easy to debug and generates fine code too. It's supported everywhere ( I used even in nuttx os ). Gcc and clang support emulation for non conforming platforms too... if you know what you are doing there is absolutely no issue in using it.

That said if this is a nim related issue I can understand.

dryajov: Oh, didn't realize it came across as a criticism, it definitely isn't a knock on you and I have no issues with TLS, @arnetheduck might have his reservations 😄 . But in this case, a TLS would really make no sense, since we already have a "singleton" object, which is the `*sub` instance.

sinkingsugar: Fair enough but guys this is a `init` proc... am I supposed to stick the singleton in it? it's a bit an anti-pattern

dryajov:
I'm a bit concerned with the implementation however, reading the spec:

```
The seqno field is a 64-bit big-endian uint that is a linearly increasing number that is unique among messages originating
from each given peer. No two messages on a pubsub topic from the same peer should have the same seqno value, however
messages from different peers may have the same sequence number, so this number alone cannot be used to address
messages. Notably the 'timecache' in use by the go implementation contains a message_id, which is constructed from the
concatenation of the seqno and the from fields. This message_id is then unique among messages. It was also proposed in
#116 to use a message_hash, however, it was noted: "a potential caveat with using hashes instead of seqnos: the peer won't
be able to send identical messages (e.g. keepalives) within the timecache interval, as they will get rejected as duplicates."
```

Specifically this line:

`No two messages on a pubsub topic from the same peer should have the same seqno value, however
messages from different peers may have the same sequence number,`

It seems to imply that the `seqno` should be somehow persisted which would definitely complicate things because we would have to bolt in some sort of storage to either libp2p or pubsub itself. Furthermore, how is this enforceable at the protocol level? How about message ordering and/or messages dropped due to validation or network errors, there doesn't seem any sane way of implementing this that would actually comply with the spec.

libp2p/protocols/pubsub/rpc/message.nim
dryajov: I didn't mean a singleton in a literal sense, as in the pattern, I mean it as in "we only have one instance of pubsub per switch", which means that we can just stick a counter in it and forget about it (more or less, read my comment here https://github.com/status-im/nim-libp2p/pull/253#issuecomment-657922172)

sinkingsugar: well yeah can just pass the seqno by value to init  😄

sinkingsugar:
> It seems to imply that the `seqno` should be somehow persisted which would definitely complicate things because we would have to bolt in some sort of storage to either libp2p or pubsub itself. Furthermore, how is this enforceable at the protocol level? How about message ordering and/or messages dropped due to validation or network errors, there doesn't seem any sane way of implementing this that would actually comply with the spec.

You mean if we disconnect/reconnect (process restart) with the same peer basically right?

sinkingsugar:
Maybe using a timestamp based approach time on first lower 32 bits , progressive N on higher 32.
This should have been in spec tho :)

dryajov:
> > It seems to imply that the `seqno` should be somehow persisted which would definitely complicate things because we would have to bolt in some sort of storage to either libp2p or pubsub itself. Furthermore, how is this enforceable at the protocol level? How about message ordering and/or messages dropped due to validation or network errors, there doesn't seem any sane way of implementing this that would actually comply with the spec.
> 
> You mean if we disconnect/reconnect (process restart) with the same peer basically right?

Yes, that is the most likely scenario.

> Maybe using a timestamp based approach time on first lower 32 bits , progressive N on higher 32.

Might not be a bad idea, but you need to keep your clocks synced (sounds like a recurrent topic nowadays ;))

sinkingsugar:
> > > It seems to imply that the `seqno` should be somehow persisted which would definitely complicate things because we would have to bolt in some sort of storage to either libp2p or pubsub itself. Furthermore, how is this enforceable at the protocol level? How about message ordering and/or messages dropped due to validation or network errors, there doesn't seem any sane way of implementing this that would actually comply with the spec.
> > 
> > 
> > You mean if we disconnect/reconnect (process restart) with the same peer basically right?
> 
> Yes, that is the most likely scenario.
> 
> > Maybe using a timestamp based approach time on first lower 32 bits , progressive N on higher 32.
> 
> Might not be a bad idea, but you need to keep your clocks synced (sounds like a recurrent topic nowadays ;))

I think that any peer 2 peer up will definitely need to keep clocks in sync somehow and one way or another. so that should not be our concern imho.

dryajov:
> Indeed, but I think that any peer 2 peer app will definitely need to keep clocks in sync somehow and one way or another. so that should not be our concern imho.

Not necessarily and it's something that cannot be implied in the spec because the implications are pretty big. Also, clock syncing isn't something that most p2p applications can count on. NTP itself is centralized and there is no off the shelf public p2p clock service (not that I know of at least) and building one would be far from trivial, because you immediately fall into some sort of distributed consensus scenario more sophisticated mechanisms are required; things like vector clocks, CRDTs and/or blockchains :).

libp2p/protocols/pubsub/rpc/message.nim
dryajov: I would actually put it here somewhere - https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/pubsub.nim#L52-L62.

stefantalpalaru:
> NTP itself is centralized

Federated: https://www.ntppool.org/zone

sinkingsugar:
I will check what they others do :)

dryajov:
> I will check what they others do :)

go seems to just use a counter - https://github.com/libp2p/go-libp2p-pubsub/blob/bac68a1ba39b568c658d60a1d18a869e1d733740/pubsub.go#L1109-L1114

sinkingsugar:
https://github.com/libp2p/go-libp2p-pubsub/blob/8945f91465bc3ba418688684a2c29c384cd79647/pubsub.go#L1109
 Indeed :D

sinkingsugar:
but your point is super valid @dryajov

sinkingsugar:
I say, let's do the same for now.. tied to pubsub, just like go..

libp2p/protocols/pubsub/rpc/message.nim
arnetheduck: > adversities towards TLS

a global, and a thread-local one at that, generally limits the options available to the user of the library - the more "helpful" features you use, the less generally useful the library becomes, because it is no longer a self-contained box but rather has a side effect on the entire application - it becomes harder to run 2 switches, the code is no longer deterministic, you can no longer test it etc etc.

such things are best left for the user of the library to decide - where the switch memory should be allocated, how it should be managed, how it should be isolated as a subsystem etc - they're free to stick the switch into a threadlocal if they want that behaviour but they can also place it on a heap or run two side-by-side without them thrashing each other. we can of course keep adding obstacles and narrow the users choice to make more "convenient" to use but that rarely ends up satisfactory in the long run - when was the last time you created a local solution that is simple (like putting the counter in the pubsub instance which is a natural place for it to be anyway), then 6 months later went back and said, man I really should have used a global variable here - let's refactor it!

arnetheduck:
> It seems to imply that the seqno should be somehow persisted

...or the peer id to be regenerated for every time you connect.

it's a bit weird in the spec indeed - more than implying persistance, it simply looks like this was overlooked. perhaps 1.1 brings more sane behaviour here?

arnetheduck:
notably, since the message cache has a fairly short time window, messages in practise only need to be unique within that window..

to be really fancy one could combine the approaches - ie hash(message) _and_ sequence number.

sinkingsugar:
@dryajov waiting your approval, if it's ok for now at least (go style)

PR #255, opened 2020-07-08
noise updates

arnetheduck:
* clear secrets explicitly
* simplify keygen
* avoid some trivial memory allocations
* fix little endian encoding of nonce

### DIFF
--- a/libp2p/protocols/secure/noise.nim
+++ b/libp2p/protocols/secure/noise.nim
@@ -12,15 +12,13 @@ import chronicles
 import bearssl
 import stew/[endians2, byteutils]
 import nimcrypto/[utils, sha2, hmac]
-import ../../stream/lpstream
+import ../../stream/[connection, streamseq]
 import ../../peerid
 import ../../peerinfo
 import ../../protobuf/minprotobuf
 import ../../utility
-import ../../stream/lpstream
 import secure,
-       ../../crypto/[crypto, chacha20poly1305, curve25519, hkdf],
-       ../../stream/bufferstream
+       ../../crypto/[crypto, chacha20poly1305, curve25519, hkdf]
 
 logScope:
   topics = "noise"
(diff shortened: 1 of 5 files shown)


libp2p/protocols/secure/noise.nim
```diff
@@ -124,13 +121,12 @@ proc encryptWithAd(state: var CipherState, ad, data: openarray[byte]): seq[byte]
   result &= tag
   trace "encryptWithAd", tag = byteutils.toHex(tag), data = result.shortLog, nonce = state.n - 1
 
-proc decryptWithAd(state: var CipherState, ad, data: openarray[byte]): seq[byte] =
+proc decryptWithAd(state: var CipherState, ad, data: openArray[byte]): seq[byte] =
   var
-    tagIn = data[^ChaChaPolyTag.len..data.high].intoChaChaPolyTag
-    tagOut = tagIn
+    tagIn = data.toOpenArray(data.len - ChaChaPolyTag.len, data.high).intoChaChaPolyTag
+    tagOut: ChaChaPolyTag
```
dryajov, line 127: where is `tagOut` being set now?

arnetheduck: `tagOut` comes from the encrypted data (it's appended as part of encryption and verified as part of decryption)

sinkingsugar: `ChaChaPoly.decrypt` is a proc(..., var _: T, ...)

sinkingsugar:
Looks good to me @arnetheduck

libp2p/protocols/secure/noise.nim
```diff
@@ -491,7 +510,7 @@ proc newNoise*(
     rng: rng,
     outgoing: outgoing,
     localPrivateKey: privateKey,
-    localPublicKey: privateKey.getKey().tryGet(),
+    localPublicKey: privateKey.getKey().tryGet().getBytes().tryGet(),
```
cheatfate, line 513: This sandwich looks ugly.

sinkingsugar: 🥪

libp2p/protocols/secure/noise.nim
dryajov: Ugh, ok, but looks fishy, it either was wrong or it's wrong now. We used to set `tagOut = tagIn`, now we aren't...

arnetheduck: the underlying chacha it calls is `out-only` for tag

arnetheduck: I guess the chacha api could be made better but whatever

PR #257, opened 2020-07-08
More gossip cleanup

dryajov:


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -125,23 +125,21 @@ template moveToMeshHelper(g: GossipSub,
                            peers = peerIds.len
 
     table[topic].excl(id) # always exclude
-    if id in g.mesh[topic]:
+    if id in g.mesh.getOrDefault(topic):
```
arnetheduck, line 128: hashset requires that init be called, so `getOrDefault` is not correct

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -249,69 +244,75 @@ proc getGossipPeers(g: GossipSub): Table[string, ControlMessage] {.gcsafe.} =
       result[id].ihave.add(ihave)
 
 proc heartbeat(g: GossipSub) {.async.} =
-  while g.heartbeatRunning:
-    try:
-      trace "running heartbeat"
+  # don't lock on every iteration
+  withLock g.heartbeatLock:
```
arnetheduck, line 248: should the sleep be locked??

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -249,69 +244,75 @@ proc getGossipPeers(g: GossipSub): Table[string, ControlMessage] {.gcsafe.} =
...
     if t in g.fanout:
-      g.fanout[t].excl(peer.id)
+      g.fanout[t].excl(id)
 
     libp2p_gossipsub_peers_per_topic_fanout
       .set(g.fanout[t].len.int64, labelValues = [t])
 
+  procCall FloodSub(g).handleDisconnect(peer)
+
 method subscribePeer*(p: GossipSub,
```
arnetheduck, line 307:
```
method subscribePeer*(
    p: GossipSub,  # 4 spaces here!
    conn: Connection) =...
```
is an indent style from python that works well - the 4 space indent sets the parameters apart from the body, but doesn't require adjustments when refactoring stuff like the name :)

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -537,31 +540,24 @@ method start*(g: GossipSub) {.async.} =
...
 
-  # interlock start to to avoid overlapping to stops
-  await g.heartbeatLock.acquire()
-
-  # setup the heartbeat interval
-  g.heartbeatRunning = true
-  g.heartbeatFut = g.heartbeat()
-
-  g.heartbeatLock.release()
+  withLock g.heartbeatLock:
```
arnetheduck, line 543: so the heartbeat lock is taken both inside and outside `g.heartbeat`? how does that work?

libp2p/protocols/pubsub/gossipsub.nim
dryajov: Not anymore apparently - https://github.com/status-im/Nim/blob/nimbus/lib/pure/collections/sets.nim#L127-L128.

libp2p/protocols/pubsub/gossipsub.nim
dryajov: You don't want two heartbeats running in concurrently, this is mainly to prevent that.

libp2p/protocols/pubsub/gossipsub.nim
dryajov: We probably don't want this one here, some mixup with @sinkingsugar refactors that I picked along the way.

libp2p/protocols/pubsub/gossipsub.nim
dryajov: We should just use a formatter to enforce common style like this across the board.

libp2p/protocols/pubsub/gossipsub.nim
sinkingsugar: But it's released every iteration before sleep in heartbeat no?

sinkingsugar: But yeah could be mixup

dryajov: Ah, it is actually correct after all, both this one and the one in `stop`.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -537,31 +540,24 @@ method start*(g: GossipSub) {.async.} =
...
 method stop*(g: GossipSub) {.async.} =
   trace "gossipsub stop"
 
   ## stop pubsub
   ## stop long running tasks
 
-  await g.heartbeatLock.acquire()
-
-  # stop heartbeat interval
   g.heartbeatRunning = false
```
dryajov, line 554: This will stop the `g.heartbeat` loop and then lock and `await g.heartbeatFut`, so the `start`, `heartbeat`, `stop` sequences look correct. I'm not 100% sure if we need them at all tho.

libp2p/protocols/pubsub/gossipsub.nim
arnetheduck: huh, there's another comment saying they are - oh well, maybe it works then.

sinkingsugar: needs INIT, docs are wrong

libp2p/protocols/pubsub/gossipsub.nim
sinkingsugar: Sleep should NOT be locked, I think it's not tho

libp2p/protocols/pubsub/gossipsub.nim
sinkingsugar: it is fine cos heartbeat won't run again since you locked and will just wait the sleep of the heartbeat, we might need to cancel the sleep tho cos now it's pretty long... 😄

libp2p/protocols/pubsub/gossipsub.nim
arnetheduck: there exists no formatter, hence it makes sense to optimize the format so that it doesn't require readjustment and aligning

libp2p/protocols/pubsub/gossipsub.nim
dryajov: I'm not sure if the locks on the `heartbeat` are adding anything, unless someone tries to accidentally call `heartbeat` twice, it should not be a problem. So I don't see any inherent concurrency issue with it anymore.

PR #258, opened 2020-07-09
Gossip tables fix

sinkingsugar:
Add peers to gossipsub table even for outgoing connections

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -474,6 +474,13 @@ method subscribe*(g: GossipSub,
                   topic: string,
                   handler: TopicHandler) {.async.} =
   await procCall PubSub(g).subscribe(topic, handler)
+
+  # add our current peers to availability table
+  for peerid, _ in g.peers:
+    if topic notin g.gossipsub:
+      g.gossipsub[topic] = initHashSet[string]()
+    g.gossipsub[topic].incl(peerid)
+  
   await g.rebalanceMesh(topic)
 
 method unsubscribe*(g: GossipSub,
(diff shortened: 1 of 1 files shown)


arnetheduck:
this turns out not being quite right - if the gossipsub table is _their_ subscriptions, we should not be adding them to gossipsub when _we_ subscribe to a topic

dryajov:
Agree, this doesn't seem to do what we want.

sinkingsugar:
How do we fill the mesh then?

sinkingsugar:
What I mean is that currently we publish to 0 peers tho.. cos mesh is empty

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -168,8 +168,10 @@ proc rebalanceMesh(g: GossipSub, topic: string) {.async.} =
         # send a graft message to the peer
         await p.sendPrune(@[topic])
         g.mesh[topic].excl(id)
-        if topic in g.gossipsub:
-          g.gossipsub[topic].incl(id)
+
+        if topic notin g.gossipsub:
```
dryajov, line 172: this shouldn't really happen, but it doesn't hurt I guess.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -474,6 +480,13 @@ method subscribe*(g: GossipSub,
                   topic: string,
                   handler: TopicHandler) {.async.} =
   await procCall PubSub(g).subscribe(topic, handler)
+
+  # add our current peers to availability table
+  for peerid, _ in g.peers:
+    if topic notin g.gossipsub:
+      g.gossipsub[topic] = initHashSet[string]()
+    g.gossipsub[topic].incl(peerid)
```
dryajov, line 488: This isn't the right place to do it, the `gossipsub` table is the table that holds a map of topics to peers, with the peers being those that previously advertised that they subscribe to the topic, what this would do is wrongly map peers to topics that they aren't subscribed to.

sinkingsugar:
Closing this as pointless since wrong

PR #259, opened 2020-07-09
[WIP] Minprotobuf refactoring

cheatfate:


libp2p/crypto/crypto.nim
```diff
@@ -222,8 +222,8 @@ proc toBytes*(key: PrivateKey, data: var openarray[byte]): CryptoResult[int] =
   ##
   ## Returns number of bytes (octets) needed to store private key ``key``.
   var msg = initProtoBuffer()
-  msg.write(initProtoField(1, cast[uint64](key.scheme)))
-  msg.write(initProtoField(2, ? key.getRawBytes()))
+  msg.write(1, cast[uint64](key.scheme))
```
arnetheduck, line 225: nice with the write. does this need cast really or would `uint64(key.scheme)` be enough?

libp2p/protobuf/minprotobuf.nim
```diff
@@ -47,11 +51,18 @@ type
...
+
+template getProtoHeader*(index: int64, wire: ProtoFieldKind): uint64 =
   ## Get protobuf's field header integer for ``index`` and ``wire``.
-  ((uint(index) shl 3) or cast[uint](wire))
+  ((uint64(index) shl 3) or cast[uint64](wire))
 
-template protoHeader*(field: ProtoField): uint =
+template getProtoHeader*(field: ProtoField): uint =
   ## Get protobuf's field header integer for ``field``.
   ((uint(field.index) shl 3) or cast[uint](field.kind))
```
arnetheduck, line 67: strictly, `field.index` should be `uint32` and range-checked when reading - https://developers.google.com/protocol-buffers/docs/proto3#assigning_field_numbers - ie 3 bits of header and 29 bits of filed numbers

libp2p/protobuf/minprotobuf.nim
```diff
@@ -127,6 +140,13 @@ proc initProtoBuffer*(data: seq[byte], offset = 0,
   result.offset = offset
   result.options = options
 
+proc initProtoBuffer*(data: openarray[byte], offset = 0,
+                      options: set[ProtoFlags] = {}): ProtoBuffer =
+  ## Initialize ProtoBuffer with shallow copy of ``data``.
```
arnetheduck, line 145: not a shallow copy (which is a good thing)

libp2p/protobuf/minprotobuf.nim
```diff
@@ -199,31 +319,412 @@ proc finish*(pb: var ProtoBuffer) =
...
+      else:
+        if not(skipValue(pb, header)):
+          output.setLen(0)
+          return false
+    else:
+      if not(skipValue(pb, header)):
+        output.setLen(0)
+        return false
+
+  if len(output) > 0:
```
arnetheduck, line 660: why is it a failure to not find a value for a repeated field? should `false` be only for when the input buffer is invalid?

libp2p/protobuf/minprotobuf.nim
```diff
@@ -199,31 +319,412 @@ proc finish*(pb: var ProtoBuffer) =
...
+    else:
+      if not(skipValue(pb, header)):
+        output.setLen(0)
+        return false
+
+  res
+
+proc getRepeatedField*[T: seq[byte]|string](data: ProtoBuffer, field: int64,
+                                            output: var seq[T]): bool =
+  var pb = data
```
arnetheduck, line 595: why not manipulate `offset` here instead of the data copy?

libp2p/protobuf/minprotobuf.nim
cheatfate: Current approach not allows to check if the message is actually valid or field is present, so right now if procedure returns `false` it means that its impossible to get value with specified `field` from the message. (It can be missing or message could be invalid).

libp2p/protobuf/minprotobuf.nim
cheatfate: Because we actually need to manipulate with `offset` but we still want to have `data` to be constant for other calls to `getField/RepeatedField/PackedRepeatedField`.

PR #260, opened 2020-07-09
clean up mesh handling logic

arnetheduck:
* gossipsub is a function of subscription messages only
* graft/prune work with mesh, get filled up from gossipsub
* fix race conditions with await
* fix exception unsafety when grafting/pruning
* fix allowing up to DHi peers in mesh on incoming graft
* fix metrics in several places

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -45,9 +45,9 @@ const GossipSubFanoutTTL* = 1.minutes
 
 type
   GossipSub* = ref object of FloodSub
-    mesh*: Table[string, HashSet[string]]      # meshes - topic to peer
-    fanout*: Table[string, HashSet[string]]    # fanout - topic to peer
-    gossipsub*: Table[string, HashSet[string]] # topic to peer map of all gossipsub peers
+    mesh*: Table[string, HashSet[string]]      # peers that we send messages to when we are subscribed to the topic
+    fanout*: Table[string, HashSet[string]]    # peers that we send messages to when we're not subscribed to the topic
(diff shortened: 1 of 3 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -68,6 +68,31 @@ declareGauge(libp2p_gossipsub_peers_per_topic_gossipsub,
...
+    table: var Table[string, HashSet[string]], topic, peerId: string) =
+  table.withValue(topic, peers):
+    peers[].excl(peerId)
+    if peers[].len == 0:
+      table.del(topic)
+
+func hasPeer(table: Table[string, HashSet[string]], topic, peerId: string): bool =
+  (topic in table) and (peerId in table[topic])
+
+func peers(table: Table[string, HashSet[string]], topic: string): int =
```
sinkingsugar, line 87: `var` table or copy will happen likely

arnetheduck: won't - see also `Table.get` itself

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -343,34 +347,39 @@ proc handleGraft(g: GossipSub,
...
-    trace "processing graft message", peer = peer.id,
-                                      topicID = graft.topicID
-
-    if graft.topicID in g.topics:
-      if g.mesh.len < GossipSubD:
-        g.mesh[graft.topicID].incl(peer.id)
+    let topic = graft.topicID
+    trace "processing graft message", topic, peerId
+
+      # If they send us a graft before they send us a subscribe, what should
```
dryajov, line 355: I would say that there is no harm in adding them to gossipsub, if anything it will keep the tables in sync.

PR #263, opened 2020-07-11
PubSubPeer tables refactor

sinkingsugar:
Remove the excessive use of `string` and `id` in favor of straight using `PubSubPeer` `ref` type.
Removes the need to resolve peers from `g.peers`

### DIFF
--- a/libp2p/protocols/pubsub/pubsub.nim
+++ b/libp2p/protocols/pubsub/pubsub.nim
@@ -30,6 +30,8 @@ declareCounter(libp2p_pubsub_validation_failure, "pubsub failed validated messag
 declarePublicCounter(libp2p_pubsub_messages_published, "published messages", labels = ["topic"])
 
 type
+  PeerTable* = Table[string, HashSet[PubSubPeer]]
+
   SendRes = tuple[published: seq[string], failed: seq[string]] # keep private
 
   TopicHandler* = proc(topic: string,
@@ -58,6 +60,16 @@ type
     validators*: Table[string, HashSet[ValidatorHandler]]
     observers: ref seq[PubSubObserver] # ref as in smart_ptr
     msgIdProvider*: MsgIdProvider      # Turn message into message id (not nil)
+
+proc hasPeerID*(t: PeerTable, topic, peerId: string): bool =
+  # unefficient but used only in tests!
+  let peers = t.getOrDefault(topic)
+  if peers.len == 0:
+    false
+  else:
+    let ps = toSeq(peers)
+    ps.any do (peer: PubSubPeer) -> bool:
+      peer.id == peerId
 
 method handleDisconnect*(p: PubSub, peer: PubSubPeer) {.base.} =
   ## handle peer disconnects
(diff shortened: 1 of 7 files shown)


libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -59,6 +61,16 @@ type
     observers: ref seq[PubSubObserver] # ref as in smart_ptr
     msgIdProvider*: MsgIdProvider      # Turn message into message id (not nil)
 
+proc hasPeerID*(t: PeerTable, topic, peerId: string): bool =
```
dryajov, line 64: If only used in tests, we should just move it to our test utils; UFCFS is a wonderful thing :)

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -43,6 +43,10 @@ type
 
   RPCHandler* = proc(peer: PubSubPeer, msg: seq[RPCMsg]): Future[void] {.gcsafe.}
 
+proc hash*(p: PubSubPeer): Hash = 
+  # int is either 32/64, so intptr basically, pubsubpeer is a ref
+  cast[int](p).hash
```
dryajov, line 48: This is generally unsafe IMO, I know that the chances of getting the same type of object on the exact same location as a previous one is very low, but it could happen, moreover, if we use an object pool/mem pool those chances become very real.

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -30,6 +30,8 @@ declareCounter(libp2p_pubsub_validation_failure, "pubsub failed validated messag
 declarePublicCounter(libp2p_pubsub_messages_published, "published messages", labels = ["topic"])
 
 type
+  PeerTable* = Table[string, HashSet[PubSubPeer]]
```
dryajov, line 33: I would actually use the `PeerID` object here instead of `PubSubPeer`, peerid is guaranteed to be cryptographically unique.

sinkingsugar:
Well PeerID is what we currently have.

To strengthen this PR I'd leave it like this in terms of hash but add equality check based on PeerID, so that when a collision happens the ID string is compared.

The pointer int64 is faster then string hash wise.

libp2p/protocols/pubsub/pubsub.nim
arnetheduck: well, it's a bit the point of ref types to simplify that model and use references so that the double-lookup is avoided - I think ref makes sense in this situation, unless we make all API string-based (ie `send(id: peerid)` which is a bit nutty in nim

libp2p/protocols/pubsub/pubsubpeer.nim
arnetheduck: in nim this is fine, but I would indeed not do it anyway - ie might as well take the first bytes of the peerid which are equally unique

sinkingsugar:
@arnetheduck ok for you too? (I added proper equality check in case of collision)

libp2p/protocols/pubsub/pubsubpeer.nim
sinkingsugar: changed it to `[pointer`] (no diff but who knows one day) and added a proper equality check in case of collision based on peerid

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -43,9 +43,34 @@ type
...
+    if a.peerInfo == nil:
+      if b.peerInfo == nil:
+        true
+      else:
+        false
+    else:
+      if b.peerInfo == nil:
+        false
+      else:
+        a.peerInfo == b.peerInfo
```
arnetheduck, line 73: this should go by id

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -43,9 +43,34 @@ type
 
   RPCHandler* = proc(peer: PubSubPeer, msg: seq[RPCMsg]): Future[void] {.gcsafe.}
 
-proc hash*(p: PubSubPeer): Hash = 
+func hash*(p: PubSubPeer): Hash = 
   # int is either 32/64, so intptr basically, pubsubpeer is a ref
-  cast[int](p).hash
+  cast[pointer](p).hash
```
arnetheduck, line 48: `hash(addr p[])` might be better here

sinkingsugar: fine but, wouldn't this defect if nil?

libp2p/protocols/pubsub/pubsubpeer.nim
sinkingsugar: good point, indeed!

libp2p/protocols/pubsub/pubsubpeer.nim
arnetheduck: not sure tbh - ok let's leave cast for now

PR #265, opened 2020-07-13
Fix connection not tracked assertion

sinkingsugar:
We often in NBC trigger
```
doAssert(conn.peerInfo.id in s.connections,
    "connection not tracked!")
```
It seems to me that it happens if muxing failed cos no available muxers and we just return not considering it an error

### DIFF
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -206,19 +206,21 @@ proc identify(s: Switch, conn: Connection) {.async, gcsafe.} =
 
     trace "identify: identified remote peer", peer = $conn.peerInfo
 
-proc mux(s: Switch, conn: Connection): Future[bool] {.async, gcsafe.} =
+proc mux(s: Switch, conn: Connection) {.async, gcsafe.} =
   ## mux incoming connection
 
   trace "muxing connection", peer = $conn
   let muxers = toSeq(s.muxers.keys)
   if muxers.len == 0:
     warn "no muxers registered, skipping upgrade flow"
-    return false
+    raise newException(CatchableError,
+      "no muxers registered, skipping upgrade flow")
 
   let muxerName = await s.ms.select(conn, muxers)
   if muxerName.len == 0 or muxerName == "na":
     debug "no muxer available, early exit", peer = $conn
-    return false
+    raise newException(CatchableError,
+      "no muxer available, early exit")
 
   # create new muxer for connection
   let muxer = s.muxers[muxerName].newMuxer(conn)
@@ -252,7 +254,6 @@ proc mux(s: Switch, conn: Connection): Future[bool] {.async, gcsafe.} =
   # store it in muxed connections if we have a peer for it
   trace "adding muxer for peer", peer = conn.peerInfo.id
   await s.storeConn(muxer, Direction.Out, handlerFut)
-  return true
 
 proc cleanupConn(s: Switch, conn: Connection) {.async, gcsafe.} =
     if isNil(conn):
@@ -333,11 +334,12 @@ proc upgradeOutgoing(s: Switch, conn: Connection): Future[Connection] {.async, g
       "unable to secure connection, stopping upgrade")
 
   trace "upgrading connection"
-  let success = await s.mux(sconn) # mux it if possible
-  if isNil(sconn.peerInfo) or not success:
+  
+  try:
+    await s.mux(sconn)
+  except:
     await sconn.close()
-    raise newException(CatchableError,
-      "unable to mux connection, stopping upgrade")
+    raise
 
   trace "succesfully upgraded outgoing connection", uoid = sconn.oid
   return sconn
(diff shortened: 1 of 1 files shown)


libp2p/switch.nim
```diff
@@ -206,19 +206,19 @@ proc identify(s: Switch, conn: Connection) {.async, gcsafe.} =
...
 
   trace "muxing connection", peer = $conn
   let muxers = toSeq(s.muxers.keys)
   if muxers.len == 0:
     warn "no muxers registered, skipping upgrade flow"
-    return
+    return false
 
   let muxerName = await s.ms.select(conn, muxers)
   if muxerName.len == 0 or muxerName == "na":
```
arnetheduck, line 219: this should also check that muxerName is in s.muxers - ie the relationship is too weak to rely on, that only "supported" muxes will be returned

libp2p/switch.nim
```diff
@@ -206,19 +206,19 @@ proc identify(s: Switch, conn: Connection) {.async, gcsafe.} =
 
     trace "identify: identified remote peer", peer = $conn.peerInfo
 
-proc mux(s: Switch, conn: Connection) {.async, gcsafe.} =
+proc mux(s: Switch, conn: Connection): Future[bool] {.async, gcsafe.} =
```
arnetheduck, line 209: does anything in the mux flow raise? if so, might as well use raises throughout - there are few things harder to understand than exceptions, but exceptions and error returns is one of them

sinkingsugar: Actually you are right, there is a raise, I will stick to raises. 
The only issue is that the caller might want to close the connection and that needs signaling... handle and re-rise seems overkill.

sinkingsugar: On the other hand the current raise will not clear the connection indeed.. for example......

sinkingsugar:
But anyway I wanna wait @dryajov cos "mux it if possible" in the comment makes me wonder if mux is necessary or not..
If it's not then there is a bigger issue cos we never call `storeConn` and that's why the assertion happens

libp2p/switch.nim
sinkingsugar: Not sure I understand this one sorry,
we take all muxer names from s.muxers above and pass it to select, that routine will do that check already afair and return ne or empty string (I fixed that recently).

arnetheduck: it's just a weird way of calling the select - with the names and then looking up in collection, and also checking for empty or "na" - why doesn't ms.select return the muxer instance and that's it or something like that?

sinkingsugar: I think that is outside of the scope of this PR, I absolutely agree tho, but requires a much big refactor.

dryajov: `select` should not return the instance of anything. Multistream-select knows nothing about mixers, secure channels, etc... All it does is signal the other end that a request for a particular string needs to be handled, it is up to the upper layers to decide what to do with the stream once negotiated.

dryajov: > this should also check that muxerName is in s.muxers - ie the relationship is too weak to rely on, that only "supported" muxes will be returned

No this is actually what the protocol expects, it will only return the strings for those ones that passed the negotiation. The relation is fine in the context of a string based protocol negotiation mechanism.

arnetheduck: well, kind of not really - ie you want the lookup and validation to be close to the resolution of the result - ie you don't want two different modules to be comparing the same string. effectively, `s.muxers[...]` performs a second lookup for the same value, and the two may go out of sync in between as code gets refactored - if instead for example a map was passed in `let muxer = s.ms.select(conn, s.muxers)`, this could be done in  a single pass with lower maintenance risk

dryajov:
Mm, interesting. I wonder how this happens at all, this was one of the intents of the `assert` either way, to catch this potentially weird scenario.

> But anyway I wanna wait @dryajov cos "mux it if possible" in the comment makes me wonder if mux is necessary or not..

Right, so muxing isn't essential and might be skipped depending on the transport. In fact all of this upgrade flow should eventually move out of the switch into the transports them self. The reason is that the transports determine whether they require a muxer or not, for example spidy/http3 already have a muxer, so transports relying on this protocols would not required additional muxing. Of course, right now this isn't the case, since we only support straight up TCP which does require an external muxer. In other words, for our use case the muxer isn't optional ATM, but future versions of `nim-libp2p` should support alternate upgrade flows.

Now, why does this happen I'm not sure - a connection will be stored regardless of whether muxing passed/failed iff there is a peer id associated with the connection. I suspect that this is some sort of raise condition that happens when a connection gets destroyed but the upgrade flow hasn't yet finished.  Maybe we need to lock both connect/disconnect flows?

sinkingsugar:
@dryajov 
> a connection will be stored regardless of whether muxing passed/failed iff there is a peer id associated with the connection

That is not the case tho

sinkingsugar:
You reworked all this part I think @dryajov , so probably this can be closed.

PR #269, opened 2020-07-14
Minprotobuf refactoring 2

cheatfate:
Transition from the result of `bool` to` Result [bool] `, so that now invalid messages and missing values can be distinguished.


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -114,7 +114,13 @@ proc handle*(p: PubSubPeer, conn: Connection) {.async.} =
           trace "message already received, skipping", peer = p.id
           continue
 
-        var msg = decodeRpcMsg(data)
+        var rmsg = decodeRpcMsg(data)
```
cheatfate, line 117: Please note this changes, before there was assumption that all messages sent through gossip are correct, so maybe `break` here should be replaced with something more meaningful.

PR #272, opened 2020-07-15
Public key infrastructure filters.

cheatfate:
This will allow to filter what public key infrastructure will be used by `nim-libp2p`, you can choose particular schemes via compilation setting `libp2p_pki_schemes`.

For example `nim-beacon-chain` could use `--d:libp2p_pki_schemes=secp256k1`, to use only `secp256k1` and disable support for `ed25519`, `ecnist` and `rsa` keys.

You can also specify any combination of schemes delimiting it with `;` or `,` or `|`.
For example `--d:libp2p_pki_schemes=secp256k1,ed25519` or `--d:libp2p_pki_schemes=secp256k1,ed25519,ecnist`.

By default all schemes are available.

This PR is also removes `Blowfish` mentions.


libp2p/crypto/crypto.nim
```diff
@@ -8,29 +8,79 @@
...
+
+  let schemes = split(list, {',', ';', '|'})
+  for item in schemes:
+    if cmpIgnoreCase(strip(item), "rsa") == 0:
+      res.incl(PKScheme.RSA)
+    elif cmpIgnoreCase(strip(item), "ed25519") == 0:
+      res.incl(PKScheme.Ed25519)
+    elif cmpIgnoreCase(strip(item), "secp256k1") == 0:
+      res.incl(PKScheme.Secp256k1)
+    elif cmpIgnoreCase(strip(item), "ecnist") == 0:
```
arnetheduck, line 34: I find ecnist vs ecdsa confusing

cheatfate: @arnetheduck ECDSA its Elliptic Curve Digital Signature Algorithm and this algorithm can work with almost any curve, so for example signature and verification with bitcoin's secp256`k`1 is also ECDSA. But initially this algorithm was used with NIST curves which are secp256`r`1, secp384`r`1 and secp521`r`1. The name i have used initially `ECDSA` was incorrect but it was name which was used by `go` code.

dryajov:
@cheatfate is this ready?

cheatfate:
Yes this PR is ready @dryajov.

This PR is also fixes  #275

PR #276, opened 2020-07-16
gossipsub fixes

arnetheduck:
* graft up to D peers
* fix logging so it's clear who is grafting/pruning who
* clear fanout when grafting

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -99,49 +99,48 @@ proc replenishFanout(g: GossipSub, topic: string) =
   trace "fanout replenished with peers", peers = g.fanout.peers(topic)
 
 proc rebalanceMesh(g: GossipSub, topic: string) {.async.} =
+  logScope:
+    topic
+
   trace "about to rebalance mesh"
+
   # create a mesh topic that we're subscribing to
 
   var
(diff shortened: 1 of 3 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -99,49 +99,48 @@ proc replenishFanout(g: GossipSub, topic: string) =
...
+    grafts = toSeq(
       g.gossipsub.getOrDefault(topic, initHashSet[PubSubPeer]()) -
       g.mesh.getOrDefault(topic, initHashSet[PubSubPeer]())
     )
 
     logScope:
-      topic = topic
       meshPeers = g.mesh.peers(topic)
-      newPeers = newPeers.len
+      grafts = grafts.len
```
dryajov, line 122: what is this doing?

arnetheduck: printing number of graft candidates I guess

dryajov: I think this is resetting the seq's size, but it's setting it to itself, so it's a noop, look fishy. I know it was there before you got to it, but seems wrong either way.

dryajov: oh, it's a logscope, nevermind.

PR #277, opened 2020-07-16
Connection manager

dryajov:
Move connection management out of switch into it's own module

### DIFF
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -25,7 +25,11 @@ type
     handle: Future[void]
 
   ConnManager* = ref object of RootObj
-    conns: Table[PeerInfo, HashSet[Connection]]
+    # NOTE: don't change to PeerInfo here
+    # the reference semantics on the PeerInfo
+    # object itself make it succeptible to
+    # copies and mangling by unrelated code.
+    conns: Table[PeerID, HashSet[Connection]]
     muxed: Table[Connection, MuxerHolder]
     cleanUpLock: Table[PeerInfo, AsyncLock]
     maxConns: int
@@ -36,7 +40,7 @@ proc newTooManyConnections(): ref TooManyConnections {.inline.} =
 proc init*(C: type ConnManager,
            maxConnsPerPeer: int = MaxConnectionsPerPeer): ConnManager =
   C(maxConns: maxConnsPerPeer,
-    conns: initTable[PeerInfo, HashSet[Connection]](),
+    conns: initTable[PeerID, HashSet[Connection]](),
     muxed: initTable[Connection, MuxerHolder]())
 
 proc contains*(c: ConnManager, conn: Connection): bool =
@@ -50,10 +54,13 @@ proc contains*(c: ConnManager, conn: Connection): bool =
   if isNil(conn.peerInfo):
     return
 
-  if conn.peerInfo notin c.conns:
+  if conn.peerInfo.peerId notin c.conns:
     return
 
-  return conn in c.conns[conn.peerInfo]
+  return conn in c.conns[conn.peerInfo.peerId]
+
+proc contains*(c: ConnManager, peerId: PeerID): bool =
+  peerId in c.conns
 
 proc contains*(c: ConnManager, muxer: Muxer): bool =
   ## checks if a muxer is being tracked by the connection
@@ -71,9 +78,6 @@ proc contains*(c: ConnManager, muxer: Muxer): bool =
     return
 
   return muxer == c.muxed[conn].muxer
-
-proc peers*(c: ConnManager): seq[PeerInfo] =
-  toSeq(c.conns.keys)
 
 proc cleanupConn(c: ConnManager, conn: Connection) {.async.} =
   ## clean connection's resources such as muxers and streams
@@ -99,18 +103,18 @@ proc cleanupConn(c: ConnManager, conn: Connection) {.async.} =
       if not(isNil(muxerHolder.handle)):
         await muxerHolder.handle
 
-    if peerInfo in c.conns:
-      c.conns[peerInfo].excl(conn)
+    if peerInfo.peerId in c.conns:
+      c.conns[peerInfo.peerId].excl(conn)
 
-      if c.conns[peerInfo].len == 0:
-        c.conns.del(peerInfo)
+      if c.conns[peerInfo.peerId].len == 0:
+        c.conns.del(peerInfo.peerId)
 
     if not(conn.peerInfo.isClosed()):
       conn.peerInfo.close()
 
   finally:
     await conn.close()
-    libp2p_peers.set(c.peers.len.int64)
+    libp2p_peers.set(c.conns.len.int64)
 
     if lock.locked():
       lock.release()
@@ -137,7 +141,7 @@ proc selectConn*(c: ConnManager,
     return
 
   let conns = toSeq(
-    c.conns.getOrDefault(peerInfo))
+    c.conns.getOrDefault(peerInfo.peerId))
     .filterIt( it.dir == dir )
 
   if conns.len > 0:
@@ -178,21 +182,29 @@ proc storeConn*(c: ConnManager, conn: Connection) =
     raise newException(CatchableError, "empty peer info")
 
   let peerInfo = conn.peerInfo
-  if c.conns.getOrDefault(peerInfo).len > c.maxConns:
+  if c.conns.getOrDefault(peerInfo.peerId).len > c.maxConns:
     trace "too many connections", peer = $conn.peerInfo,
                                   conns = c.conns
-                                  .getOrDefault(peerInfo).len
+                                  .getOrDefault(peerInfo.peerId).len
 
     raise newTooManyConnections()
 
-  if peerInfo notin c.conns:
-    c.conns[peerInfo] = initHashSet[Connection]()
+  if peerInfo.peerId notin c.conns:
+    c.conns[peerInfo.peerId] = initHashSet[Connection]()
 
-  c.conns[peerInfo].incl(conn)
+  c.conns[peerInfo.peerId].incl(conn)
 
   # launch on close listener
   asyncCheck c.onClose(conn)
-  libp2p_peers.set(c.peers.len.int64)
+  libp2p_peers.set(c.conns.len.int64)
+
+proc storeOutgoing*(c: ConnManager, conn: Connection) =
+  conn.dir = Direction.Out
+  c.storeConn(conn)
+
+proc storeIncoming*(c: ConnManager, conn: Connection) =
+  conn.dir = Direction.In
+  c.storeConn(conn)
 
 proc storeMuxer*(c: ConnManager,
                  muxer: Muxer,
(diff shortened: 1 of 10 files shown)


libp2p/connmanager.nim
```diff
@@ -0,0 +1,276 @@
...
+  if isNil(conn):
+    return
+
+  if isNil(conn.peerInfo):
+    return
+
+  if conn.peerInfo.peerId notin c.conns:
+    return
+
+  return conn in c.conns[conn.peerInfo.peerId]
```
sinkingsugar, line 60: I would really avoid mixing weird nim return rules.
I would do like I refactored here:
https://github.com/status-im/nim-libp2p/blob/gossip-one-one/libp2p/peerinfo.nim#L142-L156

This is not really just about syntax but also helps the compiler predict branches, and most of all no branch left behind.

libp2p/connmanager.nim
```diff
@@ -0,0 +1,276 @@
...
+        c.conns.del(peerInfo.peerId)
+
+    if not(conn.peerInfo.isClosed()):
+      conn.peerInfo.close()
+
+  finally:
+    await conn.close()
+    libp2p_peers.set(c.conns.len.int64)
+
+    if lock.locked():
```
sinkingsugar, line 119: if `close()` we never get unlocked here, either lock goes before or completely outside `try/catch` I think for safety

sinkingsugar: also I would use `withLock` helper, if we didn't remove it, as its interchangeable with nim std and anyway keeps track of lock unlock

libp2p/connmanager.nim
```diff
@@ -0,0 +1,276 @@
...
+
+    trace "connection cleaned up"
+
+proc onClose(c: ConnManager, conn: Connection) {.async.} =
+  ## connection close even handler
+  ##
+  ## triggers the connections resource cleanup
+  ##
+
+  await conn.closeEvent.wait()
```
sinkingsugar, line 130: We use those events often, but did we ever "audit" their ability to be reliable 😄 I wonder

libp2p/connmanager.nim
```diff
@@ -0,0 +1,276 @@
...
+  trace "triggering connection cleanup"
+  await c.cleanupConn(conn)
+
+proc selectConn*(c: ConnManager,
```
sinkingsugar, line 134: again I would refrain from mixing nim return magic/weirdness

libp2p/connmanager.nim
```diff
@@ -0,0 +1,276 @@
...
+    return
+
+  let conns = toSeq(
+    c.conns.getOrDefault(peerInfo.peerId))
+    .filterIt( it.dir == dir )
+
+  if conns.len > 0:
+    return conns[0]
+
+proc selectConn*(c: ConnManager, peerInfo: PeerInfo): Connection =
```
sinkingsugar, line 150: again I would refrain from mixing nim return magic/weirdness

libp2p/connmanager.nim
```diff
@@ -0,0 +1,276 @@
...
+  if isNil(peerInfo):
+    return
+
+  var conn = c.selectConn(peerInfo, Direction.Out)
+  if isNil(conn):
+    conn = c.selectConn(peerInfo, Direction.In)
+
+  return conn
+
+proc selectMuxer*(c: ConnManager, conn: Connection): Muxer =
```
sinkingsugar, line 164: again I would refrain from mixing nim return magic/weirdness

Like here I don't even know what happens, we return Nil? why not use `Result[Muxer,cstring]` since we got no async?

libp2p/connmanager.nim
```diff
@@ -0,0 +1,276 @@
...
+
+proc storeOutgoing*(c: ConnManager, conn: Connection) =
+  conn.dir = Direction.Out
+  c.storeConn(conn)
+
+proc storeIncoming*(c: ConnManager, conn: Connection) =
+  conn.dir = Direction.In
+  c.storeConn(conn)
+
+proc storeMuxer*(c: ConnManager,
```
sinkingsugar, line 209: not async, chance to use result/raises, consumer can just call `tryGet` and magically can throw

libp2p/connmanager.nim
```diff
@@ -0,0 +1,276 @@
...
+
+  let muxer = c.selectMuxer(conn) # always get the first muxer here
+  if not(isNil(muxer)):
+    return await muxer.newStream()
+
+proc dropPeer*(c: ConnManager, peerInfo: PeerInfo) {.async.} =
+  ## drop connections and cleanup resources for peer
+  ##
+
+  for conn in c.conns.getOrDefault(peerInfo.peerId):
```
sinkingsugar, line 260: awaiting inside a iteration is a bit dangerous for race conditions , hmmm maybe `conns` should be interlocked too cos of this.

sinkingsugar: Also `getOrDefault` is really a bad habit I started to realize.. I'd rather defect and understand why there was no record . doing that a lot in gossip 1.1

libp2p/connmanager.nim
```diff
@@ -0,0 +1,276 @@
...
+    if not(isNil(conn)):
+      await c.cleanupConn(conn)
+
+proc close*(c: ConnManager) {.async.} =
+  ## cleanup resources for the connection
+  ## manager
+  ##
+
+  for conns in toSeq(c.conns.values):
```
sinkingsugar, line 269: I guess copying mitigates races, but still not convinced 100%, maybe interlocking is better

libp2p/switch.nim
```diff
@@ -574,44 +430,27 @@ proc subscribePeer*(s: Switch, peerInfo: PeerInfo) {.async, gcsafe.} =
...
+proc publish*(s: Switch, topic: string, data: seq[byte]): Future[int] {.async.} =
   # pubslish to pubsub topic
   if s.pubSub.isNone:
-    var retFuture = newFuture[int]("Switch.publish")
-    retFuture.fail(newNoPubSubException())
-    return retFuture
+    raise newNoPubSubException()
 
-  return s.pubSub.get().publish(topic, data)
+  return await s.pubSub.get().publish(topic, data)
```
sinkingsugar, line 453: Transformation not required still

libp2p/stream/connection.nim
```diff
@@ -63,9 +70,29 @@ method initStream*(s: Connection) =
...
+    inc getConnectionTracker().closed
 
 proc `$`*(conn: Connection): string =
   if not isNil(conn.peerInfo):
     result = conn.peerInfo.id
+
+func hash*(p: Connection): Hash =
+  cast[pointer](p).hash
+
+func `==`*(a, b: Connection): bool =
```
sinkingsugar, line 84: no branch left behind 😄 
https://github.com/status-im/nim-libp2p/blob/6f638259ee81dc87078c17c12700ab5fb5d9feaa/libp2p/peerinfo.nim#L142-L156

libp2p/connmanager.nim
dryajov: I really dislike the no return on the other hand, it isn't explicit and the dangling values make code complicated to read and understand, not point in trusting the compiler if I can't follow the code. It isn't bad in this snippet, but it get bad in more involved ones.

libp2p/connmanager.nim
dryajov: close is guaranteed not to throw

libp2p/connmanager.nim
dryajov: yes, they work pretty reliably just as futures do, they are simply futures unterneat.

libp2p/connmanager.nim
dryajov: Not sure why you call it magic, I yet to see a  bug in nim with a return.

libp2p/connmanager.nim
dryajov: We can, I just haven't used results, but the `cstring` there is puzzling, why not a regular string? Also, the muxer itself never know nor care what it's string is, it's only ever relevant when selecting.

libp2p/connmanager.nim
dryajov: sure, we can try with results.

libp2p/connmanager.nim
dryajov: > awaiting inside a iteration is a bit dangerous for race conditions , hmmm maybe `conns` should be interlocked too cos of this.

Not sure, maybe we want do lock `dropPeer`, we can also do an `allFinished` as well.

libp2p/connmanager.nim
dryajov: Usually copying is fine here, but since this is close, we might want to lock. I would avoid it for now tho.

libp2p/stream/connection.nim
dryajov: really unintuitive and error prone imho.

libp2p/connmanager.nim
dryajov: > if close() fails we never get unlocked here, either lock goes before or completely outside try/catch I think for safety

close won't throw, unless it's a defect.

> also I would use withLock helper, if we didn't remove it, as its interchangeable with nim std and anyway keeps track of lock unlock

I thought we discarded `withLock` precisely because it conflicts with std?

libp2p/switch.nim
dryajov: it is, otherwise you can't use `raise` and have to return a manually constructed `Future`, this is cleaner with very little overhead if any.

libp2p/stream/connection.nim
cheatfate: It looks a little bit over-engineered. If you casting everything to pointers, you can simply do:
```nim
  cast[pointer](a) == cast[pointer](b)
```
In 3rd branch you are comparing pointers and then you are comparing values, but if pointers are equal values are equal too.

libp2p/connmanager.nim
sinkingsugar: That's a cosmetic issue tho, if we have to take that into account there are so many that a I would change 🤣, my suggestion is to consider all branches and wether you put `return` or not does not matter, just always put a value, if you omit the keyword the compiler will analyze and error quite reliably as bonus.

libp2p/connmanager.nim
sinkingsugar: I think we never resolved that conversation. 
What's the conflict tho? Nim has a type system and allows overrides. I think just the auto had to be made an explicit type.

sinkingsugar: > close is guaranteed not to throw

Can we document this somehow, I always wonder about those things, wish we had a nothrow attribute

sinkingsugar:
@dryajov btw I didn't mark anything as to resolve. They are all minor stuff. Nothing critical.

libp2p/connmanager.nim
dryajov: OK, I need to try it out more consistently. But, I understand the optimizations at play and those same optimizations are applied on return; as is an optimization to rely on the type's default value to avoid extra branches, that to me is more explicit and cleaner. 

My beef with this style is precisely style, this gets muddy really quickly and it is not immediately obvious to nim newcomers whats going on here. I always prefer explicit and "do what is expected" styles over magic optimizations and clever code - the simpler and dumber it is the easier it is to interpret by a human.

libp2p/connmanager.nim
dryajov: > What's the conflict tho? Nim has a type system and allows overrides. I think just the auto had to be made an explicit type.

Well, you're overriding a built in with an alternative that does something similar but in a completely different context, confusing at the very least. 

> Can we document this somehow, I always wonder about those things, wish we had a nothrow attribute

Well, that would be our `raises`, but it's a no-go for now, so I'm yet to come up with a succinct way of communicating it without inserting a `# does not throw` at each call site, yaiks!

PR #278, opened 2020-07-16
Channel timeout

dryajov:
add channel activity timeout

### DIFF
--- a/libp2p/muxers/mplex/lpchannel.nim
+++ b/libp2p/muxers/mplex/lpchannel.nim
@@ -274,11 +274,12 @@ proc init*(
     name: name,
     conn: conn,
     initiator: initiator,
+    isLazy: lazy,
+    timeout: timeout,
     msgCode: if initiator: MessageType.MsgOut else: MessageType.MsgIn,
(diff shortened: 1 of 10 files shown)


libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -263,3 +212,105 @@ method close*(s: LPChannel) {.async, gcsafe.} =
...
+  ##
+  ## if the timeout was hit, it means that
+  ## neither incoming nor outgoing activity
+  ## has been detected and the channel will
+  ## be reset
+  ##
+
+  try:
+    while true:
+      await sleepAsync(s.timeout)
```
sinkingsugar, line 227: This future should be started to allow cancellation.

I did something similar for gossip 1.1 (wip)

https://github.com/status-im/nim-libp2p/blob/gossip-one-one/libp2p/switch.nim#L576

dryajov: This is a tricky one because there is no one place to await for the timer task to finish, but we can and should cancel the timer itself. The problem with awaiting, is that the shutdown/close flow is not a single step but at least two in most cases. I need to think a little bit more about this.

PR #279, opened 2020-07-17
Connections and pubsub peers cleanup

dryajov:


libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -87,11 +105,14 @@ proc sendSubs*(p: PubSub,
 
     await peer.sendSubOpts(topics, subscribe)
   except CancelledError as exc:
-    p.handleDisconnect(peer)
+    if not(isNil(peer)) and not(isNil(peer.conn)):
```
sinkingsugar, line 108: if peer is `nil` something is broken tho isn't it? Thought we fixed those issues

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -266,6 +303,18 @@ proc sendHelper*(p: PubSub,
 
   return (published, failed)
 
+proc sendAndCleanup*(p: PubSub,
```
sinkingsugar, line 306: the name kind mislead me.. it gave me the illusion that the peer would be cleaned up all the time , maybe no need to mention the cleanup part.. dunno just sharing my feeling 😄

libp2p/protocols/pubsub/pubsub.nim
dryajov: this is just a bit of defencive programming here.

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -183,11 +219,9 @@ method unsubscribePeer*(p: PubSub, peerInfo: PeerInfo) {.base, async.} =
     let peer = p.peers[peerInfo.id]
 
     trace "unsubscribing from peer", peerId = $peerInfo
-    if not(isNil(peer.conn)):
+    if not(isNil(peer)) and not(isNil(peer.conn)):
```
dryajov, line 222: here however, as opposed to https://github.com/status-im/nim-libp2p/pull/279/files/378544033e73472db4761c527c4ba7bf7a5a0e0f#r456219385, can happen that the peer or connection are simply nil on arrival, which whould indicate an issue upstream, or the peer simply disconnected in the process of trying to establish a pubsub connection.

libp2p/protocols/pubsub/pubsub.nim
dryajov: hmm maybe something like `publishAndDisconnect`, I wan't to make it explicit that this helper will cleanup failed peers/streams.

PR #283, opened 2020-07-21
Handle cancellation in timeoutMonitor

sinkingsugar:
Solves some possible confusion when debugging/tracing

### DIFF
--- a/libp2p/muxers/mplex/lpchannel.nim
+++ b/libp2p/muxers/mplex/lpchannel.nim
@@ -57,7 +57,6 @@ type
     msgCode*: MessageType         # cached in/out message code
     closeCode*: MessageType       # cached in/out close code
     resetCode*: MessageType       # cached in/out reset code
-    timerFut: Future[void]        # the current timer instanse
     timerTaskFut: Future[void]    # the current timer instanse
 
 proc open*(s: LPChannel) {.async, gcsafe.}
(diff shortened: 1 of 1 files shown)


sinkingsugar:
@cheatfate looks good now?

libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -82,11 +81,8 @@ template withEOFExceptions(body: untyped): untyped =
 
 proc cleanupTimer(s: LPChannel) {.async.} =
   ## cleanup timers
-  ##
-  if not(isNil(s.timerFut)) and
-    not(s.timerFut.finished):
-    s.timerFut.cancel()
-    await s.timerTaskFut
+  if not s.timerTaskFut.finished:
```
dryajov, line 84: can we add a test for this, since it appears to be non trivial?

cheatfate: This is protection for double close which happens pretty often inside of `nim-libp2p`. `cleanupTimer()` procedure gets called from 3 different places.

For example `timeoutMonitor()` calls `reset()` which calls `cleanupTimer()`. Also somebody else could call `reset` or `close` procedures which will call `cleanupTimer()`. So this check for `timerTaskFut.finished()` only preserve you from seeing double completion Future[T] defect.

dryajov: ok, we can merge this and add the test later, this seems to be pretty straight forward after all.

PR #291, opened 2020-07-25
Audit multistream fixes

sinkingsugar:


sinkingsugar:
Ideally a state machine would be used but would complicate things a lot.
MS select on the other hand has only 2 states basically so this PR proposed changes might be ok while keeping things simple.

That was my rationale, actually I tried implementing the SM too but really was overkill, hope you agree with this vision @dryajov

PR #295, opened 2020-07-27
pubsub timeouts tuning 

dryajov:
- delayed sending in pubsub
- fine grained timeouts in pubsub
- use `atEof()` instead of `closed()` for reads

PR #303, opened 2020-08-02
misc cleanups

arnetheduck:


arnetheduck:
collected from #295 and other branches

PR #305, opened 2020-08-03
Use constant-time hex encoding/decoding procedures explicitly.

cheatfate:
Fixes https://github.com/status-im/nim-libp2p/issues/300

### DIFF
--- a/libp2p/crypto/crypto.nim
+++ b/libp2p/crypto/crypto.nim
@@ -70,13 +70,14 @@ when supported(PKScheme.Secp256k1):
 
 import ecnist, bearssl
 import ../protobuf/minprotobuf, ../vbuffer, ../multihash, ../multicodec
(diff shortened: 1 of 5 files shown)


libp2p/crypto/crypto.nim
```diff
@@ -70,13 +70,14 @@ when supported(PKScheme.Secp256k1):
 
 import ecnist, bearssl
 import ../protobuf/minprotobuf, ../vbuffer, ../multihash, ../multicodec
-import nimcrypto/[rijndael, twofish, sha2, hash, hmac, utils]
+import nimcrypto/[rijndael, twofish, sha2, hash, hmac]
+import nimcrypto/utils as ncrutils
```
arnetheduck, line 74: let's put a comment here that we use ncrutils because it is constant time (else it's easy to forget and switch to byteutils)

cheatfate: done

PR #307, opened 2020-08-03
Moving idle timeout to Connection to enable across all connection streams

dryajov:


arnetheduck:
the alternative here would be to handle timeouts from the switch - ie it has a registry of all connections so it could iterate over them every second or whatever - then each connection would have a timestamp of the last activity and that would be it. This way, we avoid creating a separate timer for every connection that takes up resources.

dryajov:
> the alternative here would be to handle timeouts from the switch - ie it has a registry of all connections so it could iterate over them every second or whatever - then each connection would have a timestamp of the last activity and that would be it. This way, we avoid creating a separate timer for every connection that takes up resources.

Yeah, but we're recreating chrono's timer mechanism at that point and we'll deff hit the same limits pretty soon, specially if most timeouts come from libp2p either way, I see a lot more room to get it wrong. IMO, we should bite the bullet and optimize chronos to the fullest, that said it actually isn't as bad right now as it used to be.

arnetheduck:
>  recreating chrono's timer

not really - when you create a loop like this you create lots and lots of events that have to be allocated, queued, scheduled, processed etc - you may not see it in _your_ code, but it happens nonetheless.

when you have a field in the connection with a timestamp, you'll loop but avoid the busy work of creating all that garbage work.

arnetheduck:
it's like copying a seq byte by byte instead of using memcpy and copy it as a buffer

dryajov:
Well, you would still have some context associated with each timer object and you would still need a heapqueu/priority queue to keep them, that's exactly what chronos uses.

arnetheduck:
a single time in switch would handle all connections - it's a trivial scheme, compared to the complexity of what goes on in this branch..

each peer has a single field, lastActivity - update this on every access with "current time"

in heartbeat (every second), compare "current time" to "last activity" - there is a single timer involved, and `O(connection)` timestamp comparisons

PR #308, opened 2020-08-04
Dial peerid

arnetheduck:


libp2p/peerinfo.nim
```diff
@@ -29,6 +29,7 @@ type
   PeerInfo* = ref object of RootObj
     peerId*: PeerID
     addrs*: seq[MultiAddress]
+    secureCodec*: string
```
dryajov, line 32: This will not hold, if the peer has several connections from different transports, they could have been secured over different channels. Same for inbound/outbound connections, they could have been secured over different mechanisms.

libp2p/switch.nim
```diff
@@ -279,101 +278,99 @@ proc upgradeIncoming(s: Switch, conn: Connection) {.async, gcsafe.} =
...
+
+    return conn
+
+  let lock = s.dialLock.mgetOrPut($peerId, newAsyncLock())
 
   try:
     await lock.acquire()
-    trace "about to dial peer", peer = id
-    conn = s.connManager.selectConn(peer.peerId)
-    if conn.isNil or (conn.closed or conn.atEof):
```
dryajov, line 295: Why is this being removed? This is a crucial part, it looks for an existing connection and uses that, removing it will force us to create many outgoing connections, which we shouldn't do.

libp2p/peerinfo.nim
arnetheduck: right

libp2p/switch.nim
arnetheduck: look a few lines up - it's just moved outside the lock - also consider what was happening before: on each internalConnect, this would start a new subscribepeer loop!

dryajov: Oh, then it's even worse, because you're breaking the locking altogether, consider when several dials are all happening at the same time, they will all create a socket potentially, this code was inside the lock to prevent that race specifically, the expected behaviour is to lock and have the rest of the dial calls await until the first dial succeeded and then have the other dials pick up the newly created connection. 

The `subscribePeer` call needs to be removed from switch and called by the application, but this requires the hooks changes to be stable.

arnetheduck: > breaking the locking altogether

locking is only relevant between `await` - ie thing of `async` as cooperative multithreading and you won't be too far off - ie a lock protects a section of code so that when `await` is called, the section in the lock is not reentered. 

the `dialLock` is special in that it is only used from within this `proc` - it means that its only purpose is to prevent `internalConnect` calls from running concurrently - but in "synchronous" sections of code, there is already no concurrency - the `selectConn` call for example is not async - thus moving it outside the lock is fine.

> needs to be removed

`asyncCheck cleanupPubSubPeer` is also called for every `internalConnect` - the problem is not constrained to `subscribePeer`

dryajov: > locking is only relevant between await - ... - ie a lock protects a section of code so that when await is called, the section in the lock is not reentered.

That's precisely what it's for

> the dialLock is special in that it is only used from within this proc - it means that its only purpose is to prevent internalConnect calls from running concurrently - but in "synchronous" sections of code, there is already no concurrency - the selectConn call for example is not async - thus moving it outside the lock is fine.

Not at all, this has nothing to do with sync/async inside the `internalConnect`, this, as you said it yourself, is to prevent concurrent dials initiated by callers. If we don't do this we will get a bunch of dials at the same time, moreover, we can't make this procedure sync because dialing is inherently async all the way down to chronos. 

Since the procedure can be called many times in a concurrent manner and we can get many concurrent sockets, we need the lock to prevent that from happening, because as I said, the assumption is to minimize the amount of physical connections.

PR #309, opened 2020-08-04
adding channel limits to mplex

dryajov:


libp2p/muxers/mplex/mplex.nim
```diff
@@ -148,6 +158,10 @@ method handle*(m: Mplex) {.async, gcsafe.} =
       case msgType:
         of MessageType.New:
           let name = string.fromBytes(data)
+          if m.getChannelList(false).len > m.maxChannCount - 1:
+            warn "too many channels created by remote peer", allowedMax = MaxChannelCount
+            raise newTooManyChannels()
```
arnetheduck, line 163: so the peer will be disconnected in this case?

dryajov: yes, it we're getting that many channels from the peer it's probably malfunctioning or malicious, and since we don't have any detailed metrics right now, we can't really make many assumptions.

arnetheduck: well, or they're keeping channels open and relying on our timeouts to close them maybe?

dryajov: That's the thing, they never hit the idle timeout, so they must be still sending something at least once every 5 minutes for inbound and every 2 minutes for outbound. But have a peer open 15K channels isn't right by any stretch and 200 channels should be plenty for any peer out there.

PR #310, opened 2020-08-04
put expensive metrics under a Nim define

stefantalpalaru:
Rationale: we can't handle that many metric samples on our remote servers.
```bash
stefan@master-01.aws-eu-central-1a.nimbus.test:~$ curl -sS localhost:9300/metrics | grep -vE '^(#|$)' | wc -l
9064
```

### DIFF
--- a/libp2p/muxers/mplex/mplex.nim
+++ b/libp2p/muxers/mplex/mplex.nim
@@ -23,7 +23,8 @@ export muxer
 logScope:
   topics = "mplex"
 
-declareGauge(libp2p_mplex_channels, "mplex channels", labels = ["initiator", "peer"])
+when defined(libp2p_expensive_metrics):
+  declareGauge(libp2p_mplex_channels, "mplex channels", labels = ["initiator", "peer"])
 
 type
   Mplex* = ref object of Muxer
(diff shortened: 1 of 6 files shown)


libp2p/muxers/mplex/mplex.nim
```diff
@@ -76,10 +77,11 @@ proc newStreamInternal*(m: Mplex,
     "channel slot already taken!")
 
   m.getChannelList(initiator)[id] = result
-  libp2p_mplex_channels.set(
-    m.getChannelList(initiator).len.int64,
-    labelValues = [$initiator,
-                   $m.connection.peerInfo])
+  when defined(libp2p_expensive_metrics):
```
arnetheduck, line 80: can't we put this behind a template in `libp2p/metrics.nim` which imports and reexports upstream metrics?

stefantalpalaru: Why obfuscate what is happening?

arnetheduck: because metrics are not supposed to break the flow of reading the actual logic of the code in question - they are a monitoring and debugging help, not a feature in and of their own - the less space they take up, the better - and repeating when defined everywhere breaks the flow

stefantalpalaru: > repeating when defined everywhere breaks the flow

Not by much. It's just one extra line and you know what it does from its first couple of words. Compare that with `declareExpensiveLibp2pGauge ...`

mratsim: template have autocompletion, "when defined(...)" do not forcing to check everywhere that we are correctly spelling them.

Also that makes it easy to add new things like a `{.gcsafe.}` in a single location.

PR #312, opened 2020-08-05
more expensive metrics

stefantalpalaru:


dryajov:
Thanks, I think you might be missing some metrics from mplex.nim.

stefantalpalaru:
I don't want to disable all of them. Just the ones with labels that may change significantly over time, generating too many unique combinations over Prometheus' default retention time of 15 days.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -57,17 +57,18 @@ type
     heartbeatRunning: bool
     heartbeatLock: AsyncLock                   # heartbeat lock to prevent two consecutive concurrent heartbeats
 
-declareGauge(libp2p_gossipsub_peers_per_topic_mesh,
-  "gossipsub peers per topic in mesh",
-  labels = ["topic"])
+when defined(libp2p_expensive_metrics):
```
arnetheduck, line 60: these should be added back, they're not expensive - ie the topic list is fairly limited (70 items)

dryajov: yes, it wouldn't hurt, but we can get them back with the define, so might not be a big issue?

stefantalpalaru: > the topic list is fairly limited (70 items)

How many times can the fork version change in 15 days?

PR #313, opened 2020-08-05
Dial peer unstable

arnetheduck:
for some reason, this causes testswitch to fail - there's nothing async in there, so the principal difference is a different order of operations (because the subscribe operation is enqueued one time fewer)


arnetheduck:
fixed in #315

PR #315, opened 2020-08-05
Fix build, add some raises

sinkingsugar:
Wip, still failing


libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -93,6 +93,8 @@ method rpcHandler*(f: FloodSub,
 
                 try:
                   await h(t, msg.data)                 # trigger user provided handler
+                except CancelledError:
+                  raise
```
arnetheduck, line 97: `raise exc` or we might hit a nim bug

PR #320, opened 2020-08-07
peer hooks -> events

arnetheduck:
* peerinfo -> peerid
* include connection direction in event
* check connection status after event
* lock connmanager lookup also when dialling peer
* clean up un-upgraded connection when upgrade fails
* await peer eventing

### DIFF
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -50,22 +50,22 @@ const
 type
     NoPubSubException* = object of CatchableError
 
-    PeerEventKind* {.pure.} = enum
+    ConnEventKind* {.pure.} = enum
       Connected, # A connection was made and securely upgraded - there may be
                  # more than one concurrent connection thus more than one upgrade
                  # event per peer.
       Disconnected # Peer disconnected - this event is fired once per upgrade
                    # when the associated connection is terminated.
 
-    PeerEvent* = object
-      case kind*: PeerEventKind
-      of PeerEventKind.Connected:
+    ConnEvent* = object
+      case kind*: ConnEventKind
+      of ConnEventKind.Connected:
         incoming*: bool
       else:
         discard
 
-    PeerEventHandler* =
-      proc(peerId: PeerID, event: PeerEvent): Future[void] {.gcsafe.}
+    ConnEventHandler* =
+      proc(peerId: PeerID, event: ConnEvent): Future[void] {.gcsafe.}
 
     Switch* = ref object of RootObj
       peerInfo*: PeerInfo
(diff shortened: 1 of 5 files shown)


libp2p/switch.nim
```diff
@@ -280,86 +291,96 @@ proc upgradeIncoming(s: Switch, conn: Connection) {.async, gcsafe.} =
...
+            except CatchableError as exc:
+              # If we failed to establish the connection through one transport,
+              # we won't succeeed through another - no use in trying again
+              await dialed.close()
+              debug "upgrade failed", exc = exc.msg
+              if exc isnot CancelledError:
+                libp2p_failed_upgrade.inc()
+              raise exc
+
+          doAssert not isNil(upgraded), "checked in upgradeOutgoing"
```
dryajov, line 352: I've no idea what this message mean, if it dies here I've no hints to work with.

libp2p/switch.nim
```diff
@@ -50,12 +50,19 @@ const
...
-    Lifecycle* {.pure.} = enum
-      Connected,
-      Upgraded,
-      Disconnected
+    PeerEventKind* {.pure.} = enum
+      Upgraded, # Peer connected and went through secure upgrade
+      Disconnected # Peer disconnected after upgrade
 
-    Hook* = proc(peer: PeerInfo, cycle: Lifecycle): Future[void] {.gcsafe.}
+    PeerEvent* = object
```
dryajov, line 57: Are this peer events or connection events? We should not confuse the two, it's a subtle but fundamental distinction. For context, the peer will have many connections (ie sockets, for some transports) associated with it and it can be considered as disconnected only after the last connection ceased to exist. It isn't clear from the code if that is the case ATM, some clarification would be appropriate. We can/should also have connection events for that reason, which is what the hooks where trying to accomplish.

The Peer lifetime is also crucial to have for an upcoming refactor of the pubsub subsystem, so if this is exactly that then it's a welcome addition, if not, then the renaming of the connection hooks into peer lifetime is at the very least confusing and should be reconsidered.

arnetheduck: well, they're events related to peers - thanks to the object and kind field, it's easy to add related events as well. updating the docs a bit - this is enough to solve what nbc needs for now.

libp2p/switch.nim
arnetheduck: well, the code randomly uses nil and exceptions when a function fails to perform its task, so this is a reminder that nil was checked in upgradeoutgoing and a second check is not needed here - this is more a reminder that if upgradeoutgoing changes, this code needs to change also

libp2p/switch.nim
dryajov: yes, it might be enough for NBC, but they interfere with changes I have in the pipeline for peer lifetime which I rely on heavily in a refactor of pubsub that I've started just recently. Can we please rename this to `connection events`, ie, connection `Upgraded` and connection `Disconnected` otherwise this is plain confusing, specially in lew of the upcoming peer lifetime changes.

libp2p/switch.nim
```diff
@@ -51,15 +51,15 @@ type
     NoPubSubException* = object of CatchableError
 
     PeerEventKind* {.pure.} = enum
```
dryajov, line 53: Sorry, I meant to make this `ConnectionEventKind`, my bad, should have been more explicit.

PR #321, opened 2020-08-07
`msg` is a reserved property name in Chronicles

zah:
Every Chronicles log record has an existing `msg` property matching
the static string supplied in the log statement. Thus, it's currently
not possible to use `msg` as the name of a user property:

https://github.com/status-im/nim-chronicles/issues/86

### DIFF
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -132,7 +132,7 @@ proc send*(
   timeout: Duration = DefaultSendTimeout) {.async.} =
   logScope:
     peer = p.id
-    msg = shortLog(msg)
+    rpcMsg = shortLog(msg)
 
   trace "sending msg to peer"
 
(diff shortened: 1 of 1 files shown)


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -132,7 +132,7 @@ proc send*(
   timeout: Duration = DefaultSendTimeout) {.async.} =
   logScope:
     peer = p.id
-    msg = shortLog(msg)
+    rpcMsg = shortLog(msg)
```
dryajov, line 135: I believe there might be many places in the codebase where `msg` is used... it's not an issue to change those, but it seems like a very big pitfall to have a reserved word in `chronicles` and not warn about it when used as a field, imo chronicles should warn about it.

arnetheduck: myeah, half of our code deals with`msg`:s, sad to not be able to use it

cheatfate: This PR fixes a problem which we currently have, `nim-libp2p` producing incorrect json log lines.

dryajov:
ok, merging... not something to bikeshed right now.

PR #322, opened 2020-08-08
Rework pubsub

dryajov:
Move pubsub out of switch and pass the switch into pubsub to allow it manage connections internally

TODO:

- peer lifetime events to properly trigger `subscribePeer`/`unsubscribePeer`

### DIFF
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -101,7 +101,7 @@ method rpcHandler*(f: FloodSub,
                   trace "exception in message handler", exc = exc.msg
 
         # forward the message to all peers interested in it
-        let published = await f.publishHelper(toSendPeers, m.messages, DefaultSendTimeout)
+        let published = await f.broadcast(toSendPeers, RPCMsg(messages: m.messages), DefaultSendTimeout)
 
         trace "forwared message to peers", peers = published
 
@@ -137,7 +137,7 @@ method publish*(f: FloodSub,
   let msg = Message.init(f.peerInfo, data, topic, f.msgSeqno, f.sign)
 
   # start the future but do not wait yet
-  let published = await f.publishHelper(f.floodsub.getOrDefault(topic), @[msg], timeout)
+  let published = await f.broadcast(f.floodsub.getOrDefault(topic), RPCMsg(messages: @[msg]), timeout)
 
   when defined(libp2p_expensive_metrics):
     libp2p_pubsub_messages_published.inc(labelValues = [topic])
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -31,12 +31,12 @@ type
 method subscribeTopic*(f: FloodSub,
                        topic: string,
                        subscribe: bool,
-                       peerId: string) {.gcsafe, async.} =
+                       peerId: PeerID) {.gcsafe, async.} =
   await procCall PubSub(f).subscribeTopic(topic, subscribe, peerId)
 
   let peer = f.peers.getOrDefault(peerId)
   if peer == nil:
-    debug "subscribeTopic on a nil peer!", peer = peerId
+    debug "subscribeTopic on a nil peer!", peer = $peerId
     return
 
   if topic notin f.floodsub:
@@ -51,16 +51,15 @@ method subscribeTopic*(f: FloodSub,
     # unsubscribe the peer from the topic
     f.floodsub[topic].excl(peer)
 
-method handleDisconnect*(f: FloodSub, peer: PubSubPeer) =
+method unsubscribePeer*(f: FloodSub, peer: PubSubPeer) =
   ## handle peer disconnects
   ##
-  
-  procCall PubSub(f).handleDisconnect(peer)
 
-  if not(isNil(peer)) and peer.peerInfo notin f.conns:
-    for t in toSeq(f.floodsub.keys):
-      if t in f.floodsub:
-        f.floodsub[t].excl(peer)
+  procCall PubSub(f).unsubscribePeer(peer)
+
+  for t in toSeq(f.floodsub.keys):
+    if t in f.floodsub:
+      f.floodsub[t].excl(peer)
 
 method rpcHandler*(f: FloodSub,
                    peer: PubSubPeer,
@@ -77,7 +76,7 @@ method rpcHandler*(f: FloodSub,
         if msgId notin f.seen:
           f.seen.put(msgId)                          # add the message to the seen cache
 
-          if f.verifySignature and not msg.verify(peer.peerInfo):
+          if f.verifySignature and not msg.verify(peer.peer):
             trace "dropping message due to failed signature verification"
             continue
 
@@ -117,11 +116,6 @@ method init*(f: FloodSub) =
 
   f.handler = handler
   f.codec = FloodSubCodec
-
-method subscribePeer*(p: FloodSub,
-                      conn: Connection) =
-  procCall PubSub(p).subscribePeer(conn)
-  asyncCheck p.handleConn(conn, FloodSubCodec)
 
 method publish*(f: FloodSub,
                 topic: string,
@@ -167,8 +161,6 @@ method unsubscribeAll*(f: FloodSub, topic: string) {.async.} =
 
 method initPubSub*(f: FloodSub) =
   procCall PubSub(f).initPubSub()
-  f.peers = initTable[string, PubSubPeer]()
-  f.topics = initTable[string, Topic]()
   f.floodsub = initTable[string, HashSet[PubSubPeer]]()
   f.seen = newTimedCache[string](2.minutes)
   f.init()
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -33,12 +33,7 @@ method subscribeTopic*(f: FloodSub,
                        subscribe: bool,
                        peerId: PeerID) {.gcsafe, async.} =
   await procCall PubSub(f).subscribeTopic(topic, subscribe, peerId)
-
   let peer = f.peers.getOrDefault(peerId)
-  if peer == nil:
-    debug "subscribeTopic on a nil peer!", peer = $peerId
-    return
-
   if topic notin f.floodsub:
     f.floodsub[topic] = initHashSet[PubSubPeer]()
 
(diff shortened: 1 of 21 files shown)


libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -51,16 +51,15 @@ method subscribeTopic*(f: FloodSub,
...
-  
-  procCall PubSub(f).handleDisconnect(peer)
 
-  if not(isNil(peer)) and peer.peerInfo notin f.conns:
-    for t in toSeq(f.floodsub.keys):
-      if t in f.floodsub:
-        f.floodsub[t].excl(peer)
+  procCall PubSub(f).unsubscribePeer(peer)
+
+  for t in toSeq(f.floodsub.keys):
```
arnetheduck, line 60:
```suggestion
  for _, v in f.floodsub.mitems(): v.excl(peer)
```

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -60,48 +61,24 @@ type
...
+  peer.subscribed = false
+  libp2p_pubsub_peers.set(p.peers.len.int64)
 
 proc sendSubs*(p: PubSub,
                peer: PubSubPeer,
                topics: seq[string],
                subscribe: bool) {.async.} =
   ## send subscriptions to remote peer
-  asyncCheck peer.sendSubOpts(topics, subscribe)
+  await peer.sendSubOpts(topics, subscribe)
```
arnetheduck, line 76: remove `await` and `async` - one future less

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -47,10 +48,10 @@ type
     handler*: seq[TopicHandler]
 
   PubSub* = ref object of LPProtocol
+    switch*: Switch                                 # the switch used to dial/connect to peers
```
arnetheduck, line 51: instead of the switch here, how about just taking a `DialCallback` closure? that'll disconnect pubsub from the switch implementation, allowing it to be more "standalone" which can help testing and break cyclic dependencies - in particular, it would enable the simple creation of tests that work the failure path

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -37,34 +37,27 @@ type
     onSend*: proc(peer: PubSubPeer; msgs: var RPCMsg) {.gcsafe, raises: [Defect].}
 
   PubSubPeer* = ref object of RootObj
-    proto*: string                      # the protocol that this peer joined from
+    switch*: Switch                     # switch instance to dial peers
+    codec*: string                      # the protocol that this peer joined from
     sendConn: Connection
-    peerInfo*: PeerInfo
+    peer*: PeerID
```
arnetheduck, line 43:
```suggestion
    peerId*: PeerID
```
easier to distinguish

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -155,41 +149,36 @@ proc send*(
...
+  try:
     trace "about to send message"
+    if not p.connected:
+      try:
+        await p.sendLock.acquire()
+        trace "no send connection, dialing peer"
+        p.sendConn = await p.switch.dial(
+          p.peer, p.codec)
+          .wait(DefaultSendTimeout) # get a send connection if there is none
+        asyncCheck p.handle(p.sendConn) # install a reader on the send connection
```
arnetheduck, line 161: comments! :+1:

libp2p/protocols/pubsub/pubsub.nim
dryajov: not a bad idea 👍

libp2p/protocols/pubsub/floodsub.nim
dryajov: This is still prone to table modified issues, even if we don't run into it again, it'll be easy to sneak in later changes.

arnetheduck: well, reading this code is weird because it looks like modification protection, even though it's not needed - ie I'd expect there to be modifications but there aren't so it's inconsistent

libp2p/protocols/pubsub/pubsub.nim
dryajov: sending is `async`

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -47,61 +48,74 @@ type
...
+
+  return broadcasted.len
 
 proc sendSubs*(p: PubSub,
                peer: PubSubPeer,
                topics: seq[string],
                subscribe: bool) {.async.} =
   ## send subscriptions to remote peer
-  asyncCheck peer.sendSubOpts(topics, subscribe)
+  discard await p.broadcast([peer], RPCMsg(subscriptions: topics.mapIt(SubOpts(subscribe: subscribe, topic: it))), DefaultSendTimeout)
```
arnetheduck, line 113: using broadcast here looks like a long and winding road compared to simply doing `peer.send`..?

arnetheduck: ie the `use-only-broadcast` idea may be fine, but this takes it over the top - semantically, we're not broadcasting here, and the extra future traffic is weird - in addition, keep in mind that the generic code above creates a fair amount of bloat - ie the compiler will generate separate/duplicate  C functions for every type in the signature multiplied by all the continuations, exception bloat etc

sinkingsugar: We use broadcast to unify the maintenance to do in the case the send fails. This was discussed with @dryajov on discord. 
Basically a pure `send` right now might not perform the required maintenance.

Maybe broadcast is not the right name btw.

Btw I'm a bit tired of hearing that nim X or Y feature should not be used. Generics are a good thing. Much easier to understand then a macro or template...

sinkingsugar: But I just realized this was an asyncCheck! That might be better stay. ( maybe that's what you meant )

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -47,61 +48,74 @@ type
...
-    let peer = conn.peerInfo
-    await conn.closeEvent.wait()
+  libp2p_pubsub_peers.set(p.peers.len.int64)
 
-    if peer in p.conns:
-      p.conns[peer].excl(conn)
-      if p.conns[peer].len <= 0:
-        p.conns.del(peer)
+proc broadcast*(p: PubSub,
+                sendPeers: HashSet[PubSubPeer] | seq[PubSubPeer] | array[1, PubSubPeer],
```
dryajov, line 76: We don't need the generic here as @arnetheduck noted, a `seq` should do just fine. This should also be private, I don't see a reason to expose it, maybe tests?

libp2p/protocols/pubsub/pubsub.nim
arnetheduck: sounds like a pure send that maintains things is needed then - ie it's always easier to reason about _one_ thing that sends, deal with the tricky issues, and then accumulate the results - trying to do both will often lead to issues when errors happen in the middle of the loop for example

dryajov: Well, that's what `broadcast` is for, it deals with sending messages to peers and updating the state of the mesh according to the outcome. `PubSubPeer`'s `send` on the other hand, deals with maintaining state for one peer.

libp2p/protocols/pubsub/pubsub.nim
sinkingsugar: Well, `array[1, PubSubPeer]` allows 0 heap allocations in the case of a single peer, also `HashSet[PubSubPeer] | seq[PubSubPeer]` is needed cos we consume both with hash and seq...

Seriously what is wrong with using generics?

libp2p/protocols/pubsub/pubsub.nim
arnetheduck: my point is more that there's no reason for the unsubscribepeer to be in a proc that _also_ deals with accumulating a result - the cleanup is done per-peer independently, so following the logic in a single-peer-send will be more easy to reason about because it will not be interspersed with peer looping logic

dryajov: Yes, but it's not possible because the peer knows nothing about the mesh state, that's all dealt by the specific pubsub implementation, ie the peer would not know which table to remove itself from. I would also much prefered this was dealt right at the point of send in the peer, but it isn't possible and this is the next best thing.

arnetheduck: this is a very common source of bugs in the libp2p code that is trivial to avoid by simply placing the logic at the abstraction level where it belongs - if a peer send fails, it's independent of other peer sends failing, in gossipsub - therefore, it belongs somewhere where only a single peer is being dealt with - the code will be much easier to read and debug if it follows this simple principle.


consider for a minute what you're doing: you're initiating a bunch of sends in a loop where you have a reference to the peer. you then save the id, forget about this peer reference, and start a new loop to do the cleanup, doing the lookup of the peer by id _again_ - not only does this introduce the potential for race conditions - it is also inefficient and harder to reason about.

libp2p/protocols/pubsub/pubsub.nim
arnetheduck: the use of generics has costs - in the case of async, these costs are significant in the sense that async itself introduces a lot of bloat that is multiplied for each generic type - in this particular case, there's hardly to gain by the generics: the sendPeers data will be copied over and over again as it travels through the async layers and closures - there are tens of heap allocations here already that async itself introduces, in addition to the return value etc. put in this perspective, it's questionable whether the bloat price is worth the difference in number of heap allocations.

arnetheduck: but whatever, it really doesn't matter that much either way - what matters is more the semantic juggling of a single send vs an aggegation of multiple sends

libp2p/protocols/pubsub/pubsub.nim
arnetheduck: > not possible

`proc send(p: PubSub, peer: PubSubPeer, msg:seq[byte]) {.async.}` ?

dryajov: @sinkingsugar @arnetheduck has a point, we can further simplify it by moving it to a pubsub `.send(...)` method.

sinkingsugar: @dryajov you took care of it I see, nothing else to add on my side right?

dryajov: We should be good for now.

dryajov:
This changes are working pretty well locally and I would like to get this merged ASAP. However, a few minor issues still exist:

- It appears that some counts for the `libp2p_gossipsub_peers_per_topic_mesh` and `libp2p_gossipsub_peers_per_topic_gossipsub` are off
- @arnetheduck suggestion of passing a callback/interface instead of the entire switch is sound, but it would require additional changes that can bloat the PR even more

Given the above, I think this can be merged and the remaining issues addressed as separate PRs

cc: @arnetheduck @sinkingsugar

libp2p/protocols/pubsub/floodsub.nim
sinkingsugar: Well, also there are no await in the loop, so what @arnetheduck proposed will be fine and faster. But end result equal anyway for now.

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -83,12 +79,13 @@ proc sendObservers(p: PubSubPeer, msg: var RPCMsg) =
...
   logScope:
     peer = p.id
+
   debug "starting pubsub read loop for peer", closed = conn.closed
   try:
     try:
       while not conn.atEof:
         trace "waiting for data", closed = conn.closed
-        let data = await conn.readLp(64 * 1024).wait(DefaultReadTimeout)
+        let data = await conn.readLp(64 * 1024)
```
sinkingsugar, line 88: is it fine to remove the timeout here?

dryajov: Yes, this was intentional, I don't think we need one there.

PR #324, opened 2020-08-10
Fix half closed

dryajov:


dryajov:
@zah this attempt to fix the issue you pointed out, let me know if this works for you.

tests/testmplex.nim
```diff
@@ -135,18 +135,20 @@ suite "Mplex":
...
+          timeout = 5.minutes
         )
         chann = LPChannel.init(1, conn, true)
 
       await chann.pushTo(("Hello!").toBytes)
-      let closeFut = chann.closeRemote()
 
       var data = newSeq[byte](6)
-      await chann.readExactly(addr data[0], 6) # this should work, since there is data in the buffer
+      await chann.readExactly(addr data[0], 3) # this should work, since there is data in the buffer
```
zah, line 146: Can you add a test where you write to a stream after the other end has half-closed it? This will ensure that we won't get an unnoticed regression for the usage pattern in the Eth2 syncing protocol.

PR #326, opened 2020-08-11
add missing mplex half closed test

dryajov:


tests/testmplex.nim
```diff
@@ -130,7 +130,40 @@ suite "Mplex":
...
+      await chann.pushTo(("Hello!").toBytes)
+
+      var data = newSeq[byte](6)
+      await chann.close() # closing channel
+      # should be able to read on local clsoe
+      await chann.readExactly(addr data[0], 3)
+      # closing remote end
+      let closeFut = chann.closeRemote()
+      # should still allow reading until buffer EOF
+      await chann.readExactly(addr data[3], 3)
```
arnetheduck, line 152: this could be split into two reads to ensure that partial readexactly works as well (since it's the last bytes)

another test is that if a readexactly for 10 bytes is issued (before the close) and only 6 bytes + eof arrives, the 10-byte read fails

PR #327, opened 2020-08-13
simplify mplex

arnetheduck:
* less async
* less copying of data
* less redundant cleanup

### DIFF
--- a/libp2p/muxers/mplex/coder.nim
+++ b/libp2p/muxers/mplex/coder.nim
@@ -46,7 +46,8 @@ proc writeMsg*(conn: Connection,
                id: uint64,
                msgType: MessageType,
                data: seq[byte] = @[]) {.async, gcsafe.} =
-  trace "sending data over mplex", id,
+  trace "sending data over mplex", oid = $conn.oid,
+                                   id,
                                    msgType,
                                    data = data.len
   var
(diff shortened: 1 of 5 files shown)


dryajov:
There are a couple of good changes there like making `newStreamInternal` sync and the use of `writeSeq` (is that a new?), but overall this seems like "making the code look more readable" for the sake of it,  not a bad thing to do, but it's adding more noise at this point and potentially masking the issues even more?

arnetheduck:
Knowing what the code does is a first step to understanding it - for example, take `cleanupChann` - its task is (more or less) to remove the channel from the channel table, so the most natural place to "schedule" that cleanup is where the channel is added to said table, so that the two are close as they balance each other - this removes the possibility that race conditions or early exceptions make the table go out of sync, as well as redundant calls to the function.

There's also changes to remove `getChannelList`, which otherwise makes a full copy of the table in some places - this scales poorly.

The removal of async also helps reason about the code, in the sense that it happens more closely to the way it's written.

dryajov:
I want to say first, that I'm not against this rafactors when things are working properly, but when we have an hard to pinpoint issue like the one we're after the constant changes to the codebase makes uncovering the issue harder, because we're constantly shifting things around. 

> Knowing what the code does is a first step to understanding it

Sure, but in making it more readable/understandable for yourself, you're rewriting it and now it's someone else's code I have to reason about. Having at least one person that knows all the ins and outs of the codebase, is pretty valuable when the codebase isn't stable.

In any case, I don't want to make it sound like I'm resisting the change, I'm raising a flag about how reasonable it is to refactor everything when there are specific issues/bugs. What is the goal? Are we trying to clean the issue up by sheer chance? I'm a bit wary of this approach.

> There's also changes to remove getChannelList, which otherwise makes a full copy of the table in some places - this scales poorly.

I was careful to return a `var` of the table, when would the copy be made?

arnetheduck:
>  What is the goal? 

the goal is to remove opportunities where async races and table maintenance mismatches happen - that inevitably involves moving things closer to each other. 

> when would the copy be made?

https://github.com/status-im/nim-libp2p/pull/327/files#diff-791568a09e61b8545da7c644adbe2f6dL144

returning `var` is a nice idea, but out of touch with reality: it's invisible at the call site hence will inevitably lead to misuse - ie sometimes it's necessary, but the copy semantics of nim make it unworkable in practise. this is also a classic way to introduce bugs: you think you're modifying the content of the entry in the table (that was returned with `var`, say `var v =table[key]; v.field = true`), but you're modifying the copy of it instead - you can't visually tell the difference when reading the code

PR #328, opened 2020-08-14
A minimal patch ensuring local finalization with Lighthouse

zah:
This is the minimal set of changes that lead to deterministic finalization with Lighthouse in a local testnet driven by the multinet scripts. I'm calling them "changes" because I don't claim this is a proper way to address the underlying problems.

1) Fixes a race during gossip initialization

2) Disables the connection inactivity monitor because the connection
   with Lighthouse is being flagged as inactive even when it delivers
   Gossip traffic.

3) Logs more events of interest

### DIFF
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -465,6 +465,12 @@ proc stop*(s: Switch) {.async.} =
   trace "switch stopped"
 
 proc muxerHandler(s: Switch, muxer: Muxer) {.async, gcsafe.} =
+  # store incoming connection
+  s.connManager.storeIncoming(muxer.connection)
+
+  # store muxer and muxed connection
+  s.connManager.storeMuxer(muxer)
+
   var stream = await muxer.newStream()
   defer:
     if not(isNil(stream)):
@@ -474,6 +480,7 @@ proc muxerHandler(s: Switch, muxer: Muxer) {.async, gcsafe.} =
     # once we got a muxed connection, attempt to
     # identify it
     await s.identify(stream)
+
     if isNil(stream.peerInfo):
       await muxer.close()
       return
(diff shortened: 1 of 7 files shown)


libp2p/switch.nim
```diff
@@ -465,6 +465,12 @@ proc stop*(s: Switch) {.async.} =
   trace "switch stopped"
 
 proc muxerHandler(s: Switch, muxer: Muxer) {.async, gcsafe.} =
+  # store incoming connection
```
dryajov, line 468: @zah this isn't correct, identify *has* to happen before adding the connection. Identify performs peer verification and if that fails, the connection should be terminated and the upgrade stopped.

zah: I figured, but we can try to undo the registration in that case perhaps? The alternative that I considered is creating a future for the muxer record, so any dials will wait for it to be properly registered.

arnetheduck: technically, we verify the peer in noise already, as part of the handshake - ie that's where peer identity is first encountered / set up

arnetheduck: ie in the peerid branch from before, that's one of the things that were firmed up, that there are fewer peerinfo objects flying around - that said, noise creates one for the peer with the right public key which gives an identity which is enough to work with on the muxer side

dryajov: We should not create a new peerinfo object either way, we should reuse/update the one that is already setup. 

We might be able to get away with running identify after muxing, but that is definitely not the expected flow.

I've also ran this changes in medalla and it doesn't seem to improve by much, which tells me that there is something else going on.

arnetheduck:
#330 sets activity flag in a few places that were missing (though the whole mechanism should probably be refactored so as to not use a timer for every `Connection` object, instead checking for timeouts for example in `mplex` etc on a heartbeat

libp2p/switch.nim
arnetheduck: > We should not create a new peerinfo object either way

this would require some pretty major refactoring - I started it in the peerid branch but it takes a bit of work as it would require a centralized repository of peerinfo objects

arnetheduck: > undo the registration

the registration should ideally be followed by announcing the connection to the world, so that nbc for example can do its status dance

zah:
The fixes were merged in other PRs

PR #329, opened 2020-08-14
buffer queue

arnetheduck:


libp2p/switch.nim
```diff
@@ -471,6 +468,16 @@ proc muxerHandler(s: Switch, muxer: Muxer) {.async, gcsafe.} =
       await stream.close()
 
   try:
+    if isNil(stream.peerInfo):
+      await muxer.close()
+      return
+
+    # store incoming connection
+    s.connManager.storeIncoming(muxer.connection)
```
zah, line 476: I've just tested this branch and it's not finalizing with Lighthouse. I expected this, because I've tested the same `storeIncoming` location in my test branch as well. It's not just `identity` that can introduce a race, but rather the very first line here (`await muxer.newStream()`) and that's why I had to put the storeIncoming lines before any other async op here. Sorry for the misleading explanation.

zah: With the additional tweak here, it does finalize.

arnetheduck: I was hoping a little though that the de-asyncified creation of mplex streams in `simplify mplex` would have helped but the `open` op is still async so .. well ..

zah: It did help in the sense that I had to make several runs to trigger the problem :) A better hiding place for the bug.

dryajov: It should be fine stirring the connection early, if the mixer close because of an error it will be cleaned up from the manager.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -157,10 +157,12 @@ proc rebalanceMesh(g: GossipSub, topic: string) {.async.} =
   # Send changes to peers after table updates to avoid stale state
   let graft = RPCMsg(control: some(ControlMessage(graft: @[ControlGraft(topicID: topic)])))
```
dryajov, line 158: should be moved inside the if.

libp2p/protocols/secure/secio.nim
```diff
@@ -236,6 +236,7 @@ method write*(sconn: SecioConn, message: seq[byte]) {.async.} =
 
     trace "Writing message", message = msg.shortLog, left, offset
     await sconn.stream.write(msg)
+    sconn.activity = true
```
dryajov, line 239: this is problematic, we should just move the flag setting down to `lpstream`, I knew I should have done it initially and it  would have probably saved some issues.

libp2p/stream/bufferstream.nim
```diff
@@ -114,199 +98,89 @@ method initStream*(s: BufferStream) =
...
-  ## wait for all data in the buffer to be consumed
-  ##
+  if data.len == 0:
+    return
 
-  trace "draining buffer", len = s.len
-  while s.len > 0:
-    await s.dataReadEvent.wait()
-    s.dataReadEvent.clear()
+  await s.readQueue.addLast(data)
```
dryajov, line 128: What happens if I push a GBs of data here? There aren't any limits on the buffer size.

arnetheduck: currently, mplex loads data into a seq that it holds in memory, then pushes that seq byte by byte to bufferstream, synchronizing for every byte.

in this PR, the bufferstream has a bounded 1-item queue, so mplex will hold the same seq in memory as it had before (which is limited), then push it in one go to the queue and will not read the next one until this one has been processed.

libp2p/muxers/mplex/mplex.nim
```diff
@@ -121,7 +121,7 @@ method handle*(m: Mplex) {.async, gcsafe.} =
       trace "stopping mplex main loop"
       await m.close()
 
-    while not m.connection.atEof:
+    while true:
```
dryajov, line 124: where do you exit?

arnetheduck: death by exception

arnetheduck:
#346 replaces this with a cleaned-up version

PR #333, opened 2020-08-16
avoid crash when subscribe is received

arnetheduck:
...by making subscribeTopic synchronous, avoiding a peer table lookup
completely.

rebalanceMesh will be called a second later - it's fine


arnetheduck:
in particular, when someone is subscribing to a topic, they will likely be checking their mesh for peers they want to subscribe to - delaying the subscription avoid some potentially unnecessary cross-traffic

PR #334, opened 2020-08-16
remove send lock

arnetheduck:
When mplex receives data it will block until a reader has processed the
data. Thus, when a large message is received, such as a gossipsub
subscription table, all of mplex will be blocked until all reading is
finished.

However, if at the same time a `dial` to establish a gossipsub send
connection is ongoing, that `dial` will be blocked because mplex is no
longer reading data - specifically, it might indeed be the connection
that's processing the previous data that is waiting for a send
connection.

There are other problems with the current code:
* If an exception is raised, it is not necessarily raised for the same
connection as `p.sendConn`, so resetting `p.sendConn` in the exception
handling is wrong
* `p.isConnected` is checked before taking the lock - thus, if it
returns false, a new dial will be started. If a new task enters `send`
before dial is finished, it will also determine `p.isConnected` is
false, then get stuck on the lock - when the previous task finishes and
releases the lock, the new task will _also_ dial and thus reset
`p.sendConn` causing a leak.


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -117,11 +116,44 @@ proc handle*(p: PubSubPeer, conn: Connection) {.async.} =
...
+    # get a send connection if there is none
+    res = await p.switch.dial(p.peerId, p.codec) # ...and here
+    if res == nil:
+      return nil
+
+    # It so happens that there may be multiple dial attempts in-flight here - we
+    # also cannot take a lock because that might block mplex backpressure
+    # handling - thus we might redundantly create more send connections than
+    # would normally be needed - we'll cycle through them as we move along..
+    p.sendConn = res
```
zah, line 148: I don't quite understand the reasoning here. Why don't you handle the case where `p.sendConn` is not nil here? (it was set by a concurrent `dial` attempt). It would be easy to close the redundant connection, return the established one and then move on.

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -117,11 +116,44 @@ proc handle*(p: PubSubPeer, conn: Connection) {.async.} =
...
   except CatchableError as exc:
     trace "Exception occurred in PubSubPeer.handle", exc = exc.msg
 
+proc getSendConn(p: PubSubPeer): Future[Connection] {.async.} =
+  # get a cached send connection or create a new one - need to call
+  # `putSendConn` after use! If one is not available, set up a new one
+  var res = p.sendConn
+
+  if not res.isNil:
+    if res.closed() or res.atEof:
```
zah, line 133: Wouldn't this code be more natural to read with an early return?

```nim
if res != nil:
  if not (res.closed or res.atEof):
    return res
  p.sendConn = nil
  await res.close() # p.sendConn might change after here
 
res = await p.switch.dial(...)
```

It's a cache after all, so an early return is a natural thing to have.

libp2p/protocols/pubsub/pubsubpeer.nim
arnetheduck: it's more or less the same: we have two send connections thus one of them has to be removed - doesn't really matter which?

arnetheduck: alright, I changed it to reuse the existing connection - it's actually slightly more simple that way because the existing connection cannot be closed at once as it might still be used in another concurrent send call..

libp2p/protocols/pubsub/pubsubpeer.nim
arnetheduck: changed it, but I'm not sure the logic is any more simple to follow

PR #335, opened 2020-08-17
work around race conditions between identify and other protocols

arnetheduck:
when identify is run on incoming connections, the connmanager tables are
updated too late for incoming connections to properly be handled

this is a quickfix that will eventually need cleaning up


arnetheduck:
this concludes the changes from #328

PR #336, opened 2020-08-17
Restore interop with Lighthouse by preventing concurrent meshsub dials

zah:
For more details, see the added commend in `getSendConn`

### DIFF
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -46,6 +46,7 @@ type
     recvdRpcCache: TimedCache[string]   # cache for already received messages
     observers*: ref seq[PubSubObserver] # ref as in smart_ptr
     subscribed*: bool                   # are we subscribed to this peer
+    dialLock: AsyncLock
 
   RPCHandler* = proc(peer: PubSubPeer, msg: seq[RPCMsg]): Future[void] {.gcsafe.}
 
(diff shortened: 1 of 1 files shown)


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -141,31 +142,36 @@ proc getSendConn(p: PubSubPeer): Future[Connection] {.async.} =
...
-  asyncCheck p.handle(newConn) # start a read loop on the new connection
-
-  return newConn
+  try:
+    # Testing has demonstrated that when we perform concurrent meshsub dials
+    # and later close one of them, other implementations such as rust-libp2p
+    # become deaf to our messages (potentially due to the clean-up associated
+    # with closing connections). To prevent this, we use a lock that ensures
+    # that only a single dial will be performed for each peer:
+    await p.dialLock.acquire()
```
arnetheduck, line 151: ok, so this is the lock that causes a deadlock in dial - while the dial is ongoing, no reads happen from the bufferstream

zah: I'll try to study this problem tomorrow, but please note that this locking is not quite the same as the previous one. The lock will be acquired only once during the initial milliseconds of each physical connection. All subsequent calls to `getSendConn` will hit the cache.

dryajov: This will still lock up, we do need the lock to prevent multiple connections from being opened, but we need to be careful not to lock up the rest of the dial flow, for this we need to make make sure that send isn't awaited at the call site.

arnetheduck: It's not so much about milliseconds, but rather about the flow in which the dial happens:
* the remote opens an mplex channel for gossip (their sendconn essentially)
* they send us data and we want to send some subscriptions / grafts to them, as a result of them sending us theirs
* from within the read loop of their mplex channel, we call `dial`, blocking the read loop
* they send us more data, so the mplex read loop gets blocked in pushTo (because the readloop reading the channel is blocked), and stops working for _all_ channels for that peer, including the one we're trying to dial

arnetheduck: in https://github.com/status-im/nim-libp2p/pull/333, I removed `rebalanceMesh` from being run when they send us their subscription list - this helps "break" the lock by not sending from within the read loop - but there are other such cases still. In general, any send from within the handler of their channel will have this issue - one thing we discussed with @dryajov was to make `send` consistently non-blocking (it would just queue up the work but not await it) - it would have to handle its own errors at that stage (which makes sense), and would systematically avoid this issue and simplify all code flows.

it would also help keep error handling more local, which in general makes the code more approachable.

zah: From the description above, my understanding is that the source of the deadlock is the act of dialing within a read loop, not strictly the additional lock that prevents or delays the dialing, is this correct?

If I have to generalize it further, the source of the deadlock is awaiting within your read loop a write operation that handles back-pressure. The fault in such a deadlock lies partly with the other peer who must be stuck in its own writing operation that prevents reading from the socket. To avoid the deadlock, you must break the back-pressure mechanism somewhere along the chain (e.g. by introducing a queue with an unlimited size as you have suggested).

The specific `pushTo` deadlock seems a bit more self-inflicted though, so more thought have to be put into it.

But in the meantime, this branch restores finalization with Lighthouse and it works fine for `eth2_network_simulation`. Shall we merge it as a step forward towards carrying more tests with Lighthouse until the deadlock issue is solved in a more sound manner? I could make some hand-wavy argument that it actually reduces the chances for the deadlock, because it prevents the start of the dial that might be problematic.

arnetheduck: well, the lock causes all dials from all sources to get stuck, in particular the one from read loop that handles the incoming connection and thus blocks `pushTo` indirectly - basically, if from within nbc we start dialling, and the incoming read loop also wants to dial (because it's sending stuff), the nbc dial will block the loop dial - in that sense, the lock is the problem, not the send.

on the other hand, if from within the read loop we never send, we will also not dial.

by not waiting for `send`, ever, we implicitly introduce an unbounded queue that breaks the deadlock - this requires `send` to handle all its errors, in particular - it also means that `send` calls will arrive _after_ the channel has closed etc, and we must thus be prepared to handle these kinds of delays and reorderings of operations, which in and of itself is a challenge.

arnetheduck: regarding probabilities, it's still possible to deadlock - the subscription table is a big send that happens in the local sim, but in the real world, the big sends will be blocks (and perhaps states) - we can merge the branch, but we'll continue seeing spurios locks of this kind - the side effects of a deadlock here are hard to diagnose - they typically cause things like block request timeouts that are hard to correlate with the lock problem (ie this is one reason why we have a hard time staying connected with peers: we deadlock, timeout then get disconnected but it looks like for example sync manager is the problem)

the buffer I put in bufferstream is ridiculously large now - 100k - we should not see it lock up for any practical reason in medalla thus I'd be ok with merging this branch as long as we include the explanation of the deadlock I wrote above, and probably open an issue to track it - finding it took quite some effort.

zah: Commends added in af0955c and issue created here:
https://github.com/status-im/nim-libp2p/issues/337

dryajov: I've added a comment here https://github.com/status-im/nim-libp2p/issues/337#issuecomment-675500710. The TLDR is that, it seems unlikely that the lock up is happening the way it was described by @arnetheduck. We need to trace this flow more carefully.

PR #338, opened 2020-08-19
work around send that may raise

arnetheduck:
`send` can raise exceptions that together with asyncCheck will
crash NBC

### DIFF
--- a/libp2p/protocols/pubsub/pubsub.nim
+++ b/libp2p/protocols/pubsub/pubsub.nim
@@ -8,7 +8,7 @@
 ## those terms.
 
 import std/[tables, sequtils, sets]
-import chronos, chronicles, metrics
+import chronos, chronicles, chronicles/chronos_tools, metrics
 import pubsubpeer,
        rpc/[message, messages],
        ../../switch,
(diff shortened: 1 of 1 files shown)


libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -197,7 +197,11 @@ method subscribePeer*(p: PubSub, peer: PeerID) {.base.} =
 
   let pubsubPeer = p.getOrCreatePeer(peer, p.codec)
   if p.topics.len > 0:
-    asyncCheck p.sendSubs(pubsubPeer, toSeq(p.topics.keys), true)
+    # TODO sendSubs may raise, but doing asyncCheck here causes the exception
+    #      to escape to the poll loop.
+    #      With a bit of luck, it may be harmless to ignore exceptions here -
+    #      some cleanup is eventually done in PubSubPeer.send
+    traceAsyncErrors p.sendSubs(pubsubPeer, toSeq(p.topics.keys), true)
```
sinkingsugar, line 204: is this deliberate so we actually crash and catch the issue?
Also not sure how much that helps given our traces are hardly containing any relevant info.

PR #339, opened 2020-08-20
No raise send

dryajov:


libp2p/stream/bufferstream.nim
```diff
@@ -129,14 +129,16 @@ proc initBufferStream*(s: BufferStream,
   if not(isNil(handler)):
     s.writeHandler = proc (data: seq[byte]) {.async, gcsafe.} =
       defer:
-        s.writeLock.release()
+        if s.writeLock.locked:
```
arnetheduck, line 132: why do you need to check that the lock is acquired? what sequence causes it not to be acquired?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -450,10 +450,14 @@ method rpcHandler*(g: GossipSub,
...
         try:
           info "sending control message", msg = respControl
-          await g.send(
+          let sent = await g.send(
             peer,
             RPCMsg(control: some(respControl), messages: messages),
             DefaultSendTimeout)
+
+          if not sent:
+            g.unsubscribePeer(peer.peerId)
```
arnetheduck, line 459: NBC (or another application using gossipsub) is supposed to call `subscribePeer` and `unsubscribePeer` - why is done here alse? Their state will be inconsistent.

libp2p/stream/bufferstream.nim
dryajov: If the lock isn't acquired `release` will throw - maybe not needed, I'm either way on this one.

libp2p/protocols/pubsub/gossipsub.nim
dryajov: > NBC (or another application using gossipsub) is supposed to call subscribePeer and unsubscribePeer

Because peers can be cleared as a result of either a failed send/broadcast or as part of normal disconnect flow. This is the old `handleDisconnect` that used to be called internally only, but with the connection events it makes sense to expose it because the lifecycle of the peer is now also exposed to the application.

arnetheduck: the point is that when `send` fails, the connection should  be closed which subsequently leads to the controlling application taking action inside the disconnect handler - unless the responsibility / ownership of managing the resource (subscription in this case) is clear, miscounts will ensue - for example, this code unsubscribes without telling NBC so nbc thinks it's subscribed but gossip doesn't - this causes logic faults in one of the the two.

libp2p/stream/bufferstream.nim
arnetheduck: well, it should be well understood at this point whether the lock is locked or not - and whether it has purpose at all? "maybe" locked locks seems like reason to be worried.

dryajov: Yeah I agree, this wont happen if the log is acquired at the right time and it doesn't happen here or anywhere, it was by chance that I hit it at some point and decided that ti would be a good check overall, but now that I think about it, this check can actually be detrimental, because it would hide issues with lock acquisitions.

libp2p/protocols/pubsub/gossipsub.nim
dryajov: Well, the *stream/channel* gets closed not the *connection/socket*, and since this stream is internal to pubsub the caller application has no way of knowing that it failed, we could call `switch.disconnect()` instead and that would drop the peer entirely and trigger `unsubscribePeer` from the peer's lifecycle connection event, but that's too drastic, because this peer might still be useful for syncing or something else, either tho we got dropped on pubsub for some other reason. In short, since the stream is internal and the caller has no access to it, it has to be done here.

arnetheduck: if it's the responsibility of PubsubPeer to unsubscribe at the right moment, it has to be the same for subscribing, else the balance is lost - this PR needs to be followed up by something that restores this balance or miscounts will happen and nbc/libp2p will have different views of the subscriptions.

arnetheduck: another way to look at this is that what the application _maybe_ should be doing is registering peers to gossipsub - then, separately, gossipsub should manage the state of individual connections to the peer - in particular, it should not pull the rug from under the application's feet and unsubscribe the peer without telling it - this cannot work.

Issue #325, opened 2020-08-11
`Incomplete data received` messages during connection upgrade


dryajov:
@arnetheduck reported an issue with `Incomplete data received` disconnects during the negotiation phase. Bellow is the log from his report:

```log
DBG 2020-08-11 17:08:23.695+02:00 Connecting to discovered peer              topics="networking" tid=668396 file=eth2_network.nim:745 peer=16*n9WhDV
DBG 2020-08-11 17:08:23.966+02:00 upgrade failed                             topics="switch" tid=668396 file=switch.nim:350 exc="Incomplete data received" peer=16*n9WhDV
DBG 2020-08-11 17:08:25.176+02:00 upgrade failed                             topics="switch" tid=668396 file=switch.nim:350 exc="Stream EOF!" peer=16*7GiXeW
DBG 2020-08-11 17:08:25.586+02:00 upgrade failed                             topics="switch" tid=668396 file=switch.nim:350 exc="Incomplete data received" peer=16*KzHJS6
DBG 2020-08-11 17:08:25.586+02:00 Unable to establish connection with peer   topics="networking" tid=668396 file=eth2_network.nim:785 peer=16*KzHJS6 errMsg="Incomplete data received"
DBG 2020-08-11 17:08:25.587+02:00 Connecting to discovered peer              topics="networking" tid=668396 file=eth2_network.nim:745 peer=16*KKwmAw
```

I was able to troubleshoot it further and it appear the message is related to failed `noise` negotiations:

(this is for a different peer, the same peer appears to be working properly for me):

```json
{"lvl":"TRC","ts":"2020-08-11 10:52:22.664-06:00","msg":"handling connection","topics":"tcptransport","tid":800159,"file":"tcptransport.nim:65","address":"34.204.178.218:9000"}
{"lvl":"TRC","ts":"2020-08-11 10:52:22.664-06:00","msg":"stream created","topics":"lpstream","tid":800159,"file":"lpstream.nim:80","oid":"5f32ccc646547f05001b6132","name":"ChronosStream"}
{"lvl":"TRC","ts":"2020-08-11 10:52:22.664-06:00","msg":"timeout","topics":"connection","tid":800159,"file":"connection.nim:80","timeout":"600000"}
{"lvl":"TRC","ts":"2020-08-11 10:52:22.664-06:00","msg":"initiating handshake","topics":"multistream","tid":800159,"file":"multistream.nim:53","codec":"\u0019/multistream/1.0.0\n"}
[...]

PR #340, opened 2020-08-20
handle lpEOF right at multistream site, to lower amount of noise

sinkingsugar:
Handle explicitly one of the scenarios behind #325 (hopefully the only one...)


arnetheduck:
hm, does this make any difference though in the message that is printed in the logs, at the higher level?

PR #341, opened 2020-08-21
Mplex slow readers

dryajov:
This PR adds a protection mechanism to prevent slow readers from hogging the entire mplex read loop. 

This situation can happen when the `BufferStream`'s buffer has been topped and no reader is consuming data from it or it isn't being consumed fast enough. When this happens, the back pressure mechanism in `BufferStream` kicks in and `pushTo` will await until enough room has been made to finish pushing the remaining data, thus blocking the entire mplex read loop. Sadly, mplex doesn't have an elegant mitigation mechanism for this and the recommended approach is to reset the channel after a timeout. Quoting the spec here https://github.com/libp2p/specs/blob/master/mplex/README.md#implementation-notes:

```
If a stream is being actively written to, the reader must take care to keep up with inbound data. Due to the lack of back pressure at the protocol level, the implementation must handle slow readers by doing one or both of:

Blocking the entire connection until the offending stream is read.
Resetting the offending stream.
For example, the go-mplex implementation blocks for a short period of time and then resets the stream if necessary.
```

### Bonus:

This should also help identifying slow readers and/or deadlocks in the consuming application.

### DIFF
--- a/libp2p/multistream.nim
+++ b/libp2p/multistream.nim
@@ -120,8 +120,12 @@ proc list*(m: MultistreamSelect,
   result = list
 
 proc handle*(m: MultistreamSelect, conn: Connection, active: bool = false) {.async, gcsafe.} =
+  logScope:
+    oid = $conn.oid
+
   trace "handle: starting multistream handling", handshaked = active
   var handshaked = active
+
   try:
     while not conn.atEof:
       var ms = string.fromBytes(await conn.readLp(1024))
--- a/libp2p/muxers/mplex/lpchannel.nim
+++ b/libp2p/muxers/mplex/lpchannel.nim
@@ -43,6 +43,10 @@ logScope:
 # more formal approach.
 #
 
+const
+  # DefaultChannelSize* = 1024 * 1024 # 1MB
+  DefaultChannelSize* = 1024 # TODO: for debugging only
+
 type
   LPChannel* = ref object of BufferStream
     id*: uint64                   # channel id
@@ -238,7 +242,7 @@ proc init*(
   conn: Connection,
   initiator: bool,
   name: string = "",
-  size: int = DefaultBufferSize,
+  size: int = DefaultChannelSize,
   lazy: bool = false,
   timeout: Duration = DefaultChanTimeout): LPChannel =
 
--- a/libp2p/muxers/mplex/mplex.nim
+++ b/libp2p/muxers/mplex/mplex.nim
@@ -25,6 +25,7 @@ logScope:
 
 const
   MaxChannelCount = 200
+  DefaultPushTimeout = 1.seconds
 
 when defined(libp2p_expensive_metrics):
   declareGauge(libp2p_mplex_channels,
@@ -38,9 +39,11 @@ type
     currentId: uint64
     inChannTimeout: Duration
     outChannTimeout: Duration
+    pushTimeout: Duration
     isClosed: bool
     oid*: Oid
-    maxChannCount: int
+    maxChanCount: int
+    chanSize: int
 
 proc newTooManyChannels(): ref TooManyChannels =
   newException(TooManyChannels, "max allowed channel count exceeded")
@@ -80,7 +83,8 @@ proc newStreamInternal*(m: Mplex,
     initiator,
     name,
     lazy = lazy,
-    timeout = timeout)
+    timeout = timeout,
+    size = m.chanSize)
 
   result.peerInfo = m.connection.peerInfo
   result.observedAddr = m.connection.observedAddr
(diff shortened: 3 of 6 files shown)


arnetheduck:
Ok, so this PR gets it a little wrong - our problem in libp2p is not a "slow" reader - our problem is a _blocked_ reader - the reader is blocked because it's deadlocked trying to set up a send channel and waiting for data to arrive on this _new_ channel. 

Effectively, this PR will reset the existing, deadlocked, channel every time we receive a subscription list and the pubsubpeer send connection is being set up.

The slow reader protection is already implemented through the connection timeout: if nothing is happening on the channel, it will be reset. There's no need to implement one more such protection.

In PubSubPeer, the `send` operation depends on mplex processing data - the only way to keep processing data in mplex while setting up a send connection _from within the reader_ is to never ever block `PubSubPeer.send` - and the way to ensure that is to turn `PubSubPeer.send` into a non-async proc that returns `void` - ie no futures involved at all.

arnetheduck:
Notably, if the accumulated processing of the chronos event loop takes more than 1s, regardless if it's the one channel that's blocked, this PR might reset _all_ channels because it will think all channels are equally blocked - this is because the `await pushTo` operation only regains control after everything on the chronos loop has been processed - all other channels, all validations etc

sinkingsugar:
> In PubSubPeer, the `send` operation depends on mplex processing data - the only way to keep processing data in mplex while setting up a send connection _from within the reader_ is to never ever block `PubSubPeer.send` - and the way to ensure that is to turn `PubSubPeer.send` into a non-async proc that returns `void` - ie no futures involved at all.

You mean `asyncCheck send()`  basically?
Most of the sends should behave like this tbh, even in sync posix `send` the operation is not really much of a blocking operation, it is just an append to the kernels buffer kind of thing. Async makes it slower for no real reason when there is no need to wait.

arnetheduck:
> You mean asyncCheck send() basically?

well, I mean... `send(pubsubpeer, data)` - without `async`, without `asyncCheck` and without `await` - inside this `send`, all exceptions should be handled and whatever processing should be done should live in its own separate task, as if done with `asyncCheck`.

this can be applied more widely, but the pubsub peer send is problematic in particular because it relies on reading from mplex channels to progress - the other send operations don't have such blocking behaviour.

zah:
To be fair, offloading work to threads is often used to "solve bugs", namely user reports about unresponsive UI or observed high-latency in network settings. It's an open question whether a beacon node can be designed to operate with complete safety on a single thread (defined as not missing any of its duties). The solution will surely boil down to breaking computational work into smaller pieces that can be interleaved by the scheduler to provide enough timer accuracy. 

Ruling out threads as a possible solution to our problems is thus something as a bet (an act on belief) and this also introduces risk for the project, which is hard to compare to the technical risks of venturing into threading in Nim.

zah:
Otherwise, one way to terminate channels that hasn't been considered perhaps is by observing the accumulate unread buffer size, but that's close enough to the proposal here I guess. Having a long timeout seems a bit inappropriate in that situations though, because this blocks other readers. The networking stack of the OS is a multiplexer after all, so we can steal some lessons from there.

arnetheduck:
> Having a long timeout seems a bit inappropriate in that situations though, because this blocks other readers.

timeouts don't help slow readers - if they're doing CPU work, they'll be blocked regardless of timers. In particular, timeouts are _only_ needed when there are bugs in the readers - merely "slow" readers are usually doing useful work - it seems wrong to waste significant resources to work around buggy code here.

> The networking stack of the OS is a multiplexer after all,

TCP has built-in flow control though - the ACK message lets the other end know how much data has been processed etc - mplex doesn't .

dryajov:
> TCP has built-in flow control though - the ACK message lets the other end know how much data has been processed etc - mplex doesn't .

Backpressure *is flow control*, that's the entire point of this and many of the flows in libp2p.

arnetheduck:
> Backpressure is flow control, 

sure - but TCP allows the other party to adapt per-stream (per tcp connection) - this is the critical feature that mplex doesn't have: you can't "pause" a single channel and keep reading the others - which is fine since mplex data is handled by a single application - the OS has a more complex problem to solve.

Because it's a single application, it's also possible to reason about resource usage in mplex readers and not add the complexity of working around reader bugs by increasing the resource usage of mplex with more and more timers and timeouts and thus penalize correct applications

dryajov:
> sure - but TCP allows the other party to adapt per-stream (per tcp connection) - this is the critical feature that mplex doesn't have: you can't "pause" a single channel and keep reading the others - which is fine since mplex data is handled by a single application - the OS has a more complex problem to solve.

Yes, and in reality this isn't due to mplex, this is due to TCP itself, this is commonly known as `head-of-line blocking`. Any multiplexing/pipelining on top of TCP will run into this and there are limited options to handle this situations. At some point you would have to do exactly what's been done here, drop the packet/channel that is blocking the entire pipe or drop the entire pipe.

PR #342, opened 2020-08-21
avoid send deadlock by not allowing send to block

arnetheduck:
This is one example of how the deadlock in gossipsub can be avoided:

by making send a non-failing operation and dealing with send errors inside send. In particular, since `send` and `PubSubPeer` is responsible for setting up the send connection, it's also reasonable that it should be responsible for sending the initial subscription message to the other peer: this would likely resolve the problem that @zah was experiencing. To do that, `PubSub` needs to become aware that a new send connection has been opened over which the list of subscribed topics must be sent - for example through a callback or event.

It would then be natural that PubSubPeer also cleans up after itself - this would make gossipsub more independent in general and it wouldn't have to rely on connection events and NBC (or other application) to remember to call this function for every connection event (it's quite weird that NBC has to do this)

The test cases are failing because they are written under the assumption that bufferstream behaves like a datagram transport when in fact it is a stream whose readOnce may return fewer bytes than requested (here, we set the limit to 10 to stress it a bit).

This branch finalizes in NBC: https://github.com/status-im/nim-beacon-chain/tree/xxxx1

### DIFF
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -101,10 +101,9 @@ method rpcHandler*(f: FloodSub,
                   trace "exception in message handler", exc = exc.msg
 
         # forward the message to all peers interested in it
-        let published = await f.broadcast(
+        let published = f.broadcast(
           toSeq(toSendPeers),
-          RPCMsg(messages: m.messages),
-          DefaultSendTimeout)
+          RPCMsg(messages: m.messages))
 
         trace "forwared message to peers", peers = published
 
@@ -140,10 +139,9 @@ method publish*(f: FloodSub,
   let msg = Message.init(f.peerInfo, data, topic, f.msgSeqno, f.sign)
 
   # start the future but do not wait yet
-  let published = await f.broadcast(
+  let published = f.broadcast(
     toSeq(f.floodsub.getOrDefault(topic)),
-    RPCMsg(messages: @[msg]),
-    timeout)
+    RPCMsg(messages: @[msg]))
 
   when defined(libp2p_expensive_metrics):
     libp2p_pubsub_messages_published.inc(labelValues = [topic])
@@ -157,13 +155,13 @@ method unsubscribe*(f: FloodSub,
   await procCall PubSub(f).unsubscribe(topics)
 
   for p in f.peers.values:
-    discard await f.sendSubs(p, topics.mapIt(it.topic).deduplicate(), false)
+    f.sendSubs(p, topics.mapIt(it.topic).deduplicate(), false)
 
 method unsubscribeAll*(f: FloodSub, topic: string) {.async.} =
   await procCall PubSub(f).unsubscribeAll(topic)
 
   for p in f.peers.values:
-    discard await f.sendSubs(p, @[topic], false)
+    f.sendSubs(p, @[topic], false)
 
 method initPubSub*(f: FloodSub) =
   procCall PubSub(f).initPubSub()
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -63,49 +63,38 @@ method unsubscribePeer*(f: FloodSub, peer: PeerID) =
 
 method rpcHandler*(f: FloodSub,
                    peer: PubSubPeer,
-                   rpcMsgs: seq[RPCMsg]) {.async.} =
-  await procCall PubSub(f).rpcHandler(peer, rpcMsgs)
+                   rpcMsg: RPCMsg) {.async.} =
+  await procCall PubSub(f).rpcHandler(peer, rpcMsg)
 
-  for m in rpcMsgs:                                  # for all RPC messages
-    if m.messages.len > 0:                           # if there are any messages
-      var toSendPeers = initHashSet[PubSubPeer]()
-      for msg in m.messages:                         # for every message
-        let msgId = f.msgIdProvider(msg)
-        logScope: msgId
+  if rpcMsg.messages.len > 0:                      # if there are any messages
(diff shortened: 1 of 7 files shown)


libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -63,50 +63,38 @@ method unsubscribePeer*(f: FloodSub, peer: PeerID) =
...
-        let published = await f.broadcast(
-          toSeq(toSendPeers),
-          RPCMsg(messages: m.messages),
-          DefaultSendTimeout)
-
-        trace "forwared message to peers", peers = published
+                   rpcMsg: RPCMsg) {.async.} =
+  await procCall PubSub(f).rpcHandler(peer, rpcMsg)
+
+  if rpcMsg.messages.len > 0:                      # if there are any messages
```
dryajov, line 69: This assumes that there will only be one `RPCMsg` per transmission, but that's an assumption that is very easy to break by a client that sends multiple `RPCMsg` messages per transmission, which wouldn't be a direct protocol violation. 

I checked `decodeRpcMsg` and it will only decode one `RPCMsg` per call and ignores any other possible `RPCMsg` messages that could be read from the pipe. This seems wrong as well, and can lead to silent packet/message drops.

arnetheduck: shouldn't each of these be separately length-prefixed?

dryajov: Not necessarily, an implementation can choose to batch several messages and in that case we would only ever process the first `RPCMsg`.

arnetheduck: actually, that's not possible to encode: protobuf is a stream of key-value, there's no end-of-message marker so the length prefix is the only way to do it

dryajov: Each message in libp2p is itself length prefixed so it is possible to batch several protobuf packets in one single libp2p message.

arnetheduck: ok - that's the part I don't understand - [here](https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/pubsubpeer.nim#L88) we read one length-prefixed chunk of data and decode it as an rpc msg - based on that, `data` cannot contain more than one message because there's no additional delimiter?

arnetheduck: also, it looks like RPC messages are independent of each other - even if they were batched somehow, they should be processed one-by-one - I think it's easier to reason about the processing done here when it's structured this way - then any looping is done outside this function which clarifies the exception handling (which is hard to reason about otherwise, mid-loop)

arnetheduck:
merging, 32-bit fail is unrelated

libp2p/protocols/pubsub/floodsub.nim
dryajov: > ok - that's the part I don't understand - here we read one length-prefixed chunk of data and decode it as an rpc msg - based on that, data cannot contain more than one message because there's no additional delimiter?

Yes, we read the length prefixed data from the channel, but then we only read the first protobuf message from it, if the buffer contains more rpc messages, those would get ignored - I think this is a bug and should be handled. 

When we read and decode the first `RPCMsg`, we should additionally return the length that was consumed from the buffer by the `decodeRpcMsg` and if there are more bytes left, we should parse it again and so on - essentially, we should call `decodeRpcMsg` in a loop until the entire buffer has been consumed.

dryajov: >  I think it's easier to reason about the processing done here when it's structured this way

Yeah, this was just the place which hinted at this issue, but I agree that it should be kept as is here and the looping should be done at the `decodeRpcMsg` call site, or even inside it.

arnetheduck: > return the length that was consumed

that's the thing - protobuf always consumes all bytes - the wire encoding is just a stream of unordered key-values for the fields, there's no specific end

dryajov:
```                                                                                               
 +---------------------------------------------------------------------------------+                 
 |      +----------------------+ +----------------------+ +----------------------+ |                 
 |buf   |                      | |                      | |                      | |                 
 |len   |       RPCMsg 1       | |       RPCMsg 2       | |        RPCMsg 3      | |                 
 |      +----------------------+ +----------------------+ +----------------------+ |                 
 +---------------------------------------------------------------------------------+                 
 ```

Yes, but like in the diagram above, if the buffer contains several `RPCMsg`'s, `decodeRpcMsg` will only parse the first one, if the buffer contains more messages, we won't ever return them from it. This is what `decodeRpcMsg` does:

```nim
proc decodeRpcMsg*(msg: seq[byte]): ProtoResult[RPCMsg] {.inline.} =
  trace "decodeRpcMsg: decoding message", msg = msg.shortLog()
  var pb = initProtoBuffer(msg)
  var rpcMsg: RPCMsg
  rpcMsg.messages = ? pb.decodeMessages()
  rpcMsg.subscriptions = ? pb.decodeSubscriptions()
  rpcMsg.control = ? pb.decodeControl()
  ok(rpcMsg)
```

arnetheduck: ok, so how does it know that `RPCMsg 1` ends and `RPCMsg 2` begins? at which byte offset does that start? keep in mind that all of `decodeMessages`, `decodeSubscriptions` etc read _all_ data. see https://developers.google.com/protocol-buffers/docs/techniques#streaming

dryajov: OK, in that case we should be good, seems like it would not be possible to batch many `RPCMsg` messages without length prefixing, which means that `readLp` would read a message at a time.

Issue #337, opened 2020-08-18
Prevent Mplex-induced deadlocks


zah:
Consider the following order of events:

1) Incoming request for a new sub-protocol stream is received from a peer P.
2) The mplex read loop invokes a handler for the protocol. 
   The handler is started with `asyncCheck`, but further writes are awaited with `channel.pushTo`.
3) As part of the logic of the handler, new dials are performed and awaited (this blocks the read loop on the `pushTo` call)
4) The peer P sends us more data, opens new streams, etc. After filling a buffer, these are also blocked on the `channel.pushTo` op of the multiplexer.
5) Some of the blocked sends from step 4 may interfere with the dials started in step 2
6) The dials never complete and the read loop is never resumed resulting in a deadlock preventing all further communication with peer P.

dryajov:
>  2. The mplex read loop invokes a handler for the protocol.
> The handler is started with asyncCheck, but further writes are awaited with channel.pushTo.

This is done for backpressure, `pushTo` blocks because we need to stop accepting data once some limit is reached.

> 3. As part of the logic of the handler, new dials are performed and awaited (this blocks the read loop on the pushTo call)

> 4. The peer P sends us more data, opens new streams, etc. After filling a buffer, these are also blocked on the channel.pushTo op of the multiplexer.

After taking a look at the current implementation of `dial`, locks are released by the time the dial from within `pubsubpeer` is being called. Right now, this only happens on a connection event, here https://github.com/status-im/nim-libp2p/blob/master/libp2p/switch.nim#L362 and the lock is released here, https://github.com/status-im/nim-libp2p/blob/master/libp2p/switch.nim#L351. We need to trace that flow again, because it seems unlikely that it will lock up the way that is being described. All other invocations of `dial` should happen outside of the dial flow. For reference, the `dial` called from within the `PubSubPeer` send aren't locked and are simply pulling an existing connection from the connection manager here https://github.com/status-im/nim-libp2p/blob/master/libp2p/switch.nim#L386-L393.

arnetheduck:
>  For reference, the dial called from within the PubSubPeer send aren't locked and are simply pulling an existing connection from the connection manager here https://github.com/status-im/nim-libp2p/blob/master/libp2p/switch.nim#L386-L393.

[...]

PR #343, opened 2020-08-21
xxx-2

arnetheduck:
This shows another way in which the deadlock can be resolved: by queueing messages in send until the send connection is ready, thus avoiding the deadlock when the send connection is being set up.

It fails tests for the same reason as #342 but finalizes in NBC: https://github.com/status-im/nim-beacon-chain/tree/xxx-2

Just like in #342, there's more to do here.


arnetheduck:
closed in favour oif #337

PR #345, opened 2020-08-31
Remove asyncCheck from codebase.

cheatfate:


dryajov:
So, we're wrapping every callsite with a `try/except` instead of adding a new proc similar to `asynCheck` that does this for us? Seems unnecessary and error prone.

libp2p/connmanager.nim
```diff
@@ -184,8 +191,9 @@ proc storeConn*(c: ConnManager, conn: Connection) =
 
   c.conns[peerId].incl(conn)
 
-  # launch on close listener
-  asyncCheck c.onClose(conn)
+  # Launch on close listener
+  # All the errors are handled inside `onClose()` procedure.
+  discard c.onClose(conn)
```
zah, line 196: Why do we use `discard` instead of `traceAsyncErrors`?

1) there are no benefits to using discard
2) you lose information in trace logs
3) defects are ignored.

dryajov: `traceAsyncErrors` is probably the right thing to do here, I agree.

libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -104,11 +92,22 @@ proc resetMessage(s: LPChannel) {.async.} =
...
 
   ## send reset message - this will not raise
-  withEOFExceptions:
+  try:
     withWriteLock(s.writeLock):
       trace "sending reset message"
-
       await s.conn.writeMsg(s.id, s.resetCode) # write reset
+  except CancelledError:
+    # This procedure is called from one place and never awaited, so there no
```
arnetheduck, line 100: :+1: for writing down why it's ignored - this helps the next developer

libp2p/connmanager.nim
cheatfate: @zah @dryajov 
First of all `traceAsyncErrors()` do not handle errors, it just logs errors, so as soon as you start care about error handling - you need to replace `traceAsyncErrors()` with proper error handlers (that's what i'm actually doing). Proper error handlers should be placed in procedures which was previously started using `traceAsyncErrors()`. So when you have error handlers you don't need `traceAsyncErrors()` (Please be patient `Defect` story will be outlined later).



> There are no benefits to using discard

There number of issues with `traceAsyncErrors()` which is not present in `discard`:
    * Because `traceAsyncErrors()` using `Future[T]` completion callback it means that this callback get scheduled only after procedure was finished and execution returned back to `poll()` call. In such case state of the world when procedure finished can be seriously different from the state of the world when `traceAsyncErrors()` handler will be scheduled.
    *  `nim-libp2p` has layered structure and all these layers connected between each other. Such approach significantly complicates understanding of `traceAsyncError()` log, because you can only understand at what layer error happens, but you can't understand what `stream` object was used there.
    * `CancelledError` is not a fatal exception, but `traceAsyncError()` handles `CancelledError` as **fatal** error.
    * `discard` is of course `unsafe` keyword but every possible `discard` which introduced by this PR was commented and all the procedures which result was discarded was properly checked/adjusted to handle errors and cancellations. In such particular case `discard` is cheaper, because it do not create any callbacks.

#### traceAsyncError() could not handle `CancelledError`.

```nim
import chronos, chronicles/chronos_tools

proc testCancel() {.async.} =
  await sleepAsync(10.seconds)

proc traceCancellation() {.async.} =
  var fut = testCancel()
  traceAsyncErrors fut
  fut.cancel()
  await sleepAsync(1.seconds)

when isMainModule:
  waitFor traceCancellation()
```

Produces such output

```
FAT 2020-09-01 11:23:04.159+03:00 Fatal exception reached                    tid=25528 err="Future operation cancelled!" stackTrace="Traceback (most recent call last)\nC:\\Projects\\cancel.nim(13) cancel\nC:\\Projects\\nim-beacon-chain\\vendor\\nim-chronos\\chronos\\asyncloop.nim(1033) waitFor\nC:\\Projects\\nim-beacon-chain\\vendor\\nim-chronos\\chronos\\asyncloop.nim(343) poll\nC:\\Projects\\nim-beacon-chain\\vendor\\nim-chronicles\\chronicles\\chronos_tools.nim(15) :anonymous\nC:\\Projects\\nim-beacon-chain\\vendor\\nim-chronicles\\chronicles.nim(169) catchOrQuit\n"
```



> You lose information in trace logs


@zah sorry but `traceAsyncError()` also uses `TRACE` logs for all `CatchableError` https://github.com/status-im/nim-chronicles/blob/master/chronicles/chronos_tools.nim#L7 . And its how exceptions got handled in `nim-libp2p` i'm not changing anything, i have wrote error handling for small amount of procedures, all other places are already used `trace` logging for `CatchableErrors`:
https://github.com/status-im/nim-libp2p/pull/345/files#diff-e0b26345ee1b1e15968d513e9b2aad60L125-L126
https://github.com/status-im/nim-libp2p/pull/345/files#diff-54deac2904add92d29f284d9eb946cc7L81-L82
https://github.com/status-im/nim-libp2p/pull/345/files#diff-791568a09e61b8545da7c644adbe2f6dL185-L186
https://github.com/status-im/nim-libp2p/pull/345/files#diff-791568a09e61b8545da7c644adbe2f6dL111-L113
https://github.com/status-im/nim-libp2p/pull/345/files#diff-b01772aeb4c446cae741c7cfa950be7eL216-L218
https://github.com/status-im/nim-libp2p/pull/345/files#diff-b01772aeb4c446cae741c7cfa950be7eL74-L79



> Defects are ignored.


`traceAsyncError()` as well as `discard` could not handle and could not swallow `Defect` exceptions

```nim
import chronos, chronicles/chronos_tools

proc testDefect() {.async.} =
  var buffer: seq[byte]
  let ch = buffer[1] # This access should raise `Defect`.

proc silenceTest() {.async.} =
  discard testDefect()

proc traceAsyncTest() {.async.} =
  traceAsyncErrors testDefect()

when isMainModule:
  waitFor silenceTest()
  # waitFor traceAsyncTest()
```

Output of `silenceTest()` is
```
C:\Projects\defect.nim(14) defect
C:\Projects\defect.nim(8) silenceTest
C:\Projects\defect.nim(5) testDefect
C:\Projects\nim-beacon-chain\vendor\nimbus-build-system\vendor\Nim\lib\system\fatal.nim(49) sysFatal
Error: unhandled exception: index out of bounds, the container is empty [IndexError]
```

Output of `traceAsyncTest()` is
```
C:\Projects\defect.nim(14) defect
C:\Projects\defect.nim(11) traceAsyncTest
C:\Projects\defect.nim(5) testDefect
C:\Projects\nim-beacon-chain\vendor\nimbus-build-system\vendor\Nim\lib\system\fatal.nim(49) sysFatal
Error: unhandled exception: index out of bounds, the container is empty [IndexError]
```

arnetheduck: the last error is an example of how the change that tried to move `CancelledError` out of the recommended exception handling hierarchy breaks existing code - `tae` will not raise on `CancelledError` with latest chonos master where those changes have been rolled back - the same breakage exists elsewhere in the codebase where previously correct code would be broken by changing the inheritance structure.

stefantalpalaru: > the last error is an example of how the change that tried to move CancelledError out of the recommended exception handling hierarchy breaks existing code

Don't you mean the first? That's the only one dealing with cancellation.

> `tae` will not raise on CancelledError with latest chonos master where those changes have been rolled back

Of course it will (when it really shouldn't, because cancellation is not an error). The problem is that it will also exit the program:

```text
FAT 2020-09-01 11:58:30.394+02:00 Fatal exception reached                    tid=22052 file=chronos_tools.nim:9 err="Future operation cancelled!" stackTrace="Traceback (most recent call last, using override)\n/mnt/sde1/storage/nim-beacon-chain-clean/tmp/test4.nim(13) test4\n/mnt/sde1/storage/nim-beacon-chain-clean/vendor/nim-chronos/chronos/asyncloop.nim(1033) waitFor\n/mnt/sde1/storage/nim-beacon-chain-clean/vendor/nim-chronos/chronos/asyncloop.nim(344) poll\n/mnt/sde1/storage/nim-beacon-chain-clean/vendor/nim-chronicles/chronicles/chronos_tools.nim(15) colonanonymous\n/mnt/sde1/storage/nim-beacon-chain-clean/vendor/nim-chronicles/chronicles.nim(169) catchOrQuit\n/mnt/sde1/storage/nim-beacon-chain-clean/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(502) getStackTrace\n"
Error: execution of an external program failed: '/mnt/sde1/storage/nim-beacon-chain-clean/tmp/test4 '
```

Now let's see why - https://github.com/status-im/nim-chronicles/blob/b60f70718f8039c5c86dfc2a4680d8c1e37cbce2/chronicles/chronos_tools.nim#L5-L10 :

```nim
proc catchOrQuit*(error: Exception) =
  if error of CatchableError:
    trace "Async operation ended with a recoverable error", err = error.msg
  else:
    fatal "Fatal exception reached", err = error.msg, stackTrace = getStackTrace()
    quit 1

proc traceAsyncErrors*(fut: FutureBase) =
  fut.addCallback do (arg: pointer):
    if not fut.error.isNil:
      catchOrQuit fut.error[]
```

So there you have it: `catchOrQuit()` should have a separate case for `CancelledError` (and probably not print anything for it). Like this:

```diff
--- a/chronicles/chronos_tools.nim
+++ b/chronicles/chronos_tools.nim
@@ -5,6 +5,8 @@ import
 proc catchOrQuit*(error: Exception) =
   if error of CatchableError:
     trace "Async operation ended with a recoverable error", err = error.msg
+  elif error of CancelledError:
+    discard
   else:
     fatal "Fatal exception reached", err = error.msg, stackTrace = getStackTrace()
     quit 1
```

arnetheduck: that's quite odd, @stefantalpalaru  - `CancelledError` inherits from `FutureError` which inherits `CatchableError` - `of CatchableError` should therefore match it without the extra condition. perhaps test your example with https://github.com/status-im/nim-beacon-chain/pull/1598 ?

arnetheduck: @stefantalpalaru you can use the following snippet to better understand how the change to move `CancelledError` out of the normal inheritance tree breaks backwards compatibility with otherwise correct code:
```
import chronos, chronicles/chronos_tools

proc xx() {.async.} =
  await sleepAsync(20.seconds)

proc testDefect(): Future[void] =
  let fut = xx()
  fut.cancel()
  fut

proc traceTest() {.async.} =
  traceAsyncErrors testDefect()
  await sleepAsync(10.seconds)

when isMainModule:
  waitFor traceTest()
```

stefantalpalaru: > CancelledError inherits from FutureError which inherits CatchableError - of CatchableError should therefore match it without the extra condition.

Not in the Chronos commit you were criticising.

> perhaps test your example with status-im/nim-beacon-chain#1598 ?

No need. I already know how that one is broken in other spots.

> the change to move CancelledError out of the normal inheritance tree breaks backwards compatibility with otherwise correct code

Yet there was not a single report of beacon_node exiting with `Fatal exception reached   [...] err="Future operation cancelled!"` in the last month, was there?

And no, `quit(1)` triggered by `traceAsyncErrors()` is not "otherwise correct code".

zah: I agree that `traceAsyncErrors` should be expanded to handle `CancelledError` which will render moot some of the arguments against using it. In essence, the code right now includes an inlined and duplicated copy of the logic of TAE spread out through the library.

OTOH, the valid argument of what @cheatfate presented is that the timing of the trace log will be different.

arnetheduck: We have the case where we want to run a task without waiting for the result - this is the moral equivalent of launching a new thread to perform some work - the right place to handle errors is clearly not the event loop of chronos so they must be handled somewhere. 

`tae` says: I'm a calling a function and I believe I don't care about exceptions in that code - in practise, this is too weak because it's hard-to-impossible to analyze code for exceptions by hand - it is also not possible to maintain this guarantee as code changes - exceptions cause no error at compile time so even if the code didn't raise exceptions before, there's no way to tell what breaks when it starts doing so.

In short, it should thus be the case that any tasks that are launched asynchronously should simply not raise exceptions - they should do all their error handling "locally" and never leak anything to the event loop - just like a thread shouldn't raise exceptions from it's "main" procedure - nobody disputes this.

It follows that `tae` is something that we should really deprecate - it promotes the poor style of raising exceptions from stand-alone tasks and dropping them silently - instead, such tasks should be checked so that they don't raise exceptions (or return values for that matter).

arnetheduck:
> So, we're wrapping every callsite with a try/except

I think this helps provide context for the reader as to which exceptions are deliberately expected vs those that just randomly bubbled up - it also provides a place to document that deliberation as well as log an appropriately leveled message - once a few "common" patterns have been identified in the code, they can be refactored away into helpers - but for example, while a `StreamEOF` is expected in many places, other random exceptions are not, and this nuance should not be lost because it makes debugging difficult (for example by logging "connection refused" as "noise stream eof" and the like)

dryajov (approved):
LGTM

PR #346, opened 2020-09-01
refactor bufferstream to use a queue

arnetheduck:
This change modifies how the backpressure algorithm in bufferstream
works - in particular, instead of working byte-by-byte, it will now work
seq-by-seq.

When data arrives, it usually does so in packets - in the current
bufferstream, the packet is read then split into bytes which are fed one
by one to the bufferstream. On the reading side, the bytes are popped of
the bufferstream, again byte by byte, to satisfy `readOnce` requests -
this introduces a lot of synchronization traffic because the checks for
full buffer and for async event handling must be done for every byte.

In this PR, a queue of length 1 is used instead - this means there will
at most exist one "packet" in `pushTo`, one in the queue and one in the
slush buffer that is used to store incomplete reads.

* avoid byte-by-byte copy to buffer, with synchronization in-between
* reuse AsyncQueue synchronization logic instead of rolling own
* avoid writeHandler callback - implement `write` method instead
* simplify EOF signalling by only setting EOF flag in queue reader
* remove BufferStream pipes (unused)

### DIFF
--- a/libp2p/muxers/mplex/lpchannel.nim
+++ b/libp2p/muxers/mplex/lpchannel.nim
@@ -13,8 +13,7 @@ import types,
        coder,
        ../muxer,
        nimcrypto/utils,
-       ../../stream/connection,
-       ../../stream/bufferstream,
+       ../../stream/[bufferstream, connection, streamseq],
        ../../peerinfo
 
 export connection
@@ -51,10 +50,13 @@ type
     initiator*: bool              # initiated remotely or locally flag
     isLazy*: bool                 # is channel lazy
     isOpen*: bool                 # has channel been opened (only used with isLazy)
+    isReset*: bool                # channel was reset, pushTo should drop data
+    pushing*: bool
     closedLocal*: bool            # has channel been closed locally
     msgCode*: MessageType         # cached in/out message code
     closeCode*: MessageType       # cached in/out close code
     resetCode*: MessageType       # cached in/out reset code
+    writeLock: AsyncLock
 
 proc open*(s: LPChannel) {.async, gcsafe.}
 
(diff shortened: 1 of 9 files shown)


sinkingsugar:
image
I like this 👍

libp2p/stream/connection.nim
```diff
@@ -77,23 +77,21 @@ method initStream*(s: Connection) =
...
     s.timerTaskFut = s.timeoutMonitor()
+    if isNil(s.timeoutHandler):
+      s.timeoutHandler = proc(): Future[void] = s.close()
 
   inc getConnectionTracker().opened
 
 method close*(s: Connection) {.async.} =
   ## cleanup timers
-  if not isNil(s.timerTaskFut) and not s.timerTaskFut.finished:
+  if not isNil(s.timerTaskFut):
```
dryajov, line 94: canceling a finished future will throw.

libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -128,52 +126,73 @@ method closed*(s: LPChannel): bool =
...
-    raise exc
-  except CatchableError as exc:
-    trace "Exception in reset", exc = exc.msg, s
+  asyncSpawn s.resetMessage()
+
+  # This should wake up any readers by pushing an EOF marker at least
+  await procCall BufferStream(s).close() # noraises, nocancels
+
+  if s.pushing:
+    # When data is being pushed, there will be two items competing for the
```
dryajov, line 161: this sounds like it would reintroduce the issue that `drainBuffer` was trying to solve?

dryajov: meaning that, in some cases, data that could have been read will be lost?

dryajov (approved):
LGTM

libp2p/muxers/mplex/lpchannel.nim
arnetheduck: it's slightly different: `drainBuffer` solved the regular controlled "close" case where the shutdown of the channel was controlled (one node signals "closed" - the other end reads all data and also signals closed etc) - `reset` is more for the extraordinary cases where something is already wrong - at that point, it's hard to reason about lost data, but generally yes, on `reset`, data will be lost - this is why it's important to be careful and only use it in extraordinary situations (timeout, underlying connection dead etc)

dryajov: Yes, that's why it was originally not used in resets and only added after you pointed out an issue in NBC, but I wonder if this issue in particular actually points to something in NBC proper, rather than the reset itself? 

In any case, I agree `reset` is destructive and should only happen in cases where we don't know what the state of the underlying socket is.

arnetheduck: well, initially it was not used anywhere and that was the issue - the eof flag was set even when there was data to process and no draining would happen - this branch takes another approach: it puts EOF on the queue together with other data - this ensures it's processed it the correct order.

`reset` upsets the order, and indeed, data gets lost - but then again, `reset` is only available internally - NBC doesn't have access to it, so it can't do anything wrong even if it wanted to.

there are still issues lurking here: there are 3 flags: `isClosed`, `isEof` and `closedLocal` for 2 properties (write-eof and read-eof) causing overlap - `BufferedStream` in particular doesn't understand or know about `closedLocal` which causes confusion and potentially race conditions - to properly implement this, one would have to support half-closed starting in `LPStream` and work up through the inheritance structure.

arnetheduck: the setup before this branch was particularly vulnerable to issues because it set flags on the global state, then called await - this would open it up for reordering issues - this is why the new approach uses the queue for EOF, to avoid these kinds of races

dryajov: > there are still issues lurking here: there are 3 flags: isClosed, isEof and closedLocal for 2 properties (write-eof and read-eof) causing overlap - BufferedStream in particular doesn't understand or know about closedLocal which causes confusion and potentially race conditions - to properly implement this, one would have to support half-closed starting in LPStream and work up through the inheritance structure.

The flags `isClosed`, `isEof` and `closedLocal` are supposed to be private, to access the state of the stream, the `closed` and `atEof` properties should be used. This was intended to allow propagating state across inheritance boundaries and for greater extensibility.

We can of course move the half-closed functionality all the way down to `LPStream`.

PR #347, opened 2020-09-01
small cleanups & docs

arnetheduck:
* simplify gossipsub heartbeat start / stop
* avoid alloc in peerid check
* stop iterating over seq after unsubscribing item (could crash)
* don't crash on missing private key with enabled sigs (shouldn't happen
but...)


dryajov (approved):
LGTM

sinkingsugar:
After this is merged I think I can start integrating into 1.1, correct @arnetheduck ? as for now that is all you had for gossip iirc?

PR #348, opened 2020-09-02
gossipsub / floodsub fixes

arnetheduck:
* remove timed cache - the window shifting already removes old messages
* ref -> object
* avoid unnecessary allocations with `[]` operator

### DIFF
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -7,16 +7,17 @@
 ## This file may not be copied, modified, or distributed except according to
 ## those terms.
 
-import sequtils, tables, sets, strutils
+import std/[sequtils, sets, tables]
 import chronos, chronicles, metrics
-import pubsub,
-       pubsubpeer,
-       timedcache,
-       peertable,
-       rpc/[messages, message],
+import ./pubsub,
+       ./pubsubpeer,
+       ./timedcache,
+       ./peertable,
+       ./rpc/[message, messages],
        ../../stream/connection,
        ../../peerid,
-       ../../peerinfo
+       ../../peerinfo,
+       ../../utility
 
 logScope:
   topics = "floodsub"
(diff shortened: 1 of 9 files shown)


arnetheduck:
* floodsub, gossipsub: don't rebroadcast messages that fail validation
(!)
* floodsub, gossipsub: don't crash when unsubscribing from unknown
topics (!)
* gossipsub: don't send message to peers that are not interested in the
topic, when messages don't share topic list
* floodsub: don't repeat all messages for each message when
rebroadcasting
* floodsub: allow sending empty data
* floodsub: fix inefficient unsubscribe
* sync floodsub/gossipsub logging
* gossipsub: include incoming messages in mcache (!)
* gossipsub: don't rebroadcast already-seen messages (!)
* pubsubpeer: remove incoming/outgoing seen caches - these are already
handled in gossipsub, floodsub and will cause trouble when peers try to
resubscribe / regraft topics (because control messages will have same
digest)
* timedcache: reimplement without timers (fixes timer leaks and extreme
inefficiency due to per-message closures, futures etc)
* timedcache: ref -> obj

sinkingsugar:
LGTM, do you have more gossip changes in the pipeline? I would like to know cos I wanna merge in 1.1

arnetheduck:
@sinkingsugar #347 fixes some more things

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -66,35 +65,34 @@ method rpcHandler*(f: FloodSub,
...
-      # forward the message to all peers interested in it
-      f.broadcast(
-        toSeq(toSendPeers),
-        RPCMsg(messages: rpcMsg.messages))
+      await handleData(f, t, msg.data)
 
-      trace "forwared message to peers", peers = toSendPeers.len
+    # In theory, if topics are the same in all messages, we could batch - we'd
+    # also have to be careful to only include validated messages
+    f.broadcast(toSeq(toSendPeers), RPCMsg(messages: @[msg]))
```
dryajov, line 94: Yes, we should potentially batch here, but should be OK for now.

dryajov (approved):
LGTM

PR #351, opened 2020-09-04
fix connection event order when connection dies early

arnetheduck:
if the connection is already closed (because the remote closes during
identfiy for example), an exception would be raised which would leave
the connection in limbo, beacuse it would not go through the rest of
internalConnect.

Also, if the connection is already closed, the disconnect event would be
scheduled before the connect event :/

### DIFF
--- a/libp2p/protocols/pubsub/rpc/messages.nim
+++ b/libp2p/protocols/pubsub/rpc/messages.nim
@@ -86,7 +86,7 @@ func shortLog*(c: ControlMessage): auto =
 
 func shortLog*(msg: Message): auto =
   (
-    fromPeer: msg.fromPeer,
+    fromPeer: msg.fromPeer.pretty,
     data: msg.data.shortLog,
     seqno: msg.seqno.shortLog,
     topicIDs: $msg.topicIDs,
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -205,10 +205,6 @@ proc disconnect*(s: Switch, peerId: PeerID): Future[void] {.gcsafe.} =
   s.connManager.dropPeer(peerId)
 
 proc upgradeOutgoing(s: Switch, conn: Connection): Future[Connection] {.async, gcsafe.} =
-  logScope:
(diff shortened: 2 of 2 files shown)


dryajov:
If the exception is left to bubble up, then the try/except on https://github.com/status-im/nim-libp2p/blob/master/libp2p/switch.nim#L332 will be excecuted, which would call the `dialed.close()`, which would invoke `connMngr`'s `onClose`, which will cleanup both the muxer and the connection that were added on lines https://github.com/status-im/nim-libp2p/blob/master/libp2p/switch.nim#L190-L191. 

Essentially, this change makes identify failures non-fatal, which can be the right thing to do, but if identify failed due to the underlying connection failing, we're just delaying the cleanup flow.

arnetheduck:
> we're just delaying the cleanup flow.

by this time, the muxer read loop is already running, and it will likely see this issue first because it's trying to set the mplex channels up (for identify, and for the other messages the peer is sending) - close gets called from multiple places per the logs - I'm trying to debug  such a case where for some reason, connections aren't removed properly on certain orderings

the non-fatal identify, I'm not sure what the right thing would be actually - it feels like identify is run as just a random, standalone attempt to grab some more info - there's been discussion of removing support for it completely in eth2 because it can be used for nefarious purposes (fingerprinting known-buggy clients pre-attack)

dryajov:
> by this time, the muxer read loop is already running, and it will likely see this issue first because it's trying to set the mplex channels up (for identify, and for the other messages the peer is sending) - close gets called from multiple places per the logs - I'm trying to debug such a case where for some reason, connections aren't removed properly on certain orderings

Yes, that can certainly happen.

> the non-fatal identify, I'm not sure what the right thing would be actually - it feels like identify is run as just a random, standalone attempt to grab some more info - there's been discussion of removing support for it completely in eth2 because it can be used for nefarious purposes (fingerprinting known-buggy clients pre-attack)

I generally agree and we should probably ignore identify, we should also make it optional then and check if identify is registered at all, and only call it if it is.

PR #356, opened 2020-09-07
mplex fixes

arnetheduck:
* close the right connection when channel send fails
* don't crash on channel id that is not unique

### DIFF
--- a/libp2p/muxers/mplex/lpchannel.nim
+++ b/libp2p/muxers/mplex/lpchannel.nim
@@ -90,7 +90,7 @@ proc resetMessage(s: LPChannel) {.async.} =
   except CancelledError:
     # This procedure is called from one place and never awaited, so there no
     # need to re-raise CancelledError.
-    trace "Unexpected cancellation while resetting channel"
+    debug "Unexpected cancellation while resetting channel", s
   except LPStreamEOFError as exc:
     trace "muxed connection EOF", exc = exc.msg, s
   except LPStreamClosedError as exc:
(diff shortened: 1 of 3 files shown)


libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -217,14 +217,14 @@ proc init*(
...
-      trace "sending data", len = data.len, chann
+      trace "sending data", len = data.len, conn, chann
 
       await conn.writeMsg(chann.id,
                           chann.msgCode,
                           data)
     except CatchableError as exc:
       trace "exception in lpchannel write handler", exc = exc.msg, chann
-      await chann.reset()
+      asyncSpawn conn.close()
```
dryajov, line 227: `close()` internally already uses `asyncSpawn`.

arnetheduck: the significant fix here is that we close the connection (which was the source of the exception) and not the channel - if we don't close the connection here, it'll stay alive as a zombie.. I suspect it would make more sense to catch this exception in `conn.write` and close the connection from inside there - this way, it's handled "locally" but this was the minimal fix.

conn right now calls  chronos' `closeWait` after a few levels of indirection and abstraction

arnetheduck: ie this is one of the cause of connection leaks in general - after it happens, the zombie connection stays in a number of tables where it shouldn't be (like nbc and gossip) and keeps causing send fails, even though it should be dead.

arnetheduck: tbh I'm not sure why the `conn.write` fails - it doesn't make a lot of sense - it looks half-closed almost, the corresponding reader doesn't die. I suspect it's a bug in the bufferstream EOF handling that sets those flags from several places - https://github.com/status-im/nim-libp2p/pull/346 simplifies this a bit by placing the eof marker on the queue as well so it gets processed in the right order.

dryajov: > the significant fix here is that we close the connection

Ah, ok - then this makes sense. 

I wonder why write fails as well, we're probably writing to it when it has already closed but it might as well be a bug in bufferstream.

dryajov (approved):

PR #357, opened 2020-09-08
add peer lifecycle events

dryajov:


libp2p/switch.nim
```diff
@@ -75,32 +82,65 @@ type
       streamHandler*: StreamHandler
       secureManagers*: seq[Secure]
       dialLock: Table[PeerID, AsyncLock]
-      ConnEvents: Table[ConnEventKind, HashSet[ConnEventHandler]]
+      connEvents: Table[ConnEventKind, HashSet[ConnEventHandler]]
```
arnetheduck, line 85: do we want to change the hashset here to a seq instead? a bit more predictable in terms of iteration order

libp2p/switch.nim
```diff
@@ -564,11 +604,11 @@ proc newSwitch*(peerInfo: PeerInfo,
     secureManagers: @secureManagers,
   )
 
-  let s = result # can't capture result
-  result.streamHandler = proc(conn: Connection) {.async, gcsafe.} = # noraises
+  let switch = result # can't capture result
```
arnetheduck, line 607: why just not use result at all?

libp2p/switch.nim
```diff
@@ -577,11 +617,27 @@ proc newSwitch*(peerInfo: PeerInfo,
...
-  result.mount(identity)
+  switch.mount(identity)
   for key, val in muxers:
-    val.streamHandler = result.streamHandler
+    val.streamHandler = switch.streamHandler
     val.muxerHandler = proc(muxer: Muxer): Future[void] =
-      s.muxerHandler(muxer)
+      switch.muxerHandler(muxer)
+
+  proc peerEventHandler(peerId: PeerID, event: ConnEvent) {.async, gcsafe.} =
```
arnetheduck, line 626: `connEventHandler`?

libp2p/switch.nim
```diff
@@ -577,11 +617,27 @@ proc newSwitch*(peerInfo: PeerInfo,
...
+    ## peer event handler
+    ##
+
+    if peerId notin switch.connManager:
```
arnetheduck, line 630: technically, can get the `connCount` here to avoid the additional lookups in the table (once for `in` and once for `count`)

libp2p/switch.nim
```diff
@@ -577,11 +617,27 @@ proc newSwitch*(peerInfo: PeerInfo,
...
+      trace "peer left", peerId
+      doAssert(event.kind == ConnEventKind.Disconnected)
+      await switch.triggerPeerEvents(peerId, PeerEvent.Left)
+    elif switch.connManager.connCount(peerId) == 1:
+      trace "peer joined", peerId
+      doAssert(event.kind == ConnEventKind.Connected)
+      await switch.triggerPeerEvents(peerId, PeerEvent.Joined)
+
+  switch.addConnEventHandler(peerEventHandler, ConnEventKind.Connected)
```
arnetheduck, line 639: won't this result in a strange order of events? if I register both a conneventhandler and a peereventhandler, I'd expect the peereventhandler to trigger before the conneventhandler (so I can create peer-related structures in there)

arnetheduck: and the reverse for disconnected, ie I'd expect all connections to be closed before peer.left triggers

dryajov: yeah, I'm not sure if this approach is going to work, but I was trying to prevent duplicating the logic in the connection events. I'll see if the order is going to be possible to guarantee and if not I'll rework it.

dryajov: > won't this result in a strange order of events? if I register both a conneventhandler and a peereventhandler, I'd expect the peereventhandler to trigger before the conneventhandler (so I can create peer-related structures in there)

Actually, I was thinking that we don't know of the peer unless the connection has finished upgrading and if we trigger the peer event before the connection event, we sort of violating this assumption, so it made sense to me to trigger the peer event after the first and last conn event have triggered. Maybe my assumption is wrong and it's ok to have the peer events trigger before and after the conn events?

libp2p/switch.nim
dryajov: not sure what you mean here?

libp2p/switch.nim
dryajov: I wouldn't mind doing that, but it is implied that a `HashSet` preserves order and avoids duplicates. The reason to use a `HashSet`, is to prevent a handler from being registered more than once, we can do the same with a seq, but it would essentially mean duplicating the `HashSet` functionality.

libp2p/switch.nim
dryajov: OK, so after some more consideration, I think the right order would be to trigger the `Joined` peer events before the first conn event and the `Left` after the last conn event for the peer. 

So the order is:

- Peer Joined
- Connected a
- Connected b
- ...
- Disconnected a
- Disconnected b
- Peer Left

libp2p/switch.nim
arnetheduck: hashset is unordered though and might change iteration order of all items on every add

libp2p/switch.nim
```diff
@@ -75,33 +82,85 @@ type
...
+proc removePeerEventHandler*(s: Switch,
+                             handler: PeerEventHandler,
+                             kind: PeerEvent) =
+  s.peerEvents.withValue(kind, handlers) do:
+    handlers[].excl(handler)
+
+proc triggerPeerEvents(s: Switch,
+                       peerId: PeerID,
+                       event: PeerEvent) {.async, gcsafe.} =
+  logScope:
```
arnetheduck, line 135: fwiw, I  stopped using `logScope` because it doesn't always work - ie it'll add to some log messages but not to all

dryajov: huh, ok... thats a fun caveat. 😭

libp2p/switch.nim
dryajov: Ah, you're right, `HashSet` is unordered, luckily there is an `OrderedSet` in the sets module, we should use that. My bad.

libp2p/connmanager.nim
```diff
@@ -277,6 +281,7 @@ proc dropPeer*(c: ConnManager, peerId: PeerID) {.async.} =
 
   for conn in conns:
     await conn.close()
+  trace "Dropped peer", peerId
```
arnetheduck, line 284:
```suggestion
```

libp2p/switch.nim
```diff
@@ -75,33 +82,86 @@ type
...
+                             handler: PeerEventHandler,
+                             kind: PeerEvent) =
+  s.peerEvents.withValue(kind, handlers) do:
+    handlers[].excl(handler)
+
+proc triggerPeerEvents(s: Switch,
+                       peerId: PeerID,
+                       event: PeerEvent) {.async, gcsafe.} =
+
+  if s.peerEvents.len <= 0 or event notin s.peerEvents:
```
arnetheduck, line 136: the `len` check is kind of redundant, `notin` will do the same

PR #360, opened 2020-09-09
fix poor performance in LRU cache

arnetheduck:
it turns out (in NBC) a heap is sufficiently slow becuase of all the
deletes that it makes more sense to go with a linked list


dryajov (approved):
LGTM

PR #367, opened 2020-09-17
add full gc collection every teardown

sinkingsugar:
Welcome to the garbage collected reality


dryajov:
Interesting, this doesn't seem to be the issue after all, tests still fail in the same place.

sinkingsugar:
`SIGSEGV` is the GC issue

sinkingsugar:
Ah sorry, forcing full collect is just to make sure it crashes 100% 😄 and avoid catching it in the future after months...

sinkingsugar:
This branch is not a fix, just an improvement on how tests test things basically.

PR #372, opened 2020-09-22
allow to omit peerId and seqno

sinkingsugar:
related to https://github.com/ethereum/eth2.0-specs/issues/1981


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -1031,7 +1031,7 @@ method rpcHandler*(g: GossipSub,
 
     g.mcache.put(msgId, msg)
 
-    if g.verifySignature and not msg.verify(peer.peerId):
+    if not g.anonymize and g.verifySignature and not msg.verify(peer.peerId):
```
arnetheduck, line 1034: there are a few more combinations to cover here:

* when the message has a signature, it must be validated - we should never propagate a message with an invalid signature - that includes discarding messages with empty bytes and other signatures / keys that can't be parsed
* when the message has a seqno, it must be valid as well (ie exactly 8 bytes long)
* if `verifySignature` is turned on, the message _must_ have a signature (and implicitly, it must be valid)
* if `anonymize` is turned on, we could either be strict (don't allow messages with signatures) or lax (allow messages with signatures) - for now, the latter is enough - we can add the strict checking at some other time.

sinkingsugar: Alright makes sense.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -1031,11 +1031,21 @@ method rpcHandler*(g: GossipSub,
 
     g.mcache.put(msgId, msg)
 
-    if g.verifySignature and not msg.verify(peer.peerId):
+    if (msg.signature.len > 0 or g.verifySignature) and not msg.verify(peer.peerId):
```
arnetheduck, line 1034: `peer.peerId` is not really needed for verify, not sure why it's there..?

arnetheduck (approved):

libp2p/protocols/pubsub/gossipsub.nim
sinkingsugar: It's indeed not used

dryajov:
Seems like this is changing again as per https://github.com/libp2p/specs/pull/294#issuecomment-697022095. We should probably hold off on it for the time being, or perhaps adopt a lax approach?

arnetheduck:
yes, the lax mode is an intermediary step that is useful now

sinkingsugar:
This PR was not complete, I said I needed a couple more hours on it.
The message cache won't work as the IDs are not generated properly.
We use the same style as go btw 
https://github.com/libp2p/go-libp2p-pubsub/blob/f7f33e10cc18b4a20542d0208aeaccb91ad64f99/pubsub.go#L965
But this doesn't work.

sinkingsugar:
Since it's merged I will continue next week. As there is something weird going on with floodsub too.

stefantalpalaru:
> This PR was not complete

Add "[WIP]" at the start of the title, while your PR is still "work in progress".

arnetheduck:
right, forgot about that - need this for nbc to do something more reasnable soon - it's not affected at this point so.,

PR #373, opened 2020-09-22
move events to conn manager

dryajov:
Moving connection & peer events to connection manager, to prepare for upcoming connection management rework.

### DIFF
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -30,6 +30,8 @@ import stream/connection,
        peerid,
        errors
 
+export connmanager
+
 logScope:
   topics = "switch"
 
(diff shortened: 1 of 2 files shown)


arnetheduck (approved):

libp2p/switch.nim
```diff
@@ -412,16 +333,16 @@ proc internalConnect(s: Switch,
     # unworthy and disconnects it
     raise newLPStreamClosedError()
 
-  await s.triggerPeerEvents(peerId, PeerEvent.Joined)
-  await s.triggerConnEvent(
+  await s.connManager.triggerPeerEvents(peerId, PeerEvent.Joined)
```
arnetheduck, line 336: it should probably be connManager that triggers these eventually

dryajov: yep, thats the plan

PR #379, opened 2020-09-26
defaultMsgIdProvider alternative/test anonymize

sinkingsugar:
defaultMsgIdProvider fixes when anonymize is on
test anonymize
various bug fixing


sinkingsugar:
Have to fix last 2 gossip tests.. no real bug just the ID now is kinda global among peers so cache is aggressive

PR #384, opened 2020-09-30
Connection limits

dryajov:
This is part 1 of several more PRs that improves connection management and segregates upgrade flows per transport, which allow for more than one transport.

This PR adds:

- rework transports to provide an accept method
  - this allows for global connection limits
- Connection manager now tracks connections even if the peer id is not know
  - which prevents leaks in case the connection failed to upgrade
  - allows limiting in flight (non upgraded) connections
  - allows replacing a connection with another connection
    - this is needed during connection upgrades

### DIFF
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -12,8 +12,8 @@ import chronos, chronicles, metrics
 import peerinfo,
        stream/connection,
        muxers/muxer,
+       utils/semaphore,
        errors
-       utils/semaphore
 
 logScope:
   topics = "libp2p connmanager"
@@ -71,7 +71,6 @@ type
     # object itself make it susceptible to
     # copies and mangling by unrelated code.
     conns: seq[Connection]
-    muxed: Table[Connection, MuxerHolder]
     connEvents: Table[ConnEventKind, OrderedSet[ConnEventHandler]]
     peerEvents: Table[PeerEventKind, OrderedSet[PeerEventHandler]]
 
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -14,6 +14,7 @@ import peerinfo,
        muxers/muxer,
        utils/semaphore,
        errors
+       utils/semaphore
 
 logScope:
   topics = "libp2p connmanager"
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -14,7 +14,6 @@ import peerinfo,
        muxers/muxer,
        utils/semaphore,
        errors
-       utils/semaphore
 
 logScope:
   topics = "libp2p connmanager"
@@ -86,9 +85,6 @@ proc connCount*(c: ConnManager, peerId: PeerID): int =
     proc(conn: Connection): bool =
       (not isNil(conn.peerInfo) and conn.peerInfo.peerId == peerId)
   ).len
-
-proc connCount*(c: ConnManager, peerId: PeerID): int =
-  c.conns.getOrDefault(peerId).len
 
 proc addConnEventHandler*(c: ConnManager,
                           handler: ConnEventHandler,
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -74,7 +74,7 @@ type
     peerEvents: Table[PeerEventKind, OrderedSet[PeerEventHandler]]
 
 proc newTooManyConnectionsError(): ref TooManyConnectionsError {.inline.} =
-  result = newException(TooManyConnectionsError, "too many connections")
+  result = newException(TooManyConnectionsError, "Too many connections")
 
 proc init*(C: type ConnManager,
            maxConns: int = MaxConnections,
@@ -339,7 +339,7 @@ proc storeConn*(c: ConnManager, conn: Connection) =
 
   let peerId = conn.peerInfo.peerId
   if c.conns.getOrDefault(peerId).len > c.maxConns:
-    debug "too many connections",
+    debug "Too many connections",
       conn, conns = c.conns.getOrDefault(peerId).len
 
     raise newTooManyConnectionsError()
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -421,7 +421,7 @@ proc trackIncomingConn*(c: ConnManager,
     trace "Tracking incoming connection"
     await c.inConnSemaphore.acquire()
     return await c.trackConn(provider, Direction.In)
-  except NilConnectionError as exc:
+  except NilConnectionError as exc: # should not be propagated
     trace "Nil connection", exc = exc.msg
     c.inConnSemaphore.release()
   except CatchableError as exc:
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -78,7 +78,7 @@ type
     peerEvents: Table[PeerEventKind, OrderedSet[PeerEventHandler]]
 
 proc newTooManyConnections(): ref TooManyConnections {.inline.} =
-  result = newException(TooManyConnections, "too many connections for peer")
+  result = newException(TooManyConnections, "too many connections")
 
 proc init*(C: type ConnManager,
            maxConns: int = MaxConnections,
@@ -421,10 +421,10 @@ proc storeConn*(c: ConnManager, conn: Connection) =
 
 proc trackConn*(c: ConnManager,
                 provider: ConnProvider,
-                dir: Direction): Future[Connection] {.async.} =
+                dir: Direction):
+                Future[Connection] {.async.} =
   var conn: Connection
   try:
-    await c.connSemaphore.acquire()
     conn = await provider()
 
     if isNil(conn):
@@ -445,14 +445,28 @@ proc trackConn*(c: ConnManager,
     raise exc
 
 proc trackIncomingConn*(c: ConnManager,
-                        provider: ConnProvider): Future[Connection] =
+                        provider: ConnProvider):
+                        Future[Connection] {.async.} =
+  ## await for a connection slot before attempting
+  ## to call the connection provider
+  ##
+
   trace "Tracking incoming connection"
-  c.trackConn(provider, Direction.In)
+  await c.connSemaphore.acquire()
+  return await c.trackConn(provider, Direction.In)
 
 proc trackOutgoingConn*(c: ConnManager,
                         provider: ConnProvider): Future[Connection] =
+  ## try acquiring a connection if all slots
+  ## are already taken, raise TooManyConnections
+  ## exception
+  ##
+
   trace "Tracking outgoing connection"
-  c.trackConn(provider, Direction.Out)
+  if not c.connSemaphore.tryAcquire():
+    raise newTooManyConnections()
+
+  return c.trackConn(provider, Direction.Out)
 
 proc storeMuxer*(c: ConnManager,
                  muxer: Muxer,
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -26,7 +26,6 @@ const
 
 type
   TooManyConnectionsError* = object of CatchableError
-  NilConnectionError* = object of CatchableError
 
   ConnProvider* = proc(): Future[Connection] {.gcsafe, closure.}
 
@@ -78,9 +77,6 @@ type
 
 proc newTooManyConnectionsError(): ref TooManyConnectionsError {.inline.} =
   result = newException(TooManyConnectionsError, "Too many connections")
-
-proc newNilConnectionError(): ref NilConnectionError {.inline.} =
-  result = newException(NilConnectionError, "Can't track nil connection")
 
 proc init*(C: type ConnManager,
            maxIncoming = MaxConnections,
@@ -384,7 +380,7 @@ proc trackConn(c: ConnManager,
     conn = await provider()
 
     if isNil(conn):
-      raise newNilConnectionError()
+      return
 
     conn.dir = dir
     trace "Got connection", conn, dir = $dir
@@ -417,17 +413,20 @@ proc trackIncomingConn*(c: ConnManager,
   ## to call the connection provider
   ##
 
+  var conn: Connection
   try:
     trace "Tracking incoming connection"
     await c.inConnSemaphore.acquire()
-    return await c.trackConn(provider, Direction.In)
-  except NilConnectionError as exc: # should not be propagated
-    trace "Nil connection", exc = exc.msg
-    c.inConnSemaphore.release()
+    conn = await c.trackConn(provider, Direction.In)
+    if isNil(conn):
+      trace "Couldn't acquire connection, releasing semaphore slot", dir = $Direction.In
+      c.inConnSemaphore.release()
   except CatchableError as exc:
     trace "Exception tracking connection", exc = exc.msg
     c.inConnSemaphore.release()
     raise exc
+
+  return conn
 
 proc trackOutgoingConn*(c: ConnManager,
                         provider: ConnProvider):
@@ -445,12 +444,18 @@ proc trackOutgoingConn*(c: ConnManager,
                                             max = c.outConnSemaphore.size
     raise newTooManyConnectionsError()
 
+  var conn: Connection
   try:
-    return await c.trackConn(provider, Direction.Out)
+    conn = await c.trackConn(provider, Direction.Out)
+    if isNil(conn):
+      trace "Couldn't acquire connection, releasing semaphore slot", dir = $Direction.Out
+      c.outConnSemaphore.release()
   except CatchableError as exc:
     trace "Exception tracking connection", exc = exc.msg
     c.outConnSemaphore.release()
     raise exc
+
+  return conn
 
 proc storeOutgoing*(c: ConnManager, conn: Connection) =
   conn.dir = Direction.Out
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -26,6 +26,8 @@ const
 type
   TooManyConnections* = object of CatchableError
 
+  ConnProvider* = proc(): Future[Connection] {.gcsafe, closure.}
+
   ConnEventKind* {.pure.} = enum
     Connected,    # A connection was made and securely upgraded - there may be
                   # more than one concurrent connection thus more than one upgrade
@@ -62,29 +64,42 @@ type
     muxer: Muxer
     handle: Future[void]
 
+  ConnHolder = object
+    conn: Connection
+    handle: Future[void]
+
   ConnManager* = ref object of RootObj
     maxConns: int
+    maxPeerConns: int
     connSemaphore*: AsyncSemaphore
     muxed: Table[Connection, MuxerHolder]
-    # NOTE: don't change to PeerInfo here
-    # the reference semantics on the PeerInfo
-    # object itself make it susceptible to
-    # copies and mangling by unrelated code.
-    conns: seq[Connection]
+    conns: seq[ConnHolder]
     connEvents: Table[ConnEventKind, OrderedSet[ConnEventHandler]]
     peerEvents: Table[PeerEventKind, OrderedSet[PeerEventHandler]]
 
-proc init*(C: type ConnManager, maxConns: int = MaxConnections): ConnManager =
+proc newTooManyConnections(): ref TooManyConnections {.inline.} =
+  result = newException(TooManyConnections, "too many connections for peer")
+
+proc init*(C: type ConnManager,
+           maxConns: int = MaxConnections,
+           maxPeerConns: int = MaxConnectionsPerPeer): ConnManager =
   C(maxConns: maxConns,
+    maxPeerConns: maxPeerConns,
     muxed: initTable[Connection, MuxerHolder](),
     connSemaphore: AsyncSemaphore.init(maxConns))
 
 proc connCount*(c: ConnManager, peerId: PeerID): int =
   c.conns
-  .filter(
-    proc(conn: Connection): bool =
-      (not isNil(conn.peerInfo) and conn.peerInfo.peerId == peerId)
+  .filter((h) =>
+      (not isNil(h.conn.peerInfo) and
+      h.conn.peerInfo.peerId == peerId)
   ).len
+
+proc peerCount*(c: ConnManager): int =
+  c.conns
+  .filter((c) => not isNil(c.conn.peerInfo))
+  .deduplicate()
+  .len
 
 proc addConnEventHandler*(c: ConnManager,
                           handler: ConnEventHandler,
@@ -277,13 +292,18 @@ proc onClose(c: ConnManager, conn: Connection) {.async.} =
 
   try:
     await conn.join()
-    trace "Connection closed, cleaning up", conn
-    await c.cleanupConn(conn)
-    c.connSemaphore.release()
+    # don't put the join in the try
+    # to prevent the finally from
+    # firing on canceled futures
+    try:
+      trace "Connection closed, cleaning up", conn
+      await c.cleanupConn(conn)
+    finally:
+      c.connSemaphore.release()
   except CancelledError:
     # This is top-level procedure which will work as separate task, so it
     # do not need to propagate CancelledError.
-    debug "Unexpected cancellation in connection manager's cleanup", conn
+    debug "Canceling close event for connection", conn
   except CatchableError as exc:
     debug "Unexpected exception in connection manager's cleanup",
           errMsg = exc.msg, conn
@@ -298,14 +318,13 @@ proc selectConn*(c: ConnManager,
   ##
 
   let conns = c.conns
-    .filter(
-      proc(conn: Connection): bool =
-        (not isNil(conn.peerInfo) and conn.peerInfo.peerId == peerId)
-    )
-    .filterIt( it.dir == dir )
+    .filter((h) =>
+      (not isNil(h.conn.peerInfo) and
+        h.conn.peerInfo.peerId == peerId))
+    .filterIt( it.conn.dir == dir )
 
   if conns.len > 0:
-    return conns[0]
+    return conns[0].conn
 
 proc selectConn*(c: ConnManager, peerId: PeerID): Connection =
   ## Select a connection for the provided giving priority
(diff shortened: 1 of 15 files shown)


libp2p/transports/tcptransport.nim
```diff
@@ -62,78 +61,72 @@ proc setupTcpTransportTracker(): TcpTransportTracker =
...
                   client: StreamTransport,
                   initiator: bool): Connection =
-  trace "handling connection", address = $client.remoteAddress
-  let conn: Connection = Connection(ChronosStream.init(client))
+  debug "Handling tcp connection", address = $client.remoteAddress,
+                                   initiator = initiator,
+                                   clients = t.clients[initiator].len
+
+  let stream = ChronosStream.init(client,
+                                  dir = if initiator: Direction.Out
```
arnetheduck, line 69: `Direction` would be better off without `None` I think - `Option[Direction]` can be used in place of `None`, where relevant

libp2p/transports/tcptransport.nim
```diff
@@ -144,34 +137,65 @@ method close*(t: TcpTransport) {.async, gcsafe.} =
...
+    except CancelledError as exc:
+      raise exc
+    except CatchableError as exc:
+      let useExc {.used.} = exc
+      debug "listen: connection setup failed", errMsg = exc.msg
+      if not transp.closed:
+        await transp.closeWait()
+  except TransportTooManyError as exc:
+    let useExc {.used.} = exc
+    warn "listen: could not accept new client, too many files opened"
```
arnetheduck, line 162: return `nil` here?

libp2p/transports/tcptransport.nim
```diff
@@ -144,34 +137,65 @@ method close*(t: TcpTransport) {.async, gcsafe.} =
...
+  except TransportOsError as exc:
+    let useExc {.used.} = exc
+    error "listen: could not accept new client, got an error",
+          errMsg = exc.msg
+    break
```
arnetheduck, line 167: `break` of what?

libp2p/daemon/daemonapi.nim
```diff
@@ -1306,3 +1306,20 @@ proc pubsubSubscribe*(api: DaemonAPI, topic: string,
   except Exception as exc:
     await api.closeConnection(transp)
     raise exc
+
+proc `$`*(pinfo: PeerInfo): string =
```
arnetheduck, line 1310: hm, `$` is better off for roundtrippable stuff - for debug / logging output, better define a `shortLog` instead, else `$` gets picked up by serialization frameworks etc thinking they can use it

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -208,15 +208,15 @@ proc send*(p: PubSubPeer, msg: RPCMsg, anonymize: bool) =
   # or malicious data on the wire - in particular, re-encoding protects against
   # some forms of valid but redundantly encoded protobufs with unknown or
   # duplicated fields
+  var mm = msg # hooks can modify the message
```
arnetheduck, line 211: this was intentionally moved into the `if`, to avoid copying the message when it's not necessary (it's a waste, and mostly we don't h ave send observers)

arnetheduck: if there are no send observers, it would in fact be better to do the object -> protobuf conversion once for all peers - alternatively, don't give send observers per-peer powers

libp2p/connmanager.nim
```diff
@@ -53,28 +58,42 @@ type
...
-  c.conns.getOrDefault(peerId).len
+  c.conns
+  .filter((h) =>
+      (not isNil(h.conn.peerInfo) and
+      h.conn.peerInfo.peerId == peerId)
+  ).len
+
+proc peerCount*(c: ConnManager): int =
+  c.conns
+  .filter((c) => not isNil(c.conn.peerInfo))
```
arnetheduck, line 94: should unidentified peers ever be allowed inside the connection manager?

libp2p/connmanager.nim
```diff
@@ -53,28 +58,42 @@ type
...
+           maxConns: int = MaxConnections,
+           maxPeerConns: int = MaxConnectionsPerPeer): ConnManager =
+  C(maxConns: maxConns,
+    maxPeerConns: maxPeerConns,
+    muxed: initTable[Connection, MuxerHolder](),
+    connSemaphore: AsyncSemaphore.init(maxConns))
 
 proc connCount*(c: ConnManager, peerId: PeerID): int =
-  c.conns.getOrDefault(peerId).len
+  c.conns
```
arnetheduck, line 86: this becomes a second source of truth, in addition to the count in the semaphore - the code is prone to have mismatches between the two because they're not updated according to the same rules

libp2p/switch.nim
```diff
@@ -385,30 +400,22 @@ proc dial*(s: Switch,
...
   trace "Dialing (new)", peerId, protos
-  let conn = await s.internalConnect(peerId, addrs)
-  trace "Opening stream", conn
-  let stream = await s.connManager.getMuxedStream(conn)
-
-  proc cleanup() {.async.} =
+  var stream: Connection
+  try:
+    stream = await s.internalConnect(peerId, addrs)
+    if not(isNil(stream)):
```
arnetheduck, line 406: these `nil`s in returns from functions are pretty annoying

libp2p/connmanager.nim
```diff
@@ -53,28 +58,42 @@ type
     muxer: Muxer
     handle: Future[void]
 
+  ConnHolder = object
+    conn: Connection
+    evtHandle: Future[void]
```
arnetheduck, line 63: re names, I think what I intended here more precisely is that `handle` is/can a passive noun, but what's being held in the variable is an active ongoing task effectively, so an active noun/ verb would be more appropriate

libp2p/switch.nim
```diff
@@ -211,12 +213,13 @@ proc upgradeOutgoing(s: Switch, conn: Connection): Future[Connection] {.async, g
     raise newException(UpgradeFailedError,
       "No peerInfo for connection, stopping upgrade")
 
+  s.connManager.updateConn(conn, sconn)
```
arnetheduck, line 216: isn't there a risk here that the un-upgraded connection gets used in a separate call to connmanager, due to async reordering?

arnetheduck: ie why are connections that are not upgraded yet held in the same connection manager structure? it seems that they should only be allowed in there _after_ they've gone through upgrade?

libp2p/switch.nim
```diff
@@ -248,14 +253,19 @@ proc upgradeIncoming(s: Switch, conn: Connection) {.async, gcsafe.} =
...
-  if (await ms.select(conn)): # just handshake
-    # add the secure handlers
-    for k in s.secureManagers:
-      ms.addHandler(k.codec, securedHandler)
-
-  # handle un-secured connections
-  # we handshaked above, set this ms handler as active
-  await ms.handle(conn, active = true)
+  try:
+    if (await ms.select(incomingConn)): # just handshake
```
arnetheduck, line 257: this should really be refactored into a `handshake` named function (in a separate PR)

tests/testswitch.nim
```diff
@@ -209,6 +210,76 @@ suite "Switch":
...
+      await allFuturesThrowing(
+        switches
+        .mapIt( it.stop() )
+        .concat( @[lestener.stop()] )
+      )
+      await allFuturesThrowing(awaiters)
+
+    waitFor(testSwitch())
+
+  test "e2e should prevent more than max conns (dialer)":
```
arnetheduck, line 248: since cancelling is prominent in this PR, cancelling a blocked dial would really need a test here to ensure that the connmanager is properly cleaned up etc

libp2p/protocols/pubsub/pubsubpeer.nim
dryajov: it breaks metrics, I can rework it to not copy.

libp2p/connmanager.nim
dryajov: yes, that allows us to properly track their counts and dispose of their resources if they close prematurelly.

libp2p/connmanager.nim
dryajov: The  semaphore counts both identified and unidentified connections whereas `countCount` returns the number of identified connections only, ie connections for which we know the peer id. This prevents inflight (unsecured/unidentified/non-upgraded) connections from exploding. The reason to do this is to make `ConnManager` the single source of truth and prevent sharing the semaphore across the switch/transport  and `ConnManager`.

libp2p/connmanager.nim
dryajov: OK, my reasoning is that this is a handle for some opaque resource, in this case it is the handle for the close event and as such, unline with a read loop for example, there isn't an ongoing action associated with it. I'm fine changing it, but I struggle to come up with a more appropriate name - maybe `cancelHandle`?

libp2p/switch.nim
dryajov: > isn't there a risk here that the un-upgraded connection gets used in a separate call to connmanager, due to async reordering?

It shouldn't and if it does then that is a logical bug, not a concurrency one. If such a bug is introduced it can only be due to the update being called at the wrong place or for the wrong connection.

> ie why are connections that are not upgraded yet held in the same connection manager structure? it seems that they should only be allowed in there after they've gone through upgrade?

I've partially addressed this in https://github.com/status-im/nim-libp2p/pull/384#discussion_r498884362, but to reiterate, `ConnManager` is intended to be the singly source of truth in terms of managing connection related resource. By consolidating all connection tracking logic in one place, we're making it easier to maintain global counts, per peer counts and eventually, add heuristics as to prioritize connections, apply throttle, etc... 

To address the specific point of "unidentified" connections, if we don't track them in the connection manager and we still want to account for them (in-flight connection), that logic would end up being spread across different modules. Which is what I initially attempted to do, but quickly realized that I'm going to have to share the semaphore across those modules that will create or upgrade connections, such as the switch (for now and the transports themself in a further refactor).

Another point to keep in mind is that the upgrade logic is moving down to each individual transport and when that happens and transports like `websockets`, `webrtc` and in particular `quic` are introduced, they will all "produce" connections in different ways and at different stages, having a connection manager that is capable of working across all those different flows is very handy.

libp2p/switch.nim
dryajov: Yep, but this is getting heavily reworked in a different PR and I want to scope this changes as much as possible.

tests/testswitch.nim
dryajov: True, cancelation tests need to be added all the way, my worry is bloating this PR even more, it's already quite a large change as it is.

dryajov: That said, maybe the tests are crucial, I'll try to add tose.

libp2p/transports/tcptransport.nim
dryajov: good catch, this used to be a while loop

libp2p/transports/tcptransport.nim
dryajov: Can be, to make it more explicit, but this will fall through and return `nil` either way for all code paths except the`try` block itself.

libp2p/connmanager.nim
dryajov: err, `onCloseHandle`?

libp2p/transports/tcptransport.nim
arnetheduck: it doesn't look intentional this way, ie more like a forgotten branch - returning nil vs exceptions remains confusing

libp2p/protocols/pubsub/pubsubpeer.nim
arnetheduck: yes, this is a hotspot in gossipsub

tests/testswitch.nim
arnetheduck: yes, I would say they are, this branch significantly depends on cancelling things

libp2p/switch.nim
arnetheduck: What I don't understand is the following:

1. `conn` is added to connectionmanager by task A - for example through a dial where the peerid was known
2. `conn` starts going through upgrade process which involves several negotiation steps, so task A hands over execution to other tasks
3. task B wants a connection to the same peer, and finds  it in connmanager - notably, before upgrade has finished.

what prevents the above scenario?

dryajov: > What I don't understand is the following:
> 
> 1. `conn` is added to connectionmanager by task A - for example through a dial where the peerid was known
> 2. `conn` starts going through upgrade process which involves several negotiation steps, so task A hands over execution to other tasks
> 3. task B wants a connection to the same peer, and finds  it in connmanager - notably, before upgrade has finished.
> 
> what prevents `conn` from being given to task B the above scenario?

The important thing to keep in mind is that I haven't really changed the existing flow all that much and things still behave very similarily to what they did prior to this changes.

- For dialing (outgoing), we hold a lock in the dial itself that prevents other dials to happen concurrently until the upgrade has finished - https://github.com/status-im/nim-libp2p/blob/tcp-limits2/libp2p/switch.nim#L278-L280.
  - I've removed the usage of `selectConn ` and instead we call [`getStream(peerId)`](https://github.com/status-im/nim-libp2p/blob/tcp-limits2/libp2p/switch.nim#L283), which won't return a stream unless the connection has upgraded (in the case of TCP, up until it got a muxer attached). This prevents raise conditions related to working directly with the connection in the Switch and instead moves that flow into the `ConnManager`
  - The requirement to keep only one outgoing connection is still in place, as we only allow to dial when we know the `peerId` (this might change in the future, but this is how it works right now).

- For incoming connections, we allow several concurrent connections to exist at the same time. However, the same logic as with the dial flow applies, `getStream` won't return anything until a muxer has been attached, which will only happen in [`muxerHandler`](https://github.com/status-im/nim-libp2p/blob/tcp-limits2/libp2p/switch.nim#L503). If a dial is initiated, it will either get a stream from an existing incoming connection, or attempt to establish a new one according to the dial flow.

This changes try hard to not modify the existing flow, rather they build on top of what's there already. This is by design and it will evolve a bit more as I rework the upgrade flow further, but right now I want to scope things as much as possible.

libp2p/connmanager.nim
arnetheduck: ah right, I think I confused it with other places where `handle` is used for the read loop - but even in this case, there is an "activity" associated with the handle - an ongoing `onClose` call which is stuck in `join` - and in fact, this is a leak as well: when the connection is removed from `conns`, `onClose` is not cancelled - this means that the future and everything it owns will never get released

arnetheduck: ping, leak?

dryajov: Well, the original flow hasn't changed, everywhere the connection is being removed there is a `conn.close()` in the vicinity (https://github.com/status-im/nim-libp2p/blob/master/libp2p/connmanager.nim#L217-L223, https://github.com/status-im/nim-libp2p/blob/master/libp2p/connmanager.nim#L374-L387), so that flow is the same, so it doesn't look like it would leak ATM.

dryajov:
@arnetheduck @cheatfate @sinkingsugar can I get some eyes on this? Also, it would be great if you could give this a try before merging to master as described in - https://github.com/status-im/nimbus-eth2/pull/1932.

arnetheduck:
what's left in this branch?

dryajov:
The core functionality for connection limiting is here, but other things such as the transports rework have been already merged. I'll rebase and get this one ready again.

dryajov:
@sinkingsugar @arnetheduck I'm planning on merging this in the next few days, please give it one last review before I do so.

sinkingsugar:
Sure! Btw can we have a devel branch up btw, master sporadically fails already and I'm getting worried for nbc stability.
I'd rather push on devel such changes personally.

dryajov:
This change needs to go into master, otherwise I can't start the required
redactors that would go into devel. So this has to be in master.

libp2p/connmanager.nim
```diff
@@ -331,35 +345,120 @@ proc storeConn*(c: ConnManager, conn: Connection) =
...
+  return conn
+
+proc trackIncomingConn*(c: ConnManager,
+                        provider: ConnProvider):
+                        Future[Connection] {.async.} =
+  ## await for a connection slot before attempting
+  ## to call the connection provider
+  ##
+
+  var conn: Connection
```
sinkingsugar, line 418: why not returning inside the try body?

libp2p/connmanager.nim
```diff
@@ -331,35 +345,120 @@ proc storeConn*(c: ConnManager, conn: Connection) =
...
+
+  trace "Tracking outgoing connection", count = c.outConnSemaphore.count,
+                                        max = c.outConnSemaphore.size
+
+  if not c.outConnSemaphore.tryAcquire():
+    trace "Too many outgoing connections!", count = c.outConnSemaphore.count,
+                                            max = c.outConnSemaphore.size
+    raise newTooManyConnectionsError()
+
+  var conn: Connection
```
sinkingsugar, line 449: return inside the try body, flow is predictable and will be safer

libp2p/connmanager.nim
```diff
@@ -62,24 +66,27 @@ type
     handle: Future[void]
 
   ConnManager* = ref object of RootObj
-    maxConns: int
-    # NOTE: don't change to PeerInfo here
-    # the reference semantics on the PeerInfo
-    # object itself make it susceptible to
-    # copies and mangling by unrelated code.
+    maxPeerConns: int
```
arnetheduck, line 69: this should be `maxConnsPerPeer` else it reads like a global limit on peers connected (regardless of direction)

libp2p/connmanager.nim
```diff
@@ -331,35 +345,120 @@ proc storeConn*(c: ConnManager, conn: Connection) =
...
+  try:
+    trace "Tracking incoming connection"
+    await c.inConnSemaphore.acquire()
+    conn = await c.trackConn(provider, Direction.In)
+    if isNil(conn):
+      trace "Couldn't acquire connection, releasing semaphore slot", dir = $Direction.In
+      c.inConnSemaphore.release()
+  except CatchableError as exc:
+    trace "Exception tracking connection", exc = exc.msg
+    c.inConnSemaphore.release()
```
arnetheduck, line 428: the release should not happen if an exception was raised before the semaphore was acquired - for example if `acquire` was cancelled before being acquired

libp2p/connmanager.nim
```diff
@@ -331,35 +345,120 @@ proc storeConn*(c: ConnManager, conn: Connection) =
...
+                        Future[Connection] {.async.} =
+  ## try acquiring a connection if all slots
+  ## are already taken, raise TooManyConnectionsError
+  ## exception
+  ##
+
+  trace "Tracking outgoing connection", count = c.outConnSemaphore.count,
+                                        max = c.outConnSemaphore.size
+
+  if not c.outConnSemaphore.tryAcquire():
```
arnetheduck, line 444: this is a bit odd, as far as limits go: I would typically want to limit the number of total connections and the number of incoming connections, but not the number of outgoing connections.. ie if I want to be connected to 100 peers and manage to make outgoing connections to all of them, there's no need for me to allow incoming connections, but I don't want to wait for incoming connections and _not_ connect to outgoing peers in the meantime.

dryajov: You also want to be able to control the number of outgoing connections, this is a pretty straightforward use case. It can be done in two ways, let the calling code worry about it or allow the library handle this internally. In this case, once connection and peer management becomes more central in libp2p, you want to have a complete view of how both of this resources are being used.

One common use case are bootstrap nodes. Bootstrap nodes have a preference for incoming connections, but might still have outgoing connections to other bootstrap nodes. In this case it makes sense to limit the amount of outgoing connections, you probably want to set 1000 incoming vs 50 outgoing, or something along this lines.

arnetheduck: but still - there's no way to implement a total peer limit with this setup: if I want 100 peers no matter the direction, the only way to do it becomes limiting both incoming and outgoing to 100, then disconnecting anything outside of that

dryajov: OK, that is a valid point. The implementation assumes limits on both directions, there needs to be a way to have a global limit as well.

For background on this PR. It was never meant to be complete solution to all potential use cases, rather a stopgap that adds the ability to control the amount of incoming outgoing connections. A complete connection management, with proper bidirectional limiting is more involved and nuanced to implement than this simple first pass. 

I would say we have to paths forward with this 

- If connection limiting is something that NBC requires right now - eg, to avoid doing constant disconnects, then I would use this for the time being. With this implementation, we can at least tune the incoming vs outgoing connections in a way that works for now. For example, given that we get a lot less outgoing connections, we could set the proportions to something like 80 incoming vs 20 outgoing.
- If NBC can continue with its current approach of disconnecting peers when it reaches the upper limit, then this PR isn't needed.

I personally think that NBC will benefit from this PR by avoiding the constants disconnects, which add quite a bit of internal churn.

libp2p/connmanager.nim
```diff
@@ -331,35 +345,120 @@ proc storeConn*(c: ConnManager, conn: Connection) =
...
+      c.outConnSemaphore.release()
+  except CatchableError as exc:
+    trace "Exception tracking connection", exc = exc.msg
+    c.outConnSemaphore.release()
+    raise exc
+
+  return conn
+
 proc storeOutgoing*(c: ConnManager, conn: Connection) =
   conn.dir = Direction.Out
```
arnetheduck, line 463: isn't the connection direction already set correctly when the connection instance is created? it would be better if it was immutable from that point onwards

dryajov: Yes, you have a point. It's probably a good idea to add an assert so that only the correct direction is added.

libp2p/connmanager.nim
```diff
@@ -331,35 +345,120 @@ proc storeConn*(c: ConnManager, conn: Connection) =
...
+  if conn.closed or conn.atEof:
+    raise newException(CatchableError, "Connection closed or EOF")
 
   if isNil(conn.peerInfo):
-    raise newException(CatchableError, "empty peer info")
+    raise newException(CatchableError, "Empty peer info")
+
+  # Launch on close listener
+  # All the errors are handled inside `onClose()` procedure.
+  asyncSpawn c.onClose(conn)
```
arnetheduck, line 358: should `onClose` run even if `c.maxPeerConns` is violated? if yes, why?

arnetheduck: also, why should it _not_ be run for connections without `peerInfo`?

dryajov: So, we don't explicitly disconnect the peer there, we only throw an exception and allow the caller to decide what needs to happen to the peer, so from that standpoint the connection, nor the peer is immediately dropped. Maybe we should drop the peer here explicitly?

libp2p/connmanager.nim
arnetheduck: yeah, I'm not sure - we simply cannot implement the nbc `maxPeers` option correctly as-is - I suspect the right thing to do for NBC is to continue disconnecting as it does now until total connection limiting is implemented in libp2p, to not impose unnecessary churn on users currently using that option.

dryajov: > not impose unnecessary churn on users currently using that option.

What would the churn for the user be?

libp2p/standard_setup.nim
```diff
@@ -27,7 +27,11 @@ proc newStandardSwitch*(privKey = none(PrivateKey),
                         transportFlags: set[ServerFlags] = {},
                         rng = newRng(),
                         inTimeout: Duration = 5.minutes,
-                        outTimeout: Duration = 5.minutes): Switch =
+                        outTimeout: Duration = 5.minutes,
+                        maxConnections = MaxConnections,
```
arnetheduck, line 31: regarding backwards compatibility, this for example would require a major version bump since it's an incompatible behavioural change - a backwards-compatible way would be to default to unlimited, but this perhaps isn't a good "default"

dryajov: > major version bump since it's an incompatible behavioural change

Sure, that's why we use versions in the first place, to signal changes, not how good we feel about it :)

libp2p/connmanager.nim
arnetheduck: well, "maxPeers" means total peers in nbc - from a practical point of view, one gets to about 30-40 with outgoing connections only after a day so to get to 160, one would have to set maxPeers to something much higher to account for the new way of measuring and above all, blocking incoming connections

libp2p/connmanager.nim
arnetheduck: the question here is what `onClose` does / should do, and to which connections? to those that were stored in the conn manager? to some connections randomly?

libp2p/connmanager.nim
dryajov: Isn't this a heuristic more than the implementation limitation? 

There are two issues at play:

1) NBC has a max number of peers, which is right now implement by disconnecting peers. This is suboptimal, because it requires performing a full connection upgrade and then tearing it down right away - this repeats continuously and it's inefficient.
2) We're getting too little outgoing connections (I suspect this is due to misconfigured NATs), which is an issue with the Eth2 network, not necessarily connection management issue. We have heuristics as to what the proportion of the incoming vs outgoing connection is, it would be up to NBC to pass those ratios correctly to libp2p.

Another approach would be to expose in/out limits as arguments in NBC? (but maybe this is the overhead you want to avoid for the user?)

libp2p/connmanager.nim
dryajov: `onClose` cleans up after the connection is closed, it works on the connection that is being passed to it. But it doesn't trigger until the connection has been added to connection manager, so it won't trigger in the case of `TooManyConnections` because it was never added to the manager in the first place.

arnetheduck: but that's a leak then: the task gets spawned before the peer connection check

dryajov: Ah, interesting. Seems like I changed it in this PR and this would indeed leak, good catch.

libp2p/connmanager.nim
dryajov: > but still - there's no way to implement a total peer limit with this setup: if I want 100 peers no matter the direction, the only way to do it becomes limiting both incoming and outgoing to 100, then disconnecting anything outside of that

So, actually there is a way of implementing both models fairly simple. Right now I'm using two semaphores one for incoming another for outgoing connections, the simplest way of implementing this would be to just share one semaphore in the case a total count is prefered - that way we can have our cake and eat it too :-)

arnetheduck: yeah, that would be nice, then we can do it properly in nbc as well

libp2p/connmanager.nim
arnetheduck: or just have a `storeconnection`  which reads `dir` from `conn`

dryajov: I think the assert is safer, in case we inadvertently set the wrong direction somewhere?

libp2p/connmanager.nim
dryajov: Yes, this needs to be fixed once the semaphore cancellations are merged, without it, unless explicitly releasing on cancelation this branch will break.

libp2p/connmanager.nim
arnetheduck: well, it says the same thing in a different way - if `dir` can't be relied upon, there will be lots of other issues as well - fewer store functions seems more simple, and whoever is calling storeoutgoing right now can potentially do the assert, if it makes sense in that context - generally, these kinds of issues appear usually with mutable data.

libp2p/connmanager.nim
arnetheduck: with shared semaphore we can do `maxPeers` correctly, but we'll eventually want to have both a global and a directional limit I suspect

dryajov: How is the global limit different from the total directional limit?

arnetheduck: well, the use case was to set a global limit and a sublimit for either direction (for example max 100 peers of which no more than 80 can be incoming / outgoing)

dryajov: Hmm, ok - but is this indispensable right now? There are numerous sophisticated limits we can think of, but I wonder if this subset will do for now?

As for adding a global limit, we probably need another semaphore.

libp2p/connmanager.nim
dryajov: I'm just trying to think of a convenient API here, this aren't anything but helpers and they are meant to do just that, removing them will not significantly affect anything. The assert is there just to make sure that the direction hasn't been changed inadvertently, which can also be accomplished by making it read-only.

dryajov:
Anything else left here that can't be addressed as a separate PR @arnetheduck?

tests/testnative.nim
```diff
@@ -25,3 +25,5 @@ import testtransport,
        testnoise,
        testpeerinfo,
        testmplex
+
+include ../libp2p/utils/semaphore
```
arnetheduck, line 29: include?

libp2p/connmanager.nim
```diff
@@ -331,42 +355,119 @@ proc storeConn*(c: ConnManager, conn: Connection) =
...
+               dir: Direction):
+               Future[Connection] {.async.} =
+  var conn: Connection
+  try:
+    conn = await provider()
+
+    if isNil(conn):
+      return
+
+    conn.dir = dir
```
arnetheduck, line 397: doesn't `provider` set this? or does the provider return garbage in the field?

libp2p/connmanager.nim
```diff
@@ -331,42 +355,119 @@ proc storeConn*(c: ConnManager, conn: Connection) =
...
 
   trace "Stored connection",
     conn, direction = $conn.dir, connections = c.conns.len
 
-proc storeOutgoing*(c: ConnManager, conn: Connection) =
-  conn.dir = Direction.Out
-  c.storeConn(conn)
+proc trackConn(c: ConnManager,
+               provider: ConnProvider,
+               dir: Direction):
```
arnetheduck, line 388: instead of passing in the direction, pass in the semaphore that is to be released?

libp2p/connmanager.nim
```diff
@@ -331,42 +355,119 @@ proc storeConn*(c: ConnManager, conn: Connection) =
...
+      except CatchableError as exc:
+        trace "Exception in semaphore monitor, ignoring", exc = exc.msg
+
+      if conn.dir == Direction.In:
+        c.inSema.release()
+      else:
+        c.outSema.release()
+
+    asyncSpawn semaphoreMonitor()
+  except CatchableError as exc:
```
arnetheduck, line 412: this is to catch exceptions in.. `provider()`? it's the only thing that potentially raises here, so it shouldn't be needed

libp2p/switch.nim
```diff
@@ -491,6 +516,9 @@ proc accept(s: Switch, transport: Transport) {.async.} = # noraises
         await conn.close()
       return
 
+  if not isNil(conn):
```
arnetheduck, line 519: this closes the... last connection that was opened?

tests/testnative.nim
dryajov: done

arnetheduck (approved):

libp2p/connmanager.nim
dryajov: Provider is a user define method, so it could be anything.

libp2p/connmanager.nim
dryajov: Since `provider` is user defined and we don't have a good way of annotating async methods with allowed exception types, this is to make sure both flows are covered.

libp2p/connmanager.nim
dryajov: Yeah, not a bad idea actually.

libp2p/connmanager.nim
arnetheduck: the point was more that neither `asyncSpawn` nor the monitor raises, so reraising the exception from `provider` makes the exception handler a no-op because `conn` will never be anything but `nil`

libp2p/connmanager.nim
```diff
@@ -199,7 +218,10 @@ proc closeMuxerHolder(muxerHolder: MuxerHolder) {.async.} =
 
   await muxerHolder.muxer.close()
   if not(isNil(muxerHolder.handle)):
-    await muxerHolder.handle # TODO noraises?
+    try:
+      await muxerHolder.handle # TODO noraises?
+    except CatchableError as exc:
+      trace "Exception in close muxer handler", exc = exc.msg
```
arnetheduck, line 224: this should be a warning at least - since the convention is to _not_ raise, it should never happen that it raises

PR #385, opened 2020-10-01
gossip pruning fixes related to #371

sinkingsugar:


arnetheduck:
it would be good to start adding tests for these scenarios - ie gossip connects to a peer, adds it to the mesh,  thus messages are sent, then the peer disconnects and it's removed from mesh - all this could be done deterministically with a testbufferstream that verifies that the messages are correctly processed and that they are broadcast as expected, then _not_ broadcast afterwards etc (ping @dryajov )

dryajov:
Yeah, I agree for most things we can use a bufferstream and write deterministic tests. A recent great addition by @arnetheduck is `getConn` https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/pubsub.nim#L140-L141, which can be used to test with mocked connections by way of bufferstream. We might need to further expose it on the pubsub interface to make use of it in tests however.

sinkingsugar:
I'm giving a bit of priority to solve NBC and gossip 1.1 right now, will add definitely more tests. Making a issue out of it.

PR #387, opened 2020-10-01
feature/expose-matcher

decanus:
Allows to mount a protocol with a specific matcher.


dryajov:
Thanks! It appears that there no tests for this where ever added, it'd be great if we can add some for the matchers.

sinkingsugar:
Maybe we should merge and add tests later? I can add it to my backlog 😄 
As I suspect this might be needed for waku.

PR #391, opened 2020-10-05
gossip MessageID as seq[byte]

sinkingsugar:


libp2p/protocols/pubsub/rpc/message.nim
```diff
@@ -28,11 +28,20 @@ const PubSubPrefix = toBytes("libp2p-pubsub:")
...
   if m.seqno.len > 0 and m.fromPeer.data.len > 0:
-    byteutils.toHex(m.seqno) & $m.fromPeer
+    res &= m.seqno
+    res &= m.fromPeer.data
   else:
-    $m.data.hash & $m.topicIDs.hash
+    var
+      dataHash = m.data.hash
+      topicHash = m.topicIDs.hash
+      bDataHash = cast[ptr UncheckedArray[byte]](addr dataHash)
```
arnetheduck, line 40: https://nim-lang.org/1.2.0/hashes.html#%21%26%2CHash%2Cint ? avoid those casts..?

sinkingsugar: ok but, 1 cast is necessary

libp2p/protocols/pubsub/rpc/message.nim
```diff
@@ -28,11 +28,20 @@ const PubSubPrefix = toBytes("libp2p-pubsub:")
 declareCounter(libp2p_pubsub_sig_verify_success, "pubsub successfully validated messages")
 declareCounter(libp2p_pubsub_sig_verify_failure, "pubsub failed validated messages")
 
-func defaultMsgIdProvider*(m: Message): string =
+func defaultMsgIdProvider*(m: Message): MessageID =
+  var res: seq[byte]
   if m.seqno.len > 0 and m.fromPeer.data.len > 0:
-    byteutils.toHex(m.seqno) & $m.fromPeer
+    res &= m.seqno
```
sinkingsugar, line 34: @arnetheduck I didn't really think much but this is wrong too in respect to compatibility with go/rust 

Everyone is still using strings:
https://github.com/libp2p/go-libp2p-pubsub/blob/a3445b756fdb68f372dadd45de65a723d0237ac8/pubsub.go#L965

I suggest we don't do anything at this point since string is a sequence of bytes anwyay.
At max just change the protobuf to seq[bytes] just for nitpicking sake but this defaultMsgIdProvider should stay `string` as it avoids a useless copy.

arnetheduck: so go uses bigendian consistently here - let's do the same (with `stew/endians2/toBytesBE`): https://github.com/libp2p/go-libp2p-pubsub/blob/a3445b756fdb68f372dadd45de65a723d0237ac8/pubsub.go#L1182

arnetheduck: when available, non-anonymous mode.. it's a good habit to never use machine-dependent int encodings in networking situations anyway.

sinkingsugar: @arnetheduck we are already going that in `Message.init` btw the `toBytesBE`.

sinkingsugar: The current code is wrong tho, compared to other implementations. I'm not sure yet how to approach this tbh

arnetheduck: right - got confused as to which hash / id this is.. so for now, if there's no seqno/peer, I think it's better to default to `sha256` instead of the nim hash as the latter runs a (comparatively) high risk of false conflicts - if they are set, we should simply match what go/rust does so that the id are computed the same way

in addition, we should check the 1.1 spec to see if there's anything about the message id - if there's not, we should clarify how the message id function relates to correct ihave/iwant processing, and open an issue in the libp2p spec repo in case having different message id functions in different implementations impacts correctness

arnetheduck (approved):

PR #395, opened 2020-10-09
Extended validators

sinkingsugar:
Gossip 1.1 spec's extended validation

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -1062,7 +1062,7 @@ method rpcHandler*(g: GossipSub,
     # g.anonymize needs no evaluation when receiving messages
     # as we have a "lax" policy and allow signed messages
 
-    if not (await g.validate(msg)):
+    if g.validate(msg).await == ValidationResult.Reject:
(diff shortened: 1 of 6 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -1062,7 +1062,7 @@ method rpcHandler*(g: GossipSub,
     # g.anonymize needs no evaluation when receiving messages
     # as we have a "lax" policy and allow signed messages
 
-    if not (await g.validate(msg)):
+    if g.validate(msg).await == ValidationResult.Reject:
```
dryajov, line 1065: Can we just do `if (await g.validate(msg)) == ValidationResult.Reject:` - seems a lot more natural and consistent.

dryajov (approved):
LGTM.

libp2p/protocols/pubsub/gossipsub.nim
sinkingsugar: hehe ok, I really hate .net and c# to start the `await\s.*` trend to be honest.. yes it became natural but in a weird way to be honest.

PR #396, opened 2020-10-09
replace nim-beacon-chain URLs with nimbus-eth2 URLs

tersec:


mratsim:
Azure Test fail due to timeout ...

PR #397, opened 2020-10-11
handle secure errors in readOnce (secure)

sinkingsugar:
As discussed in the previous meeting

### DIFF
--- a/libp2p/protocols/secure/secure.nim
+++ b/libp2p/protocols/secure/secure.nim
@@ -125,7 +125,7 @@ method readOnce*(s: SecureConn,
       except CatchableError as exc:
         (@[], exc)
 
-    if err != nil:
+    if isNil(err):
       warn "error while reading message from secure connection, closing.",  error=err.name, 
                                                                             message=err.msg, 
                                                                             connection=s
--- a/libp2p/protocols/secure/secure.nim
+++ b/libp2p/protocols/secure/secure.nim
@@ -125,7 +125,7 @@ method readOnce*(s: SecureConn,
       except CatchableError as exc:
         (@[], exc)
 
-    if isNil(err):
+    if not isNil(err):
       warn "error while reading message from secure connection, closing.",  error=err.name, 
                                                                             message=err.msg, 
                                                                             connection=s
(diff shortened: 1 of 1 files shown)


libp2p/protocols/secure/secure.nim
```diff
@@ -120,10 +120,23 @@ method readOnce*(s: SecureConn,
   doAssert(nbytes > 0, "nbytes must be positive integer")
 
   if s.buf.data().len() == 0:
-    let buf = await s.readMessage()
+    let (buf, err) = try:
+        (await s.readMessage(), nil)
+      except CatchableError as exc:
+        (@[], exc)
```
cheatfate, line 126: I think it would be better to return `(@[], exc.name, exc.msg)` because i'm not sure about lifetime of this exception object.

libp2p/protocols/secure/secure.nim
```diff
@@ -120,10 +120,23 @@ method readOnce*(s: SecureConn,
...
+
+    if err != nil:
```
cheatfate, line 128: Lets not compare with `nil` directly because it can cause problems in future (Nim changes) so its better and recommended way is `isNil(err)`.

libp2p/protocols/secure/secure.nim
sinkingsugar: Hmm I kinda wanted to just re-throw, what makes you think the reference would be collected?

sinkingsugar: will just copy the ref (add_ref) and exit the block

libp2p/protocols/secure/secure.nim
```diff
@@ -120,10 +120,23 @@ method readOnce*(s: SecureConn,
   doAssert(nbytes > 0, "nbytes must be positive integer")
 
   if s.buf.data().len() == 0:
-    let buf = await s.readMessage()
+    let (buf, err) = try:
+        (await s.readMessage(), nil)
+      except CatchableError as exc:
+        (@[], exc)
+
+    if isNil(err):
```
dryajov, line 128:
```suggestion
    if not isNil(err):
```

sinkingsugar: ah! indeed, see this construct is always messing with my brain waves... I definitely don't like it to be honest, also in nim requires more instructions likely due to the `not` and likely C compiler being fooled.

libp2p/protocols/secure/secure.nim
sinkingsugar: I tested it a bit, seems to be fine

PR #398, opened 2020-10-14
Fix first interop test on windows platforms

sinkingsugar:


sinkingsugar:
Issue was in the handler, closing too quick the stream

PR #412, opened 2020-10-21
gossipsub audit fixes

sinkingsugar:
… scores #405


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -409,7 +409,7 @@ proc rebalanceMesh(g: GossipSub, topic: string) {.async.} =
     shuffle(grafts)
 
     # sort peers by score
-    grafts.sort(byScore)
+    grafts.sort(byScore, SortOrder.Descending)
```
arnetheduck, line 412: In general, this would be a good opportunity to document the rationale behind the sorting as well - ie why should peers be sorted by score here - a comment like "sort peers by score" just repeats what the code already says but doesn't help the reader understand why the sorting makes sense, and why it's important for security

Issue #393, opened 2020-10-07
Channel leak on cancellation


arnetheduck:
Using the following script, it's possible to find channels that have not been closed properly - in particular, they are correlated with request timeouts triggered from here: https://github.com/status-im/nim-beacon-chain/blob/c5a58555efdef09cd18d2d68b62316fedced156e/beacon_chain/eth2_network.nim#L647

ie when running the script and looking for logs surrounding the leaking channels, that line pops up - what happens with a timeout is that `readChunkPayload` gets cancelled by the timer.

```python
import re,sys

re0 = re.compile(r"Creating.*channel\":\"([^\:]+:.*?)[:\"]")
re1 = re.compile(r"Closed.*s\":\"([^\"]+)")

x = set()

with open(sys.argv[1]) as f:
  data = f.readlines()
for line in data:
  for match in re0.finditer(line):
      x.add(match.group(1))

  for match in re1.finditer(line):
    x.discard(match.group(1))

for a in x:
  for line in data:
    if a in line:
      print(line, end = '')      
  print(a)
```

sinkingsugar:
I'm digging this a bit, `readChunkPayload` possible fault points, or anyway cancellation points, are a `readVarint` or, if the `readVarint` succeedes, a `readExactly`; both on a `LPStream`.
In turn they all use `readOnce`, which in this case assuming mplex it's a `LPChannel`.
`LPChannel.readOnce` does NOT handle cancellation but handles `CatchableError` alone.. this is probably a issue.

sinkingsugar:
@arnetheduck I'm a bit confused on the Open/Close stream tho, `readChunkPayload` seems to use an always opened connection to me, from `handleIncomingStream`, altho I might be confused by the amount of templates.

arnetheduck:
The message I referred to in the bug description is printed when this happens - @zah understands the eth2 flow best though - I believe it should be possible to write a smaller test case here though

zah:
The code linked by Jacek is executed only on incoming connections indeed. I can try replicating the issue locally with a custom request client that doesn't send data in time. Doing this in a loop will hopefully confirm the leak.

arnetheduck:
[...]

PR #413, opened 2020-10-23
Channel leaks

dryajov:
Attempting to fix channel leaks reported in #393.

### DIFF
--- a/libp2p/protocols/secure/secure.nim
+++ b/libp2p/protocols/secure/secure.nim
@@ -55,7 +55,7 @@ method initStream*(s: SecureConn) =
 method close*(s: SecureConn) {.async.} =
   trace "closing secure conn", s, dir = s.dir
   if not(isNil(s.stream)):
-    await s.stream.closeWithEOF()
+    await s.stream.close()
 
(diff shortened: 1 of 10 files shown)


libp2p/protocols/secure/secure.nim
```diff
@@ -52,8 +53,9 @@ method initStream*(s: SecureConn) =
   procCall Connection(s).initStream()
 
 method close*(s: SecureConn) {.async.} =
+  trace "closing secure conn", s, dir = s.dir
   if not(isNil(s.stream)):
-    await s.stream.close()
+    await s.stream.closeWithEOF()
```
arnetheduck, line 58: this looks wrong - just because we stop sending doesn't mean we've read everything there was to read? ie `closeWithEOF` should only be used when there is an in-protocol signal that all reading is finished, for example like in eth2 where you read one message and that's it

dryajov: Yeah, this was mostly an experiment to see if it synchronizes closes between the child and the parent streams. It doesn't seem like it made any difference so reverting back.

PR #416, opened 2020-11-01
Move triggers

dryajov:
Moving peer events int connmanager

### DIFF
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -269,7 +269,7 @@ proc onClose(c: ConnManager, conn: Connection) {.async.} =
           errMsg = exc.msg, conn
   finally:
     trace "Triggering peerCleanup", conn
-    await c.peerCleanup(conn)
+    asyncSpawn c.peerCleanup(conn)
 
 proc selectConn*(c: ConnManager,
                 peerId: PeerID,
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -124,19 +124,20 @@ proc triggerPeerEvents*(c: ConnManager,
                         peerId: PeerID,
                         event: PeerEvent) {.async, gcsafe.} =
 
+  trace "About to trigger peer events", peer = peerId
   if event notin c.peerEvents:
     return
 
   try:
     let count = c.connCount(peerId)
     if event == PeerEvent.Joined and count != 1:
-      trace "peer already joined", peerId, event
+      trace "peer already joined", peerId, event = $event
       return
     elif event == PeerEvent.Left and count != 0:
-      trace "peer still connected or already left", peerId, event
+      trace "peer still connected or already left", peerId, event = $event
       return
 
-    trace "triggering peer events", peerId, event
+    trace "triggering peer events", peerId, event = $event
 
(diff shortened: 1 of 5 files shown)


libp2p/switch.nim
```diff
@@ -487,44 +466,11 @@ proc muxerHandler(s: Switch, muxer: Muxer) {.async, gcsafe.} =
 
   try:
     await s.identify(muxer)
-  except CatchableError as exc:
+  except IdentifyError as exc:
```
arnetheduck, line 469: what if the remote end closes the identify connection early and a streameof happens on the connection that identify was using?

dryajov: Good point, we should handle those as well.

arnetheduck:
This would be a good time to introduce the Cancellation tests, before doing any other changes - ie that's where the current leak is, and it should be addressed with a test before anything else

dryajov:
The problem is that they are almost impossible to properly test with the current `listen` api, hence I've added them in the accepts #420 PR.

arnetheduck:
right, that's a good start! we need cancellation tests for readOnce and close as well as those are the likely culprits of the  resource leak we're seeing.

dryajov:
Yeah those are good to add as well. I'll add them in there.

libp2p/connmanager.nim
```diff
@@ -224,6 +225,32 @@ proc cleanupConn(c: ConnManager, conn: Connection) {.async.} =
 
   trace "Connection cleaned up", conn
 
+proc peerStartup(c: ConnManager, conn: Connection) {.async.} =
+  try:
+    trace "Triggering peer and connection events on connect", conn
+    let peerId = conn.peerInfo.peerId
+    await c.triggerPeerEvents(peerId, PeerEvent.Joined)
```
arnetheduck, line 232: this looks wrong - we should only trigger the Joined peer event on the _first_ connection to the peer, no? likewise for the `Left` - it would only happen when the last connection is fully cleaned up

libp2p/switch.nim
```diff
@@ -167,13 +167,9 @@ proc mux(s: Switch, conn: Connection): Future[Muxer] {.async, gcsafe.} =
   muxer.streamHandler = s.streamHandler
 
   s.connManager.storeOutgoing(conn)
-  s.connManager.storeMuxer(muxer)
-
-  # start muxer read loop - the future will complete when loop ends
-  let handlerFut = muxer.handle()
```
arnetheduck, line 173: afair, this was done so that the muxer would be available in connmanager for `handle` handlers that might run _before_ `storeMuxer`

libp2p/connmanager.nim
dryajov: This is taken care of in the `triggerPeerEvents` call - this hasn't changed at all, just moved from the switch, we also test for this and I haven't yet seen this triggered more than once.

libp2p/switch.nim
dryajov: Yes this is suspect as well and I did consider this when making the change. It might be the change that broke this PR, but I have not yet been able to reproduce it. Any chance you can try reverting this and see if it resolves the potential issues you were seeing on our infra?

PR #417, opened 2020-11-01
Revert "Channel leaks (#413)"

arnetheduck:
This reverts commit 1de1d49223505a44535fa8835d1c0c65fc4c00fb.

The leak fix seems to turn the leak into deadlock which occasionally causes nbc to freeze when requesting blocks and similar issues - this needs more investigation before going in - it could potentially be the gossipsub 1.1 upgrade but that seems less likely since the lockups are related to fetching blocks.

Don't have a perfect repro, but syncing and reconnecting wifi make the symptoms show, including large differences between libp2p and nbc peer counts and sync requests that don't end for hours (even when they should time out)

### DIFF
--- a/libp2p/muxers/mplex/lpchannel.nim
+++ b/libp2p/muxers/mplex/lpchannel.nim
@@ -66,6 +66,12 @@ proc open*(s: LPChannel) {.async, gcsafe.} =
 method closed*(s: LPChannel): bool =
   s.closedLocal
 
+proc closeUnderlying(s: LPChannel): Future[void] {.async.} =
+  ## Channels may be closed for reading and writing in any order - we'll close
+  ## the underlying bufferstream when both directions are closed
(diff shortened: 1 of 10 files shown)


dryajov:
It would be useful for further troubleshooting to provide a trace of a conversation that shows this behavior, or at least instructions on how to obtain the trace. This would be highly appreciated.

dryajov:
In addition to this, I have observed a few long lived connections (several hours), but upon inspection, this do seem to be legitimate and I see data being passed back and forth. Having a trace that identifies the issue you describe here would help filter this legit connections from pure hangs.

arnetheduck:
https://paste.c-net.org/LickingStrips shows the behaviour - you can see the dead connection in syncmanager, stuch in `D`, ie waiting for a download request to finish

PR #418, opened 2020-11-01
split stream counts by direction

dryajov:
re-adding split counts for streams


libp2p/transports/tcptransport.nim
```diff
@@ -63,7 +63,15 @@ proc connHandler*(t: TcpTransport,
                   client: StreamTransport,
                   initiator: bool): Connection =
```
arnetheduck, line 64: probably easier to just pass direction in here

PR #420, opened 2020-11-02
Rework transport to use chronos accept

dryajov:


libp2p/wire.nim
```diff
@@ -79,28 +87,46 @@ proc createStreamServer*[T](ma: MultiAddress,
                             child: StreamServer = nil,
                             init: TransportInitCallback = nil): StreamServer =
   ## Create new TCP stream server which bounds to ``ma`` address.
-  var address = initTAddress(ma)
-  if address.family in {AddressFamily.IPv4, AddressFamily.IPv6}:
-    if ma[1].tryGet().protoCode().tryGet() != multiCodec("tcp"):
-      raise newException(TransportAddressError, "Incorrect address type!")
-  result = createStreamServer(address, cbproc, flags, udata, sock, backlog,
+  if not(RTRANSPMA.match(ma)):
+    raise newException(TransportAddressError, "Incorrect address type!")
```
cheatfate, line 91: This is incorrect exception type and message, because address could be absolutely correct (for example "/ip4/127.0.0.1/udp/13000`) but it would be unsupported here, so i think message should be different.
Also it is better not use `TransportAddressError` because this is `nim-chronos` defined Exception type and used to signal about incorrect addresses. So maybe its better to rename the exception type too.

libp2p/wire.nim
```diff
@@ -79,28 +87,46 @@ proc createStreamServer*[T](ma: MultiAddress,
...
+                            flags: set[ServerFlags] = {},
+                            udata: ref T,
+                            sock: AsyncFD = asyncInvalidSocket,
+                            backlog: int = 100,
+                            bufferSize: int = DefaultStreamBufferSize,
+                            child: StreamServer = nil,
+                            init: TransportInitCallback = nil): StreamServer =
+  ## Create new TCP stream server which bounds to ``ma`` address.
+  if not(RTRANSPMA.match(ma)):
+    raise newException(TransportAddressError, "Incorrect address type!")
```
cheatfate, line 107: Same problem with exception type and message.

libp2p/transports/tcptransport.nim
```diff
@@ -72,121 +72,135 @@ proc connHandler*(t: TcpTransport,
...
+    trace "Error shutting down tcp transport", exc = exc.msg
+  finally:
+    t.running = false
+
+template withTransportErrors(body: untyped): untyped =
+  try:
+    body
+  except TransportTooManyError as exc:
+    warn "Could not create new connection, too many files opened", exc = exc.msg
+    raise newTransportClosedError(exc)
```
cheatfate, line 168: `TransportTooManyError` is not fatal/critical error, most of the server software just note about this problem in logs, but it should not crash software. Because it is possible that on next accept free socket will be available.

libp2p/wire.nim
dryajov: Interesting, this is a copy paste from you original change in https://github.com/status-im/nim-libp2p/pull/247/files#diff-7831cb8a03a4c3de33123e27a62ca63d25b578b618ea8623e72f2135ed89ea66R74 (just making sure, maybe it made sense back then but not anymore). So you suggest I remap the exception to something else or not use it at all?

libp2p/wire.nim
dryajov: Noted, I'll make thin non fatal.

libp2p/transports/tcptransport.nim
dryajov: Noted, I'll make thin non fatal.

libp2p/switch.nim
```diff
@@ -407,41 +414,55 @@ proc mount*[T: LPProtocol](s: Switch, proto: T, matcher: Matcher = nil) {.gcsafe
...
+      let conn = await transport.accept()
+      trace "Accepted an incoming connection", conn
+      asyncSpawn s.upgradeIncoming(conn) # perform upgrade on incoming connection
+    except TransportClosedError as exc:
+      debug "Transport closed", exc = exc.msg
+      return
     except CancelledError as exc:
-      raise exc
+      trace "Canceling accept loop"
+      return
```
cheatfate, line 432: In this case, this does not have a significant effect, however you introduced leak of `conn` on cancellation. So if you expect cancellation here you should close `conn` if it was created.

libp2p/switch.nim
```diff
@@ -238,20 +239,26 @@ proc upgradeIncoming(s: Switch, conn: Connection) {.async, gcsafe.} =
       await ms.handle(sconn)
 
     except CancelledError as exc:
+      trace "Canceling secure handler for incoming connection", conn
```
cheatfate, line 242: On cancel `sconn` could leak.

libp2p/wire.nim
cheatfate: I think libp2p as a high level network framework should have its own exception types, because for `chronos` this exception `TransportAddressError` means that address is incorrect, but for `libp2p` its meaning is different, its not `incorrect` but `not-supported`.

libp2p/switch.nim
```diff
@@ -407,41 +411,63 @@ proc mount*[T: LPProtocol](s: Switch, proto: T, matcher: Matcher = nil) {.gcsafe
...
+        debug "Accepted an incoming connection", conn
+        asyncSpawn s.upgradeIncoming(conn) # perform upgrade on incoming connection
+      else:
+        # A nil connection means that we might have hit a
+        # file-handle limit (or another non-fatal error),
+        # we can get one on the next try, but we should
+        # be careful to not end up in a thigh loop that
+        # will starve the main event loop, thus we sleep
+        # here before retrying.
+        await sleepAsync(100.millis) # TODO: should be configurable?
```
sinkingsugar, line 433: file handle limit is quite fatal tho imho.

libp2p/transports/tcptransport.nim
```diff
@@ -26,10 +26,8 @@ const
 type
   TcpTransport* = ref object of Transport
     server*: StreamServer
-    clients: seq[StreamTransport]
+    clients: array[bool, seq[StreamTransport]]
```
sinkingsugar, line 29: wait what? `array[bool, seq[StreamTransport]]` should be a compiler error

libp2p/switch.nim
dryajov: Not really (according to @cheatfate), we can recover from it.

libp2p/transports/tcptransport.nim
dryajov: Nope, works as expected. Handy trick for when you have only two types of objects, like `incoming == true/false`.

libp2p/switch.nim
dryajov: > fine to be careful but not sure this is really needed

Well, if we don't do await something here, it will just go into a spi and never release to the main loop if we keep getting a nil back? At least that the reasoning, but maybe not since there are other awaits in between. Not opposed to removing if it doesn't make sense.

sinkingsugar: makes sense I get the same feeling but ideally should not happen :), better be safe!

libp2p/transports/tcptransport.nim
sinkingsugar: This is very confusing, why not use a enum type instead?
Bool flags are bad in general + array is more convoluted. I get the handy trick part but its very unclear.
```
MyType = object
  kind side
  of incoming:
    inClients: seq[...]
  of outgoing:
    outClients: seq[...]
```

sinkingsugar: Altho applied to code array looks better, still a hack emulating rust `match` tho... fine for me tho in the end! 😄

libp2p/switch.nim
```diff
@@ -407,41 +411,63 @@ proc mount*[T: LPProtocol](s: Switch, proto: T, matcher: Matcher = nil) {.gcsafe
...
     except CatchableError as exc:
-      trace "Exception occurred in incoming handler", conn, msg = exc.msg
-    finally:
-      await conn.close()
-    trace "Connection handler done", conn
+      debug "Exception in accept loop, exiting", exc = exc.msg
+      if not isNil(conn):
+        await conn.closeWithEOF()
```
arnetheduck, line 437: closeWithEOF isn't really appropriate here since we don't know if the other end will be sending an EOF really - ie the connection should just be `close`:ed - if it were an LPChannel, it would need a `reset` - probably, the concept of half-closed connections needs to be moved to `LPStream` itself so that user code can distinguish between `close` and `reset` (which have unfortunate names, given decades of `shutdown` and `close` being used for this purpose)

libp2p/transports/tcptransport.nim
```diff
@@ -7,8 +7,8 @@
 ## This file may not be copied, modified, or distributed except according to
 ## those terms.
 
-import oids
-import chronos, chronicles, sequtils
+import oids, sequtils
```
arnetheduck, line 10:
```suggestion
import std/[oids, sequtils]
```

libp2p/transports/tcptransport.nim
arnetheduck: could also use the `Direction` enum, ie `array[Direction, ...]` now that it's binary

libp2p/transports/tcptransport.nim
```diff
@@ -72,121 +72,132 @@ proc connHandler*(t: TcpTransport,
...
-      # This is top-level procedure which will work as separate task, so it
-      # do not need to propogate CancelledError.
-      trace "Unexpected cancellation in transport's cleanup"
+
+      t.clients[initiator].keepItIf( it != client )
+      if not(isNil(client) and client.closed()):
+        await client.closeWait()
+
+      trace "Cleaned up client", addrs = $client.remoteAddress,
+                                 conn = $conn.oid
```
arnetheduck, line 89: better use the shortlog (just pass `conn`)

PR #425, opened 2020-11-05
Reset channels on close

dryajov:


tests/testmplex.nim
```diff
@@ -594,6 +594,468 @@ suite "Mplex":
...
+      for s in dialStreams:
+        try:
+          discard await s.readLp(1024)
+          check false
+        except LPStreamEOFError:
+          check true
+          continue
+
+      await sleepAsync(1.seconds)
+      var channelTracker = getTracker(LPChannelTrackerName)
```
arnetheduck, line 874: this could be a template that prints the tracker on leak (or reports it to unittest) like the one in helpers

tests/testmplex.nim
```diff
@@ -594,6 +594,468 @@ suite "Mplex":
...
+
+      let mplexDial = Mplex.init(conn)
+      let mplexDialFut = mplexDial.handle()
+      var dialStreams: seq[Connection]
+      for i in 0..9:
+        dialStreams.add((await mplexDial.newStream()))
+
+      for i, s in dialStreams:
+        await s.closeWithEOF()
+        await sleepAsync(10.millis)
```
arnetheduck, line 632: these sleeps should ideally not be needed - ie everything is deterministic in the test in the sense that if the close finishes, it should lead to ordering guarantees - if that's missing, it's likely an opportunity for a race condition

dryajov: yeah you right, I originally used `s.close()` which would fail to close because there where no readers/writers on the stream and the sleep is a leftover from that.

dryajov:
The two most important changes which lead to this tests are on lines https://github.com/status-im/nim-libp2p/pull/425/files#diff-3d23c942198cd88724375e828a0a972b0fc19a154f07d844afdae1a7020b1109R185 and https://github.com/status-im/nim-libp2p/pull/425/files#diff-3d23c942198cd88724375e828a0a972b0fc19a154f07d844afdae1a7020b1109R201, which make sure that the channels are reset as soon as possible if the muxed connection died. It shouldn't hurt anything, but I don't see a major change in behaviour either so I'm leaving them there right now.

tests/testmplex.nim
```diff
@@ -594,6 +594,347 @@ suite "Mplex":
...
+      let conn = await transport2.dial(transport1.ma)
+
+      let mplexDial = Mplex.init(conn)
+      let mplexDialFut = mplexDial.handle()
+      var dialStreams: seq[Connection]
+      for i in 0..9:
+        dialStreams.add((await mplexDial.newStream()))
+
+      await listenConn.close()
+      await allFuturesThrowing(
```
arnetheduck, line 922: hm, I wonder.. this could probably be written as `mapIt(dial&listen, await it.join())`

dryajov: Sure, but I honestly prefer the more consistent method call syntax.

arnetheduck: sure, it's a bit pointless to do double "all" though.. ie `allFuturesThrowing(mapit(...))` does the same at the end of the day

dryajov: ah, yeah - you're right 👍

PR #436, opened 2020-11-13
read failure

arnetheduck:
Test showing read failure on cancel


arnetheduck:
this test looked like it failed locally, but it turns out it was a timing issue in another  test :/

dryajov:
it's a good test to have either way, lets keep it.

PR #437, opened 2020-11-13
Gossip runtime params

sinkingsugar:


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -31,27 +31,10 @@ const
   GossipSubCodec* = "/meshsub/1.1.0"
   GossipSubCodec_10* = "/meshsub/1.0.0"
 
-# overlay parameters
-const
```
dryajov, line 35: I think we can keep the consts as they are in the default spec and set them as defaults, overridable in `init()`?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -174,9 +167,16 @@ proc init*(_: type[GossipSubParams]): GossipSubParams =
       pruneBackoff: 1.minutes,
       floodPublish: true,
       gossipFactor: 0.25,
+      d: 6,
```
dryajov, line 170: are this values hardcoded now?

libp2p/protocols/pubsub/gossipsub.nim
sinkingsugar: True

libp2p/protocols/pubsub/gossipsub.nim
sinkingsugar: let's keep the const params and use those in here

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -174,9 +185,16 @@ proc init*(_: type[GossipSubParams]): GossipSubParams =
       pruneBackoff: 1.minutes,
       floodPublish: true,
       gossipFactor: 0.25,
+      d: GossipSubD,
+      dLow: GossipSubDlo,
+      dHigh: GossipSubDhi,
       dScore: 4,
-      dOut: GossipSubDlo - 1,
-      dLazy: GossipSubD,
+      dOut: 4 - 1, # DLow - 1
```
dryajov, line 192: Maybe should add this as constants as well?

dryajov:
Looks good. How is the application going to provide custom params?

sinkingsugar:
> Looks good. How is the application going to provide custom params?

The usual https://github.com/status-im/nim-libp2p/blob/92fa4110c1169b0040d36e86415ebe7ea3f4cc46/libp2p/protocols/pubsub/pubsub.nim#L350
passing parameters here

dryajov (approved):

PR #439, opened 2020-11-16
fix channels not being reset

arnetheduck:
silly for loop..

* allow only one concurrent read
* fix mplex test race condition
* add some bufferstream eof tests

### DIFF
--- a/libp2p/muxers/mplex/lpchannel.nim
+++ b/libp2p/muxers/mplex/lpchannel.nim
@@ -88,17 +88,20 @@ proc reset*(s: LPChannel) {.async, gcsafe.} =
   s.readBuf = StreamSeq()
   s.pushedEof = true
 
-  for i in 0..<s.pushing:
+  let pushing = s.pushing # s.pushing changes while iterating
+  for i in 0..<pushing:
     # Make sure to drain any ongoing pushes - there's already at least one item
     # more in the queue already so any ongoing reads shouldn't interfere
     # Notably, popFirst is not fair - which reader/writer gets woken up depends
     discard await s.readQueue.popFirst()
 
-  if s.readQueue.len == 0 and s.pushing == 0:
-    # There is no push ongoing and nothing on the queue - let's place an
-    # EOF marker there so that any reader is woken up - we don't need to
-    # synchronize here
-    await s.readQueue.addLast(@[])
+  if s.readQueue.len == 0 and s.reading:
+    # There is an active reader - we just grabbed all pushes so we need to push
+    # an EOF marker to wake it up
+    try:
+      s.readQueue.addLastNoWait(@[])
+    except CatchableError:
+      raiseAssert "We just checked the queue is empty"
 
   if not s.conn.isClosed:
     # If the connection is still active, notify the other end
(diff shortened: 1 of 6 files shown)


libp2p/stream/bufferstream.nim
```diff
@@ -107,10 +109,14 @@ method readOnce*(s: BufferStream,
 
   if rbytes < nbytes:
     # There's space in the buffer - consume some data from the read queue
-    trace "popping readQueue", s, rbytes, nbytes
-    let buf = await s.readQueue.popFirst()
-
-    if buf.len == 0 or s.isEof: # Another task might have set EOF!
+    s.reading = true
```
dryajov, line 112: I tried adding this but it didn't seem to ever trigger, so I left it out for now.

libp2p/stream/bufferstream.nim
```diff
@@ -107,10 +109,14 @@ method readOnce*(s: BufferStream,
...
+    let buf =
+      try:
+        await s.readQueue.popFirst()
```
dryajov, line 115: try/finally potential issues here?

libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -88,17 +88,20 @@ proc reset*(s: LPChannel) {.async, gcsafe.} =
   s.readBuf = StreamSeq()
   s.pushedEof = true
 
-  for i in 0..<s.pushing:
+  let pushing = s.pushing # s.pushing changes while iterating
```
dryajov, line 91: Ok, looks sound. Going to take it for a drive, hopefully this is it.

libp2p/stream/bufferstream.nim
dryajov: btw, this should be moved to `LPStream`.

arnetheduck: it didn't trigger because typically it would deadlock before, in the `reset` loop. moving it to lpstream doesn't quite make sense since each stream handles this a bit differently, ie chronosstream handles the single-reader requirement "internally" already. what we really need is to expose `reset` to nbc as well because that's what it should do when its timeout is reached - notably, the nbc timeout is different than the libp2p - libp2p restarts timer on _any_ activity while nbc counts complete chunk time

libp2p/stream/bufferstream.nim
arnetheduck: as long as it's not nested and not in a loop, it's fine

arnetheduck:
* prevent concurrent `reset` calls
* reset LPChannel when read is cancelled (since data is lost)
* ensure there's one, and one only, 0-byte readOnce on EOF
* ensure that all data is returned before EOF is returned
* keep running activity monitor for half-closed channels (or they never
get closed)

libp2p/stream/bufferstream.nim
```diff
@@ -130,6 +147,12 @@ method readOnce*(s: BufferStream,
 
   s.activity = true
 
+  # We want to return 0 exactly once - after that, we'll start raising instead -
+  # this is a bit nuts in a mixed exception / return value world, but allows the
+  # consumer of the stream to rely on the 0-byte read as a "regular" EOF marker
+  # (instead of _sometimes_ getting an exception).
```
dryajov, line 153: Not sure this is necesary? Why not just raise?

arnetheduck: consistency? zero-return is customary, also because you if you have a fully regular communication, you should not be hitting _exceptions_ - it's not an exception that a channel is closed

arnetheduck: other streams _sometimes_ return 0 as well, this is mainly giving it a stable semantic that should be implemented across the board so that it's possible to reason about code flows somehow (and not having lots of different behaviours randomly)

PR #443, opened 2020-11-19
don't quit accept prematurely

dryajov:


libp2p/transports/tcptransport.nim
```diff
@@ -159,28 +159,29 @@ method stop*(t: TcpTransport) {.async, gcsafe.} =
...
   ##
 
   if not t.running:
     raise newTransportClosedError()
 
-  withTransportErrors:
+  try:
     let transp = await t.server.accept()
     return await t.connHandler(transp, Direction.In)
+  except TransportOsError as exc:
```
dryajov, line 172: @cheatfate any idea on how to filter fatal from non-fatal os errors?

cheatfate: All the `TransportOsError` are fatal, non-fatal errors only `TransportTooManyError`.

cheatfate: `TransportUseClosedError` is an error which shows that you using "closed" transport.

dryajov: So on accept, I would get something like error `(21) Incorrect Params`(or similar, just an example) and the accept loop would quit because I propagated this errors to it. I guess this error is per connection, so it's fine to ignore it here, because we don't want to shutdown the entire accepts loop just because of one peer.

dryajov:
I merged it as is for now, because otherwise incoming connections would stop due to failed connecting peers.

libp2p/transports/tcptransport.nim
cheatfate: No its not fine because this error is not recoverable, you will continue receive this errors for every possible call to `accept()`. So in such case you will just spam error log with your errors and you will not accept any new connections. For example error `Incorrect params` will not disappear because you passed incorrect parameters...

cheatfate: Usually OS accept() call operates on socket which is bound to specific address:port, and this call returns new sockets. So if its not `TransportTooManyError` you have a real problem on this server socket. And you will be unable to manage it and fix the error. So if you continue calling `accept()` you will just burn your CPU.

dryajov: OK, the accept loop exists almost immediately if I allow it to propagate this errors. If I do ignore them it seems to recover fine, however. I also get it for _some_ peers while others work just fine - so at least in practice on my machine I am seeing a different behavior from what you describe (and I assumed when originally implementing it). 

What do I look for to troubleshoot this?

cheatfate: According to man accept, possible OS returned values are:

* EAGAIN or EWOULDBLOCK  - impossible to obtain (covered by system queue).
* EBADF - Unrecoverable error.
* ECONNABORTED - Recoverable error (Looks like race condition).
* EINTR - handled inside `transport.accept()`
* EINVAL - Unrecoverable error.
* EMFILE - should be TransportTooManyError.
* ENFILE - should be TransportTooManyError.
* ENOBUFS - shoud be TransportTooManyError.
* ENOMEM - disputable, but i think its unrecoverable error.
* ENOTSOCK - should not be happened, but unrecoverable error.
* EOPNOTSUPP - Unrecoverable error.
* EPROTO - Unrecoverable error.

PR #444, opened 2020-11-22
Don't allow concurrent pushdata

dryajov:
This PR does two things:

1. Disables concurrent `pushData` calls, which 1) break backpressure because at that point the async subsystem acts as a boundless buffer 2) adds unneeded complexity to the implementation, which has been a source of multiple issues that attempt to handle corner cases introduced by a  useless/faulty requirement.
2. Corrects the reset flow to handle the following cases:

- If a push was in progress but no reader is attached we need to pop the queue
- If a read was in progress without a push/data we need to push the Eof marker to notify the reader that the channel closed
- In all other cases, there should be a data to complete a read or enough room in the queue/buffer to complete a
push.

Here is a quick truth table to illustrate the cases:

State       | Q Empty  | Q Full
------------|----------|-------
Reading     | Push Eof | Na
Pushing     | Na       | Pop

### DIFF
--- a/libp2p/muxers/mplex/lpchannel.nim
+++ b/libp2p/muxers/mplex/lpchannel.nim
@@ -90,16 +90,23 @@ proc reset*(s: LPChannel) {.async, gcsafe.} =
   s.readBuf = StreamSeq()
   s.pushedEof = true
 
-  # Took me some time to get this right.
-  # I'll leave this here in case we stumble
-  # on it again.
+  # Essentially we need to handle the following cases
+  #
+  # - If a push was in progress but no reader is
+  # attached we need to pop the queue
+  # - If a read was in progress without without a
+  # push/data we need to push the Eof marker to
+  # notify the reader that the channel closed
+  #
+  # In all other cases, there should be a data to complete
+  # a read or enough room in the queue/buffer to complete a
+  # push.
   #
   # State       | Q Empty  | Q Full
   # ------------|----------|-------
   # Reading     | Push Eof | Na
   # Pushing     | Na       | Pop
-  #
-  if not(s.reading and s.pushing): # if either a reader or pusher is missing
+  if not(s.reading and s.pushing):
     if s.reading:
       if s.readQueue.empty():
         # There is an active reader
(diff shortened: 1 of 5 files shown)


dryajov:
@sinkingsugar @arnetheduck I'd appreciate some help testing this as different issues tend to show up on different platforms and hardware, so the more testing we can get on this the better.

libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -90,20 +90,30 @@ proc reset*(s: LPChannel) {.async, gcsafe.} =
...
+  #
+  # In all other cases, there should be a data to complete
+  # a read or enough room in the queue/buffer to complete a
+  # push.
+  #
+  # State       | Q Empty  | Q Full
+  # ------------|----------|-------
+  # Reading     | Push Eof | Na
+  # Pushing     | Na       | Pop
+  if not(s.reading and s.pushing):
```
sinkingsugar, line 109: Maybe rather then booleans a single enum would be better at this point since it's almost a state machine.

libp2p/stream/bufferstream.nim
```diff
@@ -71,12 +73,12 @@ method pushData*(s: BufferStream, data: seq[byte]) {.base, async.} =
 
   # We will block here if there is already data queued, until it has been
   # processed
-  inc s.pushing
   try:
+    s.pushing = true
```
sinkingsugar, line 77: Can't say I'm happy to see that much bool flags and logic really...
We are a single threaded polled in order app ideally.
But if works let's do it, I don't see any flaw. Have to test tho properly.

sinkingsugar:
Also remove `testbufferstream` binary file please 😄 it sneaked in.

libp2p/stream/bufferstream.nim
```diff
@@ -63,6 +63,8 @@ method pushData*(s: BufferStream, data: seq[byte]) {.base, async.} =
   ##
   ## `pushTo` will block if the queue is full, thus maintaining backpressure.
   ##
+
+  doAssert(not s.pushing, "Only one concurrent push allowed")
```
arnetheduck, line 67: pusheof needs the same guard

arnetheduck:
rebased without the binaries

PR #445, opened 2020-11-23
update conn peerinfo instead of replacing

dryajov:
Integrating some more unrelated changes from `tcp-limits2`.

### DIFF
--- a/libp2p/protocols/secure/noise.nim
+++ b/libp2p/protocols/secure/noise.nim
@@ -505,13 +505,11 @@ method handshake*(p: Noise, conn: Connection, initiator: bool): Future[SecureCon
           received_key = $remotePubKey
         raise newException(NoiseHandshakeError, "Noise handshake, peer infos don't match! " & $pid & " != " & $conn.peerInfo.peerId)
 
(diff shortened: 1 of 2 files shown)


libp2p/protocols/secure/noise.nim
```diff
@@ -505,11 +505,15 @@ method handshake*(p: Noise, conn: Connection, initiator: bool): Future[SecureCon
           received_key = $remotePubKey
         raise newException(NoiseHandshakeError, "Noise handshake, peer infos don't match! " & $pid & " != " & $conn.peerInfo.peerId)
 
-    let peerInfo =
-      if conn.peerInfo != nil: conn.peerInfo
-      else: PeerInfo.init(remotePubKey)
+    let peerId = PeerID.init(remotePubKey).tryGet()
```
arnetheduck, line 508: why create peerid if it's only used in else branch below?

dryajov: I had a check to see if peerids matched where this was used, which I just removed - leftover.

arnetheduck (approved):

PR #447, opened 2020-11-23
less leak prone cleanup

dryajov:


libp2p/transports/tcptransport.nim
```diff
@@ -77,12 +77,9 @@ proc connHandler*(t: TcpTransport,
       trace "Cleaning up client", addrs = $client.remoteAddress,
                                   conn
 
-      if not(isNil(conn) and conn.closed()):
-        await conn.close()
-
       t.clients[dir].keepItIf( it != client )
-      if not(isNil(client) and client.closed()):
-        await client.closeWait()
+      await allFuturesThrowing(
```
arnetheduck, line 81: double all?

libp2p/protocols/secure/secure.nim
```diff
@@ -77,8 +78,9 @@ proc handleConn*(s: Secure,
 
   proc cleanup() {.async.} =
     try:
-      await conn.join()
-      await sconn.close()
+      await conn.join() or sconn.join()
+      await allFuturesThrowing(
```
arnetheduck, line 82: double all?

libp2p/transports/tcptransport.nim
dryajov: heh, I keep tripping on this one not sure why.

arnetheduck: too many magic helpers

dryajov: true

libp2p/protocols/secure/secure.nim
arnetheduck: I'm a bit worried about these "let's close as much as possible everywhere randomly" kind of approaches, they mask nondeterministic code flow paths that later make it harder to pinpoint races.. the interplay between the closeevent and who "owns" which connection becomes muddy

dryajov: Well, both tcptransport and secureconn have the particularity that they wrap some other resource, so when either the wrapped resource or the wrapper has been closed the other one needs to be closed as well. 

We can add additional checks to see which resource is already closed, to not call `close` unnecessarily,  but I've found that error prone and the checks are already done in the `close()` calls either way. I agree that it's not idea and we need more deterministic close flows, but I'm not sure how to make that happen with non deterministic object lifetimes and an async/event based flow.

PR #453, opened 2020-11-24
get rid of hangs on cleanup

dryajov:


libp2p/protocols/secure/secure.nim
```diff
@@ -78,8 +78,13 @@ proc handleConn*(s: Secure,
 
   proc cleanup() {.async.} =
     try:
-      await conn.join() or sconn.join()
-      await allFuturesThrowing(sconn.close(), conn.close())
+      let futs = @[conn.join(), sconn.join()]
+      await futs[0] or futs[1]
+      for f in futs:
+        if not f.finished: f.cancel # cancel outstanding join()
```
arnetheduck, line 84: why does this hang? shouldn't `close` cause all corresponding `join` to complete?

dryajov: Good question. I've seen NBC use something similar and I tried it here, which seems to work. Not sure why it would be hanging here in the first place and I haven't traced it step by step yet - maybe @cheatfate has a better explanation.

cheatfate: @dryajov i'm confused, usually i'm using `close() and then join()` but i never used `join() and then close()`. So the idea is that some tasks could be signaled about close operation. For example
```nim
proc task1(handle: Resource) {.async.} =
  echo "task1 closing handle"
  await handle.close()
  echo "task1 closed handle"
proc task2(handle: Resource) {.async.} =
  echo "task2 is waiting for the handle to close"
  await handle.join()
  echo "task2 knows that handle is closed"

var handle = newResource()
var fut1 = task1(handle)
var fut2 = task2(handle)

# So if you await for `fut2` you can be sure that `fut1` is also completed.
```

cheatfate: So i think the hang problem here is that somebody could call `cleanup()`, before calling `close()` on `sconn` and `conn`. In such way `sconn.join()` and `conn.join()` could hang await for corresponding `sconn.close()` and `conn.close()`.

dryajov: This is essentially an `onClose` event that triggers when the `.join()` completes which makes sure that the wrapped object is also closed on the wrapper's close and vice versa.

dryajov: Hmm, `cleanup` is a closure https://github.com/status-im/nim-libp2p/pull/453/files#diff-791f6d27abc4d39d7ef3572dcb46cb126db00f8eeb1cac22cf87d2e32f235592R74-R84 and it gets spawned right bellow this code https://github.com/status-im/nim-libp2p/pull/453/files#diff-791f6d27abc4d39d7ef3572dcb46cb126db00f8eeb1cac22cf87d2e32f235592R95-R97

arnetheduck:
but wait.. without understanding _why_ this works, it seems a bit risky to merge? it just adds another possible code flow order?

dryajov:
I’ve tested it pretty well across different environments and it does have a positive effect, without it I get hangs. I do plan to investigate further, but I want to try it out on as many environments as possible. The problem is that tracing the entire execution is very time consuming.

PR #456, opened 2020-11-26
rework peer event to take an initiator flag

dryajov:


libp2p/connmanager.nim
```diff
@@ -42,10 +42,17 @@ type
...
 
-  PeerEvent* {.pure.} = enum
+  PeerEventKind* {.pure.} = enum
     Left,
     Joined
 
+  PeerEvent* = object
+    case kind*: PeerEventKind
+      of PeerEventKind.Joined:
+        initiator*: bool
```
arnetheduck, line 52: so.. same question as usual, why not a `Direction`?

dryajov: I feel that the flag makes more sense here semantically, because it's not about direction, it is who was the initiator of the "session", which is per peer. But it could be `Direction` as well.

PR #458, opened 2020-11-27
add metrics into chronosstream to identify peers agents

sinkingsugar:


dryajov:
So the `agentVersion` field seems to contain string separated by `/`.

```
"Lighthouse/v0.3.4-a60ab4ef/x86_64-linux"
"Lighthouse/v0.3.5-0b556c44/x86_64-linux"
"Lighthouse/v0.3.5-0c2c2cef/x86_64-linux"
"Lighthouse/v0.3.5-3b405f10+/x86_64-linux"
"Lighthouse/v0.3.5-3b405f10/aarch64-linux"
"Lighthouse/v0.3.5-3b405f10/x86_64-linux"
"Lighthouse/v0.3.5-426b3001/x86_64-linux"
"Lighthouse/v0.3.5-48f73b21/x86_64-linux"
"Lighthouse/v0.3.5-5828ff12/x86_64-linux"
"Lighthouse/v0.3.5-59b2247a/x86_64-linux"
"Lighthouse/v0.3.5-7e4ee58/x86_64-linux"
"Lighthouse/v0.3.5-7e4ee587/x86_64-linux"
"Lighthouse/v0.3.5-7e4ee5872/x86_64-linux"
"Lighthouse/v0.3.5-c5e97b9b/aarch64-linux"
"Lighthouse/v0.3.5-c5e97b9b/x86_64-macos"
"Lighthouse/v0.3.5-ccca4707c/x86_64-linux"
"Lighthouse/v0.3.5-d727e55a/x86_64-linux"
"Lighthouse/v1.0.0-c6baa0e/aarch64-linux"
"Lighthouse/v1.0.0-c6baa0e/x86_64-linux"
"Lighthouse/v1.0.0-c6baa0ee+/x86_64-linux"
"Lighthouse/v1.0.0-c6baa0ee/aarch64-linux"
"Lighthouse/v1.0.0-c6baa0ee/x86_64-linux"
"Lighthouse/v1.0.0-c6baa0eed/x86_64-linux"
"Prysm/v1.0.0-beta.2/f361450e8da18033f5458b018e3e6784295f50ca"
"Prysm/v1.0.0-beta.3/f361450e8da18033f5458b018e3e6784295f50ca"
"Prysm/v1.0.0/c7dd33431fb23f2fa62c1617f5c735fd7104ff0c"
"Prysm/{STABLE_GIT_TAG}/71c6164c42dac9874087abf25022e3b4615de5db"
"Prysm/{STABLE_GIT_TAG}/{STABLE_GIT_COMMIT}"
"nim-libp2p/0.0.1"
"teku/teku/v20.11.0+1-geecd64a/linux-x86_64/-privatebuild-openjdk64bitservervm-java-14"
"teku/teku/v20.11.0+2-g47116f7/linux-x86_64/-ubuntu-openjdk64bitservervm-java-11"
"teku/teku/v20.11.0-RC1+13-g68c5e19/linux-x86_64/corretto-java-11"
"teku/teku/v20.11.0-RC2+20-g1ec3a9e/linux-x86_64/-ubuntu-openjdk64bitservervm-java-11"
"teku/teku/v20.11.0/linux-x86_64/oracle_openjdk-java-14"
```

libp2p/stream/chronosstream.nim
```diff
@@ -65,11 +68,27 @@ template withExceptions(body: untyped) =
     # TODO https://github.com/status-im/nim-chronos/pull/99
     raise newLPStreamEOFError()
 
+proc trackPeerIdentity(s: ChronosStream) =
+  if not s.tracked:
+    if not isNil(s.peerInfo) and s.peerInfo.agentVersion.len > 0:
+      let name = s.peerInfo.agentVersion.split(" ")[0]
```
dryajov, line 74: needs to split by `/`

libp2p/stream/chronosstream.nim
```diff
@@ -21,6 +21,9 @@ const
 type
   ChronosStream* = ref object of Connection
     client: StreamTransport
+    tracked: bool
+
+declarePublicCounter(libp2p_peers_identity, "peers identities", labels = ["agent"])
```
dryajov, line 26: We also want to track the total amount of bytes sent/rcvd, this is probably better as a gauge.

libp2p/stream/chronosstream.nim
```diff
@@ -65,11 +68,27 @@ template withExceptions(body: untyped) =
...
+      let name = s.peerInfo.agentVersion.split(" ")[0]
+      libp2p_peers_identity.dec(labelValues = [name])
+      s.tracked = false
+
 method readOnce*(s: ChronosStream, pbytes: pointer, nbytes: int): Future[int] {.async.} =
   if s.atEof:
     raise newLPStreamEOFError()
 
   withExceptions:
+    s.trackPeerIdentity()
```
dryajov, line 90: we need to track to amount of bytes received.

libp2p/stream/chronosstream.nim
```diff
@@ -81,6 +100,8 @@ method write*(s: ChronosStream, msg: seq[byte]) {.async.} =
     return
 
   withExceptions:
+    s.trackPeerIdentity()
```
dryajov, line 103: need to track the amount of bytes sent.

libp2p/stream/chronosstream.nim
dryajov: this should also be moved after the read, same for write, should only trigger after the write happened.

sinkingsugar: good point

libp2p/stream/chronosstream.nim
```diff
@@ -72,26 +75,25 @@ proc trackPeerIdentity(s: ChronosStream) =
   if not s.tracked:
     if not isNil(s.peerInfo) and s.peerInfo.agentVersion.len > 0:
       # / seems a weak "standard" so for now it's reliable
-      let name = s.peerInfo.agentVersion.split("/")[0]
-      libp2p_peers_identity.inc(labelValues = [name])
+      s.shortAgent = s.peerInfo.agentVersion.split("/")[0]
```
dryajov, line 78: this might crash if it returns nil

sinkingsugar: well should not happen as len is checked tho

sinkingsugar: `"no slashes".split("/")[0] == "no slashes"` afaik, will test

sinkingsugar: works

libp2p/stream/chronosstream.nim
```diff
@@ -65,13 +71,29 @@ template withExceptions(body: untyped) =
     # TODO https://github.com/status-im/nim-chronos/pull/99
     raise newLPStreamEOFError()
 
+proc trackPeerIdentity(s: ChronosStream) =
+  if not s.tracked:
+    if not isNil(s.peerInfo) and s.peerInfo.agentVersion.len > 0:
+      # / seems a weak "standard" so for now it's reliable
+      s.shortAgent = s.peerInfo.agentVersion.split("/")[0]
```
arnetheduck, line 78: we need to limit this to a small (5) number of agents, and "others", else remotes can create metric labels

dryajov: Yeah, lets just add a list of known agent names and go with that. This list https://github.com/status-im/nim-libp2p/pull/458#issuecomment-734601883, should have most/all the current agent strings, so let's just take whatever comes before the fist `/` as known agent names and go with that.

libp2p/stream/chronosstream.nim
sinkingsugar: I added a `strdefine` for `KnownLibP2PAgents`, can be now defined in user apps.

dryajov:
Just an overall observation, this is a temporary kludge and metrics in general need some serious cleanup, but this should help tracking and identifying recent issues with traffic spikes and stability.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -872,22 +905,22 @@ method unsubscribePeer*(g: GossipSub, peer: PeerID) =
     # also try to remove from explicit table here
     g.explicit.removePeer(t, pubSubPeer)
 
-    when defined(libp2p_expensive_metrics):
+    if KnownLibP2PTopicsSeq.contains(t):
```
arnetheduck, line 908: let's catch all other subscriptions in `other` or something, so we can track how many topics get subscribed - ie this would allow us to track topic oversubscription

sinkingsugar: I could not find an acceptable way and so decided to skip this.

So what you propose is to anyway increase the counter summing up all other topics? in that case we could make observations based on averages etc? ( A bit like my dirty current metric on master )

arnetheduck: well, the point would be to detect spikes that can be correlated with other issues

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -834,10 +864,13 @@ proc heartbeat(g: GossipSub) {.async.} =
 
       let peers = g.getGossipPeers()
       for peer, control in peers:
-        g.peers.withValue(peer.peerId, pubsubPeer) do:
-          g.send(
-            pubsubPeer[],
-            RPCMsg(control: some(control)))
+        # only ihave from here
+        for ihave in control.ihave:
+          if KnownLibP2PTopicsSeq.contains(ihave.topicID):
```
arnetheduck, line 869: also, the application should probably be able to specify the list of topics - take eth2 for example that has dynamic topics based on fork id

libp2p/protocols/pubsub/gossipsub.nim
sinkingsugar: Sounds good

libp2p/protocols/pubsub/gossipsub.nim
sinkingsugar: you mean at runtime, by the application code?

cos atm for example my build line is:
```
make NIMFLAGS="-d:insecure -d:KnownLibP2PAgents=lighthouse,prysm,teku,nimbus,nim-libp2p -d:KnownLibP2PTopics=/eth2/3b088795/beacon_block/ssz_snappy,/eth2/b5303f2a/beacon_block/ssz_snappy" NODE_PARAMS="--rpc-address:0.0.0.0" pyrmont
```

sinkingsugar:
> hmm , it's better if these metrics are collected in local variables then set at the end - because metrics runs on a different thread, the "intermediate" counts may become observable

I was wondering if that would be relevant or not, so far it worked without issues but it is true that it might happen as I guess it's just interlocked ops.

sinkingsugar:
> uhh, why not just local variables in `rebalanceMesh`?

well, it's called for every topic `proc rebalanceMesh(g: GossipSub, topic: string)` and we update metrics that span many topics

arnetheduck:
> > uhh, why not just local variables in `rebalanceMesh`?
> 
> well, it's called for every topic `proc rebalanceMesh(g: GossipSub, topic: string)` and we update metrics that span many topics

hm, maybe a `TopicStats` that can be passed to `rebalanceMesh(..,, ts: var TopicStats)` then? avoids the ugly resetting

sinkingsugar:
> > > uhh, why not just local variables in `rebalanceMesh`?
> > 
> > 
> > well, it's called for every topic `proc rebalanceMesh(g: GossipSub, topic: string)` and we update metrics that span many topics
> 
> hm, maybe a `TopicStats` that can be passed to `rebalanceMesh(..,, ts: var TopicStats)` then? avoids the ugly resetting

well ok that would work too, I will replace the `metrics: bool` with a `ptr` of this stats state and use that as signal if we need metrics on that rebalance or not as well.

sinkingsugar:
anyone against merging this?

README.md
```diff
@@ -151,12 +151,24 @@ Packages that exist in the original libp2p specs and are under active developmen
...
+#### enable expensive metrics:
 
 ```bash
 nim c -d:libp2p_expensive_metrics some_file.nim
 ```
 
+#### use identify metrics
+
+```bash
+nim c -d:libp2p_agents_metrics -d:KnownLibP2PAgents=nimbus,lighthouse,prysm,teku some_file.nim
```
dryajov, line 163: Is there any way of enabling this metrics at runtime? I'm not sure if they are all compile time, but maybe something like what NBC does with runtime log levels would be useful?

dryajov: we can add an RPC endpoint to NBC to toggle those.

dryajov:
So, I find this additions super useful, but I think we're ignoring the larger issue of why we can't collect them all at the same time, ie, why does the metrics thread breaks with expensive metrics enabled? @stefantalpalaru

stefantalpalaru:
> why does the metrics thread breaks with expensive metrics enabled?

It does? I never saw that.

The problem with a high number of metrics was Prometheus I/O on the AWS fleet, where a single Prometheus server receives metrics from multiple hosts and writes them to the same db.

dryajov:
> It does? I never saw that.

Yeah, definitely at least for me it does on multiple environments. If you enable expensive metrics right now it should show up pretty soon, within the hour usually.

> The problem with a high number of metrics was Prometheus I/O on the AWS fleet, where a single Prometheus server receives metrics from multiple hosts and writes them to the same db.

Yeah, that's well understood and they should not be consumed directly by prometheus (is there any way of making prometheus ignore some metrics?), but it's still useful to be able to request them from the endpoint.

stefantalpalaru:
> is there any way of making prometheus ignore some metrics?

Yes, but it's a complication: https://www.robustperception.io/dropping-metrics-at-scrape-time-with-prometheus

sinkingsugar:
yeah I get your point @dryajov but indeed having control on our side is not bad either imho.
I really want to deploy those on the testnet fleet, can we merge it 😄 ?
@stefantalpalaru if you have time can you give it a quick look maybe? also if this gets into auto bump will fleet receive this flawlessly?

README.md
sinkingsugar: that is true, I will just make them a `var` rather then `const`

dryajov (approved):

PR #460, opened 2020-11-27
mitigate high traffic - drop faulty peers

dryajov:
Doesn't seem to have any adverse effects on overall peer count or stability.


dryajov:
This needs passing a per peer connection parameter, mostly to satisfy tests.

dryajov:
This will decrease the number of physical connections per peer, which should be OK - worth a try, wdyt @Menduist?

Menduist:
Updated on latest unstable to see how it goes.

Don't have a problem with this, though in the future, we should replace this by a "peerid / transport pair" limitation (eg session), since we could have more connections to a single peer on different transports

arnetheduck:
This LGTM and seems harmless, is there anything blocking it?

arnetheduck (dismissed):

arnetheduck:
For context, the switch to 1 connection per peer is motivated by the following background:

Parts of libp2p, for example gossipsub, index peer activity by peer id and rely on message ordering for correctness - in the case of gossipsub, this means for example that prune must follow graft, and unsubscribe must follow subscribe - if these messages are reordered, difficult-to-debug scenarios ensue and it becomes more difficult to reason about the correctness of the mesh - we have had quite a lot of mesh trouble in the past and continue to have it, albeit more rarely - but recently, we've seen happening that a single peer had  5 connections _and_ the node had extensive mesh trouble.

The idea here is to take a step back and use a single connection per peer until the code generally has been reviewed to be capable of handling multiple connections and the message reordering that results. We should also examine the reasons for allowing multiple connections in the first place: there are some hand-wavey arguments about better performance or redundancy, but these should really be substantiated and made more precise: allowing multiple connections per peer significantly increases the surface area for bugs.

Menduist (approved):

arnetheduck:
> Well, as the name of the PR implies, we're dropping faulty peers that in the context of TCP transport and mplex combo, would attempt to open more than one socket connection

the original version of this PR did not block concurrent connections though - it allowed 2 - after we investigated some real-world issues, the conclusion was that we need to limit this to 1 connection until the rest of libp2p has been adapted to better reason about out-of-order messages

PR #461, opened 2020-11-27
move dialing logic to it's own proc to avoid try/finally bugs

dryajov:


libp2p/switch.nim
```diff
@@ -257,6 +257,45 @@ proc upgradeIncoming(s: Switch, incomingConn: Connection) {.async, gcsafe.} = #
...
+        trace "Dialing address", address = $a, peerId
+        let dialed = try:
+            await t.dial(a)
+          except CancelledError as exc:
+            debug "Dialing canceled", msg = exc.msg, peerId
+            raise exc
+          except CatchableError as exc:
+            debug "Dialing failed", msg = exc.msg, peerId
+            libp2p_failed_dials.inc()
+            continue # Try the next address
```
arnetheduck, line 277: won't this continue cause issues?

dryajov: it doesn't have any finally in that block anymore or will this also cause issues?

dryajov: but it's also not needed anymore

dryajov: ok, but the question still stands, would it cause issue or is it unknown at this point?

arnetheduck: right, it was finally blocks that got skipped in this case - maybe works, but a _test_ would confirm it

dryajov: Hmm, yeah a test would be good - I think we can add one now that we have `accept` and control over the incoming flow.

libp2p/switch.nim
```diff
@@ -281,45 +320,9 @@ proc internalConnect(s: Switch,
...
-              raise exc
-
-          doAssert not isNil(upgraded), "connection died after upgradeOutgoing"
-
-          conn = upgraded
-          debug "Dial successful", conn, peerInfo = conn.peerInfo
-          break
+    conn = await s.dialAndUpgrade(peerId, addrs)
   finally:
     if lock.locked():
```
arnetheduck, line 327: so this if shouldn't really be here?

arnetheduck: or maybe it should.. if acquire is cancelled, I guess.. another way to see it is that acquire should not be in the try/catch

dryajov: If `acquire` isn't in the try/catch, then you run the risk of never releasing.

dryajov: that is for a well functioning try/catch.

arnetheduck: acquire doesn't raise (except cancellation) so it can be outside, but it doesn't really matter - the if is needed when the acquire is in the try/catch because on cancellation, it will not be acquired

PR #466, opened 2020-11-30
allow concurrent closeWithEOF

dryajov:
Fixes the double read assert in BufferStream (credits to @cheatfate for suggesting this)


arnetheduck:
pragmatic

PR #468, opened 2020-12-02
Gossip stateful tests

sinkingsugar:
requires `identify-metrics` branch #458

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -830,6 +830,8 @@ proc heartbeat(g: GossipSub) {.async.} =
       let peers = g.getGossipPeers()
       for peer, control in peers:
         g.peers.withValue(peer.peerId, pubsubPeer) do:
+          # only ihave from here
+          libp2p_pubsub_broadcast_ihave.inc(control.ihave.len.int64)
           g.send(
             pubsubPeer[],
             RPCMsg(control: some(control)))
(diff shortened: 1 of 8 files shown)


dryajov:
@sinkingsugar can we split the PR into metrics and testing? 

I would also note that I'm not sure about testing approach in general, I would have thought we'd use a similar approach to the already existing tests in https://github.com/status-im/nim-libp2p/blob/master/tests/pubsub/testgossipinternal.nim?

arnetheduck:
> similar approach to the already existing tests

I find this approach a bit dubious - an alternative idea would be to extract a pure, non-async gossip-and-mesh data structure that could be tested independently and without all the mess - gossipsub would then be a thin async shell around this core.

dryajov:
> > similar approach to the already existing tests
> 
> I find this approach a bit dubious - an alternative idea would be to extract a pure, non-async gossip-and-mesh data structure that could be tested independently and without all the mess - gossipsub would then be a thin async shell around this core.

Yeah, I agree not a bad idea overall. We've lots of tech debt and a cleanup and refactor is in order either way.

sinkingsugar:
I'm a bit surprised about your opinions, this approach is exactly inspired by metrics and the way they showed issues to me.
It's simple and requires no artificial code at all.
Can you please enlighten me on what is wrong with it?

sinkingsugar:
> @sinkingsugar can we split the PR into metrics and testing?

Well it's based on the metrics to not miss locations but yeah will do.

sinkingsugar:
> @sinkingsugar can we split the PR into metrics and testing?
> 
> I would also note that I'm not sure about testing approach in general, I would have thought we'd use a similar approach to the already existing tests in https://github.com/status-im/nim-libp2p/blob/master/tests/pubsub/testgossipinternal.nim?

I don't like that approach so much personally, it's not abstract nor reflective, every time a change in the main code is done, a fix in the test (and there are not tests for the test) is needed.

dryajov:
> I don't like that approach so much personally, it's not abstract nor reflective, every time a change in the main code is done, a fix in the test (and there are not tests for the test) is needed.

Yes, I agree that this isn't ideal either and that's why I completely agree with @arnetheduck that we need a refactor that will most of all benefit testing. By splitting gossip into several components it will allow us to test them in isolation without having to resort to the sort of hacks that this PR and admittedly the internal tests are attempting to workaround.

PR #470, opened 2020-12-03
clean up peerid

arnetheduck:
* fix dangling cstring on error return
* remove some useless inlines
* less mallocs in shortlog
* proc -> func
* rename test

### DIFF
--- a/libp2p/peerid.nim
+++ b/libp2p/peerid.nim
@@ -11,20 +11,19 @@
 
 {.push raises: [Defect].}
 
-import std/hashes
-import chronicles
-import nimcrypto/utils, stew/base58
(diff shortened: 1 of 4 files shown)


arnetheduck:
I'm not sure what the azure issue is - looks unrelated?

dryajov:
Is this just an optimization/cleanup or was there some specific issue this is trying to fix?

dryajov:
The azure issue is unrelated but I've seen this fail sporadically a few times so I need to look into what's causing it.

arnetheduck:
well, like the description says, it fixes a cstring pointing to a local string in a return type - the rest is mostly style fixes and a few minor optimizations

Issue #469, opened 2020-12-03
Race condition in pubsub


arnetheduck:
Using await in a for loop will fail when the subscription sequence changes while looping:

https://github.com/status-im/nim-libp2p/blob/e9d4679059fa8a6bcc2b456f61687a190ddf0df1/libp2p/protocols/pubsub/pubsub.nim#L175

In general, using `await` in a for loop with `state` is broken - the codebase needs to be reviewed for more cases like this

PR #471, opened 2020-12-04
Race condition in pubsub #469

sinkingsugar:


libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -169,10 +169,13 @@ proc getOrCreatePeer*(
 proc handleData*(p: PubSub, topic: string, data: seq[byte]): Future[void] {.async.} =
   if topic notin p.topics: return # Not subscribed
 
-  for h in p.topics[topic].handler:
+  # gather all futures without yielding to scheduler
+  let futures = p.topics[topic].handler.mapIt(it(topic, data))
+  # await each future and check for errors
+  for fut in futures:
```
dryajov, line 175: `allFinished()` - we can inspect each future for errors after, because otherwise we're essentially forcing sequential execution.

sinkingsugar: ok, wonder what happens on cancelled tho...

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -169,15 +169,13 @@ proc getOrCreatePeer*(
...
-    try:
-      await h(topic, data)
-    except CancelledError as exc:
-      raise exc
-    except CatchableError as exc:
-      # Handlers should never raise exceptions
-      warn "Error in topic handler", msg = exc.msg
+  # gather all futures without yielding to scheduler
+  var futs = p.topics[topic].handler.mapIt(it(topic, data))
+  futs = await allFinished(futs)
```
arnetheduck, line 174: `allFinished` here is redundant - this code would be better written as:
```
for f in futs:
  try:
    await f
  except CancelledError: 
   ...
```

using `fut.failed` like in the PR also changes the cancellation behaviour

sinkingsugar: @arnetheduck well https://github.com/status-im/nim-libp2p/pull/471#discussion_r536246637 😄

arnetheduck: so: `p.topics[topic].handler.mapIt(it(topic, data))` _starts_ all work so that it can proceed concurrently - `allFinished` adds an extra future that completes when _all_ futures have finished but is ultimately indistinguishable from simply iterating over the started futures and awaiting them serially - there's no advantage to calling `allFinished` unless the work that is being performed has dependencies and you can't start awaiting any of the futures until all have finished

libp2p/protocols/pubsub/pubsub.nim
arnetheduck: > otherwise we're essentially forcing sequential execution.

not quite, the execution remains concurrent even with the for loop - `allFinished` ensures that all work has completed before iterating over the finished futures so this ends up being _less_ concurrent (because the work in the for loop cannot start "early") - arguably, it's sometimes better to be less concurrent because there  are fewer opportunities for bugs.

also, when one future is cancelled, it's likely _all_ remaining futures should be cancelled, though this is tricky - @cheatfate might have some better insights here

libp2p/protocols/pubsub/pubsub.nim
cheatfate: Advantage in number of async steps to complete, so if you will call `await f` N times you will finish in N*2 async steps, if you call `allFinished(futs)` you can finish in 2 steps.

cheatfate: So if you able to replace
```nim
for fut in futs:
  await fut
```
do it with `allFinished()`, `allFutures()`.

sinkingsugar: Yep in fact I liked my original approach but we keep bouncing this topic a lot 😄 so I wanted to avoid any discussion and used allfinished altho I wasn't really happy about it.
Of course @cheatfate is right about the numbers of steps. but cancellation worries me with allfinished.

cheatfate: Old approach do not cares about cancellation too, because if cancellation will happen only one (current) Future will be cancelled, all other Futures in array will not be cancelled...

sinkingsugar: what is the best approach to deal with cancellations then in this case? any suggestions?

libp2p/protocols/pubsub/pubsub.nim
dryajov: Well, yeah we spawn the futures concurrently in the `for` above and we're only awaiting sequentially so I guess it's a fair correction.

dryajov (approved):

libp2p/protocols/pubsub/pubsub.nim
cheatfate: If its important to notify all the Futures in `futs` - do it, if you do not care - ignore it.
```nim
  # gather all futures without yielding to scheduler
  trace "triggering handler", topicID = topic
  var futs = p.topics[topic].handler.mapIt(it(topic, data))
  try:
    futs = await allFinished(futs)
  except CancelledError:
    for fut in futs:
      if not(fut.finished):
        fut.cancel()
```

sinkingsugar: @cheatfate Ah.. so the cancellation would arrive on `allFinished`? this makes it much easier... would that work even if a inner future is cancelled?

PR #473, opened 2020-12-07
streamline socket read/write hot path

arnetheduck:
This avoids some unnecessary memory copying on the hot path of noise /
mplex, as well as getting rid of a few futures - profiling shows that
this is one of the main culprits of small memory allocations, which
makes sense - this is where gossip fan-out happens.

* fewer futures (and corresponding closures) when sending lpchannel
messages
* avoid data copies when encrypting and framing noise messages
* avoid copying tuple when reading noise data (poor c codegen)
* fix setting eof flag in secure read

### DIFF
--- a/libp2p/protocols/secure/noise.nim
+++ b/libp2p/protocols/secure/noise.nim
@@ -427,34 +427,47 @@ method readMessage*(sconn: NoiseConnection): Future[seq[byte]] {.async.} =
       dumpMessage(sconn, FlowDirection.Incoming, [])
     trace "Received 0-length message", sconn
 
+
+proc encryptFrame(
+    sconn: NoiseConnection, cipherFrame: var openArray[byte], src: openArray[byte]) =
+  # Frame consists of length + cipher data + tag
+  doAssert src.len <= MaxPlainSize
+  doAssert cipherFrame.len == 2 + src.len + sizeof(ChaChaPolyTag)
+
+  cipherFrame[0..<2] = toBytesBE(uint16(src.len + sizeof(ChaChaPolyTag)))
+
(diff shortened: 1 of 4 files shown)


libp2p/protocols/secure/noise.nim
```diff
@@ -427,10 +437,24 @@ method write*(sconn: NoiseConnection, message: seq[byte]): Future[void] {.async.
...
 
-    await sconn.stream.writeFrame(cipher)
+    var
+      cipherFrame =
+        newSeqUninitialized[byte](2 + chunkSize + sizeof(ChaChaPolyTag))
+
+    # Frame consists of length + cipher data + tag
+    cipherFrame[0..<2] = toBytesBE(uint16(chunkSize + sizeof(ChaChaPolyTag)))
+
+    copyMem(addr cipherFrame[2], unsafeAddr message[offset], chunkSize)
```
zah, line 448: What's the purpose of the `offset` variable here? Any value besides 0 would be unsafe in this `copyMem` call.

arnetheduck: `offset` marks the start of the chunk within `message` - it's increased by `chunkSize` on every loop iteration and `chunkSize` is capped by `message.len` - why is it unsafe?

I'm slightly uncomfortable with `copyMem` here, but unfortunately, the std lib does an extremely slow byte-by-byte iteration here that turns up in benchmarking

zah (approved):

libp2p/protocols/secure/noise.nim
zah: Apologies, I've failed to notice some of the hidden lines in the diff.

PR #474, opened 2020-12-09
Use Travis only for ARM64

mratsim:
See https://github.com/status-im/nimbus-eth2/pull/2167

Trying to revivie ARM64 on Travis and we will reserve our Travis credits for that.


arnetheduck:
@mratsim was the bearssl submodule intentionally added here?

mratsim:
arg, no, must have been leftover when I git pulled.

PR #475, opened 2020-12-10
Concurrent accepts

dryajov:


libp2p/transports/tcptransport.nim
```diff
@@ -142,6 +142,8 @@ method stop*(t: TcpTransport) {.async, gcsafe.} =
   ## stop the transport
   ##
 
+  t.running = false # mark stopped as soon as possible
```
arnetheduck, line 145: `if t.running: return`

libp2p/switch.nim
```diff
@@ -425,17 +428,18 @@ proc accept(s: Switch, transport: Transport) {.async.} = # noraises
...
         # file-handle limit (or another non-fatal error),
         # we can get one on the next try, but we should
         # be careful to not end up in a thigh loop that
         # will starve the main event loop, thus we sleep
         # here before retrying.
         await sleepAsync(100.millis) # TODO: should be configurable?
+        continue
+
+      debug "Accepted an incoming connection", conn
+      await s.upgradeIncoming(conn) or conn.upgraded # wait for connection to by upgraded
```
arnetheduck, line 442: how can the connection be upgraded without upgradeIncoming completing? why are both futures needed here?

dryajov: `upgradeIncoming` doesn't complete until the connection exists because it awaits the multistream handler which dispatches calls to user protocol handlers here - https://github.com/status-im/nim-libp2p/blob/c788a6a3c0ca614be4c67cc566aee6877da0250b/libp2p/switch.nim#L356, that's why I'm adding another event to the connection because the upgrade completes in the `muxerHandler` here https://github.com/status-im/nim-libp2p/blob/c788a6a3c0ca614be4c67cc566aee6877da0250b/libp2p/switch.nim#L624 (when added to the `ConnectionManager`).

arnetheduck: seems easier to fail `conn.upgraded` with an exception then, `iff` upgrade fails in such a way that `upgradeIncoming` ends before `upgraded`

arnetheduck: basically, we should _never_ leave futures around that don't finish -  that's a prime source of async leaks

arnetheduck:
hm, while it makes sense to have concurrent upgrades, I'm not convinced it's necessary to have concurrent accepts actually - getting a socket is almost free - as long as it hands off the rest of the upgrade, one "task" is enough

libp2p/switch.nim
dryajov: Sure, cleaning up/canceling the `.upgraded` future on a failed `upgradeIncoming` sounds like a reasonable thing to do.

dryajov:
> hm, while it makes sense to have concurrent upgrades, I'm not convinced it's necessary to have concurrent accepts actually - getting a socket is almost free - as long as it hands off the rest of the upgrade, one "task" is enough

Yeah I mostly agree that concurrent accepts are a bit tangential to concurrent upgrades and we already have concurrent upgrades in our latests master.  This PR however attempts to throttle the amount of concurrent/inflight upgrades in addition (or with the help of) concurrent accepts. See below for a more elaborate answer.

The current behaviour is a bit unfair in how it handles incoming vs outgoing connections

- For dials/outgoing connections, the dialer needs to wait for the connection to upgrade before it is returned to the caller, which in turn means that there are only as many concurrent dials as there are workers, as it is the case with NBC for example. 
- Incoming connections on the other hand, don't have any rate limiting/bounds and will continue accepting and concurrently upgrading connections as they come in until the upper connection limit is reached. This translates in having several times more incoming connections than outgoing ones, which gives us less control over which peers we're talking to.

This PR tries to address this issues by emulating a similar behavior for incoming connections as for outgoing ones. The idea is that there would be a number of concurrent accept workers which would await for `upgradeIncoming` or `conn.upgraded` to complete, which would signal that the connection either finished or failed upgrading, at which point the worker would resume and accept a new connection.  

However it seems like chronos doesn't support concurrent accept calls as it is right now, which is either a bug or a design limitation. Of course there are other ways of achieving this behaviour and it's something to potentially explore now that I've ran into issues with chronos, but it would be good to understand what the particular issue is.

In any case, in its current state, this PR requires either figuring out what the issue in chronos is or reworking it to use a different approach to throttling connection upgrades.

libp2p/switch.nim
dryajov: I just though that this `or` construct isn't necessary at all as I can just:

```nim
asyncSpawn s.upgradeIncoming(conn)
await conn.upgraded
``` 

and simply complete/fail the `conn.upgraded` if upgrade failed.

dryajov:
So as it turns out, chronos does not support concurrent accepts so an alternative solution is required. 

As a side note, it would be great to add an assert to chronos to prevent multiple accepts from being issued on the same socket @cheatfate.

dryajov:
closing in favor of #489

PR #476, opened 2020-12-10
Gossip improvements

sinkingsugar:
* If score weights are 0.0 avoid computing scores
* remove `{.async.}` from rebalance mesh

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -660,6 +660,7 @@ func `/`(a, b: Duration): float64 =
 
 proc colocationFactor(g: GossipSub, peer: PubSubPeer): float64 =
   if peer.connections.len == 0:
+    debug "colocationFactor, no connections", peer
     0.0
   else:
     let
@@ -667,6 +668,7 @@ proc colocationFactor(g: GossipSub, peer: PubSubPeer): float64 =
       ipPeers = g.peersInIP.getOrDefault(address)
       len = ipPeers.len.float64
     if len > g.parameters.ipColocationFactorThreshold:
+      debug "colocationFactor over threshold", peer, address, len
       let over = len - g.parameters.ipColocationFactorThreshold
       over * over
     else:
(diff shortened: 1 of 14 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -529,6 +529,7 @@ proc rebalanceMesh(g: GossipSub, topic: string) {.async.} =
 
       trace "pruning", prunes = prunes.len
       for peer in prunes:
+        trace "pruning peer on rabalance", peer, score = peer.score
```
arnetheduck, line 532:
```suggestion
        trace "pruning peer on rebalance", peer, score = peer.score
```

sinkingsugar:
Will remove gossip 1.0 in another PR I think,
since this one is a bit urgent a this point due to mcache fix

sinkingsugar:
Coverage is poor cos scoring is turned off by default, will address that next.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -942,9 +950,9 @@ proc punishPeer(g: GossipSub, peer: PubSubPeer, topics: seq[string]) =
     # ensure we init a new topic if unknown
     let _ = g.topicParams.mgetOrPut(t, TopicParams.init())
     # update stats
-    var tstats = g.peerStats[peer].topicInfos.getOrDefault(t)
-    tstats.invalidMessageDeliveries += 1
-    g.peerStats[peer].topicInfos[t] = tstats
+    g.peerStats.withValue(peer, stats):
```
arnetheduck, line 953: unlike the previous version, this will not add a _new_ entry to the table if it's missing.. intended?

sinkingsugar: yep it's fine

sinkingsugar: actually I see your point for this specific line, let me review it.

sinkingsugar: fixed, nice catch, the table was populated on graft but it was very possible that we received and we definitely needed to score prior to that

arnetheduck (approved):

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -318,7 +318,7 @@ method onNewPeer(g: GossipSub, peer: PubSubPeer) =
     discard
 
 proc grafted(g: GossipSub, p: PubSubPeer, topic: string) =
-  g.peerStats.withValue(p, stats) do:
+  g.peerStats.withValue(p, stats):
```
dryajov, line 321: Interesting, does removing the `do` has any noticeable effect here?

arnetheduck: `do` is needed when passing two blocks to withValue - otherwise it makes no difference (this is likely a copy-paste leftover)

PR #480, opened 2020-12-11
first pass, use results for Cid module

sinkingsugar:


libp2p/cid.nim
```diff
@@ -66,66 +69,76 @@ proc initCidCodeTable(): Table[int, MultiCodec] {.compileTime.} =
...
-      cid.cidver = CIDv0
-      cid.mcodec = multiCodec("dag-pb")
-      cid.hpos = 0
-      cid.data = initVBuffer(data)
-      result = CidStatus.Success
-  if cid.cidver == CIDvIncorrect:
+template orError*(exp: untyped, err: untyped): untyped =
+  (exp.mapErr do (_: auto) -> auto: err)
+
+proc decode(data: openarray[byte], cid: var Cid): Result[void, CidError] =
```
arnetheduck, line 75: why not return the cid instead?

sinkingsugar: Sure why not.

arnetheduck (approved):

PR #481, opened 2020-12-14
use stew/leb128

arnetheduck:
* avoids multiple reallocations in writeLp
* simplifies varint implementation
* remove vbuffer.length (unused)

depends on https://github.com/status-im/nim-stew/pull/66


dryajov:
@arnetheduck build is failing on window due to missing `stew` version?

PR #487, opened 2020-12-15
avoid creating prune message unless we're pruning

arnetheduck:


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -812,20 +812,21 @@ proc heartbeat(g: GossipSub) {.async.} =
         let gossipPeers = g.gossipsub.getOrDefault(t)
         # this will be changed by rebalance but does not matter
         totalMeshPeers += meshPeers.len
-        totalGossipPeers += gossipPeers.len
+        totalGossipPeers += g.gossipsub.peers(t)
```
dryajov, line 815: General remark - `PeerTable` uses the `table[topic]` syntax, and even tho we do check for the existence of the key, `table.getOrDefault()` is safer in this case. We should change `PeerTable` to use `getOrDefault()` in any case.

dryajov (approved):
LGTM

libp2p/protocols/pubsub/gossipsub.nim
arnetheduck: I'm not sure how it's safer, _since_ it checks existance? getordefault forces the hash set to be allocated every time even if there are entries, it's not.. ideal.

dryajov: I agree its not ideal performance wise, but it's _safer_ precisely because it will return a default object and not throw on missing key. It is unfortunate that we've had to use the `getOrDefault()` workaround to cover for all the cases, but from past experience it's the one construct that has shown to be the safest.

arnetheduck: it's not safer - it masks real issues - it has crashed less when the `in` check was not done, or it was done then an `await` happened and the table changed while waiting

sinkingsugar:
I think this is also changed in the metrics branch but let's give quicker priority to this merge since metrics still need some more work

PR #488, opened 2020-12-15
track stew head

dryajov:


arnetheduck:
I suspect this won't make any difference, nimble is too broken to track head correctly - as soon as things get merged, it just grabs the latest commit if you start with a clean nimcache

arnetheduck:
the best way to track stew is to simply depend on `stew`

PR #489, opened 2020-12-16
Concurrent upgrades

dryajov:
This PR adds inflight connection throttling for incoming connections. 

It includes:

- Adds an `upgraded` event to connections to signal when it has been upgraded/failed
  - This is needed for incoming connections because the initiator controls the upgrade flow
- Incoming connection upgrade throttling to match how dial works and prevent them from taking up all the available slots

### DIFF
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -240,14 +240,16 @@ proc cleanupConn(c: ConnManager, conn: Connection) {.async.} =
 proc onConnUpgraded(c: ConnManager, conn: Connection) {.async.} =
   try:
     trace "Triggering connect events", conn
+    # NOTE: make sure the upgrade event
+    # happens *before* any other events
+    # are triggered
+    conn.upgraded.complete()
     let peerId = conn.peerInfo.peerId
     await c.triggerPeerEvents(
       peerId, PeerEvent(kind: PeerEventKind.Joined, initiator: conn.dir == Direction.Out))
 
     await c.triggerConnEvent(
       peerId, ConnEvent(kind: ConnEventKind.Connected, incoming: conn.dir == Direction.In))
-
-    conn.upgraded.complete()
   except CatchableError as exc:
     # This is top-level procedure which will work as separate task, so it
     # do not need to propagate CancelledError and should handle other errors
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -19,7 +19,8 @@ logScope:
 
 declareGauge(libp2p_peers, "total connected peers")
 
-const MaxConnectionsPerPeer = 5
+const
+  MaxConnectionsPerPeer = 5
 
 type
   TooManyConnections* = object of CatchableError
@@ -236,14 +237,17 @@ proc cleanupConn(c: ConnManager, conn: Connection) {.async.} =
 
   trace "Connection cleaned up", conn
 
-proc peerStartup(c: ConnManager, conn: Connection) {.async.} =
+proc onConnUpgraded(c: ConnManager, conn: Connection) {.async.} =
   try:
     trace "Triggering connect events", conn
     let peerId = conn.peerInfo.peerId
     await c.triggerPeerEvents(
       peerId, PeerEvent(kind: PeerEventKind.Joined, initiator: conn.dir == Direction.Out))
+
     await c.triggerConnEvent(
       peerId, ConnEvent(kind: ConnEventKind.Connected, incoming: conn.dir == Direction.In))
+
+    conn.upgraded.complete()
   except CatchableError as exc:
     # This is top-level procedure which will work as separate task, so it
     # do not need to propagate CancelledError and should handle other errors
@@ -384,7 +388,7 @@ proc storeMuxer*(c: ConnManager,
   trace "Stored muxer",
     muxer, handle = not handle.isNil, connections = c.conns.len
 
-  asyncSpawn c.peerStartup(muxer.connection)
+  asyncSpawn c.onConnUpgraded(muxer.connection)
 
 proc getStream*(c: ConnManager,
                      peerId: PeerID,
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -240,7 +240,10 @@ proc cleanupConn(c: ConnManager, conn: Connection) {.async.} =
 proc onConnUpgraded(c: ConnManager, conn: Connection) {.async.} =
   try:
     trace "Triggering connect events", conn
+    doAssert(not isNil(conn.upgraded),
+      "The `upgraded` event hasn't been properly initialized!")
     conn.upgraded.complete()
+
     let peerId = conn.peerInfo.peerId
     await c.triggerPeerEvents(
       peerId, PeerEvent(kind: PeerEventKind.Joined, initiator: conn.dir == Direction.Out))
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -240,9 +240,6 @@ proc cleanupConn(c: ConnManager, conn: Connection) {.async.} =
 proc onConnUpgraded(c: ConnManager, conn: Connection) {.async.} =
   try:
     trace "Triggering connect events", conn
-    # NOTE: make sure the upgrade event
-    # happens *before* any other events
-    # are triggered
     conn.upgraded.complete()
     let peerId = conn.peerInfo.peerId
     await c.triggerPeerEvents(
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -437,8 +437,8 @@ proc accept(s: Switch, transport: Transport) {.async.} = # noraises
       # remember to always release the slot when
       # the upgrade succeeds or fails, this is
       # currently done by the `upgradeMonitor`
-      await upgrades.acquire()
-      conn = await transport.accept()
+      await upgrades.acquire()        # first wait for an upgrade slot to become available
+      conn = await transport.accept() # next attempt to get a connection
       if isNil(conn):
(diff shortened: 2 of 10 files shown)


libp2p/switch.nim
```diff
@@ -417,30 +426,48 @@ proc mount*[T: LPProtocol](s: Switch, proto: T, matcher: Matcher = nil) {.gcsafe
...
-      else:
+      if isNil(conn):
         # A nil connection means that we might have hit a
         # file-handle limit (or another non-fatal error),
         # we can get one on the next try, but we should
         # be careful to not end up in a thigh loop that
         # will starve the main event loop, thus we sleep
         # here before retrying.
+        trace "Unable to get a connection, sleeping"
         await sleepAsync(100.millis) # TODO: should be configurable?
```
arnetheduck, line 447: try/catch because may be cancelled, in which case semaphore must be released - no finally in loop though

libp2p/switch.nim
```diff
@@ -417,30 +426,48 @@ proc mount*[T: LPProtocol](s: Switch, proto: T, matcher: Matcher = nil) {.gcsafe
...
+        upgrades.release()
+        continue
+
+      debug "Accepted an incoming connection", conn
+      proc upgradeMonitor() {.async.} =
+        ## monitor connection for upgrades
+        ##
+        try:
+          await conn.upgraded.wait(30.seconds) # wait for connection to by upgraded
```
arnetheduck, line 456: so this implicitly becomes an upgrade timeout? like this overlaps with the general connection timeout monitor, creating two competing timeout mechanisms

libp2p/switch.nim
dryajov: Hmm, isn't canceling only possible on the entire `accept`?

libp2p/switch.nim
```diff
@@ -417,30 +426,48 @@ proc mount*[T: LPProtocol](s: Switch, proto: T, matcher: Matcher = nil) {.gcsafe
...
+          trace "Connection upgrade succeeded"
+        except CatchableError as exc:
+          trace "Exception awaiting connection upgrade", exc = exc.msg, conn
+          if not(isNil(conn)) and not(conn.closed):
```
arnetheduck, line 460: `conn` can't be nil here and doesn't need closed check really (close does that)

libp2p/switch.nim
dryajov: There is no upgrade timeout right now, so yes this is an implicit upgrade timeout.

libp2p/switch.nim
arnetheduck: yeah, but you don't want to leave the semaphore hanging, ever, same as you don't want to leave other unfinished futures around

dryajov: Those are canceled as well tho? There is (should be?) teardown process where all enclosed futures get cancelled as well.

arnetheduck: I meant that the semaphore will not be released - it may be harmless in this particular case but it's a ticking bomb not to release the semaphore because the code will get copy-pasted and refactored to places where it matters - the point is that with semaphores and locks, you never ever want to leave them in an acquired state no matter the flow - I mentioned the futures because they're similar from that point of view

libp2p/switch.nim
arnetheduck: so it looks like a feature that should at least be shared between incoming and outgoing upgrades?

dryajov: Well, this is actually more about who initiates the flow. In the case of a `dial/connect`, "we" are the initiator and we can make decisions when to stop it - either due to a timeout or some other reasons. In other words "we" control the flow of the upgrade. For incoming connections, we're not the initiator, the remote is, we only respond to it's upgrade requests, so this is a guard to prevent it from hijacking a connection in case it "hangs".

arnetheduck: fair enough, deserves a comment

libp2p/switch.nim
dryajov: yeah, I guess it's a good idea to release on cancellation either way

libp2p/switch.nim
```diff
@@ -417,30 +426,60 @@ proc mount*[T: LPProtocol](s: Switch, proto: T, matcher: Matcher = nil) {.gcsafe
...
+        ##
+        try:
+          # Since we don't control the flow of the
+          # upgrade, this timeout guarantees that a
+          # "hanged" remote doesn't hold the upgrade
+          # forever
+          await conn.upgraded.wait(30.seconds) # wait for connection to be upgraded
+          trace "Connection upgrade succeeded"
+        except CatchableError as exc:
+          if not isNil(conn): # for some reason, this can be nil
```
dryajov, line 466: This `nil` is puzzling me and it also happens almost immediately so quite reproducible, and yet there isn't anything that would `nillify` the connection explicitly and the closure scope should still be holding onto it - possible GC bug?

arnetheduck: eeh... right, nim has some pretty weird closure rules - wonder for example if a new instance is allocated on every iteration - @zah?

arnetheduck: the more reliable way to write things is to not use a closure and pass the variable in explicitly, then you don't get hit by gotchas

libp2p/connmanager.nim
```diff
@@ -236,12 +237,17 @@ proc cleanupConn(c: ConnManager, conn: Connection) {.async.} =
 
   trace "Connection cleaned up", conn
 
-proc peerStartup(c: ConnManager, conn: Connection) {.async.} =
+proc onConnUpgraded(c: ConnManager, conn: Connection) {.async.} =
   try:
     trace "Triggering connect events", conn
+    # NOTE: make sure the upgrade event
+    # happens *before* any other events
+    # are triggered
```
arnetheduck, line 245: why is this important? ie anything that's waiting on this future may run either before or after the peer event trigger below

dryajov:
Because it makes it less likely for this event to interleave with a
possible disconnect triggered from the event handlers.

arnetheduck:
> Because it makes it less likely for this event to interleave with a possible disconnect triggered from the event handlers.
> […](#)


less likely doesn't really matter though - as in, it just makes the bugs less frequent / harder to detect - the code must be written in such a way that the order _doesn't_ matter and the comment would do well to point that out, rather than advertising a false sense of order

dryajov:
> > Because it makes it less likely for this event to interleave with a possible disconnect triggered from the event handlers.
> > […](#)
> 
> less likely doesn't really matter though - as in, it just makes the bugs less frequent / harder to detect - the code must be written in such a way that the order _doesn't_ matter and the comment would do well to point that out, rather than advertising a false sense of order

It definitely doesn't guarantee anything but it reduces the likelihood of triggering error paths, but other than that and being cautious, it doesn't have an effect.

libp2p/connmanager.nim
```diff
@@ -236,12 +237,14 @@ proc cleanupConn(c: ConnManager, conn: Connection) {.async.} =
 
   trace "Connection cleaned up", conn
 
-proc peerStartup(c: ConnManager, conn: Connection) {.async.} =
+proc onConnUpgraded(c: ConnManager, conn: Connection) {.async.} =
   try:
     trace "Triggering connect events", conn
+    conn.upgraded.complete()
```
arnetheduck, line 243: doesn't this need a nil check?

dryajov: Shouldn't it gets initialized in `Connection`, but we can add one for good measure.

dryajov:
@arnetheduck anything else you think needs to be done here or you're fine merging it now?

libp2p/switch.nim
```diff
@@ -478,6 +510,14 @@ proc stop*(s: Switch) {.async.} =
     except CatchableError as exc:
       warn "error cleaning up transports", msg = exc.msg
 
+  let stopped = await allFuturesThrowing(s.acceptFuts)
```
arnetheduck, line 513: so.. the hope is that none of the accept futures raises? because if any of them did raise, we'll not finish the stop logic here?

dryajov: If it raises it will propagate up to the caller in which case the caller and the caller will have to decide what can be done about it, but we can attempt to handle it here as well given that there probably isn't much else to do.

arnetheduck: the point is more that the caller _cannot_ reason about what to do if `stop` fails, typically - you can't describe what actions are appropriate because the information is internal to the switch module - specially after stop _partially_ has succeeded (some of the futures may have completed and some not) - this is often the case with these kinds of exceptions that randomly bubble up from composite operations - this is also why raising an exception in a for loop rarely makes sense: there's no way for the caller to reason about the partial success - this is something we keep coming back to, and that has been the source of numerous bugs in libp2p: aborting operations mid-way often means some sort of cleanup must happen - this is specially true of composite operations and it is exceedingly rare for these not to require local error handling and cleanup - to the point that it should be motivated explicitly.

libp2p/switch.nim
```diff
@@ -510,8 +510,12 @@ proc stop*(s: Switch) {.async.} =
     except CatchableError as exc:
       warn "error cleaning up transports", msg = exc.msg
 
-  let stopped = await allFuturesThrowing(s.acceptFuts)
-    .withTimeout(1.seconds)
+  var stopped: bool
```
arnetheduck, line 513:
```
var stopped = try: 
    await all... 
  except ...: 
    trace "..."
    false
```

arnetheduck: this style makes the compiler ensure that all branches get an explicit value for `stopped`

libp2p/switch.nim
dryajov: Cool

arnetheduck (approved):

PR #490, opened 2020-12-17
Gossipsub refactor

sinkingsugar:
Remove 1.0
Re-enable score
Avoid storing pubsubpeers in score tables


sinkingsugar:
> this comes out wrong though - we'll first send unsubscribe then prune which is the wrong order - ideally it would prune/unsub in the same message

Ah you are right! let's just forget the method and write that broadcast in gossipsub...

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -682,16 +684,20 @@ proc updateScores(g: GossipSub) = # avoid async
   trace "updating scores", peers = g.peers.len
 
   let now = Moment.now()
-  var evicting: seq[PubSubPeer]
+  var evicting: seq[PeerID]
 
-  for peer, stats in g.peerStats.mpairs:
+  for peerId, stats in g.peerStats.mpairs:
+    let peer = g.peers.getOrDefault(peerId)
```
arnetheduck, line 690: this means we'll never evict disconnected peers from the stats table

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -1012,7 +1021,7 @@ proc handleGraft(g: GossipSub,
 
       continue
 
-    if peer notin g.peerStats:
+    if peer.peerId notin g.peerStats:
       g.onNewPeer(peer)
```
arnetheduck, line 1025: this looks odd - `onNewPeer` becomes ambiguous as to whether it's a new peer subscribing or new to stats - would be better to move this to a `initPeerStats()` function that gets called from both `onNewPeer` and here

libp2p/protocols/pubsub/gossipsub.nim
sinkingsugar: why? They expire, check a few lines under.

PR #497, opened 2020-12-23
Gossip subscription improvements

sinkingsugar:


sinkingsugar:
It's kinda ready but made it into draft cos I have some issues with nimbus I wanna investigate first.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -1435,3 +1461,5 @@ method initPubSub*(g: GossipSub) =
   g.lastFanoutPubSub = initTable[string, Moment]()  # last publish time for fanout topics
   g.gossip = initTable[string, seq[ControlIHave]]() # pending gossip
   g.control = initTable[string, ControlMessage]()   # pending control messages
+  var rng = newRng()
+  g.randomBytes = Curve25519Key.random(rng[]).getBytes()
```
arnetheduck, line 1465: just use the RNG directly? no need to go via curve25519...

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -55,6 +55,9 @@ type
   MsgIdProvider* =
     proc(m: Message): MessageID {.noSideEffect, raises: [Defect], nimcall, gcsafe.}
 
+  SubscriptionValidator* =
+    proc(topic: string): bool {.noSideEffect, raises: [Defect], nimcall, gcsafe.}
```
arnetheduck, line 59: hm, this is not sufficient to implement a validator that limits the number of subscriptions - it should probably be called for both sub and unsub, with a bool.. the question then would be what to do with failed unsub validations , hm

libp2p/protocols/pubsub/gossipsub.nim
sinkingsugar: forgot about it, that was supposed to be temporary indeed!

libp2p/protocols/pubsub/pubsub.nim
sinkingsugar: well currently it is called indeed for sub and unsubs btw. cos it's invoked in `subscribeTopic` that deals with both cases.

sinkingsugar: If the topic filters do not change during the app life-time we should have no real issue tho I think.
This filter should be used to filter really unknown topics.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -1126,7 +1144,7 @@ method rpcHandler*(g: GossipSub,
   for msg in rpcMsg.messages:                         # for every message
     let msgId = g.msgIdProvider(msg)
 
-    if g.seen.put(msgId):
+    if g.seen.put(msgId & g.randomBytes):
```
arnetheduck, line 1147: document here why the randombytes are added

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -171,15 +175,15 @@ proc handleData*(p: PubSub, topic: string, data: seq[byte]): Future[void] {.asyn
 
   # gather all futures without yielding to scheduler
   var futs = p.topics[topic].handler.mapIt(it(topic, data))
-  
+
```
arnetheduck, line 178: empty lines

arnetheduck: ah right, this removes empty lines :+1:

libp2p/protocols/pubsub/pubsub.nim
dryajov: is triggering this on `unsubscribe` really needed - even if a peer sends us an unknown topic to unsubscribe from, is it really a fault?

sinkingsugar: Well that is true, but mostly for coherence, logging and scoring (negatively).

arnetheduck: well, it covers the current use case but I'd say more use cases are covered if we specify the exact operation - or indeed, we make it a subscribe-only thing in which case we should add a few more "basic" limits to libp2p itself, including maximum subscription counts - or a score penalty for each subscription (which would be "regained" by using the subscription in a useful way such as gossipping useful message and getting score bonuses)

sinkingsugar:
merging master seems to have started causing issues...
For the rest this is good to go, minus the pending conversation.

PR #499, opened 2021-01-07
Address Book POC implementation

### LINKED ISSUE #498: Proof of Concept: Peer Store with Address Book
### Problem

Several issues, in both nim-waku and nim-libp2p, express the need for peer and connection management in the Nim implementation of libp2p. These efforts may benefit from a simple and standalone Peer Store where information about known peers can be kept and managed. This is closely related to the Peer Store implementations in [go](https://github.com/libp2p/go-libp2p-peerstore) and [js](https://github.com/libp2p/js-libp2p/tree/master/src/peer-store).

Some functions of a Peer Store would be to:
- provide a register of all "known" peers in different "books"/maps, including addresses, public keys, supported protocols, performance metrics and connection metadata. Sophisticated peer scoring and ageing mechanisms can be added over time.
- provide API calls to add/remove/modify/get peers in an agnostic way (whether via discovery/bootstrapping/file stream/explicit external API call)
- allow interested users to register event listeners to be updated about events such as the modification/removal/addition of peers, changing scores, etc.
- eventually add some persistence over restarts

Peer management and peer discovery can then be built on top of this store. For example, the `identify`-protocol can plug into this store to update the keys of identified peers, a manager could ensure peers for the `WakuRelay`-protocol remain connected, and a DHT discovery process could add addresses to the registry.

### POC

This issue suggests a POC implementation of a Peer Store that consists of only an Address Book where the multiaddresses of peers can be stored.

#### Acceptance Criteria

- Implements a Peer Store with an Address Book mapping peers to their multiaddresses
- Exposes an API to add, delete, get and modify the multiaddresses of known peers
- Allows clients to register a listener/handler from where they can be notified of any changes to known addresses.

### Linked issues:

https://github.com/status-im/nim-libp2p/issues/438, https://github.com/status-im/nim-libp2p/issues/140, https://github.com/status-im/nim-waku/issues/265, https://github.com/status-im/nim-waku/issues/266

cc @oskarth @dryajov

jm-clius:
This PR closes #498

It introduces a POC version of a Peer Store with Address Book, based on similar implementations in [go](https://github.com/libp2p/go-libp2p-peerstore) and [js](https://github.com/libp2p/js-libp2p/tree/master/src/peer-store).

The POC:
- Implements a simple AddressBook API to add, get, set and delete multiaddresses for peers
- Implements a simple PeerStore API to get and delete stored information for peers across all books (although the address book is the only one implemented so far)
- Allows clients to add their own listeners to be notified of peer store events. Each listener contains a number of handlers that are triggered on changes in the peer store. For POC purposes, I have only added a handler for multiaddr changes on a peer.

### Suggested next steps:
1. Gain consensus both on general approach of using a Peer Store and specifics of implementation (e.g. how listeners work, API usability, etc.)
2. Implement KeyBook, ProtoBook and rudimentary MetadataBook (likely in this order)
3. POC integration of Peer Store(s) in other modules. `WakuRelay` is a possible candidate here to implement a local Peer Store until it's fully integrated elsewhere in `nim-libp2p`
4. TBD integration of a peer store in `nim-libp2p` (e.g. by building a peer manager, discovery interface, etc.)

### DIFF
--- a/libp2p/peerstore.nim
+++ b/libp2p/peerstore.nim
@@ -1,0 +1,140 @@
+## Nim-LibP2P
+## Copyright (c) 2021 Status Research & Development GmbH
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+import
+  std/[tables, sets, sequtils],
+  ./peerid,
+  ./multiaddress
+
+type
+  ##############################
+  # Listener and handler types #
+  ##############################
+  
+  AddrChangeHandler* = proc(peerId: PeerID, multiaddrs: HashSet[MultiAddress])
+
+  EventListener* = object
+    # Listener object with client-defined handlers for peer store events
+    addrChange*: AddrChangeHandler
+    # @TODO add handlers for other event types
+  
+  #########
+  # Books #
+  #########
+
+  # Each book contains a book (map) and event handler(s)
+  AddressBook* = object
+    book*: Table[PeerID, HashSet[MultiAddress]]
+    addrChange: AddrChangeHandler
+  
+  ####################
+  # Peer store types #
+  ####################
+
+  PeerStore* = ref object of RootObj
+    addressBook*: AddressBook
+    listeners: ref seq[EventListener]
+  
+  StoredInfo* = object
+    # Collates stored info about a peer
+    ## @TODO include data from other stores once added
+    peerId*: PeerID
+    addrs*: HashSet[MultiAddress]
+
+proc init(T: type AddressBook, addrChange: AddrChangeHandler): AddressBook =
+  T(book: initTable[PeerId, HashSet[MultiAddress]](),
+    addrChange: addrChange)
+
+proc init*(T: type PeerStore): PeerStore =
+  var listeners: ref seq[EventListener]
+  new(listeners)
+
+  listeners[] = newSeq[EventListener]()
+
+  proc addrChange(peerId: PeerID, multiaddrs: HashSet[MultiAddress]) =
+    # Notify all listeners of change in multiaddr
+    for listener in listeners[]:
+      listener.addrChange(peerId, multiaddrs)
+  
+  T(addressBook: AddressBook.init(addrChange),
+    listeners: listeners)
+
+####################
+# Address Book API #
+####################
+
+proc get*(addressBook: AddressBook,
+          peerId: PeerID): HashSet[MultiAddress] =
+  ## Get the known addresses of a provided peer.
+  
+  addressBook.book.getOrDefault(peerId,
+                                initHashSet[MultiAddress]())
+
+proc add*(addressBook: var AddressBook,
+          peerId: PeerID,
+          multiaddr: MultiAddress) = 
+  ## Add known multiaddr of a given peer. If the peer is not known, 
+  ## it will be set with the provided multiaddr.
+  
+  addressBook.book.mgetOrPut(peerId,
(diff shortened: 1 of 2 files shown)


libp2p/peerstore.nim
```diff
@@ -0,0 +1,140 @@
...
+  std/[tables, sets, sequtils],
+  ./peerid,
+  ./multiaddress
+
+type
+  ##############################
+  # Listener and handler types #
+  ##############################
+  
+  AddrChangeHandler* = proc(peerId: PeerID, multiaddrs: HashSet[MultiAddress])
```
oskarth, line 20: Is this communicating change per se (as in a delta) or is the set here the sum of all, existing and new, multiaddresses?

oskarth (approved):
Semantic wise looks reasonable to me! I'll defer to @dryajov re nim-libp2p specifics.

libp2p/peerstore.nim
```diff
@@ -0,0 +1,140 @@
...
+    book*: Table[PeerID, HashSet[MultiAddress]]
+    addrChange: AddrChangeHandler
+  
+  ####################
+  # Peer store types #
+  ####################
+
+  PeerStore* = ref object of RootObj
+    addressBook*: AddressBook
+    listeners: ref seq[EventListener]
```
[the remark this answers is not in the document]
jm-clius: Thanks, make sense. Fixed in 29dcc33.

libp2p/peerstore.nim
```diff
@@ -0,0 +1,140 @@
...
+    ## @TODO include data from other stores once added
+    peerId*: PeerID
+    addrs*: HashSet[MultiAddress]
+
+proc init(T: type AddressBook, addrChange: AddrChangeHandler): AddressBook =
+  T(book: initTable[PeerId, HashSet[MultiAddress]](),
+    addrChange: addrChange)
+
+proc init*(T: type PeerStore): PeerStore =
+  var listeners: ref seq[EventListener]
```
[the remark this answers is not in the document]
jm-clius: Fixed in 29dcc33.

libp2p/peerstore.nim
jm-clius: Ah, yes, this could have been clearer. It's not merely the delta, but communicates all known multiaddrs per peer whenever there's a change as per [js-implementation](https://github.com/libp2p/js-libp2p/blob/master/doc/API.md#known-multiaddrs-for-a-peer-change).

jm-clius:
@dryajov any further changes from your side, or can this be merged?

libp2p/peerstore.nim
```diff
@@ -0,0 +1,142 @@
...
+  ./multiaddress
+
+type
+  ##############################
+  # Listener and handler types #
+  ##############################
+  
+  AddrChangeHandler* = proc(peerId: PeerID, multiaddrs: HashSet[MultiAddress])
+
+  EventListener* = object
```
dryajov, line 22: Not sure if we need an object to encapsulate the event's callback, it doesn't cary any additional information and can be easily added as parameter to the callback later on.

dryajov:
@jm-clius I think this looks close enough to what's in js and go and I think this is a good start. I would not merge it to master right not tho and instead I will create an `unstable` branch where all new features/development should happen first - master should remain stable and only get hotfixes and stable features. Please hold off on merging for now I'll do that once I create the `unstable` branch.

dryajov:
@jm-clius I've added an unstable branch, please repoint this PR there.

libp2p/peerstore.nim
jm-clius: Ah, sorry, I somehow missed this comment.

I see your point. The encapsulating "listener" was created to hold all event handlers that may be defined for other books (`keyChange`, `protoChange`, etc.) so that the Peer Store wouldn't need a separate collection for each type of handler. Can easily change to such an implementation though.

(Going to merge this for now to `unstable` and address this in future PR once more books are added.)

PR #502, opened 2021-01-12
Gossipsub wip

sinkingsugar:
* cleanup unused `connections` seq
* add a flag to disconnect very bad peers in score terms

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -145,6 +145,8 @@ type
 
     directPeers*: seq[PeerId]
 
+    disconnectBadPeers*: bool
+
   GossipSub* = ref object of FloodSub
     mesh*: PeerTable                           # peers that we send messages to when we are subscribed to the topic
     fanout*: PeerTable                         # peers that we send messages to when we're not subscribed to the topic
@@ -230,6 +232,7 @@ proc init*(_: type[GossipSubParams]): GossipSubParams =
       ipColocationFactorThreshold: 1.0,
       behaviourPenaltyWeight: -1.0,
       behaviourPenaltyDecay: 0.999,
+      disconnectBadPeers: false
     )
 
 proc validateParameters*(parameters: GossipSubParams): Result[void, cstring] =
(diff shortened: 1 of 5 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -714,13 +717,22 @@ func `/`(a, b: Duration): float64 =
     fb = float64(b.nanoseconds)
   fa / fb
 
+proc disconnectPeer(peer: PubSubPeer) {.async.} =
+  if peer.sendConn != nil:
+    try:
+      await peer.sendConn.close()
```
dryajov, line 723: I think you want to use `switch.disconnect(peerId)` here, otherwise you're simply closing an mplex channel.

sinkingsugar: thanks! I was gonna ask you indeed

kdeme:
commit d25ded7 fixes also https://github.com/status-im/nimbus-eth2/issues/2233

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -714,24 +716,56 @@ func `/`(a, b: Duration): float64 =
...
+            else:
+              peer.shortAgent = "unknown"
+            peer.shortAgent
+          else:
+            "unknown"
+    libp2p_gossipsub_bad_score_disconnection.inc(labelValues = [agent])
+  else:
+    libp2p_gossipsub_bad_score_disconnection.inc(labelValues = ["unknown"])
+
+  if peer.sendConn != nil:
```
dryajov, line 739: Interesting, I imagine that this will be true for most peers? For example, we might have a send connection, but the peer might be faulty/bad or have flagged _us_ as such, in which case sending to it is pointless... So, maybe we shouldn't use this as a criterion?

libp2p/switch.nim
```diff
@@ -244,7 +244,9 @@ proc upgradeIncoming(s: Switch, incomingConn: Connection) {.async, gcsafe.} = #
       await ms.handle(cconn)
     except CatchableError as exc:
       debug "Exception in secure handler during incoming upgrade", msg = exc.msg, conn
-      if not cconn.upgraded.finished:
+      if  not isNil(cconn) and
```
dryajov, line 247: Ah, yeah - I guess, if the peer/connection is being cleared up, this can become `Nil` at any time. Not fun, but a connection is a shared resource, so this is probably unavoidable. 

Maybe, we should hide the upgraded future behind a property and do the checks there instead of the ugly `and`? Don't have to be this PR, can be cleaned up later.

libp2p/protocols/pubsub/gossipsub.nim
dryajov: also, might be a good idea to cherry pick into its own PR.

libp2p/switch.nim
dryajov: also, might be a good idea to cherry pick into its own PR.

sinkingsugar: yeah I asked if this should be cherry picked but not much interest, I'm going to merge this today anyway I think!

sinkingsugar: still not sure about the nil to be fair, just at this point better be safe than sorry :D ?

libp2p/protocols/pubsub/gossipsub.nim
sinkingsugar: Just treating it as a unsafe pointer and being safe I guess :) you make a fair point tho

sinkingsugar:
Ok this can be wrapped so I can work on nimbus side, any objections?

sinkingsugar:
Merging since it already landed in unstable nimbus

PR #503, opened 2021-01-12
Semaphore cancellations

dryajov:


libp2p/utils/semaphore.nim
```diff
@@ -18,8 +18,8 @@ logScope:
 type
   AsyncSemaphore* = ref object of RootObj
     size*: int
-    count*: int
-    queue*: seq[Future[void]]
+    count: int
```
arnetheduck, line 21: `lock` exposes `locked` - semaphore should probably expose `count` as read-only

libp2p/utils/semaphore.nim
```diff
@@ -65,7 +74,8 @@ proc release*(s: AsyncSemaphore) =
                             queue = s.queue.len
 
     if s.queue.len > 0:
-      var fut = s.queue.pop()
+      var fut = s.queue[0]
```
arnetheduck, line 77: `acquire` needs a while loop:

```
let sem = Semaphore.init(1)
let a = sem.acquire()  # acquired
let b = sem.acquire() # blocked

sem.release() # future b is completed
let c = sem.acquire() # oops, both b and c completed
```

arnetheduck: actually never mind, this example is is wrong

libp2p/utils/semaphore.nim
```diff
@@ -19,7 +19,7 @@ type
   AsyncSemaphore* = ref object of RootObj
     size*: int
     count*: int
-    queue*: seq[Future[void]]
+    queue: seq[Future[void]]
 
 proc init*(T: type AsyncSemaphore, size: int): T =
```
arnetheduck, line 24: this should be called `new` because AsyncSemaphore is a `ref object` - alternatively, it should be called `newAsyncSemaphore` because that's what all chronos primitives are called

dryajov: Err, I'm a bit confused with the moving target of constructor naming, you've always pushed for `init` as the end all constructor/initializer name why the sudden change to `new`? 

(btw, I understand the consistency argument and that's fine, but still you just changed the rules :-) ?)

arnetheduck: `init` for `object`, `new` for `ref object` - no change here - it's just that I almost never use `ref object`, they're semantically problematic

libp2p/utils/semaphore.nim
arnetheduck: read-only though

tests/testsemaphore.nim
```diff
@@ -1,6 +1,7 @@
 import random
 import chronos
-import ../libp2p/utils/semaphore
+
+include ../libp2p/utils/semaphore # include to avoid exposing private members
```
arnetheduck, line 4: can import now

libp2p/utils/semaphore.nim
dryajov: Everything in libp2p is a `ref object` and yet we've had numerous discussions to change it to init. OK, I'll remember to use that convention from now on.

tests/testsemaphore.nim
dryajov: I'm still using `queue` which is private

libp2p/utils/semaphore.nim
arnetheduck: these discussions have generally been preceded by a discussion to turn some refs into plain objects - ie it's always possible to create a `ref` from an `object` - the other way around is not possible - so generally, unless there are ownership reasons (which _usually_ are false), it's better to use `object` and explicitly say `ref Xxx` where necessary

tests/testsemaphore.nim
arnetheduck: ...which is dubious - that means that the testing of the API is incomplete or taking shortcuts - testing the private implementation details means that it's not possible to use unit tests to detect backwards compatibility issues - this is a useful point to consider before 1.0: how to know if something is backwards compatible - if you need to change a test, it's _not_ backwards compatible for sure, but this litmus test depends on the tests not using hacks like this.

libp2p/utils/semaphore.nim
dryajov: Dynamic dispatch requires `ref` objects, even if using an ad-hoc dispatch table, that why libp2p uses `ref` for the most part.

tests/testsemaphore.nim
dryajov: I usually avoid private api tests, but in some narrow cases it makes sense. In this case it isn't 100% indispensable anymore and was used initially as a debugging/sanity mechanism which can be removed now.

As for versioning, I want to again stress - IMO, versions don't communicate stability, they communicate changes, associating software stability with a version number doesn't scale to automated package management for example.

arnetheduck: > stability

tests don't communicate "stability": they indicate the lack of instability in a subset of the application - but that aside, they also work as backwards incompatibility detectors: if you need to change a test that tests public api only, you're not backwards compatible and therefore need to bump the major version - it's a tool to help judge the impact of a change.

dryajov: I meant it as a general remark about versions, I don't disagree with the fact that tests communicate backwards compatibility.

dryajov:
@arnetheduck anything else left here or is this good to merge?

libp2p/utils/semaphore.nim
```diff
@@ -65,7 +76,8 @@ proc release*(s: AsyncSemaphore) =
                             queue = s.queue.len
 
     if s.queue.len > 0:
-      var fut = s.queue.pop()
+      var fut = s.queue[0]
+      s.queue.del(0)
```
arnetheduck, line 80: `del` reorders the callbacks, you want `delete` here. the ordering test is too short to detect it - @cheatfate pointed out that the reordering might lead to starvation of some acquires if something else keeps getting ahead in the queue.

dryajov: That's only the case if deleting from the middle of the sequence? I would expect taking from the tip should keep the same order?

arnetheduck: it takes the last item and puts it first, because you're here deleting the first item. effectively, you're doing `pop` except for the first callback.

dryajov: heh, you're right:

```nim
proc del*[T](x: var seq[T], i: Natural) {.noSideEffect.} =
  ## Deletes the item at index `i` by putting ``x[high(x)]`` into position `i`.
  ##
  ## This is an `O(1)` operation.
  ##
  ## See also:
  ## * `delete <#delete,seq[T][T],Natural>`_ for preserving the order
  ##
  ## .. code-block:: Nim
  ##  var i = @[1, 2, 3, 4, 5]
  ##  i.del(2) # => @[1, 2, 5, 4]
  let xl = x.len - 1
  movingCopy(x[i], x[xl])
  setLen(x, xl)
```

PR #505, opened 2021-01-15
Feat/peerstore impl

### LINKED ISSUE #504: Peer Store implementation
#### Background

Building on the POC Peer Store created in #498 with only an `AddressBook`, we can now continue the implementation by adding a `KeyBook`, `ProtoBook` and `MetadataBook`.

#### Acceptance Criteria
Adds the following books:
- `KeyBook`: maps peers to their known public keys
- `ProtoBook`: maps peers to the protocols they support
- `MetadataBook`: maps peers to their metadata. The metadata for each peer is stored as a collection of key-value pairs, where each key is a `string` and each `value` a sequence of `bytes`.

Each book must define methods to add, set, get and delete book entries. Clients should be able to register handlers to be notified of changes in each book.

#### References
Peer Store is based on similar implementation in [go](https://github.com/libp2p/go-libp2p-peerstore) and [js](https://github.com/libp2p/js-libp2p/tree/master/src/peer-store)

cc @dryajov

jm-clius:
This PR closes #504.

It adds a `KeyBook`, `ProtoBook` and `MetadataBook` to the [PoC Peer Store implementation](https://github.com/status-im/nim-libp2p/issues/498).

#### Miscellaneous changes
- Extracted handlers from an encapsulating `EventListener` object (as per [this conversation](https://github.com/status-im/nim-libp2p/pull/499#discussion_r556651834))

#### Suggested next steps:
- Coordinate on best way to provide rudimentary peer management using the Peer Store. `WakuRelay` will specifically benefit from something like a `dialer` interface from where dialed peers can be added to a store, reconnected when connection dropped, etc. Parts of this may overlap with efforts elsewhere in `nim-libp2p`, so proper coordination is necessary. I will therefore propose any next steps as PoC features in the libp2p repo to gain consensus before proceeding.

### DIFF
--- a/libp2p/peerstore.nim
+++ b/libp2p/peerstore.nim
@@ -14,21 +14,14 @@ import
   ./multiaddress
 
 type
-  ##############################
-  # Listener and handler types #
-  ##############################
+  #################
+  # Handler types #
+  #################
   
   AddrChangeHandler* = proc(peerId: PeerID, multiaddrs: HashSet[MultiAddress])
   ProtoChangeHandler* = proc(peerId: PeerID, protos: HashSet[string])
   KeyChangeHandler* = proc(peerId: PeerID, publicKey: PublicKey)
   MetadataChangeHandler* = proc(peerId: PeerID, metadata: Table[string, seq[byte]])
-
-  EventListener* = object
-    # Listener object with client-defined handlers for peer store events
-    addrChange*: AddrChangeHandler
-    protoChange*: ProtoChangeHandler
-    keyChange*: KeyChangeHandler
-    metadataChange*: MetadataChangeHandler
   
   #########
   # Books #
@@ -60,7 +53,10 @@ type
     protoBook*: ProtoBook
     keyBook*: KeyBook
     metadataBook*: MetadataBook
-    listeners: seq[EventListener]
+    addrChangeHandlers*: seq[AddrChangeHandler]
+    protoChangeHandlers*: seq[ProtoChangeHandler]
+    keyChangeHandlers*: seq[KeyChangeHandler]
+    metadataChangeHandlers*: seq[MetadataChangeHandler]
   
   StoredInfo* = object
     # Collates stored info about a peer
@@ -87,27 +83,30 @@ proc init(T: type MetadataBook, metadataChange: MetadataChangeHandler): Metadata
     metadataChange: metadataChange)
 
 proc init*(p: PeerStore) =
-  p.listeners = newSeq[EventListener]()
+  p.addrChangeHandlers = newSeq[AddrChangeHandler]()
+  p.protoChangeHandlers = newSeq[ProtoChangeHandler]()
+  p.keyChangeHandlers = newSeq[KeyChangeHandler]()
+  p.metadataChangeHandlers = newSeq[MetadataChangeHandler]()
 
   proc addrChange(peerId: PeerID, multiaddrs: HashSet[MultiAddress]) =
     # Notify all listeners of change in multiaddr
-    for listener in p.listeners:
-      listener.addrChange(peerId, multiaddrs)
+    for handler in p.addrChangeHandlers:
+      handler(peerId, multiaddrs)
   
   proc protoChange(peerId: PeerID, protos: HashSet[string]) =
     # Notify all listeners of change in proto
-    for listener in p.listeners:
-      listener.protoChange(peerId, protos)
+    for handler in p.protoChangeHandlers:
+      handler(peerId, protos)
   
   proc keyChange(peerId: PeerID, publicKey: PublicKey) =
     # Notify all listeners of change in public key
-    for listener in p.listeners:
-      listener.keyChange(peerId, publicKey)
+    for handler in p.keyChangeHandlers:
+      handler(peerId, publicKey)
   
   proc metadataChange(peerId: PeerID, metadata: Table[string, seq[byte]]) =
     # Notify all listeners of change in public key
-    for listener in p.listeners:
-      listener.metadataChange(peerId, metadata)
+    for handler in p.metadataChangeHandlers:
+      handler(peerId, metadata)
   
   p.addressBook = AddressBook.init(addrChange)
   p.protoBook = ProtoBook.init(protoChange)
@@ -299,11 +298,17 @@ proc deleteValue*(metadataBook: var MetadataBook,
 # Peer Store API #
 ##################
 
-proc addListener*(peerStore: PeerStore,
-                  listener: EventListener) =
-  ## Register event listener to notify clients of changes in the peer store
+proc addHandlers*(peerStore: PeerStore,
+                  addrChangeHandler: AddrChangeHandler,
+                  protoChangeHandler: ProtoChangeHandler,
+                  keyChangeHandler: KeyChangeHandler,
+                  metadataChangeHandler: MetadataChangeHandler) =
+  ## Register event handlers to notify clients of changes in the peer store
   
-  peerStore.listeners.add(listener)
+  peerStore.addrChangeHandlers.add(addrChangeHandler)
+  peerStore.protoChangeHandlers.add(protoChangeHandler)
+  peerStore.keyChangeHandlers.add(keyChangeHandler)
+  peerStore.metadataChangeHandlers.add(metadataChangeHandler)
 
 proc delete*(peerStore: PeerStore,
              peerId: PeerID): bool =
--- a/libp2p/peerstore.nim
+++ b/libp2p/peerstore.nim
@@ -9,6 +9,7 @@
 
 import
   std/[tables, sets, sequtils],
+  ./crypto/crypto,
   ./peerid,
   ./multiaddress
 
@@ -18,11 +19,16 @@ type
   ##############################
   
   AddrChangeHandler* = proc(peerId: PeerID, multiaddrs: HashSet[MultiAddress])
+  ProtoChangeHandler* = proc(peerId: PeerID, protos: HashSet[string])
+  KeyChangeHandler* = proc(peerId: PeerID, publicKey: PublicKey)
+  MetadataChangeHandler* = proc(peerId: PeerID, metadata: Table[string, seq[byte]])
 
   EventListener* = object
     # Listener object with client-defined handlers for peer store events
     addrChange*: AddrChangeHandler
-    # @TODO add handlers for other event types
+    protoChange*: ProtoChangeHandler
+    keyChange*: KeyChangeHandler
+    metadataChange*: MetadataChangeHandler
   
   #########
   # Books #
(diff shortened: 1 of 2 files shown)


libp2p/peerstore.nim
```diff
@@ -33,33 +32,86 @@ type
...
   # Peer store types #
   ####################
 
   PeerStore* = ref object of RootObj
     addressBook*: AddressBook
-    listeners: seq[EventListener]
+    protoBook*: ProtoBook
+    keyBook*: KeyBook
+    metadataBook*: MetadataBook
+    addrChangeHandlers*: seq[AddrChangeHandler]
```
dryajov, line 56: Hmm, wouldn't it make more sense to move the change handlers into the specific `*Book` types? For example, I'm not sure what the functionality of `KeyBook.keyChange` is and how  it relates to `keyChangeHandlers`? Wouldn't it make more sense to move  `keyChangeHandlers` under `KeyBook` and get rid of `keyChange`, or am I missing something?

jm-clius: I agree. :) Improvement in ff7d081

The idea was to create some level of indirection between the multiple _public_ handlers that different clients can register on the Peer Store and the single, _private_ handler that the peer store itself registers per *Book when it initialises those books. Each book then only notifies its encompassing Peer Store of changes and the Peer Store in turn notifies its multiple registered clients (perhaps enforcing optional policy). Currently this serves no more than a conceptual purpose, so it will be cleaner to just move the handlers to the individual books. Thanks!

dryajov: Yeah, I'm not sure if the "proxy" handler makes sense there - but I understand the idea behind it now. 

My perspective on the `PeerStore` is a wrapper that exposes specific `register*Handler` methods to each underlying `*Book`. I think it's cleaner and simpler to reason about, but I'm willing to reconsider it when we might require this functionality.

libp2p/peerstore.nim
```diff
@@ -31,35 +30,59 @@ type
...
     addrs*: HashSet[MultiAddress]
+    protos*: HashSet[string]
+    publicKey*: PublicKey
+    metadata*: Table[string, seq[byte]]
 
-proc init(T: type AddressBook, addrChange: AddrChangeHandler): AddressBook =
+proc init(T: type AddressBook): AddressBook =
   T(book: initTable[PeerId, HashSet[MultiAddress]](),
-    addrChange: addrChange)
+    changeHandlers: newSeq[AddrChangeHandler]())
```
dryajov, line 67: Just a small remark - there is no need to initialize the sequence explicitly with `newSeq[T]()`, `seq[T]` does auto initialization, just like a `string` does.

jm-clius: Thanks! And addressed: 926d340

dryajov:
@sinkingsugar would love your input on this.

libp2p/peerstore.nim
```diff
@@ -31,35 +30,55 @@ type
...
+  ProtoBook* = object
+    book*: Table[PeerID, HashSet[string]]
+    changeHandlers: seq[ProtoChangeHandler]
+  
+  KeyBook* = object
+    book*: Table[PeerID, PublicKey]
+    changeHandlers: seq[KeyChangeHandler]
+  
+  MetadataBook* = object
+    book*: Table[PeerID, Table[string, seq[byte]]]
```
oskarth, line 44: Is this completely arbitrary? What do we imagine storing here? (Generally seems useful, mostly curious)

libp2p/peerstore.nim
```diff
@@ -107,36 +129,206 @@ proc set*(addressBook: var AddressBook,
...
+#####################
+
+proc get*(protoBook: ProtoBook,
+          peerId: PeerID): HashSet[string] =
+  ## Get the known protocols of a provided peer.
+  
+  protoBook.book.getOrDefault(peerId,
+                              initHashSet[string]())
+
+proc add*(protoBook: var ProtoBook,
```
oskarth, line 148: At what point will this be called? Who is responsible for calling it and at what point in the lifecycle?

libp2p/peerstore.nim
```diff
@@ -107,36 +129,206 @@ proc set*(addressBook: var AddressBook,
...
+             peerId: PeerID): bool =
+  ## Delete the provided peer from the book.
+  
+  if not keyBook.book.hasKey(peerId):
+    return false
+  else:
+    keyBook.book.del(peerId)
+    return true
+
+proc set*(keyBook: var KeyBook,
```
oskarth, line 204: Ditto above, where is this called and who is responsible for it? Upon initial connection in nim-libp2p or somewhere else?

libp2p/peerstore.nim
```diff
@@ -9,20 +9,19 @@
...
+  # Handler types #
+  #################
   
   AddrChangeHandler* = proc(peerId: PeerID, multiaddrs: HashSet[MultiAddress])
-
-  EventListener* = object
-    # Listener object with client-defined handlers for peer store events
-    addrChange*: AddrChangeHandler
-    # @TODO add handlers for other event types
+  ProtoChangeHandler* = proc(peerId: PeerID, protos: HashSet[string])
```
oskarth, line 22: Is FooChangeHandler the common name here? Nitpick, but something like FooUpdateHandler might be a tiny bit more descriptive, "Change" to me suggests something could be breaking.

libp2p/peerstore.nim
jm-clius: Yes, it's supposed to be a general purpose key-value metadata store per peer. In my mind we can use it to keep performance related metrics, such as peer age, throughput, bandwidth, etc. In the go- and js implementations it's also used to store peer location, libp2p agent version and other peer-related metadata determined during the `identify` procedure (see e.g. https://github.com/libp2p/js-libp2p/pull/800).

libp2p/peerstore.nim
jm-clius: Since a Peer Store provides a "standalone" API it can theoretically be instantiated and added to (at any time) by any module seeking to provide some level of peer management.
Eventually I envision it to be integrated in such a way that there is a single libp2p instance, with the following integration points (based on [js-implementation](https://github.com/libp2p/js-libp2p/tree/master/src/peer-store#submitting-records-to-the-peerstore)):
- a `Dialer` (TBD) adds "bootstrap" or statically defined peers when these are dialed. This is most likely the first point of integration for `WakuRelay` as well. The `Dialer` would also be responsible for maintaining connections and redialing peers when dropped.
- the `identify` protocol adds information about identified peers.
- Discovery/DHT mechanisms adds information about discovered peers.

libp2p/peerstore.nim
jm-clius: I agree. `change` was selected to stay as close as possible to [similar events and naming](https://github.com/libp2p/js-libp2p/tree/master/src/peer-store#events) in the js-implementation, but I am happy to update naming to `update` :)

sinkingsugar:
> @sinkingsugar would love your input on this.

LGTM, Very good style!
I noticed no `raises` which is good and so we can probably add on top `{.push raises: [Defect].}`

sinkingsugar (approved):

libp2p/peerstore.nim
sinkingsugar: I'm not 100% sure about `seq[byte]` as it can just hold pure data, no references etc.. but for now it's just fine, it's something that can be fixed when we figure out better.

libp2p/peerstore.nim
dryajov: I think the naming convention makes sense here.

libp2p/peerstore.nim
dryajov: > I noticed no raises which is good and so we can probably add on top {.push raises: [Defect].}

Ah, great catch, lets get that added right at the top of the file @jm-clius!

dryajov: In fact, to clarify - there rule of thumb is that we use this style everywhere possible, the only place/reason it can't be used is in the presence of async, non-async apis should default to this style.

dryajov: (off topic, obviously this was meant to go under https://github.com/status-im/nim-libp2p/pull/505#issuecomment-766785882, but when I do "Quote reply" it puts it under the wrong thread 😨. I've noticed github doing this twice already - user error or bug?)

dryajov: > Is this completely arbitrary? What do we imagine storing here? (Generally seems useful, mostly curious)

Good point. I think there will be many metadata types and it's probably better to have a parameterizable type here like:

```nim
  MetadataBook*[T] = object
    book*: Table[PeerID, Table[string, T]]
```

Another approach is to have a base `PeerBookEntry*[T]` type that everything inherits from and then simply define specialized `PeerBookEntry*[T]` per book. So for example, 

```nim
type 
  PeerBookEntry*[T] = object of RootObj
    entry*: T

  AddressEntry* = object of PeerBookEntry[MultiAddress]

  AddressBook* = object
    book*: Table[PeerID, HashSet[AddressEntry]]
    addrChange: AddrChangeHandler
    changeHandlers: seq[AddrChangeHandler]
```

Then use that to build arbitrary metadata entries...

jm-clius: Ok, great suggestions and I especially like the consistency of the second suggestion. To be clear, do you then suggest that specialized entry objects be defined for `AddressEntry`, `ProtoEntry` and `KeyEntry`, but that the `MetadataBook` maintains the generic `PeerBookEntry[T]` so peer book clients can define their own entry types?

oskarth (approved):

libp2p/peerstore.nim
sinkingsugar: I like the generic entry yes!, maybe codegen will be bloaty but that's not our fault (nim things)

dryajov: > Ok, great suggestions and I especially like the consistency of the second suggestion. To be clear, do you then suggest that specialized entry objects be defined for AddressEntry, ProtoEntry and KeyEntry, but that the MetadataBook maintains the generic PeerBookEntry[T] so peer book clients can define their own entry types?

Yep, precisely. In fact, we can make `MetadataBook` be the base for the more specialised book types as well. The only thing I'm not content with is the implicit `entry*: T` in `PeerBookEntry`, I wish Nim had a more concise syntax for that sort of thing.

jm-clius: Yup, this is a very reasonable suggestion, since addresses, keys and protos can be seen as specialised types of metadata in any case.

My only concern would be that the API should still make explicit the subtle differences between the books. For example, an AddressBook maps a `peerId` to a _collection_ of `addrs` (and therefore should allow clients to `add` individual `addrs` to each collection), whereas a KeyBook only allows a single `PublicKey` entry per `peerId` (and therefore only allows `set`ting a singular entry).

I've attempted to do the above in 12e0716. This significantly changes (and simplifies) the `PeerStore` by making extensive use of generics. Each `PeerStore` still consists of a specialised `addressBook`, `keyBook` and `protoBook`, but now also contains one "any-purpose" `metadataBook` of generic type `PeerBook[T]` (this can easily be changed to a `seq` of such any-purpose books if there is a need for extensibility). Of course, this does not really allow for run-time object variance in the `metadataBook` (which I don't think generics are suitable for, in any case?), but does allow clients to specify their own metadata `type` when instantiating a new `PeerStore`. I'm not exactly happy with the `PeerStore` itself being genericised, but since it requires a concrete type when constructing the `metadataBook` I don't really see a way around this. I'll be glad for any suggestions to make this even more flexible. cc @sinkingsugar @dryajov

dryajov: Honestly, the concept of a free form metadata store doesn't make much sense in this context and I would just leave it out, allowing anyone who needs additional information to just extend the `PeerBook` and `PeerStores` accordingly.

dryajov (approved):

PR #507, opened 2021-01-19
Split upgrade flow

dryajov:
This refactors the upgrade flow from the switch into its own module. This is the first step to allow multi-transport support.

### DIFF
--- a/libp2p/upgrademngrs/muxedupgrade.nim
+++ b/libp2p/upgrademngrs/muxedupgrade.nim
@@ -23,7 +23,7 @@ type
     muxers*: Table[string, MuxerProvider]
     streamHandler*: StreamHandler
 
-proc identify(u: MuxedUpgrade, muxer: Muxer) {.async, gcsafe.} =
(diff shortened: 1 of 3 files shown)


libp2p/upgrademngrs/muxedupgrade.nim
```diff
@@ -0,0 +1,216 @@
+## Nim-LibP2P
+## Copyright (c) 2019 Status Research & Development GmbH
```
arnetheduck, line 2: year

libp2p/upgrademngrs/muxedupgrade.nim
```diff
@@ -0,0 +1,216 @@
...
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+import tables, sequtils
```
arnetheduck, line 10: `std/`

PR #510, opened 2021-01-28
add base exception class and fix hierarchy

dryajov:
This adds a base exception class and fixes the exception hierarchy as a first pass. Leaving this as a draft for discussion first and foremost.

### DIFF
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -25,7 +25,7 @@ const
   MaxConnectionsPerPeer* = 5
 
 type
-  TooManyConnectionsError* = object of CatchableError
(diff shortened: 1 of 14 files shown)


arnetheduck:
this is a first step - before it's useful, it must:

* catch all exceptions from other layers and translate every time it calls some other layer
* never raise other errors

dryajov:
> this is a first step - before it's useful, it must:
> 
> * catch all exceptions from other layers and translate every time it calls some other layer
> * never raise other errors

Yes, I agree with both points, as we discussed. But I want to make sure that the hierarchy is right before I make further changes.

Issue #414, opened 2020-10-29
Empty message in gossipsub 1.1 (might be in 1.0 too)


dryajov:
It appears that we're still getting empty messages in gossip/floodsub:

`{"lvl":"TRC","ts":"2020-10-28 18:20:10.904-06:00","msg":"decoded msg from peer","topics":"pubsubpeer","tid":3554702,"file":"pubsubpeer.nim:115","conn":"16*EJ9jjk:5f9a0ab8ebb9805379c87060","peer":"16*EJ9jjk","closed":false,"msg":{"subscriptions":[],"messages":[],"control":{"ihave":[],"iwant":[],"graft":[],"prune":[]}}}`

Here is a more complete trace of the interaction:

```json
{"lvl":"TRC","ts":"2020-10-28 18:20:08.609-06:00","msg":"Stream created","topics":"lpstream","tid":3554702,"file":"lpstream.nim:95","s":"5f9a0ab8ebb9805379c87060","objName":"LPChannel","dir":"In"}
{"lvl":"TRC","ts":"2020-10-28 18:20:08.609-06:00","msg":"Monitoring for timeout","topics":"connection","tid":3554702,"file":"connection.nim:80","s":"5f9a0ab8ebb9805379c87060","timeout":{"value":300000000000}}
{"lvl":"TRC","ts":"2020-10-28 18:20:08.609-06:00","msg":"BufferStream created","topics":"bufferstream","tid":3554702,"file":"bufferstream.nim:81","s":"5f9a0ab8ebb9805379c87060"}
{"lvl":"TRC","ts":"2020-10-28 18:20:08.609-06:00","msg":"Created new lpchannel","topics":"mplex","tid":3554702,"file":"lpchannel.nim:236","s":"16*EJ9jjk:5f9a0ab8ebb9805379c87060:2","id":2,"initiator":false}
{"lvl":"TRC","ts":"2020-10-28 18:20:08.610-06:00","msg":"Creating new channel","topics":"mplex","tid":3554702,"file":"mplex.nim:100","m":"16*EJ9jjk:5f9a0ab7ebb9805379c87057","channel":"16*EJ9jjk:5f9a0ab8ebb9805379c87060:2","id":2,"initiator":false,"name":"2"}
[...]

PR #512, opened 2021-02-02
Punish empty gossip messages

sinkingsugar:
Solve #414


sinkingsugar:
Looks like the go daemon is sending those empty messages during floodsub testing.

sinkingsugar:
I'm still unsure about this.. a empty message might have a meaning for some apps and sadly we don't make distinction at PB level (0 len vs not there)

dryajov:
I would say we compare with the other implementations and see what they are doing. Worst case, we can just leave it as it is.

PR #518, opened 2021-02-05
Merge unstable

dryajov:
This is fairly innocuous as it isn't being used internally by libp2p right now. However, other projects such as waku and potentially dagger, are already depending on it.


sinkingsugar:
Totally indeed. Fine for me!

PR #519, opened 2021-02-05
Remove secio usage and cleanup exports

dryajov:
Note, as the title state, this removes secio usage and would `quit` if it's attempted to be used with `newStandardSwitch` - maybe quitting is a bit too drastic?

### DIFF
--- a/libp2p.nim
+++ b/libp2p.nim
@@ -8,20 +8,26 @@
 ## those terms.
 
 import
-  libp2p/[daemon/daemonapi,
-          daemon/transpool,
-          protobuf/minprotobuf,
+  libp2p/[protobuf/minprotobuf,
+          muxers/muxer,
+          muxers/mplex/mplex,
+          stream/lpstream,
+          stream/bufferstream,
+          stream/connection,
+          transports/transport,
+          transports/tcptransport,
+          protocols/secure/noise,
           varint,
           switch,
           peerid,
           peerinfo,
-          stream/lpstream,
-          stream/bufferstream,
-          stream/connection,
           multiaddress,
           crypto/crypto]
 
+import bearssl
+
 export
-  daemonapi, transpool, minprotobuf, varint,
-  switch, peerid, peerinfo, connection,
-  multiaddress, crypto, lpstream, bufferstream
+  minprotobuf, varint,switch, peerid, peerinfo,
+  connection, multiaddress, crypto, lpstream,
+  bufferstream, bearssl, muxer, mplex, transport,
+  tcptransport, noise
--- a/libp2p/standard_setup.nim
+++ b/libp2p/standard_setup.nim
@@ -1,10 +1,10 @@
 import options, tables
 import chronos
-import libp2p
-
-import connmanager
+import ../libp2p
 
 export libp2p
+
+import connmanager
 
 type
   SecureProtocol* {.pure.} = enum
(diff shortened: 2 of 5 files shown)


sinkingsugar (approved):
Fine for me about the quit.
Also LGTM

libp2p/standard_setup.nim
```diff
@@ -54,7 +48,7 @@ proc newStandardSwitch*(privKey = none(PrivateKey),
     of SecureProtocol.Noise:
       secureManagerInstances &= newNoise(rng, seckey).Secure
     of SecureProtocol.Secio:
-      secureManagerInstances &= newSecio(rng, seckey).Secure
+      quit("Secio is deprecated!") # use of secio is unsafe
```
arnetheduck, line 51: instead, might be able to mark the  `Secio` enum value with `{.deprecated.}` - if not, might want to provide instructions for how to use secio anyway since it's needed to access old go implementations like ipfs etc

libp2p.nim
```diff
@@ -8,20 +8,26 @@
...
 
+import bearssl
+
 export
-  daemonapi, transpool, minprotobuf, varint,
-  switch, peerid, peerinfo, connection,
-  multiaddress, crypto, lpstream, bufferstream
+  minprotobuf, varint,switch, peerid, peerinfo,
+  connection, multiaddress, crypto, lpstream,
+  bufferstream, bearssl, muxer, mplex, transport,
```
arnetheduck, line 32: not sure internals like bearssl and varint necessarily should be exported, they're not part of the public api exposed in this module and pollute the global namespace with many ambiguous symbols - the rule is basically "anything that forms part of the public api should be exported"

libp2p/standard_setup.nim
dryajov: AFAIK, `secio` has been removed everywhere, since it's been marked insecure

- js https://github.com/ipfs/js-ipfs/commit/5f5ef7ee6cc6dc634cc6adbede0602492490a85d
- go https://github.com/ipfs/go-ipfs/commit/1a0474bb0d3821d75573bc990267841fcb729778

libp2p.nim
dryajov: I'm fine with `bearssl`, but though I expose it since it's now used as the default RNG provider. 

Varint seems like a good candidate to export tho, since it's used extensively across libp2p and can be useful for custom parsers and such?

arnetheduck: > I'm fine with bearssl, but though I expose it since it's now used as the default RNG provider.

ok - if any  bear type appears in public API needed by this module, then yes

> Varint seems like a good candidate to export tho,

depends - ie libp2p consumes varint but presents a more sane API to to user that doesn't expose any varints, so I would tend towards no, for pollution reasons (ie if the user is using another varint library, of which there are plenty, they'll get into symbol trouble which is annoying and easily causes breakage

dryajov: > depends - ie libp2p consumes varint but presents a more sane API to to user that doesn't expose any varints, so I would tend towards no, for pollution reasons (ie if the user is using another varint library, of which there are plenty, they'll get into symbol trouble which is annoying and easily causes breakage

OK, I don't disagree since it can be imported implicitly.

dryajov:
> ok - if any bear type appears in public API needed by this module, then yes

Yeah, we need `BrHmacDrbgContext` which seems to be part of bearssl, is returned by `newRng()`.

libp2p.nim
dryajov: > ok - if any bear type appears in public API needed by this module, then yes

Yeah, we need `BrHmacDrbgContext` which seems to be part of bearssl, is returned by `newRng()`.

PR #520, opened 2021-02-08
Cancellation tests.

cheatfate:
This is approach which i'm recommending to use to check if public API procedure properly supports cancellation.

From my point of view library's public API procedure should not swallow exceptions and should not raise errors instead (but its not against rules of cancellation). The only thing which i want to achieve is to have consistent behavior on `cancellation` (for example doesn't matter on what iteration but `dial` should always single "finished" state).

Also as you can see transports and secure connections are leaking.

This PR could be merged only when all the problems will be fixed.

### DIFF
--- a/tests/helpers.nim
+++ b/tests/helpers.nim
@@ -69,6 +69,19 @@ template asyncTest*(name: string, body: untyped): untyped =
         body
     )())
 
+template asyncCancelTest*(name: string, body: untyped): untyped =
+  test name:
+    var counter = 0
+    while true:
+      let res = waitFor((
+        proc(n: int): Future[bool] {.async, gcsafe.} =
+          var testIteration {.inject.} = n
+          body
+      )(counter))
+      if res:
+        break
+      inc(counter)
+
 type RngWrap = object
   rng: ref BrHmacDrbgContext
 
--- a/tests/testcancel.nim
+++ b/tests/testcancel.nim
@@ -1,0 +1,110 @@
+{.used.}
+
+import unittest, options, sequtils
+import chronos, chronicles
+import stew/byteutils
+import nimcrypto/sysrand
+import ../libp2p/[errors,
+                  switch,
+                  multistream,
+                  standard_setup,
+                  stream/bufferstream,
+                  stream/connection,
+                  multiaddress,
+                  peerinfo,
+                  crypto/crypto,
+                  protocols/protocol,
+                  protocols/secure/secure,
+                  muxers/muxer,
+                  muxers/mplex/lpchannel,
+                  stream/lpstream,
+                  stream/chronosstream,
+                  transports/tcptransport]
+import ./helpers
+
+const
+  TestCodec = "/test/proto/1.0.0"
+
+type
+  TestProto = ref object of LPProtocol
+
+suite "Cancellation test suite":
+  teardown:
+    checkTrackers()
+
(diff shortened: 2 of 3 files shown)


sinkingsugar:
I fixed some of those leaks but could not get all of them yet, one last transport is leaking.

But the issue here is quite huge..
I was proposing to find a way to make `CancelledError` not a `CatchableError` but @cheatfate opposed asking that we should refactor libp2p to not use `CatchableError`. He's right but it's a huge amount of work and testing necessary.

Opinions?
cc @dryajov 

p.s.
This test is good but it's testing only one proc..... testing everything would be quite an achievement.

dryajov:
I agree and I started on it here https://github.com/status-im/nim-libp2p/pull/510. But curious, what leaks are there, you have a PR that demonstrates it?

dryajov:
Btw, I believe it's totally worth to spend the time doing it and it does force use to go over all of the code and cleanup a few thing. I would say this is something to do before releasing v1.

libp2p/muxers/muxer.nim
```diff
@@ -67,9 +67,9 @@ method init(c: MuxerProvider) =
       if not isNil(c.muxerHandler):
         futs &= c.muxerHandler(muxer)
 
-      checkFutures(await allFinished(futs))
-    except CancelledError as exc:
-      raise exc
+      await allFuturesThrowing(futs)
```
dryajov, line 70: hm, I wonder why would this leak here at all, we still don't do anything if any of them raised, because we need to handle all errors in the `handler` proc.

libp2p/upgrademngrs/muxedupgrade.nim
```diff
@@ -59,6 +59,8 @@ proc mux*(u: MuxedUpgrade, conn: Connection): Future[Muxer] {.async, gcsafe.} =
 
   try:
     await u.identify(muxer)
+  except CancelledError:
```
dryajov, line 62: This causes issues with `raises` because the type of the exception is lost on `raise`.

libp2p/upgrademngrs/muxedupgrade.nim
```diff
@@ -121,7 +123,10 @@ method upgradeIncoming*(u: MuxedUpgrade, incomingConn: Connection) {.async, gcsa
     except CatchableError as exc:
       debug "Exception in secure handler during incoming upgrade", msg = exc.msg, conn
       if not cconn.isUpgraded:
-        cconn.upgrade(exc)
+        # signal failure
+        cconn.upgrade(failed = exc)
+      if exc of CancelledError:
```
dryajov, line 128: We can't reraise the cancellation here, because is a spawned proc, there is nothing to propagate it to... With the `raises: [Defect]` this would have been obvious.

libp2p/switch.nim
```diff
@@ -329,9 +331,12 @@ proc accept(s: Switch, transport: Transport) {.async.} = # noraises
       debug "Accepted an incoming connection", conn
       asyncSpawn upgradeMonitor(conn, upgrades)
       asyncSpawn s.upgrade.upgradeIncoming(conn)
-    except CancelledError as exc:
+    except CancelledError:
```
dryajov, line 334: The `as exc:` is needed for now, as there are multiple bugs surrounding it's lack.

PR #521, opened 2021-02-08
support compilation with Nim-1.4 HEAD

stefantalpalaru:


dryajov (approved):
LGTM.

We might want to add 1.4.x to the compiler matric in github actions...

PR #531, opened 2021-02-12
[SEC] gossipsub - rebalanceMesh may prune up to D_lo on oversubscription

### LINKED ISSUE #403: [SEC] gossipsub - rebalanceMesh may prune up to D_lo on oversubscription
## Description

when the peercount for a topic exceeds the high count `D_hi` rebalanceMesh attempts to prune peers retaining up to `d_Score` high scoring peers:

> When pruning because of oversubscription, the peer keeps the best D_score scoring peers and selects the remaining peers to keep at random. This protects the mesh from takeover attacks and ensures that the best scoring peers are kept in the mesh. At the same time, we do keep some peers as random so that the protocol is responsive to new peers joining the mesh. The selection is done under the constraint that D_out peers are outbound connections; if the scoring plus random selection does not result in enough outbound connections, then we replace the random and lower scoring peers in the selection with outboud connection peers.

https://github.com/status-im/nim-libp2p/blob/556213abf4bcfb33e27d28e5148619343df05dbb/libp2p/protocols/pubsub/gossipsub.nim#L460-L503

* nodes to be pruned are sorted by score (ascending). the `prunes` sequence includes direct-peers. It is unclear from the spec whether direct peers should be included in the heartbeat pruning process. However, peers might still be tracked in `g.explcit` even though they're not part of `g.mesh` anymore which may be wrong.

* If the number of peers is greater than the number of high score peers to keep (`dScore`) it is first attempted to prune inbound and then outbound peers. The problem with this approach is that if the number of peers may drop significantly below `GossipSubD`.
  * example values: `dScore=4`, `GossipSubD=6`, `dOut=3`, `dHi=12` and `mesh.peers=13 -> prunes=13-4=9 (6 inbound, 3 outbound)` will result in 6 inbound and 3 outbound to be pruned resulting in 13-9=4 peerings to be left which is exactly `dLo=4`. at this point it is up to the opportunistic grafting to make up for that loss which might not be guaranteed. With the default of `dLo=4` we exactly land on the low mark.

* `to_be_pruned` and `to_be_grafted` peers are immediately accounted for in `g.mesh` and this value is used to determine if the current p2p network graph conditions are met. The `graft` and `prune` messages for recently grafted/pruned peers, however, have not yet been sent out. They are added to a `grafts`/`prunes` sequence and broadcast at the end. Now there is a slight chance that a newly grafted node is immediately pruned e.g. when `outbound(g.mesh.peers(topic))< d_Out` which may result in a connection to be grafted just to immediately prune by `g.mesh.peers(topic) > GossipSubDhi` afterwards (immediately grafted in opportunistic grafting after just being pruned). (note: this assumes a code change according to status-im/nimbus-eth2#1871) 

## Exploit Scenario

weakened sybil attacks protection due to inaccurate peer accounting which may lead to opportunities for an attacker to join the nodes peering.

## Mitigation Recommendation

* consider removing direct-peers from the prune candidate list. (spec is unclear on that but it does not sound to be a good idea to remove 'trusted' peerings)
* ensure to never fall below `D_lo` peers when pruning (re-verify graph params at the end). Ensure to meet the `D_out` requirement when pruning due to oversubscription.
  * ensure that there are at least `D_out` peers first and rebalance to `GossipSubD` after that.
* avoid sending out graft and prune messages to the same node within one heartbeat

sinkingsugar:
Fixes #403

### DIFF
--- a/libp2p/protocols/pubsub/peertable.nim
+++ b/libp2p/protocols/pubsub/peertable.nim
@@ -41,6 +41,3 @@ func peers*(table: PeerTable, topic: string): int =
     table[topic].len
   else:
     0
-
-func getPeers*(table: Table[string, HashSet[string]], topic: string): HashSet[string] =
-  table.getOrDefault(topic, initHashSet[string]())
--- a/tests/pubsub/testgossipinternal.nim
+++ b/tests/pubsub/testgossipinternal.nim
@@ -509,3 +509,59 @@ suite "GossipSub internal":
 
     await conn.close()
     await gossipSub.switch.stop()
+
+  asyncTest "rebalanceMesh Degree Hi - audit scenario":
+    let gossipSub = TestGossipSub.init(newStandardSwitch())
+    gossipSub.parameters.dScore = 4
+    gossipSub.parameters.d = 6
+    gossipSub.parameters.dOut = 3
+    gossipSub.parameters.dHigh = 12
+    gossipSub.parameters.dLow = 4
+
+    let topic = "foobar"
+    gossipSub.mesh[topic] = initHashSet[PubSubPeer]()
+    gossipSub.topicParams[topic] = TopicParams.init()
+
+    var conns = newSeq[Connection]()
+    gossipSub.gossipsub[topic] = initHashSet[PubSubPeer]()
+    for i in 0..<6:
+      let conn = newBufferStream(noop)
+      conn.dir = Direction.In
+      conns &= conn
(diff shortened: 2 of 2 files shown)


dryajov (approved):
LGTM. 

The majority of this tests would probably need to be rewritten using the new public APIs from the split modules.

sinkingsugar:
You mean just refactor the include out basically? 
Just a refactor, should be trivial at this point, we can do that after v1 I guess even. either or.

dryajov:
> You mean just refactor the include out basically?

Yeah, exactly and you're right, doesn't have to be for v1.

PR #533, opened 2021-02-22
investigate missing coverage

sinkingsugar:


sinkingsugar:
@dryajov we were missing some folders , expect a big drop of coverage 😄 
I will try to make it less script error prone... not so easy with the way nim and cache works.

sinkingsugar:
actually it should not drop much, was just missing the new gossip folders

dryajov:
We have plenty of tests, by no means complete, but not terrible either :)

dryajov:
> We have plenty of tests, by no means complete, but not terrible either :)

Let's work to increase it to > 90%.

PR #535, opened 2021-02-23
[SEC] gossipsub - handleIHAVE/handleIWANT recommendations & notes

### LINKED ISSUE #400: [SEC] gossipsub - handleIHAVE/handleIWANT recommendations & notes
## Description

* consider checking for `<=` instead of a strict comparison in case `iHaveBudget`/`iWantBudget` may be negative (e.g. due to concurrent processing of messages from the same peer. Is it guaranteed that this is not concurrent?)
https://github.com/status-im/nim-libp2p/blob/556213abf4bcfb33e27d28e5148619343df05dbb/libp2p/protocols/pubsub/gossipsub.nim#L982-L983

* consider limiting the number of individual `msgids` allowed per Ihave message. At the moment the budget only considers the `ControlIHave` message. messageIds are unrestricted which may open up potential for DoS scenarios. Consider keeping track of the number of ASKs a client provides over a short period of time and reduce the score if the node misbehaves by flood-asking.
https://github.com/status-im/nim-libp2p/blob/556213abf4bcfb33e27d28e5148619343df05dbb/libp2p/protocols/pubsub/gossipsub.nim#L985-L988

* consider deduplicating messageId's provided with IHave/IWant messages before processing them further (but after checking budget). For example, an `IHave` may contain the same message id multiple times which would result in a `IWant` requesting the same message id multiple times. Consider limiting the number of messageids allowed for ControlIWant.
https://github.com/status-im/nim-libp2p/blob/556213abf4bcfb33e27d28e5148619343df05dbb/libp2p/protocols/pubsub/gossipsub.nim#L986-L994

* the `iWantBudget` takes submessages into account while the `iHaveBudget` only counts the ControlIHave rpc message.

* consider shuffling `result.messageIDs` before sending it out to increase the likelihood of getting an answer if the peer truncates the list due to internal size restrictions.

* `handleIWant` should return early at the beginning if the budget is already `<=0`

sinkingsugar:
Fixes #400

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -199,12 +199,14 @@ proc handleIHave*(g: GossipSub,
     # TODO review deduplicate algorithm
     # * https://github.com/nim-lang/Nim/blob/5f46474555ee93306cce55342e81130c1da79a42/lib/pure/collections/sequtils.nim#L184
     # * it's probably not efficient and might give preference to the first dupe
-    var deIhaves = ihaves.deduplicate()
-    for ihave in deIhaves.mitems:
+    let deIhaves = ihaves.deduplicate()
+    for ihave in deIhaves:
       trace "peer sent ihave",
         peer, topic = ihave.topicID, msgs = ihave.messageIDs
       if ihave.topicID in g.mesh:
-        for m in ihave.messageIDs:
+        # also avoid duplicates here!
+        let deIhavesMsgs = ihave.messageIDs.deduplicate()
+        for m in deIhavesMsgs:
           let msgId = m & g.randomBytes
           if msgId notin g.seen:
             if peer.iHaveBudget > 0:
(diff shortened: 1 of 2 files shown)


sinkingsugar:
Includes a deduplication fix as well

sinkingsugar:
@dryajov quick look when you have time so we can merge please!

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -199,12 +199,14 @@ proc handleIHave*(g: GossipSub,
     # TODO review deduplicate algorithm
     # * https://github.com/nim-lang/Nim/blob/5f46474555ee93306cce55342e81130c1da79a42/lib/pure/collections/sequtils.nim#L184
     # * it's probably not efficient and might give preference to the first dupe
-    var deIhaves = ihaves.deduplicate()
-    for ihave in deIhaves.mitems:
+    let deIhaves = ihaves.deduplicate()
```
dryajov, line 202: maybe we should just use a `HashSet`?

dryajov (approved):

libp2p/protocols/pubsub/gossipsub/behavior.nim
sinkingsugar: good point but the ideal/regular size is small tho, so hashset would be just a burden, in the case of a rogue payload could be big and things would be O(n) tho

sinkingsugar:
There is already a comment about `deduplicate` investigation, I think it's fairly low priority for now but to keep in mind.

PR #539, opened 2021-03-02
Gossipsub fixes and Initiator flagging fixes

sinkingsugar:
This PR solves very critical issues in gossipsub, specifically related to incoming connections and peers quality.

### DIFF
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -151,11 +151,7 @@ proc handle*(p: PubSubPeer, conn: Connection) {.async.} =
     debug "exiting pubsub read loop",
       conn, peer = p, closed = conn.closed
 
-proc connectOnce(p: PubSubPeer): Future[bool] {.async.} =
-  # I hate using result but return would fail with
-  # compiler error: Error: unhandled exception: closureiters.nim(828, 11) `ctx.nearestFinally != 0`  [AssertionError]
-  result = false
-
+proc connectOnce(p: PubSubPeer): Future[void] {.async.} =
   try:
     let newConn = await p.getConn()
     if newConn.isNil:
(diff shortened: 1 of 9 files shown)


libp2p/stream/connection.nim
```diff
@@ -33,6 +33,7 @@ type
     peerInfo*: PeerInfo
     observedAddr*: Multiaddress
     upgraded*: Future[void]
+    outbound*: bool
```
arnetheduck, line 36: why do we have both `dir` as a direction enum and outbound as a `bool`?

sinkingsugar: well as you can see from the diff I used to use `dir`, but that's not the actual wire direction.

arnetheduck: well, the point was more, why is it not using the direction enum?

also, I presume that what we want to capture here is whether or not the underlying connection is outbound or not, and not the mplex stream on top - this should be documented clearly along with a reason why / pointer to the spec, because it's hard to follow. 

finally, the name could be used to indicate some of this as well, ie `transportDirection` or something - conceptually, it goes a bit against the abstraction of channels and connections being layered - basically, once a transport is established, one doesn't "normally" care in what direction the underlying stream was established, or indeed how many layers of envelopes there are (in theory, on could run mplex inside mplex inside noise)

sinkingsugar:
There are some peer counting issues, I might know where the issue is but I need to give it a test

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -189,9 +195,12 @@ proc connectImpl(p: PubSubPeer) {.async.} =
     # issue so we try to get a new on
     while true:
       await connectOnce(p)
-
+  except CancelledError:
+      raise
```
dryajov, line 199: You probably want to drop the peer on cancellation as well.

sinkingsugar:
Issues fixed, good to go now.

dryajov (approved):
LGTM

PR #542, opened 2021-03-09
Split dialer

dryajov:
- Move dialing logic from switch to its own module
- Allow transports drive their own upgrade flow by exposing `upgradeIncoming` and `upgradeOutgoing`

TODO:
- [ ] The dialer is now it's own module so it can be tested in isolation, but for now it's covered as part of the switch tests

### DIFF
--- a/libp2p/dial.nim
+++ b/libp2p/dial.nim
@@ -1,0 +1,38 @@
+## Nim-LibP2P
+## Copyright (c) 2021 Status Research & Development GmbH
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+import chronos
+import peerid,
+       stream/connection
+
+type
+  Dial* = ref object of RootObj
+
+method connect*(
+  self: Dial,
+  peerId: PeerID,
+  addrs: seq[MultiAddress]) {.async, base.} =
+  ## attempt to create establish a connection
+  ## with a remote peer
+  ##
+
+  doAssert(false, "Not implemented!")
+
+method dial*(
+  self: Dial,
+  peerId: PeerID,
+  protos: seq[string]): Future[Connection] {.async, base.} =
+  doAssert(false, "Not implemented!")
+
+method dial*(
+  self: Dial,
+  peerId: PeerID,
+  addrs: seq[MultiAddress],
+  protos: seq[string]): Future[Connection] {.async, base.} =
+  doAssert(false, "Not implemented!")
--- a/libp2p/dial.nim
+++ b/libp2p/dial.nim
@@ -18,8 +18,8 @@ method connect*(
   self: Dial,
   peerId: PeerID,
   addrs: seq[MultiAddress]) {.async, base.} =
-  ## attempt to create establish a connection
-  ## with a remote peer
+  ## connect remote peer without negotiating
+  ## a protocol
   ##
 
   doAssert(false, "Not implemented!")
@@ -28,6 +28,10 @@ method dial*(
   self: Dial,
   peerId: PeerID,
   protos: seq[string]): Future[Connection] {.async, base.} =
+  ## create a protocol stream over an
+  ## existing connection
+  ##
+
   doAssert(false, "Not implemented!")
 
 method dial*(
@@ -35,4 +39,8 @@ method dial*(
   peerId: PeerID,
   addrs: seq[MultiAddress],
   protos: seq[string]): Future[Connection] {.async, base.} =
+  ## create a protocol stream and establish
+  ## a connection if one doesn't exist already
+  ##
+
   doAssert(false, "Not implemented!")
--- a/libp2p/dialer.nim
+++ b/libp2p/dialer.nim
@@ -152,8 +152,8 @@ method connect*(
   self: Dialer,
   peerId: PeerID,
   addrs: seq[MultiAddress]) {.async.} =
-  ## attempt to create establish a connection
-  ## with a remote peer
+  ## connect remote peer without negotiating
+  ## a protocol
   ##
 
   if self.connManager.connCount(peerId) > 0:
@@ -177,6 +177,10 @@ method dial*(
   self: Dialer,
   peerId: PeerID,
   protos: seq[string]): Future[Connection] {.async.} =
+  ## create a protocol stream over an
+  ## existing connection
+  ##
+
   trace "Dialing (existing)", peerId, protos
   let stream = await self.connManager.getStream(peerId)
   if stream.isNil:
@@ -189,6 +193,10 @@ method dial*(
   peerId: PeerID,
   addrs: seq[MultiAddress],
   protos: seq[string]): Future[Connection] {.async.} =
+  ## create a protocol stream and establish
+  ## a connection if one doesn't exist already
+  ##
+
   var
     conn: Connection
     stream: Connection
--- a/libp2p/dialer.nim
+++ b/libp2p/dialer.nim
@@ -1,0 +1,233 @@
+## Nim-LibP2P
+## Copyright (c) 2021 Status Research & Development GmbH
+## Licensed under either of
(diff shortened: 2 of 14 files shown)


sinkingsugar:
had a quick look, will check it better again tho.
Shouldn't this be targeting unstable tho?
It turned out to be a big refactor so should be not too dangerous but still it touched many components.

dryajov:
> had a quick look, will check it better again tho.
> Shouldn't this be targeting unstable tho?
> It turned out to be a big refactor so should be not too dangerous but still it touched many components.

Sure, we can move it to unstable, but consider that this will need to be merged to master relatively quickly, because some projects (waku), need support for additional transports which this PR is addressing in part.

sinkingsugar:
> > had a quick look, will check it better again tho.
> > Shouldn't this be targeting unstable tho?
> > It turned out to be a big refactor so should be not too dangerous but still it touched many components.
> 
> Sure, we can move it to unstable, but consider that this will need to be merged to master relatively quickly, because some projects (waku), need support for additional transports which this PR is addressing in part.

Well I'd suggest Waku using unstable in such case, I mean we should consider that Nimbus is high security and either leave master as "safe" or create a "safe" branch.
And by "safe" I mean slow pace on big changes in order to be noticed and tested on the testnets.

dryajov:
> > > had a quick look, will check it better again tho.
> > > Shouldn't this be targeting unstable tho?
> > > It turned out to be a big refactor so should be not too dangerous but still it touched many components.
> > 
> > 
> > Sure, we can move it to unstable, but consider that this will need to be merged to master relatively quickly, because some projects (waku), need support for additional transports which this PR is addressing in part.
> 
> Well I'd suggest Waku using unstable in such case, I mean we should consider that Nimbus is high security and either leave master as "safe" or create a "safe" branch.
> And by "safe" I mean slow pace on big changes in order to be noticed and tested on the testnets.

This is actually, a very important discussion to have, one we should address in more detail, but suffice to say that we need a consistent way of moving things forward without affecting the users of this library. Nimbus, although a very important consumer of the library is one of (very soon) many such consumer (waku and dagger just to name a couple and I'm sure others will follow from both, inhouse projects from status and third parties). 

There are several issues with keeping changes like this siloed, bit roth being one of them, but more importantly without having a way of testing `unstable` with Nimbus (or others) is of not much use on it's own. On the other hand, Nimbus doesn't have to jump on this change right away, because it can always pin to a specific version/commit, just like others will have to at some point.

So, what are our options? Let me list a few ideas to consider/discuss.

- First, each project needs to take care of its own testing and keep track of which versions are stable for them. 
- Second, it probably makes sense to start splitting out some parts of the library out. For example, there is no reason to keep Pubsub as part of this repo, because it is a large enough chunk of functionality that makes sense to evolve separately, same for things like DHTs, Relays and anything else that is large enough and self contained.
- Third, we need a way to test `unstable` with our other projects. There is no point in keeping it around unless we can test against it and this is where I ask for your help/input - do we have a dedicated box where we can test this with Nimbus? If not we need to request one from @jakubgs (or whoever else needs to approve it).

I'll switch the merge to `unstable` for now, but like I said above, let's have a way of testing it so we can merge back to master safely and soon.

arnetheduck:
> start splitting some parts of the library out

this would be nuts - it's nearly impossible to compose a functioning cross-repo version already (nim + chronos + ssl etc) - adding multiple repos will further exacerbate these issues, and will make it harder to fix many of the underlying architectural causes for the issues we're seeing today (leaks, poor error handling etc) which require cross-module fixes.

>  other projects

if a change breaks libp2p for the existing projects such as Nimbus, something is _wrong already_ - you don't need more projects to prove that.

> unstable

indeed, having `unstable` in nim-libp2p feels off - mostly because there's no `stable` yet, really (from either of api/abi/bugs point of view) - having an unstable at this stage basically promotes writing code that breaks existing projects - if anything, it would make more sense to keep long-running feature branches and gradually move the stable parts of those feature branches into a stable branch and keep rebasing those feature branches on top of that stable branch.

doing so will allow special cases that need access to the feature to use that feature in isolation while the rest of libp2p stays stable and without massive disruptive changes - this also fits better into the existing infrastructure where there are special libp2p tests of nimbus - it's not a problem to spin up more such branches, but it seems hard enough to keep one libp2p branch of nimbus stable enough - not sure adding more is going to make it easier.

dryajov:
> this would be nuts - it's nearly impossible to compose a functioning cross-repo version already (nim + chronos + ssl etc) - adding multiple repos will further exacerbate these issues, and will make it harder to fix many of the underlying architectural causes for the issues we're seeing today (leaks, poor error handling etc) which require cross-module fixes.

> you don't need more projects to prove that.

But there is also 0 reasons to have things like gossip be part of the core and re-ship the core every time something in gossip changes. For all intents and purposes Gossipsub is just like Nimbus in this sense, it's just an application on top of the libp2p core it just happens to have a spec from the libp2p project instead of Ethereum.

>  if anything, it would make more sense to keep long-running feature branches and gradually move the stable parts of those feature branches into a stable branch and keep rebasing those feature branches on top of that stable branch.

Err, what's wrong with having `unstable` in general? If waku needs to use features X,Y,Z and Dagger needs features A,B,C - then how are this projects suppose to operate without manually keeping track of this branches? This adds unnecessary burden to other projects. We need to keep everybody's needs in mind and not specialize for a particular project's needs.

There is nothing wrong with having both, a short lived future branch that gets tested in isolation and a longer lived `unstable` branch with features that will make it into master soonish. 

If it will really make you happy, we can always change the `unstable` to something else ;-)

The point of my first comment was to say that `unstable` needs to be tested just as much as `master` does and all we need for that is a meager aws instance to deploy this too... Don't think it's too much to ask?

dryajov:
Forgot to add, that I tend to agree that splitting now is probably premature, but it is something that will need on the medium to long run, once we're a bit more stable.

arnetheduck:
> If waku needs to use features X,Y,Z and Dagger needs features A,B,C - then how are this projects supposed to operate without manually keeping track of this branches?

so.. which of `Dagger` and `Waku` should use.. `unstable` when they want competing feature sets that are unstable/unusable together? 

ie the stable/unstable split in nimbus-eth2 is intended for _end users_ vs developers - libp2p has no end users - `unstable` doesn't solve any particular demand beyond what feature branches already solve, which more or less is "hold this code for me while I test it" - it does introduce risk though that more features progress along incompatible branches because they don't get stressed by actual usage.

> servers

it's generally not a problem to set up more servers assuming that they will get the attention they deserve (because by and large, we can tolerate some downtime due to bugs, but to meaningfully test things, they need to be running validators that the community ends up depending on)  - the easiest way is to open a concrete proposal on https://github.com/status-im/infra-nimbus after having created a corresponding auto-bump - the prater testnet setup will be happening soon: https://github.com/status-im/infra-nimbus/issues/41

dryajov:
> they want competing feature sets

Why would they be competing?

>  that are unstable/unusable together?

Not necessarily, they might be unrelated but live in the same codebase. For example this dialer split that allows more transports to be used and some unrelated change to pubsub/gossipsub. This is also further strengthens the point about separating  functionally unrelated pieces into their own repos/packages, the more isolation the less we have to worry about this, but we obviously do for now.

> ie the stable/unstable split in nimbus-eth2 is intended for end users vs developers - libp2p has no end users - unstable doesn't solve any particular demand beyond what feature branches already solve, which more or less is "hold this code for me while I test it" - it does introduce risk though that more features progress along incompatible branches because they don't get stressed by actual usage.

`unstable` in libp2p means features that are big enough that require testing without disrupting `master`. What I ultimately want to avoid, is to merge large changes to master without first testing them rigorously to avoid conflicting with potential hot-fixes and less disruptive but important changes that need to be shipped ASAP.

dryajov:
BTW, we don't necessarily need to split the repo itself, as long as they can be treated as separate packages and shipped independently, ie a monorepo, but it's not possible right now because we lack a reliable package manager and the monorepo is not viable with submodules which breaks NBC right now. I'm also not a big fan of monorepos in general.

dryajov:
Issue to create a new testing box is created here - https://github.com/status-im/infra-nimbus/issues/43

sinkingsugar:
In any case I added an unstable auto bump (draft) to nimbus. Not meant to be merged but meant to pass finalization testing.

sinkingsugar (approved):
LGTM, would be nice to test it on nbc soon

PR #546, opened 2021-03-15
add mounted protos to identify message

dryajov:


arnetheduck:
hm, this causes some privacy issues  because we're now leaking information that is not strictly necessary for the client to function, which might be used for example to do targeted bug exploits

dryajov:
Depends on clients, waku apparently uses this and there might other apps that rely on it.

sinkingsugar:
Is this by spec? this indeed opens up fingerprinting quite easily.

dryajov:
> Is this by spec?

Yes, it is in the spec.

> this indeed opens up fingerprinting quite easily.

It doesn't open anything, at worst it adds a new way to fingerprint, but at this point the information being disclosed elsewhere is much more explicit than this (ENR and similar, multistream `Ls` message or just plain probing the client for random stuff), on the other hand, this will hinder interop for other apps. If you want to disable this for NBC specifically, then by all means we can probably add a flag to control it, but in that case I suggest disabling the `Ls` message on multistream select as well.

sinkingsugar:
True, `ls` is still there anyway

PR #547, opened 2021-03-19
Raises

dryajov:
First attempt to use `raises` in `async` procs, uses https://github.com/status-im/nim-chronos/pull/166.


dryajov:
I'm not sure how to make nim pick the correct version of chronos when using the branch name, it seems to ignore the branch in the nimble file and just pick whatever looks like the latest numeric version so CI is broken for now. But it works in the nimbus repo if the proper branches in chronos (`eh-tracking`) and stew (latest head) are checked out.

arnetheduck:
It might make the transition more smooth if the refactoring to raise a single exception type was separated from the chronos exception adaptation.

> pick the correct version

it is using the correct chronos branch - the error is legit, you need to annotate _all_ `method` with the same `raises` annotations or compiler will freak out.

dryajov:
> it is using the correct chronos branch - the error is legit

It's not, it's picking the wrong chronos and this is verifiable by `nimble path chronos`. 

> you need to annotate _all_ `method` with the same `raises` annotations or compiler will freak out.

This is well understood and it is what the branch is attempting to do. However, since `eh-tracking` is/was still changing, additional issues might have been introduced from the point when this branch was created.

arnetheduck:
> It's not,

no idea what `nimble` does locally, it tends to give a random version then, depending on what's in your nimble cache, but the error message in CI is legit for the eh-tracking branch (it's complaining about `CatchableError` and not `Exception`)

libp2p/stream/connection.nim
```diff
@@ -53,7 +56,8 @@ proc onUpgrade*(s: Connection) {.async.} =
   if not isNil(s.upgraded):
     await s.upgraded
 
-func shortLog*(conn: Connection): string =
+func shortLog*(conn: Connection): string
+  {.raises: [Defect, ValueError].} =
```
arnetheduck, line 60: `shortLog` should never raise - in fact it doesn't _actually_ except that `&` raises - the right thing to do here is to catch the `ValueError` and turn it into a `Defect`

arnetheduck:
> attempting to do

`LPChannel.readOnce` is an example of a `method` that is not annotated in this branch (annotating the base is not enough) - there's plenty of them in the branch though that remain unannotated - the CI error I happened to look at was because of this.

libp2p/connmanager.nim
```diff
@@ -105,15 +109,34 @@ proc addConnEventHandler*(c: ConnManager,
...
 
-  if isNil(handler): return
-  c.connEvents.mgetOrPut(kind,
-    initOrderedSet[ConnEventHandler]()).incl(handler)
+  try:
+    if isNil(handler): return
+    c.connEvents.mgetOrPut(kind,
+      initOrderedSet[ConnEventHandler]()).incl(handler)
+  except Exception as exc:
+    # TODO: there is an Exception being raised
```
arnetheduck, line 117: this is probably https://github.com/nim-lang/Nim/issues/17382 - looks like a nim bug, can use it in the comment to highlight where it happens

arnetheduck:
looks like there's a copy of `unittest2` in the branch - this can be replaced with `requires "https://github.com/status-im/nim-unittest2.git#head"`

dryajov:
> > attempting to do
> 
> `LPChannel.readOnce` is an example of a `method` that is not annotated in this branch (annotating the base is not enough) - there's plenty of them in the branch though that remain unannotated - the CI error I happened to look at was because of this.

There is a `{.push raises: [Defect].}` enabled for the entire module, shouldn't that apply to all the methods?

dryajov:
> looks like there's a copy of `unittest2` in the branch - this can be replaced with `requires "https://github.com/status-im/nim-unittest2.git#head"`

Yes, I ran into the issue with unhandled `Exception` coming from the formaters which I worked around locally.

arnetheduck:
>  all the methods?

no, it only applies to `proc` from what I can tell (might be another bug, who knows) - each and every `method` needs an explicit raises annotation (that matches the base exactly).

arnetheduck:
> Did you run it 

yes, but that doesn't mean it works the same with unit tests and nimble etc - the eh tracking issues issues sometimes seem to depend on the order in which the compiler sees the code which might be different depending on include path order etc - it's easiest to just annotate everything consistently

dryajov:
> > all the methods?
> 
> no, it only applies to `proc` from what I can tell

Seems to work for `methods` just as well from my experiments with this branch. However if it's not the case and you have a concrete case, we need to report it upstream and apply the fix accordingly.

libp2p/connmanager.nim
arnetheduck: also, it incorrectly raises whatever `handler` raises - since you see `Exception` here it means `handler` itself is missing a correct raises signature

arnetheduck:
> Seems to work for methods just as well from my experiments with this branch. 

ok, maybe I hit something else, not sure about this one actually

dryajov:
> > Did you run it
> 
> yes, but that doesn't mean it works the same with unit tests and nimble etc - 

Well, just to clarify, the point of this branch was to get it to a) compile and b) to pass the unittests. I'm sure that we'll uncover many more issues when using with our other projects.


> the eh tracking issues issues sometimes seem to depend on the order in which the compiler sees the code which might be different depending on include path order etc - it's easiest to just annotate everything consistently

Well, I'm not arguing with this, but it seems to deviate with the already established practice of module level annotations. If you follow the 


> > Seems to work for methods just as well from my experiments with this branch.
> 
> ok, maybe I hit something else, not sure about this one actually

But it also seemed like I hadn't yet pushed a fix for annotating lpstream methods, so they are now. I think I did it procesilly because the compiler got messed up when doing different overrides/overloads.

libp2p/stream/connection.nim
dryajov: Yeah, quick and dirty fix to get things to compile on first pass, but I suspected that this wasn't the right approach.

libp2p/connmanager.nim
dryajov: `ConnEventHandler` is defined as:

```nim
  ConnEventHandler* =
    proc(peerId: PeerID, event: ConnEvent): Future[void]
      {.gcsafe, raises: [Defect].}
```

But interestingly, it seems to ignore that when actually declaring the handler which is declared in https://github.com/status-im/nim-libp2p/blob/unstable/tests/testswitch.nim#L249. Another bug?

libp2p/daemon/daemonapi.nim
```diff
@@ -570,7 +576,8 @@ proc newDaemonApi*(flags: set[P2PDaemonFlags] = {},
                    gossipsubHeartbeatDelay = 0,
                    peersRequired = 2,
                    logFile = "",
-                   logLevel = IpfsLogLevel.Debug): Future[DaemonAPI] {.async.} =
+                   logLevel = IpfsLogLevel.Debug): Future[DaemonAPI]
+                   {.async, raises: [Defect, DaemonLocalError].} =
```
arnetheduck, line 580: actually, this is where things get messy with the current chronos patch: the correct annotation for `async` proc's is actually `raises: [Defect]` - I've described it a bit here: https://github.com/status-im/nim-chronos/pull/166/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R225

effectively, the `eh-tracking` branch doesn't work with "logically correct" raises annotations (yet) - it looks at the raises of the transformed function - this will need more work in the future - for now, just put `push raises: [Defect]` on top and nothing on `async` functions..

dryajov: Hmm, actually it doesn't seem to be the case and if the exception is omitted from the annotation it complains about a leaking generic `Exception`, but  not the concrete type however, after adding the concrete exception type it seems to silence it. It would be good if you try it yourself on that module since I did initially think that all we can work with is `Defect`.

Here is the wording from that link for convenience.

```
### Exception effects

`chronos` currently offers minimal support for exception effects and `raises`
annotations. In general, during the `async` transformation, a generic
`except CatchableError` handler is added around the entire function being
transformed, in order to catch any exceptions and transfer them to the `Future`.
Because of this, the effect system thinks no exceptions are "leaking" because in
fact, exception _handling_ is deferred to when the future is being read.

Effectively, this means that while code can be compiled with
`{.push raises: [Defect]}`, the intended effect propoagation and checking is
**disabled** for `async` functions.
```

dryajov:
@arnetheduck @stefantalpalaru  `unittest2` seems to use `setProgramResult`, but I can't find it defined anywhere in the nim sources and it fails to build in some places. It seems to be a recent addition as of 1.4.x - https://nim-lang.org/docs/exitprocs.html?

stefantalpalaru:
> It seems to be a recent addition as of 1.4.x

Good catch. It comes from merging the latest "unittest" code. We do test on "version-1-2", "version-1-4" and "devel" branches on Nim, but that code path is never taken in our tests (it would require calling `fail()` outside a unit test).

dryajov:
> > It seems to be a recent addition as of 1.4.x
> 
> Good catch. It comes from merging the latest "unittest" code. We do test on "version-1-2", "version-1-4" and "devel" branches on Nim, but that code path is never taken in our tests (it would require calling `fail()` outside a unit test).

It actually does and it seems to happen when using `check` outside of a `test` template, this is useful with parameterized tests, which is where I'm running into it. You can see it for yourself in the `testinterop.nim` file in this branch.

stefantalpalaru:
Fixed in https://github.com/status-im/nim-unittest2/pull/4

libp2p/daemon/daemonapi.nim
sinkingsugar: That's weird, maybe cos its a closure? Did you try extracting a proc?

arnetheduck:
since you're working on this, see if you can make the tests work in strict mode as well (https://github.com/status-im/nim-chronos/#exception-effects)

libp2p/daemon/daemonapi.nim
```diff
@@ -7,9 +7,12 @@
 ## This file may not be copied, modified, or distributed except according to
 ## those terms.
 
+{.push raises: [Defect].}
+
 ## This module implementes API for `go-libp2p-daemon`.
 import std/[os, osproc, strutils, tables, strtabs]
 import chronos, chronicles
+import chronos/selectors2 # TODO: workaround for missing `IOSelectorsException`
```
arnetheduck, line 15: this should hopefully no longer be needed

libp2p/dialer.nim
```diff
@@ -176,7 +177,8 @@ proc negotiateStream(
 method dial*(
   self: Dialer,
   peerId: PeerID,
-  protos: seq[string]): Future[Connection] {.async.} =
+  protos: seq[string]): Future[Connection]
+  {.async, raises: [Defect, DialFailedError].} =
```
arnetheduck, line 181: ditto, anything async should only have `{.raises: [Defect].}`

libp2p/protocols/secure/secio.nim
```diff
@@ -71,9 +74,13 @@ type
...
 func shortLog*(conn: SecioConn): auto =
-  if conn.isNil: "SecioConn(nil)"
-  elif conn.peerInfo.isNil: $conn.oid
-  else: &"{shortLog(conn.peerInfo.peerId)}:{conn.oid}"
+  try:
+    if conn.isNil: "SecioConn(nil)"
+    elif conn.peerInfo.isNil: $conn.oid
+    else: &"{shortLog(conn.peerInfo.peerId)}:{conn.oid}"
+  except ValueError as exc:
+    raise newException(Defect, exc.msg)
```
arnetheduck, line 82: `raiseAssert exc.msg` is a useful shortcut

libp2p/switch.nim
```diff
@@ -260,21 +272,21 @@ proc newSwitch*(peerInfo: PeerInfo,
   switch.mount(identity)
   return switch
 
-proc isConnected*(s: Switch, peerInfo: PeerInfo): bool
-  {.deprecated: "Use PeerID version".} =
-  not isNil(peerInfo) and isConnected(s, peerInfo.peerId)
+# proc isConnected*(s: Switch, peerInfo: PeerInfo): bool
```
sinkingsugar, line 275: remove? uncomment?

.github/workflows/ci.yml
```diff
@@ -96,20 +97,20 @@ jobs:
           echo "::set-output name=nim::$nimHash"
           echo "::set-output name=csources::$csourcesHash"
 
-      - name: Restore prebuilt Nim from cache
-        id: nim-cache
-        uses: actions/cache@v1
-        with:
-          path: nim
-          key: "nim-${{ matrix.target.os }}-${{ matrix.target.cpu }}-${{ steps.versions.outputs.nim }}"
+      # - name: Restore prebuilt Nim from cache
```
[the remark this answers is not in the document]
dryajov: ah, good catch

libp2p/connmanager.nim
```diff
@@ -139,15 +160,33 @@ proc addPeerEventHandler*(c: ConnManager,
...
   ##
 
-  if isNil(handler): return
-  c.peerEvents.mgetOrPut(kind,
-    initOrderedSet[PeerEventHandler]()).incl(handler)
+  try:
+    if isNil(handler): return
+    c.peerEvents.mgetOrPut(kind,
+      initOrderedSet[PeerEventHandler]()).incl(handler)
+  except Exception as exc:
```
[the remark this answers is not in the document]
dryajov: Well, languages/std libs have bugs (not only in Nim), if this is it, then we'll probably update the comment to something that points it out, but for now, I've no idea what the actual issue is, so....

libp2p/connmanager.nim
```diff
@@ -169,13 +208,14 @@ proc triggerPeerEvents*(c: ConnManager,
     trace "triggering peer events", peerId, event = $event
 
     var peerEvents: seq[Future[void]]
-    for h in c.peerEvents[event.kind]:
-      peerEvents.add(h(peerId, event))
+    try:
+      for h in c.peerEvents[event.kind]:
+        peerEvents.add(h(peerId, event))
+    except Exception as exc:
```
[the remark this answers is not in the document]
dryajov: mm, this is most likely a bug/oversight in std lib and I'd rather it be as explicit as possible so we can clean it up once/if the issue is fixed upstream.

libp2p/switch.nim
dryajov: Probably remove, it's been deprecated for a while and we don't use it anywhere anymore, but waku might for example...

libp2p/connmanager.nim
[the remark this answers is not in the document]
sinkingsugar: Oh yeah I totally get it

libp2p/connmanager.nim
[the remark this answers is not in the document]
sinkingsugar: Maybe comment each of those cases then

dryajov:
I'm a bit conflicted about this branch - it gets us very close to _proper_ raises annotations, but as @arnetheduck points out, chronos only allows for `Defect` right now, so those are neither useful nor guaranteed to be correct. 

When I started implementing this branch, a few of the specific exceptions would be flagged and adding them to the raises list fixed the issues, but most likely it was something else either in my code or simply because I was working against the in progress chronos branch.

In any case, there is a lot of good cleanup here and like I said most of the specific raises annotations actually do make sense, but alas aren't currently verifiable/enforceable. 

To move this forward, we can either 

- extract all the changes sans the specific exception type annotations into its own branch and leave this one around for when chronos can handle specific annotations
- merge this as is and fix the issues once chronos adds the ability to handle the specific annotations

The former option would simply mean removing `raises: [...]` from all methods/procs and I'm inclined to say it's probably the safest thing to do right now.

dryajov:
> I'm a bit conflicted about this branch - it gets us very close to _proper_ raises annotations, but as @arnetheduck points out, chronos only allows for `Defect` right now, so those are neither useful nor guaranteed to be correct.
> 
> When I started implementing this branch, a few of the specific exceptions would be flagged and adding them to the raises list fixed the issues, but most likely it was something else either in my code or simply because I was working against the in progress chronos branch.
> 
> In any case, there is a lot of good cleanup here and like I said most of the specific raises annotations actually do make sense, but alas aren't currently verifiable/enforceable.
> 
> To move this forward, we can either
> 
> * extract all the changes sans the specific exception type annotations into its own branch and leave this one around for when chronos can handle specific annotations
> * merge this as is and fix the issues once chronos adds the ability to handle the specific annotations
> 
> The former option would simply mean removing `raises: [...]` from all methods/procs and I'm inclined to say it's probably the safest thing to do right now.

Actually, I took another look at this and I think that this PR should generally work as expected. Here is an example of an expanded async proc:

```nim
proc readExactly*(s: LPStream; pbytes: pointer; nbytes: int): Future[void] {.
    raises: [Defect, LPStreamEOFError, LPStreamIncompleteError], stackTrace: off.} =
  var chronosInternalRetFuture = newFuture[void]("readExactly")
  iterator readExactly_65120132(): FutureBase {.closure,
      raises: [Defect, CatchableError].} =
    block:
      template result(): auto {.used.} =
        {.fatal: "You should not reference the `result` variable inside a void async proc".}

      if s.atEof:
        raise newLPStreamEOFError()
...
```

`readExactly` can only raise `[Defect, LPStreamEOFError, LPStreamIncompleteError]`, the iterator `readExactly_65120132` can raise `[Defect, CatchableError]`, which is less specific and allows any exception of types `Defect` and `CatchableError` to escape.

Any proc called by the iterator can raise an exception of type `Defect` and `CatchableError`, but `readExactly` itself or the procs called by it can only raise `[Defect, LPStreamEOFError, LPStreamIncompleteError]`, so on and so forth, thus the chain of exception annotations propagates correctly across all invocations of `async`.

sinkingsugar:
> > I'm a bit conflicted about this branch - it gets us very close to _proper_ raises annotations, but as @arnetheduck points out, chronos only allows for `Defect` right now, so those are neither useful nor guaranteed to be correct.
> > When I started implementing this branch, a few of the specific exceptions would be flagged and adding them to the raises list fixed the issues, but most likely it was something else either in my code or simply because I was working against the in progress chronos branch.
> > In any case, there is a lot of good cleanup here and like I said most of the specific raises annotations actually do make sense, but alas aren't currently verifiable/enforceable.
> > To move this forward, we can either
> > 
> > * extract all the changes sans the specific exception type annotations into its own branch and leave this one around for when chronos can handle specific annotations
> > * merge this as is and fix the issues once chronos adds the ability to handle the specific annotations
> > 
> > The former option would simply mean removing `raises: [...]` from all methods/procs and I'm inclined to say it's probably the safest thing to do right now.
> 
> Actually, I took another look at this and I think that this PR should generally work as expected. Here is an example of an expanded async proc:
> 
> ```nim
> proc readExactly*(s: LPStream; pbytes: pointer; nbytes: int): Future[void] {.
>     raises: [Defect, LPStreamEOFError, LPStreamIncompleteError], stackTrace: off.} =
>   var chronosInternalRetFuture = newFuture[void]("readExactly")
>   iterator readExactly_65120132(): FutureBase {.closure,
>       raises: [Defect, CatchableError].} =
>     block:
>       template result(): auto {.used.} =
>         {.fatal: "You should not reference the `result` variable inside a void async proc".}
> 
>       if s.atEof:
>         raise newLPStreamEOFError()
> ...
> ```
> 
> `readExactly` can only raise `[Defect, LPStreamEOFError, LPStreamIncompleteError]`, the iterator `readExactly_65120132` can raise `[Defect, CatchableError]`, which is less specific and allows any exception of types `Defect` and `CatchableError` to escape.
> 
> Any proc called by the iterator can raise an exception of type `Defect` and `CatchableError`, but `readExactly` itself or the procs called by it can only raise `[Defect, LPStreamEOFError, LPStreamIncompleteError]`, so on and so forth, thus the chain of exception annotations propagates correctly across all invocations of `async`.

As I said on discord, I think only the inner iterator matters in fact, and that is the root of the issue.

arnetheduck:
To understand why the annotation is wrong, you need to read the whole transformed function:
```nim
proc readExactly*(rstream: AsyncStreamReader; pbytes: pointer; nbytes: int): Future[
    void] {.stackTrace: off.} =
  var chronosInternalRetFuture = newFuture[void]("readExactly")
  iterator readExactly_25840371(): FutureBase {.closure,
      raises: [Defect, CatchableError, Exception].} =
    block:
      template result(): auto {.used.} =
        {.fatal: "You should not reference the `result` variable inside a void async proc".}
    ...
    complete(chronosInternalRetFuture)

  var nameIterVar`gensym25840373 = readExactly_25840371
  {.push, stackTrace: off.}
  var readExactly_continue_25840372: proc (udata`gensym25840374: pointer) {.gcsafe,
      raises: [Defect].}
  readExactly_continue_25840372 = proc (udata`gensym25840375: pointer) {.gcsafe,
      raises: [Defect].} =
    try:
      if not (nameIterVar`gensym25840373.finished()):
        var next`gensym25840376 = nameIterVar`gensym25840373()
        while (not next`gensym25840376.isNil()) and
            next`gensym25840376.finished():
          next`gensym25840376 = nameIterVar`gensym25840373()
          if nameIterVar`gensym25840373.finished():
            break
        if next`gensym25840376 == nil:
          if not (chronosInternalRetFuture.finished()):
            const
              msg`gensym25840377 = "Async procedure (&" & "readExactly" &
                  ") yielded `nil`, " &
                  "are you await\'ing a `nil` Future?"
            raiseAssert msg`gensym25840377
        else:
          next`gensym25840376.addCallback(readExactly_continue_25840372)
    except CancelledError:
      chronosInternalRetFuture.cancelAndSchedule()
    except CatchableError as exc`gensym25840378:
      chronosInternalRetFuture.fail(exc`gensym25840378)
    except Exception as exc`gensym25840379:
      if exc`gensym25840379 of Defect:
        raise (ref Defect)(exc`gensym25840379)
      chronosInternalRetFuture.fail((ref ValueError)(msg: exc`gensym25840379.msg,
          parent: exc`gensym25840379))
  readExactly_continue_25840372(nil)
  {.pop.}
  return chronosInternalRetFuture
```

In particular, you need to understand that the transformed `readExactly` that gets called never raises - in fact, it would be wrong for it to do so - the contract of functions that return a `Future` is that they don't raise themselves but rather move all exceptions _inside_ the future - you can call these functions using `await`, or not, and you can call them outside `async`, or inside - it doesn't matter - they return a future and _must never_ raise themselves - this would create two error paths for them, and this is wrong - hence the correct annotation on them is `raises: [Defect]`. 

If you read on, you will notice that `readExactly_continue` catches all Catchables and puts them in the future, for this reason. If you were to create a function that returns a `Future` "manually" without `{.async.}` it would be your responsibility to ensure this function doesn't raise and instead does the same try/except.

By adding the annotation, you're creating a lie that will impact code down the line, because `readExactly`, after having been transformed, _doesn't_ raise - downstream code will pass this misinformation on and code that uses libp2p will have to add workarounds to clean up this lie.

arnetheduck:
> extract all the changes sans the specific exception type annotations into its own branch and leave this one around for when chronos can handle specific annotations

this is the path forward right now that causes the least issues for future code - it would be good to put the code in a cleaned up branch so we can remove the hacks from nbc as well - a new PR is probably appropriate because there's a lot of comments in this branch that come from pre-release versions of chronos 3.

arnetheduck:
Obsoleted by https://github.com/status-im/nim-libp2p/pull/1050 et al.

PR #549, opened 2021-03-22
put conditional locks on nbc bump

sinkingsugar:


.github/workflows/ci.yml
```diff
@@ -207,6 +207,7 @@ jobs:
           nimble test
 
   bumpNBC-stable:
+    if: github.ref == 'refs/heads/master'
```
dryajov, line 210: We should change to `master` to `stable`.

jakubgs: Agreed.

sinkingsugar: I propose we keep master for backward compatibility and just leave it as it is tho, does that work for you @dryajov ?
And of course set stable as default.

sinkingsugar: Altho dunno, ref might be enough, someone pulling master might be better off with error message to know they have to change.

PR #550, opened 2021-03-22
Fix minor exception issues

arnetheduck:
Makes code compatible with
https://github.com/status-im/nim-chronos/pull/166 without requiring it.

### DIFF
--- a/libp2p/stream/connection.nim
+++ b/libp2p/stream/connection.nim
@@ -41,7 +41,7 @@ proc isUpgraded*(s: Connection): bool =
   if not isNil(s.upgraded):
     return s.upgraded.finished
 
-proc upgrade*(s: Connection, failed: ref Exception = nil) =
+proc upgrade*(s: Connection, failed: ref CatchableError = nil) =
   if not isNil(s.upgraded):
     if not isNil(failed):
       s.upgraded.fail(failed)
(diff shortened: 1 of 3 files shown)


dryajov:
Should probably go against `unstable`, unless `eh-tracking` is going to be merged pretty soon? Otherwise, LGTM.

arnetheduck:
this is a pure bugfix on the libp2p side - it is wrong to leak these exceptions regardless of chronos version - the idea would be to bump libp2p in nbc _before_ the eh branch so as to make rollbacks easier

dryajov:
> this is a pure bugfix on the libp2p side - it is wrong to leak these exceptions regardless of chronos version - the idea would be to bump libp2p in nbc _before_ the eh branch so as to make rollbacks easier

Yeah, not opposed. Lets merge it.

arnetheduck:
https://github.com/status-im/nim-libp2p/pull/547 is the proper fix really - it would be interesting to see how far strict vs non-strict compatibility can be taken, ie if code can be compatible with both - I'm not sure actually, but it's plausible. 

My recommendation would be to get this little fix into `master`, then, uh, finish https://github.com/status-im/nim-libp2p/pull/510 on `master` as well, then finish this branch on `master`, then do whatever with unstable - ie this exception stuff is pure bugfixing really - it makes the code more strict which makes it more compatible - throwing `unstable` into the mix just makes things confusing

PR #551, opened 2021-03-23
use a builder pattern to build the switch

sinkingsugar:


libp2p/builders.nim
```diff
@@ -0,0 +1,206 @@
...
+  var b = builder
+  b.baddress = address
+  b
+
+proc secureManager*(builder: SwitchBuilder, secureManager: SecureProtocol): SwitchBuilder =
+  var b = builder
+  b.bsecureManagers &= secureManager
+  b
+
+proc transportFlag*(builder: SwitchBuilder, flag: ServerFlags): SwitchBuilder =
```
dryajov, line 68: Hmm, I think we need to differentiate which transport is being built here. I would probably use something like `withTcpTransport(flags = ServerFlags)`. Same for muxers and secure managers. The `with` is not necessary, but I've seen this convention used pretty widely, but I'm either way on it.

sinkingsugar: Good point,
About the `with`, rust builders generally don't have any extra words.
Some random example I stumbled upon in the last months:
https://docs.rs/rapier3d/0.6.1/rapier3d/dynamics/struct.RigidBodyBuilder.html
https://docs.rs/web3/0.15.0/web3/contract/deploy/struct.Builder.html
That was my style rationale.

dryajov: > About the with, rust builders generally don't have any extra words.

I do confess that I like `with` better than just the bare word and nim tends to use camel case vs snake case in rust. But, like I said it's a small detail and I'm either way on it.

libp2p/builders.nim
```diff
@@ -0,0 +1,210 @@
...
+
+export
+  switch, peerid, peerinfo, connection, multiaddress, crypto
+
+type
+  SecureProtocol* {.pure.} = enum
+    Noise,
+    Secio {.deprecated.}
+
+  SwitchBuilder* {.byref.} = object
```
dryajov, line 17: I think that using a heap ref is preferable here, it's a short lived object and we can pretty much guarantee it's lifetime by setting it to nil after use.

sinkingsugar: why would you do such thing? 
it's a small descriptor, stack is its natural place, the extra copies are just nim fault and they definitely won't change anything in this instance.
`byref` is just a bit less copying around, but please don't make me use a `ref` here.. it's just wrong.

sinkingsugar: it's true that nim makes it clumsy but this should not make us write blasphemous code imho.

sinkingsugar: I wish `var` return worked but it would not compile when used functionally btw, think I added a comment about it.

dryajov: Heh, I was mostly being pragmatic about avoiding potentially problematic features. If byref works I'm fine with it.

sinkingsugar:
should this target unstable btw? Not a breaking change as the old proc is left there but the internals changed.

sinkingsugar:
this is confusing , I see no errors on actions...

sinkingsugar:
Ah the close-reopen trick seems to not work

dryajov:
Triggering rebuild from the actions usually works best for me.

dryajov:
> should this target unstable btw? Not a breaking change as the old proc is left there but the internals changed.

I think it's fine to leave it in stable. We're not changing the api and the change itself isn't too big.

libp2p/builders.nim
sinkingsugar: Switched to the way you wanted it btw

libp2p/protocols/identify.nim
```diff
@@ -101,16 +106,17 @@ proc decodeMsg*(buf: seq[byte]): Option[IdentifyInfo] =
...
+proc newIdentify*(peerInfo: PeerInfo, provider: CurrentAddressProvider = nil): Identify =
   new result
   result.peerInfo = peerInfo
+  result.provider = provider
   result.init()
 
 method init*(p: Identify) =
   proc handle(conn: Connection, proto: string) {.async, gcsafe, closure.} =
     try:
       trace "handling identify request", conn
```
dryajov, line 118: I think there is a misunderstanding as to what `observedAddr` does - it stores the address that the remote node has seen us connect from. It doesn't make sense to sed it here, because we essentially sending back what the remote sent us.

libp2p/protocols/identify.nim
```diff
@@ -42,8 +42,13 @@ type
     agentVersion*: Option[string]
     protos*: seq[string]
 
+  CurrentAddressProvider* =
```
dryajov, line 45: This is probably no the correct PR to work on this change, this is a bit more involved that and goes beyond just identify. Let's move it out to it's own branch.

dryajov (changes requested):
LGTM, but lets move the address provider work to a different branch.

libp2p/protocols/identify.nim
sinkingsugar: Yeah I realized that after, should be addrs. I rationalized in the issue about it.

libp2p/protocols/identify.nim
sinkingsugar: Sure, but for now it involves only identity I think. I will move it out but likely will need this branch

dryajov (approved):
LGTM again. Let's get this merged asap.

PR #552, opened 2021-03-24
Merging latest master and possible bug?

dryajov:
Merging latest master. 

There is a curious bug/issue, `doAssert` seems to raise a `ValueError` instead of an `AssertionError`?


sinkingsugar:
@dryajov I reverted the `ValueError` thing, it was a bug in chronos that was fixed.

dryajov:
Got ya, thanks!

PR #553, opened 2021-03-29
More gossip coverage

sinkingsugar:


sinkingsugar:
This raised coverage quite a lot , around 85% from 70% ish, only some rare fatal paths are not covered basically. I think it's quite enough for v1.

sinkingsugar:
Also fixed a small bug with ihave

sinkingsugar:
@dryajov can we merge this?

PR #557, opened 2021-04-05
Docs cleanup

sinkingsugar:


sinkingsugar:
Hmmm this also makes no sense 
https://github.com/status-im/nim-libp2p/blob/master/docs/GETTING_STARTED.md
A bit confusing @dryajov 
Maybe we should just get rid of this too?

sinkingsugar:
Also how to update things like:
https://our.status.im/nim-libp2p-tutorial-a-peer-to-peer-chat-example-1/ ?

sinkingsugar:
Another thing I always found confusing was the directchat sample, uses threads and async pipes.. it's very tedious.. not complex but quite hard core basically...

sinkingsugar:
For some reasons we don't support dialing anymore the way it was in the directchat sample:
```
/ip4/127.0.0.1/tcp/55505/ipfs/QmRop5a5niZ9rpkHYEsrctyZckgfL3JKXzPQLsaT8fkFao
dialing peer: /ip4/127.0.0.1/tcp/55505/p2p/QmRop5a5niZ9rpkHYEsrctyZckgfL3JKXzPQLsaT8fkFao
DBG 2021-04-05 14:40:46.183+09:00 Dialing peer                               topics="libp2p switch" tid=108696 peerId=QmR*8fkFao
DBG 2021-04-05 14:40:46.188+09:00 Dialing failed                             topics="libp2p switch" tid=108696 msg="Incorrect or unsupported address!" peerId=QmR*8fkFao
DBG 2021-04-05 14:40:46.195+09:00 Error dialing                              topics="libp2p switch" tid=108696 conn=Connection(nil) msg="Unable to establish outgoing link"
unable to dial remote peer /ip4/127.0.0.1/tcp/55505/ipfs/QmRop5a5niZ9rpkHYEsrctyZckgfL3JKXzPQLsaT8fkFao
Unable to establish outgoing link
```

sinkingsugar:
> For some reasons we don't support dialing anymore the way it was in the directchat sample:
> 
> ```
> /ip4/127.0.0.1/tcp/55505/ipfs/QmRop5a5niZ9rpkHYEsrctyZckgfL3JKXzPQLsaT8fkFao
> dialing peer: /ip4/127.0.0.1/tcp/55505/p2p/QmRop5a5niZ9rpkHYEsrctyZckgfL3JKXzPQLsaT8fkFao
> DBG 2021-04-05 14:40:46.183+09:00 Dialing peer                               topics="libp2p switch" tid=108696 peerId=QmR*8fkFao
> DBG 2021-04-05 14:40:46.188+09:00 Dialing failed                             topics="libp2p switch" tid=108696 msg="Incorrect or unsupported address!" peerId=QmR*8fkFao
> DBG 2021-04-05 14:40:46.195+09:00 Error dialing                              topics="libp2p switch" tid=108696 conn=Connection(nil) msg="Unable to establish outgoing link"
> unable to dial remote peer /ip4/127.0.0.1/tcp/55505/ipfs/QmRop5a5niZ9rpkHYEsrctyZckgfL3JKXzPQLsaT8fkFao
> Unable to establish outgoing link
> ```

@dryajov know anything about this btw?

sinkingsugar:
So, the directchat sample wants a `IPFS` kind of address including `peerId`, but in `wire.nim` we don't seem to support that anymore.

libp2p/wire.nim
```diff
@@ -33,7 +33,7 @@ proc initTAddress*(ma: MultiAddress): MaResult[TransportAddress] {.
   ## Initialize ``TransportAddress`` with MultiAddress ``ma``.
   ##
   ## MultiAddress must be wire address, e.g. ``{IP4, IP6, UNIX}/{TCP, UDP}``.
-  if TRANSPMA.match(ma):
+  if TRANSPMA.matchPartial(ma):
```
cheatfate, line 36: What the purpose of doing this change?

libp2p/wire.nim
```diff
@@ -71,7 +71,7 @@ proc connect*(ma: MultiAddress, bufferSize = DefaultStreamBufferSize,
   ## Open new connection to remote peer with address ``ma`` and create
   ## new transport object ``StreamTransport`` for established connection.
   ## ``bufferSize`` is size of internal buffer for transport.
-  if not(RTRANSPMA.match(ma)):
+  if not(RTRANSPMA.matchPartial(ma)):
```
cheatfate, line 74: What the purpose of doing this change?

libp2p/wire.nim
sinkingsugar: To fix directchat
and to allow such addresses `/ip4/127.0.0.1/tcp/55505/ipfs/QmRop5a5niZ9rpkHYEsrctyZckgfL3JKXzPQLsaT8fkFao` else would not work

libp2p/wire.nim
sinkingsugar: See above answer.
Any implications about this choice btw? It definitely fixes the issue but yeah I'm happy to discuss it. 
It used to be like this before client side.

libp2p/wire.nim
dryajov: Hmm, yes it's indeed weird that this broke, there have been no changes to this functionality AFAIK. Perhaps there is something else going on? On the other hand, I also wonder what are the implications of this change - @cheatfate you worked on this primitives and clue whats going on?

cheatfate: This is very weird, because `directchat.nim` do not use this procedures directly. From my point of view `wire.nim` should accept only addresses without any IPFS information, because `wire.nim` is about TCP,UDP,UNIX over IPv4/IPv6 networking but nothing else. Its the glue between `libp2p` and `chronos` transports.

So for example when you change `match` to `matchPartial`, then
addresses like `/ip4/127.0.0.1/tcp/80/http` become accepted, but this addresses should not be accepted by TCP transport it should be accepted by more high level type of transport which recognize `http` part of path too... So its why i have made here strict comparison. So if you need to use this procedure you should split your MultiAddress into parts and this will confirm that you understand what you are doing.

So you should fix `directchat.nim` in some other place, or even fix `directchat.nim` itself to split addresses, because it should not rely on `wire.nim` procedures (it even not imports this file).

dryajov: Ah, right - yes I completely agree, this is something that needs to happen before it makes it to wire.

sinkingsugar: This example was written with that wrong assumption in mind tho.
It's easy to understand that.
Should we rewrite it?

dryajov: Yeah, weird that it's like that, it uses to work but maybe it was wrong all along. Let's fix it to properly reflect how it should work.

README.md
```diff
@@ -18,7 +18,9 @@
 
 ## Introduction
 
-An implementation of [libp2p](https://libp2p.io/) in Nim. Also provides a Nim wrapper of the [Libp2p Go daemon](https://github.com/libp2p/go-libp2p).
+An implementation of [libp2p](https://libp2p.io/) in Nim.
+
+Also optionally provides a Nim wrapper of the [Libp2p Go daemon](https://github.com/libp2p/go-libp2p).
 
 ## Project Status
 The current native Nim libp2p implementation support is experimental and shouldn't be relied on for production use. It is under active development and contributions are highly welcomed. :)
```
dryajov, line 26: Maybe this can be changed since we're releasing v1.

dryajov: By changing, I mean reworded :)

sinkingsugar: Ah yeah :) indeed , let's not forget that

README.md
```diff
@@ -18,10 +18,12 @@
...
 ## Introduction
 
-An implementation of [libp2p](https://libp2p.io/) in Nim. Also provides a Nim wrapper of the [Libp2p Go daemon](https://github.com/libp2p/go-libp2p).
+An implementation of [libp2p](https://libp2p.io/) in Nim.
 
 ## Project Status
-The current native Nim libp2p implementation support is experimental and shouldn't be relied on for production use. It is under active development and contributions are highly welcomed. :)
+libp2p is now used in production by a few projects at [status](https://github.com/status-im), including [nimbus](https://github.com/status-im/nimbus-eth2).
+
+While far from complete, currently available componets are fairly stable.
```
arnetheduck, line 26: This doesn't inspire much confidence, I think we can be more assertive than that

PR #558, opened 2021-04-05
Builders

dryajov:


libp2p/builders.nim
```diff
@@ -0,0 +1,198 @@
...
+    protoVersion: string
+    agentVersion: string
+
+proc init*(T: type[SwitchBuilder]): T =
+  SwitchBuilder(
+    privKey: none(PrivateKey),
+    address: MultiAddress.init("/ip4/127.0.0.1/tcp/0").tryGet(),
+    secureManagers: @[],
+    tcpTransportOpts: TcpTransportOpts(),
+    rng: newRng(),
```
arnetheduck, line 47: would be better to not create resources like this eagerly, they might not be available on all systems and are a cost at creation time

libp2p/builders.nim
```diff
@@ -0,0 +1,198 @@
...
+    inTimeout: Duration
+    outTimeout: Duration
+    maxConnections: int
+    maxIn: int
+    maxOut: int
+    maxConnsPerPeer: int
+    protoVersion: string
+    agentVersion: string
+
+proc init*(T: type[SwitchBuilder]): T =
```
arnetheduck, line 41: `new` because it's a `ref object`

libp2p/standard_setup.nim
```diff
@@ -1,71 +0,0 @@
-import
```
arnetheduck, line 1: could leave an import/export in place which would allow existing code not to break

libp2p/builders.nim
dryajov: You mean `newRng()` - good point in any case.

libp2p/builders.nim
dryajov: I'll change it. But as a note, there is an issue with paramless `new`, I believe I ran into it and nim wasn't able to tell it apart from the system `new`. If I can repro it, I'll log it as an issue in the style guide.

libp2p/standard_setup.nim
dryajov: Yeah, good point and I ran into it in nimbus, will add back to avoid breaking other projects.

PR #563, opened 2021-04-13
QUIC transport

markspanbroek:
Work in progress. Adds support for QUIC as a base transport for libp2p connections. Based on the [nim-quic](https://github.com/status-im/nim-quic) library.

PR #565, opened 2021-04-20
Rework `dial` and `accept` to return `Session`

markspanbroek:
This changes `Transport.dial()` and `Transport.accept()` to return a `Session` instead of a `Connection`. A `Connection` can be obtained by calling `getStream()` on the `Session` object.

This is in preparation for integration of QUIC (#563). In QUIC opening a connection and opening a stream are separate operations.


tests/testswitch.nim
```diff
@@ -714,12 +714,12 @@ suite "Switch":
 
     await allFuturesThrowing(readers)
     await switch2.stop() #Otherwise this leeks
+    await switch1.stop()
```
dryajov, line 717: The problem of moving the stop here is that the tracker checks don't make any sense after it. What this is testing for is that after stopping `switch1`, all resources are released on both, ends - meaning that channels, secure connections and chronos streams are both tore down.

If something is tripping the trackers, it means that the teardown flow broke.

Menduist: This is likely because MacOS takes some time to close the stream on the receiving end, and thus the stream isn't closed quickly enough, since adding a sleepAsync in the middle fixes the issue (https://github.com/status-im/nim-libp2p/commit/438a757c8d793b66a275f769e9e686a2de540669 , don't mind the allFuturesThrowing moving around)

Not sure how to fix the test properly, a sleepAsync is not great but does the job

dryajov: Ah, yeah this might very well be the case, does it happen on linux & windows as well?

Menduist: No, at least not on my linux laptop

markspanbroek: Great feedback @dryajov and @Menduist. I'll remove this last commit from the PR; continuing the conversation in #592.

Menduist:
I've updated this on the latest unstable, though we should maybe hold off on this for a little while, until we have a better overview of the conn manager
(btw, we apparently need signed commits for unstable, which is good, but could you just setup the signing of commits on your side Mark? thx :) )

markspanbroek:
Latest commit adds direction parameter to Session.getStream(), so that I can support both outgoing and incoming streams in Quic.

PR #572, opened 2021-05-15
adding raises defect across the codebase

dryajov:


libp2p/builders.nim
```diff
@@ -109,22 +125,28 @@ proc withAgentVersion*(b: SwitchBuilder, agentVersion: string): SwitchBuilder =
...
 
   let
-    peerInfo = block:
+    peerInfo = try:
       var info = PeerInfo.init(seckey, [b.address])
       info.protoVersion = b.protoVersion
       info.agentVersion = b.agentVersion
       info
+    except CatchableError as exc:
+      raise newException(Defect, exc.msg)
```
arnetheduck, line 149: `raiseAssert exc.msg`

libp2p/builders.nim
```diff
@@ -109,22 +125,28 @@ proc withAgentVersion*(b: SwitchBuilder, agentVersion: string): SwitchBuilder =
...
-    raise (ref CatchableError)(msg: "Cannot initialize RNG")
+    raise newException(Defect, "Cannot initialize RNG")
+
+  let pkRes = PrivateKey.random(b.rng[])
+  if pkRes.isErr:
+    raise newException(Defect, "No RNG supplied!")
 
   let
-    seckey = b.privKey.get(otherwise = PrivateKey.random(b.rng[]).tryGet())
+    seckey = b.privKey.get(otherwise = pkRes.get())
```
arnetheduck, line 135: generally, it's better to use `.expect("reason why it's safe")` - `get` panics with a poor error message, but above all, when reading the logic, it helps make clear why it's safe to assume that there's a value stored in here

libp2p/connmanager.nim
```diff
@@ -105,15 +109,29 @@ proc addConnEventHandler*(c: ConnManager,
...
   ##
 
-  if isNil(handler): return
-  c.connEvents.mgetOrPut(kind,
-    initOrderedSet[ConnEventHandler]()).incl(handler)
+  try:
+    if isNil(handler): return
+    c.connEvents.mgetOrPut(kind,
+      initOrderedSet[ConnEventHandler]()).incl(handler)
+  except Exception as exc:
```
arnetheduck, line 116: with https://github.com/status-im/nim-libp2p/pull/573, these should no longer be needed :) That said, once `ConnEventHandler` is marked as `raises: [Defect]` the bug goes away :/

libp2p/protocols/identify.nim
```diff
@@ -47,20 +50,25 @@ type
...
-  result.write(5, protoVersion)
-  let agentVersion = if peerInfo.agentVersion.len <= 0:
-    AgentVersion
-  else:
-    peerInfo.agentVersion
-  result.write(6, agentVersion)
-  result.finish()
+  if peerInfo.publicKey.isSome:
+    let bytesRes = peerInfo.publicKey.get().getBytes()
+    if bytesRes.isErr:
```
arnetheduck, line 55: shouldn't this be treated the same as if the public key doesn't exist?

libp2p/protocols/protocol.nim
```diff
@@ -7,13 +7,17 @@
...
 
 type
-  LPProtoHandler* = proc (conn: Connection,
-                          proto: string):
-                          Future[void] {.gcsafe, closure.}
+  LPProtoHandler* = proc (
+    conn: Connection,
+    proto: string):
+    Future[void]
+    {.gcsafe, closure, raises: [Defect].}
```
arnetheduck, line 20: explicit `closure` is not really needed here

libp2p/connmanager.nim
dryajov: weird, I'm still getting the `Exception` with Nim 1.2.6.

libp2p/builders.nim
```diff
@@ -149,16 +166,19 @@ proc build*(b: SwitchBuilder): Switch =
...
+      transports = transports,
+      identity = identify,
+      muxers = muxers,
+      secureManagers = secureManagerInstances,
+      maxConnections = b.maxConnections,
+      maxIn = b.maxIn,
+      maxOut = b.maxOut,
+      maxConnsPerPeer = b.maxConnsPerPeer)
+  except CatchableError as exc:
+    raise newException(Defect, exc.msg)
```
arnetheduck, line 181: should this really be a defect? isn't this when the port is busy for example?

libp2p/wire.nim
```diff
@@ -85,14 +89,25 @@ proc createStreamServer*[T](ma: MultiAddress,
...
+    raise newException(Defect, exc.msg)
+
+  if address.isErr:
+    raise newException(MaInvalidAddress, address.error)
+
+  try:
+    return createStreamServer(address.get(), cbproc, flags, udata, sock,
+                                backlog, bufferSize, child, init)
+  except CatchableError as exc:
+    raise newException(Defect, exc.msg)
```
arnetheduck, line 110: likewise here? this looks harsh

libp2p/builders.nim
dryajov: Good point, in fact I'm propagating `LPError` already, so it should not `Defect` here.

PR #577, opened 2021-05-24
Unstable

dryajov:


dryajov:
Before merging we need to understand why memory usage differs from `master`. Here are some screenshots:

`u-libp2p-s-large-01.aws-eu-central-1a.nimbus.prater`:

Screen Shot 2021-05-25 at 1 19 42 PM

`u-libp2p-u-large-01.aws-eu-central-1a.nimbus.prater`:

Screen Shot 2021-05-25 at 1 20 01 PM

dryajov:
We continue having the same memory usage pattern differences with stable and we need to figure out what it is before merging. This needs to be done sooner rather than later because otherwise it is holding up a significant amount of work already done in `unstable`. 

@arnetheduck might have some good insights into how he troubleshooted similar issues in `stable/master`. 

Some things I've done in the past that have had varying levels of success:

- Use Nim's own tooling 
  - https://nim-lang.org/docs/estp.html
  - https://nim-lang.org/blog/2017/10/02/documenting-profiling-and-debugging-nim-code.html (high level overview of nim tooling)
- Using Chronos future tracking facilities with the `-:chronosFutureTracking` and then querying the `debug_getChronosFutures` endpoint -  this will give an idea of which futures are taking up the most time or aren't released at all
  - For example, compile nimbus with `make -j14 LOG_LEVEL="DEBUG" NIM_PARAMS="-d:chronosFutureTracking -d:insecure" beacon_node`
  - And do something like `while true; do curl -d '{"jsonrpc":"2.0","id":"id","method":"debug_getChronosFutures","params":[] }' -H 'Content-Type: application/json' localhost:9091 -s | jq '.result | (.[0] | keys_unsorted) as $keys | $keys, map([.[ $keys[] ]])[] | @csv' > "$(date +"%Y%m%d%H%M%S").log" ; sleep 120; done` to query the endpoint in Nimbus every 120 seconds
- Using valgrind and friends has varying levels of success when used with Nim's GC, but that might have changed since @arnetheduck did some work to enable it (unsure about this)?

dryajov:
> Ah, thanks, I see. Your reasoning makes sense. We'll create a separate `PeerStore` for our own books. I've also only now noticed the integration of `PeerStore` into the `switch` - great! 👍. Does mean we'll have to rethink some patterns on our side now... 😄

I believe we are better of temporarily reverting the peerstore changes as they have also broke our unstable autobump here - https://github.com/status-im/nimbus-eth2/pull/2546. The changes are totally fine, but as it is it's blocking us from troubleshooting memory issues on the test env. Another approach might be to make them backwards compatible, but I don't think it's an option for this specific changes?

dryajov:
We've decided to move forward with the merge of this branch. We're seeing the same patterns now on stable (without this patch) and this patterns aren't atypical for nimbus/libp2p.

arnetheduck:
> Before merging we need to understand why memory usage differs from master. Here are some screenshots:

if I were to guess, this node was syncing - generally though, the thing to check first _before_ doing anything else is to simply look at the logs and see what the node is up to - they are available both on the machine and in kibana.

dryajov:
> > Before merging we need to understand why memory usage differs from master. Here are some screenshots:
> 
> if I were to guess, this node was syncing - generally though, the thing to check first _before_ doing anything else is to simply look at the logs and see what the node is up to - they are available both on the machine and in kibana.

The flat memory pattern showed on every single machine besides the one running this branch.

arnetheduck:
> The flat memory pattern showed on every single machine besides the one running this branch.

well, again: what did the logs say? that should always be the first thing to check, before doing anything more fancy.

it could be a setup issue on the server, a latent bug that was exposed or anything, really.

dryajov:
> > The flat memory pattern showed on every single machine besides the one running this branch.
> 
> well, again: what did the logs say? that should always be the first thing to check, before doing anything more fancy.
> 
> it could be a setup issue on the server, a latent bug that was exposed or anything, really.

True, I agree that logs are the first thing to look at and we had looked at the logs prior to noting this as an issue and the node was synced and operating normally as far as I remember. This branch has been around for a while and we've been monitoring it on and off, this is just the latest of developments. I'm not sure if we've checked the logs recently tho - I haven't,  but we should do that.

Menduist:
I did check the logs, and everything seemed normal: attestations going through, peers going on and off, epoch ending, etc. But I don't know nimbus well enough to be sure that they were nominal

Menduist:
Calling `GC_fullCollect` more frequently solves the issue, meaning that the GC is not triggered frequently enough, causing the spikes. (eg, it's not data that is held up by something and thus un-freeable)

From what I understand of the GC, calling `GC_fullCollect` force the cycle detection, which is normally trigger using a dynamic threshold. The regular (eg, non cyclic) detection is always ran, without threshold.

If I'm right, the spikes we're seeing is some cyclic memory that is `GC`ed less frequently. Since 90% of this memory is allocated in
`PubSubPeer.handle`, `GossipSub.rpcHandler` and `Mplex.handle` (more precisely, in the future allocated by theses functions callees), it seems that for some reason, the Future objects are cyclic.

@cheatfate any idea as to why the Future objects could be cyclical referenced? They are only used in simple await style, like [here for instance](https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/secure/noise.nim#L305)

Or maybe the GC is completely wrong for some reason. (it would be easier to diagnostic this if we could access GC statistics in the metrics, but having to parse `GC_getStatistics` is bit dumb)

dryajov:
Great finds - @Menduist!

> Or maybe the GC is completely wrong for some reason. (it would be easier to diagnostic this if we could access GC statistics in the metrics, but having to parse GC_getStatistics is bit dumb)

I think we already do that and this is what you see in the `GC Heap Objects` in Grafana - @stefantalpalaru has some experience here as well I believe.

> Edit: yeah, for a cycle collection to trigger naturally, we need to use 2 times more RAM than the last post-cycle usage.. So we use about 1go of RAM in nimbus, we would have to go to 2go to trigger the cycle collection. If I understand correctly, this is really not good

But we don't wait that long since we trigger it per slot? 

We definitely want to reduce the time we hold to some of this references and get to the bottom of the circular `Futures`, but in addition to that, perhaps there is value in manually triggering `GC_fullCollect` in some strategic places in libp2p? 🤔

dryajov:
One such candidate place would be on peer disconnect.

stefantalpalaru:
We use `dumpHeapInstances()` to look at GC-allocated memory in nim-metrics: https://github.com/status-im/nim-metrics/blob/f812b268473ace7817ade20dd8e6d0574f4b5460/metrics.nim#L1079

arnetheduck:
In general, invoking the GC is an incredibly slow operation that should be avoided at all costs - that is _why_ the full GC only runs in Nim on a threshold - running it more frequently offers more of a workaround at a high cost. We run the GC deliberately at the given intervals because that's when they have the lowest impact on useful functionality.

As long as the memory in general is being reused, it's not a _significant_ issue that RSS is high - it shows the "peak" where memory was needed at some point - in terms of memory consumption, it's a lot more important that all points where memory is allocated are of a controlled size, such that these peaks cannot grow beyond a capped size.

It's certainly possible that some low-hanging fruit exist on the memory front, but unless there are regressions or significant issues found, memory usage is within acceptable bounds for now.

> One such candidate place would be on peer disconnect.

Doing a full gc (or anything else that's costly for any resource) on any remote controlled event is a significant security issue - it would be good to write down principles like this in a doc, along with a review of the classes of errors that have been found in security and bug audits over time.

It's also a bad idea in general that a _library_ would run global operations like garbage collection - this is something that an application policy should decide. Work in libraries should rather focus on making sure that they don't make unnecessary allocations or keep references around to memory they no longer need, above all if they have hotspots that can easily be addressed.

dryajov:
> It's also a bad idea in general that a library would run global operations like garbage collection - this is something that an application policy should decide.

Yes, this should not be things that a library controls - but I wasn't suggesting this happens in the library itself, rather than it's triggered on peer disconnect, through a hook for example. In any case, this was more to incite the discussion rather than a definitive solution I'm attempting to push.

> It's certainly possible that some low-hanging fruit exist on the memory front, but unless there are regressions or significant issues found, memory usage is within acceptable bounds for now.

I agree, there is nothing to suggest that something has fundamentally changed and memory usage does seem within acceptable bounds. What set the alarms off were the differences that `stable` was exhibiting as compared to `unstable`.  However `stable` has since reverted back to the more common "spiky" pattern without any new changes from libp2p's side - which suggests that it was probably circumstantial, like network and/or hardware/environment. 

I would like to get to the bottom of this and it does bother me that we can't say what set this of exactly, but at this point the question should be is this safe to merge or are we seeing something exceptionally bad? I would be inclined to say that we aren't and this should be safe to merge to unblock a few more things along the way, such as new transports and such.

We should also start a more general conversation around memory usage in libp2p, there are lots of things that can be improved on and most solutions are on the naive side, we knew that we'd have to pay this debt off at some point, this is probably a good time as any to start a more comprehensive overhaul.

Menduist:
I've confirmed that the spikes are coming from Chronos future's cyclical references, and opened an issue in chronos: https://github.com/status-im/nim-chronos/issues/203

Solving this issue should stop the weird memory pattern we're seeing here.

Menduist:
New week, new ram issues :tada: 
image

`u-libp2p-u-large-01` started to use 263 mib of ram for `seq[uint32]` since the last restart, not seen on other nodes

Current logs: seemed normal
Logs at startup: also seems normal
Metrics: compared all metrics with `u-libp2p-s-large-01`, they were all in the same ballpark
Metrics at startup: the issue appeared as soon as the program launched (well, as far as grafana can see)
The code: checked where `seq[uint32]` are used: base58 decoding, discv5, `nim-unicodedb/gen/casing.nim` and `nim-eth/eth/keyfile/keyfile.nim`. A lot of them are used on peer-supplied data, so hard to follow

I restarted the node for more diagnostics, and the problem disappeared. Restarted it once again for good measure, still no issue

So I can't make much diagnostic, I'll wait to see if it pops up again

arnetheduck:
> u-libp2p-u-large-01 started to use 263 mib of ram for seq[uint32] since the last restart, not seen on other nodes

this is a semi-known issue that probably has been around for a while, `uint32` also somewhat annoyingly covers alias types like `ValidatorIndex` - tracking it down would be good, but probably is related to something in the startup (ping @zah) than anything else

dryajov:
I'm thinking of closing this PR and creating a new one to track the potential new issues with unstable? @Menduist

Menduist:
Yes, probably best :)

PR #580, opened 2021-06-01
DNS Addresses handling

### LINKED ISSUE #571: Support for DNS addresses
## Problem statement

The [multiaddr specification](https://github.com/multiformats/multiaddr) specifies support for `dns4`, `dns6` and `dnsaddr` addresses. It is possible to [create and parse such addresses](https://github.com/status-im/nim-libp2p/blob/master/libp2p/multiaddress.nim#L407) in the Nim implementation, but there is currently no way to resolve these to dialable transport addresses. Clients trying to connect to a DNS address will encounter an [Invalid Address exception](https://github.com/status-im/nim-libp2p/blob/master/libp2p/wire.nim#L91).

DNS address resolution will be a useful feature to add for clients not using static IPs.

## Acceptance criteria

- [ ] Add a resolver for DNS multiaddrs (`dns4`, `dns6` and `dnsaddr`)

## References

- Reference resolver implementations in [js](https://github.com/multiformats/js-multiaddr/blob/master/src/resolvers/index.js) and [go](https://github.com/multiformats/go-multiaddr-dns).
- `dnsaddr` specification [here](https://github.com/multiformats/multiaddr/blob/master/protocols/DNSADDR.md)

cc @dryajov in case this has already been discussed in a different context

Menduist:
Closes #571 

This PR adds support for `dns`, `dns4`, `dns6` and `dnsaddr` MultiAddress protocols.

**/!\ Warning: resolving DNS is currently a blocking operation!**
If a peer can send a `dns` address to libp2p, it would be really easy to ddos a libp2p node. A dns resolve can easily take a few hundred milliseconds, so resolving ~10 addresses/seconds would completely block the node.

Possible workarounds:
- Block `dns` addresses coming from peers
- Use `getaddrinfo_a` which does not look great
- Implement a DNS client in chronos

Edit: ^ has been solved, see comments

New todo:
- [x] Name resolving framework
- [x] DNS async resolving
- [x] Allow user to choose the name resolver
- [x]  `dnsaddr` proper implementation
- [x] Test everything
- [x] Better error handling


libp2p/wire.nim
```diff
@@ -20,21 +20,45 @@ else:
...
 
   RTRANSPMA* = mapOr(
-    mapAnd(IP, mapEq("tcp")),
-    mapAnd(mapEq("unix"))
+    TCP,
+    UNIX
   )
 
+proc resolveDns(ma: MultiAddress, domain: Domain = Domain.AF_UNSPEC, prefix: string = ""): MaResult[TransportAddress]
+  {.raises: [Defect, ResultError[string]]} =
```
dryajov, line 34: Probably better to map to a native exception with `toException`, similar to how it's done in `peerinfo.nim`

dryajov: Hmm, actually probably better to stick with just `Results` and not mix `tryGet`.

libp2p/wire.nim
```diff
@@ -20,21 +20,45 @@ else:
...
+  try:
+    var
+      dnsbuf: array[256, byte]
+      dnsval: string
+      pbuf: array[2, byte]
+    if ma[0].tryGet().protoArgument(dnsbuf).tryGet() == 0:
```
dryajov, line 40: Probably better to use `ma[0].isOk...`

dryajov:
> Warning: resolving DNS is currently a blocking operation!

Yes, this is a major consern and we should figure out the best way to approach it, ideally this should be non-blocking in chronos, but I'm curious to hear @cheatfate's opinion on this and why it isn't non-blocking already - perhaps there are some non obvious limiation that prevents this.

libp2p/wire.nim
arnetheduck: yeah, things should either return `Result` or raise, but never both

libp2p/wire.nim
arnetheduck: since this function returns `Result`, `(? ma[0]).protoArgument...`

libp2p/wire.nim
```diff
@@ -20,21 +20,45 @@ else:
...
+      err("Invalid DNS argument")
+    else:
+      dnsval.setLen(dnsbuf.len)
+      copyMem(addr dnsval[0], addr dnsbuf[0], dnsbuf.len)
```
Menduist, line 44: Just found the `string.fromBytes` function in stew

libp2p/wire.nim
Menduist: Yeah, this is coming from the `initTAddress` function. I'll wait for #582 to be merged, and rebase on it

dryajov:
So I'm looking at [asynctools](https://github.com/cheatfate/asynctools/blob/master/asynctools/asyncdns.nim) and it seems like it's probably our best bet, we can either extract the dns resolution and adapt it to chronos or modify  [asynctools](https://github.com/cheatfate/asynctools/blob/master/asynctools/asyncdns.nim) directly to support chronos (I like the second option better, but might be a bit too much for now). I think it's also missing TXT records, but it should be fairly easy to add?

libp2p/wire.nim
```diff
@@ -20,21 +20,45 @@ else:
...
+      if ma[1].tryGet().protoArgument(pbuf).tryGet() == 0:
+        err("Incorrect port number")
+      else:
+        let port = Port(fromBytesBE(uint16, pbuf))
+        let resolvedAddresses = resolveTAddress(prefix & dnsval, port, domain)
+        if resolvedAddresses.len > 0:
+          ok(resolvedAddresses[0])
```
dryajov, line 51: Think this should return the full set of resolved addresses.

libp2p/wire.nim
```diff
@@ -20,20 +20,42 @@ else:
...
+  try:
+    var
+      dnsbuf: array[256, byte]
+      pbuf: array[2, byte]
+
+    if ma[0].`?`.protoArgument(dnsbuf).`?` == 0:
+      err("Invalid DNS argument")
+    else:
+      let dnsval = string.fromBytes(dnsbuf)
+      if ma[1].`?`.protoArgument(pbuf).`?` == 0:
```
Menduist, line 43: I like this a bit better than the `?` sandwich `(?(?ma[1]).protoArgument(pbuf))`
But tell me if you want me to use the sandwich

libp2p/wire.nim
Menduist: Yup, the issue with using multiple resolved addresses is that the resolving must happen before-hand (eg, in the dialer -> internalConnect), so that the dialer can try out the different addresses.
I'll do that then

libp2p/wire.nim
arnetheduck: no particular opinion since none of them are pretty, except that the `()` is used elsewhere, which is a good discriminator in these cases. The alternative would be to try and figure out a better chaining syntax in general, but that's a bit beyond the scope of this PR.

Menduist:
Ok, so I handled multiple DNS entries, and it's pretty terrible. I'm open to suggestion, bc not very proud of what I've done.

Basically, I resolved the DNSs in `wire` -> `connect`, which means that it's now an async proc, and that it has to try multiple connect attempts.
(Couldn't just put in `dialer`->`tryConnect`, which already tries multiple connects, since `wire`->`connect` can also be used directly with DNS addresses)

I also resolved in each `createStreamServer`, but since they're not async and I don't want to break the api, it's a blocking resolve (though should be fine for listening, I guess?)

And I also resolve in the dialer, it's just a minor optimization, it avoids resolving multiple time for multiple transports.

Needs some cleaning up & more error handling, I just wanted to check that you don't see a better option

Menduist:
Comparison of the various DNS async resolving methods we could use:
image

`asyncdns` => cheatfate's library (based on platform-specific libraries)
`Blocking function in a separate thread` => using `getaddrinfo` or equivalent in a separate thread
`Homemade DNS client` => Using a [dns client library](https://github.com/ba0f3/dnsclient.nim) to create requests manually
`DNS over HTTPS` => Also uses a DNS client library, but use a HTTPS instead of UDP

Based on this, we can see that there is no magic bullet. Then, some decisions should be left to the user (the possibility to use DoH, use a custom DNS server, etc).

So, first of all, I added a simple interface, allowing to plug any name resolver (could allow `.eth` resolving in the future, for instance). (WIP).

Now, on the available name revolvers:
For me, the priorities are, in order:
- Compiles and works everywhere
- Works as expected by the user (eg, uses the hosts & nameservers set in the OS)
- As simple as possible

So, `asyncdns` is out of the window (doesn't work for ipv6 nameservers, _heavily_ relies on distrib specific APIs, would probably not compile everywhere)

DoH is great, because it can't be blocked. But, it may not works as the user expects.

That leaves us with `separate thread` vs `homemade dns client`.

The `homemade dns client` is fully platform agnostic (since it only relies on UDP transport).
The OS's server can be retrieved by parsing `/etc/resolv.conf` on unix (which, btw, is what the [libc is doing anyway](https://github.com/bminor/glibc/blob/595c22ecd8e87a27fd19270ed30fdbae9ad25426/resolv/res_init.c#L396)), or using [GetNetworkParams](https://docs.microsoft.com/fr-fr/windows/win32/api/iphlpapi/nf-iphlpapi-getnetworkparams?redirectedfrom=MSDN) on Windows. And if this fails, we can fallback to a hardcoded DNS server.
If this also fails (meaning, the user is in a network blocking public DNS servers and is using a weird OS), the user will have to override the DNS server manually.


`Separate thread` should only require two versions: one for windows, and one for unix. But it may not be 100% compatible, especially with (very) old unixes.

I think that the `homemade dns client` should work in more cases, and is simpler to maintain and develop, so I'll go with that.

So, to wrap up: For this pr, a homemade dns client. In the future, DoH and others resolvers could be easily added.

libp2p/wire.nim
```diff
@@ -93,24 +163,30 @@ proc createStreamServer*[T](ma: MultiAddress,
...
-      flags,
-      udata,
-      sock,
-      backlog,
-      bufferSize,
-      child,
-      init)
-  except CatchableError as exc:
-    raise newException(LPError, exc.msg)
+  let addressesResolved = waitFor resolveMAddresses(@[ma])
```
dryajov, line 171: We should never call `waitFor` from a `proc` that can be invoked from another `async` proc - the behaviour is undefined for the most part.

Menduist: Oh. Then I must break the `listen` apis to make them async

libp2p.nimble
```diff
@@ -9,6 +9,7 @@ skipDirs      = @["tests", "examples", "Nim", "tools", "scripts", "docs"]
 
 requires "nim >= 1.2.0",
          "nimcrypto >= 0.4.1",
+         "https://github.com/Menduist/dnsclient.nim == 0.1.0",
```
dryajov, line 12: Should be probably moved to a status repo. 

Another point is that this doesn't help us determine the OS DNS or does it?

Menduist: I just forked it because I needed a few modifications.
But they just been [merged](https://github.com/ba0f3/dnsclient.nim/pull/5) upstream, so we can actually use the upstream repo.
Or we can fork it in status, as you want

Menduist: For the OS DNS, I parse `/etc/resolv.conf` on unix and use a function in windows (todo).
More details in my last comment

dryajov:
> We should avoid using waitFor & co in general because they are supposed to be terminal calls - calling a waitFor from an async proc is undefined, might be just better to add a sync version of resolve.

To clarify, should be avoided as a nested call - using `waitFor` as an entry point is totally fine.

arnetheduck:
> I would prefer we do the resolving at the dial level rather than connect

Important here is that no remote-supplied multi-addrs' are resoved automatically.. to that end, it seems better perhaps that this is a manual step where you resolve one multiaddr into a list of others, then pass _that_ into connect?

arnetheduck:
in other no magic automatic resolution, but rather a "resolve" step first

libp2p/wire.nim
dryajov: How about having a sync `resolve` in addition to the async one?

dryajov:
> Important here is that no remote-supplied multi-addrs' are resoved automatically.. 

Mm, not sure what you mean exactly. Do you mean that `dial` should not take `dns` addrs or that resolve should happen outside of `wire/connect`?

arnetheduck:
> Do you mean that dial should not take dns addrs

yeah - by extension, it would not happen in `connect` either

Menduist:
Ok, since I thought `wire` procs were used by end application.
Since they're not, I'll happily break everything and make all of this a lot cleaner, by indeed only resolving in the dialer and the start of nodes :)

@arnetheduck since this in now async, we block the remote supplied dns addresses?

dryajov:
@Menduist 

> I think that the homemade dns client should work in more cases, and is simpler to maintain and develop, so I'll go with that.

I'm OK with this, as long as the DNS spec/implementations/etc aren't riddled with a bunch of  corner cases?

dryajov:
> > Do you mean that dial should not take dns addrs
> 
> yeah - by extension, it would not happen in `connect` either

Hmm, well I don't disagree, but multiaddrs support DNS resolution and libp2p works at the multiaddr level, so by extension it should support DNS resolution. Otherwise this entire issue doesn't make sense and users should just pull in/consume whatever external resolution method they want.

dryajov:
@Menduist 

> Since they're not, I'll happily break everything and make all of this a lot cleaner, by indeed only resolving in the dialer and the start of nodes :)

Yeah, breaking should be fine, tho it does make sense to have `wire` support resolution as well, for completion, so I'd say that the work already done is fine, we just want to avoid `waitFor` and things like that.

libp2p/wire.nim
dryajov: The problem with making things `async` is that it can introduce subtle concurrency issues that then lead to deadlocks and such, not that it would be the case here, but just something to keep in mind.

Menduist: > How about having a sync `resolve` in addition to the async one?

That's possible, the actual resolve async is ~65 LOC, so I would just need to duplicate that and remove all async on the duplicate.

That would not be very elegant though, and would have to be done for each resolver.

Or, just use the stock resolver for sync version, but that could bring weird discrepancies

arnetheduck:
> @arnetheduck since this in now async, why block the remote supplied dns addresses?

I guess it depends on the cost and the amount of dns requests someone else can trigger - the point would be to avoid remote triggered bombs of any kind - each concurrent request continues to have costs and these need to be capped somehow.

dryajov:
> > @arnetheduck since this in now async, why block the remote supplied dns addresses?
> 
> I guess it depends on the cost and the amount of dns requests someone else can trigger - the point would be to avoid remote triggered bombs of any kind - each concurrent request continues to have costs and these need to be capped somehow.

Yeah good point, we need to be careful with this.

dryajov:
Just noting this here as well as we've discussed it during today's meeting, we're keeping OS level DNS "detection" out of scope of this PR. This PR should be only concerned with the actual resolution without attempting to determine OS level DNS settings.

Menduist:
I've got a few CI which fails because of [this](https://github.com/status-im/nim-libp2p/pull/580/checks?check_run_id=2784621121)

```
[Suite] jitter
  [OK] channel should be able to handle erratic read/writes
  [OK] channel should handle 1 byte read/write
Error: execution of an external program failed: '/Users/runner/work/nim-libp2p/nim-libp2p/tests/testnative '
/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/nimblecache-946033505/nimscriptapi_240406716.nim(187, 16)
/Users/runner/work/nim-libp2p/nim-libp2p/libp2p.nimble(42, 10) testnativeTask
/Users/runner/work/nim-libp2p/nim-libp2p/libp2p.nimble(31, 8) runTest
/Users/runner/work/nim-libp2p/nim-libp2p/nim/lib/system/nimscript.nim(260, 7) exec
/Users/runner/work/nim-libp2p/nim-libp2p/nim/lib/system/nimscript.nim(260, 7) Error: unhandled exception: FAILED: nim c --opt:speed -d:debug -d:libp2p_agents_metrics -d:libp2p_protobuf_metrics --verbosity:0 --hints:off --warning[CaseTransition]:off --warning[ObservableStores]:off --warning[LockLevel]:off -d:libp2p_pubsub_sign=true -d:libp2p_pubsub_verify=true   -d:chronicles_log_level=INFO -r tests/testnative [OSError]
```

Is this a flaky test, or an actual issue? (relauching the CI solves it)

dryajov:
> I've got a few CI which fails because of [this](https://github.com/status-im/nim-libp2p/pull/580/checks?check_run_id=2784621121)
> 
> ```
> [Suite] jitter
>   [OK] channel should be able to handle erratic read/writes
>   [OK] channel should handle 1 byte read/write
> Error: execution of an external program failed: '/Users/runner/work/nim-libp2p/nim-libp2p/tests/testnative '
> stack trace: (most recent call last)
> /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/nimblecache-946033505/nimscriptapi_240406716.nim(187, 16)
> /Users/runner/work/nim-libp2p/nim-libp2p/libp2p.nimble(42, 10) testnativeTask
> /Users/runner/work/nim-libp2p/nim-libp2p/libp2p.nimble(31, 8) runTest
> /Users/runner/work/nim-libp2p/nim-libp2p/nim/lib/system/nimscript.nim(260, 7) exec
> /Users/runner/work/nim-libp2p/nim-libp2p/nim/lib/system/nimscript.nim(260, 7) Error: unhandled exception: FAILED: nim c --opt:speed -d:debug -d:libp2p_agents_metrics -d:libp2p_protobuf_metrics --verbosity:0 --hints:off --warning[CaseTransition]:off --warning[ObservableStores]:off --warning[LockLevel]:off -d:libp2p_pubsub_sign=true -d:libp2p_pubsub_verify=true   -d:chronicles_log_level=INFO -r tests/testnative [OSError]
> ```
> 
> Is this a flaky test, or an actual issue? (relauching the CI solves it)

We usually have pretty stable and consistent tests, so I'd say there is most likely an issue.

Menduist:
Oh this means that the test failed, I thought a jitter test raised an exception which crashed the test.

OK, makes senses then :) Just failed on macos and windows, which I don't handle correctly atm

Menduist:
Ok, I think we're pretty good here.

- I moved the resolving into the transport instead of the dialer (for websocket for instance)
- I added a dummy DNS server for the tests
- I added a mock name resolver for tests (I put in libp2p directly cause it may be helpful for users to UT their things using it)

libp2p/nameresolving/dnsresolver.nim
```diff
@@ -0,0 +1,138 @@
...
+  return @[]
+
+method resolveTxt*(
+  self: DnsResolver,
+  address: string): Future[seq[string]] {.async.} =
+
+  let nameservers = self.nameServers
+  trace "Resolving TXT using DNS", address, servers = nameservers.mapIt($it)
+  for server in nameservers:
+    let response = await getDnsResponse(server, address, TXT)
```
arnetheduck, line 128: so the idea is that the first nameserver that raises an exception causes lookup to fail in general?

Menduist: Oh right, I think it should continue, I'll `try except` that :)

libp2p/nameresolving/dnsresolver.nim
```diff
@@ -107,10 +109,11 @@ method resolveIp*(
         let resp = await fut
         for answer in resp.answers:
           resolvedAddresses.incl(answer.toString())
-      except ValueError as e:
-        info "Failed to query DNS", address, error=e.msg
-      except TransportOsError as e:
+      except CancelledError as e:
+        raise
```
arnetheduck, line 113: we avoid this form because it relies on `getCurrentException` which is problematic - `raise e`

Menduist: Ah right in meant to do `except CancelledError` without `as e`, is this also an issue?

arnetheduck: yeah, it's an unfortunate design issue in Nim - best avoid it, it works "sometimes" but it's too much of a mess to remember when, this way we only have one form to debug

libp2p/nameresolving/dnsresolver.nim
```diff
@@ -0,0 +1,147 @@
...
+      requestStream = header.toStream()
+    question.toStream(requestStream)
+
+    let dataLen = requestStream.getPosition()
+    requestStream.setPosition(0)
+
+    var buf = newSeq[byte](dataLen)
+    discard requestStream.readData(addr buf[0], dataLen)
+    return buf
+  except Exception as exc:
```
dryajov, line 42: You don't want to catch the raw `Exception`. The short of it is that this is a catch all that doesn't make distinctions between fatal and recoverable exceptions - really this should have never been exposed, the long version will suck the life out of you :-)

libp2p/transports/tcptransport.nim
```diff
@@ -212,8 +214,30 @@ method dial*(
 
   trace "Dialing remote peer", address = $address
 
-  let transp = await connect(address)
-  return await self.connHandler(transp, Direction.Out)
+  let resolvedAddrs =
```
dryajov, line 217: I don't think this is where we want resolution to happen. I can see several reasons as  to why, i) it makes resolution and dialing tightly coupled, ii) it constrains it to one specific transport and if others want to use it, they have to duplicate this logic, ii) it doesn't interact well with a generic dialing engine that will in the future priorities per transport, parallel dialing and potentially other cool things...

I don't think there is a good place for it right now either TBH, so perhaps it makes sense to create an IP based type that all other related transports can derive from - TCP, WS, Quic...

Menduist: Yes, not ideal, but this is required for WSS.

Since we need to resolve the IP in libp2p to be async, and we need to keep the original hostname to send it has a `Host` header & check the certificates, the WS transport needs to have control over the resolution.

Though, I could move the resolution to `connect` in `wire`, and just make the `WSS` transport resolve before calling `connect`.
But it still won't be in the dialer. To move it to the dialer, we should create a new object which holds the hostname + resolved ip. It's a bit cumbersome though

dryajov: > Since we need to resolve the IP in libp2p to be async, and we need to keep the original hostname to send it has a Host header & check the certificates, the WS transport needs to have control over the resolution.

Hmm, you have a point, we need a way to override the hostname in WS tho, we don't have one right now.

Menduist:
@dryajov one last review, this way I merge this & integrate with ws before merging ws? Thx :)

libp2p/nameresolving/dnsresolver.nim
```diff
@@ -0,0 +1,147 @@
...
+
+  try:
+    await sock.sendTo(dnsServer, addr sendBuf[0], sendBuf.len)
+
+    await receivedDataFuture or sleepAsync(5.seconds) #unix default
+
+    if not receivedDataFuture.finished:
+      raise newException(ValueError, "DNS server timeout")
+
+    var
```
dryajov, line 76: Should this just be moved into the `datagramDataReceived`? I guess it's a matter of style, but it does look odd in terms of flow...

libp2p/nameresolving/dnsresolver.nim
```diff
@@ -0,0 +1,147 @@
...
+  for server in nameservers:
+    var responseFutures: seq[Future[Response]]
+    if domain == Domain.AF_INET or domain == Domain.AF_UNSPEC:
+      responseFutures.add(getDnsResponse(server, address, A))
+
+    if domain == Domain.AF_INET6 or domain == Domain.AF_UNSPEC:
+      let fut = getDnsResponse(server, address, AAAA)
+      if server.family == AddressFamily.IPv6:
+        trace "IPv6 DNS server, puting AAAA records first", server = $server
+        responseFutures.insert(fut)
```
dryajov, line 102: I'm curious, do other resolvers also give preference to `ipv6` addrs?

dryajov (approved):

libp2p/nameresolving/dnsresolver.nim
Menduist: The goal was to have something like that:
```nim
await sock.sendTo(xx)
let result = await sock.recv()
[use result]
```
But chronos udp isn't built like that, so I re-created that as I could. I think this is still clearer than splitting the logic in half (sending and receiving)

libp2p/nameresolving/dnsresolver.nim
Menduist: Really tricky topic, it depends a lot on os etc.

What, eg, browsers will do is connect both in ipv4 & ipv6, and if ipv6 succeeds, kill the ipv4 connection. (To give preference to ipv6, but also have a quick fallback if the connection failed). (as per the Happy Eyeball RFC)
The cost is obviously more resources usages, since you need to create two sockets everytime you connect to a new server, at least, but you get better latency in case of non-working ipv6 (on the client or server side)

I think in our case, I think that would be ill-advised, we should keep the resources usages to a minimum.
So, if we succeed to retrieve the addresses from a IPV6 DNS server, we'll put the IPV6 resolved addresses first, since we know that at least the client has a correct ipv6 configuration.

If we end up in a latency constrained situation, we should instead use a different dialer (for instance a "quick dial"), which tries multiple `ma` in parallel to be sure that we have a working connection ASAP

Though I'll improve a bit things here, by "saving" the fact that a DNS server failed to resolve for future resolves.

Menduist:
Changed my mind again, sorry

I've moved the resolving back to the dialer. For `ws`, I'll probably add a `hostname` parameter to the dialing.

libp2p/nameresolving/dnsresolver.nim
dryajov: How so? As far as I can see it's an async callback that triggers on every packet and _should_ be able to manipulate the incoming data? Like I said, this is more a matter of taste, as your approach works, but it does look odd in terms of flow. At least in my case, I would expect that logic pertaining to manipulating the incoming data happens in the callback that triggers for that data? 

Again, matter of style, just an observation.

libp2p/nameresolving/dnsresolver.nim
dryajov: What concerns me is the success ratio of Ipv6 vs Ipv4 dials. If Ipv6 fails to connect in a large amount of cases (not unreasonable, given it's uneven adoption), then the net effect would be much slower dials.

> If we end up in a latency constrained situation, we should instead use a different dialer (for instance a "quick dial"), which tries multiple ma in parallel to be sure that we have a working connection ASAP

Hmm, not sure, I don't have a strong opinion here, but other implementations of libp2p default to concurrent dials, like you note, this has advantages and disadvantages. My plan has been to add parallel/concurrent dials by default at some point, but you might be right in that this should be user configurable and not enforced.

dryajov:
> Changed my mind again, sorry
> 
> I've moved the resolving back to the dialer. For `ws`, I'll probably add a `hostname` parameter to the dialing.

Hmm, this would be a bit weird, because it would break the existing `dial` interface and expose a parameter an arbitrary parameter to other transports - a bit of an abstraction leak.

Menduist:
We can merge this on approval @dryajov

dryajov (approved):

PR #582, opened 2021-06-01
use LPError more consistently

dryajov:


libp2p/multiaddress.nim
```diff
@@ -909,12 +909,12 @@ proc append*(m1: var MultiAddress, m2: MultiAddress): MaResult[void] =
...
 
 proc `&`*(m1, m2: MultiAddress): MultiAddress {.
-     raises: [Defect, ResultError[string]].} =
+     raises: [Defect, LPError].} =
   ## Concatenates two addresses ``m1`` and ``m2``, and returns result.
   ##
   ## This procedure performs validation of concatenated result and can raise
   ## exception on error.
-  concat(m1, m2).tryGet()
+  concat(m1, m2).expect("Invalid multiaddress after concatenation!")
```
arnetheduck, line 917: expect raises a defect - either the code or the comment is wrong

libp2p/stream/lpstream.nim
```diff
@@ -47,9 +47,9 @@ type
   LPStreamIncorrectDefect* = object of Defect
   LPStreamLimitError* = object of LPStreamError
   LPStreamReadError* = object of LPStreamError
-    par*: ref CatchableError
+    par*: ref Exception
```
arnetheduck, line 50: you should never have to deal with `Exception` like this - ie a `Defect` should propagate directly at the call site, so only `CatchableError` ever needs storing

libp2p/multiaddress.nim
dryajov: Ah, I guess we can't really use `expect/get` with `toException`.

dryajov: A an observation, `raises: [Defect]` hides the fact that `expect/get` will raise a `Defect`. Not sure what's the best way to approach it is - maybe we shouldn't allow any exception to escape with `raises: []`?

PR #584, opened 2021-06-02
add ping protocol

### LINKED ISSUE #567: Support for ping protocol
`nim-libp2p` currently does not support [ping protocol](https://docs.libp2p.io/concepts/protocols/#ping)

There are several reasons why this may be a useful protocol to add, including:
1. To perform simple connectivity and aliveness checks, especially for long-lived connections
2. To be used in keep-alive mechanisms that prevent streams from timing out (see for example [this issue](https://github.com/status-im/nim-waku/issues/504) where a connection times out if peers are idle)

Protocol id is `/ipfs/ping/1.0.0`

There are reference implementations in [go](https://github.com/libp2p/go-libp2p/tree/master/p2p/protocol/ping), [js](https://github.com/libp2p/js-libp2p/tree/master/src/ping) and [rust](https://github.com/libp2p/rust-libp2p/tree/master/protocols/ping)

cc @oskarth @D4nte

Menduist:
closes #567 

Simple ping protocol implemented.
This is a low level version (eg, you can launch a ping using the `ping` proc, and it will reply to ping), like the JS one

It would be possible to add more functionality :
- Waku keepalive seems to use a handler, the `newPing` function could take a handler without much issue if required EDIT: done
- In rust, it looks like the libp2p bundles a Handler which ping automatically

### DIFF
--- a/libp2p/protocols/ping.nim
+++ b/libp2p/protocols/ping.nim
@@ -7,8 +7,7 @@
 ## This file may not be copied, modified, or distributed except according to
 ## those terms.
 
-import std/random
-import chronos, chronicles
+import chronos, chronicles, bearssl
 import ../protobuf/minprotobuf,
        ../peerinfo,
        ../stream/connection,
@@ -36,9 +35,10 @@ type
 
   Ping* = ref object of LPProtocol
     pingHandler*: PingHandler
+    rng: ref BrHmacDrbgContext
 
 proc newPing*(handler: PingHandler = nil): Ping =
-  let ping = Ping(pinghandler: handler)
+  let ping = Ping(pinghandler: handler, rng: newRng())
   ping.init()
   ping
 
@@ -73,8 +73,7 @@ proc ping*(
     randomBuf: array[PingSize, byte]
     resultBuf: array[PingSize, byte]
 
-  for i in 0..<PingSize:
-    randomBuf[i] = byte.rand()
+  p.rng[].brHmacDrbgGenerate(randomBuf)
 
   let startTime = Moment.now()
 
--- a/libp2p/protocols/ping.nim
+++ b/libp2p/protocols/ping.nim
@@ -37,7 +37,7 @@ type
     pingHandler*: PingHandler
     rng: ref BrHmacDrbgContext
 
-proc init*(T: typedesc[Ping], handler: PingHandler = nil, rng: ref BrHmacDrbgContext = newRng()): T =
+proc new*(T: typedesc[Ping], handler: PingHandler = nil, rng: ref BrHmacDrbgContext = newRng()): T =
   let ping = Ping(pinghandler: handler, rng: rng)
   ping.init()
   ping
@@ -48,6 +48,7 @@ method init*(p: Ping) =
       trace "handling ping", conn
       var buf: array[PingSize, byte]
       await conn.readExactly(addr buf[0], PingSize)
+      trace "echoing ping", conn
       await conn.write(addr buf[0], PingSize)
       if not isNil(p.pingHandler):
         await p.pingHandler(conn.peerInfo)
--- a/libp2p/protocols/ping.nim
+++ b/libp2p/protocols/ping.nim
@@ -59,9 +59,14 @@ method init*(p: Ping) =
   p.handler = handle
   p.codec = PingCodec
 
-proc ping*(p: Ping,
-           conn: Connection,
-           ): Future[Duration] {.async, gcsafe.} =
+proc ping*(
+  p: Ping,
+  conn: Connection,
+  ): Future[Duration] {.async, gcsafe.} =
+  ## Sends ping to `conn`
+  ## Returns the delay
+  ##
+
   trace "initiating ping", conn
 
   var
--- a/libp2p/protocols/ping.nim
+++ b/libp2p/protocols/ping.nim
@@ -37,8 +37,8 @@ type
     pingHandler*: PingHandler
     rng: ref BrHmacDrbgContext
 
-proc newPing*(handler: PingHandler = nil): Ping =
-  let ping = Ping(pinghandler: handler, rng: newRng())
+proc init*(T: typedesc[Ping], handler: PingHandler = nil, rng: ref BrHmacDrbgContext = newRng()): T =
+  let ping = Ping(pinghandler: handler, rng: rng)
   ping.init()
   ping
 
--- a/libp2p/protocols/ping.nim
+++ b/libp2p/protocols/ping.nim
@@ -1,0 +1,81 @@
+## Nim-LibP2P
+## Copyright (c) 2021 Status Research & Development GmbH
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+import options, random
+import chronos, chronicles
+import ../protobuf/minprotobuf,
+       ../peerinfo,
+       ../stream/connection,
+       ../peerid,
+       ../crypto/crypto,
+       ../multiaddress,
+       ../protocols/protocol,
+       ../errors
+
+logScope:
+  topics = "libp2p ping"
+
+const
+  PingCodec* = "/ipfs/ping/1.0.0"
+  PingSize = 32
+
+type
+  PingError* = object of LPError
+  WrongPingAckError* = object of LPError
+
+  Ping* = ref object of LPProtocol
+
+proc newPing*(): Ping =
+  new result
+  result.init()
+
+method init*(p: Ping) =
+  proc handle(conn: Connection, proto: string) {.async, gcsafe, closure.} =
+    try:
+      trace "handling ping", conn
+      var buf: array[PingSize, byte]
+      await conn.readExactly(addr buf[0], PingSize)
+      await conn.write(addr buf[0], PingSize)
+    except CancelledError as exc:
+      raise exc
+    except CatchableError as exc:
+      trace "exception in ping handler", exc = exc.msg, conn
+
+  p.handler = handle
+  p.codec = PingCodec
+
+proc ping*(p: Ping,
+           conn: Connection,
+           ): Future[Duration] {.async, gcsafe.} =
+  trace "initiating ping", conn
+
+  var
+    randomBuf: array[PingSize, byte]
+    resultBuf: array[PingSize, byte]
+
+  for i in 0..<PingSize:
+    randomBuf[i] = byte.rand()
+
+  let startTime = Moment.now()
+
+  trace "sending ping", conn
+  await conn.write(addr randomBuf[0], randomBuf.len)
+
+  await conn.readExactly(addr resultBuf[0], PingSize)
+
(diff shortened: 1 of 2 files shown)


libp2p/protocols/ping.nim
```diff
@@ -0,0 +1,81 @@
...
+      await conn.write(addr buf[0], PingSize)
+    except CancelledError as exc:
+      raise exc
+    except CatchableError as exc:
+      trace "exception in ping handler", exc = exc.msg, conn
+
+  p.handler = handle
+  p.codec = PingCodec
+
+proc ping*(p: Ping,
```
dryajov, line 53: minor nitpick, we usually prefer python style indentation:

```nim
proc ping*(
  p: Ping, 
  conn: Connection)...

**Comment by dryajov** [comment]
> Waku keepalive seems to use a handler, the newPing function could take a handler without much issue if required

Yeah I think a ping handler would be a useful addition especially if it's used as a keepalive probe.

**Review by dryajov — APPROVED** [review]

**Comment by arnetheduck** [comment]
>  @dryajov dryajov force-pushed the unstable branch from d83dda0 to fbe888a 11 hours ago 

force-pushing shared branches is probably best avoided - it messes up everything around them (like this PR)

File: libp2p/protocols/ping.nim (hunk 1 above)
Comment (Menduist) [reply libp2p/protocols/ping.nim]: Oh okay :)

File: libp2p/protocols/ping.nim [hunk 2]
```diff
@@ -0,0 +1,95 @@
+## Nim-LibP2P
+## Copyright (c) 2021 Status Research & Development GmbH
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+import options, random
```
arnetheduck, line 10: we avoid the `std/random` module and use bearssl instead (look for `rng`)

also, all std lib imports should be imported with an explicit `std` prefix - `std/[options, ...]`

libp2p/protocols/ping.nim
```diff
@@ -0,0 +1,95 @@
...
+  PingHandler* = proc (
+    peer: PeerInfo):
+    Future[void]
+    {.gcsafe, raises: [Defect].}
+
+  Ping* = ref object of LPProtocol
+    pingHandler*: PingHandler
+
+proc newPing*(handler: PingHandler = nil): Ping =
+  new result
```
arnetheduck, line 41: we generally avoid `result` and prefer object-style init (`let ping = Ping(pinghandler = handler); ping.init(); ping`)

libp2p/protocols/ping.nim
```diff
@@ -0,0 +1,95 @@
...
+  let startTime = Moment.now()
+
+  trace "sending ping", conn
+  await conn.write(addr randomBuf[0], randomBuf.len)
+
+  await conn.readExactly(addr resultBuf[0], PingSize)
+
+  let responseTime = Moment.now() - startTime
+
+  trace "got ping response", conn, responseTime
```
arnetheduck, line 88: we've recently started naming all timer logs `XxxDur` (for `Duration`) which makes them easier to grep / find - this avoids the ambiguity of `Time`, reserving it for a moment in time.

tests/testping.nim
```diff
@@ -0,0 +1,86 @@
...
+      pingReceivedCount = 0
+
+    asyncTeardown:
+      await conn.close()
+      await acceptFut
+      await transport1.stop()
+      await serverFut
+      await transport2.stop()
+
+    asyncTest "simple ping":
```
arnetheduck, line 59: tests should include a cancellation test / timeout test - `await pingProto.ping().withTimeout(0)` basically, with no responding peer - this will cause the original request to be cancelled which triggers an oft-misunderstood exception path.

timeouts implicitly cancel the future they were timing out - an alternative way to test cancellations is `let p = pingProto.ping(); await p.cancelAndWait()`.

ideally, there should also be a test for an invalid ping response (ie bad random data)

libp2p/protocols/ping.nim
Menduist: Seems a bit wasteful to use a crypto rand generator for a simple ping?
And std/random is already used in the libp2p, wouldn't have used it otherwise ;)

But sure, can switch to bearssl

arnetheduck: avoiding `std/random` makes the code easier to audit in general, at no real loss - ie there's nothing really wasteful about it except a few CPU cycles which - `std/random` usage is slowly getting replaced with the crypto rng throughout, also to avoid accidental uses in places where it matters more directly - the question is better asked in the other direction: what special reasons are there to use an insecure rng here? If there are, they should be documented.

dryajov:
> > @dryajov dryajov force-pushed the unstable branch from [d83dda0](https://github.com/status-im/nim-libp2p/commit/d83dda07a2ac9408cc5bfe7c9f5c9ce76189bfe8) to [fbe888a](https://github.com/status-im/nim-libp2p/commit/fbe888a3aa349e1ae36319613cf31e690cd8fd23) 11 hours ago
> 
> force-pushing shared branches is probably best avoided - it messes up everything around them (like this PR)

You can't rebase without force pushing and there isn't really, what's the difference between doing keep up with master and a shared branch - you get the same effects.

arnetheduck:
> You can't rebase 

well, that's the point - it's [poor form](https://www.google.com/search?client=firefox-b-d&q=golden+rule+of+rebasing+git) to rebase public shared branches - if there's work against them, merging becomes a lot more difficult - it's really basic git hygiene - for public shared branches, you typically create merge commits to sync up work - see stable/unstable of nimbus-eth2 for example

dryajov:
> > You can't rebase
> 
> well, that's the point - it's [poor form](https://www.google.com/search?client=firefox-b-d&q=golden+rule+of+rebasing+git) to rebase public shared branches - if there's work against them, merging becomes a lot more difficult - it's really basic git hygiene - for public shared branches, you typically create merge commits to sync up work - see stable/unstable of nimbus-eth2 for example

Yeah I agree, but I thought we're avoiding merges - I don't have anything against merging.

libp2p/protocols/ping.nim
```diff
@@ -0,0 +1,94 @@
...
+  PingHandler* = proc (
+    peer: PeerInfo):
+    Future[void]
+    {.gcsafe, raises: [Defect].}
+
+  Ping* = ref object of LPProtocol
+    pingHandler*: PingHandler
+    rng: ref BrHmacDrbgContext
+
+proc newPing*(handler: PingHandler = nil): Ping =
```
dryajov, line 40: We use `init(T: typedesc[Ping]): T` to init objects and `new(T: typedesc[Ping]): ref T` or `new(T: typedesc[Ping]): PingRef` to create object refs. We should use the `init` style here.

libp2p/protocols/ping.nim
```diff
@@ -0,0 +1,94 @@
...
+  let ping = Ping(pinghandler: handler, rng: newRng())
```
dryajov, line 41: The `rng` should be passed in. We should be able to reuse an already existing one, which is created either by the consuming application or libp2p proper, otherwise we're instantiating a new one every time.

libp2p/protocols/ping.nim
Menduist: Oh, I copied that from Identify

tests/testping.nim
```diff
@@ -0,0 +1,140 @@
...
+        #await conn.write("Hello dns!")
+        var buf: array[32, byte]
+        await conn.readExactly(addr buf[0], 32)
+        await conn.write(addr buf[0], 32)
+
+      let handlerWait = acceptHandler()
+
+      let streamTransport = await transportdialer.dial(transport.ma)
+
+      let p = pingProto2.ping(streamTransport)
```
dryajov, line 93: This is actually pretty tricky - as it is, this test isn't really testing cancellation there are no guarantees that by the time you call `cancelAndWait` the sequence hasn't already completed. What needs to happen is that right after accepting the connection we await an "empty" future and then call `p.cancelAndAwait` at least then, we know for sure that we're cancelling in the middle of the sequence and even then we're only properly testing the caller.

Unfortunately we can't test all the cancellation points, specially on the receiver end and we potentially need something like #520 to do proper fuzzing.

libp2p/protocols/ping.nim
dryajov: Yeah, that needs updating too ;)

tests/testping.nim
Menduist: My understanding is that the `cancelAndWait` will be called just after the first `await` of `ping` (so just after it sent the data).
Since chronos is single threaded, this should be always the case. To test at each step we could do something like:
```nim
for pollCount in 0..20:
   setupPingTest()
   let future = pingProto2.ping(streamTransport)
   for _ in 0..<pollCount:
      if future.finished: break
      poll()
   if future.finished: break #We actually finished the sequence
   await future.cancelAndWait()
```

This way, we're `poll`'ing one more time at each loop, and thus testing the cancellation in a further point of the sequence each time. But I haven't dug into chronos, so maybe I missed something

arnetheduck:
> Yeah I agree, but I thought we're avoiding merges - I don't have anything against merging.

we avoid merges for feature PR's because it makes them harder to revert, specially if they are made up of many small commits that do not all work. The stable->unstable commit likely wouldn't be reverted, specially not like a feature.

Reverting features should be cheap so we can do it liberally - that constraint does not apply to `unstable`

Menduist:
Hey @jm-clius, does this API looks good to you or do you want something more?

jm-clius:
> Hey @jm-clius, does this API looks good to you or do you want something more?

No, this looks good! Thanks for finishing this so quickly.

tests/testping.nim
dryajov: The idea is that we at least test it at a point where it's actually done something. Cancelling right after the ping will potentially happen before any real work has begun. Cancelling after the connection has been established will have more cleanup to do and thus will end up testing more of the flow.

I'm curious to see what the snippet actually does. It doesn't seem it will get past a specific point it the cancelation?

tests/testping.nim
```diff
@@ -0,0 +1,147 @@
...
+        proc acceptHandler() {.async, gcsafe.} =
+          let handler = Ping.new().handler
+          let conn = await transport.accept()
+          await handler(conn, "na")
+
+        let handlerWait = acceptHandler()
+
+        let streamTransport = await transportdialer.dial(transport.ma)
+
+        discard await pingProto2.ping(streamTransport)
```
dryajov, line 92: I think we should probably return the ping future here - `return pingProto2.ping(streamTransport)`, otherwise `p.cancelAndWait` is cancelling the entire test no the ping sequence.

dryajov: Btw, just curious, why `20` - how did you get to that number or is it just a safe number of iterations?

Menduist: Yeah, 20 is just a big enough number. Could be bigger to be safe, but for ping the protocol is simple enough that it takes at most ~13 polls.

I can't really `return ping` because it would mix `await` and `return Future()` in a single function, which is not possible AFAIK
But I could iterate directly in `testPing` to cancel only the ping.

Not sure how the cancel is propagated to child futures, need to take a deeper look at that

Issue #537, opened 2021-02-25
Peer and Connection Management


dryajov:
# Peer and Connection Management

> Documenting current thinking about peer and connection manager

## Peer Book

The peer book stores serializable information about a peer.

(@jm-clius worked on the peer book implementation in - https://github.com/status-im/nim-libp2p/issues/504)
(reference [go](https://github.com/libp2p/go-libp2p-peerstore) and [js](https://github.com/libp2p/js-libp2p/tree/master/src/peer-store) implementations )

- A central repository to track peers and their information
- It should be possible to update this during the lifetime of a peer
- Should allow subscribing to events
- Track peers across restarts - persistence
  - Should be contemplated, but it shouldn't be part of libp2p, at the very least however it should have an interface that allows serialization, which applications can implement however they want

## Peer Manager

Closely related to Peer Book, but not the same thing. This is more concerned with managing _connected_ peers. Overall, I see this as the action of ranking peers with active connection, keeping limits and connecting/disconnecting on demand.

The main thing to understand with peer management is that there should be a balance between allowing applications to control who they are connected too, without affecting other applications (ie protocols). This is a bit more complex than the peer book, which is more concerned with keeping up to date information about a peer. We can use the Peer Book for storing this information such as connection related stats, scores, etc...

It's worth stressing the difference between Peers and Connections. A Peer can have several Connections and also, connections are transport dependent. Some transports are stream oriented, with muxers that produce channels on top of them - like in the case of TCP and mplex. Others are lightweight connections built on top of UDP - like in the case of uTP or quic. In any case, from the application standpoint, there should not be any distinction, but from the standpoint of peer and connection management there obviously is.

This calls to make a distinction between closing connections vs disconnecting peers. Closing a connection would not drop the peer from the peer manager, ie all other protocols with connections to that peer would keep those connection. However, dropping the peer means closing **all** the connections for that peer and removing from the peer manager.

With this in mind, there are several related counts that need to be kept

- Total number of peers
- Total number of connection
- Total number of streams

This distinctions are important, because some transports,will have hundreds of connections for a peer - as it is the case of quic. While for others, will have at most a few - as in the case of TCP streams, with many muxed streams on top. Each of this counts needs to be kept separately, but each will eventually affect the peer's score, hence it's ability to stay connected or be dialed in a near future.

So the peer manager will have to take all of that into account, which isn't the case right now. For starters, muxers shouldn't even be part of the peer manager.

[...]

PR #586, opened 2021-06-04
Peer Store implement part II

Menduist:
Following #504 
Ref #537

This PR add a `PeerStore` to the `Switch`.
The PeerStore is still user-overridable to wrap it with other fields if needed (as done in Waku)

The connmanager's PeerEvent has a new type of event: `Identify`, which is triggered when the peer identifies (or `pushidentity`, in the future)
On identify, the peer informations are store inside the PeerStore through a `PeerEventHandler`

**Breaking change**
The `ConnEventHandler` & `PeerEventHandler` signatures changed: they now receive a `PeerInfo` instead of a `PeerID`

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -64,8 +64,7 @@ proc new*(T: type[SwitchBuilder]): T =
     maxOut: -1,
     maxConnsPerPeer: MaxConnectionsPerPeer,
     protoVersion: ProtoVersion,
-    agentVersion: AgentVersion,
-  )
+    agentVersion: AgentVersion)
 
 proc withPrivateKey*(b: SwitchBuilder, privateKey: PrivateKey): SwitchBuilder =
   b.privKey = some(privateKey)
@@ -181,8 +180,7 @@ proc build*(b: SwitchBuilder): Switch
     muxers = muxers,
     secureManagers = secureManagerInstances,
     connManager = connManager,
-    ms = ms
-  )
+    ms = ms)
 
   return switch
 
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -15,11 +15,11 @@ import
   crypto/crypto, transports/[transport, tcptransport],
   muxers/[muxer, mplex/mplex],
   protocols/[identify, secure/secure, secure/noise],
-  connmanager, upgrademngrs/muxedupgrade,
+  connmanager, upgrademngrs/muxedupgrade, peerstore,
   errors
 
 export
-  switch, peerid, peerinfo, connection, multiaddress, crypto, errors
+  switch, peerid, peerinfo, connection, multiaddress, crypto, errors, peerstore
 
 type
   SecureProtocol* {.pure.} = enum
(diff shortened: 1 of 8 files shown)


dryajov:
Generally looks good. The one thing to keep in mind is that `PeerInfo` is a ref type and so we have to be careful if using it as hash/set keys, but I don't think this PR affects that. I'll do a more thorough review on Monday.

Menduist:
Small update on this PR. Having a bit more time to think about it, I don't think the `PeerStore` should be user-overridable.

In the end, each protocol/application may need to store information about each peer.
So each protocol/application may need to create a custom `PeerStore`
By making the libp2p's `PeerStore` extendable, we allow *one* application or protocol to override it, but every other application/protocol will need to store their data independently anyway

So I think it's best to use the `PeerStore` to store only the libp2p "core required" data, and let each app/protocol create their own `PeerStore` with their custom `books`.

We could create a generic `PeerStore` with no book, let app/protos extend it with their books, and for the "core required" libp2p data extend it ourself. But ATM, I don't think it's super relevant, since each book will need to be managed manually anyway.

arnetheduck:
> So I think it's best to use the PeerStore to store only the libp2p "core required" data

:+1: - this makes sense in general and the peer lifecycle events are there so that the various application stores can be kept in sync. One can imagine arbitrary data being attached to a peer through a key-value store of store, but that's easy to make a mess of.

libp2p/builders.nim
```diff
@@ -64,7 +64,8 @@ proc new*(T: type[SwitchBuilder]): T =
     maxOut: -1,
     maxConnsPerPeer: MaxConnectionsPerPeer,
     protoVersion: ProtoVersion,
-    agentVersion: AgentVersion)
+    agentVersion: AgentVersion,
+  )
```
dryajov, line 68: Yaiks! can we not do the new line in this kinda calls and istead use the old style:

```nim
agentVersion: AgentVersion)
```

This triggers my OCD every time :) 

(unless this is a new addition to the style guide...)

Menduist: It's just badly-reverted changes from a50bc81, mb

dryajov:
> Small update on this PR. Having a bit more time to think about it, I don't think the `PeerStore` should be user-overridable.
> 
> In the end, each protocol/application may need to store different informations about each peer.
> So each protocol/application may need to create a custom `PeerStore`
> By making the libp2p's `PeerStore` extendable, we allow _one_ application or protocol to override it, but every other application/protocol will need to store their data independently anyway
> 
> So I think it's best to use the `PeerStore` to store only the libp2p "core required" data, and let each app/protocol create their own `PeerStore` with their custom `books`.
> 
> We could create a generic `PeerStore` with no book, let app/protos extend it with their books, and for the "core required" libp2p data extend it ourself. But ATM, I don't think it's super relevant, since each book will need to be managed manually anyway.

Yeah, I agree with the reasoning. I think this should be mergeable now 👍

dryajov (approved):

PR #587, opened 2021-06-09
Identify Push

### LINKED ISSUE #48: Implement push identify
Push identify is missing and needs to be implemented as per the spec - https://github.com/libp2p/specs/blob/master/identify/README.md#identifypush.

Menduist:
closes #48 

Draft because I haven't tested it with another libp2p implementation to check compability.
Could use a few more tests


Menduist:
Now works with libp2p-js implem, good for review

dryajov:
We do have interop tests with `Go` through our daemon API. It would be nice to see if we can enable push identify in Go and add a couple of tests for good measure.

Menduist:
Ok, I'll probably add the interop tests in a separate PR, along with some `ping` tests

PR #588, opened 2021-06-10
Fix trackers tests

Menduist:
There is a typo in `test/helpers.nim` which made `checkTracker` useless. Fixing it broke a few tests

(nim doesn't have a "unused parameter" warning? seems like a super important warning)

### DIFF
--- a/tests/testswitch.nim
+++ b/tests/testswitch.nim
@@ -718,8 +718,7 @@ suite "Switch":
     checkTracker(SecureConnTrackerName)
     checkTracker(ChronosStreamTrackerName)
 
-    await allFuturesThrowing(
-      switch1.stop())
+    await switch1.stop()
 
     # this needs to go at end
     await allFuturesThrowing(awaiters)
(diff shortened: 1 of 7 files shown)


Menduist:
Ok, fortunately it seems that it was only bad tests, not actual issues :)
I'll let you check @dryajov

dryajov (approved):
Great to have a fresh pair of eyes on the codebase - great find!

tests/testswitch.nim
```diff
@@ -712,13 +713,13 @@ suite "Switch":
       readers.add(closeReader())
 
     await allFuturesThrowing(readers)
+    await switch2.stop() #Otherwise this leeks
     checkTracker(LPChannelTrackerName)
     checkTracker(SecureConnTrackerName)
     checkTracker(ChronosStreamTrackerName)
 
     await allFuturesThrowing(
```
dryajov, line 721: I don't think we need `allFuturesThrowing` here anymore?

dryajov (approved):
LGTM

PR #592, opened 2021-06-22
Fix flaky macos switch test

Menduist:
Seen on #565 

Probably not the best way to solve it, but MacOS seem takes some time to close the stream on the receiving end, and thus the stream isn't closed quickly enough.

This PR solves it, as demonstrated here: https://github.com/status-im/nim-libp2p/runs/2887560022

### DIFF
--- a/tests/testswitch.nim
+++ b/tests/testswitch.nim
@@ -713,7 +713,9 @@ suite "Switch":
       readers.add(closeReader())
 
     await allFuturesThrowing(readers)
-    await switch2.stop() #Otherwise this leeks
+    await switch2.stop() #Otherwise this leaks
+    await sleepAsync(500.millis)
+
     checkTracker(LPChannelTrackerName)
     checkTracker(SecureConnTrackerName)
     checkTracker(ChronosStreamTrackerName)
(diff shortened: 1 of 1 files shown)


dryajov:
Good thinking, this is a definitely a common occurrence and it's the only place where we allow for sleeps in tests (tests that don't adhere to this, are misusing sleep) .

tests/testswitch.nim
```diff
@@ -713,7 +713,9 @@ suite "Switch":
       readers.add(closeReader())
 
     await allFuturesThrowing(readers)
-    await switch2.stop() #Otherwise this leeks
+    await switch2.stop() #Otherwise this leaks
+    await sleepAsync(500.millis)
```
markspanbroek, line 717: We can avoid slowing down the test run by waiting for the connection to drop:

```suggestion
    check await checkExpiring(not switch1.isConnected(switch2.peerInfo.peerID))
```

Tested here: https://github.com/status-im/nim-libp2p/runs/2892600589?check_suite_focus=true

Menduist: :+1:

PR #593, opened 2021-06-24
Websocket Transport

### LINKED ISSUE #407: Websockets support
# Problem

For Waku v2 to run in a web browser, an important design goal, there needs to be a way to communicate between the browser and a Waku node.

The current PoC (https://github.com/vacp2p/waku-web-chat/) is using a basic NodeJS implementation supporting TCP, Websockets and WebRtcStar. This then acts as a bridge between browser (WS/WRTC) and nim-waku (TCP) and relays messages between them.

However, it'd be useful to replace this with nim-waku. This will be needed to avoid duplicate work, and other protocols are req-resp, which would mean creating a full implementation of Waku v2 in NodeJS, which is obviously not desirable if we can avoid it.

# Acceptance criteria

- Websockets transport support
- Can use to support e.g. Waku Web browser
- Works in a deployed and secure environment (wss)

Menduist:
closes #407 

Full WIP

Edit: green ci!


libp2p/transports/wstransport.nim
```diff
@@ -0,0 +1,189 @@
...
+method accept*(self: WsTransport): Future[Connection] {.async, gcsafe.} =
+  ## accept a new WS connection
+  ##
+
+  if not self.running:
+    raise newTransportClosedError()
+
+  try:
+    let transp = await self.httpserver.accept()
+    #TODO could this be blocking? If so, someone could stuck our accept loop
```
dryajov, line 135: chronos is completely async and you'd need to go out of your way (e.g. use `waitFor()`) to combine it with sync code, so should be easy to spot. The only caveat is long CPU intensive tasks that don't yield to back to the main event loop.

Menduist: I meant blocking on the sense of this accept loop,
Eg, if someone start to send the http request, but doesn't follow up on a ws connection or send a malicious request, could we get stuck between accept and handle? Thus, blocking other peers from connecting

Menduist: Put in an other way, for the duration of `handleRequest`, we don't accept other peers since there is a single accept loop. So, if a malicious/bad behaving peer can make `handleRequest` slow, he could potentially block us from accepting other peers.
Thus, either `handleRequest` should always finish in a timely manner, or we should implement a better thing here to allow multiple connection requests in parallel.

This was more a note to myself, to check if `handleRequest` is instantaneous or if I need to allow parallel connection attempts :)

dryajov: Ah, that's a good question and there are several layers of indirection here. 

First there is the TCP backlog (100 in chronos by default), which means that the OS (or TCP stack) will queue up to backlog size incoming connections, this connections will time out eventually, but they won't immediately drop giving it a chance of being picked up at some point. 

The second layer is that we can have several concurrent accepts running at the same time, we don't currently do this in the switch because we haven't seen an issue with incoming connections, but this is certainly possible.

In any case, I wouldn't actually worry about this in the transport itself, the calling application should be able to decide what's a good number of concurrent accepts to run, whether it should periodically timeout, etc... This move the complexity up to a layer that can actually make informed decisions about it.

Menduist: Yes, but the specificity of WS is that the accept is a two phase process in the application, whereas for TCP for instance, when `accept` returns, you know you have a valid & ready connection
So while for TCP, having a single accept loop is sufficient because a bad behaving peer can't block it, for WS we need to consider the possibility of having the accept loop blocked.

Anyway, the only `await`s in `handleRequest` are `send`s, so I don't think a bad behaving peer could block `handleRequest`, since in the worst case, the OS will buffer the send, so no issue here AFAICT :)

dryajov: Yes, by the time we've got to `handleRequest` we've already read all the data (since it's an HTTP request and headers aren't streamed), but we can still have a malicious attacker simply not finishing writing the headers, so I think a timeout when reading headers would be appropriate - this however should go in `nim-ws`.

libp2p/transports/wstransport.nim
```diff
@@ -0,0 +1,199 @@
...
+method readOnce*(
+  s: WsStream,
+  pbytes: pointer,
+  nbytes: int):
+  Future[int] =
+  return s.session.recv(pbytes, nbytes)
+
+method write*(s: WsStream, msg: seq[byte]):
+  Future[void] {.async.} =
+  #TODO the write procs can raise LPStreamEOF
```
dryajov, line 59: Unfortunately, chronos don't yet support specific annotations on async transformations. See https://github.com/status-im/nim-libp2p/pull/547, most of what's needed to support it is already there.

libp2p/transports/wstransport.nim
```diff
@@ -0,0 +1,199 @@
...
+
+type
+  WsStream = ref object of Connection
+    session: WSSession
+
+proc init*(T: type WsStream,
+           session: WSSession,
+           dir: Direction,
+           timeout = 10.minutes,
+           observedAddr: MultiAddress = MultiAddress()): T =
```
dryajov, line 41: I can't see a need to pass `observedAddr` here. This information is usually only available after we've received a connection, specifically during identify, we _might_ learn about a new address that the remote was able to reach us over. Even if we are able to populate it from the incoming raw http request, it's still something that is internal to the transport.

dryajov: Ah, nevermind - I got confused thinking this is the transport being initialized.

libp2p/transports/wstransport.nim
```diff
@@ -0,0 +1,199 @@
...
+
+method closeImpl*(s: WsStream): Future[void] {.async.} =
+  await s.session.close()
+  await procCall Connection(s).closeImpl()
+
+type
+  WsTransport* = ref object of Transport
+    httpserver: HttpServer
+    wsserver: WSServer
+    outconnections: seq[WsStream]
```
dryajov, line 77: Shouldn't be this something similar to `clients: array[Direction, seq[WSSession]]` (similar to how we handle it in the tcptransport)?

libp2p/transports/wstransport.nim
```diff
@@ -0,0 +1,199 @@
...
+method accept*(self: WsTransport): Future[Connection] {.async, gcsafe.} =
+  ## accept a new WS connection
+  ##
+
+  if not self.running:
+    raise newTransportClosedError()
+
+  try:
+    let
+      transp = await self.httpserver.accept()
```
dryajov, line 144: Lots of potential to unify with tcp transport `accept`, specially after we've enabled multiple address listening.

libp2p/transports/wstransport.nim
Menduist: Moved upstream: https://github.com/status-im/nim-websock/issues/79

libp2p/transports/wstransport.nim
Menduist: Oups forgot the incoming connections
Though why store the incoming and outgoing in separate seq? Just for perf?

libp2p/transports/wstransport.nim
Menduist: Yep, when #598 will be more stable I'll rebase this PR on it and add the required changes :)

libp2p/transports/wstransport.nim
dryajov: Can't remember why it was done this way, but it really isn't that much overhead and it's always nice to have this things segregated.

Menduist:
Well, the interop test with go was surprisingly uneventful, so this PR should be pretty good?
I'm calling it ready for review, and we'll see if we merge it before #598, since #598 is currently stuck and no one actually need it ATM

dryajov:
I didn't see interop tests for websockets in the `testinterop.nim` file, we should definitelly have those added.

tests/testinterop.nim
```diff
@@ -288,6 +288,49 @@ suite "Interop":
     await daemonNode.close()
     await sleepAsync(1.seconds)
 
+  asyncTest "daemon -> websocket connection":
```
dryajov, line 291: Shouldn't we also have a complement to this, where we dial the go daemon on the WS?

dryajov:
LGTM, besides the missing `websocket -> daemon` test, lets add it please.

dryajov (dismissed):

dryajov (approved):

PR #594, opened 2021-06-25
Various transports improvement

Menduist:
While working on WS transport, I've seen a few possible improvements on the transport system, so here a few "fixes":
- TcpTransport.init should be named TcpTransport.new
- localAddress wasn't either used or actually declared in TcpTransport, removed (could add it back easily in the future if needed)
- upgradeOutgoing & upgradeIncoming should be the same for most transports, so moved from TcpTransport to Transport (still overridable if needed)
- removed unused field `multicodec`
- Moved E2E test to a reusable file for all transports, eg :
```nim
TcpTransport.commonTransportTest("/ip4/0.0.0.0/tcp/0")
WsTransport.commonTransportTest("/ip4/0.0.0.0/tcp/0/ws")
QuicTransport.commonTransportTest("/ip4/0.0.0.0/udp/0/quic")
```

Ideally we would make most test in the common file to avoid duplication between transport, and just leave transport-dependant test in each transport's test

cc @markspanbroek, tell me if this looks good to you, or if you have more ideas to ease the quic implementation


libp2p/transports/transport.nim
```diff
@@ -99,9 +95,3 @@ method handles*(
   # having to repeat the check in every transport
   if address.protocols.isOk:
     return address.protocols.get().filterIt( it == multiCodec("p2p-circuit") ).len == 0
-
-method localAddress*(self: Transport): MultiAddress {.base, gcsafe.} =
```
dryajov, line 103: Maybe just fix it to return the address as corresponds - we can use it in the switch I believe.

dryajov (dismissed):
Some very needed cleanup - thanks!

markspanbroek:
Looks great!

libp2p/transports/transport.nim
Menduist: The switch just get the `transport.ma` field. Though we could add `localAddress` back and make him return `transport.ma` to allow override-ability.. but we don't need to, so I would keep it that way

libp2p/transports/tcptransport.nim
```diff
@@ -114,21 +114,24 @@ proc connHandler*(self: TcpTransport,
...
+  upgrade: Upgrade): T {.deprecated: "use .new".} =
+  T.new(flags, upgrade)
+
+proc new*(
+  T: typedesc[TcpTransport],
   flags: set[ServerFlags] = {},
   upgrade: Upgrade): T =
 
-  result = T(
+  var transport = T(
```
dryajov, line 128: probably want `let`?

arnetheduck (approved):

PR #598, opened 2021-06-29
Add support for multiple addresses to transports

### LINKED ISSUE #597: The transport system doesn't allow multiple local addresses
Currently, each transport is `start`ed [for each supported address](https://github.com/status-im/nim-libp2p/blob/72ccd1d1f55764df35087b063ed38f688f517d6a/libp2p/switch.nim#L365).

This is wrong, because a transport _could_ handle multiple local address: eg, if bound to multiple local addresses, bound to multiple ports, or ipv4+ipv6.
In the current form, a transport handling multple addresses starts & stops multiple time, ignoring extraneous addresses, causing exceptions in some cases.

Not sure what is the best way to solve this, inputs are welcomed.

My ideas:
- Make each transport handle multiple addresses instead of one (each transport will have to start multiple servers, have a accept loop for each server, etc). Not a big fan of this, because this will lead to duplicated logic for each transport
- Transform the transports into factories, which are called for each applicable address. The only issue I see with this is that the `handles` method will be used both in the factory and the actual transport. Not too big a deal
- ?

I've created a branch to work on this, currently I just added a test case for this issue: https://github.com/status-im/nim-libp2p/commit/3a05eba0e72e97927bef43da870f091f86af18ef

cc @dryajov @markspanbroek

dryajov:
closes #597 

This is building on top of https://github.com/status-im/nim-libp2p/compare/fixmultiplelocaladdresses

It attempts to add multiple address support per transport without much changes or refactoring. This is the approach I always considered the most straightforward and at some point transport was built around the idea of multiple transport however, this got dropped during one of the numerous refactors and, since we didn't have another transport besides TCP it never got addressed properly. I believe that this is a flexible enough way of handling this scenario.

This is only part one of a two; the next thing would be to move most of the dialing logic to the transport as well, at which point we can make `dial` take an array of addresses as well. Eventually, most of the logic from the switch should migrate to the transports - this is more flexible given that each transport can potentially implement a radically different dialing and upgrade flow and at the same time this would help implement a more abstract dialing engine capable of handling parallel dials, backoffs and redials consistently across all transports.


libp2p/transports/tcptransport.nim
```diff
@@ -188,7 +204,20 @@ method accept*(self: TcpTransport): Future[Connection] {.async, gcsafe.} =
...
+
+    if self.acceptFuts.len <= 0:
+      return
+
+    let finished = await one(self.acceptFuts)
+    self.acceptFuts.keepItIf(
+      it != finished
+    )
+
+    let transp = await finished
```
Menduist, line 220: This looks wrong to me, but I may be wrong

From what I understand, if we have two servers:
- We fill the acceptFuts with both accepts
- One accept eventually succeed, so we remove it from the `acceptFuts`
- On the next iteration `acceptFuts` only contains the second server, and since it's len is > 0 we don't re-add the first server
- So we only listen for the second server on the second iteration

I'll add a commit to fix this, but if your code was good feel free to revert mine

Menduist:
Nice :+1: 

So you had my "option 1" in mind, and you're right, we should give as much control as possible to the transports.
I only had TCP, quic and WS in mind, which in the end are 90% similar, and that's why I didn't want duplicated logic.
But if we want to handle more exotic transports in the future, like raw ip or ble or whatever, they should indeed have as much control as possible

And if we really want to avoid duplicated logic, we can always create a "ReliableTransport" or whatever, which tcp, quic & ws extends, and only have each reliabletransport with an accept (a single server) and a dial, for instance. We'll see if that makes sense in the future

Menduist:
I've added a test in common transport with a few scenarios that would have completely failed a few commits ago

Menduist:
I removed the macos flaky cancellation test for the moment, because it's not related to this PR and is blocked ATM.

Also adapted the fix of #609.

Should be ready for review!

Menduist:
Green CI thanks to the fix https://github.com/status-im/nim-chronos/pull/220, so ready for review!

Menduist:
I fixed websocket to allow both WS & WSS address in a single transport

dryajov:
LGTM, can we get this merged?

Menduist (approved):

PR #599, opened 2021-06-30
Add examples to CI

Menduist:
WIP PR to improve the "getting started" flow

The examples are now compiled in the CI to avoid outdated examples, and a more simple "hello world" has been added


Menduist:
I've reworked and simplified the examples a bit

Should we add a more complete example? Eg, a chat based on pubsub or something like that?
The directchat doesn't really show connection management so could be interesting, what do you think @dryajov?

dryajov:
> I've reworked and simplified the examples a bit
> 
> Should we add a more complete example? Eg, a chat based on pubsub or something like that?
> The directchat doesn't really show connection management so could be interesting, what do you think @dryajov?

Sure, more examples would be great! 

We don't have to do a chat example and in fact this example ended up being a bit more involved than expected because of blocking std-io. Perhaps simpler, standalone and compilable snippets that demonstrate pubsub, how to write a new protocol, etc, would be more appropriate.

Menduist:
I've added a system which extracts nim code from markdown files, tell me if this looks good to you
If it's good, I'll do some actual tutorials instead of this stub one :)

Example: https://github.com/status-im/nim-libp2p/blob/4c4b4c3b5e2b7420a4c3b74b32ee72571bb2776b/examples/tutorial_1_connect.md

dryajov (approved):

Menduist:
I would like to add more tutorials (gossipsub etc), but this PR is already very beneficial as-is
So let's merge this and add more tutorials in the future

examples/tutorial_1_connect.md
```diff
@@ -0,0 +1,108 @@
...
+
+You can now install the latest version of `nim-libp2p`:
+```bash
+nimble install libp2p@#master
+```
+
+# A simple ping application
+We'll start by creating a simple application, which is starting two libp2p [switch](https://docs.libp2p.io/concepts/stream-multiplexing/#switch-swarm), and pinging each other using the [Ping](https://docs.libp2p.io/concepts/protocols/#ping) protocol.
+
+_You can extract the code from this tutorial by using `nim c -r tools/markdown_runner.nim examples/tutorial_1_connect.md`!_
```
dryajov, line 23: Maybe annotate with `TIP/HINT`? It took me a sec realizing that I was being hinted at something.

examples/tutorial_2_customproto.md
```diff
@@ -0,0 +1,82 @@
...
+
+We've set a [protocol ID](https://docs.libp2p.io/concepts/protocols/#protocol-ids), and created a custom `LPProtocol`. In a more complex protocol, we could use this structure to store interesting variables.
+
+A protocol generally has two part: and handling/server part, and a dialing/client part.  
+Theses two parts can be identical, but in our trivial protocol, the server will wait for a message from the client, and the client will send a message, so we have to handle the two cases separately.
+
+Let's start with the server part:
+```nim
+proc new(T: typedesc[TestProto]): T =
+  # every incoming connections will in handled in this closure
```
dryajov, line 31: Looks like a typo? `every incoming connections will **be** handled in this closure`?

examples/directchat.nim
```diff
@@ -63,143 +97,97 @@ proc dialPeer(p: ChatProto, address: string) {.async.} =
...
-      if p.connected:
-        await p.conn.writeLp(line)
+      if c.connected:
+        await c.conn.writeLp(line)
       else:
         try:
           if line.startsWith("/") and "p2p" in line:
-            await p.dialPeer(line)
+            await c.dialPeer(line)
         except:
```
dryajov, line 141: I think we're now avoiding this style due to multiple language issues. It's probably better to use `exception CatchableError as exc`.

dryajov (approved):
LGTM aside from the couple of minor comments.

examples/directchat.nim
Menduist: Yeah, it's the issue with long running PRs, I would have fixed this in the various cleanups part 1 :)

dryajov (approved):

PR #600, opened 2021-06-30
make `isLiteral` compatible with ARC and ORC

narimiran:
This change is needed when testing with `--gc:orc`.

When we're testing this package with Nim devel and ORC, we always have to apply this patch, so it might be a good idea to have it in the upstream code. (There's no change for the default GC)


Menduist:
LGTM, but could you PR against unstable and not master? Thanks :)

narimiran:
> LGTM, but could you PR against unstable and not master? Thanks :)

Here: https://github.com/status-im/nim-libp2p/pull/602

PR #602, opened 2021-06-30
make `isLiteral` compatible with ARC and ORC, take 2

narimiran:
The same as https://github.com/status-im/nim-libp2p/pull/600, but against `unstable` branch.

### DIFF
--- a/libp2p/vbuffer.nim
+++ b/libp2p/vbuffer.nim
@@ -37,10 +37,13 @@ proc len*(vb: VBuffer): int =
   doAssert(result >= 0)
 
 proc isLiteral[T](s: seq[T]): bool {.inline.} =
-  type
-    SeqHeader = object
-      length, reserved: int
-  (cast[ptr SeqHeader](s).reserved and (1 shl (sizeof(int) * 8 - 2))) != 0
+  when defined(gcOrc) or defined(gcArc):
+    false
+  else:
+    type
+      SeqHeader = object
+        length, reserved: int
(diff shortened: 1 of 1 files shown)


libp2p/vbuffer.nim
```diff
@@ -37,10 +37,13 @@ proc len*(vb: VBuffer): int =
   doAssert(result >= 0)
 
 proc isLiteral[T](s: seq[T]): bool {.inline.} =
-  type
-    SeqHeader = object
-      length, reserved: int
-  (cast[ptr SeqHeader](s).reserved and (1 shl (sizeof(int) * 8 - 2))) != 0
+  when defined(gcOrc) or defined(gcArc):
```
dryajov, line 40: Have the new semantics changed that this is now irrelevant or is it a just quick workaround to make arc/orc compile? 

As I understand this, it's trying to determine if we're dealing with simple types vs complex objects that can be shallow copied. 

Seems like a pretty ugly hack to begin? Does it makes sense to remove it altogether?

Araq: The hack was/is probably justified for the resulting speed improvements. Arc/Orc has a different internal seq/string implementation so this hack stopped working. With Arc/Orc you can get the same speed improvement in a principled manner by a strategic `sink` parameter in some constructor(s). It's not hard to find these places but currently I cannot remember them.

Menduist (approved):

Menduist (approved):

Menduist:
Sorry about all that, apparently we now need signed commit for unstable so I signed the commit for you :)

PR #603, opened 2021-07-02
Faster CI

Menduist:
This PR disable CI on push. The goal is stop running the CI twice as it currently does, since it's not necessary and the CI runs end up piling in the actions queue for hours.

To explain `pull_request` vs `push` quickly:
`push` will simply test every time you push to any branch
`pull_request` will test every time you push to a PR __as if the pull request was merged__. You can see it a "merge temporarily, run the CI, and unmerge"

For codecov, it doesn't make much sense to CI on the commit itself, so only doing it on `pull_request` seems legit

For the actual testing, it's not as simple. Only running on `pull_request` means that branches which are not PRed will not be tested. But if you actually want to test a thing without a PR, it's easy to add `push` back to the file, or run the workflow manually inside the github UI (note: to run manually, this PR (as in #603) [will have to be merged into master](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow)).

Why not run on `push` only? Because it's more interesting to test the PR as if they were merged.

I propose to set both to `pull_request` only, and if this annoys us, switch `ci` back to `push, pull_request`. But I don't think we'll even notice it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,5 @@
 name: CI
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,10 @@
 name: CI
-on: [pull_request, workflow_dispatch]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
 
 jobs:
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,5 @@
 name: CI
-on: [pull_request]
+on: [pull_request, workflow_dispatch]
 
 jobs:
   build:
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,6 +1,12 @@
 name: nim-libp2p codecov builds
 
-on: [pull_request]
+on:
+  #On push to common branches, this computes the "bases stats" for PRs
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
 
 jobs:
   GossipSub:
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,6 +1,6 @@
 name: nim-libp2p codecov builds
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   GossipSub:


stefantalpalaru:
You can get fancier and keep "push" events for your main branch(es): https://github.com/status-im/nim-metrics/blob/f812b268473ace7817ade20dd8e6d0574f4b5460/.github/workflows/ci.yml#L2-L7

That way "master" still gets tested after merging PRs in it.

Menduist:
Oh yeah, good idea

And for unstable, we always have a PR to master opened, so not needed :+1:

dryajov:
@Menduist this isn't how it works according to this discussion - https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012/4. 

The important part to grok is:

`So two workflow runs are different. On push checks the branch code, and on pull_request checks the pre-merge code. `

You aren't testing the same git hashes - `push` checks the branch, `pull_request` will check the the result of the possible merge.

Menduist:
Yes, that is was I was trying to explain, but maybe I failed :d

My point is, we don't really care about the standalone branch, we just care about the tests passing after a possible merge. So we can safely ignore the `push`, imo

stefantalpalaru:
> You aren't testing the same git hashes

You are, in a PR branch. Print `git rev-parse --short HEAD` to verify.

> pull_request will check the the result of the possible merge

No, that's just an event trigger. No mock-merge is being done there.

dryajov:
@stefantalpalaru 

> No, that's just an event trigger. No mock-merge is being done there.

Source?

Menduist:
There is a mock-merge:
[Push](https://github.com/status-im/nim-libp2p/runs/2966806991?check_suite_focus=true)
```
/usr/bin/git log -1 --format='%H'
'd5dc2bc81f599697bd5cff48ec14ef287be9c013'
```
d5d is my commit


[pull_request](https://github.com/status-im/nim-libp2p/runs/2966807453?check_suite_focus=true)
```
/usr/bin/git log -1 --format='%H'
'99f8dc3163826a7c441e7611d983b020c6d81cc5'
```
99f is not on my computer

dryajov:
So my point is that we have to be careful that we're testing what is needed. 

The first rule here is  "les is NOT more"; the second rule is that if you need to **remember to do something** to make sure it's right, then you've just "pointed a gun at your fut", it's only a matter before it sets off.

> we just care about the tests passing after a possible merge. So we can safely ignore the push, imo

Hmm, this does make sense, I hope we're not missing anything :)

stefantalpalaru:
I stand corrected. There is a merge showing up in the "Checking out the ref" phase of the "Checkout nim-libp2p" step:

```text
/usr/bin/git checkout --progress --force refs/remotes/pull/593/merge
[...]
HEAD is now at 99f8dc3 Merge d5dc2bc81f599697bd5cff48ec14ef287be9c013 into c12f00c8b6b43d3f23851e602c66cf48b053d824
```

However, if we're running the CI tests on this merge commit, also running them on the PR branch HEAD commit is redundant. There is no scenario in which the merge will pass but the HEAD will fail, so it's just a waste of CI resources.

Menduist:
I just wonder what happens if there is a conflict and the merge can't be created?

But even if it doesn't work in this case (as in, the CI fails), you just have to update your PR to the up branch, so not too bad either

stefantalpalaru:
> I just wonder what happens if there is a conflict and the merge can't be created?

Maybe the merge is done to the common commit - the one you branched from - so it's guaranteed to work.

Menduist:
Well, then this entire thing would be useless, no? Since it would always be a FF merge

mratsim:
We should do that on all of our projects since the CI is org-wide.

stefantalpalaru:
> We should do that on all of our projects since the CI is org-wide.

I proposed that and met resistance from people using non-PR branches to run the test suite in CI. That only works if the "push" event is enabled for all branches.

I, for one, don't see the point of stealth branches because I just use a "[WIP]" PR to do that, but developer comfort is more important than saving CI resources and shortening the waiting time.

Menduist:
Isn't shorting there waiting time also developer confort? ;)

We could also set to push only, but it's better to test the post-merge result

stefantalpalaru:
> We could also set to push only

That won't work for PRs made from other repos.

dryajov:
I'm fine with the changes as long as we're safe in all possible scenarios.

markspanbroek (approved):

dryajov:
Anything else we're missing here?

Menduist:
> I just wonder what happens if there is a conflict and the merge can't be created?

To answer my own question, `pull_request` won't run if there is a conflict, which may be a bit of a pain because you have to stay mergable with the base branch to have the CI run.

PR #607, opened 2021-07-14
Muxer bandwidth metrics

Menduist:
Total bytes sent in the muxer exposed to metrics

Since this is in the muxer (to differentiate the protocols flows), it's not actual bandwidth (the muxer & other layers overhead are not counted), but gives a good idea of BW usage

To keep cardinality under control, there is one label per known (mounted or dialed) protocol + a global in / out

Here it is running on nimbus for a little while, first syncing then attesting with all subnets:
image

cc @arnetheduck

### DIFF
--- a/libp2p/muxers/mplex/lpchannel.nim
+++ b/libp2p/muxers/mplex/lpchannel.nim
@@ -21,6 +21,8 @@ export connection
 logScope:
   topics = "libp2p mplexchannel"
 
+declareCounter libp2p_muxer_bytes, "total sent or received bytes", ["protocol"]
+
 ## Channel half-closed states
 ##
 ## | State    | Closed local      | Closed remote
--- a/libp2p/muxers/mplex/lpchannel.nim
+++ b/libp2p/muxers/mplex/lpchannel.nim
@@ -21,7 +21,8 @@ export connection
 logScope:
   topics = "libp2p mplexchannel"
 
-declareCounter libp2p_muxer_bytes, "total sent or received bytes", ["protocol"]
+when defined(lipp2p_network_protocols_metrics):
+  declareCounter libp2p_protocols_bytes, "total sent or received bytes", ["protocol", "direction"]
 
 ## Channel half-closed states
 ##
@@ -159,9 +160,9 @@ method readOnce*(s: LPChannel,
   ## or the reads will lock each other.
   try:
     let bytes = await procCall BufferStream(s).readOnce(pbytes, nbytes)
-    libp2p_total_bytes.inc(bytes.int64, labelValues=["in"])
-    if s.tag.len > 0:
-      libp2p_total_bytes.inc(bytes.int64, labelValues=[s.tag])
+    when defined(lipp2p_network_protocols_metrics):
+      if s.tag.len > 0:
+        libp2p_protocols_bytes.inc(bytes.int64, labelValues=[s.tag, "in"])
 
     trace "readOnce", s, bytes
     if bytes == 0:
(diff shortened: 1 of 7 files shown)


dryajov:
We should probably put this behind a flag - it's very easy to flood prometheus with too many metrics. We've ran into this issue multiple times in the past and it's why we're being very sparing with metrics, and specifically label heavy metrics.

arnetheduck:
> To keep cardinality under control, there is one label per known (mounted or dialed) protocol + a global in / out

We probably want to see separate in/out for each protocol - from there, a global can be computed

>  We should probably put this behind a flag

Is it possible for a remote to influence the number of tags? If not, this looks benign:ish, ie in the tens?

Menduist:
I made a global in/out to limit the number of metrics, with a in/out per protocol nimbus would already have 18 time series, whereas right now it only has 11.

And it's pretty easy to guess which protocol caused a in or out spike by looking at the graph, I think

And no, the remote can't add new tags, only known, mounted protocols will be counted

I could leave the global one always on (so only 2 metrics), and put the per protocol one with a CT switch

arnetheduck:
> I made a global in/out to limit the number of metrics, with a in/out per protocol nimbus would already have 18 time series, whereas right now it only has 11.

In/out for most protocols is highly uneven - ie a block request is 200 bytes in and lots of bytes out if it comes from outside - others should be more or less even or slightly biased - ie gossipsub should initially send to all neighbours but when neighbours send messages that have already been seen, this should not result in further fan-out etc - these are important hints to understand systemic issues or bugs in the implementation.

arnetheduck:
> I could leave the global one always on (so only 2 metrics), and put the per protocol one with a CT switch

it might be a good idea to give it a different name, so that the "parametrised" metrics can be summed etc using functions that collect all together

arnetheduck:
See also https://github.com/ethereum/eth2.0-metrics/pull/12/files for "standard" metric names, from which we can get inspiration and alignment

Menduist:
Ok, so I've made one always on global traffic metrics (which has the added benefit of being actual BW since it's at the bottom layer):
image

And a flag activated one, with metric per protocol per side:
image

libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -21,6 +21,9 @@ export connection
 logScope:
   topics = "libp2p mplexchannel"
 
+when defined(lipp2p_network_protocols_metrics):
+  declareCounter libp2p_protocols_bytes, "total sent or received bytes", ["protocol", "direction"]
```
arnetheduck, line 25: btw, what happens for streams / protocols like quic that presumably don't use mplex?

Menduist: We'll need to add the increments in every `lpchannel` equivalent (and probably move the declaration of the counter to `Connection` directly)

Unfortunately, we currently don't have a central place where we could add this metric, because
1) readOnce & writeOnce can be called directly and do not call the parent method
2) and if they called the parent method, that would be wrong because each layer would add to the counter, where in reality we only want to increment in the muxer

dryajov: Keep in mind, that we already have some traffic metrics, but at the chronosstream layer - https://github.com/status-im/nim-libp2p/blob/unstable/libp2p/stream/chronosstream.nim#L31-L34, this was yet another one off attempt. At this rate we'll end up with a bunch of this everywhere, I think it's time we think of a more consistent approach?

dryajov (approved):
Looks good, but this is definitely not a scalable approach long term, we've got way too many metrics spread across the entire codebase - just look at pubsub. This is getting quite a bit unmanageable (IMO) and we should start thinking about some less invasive and more scalable way of doing this. At the very least, we should catalog the ones we have.

libp2p/muxers/mplex/lpchannel.nim
Menduist: Yes, theses ones were made to be per agent ones. Maybe we could have some flags-defined labels?
So one could choose which flags is used during compilation: agent, protocol, direction, etc.

But if we activate all flags, we would have `nagent * nprotocol * ndirection` time value series, instead of `nagent + nprotocol + ndirection` one, so not ideal..

dryajov: Yeah, that's why I say that this is becoming unmanageable, we have this metrics spread across everywhere and they are now hard to track. I don't have a straightforward solution for this unfortunately and it does look like it's at a point where it requires some focused attention.

PR #608, opened 2021-07-15
Fix tls

dryajov:
@Menduist this addresses some potential race conditions during closing as well as some minor cleanup. The core of the issues related to Tls tests are being addressed in https://github.com/status-im/nim-websock/pull/84


Menduist (approved):
Thanks!

PR #609, opened 2021-07-16
Fix transport's switch start

Menduist:
The current way the switch is started is wrong, because the transport may take some time to `start` (it's async), but the switch assumes that the `t.ma` is valid and that the `t.running` is true without waiting for `start` to finish.

This can have some pretty bad consequence (for instance not starting the accept loop, as happens here https://ci.status.im/blue/organizations/jenkins/nimbus%2Fnimbus-eth2%2Fmultibranch/detail/PR-2724/1/pipeline/), so merging against master.

By the way, I never understood why we returned the futures of the transport starting? I would remove it, but it would break API. So returning an empty seq

EDIT: not returning and empty seq anymore to allow caller to check how many transports start to fail. But we could find a better way I'm sure (eg returning the number of transports which started successfully, allowing the caller to do `assert await node.start() > 0` or something like that)

### DIFF
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -203,32 +203,6 @@ proc accept(s: Switch, transport: Transport) {.async.} = # noraises
         await conn.close()
       return
 
-proc start*(s: Switch): Future[seq[Future[void]]] {.async, gcsafe.} =
-  trace "starting switch for peer", peerInfo = s.peerInfo
-  var startFuts: seq[Future[void]]
-  for t in s.transports: # for each transport
-    for i, a in s.peerInfo.addrs:
-      if t.handles(a): # check if it handles the multiaddr
-        let transpStart = t.start(a)
-        startFuts.add(transpStart)
-        try:
-          await transpStart
-          s.peerInfo.addrs[i] = t.ma # update peer's address
-          s.acceptFuts.add(s.accept(t))
-        except CancelledError:
-          raise
-        except CatchableError as exc:
-          debug "Failed to start one transport", address = $a, err = exc.msg
-          continue
-
-  proc peerIdentifiedHandler(peerInfo: PeerInfo, event: PeerEvent) {.async.} =
-    s.peerStore.replace(peerInfo)
-
-  s.connManager.addPeerEventHandler(peerIdentifiedHandler, PeerEventKind.Identified)
-
-  debug "Started libp2p node", peer = s.peerInfo
-  return startFuts # listen for incoming connections
-
 proc stop*(s: Switch) {.async.} =
   trace "Stopping switch"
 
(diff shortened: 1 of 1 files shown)


libp2p/switch.nim
```diff
@@ -209,10 +209,9 @@ proc start*(s: Switch): Future[seq[Future[void]]] {.async, gcsafe.} =
   for t in s.transports: # for each transport
     for i, a in s.peerInfo.addrs:
       if t.handles(a): # check if it handles the multiaddr
-        var server = t.start(a)
+        await t.start(a)
```
arnetheduck, line 212: this will raise an exception if start fails, unlike the previous code which would keep going - is that intended?

Menduist: This is [how nimbus](https://github.com/status-im/nimbus-eth2/blob/07a1c5716b707e7244eb312bed68c92c1fb9508d/beacon_chain/networking/eth2_network.nim#L1175-L1179) expects it

But I can change it, I can see cases were you don't mind about loosing one transport

arnetheduck: not quite - if you read the code carefully, you'll see that the previous code _never raises_ an exception - that means that nimbus will never catch one either.

In fact, the way the code is currently written, exceptions raised during `start` will silently be ignored by nimbus (which is wrong) - if an exception happens, it will be stored in one of the returned futures _inside_ the sequence, but nobody ever checks those - this needs fixing both here and in nimbus

arnetheduck: > By the way, I never understood why we returned the futures of the transport starting? I would remove it, but it would break API. So returning an empty seq

that's also the answer to why the futures are returned: if one of the accepts fails, the calling code can in theory figure out why and which connection failed - this is of course an extremely poor API to do so (because it requires knowledge about iteration order which is an implementation detail, but nonetheless).

It's a classic example of why we've moved on from exceptions in general: the semantics are hard to reason about and it's easy to miss where errors happen and why.

libp2p/switch.nim
```diff
@@ -209,9 +209,17 @@ proc start*(s: Switch): Future[seq[Future[void]]] {.async, gcsafe.} =
...
-        await t.start(a)
-        s.peerInfo.addrs[i] = t.ma # update peer's address
-        s.acceptFuts.add(s.accept(t))
+        let transpStart = t.start(a)
+        startFuts.add(transpStart)
+        try:
+          await transpStart
+          s.peerInfo.addrs[i] = t.ma # update peer's address
+          s.acceptFuts.add(s.accept(t))
+        except CancelledError:
```
arnetheduck, line 218: this introduces a leak because one one start has succeeded and cancellation happens during the second, the first start is "lost"

libp2p/switch.nim
Menduist: > not quite - if you read the code carefully, you'll see that the previous code _never raises_ an exception - that means that nimbus will never catch one either.
> 
> In fact, the way the code is currently written, exceptions raised during `start` will silently be ignored by nimbus (which is wrong) - if an exception happens, it will be stored in one of the returned futures _inside_ the sequence, but nobody ever checks those - this needs fixing both here and in nimbus

Yes, that's why I said that nimbus _expected_ for libp2p to throw, but indeed, libp2p doesn't throw :)

I think it would be better to return the number of transport that failed to start, and then let the application either quit immediately, or check inside `switch.transports` which one failed to start if they don't mind loosing a few ones

libp2p/switch.nim
```diff
@@ -250,6 +231,33 @@ proc stop*(s: Switch) {.async.} =
...
+    for i, a in s.peerInfo.addrs:
+      if t.handles(a): # check if it handles the multiaddr
+        let transpStart = t.start(a)
+        startFuts.add(transpStart)
+        try:
+          await transpStart
+          s.peerInfo.addrs[i] = t.ma # update peer's address
+          s.acceptFuts.add(s.accept(t))
+        except CancelledError:
+          await s.stop()
```
arnetheduck, line 247: technically, this will call stop on transports which have not yet started .. that's an imbalance in starts vs stops, but one can hope that's handled somewhere..

libp2p/switch.nim
```diff
@@ -250,6 +231,33 @@ proc stop*(s: Switch) {.async.} =
...
+  trace "starting switch for peer", peerInfo = s.peerInfo
+  var startFuts: seq[Future[void]]
+  for t in s.transports: # for each transport
+    for i, a in s.peerInfo.addrs:
+      if t.handles(a): # check if it handles the multiaddr
+        let transpStart = t.start(a)
+        startFuts.add(transpStart)
+        try:
+          await transpStart
+          s.peerInfo.addrs[i] = t.ma # update peer's address
```
arnetheduck, line 244: if there are two transports, won't this overwrite `addrs[i]`?

Menduist: The idea is that each transport handle a different set of addresses, but it's not handled correctly ATM, we're working on it in #598

libp2p/switch.nim
Menduist: Yes, stopping a non started transport is not an issue

libp2p/switch.nim
```diff
@@ -250,6 +231,33 @@ proc stop*(s: Switch) {.async.} =
...
+          s.acceptFuts.add(s.accept(t))
+        except CancelledError:
```
dryajov, line 246: `except CancelledError as exc: raise exc` - the alternative is highly error prone and has shown to regress in several occations.

dryajov (approved):
LGTM, good comments by @arnetheduck 👍

PR #610, opened 2021-07-22
Remove peer info

Menduist:
Really wip,
Removing peer info from the connection in favor of the PeerStore in the switch

A connection now __must__ hold a PeerID, whereas before it was optional. It seems OK with the libp2p specs, though we should update PlainText to 2.0.0 to get peerids with this protocol


libp2p/protocols/identify.nim
```diff
@@ -150,24 +157,26 @@ proc identify*(p: Identify,
...
-              local = remotePeerInfo.peerId
+              local = remotePeerId
 
         raise newException(IdentityNoMatchError, "Peer ids don't match")
+  else:
+    raise newException(IdentityInvalidMsgError, "No pubkey in identify")
 
-proc new*(T: typedesc[IdentifyPush], connManager: ConnManager): T =
-  let identifypush = T(connManager: connManager)
+proc new*(T: typedesc[IdentifyPush], handler: IdentifyPushHandler = nil): T =
```
dryajov, line 175: Should we split this out into it's own PR?

Menduist: It would duplicate a bit of work, since this PR completely breaks the identify process anyway
And right now nobody uses the IdentifyPush, so should be ok to modify it "sneakily"

But I can split if you want

Menduist:
My last commit is a bit "controversial",

Now `PeerInfo` is _only_ used for the local peer info (so as a "Private Key + addresses + protos" store, basically)
It doesn't seem to break everything, and at least it's "simpler", but if someone don't like it I can revert it easily.

Apart from that, we got green ci :tada:, but I need to check that the `shortAgent` thing is good.

dryajov:
> My last commit is a bit "controversial",
> 
> Now `PeerInfo` is _only_ used for the local peer info (so as a "Private Key + addresses + protos" store, basically)
> It doesn't seem to break everything, and at least it's "simpler", but if someone don't like it I can revert it easily.
> 
> Apart from that, we got green ci 🎉, but I need to check that the `shortAgent` thing is good.

I would unify `PeerInfo` with `StoredInfo` from the peerstore and use that for the local info as well, no need to duplicate structures if they are almost identical.

Menduist:
In the end I removed the PrivateKey from PeerInfo, made the PublicKey mandatory, and StoredInfo is now an alias to PeerInfo

Menduist:
Without further modification, `PeerInfo` needs to stay a `ref`, because it's shared by Identify & the Switch

Could change the flow a bit, but I think this PR is big enough as is

libp2p/switch.nim
```diff
@@ -250,6 +245,21 @@ proc stop*(s: Switch) {.async.} =
 
   trace "Switch stopped"
 
+proc updatePeerInfo*(
```
dryajov, line 248: Is this being called anywhere?

dryajov (dismissed):
LGTM

libp2p/switch.nim
```diff
@@ -267,26 +277,8 @@ proc newSwitch*(peerInfo: PeerInfo,
     transports: transports,
     connManager: connManager,
     peerStore: PeerStore.new(),
-    dialer: Dialer.new(peerInfo, connManager, transports, ms))
+    dialer: Dialer.new(peerInfo.peerId, connManager, transports, ms))
 
+  switch.connManager.peerStore = switch.peerStore
```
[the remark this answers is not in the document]
Menduist: You mean the connManager constructor?
The issue is that currently the connManager is created in the builder (or app), and the peerStore is created directly here.

Or you mean pass the peerStore to `newSwitch`, and though create the peerStore in the builder?

libp2p/switch.nim
Menduist: Not ATM (only required in a test), but this is in preparation to the integration of push identify in the switch.
Not done in this PR because this will change a few things too unrelated

Menduist: I've moved to PeerStore, now it's used both by the regular upgrade flow and the push identify

libp2p/protocols/pubsub/rpc/message.nim
```diff
@@ -77,9 +77,6 @@ proc init*(
     let peer = peer.get()
     msg.fromPeer = peer.peerId
     if sign:
-      if peer.keyType != KeyType.HasPrivate:
-        raise (ref LPError)(msg: "Cannot sign message without private key")
-
       msg.signature = sign(msg, peer.privateKey).expect("Couldn't sign message!")
       msg.key = peer.privateKey
```
dryajov, line 81: yaiks! this needs to be public key!

Menduist: Ah, `getKey` actually gives the publicKey, and since the chaining is on next line, it looks wrong here, but is actually giving the publicKey.
I'll do a PR to rename `getKey` to `getPublicKey` to avoid the confusion

Menduist: #621

jm-clius (approved):
Going to result in a number of interesting breaks in `nim-waku`, but otherwise LGTM! 😄

dryajov (dismissed):
LGTM

Menduist:
Last commit fixes the shortAgent - it's pretty ugly, can't wait to clean up the metrics

PR #615, opened 2021-08-05
Integrate dns resolving

### LINKED ISSUE #571: Support for DNS addresses

Menduist:
Follow up of #580
Closes #571

TLDR of current DNS state:
Actually resolving multiaddress can be tricky in our infra, because:
- For WSS, we need to keep the original hostname
- We would like to deduplicate resolved addresses to avoid redundant dial
- But in the case of WSS, a different hostname/ip pair can have a different meaning, and we should dial both
- We need to keep the resolving in the dialer (instead of the transport), because `dnsaddr` could resolve into different transports

So it seems that our only option is to resolve in the dialer, deduplicate on `hostname/ip` pair (or not deduplicate at all), and give the hostname to the transport.

Is there a better option?

### DIFF
--- a/libp2p/dialer.nim
+++ b/libp2p/dialer.nim
@@ -20,6 +20,7 @@ import dial,
        connmanager,
        stream/connection,
        transports/transport,
+       nameresolving/nameresolver,
        errors
 
 export dial, errors
@@ -41,6 +42,7 @@ type
     connManager: ConnManager
     dialLock: Table[PeerID, AsyncLock]
     transports: seq[Transport]
+    nameResolver: NameResolver
 
 proc dialAndUpgrade(
   self: Dialer,
@@ -54,9 +56,15 @@ proc dialAndUpgrade(
     transport: Transport
     address: MultiAddress
 
+  let resolvedAddresses =
+    if not isNil(self.nameResolver):
+      await self.nameResolver.resolveMAddresses(addrs)
+    else:
+      addrs
+
   for t in self.transports: # for each transport
     transport = t
-    for a in addrs:      # for each address
+    for a in resolvedAddresses:      # for each address
       address = a
       if t.handles(a):   # check if it can dial it
         trace "Dialing address", address = $a, peerId
@@ -234,9 +242,11 @@ proc new*(
   peerInfo: PeerInfo,
   connManager: ConnManager,
   transports: seq[Transport],
-  ms: MultistreamSelect): Dialer =
+  ms: MultistreamSelect,
+  nameResolver: NameResolver = nil): Dialer =
 
   T(peerInfo: peerInfo,
     connManager: connManager,
     transports: transports,
-    ms: ms)
+    ms: ms,
+    nameResolver: nameResolver)
--- a/libp2p/dialer.nim
+++ b/libp2p/dialer.nim
@@ -56,59 +56,61 @@ proc dialAndUpgrade(
     transport: Transport
     address: MultiAddress
 
-  let resolvedAddresses =
-    if not isNil(self.nameResolver):
-      await self.nameResolver.resolveMAddresses(addrs)
-    else:
-      addrs
+  for rawAddress in addrs:      # for each address
+    address = rawAddress
+    let
+      hostname = address.getHostname()
+      resolvedAddresses =
+        if isNil(self.nameResolver): @[address]
+        else: await self.nameResolver.resolveMAddress(address)
 
-  for t in self.transports: # for each transport
-    transport = t
-    for a in resolvedAddresses:      # for each address
-      address = a
-      if t.handles(a):   # check if it can dial it
-        trace "Dialing address", address = $a, peerId
-        let dialed = try:
-            libp2p_total_dial_attempts.inc()
-            # await a connection slot when the total
-            # connection count is equal to `maxConns`
-            await self.connManager.trackOutgoingConn(
-              () => transport.dial(address)
-            )
-          except TooManyConnectionsError as exc:
-            trace "Connection limit reached!"
-            raise exc
-          except CancelledError as exc:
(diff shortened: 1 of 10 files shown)


jm-clius:
Any further progress on this? 😃

Menduist:
If this is urgent, we can merge this as is, and handle WSS dns in a future PR (since the custom host is not implemented yet in nim-websock anyway)

jm-clius:
> If this is urgent, we can merge this as is, and handle WSS dns in a future PR (since the custom host is not implemented yet in nim-websock anyway)

Thanks, @Menduist. Not (yet) urgent, as far as I know, but will sync with team members on this. 
We _are_ using the [DNS resolver](https://github.com/status-im/nim-libp2p/pull/580) already and it works well. :)

Menduist:
For simplicity stake, I've chosen not to deduplicate addresses for now. If it become important in the future it's still possible to add it, but as of now, that would complicate the code for few real use cases.

libp2p/dialer.nim
```diff
@@ -54,53 +56,61 @@ proc dialAndUpgrade(
...
-            debug "Upgrade failed", msg = exc.msg, peerId
-            if exc isnot CancelledError:
-              libp2p_failed_upgrades_outgoing.inc()
-            raise exc
-
-        doAssert not isNil(conn), "connection died after upgradeOutgoing"
-        debug "Dial successful", conn, peerId = conn.peerId
-        return conn
+  for rawAddress in addrs:      # for each address
+    address = rawAddress
```
dryajov, line 60: It doesn't look like we need `var address` from above anymore and we should be able to simplify the for to `for address in addrs: ...`?

tests/testnameresolve.nim
```diff
@@ -90,18 +90,6 @@ suite "Name resolving":
 
       check testOne("/ip6/::1/tcp/0", "/ip6/::1/tcp/0")
 
-    asyncTest "test multiple resolve":
```
dryajov, line 93: why is this removed?

dryajov: oh I see, you're removing `resolveMAddresses`... hmm, ok I guess it is a better interface overall since a name can resolve to multiple addrs.

tests/testswitch.nim
```diff
@@ -901,5 +901,28 @@ suite "Switch":
       switch1.peerStore.addressBook.get(switch2.peerInfo.peerId) == switch2.peerInfo.addrs.toHashSet()
       switch2.peerStore.addressBook.get(switch1.peerInfo.peerId) == switch1.peerInfo.addrs.toHashSet()
 
-      switch1.peerStore.addressBook.get(switch2.peerInfo.peerId) == switch2.peerInfo.addrs.toHashSet()
-      switch2.peerStore.addressBook.get(switch1.peerInfo.peerId) == switch1.peerInfo.addrs.toHashSet()
+      switch1.peerStore.protoBook.get(switch2.peerInfo.peerId) == switch2.peerInfo.protocols.toHashSet()
+      switch2.peerStore.protoBook.get(switch1.peerInfo.peerId) == switch1.peerInfo.protocols.toHashSet()
+
+  asyncTest "e2e dial dns address":
```
dryajov, line 907: I think we're missing tests for hostnames that resolve to multiple addrs, it would also be nice to test this across both tcp and ws transports.

libp2p/dialer.nim
Menduist: Good point, this code wasn't working in 1.4 actually, I fixed it

tests/testnameresolve.nim
Menduist: Yeah, we could wrap `resolveMAddress` in a `resolveMAddresses`, but we wouldn't use it in libp2p since we need to keep the original hostname for each original maddress

tests/testswitch.nim
Menduist: Done

Menduist:
While adding the last test, I though about an other complication

If we have
`/dnsaddr/test1.io` which resolves to `/dns4/test2.io/tcp/64/wss`

We would ideally need to dial test2.io using the test2.io hostname, but currently we keep the original test1.io hostname, which could create issues while using wss

(to be fair, right now it wouldn't work at all since we don't allow resolving dns4 addresses inside a dnsaddr)

Someone said can of worms? :sweat_smile:
I'll leave it as is for now, but something to keep in mind..

dryajov:
> Maybe a way to handle all of this more cleanly would be to split the dnsaddr and dns* resolution in two phases, dnsaddr in the dialer and dns* in the transport

Actually, this doesn't sound like a bad idea, but I need to think about it a little more - can of worms 😭

dryajov (approved):
LGTM, even tho there might be dragons.

PR #616, opened 2021-08-18
Fix gossipsub incoming graft backoff

Menduist:
Currently, the `BackoffSlackTime` is taken into account to refuse incoming `grafts`. 

This is an issue, because a peer may send us a graft as soon as the backoff time is finished, not knowing about this 2 seconds delay, and we'll end up refusing his graft.

This PR fixes that by substracting two time the BackoffSlackTime on incoming graft request. The first time to fix the issue, and a second time to avoid race conditions.

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -105,9 +105,13 @@ proc handleGraft*(g: GossipSub,
 
       continue
 
+    # Check backingOff
+    # Ignore BackoffSlackTime here, since this only for outbound activity
+    # and subtract a second time to avoid race conditions
+    # (peers may wait to graft us as the exact instant they're allowed to)
     if  g.backingOff
           .getOrDefault(topic)
-          .getOrDefault(peer.peerId) > Moment.now():
+          .getOrDefault(peer.peerId) - (BackoffSlackTime * 2).seconds > Moment.now():
       debug "attempt to graft a backingOff peer", peer, topic
       # and such an attempt should be logged and rejected with a PRUNE
       prunes.add(ControlPrune(
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -89,7 +89,7 @@ proc handleGraft*(g: GossipSub,
     # It is an error to GRAFT on a explicit peer
     if peer.peerId in g.parameters.directPeers:
       # receiving a graft from a direct peer should yield a more prominent warning (protocol violation)
-      warn "attempt to graft an explicit peer, peering agreements should be reciprocal",
+      warn "an explicit peer attempted to graft us, peering agreements should be reciprocal",
         peer, topic
       # and such an attempt should be logged and rejected with a PRUNE
       prunes.add(ControlPrune(
@@ -112,7 +112,7 @@ proc handleGraft*(g: GossipSub,
     if  g.backingOff
           .getOrDefault(topic)
           .getOrDefault(peer.peerId) - (BackoffSlackTime * 2).seconds > Moment.now():
-      debug "attempt to graft a backingOff peer", peer, topic
+      debug "a backingOff peer attempted to graft us", peer, topic
       # and such an attempt should be logged and rejected with a PRUNE
       prunes.add(ControlPrune(
         topicID: topic,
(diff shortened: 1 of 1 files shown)


Menduist:
Should we also remove the clamp on incoming `prune`?
It was added here with few explanation https://github.com/status-im/nim-libp2p/pull/539/files#diff-4a8279b47cb86bc36bbdcfb9c0f999a7de0e62fb4ced0c51e1a8fd3aa6cf5a50
Others implementations don't have it

It seems a bit rude to ignore backoff requests from other peers, especially since we could enter a "graft infinite loop":
- Someone `prune` us with a 60s backoff
- We clamp it to 30s
- After 30s, our backoff is finished, we try to graft again
- The peer sees us as still being backing off, prune us as a punishment
- And the circle continues

arnetheduck:
> It seems a bit rude to ignore backoff requests from other peers, especially since we could enter a "graft infinite loop":

it seems reasonable perhaps to clamp it at a much higher value so as to prevent overflows and such, but not clamp it so aggressively

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -105,9 +105,13 @@ proc handleGraft*(g: GossipSub,
...
 
+    # Check backingOff
+    # Ignore BackoffSlackTime here, since this only for outbound activity
+    # and subtract a second time to avoid race conditions
+    # (peers may wait to graft us as the exact instant they're allowed to)
     if  g.backingOff
           .getOrDefault(topic)
-          .getOrDefault(peer.peerId) > Moment.now():
+          .getOrDefault(peer.peerId) - (BackoffSlackTime * 2).seconds > Moment.now():
       debug "attempt to graft a backingOff peer", peer, topic
```
arnetheduck, line 115: this log message reads like we're trying to do the grafting

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -105,9 +105,13 @@ proc handleGraft*(g: GossipSub,
 
       continue
 
+    # Check backingOff
+    # Ignore BackoffSlackTime here, since this only for outbound activity
```
arnetheduck, line 109: I don't really follow the reason we're subtracting BackoffSlackTime twice here..?

when we send a backoff time to someone, our clock starts ticking but theirs doesn't until they process the message - that means that if they follow the protocol reasonably, they will not send a graft until `backoffTime+roundrip` so there's already some "natural" slack built in - is there any harm in removing the slack time completely?

Menduist: Yes, but it could be them that sent the `backoff`
Let's say at the moment they sent the `backoff` they had 500ms roundtrip, we'll have 500ms more backoff than them.

But then, for some reason, the roundtrip improves to 100ms, and they graft us as early as they can. That would be 400ms too early for us.

The "natural" slack is jittery and not very big, that is the reason that the `BackoffSlackTime` exists:
- We make sure that we send outgoing graft 2 seconds after the time we're allowed, to avoid race condition on that side
- We make sure that we allow incoming graft 2 seconds before they're allowed, to avoid on the other side

It may be cleaner to store the backoff without the Slack time, and compute it each it time though
But for all the cases except this one, the backoff is used for outgoing and not incoming

arnetheduck: well, still: when we receive a prune, we store their graft time - when we send a prune, we store our graft time - it seems to me that the max of these two values should be used to decide when to send the next graft, no? or do we not receive any prune from them when we send a prune? can't remember.

Menduist:
> it seems reasonable perhaps to clamp it at a much higher value so as to prevent overflows and such, but not clamp it so aggressively

I've clamped to 1 day, seems like a reasonable maximum reasonable value

libp2p/protocols/pubsub/gossipsub/behavior.nim
Menduist: Yes, the peer pruning is sending his backoff, and the pruned peer doesn't have anything to say about it

dryajov (approved):

libp2p/protocols/pubsub/gossipsub/behavior.nim
dryajov: AFAIR,  pruning is send and forget and no ack is implied from the pruned peer.

dryajov:
@arnetheduck anything else to note here from your end? LGTM to me.

PR #612, opened 2021-07-28
Fixes failing peer store test in Nim 1.4.8

markspanbroek:
This test failed because the order in which the elements of a hashset are added to a seq is non-deterministic. It accidentally worked in Nim 1.2.6, but not in 1.4.8.

PR #617, opened 2021-08-21
fix: [testswitch e2e peer store] relies on hashset order

kaiserd:
testswitch can fail with the following message:

```
Check failed: storedInfo2.protos.toSeq() == switch1.peerInfo.protocols
    storedInfo2.protos.toSeq() was @["/test/proto/1.0.0", "/ipfs/id/1.0.0"]
    switch1.peerInfo.protocols was @["/ipfs/id/1.0.0", "/test/proto/1.0.0"]
```

The test assumes a specific order of items in a hashset.
This PR converts the sequence into a hashset fixing this issue locally; this makes the test complete successfully on my machine.

Is there a reason for PeerInfo managing supported protocols in a seq, while PeerStore manages it in a HashSet?

### DIFF
--- a/tests/testswitch.nim
+++ b/tests/testswitch.nim
@@ -901,5 +901,6 @@ suite "Switch":
       storedInfo1.addrs.toSeq() == switch2.peerInfo.addrs
       storedInfo2.addrs.toSeq() == switch1.peerInfo.addrs
 
-      storedInfo1.protos.toSeq() == switch2.peerInfo.protocols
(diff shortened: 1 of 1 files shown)


Menduist:
Hi, thanks for the report :)

Indeed this was a mistake that was fixed in #612
It's not yet on master, but you can use `unstable` directly without issue

PeerInfo is planned to be removed, so the discrepancy won't stay here long

PR #624, opened 2021-09-09
[WIP] CI: test multiple Nim compiler branches

stefantalpalaru:


dryajov:
I like it, but I will this run all compilers all the time? It already takes hours to run all the jobs as it is. We can probably restrict it to be ran on `master` and `unstable` only?

dryajov:
> > We can probably restrict it to be ran on master and unstable only?
> 
> Sure. Probably makes more sense as a separate workflow.

I think it would be ok if restricted to those two branches, it's a good balance between good coverage and not bringing everything to a halt. This branches aren't pushed to that often - `unstable` gets two three commits a week on average and `master` is something like one or two a month (according to our expected cadence, but has been much less so far).

Menduist:
Let's wait for chronos to work properly, then we'll fix the eventual internal issues here
Don't want to have red CI in the meantime, that would just be confusing

stefantalpalaru:
> Don't want to have red CI in the meantime

I'm already setting "continue-on-error" for the failing Nim branches, but this only works to mark those CI jobs as passing in the "Actions" tab. Everything else in the GitHub UI shows them as failing.

https://github.com/status-im/nim-libp2p/actions/workflows/multi_nim.yml

stefantalpalaru:
Converted to a daily CI job.

PR #625, opened 2021-09-21
Gossipsub: don't relay message to source

Menduist:


dryajov:
The daemon is failing on linux 386... Seems to be the same sporadic issue we've seen before. I wonder if we're simply timing out on connects? Perhaps we can bump/add some sleeps to allow connections to get established and see if this affects the failure rate?

Menduist:
Already tried to sleep, no effect unfortunately :/
When it fails, it succeeds to connect, but the peers don't see that they are both members of the topic.

I'll create a PR with my various test if you want to take a look

Also I noticed here that it failed some times on Gossipsub, and other times on pubsub. Fun

Menduist:
Uh, I guess there was no actual issues with it, just unlucky CI

Menduist:
Closed in favor of #648

PR #627, opened 2021-09-23
Validate pubsub subscriptions

Menduist:
Having an invalid `subscriptionValidator` can lead to [sneaky errors](https://github.com/status-im/nimbus-eth2/commit/9b41bb64da3d019aa6edea50ae111019096a3205), because if we subscribe to an invalid topic, we'll get grafted from time to time, but will be unable to graft peers.

This adds a warning when we subscribe to an invalid topic
EDIT: and also block subscription


libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -391,6 +391,10 @@ method onTopicSubscription*(p: PubSub, topic: string, subscribed: bool) {.base.}
   # Called when subscribe is called the first time for a topic or unsubscribe
   # removes the last handler
 
+  # Check that this is an allowed topic
+  if p.subscriptionValidator != nil and p.subscriptionValidator(topic) == false:
+    warn "Trying to subscribe to a topic not passing validation!", topic
```
arnetheduck, line 396: arguably we shouldn't add the topic to the subscription table at all in this case because it leads to further protocol violations and descoring when the others subscribe/graft

dryajov (approved):

Issue #641, opened 2021-10-29
Check peers pubsub sub -> graft -> prune -> unsub cycle


Menduist:
Currently, we allow unsubbed peers to graft us, and grafted peers to unsub. That's technically against the spec (should be strictly sub -> graft -> prune -> unsub)

Check the spec for clarity, other implementations to see what they do, and check that this doesn't lead to exploits in our implementation

PR #644, opened 2021-11-02
Score correctly on mesh peer unsub

Menduist:
Related to #641

When a peer unsub without pruning before, we currently keep it "in mesh" in the scoring mechanism, which create issues

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -182,15 +182,15 @@ method unsubscribePeer*(g: GossipSub, peer: PeerID) =
       if s[].len == 0:
         g.peersInIP.del(pubSubPeer.address.get())
 
-  for t in toSeq(g.gossipsub.keys):
-    g.gossipsub.removePeer(t, pubSubPeer)
-    # also try to remove from explicit table here
-    g.explicit.removePeer(t, pubSubPeer)
-
   for t in toSeq(g.mesh.keys):
     trace "pruning unsubscribing peer", pubSubPeer, score = pubSubPeer.score
     g.pruned(pubSubPeer, t)
     g.mesh.removePeer(t, pubSubPeer)
+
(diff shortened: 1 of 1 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -235,7 +235,12 @@ proc handleSubscribe*(g: GossipSub,
 
     # unsubscribe remote peer from the topic
     g.gossipsub.removePeer(topic, peer)
-    g.mesh.removePeer(topic, peer)
+
+    if g.mesh.hasPeer(topic, peer):
```
arnetheduck, line 239: since the proper order of events  is "remove from mesh" then "remove from gossipsub" I'd keep it that way in this code as well.. this gives "balance" when reading the code in general (sub -> graft -> prune -> unsub)

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -235,7 +235,12 @@ proc handleSubscribe*(g: GossipSub,
...
+      #against spec
+      g.mesh.removePeer(topic, peer)
+      g.pruned(peer, topic)
```
arnetheduck, line 242: this also makes me wonder: should we move `mesh.removePeer` to `pruned`?

Menduist: In some cases, pruned is called without removePeer (eg, when we unsub, we simply remove the topic from mesh)
Though this pruned is definitively error prone, has shown here

We could do it the other way around, add a custom event in the PeerTable which calls pruned for GS
But pubsub & gossipsub needs a refacto anyway

libp2p/protocols/pubsub/gossipsub.nim
Menduist: Done

dryajov (approved):

PR #645, opened 2021-11-08
Short agent fix

### LINKED ISSUE #635: `libp2p_peers_identity` only counts outgoing peers

Menduist:
closes #635 

The current "connection management" makes this operation a bit painful, because it's hard to pass data around layers (`libp2p_peers_identity` needs the shortAgent on "layer 0" (chronosstream), but the shortAgent is given on identify which is "layer 4" (chronosstream -> secureconn -> muxer -> identify), so we have to copy it around everywhere)

So dirty fix for now, hopefully will improve with better conn management

### DIFF
--- a/libp2p/muxers/mplex/mplex.nim
+++ b/libp2p/muxers/mplex/mplex.nim
@@ -97,6 +97,8 @@ proc newStreamInternal*(m: Mplex,
   result.peerId = m.connection.peerId
   result.observedAddr = m.connection.observedAddr
   result.transportDir = m.connection.transportDir
+  when defined(libp2p_agents_metrics):
+    result.shortAgent = m.connection.shortAgent
 
   trace "Creating new channel", m, channel = result, id, initiator, name
 
--- a/libp2p/upgrademngrs/muxedupgrade.nim
+++ b/libp2p/upgrademngrs/muxedupgrade.nim
@@ -173,6 +173,11 @@ proc muxerHandler(
 
   try:
     await self.identify(muxer)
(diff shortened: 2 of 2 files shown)


arnetheduck:
while we're here, it's a bit odd that we store the shortagent, really - it seems we should be storing the full agent and then shorten it for metrics? else it works poorly for situations like https://github.com/status-im/nimbus-eth2/blob/5c296c08b78ad4eae300ee05b21c1c743f144d7b/beacon_chain/rpc/rest_nimbus_api.nim#L107 where there's no cost to displaying the full string

arnetheduck (approved):

Menduist:
We are storing it.. in the peerstore :/

The big issue with Remove peer info was metrics & logs, where we try to access data that we wouldn't access normally in theses places (why would you need the peer agent in `ChronosStream.write` if it wasn't for metrics)

In the case you linked, it would make sense to have access to the peer store to retrieve that data, but for instance in `ChronosStream.write`, that doesn't make much sense to access the peerstore, so I had to copy the shortAgent around (since this field is only used for metrics, it's easier to compute it once)

shortAgent should really be considered internal

dryajov:
The `peerstore` is a perfect example of a "global" (😱) object. This has been solved in different ways at different times - real globals, singletons and later "injection" (with runtime dependency resolution). I'm not sure what is the best way to deal with this now and specially in our codebase. On the one hand we're going to keep running into this issue over and over (_almost_ guaranteed), on the other hand adding a global or a singleton is meh and I'm afraid of even googling for dependency injection in nim.

PR #647, opened 2021-11-09
Gossipsub peer exchange

Menduist:
Start of work on gossipsub PX

TODO:
- [x] Signed Envelope
- [x] Routing records
- [x] Add handler on gossipsub to allow application to connect to discovered peer
- [x] Send signed peer record as part of identify
- [x] Add received SPR to a new peer book
- [x] Send SPR from peer book on prune


Menduist:
I've run this for a little while on nimbus, and got precisely 0 SPR on prune from peers. They only send PeerId.
I guess we could hook up the peerid to ENR discovery, not sure it will actually help

libp2p/multiaddress.nim
```diff
@@ -1005,7 +1005,7 @@ proc `$`*(pat: MaPattern): string =
 proc write*(pb: var ProtoBuffer, field: int, value: MultiAddress) {.inline.} =
   write(pb, field, value.data.buffer)
 
-proc getField*(pb: var ProtoBuffer, field: int,
+proc getField*(pb: ProtoBuffer, field: int,
```
dryajov, line 1008: Why remove `var`,  no reason to pass by copy?

Menduist: All getField had their `var` remove a while ago, this was an old remainder (which was incompatible with the new getField without var)

dryajov: hmm, I can't recall the reason why it was removed? This would incur in potential copying overhead without any clear benefits?

libp2p/protocols/identify.nim
```diff
@@ -95,8 +105,11 @@ proc decodeMsg*(buf: seq[byte]): Option[IdentifyInfo] =
   let r5 = pb.getField(5, protoVersion)
   let r6 = pb.getField(6, agentVersion)
 
+  let r8 = pb.getSignedPeerRecord(8, signedPeerRecord)
+
   let res = r1.isOk() and r2.isOk() and r3.isOk() and
-            r4.isOk() and r5.isOk() and r6.isOk()
+            r4.isOk() and r5.isOk() and r6.isOk() and
+            r8.isOk()
```
dryajov, line 112: Aren't signed records optional? We shouldn't fail if it's absent.

libp2p/multiaddress.nim
arnetheduck: > pass by copy

per the same comment in the other branch, nim passes by reference, when that is appropriate, regardless of `var`/`let`

dryajov: Isn't that an artifact of the optimizer and it wont work in certain cases (not to mention that it's platform dependent and hard to predict). `var` makes it explicit and outside of the optimizers reach? Copying thousands of bytes doesn't seem to be very efficient?

arnetheduck: `var` makes explicit that the instance is mutable -it's unrelated to references - that a reference is used for either of `let` or `var` is an implementation detail (and one that nim _generally_ gets right, when it comes to passing variables). 

When the function doesn't need to mutate its parameter, we don't use `var`: https://status-im.github.io/nim-style-guide/language.vardecl.html - in part because `var` is infectious: callers must now also use `var`, where they could have used `let` or `const`.

libp2p/protocols/identify.nim
Menduist: `isOk` means "valid", to check if it's here you have to do `r8.get()` (which is done below)

Something missing will still be valid

This is a bit subtle, and why I added `getMandatoryField` a few weeks ago, to make this more readable

libp2p/multiaddress.nim
dryajov: I agree that it forces weird semantics on the caller, but it's also the only way of passing a stack object "by ref", whats the alternative?

libp2p/routing_record.nim
```diff
@@ -0,0 +1,125 @@
...
+  peerid,
+  protobuf/minprotobuf,
+  signed_envelope
+
+export peerid, multiaddress, signed_envelope
+
+## Constants relating to signed peer records
+const
+  EnvelopeDomain = multiCodec("libp2p-peer-record") # envelope domain as per RFC0002
+  EnvelopePayloadType= @[(byte) 0x03, (byte) 0x01] # payload_type for routing records as spec'ed in RFC0003
```
dryajov, line 28: This should be initialized with the appropriate multicodec, since it's already in our multicodec table.

libp2p/multiaddress.nim
arnetheduck: the alternative is to trust the compiler to do the right thing, which is what we do pretty much everywhere.

libp2p/routing_record.nim
Menduist: See in https://github.com/status-im/nim-libp2p/pull/649#discussion_r748341428
AFAICt there is no good way to extract only 0x03 0x01 (without padding) from our multiCodec implementation, and we have to use exactly that to be compatible with go-libp2p
Not sure if it's something we are missing, or if they should have padding in their value

libp2p/protocols/identify.nim
jm-clius: Perhaps in future we could also add a `hasField`, like some protobuf implementations?

Menduist: I think an ideal UX would be to just have `hasField` & `getMandatoryField` (which we could rename to `getField`)
But maybe that leads to double-parsing which can be slightly sub optimal?

Menduist: And both could potentially fail because of invalid pb

libp2p/routing_record.nim
dryajov: I don't think thats the case, `multiCodec()` will return the raw codec, not the length prefixed one. Codecs are not length prefixed, things like `Cid`, `multihash`, etc... are.

```nim
const
  CodeCodecs = initMultiCodecCodeTable()
  NameCodecs = initMultiCodecNameTable()

proc multiCodec*(name: string): MultiCodec {.compileTime.} =
  ## Generate MultiCodec from string ``name`` at compile time.
  let code = NameCodecs.getOrDefault(name, -1)
  doAssert(code != -1)
  MultiCodec(code)

proc multiCodec*(code: int): MultiCodec {.compileTime.} =
  ## Generate MultiCodec from integer ``code`` at compile time.
  let name = CodeCodecs.getOrDefault(code, "")
  doAssert(name != "")
  MultiCodec(code)
```

libp2p/protocols/identify.nim
Menduist: Moved to a new issue: https://github.com/status-im/nim-libp2p/issues/653

libp2p/routing_record.nim
```diff
@@ -127,7 +127,7 @@ proc getSignedPeerRecord*(pb: ProtoBuffer, field: int,
 proc getSignedPeerRecord*(envelopeBuf: seq[byte]): Result[PeerRecord, EnvelopeError] =
   let
     envelope = ? Envelope.decode(envelopeBuf, EnvelopeDomain)
-    spr = ? PeerRecord.decode(envelope.payload).orErr(EnvelopeInvalidProtobuf)
+    spr = ? PeerRecord.decode(envelope.payload).mapErr(x => EnvelopeInvalidProtobuf)
```
dryajov, line 130: Whats the difference with `orErr` in this instance? It looks like it's doing exactly the same in a more generic/less explicit way?

Menduist: I would use orErr, but it has only been available for 2 weeks in stew, and I would rather not force libp2p users to update stew (especially for something as easy to work-around)

Menduist:
Last few commits create a system for generic signed payload, can be reverted if we don't like them

libp2p/signed_envelope.nim
```diff
@@ -116,3 +118,49 @@ proc getField*(pb: ProtoBuffer, field: int,
       ok(true)
     else:
       err(ProtoError.IncorrectBlob)
+
+type
+  SignedPayload*[T] = object
+    # T needs to have .encode(), .decode(), .payloadType(), .domain()
```
emizzle, line 124: Could this constraint be enforced by typing `T`, eg `T: Payload` such that

```nim
type Payload = object

method encode*(self: Payload) {.base.} =
  doAssert(false, "Not implemented")

method decode*(self: Payload) {.base.} =
  doAssert(false, "Not implemented")

method payloadType*(self: Payload) {.base.} =
  doAssert(false, "Not implemented")

method domain*(self: Payload) {.base.} =
  doAssert(false, "Not implemented")
```
and signed typed would need to be derived from `Payload`, eg
```nim
type PeerRecord = object of Payload
  peerId*: PeerId
  seqNo*: uint64
  addresses*: seq[AddressInfo]
```

Menduist: The goal was to keep PeerRecord (or whatever T) "generic", ie not bound to the Envelope

I tried using a concept, works pretty well, though the error messages when the type does not match are not clear at all

cskiraly:
Concept looks like a good match, I'm just not sure about its implications on nim version compatibility. Since when we have concepts?

libp2p/signed_envelope.nim
arnetheduck: concepts are on the no-no list: unsupported and scheduled to be redone completely

Menduist: Ok, removed, worst case scenario it won't compile

emizzle: Could we instead wrap `PeerRecord` in a `Payload` type to keep `PeerRecord` from being modified, and creating a type that can be used in a constrained `SignedPayload` generic?

```nim
type Payload = object

method encode*(self: Payload) {.base.} =
  doAssert(false, "Not implemented")

method decode*(self: Payload) {.base.} =
  doAssert(false, "Not implemented")

method payloadType*(self: Payload) {.base.} =
  doAssert(false, "Not implemented")

method domain*(self: Payload) {.base.} =
  doAssert(false, "Not implemented")
```
```nim
type PeerRecordPayload = object of Payload
  record: PeerRecord
```
```nim
type
  SignedPayload*[T: Payload] = object
```

Menduist: Sure, but the goal of this system was to avoid boilerplate, so that would bring us back to square one.
Currently, if you try to use the SignedPayload on a non-compatible type, you will get errors like:
```nim
SignedPayload[int].decode(xxx)
```


```
testsigned_envelope.nim([line of the decode], 28) template/generic instantiation of `decode` from here
signed_envelope.nim(156, 43) Error: type mismatch: got <typedesc[int]>
but expected one of:
proc payloadDomain(T: typedesc[DummyPayload]): string
  first type mismatch at position: 1
  required type for T: typedesc[DummyPayload]
  but expression 'T' is of type: typedesc[int]
```

Which I think is not too bad

Worst case, we can add some static asserts to check that the type is compatible

But I don't think this is necessary: `SignedPayload` should not be used by end-users, instead they should use the types created by the library maintainer (eg, `SignedPeerRecord`), and the library maintainer should make sure that their type work in the CI or other

libp2p/protocols/identify.nim
```diff
@@ -107,29 +121,44 @@ proc decodeMsg*(buf: seq[byte]): Option[IdentifyInfo] =
       iinfo.protoVersion = some(protoVersion)
     if r6.get():
       iinfo.agentVersion = some(agentVersion)
+    if r8.get() and r1.get():
+      if iinfo.pubkey.get() == signedPeerRecord.envelope.publicKey:
+        iinfo.signedPeerRecord = some(signedPeerRecord.envelope)
     debug "decodeMsg: decoded identify", pubkey = ($pubkey).shortLog,
```
dryajov, line 127: This `debug` print looks more complex than everything else in this block, maybe we should factor it out to a template or something, because it definitely threw me off at first.

Menduist: Fixed

Menduist:
Interop tests are a bigger endeavor than I thought, merge this as-is and add them in a follow-up PR?
cc @dryajov

libp2p/routing_record.nim
```diff
@@ -87,39 +83,13 @@ proc init*(T: typedesc[PeerRecord],
...
-
-proc getSignedPeerRecord*(pb: ProtoBuffer, field: int,
-               value: var Envelope): ProtoResult[bool] {.
-     inline.} =
-  getField(pb, field, value, $EnvelopeDomain)
+proc checkValid*(spr: SignedPeerRecord): Result[void, EnvelopeError] =
+  if not spr.data.peerId.match(spr.envelope.publicKey):
+    err(EnvelopeInvalidSignature)
+  else:
+    ok()
```
emizzle, line 95: Since `spr.envelope` has a `signature` field, should we verify the `signature` agains the public key?

emizzle: Nevermind, I see this is being done in `decode`: https://github.com/status-im/nim-libp2p/pull/647/files#diff-070c4d16158c453c999e4cf95e766375df537a3c270eedc7e867ded8f9b597d4R67

Menduist: Yeah, the SPR is sub-ideal because it has redundant data, that need to be checked at every layer, which is error prone

There is some discussion at the libp2p layer to improve that

libp2p/signed_envelope.nim
```diff
@@ -116,3 +118,49 @@ proc getField*(pb: ProtoBuffer, field: int,
...
+               value: var SignedPayload[T]): ProtoResult[bool] {.
+     inline.} =
+  if not ? getField(pb, field, value.envelope, T.payloadDomain):
+    ok(false)
+  else:
+    mixin decode
+    value.data = ? T.decode(value.envelope.payload).mapErr(x => ProtoError.IncorrectBlob)
+    ok(true)
+
+proc decode*[T](
```
emizzle, line 150: We should probably include an `encode` for `SignedPayload[T]` as well?
```nim
proc encode*[T](msg: SignedPayload[T]): Result[seq[byte], CryptoError] =
  msg.envelope.encode()
```

I'm using this branch downstream in `nim-libp2p-dht` and needed the above proc to encode/decode protobufs.

dryajov (approved):

PR #648, opened 2021-11-10
Gossipsub: don't send to peers seen during validation

### LINKED ISSUE #626: Don't relay to any peer that has seen message already
https://github.com/status-im/nim-libp2p/pull/625 fixes message relay such that message is never sent to the source peer that obviously already has it.

However, because validation is asynchronous and can take time, it's quite possible that a message is received from multiple peers while it's being validated - sending the message to any of these peers is a waste.

Menduist:
Closes #626

This PRs adds a cache for currently validated message, and if a peer sends us a message which we are validating, we'll avoid forwarding him once validation is complete.

This also supersedes #625, because the source peer is put by default in the said cache.

cc @arnetheduck

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -314,8 +314,6 @@ method rpcHandler*(g: GossipSub,
       # onto the next message
       continue
 
-    g.validationSeen[msgIdSalted] = toHashSet([peer])
-
     # avoid processing messages we are not interested in
     if msg.topicIDs.allIt(it notin g.topics):
       debug "Dropping message of topic without subscription", msgId = shortLog(msgId), peer
@@ -337,6 +335,10 @@ method rpcHandler*(g: GossipSub,
 
     # g.anonymize needs no evaluation when receiving messages
     # as we have a "lax" policy and allow signed messages
+
+    # Be careful not to fill the validationSeen table
+    # (eg, pop everything you put in it)
+    g.validationSeen[msgIdSalted] = toHashSet([peer])
 
     let validation = await g.validate(msg)
 
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -309,8 +309,7 @@ method rpcHandler*(g: GossipSub,
       # score only if messages are not too old.
       g.rewardDelivered(peer, msg.topicIDs, false)
 
-      if msgIdSalted in g.validationSeen:
-        g.validationSeen[msgIdSalted].incl(peer)
+      g.validationSeen.withValue(msgIdSalted, seen): seen[].incl(peer)
 
       # onto the next message
       continue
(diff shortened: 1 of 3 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -331,7 +337,16 @@ method rpcHandler*(g: GossipSub,
     # g.anonymize needs no evaluation when receiving messages
     # as we have a "lax" policy and allow signed messages
 
+    # Be careful not to fill the validationSeen table
+    # (eg, pop everything you put in it)
+    g.validationSeen[msgIdSalted] = toHashSet([peer])
```
dryajov, line 342: wouldn't this override the previous entries?

Menduist: This code is only triggered for the first instance we receive of a message. As soon as we called `addSeen` above, every peer sending us this message will fail that condition (since we already saw it)

dryajov: How does this makes sense then, if we only do this once per message, then there is only ever going to be one peer per msg id? Seem a bit convoluted just to count one peer? What am I missing?

Menduist: Full code:
```nim
if g.addSeen(msgId):
  trace "Dropping already-seen message", msgId = shortLog(msgId), peer

  if msgIdSalted in g.validationSeen:
    g.validationSeen[msgIdSalted].incl(peer)
  
[...]
g.validationSeen[msgIdSalted] = toHashSet([peer])
```

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -331,7 +337,16 @@ method rpcHandler*(g: GossipSub,
...
+
     let validation = await g.validate(msg)
+
+    var seenPeers: HashSet[PubSubPeer]
+    discard g.validationSeen.pop(msgIdSalted, seenPeers)
+    libp2p_gossipsub_duplicate_during_validation.inc(seenPeers.len.float - 1)
```
arnetheduck, line 348:
```suggestion
    libp2p_gossipsub_duplicate_during_validation.inc(seenPeers.len.int64 - 1)
```

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -306,6 +309,9 @@ method rpcHandler*(g: GossipSub,
       # score only if messages are not too old.
       g.rewardDelivered(peer, msg.topicIDs, false)
 
+      if msgIdSalted in g.validationSeen:
```
arnetheduck, line 312:
```suggestion
      g.validationSeen.withValue(msgIdSalted, seen): seen[].incl(peer)
```

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -360,6 +375,10 @@ method rpcHandler*(g: GossipSub,
 
       await handleData(g, t, msg.data)
 
+    # Don't send it to source peer, or peers that
+    # sent it during validation
+    toSendPeers.excl(seenPeers)
```
arnetheduck, line 380: it probably makes sense to exclude the "first" sending peer separately: this impl looks correct but in the majority of cases will unnecessarily allocate a hash set - when we do the sending peer separately, an empty hash set init will not use any memory

arnetheduck (dismissed):

libp2p/protocols/pubsub/gossipsub.nim
Menduist: Oh, didn't know that an empty set was free :) - done

Menduist: Actually I'm not sure it's free, initHashSet creates a seq. Maybe there is a way to create one without init.
But maybe a hashset is not necessary anyway, as a wise men once said: "`O` notation only matters when `n` is big, and `n` is generally small"
We could probably switch a seq for so few elements

arnetheduck: use `... = HashSet[...]()` - that allocates no memory - I'm not sure about the seq - there's a risk here that if validation is slow, N can grow quite a bit

arnetheduck (approved):

arnetheduck:
let's get this to stable asap?

PR #649, opened 2021-11-11
Send signed peer record as part of identify

jm-clius:
Adding signed peer record as part of identify according to #647 and as an increment towards GossipSub Peer Exchange.

I have a couple of questions, though, mostly related to idiomatic use of `nim-libp2p`: 😃

1. Do we want to make the exchange of signed peer records configurable at all? Currently it will be populated if available on the peer info and it will always be available on the peer info, unless we encountered an error during creation.
2. It makes sense to me to cache the `signedPeerRecord` in the `PeerInfo` object, where we keep other local peer information. I'm unsure, however, where best to declare related `const`s such as the envelope `Domain` and `PayloadType`. For now they also exist in `peerinfo`. LMK if you think we should split these into a separate module.
3. I'm not too happy introducing the first use of `std/times` in `nim-libp2p` (`getTime()` for seq number), but `Moment.now()` only has meaning relative to another moment. Any other ideas?

### DIFF
--- a/libp2p/peerinfo.nim
+++ b/libp2p/peerinfo.nim
@@ -9,13 +9,18 @@
 
 {.push raises: [Defect].}
 
-import std/[options, sequtils, hashes]
-import pkg/[chronos, chronicles, stew/results]
-import peerid, multiaddress, crypto/crypto, errors
+import std/[options, sequtils, hashes, times]
+import pkg/[chronos, chronicles, stew/byteutils, stew/results]
+import peerid, multiaddress, crypto/crypto, signed_envelope, routing_record, errors
 
-export peerid, multiaddress, crypto, errors, results
+export peerid, multiaddress, crypto, signed_envelope, errors, results
 
 ## Our local peer info
+
+## Constants relating to signed peer records
+const
+  EnvelopePayloadType* = "/libp2p/routing-state-record".toBytes() # payload_type for routing records as spec'ed in RFC0003
+  EnvelopeDomain* = "libp2p-routing-record" # envelope domain as per RFC0002
 
 type
   PeerInfoError* = LPError
@@ -28,6 +33,25 @@ type
     agentVersion*: string
     privateKey*: PrivateKey
     publicKey*: PublicKey
+    signedPeerRecord*: Option[Envelope]
+
+proc createSignedPeerRecord(peerId: PeerID, addrs: seq[MultiAddress], key: PrivateKey): Result[Envelope, CryptoError] =
+  ## Creates a signed peer record for this peer:
+  ## a peer routing record according to https://github.com/libp2p/specs/blob/master/RFC/0003-routing-records.md
+  ## in a signed envelope according to https://github.com/libp2p/specs/blob/master/RFC/0002-signed-envelopes.md
+
+  # First create a peer record from the peer info
+  let peerRecord = PeerRecord.init(peerId,
+                                   getTime().toUnix().uint64, # This currently follows the recommended implementation, using unix epoch as seq no.
+                                   addrs)
+
+  # Wrap peer record in envelope and sign
+  let envelope = ? Envelope.init(key,
+                                 EnvelopePayloadType,
+                                 peerRecord.encode(),
+                                 EnvelopeDomain)
+  
+  ok(envelope)
 
 func shortLog*(p: PeerInfo): auto =
   (
@@ -52,14 +76,24 @@ proc new*(
       key.getPublicKey().tryGet()
     except CatchableError:
       raise newException(PeerInfoError, "invalid private key")
+  
+  let peerId = PeerID.init(key).tryGet()
+
+  # TODO: should using signed peer records be configurable?
+  let sprRes = createSignedPeerRecord(peerId, @addrs, key)
+  let spr = if sprRes.isOk:
+              some(sprRes.get())
+            else:
+              none(Envelope)
 
   let peerInfo = PeerInfo(
-    peerId: PeerID.init(key).tryGet(),
+    peerId: peerId,
     publicKey: pubkey,
     privateKey: key,
     protoVersion: protoVersion,
     agentVersion: agentVersion,
     addrs: @addrs,
-    protocols: @protocols)
+    protocols: @protocols,
+    signedPeerRecord: spr)
 
   return peerInfo
--- a/libp2p/protocols/identify.nim
+++ b/libp2p/protocols/identify.nim
@@ -44,6 +44,7 @@ type
     protoVersion*: Option[string]
     agentVersion*: Option[string]
     protos*: seq[string]
+    signedPeerRecord*: Option[Envelope]
 
   Identify* = ref object of LPProtocol
     peerInfo*: PeerInfo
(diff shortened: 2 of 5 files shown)


libp2p/peerinfo.nim
```diff
@@ -9,14 +9,19 @@
...
 
-export peerid, multiaddress, crypto, errors, results
+export peerid, multiaddress, crypto, signed_envelope, errors, results
 
 ## Our local peer info
 
+## Constants relating to signed peer records
+const
+  EnvelopePayloadType* = "/libp2p/routing-state-record".toBytes() # payload_type for routing records as spec'ed in RFC0003
+  EnvelopeDomain* = "libp2p-routing-record" # envelope domain as per RFC0002
```
Menduist, line 23: Actually, since the spec was done, it has been added to multicodec

`ibp2p-peer-record`
https://github.com/multiformats/multicodec/blob/master/table.csv

We should update our local list, and use this instead

libp2p/protocols/identify.nim
```diff
@@ -76,6 +77,14 @@ proc encodeMsg*(peerInfo: PeerInfo, observedAddr: Multiaddress): ProtoBuffer
   else:
     peerInfo.agentVersion
   result.write(6, agentVersion)
+
+  ## Optionally populate signedPeerRecord field.
+  ## See https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/identify/pb/identify.proto for reference.
```
Menduist, line 82: Could you put a permalink instead of master, always a pain when a link dies

libp2p/peerinfo.nim
```diff
@@ -28,6 +33,25 @@ type
...
+  ## a peer routing record according to https://github.com/libp2p/specs/blob/master/RFC/0003-routing-records.md
+  ## in a signed envelope according to https://github.com/libp2p/specs/blob/master/RFC/0002-signed-envelopes.md
+
+  # First create a peer record from the peer info
+  let peerRecord = PeerRecord.init(peerId,
+                                   getTime().toUnix().uint64, # This currently follows the recommended implementation, using unix epoch as seq no.
+                                   addrs)
+
+  # Wrap peer record in envelope and sign
+  let envelope = ? Envelope.init(key,
```
Menduist, line 49: I was thinking about this, maybe we should instead create a

Envelope.init(PrivateKey, PeerRecord)
&
Envelope.decodePeerRecord(seq[byte])
(or maybe getField is better, not sure) 

Inside peer_record.nim, and put the constants in it? (don't even need to export them)
wdyt?

libp2p/peerinfo.nim
```diff
@@ -52,14 +76,24 @@ proc new*(
       key.getPublicKey().tryGet()
     except CatchableError:
       raise newException(PeerInfoError, "invalid private key")
+  
+  let peerId = PeerID.init(key).tryGet()
+
+  # TODO: should using signed peer records be configurable?
```
Menduist, line 82: Related to https://github.com/status-im/nim-libp2p/issues/429
We can handle this as part of 429, leave as is for now

libp2p/peerinfo.nim
```diff
@@ -28,6 +33,25 @@ type
...
+    signedPeerRecord*: Option[Envelope]
+
+proc createSignedPeerRecord(peerId: PeerID, addrs: seq[MultiAddress], key: PrivateKey): Result[Envelope, CryptoError] =
+  ## Creates a signed peer record for this peer:
+  ## a peer routing record according to https://github.com/libp2p/specs/blob/master/RFC/0003-routing-records.md
+  ## in a signed envelope according to https://github.com/libp2p/specs/blob/master/RFC/0002-signed-envelopes.md
+
+  # First create a peer record from the peer info
+  let peerRecord = PeerRecord.init(peerId,
+                                   getTime().toUnix().uint64, # This currently follows the recommended implementation, using unix epoch as seq no.
```
Menduist, line 45: Are you sure Moment.now wouldn't do the trick? I though it was the same thing

libp2p/peerinfo.nim
jm-clius: Made `EnvelopeDomain` a multicodec in 0d5dfbb and added `libp2p-peer-record` to our list. Afaict we currently only use it in its stringified format.

libp2p/protocols/identify.nim
jm-clius: Yeah, thanks! Updated links in 0d5dfbb

libp2p/peerinfo.nim
jm-clius: I agree. Seems cleaner this way. Have gone ahead and moved all initialisation and decoding of signed peer record envelopes to routing_record.nim in 0d5dfbb.

This might not have been your exact intention, so LMK :)

libp2p/peerinfo.nim
jm-clius: Afaict there's no way to get a globally significant value from a `Moment`. `Moment`s are indeed captured using the epoch time, but that value is not exposed. The result is that `Moment`s are only meaningful as comparisons to other `Moment`s, e.g. to determine `Duration`s.

jm-clius: I guess we can initialise a `Moment` at value `0` and use the `Duration` up to `Moment.now()` as `seq` number?

libp2p/routing_record.nim
```diff
@@ -77,3 +84,42 @@ proc init*(T: typedesc[PeerRecord],
...
+
+  let envelope = ? Envelope.init(privateKey,
+                                 peerRecord)
+  
+  ok(envelope)
+
+proc getField*(pb: ProtoBuffer, field: int,
+               value: var Envelope): ProtoResult[bool] {.
+     inline.} =
+  getField(pb, field, value, $EnvelopeDomain)
```
Menduist, line 125: Actually the go version will use [0x03 0x01](https://github.com/libp2p/go-libp2p-core/blob/ef6e277df7422a37b1bbd92ac97f933d63d4cd97/peer/record.go#L28), so we're not compatible with it doing this. (Might be good to add SPR interop tests afterwards, I can handle it)

Not sure what's the best way to get "0x03 0x01" from the current MultiCodec implem, `MultiCodec.write` outputs a LP uint, so not good..
Worst case, we can also hardcode it

libp2p/routing_record.nim
```diff
@@ -77,3 +84,42 @@ proc init*(T: typedesc[PeerRecord],
...
+                                   getTime().toUnix().uint64, # This currently follows the recommended implementation, using unix epoch as seq no.
+                                   addresses)
+
+  let envelope = ? Envelope.init(privateKey,
+                                 peerRecord)
+  
+  ok(envelope)
+
+proc getField*(pb: ProtoBuffer, field: int,
+               value: var Envelope): ProtoResult[bool] {.
```
Menduist, line 123: We should make the RoutingRecord explicit, since there may be multiple envelope payload type in the future.

So getSignedPeerRecord?

libp2p/routing_record.nim
jm-clius: Right. Thanks. I should have checked - just assumed they were using the string representation.

Menduist (approved):
We can merge this and do the finishing touches in #647

jm-clius:
> We can merge this and do the finishing touches in #647

Cool, thanks! BTW, is the failing `CI / linux-amd64` a known issue? Seems to be related to the `go-libp2p-daemon`? Re-running CI now, but don't want to merge any new failures. :)

arnetheduck:
let's keep the default off as well, given that this might impact compatibility and needs consideration / documentation from a privacy / security point of view

Menduist:
Yes, known issue coming from go-libp2p-daemon

PR #650, opened 2021-11-12
Documentation system PoC

Menduist:
This is a PoC for a documentation idea I have.

The basic goal is to annotate every "public/stable" proc/object/etc with a pragma, and only generate doc for the public/stable stuff.

This allows:
- As a libp2p developer, looking at code, know what can you break without it being a breaking change
- For users to know easily if they use public procs or not (just look at the doc _or_ code)
- Still use nimdoc

A rough PoC is included in this PR. Since we can't generate docs for libp2p ATM, I added a stub nim example.

Doc looks like this:
image

Clicking the "Show internal procedures" reveals everything (for libp2p developers, or people not afraid of using internal stuff):
image

We could improve this a lot, eg do a slider like Dark mode for "internal". The current technique is a bit ugly, but hey, PoC.

Putting this here to get some feedback on the pragma technique


Menduist:
Closed in favor of #716

PR #651, opened 2021-11-15
Unstable

Menduist:


libp2p/protocols/secure/secio.nim
```diff
@@ -263,7 +263,7 @@ proc newSecioConn(conn: Connection,
   ## cipher algorithm ``cipher``, stretched keys ``secrets`` and order
   ## ``order``.
 
-  result = SecioConn.init(conn, conn.peerId, conn.observedAddr)
+  result = SecioConn.new(conn, conn.peerId, conn.observedAddr)
```
dryajov, line 266: secio should probably be removed for good.

dryajov (approved):
LGTM

libp2p/protobuf/minprotobuf.nim
```diff
@@ -657,6 +576,17 @@ proc getField*(pb: ProtoBuffer, field: int,
   else:
     err(res.error)
 
+proc getRequiredField*[T](pb: ProtoBuffer, field: int,
+               output: var T): ProtoResult[void] {.inline.} =
```
dryajov, line 580: Now that I think about it, it doesn't really make much sense to have a `Result` type and use a variable for output. `Result` already allows you to express both an error condition and a result value when everything went well. 

IMO a `void` result **only** makes sense when nothing is being returned from the function, in this (and similar) case, there is a return value and it should have been passed as part of the result. This resembles C and it's not very idiomatic to Nim, IMO.

Menduist: Good point

I mostly did that to stay consistent with getField, but also because the generic matching don't work solely on result type, so you would have to do
`? getMandatoryField[PeerId](5)`

(PeerId explicitely)
Not sure if it's a good think or a bad thing

dryajov: I think if you do `let peerId: PeerID = getMandatoryField(5)`, the compiler deduces the generic type? Usually Nim is pretty good at inferring this.

Menduist: I don't think type inferring works on the return type, I'll check to be sure

Menduist: Moved PB to a new issue for better follow up: https://github.com/status-im/nim-libp2p/issues/653

PR #652, opened 2021-11-15
Chronos strict exception tracking

Menduist:
closes https://github.com/status-im/nim-libp2p/issues/637

Enable strict exception tracking for chronos, and fixes where it broke

I may update the DNS library instead of handling it here

### DIFF
--- a/libp2p/daemon/daemonapi.nim
+++ b/libp2p/daemon/daemonapi.nim
@@ -761,7 +761,7 @@ proc newDaemonApi*(flags: set[P2PDaemonFlags] = {},
     except CatchableError as exc:
       raise exc
     except Exception as exc:
-      raise newException(DaemonLocalError, exc.msg)
+      raise raiseAssert exc.msg
   # Waiting until daemon will not be bound to control socket.
   while true:
     if not api.process.running():
--- a/libp2p/daemon/daemonapi.nim
+++ b/libp2p/daemon/daemonapi.nim
@@ -150,10 +150,10 @@ type
     key*: PublicKey
 
   P2PStreamCallback* = proc(api: DaemonAPI,
-                            stream: P2PStream): Future[void] {.gcsafe.}
+                            stream: P2PStream): Future[void] {.gcsafe, raises: [Defect, CatchableError].}
   P2PPubSubCallback* = proc(api: DaemonAPI,
                             ticket: PubsubTicket,
-                            message: PubSubMessage): Future[bool] {.gcsafe.}
+                            message: PubSubMessage): Future[bool] {.gcsafe, raises: [Defect, CatchableError].}
 
   DaemonError* = object of LPError
   DaemonRemoteError* = object of DaemonError
@@ -755,7 +755,13 @@ proc newDaemonApi*(flags: set[P2PDaemonFlags] = {},
 
   # Starting daemon process
   # echo "Starting ", cmd, " ", args.join(" ")
-  api.process = startProcess(cmd, "", args, env, {poParentStreams})
+  api.process = 
+    try:
+      startProcess(cmd, "", args, env, {poParentStreams})
+    except CatchableError as exc:
+      raise exc
+    except Exception as exc:
+      raise newException(DaemonLocalError, exc.msg)
   # Waiting until daemon will not be bound to control socket.
   while true:
     if not api.process.running():
@@ -900,7 +906,7 @@ proc openStream*(api: DaemonAPI, peer: PeerID,
         stream.flags.incl(Outbound)
         stream.transp = transp
         result = stream
-  except Exception as exc:
+  except CatchableError as exc:
     await api.closeConnection(transp)
     raise exc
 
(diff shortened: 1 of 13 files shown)


libp2p/daemon/daemonapi.nim
```diff
@@ -755,7 +755,13 @@ proc newDaemonApi*(flags: set[P2PDaemonFlags] = {},
...
   # Starting daemon process
   # echo "Starting ", cmd, " ", args.join(" ")
-  api.process = startProcess(cmd, "", args, env, {poParentStreams})
+  api.process = 
+    try:
+      startProcess(cmd, "", args, env, {poParentStreams})
+    except CatchableError as exc:
+      raise exc
+    except Exception as exc:
+      raise newException(DaemonLocalError, exc.msg)
```
dryajov, line 764: This should be mapped to a `Defect` instead. We have no idea what this `Exception` is and the safest thing to do is to allow the process to exit here.

libp2p/nameresolving/dnsresolver.nim
```diff
@@ -79,6 +79,10 @@ proc getDnsResponse(
     dataStream.writeData(addr rawResponse[0], rawResponse.len)
     dataStream.setPosition(0)
     return parseResponse(dataStream)
+  except CatchableError as exc:
+    raise exc
+  except Exception as exc: # DNS library can throw anything
+    raise newException(CatchableError, "DNS query error: " & exc.msg)
```
dryajov, line 85: This should be `Defect`.

libp2p/nameresolving/dnsresolver.nim
```diff
@@ -110,7 +114,10 @@ method resolveIp*(
       try:
         let resp = await fut
         for answer in resp.answers:
-          resolvedAddresses.incl(answer.toString())
+          resolvedAddresses.incl(
+            try: answer.toString()
+            except Exception as exc: raise newException(CatchableError, exc.msg)
```
dryajov, line 119: Same as in previous comment. Rule of thumb is usually to remap to `Defect` if not able to handle in place and specially when remapping `Exception` - there is simply no way of knowing what type of error we're handling and if it's possible to recover from it at all.

libp2p/nameresolving/dnsresolver.nim
```diff
@@ -110,7 +114,10 @@ method resolveIp*(
...
+          )
       except CancelledError as e:
         raise e
       except ValueError as e:
```
dryajov, line 123: We shouldn't handle `ValueError` as it's a `Defect` type.

libp2p/nameresolving/dnsresolver.nim
```diff
@@ -140,9 +147,14 @@ method resolveTxt*(
...
     let server = self.nameservers[0]
     try:
-      let response = await getDnsResponse(server, address, TXT)
-      trace "Got TXT response", server = $server, answer=response.answers.mapIt(it.toString())
-      return response.answers.mapIt(it.toString())
+      let
+        response = await getDnsResponse(server, address, TXT)
+        answers = response.answers.mapIt(
+            try: it.toString()
+            except Exception as exc: raise newException(CatchableError, exc.msg)
```
dryajov, line 154: same, needs to be `Defect`

libp2p/nameresolving/dnsresolver.nim
Menduist: `ValueError* = object of CatchableError`

Pretty sure I caught all the Defect catching attempt as part of Various Cleanups part 1

dryajov: I think it's slated to be a defect, I remember seeing this somewhere - @arnetheduck might know as well.

arnetheduck: It's a catchable, and that won't change.

libp2p/nameresolving/dnsresolver.nim
arnetheduck: fwiw, we usually type this as `raiseAssert exc.msg` - also, unless the dns library actuall_ raises exceptions that are not `CatchableError`, it's not a real problem

the most important thing here is that dns resolution should generally not happen automatically - that introduces a number of security issues and considerations that need to be taken into account in addition to language details like this - ie resolution should always be explicitly enabled to happen

Menduist: You have to pass a nameresolver to the switch for it to happen

libp2p/nameresolving/dnsresolver.nim
arnetheduck: the way to write these is:
```
try: ...
except CatcahbleError as exc: raise exc
except Exception as exc: raiseAssert exc.msg
```
unless you know that dns raises something that inherits from Exception, in which case it gets more complicated:
```
try: ...
except CatcahbleError as exc: raise exc
except Defect as exc: raise exc
except Exception as exc: raise (ref CatchableError)(msg: exc.msg, parent: exc)
```

arnetheduck (approved):

PR #657, opened 2021-11-24
Add SPR from identify to new peer book

jm-clius:
Adding received signed peer record from `identify` to new peer book according to #647 and as an increment towards GossipSub Peer Exchange.

(Note: I notice [the go implementation](https://github.com/libp2p/go-libp2p-peerstore/blob/master/pstoreds/addr_book.go#L316) stores the marshalled SPR. Perhaps we should consider doing the same, if it's significantly easier to forward as such on PRUNE? I prefer the current way (with `Envelope` object) though.)

### DIFF
--- a/libp2p/peerstore.nim
+++ b/libp2p/peerstore.nim
@@ -14,6 +14,7 @@ import
   ./crypto/crypto,
   ./protocols/identify,
   ./peerid, ./peerinfo,
+  ./routing_record,
   ./multiaddress
 
 type
@@ -53,6 +54,8 @@ type
 
     agentBook*: PeerBook[string]
     protoVersionBook*: PeerBook[string]
+
+    signedPeerRecordBook*: PeerBook[Envelope]
   
 ## Constructs a new PeerStore with metadata of type M
 proc new*(T: type PeerStore): PeerStore =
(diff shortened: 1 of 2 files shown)


libp2p/peerstore.nim
```diff
@@ -53,6 +54,8 @@ type
 
     agentBook*: PeerBook[string]
     protoVersionBook*: PeerBook[string]
+
+    signedPeerRecordBook*: PeerBook[Envelope]
```
Menduist, line 58: Maybe it would make sense to create a specialized PeerBook that check the seqNo before storing?
Ofc we can trust the peer during identify, but if we get SPR from elsewhere it would be nice to just check here, wdyt ?

Menduist:
IMO marshaling is pretty cheap, it's just some protobuffing, we don't have to cache it

libp2p/peerstore.nim
jm-clius: Ah, do you mean to just check that the `seq` no increases monotonically? I didn't initially think it necessary for info we receive via `identify`, as we implicitly trust that information as being part of the negotiation.

I understand it may be necessary if we get the SPR "from elsewhere". In this case, wouldn't we want to check the signature as well? Also, I imagined that it's only in `PRUNE` that we'll receive SPR from another source, in which case it will be up to the higher layer application to verify the records and decide whether to connect or not? I may be missing something obvious here. :)

Menduist: I though a bit more about this, I think a proc in the PeerStore that checks the seqNo and update the relevant books when we get a new SPR would be best. I can add it as part of gspx

Menduist (approved):

PR #659, opened 2021-11-25
fast path for writes

arnetheduck:
avoids several copies of the various message buffers being kept alive
for the lifetime of the future

### DIFF
--- a/libp2p/muxers/mplex/lpchannel.nim
+++ b/libp2p/muxers/mplex/lpchannel.nim
@@ -175,9 +175,9 @@ method readOnce*(s: LPChannel,
     await s.reset()
     raise exc
 
-method write*(s: LPChannel, msg: seq[byte]): Future[void] {.async.} =
-  ## Write to mplex channel - there may be up to MaxWrite concurrent writes
-  ## pending after which the peer is disconnected
+proc prepareWrite(s: LPChannel, msg: seq[byte]): Future[void] {.async.} =
+  # prepareWrite is the slow path of writing a message - see conditions in
+  # write
   if s.closedLocal or s.conn.closed:
     raise newLPStreamClosedError()
 
@@ -191,19 +191,20 @@ method write*(s: LPChannel, msg: seq[byte]): Future[void] {.async.} =
     await s.conn.close()
     return
 
-  s.writes += 1
+  if not s.isOpen:
+    await s.open()
+
+  await s.conn.writeMsg(s.id, s.msgCode, msg)
+
+proc completeWrite(
+    s: LPChannel, fut: Future[void], msgLen: int): Future[void] {.async.} =
   try:
-    if not s.isOpen:
-      await s.open()
+    s.writes += 1
 
-    # writes should happen in sequence
-    trace "write msg", s, conn = s.conn, len = msg.len
-
-    await s.conn.writeMsg(s.id, s.msgCode, msg)
-
+    await fut
     when defined(libp2p_network_protocols_metrics):
       if s.tag.len > 0:
-        libp2p_protocols_bytes.inc(msg.len.int64, labelValues=[s.tag, "out"])
+        libp2p_protocols_bytes.inc(msgLen.int64, labelValues=[s.tag, "out"])
 
     s.activity = true
   except CatchableError as exc:
@@ -213,6 +214,24 @@ method write*(s: LPChannel, msg: seq[byte]): Future[void] {.async.} =
     raise exc
   finally:
     s.writes -= 1
+
+method write*(s: LPChannel, msg: seq[byte]): Future[void] =
+  ## Write to mplex channel - there may be up to MaxWrite concurrent writes
+  ## pending after which the peer is disconnected
+
+  let
+    closed = s.closedLocal or s.conn.closed
+
+  let fut =
+    if (not closed) and msg.len > 0 and s.writes < MaxWrites and s.isOpen:
+      # Fast path: Avoid a copy of msg being kept in the closure created by
+      # `{.async.}` as this drives up memory usage - the conditions are laid out
+      # in prepareWrite
+      s.conn.writeMsg(s.id, s.msgCode, msg)
+    else:
+      prepareWrite(s, msg)
+
+  s.completeWrite(fut, msg.len)
 
 proc init*(
   L: type LPChannel,
--- a/libp2p/protocols/secure/noise.nim
+++ b/libp2p/protocols/secure/noise.nim
@@ -460,10 +460,8 @@ proc encryptFrame(
 
   cipherFrame[2 + src.len()..<cipherFrame.len] = tag
 
-method write*(sconn: NoiseConnection, message: seq[byte]): Future[void] {.async.} =
-  if message.len == 0:
-    return
-
+method write*(sconn: NoiseConnection, message: seq[byte]): Future[void] =
+  # Fast path: `{.async.}` would introduce a copy of `message`
   const FramingSize = 2 + sizeof(ChaChaPolyTag)
 
   let
@@ -479,10 +477,16 @@ method write*(sconn: NoiseConnection, message: seq[byte]): Future[void] {.async.
     let
       chunkSize = min(MaxPlainSize, left)
 
-    encryptFrame(
-      sconn,
-      cipherFrames.toOpenArray(woffset, woffset + chunkSize + FramingSize - 1),
-      message.toOpenArray(offset, offset + chunkSize - 1))
+    try:
+      encryptFrame(
+        sconn,
+        cipherFrames.toOpenArray(woffset, woffset + chunkSize + FramingSize - 1),
+        message.toOpenArray(offset, offset + chunkSize - 1))
+    except NoiseNonceMaxError as exc:
+      debug "Noise nonce exceeded"
+      let fut = newFuture[void]("noise.write.nonce")
+      fut.fail(exc)
+      return fut
 
     when defined(libp2p_dump):
       dumpMessage(
@@ -492,9 +496,12 @@ method write*(sconn: NoiseConnection, message: seq[byte]): Future[void] {.async.
     left = left - chunkSize
     offset += chunkSize
     woffset += chunkSize + FramingSize
-    sconn.activity = true
 
-  await sconn.stream.write(cipherFrames)
+  sconn.activity = true
+
+  # Write all `cipherFrames` in a single write, to avoid interleaving /
+  # sequencing issues
+  sconn.stream.write(cipherFrames)
 
 method handshake*(p: Noise, conn: Connection, initiator: bool): Future[SecureConn] {.async.} =
   trace "Starting Noise handshake", conn, initiator
--- a/libp2p/stream/chronosstream.nim
+++ b/libp2p/stream/chronosstream.nim
@@ -94,7 +94,6 @@ when defined(libp2p_agents_metrics):
 method readOnce*(s: ChronosStream, pbytes: pointer, nbytes: int): Future[int] {.async.} =
   if s.atEof:
     raise newLPStreamEOFError()
-
   withExceptions:
     result = await s.client.readOnce(pbytes, nbytes)
     s.activity = true # reset activity flag
@@ -104,31 +103,36 @@ method readOnce*(s: ChronosStream, pbytes: pointer, nbytes: int): Future[int] {.
       if s.tracked:
         libp2p_peers_traffic_read.inc(nbytes.int64, labelValues = [s.shortAgent])
 
-method write*(s: ChronosStream, msg: seq[byte]) {.async.} =
-  if s.closed:
-    raise newLPStreamClosedError()
-
-  if msg.len == 0:
-    return
-
+proc completeWrite(
+    s: ChronosStream, fut: Future[int], msgLen: int): Future[void] {.async.} =
   withExceptions:
     # StreamTransport will only return written < msg.len on fatal failures where
     # further writing is not possible - in such cases, we'll raise here,
     # since we don't return partial writes lengths
-    var written = await s.client.write(msg)
+    var written = await fut
 
-    if written < msg.len:
+    if written < msgLen:
       raise (ref LPStreamClosedError)(msg: "Write couldn't finish writing")
 
     s.activity = true # reset activity flag
-    libp2p_network_bytes.inc(msg.len.int64, labelValues = ["out"])
+    libp2p_network_bytes.inc(msgLen.int64, labelValues = ["out"])
     when defined(libp2p_agents_metrics):
       s.trackPeerIdentity()
       if s.tracked:
-        libp2p_peers_traffic_write.inc(msg.len.int64, labelValues = [s.shortAgent])
+        libp2p_peers_traffic_write.inc(msgLen.int64, labelValues = [s.shortAgent])
+
+method write*(s: ChronosStream, msg: seq[byte]): Future[void] =
+  # Avoid a copy of msg being kept in the closure created by `{.async.}` as this
+  # drives up memory usage
+  if s.closed:
+    let fut = newFuture[void]("chronosstream.write.closed")
+    fut.fail(newLPStreamClosedError())
+    return fut
+
+  s.completeWrite(s.client.write(msg), msg.len)
 
 method closed*(s: ChronosStream): bool =
-  result = s.client.closed
+  s.client.closed
 
 method atEof*(s: ChronosStream): bool =
   s.client.atEof()


dryajov:
LGTM, but do we have any measurements to show the improvements?

Menduist:
Can we hold off on this while I investigate the cycle breaking?
If we manage to break the cycle, it will encompass theses improvement, since the future will be unallocated instantly

arnetheduck:
> it will encompass theses improvement

no, it will not: when using `async`, a closure is allocated, and then the data is copied into the closure. even if you break cycles, the copy is still made (to simplify reasoning about lifetimes) _and_ the data stays alive throughout all `await` steps - for example, in `await a(f); await b(g)` both `f` and `g` are first copied into something that effectively looks like `(ref tuple)(f, g)` and kept alive to the end of the second `await`, while `f` could be released _before_ the first await and _not_ copied, without introducing any memory safety issues.

you can, by just reasoning about the code, immediately understand that the code at the very least removes 3 unnecessary copies of the message (not only the memory for the copies, but the act of copying them). this happens independently of other improvements which release memory more aggressively (it's better to not allocate than to allocate and deallocate quickly).

> measurements

https://metrics.status.im/d/pgeNfj2Wz23/nimbus-fleet-testnets?orgId=1&var-instance=metal-01.he-eu-hel1.nimbus.prater&var-container=beacon-node-prater-unstable-01&from=1637762622601&to=1637789016132

you can see how the memory profile is more "calm" - but above all you can see the improved quality of the mesh as a result of the improved performance - the server is running right at the edge of performance.

Menduist:
> it's better to not allocate than to allocate and deallocate quickly

Sure, but now we see a measurable difference because of the memory issue in chronos, but we may not see it in the future.
Theses fast path decrease readability and are more error prone, so we must be sure that they are "worth it"

Until now they clearly were, but maybe after https://github.com/status-im/nim-chronos/pull/243 we won't be able to see a measurable difference with and without them, and so they won't be worth it anymore

We can merge this, and when the chronos fix is merged, try to remove every fast path in libp2p, and see if we can still see a measurable difference. And if not, remove every fast path. And if we still see a difference, keep it

arnetheduck:
>  measurable difference

keep in mind also that the messages we're sending now are 200-300 bytes (attestations) - soon, we'll be seeing 10mb blocks instead.

because of amplification factor and all writes passing through here, this will remain a hotspot - I doubt we can go back to all these copies, neither for future eth2, nor for waku - on the contrary, we'll probably need to introduce more hotspot fixes (when decoding the rpcmsg protobuf for example)

dryajov:
I don't think there is anything particularly wrong with splitting this out, but in most cases, we can keep the nested proc without turning it into a closure by simply avoiding capturing outer scope, just pass it as arguments:

```nim
proc outer(a, b): int =
  proc inner(a, b: int): int =
    a + b

  inner(a, b)
```

arnetheduck:
> avoiding capturing outer scope,

this is easy in theory but hard in practice: it's easy to "accidentally" capture a variable (earlier versions of this PR did that, with poor outcomes) - it's possible to annotate inner functions with `{.nimcall.}` instead of putting them outside which prevents accidental captures, but this requires more magic nim knowledge about exotic pragmas - I'm fine either way with a slight preference to use fewer fancy features.

dryajov:
> > avoiding capturing outer scope,
> 
> this is easy in theory but hard in practice: it's easy to "accidentally" capture a variable (earlier versions of this PR did that, with poor outcomes) - it's possible to annotate inner functions with `{.nimcall.}` instead of putting them outside which prevents accidental captures, but this requires more magic nim knowledge about exotic pragmas - I'm fine either way with a slight preference to use fewer fancy features.

Depends on each particular use case, I agree that in most cases we can avoid inner calls without affecting readability but there are specifics where an inner call might be appropriate. But in general I agree, most are just a product of habit/laziness.

Menduist (dismissed):
> keep in mind also that the messages we're sending now are 200-300 bytes (attestations) - soon, we'll be seeing 10mb blocks instead.

That's true

Menduist:
I was thinking about theses fast paths, and I had an idea

Most of the time, theses functions will be called by doing
```nim
let buf = generateData()
await stuff.write(buf)
```

So, we don't really need to copy the buffer around. We could do something like:
```nim
method write_unsafe*(s: LPStream, msg: ptr UncheckedArray[byte], size: int): Future[void] {.base.} =
  doAssert(false, "not implemented!")

proc write*(s: LPStream, msg: seq[byte]): {.async.} =
  # Maybe there is better semantics to convert a seq to a UncheckedArray, not really pretty here
  # But having a UncheckedArray instead of a ptr allows to call toOpenArray on it
  await s.write_unsafe(cast[ptr UncheckedArray[byte]](unsafeAddr msg[0]), msg.len)
```

Which opens the possibility to call the `write_unsafe` when you're sure that the buffer will stay valid for the duration of the future, and avoids copying even if the queue is not empty

wdyt?

zah:
The API of the FastStreams offers an alternative that makes it easy to write code such that every byte of the outgoing message is written exactly once (with zero copies and very few memory allocations). If you want to make Chronos and LibP2P leaner and faster, it would be useful to consider how some of the FastStreams ideas can be incorporated. The README of FastStreams and especially the following two articles provide some very accessible explanation for the benefits of the API design:

https://github.com/status-im/nim-faststreams
https://blog.marcgravell.com/2018/07/pipe-dreams-part-1.html
https://blog.marcgravell.com/2018/07/pipe-dreams-part-2.html

Menduist:
> FastStreams

That looks interesting, but it would imply to completely rewrite this part of libp2p, which is a big undertaking


We brainstormed a bit with @dryajov, building up on my last proposal
Instead of using `UncheckedArray`, we can actually use `ref seq`, which would avoid copies while keeping it memory safe

Here are two proposals in pseudocode:
```nim
proc writeInner(lb: ref seq[byte]) {.async.} =
  # do stuff
  await send(lb)
  
proc writeLp(lb: ref seq[byte]) {.async.} =
  # do stuff
  await writeInner(lb)

# to convert a seq to a ref seq:
let rSeq = new(seq[byte])
rSeq[] = @[12.byte, 12]
writeLp(sSeq)
```
As soon as assigned our rSeq, no copy is made in the inner layers. So this at least removes every copy in libp2p, and the consuming application is free to use a `ref seq` directly to remove even more copies


Second proposal:
```nim
type LBuffer[T] = ref seq[T]

converter toLBuffer[T](s: seq[T]): LBuffer[T] =
  result = LBuffer[T](new(seq[T]))
  result[] = s

proc writeInner(lb: LBuffer[int]) {.async.} =
  # do stuff
  await send(lb)
 
proc writeLp(lb: LBuffer[int]) {.async.} =
  # do stuff
  await writeInner(lb)

writeLp(@[12, 12])
```
Builds on the same idea, but create another type for our ref seq, and make a converter for it.
This allow us to use `LBuffer` everywhere in libp2p, and the user don't have to know about it, since by default, the converter will be used.

If he wants to optimize for seq copies, he is free to also use LBuffer in his application, instead of seq



I think either of theses proposals are very nice, and remove copies outright for every scenario, while keeping a pretty good readability. wdyt?

arnetheduck:
> you're sure that the buffer will stay valid 


`chronos` offers this functionality, but so far we've opted to stay within the memory-safe bounds of the language as much as possible, as this makes it easier to reason about the correctness of the code - by taking a copy and matching the lifetime of the future to that of the buffer, we avoid a class of bugs and issues "by default" - then we can selectively introduce optimizations where they actually matter, like here in gossipsub.

> `faststreams`,

like its chronos counterpart, also steps outside of the memory-safe world, even though it offers its own .. utilities to manage the problem. introducing it amounts to a paradigm shift and a rewrite.

> `ref seq`

this opens another class of bugs, namely that of concurrent mutation, which actually is very similar from a risk point of view to the "manual" buffer management with pointers and so on - the language unfortunately doesn't have a way to express immutability through references which is unfortunate - it's a very common pattern to have data that _can_ be shared in a read-only way after its construction.

even with `ref seq`, one problem that `faststreams` does solve is that of decorating buffers with headers - after this PR, this is the main source of "avoidable" copies (compression and encryption for example are not "reasonably" avoidable) - outside of nim, the "normal" way of solving that is through "gather writes", aka `writev`, or the ability to pass a sequence of buffers to the "next" layer such that when a copy "must" be made (for encryption for example), several "input" buffers can be compacted into one. the "header" support in faststreams could be seen as a special case of this, because you still need to "preallocate" header space in faststreams - ie it's not a general solution.

Menduist:
> concurrent mutation

Yes, when I said "keeping it memory safe", I was talking about out of bounds R/Ws, obviously you still need to be sure that the buffer stays valid for the duration of the child future
If the caller cannot guarantee that, he'll have to use the non-ref version, which will copy once into a ref seq, and then for internal layers there will be no copies

> buffers with headers

_Technically_ we can already do that, by calling write twice. Since every inner layer is a bufferstream, it can handle it

Issue is, our code is very packet based right now, so each buffer would be treated separately. But we can add some buffering, or even just something like I proposed here https://github.com/status-im/nim-libp2p/pull/664#discussion_r764733123, and they will get re-combined correctly

arnetheduck:
> Issue is, our code is very packet based right now

well, it's not, in the sense that it doesn't give guarantees about packets - the libp2p protocol has a packet-based implementation, true, because packets make sense for performance, more or less, and our code follows the protocol closely so as not to introduce mechanical disharmony that results in inefficiency.. 

> buffer stays valid for the duration

well, the thing with the current construct is that we let the language provide this guarantee - faststreams breaks the language guarantees and offer a "library guarantee" instead that is weaker, and the next step is to introduce magic write functions that let you drop the language guarantees on a case-by-case basis - every such weakening is a risk so you'd want a good return on them - I'm not sure the returns are good enough to warrant work in this direction, in the nim language as it looks today, but.. _maybe_? `rust` for example tries to "regain" some of the performance loss caused by abstraction and safety guarantees by employing lifetime, ownership and mutability guarantees that don't exist in Nim - what you're suggesting certainly is doable, but it's also compromises safety in one way or another.

On the bright side, some of these guarantees will become better in future Nim versions - and some could actually be done in chronos "systemically" - two in particular: what I did in this branch was manually unroll the `async` functions into a sync part and an async part - chronos could be smart about this - the second thing I did was what could best be described as a tail call optimization that skips the last await in `return await xxx` - this is also something we could accomplish systemically at the chronos level without giving up any safety.

arnetheduck:
> Most of the time, theses functions will be called by doing

one thing that stands in the way here is that many async functions start with a conditional or error check (`if closed: raise`) - this messes up the control flow because the generation and sending of the message is the "second" step in the async state machine - the first and the last steps can usually be optimized away easily, but the middle ones are uglier / harder. In this PR, they're handled by allocating the futures manually and redundantly.

arnetheduck:
ready to make a move on this, @Menduist?

Menduist:
You can merge if you want (maybe in unstable, we'll merge to master soon enough anyway), but I still think that something like I outlined would be a better solution

I did a prototype here: https://github.com/status-im/nim-libp2p/compare/unstable...sharedbuffer?expand=1

It creates a SharedBuffer type, which is readonly once you set it up (I mean, technically you can update data in it, but you have to convert to openArray or to a pointer)
```nim
  SharedBuffer*[T] = object
    buf: ref seq[T]
    start, len: int
```
(note that the buf is not exposed)

It's also sliceable for "free", which is not really used to it's full potential right now, but could be in the future, for instance to handle the slicing in mplex & noise with zero-copy (you would still have to recombine, and thus copy, at a lower layer ofc)

I did a dumb benchmark where I spawn a GS mesh and send a 0.5mb message in it, here are the total allocated memory for each branch:
`unstable`: 413mb
`write-fast-path`: 390mb (-23mb compared to unstable)
`sharedbuffer`: 360mb (-53mb compared to unstable)

My branch needs more work in terms of usability etc, but with basically a `s/seq/SharedBuffer`, we have more gain than this branch, and with better readability/maintainability, imo

By optimizing it further, we could even reuse the same buffer for each of the copies we send in gossipsub until the secure layer, and do basically only 1 copy per send (+ 1 if the end user doesn't use a SharedBuffer directly)

arnetheduck:
> `SharedBuffer`

One reason we didn't go down this path before was that you then need to copy data _into_ the `SharedBuffer` first - as an "internal" type it makes sense in some cases, but .. once you export it as part of a public API, it becomes quite a heavy requirement for callers to follow.. it's a pragmatic solution for sure, but not without drawbacks.

arnetheduck:
> copies we send in gossipsub

mplex, afair, has connection-specific headers (because of the id field) - gossipsub already does this trick to the extent that it can ("observer" support makes this ugly, and so far hasn't been used).

It would be interesting to see where the other 30:ish MB of copies actually happen - I know many happen during parsing.

Menduist:
> need to copy data into the SharedBuffer

Sure, it's one copy for when data enters libp2p, basically. And it would have to be done at one point anyway
And we can imagine giving an API like:
```nim
let sharedBuffer = SharedBuffer.create:
  # buf is injected, and is equal to result.buf[]
  buf = newSeqUninitialized[byte](msg.len + vbytes.len)
  buf[0..<vbytes.len] = vbytes.toOpenArray()
  buf[vbytes.len..<buf.len] = msg.sbOpenArray()
s.write(sharedBuffer)
```
Both for internal & external applications that want to be careful about memory allocations, while letting the non-careful applications just use the `seq[byte]` + converter

> mplex, afair, has connection-specific headers

Yes, that would imply that `write` takes a `seq[SharedBuffer]` (so new allocation), and we only recombine the full message at the secure layer. We need to test if it's worth it, but it would be 0 copy til secure, and we would be able to reuse the same buffer for slices of each peer

arnetheduck:
full circle to `writev`, invented 40 years ago :)

Menduist:
Yes, as I said earlier, we can merge this while we don't have anything better. `SharedBuffer` seems like an interesting option for 1.2, orc would solve the issue entirely with minimal code change AFAICT: https://github.com/nim-lang/RFCs/issues/16#issuecomment-626875867

Menduist (approved):

dryajov:
`SharedBuffer` doesn't seem unreasonable to me. I think it's a cleaner and less error prone approach and it's less disruptive for 1.6.x than this manual optimizations. With orc and the new runtime ownership, we can easily migrate back to using straight `seq[byte]`. 

My issues with this tailor made optimizations is that they completely invalidate the main reason for using `async/await` in the first place - syntactic sugar. If we're going to go this route, it was probably better to stick to pure `Future` and callbacks style.

dryajov:
I'm OK merging this btw.

zah:
> like its chronos counterpart, also steps outside of the memory-safe world, even though it offers its own .. utilities to manage the problem. introducing it amounts to a paradigm shift and a rewrite.

@arnetheduck, can you please open issues in nim-fastsreams the clarify what parts of the API you consider unsafe. The public API of FastStreams's `outputs` module tries to provide only safe constructs in general (but perhaps I have forgotten some details).

> well, the thing with the current construct is that we let the language provide this guarantee - faststreams breaks the language guarantees and offer a "library guarantee" instead that is weaker

Again, I'm really not sure what you are referring to here, so documenting your concerns in an issue would be helpful.

zah:
Can you also elaborate on this argument as well:

> or the ability to pass a sequence of buffers to the "next" layer such that when a copy "must" be made (for encryption for example), several "input" buffers can be compacted into one. the "header" support in faststreams could be seen as a special case of this, because you still need to "preallocate" header space in faststreams - ie it's not a general solution.

Since FastStreams stores the accumulated data in its multi-page `Buffers` structure, most of the internal mechanisms are already equipped to handle data split into multiple buffers (similar to `writev`). As one concrete example of this, I would also mention that the buffering mechanism was specifically designed to be able to handle de-multiplexers such as mplex in efficient way without memory copies. The input stream of Mplex arrives as a sequence of "page refs" which are then directly shared with the downstream protocols by only adjusting the readable "boundaries" of the pages.

By "preallocating" headers, I assume you mean that writing your header directly to the stream leads to memory allocation within the stream? By default, FastStreams will use a known fast path in the Nim allocator by fetching a fixed size page from a free list, so the way to beat its allocation speed would be to use stack allocation or no-alloaction at all (i.e. you can pass the address of a global constant). Is this what you are referring to? It is possible to introduce support for such "unsafe pages" and this would indeed steps outside of the memory-safe world, but it's not done right now.

arnetheduck:
> parts of the API you consider unsafe

`unsafeMemoryOutput`, `PageSpan` - `grep ptr` is a good way to find them - I don't know about opening issues, they're kind of part of the vocabulary of the library, and central parts of its implementation. The problem isn't really that they exist, but rather that they "infect" everything else, so you have to reason about them when debugging. A bit like the `pointer/len` interface that `chronos` exports.

> sure what you are referring to here

What I call a "library guarantee" about memory safety is a guarantee that the author of a library provides, where they say "I tried to keep it safe, both in API and implementation". If you use the "safe" subset of the library, you hope that the library author didn't mess up, but even as an expert in the language, you don't really have good intuition about what's going on under the hood of that library (assuming it doesn't become a "vocabulary" library).

A language guarantee on the other hand _tends_ to be rooted in some more or less sound type theory that is intimately known by anyone familiar with the language, including anyone looking at libp2p for the first time. `async` is slow and makes lots of copies to uphold these guarantees because nim lacks the vocabulary to explain the safe cases to the compiler (without resorting to `ref`, though we know from the structure of the code, and indeed from this PR, that parts of the problem _can_ be solved without these concessions) - many could be avoided, but here we are.

When you create an API that works with `OutputStream` as parameter, you really have no clue what you're getting - it could be anything - a stream, a snazzy multipage memory-mapped compressing encrypting buffered socket or anything in between, so you have to assume the worst and support everything, both in your logic and your error handling. 

Every advanced feature you promote in faststreams (headers, memory mappings, multilayered buffers with advanced page handling and so on, when someone reads about them, they mentally replace "faststream does the advanced XYZ" with "_when_ something goes wrong, I will have to debug the advanced XYZ". This is also a major contributing factor to why libp2p focuses on `seq[byte]`, both in its implementation and its public API - they're dumb and slow, but you can treat each of them in isolation and they're incredibly easy to understand and debug.

I'll note that major rewrites tend to be a lot more costly than expected, as is over-generalizing and over-abstracting a problem - this PR fixes a small problem with a really, really cheap fix (it took me more time to discuss it in this issue than to find and rewrite it, simply because the code is so dumb and trivial).

> leads to memory allocation within the stream

presumably, a theoretical "optimal" solution could anticipate the the size of all headers and make a single, contiguous memory area for the data and all its headers - afair faststreams had some trick for that as well, except in this case we don't trivially know the header size up-front, without breaking abstraction. Funny enough, our performance problems are not so much in "allocating" memory (as you observe, there's a fast path), but rather in releasing it afterwards - I guess what goes up must come down as well :)

zah:
`unsafeMemoryOutput` exists in the library for completeness and symmetry with `unsafeMemoryInput`. A library such as nim-libp2p doesn't need these features and can easily stick to the safe subset of the APIs provided by FastStreams. I disagree with the notion that the mere existence of these APIs which are clearly marked as "unsafe" should be used as an argument against the adoption of the library. `PageSpan` and the pointer types used behind the scenes are internal implementation detail. They don't make the APIs unsafe, but you seem to acknowledge that and you put in the category of "library provided guarantees".

As for the general argument against sophistication, I find it pitiful that it's given so much weight. When was the last time you had to "debug a stream issue" while working with SSZ, Json, Chronicles, Snappy or any of the other libraries that use FastStreams behind the scenes? Perhaps, a stream is something that can be abstracted in a library well enough so debugging it is not necessary? (like in every mainstream programming language).

So, I'll just have to repeat my message here: If you want to make Chronos and LibP2P leaner and faster, it would be useful to consider how some of the FastStreams ideas can be incorporated. It's actually fair to say that @Menduist's `SharedBuffer` idea is doing exactly that. It replicates pretty closely FastStream's `Page` type:

https://github.com/status-im/nim-faststreams/blob/master/faststreams/buffers.nim#L13-L16

```nim
type
  SharedBuffer*[T] = object
    buf: ref seq[T]
    start, len: int 
```
vs 

```nim
type
  Page* = object
    consumedTo*: Natural
    writtenTo*: Natural
    data*: ref string
```

zah:
Rewriting everything in terms of FastStreams is surely a large undertaking, but if there was at least some appetite for it early on when I was suggesting it more strongly, we could have been in a much better position now IMO. Is essence, you are saying that understanding/debugging a "third-party" stream library may be hard sometimes, so the much better alternative for the team is to have 3 separate stream libraries instead :) I guess this way everybody gets to debug the one they are familiar with.

PR #661, opened 2021-11-25
Try all addresses before aborting start 

dryajov:


Menduist:
This goes against the "fail early" discussed in 662, so closing for now

As part of #667 we'll probably have to do something similar, though

PR #662, opened 2021-11-26
fix: remove returned Futures from switch.start

emizzle:
The proc `start` returned a seq of futures that was mean to be awaited by the caller. However, the start proc itself awaited each Future before returning it, so the ceremony requiring the caller to await the Future, and returning the Futures themselves, was unnecessary.

Remove `switch.start` return type (implicit `Future[void]`)

Update tutorials and examples to reflect the change.

### DIFF
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -234,7 +234,7 @@ proc stop*(s: Switch) {.async.} =
 
   trace "Switch stopped"
 
-proc start*(s: Switch): Future[seq[Future[void]]] {.async, gcsafe.} =
+proc start*(s: Switch) {.async, gcsafe.} =
   trace "starting switch for peer", peerInfo = s.peerInfo
   var startFuts: seq[Future[void]]
   for t in s.transports:
@@ -253,7 +253,7 @@ proc start*(s: Switch): Future[seq[Future[void]]] {.async, gcsafe.} =
 
   for s in startFuts:
     if s.failed:
-      info "Failed to start one transport", error=s.error.msg
+      error "Failed to start one transport", error=s.error.msg
 
   for t in s.transports: # for each transport
     if t.addrs.len > 0:
@@ -261,8 +261,6 @@ proc start*(s: Switch): Future[seq[Future[void]]] {.async, gcsafe.} =
       s.peerInfo.addrs &= t.addrs
 
   debug "Started libp2p node", peer = s.peerInfo
-  return startFuts # listen for incoming connections
-
 
 proc newSwitch*(peerInfo: PeerInfo,
                 transports: seq[Transport],
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -253,7 +253,10 @@ proc start*(s: Switch) {.async, gcsafe.} =
 
   for s in startFuts:
     if s.failed:
-      error "Failed to start one transport", error=s.error.msg
+      # TODO: replace this exception with a `listenError` callback. See
+      # https://github.com/status-im/nim-libp2p/pull/662 for more info.
+      raise newException(transport.TransportError,
+        "Failed to start one transport", s.error)
 
   for t in s.transports: # for each transport
     if t.addrs.len > 0:
--- a/examples/directchat.nim
+++ b/examples/directchat.nim
@@ -181,7 +181,7 @@ proc main() {.async.} =
 
   switch.mount(ChatProto.new(chat))
 
-  let libp2pFuts = await switch.start()
+  await switch.start()
 
   let id = $switch.peerInfo.peerId
   echo "PeerID: " & id
@@ -190,7 +190,6 @@ proc main() {.async.} =
     echo &"{a}/p2p/{id}"
 
   await chat.readLoop()
-  await allFuturesThrowing(libp2pFuts)
   quit(0)
 
 waitFor(main())
--- a/examples/helloworld.nim
+++ b/examples/helloworld.nim
@@ -65,9 +65,8 @@ proc main() {.async, gcsafe.} =
 
   # Start the nodes. This will start the transports
   # and listen on each local addresses
-  let
-    switch1Fut = await switch1.start()
-    switch2Fut = await switch2.start()
+  await switch1.start()
+  await switch2.start()
 
   # the node addrs is populated with it's
   # actual port during the start
@@ -87,6 +86,5 @@ proc main() {.async, gcsafe.} =
   await conn.close()
 
   await allFutures(switch1.stop(), switch2.stop()) # close connections and shutdown all transports
-  await allFutures(switch1Fut & switch2Fut) # wait for all transports to shutdown
 
 waitFor(main())
--- a/examples/tutorial_1_connect.md
+++ b/examples/tutorial_1_connect.md
@@ -73,9 +73,8 @@ We can now create our two switches:
     
   switch1.mount(pingProtocol)
 
-  let
-    switch1Fut = await switch1.start()
-    switch2Fut = await switch2.start()
+  await switch1.start()
+  await switch2.start()
 ```
 We've **mounted** the `pingProtocol` on our first switch. This means that the first switch will actually listen for any ping requests coming in, and handle them accordingly.
 
@@ -98,7 +97,6 @@ We now have a `Ping` connection setup between the second and the first switch, w
 And that's it! Just a little bit of cleanup: shutting down the switches, waiting for them to stop, and we'll call our `main` procedure:
 ```nim
   await allFutures(switch1.stop(), switch2.stop()) # close connections and shutdown all transports
-  await allFutures(switch1Fut & switch2Fut) # wait for all transports to shutdown
   
 waitFor(main())
 ```
--- a/examples/tutorial_2_customproto.md
+++ b/examples/tutorial_2_customproto.md
@@ -57,11 +57,10 @@ proc main() {.async, gcsafe.} =
   
   switch1.mount(testProto)
   
-  let
-    switch1Fut = await switch1.start()
-    switch2Fut = await switch2.start()
+  await switch1.start()
+  await switch2.start()
     
-    conn = await switch2.dial(switch1.peerInfo.peerId, switch1.peerInfo.addrs, TestCodec)
+  let conn = await switch2.dial(switch1.peerInfo.peerId, switch1.peerInfo.addrs, TestCodec)
 
   await testProto.hello(conn)
 
@@ -69,7 +68,6 @@ proc main() {.async, gcsafe.} =
   await conn.close()
 
   await allFutures(switch1.stop(), switch2.stop()) # close connections and shutdown all transports
-  await allFutures(switch1Fut & switch2Fut) # wait for all transports to shutdown
 ```
 
 This is very similar to the first tutorial's `main`, the only noteworthy difference is that we use `newStandardSwitch`, which is similar to `createSwitch` but is bundled directly in libp2p
--- a/tests/testidentify.nim
+++ b/tests/testidentify.nim
@@ -129,7 +129,6 @@ suite "Identify":
       switch2 {.threadvar.}: Switch
       identifyPush1 {.threadvar.}: IdentifyPush
       identifyPush2 {.threadvar.}: IdentifyPush
-      awaiters {.threadvar.}: seq[Future[void]]
       conn {.threadvar.}: Connection
     asyncSetup:
       switch1 = newStandardSwitch()
@@ -146,8 +145,8 @@ suite "Identify":
       switch1.mount(identifyPush1)
       switch2.mount(identifyPush2)
 
-      awaiters.add(await switch1.start())
-      awaiters.add(await switch2.start())
+      await switch1.start()
+      await switch2.start()
 
       conn = await switch2.dial(
         switch1.peerInfo.peerId,
@@ -166,9 +165,6 @@ suite "Identify":
 
       await switch1.stop()
       await switch2.stop()
-
-      # this needs to go at end
-      await allFuturesThrowing(awaiters)
 
     asyncTest "simple push identify":
       switch2.peerInfo.protocols.add("/newprotocol/")
--- a/tests/testinterop.nim
+++ b/tests/testinterop.nim
@@ -54,7 +54,7 @@ proc testPubSubDaemonPublish(gossip: bool = false, count: int = 1) {.async.} =
 
   nativeNode.mount(pubsub)
 
-  let awaiters = nativeNode.start()
+  await nativeNode.start()
   await pubsub.start()
   let nativePeer = nativeNode.peerInfo
 
@@ -90,7 +90,6 @@ proc testPubSubDaemonPublish(gossip: bool = false, count: int = 1) {.async.} =
 
   await nativeNode.stop()
   await pubsub.stop()
-  await allFutures(awaiters)
   await daemonNode.close()
 
 proc testPubSubNodePublish(gossip: bool = false, count: int = 1) {.async.} =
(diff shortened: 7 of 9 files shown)


Menduist:
I agree with this change, but see here: https://github.com/status-im/nim-libp2p/pull/609#discussion_r671147206

We may need to provide better api to see start errors before removing this

Menduist:
Even more so with https://github.com/status-im/nim-libp2p/pull/661, which can make the start fail on some addresses transparently


Maybe in start, return the number of "addrs" that failed to start, and then the application can check the array of addrs (which will only contain the one that succeeded to start) to see if it can recover?

Seems pretty flexible:
- if you want all addresses to listen no matter what, you fail instantly `if await s.start() > 0`
- if you want, for instance, tcp to always succeed, but allow ws to fail, you can fail `if addr.filter(isTcp).len != switch.addr.filter(isTcp).len`

wdyt @arnetheduck @dryajov @jm-clius @emizzle

dryajov:
Yeah, @emizzle and I discussed this particular change but I totally forgot that we found a new use the futures returned by start.

> Maybe in start, return the number of "addrs" that failed to start, and then the application can check the array of addrs (which will only contain the one that succeeded to start) to see if it can recover?

Well, I think to set the context here, we need to first understand what sort of startup behavior we want to begin with:

- `Optimistic` - continue despite any errors and only fail if all transports failed
   - this is what with get with #598 and #661
- `Pessimistic` - fail fast on first failed transport
   - probably the simplest to do
- `Hybrid` - fail the transport on first failed address, but continue with start procedure for other transports
   - this is what we have in `unstable` with #598

I like `Optimistic` with an array of the original addresses that failed to bind to any transport, which is what I think @Menduist is proposing above. This would require merging #661 and changing the return to a `Future[seq[MultiAdress]]` in `start`, containing the original address that we couldn't listen on.

arnetheduck:
why not create sockets etc as part of creating the transports? and fail there?

arnetheduck:
In general, this is the same discussion as we've had a few times re futures and error handling: you can't generally simplify a multiple-errors world into a single error without information loss and/or compromises - you can't raise exceptions in loops of any kind without messing things up if you're not careful - this applies to a loop of transports starting, as well.

Consider for example the `Pessimistic` approach - if something fails mid-way, the already-started transports must be stopped and any side effects undone (ie if someone already connected while the other transports were being started - for example because the starting itself is async -they need to be disconnected and so on) - simple is not the right word here.

dryajov:
> why not create sockets etc as part of creating the transports? and fail there?

Who creates the transports and when?

Menduist:
I would say always succeed, and let the application choose what to do with the failure

emizzle:
I may be reading things wrong, and I'm still very new to libp2p, so my apologies if this is a bit off the mark. Hoping to help brainstorm here...

Perhaps we could leave the startup behaviour decision up to the calling application. Instead of trying to reduce the result of starting multiple transports (each with potentially multiple addresses) in to a single return type (which can be lossy), could we instead store the state of the attempted transport starts/attempts inside of the transport itself, and then the calling application could inspect the switch and iterate the attempted transports and do what it likes with successes/failures and state?

We could utilise `stew/results` instead of exception handling for `Transport.start` and `Transport.accept`:
1. `Transport.start()` could return `Future[seq[Result[void, E]]]`, one for each address attempted, where E is a reason for an except failure, or possibly could contain an exception.
2. `Transport.accept` could return `Future[Result[Connection, E]]`, where E is a reason for an except failure, or possibly could contain an exception.

emizzle:
I'll close this PR for now as I don't think returning nothing is really on the table at this point. We can reopen it later if needed.

emizzle:
@dryajov and I discussed this issue in detail and have come up with a couple of possible solutions, and we think going with solution 2 below is the best approach.

## Solution 1
As @arnetheduck suggested,
> why not create sockets etc as part of creating the transports? and fail there?

We could move the socket creation in `start` to the transport constructor, allowing the caller to detect failures upon instantiation. This could then lead to no longer needing a `start` proc on the transport, leaving the caller with needing to call `accept` in the transport bootstrap.

The downside to creating a socket in the transport constructor is that it could prematurely trigger an OS's network permissions prompt when transports are being set up. To workaround this, calling applications could move transport instantiation to the point where users are ready to accept network permissions. However, from an API design point of view, maybe it wouldn't be obvious that the transport constructor attempts to create a socket.

The upside of this is that it forces the caller to handle `start` (instantiation in this solution) and `accept` separately, which currently are called together in `switch.start`. We can handle this in solution 2 instead.

In terms of bubbling exceptions to the caller, the exceptions raised during `new` and `accept` would still need to make their way to the caller so they can decide how to handle them. Again, we can handle this in solution 2 instead.

## Solution 2
`switch.start` currently iterates each transport, calling first `transport.start`, then `transport.accept`. We can split out the accept part and put it in an exported proc `switch.accept*` (currently switch has a non-exported `accept`). This provides several advantages:
1. The caller will be able to handle results from `start` and `accept` separately and make decisions based on which transports failed and why.
2. It allows the caller to inspect results from `switch.start` and make a decision to continue on calling `switch.accept`.
3. Prevents the need for tracking state in the switch and/or in the transports as both `start` and `accept` can return the result of their operations independently.

Additionally, if we revert the support for multiple addresses per transport, having each transport support only one address  allows further simplification of results returned by `transport.start` by not requiring the transport to store any kind of state or `start/accept` results for inspection by the caller. It also simplifies the current `accept` ceremony needed for multiple address support.

### Solution 2 API changes
Solution 2 would need the following api changes made:
1. `Transport.start` would return `Future[Result[void, TransportError]]`. `TransportError` is an enum of errors that could occur on the transport layer as a whole.
2. `Transport.accept` would return `Future[Result[Connection, TransportError]]`.
3. `Switch.start` would return `Future[seq[Result[void, SwitchError]]]`, containing a result for each transport started. `SwitchError` is an enum of errors that could occur on the switch layer as a whole.
4. `Switch.accept*` would be created and return `Future[seq[Result[void, SwitchError]]]`, containing a result for each `transport.accept` call.

Current implementations that call `switch.start` will now also need to inspect the results of `switch.start`, and if no errors, continue on to call `switch.accept` and inspect the results for errors.

Altogether, this will allow exceptions/error context to not be lost and returned to the caller.

### Solution 2 questions
1. @arnetheduck what level of transport exception granularity is needed when calling `switch.start`? Would an enum be sufficient for `SwitchError` and `TransportError` (via `.mapErrTo`) or would we need to bubble the actual exceptions encountered all the way up to `SwitchError`?

arnetheduck:
> what level of transport exception granularity is needed

I don't have a good answer really because whether errors should be generic or transport-specific is highly application-dependent - ie it will depend on how much they _want_ to know about the specific transports being used - here's an experiment that will answer this question in the best possible way I can think of: make a PR for nim-waku and another for nimbus-eth2 that uses the new interface and see what the fallout is.

Menduist:
We talked about this in the libp2p meeting, and the callback really starts to sound nice. The idea would be:

We have a callback, eg by default
```nim
proc callback(ma: MultiAddress, err: CatchableError) =
  raise err
```

The application has the option to override it, and in some cases, swallow the err instead of raising it, for instance
```nim
switch.listenError = proc callback(ma: MultiAddress, err: CatchableError) =
  if TCP.match(ma): raise err
```

The start doesn't return anything anymore. Instead, the callback will bubble up to the `start` caller, if not swallowed.
And the start will cleanup in case of error



This gives great fleaxibility while being pessimistic by default, which is pretty nice, I think.
The other nice thing is that we can merge this PR as-is, and add the callback in a later version

emizzle:
> So what we need is a mapping Listen multiaddress -> error

@Menduist I think we are saying the same thing 😄 My apologies for not doing a good enough job explaining the thought process. @dryajov discussed removing the multiple address per transport support such that each transport has only one address, in a 1:1 mapping. This can facilitate your next suggestion

> return a mapping MultiAddress -> CatchableError/Result/Whatever

When transport/addresses are in a 1:1 mapping, we can get an error per transport/address as `seq[Result[void, SwitchError]`, from the `switch.start` api change:
> `Switch.start` would return `Future[seq[Result[void, SwitchError]]]`, containing a result for each transport started. `SwitchError` is an enum of errors that could occur on the switch layer as a whole.

emizzle:
The callback seems like a nice approach as well.

> We have a callback, eg by default
> 
> ```nim
> proc callback(ma: MultiAddress, err: CatchableError) =
>   raise err
> ```

If we re-raise the exception by default, then we are pessimistic by default and may be missing out on getting other exceptions from other transport attempts. Maybe we should allow for an optimistic approach by default? Also, we could encapsulate the error at the switch layer which could contain the original exception as an inner exception:
```nim
type
  SwitchError object of CatchableError =
    inner*: ref CatchableError

proc callback(ma: MultiAddress, err: SwitchError) =
  discard # optimistic by default
```
If the calling application wants to be pessimistic, this would allow it to be.

We should likely still split up `switch.start` and `switch.accept` and add a callback for `swith.acceptError` for the same reasons (to prevent optimistic/pessimistic pigeonholing). So maybe we should create callbacks for both `switch.startError` (instead of `listenError`) and `switch.acceptError`?

Altogether, the changes would include:
1. revert the multiple address support per transport, to create a one-to-one mapping of transport to address
2. modify the following:

#### `switch.nim`
```nim
type
  SwitchError object of CatchableError =
    inner*: ref CatchableError
  SwitchCallback = proc (ma: MultiAddress, err: SwitchError) {.gcsafe, async.}
  Switch* = ref object of Dial
    startError*: SwitchCallback
    # ...

proc start*(s: Switch) {.async, gcsafe.} =
  # for each transport, call transport.start
  #     if err, call s.startError
  # ...
proc accept*(s: Switch) {.async, gcsafe.} =
  # for each transport, call transport.accept
  #     if err, call s.acceptError
  # ...
proc newSwitch*(...): Switch =
  # ...
  let switch = Switch(...)
  switch.startError = proc (ma: MultiAddress, err: SwitchError) = discard # default to optimistic
  switch.acceptError = proc (ma: MultiAddress, err: SwitchError) = discard # default to optimistic
```

#### `transport.nim`
```nim
type
  TransportError object of CatchableError =
    inner*: ref CatchableError

method start*(self: Transport, addrs: seq[MultiAddress]): Future[Result[void, TransportError]] {.base, async.} =
  # ...
method accept*(self: Transport, addrs: seq[MultiAddress]): Future[Result[Connection, TransportError]] {.base, async.} =
  # ...
```

Menduist:
Sorry, I forgot to reply to some of your points, indeed:

### 1:1 mapping of transports
That was originally my preferred solution to handle multiple addresses per transport, but Dmitriy changed my mind, you can follow a bit of the story here: https://github.com/status-im/nim-libp2p/issues/597. Mark also confirmed that this was a good idea iirc, because quic allows you to specify natively multiple addresses on startup, so it makes sense to let the transport handle the multiple addresses logic

Also, even if you do this, you'll still not have a 1/1 binding, for instance if an address failed because no transport handle it 

### Pessimistic by default
imo, letting an address start to fail is a fairly advanced use of libp2p: it means you have enough addresses that you don't care about one failing. The simple use case should be to fail early, because that's what most applications would expect.

For instance, currently, almost no one bother to check the return of start, hence my idea to fail start completely on error, and make it tunable for advanced applications

### Splitting start/accept
I'm not really convinced about this, I don't see much improvement from it, but maybe I missed something

accept will only fail if start failed, the two are very closely bound together. When you think about it, "starting" a transport is really "setting up his accept socket", you could dial from a transport without starting it

And actually, accept cannot even fail: we start the futures without checking if they failed or succeeded (which makes sense, since we have to wait for a connection to see if it did)

Opening this granularity _could_ make sense in some applications, eg it could be seen as "start the switch, and after a while start to accept incoming connections, or don't"
But that's again an advance use case, most users will just want to start everything and not worry about it.

As @cskiraly proposed, we can have two apis to start: one with just `start` for most users, and one with `setup`/`accept` for advance users
But that's something we can do in the future, if someone needs it



Obviously I'm not stuck on theses opinions, and thanks again for brining back this topic on the table which is one of the the thing we should change before V1 :)

dryajov:
With regards to splitting out `accept`, I think that in its current form it doesn't actually fix anything, since as pointed out by @Menduist, `accept` wont raise any errors. This is an issue, because there is no way of communicated to the user when a transport stopped accepting connection and why (for example, we might have hit an OS limit and the transport silently exited). Moreover, we don't restart failed accepts so eventually we might end up with a dead switch if all accepts failed. So all in all, this is yet another thing that needs addressing so I've opened a separate issue to track it https://github.com/status-im/nim-libp2p/issues/666 (granted we don't seem to have ran into any serious issues with it just yet).

emizzle:
> I'd rather return a value (exception, result, etc...) for the sake of keeping the codebase consistent.

@dryajov, apologies I'm a bit confused by this. Do you mean callbacks should return a value or do you mean we not use callbacks at all and instead return a value from `start` to keep consistency?

> that's again an advance use case

@Menduist makes sense, but for any advanced use case that needs to be catered for, the library needs to have the apis available, and the simplified use cases would make use of those apis. It sounds like we'd need to wait on #666 before being able to support these use cases anyway.

Putting everything together now, we are at the point where we're going to keep everything as it is and use `listenError` callbacks only, correct? We'd need to pass `listenError` to the transport as well. Probably makes the most sense to include it in `SwitchBuilder`?

Menduist:
> the library needs to have the apis available, and the simplified use cases would make use of those apis

Absolutely, but switching to a setup/listen and start calls both in the future is a non-breaking change, so we could do it whenever we need it

I would merge this PR as-is (already a big, breaking change), and create another PR to add the callback, where we can discuss specifics such as raise vs return etc (which will not be a breaking change)
And in the future, another PR to split setup & listen in two procs (again, not a breaking change), when we need it

libp2p/switch.nim
```diff
@@ -253,16 +253,14 @@ proc start*(s: Switch): Future[seq[Future[void]]] {.async, gcsafe.} =
 
   for s in startFuts:
     if s.failed:
-      info "Failed to start one transport", error=s.error.msg
+      error "Failed to start one transport", error=s.error.msg
```
Menduist, line 256: So we would throw here (and in the future, add the callback)

(Or better, get rid of startFuts and simplify the logic to a single loop)

emizzle: I think the multiple loops here are warranted as we build up a sequence of Futures and let them all complete asynchronously. A single loop would require us to await the start in each iteration.

There is now an exception raised here, with a TODO note about replacing with a `listenError` callback.

emizzle:
> create another PR to add the callback

https://github.com/status-im/nim-libp2p/issues/667

Menduist (approved):
LGTM, waiting for @dryajov review to be sure

As a recap, this PR just remove the return value from start, and fails eagerly instead
Next PRs will follow up on this

dryajov:
@emizzle 

> @dryajov, apologies I'm a bit confused by this. Do you mean callbacks should return a value or do you mean we not use callbacks at all and instead return a value from start to keep consistency?

What I mean is that we usually don't throw from callbacks as it is too error prone, specially in combination with async. This is why all callbacks are usually declared with `raises: [Defect]`. I'm suggesting that we don't re-raise, but return a new error (or the same one) form the callback, something like this:

```nim
try:
  ...
except CatchableError as exc:
  let err =  self.handleError(exc)
  if not err.isNil: 
    raise err
```

dryajov (approved):
LGTM

libp2p/switch.nim
Menduist: Yeah, in practice, launching a transport is nearly instantaneous (as proof, before a few month ago, we weren't actually waiting for them and everything was working most of the time, see https://github.com/status-im/nim-libp2p/pull/609)
so launching them in parallel is a bit useless in terms of speed (at least with current transports, but I guess with future ones too), however it makes the logic more complex

But we can also refacto that as part of #667

PR #663, opened 2021-11-26
fix (directchat): auto disconnect self once conn closed

emizzle:
When closing a connection of clientA(using the command “/disconnect”), the connection is then closed on clientA’s side. However, the connection remained open on clientB and clientB could continue sending message to clientA and clientA would receive those messages.

This PR listens for a closing connection on clientB, then also closes clientB’s connection.

### DIFF
--- a/examples/directchat.nim
+++ b/examples/directchat.nim
@@ -79,7 +79,9 @@ proc handlePeer(c: Chat, conn: Connection) {.async.} =
       c.writeStdout $conn.peerId & ": " & $str
 
   except LPStreamEOFError:
-    c.writeStdout $conn.peerId & " disconnected"
+    defer: c.writeStdout $conn.peerId & " disconnected"
+    await c.conn.close()
+    c.connected = false
 
 proc dialPeer(c: Chat, address: string) {.async.} =
   # Parse and dial address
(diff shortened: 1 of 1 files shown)


Menduist (changes requested):
https://github.com/status-im/nim-libp2p/pull/662 needs to be merged first, blocking here in the meantime

arnetheduck:
this looks off  - generally, closing is a two-phase operation where one side initiates the close and the other side performs on receiving the close message and does cleanup (which might include sending more messages) then closes its side - automatic closing prevents that from happening.

emizzle:
> https://github.com/status-im/nim-libp2p/pull/662 needs to be merged first, blocking here in the meantime 

Because #662 was closed, this has been rebased on top of unstable.

dryajov (approved):
LGTM. This can be merged regardless of #662.

examples/directchat.nim
```diff
@@ -79,7 +79,9 @@ proc handlePeer(c: Chat, conn: Connection) {.async.} =
       c.writeStdout $conn.peerId & ": " & $str
 
   except LPStreamEOFError:
-    c.writeStdout $conn.peerId & " disconnected"
+    defer: c.writeStdout $conn.peerId & " disconnected"
```
dryajov, line 82: Just a word of caution with `defer`, which is known to fail when combined with `async`.

Menduist (approved):

PR #664, opened 2021-11-26
fix: remove returned Futures from switch.start

emizzle:
The proc `start` returned a seq of futures that was mean to be awaited by the caller. However, the start proc itself awaited each Future before returning it, so the ceremony requiring the caller to await the Future, and returning the Futures themselves, was unnecessary.

Remove `switch.start` return type (implicit `Future[void]`)

Update tutorials and examples to reflect the change.


examples/tutorial_2_customproto.md
```diff
@@ -22,27 +22,33 @@ type TestProto = ref object of LPProtocol
 
 We've set a [protocol ID](https://docs.libp2p.io/concepts/protocols/#protocol-ids), and created a custom `LPProtocol`. In a more complex protocol, we could use this structure to store interesting variables.
 
-A protocol generally has two part: and handling/server part, and a dialing/client part.  
-Theses two parts can be identical, but in our trivial protocol, the server will wait for a message from the client, and the client will send a message, so we have to handle the two cases separately.
+A protocol generally has two parts: a handling/server part, and a dialing/client part. These two parts can be identical, but in our trivial protocol, the server will wait for a message from the client, and the client will send a message, so we have to handle the two cases separately.
 
-Let's start with the server part:
+Lets start with the server part:
```
Menduist, line 27: Is it Lets?

examples/tutorial_2_customproto.md
```diff
@@ -54,23 +60,23 @@ proc main() {.async, gcsafe.} =
...
   await switch1.start()
   await switch2.start()
-    
   let conn = await switch2.dial(switch1.peerInfo.peerId, switch1.peerInfo.addrs, TestCodec)
 
+  # conn is now a fully setup connection, we talk directly to the switch1 custom protocol handler
   await testProto.hello(conn)
 
-  # We must close the connection ourselves when we're done with it
+  # We must close the connections ourselves when we're done with it
```
Menduist, line 73: This sentence is talking specifically about `conn`, the connection we got from dial.

This is actually a stream and not a connection: we need to close the stream, and libp2p will handle to close the connection

examples/tutorial_2_customproto.md
```diff
@@ -54,23 +60,23 @@ proc main() {.async, gcsafe.} =
...
   await conn.close()
 
-  await allFutures(switch1.stop(), switch2.stop()) # close connections and shutdown all transports
+  await allFutures(switch1.stop(), switch2.stop()) # shutdown all transports
```
Menduist, line 76: And libp2p will close all current connections during stop, so my comment was accurate :) though it probably need better explanation

examples/tutorial_3_streamproto.md
```diff
@@ -0,0 +1,94 @@
...
+      let
+        # readLp reads length prefixed bytes and returns a buffer without the prefix
+        strData = await conn.readLp(1024)
+        str = string.fromBytes(strData)
+      echo "[client] received from server: ", str
+
+  except LPStreamEOFError:
+    # Once the remote peer has closed disconnected (closed the connection),
+    # we should also close the connection from this end as well.
+    # The only way to tell that a connection has been closed on the remote
```
Menduist, line 59: You can also use conn.closed() or conn.atEof() (which have different meanings because the connection can be half closed)

Menduist (commented):
You can also add a link to your tutorial at the end of the second one, and add it to the CI in libp2p.nimble

examples/tutorial_2_customproto.md
emizzle: Gotcha, thanks. Should the comment on line 70 be:
```
# conn is now a fully setup stream, which we use to talk directly to the switch1 custom protocol handler
```
?

examples/tutorial_2_customproto.md
emizzle: Nope, you were definitely right! Really sorry for that.

examples/tutorial_2_customproto.md
emizzle: Or maybe we should just leave it with no comment 😄

examples/tutorial_2_customproto.md
emizzle: Yep, gotcha. I think I was assuming this comment was referring to the stream (`conn` created during `dial()`).

examples/tutorial_2_customproto.md
Menduist: Yeah, we can rename it stream, and explain the flow better in part 1
I can do it if you want

examples/tutorial_3_streamproto.md
emizzle: I had tried both of these actually (and retried again since you mentioned it), and at no point inside the while loop is `conn.closed` or `conn.atEof` ever true. This is likely my lack of knowledge at how it is meant to work. What am I missing?

examples/tutorial_2_customproto.md
emizzle: I'd definitely appreciate it, for my own learning 😄 🙏 

Note, i've just force pushed (with lease), in case you've pulled the branch already.

examples/tutorial_3_streamproto.md
Menduist: Weird, need to look into that

examples/tutorial_2_customproto.md
Menduist: Did something, LMK what you think :)

examples/tutorial_3_streamproto.md
emizzle: When using `readLp`, I think reason that `atEof` is not true is because `atEof` has already been checked and will not continue to be true on subsequent reads.

emizzle:
> You can also add a link to your tutorial at the end of the second one, and add it to the CI in libp2p.nimble

Fixed up.

Also changed the way the stream checks `atEof` using `readOnce` instead of `readLp` and adjusted the comments and tutorial explanation to suit.

examples/tutorial_3_streamproto.md
Menduist: Well, that's error prone.. Don't see how a stream atEof could not be atEof later
@dryajov we agree that is something that needs to be fixed?

dryajov: `readLp` and `atEof` weren't designed to be used together. The behavior of `readLp` is such that it will always await for all the data from the length prefixed message to be read before returning, which means that there is no way of checking `atEof` or `closed` when used in conjunction with it and other similar primitives such as `readExactly`, `readLine`, `readUntil`, etc... This is why an exception is raised to signal `EOF`. `atEof` and `closed` should however be true once the exception has been triggered. In general `atEof` is useful with `readOnce`, which optimistically reads up to the size of the passed buffer otherwise it returns as much as it was able to read. So, the behavior here is correct.

examples/tutorial_1_connect.md
```diff
@@ -84,16 +84,27 @@ We can find out which port was attributed, and the resulting local addresses, by
 
 We'll **dial** the first switch from the second one, by specifying it's **Peer ID**, it's **MultiAddress** and the **`Ping` protocol codec**:
 ```nim
-  let conn = await switch2.dial(switch1.peerInfo.peerId, switch1.peerInfo.addrs, PingCodec)
+  let stream = await switch2.dial(switch1.peerInfo.peerId, switch1.peerInfo.addrs, PingCodec)
```
dryajov, line 87: What you get from the `switch.dial` might not necessarily be a `stream`, this is very protocol dependent and might be a real connection in some cases.

examples/tutorial_2_customproto.md
```diff
@@ -22,29 +22,35 @@ type TestProto = ref object of LPProtocol
...
 We've set a [protocol ID](https://docs.libp2p.io/concepts/protocols/#protocol-ids), and created a custom `LPProtocol`. In a more complex protocol, we could use this structure to store interesting variables.
 
-A protocol generally has two part: and handling/server part, and a dialing/client part.  
-Theses two parts can be identical, but in our trivial protocol, the server will wait for a message from the client, and the client will send a message, so we have to handle the two cases separately.
+A protocol generally has two parts: a handling/server part, and a dialing/client part. These two parts can be identical, but in our trivial protocol, the server will wait for a message from the client, and the client will send a message, so we have to handle the two cases separately.
 
 Let's start with the server part:
 ```nim
 proc new(T: typedesc[TestProto]): T =
   # every incoming connections will in be handled in this closure
```
dryajov, line 30:
```suggestion
  # every incoming connections will be handled in this closure
```

examples/tutorial_3_streamproto.md
```diff
@@ -0,0 +1,98 @@
...
+type TestStreamProto = ref object of LPProtocol
+```
+
+Just as we did in the last tutorial, we've set a [protocol ID](https://docs.libp2p.io/concepts/protocols/#protocol-ids), and created a custom `LPProtocol`.
+
+As in the last tutorial, we are going to handle our client and server parts separately. The server will wait for an instruction from the client, then send a stream of data to the client and finish by closing the connection. The client will listen in a loop until the server has finished and closed the connection. Then the client will close its end of the connection as well.
+
+Lets start with the server part:
+```nim
+proc new(T: typedesc[TestStreamProto]): T =
```
dryajov, line 27: I think `type` is preferred over `typedesc`?

examples/tutorial_3_streamproto.md
```diff
@@ -0,0 +1,98 @@
...
+proc streamData(p: TestStreamProto, stream: Connection) {.async.} =
+  await stream.writeLp("please send data")
+
+  # Read loop
+  while stream.atEof.not:
+
+    var buf = newSeq[byte](1024)
+    # readOnce reads length prefixed bytes into a buffer and returns the length
+    # of data read
+    let len = await stream.readOnce(addr buf[0], 1024)
```
dryajov, line 54: Don't use `readOnce` with `readLp/writeLp`. There are several reasons for that, one is that it breaks the symmetry of the two calls - there is no point in sending length prefixed data if at the end of the day you'll always read to `EOF` either way. The second is that you now cannot stream different types of messages over the same channel, since you've just made `EOF` your message marker. If the idea is to stream data, you should probably simply call `write` which writes the message as is, without length prefix.

examples/tutorial_3_streamproto.md
Menduist:
```sh
$ grep -rn "typedesc" libp2p | wc -l
121
```
I guess we have to update the entire codebase

examples/tutorial_3_streamproto.md
Menduist: Ok, either I or @emizzle misunderstood something:
```nim
  while not stream.atEof:
    try:
      let buf = await stream.readLp(1024)
      echo "[client] received from server: ", string.fromBytes(buf)
    except LPStreamEOFError:
      echo "finished: ", stream.atEof
  await stream.close()
```
Works fine (and echoes `finished: true`), so you _can_ use `atEof` with lp procs, as I expected

Though in this case you may as well just use the exception. But it's always nice to avoid a while true, imo

dryajov: Heh, this is a bit subtle - the reason it works is because you explicitly handle `LPStreamEOFError` and then check `atEof` after it fired, in this case it will be true, if however, you move the while into the `try`, you won't be able to use it as such.

examples/tutorial_1_connect.md
dryajov: Yes and no, currently all transports are multiplexed with mplex, however that doesn't have to be the case. For example, Quic will most likely bypass multiplexing at the libp2p level since it's natively multiplexed, same for WebRTC. The point is that, we cannot asume that we're getting a multiplexed stream here, even if the underlying transport is doing multiplexing, because it might actually not be the case. But maybe it makes sense to refer to this as `streams` at the high level libp2p api?

examples/tutorial_3_streamproto.md
dryajov: Yeah, the current state isn't indicative of whats preferred ;)

examples/tutorial_3_streamproto.md
Menduist: Ooohhh ok, I understand the misunderstanding.

You wanted to see if the stream was at eof before calling readLp, to avoid the exception raising. That's indeed not possible

And I though you were saying that once it throwed, isEof was still false, which I deemed wrong

So maybe a better wording would be something like "once the other end closes the connection, readLp will throw a [..] and stream.atEof will be true"

examples/tutorial_1_connect.md
emizzle: Because we are instantiated our switch using `newStandardSwitch`, this will always give us a multiplexed tcp stream (based on the current codebase), no? Or is that something is meant to be changing in the future?

examples/tutorial_3_streamproto.md
emizzle: @dryajov I've removed `readLp/writeLp`, and instead am using `write` and `readOnce` only. Wdyt?

emizzle: removed `readLp/writeLp`, and instead am using `write`, `readOnce`, and `atEof`

emizzle:
Updated tutorial 3 (streaming tutorial) to remove `readLp/writeLp` in combo with `atEof`, and instead using `write`, `readOnce`, and `atEof`.

Menduist:
I think 90% of  will

examples/tutorial_3_streamproto.md
Menduist: I personally like having brackets in the expression, because it makes it more clear that this is a generic, imo

Could we switch to `type[TestStreamProto]` instead, so that everybody is happy (well, at least me)? (across the codebase ofc)

examples/tutorial_3_streamproto.md
dryajov: LGTM

emizzle: > I think 90% of users will use the Lp variants, but since we already show it in tutorial 2, it may make sense to show the non-Lp here But it should be clear that Lp can of be used for streaming as well, since it's about what everybody does

I'll make tutorial #3 a streaming tutorial with `readLp/writeLp` and `LPStreamEOFError`, then we can make a fourth with `read/write` and `atEof`.

emizzle:
> I think 90% of users will use the Lp variants, but since we already show it in tutorial 2, it may make sense to show the non-Lp here But it should be clear that Lp can of be used for streaming as well, since it's about what everybody does

I'll make tutorial #3 a streaming tutorial with `readLp/writeLp` and `LPStreamEOFError`, then we can make a fourth with `read/write` and `atEof`.

examples/tutorial_1_connect.md
dryajov: Not really, there is no reason not to add other multiplexers or even disable the multiplexer altogether (QUIC), the point is mostly about nomenclature, so far `connection` and `stream` have been used interchangeably, and there is a good reason for that, i.e. you never know what you're getting as it's completely dependent on the transport and the multiplexer. It's a minor point and I'm not trying to make a big deal out of it, just give some perspective 👍

emizzle:
Should've updated this sooner, added a fourth tutorial as per above comment.

examples/tutorial_4_streamproto2.md
```diff
@@ -0,0 +1,106 @@
...
+```nim
+import bearssl
+import chronos
+import stew/byteutils
+
+import libp2p
+```
+This is similar to the third tutorial. Next, we'll declare our custom protocol.
+```nim
+const TestStreamCodec = "/test/stream-proto/1.0.0"
```
Menduist, line 16:
```suggestion
const TestStreamCodec = "/test/stream-proto/2.0.0"
```
Just to differentiate the two protocols :)

examples/tutorial_4_streamproto2.md
```diff
@@ -0,0 +1,106 @@
...
+Instead of using `readLp` to read data from the connection (as in the previous tutorial), we are going to use `readOnce`, which allows us to read bytes from the buffer or wait for bytes to be received.
+
+We can now create our client part:
+```nim
+proc streamData(p: TestStreamProto, stream: Connection) {.async.} =
+  let data = "please send data"
+  await stream.write(data)
+
+  # Read loop
+  while stream.atEof.not:
```
Menduist, line 58:
```suggestion
  while not stream.atEof:
```
Maybe it's just me, but I find this syntax a bit better, especially to people whom may not be introduced to nim

examples/tutorial_4_streamproto2.md
```diff
@@ -0,0 +1,106 @@
...
+
+    var buf = newSeq[byte](CHUNK_LEN_BYTES)
+    # readOnce reads bytes from internal buffer if present, otherwise waits for
+    # for bytes to be received.
```
Menduist, line 62:
```suggestion
    # for bytes to be received. Note that it will return as soon as it receive bytes, even if the message is
    # not complete.
```

Or maybe I would use readExactly here, actually?

dryajov: `readExactly` will throw if not all bytes have been consumed or hang if not enough have arrived. You usually want to use `readExactly` only when you know exactly how many bytes are coming, for example with fixed chunk sizes or length prefixed messages (a la `readLp/writeLp`)

emizzle: In tutorial 4, we aren't using `readLp/writeLp` though. And sometimes we send 16 bytes (`please send data`), sometimes 6 bytes (eg `Part 1` ... `Part 9`), and sometimes 7 bytes (eg `Part 10`). So the length isn't exactly known, right? Changing `readOnce` to `readExactly` gets this in the output:
```
[server] received from client: please send data
[client] received from server: Part 0Part 1Part
[client] received from server:  2Part 3Part 4Pa
[client] received from server: rt 5Part 6Part 7
[client] received from server: Part 8Part 9Part
```
followed by an `LPStreamIncompleteError` exception.

emizzle: That was a cool experiment to see what the difference is between `readOnce` and `readExactly`. Would make a nice tutorial 😂

examples/tutorial_4_streamproto2.md
emizzle: Updated.

examples/tutorial_4_streamproto2.md
emizzle: Good idea, done.

examples/tutorial_4_streamproto2.md
Menduist: Yes, to use readExactly you have to send fixed size messages :)

Could indeed be a nice thing to have a tutorial showing the 3 methods (readOnce, readExactly and readLp) and their usecases/drawbacks (maybe one tutorial for the 3 would be enough, bombarding with  3tutorials for the same concept might be a bit too much)

But we can do that later

However it should be noted that with readOnce, you don't know if you'll get entire messages

eg, if you sleep in the receiver, it will probably receive every message at once, I think?

arnetheduck: sleeping doesn't guarantee that you get all data - with readOnce, you roughly get one "buffer" at a time, but there are many buffers of many sizes along the way so you can't tell - when you sleep, those buffers can become full so you'll never get "everything" guaranteed

Menduist: Uh, indeed. Interestingly, in this scenario, since "lower layers" (notably the multiplexer) are packet based, if you call readOnce, you only get one packet, even if there is more bytes queued up.

I hope that nobody builds on top of this assumption, because that's a very shaky and hard to debug one.

I would even be inclined to remove it, by doing this:
```diff
--- a/libp2p/stream/bufferstream.nim
+++ b/libp2p/stream/bufferstream.nim
@@ -128,7 +128,7 @@ method readOnce*(s: BufferStream,
   # First consume leftovers from previous read
   var rbytes = s.readBuf.consumeTo(toOpenArray(p, 0, nbytes - 1))
 
-  if rbytes < nbytes and not s.isEof:
+  while rbytes < nbytes and not s.isEof:
     # There's space in the buffer - consume some data from the read queue
     s.reading = true
     let buf =
@@ -158,6 +158,8 @@ method readOnce*(s: BufferStream,
         trace "add leftovers", s, len = buf.len - remaining
         s.readBuf.add(buf.toOpenArray(remaining, buf.high))
 
+    if s.readQueue.len() == 0: break
+
   if s.isEof and s.readBuf.len() == 0:
     # We can clear the readBuf memory since it won't be used any more
     s.readBuf = StreamSeq()
```

Which gives for the tutorial, on my machine, in this day and age:
```
[server] received from client: please send data
[client] received from server: Part 0Part 1
[client] received from server: Part 2
[client] received from server: Part 3Part 4
[client] received from server: Part 5
[client] received from server: Part 6Part 7
[client] received from server: Part 8
[client] received from server: Part 9Part 10
[client] received from server: 
```
(It's still kinda packet delimited, because the readQueue in the BufferStream is packet based instead of just a stream of bytes)

Thankfully I think almost everyone uses the length prefixed variants, so it shouldn't break anything, but since this kind of bugs are a real plague: very timing dependent and random

And while we can "guarantee" that with mplex, readOnce will give you a packet, it will not be the case with other multiplexers

dryajov: > eg, if you sleep in the receiver, it will probably receive every message at once, I think?

Not really, as noted by @arnetheduck you get roughly one buffer in flight at any given time and with back pressure, you shouldn't end up in a situation where buffers overflow. It is true that this buffers are usually message based, because this is how mplex (and most other libp2p protocols) work, but you still get backpressure at the message boundaries. In other words (and you should be able to verify this), the writer won't be able to send any more data unless the read has picked up the last message from the stream.

I'm not sure what issues you're running in the above reply, but the changes don't seem correct, can you describe the behavior a little better?

examples/tutorial_2_customproto.md
```diff
@@ -22,29 +22,35 @@ type TestProto = ref object of LPProtocol
...
 proc new(T: typedesc[TestProto]): T =
-  # every incoming connections will in be handled in this closure
-  proc handle(conn: Connection, proto: string) {.async, gcsafe.} =
-    echo "Got from remote - ", string.fromBytes(await conn.readLp(1024))
-    # We must close the connections ourselves when we're done with it
-    await conn.close()
+  # every incoming connections will be handled in this closure
+  proc handle(stream: Connection, proto: string) {.async, gcsafe.} =
+    echo "[server] received from client: ", string.fromBytes(await stream.readLp(1024))
+    await stream.writeLp("Hi back!")
```
dryajov, line 33: It's probably a good idea to wrap it in a `try/except/finally`, to acquaint the reader with best practices, etc...

dryajov:
Is there anything blocking this? LGTM

PR #665, opened 2021-11-29
Customizable gossipsub backoff on unsubscribe

Menduist:
This is an alternative to https://github.com/status-im/nimbus-eth2/pull/3025

It allows applications to choose a different backoff when we unsubcribe from a gossipsub topic.

Currently, if you unsubscribe - resubscribe from a topic in less than `backoff` seconds, all of your peers that were previously in your mesh are in your backoff, and won't allow you to graft them back. Setting the `unsubcribeBackoff` to a lower value allows you to alleviate this issue.

Note that it should still be >0 seconds, or some implementations (including ours) will ignore it. However, if it's less than `BackoffSlackTime` (2 seconds), it will effectively be a 0 seconds backoff.
It should not be an issue to have a 0 seconds backoff, since you're unsubscribed and peers should not attempt to graft you until you resub anyway. But maybe to avoid "race conditions", it's a good idea to keep a few seconds as backoff, just in case. Worst case scenario, a peer may attempt to graft you before receiving your unsub, and since this is not allowed, we'll prune him again with a bigger backoff.

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -42,11 +42,11 @@ proc pruned*(g: GossipSub,
              p: PubSubPeer,
              topic: string,
              setBackoff: bool = true,
-             backoff: Duration = 0.seconds) {.raises: [Defect].} =
+             backoff = none(Duration)) {.raises: [Defect].} =
   if setBackoff:
     let
       backoffDuration =
-        if backoff.seconds > 0: backoff
+        if isSome(backoff): backoff.get()
         else: g.parameters.pruneBackoff
       backoffMoment = Moment.fromNow(backoffDuration)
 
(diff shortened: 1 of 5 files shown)


libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -38,11 +38,20 @@ proc grafted*(g: GossipSub, p: PubSubPeer, topic: string) {.raises: [Defect].} =
 
     trace "grafted", peer=p, topic
 
-proc pruned*(g: GossipSub, p: PubSubPeer, topic: string, setBackoff: bool = true) {.raises: [Defect].} =
+proc pruned*(g: GossipSub,
+             p: PubSubPeer,
+             topic: string,
+             setBackoff: bool = true,
```
arnetheduck, line 44: `Option[Duration]`?

arnetheduck (approved):

arnetheduck (approved):

zah:
Please don't merge yet until we ship 1.5.5

Menduist:
Unstable will not be merged into master, and thus get bumped in nimbus, for the next few days/weeks, so 1.5.5 won't be affected anyway

Or you have reservation about this PR in general?

zah:
We haven't yet decided on the branching point from `unstable` that will go into the release.

Issue #667, opened 2021-12-03
Add `listenError` callback to switch


emizzle:
Add a `listenError` callback to switch that is called when a transport fails to start. See https://github.com/status-im/nim-libp2p/pull/662 for more information.

Menduist:
Most likely we will go with #693 instead

PR #668, opened 2021-12-07
feat: Add callback for pessimistic transport start

emizzle:
Fixes: #667.

Add switch.listenError callback that is called when a transport fails to start. This is a pessimistic approach meaning that it will fail on the first address that fails to start in the transport, which has been discussed in #662. 

This remains in draft to kick the discussion off if we want to be optimistic inside of the transport.start (a la #661), which can be added in a new commit.


libp2p/transports/mocktransport.nim
```diff
@@ -0,0 +1,59 @@
...
+
+const STARTMOCK_DEFAULT =
+  proc(addrs: seq[MultiAddress]) {.async.} =
+    discard
+
+const STOPMOCK_DEFAULT =
+  proc() {.async.} =
+    discard
+
+const ACCEPTMOCK_DEFAULT =
```
Menduist, line 28: I find this technique a bit weird, but maybe it's just me

Why not just create a mock directly in the test, that inherits from Transport and actually use inheritance instead of theses callback?
(I should note that I'm not a big fan of OOP generally, but since we're already inheriting, we might as well use it)

From what I can tell, we only need an optional exception to raise in start, so you can create a MockTransport that just takes a exception and raise it if it's not nil
Unless you planned to use the mock more thoroughly later?

libp2p/switch.nim
```diff
@@ -284,7 +284,13 @@ proc newSwitch*(peerInfo: PeerInfo,
...
     peerStore: PeerStore.new(),
     dialer: Dialer.new(peerInfo.peerId, connManager, transports, ms, nameResolver),
-    nameResolver: nameResolver)
+    nameResolver: nameResolver,
+    listenError: listenError)
+
+  if switch.listenError.isNil:
+    switch.listenError = proc(e: ref CatchableError): ref CatchableError =
+      error "Failed to start one transport", error = e.msg
+      return nil
```
Menduist, line 293: I would actually fail by default, to keep the same behaviour as right now

libp2p/switch.nim
```diff
@@ -247,19 +251,13 @@ proc start*(s: Switch) {.async, gcsafe.} =
...
-  await allFutures(startFuts)
-
-  for s in startFuts:
-    if s.failed:
-      # TODO: replace this exception with a `listenError` callback. See
-      # https://github.com/status-im/nim-libp2p/pull/662 for more info.
-      raise newException(transport.TransportError,
-        "Failed to start one transport", s.error)
+      try:
+        await t.start(addrs)
```
Menduist, line 255: In the spirit of 661, I would give the callback as parameter to the start, and let the transport handle the errors

Btw, I completely missed that, but right now if an address fail, we'll not clean up properly (if a transport is already started, it will not be stopped). So we need to stop the switch on error

Menduist:
I had another idea, let me know what you think:

We build a `seq[(MultiAddress, CatchableError)]` (or equivalent), and put all failures in it.
Then, at the the of the start, we call the callback with the `seq`. The default one would be `if seq.len > 0: throw`

This allows for more sophisticated logic in the handle, and we just have to call/handle it once.

wdyt?

libp2p/switch.nim
dryajov: We need to pass the `listenError` to the transport, otherwise there is no way of handling this at the correct granularity. Transports can bind many addresses and any one could fail, so it only makes sense to handle it in the transport. 

Maybe, we should even consider setting the callback on the transport itself and allow each transport to use a different callback, then each transport's `start()` would throw or continue depending on what `listenError` returns.

dryajov:
> I had another idea, let me know what you think:
> 
> We build a `seq[(MultiAddress, CatchableError)]` (or equivalent), and put all failures in it. Then, at the the of the start, we call the callback with the `seq`. The default one would be `if seq.len > 0: throw`
> 
> This allows for more sophisticated logic in the handle, and we just have to call/handle it once.
> 
> wdyt?

Seems less flexible than the callback, also see my comment which is related - https://github.com/status-im/nim-libp2p/pull/668#discussion_r764147551

libp2p/switch.nim
dryajov: Hmm, thats a good point, but I think it's orthogonal to this particular issue. 

I would reframe it as - should we treat unhandled addresses as a failure case or not? Having an address that isn't handled by any one transport doesn't seem like a candidate for a failure, but we should definitely warn the user about it.

On the other hand, nothing prevents us from passing a callback to both the transports and the switch, which could be another way of allowing the consumer to handle all potential failure/exceptional scenarios.

libp2p/transports/mocktransport.nim
emizzle: > I find this technique a bit weird, but maybe it's just me

At the end of the day, it's a test, and does not have an impact on the library.

> Why not just create a mock directly in the test, that inherits from Transport

This is quite cumbersome and would need to be done for each test that needs a mock.

> Unless you planned to use the mock more thoroughly later?

That was the more the point, yes. It creates a mock that allows the `start/stop/accept` procs to be injected. We can use this mock to test anything that uses transports with a unit test oriented approach, eg the switch.

Menduist: > At the end of the day, it's a test, and does not have an impact on the library.

Ok, but it's in the libp2p folder, not tests
Right now the limit between the public Api and internal stuff is a bit blurry, but I though you meant this to be a public thing

For instance we have a public mocker for name resolution, with the idea that end applications can also use it for their unit test

Could you just move it to tests folder then, if it's "private"? As long as the filename doesn't start with a "t", it shouldn't be tested

Indeed, if it's just for tests, you can used whatever you want :)

emizzle: > Could you just move it to tests folder then, if it's "private"? As long as the filename doesn't start with a "t", it shouldn't be tested

Moved :)

libp2p/switch.nim
emizzle: Updated.

libp2p/switch.nim
emizzle: > should we treat unhandled addresses as a failure case or not

In the latest commit, I removed addresses from tcptransport that were not handled in start (https://github.com/status-im/nim-libp2p/blob/cb4298afcaf2e225b0b8df5483f636ebf12aa3a1/libp2p/transports/tcptransport.nim#L151). The idea being that after the switch calls `start`, it could compare which addresses are remaining in the transport to know the addresses that have not been handled, then treat as an error if needed. Alternatively, we could filter out unhandlable addresses in the transport constructor, and then the switch could do the comparison at any point (not just after `start`)

emizzle: > We need to pass the listenError to the transport

The latest commit now has support for this. Check out [the tests](https://github.com/status-im/nim-libp2p/pull/668/files#diff-4ce97f016e8c6ea2e9574d1355ff0a2c3657e78e3ee971556ebce4b1f5a94bdeR1040-R1180) for implementation. The two tests only cover fully pessimistic or fully optimistic (where both `switch.listenError` and `transport.listenError` are either pessimistic or optimistic), however it's possible to have partially pessimistic/optimistic where `switch.listenError` is pessimistic and `transport.listenError` is optimistic and vice versa. I just haven't had a chance to add those tests yet.

Menduist:
Maybe I wasn't clear, but the idea would be to keep the callback, just call it once at the end. Pseudocode:
```nim
proc start*(s: Switch) =
  let failures: seq[(MultiAddress, CatchableError)]
  for t in s.transport:
    failures &= await t.start(s.addrs.filterIt(t.handles(it))

  if failures.len > 0:
    let toThrow = callback(failures)
    if toThrow: raise toThrow
```

Pros:
- Allows more sophisticated logic in the callback (for instance, if both ipv4 & ipv6 failed, throw, but swallow error if one of them succeeded. This can be done with our current callback, but not elegantly)
Cons:
- If the first address failed and we want to stop on that one, we'll still start everything before shutting down, since the callback is only called at the end

libp2p/switch.nim
Menduist: Another option would be to simply gives the callback to start, like:

```nim
proc start*(s: Switch, errorHandler = ListenErrorDefault) {.async, gcsafe.} =
[...]
  await t.start(addrs, errorHandler)

[tcptransport.nim]
method start*(
  self: TcpTransport,
  addrs: seq[MultiAddress],
  errorHandler: ListenErrorCallback) {.async.} =
```
wdyt?

Not sure whats more convenient as a user

emizzle:
> Maybe I wasn't clear, but the idea would be to keep the callback, just call it once at the end. Pseudocode:
> 
> ```nim
> proc start*(s: Switch) =
>   let failures: seq[(MultiAddress, CatchableError)]
>   for t in s.transport:
>     failures &= await t.start(s.addrs.filterIt(t.handles(it))
> 
>   if failures.len > 0:
>     let toThrow = callback(failures)
>     if toThrow: raise toThrow
> ```
> 
> Pros:
> 
> * Allows more sophisticated logic in the callback (for instance, if both ipv4 & ipv6 failed, throw, but swallow error if one of them succeeded. This can be done with our current callback, but not elegantly)
>  Cons:
> * If the first address failed and we want to stop on that one, we'll still start everything before shutting down, since the callback is only called at the end

This would require the transport to start all of its addresses optimistically in order to return a seq of exceptions, right? Didn't we want it to be a) pessimistic by default and b) customisable from the caller?

dryajov:
> If the first address failed and we want to stop on that one, we'll still start everything before shutting down, since the callback is only called at the end

Yeah, to me this defeats the purpose of the callback.

emizzle:
This PR now has:
1. support for `switch.listenError` callback
2. support for `transport.listenError` callback, unique *per transport*
3. tests that have fully and partially optimistic/pessimistic cases
4. tests for changes made to tcp and ws transports to support `listenError`
5. transport `addrs` sequence only contains handleable addresses once `start` is called, which has been handled in the `Transport` base class' `start` method.
6. tests for changes made to tcp and ws transports to support filtering of unhandleable addresses

There is one particular use case for pessimism/optimism that should be pointed out as it may or may not be what is expected. When a switch is pessimistic and all its  transports are optimistic, then one might assume that after the first failure occurs in a transport, that the subsequent transport would not be started (the switch is supposed to be pessimistic). This, however, does not occur as the optimism in the transport prevents the `switch.listenError` from being called at all -- the optimistic `transport.listenError` does not return an error, and therefore an error is not re-raised inside of `switch.start`, which means `switch.listenError` will not be called. Is this a use that we want to handle or is this expected behaviour?

libp2p/transports/transport.nim
```diff
@@ -91,17 +130,10 @@ method upgradeOutgoing*(
 
   self.upgrader.upgradeOutgoing(conn)
 
-method handles*(
-  self: Transport,
-  address: MultiAddress): bool {.base, gcsafe.} =
-  ## check if transport supports the multiaddress
-  ##
+method `==`*(
```
dryajov, line 133: Not sure why you're removing `handles`, if we want to have helper operator as it is in this case, we should just provide it as a wrapper to `handles`. One reason is making operators a `method`, I'm not even sure if that makes sense and other languages prohibit this explicitly.

Menduist: I think he just moved it above to be able to use it in start

libp2p/transports/transport.nim
```diff
@@ -21,27 +21,66 @@ logScope:
...
   addrs: seq[MultiAddress]) {.base, async.} =
   ## start the transport
   ##
 
   trace "starting transport on addrs", address = $addrs
-  self.addrs = addrs
+  self.addrs = addrs.filter(proc(ma: MultiAddress): bool =
+    let handled = self.handles(ma)
+    if not handled:
+      trace "Invalid address detected, skipping!", address = ma
```
Menduist, line 82: This should never happen, since we filter in switch.start, no?

libp2p/switch.nim
```diff
@@ -256,21 +258,29 @@ proc start*(s: Switch) {.async, gcsafe.} =
...
-  (ref SwitchListenError)(msg: "Failed to start one transport", parent: parent)
+proc newSwitchListenError*(
+    t: Transport,
+    parent: ref TransportListenError = nil): ref SwitchListenError =
+
+  (ref SwitchListenError)(msg: "Failed to start one transport",
+                          transport: t,
+                          parent: parent)
 
 const ListenErrorDefault =
```
dryajov, line 277: I'm not sure if it makes sense to provide a switch level callback at this point. If every transport is already given the chance to delegate error handling to the user, it's probably better to just asume that if a transport raised after the user inspected the error and decided to propagate it, there is no reason in keeping with the startup sequence. In other words, the switch will always abort if one of the transports failed, but the transport can decide wether the error is fatal or not. 

That way we eliminate the complexity involved in having two sets of callbacks at different levels, the added bonus is that transports can still handle errors individually on a per transport basis, with separate callbacks (or the same one if desired).

IMO, this would be a lot cleaner overall.

libp2p/transports/transport.nim
emizzle: Hmm good point. `tcptransport.start` had logic to skip unhandled addresses, but I guess that can just be removed instead of moving it to the transport base class.

libp2p/transports/transport.nim
emizzle: Yep, it was moved above, but I'll put it back since we don't need to do any filtering in the base class `start` (see comment below) :)

libp2p/switch.nim
emizzle: I definitely agree this will simplify things quite a lot. Will update.

emizzle:
`switch.listenError` has been removed, along with filtering of addresses in transport's base `handles`.

libp2p/transports/transport.nim
dryajov: I originally thought that this was replacing the `handles` method, but I realize now it's something else. Where in the tests are we using this?

dryajov:
I'm still a bit conflicted about how we're approaching this from an API standpoint. 

In our last call @Menduist brought up a good point about passing the callback only to the transports, which doesn't allow us to handle switch specific issues, the main example used was addresses that don't have a transport to handle them. This in turn raises an additional question of what is really a fatal vs a non fatal error and at which point should we detect it? So I have this two questions right now:

1. Should a missing address (or similar) be considered a fatal error?
2. What is the appropriate stage to detect and report this kind of errors?

I don't have a good answer for the first question, but on the second I'm of the opinion that **errors should be detected and reported/handled as soon as possible**. In the case of missing transports, we can detect it a lot earlier than calling `start`, either before instantiating the `Switch` or during instantiation, with an `assert` or similar. This goes for must everything that can be detected early and we shouldn't allow instantiation with incorrect/unwanted states in general.

The second point that @Menduist brought up was in regards to https://github.com/status-im/nim-libp2p/pull/668#discussion_r765708049, which allows passing the callback to `start` instead the constructor. I don't have a strong opinion here, except that passing the callback during instantiation allows us to clearly separate the two stages, for example, a user might instantiate the switch in one palace and hold on to the instance until starting it in a completely different part of the application? Dunno... just my 2c on this one.

emizzle:
> @emizzle we don't need to mock connections, instead we use the BufferStream abstraction for that. Please take a look at `test/helpers.nim` where a wrapper `TestBufferStream` is defined which allows you to set a writer handler. In order to "write" data to the Connection backed by the BufferStream use `await conn.pushData(@[...])`. BufferStream inherits Connection, so it can be already used interchangeably everywhere you need a stream, for more examples take a look at `testmplex.nim` or I can provide some if you need.

Thanks, I'll remove the mock connection and use a `TestBufferStream` instead in the tests. The point of having the mock connection was to attempt to remove as much logic as possible that was not being tested. Using a `TestBufferStream` didn't seem to remove as much logic because of the parent type `BufferStream` still have fleshed out implementations that might be called. But it appears that `BufferStream` is well covered in unit tests, so using it in tests should be less of an issue. In other words, if something in `BufferStream` has broken, then those tests would pick it up.

emizzle:
> I'm still a bit conflicted about how we're approaching this from an API standpoint.
> 
> In our last call @Menduist brought up a good point about passing the callback only to the transports, which doesn't allow us to handle switch specific issues, the main example used was addresses that don't have a transport to handle them. This in turn raises an additional question of what is really a fatal vs a non fatal error and at which point should we detect it? So I have this two questions right now:
> 
> 1. Should a missing address (or similar) be considered a fatal error?
> 2. What is the appropriate stage to detect and report this kind of errors?
> 
> I don't have a good answer for the first question, but on the second I'm of the opinion that **errors should be detected and reported/handled as soon as possible**. In the case of missing transports, we can detect it a lot earlier than calling `start`, either before instantiating the `Switch` or during instantiation, with an `assert` or similar. This goes for must everything that can be detected early and we shouldn't allow instantiation with incorrect/unwanted states in general.

I like the idea of asserting during instantiation for things that are required for the switch and transports to work. If a switch needs one or more transports to function, (and will not function otherwise) we could assert this in the switch constructor. If a transport needs one or more addresses to function (and will not function otherwise), we could assert this in the transport constructor. This handles detecting and reporting errors as soon as possible and it also describes expected behaviour of the switch and transport.

I'm quite possibly missing some context, but it feels like the issue here is whether we should allow a switch to be instantiated without transports or addresses, then allow the consumer to add transports and addresses afterwards as needed, prior to calling start. Are there particular use cases in mind (now or in the future) in which we would be interested in supporting this? If it needs to be supported, maybe we could use a `booldefine` to allow consumers to disable the defect at compilation, at which point they could instantiate a switch and start to build it as needed after instantiation.

emizzle:
> The second point that @Menduist brought up was in regards to [#668 (comment)](https://github.com/status-im/nim-libp2p/pull/668#discussion_r765708049), which allows passing the callback to `start` instead the constructor. I don't have a strong opinion here, except that passing the callback during instantiation allows us to clearly separate the two stages, for example, a user might instantiate the switch in one palace and hold on to the instance until starting it in a completely different part of the application? Dunno... just my 2c on this one.

Agreed, I don't have an entirely strong opinion either direction. It is an interesting thought that if a switch in created in ModuleA and injected in to ModuleB where start is called, then there may be some context missing in ModuleB to reason about an error. However, the way it works right now, `switch.listenError` can be provided during instantiation or overridden at any point, so ModuleB could override `listenError` prior to calling start if needed. This setup seems to provide a fair bit of flexibility. Providing `listenError` as a parameter to `start` would remove that flexibility as well as instantiation context  (eg ModuleA context) in order to reason about an error.

libp2p/transports/transport.nim
emizzle: After removing `swithc.listenError` and associated tests, `transport.==` is no longer used in the tests, so it has been now been removed. Good catch!

emizzle:
[Latest commit](https://github.com/status-im/nim-libp2p/pull/668/commits/d4a65503ac1dbd89e208730261e29671248367e1) adds two things:
1. During switch instantiation, raises defect if no transports or addresses supplied
2. During transport.start, raises defect if no addresses supplied.

I implemented the points discussed above because sometimes it's easier to reason once written out.

emizzle:
Anyone know why CI is failing for `testnative` on Linux? I've run this in a Ubuntu 20.04.3 VM using nim `1.4.8` and `1.2.16` and had no failures...

Menduist:
Can't repro locally either

emizzle:
All tests passing.

libp2p/errors.nim
```diff
@@ -9,6 +9,8 @@ type
   # Base exception type for libp2p
   LPError* = object of CatchableError
 
+  LPDefect* = object of CatchableError
```
Menduist, line 12: I find this a bit confusing
I get the idea of "a libp2p non recoverable error may not be an app non recoverable error, so let's make a fake defect"

But that's more confusing than anything, imo

libp2p/switch.nim
```diff
@@ -234,9 +235,8 @@ proc stop*(s: Switch) {.async.} =
 
   trace "Switch stopped"
 
-proc start*(s: Switch) {.async, gcsafe.} =
+proc start*(s: Switch) {.async, gcsafe, raises: [Defect, TransportListenError].} =
```
Menduist, line 238: Async raises are not supported ATM, and adding them is more error inducing than anything else

libp2p/transports/tcptransport.nim
```diff
@@ -118,18 +118,20 @@ proc connHandler*(self: TcpTransport,
...
+    upgrader: upgrade,
+    listenError: listenError)
 
   inc getTcpTransportTracker().opened
   return transport
 
 method start*(
   self: TcpTransport,
-  addrs: seq[MultiAddress]) {.async.} =
+  addrs: seq[MultiAddress]) {.async, raises: [Defect, TransportListenError].} =
```
Menduist, line 134: Ditto

libp2p/transports/transport.nim
```diff
@@ -21,26 +21,52 @@ logScope:
...
+      ma: MultiAddress,
+      err: ref CatchableError): Future[ref TransportListenError] {.async.} =
+
+    return newTransportListenError(ma, err)
+
 method start*(
   self: Transport,
-  addrs: seq[MultiAddress]) {.base, async.} =
+  addrs: seq[MultiAddress])
+  {.base, async, raises: [Defect, TransportDefect, TransportListenError].} =
```
Menduist, line 61: Ditto

libp2p/transports/wstransport.nim
```diff
@@ -90,7 +90,7 @@ proc secure*(self: WsTransport): bool =
 
 method start*(
   self: WsTransport,
-  addrs: seq[MultiAddress]) {.async.} =
+  addrs: seq[MultiAddress]) {.async, raises: [Defect, TransportListenError].} =
```
Menduist, line 93: Ditto

libp2p/errors.nim
emizzle: I totally agree, will change it use `LPError` instead. Not sure what the line of reasoning was here. With the callbacks, the calling applications can re-raise a defect if needed.

libp2p/switch.nim
emizzle: Good catch, thanks for picking that up. The raises pragma has been removed and a TODO put in its place.

libp2p/transports/tcptransport.nim
emizzle: Ditto 😄

libp2p/transports/transport.nim
emizzle: Ditto 😄

libp2p/transports/wstransport.nim
emizzle: Ditto 😄

libp2p/switch.nim
```diff
@@ -274,6 +263,13 @@ proc newSwitch*(peerInfo: PeerInfo,
                 ms: MultistreamSelect,
                 nameResolver: NameResolver = nil): Switch
                 {.raises: [Defect, LPError].} =
+
+  if transports.len == 0:
+    raise newException(LPError, "Provide at least one transport")
+
+  if peerInfo.addrs.len == 0:
+    raise newException(LPError, "Provide at least one address")
```
Menduist, line 271: imo, starting a switch without address is valid (outbound only)

Menduist:
I'll admit, I'm still not pleased with the state of this PR.
We still don't raise an error when an address is not handled by any transport, and that would imply yet-another-callback

I don't find it very user friendly, you have to make a lot of different callbacks without clear benefits.

ATM, I don't have anything better to propose (outside what I've already proposed here), so we may have to brainstorm a bit

Menduist:
We brainstormed a bit with @dryajov and this is what we came up with: https://github.com/status-im/nim-libp2p/issues/693
(I think someone proposed that a while ago actually)

I think it's more elegant and powerful than the handler approach, sorry that we launched you on the wrong path

richard-ramos:
Closing this PR as it is stale. @emizzle feel free to reopen, altho it seems based on the discussion that #693 superseded it

PR #670, opened 2021-12-09
Circuit relay v1

lchenut:
A simple way to make it work at the moment:
```nim
proc main() {.async, gcsafe.} =
  let
    rng = newRng()
    src = newStandardSwitch(rng=rng)
    rel = newStandardSwitch(rng=rng)
    address = MultiAddress.init("/ip4/127.0.0.1/tcp/0").tryGet()
    addressRelay = MultiAddress.init("/ip4/127.0.0.1/tcp/0/p2p-circuit").tryGet()
    dst = SwitchBuilder
      .new()
      .withRng(rng)
      .withAddresses(@[ address, addressRelay ])
      .withTcpTransport()
      .withMplex()
      .withNoise()
      .build()
    pingProtocol = Ping.new(rng=rng)
    relaySrc = Relay.new(src, false)
    relayRel = Relay.new(rel, true)
    relayDst = Relay.new(dst, false)

  src.mount(relaySrc)
  rel.mount(relayRel)
  dst.mount(relayDst)
  dst.mount(pingProtocol)
  src.transports &= RelayTransport.new(relaySrc)
  ((Dialer)src.dialer).transports = src.transports
  dst.transports &= RelayTransport.new(relayDst)
  ((Dialer)dst.dialer).transports = dst.transports

  let
    srcFut = await src.start()
    relFut = await rel.start()
    dstFut = await dst.start()

  let ma = MultiAddress.init($rel.peerInfo.addrs[0] & "/p2p/" & $rel.peerInfo.peerId & "/p2p-circuit/p2p/" & $dst.peerInfo.peerId).tryGet()
  await src.connect(rel.peerInfo.peerId, rel.peerInfo.addrs)
  await rel.connect(dst.peerInfo.peerId, dst.peerInfo.addrs)
  let conn = await src.dial(dst.peerInfo.peerId, @[ ma ], PingCodec)

  echo "ping: ", await pingProtocol.ping(conn)

  await allFutures(src.stop(), rel.stop(), dst.stop())
  await allFutures(srcFut & relFut & dstFut)
waitFor(main())
```
Few notes:
- The destination having two multiaddresses is clearly undesirable but it's a temporary workaround because without circuit-relay address, the transport won't start. 
- Speaking of temporary workaround, Dial.transports becoming public is one.
- It may be interesting to create an access control list as it is implemented in [the go-libp2p version](https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/circuitv1/relay/relay.go#L133-L137) even though there's no mention of it in the specs. It's relatively easy to implement.

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -14,7 +14,7 @@ import
   switch, peerid, peerinfo, stream/connection, multiaddress,
   crypto/crypto, transports/[transport, tcptransport],
   muxers/[muxer, mplex/mplex],
-  protocols/[identify, secure/secure, secure/noise],
+  protocols/[identify, secure/secure, secure/noise, relay],
   connmanager, upgrademngrs/muxedupgrade,
   nameresolving/nameresolver,
   errors
@@ -48,6 +48,8 @@ type
     protoVersion: string
     agentVersion: string
     nameResolver: NameResolver
+    isCircuitRelay: bool
+    circuitRelayCanHop: bool
 
 proc new*(T: type[SwitchBuilder]): T =
 
@@ -64,7 +66,8 @@ proc new*(T: type[SwitchBuilder]): T =
     maxOut: -1,
     maxConnsPerPeer: MaxConnectionsPerPeer,
     protoVersion: ProtoVersion,
-    agentVersion: AgentVersion)
+    agentVersion: AgentVersion,
+    isCircuitRelay: false)
 
 proc withPrivateKey*(b: SwitchBuilder, privateKey: PrivateKey): SwitchBuilder =
   b.privKey = some(privateKey)
@@ -139,6 +142,11 @@ proc withNameResolver*(b: SwitchBuilder, nameResolver: NameResolver): SwitchBuil
   b.nameResolver = nameResolver
   b
 
+proc withRelayTransport*(b: SwitchBuilder, canHop: bool): SwitchBuilder =
+  b.isCircuitRelay = true
+  b.circuitRelayCanHop = canHop
+  b
+
 proc build*(b: SwitchBuilder): Switch
   {.raises: [Defect, LPError].} =
 
@@ -196,6 +204,11 @@ proc build*(b: SwitchBuilder): Switch
     connManager = connManager,
     ms = ms,
     nameResolver = b.nameResolver)
+
+  if b.isCircuitRelay:
+    let relay = Relay.new(switch, b.circuitRelayCanHop)
+    switch.mount(relay)
+    switch.addTransport(RelayTransport.new(relay, muxedUpgrade))
 
   return switch
 
--- a/libp2p/protocols/relay.nim
+++ b/libp2p/protocols/relay.nim
@@ -1,0 +1,487 @@
+## Nim-LibP2P
+## Copyright (c) 2022 Status Research & Development GmbH
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+{.push raises: [Defect].}
+
+import options
+import sequtils, strutils, tables
+import chronos, chronicles
+
+import ../peerinfo,
+       ../switch,
+       ../multiaddress,
+       ../stream/connection,
+       ../protocols/protocol,
+       ../transports/transport,
+       ../utility,
+       ../errors
+
+const
+  RelayCodec* = "/libp2p/circuit/relay/0.1.0"
+  MsgSize* = 4096
+  MaxCircuit* = 1024
+  MaxCircuitPerPeer* = 64
+
+logScope:
+  topics = "libp2p relay"
+
+type
+  RelayType* = enum
+    Hop = 1
+    Stop = 2
+    Status = 3
+    CanHop = 4
+  RelayStatus* = enum
+    Success = 100
+    HopSrcAddrTooLong = 220
+    HopDstAddrTooLong = 221
+    HopSrcMultiaddrInvalid = 250
+    HopDstMultiaddrInvalid = 251
+    HopNoConnToDst = 260
+    HopCantDialDst = 261
+    HopCantOpenDstStream = 262
+    HopCantSpeakRelay = 270
+    HopCantRelayToSelf = 280
+    StopSrcAddrTooLong = 320
+    StopDstAddrTooLong = 321
+    StopSrcMultiaddrInvalid = 350
+    StopDstMultiaddrInvalid = 351
+    StopRelayRefused = 390
+    MalformedMessage = 400
+
+  RelayError* = object of LPError
+
+  RelayPeer* = object
+    peerId*: PeerID
+    addrs*: seq[MultiAddress]
+
+  AddConn* = proc(conn: Connection): Future[void] {.gcsafe, raises: [Defect].}
+
+  RelayMessage* = object
+    msgType*: Option[RelayType]
+    srcPeer*: Option[RelayPeer]
+    dstPeer*: Option[RelayPeer]
+    status*: Option[RelayStatus]
+
+  Relay* = ref object of LPProtocol
+    switch*: Switch
+    peerId: PeerID
+    dialer: Dial
+    canHop: bool
+    streamCount: int
+    hopCount: CountTable[PeerId]
+
+    addConn: AddConn
+
+    maxCircuit*: int
+    maxCircuitPerPeer*: int
+    msgSize*: int
+
+proc encodeMsg*(msg: RelayMessage): ProtoBuffer =
+  result = initProtoBuffer()
+
+  if isSome(msg.msgType):
+    result.write(1, msg.msgType.get().ord.uint)
+  if isSome(msg.srcPeer):
+    var peer = initProtoBuffer()
+    peer.write(1, msg.srcPeer.get().peerId)
+    for ma in msg.srcPeer.get().addrs:
+      peer.write(2, ma.data.buffer)
+    peer.finish()
+    result.write(2, peer.buffer)
+  if isSome(msg.dstPeer):
+    var peer = initProtoBuffer()
+    peer.write(1, msg.dstPeer.get().peerId)
+    for ma in msg.dstPeer.get().addrs:
+      peer.write(2, ma.data.buffer)
+    peer.finish()
+    result.write(3, peer.buffer)
+  if isSome(msg.status):
+    result.write(4, msg.status.get().ord.uint)
+
+  result.finish()
+
+proc decodeMsg*(buf: seq[byte]): Option[RelayMessage] =
+  var
+    rMsg: RelayMessage
+    msgTypeOrd: uint32
+    src: RelayPeer
+    dst: RelayPeer
+    statusOrd: uint32
+    pbSrc: ProtoBuffer
+    pbDst: ProtoBuffer
+
+  let
+    pb = initProtoBuffer(buf)
+    r1 = pb.getField(1, msgTypeOrd)
+    r2 = pb.getField(2, pbSrc)
+    r3 = pb.getField(3, pbDst)
+    r4 = pb.getField(4, statusOrd)
+
+  if r1.isErr() or r2.isErr() or r3.isErr() or r4.isErr():
+    return none(RelayMessage)
+
+  if r2.get() and
+     (pbSrc.getField(1, src.peerId).isErr() or
+      pbSrc.getRepeatedField(2, src.addrs).isErr()):
+    return none(RelayMessage)
+
+  if r3.get() and
+     (pbDst.getField(1, dst.peerId).isErr() or
+      pbDst.getRepeatedField(2, dst.addrs).isErr()):
+    return none(RelayMessage)
+
+  if r1.get(): rMsg.msgType = some(RelayType(msgTypeOrd))
+  if r2.get(): rMsg.srcPeer = some(src)
+  if r3.get(): rMsg.dstPeer = some(dst)
+  if r4.get(): rMsg.status = some(RelayStatus(statusOrd))
+  some(rMsg)
+
+proc sendStatus*(conn: Connection, code: RelayStatus) {.async, gcsafe.} =
+  trace "send status", status = $code & "(" & $ord(code) & ")"
+  let
+    msg = RelayMessage(
+      msgType: some(RelayType.Status),
+      status: some(code))
+    pb = encodeMsg(msg)
+
+  await conn.writeLp(pb.buffer)
+
+proc handleHopStream(r: Relay, conn: Connection, msg: RelayMessage) {.async, gcsafe.} =
+  r.streamCount.inc()
+  defer:
+    r.streamCount.dec()
+
+  if r.streamCount > r.maxCircuit:
+    trace "refusing connection; too many active circuit"
+    await sendStatus(conn, RelayStatus.HopCantSpeakRelay)
+    return
+
+  proc checkMsg(): Result[RelayMessage, RelayStatus] =
+    if not r.canHop:
+      return err(RelayStatus.HopCantSpeakRelay)
+    if msg.srcPeer.isNone:
+      return err(RelayStatus.HopSrcMultiaddrInvalid)
+    let src = msg.srcPeer.get()
+    if src.peerId != conn.peerId:
+      return err(RelayStatus.HopSrcMultiaddrInvalid)
+    if msg.dstPeer.isNone:
+      return err(RelayStatus.HopDstMultiaddrInvalid)
+    let dst = msg.dstPeer.get()
+    if dst.peerId == r.switch.peerInfo.peerId:
+      return err(RelayStatus.HopCantRelayToSelf)
+    if not r.switch.isConnected(dst.peerId):
+      trace "relay not connected to dst", dst
+      return err(RelayStatus.HopNoConnToDst)
+    ok(msg)
+
+  let check = checkMsg()
+  if check.isErr:
+    await sendStatus(conn, check.error())
+    return
+  let
+    src = msg.srcPeer.get()
+    dst = msg.dstPeer.get()
+
+  # TODO: if r.acl # access control list
+  #       and not r.acl.AllowHop(src.peerId, dst.peerId)
+  #         sendStatus(conn, RelayStatus.HopCantSpeakRelay)
+
+  r.hopCount.inc(src.peerId)
+  r.hopCount.inc(dst.peerId)
+  defer:
+    r.hopCount.inc(src.peerId, -1)
+    r.hopCount.inc(dst.peerId, -1)
+
+  if r.hopCount[src.peerId] > r.maxCircuitPerPeer:
+    trace "refusing connection; too many connection from src", src, dst
+    await sendStatus(conn, RelayStatus.HopCantSpeakRelay)
+    return
+
+  if r.hopCount[dst.peerId] > r.maxCircuitPerPeer:
+    trace "refusing connection; too many connection to dst", src, dst
+    await sendStatus(conn, RelayStatus.HopCantSpeakRelay)
+    return
+
+  let connDst = try:
+    await r.switch.dial(dst.peerId, @[RelayCodec])
+  except CancelledError as exc:
+    raise exc
+  except CatchableError as exc:
+    trace "error opening relay stream", dst, exc=exc.msg
+    await sendStatus(conn, RelayStatus.HopCantDialDst)
+    return
+  defer:
+    await connDst.close()
+
+  let msgToSend = RelayMessage(
+    msgType: some(RelayType.Stop),
+    srcPeer: some(src),
+    dstPeer: some(dst),
+    status: none(RelayStatus))
+
+  let msgRcvFromDstOpt = try:
+    await connDst.writeLp(encodeMsg(msgToSend).buffer)
+    decodeMsg(await connDst.readLp(r.msgSize))
+  except CancelledError as exc:
+    raise exc
+  except CatchableError as exc:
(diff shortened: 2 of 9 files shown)


libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+
+  r.streamCount += 1
+  r.hopCount.inc(src.peerId)
+  r.hopCount.inc(dst.peerId)
+  defer:
+    r.streamCount -= 1
+    r.hopCount.inc(src.peerId, -1)
+    r.hopCount.inc(dst.peerId, -1)
+
+  if r.streamCount > r.maxCircuit:
```
dryajov, line 186: This should probably be at the top, before we parse the message and all that.

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+
+method accept*(self: RelayTransport): Future[Connection] {.async, gcsafe.} =
+  result = await self.relay.queue.popFirst()
+
+# TODO: dial / handles active relay:
+# /ipv4/127.0.0.1/tcp/5002/p2p/QmRelay/p2p-circuit/ipv4/127.0.0.1/tcp/5003/p2p/QmTarget
+# dstPeerId = QmTarget
+# dstAddrs = /ipv4/127.0.0.1/tcp/5003
+#
+# TODO: Must find a way to solve the handles problem, currently the dst switch is created
```
dryajov, line 431: Not exactly sure what the problem is, it's should be totally valid to start with the `/p2p-circuit` address, that might be another way of announcing circuit capabilities in the future.

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+  result = await self.relay.dialPeer(conn, dstPeerId, @[])
+
+method dial*(
+  self: RelayTransport,
+  hostname: string,
+  address: MultiAddress): Future[Connection] {.async, gcsafe.} =
+  result = await self.dial(address)
+
+method handles*(self: RelayTransport, ma: MultiAddress): bool {.gcsafe} =
+  result=false
```
dryajov, line 460: `bools` are false by default

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+  if ma.protocols.isOk:
+    let sma = toSeq(ma.items())
+    if sma.len >= 3:
+      result=(CircuitRelay.match(sma[^2].get()) and P2PPat.match(sma[^1].get())) or CircuitRelay.match(sma[^1].get())
+  trace "Handles return", ma, result
+
+proc new*(T: typedesc[RelayTransport], relay: Relay): T =
+  T(relay: relay, upgrader: relay.switch.transports[0].upgrader)
```
dryajov, line 468: There isn't any need to do this, you can safelly override `upgradeIncoming` and `upgradeOutgoing` to do nothing and it would work as expected.

dryajov (changes requested):
Hey @lchenut! Thank you so much for taking the time to work on this! Great job and appologies for taking this long to respond from my side!

So far everything looks pretty good sans a few minor things I've already commented on. In any case I'll give it another pass soon.

One crussial thing that is missing are interop tests. We have examples of how we do this in the `testinterop.nim` file, we run them against a go libp2p daemon instance. It's very usefull in catching any inconsistencies and validate how well supported the protocol. Let me know if you need any help navigating it.

Cheers!

tests/testrelay.nim
```diff
@@ -0,0 +1,242 @@
...
+    srcFuture {.threadVar.}: seq[Future[void]]
+    dstFuture {.threadVar.}: seq[Future[void]]
+    relFuture {.threadVar.}: seq[Future[void]]
+    conn {.threadVar.}: Connection
+    msg {.threadVar.}: ProtoBuffer
+    rcv {.threadVar.}: Option[RelayMessage]
+    pingProtocol {.threadVar.}: Ping
+
+  proc createMsg(
+    msgType: Option[RelayType] = none[RelayType](),
```
dryajov, line 37: The idiomatic way to writ this would be `msgType = RelayType.none`

libp2p/protocols/relay.nim
Menduist: Edit: we actually need to upgrade, though adding support in `builders.nim` is a better solution

libp2p/protocols/relay.nim
Menduist: Actually, I just re-read this, and I think the issue is that you shouldn't supply the addresses manually at start, since you don't know which relay will relay to you

So I think we shouldn't have any p2p-circuit address in our multiaddress list (until we handle the reservations from circuit relay v2), and should work as-is?

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+
+  result.finish()
+
+proc decodeMsg*(buf: seq[byte]): Option[RelayMessage] =
+  var
+    rMsg: RelayMessage
+    msgTypeOrd: uint32
+    src: RelayPeer
+    dst: RelayPeer
+    status: RelayStatus
```
Menduist, line 107: Unused?

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+    statusOrd: uint32
+    pb = initProtoBuffer(buf)
```
Menduist, line 109: Can be moved to the let

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+    if r6.get():
+      rMsg.status = some(RelayStatus(statusOrd))
+    some(rMsg)
+  else:
+    none[RelayMessage]()
+
+proc sendStatus*(conn: Connection, code: RelayStatus) {.async, gcsafe.} =
+  trace "send status", status = $code & "(" & $ord(code) & ")"
+
+  var
```
Menduist, line 137: Can be a let

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+    return
+  let
+    src = msg.srcPeer.get()
+    dst = msg.dstPeer.get()
+
+  # TODO: if r.acl # access control list?
+  #       and not r.acl.AllowHop(src.peerId, dst.peerId)
+  #         sendStatus(conn, RelayStatus.HopCantSpeakRelay)
+
+  r.streamCount += 1
```
Menduist, line 178:
```suggestion
  r.streamCount.inc()
```

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+  r.hopCount.inc(src.peerId)
+  r.hopCount.inc(dst.peerId)
+  defer:
+    r.streamCount -= 1
+    r.hopCount.inc(src.peerId, -1)
```
Menduist, line 183: There should be a dec, no?

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+    return
+
+  trace "get a relay connection", src, conn
+
+  await sendStatus(conn, RelayStatus.Success)
+  await r.queue.addLast(conn)
+  await conn.join()
+
+proc handleCanHop(r: Relay, conn: Connection, msg: RelayMessage) {.async, gcsafe.} =
+  if r.canHop:
```
Menduist, line 311:
```nim
await sendStatus(conn,
  if r.canHop:
    RelayStatus.Success
  else:
    RelayStatus.HopCantSpeakRelay
)
```

Less code duplication

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+    raise newException(RelayError, "Hop can't open destination stream")
+  result = conn
+
+#
+# Relay Transport
+#
+
+const
+  CircuitRelay = mapEq("p2p-circuit")
+  P2PPat = mapEq("p2p")
```
Menduist, line 404: We can move that to multiaddress.nim

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+method dial*(self: RelayTransport, ma: MultiAddress): Future[Connection] {.async, gcsafe.} =
+  let
+    sma = toSeq(ma.items())
+    relayAddrs = sma[0..sma.len-4].mapIt(it.tryGet()).foldl(a & b)
+  var
+    relayPeerId: PeerId
+    dstPeerId: PeerId
+  if not relayPeerId.init(($(sma[^3].get())).split('/')[2]):
+    # error
+    discard
```
Menduist, line 444: This will not do anything? Better to do something with the error

Menduist (commented):
Some nitpicks

libp2p/protocols/relay.nim
dryajov: Ah, I see - yes, we might allow an empty/no addresses at start, that shouldn't be a big issue.

libp2p/protocols/relay.nim
dryajov: Ah, you're right, we need to upgrade the connection and correct, we can just pass it as a parameter. 

I wonder if we can pass the `transports` array as well without having to expose it in the switch - we should be able to too.

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,471 @@
...
+
+  try:
+    await sendStatus(conn, RelayStatus.Success)
+  except:
+    trace "error writing relay response", error=getCurrentExceptionMsg()
+    return
+
+  trace "relaying connection", src, dst
+
+  var
```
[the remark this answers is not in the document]
lchenut: The buffers have to be initialized outside the loop, the reads that have a reference to these buffers are outside the scope of the loop. The best I can do for the moment is to create a separate bridge procedure, to use the buffers outside the main function.

Eventually, a good solution could be to use (if it exists) or create (if not) a bridge function from chronos, but I'm scared of this library right now.

lchenut:
@dryajov @Menduist, I corrected all the errors / nitpicks and I added the interop tests. The only thing I left open is the dirty buffers in the bridge part of the relay as I wanted your advice on this point.
The new way to make it works could be this:
```nim
proc main() {.async, gcsafe.} =
  let
    CustomProtoCodec = "/customProto"
    rng = newRng()
    src = SwitchBuilder.new()
      .withRng(rng)
      .withAddresses(@[ MultiAddress.init("/ip4/127.0.0.1/tcp/0").tryGet() ])
      .withTcpTransport()
      .withMplex()
      .withNoise()
      .withRelayTransport(canHop=false)
      .build()
    rel = SwitchBuilder.new()
      .withRng(rng)
      .withAddresses(@[ MultiAddress.init("/ip4/127.0.0.1/tcp/0").tryGet() ])
      .withTcpTransport()
      .withMplex()
      .withNoise()
      .withRelayTransport(canHop=true)
      .build()
    dst = SwitchBuilder.new()
      .withRng(rng)
      .withAddresses(@[ MultiAddress.init("/ip4/127.0.0.1/tcp/0").tryGet() ])
      .withTcpTransport()
      .withMplex()
      .withNoise()
      .withRelayTransport(canHop=false)
      .build()

  # Create a custom protocol
  var proto = new LPProtocol
  proto.handler = proc (conn: Connection, customProto: string) {.async.} =
    echo "Destination Read From Source: ", string.fromBytes(await conn.readLP(1024))
    await conn.writeLp("I'm out")
    await conn.close()
  proto.codec = CustomProtoCodec

  dst.mount(proto)

  await src.start()
  await rel.start()
  await dst.start()

  let ma = MultiAddress.init($rel.peerInfo.addrs[0] & "/p2p/" & $rel.peerInfo.peerId & "/p2p-circuit/p2p/" & $dst.peerInfo.peerId).tryGet()
  # /ip4/.../tcp/.../p2p/QmRelay/p2p-circuit/p2p/QmDestination
  await src.connect(rel.peerInfo.peerId, rel.peerInfo.addrs)
  await rel.connect(dst.peerInfo.peerId, dst.peerInfo.addrs)
  let conn = await src.dial(dst.peerInfo.peerId, @[ ma ], CustomProtoCodec)

  await conn.writeLp("Hello")
  echo "Source Read From Destination: ", string.fromBytes(await conn.readLp(1024))

  await allFutures(src.stop(), rel.stop(), dst.stop())
waitFor(main())
```

Few more points:
- [This test](https://github.com/lchenut/nim-libp2p/blob/f6d6a5c630d645615d5ba6f7948b4dfaa9c88d03/tests/testrelay.nim#L290-L340) is really redundant with the previous one, I was originally only testing the withRelayTransport of the builder and decide to keep it (more tests is (nearly) always better)
- Speaking of the builder, [this part](https://github.com/lchenut/nim-libp2p/blob/f6d6a5c630d645615d5ba6f7948b4dfaa9c88d03/libp2p/builders.nim#L208-L212) is really dirty (still cleaner than the old version though) and hiding the Relay behind the switch prevents us to change the three configurable variables (number maximum of circuit, number maximum of circuit per peer and message size)
- There's always this ACL thing, not mentioned in the spec, that [the go version does](https://github.com/libp2p/go-libp2p/blob/0e72b4832bb3949d15291aa0ba7883040691a24a/p2p/protocol/circuitv1/relay/relay.go#L180-L184)
- [This is my bad, I fix it now](https://github.com/lchenut/nim-libp2p/blob/f6d6a5c630d645615d5ba6f7948b4dfaa9c88d03/libp2p/protocols/relay.nim#L489-L493)

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,491 @@
...
+  except CatchableError as exc:
+    trace "error writing relay response", exc=exc.msg
+    return
+
+  trace "relaying connection", src, dst
+
+  # TODO: Eventually create a similar procedure somewhere in the chronos library
+  proc bridge(conn: Connection, connDst: Connection) {.async.} =
+    var
+      bufSrcToDst: array[4096, byte]
```
dryajov, line 282: I would use the `{.noInit.}` pragma here and then `zeroMem` every time I finished using the buffer in the loop.

Menduist: Not sure I understand the issue here, this code can't leak any data unless `read` returns a wrong value, which would be catastrophic anyway
And even if a bit of the buffer was somehow reused, that would be in the same relayed connection, so it would only resend some old data to the same peer

And even if all of that happened, since the connection from A to B through R is secured anyway, you wouldn't be able to extract any useful data? R could post the entire exchange on Facebook and nothing bad would happen

The only thing we have to be careful about is to not leak our internal memory, so we must be careful to zero mem the buffers before any usage (which is done implicitly right now)

lchenut: After a little thought and more deeper inspection on how pipes work kernel-side (with a buffer, more or less the same thing I did) I must agree with @Menduist, there's not really a better solution.

A `const bufferSize = 4096` or `bufSrcToDst.high + 1` are good ways to avoid the repeat of the 4096 that could be problematic if someone updates this section of the code.

And I understand the use of the `zeroMem` at the end of the loop, even though, like Menduist pointed out, the connection is secured so it's a bit superfluous, but we never know what can happen. The `{.noInit.}` pragma on another hand is less effective to secure this part, if the code fails during the first loop, it could be dangerous with the buffer uninitialized.

lchenut:
@dryajov @Menduist I fixed the few minor errors I mentioned at the last meeting. If you could review the code that would be great.

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,500 @@
...
+    res: bool
+
+  let
+    pb = initProtoBuffer(buf)
+    r1 = pb.getField(1, msgTypeOrd)
+    r2 = pb.getField(2, pbSrc)
+    r3 = pb.getField(3, pbDst)
+    r4 = pb.getField(4, statusOrd)
+
+  res = r1.isOk() and r2.isOk() and r3.isOk() and r4.isOk()
```
Menduist, line 125: Why not

```nim
if not res: return none(RelayMessage)
```

And the follow ifs can be just `if r2.get()`

And no need for the last one

Menduist: Actually, I would do something like:
```nim
let res = r1.isOk() and r2.isOk() and r3.isOk() and r4.isOk()
if not res: return none(RelayMessage)

if r2.get() and
   pbSrc.getField(1, sec.peerId) == false or
   pbSrc.getRepeatedField(2, sec.addrs) == false:
     return none(RelayMessage)

[..]
some(
```
Or something similar

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,500 @@
...
+  else:
+    RelayMessage.none
+
+proc sendStatus*(conn: Connection, code: RelayStatus) {.async, gcsafe.} =
+  trace "send status", status = $code & "(" & $ord(code) & ")"
+  let
+    msg = RelayMessage(
+      msgType: some(RelayType.Status),
+      srcPeer: none(RelayPeer),
+      dstPeer: none(RelayPeer),
```
Menduist, line 158: none is the default value, no need to specify them I think

libp2p/protocols/relay.nim
lchenut: To be fair, it was one of the first function I wrote and I was (maybe?) influenced by the code of "identify" at that time.
I could change it, but I'll change the identify decodeMsg function aswell.

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,500 @@
...
+    return
+
+  let msgRcvFromDstOpt = try:
+    decodeMsg(await connDst.readLp(r.msgSize))
+  except CancelledError as exc:
+    raise exc
+  except CatchableError as exc:
+    trace "error reading stop response", exc=exc.msg
+    await sendStatus(conn, RelayStatus.HopCantOpenDstStream)
+    return
```
Menduist, line 252: Maybe this could be a single try / except, instead of 3

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,500 @@
...
+
+  if msgRcvFromDst.status.isNone or msgRcvFromDst.status.get() != RelayStatus.Success:
+    trace "relay stop failure", status=msgRcvFromDst.status
+    await sendStatus(conn, RelayStatus.HopCantOpenDstStream)
+    return
+
+  try:
+    await sendStatus(conn, RelayStatus.Success)
```
Menduist, line 271: Why is this sendStatus in a try / except and not the other ones?

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,500 @@
...
+      let msgOpt = decodeMsg(await conn.readLp(r.msgSize))
+
+      if msgOpt.isNone:
+        await sendStatus(conn, RelayStatus.MalformedMessage)
+        return
+      else:
+        trace "relay handle stream", msg = msgOpt.get()
+      let msg = msgOpt.get()
+
+      if msg.msgType.get() == RelayType.Hop:
```
Menduist, line 368: Can't use a case?

```nim
case msg.msgType.get():
of Hop: ###
```

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,500 @@
...
+    trace "Relay transport already running"
+    return
+
+  await procCall Transport(self).start(ma)
+  trace "Starting Relay transport"
+
+method stop*(self: RelayTransport) {.async, gcsafe.} =
+  self.running = false
+  while not self.relay.queue.empty():
+    await self.relay.queue.popFirstNoWait().close()
```
Menduist, line 461: This looks a bit risky: a peer might "slip in" during this procedure

Could you block the relay somehow before closing the connections?

Actually, it might make sense to move the `queue` to the transport, and add a callback to the Relay instead.
If the callback is nil, the relay refuses the connection, and the callback would be set to nil at the start of stop

And the transport start would hook in the callback to the transport queues

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,500 @@
...
+
+method accept*(self: RelayTransport): Future[Connection] {.async, gcsafe.} =
+  result = await self.relay.queue.popFirst()
+
+# TODO: dial / handles active relay:
+# /ipv4/127.0.0.1/tcp/5002/p2p/QmRelay/p2p-circuit/ipv4/127.0.0.1/tcp/5003/p2p/QmTarget
+# dstPeerId = QmTarget
+# dstAddrs = /ipv4/127.0.0.1/tcp/5003
```
Menduist, line 469: As discussed somewhere, the active relay has been deprecated, so we can safely remove this

libp2p/protocols/relay.nim
```diff
@@ -0,0 +1,500 @@
...
+proc dial*(self: RelayTransport, ma: MultiAddress): Future[Connection] {.async, gcsafe.} =
+  let
+    sma = toSeq(ma.items())
+    relayAddrs = sma[0..sma.len-4].mapIt(it.tryGet()).foldl(a & b)
```
Menduist, line 473: Careful when using this kind of operator, they can raise a KeyError which would crash the entire program
Could you add some tests with invalid multiaddress, to be sure this is handled properly

libp2p/protocols/relay.nim
lchenut: The last two try were merged in a previous version, I separated them just to make the trace more accurate.

libp2p/builders.nim
```diff
@@ -197,6 +205,11 @@ proc build*(b: SwitchBuilder): Switch
     ms = ms,
     nameResolver = b.nameResolver)
 
+  if b.isCircuitRelay:
+    let relay = Relay.new(switch, b.circuitRelayCanHop)
+    switch.mount(relay)
+    switch.addTransport(RelayTransport.new(relay, muxedUpgrade))
```
Menduist, line 211: Actually, now that this is cleaner, we could even remove it from the builder, and let the user build it himself (like gossipsub)

Menduist (commented):
Getting close!

dryajov (approved):
Great job @lchenut! LGTM!

PR #672, opened 2021-12-10
harden `ProtoBuffer.finish()` assert

etan-status:
This hardens the length check in `ProtoBuffer.finish` to account for
passed `options`.


etan-status:
Ah, incorrect base branch. Rebasing.

Menduist:
Thanks - tried to do it myself, be because of signed commits & the fact that master has a commit which is not in unstable, can't do it cleanly :)

If you can do the same for #671 that would be perfect

etan-status:
Rebased.

Menduist (approved):

PR #674, opened 2021-12-13
Unstable

Menduist:
Multiple addresses ran smoothly for 2 weeks


emizzle:
Next time, should we we aim to use a rebase when merging (instead of merge commits) to keep the commit history clean? (sorry to be *that* guy 😬)

Menduist:
Right now, theses merge commits are the closest thing we have to a "release"

In the future we would use tag + FF merge (have to change a bit our workflows and CI scripts, will be done before V1)
However, we must not rebase unstable, because it's a shared branch, and that would break every PR open pointing to it (>10)

emizzle:
Apart from CI, maybe i'm missing something, but couldn't the branches pointing to unstable rebase on top of unstable before they're merged?

PR #677, opened 2021-12-16
Pin dependencies for CI

Menduist:
This PR adds a ~lock~pin-file for CI.
Over the last two weeks, our CI has been broken two times by dependencies updates, and in the second occurrence, I have no idea what caused it

While we must make sure that we stay up to date with our dependencies, we should avoid having a broken CI. So, we'll instead do bump PR to update our dependencies manually, and fix the issues in the bump directly, instead of breaking all of our branches while it's not fixed

We can't use the nimble lockfile, and switching to NBS is overkill, so I did a simple script `nimble pin` that installs every dependencies using nimble, and stores to `.pinned` which dependencies has been installed (with their version)
`nimble install_pinned` will read this file, and install the pinned version of every dependency.

The CI will use `install_pinned`, except the daily multi_nim job, which already uses nim "head" branches, so it might as well the head of everything.

In the future, it would be pretty easy to create the bumps automatically, by making a daily job that runs `nimble pin`, and if there is a diff in `.pinned`, opens a PR.


libp2p.nimble
```diff
@@ -94,3 +94,47 @@ task examples_build, "Build the samples":
...
+  var allDeps = newSeq[string]()
+  for dependency in listDirs("nimbledeps/pkgs"):
+    #echo dependency & "/nimblemeta.json"
+    let fileContent = readFile(dependency & "/nimblemeta.json")
+    var
+      prefix = ""
+      url = ""
+      version = ""
+    discard scanf(fileContent, "$*url\":\"$+\"", prefix, url)
+    discard scanf(fileContent, "$*vcsRevision\":\"$+\"", prefix, version)
```
Menduist, line 117: parseJson doesn't work in nimscript in nim 1.2.X

libp2p.nimble
```diff
@@ -94,3 +94,47 @@ task examples_build, "Build the samples":
...
+    #if version.len == 0:
+    #  echo execCmdEx("ls")
```
Menduist, line 119: In some unknown case (like with dnsclient), the version will not be in the nimblemeta.json file.
I would like to do `git ls-remote https://github.com/ba0f3/dnsclient.nim` and capture the output to get a pinned version, but I couldn't find a way to read the stdout of a command in nimscript.

In the meantime, I put the missing ones manually in the .pinned file, which is not ideal, but works

libp2p.nimble
Menduist: Fixed by switching the pinner to a .nim file

libp2p.nimble
Menduist: Fixed by switching the pinner to a .nim file

dryajov:
OK, this is definitely not ideal, but worth trying out.

dryajov (approved):

arnetheduck:
one regression here is that when deps change in an incompatible way, libp2p will not notice - ie up to this point, it's been a feature that we get the latest deps automatically in _libraries_ and we pin applications

arnetheduck:
this is similar to how, in langauges with working package managers, one does typically [not pin dependencies for libraries](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) - only for end binaries - I guess test binaries are a bit of a grey area

Menduist:
We'll notice in the daily job, which is even better since it's ran even if we don't change anything
And we can add an autobump if it proves necessary

We only lock for CI for our internal use basically. My long term master-plan would be:
- Each of our dep is in the form: `dep >= minimumVersion & <= latestBump`
- A CI runs on the minimumVersion every once in a while, to assure that we do support it
- The latestBump gets updated everytime we merge an (auto)bump
- Our regular CI runs like on the locked (meaning latestBump) version

But ATM, we have no way to retrieve the minimumVersion with nimble, and updating the latestBump automatically is a bit involved for the PR

dryajov:
> this is similar to how, in langauges with working package managers, one does typically [not pin dependencies for libraries](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) - only for end binaries - I guess test binaries are a bit of a grey area

Yeah, I agree that transitory dependencies such as libp2p shouldn't have lock files that influence top level dependencies, but it still needs to guarantee which transitory dependencies it it self has shown to work with. It's only advisory as far as the top level consumer is concerned, but it allows the library maintainer to set minimal requirements and provide guidelines as to which dependencies the top level project **can** use in combination with the library.

PR #678, opened 2021-12-20
expose more libp2p performance and queuing metrics (WIP)

cskiraly:
This goal of this PR is to add metrics at the libp2p level allowing deeper insight into message dynamics.

This PR is currently Work In Progress


libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -233,12 +234,15 @@ proc handleIWant*(g: GossipSub,
         trace "peer sent iwant", peer, messageID = mid
         let msg = g.mcache.get(mid)
         if msg.isSome:
+          libp2p_gossipsub_mcache_hit.observe(1)
```
Menduist, line 237: Isn't this the same as the number of IWANT sent?
I guess not if we have multiple request in a single IWANT

Menduist:
I like the idea of having a libp2p grafana dashboard, though maybe it could go in the `tools` folder?

dryajov:
> I like the idea of having a libp2p grafana dashboard, though maybe it could go in the `tools` folder?

My problem with this has always been that you can't really use it directly with libp2p since there is no executable and there doesn't seem to be any way of importing definitions from a top level project, which is a bit of a shame.

libp2p/protocols/pubsub/gossipsub/behavior.nim
cskiraly: IWANT received would be, but it's not the same. I'm exposing both on the committed dashboard, and numbers are surprisingly different. So much that there might be some scaling issue as well.

cskiraly:
> 

We could have some demo DummyApp that e.g. gossips a bit. Obviously not as integral part of the code, but in a separate folder.  Or maybe some network performance testing app that sends around data.

libp2p/protocols/pubsub/gossipsub/behavior.nim
Menduist: Oh, we are in the handleIWant, I thought we were in the handeIHave
Having very different numbers would indicate that a peer sending a IWant generally missed a lot of messages

But if the ratio is bigger than 15 (iWantBudget iirc), it's a bug sumewhere

Menduist:
> can't really use it directly with libp2p

Sure, but that could be a nice "first dashboard" for anyone using libp2p for their project, or even a dedicated dashboard for troubleshooting libp2p, which could be used by any project

arnetheduck:
+1 for a libp2p metrics dashboard that exposes as many metrics as possible - all apps _using_ libp2p will benefit from it because it shows the full measurement capabilities available - they can import it in a "subsection" of their "app dashboard" or use it as copy-paste source

cskiraly:
Should I rebase it on top of some current branch? I've intentionally started from the commit used in Nimbus-eth2 stable branch.

Menduist:
Nimbus unstable should currently point to master/unstable (except the last commit)
Use unstable as a target here

dryajov:
> +1 for a libp2p metrics dashboard that exposes as many metrics as possible - all apps _using_ libp2p will benefit from it because it shows the full measurement capabilities available - they can import it in a "subsection" of their "app dashboard" or use it as copy-paste source

Oh! So you can import the definitions in grafana? That is indeed awesome.

arnetheduck:
>  import the definitions in grafana

https://grafana.com/docs/grafana/latest/panels/panel-library/

or you can just copy-paste the json snippet that is available via buttons in the grafana ui.

cskiraly:
@Menduist , I would say this is ready for merge.

libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -23,6 +23,10 @@ logScope:
 
 when defined(libp2p_network_protocols_metrics):
   declareCounter libp2p_protocols_bytes, "total sent or received bytes", ["protocol", "direction"]
+  declareHistogram libp2p_protocols_qlen, "message queue length", ["protocol"],
```
Menduist, line 26: Does it really make sense to have theses metrics be per protocol?
Since the queue is global, they should be about the same for every protocol, no?

dryajov: The queue should be per stream tho, which can be mapped to a protocol.

dryajov (approved):
Anything stopping this from being merged?

libp2p/muxers/mplex/lpchannel.nim
Menduist: Having it be per-protocol means 12 metrics instead of one, we have to be sure it's worth it

arnetheduck:
We talked about this a little bit with tanguy - libp2p is clearly layered and it would make sense that the exposed metrics have a structure that follows these layers "naturally" - for example, `mplex` is a layer and metrics that relate to it would have "mplex" in their name - the "layering" would also help determine the granularity of metrics etc.

With that in mind, it's good to be mindful that it's easier to add metrics than it is to remove them, though it's not a strict requirement that obsolete metrics hang around - it's a "soft" compatibility thing.

arnetheduck:
the other point about the layering was that one should be able to read a story from the metrics - just like one API uses another and translates the abstraction levels, so should the metrics relate to each other layer by layer

arnetheduck:
oh, sorry, `mplex` was just the first layer that came to mind, mostly because it's in the middle :)

libp2p/muxers/mplex/lpchannel.nim
dryajov: You're right, there are limits to how many metrics we can collect in practice, even for tracing/debugging purposes, for example prometheus gets quickly overwhelmed if too many metrics are being pushed to it.

dryajov:
> Though, the metric added here is a bit different It's not really tied to mplex, we just put it there because that's where the "buffering" happens to be
> 
> The goal of this metric is only to know how long our queue is. And this queue will be per-peer, basically
> 
> Hence my question. My gut feel is that if anything, having it be per-protocol will be error inducing, since the buffering is not actually per-protocol, but per-peer. But if it exhibits interesting behaviors, happy to keep it that way

Good point, in this context it does make sense to make it per peer rather than per protocol.

I agree with @arnetheduck about layering and in general would like to see our metrics re-engineered both at the level of information we collect as well as the overall architecture. We're quickly reaching the limits of our simplistic approach and its only becoming harder to keep track as we keep adding more.

cskiraly:
> > Though, the metric added here is a bit different It's not really tied to mplex, we just put it there because that's where the "buffering" happens to be
> > The goal of this metric is only to know how long our queue is. And this queue will be per-peer, basically
> > Hence my question. My gut feel is that if anything, having it be per-protocol will be error inducing, since the buffering is not actually per-protocol, but per-peer. But if it exhibits interesting behaviors, happy to keep it that way
> 
> Good point, in this context it does make sense to make it per peer rather than per protocol.
> 
> I agree with @arnetheduck about layering and in general would like to see our metrics re-engineered both at the level of information we collect as well as the overall architecture. We're quickly reaching the limits of our simplistic approach and its only becoming harder to keep track as we keep adding more.

Indeed, it would be great to have metrics per layer with some naming convention, and feature flags for anything that is somewhat heavy in processing. 
- For the queuing related metrics, they are in mplex, so I would rename them libp2p_mplex_* (or libp2p_muxers_* ?). Per protocol queue metrics indeed does not make sense here. Per peer is instead heavy, so that should definitely be behind a flag. I change it to be just a simple metric by default, and see whether a per peer behind a compile time flag makes sense.
- Duplicate related metrics are already called libp2p_gossipsub_*

libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -187,6 +191,8 @@ proc prepareWrite(s: LPChannel, msg: seq[byte]): Future[void] {.async.} =
   if s.writes >= MaxWrites:
     debug "Closing connection, too many in-flight writes on channel",
       s, conn = s.conn, writes = s.writes
+    when defined(libp2p_network_protocols_metrics):
+        libp2p_protocols_qlenclose.inc(labelValues=[s.tag])
```
Menduist, line 195:
```suggestion
    when defined(libp2p_network_protocols_metrics):
      libp2p_protocols_qlenclose.inc(labelValues=[s.tag])
```
ocd

libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -21,6 +21,12 @@ export connection
 logScope:
   topics = "libp2p mplexchannel"
 
+when defined(libp2p_mplex_metrics):
```
Menduist, line 24: Actually, those are pretty cheap, we can enable them by default
Apart from that, LGTM

cskiraly: 2 of them are histograms. Still not a lot of metrics, but maybe we should keep them optional to avoid extra load on every message processing. Let me know, I can keep it as is or remove the `define`.

Menduist:
Sorry I don't remember, what was the state of this? @cskiraly

cskiraly:
> Sorry I don't remember, what was the state of this? @cskiraly

I think we've resolved everything, and it can be merged.

Menduist (dismissed):

arnetheduck:
>  it would be great to have metrics per layer with some naming convention, 

how do these metrics mesh with the other metrics already present? adding metrics is easier than removing them (for backwards compat reasons) - are they named and do they fit current or future planned conventions? were the existing metrics reviewed for overlap?

cskiraly:
3 groups of metrics:
- libp2p_mplex_: here we add metrics related to queuing that happens in code. These metrics are behind `when defined(libp2p_mplex_metrics)` to avoid overhead when not looking into queuing behavior
- libp2p_gossipsub_: new metrics are here to verify assumptions on overhead due to blind push behaviour, a protocol choice that has its reasons and costs.
- libp2p_gossipsub_mcache_: new metric to provide insight into the pull behaviour of the protocol

Naming is the best I can come up with.
I haven't seen anything I would consider an overlap.

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -25,6 +25,7 @@ declareGauge(libp2p_gossipsub_no_peers_topics, "number of topics in mesh with no
 declareGauge(libp2p_gossipsub_low_peers_topics, "number of topics in mesh with at least one but below dlow peers")
 declareGauge(libp2p_gossipsub_healthy_peers_topics, "number of topics in mesh with at least dlow peers (but below dhigh)")
 declareCounter(libp2p_gossipsub_above_dhigh_condition, "number of above dhigh pruning branches ran", labels = ["topic"])
+declareSummary(libp2p_gossipsub_mcache_hit, "ratio of successful IWANT message cache lookups", labels = ["topic"])
```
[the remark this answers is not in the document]
cskiraly: not set. This would make sense as a per-topic meric, but I see IHAVE and other messages have a topicID field, but IWANT does not. 
If we still have the message in mcache, we could get the topic ID. But if it is missing, we just simply don't know, I'm afraid.
Consequence: if you agree, lets just remove labels here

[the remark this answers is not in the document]
arnetheduck: if you don't set declared labels, metrics will crash at runtime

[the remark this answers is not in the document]
Menduist: Yeah, let's remove it

Menduist (approved):

PR #680, opened 2021-12-22
Gossipsub: process messages concurrently

### LINKED ISSUE #679: Add more concurrency to Gossipsub validation
One issue with the way that validation is set up is that in libp2p, the "next" gossip message from a single node is not processed until the current one has been validated - this decreases concurrency, and therefore also the ability to build meaningful batches - this is specially problematic for lazy aggregation because one node is likely to be sending multiple messages for the same data when dealing with an attestation burst: all attestations on a single subnet will typically have the same message being signed, and nodes listen to particular subnets

_Originally posted by @arnetheduck in https://github.com/status-im/nimbus-eth2/issues/3212#issuecomment-998940551_

Menduist:
Closes #679

First dumb version, each message processing is asyncSpawn. Probably good enough?

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -319,8 +319,6 @@ proc validateAndRelay(g: GossipSub,
       g.floodsub.withValue(t, peers): toSendPeers.incl(peers[])
       g.mesh.withValue(t, peers): toSendPeers.incl(peers[])
 
-      await handleData(g, t, msg.data)
-
     # Don't send it to source peer, or peers that
     # sent it during validation
     toSendPeers.excl(peer)
@@ -331,10 +329,14 @@ proc validateAndRelay(g: GossipSub,
     g.broadcast(toSendPeers, RPCMsg(messages: @[msg]))
     trace "forwared message to peers", peers = toSendPeers.len, msgId, peer
     for topic in msg.topicIDs:
+      if topic notin g.topics: continue
+
       if g.knownTopics.contains(topic):
         libp2p_pubsub_messages_rebroadcasted.inc(toSendPeers.len.int64, labelValues = [topic])
       else:
         libp2p_pubsub_messages_rebroadcasted.inc(toSendPeers.len.int64, labelValues = ["generic"])
+
+      await handleData(g, topic, msg.data)
   except CatchableError as exc:
     info "validateAndRelay failed", msg=exc.msg
 
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -282,6 +282,62 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
       peer,
       RPCMsg(control: some(respControl), messages: messages))
 
+proc validateAndRelay(g: GossipSub,
+                      msg: Message,
+                      msgId, msgIdSalted: MessageId,
+                      peer: PubSubPeer) {.async.} =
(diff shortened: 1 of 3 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -282,6 +282,62 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
...
+
+    var toSendPeers = HashSet[PubSubPeer]()
+    for t in msg.topicIDs:                      # for every topic in the message
+      if t notin g.topics:
+        continue
+
+      g.floodsub.withValue(t, peers): toSendPeers.incl(peers[])
+      g.mesh.withValue(t, peers): toSendPeers.incl(peers[])
+
+      await handleData(g, t, msg.data)
```
arnetheduck, line 322: looks to me like `toSendPeers` can go stale between calls to `handleData` - calling it should likely be done in a separate loop

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -282,6 +282,62 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
...
+      debug "Dropping message after validation, reason: ignore",
+        msgId = shortLog(msgId), peer
+      return
+    of ValidationResult.Accept:
+      discard
+
+    # store in cache only after validation
+    g.mcache.put(msgId, msg)
+
+    g.rewardDelivered(peer, msg.topicIDs, true)
```
arnetheduck, line 312: what happens if the peer is .. disconnected somewhere in the middle here?

Menduist: Shouldn't be an issue, we'll just ignore the score

libp2p/protocols/pubsub/gossipsub.nim
Menduist: Move handleData to the end, the broadcasting part is already asyncSpawned anyway

dryajov:
> First dumb version, each message processing is asyncSpawn. Probably good enough?

Should be fine, we handle all the error conditions in the rpc loop either way and there isn't anything you could do with the potentially failed future either way.

dryajov (dismissed):

dryajov:
Careful with the failing tests, it might have broken some timing assumptions around message order/arrival... should be fixed accordingly but shouldn't necessarily be cause for concern for stuff already in the wild.

Menduist:
It's floodsub which is failing, and this PR only impacts Gossipsub, but you never know with flaky stuff.. I'll check tmr

Menduist:
I added a `flakyAsyncTest` template, which will retry a flaky test multiple time.

Tired of being slowed down by go-libp2p-daemon bugs
If it LGTY we can merge, otherwise I'll just revert it and hope my other test to fix flakyness work

dryajov:
> I added a `flakyAsyncTest` template, which will retry a flaky test multiple time.
> 
> Tired of being slowed down by go-libp2p-daemon bugs (yesterday, it was failing consistently on macos, even on other branches, and now everything work fine) If it LGTY we can merge, otherwise I'll just revert it and hope my other test to fix flakyness work EDIT: it did not

I'd be careful, flaky tests are more often than not a symptom of something broken.

Menduist:
Yes, the go-libp2p-daemon is broken, it's not new

The test on which I added the flaky option is on the daemon test, not even the interop one

Menduist:
Merging this so we can test it over the holidays on the testfleet, we can always revert the flaky thing later

PR #682, opened 2021-12-23
[Test] Test bumped libp2p daemon

Menduist:
I bumped go-libp2p-daemon deps here: https://github.com/Menduist/go-libp2p-daemon/tree/bumpdeps
Trying it here to see if it solve a bit of flakyness

If it works nicely, I'll open a PR uptstream


Menduist:
Ok so it's really a bug in the daemon

PR #684, opened 2022-01-06
CI: test with multiple Nim versions by default

stefantalpalaru:


stefantalpalaru:
Testing ORC doubles the run time, for Nim-1.6 and devel. Should we compromise on only running "test_slim" with it?

stefantalpalaru:
The "e2e should allow multiple local addresses" test failed silently on Windows, with Nim-1.4 and 1.6 and the system was waiting on an interactive prompt, for some reason. Adding a timeout.

Menduist:
I though we settled on only testing multiple versions during daily CI to avoid having a too long CI?

With the current system, we already get >8h queues of ci on busy days, so I would try to minimize this has much has possible

Unless this is fully parallel and doesn't take more time? But since the daily ci takes >1h instead of ~30 minutes for the regular one, I don't think so

Maybe a middle ground would be to only compile with 1.2 & devel on regular CI, and do the full one daily?

stefantalpalaru:
The daily test only runs on the "master" branch and is hidden. I just found out it passes on all compiler versions yesterday.

Only testing Nim-1.2 and -devel is a good compromise, so I implemented that. Yes, all these CI jobs run in parallel, but they do add up in the queue of used hosts, which seems to be per GitHub account.

Also leaving ORC testing to just the daily CI job, to lighten the regular CI load.

I disabled that problematic test on Windows, because I saw it failing in the same spot, on multiple compiler versions, and locking up the host.

Menduist:
> hidden

Don't worry, I get nice emails everytime it fails :)
We could also run unstable though, indeed

Indeed we have windows lockups currently, that has been caused by a dependency update, but I don't know which one unfortunately
That's the reason I implemented lockfiles actually, to avoid it happening again :)
Need to investigate

stefantalpalaru:
The failing Windows test is always associated with websockets.

Menduist:
Yeah, but nim-webock wasn't updated significantly since ~November, and the bug popped up in December, so it's a transiant dep unfortunately

And the two test aren't duplicate, they just are at different layers ;)

.github/workflows/ci.yml
```diff
@@ -174,8 +160,14 @@ jobs:
         run: |
           V=1 bash scripts/build_p2pd.sh p2pdCache 124530a3
 
-      - name: Run nim-libp2p tests
+      - name: Run tests
         run: |
+          if [[ "${{ matrix.target.os }}" == "windows" ]]; then
+            # https://github.com/status-im/nimbus-eth2/issues/3121
+            export NIMFLAGS="-d:nimRawSetjmp"
+          fi
```
Menduist, line 168: Shouldn't this be set in the `config.nims` instead?

stefantalpalaru: Maybe, but not all projects have one and it's more important to me to be able to use "--skipParentCfg" than concentrating all compiler flags in a single place.

Menduist:
@stefantalpalaru sorry I missed that, but why aren't we testing with windows-i386 anymore?

stefantalpalaru:
Because we already have Linux i386. No point in having more than one 32-bit platform in CI, when we don't actually deploy on it anyway.

PR #687, opened 2022-01-20
Semaphore rework

Menduist:
This is an alternative to https://github.com/status-im/nim-libp2p/pull/629

This PR modifies the way the maxIn, maxOut & maxConnections are handled.

Previously, this was the logic:
```nim
  if maxIn > 0 or maxOut > 0:
    inSema = newAsyncSemaphore(maxIn)
    outSema = newAsyncSemaphore(maxOut)
  elif maxConnections > 0:
    inSema = newAsyncSemaphore(maxConnections)
    outSema = inSema
```

Now:
```nim
    inSema: newAsyncSemaphore(maxIn),
    outSema: newAsyncSemaphore(maxOut),
    connSema: newAsyncSemaphore(maxConnections)
```

In practice, this means that there is always a global & per direction limit.
We now have to acquire two semaphore, but I don't think it can create liveness issues.

This new method seems more elegant (& flexible) than the current system, and than 629, imo

Second feature of this PR: you can now enable forceDial on the connManager.
This means that outgoing connections will only fail if the outSema is full, the connSema will be forceAcquired.

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -59,8 +59,8 @@ proc new*(T: type[SwitchBuilder]): T =
     addresses: @[address],
     secureManagers: @[],
     maxConnections: MaxConnections,
-    maxIn: -1,
-    maxOut: -1,
+    maxIn: MaxConnections,
+    maxOut: MaxConnections,
     maxConnsPerPeer: MaxConnectionsPerPeer,
     protoVersion: ProtoVersion,
     agentVersion: AgentVersion)
@@ -206,8 +206,8 @@ proc newStandardSwitch*(
   inTimeout: Duration = 5.minutes,
   outTimeout: Duration = 5.minutes,
   maxConnections = MaxConnections,
-  maxIn = -1,
-  maxOut = -1,
+  maxIn = MaxConnections,
+  maxOut = MaxConnections,
   maxConnsPerPeer = MaxConnectionsPerPeer,
   nameResolver: NameResolver = nil): Switch
   {.raises: [Defect, LPError].} =
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -46,6 +46,7 @@ type
     maxConnsPerPeer: int
     protoVersion: string
     agentVersion: string
+    forceDial: bool
     nameResolver: NameResolver
 
 proc new*(T: type[SwitchBuilder]): T =
@@ -59,8 +60,8 @@ proc new*(T: type[SwitchBuilder]): T =
     addresses: @[address],
     secureManagers: @[],
     maxConnections: MaxConnections,
-    maxIn: MaxConnections,
-    maxOut: MaxConnections,
+    maxIn: -1,
+    maxOut: -1,
     maxConnsPerPeer: MaxConnectionsPerPeer,
     protoVersion: ProtoVersion,
     agentVersion: AgentVersion)
@@ -109,6 +110,10 @@ proc withRng*(b: SwitchBuilder, rng: ref BrHmacDrbgContext): SwitchBuilder =
 
 proc withMaxConnections*(b: SwitchBuilder, maxConnections: int): SwitchBuilder =
   b.maxConnections = maxConnections
+  b
+
+proc withForceDial*(b: SwitchBuilder, forceDial: bool): SwitchBuilder =
+  b.forceDial = forceDial
   b
 
 proc withMaxIn*(b: SwitchBuilder, maxIn: int): SwitchBuilder =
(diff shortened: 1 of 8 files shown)


libp2p/builders.nim
```diff
@@ -166,7 +171,9 @@ proc build*(b: SwitchBuilder): Switch
 
   let
     identify = Identify.new(peerInfo)
-    connManager = ConnManager.new(b.maxConnsPerPeer, b.maxConnections, b.maxIn, b.maxOut)
+    maxIn = if b.maxIn >= 0: b.maxIn else: b.maxConnections
+    maxOut = if b.maxOut >= 0: b.maxOut else: b.maxConnections
+    connManager = ConnManager.new(b.maxConnsPerPeer, b.maxConnections, maxIn, maxOut, b.forceDial)
```
arnetheduck, line 176: why state and not an argument to the `dial` function? I can imagine applications wanting to control this on a dial-by-dial basis

libp2p/connmanager.nim
```diff
@@ -440,19 +434,23 @@ proc trackIncomingConn*(c: ConnManager,
   try:
     trace "Tracking incoming connection"
     await c.inSema.acquire()
+    await c.connSema.acquire()
     conn = await c.trackConn(provider, c.inSema)
     if isNil(conn):
       trace "Couldn't acquire connection, releasing semaphore slot", dir = $Direction.In
       c.inSema.release()
+      c.connSema.release()
```
arnetheduck, line 442: release in opposite order to acquire

dryajov:
As I mentioned in https://github.com/status-im/nim-libp2p/pull/629#pullrequestreview-855755982, the changes made to the semaphore are incorrect they break the symmetry of the `acquire/release` calls and it will end up releasing the wrong future in the incorrect order.

libp2p/builders.nim
```diff
@@ -166,7 +166,9 @@ proc build*(b: SwitchBuilder): Switch
 
   let
     identify = Identify.new(peerInfo)
-    connManager = ConnManager.new(b.maxConnsPerPeer, b.maxConnections, b.maxIn, b.maxOut)
+    maxIn = if b.maxIn >= 0: b.maxIn else: b.maxConnections
```
dryajov, line 169: What happens if `maxIn/maxOut > maxConnections`?

dryajov (changes requested):
I can't really see how this is an improvement over whats there, specially with the addition of the second semaphore, which unless you allow `maxIn/maxOut > maxConnections` (which doesn't really make sense), is exactly the same thing as whats currently in place. 

The `forceDial` flag is fine, but at that point and _specially_ with the way that `forceAcquire` is implemented, makes the semaphore into a simple counter. Why then not, simply use a counter if thats the desired behavior?

Lastly, the changes to the semaphore as they are right now, render the semaphore completely broken (as per https://github.com/status-im/nim-libp2p/pull/629#discussion_r789117834). Again, if all you want to do is count resources, not synchronize them, then why use a semaphore?

Menduist:
> is exactly the same thing as whats currently in place.

It's the same thing, except that it's simpler to reason about (always two semaphores, instead of having different cases depending on `maxIn > 0 or maxOut > 0`)

It also opens a few possibilities, for instance:
MaxConn = 100, MaxIn = 90, MaxOut = 30
Will cap to 100 peers, but always leaving at least 10 slots, up to 30 for outgoing peers, which is not possible with the current system

It's not miles better than the current one, but a bit better.
However, it's IMO better than the proposal in 629, which felt ad-hoc and more complex to grasp

With forceDial, we could setup nimbus like this:
MaxIn = MaxOut = MaxConn = 70
Which gives 70 as a target, while allowing outgoing connections to go above that, while still being limited to 70 total outgoing connections.

> makes the semaphore into a simple counter. Why then not, simply use a counter if thats the desired behavior?

You'll notice that the new semaphore is used exactly like the old one. Eg, as a "counter" for outgoing connections, but as a sync mechanism for incoming connections

I guess we could use a simple counter for the `outSema`, but I like the consistency here. We could also in the future add a `slowDial`, that would wait for an outgoing slot


> render the semaphore completely broken

I mean, the instant you add a notion of "forceAcquire" to a semaphore, you've broken it, by allowing it to become negative.
My version tries to restore a natural state (>= 0) as quickly as possible, by not releasing new slot as long as it's negative, which I think make sense. The asyncSpawn version would, in practice, let more undesired incoming peers slip through, which I think is not great, but we can probably deal with both

libp2p/builders.nim
Menduist: Without forceDial, you'll always be capped by the lowest value. With forceDial, you can go over maxConnections

Menduist:
Superseded by #696

Issue #642, opened 2021-11-02
Allow msgIdProvider to fail


Menduist:
In pubsub, allow msgIdProvider to fail (either through a exception, or a ValidationResult), and if it fails, completely skip the message (don't store it in the seen cache, etc)

More info: https://github.com/status-im/nimbus-eth2/issues/3043

PR #688, opened 2022-01-21
feat: allow msgIdProvider to fail

emizzle:
Closes: #642.

Changes the return type of the msgIdProvider to `Result[MessageID, string]` so that message id generation can fail.

String error type was chosen as this `msgIdProvider` mainly because the failed message id generation drops the message and logs the error provided. Because `msgIdProvider` can be externally provided by library consumers, an enum didn’t make sense and a object seemed to be overkill. Exceptions could have been used as well, however, in this case, Result ergonomics were warranted and prevented wrapping quite a large block of code in try/except.

The `defaultMsgIdProvider` function previously allowed message id generation to fail silently for use in the tests: when seqno or source peerid were not valid, the message id generated was based on a hash of the message data and topic ids. The silent failing was moved to the `defaultMsgIdProvider` used only in the tests so that it could not fail silently in applications.

Unit tests were added for the `defaultMsgIdProvider`.

### DIFF
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -100,6 +100,7 @@ method rpcHandler*(f: FloodSub,
     if msgIdResult.isErr:
       debug "Dropping message due to failed message id generation",
         error = msgIdResult.error
+      # TODO: descore peers due to error during message validation (malicious?)
       continue
 
     let msgId = msgIdResult.get
@@ -193,7 +194,7 @@ method publish*(f: FloodSub,
     msgIdResult = f.msgIdProvider(msg)
 
   if msgIdResult.isErr:
-    debug "Error generating message id, skipping publish",
+    trace "Error generating message id, skipping publish",
       error = msgIdResult.error
     return 0
 
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -96,7 +96,13 @@ method rpcHandler*(f: FloodSub,
     f.handleSubscribe(peer, sub.topic, sub.subscribe)
 
   for msg in rpcMsg.messages:                         # for every message
-    let msgId = f.msgIdProvider(msg)
+    let msgIdResult = f.msgIdProvider(msg)
+    if msgIdResult.isErr:
+      debug "Dropping message due to failed message id generation",
+        error = msgIdResult.error
+      continue
+
+    let msgId = msgIdResult.get
 
     if f.addSeen(msgId):
       trace "Dropping already-seen message", msgId, peer
@@ -184,7 +190,14 @@ method publish*(f: FloodSub,
         Message.init(none(PeerInfo), data, topic, none(uint64), false)
       else:
         Message.init(some(f.peerInfo), data, topic, some(f.msgSeqno), f.sign)
-    msgId = f.msgIdProvider(msg)
+    msgIdResult = f.msgIdProvider(msg)
+
+  if msgIdResult.isErr:
+    debug "Error generating message id, skipping publish",
+      error = msgIdResult.error
+    return 0
+
+  let msgId = msgIdResult.get
 
   trace "Created new message",
     msg = shortLog(msg), peers = peers.len, topic, msgId
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -367,6 +367,7 @@ method rpcHandler*(g: GossipSub,
     if msgIdResult.isErr:
       debug "Dropping message due to failed message id generation",
         error = msgIdResult.error
+      # TODO: descore peers due to error during message validation (malicious?)
       continue
 
     let
@@ -515,8 +516,9 @@ method publish*(g: GossipSub,
     msgIdResult = g.msgIdProvider(msg)
 
   if msgIdResult.isErr:
-    debug "Dropping message due to failed message id generation",
+    trace "Error generating message id, skipping publish",
       error = msgIdResult.error
+    libp2p_gossipsub_failed_publish.inc()
     return 0
 
   let msgId = msgIdResult.get
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -362,8 +362,15 @@ method rpcHandler*(g: GossipSub,
 
   for i in 0..<rpcMsg.messages.len():                         # for every message
     template msg: untyped = rpcMsg.messages[i]
+    let msgIdResult = g.msgIdProvider(msg)
+
+    if msgIdResult.isErr:
+      debug "Dropping message due to failed message id generation",
+        error = msgIdResult.error
+      continue
+
     let
-      msgId = g.msgIdProvider(msg)
+      msgId = msgIdResult.get
       msgIdSalted = msgId & g.seenSalt
 
     # addSeen adds salt to msgId to avoid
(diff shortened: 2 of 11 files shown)


libp2p/protocols/pubsub/rpc/message.nim
```diff
@@ -28,16 +28,12 @@ const PubSubPrefix = toBytes("libp2p-pubsub:")
...
 
-func defaultMsgIdProvider*(m: Message): MessageID =
-  let mid =
-    if m.seqno.len > 0 and m.fromPeer.data.len > 0:
-      byteutils.toHex(m.seqno) & $m.fromPeer
-    else:
-      # This part is irrelevant because it's not standard,
-      # We use it exclusively for testing basically and users should
-      # implement their own logic in the case they use anonymization
-      $m.data.hash & $m.topicIDs.hash
```
emizzle, line 39: Moved this part to the tests so that it does not fail silently.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -505,7 +512,14 @@ method publish*(g: GossipSub,
         Message.init(none(PeerInfo), data, topic, none(uint64), false)
       else:
         Message.init(some(g.peerInfo), data, topic, some(g.msgSeqno), g.sign)
-    msgId = g.msgIdProvider(msg)
+    msgIdResult = g.msgIdProvider(msg)
+
+  if msgIdResult.isErr:
+    debug "Dropping message due to failed message id generation",
```
Menduist, line 518:
```suggestion
    debug "Error generating message id, skipping publish",
```

Menduist: And theses ones should probably be some INFOs, actually

Menduist (commented):
Looks pretty good!

A benefit of exceptions instead of a Result would have been to keep backward compat, but that shouldn't block us

wdyt @arnetheduck?

libp2p/protocols/pubsub/gossipsub.nim
arnetheduck: no, that would open up an attack vector: generating a message id happens _before_ pretty much every other validation like deduplication etc - in fact, it's where an untrusted external blob first meets the application and we need to treat it with care - in eth2, the message id generation function is quite complex and if it fails, it's usually due to an attack - therefore, rather than a string, I see a ValidationResult (or similar) `enum` as the way forward so that libp2p can penalise the peer for sending it.

Menduist: That's in publish, so to hit this one, it's the application doing something bad

Though, good point for the ones in rpcHandler. Maybe switch to a trace and add a metric instead

arnetheduck: right, this is publish - hm, yeah, if it fails here, it's almost the case that it warrants an error being passed on to the caller - I'm not sure callers have an adequate error handling policy at this stage though because `publish` is sort of best-effort..

this reminds me of https://github.com/status-im/nimbus-eth2/issues/3269 which basically says that libp2p is the wrong abstraction level to be making log level decisions

dryajov (dismissed):
LGTM :)

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -85,7 +85,7 @@ type
   TopicPair* = tuple[topic: string, handler: TopicHandler]
 
   MsgIdProvider* =
-    proc(m: Message): MessageID {.noSideEffect, raises: [Defect], gcsafe.}
+    proc(m: Message): Result[MessageID, string] {.noSideEffect, raises: [Defect], gcsafe.}
```
arnetheduck, line 88: so per discussion elsewhere, this should return a `ValidationResult` so that gossip can penalize peers that send junk

emizzle: Updated.

emizzle:
@arnetheduck @Menduist thanks for the reviews. This has been updated so that `MsgIdProivder` return type is `Result[MessageId, ValidationResult]`. The `ValidationResult` type had to be moved to `/errors.nim` to avoid circular dependencies.

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -96,7 +96,13 @@ method rpcHandler*(f: FloodSub,
     f.handleSubscribe(peer, sub.topic, sub.subscribe)
 
   for msg in rpcMsg.messages:                         # for every message
-    let msgId = f.msgIdProvider(msg)
+    let msgIdResult = f.msgIdProvider(msg)
+    if msgIdResult.isErr:
+      debug "Dropping message due to failed message id generation",
```
arnetheduck, line 101: although it doesn't necessarily have to be part of this PR, we should at least note in a TODO that we need to descore peers for which reject is returned here.

Also, seeing this code makes me think of another potential polishing feature for the future: turn `ValidationResult` into `ValidationError` without accept, because `err(ValidationResult.Accept` is not really meaningful.

regarding backwards compat, that can actually be done quite easily @Menduist - the enrolling function that takes a validation callback can have overloads for the old signatures that translate into the new signature.

arnetheduck (dismissed):

libp2p/errors.nim
```diff
@@ -9,6 +9,9 @@ type
   # Base exception type for libp2p
   LPError* = object of CatchableError
 
+  ValidationResult* {.pure.} = enum
+    Accept, Reject, Ignore
+
```
Menduist, line 14: If we can avoid having this here, this is really gossipsub centric, and gossipsub shouldn't "escape" the pubsub folder
I know, cyclic imports are a pain :/

libp2p/protocols/pubsub/gossipsub.nim
Menduist: up on this

libp2p/protocols/pubsub/floodsub.nim
emizzle: Todo added 👍 

Descore peers: https://github.com/status-im/nim-libp2p/issues/692

`ValidationResult` -> `ValidationError` issue: https://github.com/status-im/nim-libp2p/issues/691

libp2p/protocols/pubsub/gossipsub.nim
emizzle: @Menduist changed to `trace` and metric added 👍

libp2p/errors.nim
emizzle: Moved `ValidationResult` to a new file `libp2p/protocols/pubsub/errors.nim`.

emizzle:
All changes should be addressed. My apologies, should have added a comment after addressing the changes.

tests/pubsub/testmcache.nim
```diff
@@ -5,19 +5,21 @@ import stew/byteutils
...
   PeerId.init(PrivateKey.random(ECDSA, rng[]).get()).get()
 
+const MsgIdGenFail = "msg id gen failure"
+
 suite "MCache":
   test "put/get":
     var mCache = MCache.init(3, 5)
     var msg = Message(fromPeer: randomPeerId(), seqno: "12345".toBytes())
-    let msgId = defaultMsgIdProvider(msg)
+    let msgId = defaultMsgIdProvider(msg).expect(MsgIdGenFail)
```
arnetheduck, line 22: generally, `expect` expects a "positive" message: `.expect("it worked")` because it will negate it in the error string that it actually print - right now, the code reads as if we "expect message id generation failure" which is off

arnetheduck (approved):

dryajov (approved):

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -11,7 +11,8 @@
 
 import std/[tables, sequtils, sets, strutils]
 import chronos, chronicles, metrics, bearssl
-import ./pubsubpeer,
+import ./errors as pubsub_errors,
```
Menduist, line 14: We missed that during review, but `pubsub_errors` should have been exported to limit the backward compat breaking (though is PR is already breaking backward compatibility, so it's not too bad)

emizzle: Fixed in https://github.com/status-im/nim-libp2p/pull/697.

tests/pubsub/testmcache.nim
emizzle: Fixed in https://github.com/status-im/nim-libp2p/pull/697.

PR #689, opened 2022-01-28
Fix websocket EOF reading exception

Menduist:
by mapping AsyncStreamIncompleteError to an EOF

### DIFF
--- a/libp2p/transports/wstransport.nim
+++ b/libp2p/transports/wstransport.nim
@@ -49,15 +49,23 @@ proc new*(T: type WsStream,
   stream.initStream()
   return stream
 
+template mapExceptions(body: untyped) =
+  try:
+    body
+  except AsyncStreamIncompleteError:
+    raise newLPStreamEOFError()
+  except AsyncStreamUseClosedError:
+    raise newLPStreamEOFError()
+  except WSClosedError:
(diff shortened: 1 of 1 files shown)


libp2p/transports/wstransport.nim
```diff
@@ -53,7 +53,12 @@ method readOnce*(
   s: WsStream,
   pbytes: pointer,
   nbytes: int): Future[int] {.async.} =
-  let res = await s.session.recv(pbytes, nbytes)
+  let res =
```
dryajov, line 56: We probably want to handle all the exceptions as in https://github.com/status-im/nim-libp2p/blob/unstable/libp2p/stream/chronosstream.nim#L67, also maybe a good idea to move to `/streams`?

Menduist: Done

I like to keep it self-contained, especially since it's that small (<50 lines for the stream)
But maybe in the future if someone wants to use WsStream without the transport?

dryajov: Thanks! Yeah, thats the logic that the streams are their own thing, we might for example use them in other places, like codex/dagger. We're already using `BufferStream` for example.

dryajov (approved):

PR #690, opened 2022-01-28
Fix identify log for json_sink

Menduist:
ref https://github.com/status-im/nimbus-eth2/issues/3336

### DIFF
--- a/libp2p/protocols/identify.nim
+++ b/libp2p/protocols/identify.nim
@@ -9,8 +9,8 @@
 
 {.push raises: [Defect].}
 
-import options
-import chronos, chronicles
+import std/[sequtils, strutils]
+import chronos, chronicles, json_serialization/std/options
 import ../protobuf/minprotobuf,
        ../peerinfo,
        ../stream/connection,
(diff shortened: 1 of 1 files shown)


libp2p/protocols/identify.nim
```diff
@@ -9,8 +9,8 @@
 
 {.push raises: [Defect].}
 
-import options
-import chronos, chronicles
+import std/[sequtils, strutils]
+import chronos, chronicles, json_serialization/std/options
```
dryajov, line 13: Hmm, this seems weird, _do we have to_ import this explicitly or is this a workaround?

Menduist: The default option will be serialized into: `Some("text")` (and the " will break json) 
This imports contains a wrapper which will correctly serialize it.

Other option is how I had to do for the observed address

Menduist: Not happy with it either, since it's the first time we mention json_serialization in libp2p (it's a transient dependency)
Oh well, I'll probably switch to the manual technique next week

dryajov (approved):

PR #695, opened 2022-02-21
Unstable

Menduist:
This "release" contains a breaking change for gossipsub msgId generation: #688


Menduist:
Note for later: don't merge old PR without checking they still work (looking at you #638)

tests/commontransport.nim
```diff
@@ -137,6 +137,10 @@ proc commonTransportTest*(name: string, prov: TransportProvider, ma: string) =
       await transport1.stop()
 
     asyncTest "e2e should allow multiple local addresses":
+      when defined(windows):
+        # this randomly locks the Windows CI job
```
dryajov, line 141: Huh, maybe on windows it could use a timeout to allow stuff to cleanup between runs. Usually, issues that we've identified on windows (or other platforms) individually, have ended up being real problems.

PR #696, opened 2022-02-22
Force dial

Menduist:
This PR allows applications to dial peers even if we reached the connection limit

A new semaphore slot will be temporarily created, so the connection can never fail because of peer limitation

Alternative for #687, needed for https://github.com/status-im/nimbus-eth2/pull/3346


libp2p/connmanager.nim
```diff
@@ -463,9 +464,12 @@ proc trackOutgoingConn*(c: ConnManager,
                                         max = c.outSema.size
 
   if not c.outSema.tryAcquire():
-    trace "Too many outgoing connections!", count = c.outSema.count,
-                                            max = c.outSema.size
-    raise newTooManyConnectionsError()
+    if forceDial:
+      c.outSema.forceAcquire()
```
arnetheduck, line 468: why not always `forceAquire`? what does the "try" buy?

libp2p/utils/semaphore.nim
```diff
@@ -77,13 +78,14 @@ proc release*(s: AsyncSemaphore) =
     trace "Releasing slot", available = s.count,
                             queue = s.queue.len
 
+    s.count.inc
```
arnetheduck, line 81: this doesn't guarantee that count goes above 0 - should it really complete queued futures then?

libp2p/utils/semaphore.nim
```diff
@@ -77,13 +78,14 @@ proc release*(s: AsyncSemaphore) =
...
     if s.queue.len > 0:
       var fut = s.queue[0]
       s.queue.delete(0)
       if not fut.finished():
         fut.complete()
+        s.count.dec
```
arnetheduck, line 87: as general hygiene, it's probably better to mutate the state _before_ "releasing" the work that depends on the state update to have happened, even if de-facto, the work that gets released happens later

libp2p/utils/semaphore.nim
Menduist: That's how the incoming doesn't get starved: the slot "created" by a `forceAcquire` can be reused for incoming connections.
Of course, after a while this slot will be deleted (because more stuff will be released than `acquire`d)

libp2p/utils/semaphore.nim
Menduist: Yeah, this proc is not async, but better be clean

libp2p/utils/semaphore.nim
arnetheduck: ok - hm, that is a little bit more permissive than I imagined it - it's a good point about starvation though, it does give more balance in return for longer periods of "non-compliance" - let's document  this in code in case it causes trouble we didn't anticipate

libp2p/utils/semaphore.nim
```diff
@@ -77,13 +78,14 @@ proc release*(s: AsyncSemaphore) =
     trace "Releasing slot", available = s.count,
                             queue = s.queue.len
 
+    s.count.inc
     if s.queue.len > 0:
       var fut = s.queue[0]
       s.queue.delete(0)
       if not fut.finished():
```
arnetheduck, line 85: this code looks odd btw: why would there be finished futures in the queue?

arnetheduck: and as a follow-up, if a future is finished already, should the "next" future be completed instead?

Menduist: A future could be cancelled
Not sure why we don't try another, I switched to a loop and every tests pass, @dryajov?

arnetheduck: the cancelled futures remove themselves from the queue in their cancellation callback, so that _should_ never happen..
what could happen is that the user "manually" completes the future but that's not how one uses futures in general so..

Menduist: Right, missed that part

arnetheduck: if we can't figure out a reason, might be a good idea to assert with a message that the future was unexpectedly completed - by extension, a loop would no longer be needed - if this happens, I get the feeling we might get an imbalance in acquires and releases which either leads to lockups or overallocation

arnetheduck (approved):

libp2p/utils/semaphore.nim
dryajov: Since the counter and the queue size can now diverge due to the `forceAcquire` contraption, we might as well have to use a loop in `release`.

PR #697, opened 2022-02-23
fixes from #688

emizzle:
1. tests: invert expect message logic
2. export pubsub_errors for backward compatibility


Menduist (approved):
Thanks!

PR #699, opened 2022-03-02
WS Accept timeout

Menduist:
Merge https://github.com/status-im/nim-websock/pull/102 first

### DIFF
--- a/libp2p/transports/wstransport.nim
+++ b/libp2p/transports/wstransport.nim
@@ -338,4 +338,4 @@ proc new*(
     flags = flags,
     factories = @factories,
     rng = rng,
-    handshakeTimeout: handshakeTimeout)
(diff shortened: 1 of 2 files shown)


libp2p/transports/wstransport.nim
```diff
@@ -331,11 +331,11 @@ proc new*(
...
 
-  T(
-    upgrade: upgrade,
-    tlsPrivateKey: nil,
-    tlsCertificate: nil,
-    flags: flags,
-    factories: @factories,
-    rng: rng,
-    handshakeTimeout: handshakeTimeout)
+  T.new(
```
dryajov, line 334: You don't need to call `new` explicitly on a `ref` type, it will always be created on the heap.

Menduist: It's to call the procedure above

PR #700, opened 2022-03-09
Peer store refacto

Menduist:
Improves the Peer Store:

There is now a global PeerStore structure (instead of having one for libp2p, one for waku, etc)
To create a PeerBook:
```nim
type StateBook = ref object of PeerBook[MagicState]
switch.peerStore[StateBook][peerId] = Rocking
```

The operators have been nim-ified ([], []= & del)

And a pruner system, that will only keep a certain amount of dead peers in memory

Also, switch to seqs instead of HashSet for the inner books (eg `PeerBook[seq[MultiAddress]]`) because we generally don't have enough elements to justify the overhead of a HashSet

LMKWYT


libp2p/peerstore.nim
```diff
@@ -80,83 +75,82 @@ proc set*[T](peerBook: var PeerBook[T],
...
+
+  let cleanupPos = peerStore.toClean.find(info.peerId)
+  if cleanupPos >= 0:
+    peerStore.toClean.delete(cleanupPos)
+
+proc cleanup*(
+  peerStore: PeerStore,
+  peerId: PeerId) =
+
+  if peerStore.capacity <= 0:
```
dryajov, line 149: usually `-1` would mean infinite capasity, seems to be reversed here?

libp2p/connmanager.nim
```diff
@@ -307,6 +307,9 @@ proc peerCleanup(c: ConnManager, conn: Connection) {.async.} =
     await c.triggerConnEvent(
       peerId, ConnEvent(kind: ConnEventKind.Disconnected))
     await c.triggerPeerEvents(peerId, PeerEvent(kind: PeerEventKind.Left))
+
+    if not(c.peerStore.isNil):
+      c.peerStore.cleanup(peerId)
```
jm-clius, line 312: It makes sense to clean up peers that were disconnected. If Waku were to use the native `switch` peer store, we'd need to change some peer persistence logic though - we do currently track disconnected peers in our peer store (a `DisconnectBook`), mostly because it's a convenient way to determine a back off period before attempting to reconnect to such peers after a restart. I think we need to do something else here in any case, though, and can't think of any other good reasons to keep peers in the store after disconnection.

jm-clius (approved):
Makes sense to me. Thanks!

libp2p/connmanager.nim
Menduist: Yes, currently this will keep a certain amount of disconnected peers (peerStore.capacity which the app can choose), and then remove them from every book.

I was wondering if you needed more sophisticated logic (eg, have a per-book settings on how much dead peer to save)
Happy to brainstorm a bit on your needs :)

libp2p/peerstore.nim
Menduist: Added this,
> 0: regular capacity
0: no capacity
< 0: infinite

dryajov:
Whats missing here?

libp2p/builders.nim
```diff
@@ -187,6 +192,9 @@ proc build*(b: SwitchBuilder): Switch
   if isNil(b.rng):
     b.rng = newRng()
 
+  if isNil(b.peerStore):
```
dryajov, line 195: If `PeerStore` is required now, we don't need a `withPeerStore` anymore, at least not one that takes an instance only the non-default params needed to construct it? Otherwise, this construction flow doesn't make much sense.

dryajov:
LGTM, besides the convoluted construction flow. Needs to be changed before merging.

PR #702, opened 2022-03-17
simplify per-topic metric collection

cskiraly:
removing some of the repeated metrics code for per-topic metrics in PubSub and co.

### DIFF
--- a/libp2p/protocols/pubsub/pubsub.nim
+++ b/libp2p/protocols/pubsub/pubsub.nim
@@ -127,6 +127,15 @@ method unsubscribePeer*(p: PubSub, peerId: PeerId) {.base.} =
 
   libp2p_pubsub_peers.set(p.peers.len.int64)
 
+func topicMetricInc*(
+  p: PubSub,
+  topic: string,
+  counter: Collector | type IgnoredCollector,
+  amount: int64|float64 = 1) =
+  ## helper for increasing per-topic metrics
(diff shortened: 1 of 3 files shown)


Menduist:
Discussion on discord:
Tanguy — Today at 1:01 PM
Mhh, fair
My only issue with the current technique is that it only works for inc (and to be fair, that's all we have right now)

Maybe have a
`proc metricTopicLabel(p: PubSub, t: string): string = p.knownTopics.containsOr(topic, "generic")`


and then
`libp2p_pubsub_broadcast_subscriptions.inc(npeers, labelValues = p.metricTopicLabel(topic))`
?
cskiraly — Today at 2:15 PM
sounds like a good compromise.
Tanguy — Today at 2:18 PM
Btw, we could also remove the knownTopics from PubSub completely, it's just a copy from a constant in pubsub.nim
That may simplify the code a little bit (don't need the PubSub in metricTopicLabel)

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -127,6 +127,15 @@ method unsubscribePeer*(p: PubSub, peerId: PeerId) {.base.} =
 
   libp2p_pubsub_peers.set(p.peers.len.int64)
 
+func topicMetricInc*(
```
dryajov, line 130: maybe should make it a `template`?

dryajov:
very nice, much needed cleanup.

PR #704, opened 2022-04-04
Yamux

Menduist:
Mplex is getting pushed over by yamux: https://github.com/libp2p/specs/pull/402

So I'm starting to implem yamux here. Since it's not urgent, don't expect fast dev on it, I just do some bits from time to time

- [X] Wire encoding / decoding
- [X] Stream handling
- [X] Stream pressure management (window system)
- [x] Stream reset
- [x] Half closed states
- [x] Interop test


mxinden:
:wave:  author of https://github.com/libp2p/specs/pull/402 here.

I am sorry that we don't have a dedicated Yamux specification. Feel free to ping me in case there are any questions.

arnetheduck:
more interesting than yamux would be to finish quic which has its own stream/backpressure handling - mplex works "good enough" for now and if we're going to switch to something, yamux feels .. like a bit of an unnecessary step.

Menduist:
Quic can't be used everywhere (websocket etc), so we'll most likely always need a "manual" muxer, might as well have a slightly better one

Also, the Quic implem is currently "softly" blocked, because bearssl isn't currently supporting TLS1.3 (support is coming)

If someone need it urgently, we would have to wrap another SSL library, only to revert it back to bearssl once support is here, so it would be a bit of wasted work

(currently there is no official way to make Quic work with another crypto layer in libp2p)

libp2p/muxers/yamux/yamux.nim
```diff
@@ -198,55 +204,68 @@ proc setupHeader(channel: YamuxChannel, header: var YamuxHeader) =
...
+      channel.bytesSent.inc(toSend)
+      await channel.conn.write(sendBuffer)
+      channel.sentData.fire()
+    channel.isSending = false
 
 method write*(channel: YamuxChannel, msg: seq[byte]) {.async.} =
   if channel.closedLocally or channel.isReset:
     raise newLPStreamEOFError()
 
+  let blockUntil = channel.bytesSent + msg.len().uint64
```
Menduist, line 261:
```suggestion
  let blockUntil = channel.bytesSent + channel.sendQueueBytes(false) + msg.len().uint64
```

libp2p/muxers/yamux/yamux.nim
```diff
@@ -198,55 +204,68 @@ proc setupHeader(channel: YamuxChannel, header: var YamuxHeader) =
...
   channel.sendQueue.add(msg)
   await channel.trySend()
 
-  while channel.sendWindow == 0 and channel.sendQueue.len > 0:
+  while blockUntil > channel.bytesSent:
     # Block until there is room
```
Menduist, line 266:
```suggestion
    # Block until that has been sent room
```

libp2p/muxers/yamux/yamux.nim
```diff
@@ -0,0 +1,419 @@
...
+
+proc reset(channel: YamuxChannel) {.async.} =
+  if not channel.isReset:
+    channel.isReset = true
+    channel.sendQueue = @[]
+    channel.recvQueue = @[]
+    channel.sendWindow = 0
+    channel.recvWindow = 0
+    if not channel.closedRemotely.done():
+      await channel.conn.write(toSeq(YamuxHeader.data(channel.id, 0, {Rst}).encode()))
```
Menduist, line 151:
```suggestion
      await channel.conn.write(YamuxHeader.data(channel.id, 0, {Rst}))
```

tests/commoninterop.nim
```diff
@@ -0,0 +1,592 @@
...
+      proto.handler = nativeHandler
+      proto.codec = protos[0] # codec
+
+      let wsAddress = MultiAddress.init("/ip4/127.0.0.1/tcp/0/ws").tryGet()
+
+      let nativeNode = SwitchBuilder
+        .new()
+        .withAddress(wsAddress)
+        .withRng(crypto.newRng())
+        .withMplex()
```
Menduist, line 302: This will always use mplex

lchenut: I completely forgot this one... It's fix

libp2p/muxers/yamux/yamux.nim
```diff
@@ -0,0 +1,483 @@
...
+        if MsgFlags.Syn in header.flags:
+          if header.streamId in m.channels:
+            debug "Trying to create an existing channel, skipping", id=header.streamId
+          else:
+            if header.streamId in m.flushed:
+              m.flushed.del(header.streamId)
+            if header.streamId mod 2 == m.currentId mod 2:
+              raise newException(YamuxError, "Peer used our reserved stream id")
+            let newStream = m.createStream(header.streamId, false)
+            await newStream.write(@[])
```
Menduist, line 423:
```suggestion
            await newStream.open()
```

libp2p/muxers/yamux/yamux.nim
```diff
@@ -0,0 +1,483 @@
...
+          futures.add(fut)
+          channel.sendQueue.delete(0)
+        inBuffer.inc(bufferToSend)
+
+      trace "build send buffer", h = $header, msg=string.fromBytes(sendBuffer[12..^1])
+      channel.sendWindow.dec(toSend)
+      try: await channel.conn.write(sendBuffer)
+      except LPStreamEOFError as exc:
+        for fut in futures.items():
+          fut.fail(exc)
```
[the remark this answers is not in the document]
lchenut: No because the futures are deleted from the channel.sendQueue.

libp2p/muxers/yamux/yamux.nim
```diff
@@ -0,0 +1,485 @@
...
+        trace "send window empty"
+        if channel.sendQueueBytes(true) > channel.maxRecvWindow:
+          await channel.reset(true)
+        break
+      if channel.sendQueue[0].data.len == 0:
+        var header = YamuxHeader.data(channel.id, 0)
+        channel.setupHeader(header)
+        let sendBuffer = toSeq(header.encode())
+        trace "send empty buffer", h = $header
+        try: await channel.conn.write(sendBuffer)
```
Menduist, line 280: You really don't like my `channel.conn.write(header)` :smile:

lchenut: I tend to forget that this proc exists

Menduist:
We can merge this once CI is green, will need to follow up on the finishing touches (merge the relevant part of the mplex code)
Since this will touch on mplex, makes more sense to do in another PR instead of bloating this one

mxinden:
Great to see this merged! :tada:

PR #708, opened 2022-04-05
Optimize rebalanceMesh

Menduist:
closes https://github.com/status-im/nim-libp2p/issues/705

cc @arnetheduck, if you have a test bench already setup, could you check if this is now optimized enough?

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -311,6 +311,7 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
     prunes, grafts: seq[PubSubPeer]
     npeers = g.mesh.peers(topic)
     defaultMesh: HashSet[PubSubPeer]
+    backingOff = g.backingOff.getOrDefault(topic)
 
   if npeers  < g.parameters.dLow:
     trace "replenishing mesh", peers = npeers
@@ -329,7 +330,7 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
               # don't pick explicit peers
               it.peerId notin g.parameters.directPeers and
               # and avoid peers we are backing off
-              it.peerId notin g.backingOff.getOrDefault(topic):
+              it.peerId notin backingOff:
             result.add(it)
     var candidates =
       try: getCandidates(g.mesh[topic])
@@ -371,7 +372,7 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
               # don't pick explicit peers
               it.peerId notin g.parameters.directPeers and
               # and avoid peers we are backing off
-              it.peerId notin g.backingOff.getOrDefault(topic):
+              it.peerId notin backingOff:
             result.add(it)
 
     var candidates =
@@ -452,6 +453,8 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
         g.pruned(peer, topic)
         g.mesh.removePeer(topic, peer)
 
+      backingOff = g.backingOff.getOrDefault(topic)
+
   # opportunistic grafting, by spec mesh should not be empty...
   if g.mesh.peers(topic) > 1:
     var peers = toSeq(try: g.mesh[topic] except KeyError: raiseAssert "have peers")
@@ -474,7 +477,7 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
                 # don't pick explicit peers
                 it.peerId notin g.parameters.directPeers and
                 # and avoid peers we are backing off
-                it.peerId notin g.backingOff.getOrDefault(topic):
+                it.peerId notin backingOff:
               result.add(it)
 
               # by spec, grab only 2
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -317,24 +317,22 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
     trace "replenishing mesh", peers = npeers
     # replenish the mesh if we're below Dlo
 
-    # This is a workaround to get the HashSet from the table
-    # without creating a copy.
-    proc getCandidates(currentMesh: var HashSet[PubSubPeer]): seq[PubSubPeer] =
-      g.gossipSub.withValue(topic, peerList):
-        for it in peerList[]:
-          if
-              it.connected and
-              # avoid negative score peers
-              it.score >= 0.0 and
-              it notin currentMesh and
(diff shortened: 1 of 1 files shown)


arnetheduck:
> if you have a test bench already setup,

no formal test really, I just ran a mainnet node and collected metrics via vtune for a few minutes (for other reasons, the ticket was mostly a collateral observation) - `rebalanceMesh` appears often enough that it's hard to miss :)

arnetheduck:
the other way to phrase this is that every improvement is better than the status quo, which gets bad with lots of peers.

arnetheduck (dismissed):

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -310,18 +310,21 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
...
-      it.peerId notin g.parameters.directPeers and
-      # and avoid peers we are backing off
-      it.peerId notin g.backingOff.getOrDefault(topic)
-    )
+    var candidates: seq[PubSubPeer]
+    let currentMesh = g.mesh.getOrDefault(topic)
+    g.gossipSub.withValue(topic, peerList):
+      for it in peerList[]:
+        if
+          it notin currentMesh and
```
arnetheduck, line 318: this should be moved below the simple checks so as to avoid the (more expensive) hash table lookup

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -310,18 +310,21 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
...
+          it.connected and
+          # avoid negative score peers
+          it.score >= 0.0 and
+          # don't pick explicit peers
+          it.peerId notin g.parameters.directPeers and
+          # and avoid peers we are backing off
+          it.peerId notin g.backingOff.getOrDefault(topic):
+
+          candidates.add(it)
```
arnetheduck, line 327: as a general thing, I like when there's a 4-space indent of the if-stuff so that the 2-space indent of the code gets differentiated, indent-wise

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -446,22 +450,24 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
...
+            it.score >= median.score and
+            it notin currentMesh and
+            # don't pick explicit peers
+            it.peerId notin g.parameters.directPeers and
+            # and avoid peers we are backing off
+            it.peerId notin g.backingOff.getOrDefault(topic):
+              avail.add(it)
+
+              # by spec, grab only 2
+              if avail.len > 1:
```
arnetheduck, line 469: I do it the other way around :) conditions get 4 spaces, if block gets 2

Menduist:
On my test bench in prater, this PR is about 2.5 faster than the current implementation
It's still inefficient because it copies `npeers` & `backingOff`, but that's going to be tricky to optimize until we get the "viewtypes" from nim 1.6

libp2p/protocols/pubsub/gossipsub/behavior.nim
arnetheduck: well, out guide says 2 space indent per block - since the block of the  if is the code and not the condition, it's more logical to move the condition an extra level - same goes for `proc` definitions - body = 2 spaces, parameters = 4 spaces - it's probably the best one can do with indent-based languages which also is why python auto-formatters tend to do it this way

arnetheduck (approved):

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -310,22 +310,31 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
...
+              # avoid negative score peers
+              it.score >= 0.0 and
+              it notin currentMesh and
+              # don't pick explicit peers
+              it.peerId notin g.parameters.directPeers and
+              # and avoid peers we are backing off
+              it.peerId notin backingOff:
+            result.add(it)
+    var candidates =
+      try: getCandidates(g.mesh[topic])
```
dryajov, line 336: You shouldn't get `KeyError` if using `withValue`, as a consequence I don't think you need the `getCandidates` closure here.

dryajov: Ah, I see that you're accessing `g.mesh[topic]` which will throw, but a better approach is to use the two body `withValue`:

```nim
  ##   table.withValue(key, value) do:
  ##     # block is executed only if ``key`` in ``t``
  ##     value.name = "username"
  ##     value.uid = 1000
  ##   do:
  ##     # block is executed when ``key`` not in ``t``
  ##     raise newException(KeyError, "Key not found")
```

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -346,39 +355,46 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
...
+              it notin currentMesh and
+              # avoid negative score peers
+              it.score >= 0.0 and
+              # don't pick explicit peers
+              it.peerId notin g.parameters.directPeers and
+              # and avoid peers we are backing off
+              it.peerId notin backingOff:
+            result.add(it)
+
+    var candidates =
```
dryajov, line 378: Same comment as in https://github.com/status-im/nim-libp2p/pull/708/files#r875077380

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -446,22 +464,29 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
...
+                it.peerId notin g.parameters.directPeers and
+                # and avoid peers we are backing off
+                it.peerId notin backingOff:
+              result.add(it)
+
+              # by spec, grab only 2
+              if result.len > 1:
+                break
+
+      let avail =
```
dryajov, line 487: Same as in https://github.com/status-im/nim-libp2p/pull/708/files#r875077380

dryajov (approved):
LGTM, but I would rework the `try/except` flows with the double body `withValue` as indicated in https://github.com/status-im/nim-libp2p/pull/708/files#r875077380

libp2p/protocols/pubsub/gossipsub/behavior.nim
Menduist: The goal of the closure is to create a "lent variable", but since that's only possible in proc parameters, we have to create a closure
Not sure if withValue will work here (ie, not copy the seqs), I'll try

dryajov: `withValue` will avoid copies.

Menduist: Switched to withValue
Since it works with pointer, I switched everything to pointer, which is less safe, but avoids the closure :shrug:

PR #709, opened 2022-04-05
Gossipsub scoring fixes

Menduist:
We were lacking implementation for two parameters that changed the scoring completely:
decayInterval: was pegged to the heartbeat, so the decay speed was wrong
meshMessageDeliveriesWindow: every peer was gaining score for delivering, even if it was 5 minutes late

Also added a little check, to avoid peers sending us duplicate, hoping to get more points

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/scoring.nim
+++ b/libp2p/protocols/pubsub/gossipsub/scoring.nim
@@ -253,6 +253,12 @@ proc updateScores*(g: GossipSub) = # avoid async
 
   trace "updated scores", peers = g.peers.len
 
+proc scoringHeartbeat*(g: GossipSub) {.async.} =
+  while g.heartbeatRunning:
+    trace "running scoring heartbeat", instance = cast[int](g)
+    g.updateScores()
+    await sleepAsync(g.parameters.decayInterval)
+
 proc punishInvalidMessage*(g: GossipSub, peer: PubSubPeer, topics: seq[string]) =
   for tt in topics:
     let t = tt
--- a/libp2p/protocols/pubsub/gossipsub/scoring.nim
+++ b/libp2p/protocols/pubsub/gossipsub/scoring.nim
@@ -255,9 +255,10 @@ proc updateScores*(g: GossipSub) = # avoid async
 
 proc scoringHeartbeat*(g: GossipSub) {.async.} =
   while g.heartbeatRunning:
+    let sleepFuture = sleepAsync(g.parameters.decayInterval)
     trace "running scoring heartbeat", instance = cast[int](g)
     g.updateScores()
-    await sleepAsync(g.parameters.decayInterval)
+    await sleepFuture
 
 proc punishInvalidMessage*(g: GossipSub, peer: PubSubPeer, topics: seq[string]) =
   for tt in topics:
--- a/libp2p/protocols/pubsub/gossipsub/scoring.nim
+++ b/libp2p/protocols/pubsub/gossipsub/scoring.nim
@@ -13,7 +13,7 @@ import std/[tables, sets, options]
 import chronos, chronicles, metrics
 import "."/[types]
 import ".."/[pubsubpeer]
-import "../../.."/[peerid, multiaddress, utility, switch]
+import "../../.."/[peerid, multiaddress, utility, switch, utils/heartbeat]
 
 declareGauge(libp2p_gossipsub_peers_scores, "the scores of the peers in gossipsub", labels = ["agent"])
 declareCounter(libp2p_gossipsub_bad_score_disconnection, "the number of peers disconnected by gossipsub", labels = ["agent"])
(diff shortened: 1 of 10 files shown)


libp2p/protocols/pubsub/gossipsub/scoring.nim
```diff
@@ -253,6 +253,12 @@ proc updateScores*(g: GossipSub) = # avoid async
 
   trace "updated scores", peers = g.peers.len
 
+proc scoringHeartbeat*(g: GossipSub) {.async.} =
+  while g.heartbeatRunning:
+    trace "running scoring heartbeat", instance = cast[int](g)
+    g.updateScores()
+    await sleepAsync(g.parameters.decayInterval)
```
arnetheduck, line 260: not the most important thing, but writing sleep loops this way introduces drift - ideally it would do something like `next = next + interval; sleep(now-next)` but guarded against negatives or the scoring will be off (same for heartbeat)

libp2p/protocols/pubsub/gossipsub/scoring.nim
```diff
@@ -253,6 +253,12 @@ proc updateScores*(g: GossipSub) = # avoid async
 
   trace "updated scores", peers = g.peers.len
 
+proc scoringHeartbeat*(g: GossipSub) {.async.} =
+  while g.heartbeatRunning:
```
arnetheduck, line 257: `await  sleepasync` will raise on `scoringHeartbeatFut.cancel` so the boolean is redundant

libp2p/protocols/pubsub/gossipsub/scoring.nim
Menduist: Indeed, I would have done the proper loop if there was any await here, but you're right, better safe than sorry
Fixed

libp2p/protocols/pubsub/gossipsub/scoring.nim
arnetheduck: `while not scoringHeartbeatFut.finished` then

arnetheduck (dismissed):

arnetheduck (approved):

libp2p/protocols/pubsub/gossipsub/scoring.nim
Menduist: Actually, the drift was causing issues in CI, which is rather unexpected since this should be a very quick process
Maybe there is some jitter in sleepAsync, or somewhere else

Anyway, I've added a proper heartbeat system, which takes care of the drift properly

Menduist:
"Fun" discovery of the day, MacOs seems to have a really bad scheduler, which is really jittery when it comes to timing things:
```c
#include <time.h>
#include <stdio.h>
#include <unistd.h>
int main()
{
        while (1) {
                struct timespec tstart={0, 0}, tend={0,0};
                clock_gettime(CLOCK_MONOTONIC, &tstart);
                usleep(100000);
                clock_gettime(CLOCK_MONOTONIC, &tend);
                if (tend.tv_nsec > tstart.tv_nsec)
                        printf("%.5f ms\n", (1.0e-6*tend.tv_nsec) - (1.0e-6*tstart.tv_nsec));
        }
}
```

Output on linux:
```
100.12988 ms
100.07913 ms
100.12629 ms
100.12398 ms
100.12010 ms
100.07552 ms
100.12226 ms
100.09175 ms
```
Up to 0.13 ms of delay, MacOs (on a M1 laptop):
```
105.03300 ms
101.10400 ms
101.46400 ms
100.52100 ms
105.01700 ms
102.77800 ms
```
Up to 5 ms (I've seen up to 7 in a VM)

We see the same random delays in chronos with `asyncSleep`, and my test were sleeping for 10ms, so a 7 ms delay completely throw them off.
Fixed my test to sleep for at least 30ms

arnetheduck:
> Fixed my test to sleep for at least 30ms

you'll get to the point eventually where you'll realise that sleep and deterministic tests don't go well together - this is just a step on that journey :stuck_out_tongue:

libp2p/utils/heartbeat.nim
```diff
@@ -0,0 +1,27 @@
...
+{.push raises: [Defect].}
+
+import sequtils
+import chronos, chronicles
+
+export chronicles
+
+template heartbeat*(name: string, interval: Duration, body: untyped): untyped =
+  var nextHeartbeat = Moment.now()
+  while true:
```
dryajov, line 19: Should there be a flag instead of just `true`?

dryajov:
> "Fun" discovery of the day, MacOs seems to have a really bad scheduler, which is really jittery when it comes to timing things:

Why do you think we've set most timeout to 500ms+ in our tests :) 

On a serious note, nice find, I suspected this but never confirmed it.

libp2p/utils/heartbeat.nim
Menduist: I think it's easier / less error prone to only rely on cancellation. +, in this template we're forced to await on each iteration, so a "busy loop" is not possible

Menduist:
Merging
If the tests causes CI flakiness, we can always remove them, but can't see how to do them without sleeping, since the goal of the test is precisely to test the timing :/

PR #717, opened 2022-05-11
Circuit relay v2

lchenut:
### _OUT OF DATE_
Just a little comment here, I'll write a more detailed one by the end of the week:
The circuit relay v2 is, on theory, nearly finished. The last things to do are listed at the beginning of the file. And I will (obviously) add a lot of tests but for this current version you could try it by using a code like this:

```nim
proc main() {.async, gcsafe.} =
  let
    rng = newRng()
    src = SwitchBuilder.new()
      .withRng(rng)
      .withAddresses(@[ MultiAddress.init("/ip4/127.0.0.1/tcp/0").tryGet() ])
      .withTcpTransport()
      .withMplex()
      .withNoise()
      .build()
    rel = SwitchBuilder.new()
      .withRng(rng)
      .withAddresses(@[ MultiAddress.init("/ip4/127.0.0.1/tcp/0").tryGet() ])
      .withTcpTransport()
      .withMplex()
      .withNoise()
      .build()
    dst = SwitchBuilder.new()
      .withRng(rng)
      .withAddresses(@[ MultiAddress.init("/ip4/127.0.0.1/tcp/0").tryGet() ])
      .withTcpTransport()
      .withMplex()
      .withNoise()
      .build()

  const CustomProtoCodec = "/test"
  var proto = new LPProtocol
  proto.handler = proc(conn: Connection, proto: string) {.async.} =
    echo "Dst Rcv: ", string.fromBytes(await conn.readLp(1024))
    await conn.writeLp("test 2")
    await conn.close()
  proto.codec = CustomProtoCodec

  let clSrc = Client.new(src)
  let rv2 = RelayV2.new(rel)
  let clDst = Client.new(dst)

  src.mount(clSrc)
  rel.mount(rv2)
  dst.mount(clDst)

  dst.mount(proto)

  # this part is temporary
  src.transports &= RelayV2Transport.new(clSrc, src.transports[0].upgrader)
  ((Dialer)src.dialer).transports = src.transports
  dst.transports &= RelayV2Transport.new(clDst, dst.transports[0].upgrader)
  ((Dialer)dst.dialer).transports = dst.transports
 
  await rv2.start()
  await src.start()
  await rel.start()
  await dst.start()

  await src.connect(rel.peerInfo.peerId, rel.peerInfo.addrs)
  await rel.connect(dst.peerInfo.peerId, dst.peerInfo.addrs)

  # One of the TODO, create a "RelayManager" or something to re-reserve at the end of the TTL
  discard await clDst.reserve(rel.peerInfo)

  let ma = MultiAddress.init($rel.peerInfo.addrs[0] & "/p2p/" & $rel.peerInfo.peerId & "/p2p-circuit/p2p/" & $dst.peerInfo.peerId).tryGet()
  var conn: Connection
  try:
    conn = await src.dial(dst.peerInfo.peerId, @[ ma ], CustomProtoCodec)
  except:
    echo getCurrentExceptionMsg()

  await conn.writeLp("test 1")
  echo "Src Rcv: ", string.fromBytes(await conn.readLp(1024))

  await allFutures(src.stop(), rel.stop(), dst.stop())

waitFor(main())
```


libp2p/protocols/relayv2/relayv2.nim
```diff
@@ -0,0 +1,860 @@
...
+  trace "reserving relay slot for", pid
+
+  let msg = rv2.createHopMessage(pid, expire)
+  if msg.isErr():
+    trace "error signing the voucher", error = error(msg), pid
+    # conn.reset()
+    return
+  let pb = encodeHopMessage(msg.get())
+
+  try:
```
Menduist, line 359: We can probably get rid of this try / except, since the one in handleStream does the same thing
(may be valid for other try / excepts)

libp2p/protocols/relayv2/relayv2.nim
```diff
@@ -0,0 +1,860 @@
...
+    return
+
+  let msgRcvFromDstOpt = try:
+    decodeStopMessage(await connDst.readLp(rv2.msgSize))
+  except CancelledError as exc:
+    raise exc
+  except CatchableError as exc:
+    trace "error reading stop response", exc=exc.msg
+    await handleHopError(conn, RelayV2Status.ConnectionFailed)
+    return
```
Menduist, line 441: We can merge theses two

libp2p/protocols/relayv2/relayv2.nim
```diff
@@ -0,0 +1,860 @@
...
+    trace "unexpected stop response, not a status message",
+          msgType = msgRcvFromDst.msgType
+    await handleHopError(conn, RelayV2Status.ConnectionFailed)
+    return
+
+  if msgRcvFromDst.status.isNone() or
+     msgRcvFromDst.status.get() != RelayV2Status.Ok:
+    trace "relay stop failure", status = msgRcvFromDst.status
+    await handleHopError(conn, RelayV2Status.ConnectionFailed)
+    return
```
Menduist, line 459: Actually, from dial to here could be put in a separate procedure (eg stopPeer) without any try / expect / if isNone
and handle the failure only once

libp2p/protocols/relayv2/relayv2.nim
```diff
@@ -0,0 +1,860 @@
...
+    trace "end relaying", bytesSendFromSrcToDst, bytesSendFromDstToSrc
+
+    await futSrc.cancelAndWait()
+    await futDst.cancelAndWait()
+  await bridge(RelayConnection.new(conn, rv2.limit.duration, rv2.limit.data),
+               RelayConnection.new(connDst, rv2.limit.duration, rv2.limit.data))
+
+
+proc heartbeat(rv2: RelayV2) {.async.} =
+  while true:
```
Menduist, line 520: I've just merged a `heartbeat` helper in unstable, see here: https://github.com/status-im/nim-libp2p/blob/991549f3913edeac6c34ee0c07bab31365d53c48/libp2p/protocols/pubsub/gossipsub.nim#L570

libp2p/protocols/relayv2/relayv2.nim
```diff
@@ -0,0 +1,860 @@
...
+    msgSize: msgSize)
+
+  proc handleStream(conn: Connection, proto: string) {.async, gcsafe.} =
+    try:
+      let msgOpt = decodeHopMessage(await conn.readLp(rv2.msgSize))
+
+      if msgOpt.isNone():
+        await handleHopError(conn, RelayV2Status.MalformedMessage)
+        return
+      else:
```
Menduist, line 556: No need for this else

libp2p/protocols/relayv2/relayv2.nim
```diff
@@ -0,0 +1,860 @@
...
+
+  rv2.heartbeatFut.cancel()
+
+# Client side
+
+type
+  Client* = ref object of LPProtocol
+    switch: Switch
+    queue: AsyncQueue[Connection]
+  ReserveError* = enum
```
Menduist, line 597: Not used

libp2p/protocols/relayv2/relayv2.nim
```diff
@@ -0,0 +1,860 @@
...
+    ReserveFailedConnection,
+    ReserveMalformedMessage,
+    ReserveBadType,
+    ReserveBadStatus,
+    ReserveMissingReservation,
+    ReserveBadExpirationDate,
+    ReserveInvalidVoucher,
+    ReserveUnexpected
+
+proc handleStopError(conn: Connection, code: RelayV2Status) {.async.} =
```
Menduist, line 608: rename to sendStopError or similar

libp2p/protocols/relayv2/relayv2.nim
```diff
@@ -0,0 +1,860 @@
...
+    limitDuration = msg.limit.get().duration
+    limitData = msg.limit.get().data
+
+  trace "incoming relay connection", src, conn
+
+  let
+    msg = StopMessage(
+      msgType: StopMessageType.Status,
+      peer: none(RelayV2Peer),
+      limit: none(RelayV2Limit),
```
Menduist, line 648: Can be removed

libp2p/protocols/relayv2/relayv2.nim
```diff
@@ -0,0 +1,860 @@
...
+
+proc init*(cl: Client) =
+  proc handleStream(conn: Connection, proto: string) {.async, gcsafe.} =
+    try:
+      let msgOpt = decodeStopMessage(await conn.readLp(MsgSize))
+
+      if msgOpt.isNone():
+        await handleHopError(conn, RelayV2Status.MalformedMessage)
+        return
+      else:
```
Menduist, line 670: no need for this else

libp2p/protocols/relayv2/relayv2.nim
```diff
@@ -0,0 +1,860 @@
...
+    raise exc
+  except CatchableError as exc:
+    trace "error opening relay stream", exc=exc.msg, relayPI
+    raise newException(ReservationError, exc.msg)
+  defer:
+    await conn.close()
+
+  try:
+    await conn.writeLp(pb.buffer)
+    msgOpt = decodeHopMessage(await conn.readLp(MsgSize))
```
Menduist, line 721:
```suggestion
  let msg = try:
    await conn.writeLp(pb.buffer)
    decodeHopMessage(await conn.readLp(MsgSize)).get()
```

libp2p/protocols/relayv2/relayv2.nim
```diff
@@ -0,0 +1,860 @@
...
+  except CatchableError as exc:
+    trace "error writing reservation message", exc=exc.msg
+    raise newException(ReservationError, exc.msg)
+
+  if msgOpt.isNone():
+    trace "Malformed message from relay"
+    raise newException(ReservationError, "Malformed message from relay")
+  msg = msgOpt.get()
+  if msg.msgType != HopMessageType.Status:
+    trace "unexpected relay response type", msgType = msg.msgType
```
Menduist, line 733: All theses trace can be removed, since the exception handler will show the error anyway

Menduist (commented):
Nice!

lchenut:
Here we are, the circuit relay v2!
First things first, here's how you make it work
```nim
import chronos
import stew/byteutils
import libp2p/[switch, builders]
import libp2p/protocols/relay/[relay, client]

proc main() {.async.} =
  let customProtoCodec = "/test"
  var proto = new LPProtocol
  proto.codec = customProtoCodec
  proto.handler = proc(conn: Connection, proto: string) {.async.} =
    assert "test1" == string.fromBytes(await conn.readLp(1024))
    await conn.writeLp("test2")
    assert "test3" == string.fromBytes(await conn.readLp(1024))
    await conn.writeLp("test4")

  let
    clSrc = Client.new()
    clDSt = Client.new()
    swSrc = SwitchBuilder.new()
      .withRng(newRng())
      .withAddresses(@[ MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet() ])
      .withTcpTransport()
      .withMplex()
      .withNoise()
      .withCircuitRelayTransport(clSrc)
      .build()
    swDst = SwitchBuilder.new()
      .withRng(newRng())
      .withAddresses(@[ MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet() ])
      .withTcpTransport()
      .withMplex()
      .withNoise()
      .withCircuitRelayTransport(clDst)
      .build()
    swRel = newStandardSwitch()
    relay = Relay.new(swRel)
    addrs = MultiAddress.init($swRel.peerInfo.addrs[0] & "/p2p/" &
                              $swRel.peerInfo.peerId & "/p2p-circuit/p2p/" &
                              $swDst.peerInfo.peerId).get()

  swRel.mount(relay)
  swDst.mount(proto)

  await relay.start()
  await swRel.start()
  await swSrc.start()
  await swDst.start()

  await swSrc.connect(swRel.peerInfo.peerId, swRel.peerInfo.addrs)
  await swDst.connect(swRel.peerInfo.peerId, swRel.peerInfo.addrs)

  let rsvp = await clDst.reserve(swRel.peerInfo.peerId, swRel.peerInfo.addrs)
  let conn = await swSrc.dial(swDst.peerInfo.peerId, @[ addrs ], customProtoCodec)

  await conn.writeLp("test1")
  assert "test2" == string.fromBytes(await conn.readLp(1024))
  await conn.writeLp("test3")
  assert "test4" == string.fromBytes(await conn.readLp(1024))

  await allFutures(relay.stop(), swSrc.stop(), swDst.stop(), swRel.stop())

waitFor(main())
```
Few notes:
- The Relay LPProtocol is a merge between the RelayV1 and the RelayV2, it can establish a relayed connection between two nodes using both relayv1 and relayv2 protocols (and manage reservation for the V2)
- Similarly, the Client LPProtocol can dial with (or accept a connection from) a relayv1 (or v2), as well as reserve a slot on a relayv2.
- The Relay.start / Relay.stop procedure are only there to start / stop the deletesReservation procedure (for the RelayV2), there's maybe a more elegant solution but I didn't find one satisfying.
- The RelayTransport tries to dial with the relayV2 protocol first, then, if it fails, the v1.
- Some things related to the reservation are weird, such as: the voucher is useless right know even according to the spec `The intention is to eventually require the use of reservation vouchers for dialing relay addresses, but this is not currently enforced so the vouchers are only advisory.`; one node that reserves a slot on a relay could theoretically get an infinite number of connection during the times of the reservation instead of one reservation = one connection; the relay have to send the limit (in time and data) of a connection during the reservation but it re-send it again at the beginning of a relayed connection.

On my implementation on itself: 
- I find the way I set and check a maximum of relayed connection a bit unsatisfying, [especially comparing it with the go one](https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/circuitv2/relay/constraints.go).
- I didn't write the interop test for the V2 yet, because I don't think there's a way to setup a relayv2 node (or client relayv2) currently with the go-libp2p daemon.
- I used std/times instead of chronos/timer for all the reservation part because it's impossible with chronos to get an UTC UNIX time.
- There's currently no way to automate the re-reservation, a ClientManager could be a solution.

libp2p/builders.nim
```diff
@@ -49,8 +50,7 @@ type
     agentVersion: string
     nameResolver: NameResolver
     peerStoreCapacity: Option[int]
-    isCircuitRelay: bool
-    circuitRelayCanHop: bool
+    circuitRelayClient: Client
```
Menduist, line 53: Can't we be in a "relay only" scenario? (with no client)

libp2p/protocols/relay/client.nim
```diff
@@ -0,0 +1,266 @@
...
+import chronos, chronicles, options
+import ./messages,
+       ./rconn,
+       ./utils,
+       ../../peerinfo,
+       ../../switch,
+       ../../multiaddress,
+       ../../stream/connection
+
+import std/times
```
Menduist, line 19: Nitpick, but generally, we import in this order:
- std
- packages
- local files
(also applies to the `options` above)

libp2p/protocols/relay/client.nim
```diff
@@ -0,0 +1,266 @@
...
+
+logScope:
+  topics = "libp2p circuit relay client"
```
Menduist, line 22: Each word in this string is a different topic, so it should be `libp2p relay relay-client` or equivalent

libp2p/protocols/relay/client.nim
```diff
@@ -0,0 +1,266 @@
...
+
+type
+  ClientError* = object of LPError
+  ReservationError* = object of ClientError
+  RelayV1DialError* = object of ClientError
+  RelayV2DialError* = object of ClientError
+  ClientAddConn* = proc(conn: Connection,
+                        duration: uint32,
+                        data: uint64): Future[void] {.gcsafe, raises: [Defect].}
+  Client* = ref object of LPProtocol
```
Menduist, line 34: `Client` is a too generic name for something exposed to users, consider renaming to `RelayClient`

Also, mounting the RelayV1 twice (once here and once in `Relay`) is going to cause issues when you want to both be a client & server

Menduist (changes requested):
See my last comment,
we need to enable 3 scenarios:
- Just a client
- Just a server
- Server & client

libp2p/protocols/relay/client.nim
```diff
@@ -0,0 +1,290 @@
...
+    voucher*: Option[Voucher] # optional, reservation voucher
+    limitDuration*: uint32    # seconds
+    limitData*: uint64        # bytes
+
+proc sendStopError(conn: Connection, code: StatusV2) {.async.} =
+  trace "send stop status", status = $code & " (" & $ord(code) & ")"
+  let msg = StopMessage(msgType: StopMessageType.Status, status: some(code))
+  await conn.writeLp(encode(msg).buffer)
+
+proc handleConnect(cl: RelayClient, conn: Connection, msg: StopMessage) {.async.} =
```
Menduist, line 53: It's a bit confusing to have two procedures named `handleConnect` (one here, one in the relay)
Maybe this one could be called `handleRelayedConnection`?

libp2p/protocols/relay/client.nim
```diff
@@ -0,0 +1,290 @@
...
+    pb = encode(msg)
+
+  trace "incoming relay connection", src
+
+  if cl.addConn == nil:
+    await sendStopError(conn, StatusV2.ConnectionFailed)
+    await conn.close()
+    return
+  await conn.writeLp(pb.buffer)
+  # This sound redundant but the callback could, in theory, be set to nil during
```
Menduist, line 75: could be set to nil

libp2p/protocols/relay/client.nim
```diff
@@ -0,0 +1,290 @@
...
+  # conn.writeLp so it's safer to double check
+  if cl.addConn != nil: await cl.addConn(conn, limitDuration, limitData)
+  else: await conn.close()
+
+proc reserve*(cl: RelayClient, peerId: PeerId, addrs: seq[MultiAddress]): Future[Rsvp] {.async.} =
```
Menduist, line 80: `addrs` should be optional (if we're already connected to the peer)

libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -247,6 +247,9 @@ method write*(s: LPChannel, msg: seq[byte]): Future[void] =
 
   s.completeWrite(fut, msg.len)
 
+method isCircuitRelay*(s: LPChannel): bool =
```
dryajov, line 250: nitpick, but I would name this `relayed`.

libp2p/dial.nim
```diff
@@ -17,6 +17,10 @@ import peerid,
 type
   Dial* = ref object of RootObj
 
+  DialRes* = tuple
```
dryajov, line 20: What is the reason to return the dialed protocol?

lchenut: For this v2 version, I merged the relayv1 and the relayv2, and they use two (three actually, but let's say two) different protocols. So I need to know what protocol is used when I dial in the [relay transport](https://github.com/lchenut/nim-libp2p/blob/566223a05ca0f30c1b0801df3748ec3c719787cf/libp2p/protocols/relay/rtransport.nim#L67-L78) for example.

examples/circuitrelay.nim
```diff
@@ -0,0 +1,76 @@
...
+    swDst = createCircuitRelaySwitch(clDst)
+
+    # Create a relay address to swDst using swRel as the relay
+    addrs = MultiAddress.init($swRel.peerInfo.addrs[0] & "/p2p/" &
+                              $swRel.peerInfo.peerId & "/p2p-circuit/p2p/" &
+                              $swDst.peerInfo.peerId).get()
+
+  swDst.mount(proto)
+
+  await relay.start()
```
Menduist, line 50: I'm not very fond of this new / start / stop.

Other idea:
- We `start` in `new`
- We add a `stop` method to the LPProtocol (can also be used for the gossipsub stop)
- Either in the switch stop or a newly created multistream.stop, we stop each mounted protocol

libp2p/builders.nim
```diff
@@ -152,9 +151,8 @@ proc withNameResolver*(b: SwitchBuilder, nameResolver: NameResolver): SwitchBuil
   b.nameResolver = nameResolver
   b
 
-proc withRelayTransport*(b: SwitchBuilder, canHop: bool): SwitchBuilder =
-  b.isCircuitRelay = true
-  b.circuitRelayCanHop = canHop
+proc withCircuitRelay*(b: SwitchBuilder, r: Relay): SwitchBuilder =
```
Menduist, line 154: If we do as my last comment suggest, we could do `r = Relay.new()`, since for relay only setup, the user doesn't need the relay

libp2p/muxers/mplex/lpchannel.nim
Menduist: Instead of this method, we could have a `getWrappedConnection`, which would return `s.conn` in this case. It would be more versatile

libp2p/protocols/relay/client.nim
```diff
@@ -0,0 +1,290 @@
...
+
+  if msg.msgType != HopMessageType.Status:
+    raise newException(ReservationError, "Unexpected relay response type")
+  if msg.status.get(UnexpectedMessage) != Ok:
+    raise newException(ReservationError, "Reservation failed")
+  if msg.reservation.isNone():
+    raise newException(ReservationError, "Missing reservation information")
+
+  let reservation = msg.reservation.get()
+  if now().utc > reservation.expire.int64.fromUnix.utc:
```
Menduist, line 102: Check this line for possible defects (for instance, I suspect a reservation.expire > int64.max will crash, maybe fromUnix & utc also have some shenanigans)

libp2p/protocols/relay/client.nim
```diff
@@ -0,0 +1,290 @@
...
+  let msgRcvFromRelay = msgRcvFromRelayOpt.get()
+  if msgRcvFromRelay.msgType.isNone or msgRcvFromRelay.msgType.get() != RelayType.Status:
+    trace "unexcepted relay stop response", msgType = msgRcvFromRelay.msgType
+    await sendStatus(conn, StatusV1.HopCantOpenDstStream)
+    raise newException(RelayV1DialError, "Hop can't open destination stream")
+
+  if msgRcvFromRelay.status.isNone or msgRcvFromRelay.status.get() != StatusV1.Success:
+    trace "relay stop failure", status=msgRcvFromRelay.status
+    await sendStatus(conn, StatusV1.HopCantOpenDstStream)
+    raise newException(RelayV1DialError, "Hop can't open destination stream")
```
Menduist, line 161:
```suggestion
  try:
    if msgRcvFromRelayOpt.isNone:
      raise newException(RelayV1DialError, "Hop can't open destination stream")
      
    let msgRcvFromRelay = msgRcvFromRelayOpt.get()
    if msgRcvFromRelay.msgType.get(RelayStatus.CanHop) != RelayType.Status:
      raise newException(RelayV1DialError, "Hop can't open destination stream: wrong msg type")
      
    if msgRcvFromRelay.status.get(StatusV1.MalformedMessage) != StatusV1.Success:
      raise newException(RelayV1DialError, "Hop can't open destination stream: status failed")
  except RelayV1DialError as exc:
    await sendStatus(conn, StatusV1.HopCantOpenDstStream)
    raise exc
```

libp2p/muxers/mplex/lpchannel.nim
lchenut: I changed it to getWrapped and I used it [here](https://github.com/status-im/nim-libp2p/pull/717/files#diff-bd5997eecbb00b0d3a8a5121afba90ad5d7800037d398c5d0da2bd9663a53414R98-R104), isCircuitRelay/isRelayed could not really be re-used, getWrapped could eventually be.

libp2p/protocols/relay/client.nim
```diff
@@ -0,0 +1,291 @@
+## Nim-LibP2P
+## Copyright (c) 2022 Status Research & Development GmbH
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
```
Menduist, line 8:
```suggestion
# Nim-LibP2P
# Copyright (c) 2022 Status Research & Development GmbH
# Licensed under either of
#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
# at your option.
# This file may not be copied, modified, or distributed except according to
# those terms.
```

Menduist: Same for other files

Menduist (approved):

Menduist:
Let's fix the CI first

Menduist (approved):

PR #724, opened 2022-05-27
Faster CI

Menduist:
Currently, we compile once with TRACE & json logs, and a second time to actually run the test.
Instead, this branch will compile once with TRACE & json, and use runtime filtering to only show NOTICE and above

~~Also, use choosenim instead of recompiling the Nim compiler every time~~

Lastly, included a small fix for codecov

Overall, the CI is about 40% faster


dryajov:
I'm not sure about `choosenim`, if anything we should cache the compiler build and only recompile when needed.

Menduist:
Caching can be helpful, but is quite limited: it's limited to the current branch (meaning each new PR will have to rebuild everything), and only last for a week (most of our PRs stays open for multiple weeks / months)

What's the worry with choosenim? Note that the daily job still compile from scratch, with more nim versions, every day

arnetheduck:
> What's the worry with choosenim? Note that the daily job still compile from scratch, with more nim versions, every day

we keep our own nim repo/branch such that when a problem happens, we have the infrastructure ready on standby without waiting for upstream - don't think choosenim can be convinced to build / use that

arnetheduck:
> break our CI

the scenario is a supply chain issue in which the repo is compromised somehow in a way which doesn't break the build but introduces a vulnerability - ie we audit every nim upgrade "manually", verifying that nothing undue has sneaked in (such as, for example, an increased risk of hash table collisions if we're not careful with the `hash` changes in 1.6)

Menduist:
That may be the case for nimbus, but for nim-libp2p, we don't control which nim version the user is gonna choose anyway

Menduist:
Moved the choosenim thing to a new PR

I still think it make sense to use whatever fastest as a "push" CI, and do deeper / broader tests in the daily CI

PR #725, opened 2022-06-01
feat(transport): add experimental QUIC Transport (not production ready)

Menduist:
Our quic effort is blocked by bearssl not supporting TLS1.3, but since Mark did most of the work to implement Quic here: https://github.com/status-im/nim-libp2p/pull/563 and on nim-quic, this PR is going to bring encryption-less Quic into nim-libp2p
This allows us to test it, and make sure it doesn't bitrot.

Heavily WiP:
- [X] Extract code from #563
- [X] Create custom muxer & upgrader
- [X] Basic E2E switch test working
- [x] Update nim-quic to get address informations in libp2p (for `observed address` and port 0 resolving)
- [ ] More tests
- [ ] Cleanup

Since encryption is not yet supported, we're not compatible with any other libp2ps, and have to rely on home made protocols to retrieve the peer's id


libp2p/transports/quictransport.nim
```diff
@@ -0,0 +1,242 @@
...
+    body
+  except QuicError:
+    raise newLPStreamEOFError()
+
+method readOnce*(stream: QuicStream,
+                 pbytes: pointer,
+                 nbytes: int): Future[int] {.async.} =
+  if stream.cached.len == 0:
+    stream.cached = await mapExceptions(stream.stream.read())
+  if stream.cached.len <= nbytes:
```
dryajov, line 47: `readOnce` is best effort, it reads up to `nbytes` and returns as much as it could read. We should simply return whatever `stream.stream.read()` returns...

Menduist:
Unfortunately, nim-quic relies on questionable which doesn't work on nim devel
(ref https://github.com/status-im/questionable/issues/17)

So even though this branch works, it should not be merged until this get fixed
(otherwise, this will break our daily CI tests)

dryajov:
We should not be using `questionable` in a low level library such as `nim-quic`, in fact, we should keep the amount of external dependencies to an absolute minimum.

Questionable and the likes, are totally fine in a high level application, where the amount of "busyness" related logic might become high (Codex, Nimbus, etc...), but not in a low level dependency such as `libp2p` or a transport.

libp2p/transports/quictransport.nim
```diff
@@ -0,0 +1,224 @@
...
+# Stream
+type QuicStream* = ref object of P2PConnection
+  stream: Stream
+  cached: seq[byte]
+
+proc new(
+    _: type QuicStream, stream: Stream, oaddr: Opt[MultiAddress], peerId: PeerId
+): QuicStream =
+  let quicstream = QuicStream(stream: stream, observedAddr: oaddr, peerId: peerId)
+  procCall P2PConnection(quicstream).initStream()
```
lchenut, line 34: You should specify an objName for `quicstream` before calling `initStream`. https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/stream/connection.nim#L56-L57

libp2p/transports/quictransport.nim
```diff
@@ -0,0 +1,224 @@
...
+  quicstream
+
+template mapExceptions(body: untyped) =
+  try:
+    body
+  except QuicError:
+    raise newLPStreamEOFError()
+  except CatchableError:
+    raise newLPStreamEOFError()
```
lchenut, line 43: After a recent discussion with @Ivansete-status, I think those error need a small rework (not in this PR though). But nevertheless, you should try to use [one of those error](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/stream/lpstream.nim#L52-L61) here

libp2p/transports/quictransport.nim
```diff
@@ -0,0 +1,224 @@
...
+method closeImpl*(stream: QuicStream) {.async: (raises: []).} =
+  try:
+    await stream.stream.close()
+  except CatchableError as exc:
+    discard
+  await procCall P2PConnection(stream).closeImpl()
+
+# Session
+type QuicSession* = ref object of P2PConnection
+  connection: QuicConnection
```
lchenut, line 74: I'm not sure to understand why you need two P2PConnections here (with QuicSession and QuicStream). It's a bit confusing.

And I think it's better to create your custom `Connection` in libp2p/streams instead of here.

libp2p/transports/quictransport.nim
```diff
@@ -0,0 +1,224 @@
...
+logScope:
+  topics = "libp2p quictransport"
+
+type
+  P2PConnection = connection.Connection
+  QuicConnection = quic.Connection
+
+# Stream
+type QuicStream* = ref object of P2PConnection
+  stream: Stream
```
lchenut, line 27: I'm not a huge fan of this un-prefixed `Stream`. Maybe renaming it `QuicStream` and renaming the current `QuicStream` into `LPQuicStream` will be clearer.

libp2p/transports/quictransport.nim
```diff
@@ -0,0 +1,224 @@
...
+method getWrapped*(self: QuicSession): P2PConnection =
+  nil
+
+# Muxer
+type QuicMuxer = ref object of Muxer
+  quicSession: QuicSession
+  handleFut: Future[void]
+
+method newStream*(
+    m: QuicMuxer, name: string = "", lazy: bool = false
```
lchenut, line 101: I don't think the `lazy` flag is an option here. Maybe add a log if someone tries to use it.

libp2p/transports/quictransport.nim
```diff
@@ -0,0 +1,224 @@
...
+    doAssert(chann.closed, "connection not closed by handler!")
+  except CatchableError as exc:
+    trace "Exception in mplex stream handler", msg = exc.msg
+    await chann.close()
+
+method handle*(m: QuicMuxer): Future[void] {.async: (raises: []).} =
+  try:
+    while not m.quicSession.atEof:
+      let incomingStream = await m.quicSession.getStream(Direction.In)
+      asyncSpawn m.handleStream(incomingStream)
```
lchenut, line 125: If you find an alternative to the asyncspawn, it would be great.

libp2p/transports/quictransport.nim
```diff
@@ -0,0 +1,224 @@
...
+    await m.quicSession.close()
+    m.handleFut.cancel()
+  except CatchableError as exc:
+    discard
+
+# Transport
+type QuicUpgrade = ref object of Upgrade
+
+type QuicTransport* = ref object of Transport
+  listener: Listener
```
lchenut, line 140: Same as above, not a fan of un-prefixed object like that. QuicListener is great.

libp2p/transports/quictransport.nim
```diff
@@ -0,0 +1,224 @@
...
+func new*(_: type QuicTransport, u: Upgrade): QuicTransport =
+  QuicTransport(upgrader: QuicUpgrade(ms: u.ms))
+
+method handles*(transport: QuicTransport, address: MultiAddress): bool =
+  if not procCall Transport(transport).handles(address):
+    return false
+  QUIC_V1.match(address)
+
+method start*(transport: QuicTransport, addrs: seq[MultiAddress]) {.async.} =
+  doAssert transport.listener.isNil, "start() already called"
```
lchenut, line 152: Should this really be a `doAssert`? Raising an LPError should be a better idea.

libp2p/transports/quictransport.nim
```diff
@@ -0,0 +1,224 @@
...
+  await procCall Transport(transport).start(addrs)
+  transport.addrs[0] =
+    MultiAddress.init(transport.listener.localAddress(), IPPROTO_UDP).tryGet() &
+    MultiAddress.init("/quic-v1").get()
+  transport.running = true
+
+method stop*(transport: QuicTransport) {.async.} =
+  if transport.running:
+    for c in transport.connections:
+      await c.close()
```
lchenut, line 164: What about something like that: `await allFutures(transport.connections.mapIt(it.close))`

libp2p/transports/quictransport.nim
```diff
@@ -0,0 +1,224 @@
...
+  proc onClose() {.async.} =
+    await conres.join()
+    transport.connections.keepItIf(it != conres)
+    trace "Cleaned up client"
+
+  asyncSpawn onClose()
+  return conres
+
+method accept*(transport: QuicTransport): Future[P2PConnection] {.async.} =
+  doAssert not transport.listener.isNil, "call start() before calling accept()"
```
lchenut, line 191: You should probably check if `transport.running == true`

libp2p/transports/quictransport.nim
```diff
@@ -0,0 +1,224 @@
...
+  return transport.wrapConnection(connection)
+
+method dial*(
+    transport: QuicTransport,
+    hostname: string,
+    address: MultiAddress,
+    peerId: Opt[PeerId] = Opt.none(PeerId),
+): Future[P2PConnection] {.async, gcsafe.} =
+  let connection = await dial(initTAddress(address).tryGet)
+  return transport.wrapConnection(connection)
```
lchenut, line 202: You should probably wait if `transport.running == true`

libp2p/transports/quictransport.nim
diegomrsantos: Is there any semantic difference? Not sure if it adds enough value to justify the change.

libp2p/transports/quictransport.nim
diegomrsantos: This is defined in nim-quic.

lchenut (approved):
Approved in order to work on the comments in separate/smaller PRs.

PR #726, opened 2022-06-07
WIP: gossipsub: add optional per-peer delay metrics

cskiraly:
This is to add more delay related metrics to evaluate GossipSub improvement proposals.

See https://github.com/libp2p/specs/pull/413 for related discussions


Menduist:
Just a note, we won't be able to enable per-peer metric on the fleet (too expensive)
We can still get this in though

PR #728, opened 2022-06-09
Remove Br prefix from BearSSL

Menduist:
ref https://github.com/status-im/nim-bearssl/pull/27


libp2p/crypto/chacha20poly1305.nim
```diff
@@ -17,17 +17,10 @@
 
 {.push raises: [Defect].}
 
-import bearssl
+import bearssl/abi/bearssl_block
```
arnetheduck, line 20:
```suggestion
import bearssl/blockx
```

libp2p/crypto/crypto.nim
```diff
@@ -69,7 +69,7 @@ when supported(PKScheme.Secp256k1):
 # We are still importing `ecnist` because, it is used for SECIO handshake,
 # but it will be impossible to create ECNIST keys or import ECNIST keys.
 
-import ecnist, bearssl
+import ecnist, bearssl/rand, bearssl/hash as bhash, bearssl/abi/intx
```
arnetheduck, line 72: `intx` should be fixed in bearssl instead - ie it should be imported by any `abi/bearss_xxx` that requires access to `int` symbols - what was the error message exactly?

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -226,6 +226,6 @@ method initPubSub*(f: FloodSub)
   procCall PubSub(f).initPubSub()
   f.seen = TimedCache[MessageID].init(2.minutes)
   f.seenSalt = newSeqUninitialized[byte](sizeof(Hash))
-  brHmacDrbgGenerate(f.rng[], f.seenSalt)
+  hmacDrbgGenerate(f.rng[], addr f.seenSalt[0], uint(f.seenSalt.len))
```
arnetheduck, line 229: this is unnecessary, there's an `openArray` version in `bearssl/rand`

arnetheduck:
to really reap the benefits, all `import bearssl` should systematically be removed from the codebase and replaced with specific imports (and reexports, for public symbols)

Menduist:
I think we can merge this once the bearssl PR is merged, and switch to the new `rng` once the generic system is done

libp2p/crypto/ed25519/ed25519.nim
```diff
@@ -13,7 +13,8 @@
 
 {.push raises: Defect.}
 
-import constants, bearssl
+import bearssl/rand
```
arnetheduck, line 16: should export `rand` here too since there's a public symbol that references it - this way, the tests and examples no longer explicitly have to import `bearssl` and it can be removed completely, making the lib easier to use

arnetheduck (dismissed):

arnetheduck:
one more `bearssl` in `libp2p.nim`

arnetheduck (approved):

PR #729, opened 2022-06-09
Change dial return to take into account the protocol

lchenut:
As the title said, currently there's no way to know which protocol is accepted when we dial using multiple protocols.

### DIFF
--- a/libp2p/dial.nim
+++ b/libp2p/dial.nim
@@ -17,6 +17,10 @@ import peerid,
 type
   Dial* = ref object of RootObj
 
+  DialRes* = tuple
+    conn: Connection
+    proto: string
+
 method connect*(
   self: Dial,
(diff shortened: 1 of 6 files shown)


Menduist:
Make sense to me, @dryajov you had some reservations?

libp2p/dial.nim
```diff
@@ -17,6 +17,10 @@ import peerid,
 type
   Dial* = ref object of RootObj
 
+  DialRes* = tuple
+    conn: Connection
+    proto: string
```
Menduist, line 22:
```suggestion
    protocol: string
```

Menduist:
Might as well break before v1?

lchenut:
As it's not urgent and it's not a breaking change anymore. I'll just add it on the circuit relay v2.

PR #730, opened 2022-06-10
Start/stop protocols

lchenut:
Starting/stopping a switch now starts/stops all protocols mounted on that switch.
Currently, there's only gossipsub that effectively starts something but there's might be others (looking at you circuit relayv2) in a near future.

In addition to this change, I added a proc addDirectPeers to make all the gossipsub tests worked.

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -566,22 +566,29 @@ method publish*(g: GossipSub,
 
   return peers.len
 
+proc maintainDirectPeer(g: GossipSub, id: PeerId, addrs: seq[MultiAddress]) {.async.} =
+  let peer = g.peers.getOrDefault(id)
+  if isNil(peer):
+    trace "Attempting to dial a direct peer", peer = id
+    try:
+      # dial, internally connection will be stored
+      let _ = await g.switch.dial(id, addrs, g.codecs)
+      # populate the peer after it's connected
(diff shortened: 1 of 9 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -566,22 +566,29 @@ method publish*(g: GossipSub,
 
   return peers.len
 
+proc maintainDirectPeer(g: GossipSub, id: PeerId, addrs: seq[MultiAddress]) {.async.} =
+  let peer = g.peers.getOrDefault(id)
+  if isNil(peer):
+    trace "Attempting to dial a direct peer", peer = id
+    try:
+      # dial, internally connection will be stored
+      let _ = await g.switch.dial(id, addrs, g.codecs)
```
dryajov, line 575: this can probably use `connect` instead of `dial`?

lchenut:
> Nonetheless, we should start instantly new protocols if they are mounted while the switch is already running @lchenut

To summarize our discussion on this subject, starting a protocol during mounting if the switch is already running causes a lot of unwanted side effect. So I changed the perspective and forbid the unstarted protocols from being mounted if the switch is running.

Menduist (approved):

Menduist:
For future reference, goal is to handle the "mounting a protocol with a switch already running" mess once someone need it, and shelve it in the meantime
Currently, the to mount a protocol on a running switch, you need to start it manually, which is not ideal

PR #735, opened 2022-06-30
lower log level of no-peers message

arnetheduck:
the number of peers sent to is returned from the function - the caller
can log this, if need be


arnetheduck:
also, looks like someone is not trimming whitespace ;)

Menduist (approved):

PR #736, opened 2022-06-30
Use choosenim for regular CI

Menduist:
Extracted from #724


dryajov:
Are we sure we want to do this? I know that the previous setup is not the prettiest, but it gives us a lot of control, I've no idea what `jiro4989/setup-nim-action` is actually doing and how to debug it. I'm not opposed, but I'd be inclined to say we aren't actually improving anything with this change?

Menduist:
Can be revisited later, but now at least the compiler is cached so we don't compile every time

PR #739, opened 2022-07-08
Autonat protocol

lchenut:
Implementation of the [AutoNAT Protocol](https://github.com/libp2p/specs/tree/master/autonat)


mxinden:
Really cool to see this happening! Let me know in case there are any questions in regards to the specification.

Also //CC @elenaf9 who implemented AutoNAT in rust-libp2p https://github.com/libp2p/rust-libp2p/pull/2262/

libp2p/protocols/autonat.nim
```diff
@@ -0,0 +1,276 @@
...
+    return conn.sendResponseError(InternalError, "Expected an IP address")
+
+  var addrs = initHashSet[MultiAddress]()
+  addrs.incl(conn.observedAddr)
+  for ma in peerInfo.addrs:
+    let maFirst = ma[0]
+    if maFirst.isErr() or (not IP4.match(maFirst.get()) and
+                           not IP6.match(maFirst.get())):
+      continue
+    addrs.incl(ma)
```
elenaf9, line 246: If I  understand it correctly, you are currently dialing any address that is sent in `PeerInfo.addrs` (up to the `ArbitraryLimit`). Note that for DoS protection the server must only dial addresses that are based on `observedAddr`. Quoting from the AutoNAT specs: 
> In order to prevent attacks like the one described in [RFC 3489, Section 12.1.1](https://www.rfc-editor.org/rfc/rfc3489#section-12.1.1) (..), implementations MUST NOT dial any multiaddress unless it is based on the IP address the requesting node is observed as. This restriction as well implies that implementations MUST NOT accept dial requests via relayed connections as one can not validate the IP address of the requesting node.

In go-libp2p and rust-libp2p we solve this by replacing each addresses' IP (in your case `maFirst`) with the IP of `observedAddr`:

- [1] Logic in Go: https://github.com/libp2p/go-libp2p/blob/3d351e4ed396917866f464f2ab90c838fe7d92a2/p2p/host/autonat/svc.go#L154-L168
- [2] Logic in Rust:  https://github.com/libp2p/rust-libp2p/blob/d0da3a0973431babba94fffd3ffd7723e0b09837/protocols/autonat/src/behaviour/as_server.rs#L313-L352

elenaf9 (commented):
Very cool!

libp2p/protocols/autonat.nim
Menduist: @elenaf9 @mxinden should we update the spec to reflect the fact that wrong addresses will be "fixed" using the observed address?

The current spec really reads as "any address not matching the observed address will be __skipped__"

mxinden: I am in favor of such change @Menduist. Would you mind opening a pull request?

libp2p/dialer.nim
```diff
@@ -185,6 +186,45 @@ proc negotiateStream(
...
+          trace "Dialing address", address = $a, peerId, hostname
+          try:
+            let dialed = await transport.dial(hostname, address)
+            defer: await dialed.close()
+            # make sure to assign the peer to the connection
+            dialed.peerId = peerId
+            # also keep track of the connection's bottom unsafe transport direction
+            # required by gossipsub scoring
+            dialed.transportDir = Direction.Out
+            let sconn = await transport.upgrader.secure(dialed)
```
Menduist, line 219: Why did you use `await transport.upgrader.secure(dialed)` instead of `await transport.upgradeOutgoing(dialed)`?

Menduist:
@lchenut I just pushed a commit which extracts the conn manager tracking from dialAndUpgrade, you should now be able to use dialAndUpgrade in canDial

libp2p/dialer.nim
```diff
@@ -185,6 +178,45 @@ proc negotiateStream(
 
   return conn
 
+method canDial*(
+  self: Dialer,
+  peerId: PeerId,
+  addrs: seq[MultiAddress],
+  protos: seq[string]): Future[MultiAddress] {.raises: [Defect], async.} =
```
Menduist, line 185: protos is not used?

libp2p/protocols/autonat.nim
```diff
@@ -0,0 +1,288 @@
...
+    raise newException(AutonatError, "Bad status " &
+                                      $response.status & " " &
+                                      response.text.get(""))
+  if response.ma.isNone():
+    raise newException(AutonatError, "Missing address")
+  return response.ma.get()
+
+proc doDial(a: Autonat, conn: Connection, addrs: seq[MultiAddress]) {.async.} =
+  try:
+    let ma = await Dialer(a.switch.dialer).canDial(conn.peerId, addrs, @[AutonatCodec])
```
Menduist, line 221: Let's add `canDial` (or maybe `tryDial` is a more descriptive name) to the Dial interface instead of doing this

libp2p/protocols/autonat.nim
```diff
@@ -0,0 +1,288 @@
...
+  if hostIp.isErr() or (not IP4.match(hostIp.get()) and
+                        not IP6.match(hostIp.get())):
+    trace "wrong observed address", address=conn.observedAddr
+    return conn.sendResponseError(InternalError, "Expected an IP address")
+  var addrs = initHashSet[MultiAddress]()
+  addrs.incl(conn.observedAddr)
+  for ma in peerInfo.addrs:
+    let maFirst = ma[0]
+    if maFirst.isErr() or (not IP4.match(maFirst.get()) and
+                           not IP6.match(maFirst.get())):
```
Menduist, line 245:
```suggestion
    if maFirst.isErr() or not IP.match(maFirst.get()):
```

Menduist: Same above

libp2p/protocols/autonat.nim
```diff
@@ -0,0 +1,288 @@
...
+       ../multiaddress,
+       ../peerid,
+       ../errors
+
+logScope:
+  topics = "libp2p identify"
+
+const
+  AutonatCodec* = "/libp2p/autonat/1.0.0"
+  ArbitraryLimit = 128 # TODO: make it a configuration variable or smthg
```
Menduist, line 25: :smile: let's give it a better name though
8 seems like a saner default too

Menduist:
Ok, last detail before my approval, we should use a semaphore to limit the number of concurrent dial backs to avoid attacks

1 max seems like a sane default

libp2p/multiaddress.nim
```diff
@@ -548,6 +548,28 @@ proc protoAddress*(ma: MultiAddress): MaResult[seq[byte]] =
   buffer.setLen(res)
   ok(buffer)
 
+proc len*(ma: MultiAddress): MaResult[int] =
```
Menduist, line 551: Couldn't we do

```nim
proc len*(ma: MultiAddress): MaResult[int] =
  var count = 0
  for part in ma:
    if part.isErr: err(part.err) # or similar
    count.inc
  ok(count)
```

A tad more wasteful, but for less error prone, I think

libp2p/multiaddress.nim
```diff
@@ -587,10 +609,61 @@ proc getPart(ma: MultiAddress, index: int): MaResult[MultiAddress] =
     inc(offset)
   ok(res)
 
+proc getParts[U, V](ma: MultiAddress, slice: HSlice[U, V]): MaResult[MultiAddress] =
```
Menduist, line 612: Ditto, if we can reuse existing function, that would be nice

Menduist (approved):

libp2p/protocols/autonat.nim
```diff
@@ -0,0 +1,291 @@
...
+  if msg.dial.isNone() or msg.dial.get().peerInfo.isNone():
+    return conn.sendResponseError(BadRequest, "Missing Peer Info")
+  let peerInfo = msg.dial.get().peerInfo.get()
+  if peerInfo.id.isSome() and peerInfo.id.get() != conn.peerId:
+    return conn.sendResponseError(BadRequest, "PeerId mismatch")
+
+  let hostIp = conn.observedAddr[0]
+  if hostIp.isErr() or not IP.match(hostIp.get()):
+    trace "wrong observed address", address=conn.observedAddr
+    return conn.sendResponseError(InternalError, "Expected an IP address")
```
elenaf9, line 243: I don't know if nim-libp2p already supports the relay-protocol, but if the observed address is a relayed address (either because the local peer is listening via a relay or the remote is) the dial should be rejected as well. 
a) not sure how it is implemented in nim-libp2p, but at least in rust-libp2p the observed address has the form /\<relay-ip>/\<relay-port>/p2p-circuit/\<remote-ip>, so the `hostIp` that you use here is actually the relays IP
b) even if the relay would inform us of the remote's IP, we can not trust that this information is correct

lchenut: We talked about it with @Menduist, trying to figure out if we have to refuse connections with relayed observed addresses or skip relayed addresses in the PeerInfo.addrs or both and ended up doing nothing without checking what the rust-libp2p or go-libp2p did (and forgetting about it aswell).

Well, that should be fixed now, thanks again.

Menduist (approved):

Menduist (approved):

Menduist (approved):

PR #741, opened 2022-07-18
Replace table by seq for storing muxers

lchenut:
A simple change in the way to store `MuxerProvider` to remove the randomness of the Table (it could have been an OrderedTable, but I think it's a bit silly to use that for a maximum of two elements).
Now the first muxer added in the switch builder is always the one used/tried first.

### DIFF
--- a/libp2p/upgrademngrs/muxedupgrade.nim
+++ b/libp2p/upgrademngrs/muxedupgrade.nim
@@ -27,7 +27,7 @@ type
 
 proc getMuxerByCodec(self: MuxedUpgrade, muxerName: string): MuxerProvider =
   for m in self.muxers:
-    if m.codecs[0] == muxerName:
+    if muxerName in m.codecs:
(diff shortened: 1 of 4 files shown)


libp2p/upgrademngrs/muxedupgrade.nim
```diff
@@ -22,9 +22,14 @@ logScope:
 
 type
   MuxedUpgrade* = ref object of Upgrade
-    muxers*: Table[string, MuxerProvider]
+    muxers*: seq[MuxerProvider]
     streamHandler*: StreamHandler
 
+proc getMuxerByCodec(self: MuxedUpgrade, muxerName: string): MuxerProvider =
+  for m in self.muxers:
+    if m.codecs[0] == muxerName:
```
Menduist, line 30:
```suggestion
    if muxerName in m.codecs:
```

PR #742, opened 2022-07-19
make it more Nim 1.4+ compatible

narimiran:
A continuation of https://github.com/status-im/nimbus-eth2/pull/3888

Rationale (from that link):

> Currently (the latest commit in unstable branch), the CI log for the 1.6 branch has **107.000 lines**.
Removing every hint about cannot raise 'Defect' (including the ones in the vendor directory) leaves **16.000 lines**. 85% off.

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -16,7 +16,10 @@ runnableExamples:
    # etc
    .build()
 
-{.push raises: [Defect].}
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
 
 import
   options, tables, chronos, chronicles, sequtils,
--- a/config.nims
+++ b/config.nims
@@ -2,8 +2,6 @@
 if dirExists("nimbledeps/pkgs"):
   switch("NimblePath", "nimbledeps/pkgs")
 
-when (NimMajor, NimMinor) > (1, 2):
-  switch("hint", "XCannotRaiseY:off")
 # begin Nimble config (version 1)
 when fileExists("nimble.paths"):
   include "nimble.paths"
--- a/libp2p/cid.nim
+++ b/libp2p/cid.nim
@@ -9,7 +9,10 @@
 
 ## This module implementes CID (Content IDentifier).
 
-{.push raises: [Defect].}
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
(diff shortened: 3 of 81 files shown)


Menduist (dismissed):

narimiran:
> /Users/runner/work/nim-libp2p/nim-libp2p/libp2p/stream/bufferstream.nim(114, 49) Error: Defect can raise an unlisted exception: Defect

For all `nim devel` CIs. But that line hasn't been changed:

```nim
method atEof*(s: BufferStream): bool {.raises: [Defect].} =
```

Menduist:
That's because the base method probably raises: [], and the sub method had Defect explicitly

This will probably require some cleanup work

narimiran:
> That's because the base method probably raises: [], and the sub method had Defect explicitly

Correct.

> This will probably require some cleanup work

See my [second commit](https://github.com/status-im/nim-libp2p/pull/742/commits/d4d767a391af64cb205aa89868df3dbb56b34923):
- removed explicit `{.raises: [Defect].}` pragma for some child-methods
- hack-ish way to fix `method init` in `tests/testnoise.nim` — I've tried several other solutions, but nothing until this one had worked for both Nim 1.2 and Nim devel.

libp2p/builders.nim
```diff
@@ -16,7 +16,10 @@ runnableExamples:
    # etc
    .build()
 
-{.push raises: [Defect].}
+when (NimMajor, NimMinor) < (1, 4):
```
dryajov, line 19: There is https://github.com/markspanbroek/upraises precisely to work around this, I think it's thin enough that it can be used without much issues?

narimiran: `upraises` was also mentioned and rejected by @arnetheduck in the [nimbus PR](https://github.com/status-im/nimbus-eth2/pull/3888), because:

> ... they introduce a lot of complexity for what ultimately is a cosmetic problem that will go away soon

Menduist (approved):

Menduist:
If this branch works well, you can probably remove the
```nim
when (NimMajor, NimMinor) > (1, 2):
  switch("hint", "XCannotRaiseY:off")
```
from config.nims (I guess now it should be a rare warning)

narimiran:
> you can probably remove (...)

Done.

And fixed some new files with the same change as for all the other files.

Menduist (approved):

narimiran:
Merged, before it bit-rots again.

PR #743, opened 2022-07-21
--styleCheck:error

tersec:


Menduist (dismissed):

libp2p/peerinfo.nim
```diff
@@ -1,11 +1,12 @@
...
-# This file may not be copied, modified, or distributed except according to
-# those terms.
+## Nim-LibP2P
+## Copyright (c) 2019-2022 Status Research & Development GmbH
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
```
Menduist, line 8:
```suggestion
# Nim-LibP2P
# Copyright (c) 2019-2022 Status Research & Development GmbH
# Licensed under either of
#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
# at your option.
# This file may not be copied, modified, or distributed except according to
# those terms.
```

Menduist: We use nimdoc in libp2p, so we have a single `#` to hide it in the docs

tersec: I ended up just removing this change, since there was already a copyright year update.

libp2p/protocols/pubsub/rpc/protobuf.nim
```diff
@@ -132,8 +132,8 @@ proc decodeGraft*(pb: ProtoBuffer): ProtoResult[ControlGraft] {.
 
   trace "decodeGraft: decoding message"
   var control = ControlGraft()
-  if ? pb.getField(1, control.topicId):
-    trace "decodeGraft: read topicId", topic_id = control.topicId
+  if ? pb.getField(1, control.topicID):
```
Menduist, line 135: Erm, we switched to PeerId as part of the first styleCheck cleanup, so we should try to stay consistent with the `Id`, if possible (also applies to `MessageID`)

Not sure why theses were not catch on the first pass, sorry about that

tersec: https://github.com/status-im/nim-libp2p/pull/743/commits/90de6d56657ee6d3fd86faa2f6ed14aae06ec0ab

Got those two in particular. There are some similar but not-exactly-identical identifiers (plural versions, for example) which I didn't change, because better not to do more than people want then revert.

Menduist: I mean, it's now or never to fix these, so feel free to fix any inconsistencies you find

Menduist (approved):

libp2p/protocols/pubsub/rpc/protobuf.nim
tersec: Okay, with this kind of diffuse change, probably worth merging this PR and doing the rest in another PR.

PR #747, opened 2022-07-26
ci: test Nim 1.6 branch

narimiran:


Menduist (dismissed):
We originally removed it to have a faster CI, but now that nimbus is on 1.6 and waku / codex on 1.2, we can have both during the transition period..

narimiran:
> We originally removed it to have a faster CI

IMO (and I probably don't know all the details, so please correct me), testing on 1.6 is more important than testing on `devel`.
So, if we want faster CIs by having just two branches, we can drop `devel` and keep 1.6.

Menduist:
SGTM to drop devel (we still test it in the daily job), go ahead

Menduist (dismissed):

Menduist:
Well, this PR would have saved us a few days earlier :sweat_smile: https://github.com/status-im/nim-libp2p/pull/743 seem to have broken 1.6 (but not devel)
I'll fix it here

PR #749, opened 2022-07-29
ConnManager connection tracking refacto

Menduist:
Simpler API which avoids the "ConnProvider" callback
Saner IMO


dryajov:
Great refacto! I think it should work. One of the concerns of the original version was to avoid dangling connections due to some concurrency issue, hence the overly complicated approach. If this works, the its definitely much better.

lchenut:
It's a clever way to do it. I like it!

lchenut (approved):

PR #751, opened 2022-08-02
RendezVous Protocol

lchenut:
First draft of a discovery interface (discussed [here](https://github.com/status-im/nim-libp2p/issues/140#issuecomment-1173930215)). As the interface is not useful without discovery mechanisms, I also worked on an implementation of the [RendezVous protocol](https://github.com/libp2p/specs/blob/master/rendezvous/README.md).

As I said, it's a first draft, so here a non-exhaustive list of what remains to be done:
- [x] Finish the rendezvous client request
- [x] Bind the request / advertise function to a RendezVousInterface
- [x] Create a DiscoveryManager
- [x] Tests

More discovery mechanisms should be added lately to test the discovery manager more thoroughly.

As the discovery manager is a important thing, I move it in another PR, this one is now just the RendezVous Protocol


libp2p/protocols/discovery/rendezvous.nim
```diff
@@ -0,0 +1,348 @@
...
+    registerEvent: AsyncEvent # TODO: to raise during the heartbeat
+    # + make the heartbeat sleep duration "smarter"
+    sema: AsyncSemaphore
+    peers: seq[PeerId]
+    switch: Switch
+
+proc checkPeerRecord(spr: Option[seq[byte]], peerId: PeerId): Result[void, string] =
+  if spr.isNone():
+    return err("Empty peer record")
+  let signedEnv = SignedPeerRecord.decode(spr.get())
```
Menduist, line 116: You can probably use a combination of `?` and `mapErr(x => $x)`

libp2p/protocols/discovery/rendezvous.nim
```diff
@@ -0,0 +1,348 @@
...
+    cookie = if d.cookie.isSome():
+      try: Protobuf.decode(d.cookie.get(), type(Cookie))
+      except ProtobufReadError: return conn.sendDiscoverResponseError(InvalidCookie)
+    else: Cookie(offset: rdv.registered.low().uint64, ns: ns)
+    indexes = if ns != "":
+        try: rdv.indexes[nsSalted].s
+        except KeyError: return conn.sendDiscoverResponseError(InvalidNamespace)
+      else: toSeq(rdv.registered.low()..rdv.registered.high())
+  if indexes.len() == 0:
+    return conn.sendDiscoverResponse(@[], Cookie())
```
Menduist, line 226: Shouldn't we send a None Cookie in this case? Not sure it's worth the trouble though

libp2p/protocols/discovery/rendezvous.nim
```diff
@@ -0,0 +1,348 @@
...
+          offset = index
+          break
+        limit.dec()
+        rdv.registered[index].data
+  conn.sendDiscoverResponse(s, Cookie(offset: offset.uint64, ns: ns))
+
+proc advertisePeer(rdv: RendezVous, peer: PeerId, msg: seq[byte]) {.async.} =
+  let conn = await rdv.switch.dial(peer, RendezVousCodec)
+  defer: await conn.close()
+  await conn.writeLp(msg).wait(chronos.milliseconds(2500))
```
Menduist, line 241: This will raise an exception on timeout, but we are in a asyncSpawned context, so any uncaught exception will be raise as a defect

Another option might be to start a `let timeout = sleepAsync(5.seconds)` at the beginning of the procedure, and do
```nim
let writing = conn.writeLp(msg)
await writing or timeout
if timeout.finished:
  writing.cancel()
  return
```

libp2p/protocols/discovery/rendezvous.nim
```diff
@@ -0,0 +1,348 @@
...
+  let
+    buf = await conn.readLp(4096).wait(chronos.milliseconds(2500))
+    msgRecv = Protobuf.decode(buf, type(Message))
+  if msgRecv.msgType.isNone() or msgRecv.msgType.get() != MessageType.RegisterResponse:
```
Menduist, line 245:
```suggestion
  if msgRecv.msgType != some MessageType.RegisterResponse:
```

libp2p/protocols/discovery/rendezvous.nim
```diff
@@ -0,0 +1,348 @@
...
+    trace "Unexpected register response", peer, msgType = msgRecv.msgType
+  elif msgRecv.registerResponse.isNone() or
+       msgRecv.registerResponse.get().status.isNone() or
+       msgRecv.registerResponse.get().status.get() != ResponseStatus.Ok:
+    trace "Refuse to register", peer, response = msgRecv.registerResponse
+  rdv.sema.release()
```
Menduist, line 251: Should be in a defer / finally

libp2p/protocols/discovery/rendezvous.nim
```diff
@@ -0,0 +1,348 @@
...
+  let
+    r = Register(ns: some(ns), signedPeerRecord: some(sprBuff.get()), ttl: some(ttl))
+    msg = Protobuf.encode(Message(msgType: some(MessageType.Register), register: some(r)))
+  rdv.save(ns, rdv.switch.peerInfo.peerId, r, ttl)
+  for peer in rdv.peers:
+    try:
+      await rdv.sema.acquire()
+      asyncSpawn rdv.advertisePeer(peer, msg)
+    except CancelledError as exc:
+      rdv.sema.release()
```
Menduist, line 268: That's not correct, the only point where we can get cancelled / have exception here is in the acquire (asyncSpawn doesn't block)

libp2p/protocols/discovery/rendezvous.nim
lchenut: It's the same thing

Menduist:
btw lets move this to protocols/discovery

libp2p.nimble
```diff
@@ -16,7 +16,8 @@ requires "nim >= 1.2.0",
          "metrics",
          "secp256k1",
          "stew#head",
-         "websock"
+         "websock",
+         "protobuf_serialization"
```
Menduist, line 20: Do we still need this?

libp2p/protocols/rendezvous.nim
```diff
@@ -0,0 +1,648 @@
...
+export protobuf_serialization
+export chronicles
+
+logScope:
+  topics = "libp2p discovery rendezvous"
+
+const
+  RendezVousCodec* = "/rendezvous/1.0.0"
+  MinimumTTL = 2'u64 * 60 * 60
+  MaximumTTL = 72'u64 * 60 * 60
```
Menduist, line 38: Why not use durations?

libp2p/protocols/rendezvous.nim
```diff
@@ -0,0 +1,648 @@
...
+          rdv.indexes[nsSalted]
+        except KeyError:
+          await conn.sendDiscoverResponseError(InvalidNamespace)
+          return
+      else: toSeq(rdv.registered.low()..rdv.registered.high())
+  if indexes.len() == 0:
+    await conn.sendDiscoverResponse(@[], Cookie())
+    return
+  var offset = indexes[^1]
+  let n = Moment.now()
```
Menduist, line 428: Why not something like `indexes.keepItIf(it.uint64 > cookie.offset)` instead of filtering in the loop?

libp2p/protocols/rendezvous.nim
```diff
@@ -0,0 +1,648 @@
...
+  let s = collect(newSeq()):
+      for index in indexes:
+        var reg = rdv.registered[index]
+        if reg.expiration < n or index.uint64 <= cookie.offset: continue
+        if limit == 0:
```
Menduist, line 433: This could be the first check of this loop to avoid a few useless iterations

libp2p/protocols/rendezvous.nim
```diff
@@ -0,0 +1,648 @@
...
+  let
+    nsSalted = d.ns & rdv.salt
+    cookie =
+      if d.cookie.isSome():
+        try:
+          Cookie.decode(d.cookie.get()).get()
+        except CatchableError:
+          await conn.sendDiscoverResponseError(InvalidCookie)
+          return
+      else: Cookie()
```
Menduist, line 415: I think we said at some point that we should start with a random cookie to avoid that everyone gets the same peers?

libp2p/protocols/rendezvous.nim
```diff
@@ -0,0 +1,648 @@
...
+    sema = newAsyncSemaphore(1)
+  if sprBuff.isErr() or ns.len notin 1..255 or ttl notin MinimumTTL..MaximumTTL:
+    return
+  let
+    r = Register(ns: ns, signedPeerRecord: sprBuff.get(), ttl: some(ttl))
+    msg = encode(Message(msgType: MessageType.Register, register: some(r)))
+  rdv.save(ns, rdv.switch.peerInfo.peerId, r)
+  for peer in rdv.peers:
+    await rdv.sema.acquire()
+    sema.forceAcquire()
```
Menduist, line 479: Is this a hacky way to wait for all of them to be done? Lol
Instead you could every advertisePeer in a seq[Future[void]] and use await allFutures or similar

libp2p/protocols/rendezvous.nim
```diff
@@ -0,0 +1,648 @@
...
+        none(seq[byte])
+    await conn.writeLp(encode(Message(
+      msgType: MessageType.Discover,
+      discover: some(d))).buffer)
+    let
+      buf = await conn.readLp(65536)
+      msgRcv = Message.decode(buf).get()
+    if msgRcv.msgType != MessageType.DiscoverResponse or
+       msgRcv.discoverResponse.isNone():
+      trace "Unexpected discover response", msgType = msgRcv.msgType
```
Menduist, line 520: debug

libp2p/protocols/rendezvous.nim
```diff
@@ -0,0 +1,648 @@
...
+
+proc new*(T: typedesc[RendezVous],
+         switch: Switch,
+         rng: ref HmacDrbgContext = newRng()): T =
+  let rdv = T.new(rng)
+  rdv.setup(switch)
+  return rdv
+
+proc deletesRegister(rdv: RendezVous) {.async.} =
+  # TODO replace 1 minutes by the shortest time to sleep before a registration finished
```
Menduist, line 630: Don't think we need to

libp2p/protocols/rendezvous.nim
```diff
@@ -0,0 +1,648 @@
...
+    peerId: PeerId
+    data: Register
+
+  RegisteredSeq = object
+    s: seq[RegisteredData]
+    offset: uint64
+
+  RendezVous* = ref object of LPProtocol
+    indexes: Table[string, seq[int]]
+    registered: OffsettedSeq[RegisteredData]
```
Menduist, line 292: Maybe add a comment to explain those

tests/testrendezvous.nim
```diff
@@ -0,0 +1,80 @@
...
+  SwitchBuilder.new()
+    .withRng(newRng())
+    .withAddresses(@[ MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet() ])
+    .withTcpTransport()
+    .withMplex()
+    .withNoise()
+    .withRendezVous(rdv)
+    .build()
+
+suite "RendezVous":
```
Menduist, line 20: Looks a bit light

libp2p/protocols/rendezvous.nim
Menduist: Or maybe we can just shuffle the response, but meh

libp2p/protocols/rendezvous.nim
lchenut: No, it's a hacky way to send only five requests at the same time. But I agree that the two semaphores acquire could be better in advertisePeer than here.

Menduist: You only need a single semaphore: `rdv.sema.acquire` at the beginning of `advertisePeer`, release it at the end.
Start every `advertisePeer` at once, and wait for all of them here

libp2p/protocols/rendezvous.nim
```diff
@@ -0,0 +1,648 @@
...
+proc discover(rdv: RendezVous, conn: Connection, d: Discover) {.async.} =
+  trace "Received Discover", peerId = conn.peerId, ns = d.ns
+  if d.ns.len notin 0..255:
+    await conn.sendDiscoverResponseError(InvalidNamespace)
+    return
+  var limit = min(DiscoverLimit, d.limit.get(DiscoverLimit))
+  let
+    nsSalted = d.ns & rdv.salt
+    cookie =
+      if d.cookie.isSome():
```
Menduist, line 409: The ns from the cookie isn't checked?

Menduist (approved):

PR #753, opened 2022-08-04
Better dnsaddr resolving

Menduist:
Also tries multiples address on dial failure

ref https://github.com/status-im/nim-libp2p/pull/615#issuecomment-958851810

### DIFF
--- a/libp2p/multiaddress.nim
+++ b/libp2p/multiaddress.nim
@@ -530,7 +530,7 @@ proc protoArgument*(ma: MultiAddress,
           err("multiaddress: Decoding protocol error")
         else:
           ok(res)
-      elif proto.kind in {Length, Path}:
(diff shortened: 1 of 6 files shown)


libp2p/multiaddress.nim
```diff
@@ -541,6 +541,17 @@ proc protoArgument*(ma: MultiAddress,
       else:
         ok(res)
 
+proc protoArgument*(ma: MultiAddress): MaResult[seq[byte]] =
```
lchenut, line 544: protoArgument here and protoAddress (really bad name) below seems to do the same thing.
Either depecrate one or merge the two procedures into one.

Menduist: Done

lchenut (approved):

PR #754, opened 2022-08-24
Keep connection alive when peer doesn't support pubsub

Menduist:
Since this was introduced, nimbus got better peer management, and will eventually kick a useless peer (ie not supporting GS) if required

The current behavior is obviously wrong, because nimbus & waku won't peer with light clients not supporting GS for instance


Menduist:
Tried to add a log to see if this still happens in practice, unfortunately it's hard to distinguish regular situations from irregular ones

lchenut (approved):

PR #756, opened 2022-08-31
allow connection to a peer with unknown PeerId

arnetheduck:
Useful when connecting to a noise peer based on IP address alone

### DIFF
--- a/libp2p/dial.nim
+++ b/libp2p/dial.nim
@@ -34,7 +34,7 @@ method connect*(
 method connect*(
   self: Dial,
   addrs: seq[MultiAddress]): Future[PeerId] {.async, base.} =
-  ## Connects to a peer and retrieve it's PeerId
+  ## Connects to a peer and retrieve its PeerId
 
   doAssert(false, "Not implemented!")
 
--- a/libp2p/dial.nim
+++ b/libp2p/dial.nim
@@ -31,6 +31,13 @@ method connect*(
 
   doAssert(false, "Not implemented!")
 
+method connect*(
+  self: Dial,
+  addrs: seq[MultiAddress]): Future[PeerId] {.async, base.} =
+  ## Connects to a peer and retrieve it's PeerId
+
+  doAssert(false, "Not implemented!")
+
 method dial*(
   self: Dial,
   peerId: PeerId,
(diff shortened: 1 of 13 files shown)


Menduist:
The new API for this:
```nim
check: (await switch2.connect(switch1.peerInfo.addrs)) == switch1.peerInfo.peerId
```

Since it returns the `PeerId`, it should be pretty obvious what it does, and can't be used unintentionally insteado f the original `connect` (which returns void)

libp2p/dial.nim
```diff
@@ -31,6 +31,13 @@ method connect*(
 
   doAssert(false, "Not implemented!")
 
+method connect*(
+  self: Dial,
+  addrs: seq[MultiAddress]): Future[PeerId] {.async, base.} =
+  ## Connects to a peer and retrieve it's PeerId
```
arnetheduck, line 37:
```suggestion
  ## Connects to a peer and retrieve its PeerId
```

libp2p/dialer.nim
```diff
@@ -101,20 +98,22 @@ proc dialAndUpgrade(
...
   addrs: seq[MultiAddress],
   forceDial: bool):
   Future[Connection] {.async.} =
-  if self.localPeerId == peerId:
+  if Opt.some(self.localPeerId) == peerId:
     raise newException(CatchableError, "can't dial self!")
 
   # Ensure there's only one in-flight attempt per peer
-  let lock = self.dialLock.mgetOrPut(peerId, newAsyncLock())
+  let lock = self.dialLock.mgetOrPut(peerId.get(default(PeerId)), newAsyncLock())
```
arnetheduck, line 109: would it make sense to block on addresses here instead?

libp2p/dialer.nim
```diff
@@ -101,20 +98,22 @@ proc dialAndUpgrade(
...
   try:
     await lock.acquire()
 
     # Check if we have a connection already and try to reuse it
-    var conn = self.connManager.selectConn(peerId)
+    var conn =
```
arnetheduck, line 114: where do we check that after a discovery of peerid, we don't have too many connections to the same peer?

Menduist: https://github.com/status-im/nim-libp2p/blob/e6440c43c211d930d72d6189d4bca67fef7276b3/libp2p/connmanager.nim#L387-L390

arnetheduck (commented):
lgtm, can't approve my own PR :)

libp2p/dialer.nim
Menduist: Don't have a strong opinion, though it seems more tricky to do

PR #764, opened 2022-09-07
Add onion3 multiaddr support

diegomrsantos:
Add onion3 multiaddr support

### DIFF
--- a/libp2p/multiaddress.nim
+++ b/libp2p/multiaddress.nim
@@ -222,6 +222,40 @@ proc onionVB(vb: var VBuffer): bool =
   if vb.readArray(buf) == 12:
     result = true
 
+proc onion3StB(s: string, vb: var VBuffer): bool =
+  try:
+    var parts = s.split(':')
+    if len(parts) != 2:
+      return false
+    if len(parts[0]) != 56:
+      return false
+    var address = Base32Lower.decode(parts[0].toLowerAscii())
+    var nport = parseInt(parts[1])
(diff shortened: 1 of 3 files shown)


libp2p/multiaddress.nim
```diff
@@ -363,6 +402,10 @@ const
       mcodec: multiCodec("onion"), kind: Fixed, size: 10,
```
Menduist, line 402: Can you fix this to 12 while we're at it?

diegomrsantos: Thought about that, but is it fine to do that in this PR?

Menduist: Yes go ahead, better than doing a PR for 1 LoC imo

Menduist:
Looks good!

libp2p/multiaddress.nim
diegomrsantos: Maybe tests for the fix should also be added?

Menduist: Yes good point, maybe we should do another PR with:
- The fix
- Roundtrip tests (need to replace `ipfs` by `p2p` in the test inputs)
- Roundtrip tests "per part": `ma[0..^1].tryGet().mapIt($it).join('')` or something like that
The last one would have caught the bug in the first place

Menduist (dismissed):

Menduist (approved):

Menduist (approved):

PR #768, opened 2022-09-14
Add codex & waku to autobump

Menduist:
Similar to nimbus: https://github.com/status-im/nimbus-eth2/pull/3916
Will run CI on waku & codex for each unstable push on libp2p

Also opens the possibility to run a few nodes of each project with libp2p's unstable at all times, which is very useful


jm-clius (approved):
Ah, cool. Take it we'll know of breaking changes earlier 😄

PR #769, opened 2022-09-14
Move relay & autonat to connectivity folder

Menduist:
Trying to keep things tidy, since in the end we will probably have >5 connectivity protocols


lchenut (dismissed):
This is a good idea. But it might be interesting to merge all the relay files into one so that you only have one relay.nim to import instead of six different files. I'll do it one day.

PR #772, opened 2022-09-21
Make observedAddr optional

diegomrsantos:


libp2p/protocols/connectivity/autonat.nim
```diff
@@ -241,15 +244,21 @@ proc handleDial(a: Autonat, conn: Connection, msg: AutonatMsg): Future[void] =
   if peerInfo.id.isSome() and peerInfo.id.get() != conn.peerId:
     return conn.sendResponseError(BadRequest, "PeerId mismatch")
 
-  var isRelayed = conn.observedAddr.contains(multiCodec("p2p-circuit"))
+  let observedAddr =
+    if conn.observedAddr.isNone:
+      return conn.sendResponseError(BadRequest, "Missing observed address")
+    else:
+      conn.observedAddr.get()
```
Menduist, line 251:
```suggestion
  if conn.observedAddr.isNone:
    return conn.sendResponseError(BadRequest, "Missing observed address")
  let observedAddr = conn.observedAddr.get()
```
Maybe

Menduist (approved):

libp2p/dial.nim
```diff
@@ -12,6 +12,7 @@ when (NimMajor, NimMinor) < (1, 4):
 else:
   {.push raises: [].}
 
+import std/options
```
arnetheduck, line 15: prefer `results.Opt`

diegomrsantos: I had this error 

`/nim-libp2p/nimbledeps/pkgs/stew-#598246620da5c41d0e92a8dd6aab0755381b21cd/stew/results.nim(903, 9) Error: the field 'o' is not accessible.`

with the code below

```
proc new*(T: type WsStream,
           session: WSSession,
           dir: Direction,
           timeout = 10.minutes,
           observedAddr: Opt[MultiAddress] = Opt.none(MultiAddress)): T =
```

Menduist: I don't think we can work around this one easily, it's probably because `new` is generic

Menduist: Anyway this is private API, we're free to switch to something `Opt` whenever we want

Menduist (dismissed):

libp2p/dial.nim
Menduist: Unless we can easily remove the optional parameter? @diegomrsantos (as in, make it obligatory)

arnetheduck: can you push a branch with the code that gives compile error?

diegomrsantos: > can you push a branch with the code that gives compile error?

https://github.com/status-im/nim-libp2p/compare/opt-compilation-error

arnetheduck: thanks - so this is no the same as 21 - it's a separate compiler bug for when evaluation happens for a default - oddly, it can be worked around by making `o` public in `Result` - there are many compiler bugs surrounding default parameters unfortunately - `observedAddr: Opt[MultiAddress] = default(Opt[MultiAddress])` "should" also work with the same effect, but fails with a different compiler error - it would perhaps make sense to try to pull this out into a smaller test case and upstream a bug report

Menduist: @arnetheduck I never quite understood if template are supposed to work when accessing private fields
Anyway, if they are, it's currently broken in many ways (21, this one, etc)
Maybe we should do an "epic issue" with every case we know of where is doesn't work?

arnetheduck: > I never quite understood if template are supposed to work when accessing private fields

the public/private part is supposed to be evaluated in the module where the template is declared - it's used to "open" access to otherwise private fields (for example like `Result` does, to gate each `get` with an error/value check

lchenut (dismissed):

libp2p/dial.nim
```diff
@@ -69,5 +69,5 @@ method addTransport*(
 method tryDial*(
   self: Dial,
   peerId: PeerId,
-  addrs: seq[MultiAddress]): Future[MultiAddress] {.async, base.} =
+  addrs: seq[MultiAddress]): Future[Opt[MultiAddress]] {.async, base.} =
```
arnetheduck, line 72: when using opt, don't forget to import/export `results` - it might work now, but if a module you depend on stops exporting results, this is break - always import what is used directly in the module and export that which appears in public interfaces

diegomrsantos: Thanks for letting me know. I've pushed a new commit.

lchenut (approved):

PR #780, opened 2022-09-29
Handle trying to write empty byte seq

diegomrsantos:


libp2p/stream/chronosstream.nim
```diff
@@ -130,6 +130,9 @@ proc completeWrite(
 method write*(s: ChronosStream, msg: seq[byte]): Future[void] =
   # Avoid a copy of msg being kept in the closure created by `{.async.}` as this
   # drives up memory usage
+  if msg.len == 0:
+    warn "Empty seq, nothing to write"
```
Menduist, line 134: Not even sure we need to log anything here, we should check what are the best practices here
Posix `write` just return `0 bytes written`, and doesn't consider this an error.

Menduist (dismissed):
Since this is blocking tor, we can merge this and think about logging when the need arises

libp2p/stream/chronosstream.nim
diegomrsantos: It isn't an error, execution will continue normally. I thought it was a good idea to add a warning because there might be cases when that's not really what the caller really wanted to do. What do you think?

Menduist: Let's put a `trace` instead, and we'll find out later if it was a bad idea :)

warn can be really bad, because if a peer can trigger a condition where you send an empty message, he could fill your logs & ddos you in some way. Still a lot better than crashing as we do now, but still not ideal

We just have to be careful to stay consistent: if ChronosStream allows empty messages, other forms of connection should too
But we can check that later, just open an issue so that we don't forget

diegomrsantos: Makes sense. Done.

Menduist (approved):

libp2p/stream/chronosstream.nim
diegomrsantos: https://github.com/status-im/nim-libp2p/issues/781

PR #786, opened 2022-10-21
Revert "Remove shallow copies"

Menduist:
Reverts status-im/nim-libp2p#782

Don't think it's the root of the CI issues, but why not


Menduist:
Indeed not

PR #788, opened 2022-10-21
Temporarily remove failing test

Menduist:
This test started failing out of the blue in the CI on linux with Nim 1.2 & 1.6.
It doesn't seem to be caused by new nim commits, nor any dependency change.
I also cannot reproduce it locally

I don't have time to investigate this right now, we'll open a revert PR as soon as this one is merged and investigate there, but we can't keep our CI locked up in the meantime

### DIFF
--- a/tests/pubsub/testgossipsub.nim
+++ b/tests/pubsub/testgossipsub.nim
@@ -715,57 +715,56 @@ suite "GossipSub":
 
     await allFuturesThrowing(nodesFut)
 
-  #TODO fix this test, see #788
-  #asyncTest "e2e - GossipSub with multiple peers (sparse)":
-  #  var runs = 10
+  asyncTest "e2e - GossipSub with multiple peers (sparse)":
+    var runs = 10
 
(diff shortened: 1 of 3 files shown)


tests/pubsub/testgossipsub.nim
```diff
@@ -665,55 +665,56 @@ suite "GossipSub":
...
-      check:
-        "foobar" in gossip.gossipsub
-
-    await allFuturesThrowing(
-      nodes.mapIt(
-        allFutures(
-          it.switch.stop())))
-
-    await allFuturesThrowing(nodesFut)
+  #TODO fix this test, see #787
```
Menduist, line 668:
```suggestion
  #TODO fix this test, see #788
```

lchenut (dismissed):

lchenut (dismissed):

Menduist:
Update: the failing test is most likely caused by the latest ubuntu release `20221018.2`

PR #792, opened 2022-10-27
Remove sleepAsync from tests/testswitch

Menduist:
It's flaky

We should aim to remove as many sleepAsync as possible, as they are generally synonyms with flaky tests. Exception being tests actually based on time

### DIFF
--- a/tests/testswitch.nim
+++ b/tests/testswitch.nim
@@ -252,9 +252,7 @@ suite "Switch":
     checkTracker(LPChannelTrackerName)
     checkTracker(SecureConnTrackerName)
 
-    await sleepAsync(1.seconds)
-
-    check:
+    check: await: checkExpiring:
       startCounts ==
         @[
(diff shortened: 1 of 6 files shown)


tests/testswitch.nim
```diff
@@ -625,42 +620,6 @@ suite "Switch":
...
-        discard
-
-    let handlerWait = acceptHandler()
-    let switch = newStandardSwitch(secureManagers = [SecureProtocol.Noise])
-
-    await switch.start()
-
-    var peerId = PeerId.init(PrivateKey.random(ECDSA, rng[]).get()).get()
-    let connectFut = switch.connect(peerId, transport.addrs)
-    await sleepAsync(500.millis)
```
Menduist, line 652: This test was not actually testing cancellation since the `connect` was finished far before 500 milliseconds

lchenut (approved):

PR #801, opened 2022-11-15
Fix example link from file rename

shyba:
Thanks for the project,
Just a minor broken link I found while reading the examples

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -2,5 +2,5 @@
 
 Welcome to the nim-libp2p documentation!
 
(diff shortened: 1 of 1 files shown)


Menduist:
Thanks for the PR! `.md` is used here because of the website (which translates the .nim into .md): https://status-im.github.io/nim-libp2p/docs/

But indeed, this breaks on github, so we'll have to find a better way

shyba:
Oh, sorry, I missed that.
The rendered version is awesome!

Closing as invalid.

PR #803, opened 2022-11-16
Reuseable install_nim action

Menduist:
This PR creates a reusable `install_nim` action that builds nim using the NBS
It will cache the built nim version for the specific arch & os.

The daily CI will run on `unstable`, delete those cached build, and re-create them every day (to make sure we have up to date nim versions)

If this action works well, it will be moved to the NBS, to be reused by other repos

Also note, only one CI run is allowed per PR now, the rest will be cancelled.

### DIFF
--- a/.github/workflows/multi_nim.yml
+++ b/.github/workflows/multi_nim.yml
@@ -5,7 +5,15 @@ on:
   workflow_dispatch:
 
 jobs:
+  delete-cache:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: snnaplab/delete-branch-cache-action@v1
+        with:
+          ref: unstable
+
   build:
+    needs: delete-cache
     timeout-minutes: 120
     strategy:
       fail-fast: false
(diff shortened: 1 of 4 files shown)


.github/workflows/multi_nim.yml
```diff
@@ -46,107 +52,14 @@ jobs:
...
-        run: |
-          curl -O -L -s -S https://raw.githubusercontent.com/status-im/nimbus-build-system/master/scripts/build_nim.sh
-          env MAKE="${MAKE_CMD} -j${ncpu}" ARCH_OVERRIDE=${PLATFORM} NIM_COMMIT=${{ matrix.branch }} \
-            QUICK_AND_DIRTY_COMPILER=1 QUICK_AND_DIRTY_NIMBLE=1 CC=gcc \
-            bash build_nim.sh nim csources dist/nimble NimBinaries
-          echo '${{ github.workspace }}/nim/bin' >> $GITHUB_PATH
+          os: ${{ matrix.target.os }}
+          shell: ${{ matrix.shell }}
+          nim_branch: ${{ matrix.branch }}
+          cpu: ${{ matrix.target.branch }}
```
Menduist, line 62:
```suggestion
          cpu: ${{ matrix.target.cpu }}
```

PR #807, opened 2022-11-23
Upgrade flow refactoring

Menduist:
This PR simplifies a lot the upgrade flow

The main thing is that `upgradeIncoming` and `upgradeOutgoing` now return the `Muxer`, that the switch / dialer will then pass to the connManager to be stored.
Previously, this relied on the upgrade manager storing the connection itself at the end of the process

Other things:
- connManager now only store the muxer (instead of muxer & connection). Connection is available in muxer.connection
- MultiStreamselect is now usable without holding the instance (since it's stateless anyway)
- Identify is now handled by the PeerStore instead of the upgrademanager
- Removed the "upgrade" future from the Connection which is no longer necessary
- Unified outgoing & incoming upgrade flows
- Simplified MuxerProvider

Sorry, big PR, but one thing lead to another


libp2p/connmanager.nim
```diff
@@ -67,16 +66,11 @@ type
...
-    muxer: Muxer
-    handle: Future[void]
-
   ConnManager* = ref object of RootObj
     maxConnsPerPeer: int
     inSema*: AsyncSemaphore
     outSema*: AsyncSemaphore
-    conns: Table[PeerId, HashSet[Connection]]
-    muxed: Table[Connection, MuxerHolder]
+    muxed: Table[PeerId, seq[Muxer]]
```
dryajov, line 73: how about non-muxed connections, like quic?

libp2p/multistream.nim
```diff
@@ -43,21 +42,21 @@ type
...
-proc select*(m: MultistreamSelect,
+proc select*(_: MultistreamSelect | type MultistreamSelect,
              conn: Connection,
              proto: seq[string]):
              Future[string] {.async.} =
-  trace "initiating handshake", conn, codec = m.codec
+  trace "initiating handshake", conn, codec = Codec
   ## select a remote protocol
-  await conn.write(m.codec) # write handshake
+  await conn.writeLp(Codec & "\n") # write handshake
```
dryajov, line 59: not a biggie, but the bare `write` was there to avoid adding the len-prefix on each write, since the messages are fixed length regardless...

libp2p/connmanager.nim
Menduist: Quic is still muxed, it's just native, so the muxer will just call the native procedures
https://github.com/status-im/nim-libp2p/pull/725/files#diff-e6732ffe4393737472ba25d1a6f85933cd5b5b2c85c7356a9684e05e4de8eabcR87-R92

I think it's a safe assumption to make that every libp2p is going to be muxed (otherwise we loose a big benefit of libp2p)

dryajov: I guess it can be bypassed... should be fine, I think overall it makes things a lot simpler.

libp2p/multistream.nim
dryajov: Yeah, I think it's a small optimization that bring quite a bit of confusion... probably not worth the trouble overall.

libp2p/multistream.nim
```diff
@@ -129,68 +128,90 @@ proc list*(m: MultistreamSelect,
...
-
-      trace "handle: got request", conn, ms
-      if ms.len() <= 0:
-        trace "handle: invalid proto", conn
-        await conn.write(Na)
-
-      if m.handlers.len() == 0:
-        trace "handle: sending `na` for protocol", conn, protocol = ms
-        await conn.write(Na)
-        continue
```
diegomrsantos, line 153: Why this has been removed?

libp2p/peerstore.nim
```diff
@@ -70,11 +75,15 @@ type
 
   PeerStore* {.public.} = ref object
     books: Table[string, BasePeerBook]
+    identify: Identify
```
diegomrsantos, line 78: Should the PeerStore have a protocol?

libp2p/connmanager.nim
```diff
@@ -67,19 +66,14 @@ type
...
     connEvents: array[ConnEventKind, OrderedSet[ConnEventHandler]]
     peerEvents: array[PeerEventKind, OrderedSet[PeerEventHandler]]
-    expectedConnections: Table[PeerId, Future[Connection]]
+    expectedConnections: Table[PeerId, Future[Muxer]]
```
diegomrsantos, line 76: Why change this to a Muxer?

libp2p/multistream.nim
```diff
@@ -54,13 +53,13 @@ template validateSuffix(str: string): untyped =
     else:
       raise newException(MultiStreamError, "MultistreamSelect failed, malformed message")
 
-proc select*(m: MultistreamSelect,
+proc select*(_: MultistreamSelect | type MultistreamSelect,
```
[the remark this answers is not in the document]
Menduist: `select` is "state-less", this change enables us to do `MultistreamSelect.select(xx)` instead of requiring an instance to do so

libp2p/multistream.nim
Menduist: This was moved to the handle line 131

libp2p/peerstore.nim
Menduist: "who is responsible for Identify" is a good question, previously this was a mix of the dialer, upgrader & connMngr. With this PR, this unifies it to "just the PeerStore", but we can change it to something else later

libp2p/connmanager.nim
Menduist: A Muxer now represent a "physical connection"

libp2p/connmanager.nim
```diff
@@ -201,14 +195,6 @@ proc triggerPeerEvents*(c: ConnManager,
     return
 
   try:
-    let count = c.connCount(peerId)
-    if event.kind == PeerEventKind.Joined and count != 1:
-      trace "peer already joined", peer = peerId, event = $event
-      return
-    elif event.kind == PeerEventKind.Left and count != 0:
-      trace "peer still connected or already left", peer = peerId, event = $event
-      return
```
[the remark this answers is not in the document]
Menduist: This logic has been moved to the call sites of triggerPeerEvents

libp2p/connmanager.nim
```diff
@@ -221,215 +207,153 @@ proc triggerPeerEvents*(c: ConnManager,
...
-proc contains*(c: ConnManager, conn: Connection): bool =
-  ## checks if a connection is being tracked by the
-  ## connection manager
-  ##
-
-  if isNil(conn):
-    return
-
-  return conn in c.conns.getOrDefault(conn.peerId)
-
```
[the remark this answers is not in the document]
Menduist: ConnManager no longer tracks "Connections", just the "Muxer" (which represent physical connections)

libp2p/multistream.nim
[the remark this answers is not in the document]
Menduist: Backward compat

libp2p/multistream.nim
```diff
@@ -129,68 +128,90 @@ proc list*(m: MultistreamSelect,
...
+      if not found:
+        for matcher in matchers:
+          if matcher(ms): found = true
+      if found:
+        trace "found handler", conn, protocol = ms
+        await conn.writeLp(ms & "\n")
+        conn.protocol = ms
+        return ms
+      trace "no handlers", conn, protocol = ms
+      await conn.writeLp(Na)
```
diegomrsantos, line 180:
```suggestion
    elif ms in protos or matchers.anyIt(it(ms)):
      trace "found handler", conn, protocol = ms
      await conn.writeLp(ms & "\n")
      conn.protocol = ms
      return ms
    else:
      trace "no handlers", conn, protocol = ms
      await conn.writeLp(Na)
```

libp2p/muxers/muxer.nim
```diff
@@ -32,61 +32,34 @@ type
...
+  if isNil(m): "nil"
+  else: shortLog(m.connection)
 chronicles.formatIt(Muxer): shortLog(it)
 
 # muxer interface
 method newStream*(m: Muxer, name: string = "", lazy: bool = false):
   Future[Connection] {.base, async, gcsafe.} = discard
-method close*(m: Muxer) {.base, async, gcsafe.} = discard
+method close*(m: Muxer) {.base, async, gcsafe.} =
+  if not isNil(m.connection):
```
diegomrsantos, line 55: should we check if `m` is `nil`?

libp2p/connmanager.nim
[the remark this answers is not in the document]
diegomrsantos: where?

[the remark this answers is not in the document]
Menduist:
```nim
    let muxers = c.muxed.getOrDefault(peerId).filterIt(it != mux)
    if muxers.len > 0:
      c.muxed[peerId] = muxers
    else:
      c.muxed.del(peerId)
      libp2p_peers.set(c.muxed.len.int64)
      await c.triggerPeerEvents(peerId, PeerEvent(kind: PeerEventKind.Left)) # will only happen when count == 0
```

```nim
  if newPeer:
    asyncSpawn c.triggerPeerEvents(
      peerId, PeerEvent(kind: PeerEventKind.Joined, initiator: dir == Direction.Out))
```

Though there might be a bug lingering somewhere around here

libp2p/connmanager.nim
```diff
@@ -67,19 +66,14 @@ type
...
-    muxer: Muxer
-    handle: Future[void]
-
   ConnManager* = ref object of RootObj
     maxConnsPerPeer: int
     inSema*: AsyncSemaphore
     outSema*: AsyncSemaphore
-    conns: Table[PeerId, HashSet[Connection]]
-    muxed: Table[Connection, MuxerHolder]
+    muxed: Table[PeerId, seq[Muxer]]
```
diegomrsantos, line 73: Why change the HashSet to a Seq?

libp2p/connmanager.nim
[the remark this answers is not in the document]
diegomrsantos: This part seems to be correct.

libp2p/connmanager.nim
diegomrsantos: Additionally, it should guarantee there are no duplications.

libp2p/peerstore.nim
diegomrsantos: I was thinking about `PeerInfo`, but it isn't in the `dialer`. How about having it in the `dialer` itself?

libp2p/dialer.nim
```diff
@@ -147,29 +149,21 @@ proc dialAndUpgrade(
...
-    return Opt.none(Connection)
+proc tryReusingConnection(self: Dialer, peerId: PeerId): Future[Opt[Muxer]] {.async.} =
+  let muxer = self.connManager.selectMuxer(peerId)
+  if muxer == nil:
+    return Opt.none(Muxer)
 
-  if conn.atEof or conn.closed:
-    # This connection should already have been removed from the connection
-    # manager - it's essentially a bug that we end up here - we'll fail
-    # for now, hoping that this will clean themselves up later...
```
diegomrsantos, line 158: Was it a bug? Has it been fixed?

libp2p/dialer.nim
```diff
@@ -179,32 +173,30 @@ proc internalConnect(
...
-    if conn.closed() or conn.atEof():
-      # This can happen when the other ends drops us
-      # before we get a chance to return the connection
-      # back to the dialer.
-      trace "Connection dead on arrival", conn
-      raise newLPStreamClosedError()
-
-    return conn
+    try:
+      await self.peerStore.identify(muxed)
```
diegomrsantos, line 192: If we are adding the `peerStore` here only cause of `Identify`, could we move the `Identify` here instead?

libp2p/muxers/muxer.nim
Menduist: I think it's common that calling `someProcedure(nil)` crashes, so I don't think so

libp2p/dialer.nim
Menduist: It's also used in the switch

libp2p/dialer.nim
Menduist: Seems like it

libp2p/connmanager.nim
Menduist: This seq should stay very small, so I don't worry too much about efficiency
A HashSet can introduce surprising ordering behavior, on the other hand

Menduist:
This PR is now stable on nimbus, anything else to change?

diegomrsantos (approved):

lchenut (approved):

PR #808, opened 2022-11-24
Fix: big messages in pubsub

Menduist:
They are currently capped to 4mb

### DIFF
--- a/libp2p/protocols/pubsub/rpc/protobuf.nim
+++ b/libp2p/protocols/pubsub/rpc/protobuf.nim
@@ -304,14 +304,15 @@ proc decodeMessages*(pb: ProtoBuffer): ProtoResult[seq[Message]] {.inline.} =
   if ? pb.getRepeatedField(2, msgpbs):
     trace "decodeMessages: read messages", count = len(msgpbs)
     for item in msgpbs:
-      msgs.add(? decodeMessage(initProtoBuffer(item)))
+      # size is constrained at the network level
+      msgs.add(? decodeMessage(initProtoBuffer(item, maxSize = uint.high)))
   else:
     trace "decodeMessages: no messages found"
   ok(msgs)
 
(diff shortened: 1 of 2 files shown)


lchenut (approved):

libp2p/protocols/pubsub/rpc/protobuf.nim
```diff
@@ -304,14 +304,15 @@ proc decodeMessages*(pb: ProtoBuffer): ProtoResult[seq[Message]] {.inline.} =
   if ? pb.getRepeatedField(2, msgpbs):
     trace "decodeMessages: read messages", count = len(msgpbs)
     for item in msgpbs:
-      msgs.add(? decodeMessage(initProtoBuffer(item)))
+      # size is constrained at the network level
+      msgs.add(? decodeMessage(initProtoBuffer(item, maxSize = uint.high)))
```
dryajov, line 308: Do we really want messages this high by default? It should probably be configurable, but `2^64` opens up quite a big attack vector...

Menduist: As said in the comment, the size is constrained in the network layer:
https://github.com/status-im/nim-libp2p/blob/c43aacdc81ea044518b24e82ff942671314e3dc8/libp2p/protocols/pubsub/pubsubpeer.nim#L126

arnetheduck: wouldn't it then make sense to remove maxSize from initProtobuffer? it feels like the wrong layer to be checking for large messages (there is no compression bomb for example - quite the opposite since protobuf generally adds overhead, except for small integers) - this means that whatever overflow we check for *must* be checked earlier.

Menduist: Probably, yes

PR #809, opened 2022-11-25
Fix PubSub subscribe on connection race condition

Menduist:
This PR fixes a subtle race condition that I discovered while working on https://github.com/status-im/nim-libp2p/pull/807. Since 807 renders the upgrade "faster" (less callback and asyncSpawn etc), it's highlights some bugs, which is cool

I think this race condition can also happen without 807.
1 - A peer connect to us, and opens the pubsub stream very fast, preloaded with the list of topic its subscribed to
2 - We see that we have a topic in common, so we GRAFT him
3 - "sendConn" stream is setup.

Most of the time, 3 happens before 2. But when 2 happens first, the GRAFT is simply swallowed by this:
https://github.com/status-im/nim-libp2p/blob/1711c204eab2596e4bd42b162d630bc262704daf/libp2p/protocols/pubsub/pubsubpeer.nim#L253
and we end up with an assymetric mesh.

This PR simply wait for both direction to be setup before handling the peer messages.
It's a simplistic fix, and doesn't handle more complex cases (ie, the send stream getting closed, but not the receive one)
The better fix would be to get rid of this silent swallow, and make sure it never happens, but that's a much bigger effort

### DIFF
--- a/libp2p/protocols/pubsub/pubsub.nim
+++ b/libp2p/protocols/pubsub/pubsub.nim
@@ -202,7 +202,7 @@ proc broadcast*(
     # Fast path that only encodes message once
     let encoded = encodeRpcMsg(msg, p.anonymize)
     for peer in sendPeers:
-      peer.sendEncoded(encoded)
+      asyncSpawn peer.sendEncoded(encoded)
 
 proc sendSubs*(p: PubSub,
                peer: PubSubPeer,
@@ -307,8 +307,6 @@ proc getOrCreatePeer*(
   # metrics
   libp2p_pubsub_peers.set(p.peers.len.int64)
 
-  pubSubPeer.connect()
-
   return pubSubPeer
 
 proc handleData*(p: PubSub, topic: string, data: seq[byte]): Future[void] =
@@ -382,7 +380,8 @@ method subscribePeer*(p: PubSub, peer: PeerId) {.base, gcsafe.} =
   ## messages
   ##
 
-  discard p.getOrCreatePeer(peer, p.codecs)
+  let pubSubPeer = p.getOrCreatePeer(peer, p.codecs)
+  pubSubPeer.connect()
 
 proc updateTopicMetrics(p: PubSub, topic: string) =
   # metrics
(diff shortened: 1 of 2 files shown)


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -182,6 +186,8 @@ proc connectOnce(p: PubSubPeer): Future[void] {.async.} =
 
     await handle(p, newConn)
   finally:
+    if not p.connectedFut.finished():
+      p.connectedFut.fail(newException(LPError, "can't establish conn"))
```
diegomrsantos, line 190: Is it possible to create a test?

Menduist: Testing the race condition is a bit tricky as it's quite random
Though when we'll merge 807, it will be a lot more frequent (but still random)

diegomrsantos: Should it happen if we repeat it a large enough number of times?

Menduist: I'll do a better fix & try to create a test by manually slowing down the connection

lchenut (approved):

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -382,7 +380,8 @@ method subscribePeer*(p: PubSub, peer: PeerId) {.base, gcsafe.} =
   ## messages
   ##
 
-  discard p.getOrCreatePeer(peer, p.codecs)
+  let pubSubPeer = p.getOrCreatePeer(peer, p.codecs)
```
diegomrsantos, line 383: This creates a remote peer `pubsubpeer` but with the local peer `p.codes` and adds the remote peer to the local peer's peers table. When this executes https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/pubsub.nim#L379 the remote peer continues with the wrong codecs, as it's already present in the table.

PR #811, opened 2022-11-29
Limit number of streams per protocol per peer

Menduist:
Each protocol can specify a maximum number of (inbounds) streams per Peer (not connection!). Above that, new streams will be closed directly.

This default to 10, could be finer tuned for each libp2p protocol, but I think it's pretty good default

### DIFF
--- a/libp2p/multistream.nim
+++ b/libp2p/multistream.nim
@@ -33,20 +33,19 @@ type
 
   MultiStreamError* = object of LPError
 
-  HandlerHolder* = object
+  HandlerHolder* = ref object
     protos*: seq[string]
     protocol*: LPProtocol
     match*: Matcher
+    openedStreams: CountTable[PeerId]
 
   MultistreamSelect* = ref object of RootObj
     handlers*: seq[HandlerHolder]
     codec*: string
-    openedStreams: Table[PeerId, CountTable[LPProtocol]]
 
 proc new*(T: typedesc[MultistreamSelect]): T =
   T(
     codec: MSCodec,
-    openedStreams: initTable[PeerId, CountTable[LPProtocol]]()
   )
 
 template validateSuffix(str: string): untyped =
@@ -174,25 +173,22 @@ proc handle*(m: MultistreamSelect, conn: Connection, active: bool = false) {.asy
           if (not isNil(h.match) and h.match(ms)) or h.protos.contains(ms):
             trace "found handler", conn, protocol = ms
 
-            let countTable = m.openedStreams.mgetOrPut(
-                                  conn.peerId, initCountTable[LPProtocol]())
-            let
-              protocol = h.protocol
-              maxIncomingStreams =
-                protocol.maxIncomingStreams.get(DefaultMaxIncomingStreams)
-            if countTable.getOrDefault(protocol) >= maxIncomingStreams:
+            var protocolHolder = h
+            let maxIncomingStreams =
+                  protocolHolder.protocol.maxIncomingStreams.get(DefaultMaxIncomingStreams)
+            if protocolHolder.openedStreams.getOrDefault(conn.peerId) >= maxIncomingStreams:
               debug "Max streams for protocol reached, blocking new stream",
                 conn, protocol = ms, maxIncomingStreams
               return
-            m.openedStreams[conn.peerId].inc(protocol)
+            protocolHolder.openedStreams.inc(conn.peerId)
             try:
               await conn.writeLp(ms & "\n")
               conn.protocol = ms
-              await protocol.handler(conn, ms)
+              await protocolHolder.protocol.handler(conn, ms)
             finally:
-              m.openedStreams[conn.peerId].inc(protocol, -1)
-              if m.openedStreams[conn.peerId].len == 0:
-                m.openedStreams.del(conn.peerId)
+              protocolHolder.openedStreams.inc(conn.peerId, -1)
+              if protocolHolder.openedStreams[conn.peerId] == 0:
+                protocolHolder.openedStreams.del(conn.peerId)
             return
         debug "no handlers", conn, protocol = ms
         await conn.write(Na)
--- a/libp2p/multistream.nim
+++ b/libp2p/multistream.nim
@@ -174,8 +174,7 @@ proc handle*(m: MultistreamSelect, conn: Connection, active: bool = false) {.asy
             trace "found handler", conn, protocol = ms
 
             var protocolHolder = h
-            let maxIncomingStreams =
-                  protocolHolder.protocol.maxIncomingStreams.get(DefaultMaxIncomingStreams)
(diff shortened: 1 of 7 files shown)


libp2p/multistream.nim
```diff
@@ -169,9 +173,25 @@ proc handle*(m: MultistreamSelect, conn: Connection, active: bool = false) {.asy
...
-            await conn.writeLp(ms & "\n")
-            conn.protocol = ms
-            await h.protocol.handler(conn, ms)
+
+            let countTable = m.openedStreams.mgetOrPut(
+                                  conn.peerId, initCountTable[LPProtocol]())
+            let
+              protocol = h.protocol
+              maxStreams = protocol.maxStreams.get(DefaultMaxStreams)
+            if countTable.getOrDefault(protocol) > maxStreams:
```
diegomrsantos, line 182: Should it be` >=`?

libp2p/protocols/protocol.nim
```diff
@@ -26,11 +30,16 @@ type
     codecs*: seq[string]
     handler*: LPProtoHandler ## this handler gets invoked by the protocol negotiator
     started*: bool
+    maxStreams*: Opt[int]
```
diegomrsantos, line 33: Can't this be an `int` with a default value of `DefaultMaxStreams`?

libp2p/protocols/protocol.nim
```diff
@@ -26,11 +30,16 @@ type
...
 
 method init*(p: LPProtocol) {.base, gcsafe.} = discard
 method start*(p: LPProtocol) {.async, base.} = p.started = true
 method stop*(p: LPProtocol) {.async, base.} = p.started = false
 
+proc hash*(protocol: LPProtocol): Hash =
```
diegomrsantos, line 39: Is this necessary to use LPProtocol as a key in the table?

tests/testmultistream.nim
```diff
@@ -278,6 +278,68 @@ suite "Multistream select":
...
+      check: (await msDial.select(conn, "/test/proto/1.0.0")) == true
+      check: string.fromBytes(await conn.readLp(1024)) == "Hello!"
+      await conn.close()
+
+    # Fill up the 5 allowed streams
+    var dialers: seq[Future[void]]
+    for _ in 0..5:
+      dialers.add(connector())
+
+    # This one will fail somehow
```
diegomrsantos, line 325: What do you mean by "will fail somehow"?

tests/testmultistream.nim
```diff
@@ -278,6 +278,68 @@ suite "Multistream select":
...
+    expect(CatchableError):
+      waitFor(connector())
+    # check that the dialers aren't finished
+    check: (await dialers[0].withTimeout(10.milliseconds)) == false
```
diegomrsantos, line 329: Why the first dialer shouldn't be finished?

libp2p/protocols/protocol.nim
Menduist: Default values for object just landed for nim 2.0: https://github.com/nim-lang/Nim/pull/20480
So we will probably be able to use them in 2 years, once we drop support for 1.6

In the meantime, that's the best way to do this, imo (other solution would be to use `0` as a default = `10`, but I find this not great)

libp2p/protocols/protocol.nim
Menduist: We could also use the list of codecs, but that might do weird things when the user uses a [matcher](https://github.com/status-im/nim-libp2p/blob/b3d9360dfcc3c395cf0768567685e923587e3b77/libp2p/multistream.nim#L39)?
Or maybe we just put the CountTable directly in LPProtocol, idk, wdyt?

tests/testmultistream.nim
Menduist: The stream will get closed during negotiation, so `select` will pop an error

tests/testmultistream.nim
Menduist: Because they are all blocked by `blocker`, see `testHandler`

tests/testmultistream.nim
```diff
@@ -278,6 +278,68 @@ suite "Multistream select":
...
+      check: string.fromBytes(await conn.readLp(1024)) == "Hello!"
+      await conn.close()
+
+    # Fill up the 5 allowed streams
+    var dialers: seq[Future[void]]
+    for _ in 0..5:
+      dialers.add(connector())
+
+    # This one will fail somehow
+    expect(CatchableError):
```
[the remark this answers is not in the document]
Menduist: Yes because we opened too many streams for that protocol, so we refuse this one

libp2p/multistream.nim
Menduist: Yup, fixed

libp2p/protocols/protocol.nim
diegomrsantos: Is it possible to use the object's name?

tests/testmultistream.nim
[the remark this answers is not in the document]
diegomrsantos: How do you know the difference? Seems the proc always return nothing.

libp2p/protocols/protocol.nim
diegomrsantos: can't we use a new proc?

tests/testmultistream.nim
diegomrsantos: Nitpicking, but only one dialer is being checked.

libp2p/protocols/protocol.nim
Menduist: We would have to update every protocol to call this `new`, but some protocols are in the users codebase, so we can't :/

tests/testmultistream.nim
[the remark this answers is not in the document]
Menduist: It will raise an exception

diegomrsantos (dismissed):

mxinden:
Thanks @Menduist!

libp2p/multistream.nim
```diff
@@ -169,9 +172,22 @@ proc handle*(m: MultistreamSelect, conn: Connection, active: bool = false) {.asy
...
+              debug "Max streams for protocol reached, blocking new stream",
+                conn, protocol = ms, maxIncomingStreams
+              return
+            protocolHolder.openedStreams.inc(conn.peerId)
+            try:
+              await conn.writeLp(ms & "\n")
+              conn.protocol = ms
+              await protocolHolder.protocol.handler(conn, ms)
+            finally:
+              protocolHolder.openedStreams.inc(conn.peerId, -1)
```
arnetheduck, line 188: there's `dec` too..

Menduist: Not on CountTables, see https://github.com/nim-lang/Nim/pull/19256 for discussion

arnetheduck: doh!

PR #813, opened 2022-11-30
Less warnings

Menduist:
Also see https://github.com/status-im/nim-chronos/pull/332


libp2p/utility.nim
```diff
@@ -65,3 +65,16 @@ when defined(libp2p_agents_metrics):
   const
     KnownLibP2PAgents* {.strdefine.} = "nim-libp2p"
     KnownLibP2PAgentsSeq* = KnownLibP2PAgents.safeToLowerAscii().tryGet().split(",")
+
+template tryException*(body: untyped): untyped =
```
Menduist, line 69: Needs a more descriptive name
exceptionToAssert? Idk

tests/helpers.nim
```diff
@@ -55,10 +55,6 @@ template checkTrackers*() =
     if tracker.isLeaked():
       checkpoint tracker.dump()
       fail()
-  # Also test the GC is not fooling with us
-  try:
-    GC_fullCollect()
-  except: discard
```
Menduist, line 61: Let's keep this to be safe

tests/testmultiaddress.nim
```diff
@@ -332,7 +332,7 @@ suite "MultiAddress test suite":
       var r = false
```
Menduist, line 332: Why not expect(CatchableError) instead?

arnetheduck:
see also: https://github.com/nim-lang/Nim/pull/21728

tests/testnative.nim
```diff
@@ -1,3 +1,5 @@
+{.warning[UnusedImport]: off.}
```
Menduist, line 1: I think `{.used.}` should take care of this, maybe its missing from some files

libp2p/utility.nim
```diff
@@ -65,3 +65,16 @@ when defined(libp2p_agents_metrics):
...
+  block:
+    var res: type(body)
+    when defined(nimHasWarnBareExcept):
+      {.push warning[BareExcept]:off.}
+    try:
+      res = body
+    except CatchableError as exc: raise exc
```
Menduist, line 76: Maybe we should add a
`except Defect as exc: raise exc`
Too, to avoid catching defects

tests/helpers.nim
diegomrsantos: How does it increase safety? It is used only in tests and apparently, it is not necessary. We can always add it back If we start having issues.

libp2p/utility.nim
diegomrsantos: I didn't like the name either, thanks for the suggestion.

tests/testnative.nim
arnetheduck: the latter

libp2p/utility.nim
diegomrsantos: And then making `except Exception: discard`?

Menduist: I'd rather assert on this unexpected case than swallow it

tests/helpers.nim
Menduist: Honestly I have no idea why this is here, and the git blame doesn't really explains it either
But I think it's better to keep it, it doesn't cost anything and may cause weird issue that someone already had to go through

diegomrsantos: I have no problem leaving it, but IMHO it's better to try to make the codebase cleaner and remove code that no one can explain why it is there, than leaving it. Especially in tests where the risks are low. I would even argue that it is safer to let it crash and fix the root cause than run a full GC and swallow whatever is raised.

Menduist: I don't think GC_fullCollect can actually raise anything
Issues it might cause are more related to flaky tests etc
You can just use exceptionToAssert to be safe regarding exceptions

diegomrsantos: I reverted it.

libp2p/crypto/minasn1.nim
```diff
@@ -448,21 +449,21 @@ proc getLength(ab: var Asn1Buffer): Asn1Result[uint64] =
...
     if b == 0x80'u8:
       return err(Asn1Error.Indefinite)
     if b == 0xFF'u8:
       return err(Asn1Error.Incorrect)
-    let octets = cast[uint64](b and 0x7F'u8)
+    let octets = safeConvert[uint64](b and 0x7F'u8)
     if octets > 8'u64:
       return err(Asn1Error.Overflow)
-    if ab.isEnough(int(octets)):
+    if ab.isEnough(cast[int](octets)): # FIXME uint64 to int potentially unsafe
```
Menduist, line 462: This is safe thanks to the if above, we know that `octets` will fit in `int`

libp2p/crypto/minasn1.nim
```diff
@@ -519,15 +520,15 @@ proc read*(ab: var Asn1Buffer): Asn1Result[Asn1Field] =
         if length != 1:
           return err(Asn1Error.Incorrect)
 
-        if not ab.isEnough(int(length)):
+        if not ab.isEnough(cast[int](length)): # FIXME uint64 to int potentially unsafe
```
Menduist, line 523: In these cases we shouldn't cast since this would fail silently by producing junk value

libp2p/crypto/rsa.nim
```diff
@@ -24,6 +24,8 @@ import stew/[results, ctops]
 # We use `ncrutils` for constant-time hexadecimal encoding/decoding procedures.
 import nimcrypto/utils as ncrutils
 
+import ../utility
```
Menduist, line 27: Not needed?

libp2p/utility.nim
```diff
@@ -65,3 +65,23 @@ when defined(libp2p_agents_metrics):
   const
     KnownLibP2PAgents* {.strdefine.} = "nim-libp2p"
     KnownLibP2PAgentsSeq* = KnownLibP2PAgents.safeToLowerAscii().tryGet().split(",")
+
+template safeConvert*[T: SomeInteger, S: Ordinal](a: S): T =
```
Menduist, line 69:
```suggestion
template safeConvert*[T: SomeInteger, S: Ordinal](a: S): T =
   # Converts `s` to `T` iff it can't trigger defects
```
Or something like that?

diegomrsantos: It's not even about defects, this compiles, and prints 18446744073709551615
```nim
let i = -1
echo uint64(i)
```

libp2p/crypto/minasn1.nim
diegomrsantos: what should be done?

Menduist: In this case, I think we should return a int64 from `getLength` (and an error if it overflows), that would avoid all of those

libp2p/crypto/minasn1.nim
```diff
@@ -355,7 +356,7 @@ proc asn1EncodeOid*(dest: var openArray[byte], value: openArray[int]): int =
   var res = 1
   var oidlen = 1
   for i in 2..<len(value):
-    oidlen += asn1EncodeTag(buffer, cast[uint64](value[i]))
+    oidlen += asn1EncodeTag(buffer, cast[uint64](value[i])) # FIXME int to uint64 potentially unsafe
```
Menduist, line 359: It seems like this proc could take a `value: openArray[byte]` and it work without the cast

libp2p/crypto/minasn1.nim
diegomrsantos: Please review the new code

libp2p/crypto/minasn1.nim
diegomrsantos: There's a `proc asn1EncodeOid*(dest: var openArray[byte], value: openArray[byte]): int` defined right below. There's nothing using the one with int, I'm removing it.

libp2p/crypto/minasn1.nim
```diff
@@ -519,15 +498,15 @@ proc read*(ab: var Asn1Buffer): Asn1Result[Asn1Field] =
...
+        if not ab.isEnough(length):
           return err(Asn1Error.Incomplete)
 
         let b = ab.buffer[ab.offset]
         if b != 0xFF'u8 and b != 0x00'u8:
            return err(Asn1Error.Incorrect)
 
         field = Asn1Field(kind: Asn1Tag.Boolean, klass: aclass,
-                          index: ttag, offset: int(ab.offset),
+                          index: ttag, offset: safeConvert[int](ab.offset),
```
lchenut, line 509: `ab.offset` is already an int, why is it convert?

lchenut (approved):

PR #814, opened 2022-12-05
Autonat service

diegomrsantos:
This a service which uses the Autonat protocol to estimate a node network reachability on the internet.


libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,128 @@
...
+      let ma = await self.autonat.dialMe(peerId)
+      (NetworkReachability.Reachable, true)
+    except AutonatUnreachableError:
+      (NetworkReachability.NotReachable, true)
+    except AutonatError:
+      (NetworkReachability.Unknown, false)
+  await self.handleAnswer(ans)
+  return known
+
+proc askConnectedPeers(self: AutonatService, switch: Switch) {.async.} =
```
alrevuelta, line 97: As an idea I think we can simplify this by removing `handleAnswer` and not store the states in `t`.

very very roughly
```nim
var states: seq[NetworkState]
for peer in peers:
  if peersToAsk == 0:
    break
  let state = await askPeer(self, switch, peer)
  if state != Unknown:
    peersToAsk -= 1
    states.add(state)

if states.countIt(it == Reachable) > maxConfidence:
   ...
```
and then trigger the `newStatusHandler`.

libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,128 @@
...
+    self.t.inc(ans)
+  else:
+    if self.t[self.networkReachability] > 0:
+      self.t.inc(self.networkReachability, -1)
+    if self.t[ans] < self.maxConfidence:
+      self.t.inc(ans)
+    if self.t[ans] == self.maxConfidence or self.t[self.networkReachability] == 0:
+      self.networkReachability = ans
+
+  if self.t[self.networkReachability] == self.maxConfidence:
```
alrevuelta, line 77: just a question, since im not sure I get this. When are we calling the `newStatusHandler`. I am expecting it to be called every time the status changes, so unsure how the `maxConfidence` applies here.

libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,128 @@
...
+  await askConnectedPeers(self, switch)
+
+method stop*(self: AutonatService, switch: Switch): Future[bool] {.async, public.} =
+  let hasBeenStopped = await procCall Service(self).stop(switch)
+  if hasBeenStopped and self.scheduleInterval.isSome and self.registerLoop != nil:
+    self.registerLoop.cancel()
+    self.registerLoop = nil
+  return hasBeenStopped
+
+proc onNewStatuswithMaxConfidence*(self: AutonatService, f: NewStatusHandler) =
```
alrevuelta, line 127: not sure how `maxConfidence` applies here. How about `setNewStatusHandler`?

tests/testautonatservice.nim
```diff
@@ -0,0 +1,150 @@
...
+    let switch1 = createSwitch()
+    let switch2 = createSwitch()
+    let switch3 = createSwitch()
+    let switch4 = createSwitch()
+
+    let autonatStub = AutonatStub.new(expectedDials = 3)
+    autonatStub.returnSuccess = false
+
+    let autonatService = AutonatService.new(autonatStub, newRng(), some(1.seconds))
+    switch1.addService(autonatService)
```
alrevuelta, line 45: not related per se to the test, but does it make sense to pass the `autonatService` in the switch builder `.withAutonat()`?

So somthing like `.withAutonat(autonatService)` so that we avoid the calling `addService`?

tests/testautonatservice.nim
```diff
@@ -0,0 +1,150 @@
...
+    await switch1.connect(switch4.peerInfo.peerId, switch4.peerInfo.addrs)
+
+    await autonatStub.finished
+
+    check autonatService.networkReachability() == NetworkReachability.Reachable
+
+    await allFuturesThrowing(
+      switch1.stop(), switch2.stop(), switch3.stop(), switch4.stop())
+
+  asyncTest "Autonat Service Full Reachability test":
```
alrevuelta, line 95: wondering if we could have one test case without the stub. I mean, I understand `Unknown` and `NotReachable` states might be tricky to simulate, but the `Reachable` state should be possible to get it without a stub right?

And its perhaps closer to reality?

libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,128 @@
...
+    if await askPeer(self, switch, peer):
+      peersToAsk -= 1
+
+proc register(service: AutonatService, switch: Switch, interval: Duration) {.async.} =
+  heartbeat "Register AutonatService run", interval:
+    await service.run(switch)
+
+method setup*(self: AutonatService, switch: Switch): Future[bool] {.async.} =
+  let hasBeenSettedUp = await procCall Service(self).setup(switch)
+  if hasBeenSettedUp and self.scheduleInterval.isSome:
```
alrevuelta, line 113: should we force a `scheduleInterval` to be mandatory?

alrevuelta (commented):
looks great, just some minor comments.

libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,128 @@
...
+    t: initCountTable[NetworkReachability](),
+    autonat: autonat,
+    rng: rng,
+    numPeersToAsk: numPeersToAsk,
+    maxConfidence: maxConfidence)
+
+proc networkReachability*(self: AutonatService): NetworkReachability {.inline.} =
+  return self.networkReachability
+
+proc handleAnswer(self: AutonatService, ans: NetworkReachability) {.async.} =
```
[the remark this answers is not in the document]
diegomrsantos: See previous answer.

libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,128 @@
...
+import ../discovery/[rendezvousinterface, discoverymngr]
+import ../utils/heartbeat
+import ../crypto/crypto
+
+type
+  AutonatService* = ref object of Service
+    registerLoop: Future[void]
+    scheduleInterval: Option[Duration]
+    networkReachability: NetworkReachability
+    t: CountTable[NetworkReachability]
```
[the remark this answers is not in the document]
diegomrsantos: I believe this state must outlive `run` calls. Am I missing something?

libp2p/services/autonatservice.nim
diegomrsantos: The status can also change with zero confidence, I thought it wouldn't make sense to call in this case. Wdyt?

tests/testautonatservice.nim
diegomrsantos: Good point, thanks for reminding me about this. Autonat and AutonatService are different things. We are planning to implement clearer abstractions soon. For now, perhaps something like `withServices(...)`?

tests/testautonatservice.nim
diegomrsantos: That was what I did initially, but then I had to use `sleepAsync` and I had issues with it on windows. In the end I thought that Autonat should be tested in the testautonat instead. Wdyt?

libp2p/services/autonatservice.nim
diegomrsantos: Most likely.

libp2p/services/autonatservice.nim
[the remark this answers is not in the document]
alrevuelta: Ah perhaps im missing something. What I had in mind is every `interval` we ask `x` peers and if `y` flags us as `Reachable`, that's it. So in the next interval, I think we don't need the `t`.

Of course I'm assuming that we do something like [this](https://github.com/status-im/nim-libp2p/pull/814#discussion_r1042073528). So within the same function we await until all peers have replied.

libp2p/services/autonatservice.nim
alrevuelta: mm I mean what i expected is that for every `interval` we call `newStatusHandler` with the new status. It can be `unknown`, `unreachable` or `rechable` depending on the result (which depends on `maxConfidence`) but is called no matter what.

tests/testautonatservice.nim
alrevuelta: ah right. `withServices()` sounds good, but totally up to you :)
nothing against the current implementation though then.

tests/testautonatservice.nim
alrevuelta: mm right, that already tested.

libp2p/services/autonatservice.nim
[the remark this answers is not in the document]
diegomrsantos: Now I understand what you mean. If in a run we ask 5 peers, 2 say NotReachable, 2 say Reachable and 1 says Unknown. Should we keep the current status, discard all answers and start from scratch in the next run? 

Btw I think my current algo is wrong. It would end up with an empty table in the case above and it's not what we probably want.

libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,127 @@
...
+  AutonatService* = ref object of Service
+    registerLoop: Future[void]
+    scheduleInterval: Duration
+    networkReachability: NetworkReachability
+    t: CountTable[NetworkReachability]
+    autonat: Autonat
+    newStatusHandler: NewStatusHandler
+    rng: ref HmacDrbgContext
+    numPeersToAsk: int
+    maxConfidence: int
```
alrevuelta, line 35: one last comment, not sure if nitpick. `minConfidence` is a better naming? As i see it is the minimum amount of states that we use to consider that the node is `Reachable` or `NotReachable`

alrevuelta (approved):
lgtm! thanks!

libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,154 @@
...
+
+method stop*(self: AutonatService, switch: Switch): Future[bool] {.async, public.} =
+  let hasBeenStopped = await procCall Service(self).stop(switch)
+  if hasBeenStopped:
+    if not isNil(self.registerLoop):
+      self.registerLoop.cancel()
+      self.registerLoop = nil
+    if not isNil(self.newConnectedPeerHandler):
+      switch.connManager.removePeerEventHandler(self.newConnectedPeerHandler, PeerEventKind.Joined)
+  return hasBeenStopped
```
Menduist, line 151: This is very confusing, to refacto during the service rework

libp2p/switch.nim
```diff
@@ -74,6 +74,28 @@ type
...
+  if self.inUse:
+    warn "service setup has already been called"
+    return false
+  self.inUse = true
+  return true
+
+method run*(self: Service, switch: Switch) {.base, async, gcsafe, public.} =
+  doAssert(false, "Not implemented!")
+
+method stop*(self: Service, switch: Switch): Future[bool] {.base, async, gcsafe, public.} =
```
Menduist, line 93: These shouldn't be public

tests/helpers.nim
```diff
@@ -117,3 +117,10 @@ proc checkExpiringInternal(cond: proc(): bool {.raises: [Defect], gcsafe.} ): Fu
 
 template checkExpiring*(code: untyped): untyped =
   check await checkExpiringInternal(proc(): bool = code)
+
+type
+  Awaiter* = ref object of RootObj
```
Menduist, line 122: Why this wrapper?

libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,154 @@
...
+    minConfidence: minConfidence,
+    dialTimeout: dialTimeout)
+
+proc networkReachability*(self: AutonatService): NetworkReachability {.inline.} =
+  return self.networkReachability
+
+proc handleAnswer(self: AutonatService, ans: NetworkReachability) {.async.} =
+
+  if self.answerDeque.len == self.maxQueueSize:
+    self.answerDeque.popFirst()
```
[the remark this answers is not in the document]
diegomrsantos: what would be the benefit?

libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,154 @@
...
+
+  self.answerDeque.addLast(ans)
+
+  let reachableCount = toSeq(self.answerDeque.items).countIt(it == NetworkReachability.Reachable)
```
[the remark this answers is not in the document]
diegomrsantos: Docs say they use a seq https://nim-lang.org/docs/deques.html. We could create an issue to add a toSeq proc that could potentially be cheap. Anyway, I believe using toSeq has negligible performance issue as the size is very small.

libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,154 @@
...
+  else:
+    let notReachableCount = toSeq(self.answerDeque.items).countIt(it == NetworkReachability.NotReachable)
+    let confidence = notReachableCount / self.maxQueueSize
+    if confidence >= self.minConfidence:
+      self.networkReachability = NetworkReachability.NotReachable
+      self.confidence = some(confidence)
+      libp2p_autonat_reachability_confidence.set(confidence, ["NotReachable"])
+    else:
+      self.networkReachability = NetworkReachability.Unknown
+      self.confidence = none(float)
```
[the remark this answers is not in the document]
Menduist: Right, edited to be less clever

libp2p/switch.nim
diegomrsantos: why not?

tests/helpers.nim
diegomrsantos: The trick to use a global inside a gcsafe.

libp2p/services/autonatservice.nim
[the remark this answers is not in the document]
diegomrsantos: My idea here was to be more explicit about what was happening.

libp2p/services/autonatservice.nim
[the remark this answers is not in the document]
diegomrsantos: Can't see the difference from before regarding cleverness. Actually, your first version was right and my initial comment was wrong so I deleted it.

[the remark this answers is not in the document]
Menduist: Yeah my initial version was misleading, fixed it a bit, my point is we can make something explicit & DRY

libp2p/switch.nim
Menduist: public = API meant for the user that won't change until v2
ATM services are not stable (we just introduced them), and not meant for the user (we just introduced them)

tests/helpers.nim
Menduist: That seems very weird, `Future` is already a `ref object of RootObj`, so this shouldn't change anything, and you are not even using it as a global

libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,151 @@
...
+  const reachabilityPriority = [Reachable, NotReachable]
+  for reachability in reachabilityPriority:
+    let confidence = self.answers.countIt(it == reachability) / self.maxQueueSize
+    libp2p_autonat_reachability_confidence.set(value = confidence, labelValues = [$reachability])
+    if self.confidence.isNone and confidence >= self.minConfidence:
+      self.networkReachability = reachability
+      self.confidence = some(confidence)
+
+  if not isNil(self.statusAndConfidenceHandler):
+    await self.statusAndConfidenceHandler(self.networkReachability, self.confidence)
```
Menduist, line 90: When ran inside `askConnectedPeers`, this will fire the callback `numPeersToAsk` times (even more if we have unknowns). This is minor, but maybe we could call the callback at the end of `askConnectedPeers` & `newConnectedPeerHandler` instead, wdyt?

libp2p/services/autonatservice.nim
```diff
@@ -0,0 +1,151 @@
...
+    NotReachable, Reachable, Unknown
+
+  StatusAndConfidenceHandler* = proc (networkReachability: NetworkReachability, confidence: Option[float]): Future[void]  {.gcsafe, raises: [Defect].}
+
+proc new*(
+  T: typedesc[AutonatService],
+  autonat: Autonat,
+  rng: ref HmacDrbgContext,
+  scheduleInterval: Option[Duration] = none(Duration),
+  disableAskNewConnectedPeers = false,
```
Menduist, line 51: `askNewConnectedPeers = true` instead? Negation feels a bit weird, even though it's the default

tests/stubs/autonatstub.nim
```diff
@@ -0,0 +1,40 @@
...
+    proc tryFinish() =
+      if self.dials == self.expectedDials:
+        self.finished.complete()
+
+    if self.returnSuccess:
+      tryFinish()
+      return MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet()
+    else:
+      tryFinish()
+      raise newException(AutonatUnreachableError, "")
```
Menduist, line 40:
```suggestion
    if self.dials == self.expectedDials:
      self.finished.complete()

    if self.returnSuccess:
      return MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet()
    else:
      raise newException(AutonatUnreachableError, "")
```

am I missing something obvious? :sweat_smile:

tests/helpers.nim
```diff
@@ -116,4 +116,4 @@ proc checkExpiringInternal(cond: proc(): bool {.raises: [Defect], gcsafe.} ): Fu
       await sleepAsync(1.millis)
 
 template checkExpiring*(code: untyped): untyped =
-  check await checkExpiringInternal(proc(): bool = code)
+  check await checkExpiringInternal(proc(): bool = code)
```
Menduist, line 119: Your IDE seems to be configured to remove newlines at the end of files, could you make it have newlines instead? Just creates noise here

Menduist (dismissed):
Only minor tweaks left, overall LGTM!

libp2p/services/autonatservice.nim
diegomrsantos: Good catch. It introduces some repetition, but it is probably better.

tests/helpers.nim
diegomrsantos: I think I deleted it by accident

libp2p/connmanager.nim
```diff
@@ -110,6 +110,13 @@ proc new*(C: type ConnManager,
 proc connCount*(c: ConnManager, peerId: PeerId): int =
   c.conns.getOrDefault(peerId).len
 
+proc connectedPeers*(c: ConnManager, dir: Direction): seq[PeerId] =
+  var peers = newSeq[PeerId]();
```
Menduist, line 114:
```suggestion
  var peers = newSeq[PeerId]()
```

Menduist (approved):

PR #817, opened 2022-12-08
Allow connect with known PeerId in MA

### LINKED ISSUE #774: allow connecting to addresses with p2p id
When passing an address like `/ip4/192.168.1.106/tcp/10634/p2p/16Uiu2HAmTaEoEpEFkrCKfU5srxYrgkgBz4e2LJDhdod4VwgH2DkC` to `connect`, it should succeed to connect and verify that the peer id matches the given on.

Instead, the address gets dropped here because of the full match requirement:

https://github.com/status-im/nim-libp2p/blob/1de7508b64e29c2bcad7a393c36cf8390a8c50fe/libp2p/transports/tcptransport.nim#L261

Menduist:
closes #774

### DIFF
--- a/libp2p/dial.nim
+++ b/libp2p/dial.nim
@@ -36,7 +36,8 @@ method connect*(
 
 method connect*(
   self: Dial,
-  addrs: seq[MultiAddress]): Future[PeerId] {.async, base.} =
+  addrs: seq[MultiAddress],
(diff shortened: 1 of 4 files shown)


Menduist:
Last commit switches the `connect` to only allow a single MA, otherwise weird stuff can happen when mixing MA with/without PeerId or MAs with different PeerIds

lchenut (approved):

PR #819, opened 2022-12-13
Autorelay service

lchenut:
A service to send and manage reservations with relay peers.
It misses some more tests and some iterations over the Service object but it works.


libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,110 @@
...
+import ../../switch,
+       relay/[client, utils]
+
+logScope:
+  topics = "libp2p autorelay"
+
+type
+  OnReservationHandler = proc (ma: MultiAddress): Future[void] {.gcsafe, raises: [Defect].}
+
+  AutoRelay* = ref object of Service
```
diegomrsantos, line 25: Should we name it AutoRelayService?

libp2p/switch.nim
```diff
@@ -108,6 +127,12 @@ method addTransport*(s: Switch, t: Transport) =
   s.transports &= t
   s.dialer.addTransport(t)
 
+method addService*(switch: Switch, service: Service) =
```
diegomrsantos, line 130: I removed this in my pr and added a `withServices` in the builder

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,110 @@
...
+    let
+      rsvp = await self.client.reserve(relayPid).wait(chronos.seconds(5))
+      relayedAddr = MultiAddress.init($(rsvp.addrs[0]) &
+                                  "/p2p-circuit/p2p/" &
+                                  $selfPid).tryGet()
+    await self.onReservation(relayedAddr)
+    await sleepAsync chronos.seconds(rsvp.expire.int64 - times.now().utc.toTime.toUnix)
+
+method setup*(self: AutoRelay, switch: Switch) {.async, gcsafe.} =
+  if self.inUse:
```
diegomrsantos, line 45: I did this in the autonatservice:

```
  let hasBeenSettedUp = await procCall Service(self).setup(switch)
  if hasBeenSettedUp:
    self.newConnectedPeerHandler = proc (peerId: PeerId, event: PeerEvent): Future[void] {.async.} =
      discard askPeer(self, switch, peerId)
    switch.connManager.addPeerEventHandler(self.newConnectedPeerHandler, PeerEventKind.Joined)
    if self.scheduleInterval.isSome():
      self.registerLoop = register(self, switch, self.scheduleInterval.get())
  return hasBeenSettedUp
```

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,110 @@
...
+    onReservation: OnReservationHandler
+
+proc reserveAndUpdate(self: AutoRelay, relayPid: PeerId, selfPid: PeerId) {.async.} =
+  while self.running:
+    let
+      rsvp = await self.client.reserve(relayPid).wait(chronos.seconds(5))
+      relayedAddr = MultiAddress.init($(rsvp.addrs[0]) &
+                                  "/p2p-circuit/p2p/" &
+                                  $selfPid).tryGet()
+    await self.onReservation(relayedAddr)
```
diegomrsantos, line 41: can onReservation be nil?

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,110 @@
...
+
+type
+  OnReservationHandler = proc (ma: MultiAddress): Future[void] {.gcsafe, raises: [Defect].}
+
+  AutoRelay* = ref object of Service
+    running: bool
+    runner: Future[void]
+    client: RelayClient
+    npeers: int
+    peers: Table[PeerId, Future[void]]
```
diegomrsantos, line 30: where is it initialized?

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,110 @@
...
+
+proc reserveAndUpdate(self: AutoRelay, relayPid: PeerId, selfPid: PeerId) {.async.} =
+  while self.running:
+    let
+      rsvp = await self.client.reserve(relayPid).wait(chronos.seconds(5))
+      relayedAddr = MultiAddress.init($(rsvp.addrs[0]) &
+                                  "/p2p-circuit/p2p/" &
+                                  $selfPid).tryGet()
+    await self.onReservation(relayedAddr)
+    await sleepAsync chronos.seconds(rsvp.expire.int64 - times.now().utc.toTime.toUnix)
```
diegomrsantos, line 42: what would we do if we need to mock `now()` in tests?

libp2p/protocols/connectivity/autorelay.nim
diegomrsantos, line 30: Can we have a more descriptive name here? Maybe `relayPeers` if I understood it correctly.

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,110 @@
...
+
+method setup*(self: AutoRelay, switch: Switch) {.async, gcsafe.} =
+  if self.inUse:
+    warn "Autorelay setup has already been called"
+    return
+  self.inUse = true
+  proc handlePeer(peerId: PeerId, event: PeerEvent) {.async.} =
+    if event.kind == Left and peerId in self.peers:
```
diegomrsantos, line 50: What does `event.kind == Left` mean?

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,110 @@
...
+  topics = "libp2p autorelay"
+
+type
+  OnReservationHandler = proc (ma: MultiAddress): Future[void] {.gcsafe, raises: [Defect].}
+
+  AutoRelay* = ref object of Service
+    running: bool
+    runner: Future[void]
+    client: RelayClient
+    npeers: int
```
diegomrsantos, line 29: maybe `numRelays`?

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,110 @@
...
+                                  $selfPid).tryGet()
+    await self.onReservation(relayedAddr)
+    await sleepAsync chronos.seconds(rsvp.expire.int64 - times.now().utc.toTime.toUnix)
+
+method setup*(self: AutoRelay, switch: Switch) {.async, gcsafe.} =
+  if self.inUse:
+    warn "Autorelay setup has already been called"
+    return
+  self.inUse = true
+  proc handlePeer(peerId: PeerId, event: PeerEvent) {.async.} =
```
diegomrsantos, line 49: How about?
  
```
  proc handlePeerJoined(peerId: PeerId, event: PeerEvent) {.async.} =
    if self.peers.len < self.npeers:
      self.peerJoined.fire()  

proc handlePeerLeft(peerId: PeerId, event: PeerEvent) {.async.} =
    if peerId in self.peers:
      self.peers[peerId].cancel()
    elif event.kind == Joined and self.peers.len < self.npeers:
      self.peerJoined.fire()
  switch.addPeerEventHandler(handlePeerJoined, Joined)
  switch.addPeerEventHandler(handlePeerLeft, Left)
```

libp2p/protocols/connectivity/autorelay.nim
lchenut: You're right, it's clearer

libp2p/protocols/connectivity/autorelay.nim
lchenut: Same here, it's clearer

libp2p/protocols/connectivity/autorelay.nim
lchenut: It doesn't need to be, you can do 
```nim
var x: Table[int, string]
x[12] = "twelve"
```

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,110 @@
...
+    client: RelayClient
+    npeers: int
+    peers: Table[PeerId, Future[void]]
+    peerJoined: AsyncEvent
+    onReservation: OnReservationHandler
+
+proc reserveAndUpdate(self: AutoRelay, relayPid: PeerId, selfPid: PeerId) {.async.} =
+  while self.running:
+    let
+      rsvp = await self.client.reserve(relayPid).wait(chronos.seconds(5))
```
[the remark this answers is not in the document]
lchenut: We aren't handling ReservationError either, but the purpose of reserveAndUpdate is to run indefinitely or to crash.
I check if it fail [here](https://github.com/status-im/nim-libp2p/blob/autorelay/libp2p/protocols/connectivity/autorelay.nim#L62).

libp2p/protocols/connectivity/autorelay.nim
lchenut: I thought about it but I'm not 100% sure on how to do it.
Maybe something like `method getAddresses(self: AutoRelay): seq[MultiAddress]` could be a good alternative in case there's no callback.

diegomrsantos: I think we can check if it is `nil` before calling it, cause the service can be created like that `AutoRelay.new(3, client, nil)`

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,110 @@
...
+  switch.addPeerEventHandler(handlePeer, Joined)
+  switch.addPeerEventHandler(handlePeer, Left)
+
+method innerRun(self: AutoRelay, switch: Switch) {.async, gcsafe.} =
+  while true:
+    # Remove peers that failed
+    var peersToRemove: seq[PeerId]
+    for k, v in self.peers:
+      if v.finished():
+        peersToRemove.add(k)
```
diegomrsantos, line 63: can we replace this line by `self.peers.del(k)` and get rid of `peersToRemove`?

libp2p/protocols/connectivity/autorelay.nim
Menduist: If there is no other way to get the addresses, giving `nil` should be invalid and crash
If there is another way, we need to protect here indeed

libp2p/protocols/connectivity/autorelay.nim
lchenut: No it doesn't work, it raises `the length of the table changed while iterating over it [AssertionDefect]`.
And as there's no removeIf in std/tables, I had to do something inelegant.

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,117 @@
...
+  while self.running:
+    let
+      rsvp = await self.client.reserve(relayPid).wait(chronos.seconds(5))
+      relayedAddr = MultiAddress.init($(rsvp.addrs[0]) &
+                                  "/p2p-circuit/p2p/" &
+                                  $selfPid).tryGet()
+    if not self.onReservation.isNil():
+      await self.onReservation(relayedAddr)
+    self.relayAddresses[relayPid] = relayedAddr
+    await sleepAsync chronos.seconds(rsvp.expire.int64 - times.now().utc.toTime.toUnix)
```
Menduist, line 45: Subtract some margin here, need to take into account random delays

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,117 @@
...
+  self.runner = self.innerRun(switch)
+
+method stop*(self: AutoRelayService, switch: Switch): Future[bool] {.async, gcsafe.} =
+  let hasBeenStopped = await procCall Service(self).stop(switch)
+  if hasBeenStopped:
+    self.running = false
+    self.runner.cancel()
+  return hasBeenStopped
+
+method getAddresses*(self: AutoRelayService): seq[MultiAddress] =
```
Menduist, line 107: Why method?

libp2p/protocols/connectivity/autorelay.nim
lchenut: Done!

libp2p/protocols/connectivity/autorelay.nim
lchenut: I added a `proc getAddresses*(self: AutoRelayService): seq[MultiAddress]` so it's no longer the only way... So it's protected now.

libp2p/protocols/connectivity/autorelay.nim
lchenut: Done!

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,131 @@
...
+      if self.relayPeers[k].finished():
+        self.relayPeers.del(k)
+        self.relayAddresses.del(k)
+        if not self.onReservation.isNil():
+          await self.onReservation(toSeq(self.relayAddresses.values))
+        # To avoid ddosing our peers in certain conditions
+        self.backedOff.add(k)
+        asyncSpawn self.manageBackedOff(k)
+
+    # Get all connected relayPeers
```
diegomrsantos, line 87: How about a new proc with the same name as in the comment?

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,131 @@
...
+    var connectedPeers = switch.connectedPeers(Direction.Out)
+    connectedPeers.keepItIf(RelayV2HopCodec in switch.peerStore[ProtoBook][it] or
+                            it notin self.relayPeers)
```
diegomrsantos, line 90: How about a new proc with the same name as in the comment?

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,131 @@
...
+    let peers = toSeq(self.relayPeers.keys())
+    for k in peers:
+      if self.relayPeers[k].finished():
+        self.relayPeers.del(k)
+        self.relayAddresses.del(k)
+        if not self.onReservation.isNil():
+          await self.onReservation(toSeq(self.relayAddresses.values))
+        # To avoid ddosing our peers in certain conditions
+        self.backedOff.add(k)
+        asyncSpawn self.manageBackedOff(k)
```
diegomrsantos, line 85: How about a new proc with the same name as in the comment?

libp2p/protocols/connectivity/autorelay.nim
```diff
@@ -0,0 +1,131 @@
...
+      if self.relayPeers.len() >= self.numRelays:
+        break
+      self.relayPeers[relayPid] = self.reserveAndUpdate(relayPid, switch.peerInfo.peerId)
+    let peersFutures = toSeq(self.relayPeers.values())
+
+    if self.relayPeers.len() < self.numRelays:
+      self.peerAvailable.clear()
+      await one(peersFutures) or self.peerAvailable.wait()
+    else:
+      discard await one(peersFutures)
```
diegomrsantos, line 103:
```suggestion
    proc getNewConnectedRelayPeers(): seq[PeerId] =
      var connectedPeers = switch.connectedPeers(Direction.Out)
      connectedPeers.keepItIf(RelayV2HopCodec in switch.peerStore[ProtoBook][it] or
                              it notin self.relayPeers)
      return connectedPeers

    if self.relayPeers.len() < self.numRelays:
      var connectedPeers = getNewConnectedRelayPeers()
      self.rng.shuffle(connectedPeers)

      let numMissingRelayPeers = self.numRelays - self.relayPeers.len()
      for idx in 0..<min(connectedPeers.len, numMissingRelayPeers):
        let relayPid = connectedPeers[idx]
        self.relayPeers[relayPid] = self.reserveAndUpdate(relayPid, switch.peerInfo.peerId)

      if self.relayPeers.len() < self.numRelays:
            self.peerAvailable.clear()
            await one(toSeq(self.relayPeers.values())) or self.peerAvailable.wait()

    if self.relayPeers.len() == self.numRelays:
      discard await one(toSeq(self.relayPeers.values()))
```

How about sth like this? `getNewConnectedRelayPeers` could be moved outside.

libp2p/services/autorelayservice.nim
```diff
@@ -0,0 +1,138 @@
...
+  while self.running:
+    let
+      rsvp = await self.client.reserve(relayPid).wait(chronos.seconds(5))
+      relayedAddr = MultiAddress.init($(rsvp.addrs[0]) &
+                                  "/p2p-circuit/p2p/" &
+                                  $selfPid).tryGet()
+      ttl = rsvp.expire.int64 - times.now().utc.toTime.toUnix
+    if ttl <= 60:
+      # A reservation under a minute is basically useless
+      break
```
diegomrsantos, line 49: Do we really need this?

lchenut: First of all, a very short reservation is really useless, and secondly, if we accept this type of reservation, when we subtract a certain margin in order to avoid having a period without a reservation, there's a possibility where you simply send reservations without any sleep time. So with that in mind  and without any spec whatsoever, I thought it was better to put this limit

libp2p/services/autorelayservice.nim
```diff
@@ -0,0 +1,138 @@
...
+        self.relayAddresses.del(k)
+        if not self.onReservation.isNil():
+          self.onReservation(toSeq(self.relayAddresses.values))
+        # To avoid ddosing our peers in certain conditions
+        self.backingOff.add(k)
+        asyncSpawn self.manageBackedOff(k)
+
+    # Get all connected relayPeers
+    var connectedPeers = switch.connectedPeers(Direction.Out)
+    connectedPeers.keepItIf(RelayV2HopCodec in switch.peerStore[ProtoBook][it] or
```
Menduist, line 93: and?

libp2p/services/autorelayservice.nim
```diff
@@ -0,0 +1,138 @@
...
+                            it notin self.relayPeers or
+                            it notin self.backingOff)
+    self.rng.shuffle(connectedPeers)
+
+    for relayPid in connectedPeers:
+      if self.relayPeers.len() >= self.numRelays:
+        break
+      self.relayPeers[relayPid] = self.reserveAndUpdate(relayPid, switch.peerInfo.peerId)
+    let peersFutures = toSeq(self.relayPeers.values())
```
Menduist, line 102: Unused?

libp2p/services/autorelayservice.nim
```diff
@@ -0,0 +1,137 @@
...
+                            it notin self.backingOff)
+    self.rng.shuffle(connectedPeers)
+
+    for relayPid in connectedPeers:
+      if self.relayPeers.len() >= self.numRelays:
+        break
+      self.relayPeers[relayPid] = self.reserveAndUpdate(relayPid, switch.peerInfo.peerId)
+
+    self.reservationFailed.clear()
+    self.peerAvailable.clear()
```
Menduist, line 104:
```suggestion

    self.reservationFailed.clear()
    self.peerAvailable.clear()

    # Get all connected relayPeers
    var connectedPeers = switch.connectedPeers(Direction.Out)
    connectedPeers.keepItIf(RelayV2HopCodec in switch.peerStore[ProtoBook][it] and
                            it notin self.relayPeers and
                            it notin self.backingOff)
    self.rng.shuffle(connectedPeers)

    for relayPid in connectedPeers:
      if self.relayPeers.len() >= self.numRelays:
        break
      self.relayPeers[relayPid] = self.reserveAndUpdate(relayPid, switch.peerInfo.peerId)

```

libp2p/services/autorelayservice.nim
```diff
@@ -0,0 +1,137 @@
...
+    if self.relayPeers.len() < self.numRelays:
+      await self.reservationFailed.wait() or self.peerAvailable.wait()
```
Menduist, line 106: I'm not a big fan of this reservationFailed, seems like a nest of synchronization bugs

Would something like this keep everyone happy?:
```nim
    let reservationFailed = anyFuture(self.relayPeers.values)
    if self.relayPeers.len() < self.numRelays:
      await reservationFailed or self.peerAvailable.wait()
    else:
      await reservationFailed
```

libp2p/services/autorelayservice.nim
```diff
@@ -0,0 +1,137 @@
...
+                                  $selfPid).tryGet()
+      ttl = rsvp.expire.int64 - times.now().utc.toTime.toUnix
+    if ttl <= 60:
+      # A reservation under a minute is basically useless
+      break
+    if relayPid notin self.relayAddresses or self.relayAddresses[relayPid] != relayedAddr:
+      self.relayAddresses[relayPid] = relayedAddr
+      if not self.onReservation.isNil():
+        self.onReservation(toSeq(self.relayAddresses.values))
+    await sleepAsync chronos.seconds(max(0, ttl - 30))
```
Menduist, line 54:
```suggestion
    await sleepAsync chronos.seconds(ttl - 30)
```

libp2p/services/autorelayservice.nim
lchenut: First of all, this change is correct, that's not why I unresolved it, but it raises some concerns according the addPeerEventHandler:
Right now, the peer event handler is called before the Identify. And it causes one problem: the peerStore isn't fill with anything. So RelayV2HopCodec will never be in the `peerStore[ProtoBook]`. I tried to make it work by calling the handlers after the Identify, but because the handlers are asyncspawn, it causes some race conditions.

I probably didn't think of all the different possibilities to fix this but there's two possible ways:
- await the handlers instead of asyncspawning them (it might still causes some race condition though, I didn't test it heavily)
- make the handler synchronous (but that'll broke a lot of thing in the relayv2 or the autonat for example)

Wdyt?

libp2p/services/autorelayservice.nim
```diff
@@ -0,0 +1,134 @@
...
+    backingOff: seq[PeerId]
+    peerAvailable: AsyncEvent
+    onReservation: OnReservationHandler
+    rng: ref HmacDrbgContext
+
+proc reserveAndUpdate(self: AutoRelayService, relayPid: PeerId, selfPid: PeerId) {.async.} =
+  while self.running:
+    let
+      rsvp = await self.client.reserve(relayPid).wait(chronos.seconds(5))
+      relayedAddr = MultiAddress.init($(rsvp.addrs[0]) &
```
Menduist, line 41: Actually, we need to do this for each public address of the relay..

Menduist (approved):

PR #825, opened 2022-12-21
Add autonat service example

alrevuelta:
Add autonat service example

### DIFF
--- a/examples/tutorial_7_autonatservice.nim
+++ b/examples/tutorial_7_autonatservice.nim
@@ -1,0 +1,100 @@
+# Autonat Example
+
+# This example shows how to use the autonat service. This service allows a given
+# peer to constantly request other peers to check if it is reachable or not. It can
+# be configured with the following parameters:
+# -scheduleInterval: how often to check if the peer is reachable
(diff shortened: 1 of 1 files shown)


Menduist (commented):
I don't think this should be a "tutorial", you can rename it autonatservice.nim

You should also add your file here:
https://github.com/status-im/nim-libp2p/blob/63e1872516cdbd2c843e5297aa36616a531ac2cd/libp2p.nimble#L100

to make it available on the website, here:
https://github.com/status-im/nim-libp2p/blob/63e1872516cdbd2c843e5297aa36616a531ac2cd/libp2p.nimble#L108
to be built as part of CI, and here:

https://github.com/status-im/nim-libp2p/blob/63e1872516cdbd2c843e5297aa36616a531ac2cd/mkdocs.yml#L51

To be in the navbar of the website

alrevuelta:
Actually now that we have [this test](https://github.com/status-im/nim-libp2p/blob/unstable/tests/testautonatservice.nim#L153-L189) with the real autonat service (not the stub like before) this example is not that relevant. Closing.

PR #829, opened 2022-12-22
Add Autonat timeout

diegomrsantos:


Menduist (dismissed):

libp2p/services/autonatservice.nim
```diff
@@ -52,7 +52,7 @@ proc new*(
   numPeersToAsk: int = 5,
   maxQueueSize: int = 10,
   minConfidence: float = 0.3,
-  dialTimeout = 5.seconds): T =
+  dialTimeout = 30.seconds): T =
```
alrevuelta, line 55: as I understand it, the autonat timeout has to be lower than the autonatservice timeout, right?

should we add a check to enforce this? otherwise we risk that someone missconfigure it?

diegomrsantos: I'm not sure. Wdyt @Menduist?

Menduist: We can check against the local value, but not against remote values, even less values of other implementations

Menduist (approved):

PR #831, opened 2022-12-23
Ignore unknown answers

diegomrsantos:


libp2p/services/autonatservice.nim
```diff
@@ -75,9 +75,11 @@ proc callHandler(self: AutonatService) {.async.} =
 
 proc handleAnswer(self: AutonatService, ans: NetworkReachability) {.async.} =
 
-  if self.answers.len == self.maxQueueSize:
-    self.answers.popFirst()
+  if ans == Unknown:
+    return
 
+  if self.answers.len == self.maxQueueSize:
+      self.answers.popFirst()
```
Menduist, line 82:
```suggestion
    self.answers.popFirst()
```

Menduist (approved):

alrevuelta (approved):
lgtm! thanks!

We can leave it for another PR, but I think it would be nice to have a couple of simple tests wihtout the autonat stub, so using a real autonat. I would suggest just two tests:
* 3 peers, all of them with autonat. peer1 connects to peer2/3. We can see its reachable with 100% confidence.
* 3 peers,  where peer2 supports autonat but peer3 does not. Here we should see that node1 is also labeled as Reachable with 100 confidence, since peer3 responses are dicarded due to not supporting autonat.

imho the stub is only useful for scenarios that we can't reproduce, like NotReachable. For the rest, the closer we get to a real scenario, the better.

diegomrsantos:
> lgtm! thanks!
> 
> We can leave it for another PR, but I think it would be nice to have a couple of simple tests wihtout the autonat stub, so using a real autonat. I would suggest just two tests:
> 
> * 3 peers, all of them with autonat. peer1 connects to peer2/3. We can see its reachable with 100% confidence.
> * 3 peers,  where peer2 supports autonat but peer3 does not. Here we should see that node1 is also labeled as Reachable with 100 confidence, since peer3 responses are dicarded due to not supporting autonat.
> 
> imho the stub is only useful for scenarios that we can't reproduce, like NotReachable. For the rest, the closer we get to a real scenario, the better.

I agree. That's how I started testing, but it was harder to know when to finish the test. I used `sleepAsync` which worked on Linux and Mac, but didn't with Nim 1.2 on Windows.  Using `sleepAsync` doesn't seem like a reliable way to do that anyway. It was easier to solve it with the stub, but there might be another solution too.

Menduist (approved):

Menduist:
Can't you wait for the callback to be called?

diegomrsantos:
> Can't you wait for the callback to be called?

Probably yes, a lot of refactoring happened since then.

PR #832, opened 2022-12-23
move store muxer

lchenut:
The "peer callback" will now get called _after_ identify
This leads to a slight delay, but makes sure that the upgrade is completely finished before going to user code.
We could add an "early callback" in the future if we want to let user get a partially upgraded connection


Menduist (dismissed):
If CI passes, let's merge this to see how it behaves in the real world

lchenut:
Closed due to redundancy with https://github.com/status-im/nim-libp2p/pull/807

PR #834, opened 2023-01-03
Autonat refactoring

diegomrsantos:
The initial motivation was to fix the circular dependency between switch and autonat.


lchenut (approved):
LGTM
This makes me think that I might do something similar for the relay.

PR #835, opened 2023-01-03
Fix nim devel compilation

Menduist:


Menduist:
Fixed by https://github.com/status-im/nimbus-build-system/pull/55

PR #837, opened 2023-01-05
Refactor dialer and upgrade

diegomrsantos:
Splits `dialAndUpgrade` into smaller procs and hopefully makes it easier to understand.

### DIFF
--- a/libp2p/dialer.nim
+++ b/libp2p/dialer.nim
@@ -47,52 +47,87 @@ type
     transports: seq[Transport]
     nameResolver: NameResolver
 
-proc dialAndUpgrade(
+proc tryDial(
+  peerId: Opt[PeerId],
+  hostname: string,
+  address: MultiAddress,
+  transport: Transport):
+  Future[Connection] {.async.} =
+  trace "Dialing address", address, peerId, hostname
+  result =
+    try:
+      libp2p_total_dial_attempts.inc()
+      await transport.dial(hostname, address)
+    except CancelledError as exc:
+      debug "Dialing canceled", msg = exc.msg, peerId
+      raise exc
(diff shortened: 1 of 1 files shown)


Menduist (changes requested):
I find this even more confusing, too many procedures, and some with the same name (I know it's already the case today, but this PR makes it worst IMO)

We need to create a better terminology to distinguish, dial, connect, upgrade, etc

Maybe we say that:
connect = dial + upgrade
(then "Dialer" is not a proper name anymore)
or
connect = dial = "open connection" + upgrade
idk

Then maybe we have:
- tryDial & tryUpgrade (merged in a single proc) -> dialAndUpgrade / dialOneAddress
- tryDialAndUpgrade (line 119, 106 & 87 merged in a single proc) -> dialAddresses
- dialAndUpgrade -> dialDnsAddresses

diegomrsantos:
> I find this even more confusing, too many procedures, and some with the same name (I know it's already the case today, but this PR makes it worst IMO)
> 
> We need to create a better terminology to distinguish, dial, connect, upgrade, etc
> 
> Maybe we say that: connect = dial + upgrade (then "Dialer" is not a proper name anymore) or connect = dial = "open connection" + upgrade idk
> 
> Then maybe we have:
> 
> * tryDial & tryUpgrade (merged in a single proc) -> dialAndUpgrade / dialOneAddress
> * tryDialAndUpgrade (line 119, 106 & 87 merged in a single proc) -> dialAddresses
> * dialAndUpgrade -> dialDnsAddresses

I'm going to close this PR. I still think the original code is hard to understand, but I realized the code in this PR is not equivalent to the original one and if we want them to be, the refactoring doesn't add much value after all.

PR #838, opened 2023-01-06
Using nim-protobuf-serialization for the RendezVous protocol

lchenut:


Menduist:
Even without enum support, that looks miles better! Thanks @arnetheduck

.pinned
```diff
@@ -1,15 +1,17 @@
 bearssl;https://github.com/status-im/nim-bearssl@#a647994910904b0103a05db3a5ec1ecfc4d91a88
 chronicles;https://github.com/status-im/nim-chronicles@#32ac8679680ea699f7dbc046e8e0131cac97d41a
-chronos;https://github.com/status-im/nim-chronos@#75d030ff71264513fb9701c75a326cd36fcb4692
+chronos;https://github.com/status-im/nim-chronos@#e9f8baa6ee2e21ff8e6b6c0ce0c22368cdd9e758
+combparser;https://github.com/PMunch/combparser@#ba4464c005d7617c008e2ed2ebc1ba52feb469c6
```
arnetheduck, line 4: not needed

libp2p/protocols/rendezvous.nim
```diff
@@ -40,244 +41,58 @@ const
...
-  if r5.get():
-    msg.discover = Discover.decode(pbd.buffer)
-    if msg.discover.isNone(): return none(Message)
-  if r6.get():
-    msg.discoverResponse = DiscoverResponse.decode(pbdr.buffer)
-    if msg.discoverResponse.isNone(): return none(Message)
-  some(msg)
-
+const
+  RdvRegister = 0'u64
```
arnetheduck, line 45: u32 for enums

libp2p/protocols/rendezvous.nim
```diff
@@ -312,39 +127,39 @@ proc checkPeerRecord(spr: seq[byte], peerId: PeerId): Result[void, string] =
...
 
 proc sendRegisterResponseError(conn: Connection,
-                               status: ResponseStatus,
+                               status: uint64,
                                text: string = "") {.async.} =
-  let msg = encode(Message(
-        msgType: MessageType.RegisterResponse,
-        registerResponse: some(RegisterResponse(status: status, text: some(text)))))
-  await conn.writeLp(msg.buffer)
+  let msg = Protobuf.encode(Message(
```
arnetheduck, line 138: cc @Menduist this is where you'd use a streaming writer instead - ie each of these functions is pointlessly inefficient

Menduist: So you would imagine a `Protobuf.streamEncode` macro that takes
```nim
Protobuf.streamEncode(Message(
  msgType: RdvRegisterResponse,
  registerResponse: RegisterResponse(status: status, text: text))
)
# and transforms into
block:
  var pb = initProtoBuf()
  pb.writeField(1, RdvRegisterResponse)
  pb.writeField(2, Protobuf.streamEncode(RegisterResponse(status: status, text: text))
  pb.finish()
  pb.buffer
```
? That shouldn't be that hard to do

arnetheduck: Well, or some variations on that theme:
```nim
Protobuf.encode(Message, encoder):
  encoder.write(fieldname1, value)
  encoder.write(somesubobject, subencoder):
    subencoder.write(subfield, value)
```

```nim
Protobuf.encode(Message, encoder)
    .field1(value).
    .field2(subobject):
       suboject
          .subField(value)
    .field3(value)
```

not sure, haven't really looked into it, though the obvious point is to grab the field number and encoding from the object definition and for that you need the type and the field name - the rest is really a syntax game.

There are some caveats though - streaming inherently becomes more "freeform" in that you can repeat the same field multiple times (this is valid in pb even for non-repeated) and there are some quirks with packed fields (which are length-prefixed and therefore would need an intermediate store - either the fancy but buggy faststreams buffers or something equivalent.

arnetheduck: When I look at your example, maybe transforming the object constructor isn't that bad :) That said, similar caveats apply: for example, to correctly encode a nested object you need to evaluate it twice: once to know if it has any non-default values and the second time to actually write the bytes - do you then evaluate the "source" twice or introduce a temporary buffer? tradeoffs, tradeoffs...

.pinned
lchenut: @arnetheduck I am reopening this comment as it appears that combparser is indeed needed.
It's used by protobuf-ser [here](https://github.com/status-im/nim-protobuf-serialization/blob/master/protobuf_serialization/files/proto_parser.nim) and is therefore installed / needed by the nim-libp2p. I don't know if it's an oversight on your part or if you eventually want to remove the use of combparser.
Or maybe I'm just missing something.

arnetheduck: as long as you don't use `import_proto3` it's not needed - ie that file should not be imported by default, per the nps readme.

Menduist: The pinned file is generated from the nimble dependency, so it will still be bringed in, as long as nimble doesn't support optional deps

PR #843, opened 2023-01-17
feat(peerEvents): add a peerEvent Identified

lchenut:
This PR was originally creates to remove a temporary sleepAsync which was a temporary patch against races conditions which were, in theory, fixed by #807.
As #807 didn't solved the problem (explained in details in [this comment](https://github.com/vacp2p/nim-libp2p/pull/843#issuecomment-2165659939)), the goal of this current PR changed to add a peerEvent callback in order to safely remove the sleepAsync.

As a side note, I also remove some duplicate code. Services were indeed being stopped twice when a switch stopped generating warnings each time this happened.


lchenut:
After a bit of research on this PR, here's why it exists:
This PR removes `await sleepAsync(200.millis)` after `await self.peerAvailable.wait()`. The problem is that removing it creates a race condition. The asyncEvent `peerAvailable` is fired when a new peer joins. Once it's fired the loop (in the proc innerRun) restarts from the beginning until these two lines of code:
```nim
    var connectedPeers = switch.connectedPeers(Direction.Out)    
    connectedPeers.keepItIf(RelayV2HopCodec in switch.peerStore[ProtoBook][it] and    
```
We get the newly connected peers and only keeping it if RelayV2HopCodec is in the `switch.peerStore`. The issue here is that the peerStore may not be up to date at this point, hence the sleepAsync(200.millis).

The way it's currently handled is certainly not a perfect solution, but if we want to remove this sleep, we either have to change when the peerStore is filled or when the `handlePeerJoined` callback is called. Either way, after looking at it, it'll change the call order of different asynchronous procedure and could cause other race conditions if done badly.

With that in mind, I'm not sure what to do about this PR... What do you think @diegomrsantos @kaiserd?

libp2p/connmanager.nim
```diff
@@ -48,10 +48,11 @@ type
   PeerEventKind* {.pure.} = enum
     Left
     Joined
+    Identified
 
   PeerEvent* = object
     case kind*: PeerEventKind
-    of PeerEventKind.Joined:
+    of PeerEventKind.Joined, PeerEventKind.Identified:
```
diegomrsantos, line 55: why do we need this?

libp2p/dialer.nim
```diff
@@ -195,6 +195,10 @@ proc internalConnect(
     try:
       self.connManager.storeMuxer(muxed)
       await self.peerStore.identify(muxed)
+      asyncSpawn self.connManager.triggerPeerEvents(
```
diegomrsantos, line 198: why `asyncSpawn`?

libp2p/services/autorelayservice.nim
```diff
@@ -66,7 +66,7 @@ method setup*(self: AutoRelayService, switch: Switch): Future[bool] {.async.} =
 
   let hasBeenSetUp = await procCall Service(self).setup(switch)
   if hasBeenSetUp:
-    proc handlePeerJoined(peerId: PeerId, event: PeerEvent) {.async.} =
+    proc handlePeerIdentified(peerId: PeerId, event: PeerEvent) {.async.} =
       trace "Peer Joined", peerId
```
diegomrsantos, line 70: should we update this?

libp2p/switch.nim
```diff
@@ -204,6 +204,9 @@ proc upgrader(switch: Switch, trans: Transport, conn: Connection) {.async.} =
   let muxed = await trans.upgrade(conn, Opt.none(PeerId))
   switch.connManager.storeMuxer(muxed)
   await switch.peerStore.identify(muxed)
+  asyncSpawn switch.connManager.triggerPeerEvents(
```
diegomrsantos, line 207: why `asyncSpawn`?

libp2p/switch.nim
```diff
@@ -295,9 +298,6 @@ proc stop*(s: Switch) {.async, public.} =
     except CatchableError as exc:
       warn "error cleaning up transports", msg = exc.msg
 
-  for service in s.services:
```
diegomrsantos, line 298: Could you please mention this in the description?

libp2p/dialer.nim
lchenut: Because we do not know what's the callback will be. It could possibly be an infinite loop for example. AsyncSpawn is safer to avoid this kind of behavior.

libp2p/services/autorelayservice.nim
lchenut: I'm not sure what do you mean?

libp2p/switch.nim
lchenut: Same as above.

libp2p/switch.nim
lchenut: done

libp2p/connmanager.nim
diegomrsantos: It might be fine as this case is minor, but it's better to add functionality only when needed.

libp2p/dialer.nim
diegomrsantos: Is it? Running an infinite loop in the background doesn't sound like a good alternative. How about we use `await` with a timeout?

lchenut: What I mean by that is that the peerEvent is public, we do not control what libp2p user could do here.
Another point to consider is that this `triggerPeerEvent(Identified)` is called in important part of the code. Let's take a case where the callback do a lot of computation (let's say 200ms), every time we dial a peer or we accept a connection, 200 ms is spent waiting for the callback to finish before anything else can be done.
And last point, if you look at how the other events are called, they also use asyncSpawn (https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/connmanager.nim#L310-L317)

diegomrsantos: I'm still not convinced. AFAIK, chronos is single-threaded and once this task is executed it won't yield control until it stops running. It's not executed in parallel.

lchenut: In the case of the scenario of 200 milliseconds it'll slow down every dial by this amount. I think that it's really preferable to give the newly created connection to the user directly and await for these after the dial (or accept) is completed.

lchenut: After a private conversation, it's indeed better to not use asyncSpawn here :+1: 
It's now change to `await` instead

diegomrsantos (approved):

PR #844, opened 2023-01-19
Test CI

yyoncho:


.github/workflows/ci-latest-packages.yml
```diff
@@ -0,0 +1,79 @@
+name: latest version of packages
```
Menduist, line 1: Currently, the daily CI is using the latest version of packages, and the regular CI is using locked version
I think we can keep the same logic here

.github/workflows/ci.yml
```diff
@@ -73,20 +73,6 @@ jobs:
         run: |
           V=1 bash scripts/build_p2pd.sh p2pdCache 124530a3
 
-      - name: Restore deps from cache
```
Menduist, line 76: Can't we cache the deps in the same way?

.github/workflows/codecov.yml
```diff
@@ -30,27 +30,19 @@ jobs:
...
           sudo apt-get update
           sudo apt-get install -y lcov build-essential git curl
           mkdir coverage
           export NIMFLAGS="--lineDir:on --passC:-fprofile-arcs --passC:-ftest-coverage --passL:-fprofile-arcs --passL:-ftest-coverage"
-          nimble testnative
-          nimble testpubsub
-          nimble testfilter
+          nimble testnative --lock-file=nimble-version-1-6.lock
+          nimble testpubsub --lock-file=nimble-version-1-6.lock
+          nimble testfilter --lock-file=nimble-version-1-6.lock
```
Menduist, line 45: Does that mean that everytime we want to run tests we'll have to do
```
nimble testnative --lock-file=nimble-version-1-6.lock
# instead of just
nimble testnative
```
?

libp2p.nimble
```diff
@@ -19,11 +19,15 @@ requires "nim >= 1.2.0",
          "websock",
          "unittest2 >= 0.0.5 & < 0.1.0"
 
+if (NimMajor, NimMinor) > (1, 2):
+  requires "nimpng#HEAD", "nico"
```
Menduist, line 23: That is not correct, these packages are just needed for the tests, they are not a dependency of libp2p itself

libp2p.nimble
```diff
@@ -109,52 +99,23 @@ task examples_build, "Build the samples":
...
-          fileName.endsWith(it[1].split('#')[^1]) # nimble for nim 1.X
-        )
-      ) == false or
-      fileName.split('-')[^1].len < 20: # safegard for nimble for nim 1.X
-        rmDir(dependency)
-
-task unpin, "Restore global package use":
-  rmDir("nimbledeps")
+task test_slim, "Runs the (slimmed down) test suite":
+  `testnative Task`()
```
Menduist, line 109: Why is that change necessary? To keep some environment?

libp2p.nimble
```diff
@@ -109,52 +99,23 @@ task examples_build, "Build the samples":
...
-const PinFile = ".pinned"
-task pin, "Create a lockfile":
-  # pinner.nim was originally here
-  # but you can't read output from
-  # a command in a nimscript
-  exec "nim c -r tools/pinner.nim"
-
-import sequtils
-import os
-task install_pinned, "Reads the lockfile":
```
Menduist, line 130: As discussed on discord, I'd rather keep this to have compat with old nimble, and maybe do
```nim
task install_pinned "Reads the lockfile":
  when nimbleIsCompatibleWithNewSystem:
    exec "nimble install -dy" # or whatever
  else:
    # the old code
```

libp2p.nimble
yyoncho: To keep the params passed to nimble.

libp2p.nimble
yyoncho: I will see what we can do. I think we can go with test scope, but I will check if it will work.

.github/workflows/codecov.yml
yyoncho: You should be able to do `nimble setup --lock-file=nimble-version-1-6.lock` and from now on it won't need `--lock-file=nimble-version-1-6.lock` . `nimble setup` call will create `nimble.paths` and they will be used when there is no lock file specified.

.github/workflows/ci.yml
yyoncho: Yes, we can - given the fact that we are using lock files then we can just cache the `~/.nimble/` dir unconditionally.

Menduist: Ok nice (we should still use the hash of the lockfile has a key to avoid using the wrong cache though)

yyoncho: There is no need to do that - it doesn't matter what you have in initially ~/.nimble/pkgs it will always either download the correct version of the package version or it will reuse it if it is already present.

libp2p.nimble
yyoncho: With current nimble functionality the way to fix the issue is to use taskRequires for each test ask and add the deps like this: 

``` nim
if (NimMajor, NimMinor) > (1, 2):
   taskRequires "test", "nimpng#HEAD", "nico"
   taskRequires "testnative", "nimpng#HEAD", "nico"
   taskRequires "<whatever>", "nimpng#HEAD", "nico"
```

A possible alternative is to expose the action name from here: https://github.com/nim-lang/nimble/blob/master/src/nimblepkg/nimscriptapi.nim#L82

and do something like:

``` nim 
if (NimMajor, NimMinor) > (1, 2) and actionName not in {"build", "install"}:
   requires "nimpng#HEAD", "nico"
```


cc @zah

Menduist: To be more precise, it's only required by `nimble examples_build `

.github/workflows/ci.yml
Menduist: Sure, the nice thing about using the hash as key is that you can have multiple lockfiles deps cached in parallel, instead of having a single `pkgs` folder which may or may not be the correct one

.github/workflows/codecov.yml
Menduist: Maybe it's cleaner to do nimble setup in the CI too instead of passing the lock file everytime?

yyoncho: this is what I was doing initially, but @zah opposed based on the argument that `setup` is something that only the user should do for local development.

Menduist: Ok, don't care particularly about it, but is there a specific reason it's only for users?

yyoncho: I am not 100% sure, I guess it is related to polluting the workspace with nimble.paths file. I will let @zah elaborate.

zah: It's a matter of hygiene. If some of the Nimble commands don't work when `nimble setup` hasn't been executed, this means that we've left a bug somewhere in Nimble.

The primary goal of `nimble setup` is to enable the local use of the `nim` executable (and other tools such as `nimsuggest`) without wrapping their executions in a `nimble` command that setups the necessary paths.

PR #845, opened 2023-01-19
Add expected connections to connmngr

Menduist:
This system allows you to wait for a connection from a specific peer. It will bypass the `maxConnectionsPerPeer`, since you know you are waiting for him (but not the global limit of connections)

This is useful for AutoNat, HolePunching, maybe others

Also.. turns out the maxConnectionsPerPeer was off by one (hurray). I fixed it here, but maybe we should leave it as is to avoid changing the behavior. idk
EDIT: reverted, breaks to many things

### DIFF
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -411,10 +411,9 @@ proc storeConn*(c: ConnManager, conn: Connection)
 
   let peerId = conn.peerId
 
-  if peerId in c.expectedConnections:
-    let fut = c.expectedConnections.getOrDefault(peerId)
-    if not fut.finished:
-      fut.complete(conn)
+  if peerId in c.expectedConnections and
+     not(c.expectedConnections.getOrDefault(peerId).finished):
+      c.expectedConnections.getOrDefault(peerId).fut.complete(conn)
   elif c.conns.getOrDefault(peerId).len > c.maxConnsPerPeer:
     debug "Too many connections for peer",
       conn, conns = c.conns.getOrDefault(peerId).len
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -417,7 +417,7 @@ proc storeConn*(c: ConnManager, conn: Connection)
   if peerId in c.expectedConnections:
     for future in c.expectedConnections.getOrDefault(peerId):
       future.complete(conn)
-  elif c.conns.getOrDefault(peerId).len >= c.maxConnsPerPeer:
+  elif c.conns.getOrDefault(peerId).len > c.maxConnsPerPeer:
     debug "Too many connections for peer",
       conn, conns = c.conns.getOrDefault(peerId).len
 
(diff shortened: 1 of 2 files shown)


libp2p/connmanager.nim
```diff
@@ -396,7 +410,11 @@ proc storeConn*(c: ConnManager, conn: Connection)
     raise newException(LPError, "Connection closed or EOF")
 
   let peerId = conn.peerId
-  if c.conns.getOrDefault(peerId).len > c.maxConnsPerPeer:
+
+  if peerId in c.expectedConnections and
+     not(c.expectedConnections.getOrDefault(peerId).finished):
+      c.expectedConnections.getOrDefault(peerId).fut.complete(conn)
```
diegomrsantos, line 416: Is `getOrDefault` necessary? Does `peerId in c.expectedConnections` imply always the presence of a value for the key?

diegomrsantos, line 416:
```suggestion
  if peerId in c.expectedConnections and
     not(c.expectedConnections.getOrDefault(peerId).finished):
      c.expectedConnections.getOrDefault(peerId).complete(conn)
```

Menduist: This is just to avoid the `KeyError` that `[]` can raise (even though it can't actually raise it since we checked)
Ideally we would have a `[]` that raises a defect if the value is not here to avoid this issue

tests/testconnmngr.nim
```diff
@@ -96,7 +96,7 @@ suite "Connection Manager":
     await connMngr.close()
 
   asyncTest "get conn with direction":
-    let connMngr = ConnManager.new()
+    let connMngr = ConnManager.new(maxConnsPerPeer = 2)
```
diegomrsantos, line 99: Not sure if it's a good idea to add this at the moment. Even if we lower this value to 1, the test will still pass.

tests/testconnmngr.nim
```diff
@@ -187,10 +187,45 @@ suite "Connection Manager":
...
     expect TooManyConnectionsError:
       connMngr.storeConn(conns[0])
+
+    await connMngr.close()
+
+    await allFuturesThrowing(
+      allFutures(conns.mapIt( it.close() )))
+
+  asyncTest "expect connection from peer":
+    let connMngr = ConnManager.new(maxConnsPerPeer = 0)
```
diegomrsantos, line 197: This is even more confusing, IMHO. If we don't want to fix this bug in this PR, I would keep it as 1 and add a comment explaining the bug with a FIXME.

Menduist: Except that we really need a limit of 1 for this test to pass

diegomrsantos: I mean, with a limit of 0, this test should fail, shouldn't it?

Menduist (commented):
Can't approve, but I approve :)

diegomrsantos (approved):

PR #849, opened 2023-01-20
Connect is able to force a new connection

diegomrsantos:


libp2p/dialer.nim
```diff
@@ -207,15 +213,16 @@ method connect*(
   self: Dialer,
   peerId: PeerId,
   addrs: seq[MultiAddress],
-  forceDial = false) {.async.} =
+  forceDial = false,
+  forceNewConnection = false) {.async.} =
```
Menduist, line 217: Nitpick, but maybe `reuseConnection = true` is better?

Menduist (dismissed):
Apart from nitpick, LGTM

libp2p/dialer.nim
diegomrsantos: It seems better, but on the other hand it is different from the existing pattern "forceDial".

Menduist (approved):

Issue #850, opened 2023-01-23
gossipsub: Lazy sending


arnetheduck:
Consider:

* peer A sends a message to us which is sent to validation - peer B sends the same message to us and when the message validates, we skip sending it to peer B
* however, we still send the data to peer C and D
* sending data to peed C takes time - for example, streaming 1mb to peer C takes 1s on a 10mbit line
* peer D completes sending the same message to us

In the above, we can skip sending to peer D as well - to make this work:
* track the completion of "sending" to peer C
* when the future completes, check if sending to D is still relevant

Problems:

* C might be slow which increases latency of sending to D
  * send to `Dmin` peers, then use lazy sending for the rest

Menduist:
That seems a bit scary to me, might open a new category of attacks ("slow peers" which lowers our scores regarding legit peers since we become slow)

We could however cancel the send if we receive the duplicate while the message was queued

arnetheduck:
well, there are already cases where we run with `Dmin` peers - ie we consider that "ok" - we can further limit damage by adding a "wait-for-send-timeout" in the low 100's of ms (or stagger sending after `Dmin` - the savings here are significant though, ie peer A/B case that we already do today gives a noticable reduction in processing/bandwidth

Menduist:
> well, there are already cases where we run with Dmin peers - ie we consider that "ok"

My worry isn't that we only send to Dmin peers (though that will also have an impact on the time it takes to reach the full network), but that other peers which are in our mesh will think that we are slow, and thus descore us

arnetheduck:
> our mesh will think that we are slow,

there are 2 cases here:
* we run out of bandwidth -> sending in parallel slows down everyone because of multuplexing effects -> worst case for all -> lazy sending allows some peers to receive the whole message faster -> better scoring overall
* we hit the timeout before we saturate bandwidth -> the `>dmin` peers receive the message later than they would have already - we have to tune the timeout parameter such that we are withing safe "Slowness" limits with respect to scoring

Menduist:
> lazy sending allows some peers to receive the whole message faster -> better scoring overall

Good point. I would still like to do this in two steps, #851 which just cancels outgoing messages, and a second PR that caps the inflight requests to dmin (with a timeout)

I think the cancel of inflight will naturally send less stuff if you are the one being slow, since more stuff will be queued. But we can measure & verify that

Menduist:
[...]

PR #851, opened 2023-01-24
GossipSub: cancel inflight msgs when receiving duplicate

Menduist:
ref #850

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -43,6 +43,7 @@ logScope:
 declareCounter(libp2p_gossipsub_failed_publish, "number of failed publish")
 declareCounter(libp2p_gossipsub_invalid_topic_subscription, "number of invalid topic subscriptions that happened")
 declareCounter(libp2p_gossipsub_duplicate_during_validation, "number of duplicates received during message validation")
+declareCounter(libp2p_gossipsub_duplicate_during_broadcast, "number of duplicates received during message broadcast")
 declareCounter(libp2p_gossipsub_duplicate, "number of duplicates received")
 declareCounter(libp2p_gossipsub_received, "number of messages received (deduplicated)")
 
@@ -291,9 +292,27 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
         libp2p_pubsub_broadcast_prune.inc(labelValues = ["generic"])
 
     trace "sending control message", msg = shortLog(respControl), peer
-    g.send(
+    asyncSpawn g.send(
       peer,
       RPCMsg(control: some(respControl), messages: messages))
+
+proc lazyBroadcast(
+    g: GossipSub,
+    peers: seq[PubSubPeer],
+    msgIdSalted: MessageId,
+    msg: RPCMsg) {.async.} =
+  let
+    futsTable = g.cancellableBroadcast(peers, msg)
+    futs = toSeq(futsTable.values)
+  g.sendingFutures[msgIdSalted] = futsTable
+  await allFutures(futs)
+  g.sendingFutures.del(msgIdSalted)
+
+  # This isn't equal to the amount of saved bandwidth,
+  # because lower layer may send the data even after
+  # cancellation
+  let cancelledCount = futs.countIt(it.cancelled)
+  libp2p_gossipsub_duplicate_during_broadcast.inc(cancelledCount.int64)
 
 proc validateAndRelay(g: GossipSub,
                       msg: Message,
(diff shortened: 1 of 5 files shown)


arnetheduck:
this carries the risk that the other end disconnects / descores due to the stream being reset mid-sending - I'm not sure it's a good idea in general - ie once the channel is open and streaming, interrupting the send looks like we're just trying to waste bandwidth

Menduist:
Cancelling a write will not reset a stream, it will just "unqueue" the message if that is still possible

However, for very big messages, it might make sense to a open a new stream just for it (when we'll have 0rtt negotiation), and being able to reset it mid-air.

arnetheduck:
> However, for very big messages,

right, right now, I'm in the middle of discussions of features like this for 1mb-sized messages - partial streaming is highly likely - a similar feature is to not send messages to peers that have an in-progress send of potentially the same message via some kind of prefix comparison (or maybe by including the computed message id in the "header" of the stream.

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -192,24 +189,50 @@ proc broadcast*(
...
+    peers = sendPeers.len, msg = shortLog(msg)
+
+  if anyIt(sendPeers, it.hasObservers):
+    for peer in sendPeers:
+      result[peer.peerId] = p.send(peer, msg)
+  else:
+    # Fast path that only encodes message once
+    let encoded = encodeRpcMsg(msg, p.anonymize)
+    for peer in sendPeers:
+      result[peer.peerId] = peer.sendEncoded(encoded)
```
arnetheduck, line 228: doesn't this already synchronously put bytes in the send buffer?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -291,10 +292,28 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
...
 
+proc lazyBroadcast(
+    g: GossipSub,
+    peers: seq[PubSubPeer],
+    msgIdSalted: MessageId,
+    msg: RPCMsg) {.async.} =
+  let
+    futsTable = g.cancellableBroadcast(peers, msg)
+    futs = toSeq(futsTable.values)
+  g.sendingFutures[msgIdSalted] = futsTable
```
arnetheduck, line 307: can't there be two inflight lazy broadcasts of the same message?

Menduist: That shouldn't happen, I may add an assert to be sure

libp2p/protocols/pubsub/pubsub.nim
Menduist: With mplex, that will indeed go synchronously through every layer, down to chronos
I thought that chronos supported write cancellation, apparently not, quoting cheatfate:
> There is a big problem with write and i wish not to solve it today or tomorrow, because write is made using queue, so cancelling write in the middle should remove this write call from the queue, but it will break cross-platform compatibility, because on Windows this queue is in system, so i can't freely remove write call from the middle of the queue...

So we could add cancellation support to linux at least

With yamux, there is an intermediary queue before chronos (since there is per-stream backpressure), that does support cancellation:
https://github.com/status-im/nim-libp2p/blob/351bda2b56e4389a91fb42199621b3617bdbbdd3/libp2p/muxers/yamux/yamux.nim#L285

So this PR only makes sense if we get chronos support for write cancellation on linux, or enable yamux in nimbus :/

Menduist:
> a similar feature is to not send messages to peers that have an in-progress send of potentially the same message via some kind of prefix comparison

Yeah, seems like the ideal solution would be a new protocol that opens a stream eagerly, sends the message id & the message

The receiving end can reset the stream at any point if he gets the message first somewhere else (or if many peers are currently sending him the same message)

libp2p/protocols/pubsub/pubsub.nim
Menduist: Getting TCP cancellation support in chronos seems less complex than expected: https://github.com/status-im/nim-chronos/pull/353
I'll wait for reviews to be sure

arnetheduck:
> Yeah, seems like the ideal solution would be a new protocol that opens a stream eagerly, sends the message id & the message

actually, if we start receiving a message from someone we *could* actually look at the first few bytes and delay send iff they match whatever is being sent - basically, when we perform the send, we'd compare the message being sent against whatever partial bytes that have been received - a bit tricky perhaps because of protobuf and its unordered fields, but perhaps doable. easier to do with explicit support perhaps.

Menduist:
This is not actually doable, since cancelling an outbound message is unpractical

PR #855, opened 2023-02-07
Dialing addrs concurrently in autonat

diegomrsantos:
There's also some change in the log levels. The motivation is to provide more useful information when running the system in debug. Currently, most of the logs use trace and it is unrealistic to run the system in a non-local environment with this log level due to the huge amount of data created.

### DIFF
--- a/libp2p/protocols/connectivity/autonat/server.nim
+++ b/libp2p/protocols/connectivity/autonat/server.nim
@@ -77,9 +77,6 @@ proc tryDial(autonat: Autonat, conn: Connection, addrs: seq[MultiAddress]) {.asy
   except CancelledError as exc:
     raise exc
   except AllFuturesFailedError as exc:
-    for f in futs:
-      if not f.finished():
-        f.cancel()
     debug "All dial attempts failed", addrs, exc = exc.msg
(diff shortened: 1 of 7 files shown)


libp2p/protocols/connectivity/autonat/server.nim
```diff
@@ -60,21 +60,34 @@ proc sendResponseOk(conn: Connection, ma: MultiAddress) {.async.} =
...
     if ma.isSome:
       await conn.sendResponseOk(ma.get())
     else:
       await conn.sendResponseError(DialError, "Missing observed address")
   except CancelledError as exc:
     raise exc
+  except AllFuturesFailedError as exc:
+    for f in futs:
+      if not f.finished():
+        f.cancel()
```
Menduist, line 82: This should happen in finally (or ideally, in finally and just after the anyCompleted)

diegomrsantos: Why finally? But it's in the wrong place indeed. I meant to add it in `except AsyncTimeoutError`.

diegomrsantos: You are right, we also should do that when one future completes.

Menduist (approved):

PR #873, opened 2023-03-07
Transport hole punching

diegomrsantos:
Previously, when dialing a peer using the TcpTransport, the OS randomly chose the socket local port. It means that when an internal peer dials a remote one, the latter will always observe a random port. But to do TCP hole punching, we need a stable and predictable public port that will be advertised to the peers we want to create a TCP simultaneous open. This is known as the NAT port prediction problem. 

To solve this problem,  we add the possibility of binding to the TCP listening port before dialing remote peers in this PR. The TCP transport will always bind when the peer's network reachability is "not reachable" and not bind otherwise. We assume the NAT uses the port preservation scheme, i.e. it maps the source port of the internal peer to the same public port. 

More info:

- https://en.wikipedia.org/wiki/TCP_hole_punching
- https://ttcplinux.sourceforge.net/documents/one/tcpstate/tcpstate.html#:~:text=Simultaneous%20Open,active%20open%20on%20both%20sides%22.


.pinned
```diff
@@ -1,6 +1,6 @@
 bearssl;https://github.com/status-im/nim-bearssl@#acf9645e328bdcab481cfda1c158e07ecd46bd7b
 chronicles;https://github.com/status-im/nim-chronicles@#32ac8679680ea699f7dbc046e8e0131cac97d41a
-chronos;https://github.com/status-im/nim-chronos@#f7835a192b45c37e97614d865141f21eea8c156e
+chronos;https://github.com/status-im/nim-chronos@#d488db3324aed2f096a63a926e3f709d6975e4df
```
Menduist, line 3: We can't merge until the PR is merged, unfortunately: https://github.com/status-im/nim-chronos/pull/362

libp2p/protocols/connectivity/autonat/core.nim
```diff
@@ -58,6 +58,9 @@ type
     dial*: Option[AutonatDial]
     response*: Option[AutonatDialResponse]
 
+  NetworkReachability* {.pure.} = enum
+    NotReachable, Reachable, Unknown
```
Menduist, line 62:
```suggestion
    Unknown, NotReachable, Reachable
```
Unknown should be the default value of this enum, this would also avoid having to set it explicitly to Unknown in the TCP constructor

Menduist (approved):

PR #875, opened 2023-03-14
GossipSub: Better IWANT handling

Menduist:
Currently, IWANTs are only limited by the budget (25 messages per heartbeat)
That's quite random, and causes some issues when peer are asking for more than 25 messages per HB

This PR changes the spam protection to only allow one IWANT per message sent in each IHAVE
In the future, we could also avoid sending an IHAVE twice for the same message. This PR opens the door to that

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -31,7 +31,7 @@ declareGauge(libp2p_gossipsub_no_peers_topics, "number of topics in mesh with no
 declareGauge(libp2p_gossipsub_low_peers_topics, "number of topics in mesh with at least one but below dlow peers")
 declareGauge(libp2p_gossipsub_healthy_peers_topics, "number of topics in mesh with at least dlow peers (but below dhigh)")
 declareCounter(libp2p_gossipsub_above_dhigh_condition, "number of above dhigh pruning branches ran", labels = ["topic"])
-declareSummary(libp2p_gossipsub_mcache_hit, "ratio of successful IWANT message cache lookups")
+declareGauge(libp2p_gossipsub_received_iwants, "received iwants", labels = ["kind"])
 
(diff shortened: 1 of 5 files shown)


libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -275,25 +275,19 @@ proc handleIWant*(g: GossipSub,
   var messages: seq[Message]
   if peer.score < g.parameters.gossipThreshold:
     trace "iwant: ignoring low score peer", peer, score = peer.score
-  elif peer.iWantBudget <= 0:
-    trace "iwant: ignoring out of budget peer", peer, score = peer.score
   else:
-    let deIwants = iwants.deduplicate()
```
Menduist, line 281: I've remove deduplication since this now happens naturally, though the computation is a tad more expensive

arnetheduck:
what's the upper bound on memory usage for this?

Menduist:
Everytime we send an IHAVE, the message id will be stored for `historyLength`

Plugging in realistic values:
historyLength: 6 = 4.2 seconds
Each messages gossiped to: ~20 peers
Msgs per seconds: 1300
Msg id size: 32 bytes

1300 * 4.2 * 20 * 32 = 3.49 MB, as a worst case scenario

arnetheduck:
> Having this data is also useful for future bandwidth optimization (and already optimize a bit, since a peer can now only IWANT us a single time per message, instead of multiple times currently)

yeah, I was thinking about that vs a simple counter that per-peer says how many IHAVE's we've sent them recently, but if it's in the single-digit MB range it seems reasonable

lchenut (approved):

Menduist:
I started this PR using a counter, but then with wildly different messages size it gets pretty bad (I send you 10 IHAVE for 100 bytes msgs, you reply with 10 IWANT for 1 mb msgs)
So I tried to do a counter of message size, but it also gets messy-ish, because we have to find the message in the cache to get its size, which is a tad expensive

PR #876, opened 2023-03-21
Yamux automatic window scaling

Menduist:
For better performance, the yamux window will be scaled to the bigger "read" that happened on that channel.
Could be improved later to handle window reduction, and faster to react


Menduist:
Yup, I've read it while originally implementing yamux :)
Definitevily something we'll have to implement too once "high bandwidth" becomes a priority, but for now, our priority is just "burst of high traffics" (like ethereum blocks every 12 seconds), which doesn't require to be as fancy

richard-ramos:
Closing as PR is stale, and focus should be on quic

PR #877, opened 2023-03-22
Does not allow an empty MA

diegomrsantos:


Menduist (dismissed):

Menduist:
Just linking discussions here: https://github.com/multiformats/rust-multiaddr/issues/85

Menduist (approved):

PR #878, opened 2023-03-28
Add getters for conns and streams

alrevuelta:
* Add getter function to access the connections and streams within each connection.
* This would allow nim-libp2p users to have a better view on the connections/streams they have, giving access to its direction, protocol being used and other relevant fields.

Closes https://github.com/status-im/nim-libp2p/issues/866

### DIFF
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -113,6 +113,9 @@ proc connectedPeers*(c: ConnManager, dir: Direction): seq[PeerId] =
       peers.add(peerId)
   return peers
 
+proc getConnections*(c: ConnManager): Table[PeerId, seq[Muxer]] =
+  return c.muxed
+
 proc addConnEventHandler*(c: ConnManager,
                           handler: ConnEventHandler,
                           kind: ConnEventKind) =
--- a/libp2p/muxers/muxer.nim
+++ b/libp2p/muxers/muxer.nim
@@ -63,3 +63,5 @@ proc new*(
 
   let muxerProvider = T(newMuxer: creator, codec: codec)
   muxerProvider
+
+method getStreams*(m: Muxer): seq[Connection] {.base.} = doAssert false, "not implemented"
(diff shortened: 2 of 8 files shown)


libp2p/muxers/muxer.nim
```diff
@@ -63,3 +63,5 @@ proc new*(
 
   let muxerProvider = T(newMuxer: creator, codec: codec)
   muxerProvider
+
+method getStreams*(m: Muxer): seq[Connection] {.base.} = doAssert false, "not implemented"
```
Menduist, line 67: This will also need an implementation for yamux, around here:
https://github.com/status-im/nim-libp2p/blob/af5299f26c195540d3fbe0f1a20742658f8a6220/libp2p/muxers/yamux/yamux.nim#L509

Should be similar to mplex, using `m.channels`

libp2p/connmanager.nim
```diff
@@ -113,6 +113,9 @@ proc connectedPeers*(c: ConnManager, dir: Direction): seq[PeerId] =
       peers.add(peerId)
   return peers
 
+proc getConnections*(c: ConnManager): Table[PeerId, seq[Muxer]] =
```
Menduist, line 116: Should be easy to add a test around here:
https://github.com/status-im/nim-libp2p/blob/af5299f26c195540d3fbe0f1a20742658f8a6220/tests/testconnmngr.nim#L31-L43

tests/testmplex.nim
```diff
@@ -710,9 +711,10 @@ suite "Mplex":
 
       let mplexDial = Mplex.new(conn)
       let mplexDialFut = mplexDial.handle()
-      var dialStreams: seq[Connection]
-      for i in 0..9:
-        dialStreams.add((await mplexDial.newStream()))
+      var dialStreams = toSeq(0..9).mapIt(await mplexDial.newStream())
+
+      check:
+        isEqual(dialStreams, mplexDial.getStreams())
```
alrevuelta, line 717: note that the order is not guaranteeed due to how its implemented

```
method getStreams*(m: Mplex): seq[Connection] =
  for c in m.channels[false].values: result.add(c)
  for c in m.channels[true].values: result.add(c)
```

Menduist (approved):

PR #879, opened 2023-04-03
Fix traffic metrics

Menduist:
Thanks @alrevuelta for the report


lchenut (approved):

alrevuelta:
Indeed it works now, thanks!

Image showing before/after fix. Before, nim-libp2p metrics differed from cadvisor (same pattern but amplified). After 14:05 (see time axis), I deployed with the fix in this PR and the bandwidths now match.
image

(Running nwaku)

PR #880, opened 2023-04-11
Add PeerId specs key decoding tests

Menduist:
https://github.com/libp2p/specs/pull/537


Menduist:
Upstream ended up using our test for specs

PR #881, opened 2023-04-13
Invalid MA is ignored

diegomrsantos:


libp2p/multiaddress.nim
```diff
@@ -1128,6 +1131,5 @@ proc getRepeatedField*(pb: ProtoBuffer, field: int,
       if ma.isOk():
         value.add(ma.get())
       else:
-        value.setLen(0)
-        return err(ProtoError.IncorrectBlob)
+        debug "Not supported MultiAddress in blob", ma = item
```
Menduist, line 1134: This doesn't seems like the best solution
Maybe create a `tryGetRepeatedField`, that returns the number of failed parses? This way the caller can handle this however it wants

Menduist: If the current version is not relevant anywhere, then we should change this proc to return the number of failed parses (or eq)

Menduist: And update every usage accordingly (and check it's not used by nimbus & co)

diegomrsantos: pushed changes

Menduist (approved):

PR #883, opened 2023-04-25
Fix gossipsub dOut handling

Menduist:
Thanks @alrevuelta for the report

dOut wasn't actually looking at the current number of outgoing peers, it just grafted up to dOut everytime

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -164,7 +164,8 @@ proc handleGraft*(g: GossipSub,
     # If they send us a graft before they send us a subscribe, what should
     # we do? For now, we add them to mesh but don't add them to gossipsub.
     if topic in g.topics:
-      if g.mesh.peers(topic) < g.parameters.dHigh or peer.outbound:
+      if g.mesh.peers(topic) < g.parameters.dHigh or (peer.outbound and
+          g.mesh.getOrDefault(topic).countIt(it.outbound) < g.parameters.dOut):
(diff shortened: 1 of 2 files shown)


libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -399,7 +400,7 @@ proc rebalanceMesh*(g: GossipSub, topic: string, metrics: ptr MeshMetrics = nil)
     candidates.sort(byScore, SortOrder.Descending)
 
     # Graft peers so we reach a count of D
```
Menduist, line 402:
```suggestion
    # Graft outgoing peers so we reach a count of dOut
```

libp2p/protocols/pubsub/peertable.nim
```diff
@@ -53,3 +53,10 @@ func peers*(table: PeerTable, topic: string): int =
     except KeyError: raiseAssert "checked with in"
   else:
     0
+
+func outboundPeers*(table: PeerTable, topic: string): int =
+  if topic in table:
+    try: table[topic].countIt(it.outbound)
```
alrevuelta, line 59: out of curiosity, is the `try` needed? `if topic in table` should protect from trying to access a non existing key? so `KeyError` should never happen?

Menduist: It should never happen indeed, but the compiler thinks it can

I think this all proc could just be `table.getOrDefault(topic).countIt(it.outbound)`, but wanted to keep it consistent with the one above

lchenut (approved):

PR #886, opened 2023-05-05
GossipSub: remove peer if we can't establish sendConn

Menduist:
Currently, if we can't establish a sendConn, we will swallow messages, which can lead to asymmetric mesh and useless memory usage
This PR changes the behavior to consider the peer disconnected if we can't establish a sendConn in 5 seconds


diegomrsantos:
how about tests?

Menduist:
This PR is very hard to test because it only happens with buggy clients

PR #891, opened 2023-05-16
Try a direct connection only if there isn't one already

diegomrsantos:


libp2p/services/hpservice.nim
```diff
@@ -71,22 +96,10 @@ method setup*(self: HPService, switch: Switch): Future[bool] {.async.} =
...
-            natAddrs =  switch.peerInfo.listenAddrs.mapIt(switch.peerStore.guessDialableAddr(it))
-          await dcutrClient.startSync(switch, peerId, natAddrs)
-          await sleepAsync(2000.milliseconds) # grace period before closing relayed connection
-          await conn.close()
-      except CatchableError as err:
-        debug "Hole punching failed during dcutr", err = err.msg
+    proc newConnectedPeerHandler(peerId: PeerId, event: PeerEvent) {.async.} =
+      await newConnectedPeerHandlerAux(self, switch, peerId, event)
+
+    self.newConnectedPeerHandler = newConnectedPeerHandler
```
Menduist, line 102:
```suggestion
    self.newConnectedPeerHandler = proc (peerId: PeerId, event: PeerEvent) {.async.} =
      await newConnectedPeerHandlerAux(self, switch, peerId, event)
```
And you cane rename newConnectedPeerHandlerAux to onNewPeer or something like that

libp2p/services/hpservice.nim
```diff
@@ -63,6 +63,31 @@ proc tryStartingDirectConn(self: HPService, switch: Switch, peerId: PeerId): Fut
...
+    # Get all connections to the peer. If there is at least one non-relayed connection, return.
+    var relayedConnt: Connection
+    for muxer in switch.connManager.getConnections()[peerId]:
+      let conn = muxer.connection
+      if not isRelayed(conn):
+        return
+      elif conn.transportDir == Direction.In:
+        relayedConnt = conn
+
+    if not isNil(relayedConnt):
```
Menduist, line 77:
```suggestion
    let connections = switch.connManager.getConnections()[peerId].mapIt(it.connection)
    if connections.anyIt(not isRelayed(it)):
      return
    let incomingRelays = connections.filterIt(it.transportDir == Direction.In)
    if incomingRelays.len == 0:
      return
    
    let relayedConn = incomingRelays[0]
```
(need to de-ident the rest of the code)
wdyt?

Menduist (approved):
LGTM besides style questions

libp2p/services/hpservice.nim
diegomrsantos: Not sure if this is better

Menduist: I think the current version is slightly confusing, I tried to convey the intention more clearly

diegomrsantos: done

Menduist (approved):

PR #892, opened 2023-05-17
Refinement of Hole Punching Service

diegomrsantos:
This PR incorporates several updates to the Hole Punching Service.

Key Changes:

1. The `dcutr` sync message doesn't send addresses anymore.

2. The `dcutr` server upgrades as the incoming side instead of the client. It was necessary in order to work with go-libp2p.

3. The `dcutr` version has been updated to exclude the version number as it is in the spec.

4. A new condition has been added during the direct connection attempt to check if an address contains the "p2p-circuit". If it does, or if an error is encountered, the function will continue without further processing.

5. The initTAddress has been fixed to not accept DNS addresses. Tests were added.


Menduist (approved):

tests/testhpservice.nim
```diff
@@ -81,10 +80,13 @@ suite "Hole Punching":
...
 
     let privatePeerSwitch = createSwitch(relayClient, hpservice)
+    let peerSwitch = createSwitch()
     let switchRelay = createSwitch(Relay.new())
 
-    await allFutures(switchRelay.start(), privatePeerSwitch.start(), publicPeerSwitch.start())
+    await allFutures(switchRelay.start(), privatePeerSwitch.start(), publicPeerSwitch.start(), peerSwitch.start())
 
+    await privatePeerSwitch.connect(peerSwitch.peerInfo.peerId, peerSwitch.peerInfo.addrs) # for autonat
+    await sleepAsync(200.millis) # wait for autorelay to kick in
```
Menduist, line 89: Try replacing with checkExpiring
And check if its a bug with auto relay @lchenut

diegomrsantos: Same issue as before in Linux. Maybe my theory was wrong.

libp2p/services/hpservice.nim
```diff
@@ -52,6 +51,9 @@ proc tryStartingDirectConn(self: HPService, switch: Switch, peerId: PeerId): Fut
   await sleepAsync(500.milliseconds) # wait for AddressBook to be populated
   for address in switch.peerStore[AddressBook][peerId]:
     try:
+      let isRelayed = address.contains(multiCodec("p2p-circuit"))
+      if isRelayed.isErr() or isRelayed.get():
+        continue
```
Menduist, line 56: Why not reuse `isRelayed()`?

diegomrsantos: Do you mean the proc in relay.nim? If so, it is for a connection, this one is for a MA.

PR #895, opened 2023-05-23
Populate peerstore before triggering events

alrevuelta:
* Peerstore was populated after triggering peer connection events. This caused that when reading the peerstore `addPeerEventHandler` handler for the inbound connecting `peerId`, its fields were not yet populated.
* Add testcase showcasing the issue. Without the changes made in this PR, that test would fail.


alrevuelta:
thanks, closing then.

PR #898, opened 2023-05-30
Fail only if all addresses in PeerRecord are invalid

diegomrsantos:
Related to:

* https://github.com/waku-org/nwaku/issues/1768

### DIFF
--- a/libp2p/routing_record.nim
+++ b/libp2p/routing_record.nim
@@ -51,10 +51,12 @@ proc decode*(
     for address in addressInfos:
       var addressInfo = AddressInfo()
       let subProto = initProtoBuffer(address)
-      if ? subProto.getField(1, addressInfo.address) == false:
-        return err(ProtoError.RequiredFieldMissing)
+      let f = subProto.getField(1, addressInfo.address)
+      if f.isOk() and f.get():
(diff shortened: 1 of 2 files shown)


libp2p/routing_record.nim
```diff
@@ -51,10 +51,12 @@ proc decode*(
...
       let subProto = initProtoBuffer(address)
-      if ? subProto.getField(1, addressInfo.address) == false:
-        return err(ProtoError.RequiredFieldMissing)
+      let f = subProto.getField(1, addressInfo.address)
+      if f.isOk() and f.get():
+          record.addresses &= addressInfo
 
-      record.addresses &= addressInfo
+    if record.addresses.len == 0:
+      return err(ProtoError.RequiredFieldMissing)
```
Menduist, line 59: This shouldn't fail if addressInfos is empty

diegomrsantos: It doesn't, added a test.

Menduist (approved):

jakubgs:
It's a good idea to include a reference to the https://github.com/waku-org/nwaku/issues/1768 issue in both he commit and the PR description.

PR #908, opened 2023-06-06
use `bearssl` `new` instead of `newRng`

etan-status:
`bearssl` meanwhile has a copy of `newRng`, so the own one is no longer needed. Replace `newRng` with `HmacDrbgContext.new()` and bump bearssl. Also see https://github.com/status-im/nim-websock/pull/132


libp2p/builders.nim
```diff
@@ -242,9 +241,6 @@ proc build*(b: SwitchBuilder): Switch
   if b.secureManagers.len == 0:
     b.secureManagers &= SecureProtocol.Noise
 
-  if isNil(b.rng):
-    b.rng = newRng()
-
```
etan-status, line 247: Unreachable, due to the assert above

etan-status:
Probably not worth, given this leaks even into client code: https://github.com/libp2p/test-plans

arnetheduck:
`newRng` can simply call `HmacDrbgContext.new()`, no?

etan-status:
Yeah, that would be an alternative. There were issues with `newRng` conflicts across the various modules that offer this function. See https://github.com/status-im/nim-websock/pull/132

Was thinking about `HmacDrbgContext.new()`, but realistically, that also requires a `nim-bearssl` bump. I don't want to be the one introducing flakiness because someone on an older 0.1.5 commit doesn't have the override for `HmacDrbgContext.new()` but instead uses the system `new` that allows a zeroed context instead, randomly, depending on what nimble cached :-)

Issue #854, opened 2023-02-07
GossipSub: improve flood publish


Menduist:
Flood publish:
:white_check_mark: faster propagation of messages
:white_check_mark: better sybil resistance (the first hop is the most fragile one)
:cry: bandwidth usage (scales linearly with number of subbed connected peers)
:cry: privacy (when you send a message out of your mesh, you reveal that you are the source)

The bandwidth usage will soon become a bottleneck in ethereum, and we need to find better solutions. Some leads:
- Limit the number of peers we flood publish to (reuse `gossipFactor`?)
- Rely on flooding `IHAVE` instead of the full message?
- ?

Will also need adjustments at the specs level

arnetheduck:
* stagger sending so that some peers receive the (full) message earlier

Menduist:
Sure, whatever we end up doing for #850 will also apply to flood publishing

Menduist:
The performance part has been addressed in #911
The privacy part is a harder task to achieve

diegomrsantos:
Flood publishing has been addressed as described in the comment above. 

Other improvements related to bandwidth and latency were addressed in:

* https://github.com/vacp2p/nim-libp2p/pull/934
* https://github.com/vacp2p/nim-libp2p/pull/1051
* https://github.com/vacp2p/nim-libp2p/pull/1077
* https://github.com/vacp2p/nim-libp2p/pull/1083 (waiting review)

Can we consider this issue solved? @arnetheduck @kaiserd

PR #911, opened 2023-06-09
GossipSub: Limit flood publishing

Menduist:
Part of #854

This PR adds a limit to the flood publishing, so that we spend at most 1 heartbeat flood publishing
Of course, this requires to know our bandwidth, which is currently hardcoded to 100 mbps, but this will be replaced by an actual estimate later

With 700ms heartbeat and 1mb message size, we will send it to a maximum of 17 peers (including the mesh)

The "1 heartbeat" period is used because after that, we will probably be responding to IWANTs

Next step for flood publishing will be to integrate https://github.com/status-im/nim-libp2p/issues/850 (staggered sending), which will not only give us the bandwidth estimate, but also send first to the mesh and then the flood.


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -488,20 +488,28 @@ method publish*(g: GossipSub,
...
+
+  if topic in g.topics: # if we're subscribed use the mesh
+    peers.incl(g.mesh.getOrDefault(topic))
+
   if g.parameters.floodPublish:
+    let
+      msgSize = data.len
+      bandwidth = 25_000_000 #TODO replace with bandwidth estimate
+      msToTransmit = max(msgSize div (bandwidth div 1000), 1)
+      maxFloodPublish =
```
arnetheduck, line 502: no need for a `min`?

arnetheduck: consider the case where we're not subscribed to the topic - sending to 0 peers seems a bit harsh then...

Menduist: Right, that won't happen before >50mb, but still good to cover it

arnetheduck:
> The "1 heartbeat" period is used because after that, we will probably be responding to IWANTs

after 1 heartbeat, should we sent an IHAVE to all peers we didn't flood it to?

libp2p/protocols/pubsub/gossipsub.nim
arnetheduck: thought a bit more about this, and I think we want to put the minimum at dmin at least .. 

it feels very risky to send only to one peer - if that peer is slow, it'll delay the message for all peers by a heartbeat _and_ it relies a bit too heavily on the IHAVE/IWANT mechanism to recover, ie there's no redundancy...

I know there's a cost here for sending big messages from slow peers, but I think the risk for normal/highbandwidth peers is more real and costly.

Menduist:
> after 1 heartbeat, should we sent an IHAVE to all peers we didn't flood it to?

Not sure thats wise, most likely they will all reply with IWANTs, and we will be back to square one

libp2p/protocols/pubsub/gossipsub.nim
arnetheduck: this also implies that rather than treating this as a per-send thing, we should have a global rate limiter that triggers for all sends, specially iwants so that we don't succumb to iwant-overload while doing parallel sending here

Menduist: The minimum is currently dLow (it will be caught by the `if peers.len < g.parameters.dLow:` below)

arnetheduck:
> Not sure thats wise, most likely they will all reply with IWANTs, and we will be back to square one

hm, what might make sense here is an exponential increase per heartbeat .. if we send message to 4 peers, we iwant 4 peers at the heartbeat, then 8 and so on for each heartbeat that passes - that should help avoid eclipsing while at the same time allowing the message to spread before

libp2p/protocols/pubsub/gossipsub.nim
arnetheduck: aren't fanout peers populated only when we're not subscribed? also as such, there might not be enough of them in the fanout table either?

Menduist: We also use them when we can't publish to enough peers (ie bad mesh), and they are replenished when < dLow

arnetheduck: > We also use them when we can't publish to enough peers (ie bad mesh), and they are replenished when < dLow


so sending to `g.gossipsub[topic]` as a last resort wouldn't be appropriate?

Menduist: This should be the same as using the fanout, the fanout is only a small cache of `gossipsub` to have more stable diffusion routes, basically

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -488,32 +488,37 @@ method publish*(g: GossipSub,
...
+  # add always direct peers
+  peers.incl(g.explicit.getOrDefault(topic))
+
+  if topic in g.topics: # if we're subscribed use the mesh
+    peers.incl(g.mesh.getOrDefault(topic))
+
   if g.parameters.floodPublish:
+    let
+      msgSize = data.len
+      bandwidth = 25_000_000 #TODO replace with bandwidth estimate
```
arnetheduck, line 500: I suspect we need to make this a (debug) parameter until we have something better - ie in nimbus-eth2, this would be a hidden command-line parameter that we pull out in case of weird behavior - hidden, because obviously it's something that should go away at some point

Menduist:
@diegomrsantos not sure I like the refacto you did, the splits seems a bit random, and the flow became harder to follow IMO

Not sure `publishMessage` is relevant, but we could do something like
```nim
if g.parameters.floodPublish:
  addFloodPublishPeers(g, topic, data)

if peers.len < g.parameters.dLow:
  # not subscribed, or bad mesh, send to fanout peers
  addFanoutPeers(g, topic, peers)
```

diegomrsantos:
@Menduist feel free to remove or modify it.

diegomrsantos (approved):

PR #912, opened 2023-06-09
Add GossipSub ping

Menduist:
The gossipsub ping will be used for various things as part of the https://github.com/status-im/nim-libp2p/issues/850

Introducing it today will enable us to test said optimizations on real networks to make sure they are viable

It is capped to make sure it doesn't introduce any security issues

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -156,6 +156,7 @@ method onNewPeer(g: GossipSub, peer: PubSubPeer) =
     peer.behaviourPenalty = stats.behaviourPenalty
 
   peer.iHaveBudget = IHavePeerBudget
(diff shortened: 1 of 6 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -352,6 +353,9 @@ proc validateAndRelay(g: GossipSub,
 method rpcHandler*(g: GossipSub,
                   peer: PubSubPeer,
                   rpcMsg: RPCMsg) {.async.} =
+  if rpcMsg.ping.len in 1..<64 and peer.pingBudget > 0:
```
diegomrsantos, line 356: How about making `64` a const like `MaxPingLengthInBytes`?

diegomrsantos (approved):

PR #913, opened 2023-06-09
Consider dns as public address

diegomrsantos:


libp2p/wire.nim
```diff
@@ -213,3 +213,13 @@ proc getLocalAddress*(sock: AsyncFD): TransportAddress =
...
+
+proc isPublicMA*(ma: MultiAddress): bool =
+  if DNS.matchPartial(ma):
+    return true
+
+  let hostIP = initTAddress(ma)
+  if hostIP.isOk() and hostIP.get().isGlobal():
+    return true
+
+  return false
```
Menduist, line 225:
```suggestion
  let hostIP = initTAddress(ma).valueOr: return false
  return hostIp.isGlobal()
```

diegomrsantos: There's a typo in `hostIp`, but pushed a commit with it.

libp2p/services/hpservice.nim
```diff
@@ -32,10 +32,10 @@ type
     autonatService: AutonatService
     isPublicIPAddrProc: IsPublicIPAddrProc
 
-  IsPublicIPAddrProc* = proc(ta: TransportAddress): bool {.gcsafe, raises: [].}
+  IsPublicIPAddrProc* = proc(ma: MultiAddress): bool {.gcsafe, raises: [].}
```
Menduist, line 35: Is it still necessary btw? Seems we can now use `/dns/localhost` in test to emulate a public node

Menduist (dismissed):

libp2p/services/hpservice.nim
diegomrsantos: True, but it's a bit cumbersome to set it.

diegomrsantos: Done.

tests/stubs/switchstub.nim
```diff
@@ -26,12 +35,15 @@ method connect*(
  forceDial = false,
  reuseConnection = true,
  upgradeDir = Direction.Out) {.async.} =
-  if (self.connectStub != nil):
-    await self.connectStub()
+  if (self.connectStub != nil) and self.numStubExecutions < self.numStubToExecute:
```
Menduist, line 38: Why is this necessary? This logic could go in the `connectStub`

diegomrsantos: Improved it.

Menduist (approved):

PR #917, opened 2023-06-15
Bumper: fix case where target is up to date

Menduist:
```
Your branch is up to date with 'origin/unstable'.

nothing to commit, working tree clean
Error: Process completed with exit code 1.
```

Can happen if we start the job manually

### DIFF
--- a/.github/workflows/bumper.yml
+++ b/.github/workflows/bumper.yml
@@ -24,13 +24,14 @@ jobs:
           repository: ${{ matrix.target.repo }}
           ref: ${{ matrix.target.branch }}
           path: nbc
-          submodules: true
           fetch-depth: 0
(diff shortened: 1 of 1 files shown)


.github/workflows/bumper.yml
```diff
@@ -24,21 +24,22 @@ jobs:
...
           path: nbc
-          submodules: true
           fetch-depth: 0
           token:  ${{ secrets.ACTIONS_GITHUB_TOKEN }}
 
       - name: Checkout this ref
         run: |
-          cd nbc/vendor/nim-libp2p
+          cd nbc
+          git submodule update vendor/nim-libp2p
```
Menduist, line 33:
```suggestion
          git submodule update --init vendor/nim-libp2p
```

PR #920, opened 2023-06-23
GossipSub Traffic scoring

diegomrsantos:
The main idea here is to eventually disconnect peers that send us too much useless data in a short amount of time (potentially DoSing us). The definition of useless we chose depends if the singing mode is enabled or not. If so, it is everything but the messages themselves. Otherwise, it is everything but the data and the topicIDs inside the message. For now, we just log and measure the number of times peers are above their rate limit - the max amount of useless data they can send us for a given interval.


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -132,8 +135,11 @@ proc handle*(p: PubSubPeer, conn: Connection) {.async.} =
         trace "read data from peer",
           conn, peer = p, closed = conn.closed,
           data = data.shortLog
-
+        # In this way we count even ignored fields by protobuf
+        let msgSize = sizeof(data)
```
Menduist, line 139:
```suggestion
        let msgSize = data.len
```

sizeof is something different:
```nim
echo sizeof(newSeq[byte](10000))
# echoes 16
```

libp2p/protocols/pubsub/gossipsub/scoring.nim
```diff
@@ -244,10 +252,11 @@ proc updateScores*(g: GossipSub) = # avoid async
 
     trace "updated peer's score", peer, score = peer.score, n_topics, is_grafted
 
-    if g.parameters.disconnectBadPeers and stats.score < g.parameters.graylistThreshold and
-        peer.peerId notin g.parameters.directPeers:
-      debug "disconnecting bad score peer", peer, score = peer.score
-      asyncSpawn(g.disconnectPeer(peer))
+    if not g.tryDisconnectBadPeer(peer, stats.score, g.parameters.graylistThreshold):
+      if peer.totalTraffic > 0:
```
Menduist, line 256: So this is only tested during heartbeats?

diegomrsantos: yes, but I think we can do it in `punishInvalidMessage`, right?

diegomrsantos: Even better, maybe we can do it in method `rpcHandler*` implementations.

libp2p/protocols/pubsub/gossipsub/scoring.nim
```diff
@@ -27,6 +28,10 @@ declareGauge(libp2p_gossipsub_peers_score_invalidMessageDeliveries, "Detailed go
 declareGauge(libp2p_gossipsub_peers_score_appScore, "Detailed gossipsub scoring metric", labels = ["agent"])
 declareGauge(libp2p_gossipsub_peers_score_behaviourPenalty, "Detailed gossipsub scoring metric", labels = ["agent"])
 declareGauge(libp2p_gossipsub_peers_score_colocationFactor, "Detailed gossipsub scoring metric", labels = ["agent"])
+declareGauge(libp2p_gossipsub_peers_invalidIgnoredTrafficMB, "Invalid Ignored Traffic (MB)", labels = ["agent"])
```
Menduist, line 31: https://prometheus.io/docs/practices/naming/
Grafana will automatically handle the units if this named properly

libp2p/protocols/pubsub/gossipsub/scoring.nim
```diff
@@ -85,27 +90,48 @@ proc colocationFactor(g: GossipSub, peer: PubSubPeer): float64 =
...
     asyncSpawn(g.disconnectPeer(peer))
+    return true
+
+  return false
+
+proc disconnectIfBadTrafficPeer*(g: GossipSub, peer: PubSubPeer) =
+  if peer.totalTraffic > 0:
+    # dividing in this way to avoid integer overflow
+    let agent = getAgent(peer)
+    let invalidTrafficRatio = float64(peer.invalidTraffic) / float64(peer.totalTraffic)
```
Menduist, line 125: We need to have these value expire at some point (or continuously reduced), otherwise I can be a good peer for a few days, and then use my accumulated score to spam you heavily in a short amount of time

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -65,6 +65,9 @@ type
     maxMessageSize: int
     appScore*: float64 # application specific score
     behaviourPenalty*: float64 # the eventual penalty score
+    totalTraffic*: int64
+    invalidTraffic*: int64 # data that was parsed by protobuf but was invalid
+    invalidIgnoredTraffic*: int64 # data that couldn't be parsed by protobuf
```
Menduist, line 70: My initial proposal just included "usefulTraffic" (ie sum of Valid `Message`) vs "totalTraffic"
What is the benefit of this version? Seems that we are still exposed to spam on the known protobuf fields (ie I could send 1000 subscriptions or whatever)

tests/pubsub/testmessage.nim
```diff
@@ -73,3 +73,13 @@ suite "Message":
     check:
       msgIdResult.isErr
       msgIdResult.error == ValidationResult.Reject
+
+  test "byteSize for Message":
```
lchenut, line 77: I'm pretty sure the test is unfinished.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -370,9 +372,57 @@ proc validateAndRelay(g: GossipSub,
...
+      # debug "Peer disconnected", peer, msgSize, uselessAppBytesNum
+      # raise newException(PeerRateLimitError, "Peer sent too much useless application data and it's above rate limit.")
+
 method rpcHandler*(g: GossipSub,
                   peer: PubSubPeer,
-                  rpcMsg: RPCMsg) {.async.} =
+                  data: seq[byte]) {.async.} =
+  let msgSize = data.len
+  let rpcMsgResult = decodeRpcMsg(data)
+  var rpcMsg = rpcMsgResult.valueOr:
```
lchenut, line 415: To be consistent with floodsub.nim: `var rpcMsg = decodeRpcMsg(data).valueOr:`
... or `let rpcMsgResult = ...` in floodsub.nim, either solution works for me.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -370,9 +372,57 @@ proc validateAndRelay(g: GossipSub,
...
+
+    raise newException(CatchableError, "Peer msg couldn't be decoded")
+
+  let usefulMsgBytesNum =
+    if g.verifySignature:
+      byteSize(rmsg.messages)
+    else:
+      dataAndTopicsIdSize(rmsg.messages)
+
+  var uselessAppBytesNum =  msgSize - usefulMsgBytesNum
```
lchenut, line 398:
```suggestion
  var uselessAppBytesNum = msgSize - usefulMsgBytesNum
```

tests/pubsub/testmessage.nim
diegomrsantos: somehow the check got lot

libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -116,3 +116,86 @@ func shortLog*(m: RPCMsg): auto =
...
+  return total
+
+proc byteSize*(iwant: seq[ControlIWant]): int =
+  var total = 0
+  for item in iwant:
+    for msgId in item.messageIds:
+      total += msgId.len
+  return total
+
+proc `$` (msg: PeerInfoMsg): string =
```
Menduist, line 149: Why is all of this necessary?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -75,7 +76,8 @@ proc init*(_: type[GossipSubParams]): GossipSubParams =
       behaviourPenaltyDecay: 0.999,
       disconnectBadPeers: false,
       enablePX: false,
-      bandwidthEstimatebps: 100_000_000 # 100 Mbps or 12.5 MBps
+      bandwidthEstimatebps: 100_000_000, # 100 Mbps or 12.5 MBps
+      uselessAppBytesRateConfOpt: Opt.none(tuple[bytes: int, interval: Duration])
```
Menduist, line 80: We need a better name for this, overheadRateLimit? Not very pleased by this one either

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -370,9 +372,57 @@ proc validateAndRelay(g: GossipSub,
...
+  # In this way we count even ignored fields by protobuf
+
+  var rmsg = rpcMsgOpt.valueOr:
+    peer.uselessAppBytesRateOpt.withValue(uselessAppBytesRate):
+      if not uselessAppBytesRate.tryConsume(msgSize):
+        libp2p_gossipsub_peers_rate_limit_disconnections.inc(labelValues = [peer.getAgent()]) # let's just measure at the beginning for test purposes.
+        debug "Peer sent a msg that couldn't be decoded and it's above rate limit", peer, uselessAppBytesNum = msgSize
+        # discard g.disconnectPeer(peer)
+        # debug "Peer disconnected", peer, uselessAppBytesNum = msgSize
+        # raise newException(PeerRateLimitError, "Peer sent a msg that couldn't be decoded and it's above rate limit")
```
Menduist, line 388: With this repetition with line 403?

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -107,7 +111,7 @@ func outbound*(p: PubSubPeer): bool =
   else:
     false
 
-proc recvObservers(p: PubSubPeer, msg: var RPCMsg) =
+proc recvObservers*(p: PubSubPeer, msg: var RPCMsg) =
```
Menduist, line 114: We could remove this from the pubsubpeer and directly handle it in pubsub
The reason it was here is just because the decoding happened to be here

libp2p/protocols/pubsub/rpc/messages.nim
diegomrsantos: It isn't. Added to see the actual content of the big messages as it doesn't show in the short log. Can remove it.

libp2p/protocols/pubsub/gossipsub.nim
diegomrsantos: The messages and data logged are slightly different.

libp2p/protocols/pubsub/pubsubpeer.nim
diegomrsantos: Should we also change it ti PubSub here?

```
  PubSubObserver* = ref object
    onRecv*: proc(peer: PubSubPeer; msgs: var RPCMsg) {.gcsafe, raises: [].}
```

libp2p/protocols/pubsub/gossipsub.nim
diegomrsantos: nonEssentialByteRateLimiter?

libp2p/protocols/pubsub/pubsubpeer.nim
Menduist: No, that would break the API (and wouldn't really make sense)

diegomrsantos: This is called in the proc you mentioned.

diegomrsantos: By removing this proc from `PubSubPeer` and directly handling it in PubSub, do you mean getting completely rid of the PubSubpeer instance? The `onRecv` call still requires `PubSubPeer` as an argument.

Menduist: I'm just talking about the observers, you can remove PubSubPeer.observers, and move this proc to PubSub which also holds the observers

diegomrsantos: `observers` is also used in `send`. Should I move everything?

Menduist: Oh, I didn't realize
If not too bloody, yeah

diegomrsantos: We'll work on that in a future PR.

libp2p/protocols/pubsub/rpc/messages.nim
Menduist: Yeah, if not necessary, let's remove it

libp2p/protocols/pubsub/gossipsub.nim
Menduist: These are not non essential though, it's more like overhead

diegomrsantos: Used `overheadRateLimit`.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -78,7 +79,8 @@ proc init*(_: type[GossipSubParams]): GossipSubParams =
       disconnectBadPeers: false,
       enablePX: false,
       bandwidthEstimatebps: 100_000_000, # 100 Mbps or 12.5 MBps
-      iwantTimeout: 3 * GossipSubHeartbeatInterval
+      iwantTimeout: 3 * GossipSubHeartbeatInterval,
+      overheadRateLimitConfOpt: Opt.none(tuple[bytes: int, interval: Duration])
```
Menduist, line 83:
```suggestion
      overheadRateLimit: Opt.none(tuple[bytes: int, interval: Duration])
```

diegomrsantos: This is different from the actual rate limit, it is a config. Adding the Opt makes using `withValue` cleaner.

Menduist: Everything in that structure is a config, it's even name GossipSubParams
And since when do we put the type of a variable in its name?

diegomrsantos: Done

libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -116,3 +116,32 @@ func shortLog*(m: RPCMsg): auto =
     messages: mapIt(m.messages, it.shortLog),
     control: m.control.get(ControlMessage()).shortLog
   )
+
+proc byteSize*(msg: Message): int =
```
Menduist, line 120: Same issue as https://github.com/status-im/nim-libp2p/pull/944#discussion_r1322726849, but I guess we can unify that after merging this PR

Menduist (approved):

PR #925, opened 2023-07-04
Perf protocol

lchenut:
First draft of the [perf](https://github.com/libp2p/specs/blob/master/perf/perf.md) protocol.

A few tests are necessary, but overall I think it's correct.


libp2p/protocols/perf.nim
```diff
@@ -0,0 +1,113 @@
...
+  while size > 0:
+    let toWrite = min(size, PerfSize)
+    await conn.write(buf[0..<toWrite])
+    size -= toWrite
+
+  await conn.join()
+  
+  let duration = Moment.now() - start
+  trace "got upload benchmark duration", conn, duration
+  return duration
```
mxinden, line 113: Note that to integrate with https://github.com/libp2p/test-plans/tree/master/perf you would need to support both `--upload-bytes` and `--download-bytes` within a single run. In other words, you would need to be able to both upload and download on one stream.

mxinden (commented):
:rocket: thanks for working on this!

libp2p/protocols/perf.nim
lchenut: Thanks for the input, I'll do it right away.

diegomrsantos:
@lchenut maybe we can slipt it in client and server files?

libp2p/protocols/perf.nim
```diff
@@ -0,0 +1,139 @@
...
+    size -= toWrite
+
+  await conn.close()
+  await conn.join()
+  
+  let duration = Moment.now() - start
+  trace "got upload benchmark duration", conn, duration
+  return duration
+
+proc perf*(conn: Connection, sizeToWrite: uint64 = 0, sizeToRead: uint64 = 0):
```
diegomrsantos, line 115: are you planning to refactor this to use `perfUpload` and `perfDownload`?

lchenut: Hum... I'm planning to refactor `perfUpload` and `perfDownload` to use `perf` actually

diegomrsantos (approved):

mxinden:
:rocket: let me know in case you need any help integrating this into https://github.com/libp2p/test-plans/tree/master/perf.

PR #930, opened 2023-07-10
Fix crash on empty write

Menduist:


libp2p/stream/chronosstream.nim
```diff
@@ -129,7 +129,9 @@ method write*(s: ChronosStream, msg: seq[byte]): Future[void] =
   # drives up memory usage
   if msg.len == 0:
     trace "Empty byte seq, nothing to write"
-    return
+    let fut = newFuture[void]("chronosstream.write.empty")
```
arnetheduck, line 132: fyi this can now be spelled `Future[void].completed("...")`

arnetheduck (approved):

libp2p/stream/chronosstream.nim
arnetheduck: also, maybe it would make sense to reorder this check with the closed on - ie a zero-write to a closed channel should still cause a closed exception

PR #937, opened 2023-07-28
GossipSub: stagger sending using TCP internals

Menduist:
An attempt to implement stagger sending (#850) by accessing the TCP internal data instead of trying to compute the bandwidth ourself.

Turns out, this is (probably) not a good idea.

On the open internet, most routers (>90% according to `An Internet-Wide Analysis of Traffic Policing`) will buffer packets before dropping them when they are overloaded.

Since linux & other bases its default congestion algorithm (cubic for TCP) on packet losses, it means that the bandwidth estimate will be completely wrong if we don't reach the threshold of "dropping packets" in the routers connecting us to the other peer.

So if we send a 1MB at 1gbits to a router limited to 1mbps, he will actually buffer that data, and send it at 1mbps. TCP, seeing no loss, will think it has 1gbits available. We need to send more data than the router can buffer to actually have losses, and for TCP to realize it has less bandwidth than 1gbits. The exception to this is if the network supports [ECN](https://en.wikipedia.org/wiki/Explicit_Congestion_Notification), but it seems it's not widespread enough to rely on it

We would need more measurement on real networks, but from my limited testing, the intermediary buffers are always big enough to store a full block for each peer, ie ethereum is not using enough bandwidth to overload them.

So seems the only way to implement stagger sending is to use a congestion algorithm based on latency instead of loss (like BBR), or re-implement a latency-based bandwidth estimator on top of TCP (like https://github.com/status-im/nim-libp2p/commit/fd6874e6474518a475595260996530dda8a358ff)

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -22,6 +22,7 @@ import ./pubsub,
        ./rpc/[messages, message],
        ../protocol,
        ../../stream/connection,
+       ../../utils/semaphore,
        ../../peerinfo,
        ../../peerid,
        ../../utility,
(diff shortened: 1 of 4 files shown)


Menduist:
This PR is just here for documentation, not meant to be merged

mxinden:
The [libp2p perf dashboard](https://observablehq.com/@libp2p-workspace/performance-dashboard) shows that rust-libp2p can not (even close) exhaust the bandwidth of a larger machine with a single TCP flow. This is documented in https://github.com/libp2p/rust-yamux/issues/162.

I am curious about nim-libp2p. In other words, I wonder whether there are any low-hanging fruits in nim-libp2p itself, before diving deep into the OS networking stack, e.g. depending on a specific congestion control algorithm like BBR(2).

@Menduist did you have a chance to [integrating nim-libp2p with the libp2p performance tooling](https://github.com/libp2p/test-plans/blob/master/perf/README.md#adding-a-new-implementation)? Is there anything I can do to help?

mxinden:
> So here the goal is not to get the highest throughput, but to know what our throughput is.

Especially since I would expect throughput to vary with each destination and vary with competing network traffic, thus in time, knowing or estimating ones throughput seems impossible.

mxinden:
> Also, ethereum network still uses "dumb" mplex, which should be a lot better than yamux at filling the TCP pipe

At filling it, yes. At reducing the amount of data in-flight, i.e. in buffers, no, given that it does not implement per-stream backpressure.

> Regarding the performance dashboard, @lchenut is on it, but he is also on holidays :D

That is good to hear.

Menduist:
> knowing or estimating ones throughput seems impossible

Well, that "impossible task" is achieved by TCP/QUIC/whatever every time you connect to a website, so it should be possible :stuck_out_tongue:
The goal here is just to create a super simple congestion control system, doesn't have to be perfect

The first prototype of "stagger sending" is simply waiting for an ACK of a peer before relaying to the next one, and its already showing great results

> At reducing the amount of data in-flight, i.e. in buffers, no

Yes, but the task at hand here is "sending a single message, as fast as possible, on an ~idle connection", no per-stream backpressure required, so mplex will always outperform a naive yamux implementation, or be on par with a fancy one with BDP estimation

PR #941, opened 2023-08-08
Bandwidth estimate as a parameter

diegomrsantos:
This is to replace the hard-coded value in https://github.com/status-im/nim-libp2p/pull/911

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -75,7 +75,7 @@ proc init*(_: type[GossipSubParams]): GossipSubParams =
       behaviourPenaltyDecay: 0.999,
       disconnectBadPeers: false,
       enablePX: false,
-      bandwidthEstimateMbps: 100
+      bandwidthEstimatebps: 100_000_000 # 100 Mbps or 12.5 MBps
     )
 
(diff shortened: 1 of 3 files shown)


lchenut (approved):

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -74,7 +74,8 @@ proc init*(_: type[GossipSubParams]): GossipSubParams =
       behaviourPenaltyWeight: -1.0,
       behaviourPenaltyDecay: 0.999,
       disconnectBadPeers: false,
-      enablePX: false
+      enablePX: false,
+      bandwidthEstimateMbps: 100
```
arnetheduck, line 78: 0 should disable flood-limiting completely

lchenut (approved):

Menduist:
What's blocking this?

libp2p/protocols/pubsub/gossipsub.nim
arnetheduck: I find it more flexible and easier when API:s use basic units rather than pre-baked multiples - there's more flexibility this way for API users (ie imagine working on an embedded system with more aggressive bandwidth constraints).

This is somewhat different from when working with "normal users" or end-user UX.

PR #944, opened 2023-08-21
Split msgs in iwant response if bigger than limit

diegomrsantos:
closes https://github.com/status-im/nim-libp2p/issues/887


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -292,7 +292,26 @@ proc send*(p: PubSubPeer, msg: RPCMsg, anonymize: bool) {.raises: [].} =
     sendMetrics(msg)
     encodeRpcMsg(msg, anonymize)
 
-  asyncSpawn p.sendEncoded(encoded)
+  var sentMessages: seq[RPCMsg]
+  # Check if the encoded message size exceeds the maxMessageSize
+  if encoded.len > p.maxMessageSize:
+    var controlSent = false
+    # Split the RPCMsg into individual messages and send them separately
```
arnetheduck, line 299: this feels like it should be a responsibility of the caller - otherwise, we should change the `send` function tot take a list of `message` instead of a full `RPCMsg` - also, a single message might also be bigger than the limit so the utility of this change seems limited compared to the complexity it brings.

diegomrsantos: I agree. I only did it here to use the encoded message and not add the overhead of calculating the message size. 

If individual messages bigger than the size limit are a common case, she should revisit the message limit. This change allows us to send messages below the size limit that can be sent individually but not as a group.

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -272,7 +272,7 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte]) {.raises: [], async.} =
 
     await conn.close() # This will clean up the send connection
 
-proc send*(p: PubSubPeer, msg: RPCMsg, anonymize: bool) {.raises: [].} =
+proc send*(p: PubSubPeer, msg: RPCMsg, anonymize: bool): seq[RPCMsg] {.raises: [].} =
```
Menduist, line 275: Most of the time, sendEncoded is used directly here:
https://github.com/status-im/nim-libp2p/blob/c6aa085e98e7526cb8d4415cb9a7f886e6dcab30/libp2p/protocols/pubsub/pubsub.nim#L199-L202

So this fix won't be enough

diegomrsantos: I believe the code related to an iwant response uses `send` https://github.com/status-im/nim-libp2p/blob/0cdfb0fedcb5f19199c3f3607dd6ff67740fb937/libp2p/protocols/pubsub/gossipsub.nim#L298.

But I can also move this solution to `sendEncoded`, as it is called by send. Wdyt?

Menduist: Good point, no preference here

diegomrsantos: Btw in the current implementation I split the msgs after `p.sendObservers(mm)` and `sendMetrics(mm)`. Do those procs require splitting too?

arnetheduck: apart from the oddity of abstraction noted below, there is quite a bit of overhead with this solution - assuming the rpcmsg is large, copying it, returning the parts in a seq and so on requires a lot of allocations and deallocations - it worsens the case that is already bad. It is exacerbated by the fact that the rpcmsg type is made up of many small seqs, all of which require separate allocation and copying.

libp2p and in particular the rpcmsg construct and its serialization / deserialization is already the hotspot for gossipsub and nimbus-eth2 - in terms of cpu and memory usage, it is responsible for a significant part of the total usage (something like 20% of all cpu goes to copying rpcmsg around).

The `sendEncoded` optimisation was at the time a significant improvement to this flow, in terms of reducing this overhead.

One way to avoid this effect is to serialize the message piece by piece into a single buffer and remember the boundaries at which each part begins and ends then use those boundaries to split the sending - the way protobuf is constructed, this would work.

arnetheduck: To illustrate this better, here's a flame graph of nimbus in action - the red parts are gossipsub doing operations on rpcmsg and varint - it also has knock-on effects elsewhere because all that memory must be reclaimed by the GC - with better efficiency, the GC has less work to do.

image

Menduist: I agree that I don't see why this proc needs to return the list of RPCMsgs

However the splitting of messages will only happen in a rare case (when replying to an IWANT requesting more messages than can fit in a RPC message), and that case is already safe against DDoS attacks (we only reply to IWANT when we actually sent an IHAVE for this message id to this peer)

So I wouldn't try to make it _that_ optimized

arnetheduck: > will only happen in a rare case

until a new case is added and this function happens to be the simplest one to call.

> So I wouldn't try to make it that optimized

the structural problem here is that it makes a slow case _even slower_ - keeping track of boundaries on the other hand and "sending as you go" doesn't have this effect... ie it would look something like this:
```
var buf: seq[byte]
for submsg in rpcmsg:
  let pos = buf.len
  buf.addEncoded(submsg))
  if overlimit(buf):
    send(buf.slice(0, pos))
    buf.delete(0, pos)
send(buf)
```

diegomrsantos: > I agree that I don't see why this proc needs to return the list of RPCMsgs

It was to test the msgs were split. It's not necessary anymore with the e2e test.

diegomrsantos: > the structural problem here is that it makes a slow case even slower - keeping track of boundaries on the other hand and "sending as you go" doesn't have this effect... ie it would look something like this:

In this solution, each sub-message is sent separately, even when the total message size is below the limit. Is it fine?

arnetheduck: it's not if you look at the loop carefully: it only injects extra sends when the buf is full

diegomrsantos: > libp2p and in particular the rpcmsg construct and its serialization / deserialization is already the hotspot for gossipsub and nimbus-eth2 - in terms of cpu and memory usage, it is responsible for a significant part of the total usage (something like 20% of all cpu goes to copying rpcmsg around).

In the original code, each RPCMsg gets its own new buffer, leading to multiple memory allocations and send operations. Your suggestion uses a single, re-used buffer for all messages, appending to it as needed

It seems this implementation is more focused on sending efficiency and better buffer management rather than avoiding the creation of new RPCMsg objects or serialization. 

What needs to be sent is an `RPCMsg` containing `Messages`. Ideally, we would know the optimal way of splitting an RPCMsg into X new ones where they all are slightly below the max size. But to do that we need to create an `RPCMsg`, add `Messages` and calculate the size. We decided to avoid this additional overhead and just create an `RPCMsg` for each  `Message`. If I understood it correctly your suggestion could optimize the sending of those messages - send them in a batch instead of one by one, but it doesn't seem to improve the creation and serialization of `RPCMsg` messages.

arnetheduck: > But to do that we need to create an RPCMsg, add Messages, and calculate its size.

the size can be pre-computed - this would be the other way to write the above loop (this is what most protobuf encoders do btw - they precompute size then allocate memory in a single shot).

the overhead of `rpcmsg` itself is static so it can easily be accounted for - this would be the other way to implement this, ie precompute sizes and split based on them and the rpcmsg overhead itself.

Efficiency is indeed key in gossipsub because of its multiplicative nature - any overhead quickly explodes as message counts go up - at this point, we should be mindful of it and indeed, one of the best things we could do in terms of overall efficiency of gossipsub (and therefore nimbus) would be to improve the efficiency of rpcmsg encoding and decoding in general.

diegomrsantos: so how about this?

1 - We create a len proc for `RPCMsg` and all other types used within it that don't have one already.
2 - We use those procs to estimate the serialized size of a RPCMsg and split it based on that. Let's say len(serializedRPCMsg) = 1.1 * len(rpcMsg) - guessing a 10% Protobuf overhead.
3 - We send each RPCMsg individually 

A pseud-code below:

```
Function splitRPCMsg(rpcMsg: RPCMsg, maxSize: int) -> List[RPCMsg]:
    Initialize newRPCMsgs as an empty list of RPCMsg objects
    Initialize currentRPCMsg as a new RPCMsg object with the same non-message fields as rpcMsg
    Initialize currentSize = len(currentRPCMsg)  # This includes the size of control and other fields

    For each msg in rpcMsg.messages:
        msgSize = len(msg)

        # Check if adding the next message will exceed maxSize
        If (currentSize + msgSize) * 1.1 > maxSize:
            Add currentRPCMsg to newRPCMsgs
            Initialize currentRPCMsg as a new, empty RPCMsg object
            currentSize = 0

        # Add the message to the current RPCMsg
        Add msg to currentRPCMsg.messages
        currentSize += msgSize

    # Check if there is a non-empty currentRPCMsg left to be added
    If currentSize > 0:
        Add currentRPCMsg to newRPCMsgs

    Return newRPCMsgs
```

arnetheduck: this can be done deterministically, there's code in nim-protobuf-serialization showing how - also, unless there's need for it, better to return the encoded messages (or even better, turn this into an `iterator` which yields the encoded messages so that there is no complex `seq` involved

Menduist: Honestly we don't have to be super precise, even if we just split once we reach `limit / 2` without counting pb overhead, each message would still be >500kb with probably <50 bytes overhead (or 0.01% overhead, that's assuming 1mb max message size)
That's why I also proposed originally to split in the handleIWant directly (or somewhere like that). Even just sending each reply in a separate message wouldn't have crazy overhead

diegomrsantos: We don't use nim-protobuf-serialization in libp2p. Tried importing it and using `computeObjectSize`, but got errors about lack of pragma annotation and so on.

libp2p/protocols/pubsub/pubsubpeer.nim
diegomrsantos: Now we can split the messages before calling `send`. But then we need to create a `seq` to pass it. Wdyt @arnetheduck ?

libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -116,3 +116,36 @@ func shortLog*(m: RPCMsg): auto =
     messages: mapIt(m.messages, it.shortLog),
     control: m.control.get(ControlMessage()).shortLog
   )
+
+proc len(peerInfo: PeerInfoMsg): int =
+  return peerInfo.peerId.len + peerInfo.signedPeerRecord.len
```
arnetheduck, line 121:
```suggestion
  peerInfo.peerId.len + peerInfo.signedPeerRecord.len
```
..

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -272,9 +272,31 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte]) {.raises: [], async.} =
 
     await conn.close() # This will clean up the send connection
 
-proc send*(p: PubSubPeer, msg: RPCMsg, anonymize: bool) {.raises: [].} =
-  trace "sending msg to peer", peer = p, rpcMsg = shortLog(msg)
+iterator splitRPCMsg(peer: PubSubPeer, rpcMsg: RPCMsg, maxSize: int, anonymize: bool): seq[byte] =
+  var currentRPCMsg = RPCMsg(subscriptions: rpcMsg.subscriptions, control: rpcMsg.control,
```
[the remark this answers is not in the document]
diegomrsantos: I believe when no split is needed the iterator isn't called:

```
  if encoded.len > p.maxMessageSize:
    for encodedSplitMsg in splitRPCMsg(p, msg, p.maxMessageSize, anonymize):
      asyncSpawn p.sendEncoded(encodedSplitMsg)
  else:
    # If the message size is within limits, send it as is
    debug "sending msg to peer", peer = p, rpcMsg = shortLog(msg)
    asyncSpawn p.sendEncoded(encoded)
```

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -272,9 +272,30 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte]) {.raises: [], async.} =
...
```
Menduist, line 276: This code (& the len part) seems very prone to failure, if we add a field to RPCMsg, update a field, etc
We should at the very least add a comment to the RPCMsg declaration to warn about that, and ideally find a better solution

It seems that the best solution would be to use nim-protobuf-ser, but we can't yet unfortunately

As an alternative, wouldn't it be simpler to modify `encodeRpcMsg`? It could take a `maxSize`, and return a `seq[seq[byte]]`, or even just concatenate the PBs directly internally (would require some tweaks to the pubsubpeer)

tests/pubsub/utils.nim
```diff
@@ -9,16 +9,39 @@ import chronos, stew/[byteutils, results]
...
+
+  onNewPeer(p, pubSubPeer)
+  pubSubPeer
+
+proc randomPeerId*(): PeerId =
+  try:
+    PeerId.init(PrivateKey.random(ECDSA, rng[]).get()).tryGet()
+  except CatchableError as exc:
+    raise newException(Defect, exc.msg)
+
```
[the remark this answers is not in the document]
diegomrsantos: This already existed, I just moved it to this file.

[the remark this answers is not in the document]
diegomrsantos: I did it cause I tried to use `getPubSubPeer` in my test, but it isn't necessary anymore. I tried to revert it, but I got multiple errors as testgossipinternal calls `onNewPeer` and it works only because gossipsub file was included there. I can no longer do that as I import utils in testgossipinternal now and it causes an error. Yeah, I know, messy.

libp2p/protocols/pubsub/pubsubpeer.nim
diegomrsantos: I think adding the comment is a good and easy-to-do improvement. But not sure about splitting in `encodeRpcMsg`, it seems too much abstraction leakage.

diegomrsantos: Added the comment

Menduist (approved):
Great job!

PR #945, opened 2023-09-01
Add validationStrategy Paralel/Seq + ordered validators

alrevuelta:
Description:
* Uses an `OrderedSet` instead to store gossipsub `validators`, which now guarantees that the insertion order is remembered.
* Adds `validationStrategy` which defines how validators are executed:
  * `Paralel`: It is the previous and default behaviour without any change, ensuring backwards compatibility.
  * `Sequential`: New validation strategy introduced in this PR, which allows to run the validators in sequence, so that validator `n+1` is not executed if `n` was rejected. This can be useful when dealing with "heavy" validators that shouldn't not be executed if a previous one failed. Execution order is the same as "insertion" order.

As a rule of thumb on which strategy to choose:
* `Paralel` may validate messages faster, which can translate into slightly lower network delays. But when i) having some "heavy" validator and ii) under a spamming attack, it can lead to higher mem/cpu usage.
* `Sequential` might be a bit slower, but since it early rejects the messages, we save executing the next validators under.


libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -513,26 +517,39 @@ method removeValidator*(p: PubSub,
...
+          of ValidationStrategy.Paralel:
+            pending.add(validator(topic, message))
+          of ValidationStrategy.Sequential:
+            let validatorRes = await validator(topic, message)
+            # early break on first Reject/Ignore
+            if validatorRes != ValidationResult.Accept:
+              result = validatorRes
+              # TODO: Wrong, actually need to break two loops
+              # but don't get why message.topicIds is a seq. This
+              # will just work with message.topicIds.len = 1
```
alrevuelta, line 539: @Menduist Not sure I understand why message.topicIds is a seq. Guess a message can only come from one topic?

But apart from that, any idea on how to break these two loops in here? In python you have the `for ... else` but unsure in nim.

```
for i in l1:
    for j in l2:
        print(i, j)
        if i == 2 and j == 20:
            break
    else:
        continue
    break
```

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -85,6 +85,9 @@ type
   TopicHandler* {.public.} = proc(topic: string,
                        data: seq[byte]): Future[void] {.gcsafe, raises: [].}
 
+  ValidationStrategy* {.pure, public.} = enum
+    Paralel, Sequential
```
rymnc, line 89:
```suggestion
    Parallel, Sequential
```
typo :)

libp2p/protocols/pubsub/pubsub.nim
Menduist: The specs allow multiple topics per message, but it's a minefield.

You can use a named block that you can break

alrevuelta: ah right, done c1fe209b3dcc10aff362c9ae3a218f5622aaaf5f

libp2p/protocols/pubsub/pubsub.nim
alrevuelta: wops, done c1fe209b3dcc10aff362c9ae3a218f5622aaaf5f

arnetheduck:
this can be implemented client-side by adding a single validator that does sequential validation of several sub validators - this also allows mixed strategies where some validators are parallel and some are serial - as is, a "global" strategy seems suboptimal and not really needed

alrevuelta:
> this can be implemented client-side by adding a single validator that does sequential validation of several sub validators - this also allows mixed strategies where some validators are parallel and some are serial - as is, a "global" strategy seems suboptimal and not really needed

Your call :) We are indeed fixing this by merging validators into one. https://github.com/waku-org/nwaku/pull/1966
But then to be consisetnt with your idea, shouldn't nim-libp2p just allow a single validator? Would break backwards compatibility I guess, but validation is already opinionated.

dryajov:
Agree on `OrderedSet`, validator order should be preserved, I'd keep parallel vs sequential out of the core, the current model allows for either strategy to be implemented externally as @arnetheduck points out.

Menduist:
Though an OrderedSet doesn't mean anything if they are executed in parallel, since they will all be started simultaneously no matter what

I don't have a strong opinion here honestly

arnetheduck:
>  validation is already opinionated.

it models validators as being independent one of the other (as hinted by their API - ie nowhere is a relationship between the validators implied) - the parallelization is merely a "freedom" / optimization gained from modelling them that way and not really an opinion.

>  shouldn't nim-libp2p just allow a single validator

when an application uses gossipsub, that application might have several independent "modules" fed by the same gossipsub instance - to maintain independence, these modules could each add their own validators without knowing of each other (ie imagine an application that has a "plugin" model which enables or disables features based on configuration).

The "independent" validator model permits such applications to be supported easily, and this again is a natural consequence of modelling validators as being independent.

Moving to a single-validator model would not just be a change of validation - it would have effects on instantiation strategies of the gossipsub instance itsel - doable, but certainly a bigger deviation.

arnetheduck:
> Though an OrderedSet doesn't mean anything if they are executed in parallel, since they will all be started simultaneously no matter what

the nuance here is that they are started serially _actually_ - they then might reorder themselves as they `await` different things..

I'm not sure tbh whether it's a good idea or not to rely on this. It feels like an implementation detail more than something to rely upon.

PR #949, opened 2023-09-13
GS: Relay messages to direct peers

Menduist:
Currently, only the published messages are relayed to direct peers
However the spec states that all valid messages should be

Also, forceDial the direct peers to dial them even if we are full

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -614,7 +614,7 @@ method publish*(g: GossipSub,
 proc maintainDirectPeer(g: GossipSub, id: PeerId, addrs: seq[MultiAddress]) {.async.} =
   if id notin g.peers:
     trace "Attempting to dial a direct peer", peer = id
-    if g.switch.connected(id):
+    if g.switch.isConnected(id):
       info "We are connected to a direct peer, but GossipSub isn't setup with him!", id
       return
     try:
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -614,7 +614,7 @@ proc maintainDirectPeer(g: GossipSub, id: PeerId, addrs: seq[MultiAddress]) {.as
   if id notin g.peers:
     trace "Attempting to dial a direct peer", peer = id
     if g.switch.isConnected(id):
-      info "We are connected to a direct peer, but it isn't a GossipSub peer!", id
+      warn "We are connected to a direct peer, but it isn't a GossipSub peer!", id
       return
     try:
       await g.switch.connect(id, addrs, forceDial = true)
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -338,6 +338,9 @@ proc validateAndRelay(g: GossipSub,
       g.floodsub.withValue(t, peers): toSendPeers.incl(peers[])
       g.mesh.withValue(t, peers): toSendPeers.incl(peers[])
 
+      # add direct peers
+      toSendPeers.incl(g.subscribedDirectPeers.getOrDefault(t))
+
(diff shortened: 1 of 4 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -608,11 +612,13 @@ method publish*(g: GossipSub,
   return peers.len
 
 proc maintainDirectPeer(g: GossipSub, id: PeerId, addrs: seq[MultiAddress]) {.async.} =
-  let peer = g.peers.getOrDefault(id)
-  if isNil(peer):
+  if id notin g.peers:
     trace "Attempting to dial a direct peer", peer = id
+    if g.switch.isConnected(id):
+      info "We are connected to a direct peer, but GossipSub isn't setup with him!", id
```
diegomrsantos, line 618:
```suggestion
      info "We are connected to a direct peer, but it isn't a GossipSub peer!", id
```

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -608,11 +612,13 @@ method publish*(g: GossipSub,
...
+      return
     try:
-      await g.switch.connect(id, addrs)
+      await g.switch.connect(id, addrs, forceDial = true)
```
diegomrsantos, line 621: if we are already connected, why are we connecting to it again?

libp2p/protocols/pubsub/gossipsub.nim
diegomrsantos: Btw should we do something more in this case? If it's unexpected, maybe it should be a warn.

Menduist: It's basically a user error, but not critical, so info seems like a good fit

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -106,10 +106,10 @@ proc handleGraft*(g: GossipSub,
     let topic = graft.topicId
     trace "peer grafted topic", peer, topic
 
-    # It is an error to GRAFT on a explicit peer
+    # It is an error to GRAFT on a direct peer
     if peer.peerId in g.parameters.directPeers:
       # receiving a graft from a direct peer should yield a more prominent warning (protocol violation)
-      warn "an explicit peer attempted to graft us, peering agreements should be reciprocal",
+      warn "a direct peer attempted to graft us, peering agreements should be reciprocal",
```
arnetheduck, line 112: this can cause a log spam issue if they keep doing this

Menduist: I think we are supposed to trust them somewhat
The other solution would be to remove them from the trusted set in that case
(descoring isn't an option as we will never kick them due to bad score)

arnetheduck: hm, right about the trust - let's make a comment about it

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -359,6 +359,10 @@ proc validateAndRelay(g: GossipSub,
           break
     toSendPeers.excl(seenPeers)
 
+    # add always direct peers
+    for topic in msg.topicIds:
+      toSendPeers.incl(g.subscribedDirectPeers.getOrDefault(topic))
```
arnetheduck, line 364: I'll assume this gets filtered somewhere so we don't send to peers that have already seen the message?

arnetheduck (approved):

libp2p/protocols/pubsub/gossipsub.nim
Menduist: It wasn't, fixed

libp2p/protocols/pubsub/gossipsub.nim
diegomrsantos: I'd use debug or warn.

diegomrsantos (approved):

PR #959, opened 2023-10-11
Remove ConnManager from Upgrade

Menduist:
This is no longer necessary since #807


diegomrsantos:
should we close it?

Menduist:
Before #807, the ConnManager was required in the upgrade, because it was storing the connection
After 807, the connection is stored elsewhere, so the upgrade doesn't need a reference to the ConnManager

diegomrsantos (approved):

PR #967, opened 2023-10-19
Fix WS transport when the connection aborts

lchenut:
Should fix https://github.com/waku-org/nwaku/issues/2140


Ivansete-status (approved):
It looks reasonable to me, thanks!

Aside from that, this is the second time we have encountered an issue with an exception not being properly captured. The last time I remember it was related to the `TransportOsError' exception.

As a further enhancement, I suggest trying to capture more generic exceptions at the end of the block.
CC: @Menduist

PR #968, opened 2023-10-19
fix: move workflows for Nim Devel and legacy i386 from "Daily"

romanzac:
We have "unstable" as a default branch. Others may expect from it more stability than its name suggest. I propose to make "Daily" workflow more reliable by moving tests with Nim Devel to separate (experimental?) workflow.

### DIFF
--- a/.github/workflows/multi_nim_common.yml
+++ b/.github/workflows/multi_nim_common.yml
@@ -1,0 +1,83 @@
+name: daily-common
+
+on:
+  workflow_call:
+    inputs:
+      nim-branch:
+        description: 'Nim branch'
+        required: true
+        type: string
+      platform:
+        description: 'Platform'
+        required: true
+        type: string
+
+jobs:
+  delete-cache:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: snnaplab/delete-branch-cache-action@v1
+
+  matrix-prep:
+
+  build:
+    needs: delete-cache
+    timeout-minutes: 120
+    strategy:
+      fail-fast: false
+      matrix:
+        target: ${{ fromJSON(inputs.platform) }}
+        branch: ${{ fromJSON(inputs.nim-branch) }}
+        include:
+          - target:
+              os: linux
+            builder: ubuntu-20.04
+            shell: bash
+          - target:
+              os: macos
+            builder: macos-12
+            shell: bash
+          - target:
+              os: windows
+            builder: windows-2019
+            shell: msys2 {0}
+
+    defaults:
+      run:
+        shell: ${{ matrix.shell }}
+
+    name: '${{ matrix.target.os }}-${{ matrix.target.cpu }} (Nim ${{ matrix.branch }})'
+    runs-on: ${{ matrix.builder }}
+    continue-on-error: ${{ matrix.branch == 'devel' || matrix.branch == 'version-2-0' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Nim
+        uses: "./.github/actions/install_nim"
+        with:
+          os: ${{ matrix.target.os }}
+          shell: ${{ matrix.shell }}
+          nim_branch: ${{ matrix.branch }}
+          cpu: ${{ matrix.target.cpu }}
+
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '~1.15.5'
+          cache: false
+
+      - name: Install p2pd
+        run: |
+          V=1 bash scripts/build_p2pd.sh p2pdCache 124530a3
+
+      - name: Run tests
+        run: |
+          nim --version
+          nimble --version
+          nimble install -y --depsOnly
+          NIMFLAGS="${NIMFLAGS} --mm:refc" nimble test
+          if [[ "${{ matrix.branch }}" == "devel" ]]; then
+            echo -e "\nTesting with '--mm:orc':\n"
+            NIMFLAGS="${NIMFLAGS} --mm:orc" nimble test
+          fi
--- a/.github/workflows/multi_nim_devel.yml
+++ b/.github/workflows/multi_nim_devel.yml
@@ -1,4 +1,4 @@
-name: Daily
+name: Nim Devel
 on:
   schedule:
     - cron: "30 6 * * *"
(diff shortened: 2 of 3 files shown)


.github/workflows/multi_nim_devel.yml
```diff
@@ -0,0 +1,79 @@
+name: Daily - Nim Devel
```
diegomrsantos, line 1: This file is basically a duplicate of the previous one?

romanzac: Please have a look also at the first commit. Files may look similar, but "Daily" is looking at present problems and the "Daily - Nim Devel" is looking into future potential problems. This separation would reduce our need for immediate attention. This would give us some mental space to schedule new features development rather than chasing potential issues. I am not the only person in the team, so also depends on you guys which way would you like to go ?

diegomrsantos: The comment was more related to the technical aspect of the proposed change.

romanzac: What would you like to change at the moment ?

diegomrsantos: Avoid the whole duplication of the file.

romanzac: Please have a look at latest commit.

arnetheduck:
`i386` is there to ensure that libp2p continues to work with 32-bit platforms in general (ie arm) - as of the last time someone looked into this, it was the least bad 32-bit option available in CI and it should be maintained as such or replaced by another 32-bit CI.

romanzac:
> `i386` is there to ensure that libp2p continues to work with 32-bit platforms in general (ie arm) - as of the last time someone looked into this, it was the least bad 32-bit option available in CI and it should be maintained as such or replaced by another 32-bit CI.

@arnetheduck This PR will still keep running Daily for i386, just in specialized workflow for legacy platforms. If you could name at least one recent 32bit platform, which has some public adoption, we could definitely discuss to include it in main workflow and provide first grade support. You would probably agree we don't have unlimited resources to support all platforms equally. I would even add, being selective and timely on what we support to which extent could make or break usability of something so essential like p2p library.

arnetheduck:
> If you could name at least one recent 32bit platform

raspberry pi, android.

romanzac:
> > If you could name at least one recent 32bit platform
> 
> raspberry pi, android, openwrt (ie most home wifi routers out there)

Let me discuss those internally within Status. Thanks.

.github/workflows/multi_nim_common.yml
```diff
@@ -0,0 +1,83 @@
...
+
+      - name: Install p2pd
+        run: |
+          V=1 bash scripts/build_p2pd.sh p2pdCache 124530a3
+
+      - name: Run tests
+        run: |
+          nim --version
+          nimble --version
+          nimble install -y --depsOnly
```
diegomrsantos, line 78: why aren't we using `nimble install_pinned` here? @lchenut @Menduist

.github/workflows/multi_nim_common.yml
```diff
@@ -0,0 +1,83 @@
...
+          NIMFLAGS="${NIMFLAGS} --mm:refc" nimble test
+          if [[ "${{ matrix.branch }}" == "devel" ]]; then
+            echo -e "\nTesting with '--mm:orc':\n"
+            NIMFLAGS="${NIMFLAGS} --mm:orc" nimble test
+          fi
```
diegomrsantos, line 83: Add a new line, please.

diegomrsantos (commented):
I think it is good, just added two small comments

.github/workflows/multi_nim_common.yml
romanzac: "nimble install_pinned" will be soon replaced by nimble.lock file which I have demonstrated in discord is working well. We will soon have new branch with other name, which will require some kind of a pinning to be later able to make regular releases.

.github/workflows/multi_nim_common.yml
romanzac: Added in https://github.com/status-im/nim-libp2p/pull/968/commits/c8ad908b5e8834107c7dc3b547637ff10149bb0d

.github/workflows/multi_nim_common.yml
romanzac: I would suggest we leave deps installation as it was previously. To pin or lock deps requires upfront testing, which at the current situation, with so many flaky tests, is not possible right now.

lchenut (dismissed):

diegomrsantos (changes requested):
Please, address my question.

.github/workflows/multi_nim_common.yml
romanzac: > why aren't we using `nimble install_pinned` here? @lchenut @Menduist

I don't know why we are not using 'nimble install_pinned' in actual Daily workflow.

Menduist: The daily workflow is here to test regressions in our dependencies, so it always tests the latest versions of everything

Regarding nimble lock, it didn't exist back when we introduced the pinning system, but would indeed be good to migrate to it if it is stable enough. Also see https://github.com/status-im/nim-libp2p/pull/844

diegomrsantos: It makes sense, but two points:

1.  What is the purpose of multi_nim_common.yml? It's not related to the regular tests in the PRs? Is it only for Daily?
2. Nimble `install` seems not to work properly and I am not sure removing all ranges is an acceptable strategy. See https://github.com/status-im/nim-bearssl/pull/52 and https://github.com/nim-lang/nimble/issues/1162.

romanzac: 1. The purpose of multi_nim.yml ("Daily" workflow) is to cover mainstream platforms Linux, Mac, and Windows on amd64
2. Would it be acceptable for you we can discuss and try to identify "best shot" solution in upcoming PRs ?

diegomrsantos (dismissed):

.github/workflows/multi_nim_common.yml
diegomrsantos: I think it is fine to merge it as it is, but it seems you need to sign your commits.

romanzac: I can sing new commits now, but signing old commits is not working for me yet. Do you have any suggestion what to about it ?

diegomrsantos: I think you can squash the old commits

diegomrsantos (approved):

.github/workflows/multi_nim_common.yml
diegomrsantos: Have you tried something like this https://superuser.com/questions/397149/can-you-gpg-sign-old-commits ?

romanzac: Thanks a lot for advice. I have tried, at the end I had to drop all unverified commits and submit all changes in the new commit. Please have a look.

diegomrsantos (approved):

PR #969, opened 2023-10-23
idontwant with d=dlow-1 and iwant to pull messages!

ufarooqstatus:
Only for large messages. We relay large messages only to dlow-1 peers and send idontwant message to others. missing peers can use iwant to pull the large message

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -46,6 +46,9 @@ declareCounter(libp2p_gossipsub_saved_bytes, "bytes saved by gossipsub optimizat
 declareCounter(libp2p_gossipsub_duplicate, "number of duplicates received")
 declareCounter(libp2p_gossipsub_received, "number of messages received (deduplicated)")
 
+export libp2p_gossipsub_duplicate_during_validation, libp2p_gossipsub_idontwant_saved_messages, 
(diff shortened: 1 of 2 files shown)


arnetheduck:
While this is an interesting idea:

* why `IDONTWANT`? Why not `IHAVE`?
* since sending the large messages takes time, it seems likely that peers will request it in most cases as they receive the notification before they receive the message - are there simulations looking into this effect?

ufarooqstatus:
> * why `IDONTWANT`? Why not `IHAVE`?

IDontWant has two advantages over IHave: 
1) we need to maintain [much smaller cache](https://github.com/vacp2p/nim-libp2p/blob/unstable/libp2p/protocols/pubsub/pubsubpeer.nim#L77).
2) On getting 1st iHave, the receiver generates IWant, although it is very likely that some other peer is already sending the same message

ufarooqstatus:
> * since sending the large messages takes time, it seems likely that peers will request it in most cases as they receive the notification before they receive the message - are there simulations looking into this effect?

[Yes, and limited forwarding with/without IHaves showing better results for messages beyond 600KB](https://www.notion.so/Limited-Forwarding-Can-Reduce-Latency-for-Large-Messages-a1d7797313694f0c856b0330bec77ae0#70563bd453944337b92e87674d5e2055)

PR #985, opened 2023-11-15
fix: doc workflow

diegomrsantos:
Use nim 1.6.x as nimble install_pinned does not work with nim 2.0.0 and nimble 0.14.2


romanzac:
Looks alright to me. We don't have lock file working yet, nimble install_pinned is the option for now.
Is the documentation same when I generate it with Nim 1.6 or Nim 2.0 ?

PR #992, opened 2023-11-21
fix(yamux): yamux uses wrong direction during dcutr

diegomrsantos:
During DCUtR, the listener initiates the direct connection attempt and must connect to the other peer as the incoming upgrade direction. Yamux must use this direction to decide on using odd or even stream IDs, not the underlying connection direction -which in this case is TCP (or another transport) outgoing as the peer is initiating the connection.

This is an unusual case where a peer initiates a connection, but should treat it as incoming. The reason is probably that DCUtR happens when the other remote peer dials this peer relayed address, therefore the original connection was in fact incoming. 

Related:
https://github.com/libp2p/specs/tree/master/yamux
https://github.com/libp2p/specs/blob/master/relay/DCUtR.md


diegomrsantos:
This is replaced by https://github.com/status-im/nim-libp2p/pull/994

PR #1009, opened 2024-01-15
make relayed messages non priority

diegomrsantos:
This PR is still in an exploratory phase and the base idea is to give a lower priority to the messages received from other peers that need to be relayed than the messages published by the peer.


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -245,29 +270,15 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte]) {.async.} =
...
-  except CatchableError as exc: # never cancelled
-    # Because we detach the send call from the currently executing task using
-    # asyncSpawn, no exceptions may leak out of it
-    trace "Unable to send to remote", conn, msg = exc.msg
-    # Next time sendConn is used, it will be have its close flag set and thus
-    # will be recycled
-
-    await conn.close() # This will clean up the send connection
+  if isHighPriority:
+    await p.rpcmessagequeue.priorityQueue.put(msg)
```
arnetheduck, line 274: in general, the queue for priority messages is not needed - one can simply send the message and put its future on a `deque` to achieve the same effect.

For non-priority messages, these are put on a queue like in this PR. The "worker loop" for non-priority messages then becomes:
```nim
sendQueue: var Deque[Future[void]]

proc addSendQueue(msg: Future[void]) = 
  while sendQueue.len > 0 and senQueue[0].finished:
    discard sendQueue.popFirst()
  if not msg.finished: sendQueue.addLast(msg)

proc send(msg): Future[void] = ...

while true:
   let msg = await nonPrioQueue.popFirst()
   while sendQueue.len > 0:
     await sendQueue[^1]
   addSendQueue(send(msg))
```

Among other things, this avoids latency for publishing prio message as well as the additional memory traffic resulting from message copies.

I'm also not sure what `messageAvailable` is doing here - `await nonPriorityQueue.popFirst()` already wakes up whenever there's a message.

diegomrsantos: I think the priority queue is for the case when we're trying to publish messages faster than we can send them. In this case we'll need to either drop or slow down somehow. How are we going to achieve the same without an explicit queue?

diegomrsantos: >  this avoids latency for publishing prio message
Could you please elaborate more on how the current code causes this latency? Currently, we don't send a non-priority message as long as there's a priority message to be sent.

arnetheduck: > I think the priority queue is for the case when we're trying to publish messages faster than we can send them. In this case we'll need to either drop or slow down somehow. How are we going to achieve the same without an explicit queue?

There is still a queue, but it is implicit: it exists in the form of pending futures - similar to how it happens now - `asyncSpawn` puts things in a hidden implicit queue

Pending futures can be cancelled - ie the `deque` I suggest above does exactly that.

arnetheduck: > Could you please elaborate more on how the current code causes this latency? Currently, we don't send a non-priority message as long as there's a priority message to be sent.

In this PR, you copy the message into `priorityQueue` which wakes up `processMessages` which removes the message from the `priorityQueue` and adds it to the (hidden) send queue of the connection - the time it takes to wake up `processMessages` is added latency (and additional copies of the message) - it is at least one iteration of the event poll loop which can take several milliseconds (or worse, if there's a lot of work).

diegomrsantos: ok, thanks for the explanation. How will the new proposed solution prioritize the published msgs over the relayed ones? Will we check the sendQueue for priority messages and send non-priority ones only when this queue is empty?

arnetheduck: yes, that's how it could work - not added in my snippet is that priority messages also get sent via `sendQueue`

diegomrsantos:
```nim
   while sendQueue.len > 0:
     await sendQueue[^1]
     clearSendQueue()
```

seems this will keep sending non-priority messages regardless of how many priority messages are waiting to be sent.

diegomrsantos:
This has been replaced by https://github.com/status-im/nim-libp2p/pull/1015

PR #1015, opened 2024-01-29
feat: message prioritization with immediate peer-published dispatch and queuing for other msgs

diegomrsantos:
**Motivation:** We are assuming the local peer might be spending too much time/bandwidth relaying messages rather than publishing its own messages, which could increase latency for the latter.

**Objective:** To enhance efficiency in message processing by prioritizing immediate dispatch of peer-published messages, and managing other messages through a queuing system.

**Changes:**

**Immediate Dispatch for Peer-Published Messages:** Ensures top priority by immediately sending messages published by the peer.
**Queued Handling for IWANT Replies and Relay Messages:** Introduces a unified queuing system for both replies to "IWANT" messages and messages from other peers that need relaying. These are processed after peer-published messages.
**Asynchronous Task Queue Tracking:** Implements an additional queue to monitor asynchronous tasks associated with each published message. The queue for IWANT replies and relay messages is processed once this is cleared.

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -306,6 +306,7 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
     g.send(
       peer,
       RPCMsg(control: some(respControl)), true)
+    # iwant replies have lower priority
     g.send(
       peer,
       RPCMsg(messages: messages), false)
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -220,6 +220,8 @@ method unsubscribePeer*(g: GossipSub, peer: PeerId) =
     for topic, info in stats[].topicInfos.mpairs:
       info.firstMessageDeliveries = 0
 
+  pubSubPeer.stopSendNonPriorityTask()
+
   procCall FloodSub(g).unsubscribePeer(peer)
 
 proc handleSubscribe*(g: GossipSub,
@@ -303,7 +305,10 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
     trace "sending control message", msg = shortLog(respControl), peer
     g.send(
       peer,
-      RPCMsg(control: some(respControl), messages: messages))
+      RPCMsg(control: some(respControl)), true)
+    g.send(
+      peer,
+      RPCMsg(messages: messages), false)
 
 proc validateAndRelay(g: GossipSub,
                       msg: Message,
@@ -370,7 +375,7 @@ proc validateAndRelay(g: GossipSub,
 
     # In theory, if topics are the same in all messages, we could batch - we'd
     # also have to be careful to only include validated messages
-    g.broadcast(toSendPeers, RPCMsg(messages: @[msg]))
+    g.broadcast(toSendPeers, RPCMsg(messages: @[msg]), false)
     trace "forwarded message to peers", peers = toSendPeers.len, msgId, peer
     for topic in msg.topicIds:
       if topic notin g.topics: continue
@@ -441,7 +446,7 @@ method rpcHandler*(g: GossipSub,
   peer.recvObservers(rpcMsg)
 
   if rpcMsg.ping.len in 1..<64 and peer.pingBudget > 0:
-    g.send(peer, RPCMsg(pong: rpcMsg.ping))
+    g.send(peer, RPCMsg(pong: rpcMsg.ping), true)
     peer.pingBudget.dec
   for i in 0..<min(g.topicsHigh, rpcMsg.subscriptions.len):
     template sub: untyped = rpcMsg.subscriptions[i]
@@ -655,7 +660,7 @@ method publish*(g: GossipSub,
 
   g.mcache.put(msgId, msg)
 
-  g.broadcast(peers, RPCMsg(messages: @[msg]))
+  g.broadcast(peers, RPCMsg(messages: @[msg]), true)
 
   if g.knownTopics.contains(topic):
     libp2p_pubsub_messages_published.inc(peers.len.int64, labelValues = [topic])
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -281,12 +281,28 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
   respControl.prune.add(g.handleGraft(peer, control.graft))
   let messages = g.handleIWant(peer, control.iwant)
 
-  if
-    respControl.prune.len > 0 or
-    respControl.iwant.len > 0 or
-    messages.len > 0:
-    # iwant and prunes from here, also messages
+  let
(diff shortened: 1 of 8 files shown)


diegomrsantos:
@kaiserd I tested it running a local Nimbus node on Holenski, the metrics seems fine. But I see huge spikes in the non-priority queue size (can be hundreds of thousands for some nodes. I'm subscribing to all subnets to have enough traffic. It'd be good to test with a validator node as well.

diegomrsantos:
My idea with this PR was to make the difference in the implementation more explicit and potentially testing using each one easier. I'm gonna squash the commits here and close the original PR later.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -303,7 +305,11 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
     trace "sending control message", msg = shortLog(respControl), peer
     g.send(
       peer,
-      RPCMsg(control: some(respControl), messages: messages))
+      RPCMsg(control: some(respControl)), true)
+    # iwant replies have lower priority
```
arnetheduck, line 309: there might not be any messages or controls to send, ie either of these may be empty.

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -262,6 +275,27 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte]) {.async.} =
 
     await conn.close() # This will clean up the send connection
 
+proc sendEncoded*(p: PubSubPeer, msg: seq[byte], isHighPriority: bool = false) {.async.} =
```
arnetheduck, line 278: the `async` here introduces an additional copy of `msg` and is redundant given that all callers `asyncSpawn` the send more or less..

In general, the consequence of using `asyncSpawn` is that `await p.rpcmessagequeue.nonPriorityQueue.addLast(msg)` loses its meaning and effectively disables any limits placed on the queue (because all the spawned tasks will be stuck here).

Given that the policy of the queue is time-based (and not slot-based), it's better to make this function non-async and instead use `addLastNoWait` - it's guaranteed not to raise if the queue is unbounded.

To understand why `send` is non-blocking (not async), we need to go back to #342 and look at how it avoids a deadlock between sending and waiting for data

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -330,6 +364,45 @@ proc canAskIWant*(p: PubSubPeer, msgId: MessageId): bool =
...
+      libp2p_gossipsub_priority_queue_size.dec(labelValues = [$p.peerId])
+    discard p.rpcmessagequeue.sendPriorityQueue.popFirst()
+
+proc sendNonPriorityTask(p: PubSubPeer) {.async.} =
+  while true:
+     while p.rpcmessagequeue.sendPriorityQueue.len > 0:
+       await p.rpcmessagequeue.sendPriorityQueue[0]
+       p.clearSendPriorityQueue()
+     # we send non-priority messages only if there are no pending priority messages
+     let msg = await p.rpcmessagequeue.nonPriorityQueue.popFirst()
```
arnetheduck, line 379: while waiting here, a priority message might have been added - this await should be at the top of the while

libp2p/protocols/pubsub/gossipsub.nim
diegomrsantos: Won't line 284 handle that?

arnetheduck: 284 is an `or` condition

diegomrsantos: yeah, seems to be correct now.

libp2p/protocols/pubsub/pubsubpeer.nim
diegomrsantos: But then we won't clear the priority queue if there's no non-priority msg. Actually, it's true for both solutions.

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -262,6 +275,27 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte]) {.async.} =
...
+  if msg.len <= 0:
+    debug "empty message, skipping", p, msg = shortLog(msg)
+    return
+
+  if msg.len > p.maxMessageSize:
+    info "trying to send a msg too big for pubsub", maxSize=p.maxMessageSize, msgSize=msg.len
+    return
+
+  if isHighPriority:
+    p.rpcmessagequeue.sendPriorityQueue.addLast(p.sendMsg(msg))
```
arnetheduck, line 290: `sendMsg` might return a finished future

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -330,6 +364,45 @@ proc canAskIWant*(p: PubSubPeer, msgId: MessageId): bool =
...
+proc clearSendPriorityQueue(p: PubSubPeer) =
+  while p.rpcmessagequeue.sendPriorityQueue.len > 0 and p.rpcmessagequeue.sendPriorityQueue[0].finished:
+    when defined(libp2p_expensive_metrics):
+      libp2p_gossipsub_priority_queue_size.dec(labelValues = [$p.peerId])
+    discard p.rpcmessagequeue.sendPriorityQueue.popFirst()
+
+proc sendNonPriorityTask(p: PubSubPeer) {.async.} =
+  while true:
+     while p.rpcmessagequeue.sendPriorityQueue.len > 0:
+       await p.rpcmessagequeue.sendPriorityQueue[0]
```
arnetheduck, line 376: it's better to wait for the last item - if you have 5 items here that are sent in order, you will necessarily have to wait for all of them.

also, the queue should be cleared before await so that you don't await a finished future unnecessarily

libp2p/protocols/pubsub/pubsubpeer.nim
arnetheduck: the prio queue can be cleared before adding to it (also)

libp2p/protocols/pubsub/pubsubpeer.nim
diegomrsantos: Could please elaborate on what's the difference between waiting the first or the last? The current approach seems more intuitive to me. 

If we clear first we also might call the clear proc unnecessarily as the first future might not have finished yet.

arnetheduck: the logic should be:
* remove all finished futures from the queue
* if the queue is not empty, wait for the _last_ remaining future
* repeat until empty

this minimizes the number of times we have to wait for something (each wait = performance cost) - we will never wait for a finished future and by waiting for the last one, all that come before it are guaranteed to be finished already (since sends are processed in order).

diegomrsantos: thanks for explaining it, it makes sense.

libp2p/protocols/pubsub/pubsubpeer.nim
diegomrsantos: We `asyncSpawn` `peer.sendEncoded` at pubsub:204

arnetheduck (approved):

PR #1017, opened 2024-01-31
feat: drop msgs to be relayed waiting for too long in the queue 

diegomrsantos:
This PR adds the possibility of configuring the maximum time a message can stay in the non-priority queue and still be considered relevant to be sent to the respective remote peer. If the message has been waiting for this duration or longer at the time is dequeued it will be discarded and not sent.

### DIFF
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -417,7 +417,7 @@ proc stopSendNonPriorityTask*(p: PubSubPeer) =
       libp2p_gossipsub_priority_queue_size.set(labelValues = [$p.peerId], value = 0)
       libp2p_gossipsub_non_priority_queue_size.set(labelValues = [$p.peerId], value = 0)
 
-proc new(T: typedesc[RpcMessageQueue], maxDurationInNonPriorityQueue = 500.milliseconds): T =
+proc new(T: typedesc[RpcMessageQueue], maxDurationInNonPriorityQueue = 1.seconds): T =
   return T(
     sendPriorityQueue: initDeque[Future[void]](),
     nonPriorityQueue: newAsyncQueue[Ttlmessage](),
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -372,11 +372,11 @@ proc clearSendPriorityQueue(p: PubSubPeer) =
 
 proc sendNonPriorityTask(p: PubSubPeer) {.async.} =
   while true:
+     # we send non-priority messages only if there are no pending priority messages
+     let msg = await p.rpcmessagequeue.nonPriorityQueue.popFirst()
      while p.rpcmessagequeue.sendPriorityQueue.len > 0:
        await p.rpcmessagequeue.sendPriorityQueue[0]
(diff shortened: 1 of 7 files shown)


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -53,10 +56,16 @@ type
   DropConn* = proc(peer: PubSubPeer) {.gcsafe, raises: [].} # have to pass peer as it's unknown during init
   OnEvent* = proc(peer: PubSubPeer, event: PubSubPeerEvent) {.gcsafe, raises: [].}
 
+  Ttlmessage* = object
+    msg*: seq[byte]
+    ttl*: Moment
```
AlejandroCabeza, line 61: I think `ttl` doesn't fit the usage we're making of it. To me, `ttl` suggests either the amount of time the message is allowed to live (time increment, e.g.: `5s`) or the timestamp when the message should expire.

Given you are saving the message generation timestamp (which I fully agree with), I'd use a different name as that might lead to confusion.

strinnityk (commented):
At first glance it LGTM. 
Maybe it'd be good to take into account Ludovic's opinion, given he's more familiar with the repo?

libp2p/protocols/pubsub/pubsubpeer.nim
diegomrsantos: Right, I think I'll change it to `timestamp`.

AlejandroCabeza: I would argue, though, that "timestamp" can be inferred from the type (`Moment`). I would use something like `created` or `createdAt`, because `timestamp` alone could lead to a missunderstanding between "creation timestamp" and "deletion date".

diegomrsantos: I thought about `created`, but it'd also be misleading, that's not the time when the message was created, but added to the queue. How about: 

```
  QueuedMessage* = object
    msg*: seq[byte]
    addedAt*: Moment
```

AlejandroCabeza: Right, agree. Said that one just because I'm used to the SQL standard. The one you said works perfectly, very explicit :)

kaiserd (commented):
Thanks! Looks good.

Is there a reference as to why 1s was chosen as the default.
How do other implementations / the spec handle this?

If you did not do this yet, I suggest first testing the base PR in a testnet, followed by another run with this PR merged into the base PR [forward-msgs-non-priority-v2](https://github.com/status-im/nim-libp2p/tree/forward-msgs-non-priority-v2)

diegomrsantos:
@kaiserd I've been testing all PRs running a local Nimbus client on Holesky. Everything seems fine. The only implementation with this feature is rust-libp2p and last time I checked they were testing it on a branch https://github.com/sigp/rust-libp2p/pull/555.

They use 500ms as the default value. I tried it first but I saw a large number os messages being dropped. I believe this value is network/application specific and could be hard to guess a default value. 1s was just an initial attempt, I welcome feedback here.

diegomrsantos:
Those are the other features they implemented https://github.com/sigp/lighthouse/issues/4918#issuecomment-1843757454

richard-ramos:
Closing PR as it is stale

PR #1018, opened 2024-01-31
test(flaky): Log checkExpiring failure

AlejandroCabeza:
Changed `check` to `assert` to leverage the capability of returning a message on failure, as `check` does not have that capability (that I know of) and using logs (e.g.: error, debug) was not working for me.


libp2p.nimble
```diff
@@ -12,7 +12,7 @@ requires "nim >= 1.6.0",
          "dnsclient >= 0.3.0 & < 0.4.0",
          "bearssl >= 0.1.4",
          "chronicles >= 0.10.2",
-         "chronos >= 3.0.6",
+         "chronos >= 3.0.6 & < 4.0.0",
```
diegomrsantos, line 15: Please, let's avoid adding unrelated changes to the PRs.

tests/helpers.nim
```diff
@@ -122,7 +122,8 @@ proc checkExpiringInternal(cond: proc(): bool {.raises: [], gcsafe.} ): Future[b
       await sleepAsync(1.millis)
 
 template checkExpiring*(code: untyped): untyped =
-  check await checkExpiringInternal(proc(): bool = code)
+  let result = await checkExpiringInternal(proc(): bool = code)
+  assert result, "[TIMEOUT] Test failed due to the check timeout. Consider adjusting it."
```
diegomrsantos, line 126: I believe this is gonna change the current output that shows the current and expected values.

diegomrsantos: @AlejandroCabeza could you please make sure the current behavior isn't changed? Isn't it possible to add this message to the check proc?

libp2p.nimble
AlejandroCabeza: I'll move it.

tests/helpers.nim
AlejandroCabeza: As far as I know, there's no way to add a message to the `check` template; hence the change. That's why I suggested the possibility of PRing the `unittest` library with said change, in the case somebody's able to implement it.

In any case, and answering your first comment: I think this shouldn't change the output, at least it doesn't seem that way. Having a closer look at `checkExpiring` you can notice the previous check was done over the returning value of `checkExpiringInternal`; which in turn is a simplified `bool`.

I believe the better output message that `check` generally outputs is due to it receving the raw conditions and parsing the syntax tree; which is not possible as this is currently defined. Due to that, I also believe `assert` will not change in any way the behaviour previously defined.

diegomrsantos: You're right, it shouldn't change the current behavior. But it'd probably be better to make `checkExpiring` behave like `check`, if it's possible to do that. Timeout isn't the only possible failure if the conditions are always false. Does it make sense?

AlejandroCabeza: It'd be nice to make it behave like so, but I don't consider myself capable of manipulating Nim's syntax trees just yet.

diegomrsantos: My last comment is wrong. When the conditions are always false, it'll always fail with a timeout, but we won't see what the different values are.

diegomrsantos: we can try to do it in another PR.

diegomrsantos (approved):
Looks good, thanks for that. Maybe only the tittle isn't up-to-date.

PR #1019, opened 2024-02-01
Defining pubsub's max message size as a global const

gabrielmer:
Moving the maximum message size value as a global constant so the value can be accessed and reused across the codebase

### DIFF
--- a/libp2p/protocols/pubsub/pubsub.nim
+++ b/libp2p/protocols/pubsub/pubsub.nim
@@ -46,6 +46,7 @@ logScope:
 const
   KnownLibP2PTopics* {.strdefine.} = ""
   KnownLibP2PTopicsSeq* = KnownLibP2PTopics.toLowerAscii().split(",")
+  MaxMessageSize* = 1024 * 1024 # Maximum message size in bytes
 
 declareGauge(libp2p_pubsub_peers, "pubsub peer instances")
 declareGauge(libp2p_pubsub_topics, "pubsub subscribed topics")
@@ -553,7 +554,7 @@ proc init*[PubParams: object | bool](
   sign: bool = true,
   msgIdProvider: MsgIdProvider = defaultMsgIdProvider,
   subscriptionValidator: SubscriptionValidator = nil,
-  maxMessageSize: int = 1024 * 1024,
+  maxMessageSize: int = MaxMessageSize,
   rng: ref HmacDrbgContext = newRng(),
   parameters: PubParams = false): P
   {.raises: [InitializationError], public.} =


diegomrsantos:
Could you please elaborate on "reused across the codebase"?

gabrielmer:
> Could you please elaborate on "reused across the codebase"?

Yes, in my specific case I want to check before sending a message if it's greater than the allowed limit. The idea is to avoid hardcoding 1024*1024 and referring to a global const, so if that value changes we don't introduce a bug.

diegomrsantos:
This seems to be a parameter and  1024*1024 is the default value. Can you pass the value you want to use instead (that could be  1024*1024 too)?

Ivansete-status (dismissed):
LGTM! Thanks for it! :100: 

@diegomrsantos - this is very interesting from the point of view of nim-waku. From there, we  might need to refer to the default maximum message size considered by `libp2p`. And having this value as a `const` item will help us to be aligned to it. In other words, we cannot use a hard-coded value in the `nwaku` code base, and therefore we need to use that compilation constraint.

Let me elaborate a bit more with an example of using a hard-coded value. Let's imagine that we hard-code the max msg size in the `nwaku` code base to 1MiB. In this situation, let's imagine that the default max libp2p msg size is changed to 512KiB. In that case, the `nwaku` code base will be incurring in a discrepancy because it wouldn't be considering the correct default msg size.

I hope that helps:) If not, I'm happy to comment further.
Cheers

diegomrsantos:
@Ivansete-status thank you for elaborating on it, but I still don't understand why Waku needs to use Gossipsub default max message size in its codebase. Are you happy using whatever value it happens to have? Wouldn't it be better if Waku chose a value (which could be the same as the default) aligned with its needs and use it as the parameter value when creating the Gossipsub instance?

diegomrsantos:
But I don't have a strong opinion on that, feel free to merge it.

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -46,6 +46,7 @@ logScope:
 const
   KnownLibP2PTopics* {.strdefine.} = ""
   KnownLibP2PTopicsSeq* = KnownLibP2PTopics.toLowerAscii().split(",")
+  MaxMessageSize* = 1024 * 1024 # Maximum message size in bytes
```
lchenut, line 49:
```suggestion
  PubsubMaxMessageSize* = 1024 * 1024 # Maximum message size in bytes
```
`MaxMessageSize` is too common, if you want to make it a global, it is not specific enough, in my opinion.

diegomrsantos:
Actually, we also have a `maxMessageSize` and `pubsubpeer.nim` file. I don't know the reason for the duplication, it might be important to analyze it before merging this PR.

libp2p/protocols/pubsub/pubsub.nim
gabrielmer: Just pushed a commit naming it `DefaultPubsubMaxMessageSize` also to make clear that it's the default value and not a hard limit.

It got a little verbose though, let me know if `PubsubMaxMessageSize` is better and I change it

arnetheduck:
er, given this is a default for a runtime value, it doesn't make much sense checking against the constant - nwaku should be checking against the runtime value at the time of instantiation - if nwaku wants to use a constant, it should use its own constant both when instantiating libp2p and nwaku.

arnetheduck:
more broadly, the thing nwaku needs to be wary of is if different nwaku nodes have different max message size values - for example when upgrading from one version to another - such a change generally must see a coordinated release / update

gabrielmer:
> er, given this is a default for a runtime value, it doesn't make much sense checking against the constant - nwaku should be checking against the runtime value at the time of instantiation - if nwaku wants to use a constant, it should use its own constant both when instantiating libp2p and nwaku.

Makes sense! Not sure why in nwaku we use the default value of nim-libp2p and not a value set by us in the first place.
Is it by the assumption that the default value should be a reasonable and well-tested one? @Ivansete-status @jm-clius

jm-clius:
> Not sure why in nwaku we use the default value of nim-libp2p

We don't afaict. Where default values in nim-libp2p is considered reasonable we don't necessarily make this configurable or an explicit constant in Waku, but in the case of max message size it is (and should be) a Waku constant:

Constant here - https://github.com/waku-org/nwaku/blob/master/waku/waku_core/message/default_values.nim#L8
Used to initialise libp2p pubsub here - https://github.com/waku-org/nwaku/blob/e4e147bcbb71366b159d0f2dcfddfe85fa85b1af/waku/waku_relay/protocol.nim#L168

PR #1020, opened 2024-02-01
fix(dependencies): Clamping

AlejandroCabeza:
Clamp and fix dependency versions so as to avoid running into direct inconsistencies.


arnetheduck:
these clampings, like all clampings before them, will break the CI of downstream projects that use nimble to load libp2p. In the present state of affairs, only "end projects" should specify versions, ie applications like waku/nimbus/etc.

libp2p.nimble
```diff
@@ -8,16 +8,16 @@ license       = "MIT"
...
-         "chronicles >= 0.10.2",
-         "chronos >= 3.0.6",
-         "metrics",
-         "secp256k1",
-         "stew#head",
-         "websock",
-         "unittest2"
+         "bearssl >= 0.1.4 & < 1.0.0",
+         "chronicles >= 0.10.2 & < 1.0.0",
+         "chronos >= 3.0.6 & < 4.0.0",
```
arnetheduck, line 15: why is this clamped for example? when going from clamped to non-clamped, you first need to verify that the newer version indeed is a problem (which it is not in several cases here) -  putting random version numbers here is a regression from the status quo

PR #1021, opened 2024-02-01
docs(general): Improve docs

AlejandroCabeza:
Improve documentation related to go-libp2p-daemon building, and nim version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ nimble install libp2p
 You'll find the nim-libp2p documentation [here](https://status-im.github.io/nim-libp2p/docs/).
 
 ### Testing
-Remember you'll need to build the `go-libp2p-daemon` binary to run the `nim-libp2p` tests. They're required for the interop tests.
+Remember you'll need to build the `go-libp2p-daemon` binary to run the `nim-libp2p` tests.
 To do so, please follow the installation instructions in [daemonapi.md](examples/go-daemon/daemonapi.md).
 
 ## Modules
--- a/README.md
+++ b/README.md
@@ -50,8 +50,8 @@ nimble install libp2p
 ## Getting Started
 You'll find the nim-libp2p documentation [here](https://status-im.github.io/nim-libp2p/docs/).
 
-### Go-Libp2p Daemon
-Remember you'll need to build the `go-libp2p-daemon` binary to use `nim-libp2p`.
+### Testing
+Remember you'll need to build the `go-libp2p-daemon` binary to run the `nim-libp2p` tests. They're required for the interop tests.
 To do so, please follow the installation instructions in [daemonapi.md](examples/go-daemon/daemonapi.md).
 
 ## Modules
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -40,18 +41,20 @@ Learn more about libp2p at [**libp2p.io**](https://libp2p.io) and follow libp2p'
...
 nimble install libp2p
 ```
 
 ## Getting Started
 You'll find the nim-libp2p documentation [here](https://status-im.github.io/nim-libp2p/docs/).
 
-**Go Daemon:**
-Please find the installation and usage intructions in [daemonapi.md](examples/go-daemon/daemonapi.md).
+### Go-Libp2p Daemon
+Remember you'll need to build the `go-libp2p-daemon` binary to use `nim-libp2p`.
```
diegomrsantos, line 54: I think `go-libp2p-daemon` is only necessary for running the tests, isn't it?

AlejandroCabeza: Is that so? Why is it needed to run the tests, then?

diegomrsantos: @Menduist told me this once:

> Why do we have testdaemon? Is it necessary?
Currently the only way we have to test pubsub interopability is using the go-daemon, that's why it's still here
Historically, nim-libp2p began has a simple go-daemon wrapper

AlejandroCabeza: Added a note for that.

examples/go-daemon/daemonapi.md
```diff
@@ -1,15 +1,41 @@
...
+If everything goes correctly, the binary (`p2pd`) should be built and placed in the correct directory.
+If you find any issues, please head into our discord and ask for our asistance.
+
+After successfully building the binary, remember to add it to your path so it can be found. You can do that by running:
+```sh
+export PATH="$PATH:$HOME/go/bin"
+```
+> **Tip:** To make this change permanent, add the command above to your `.bashrc` file.
+
+## Manual
```
diegomrsantos, line 38: Could we update the manual part as well? I'm not sure it is correct.

AlejandroCabeza: Could you elaborate on what's wrong? Aside from the first block which relates to "installing nim-libp2p", that obviously doesn't belong here.
I haven't manually installed `go-libp2p-daemon`, so I don't know what's missing/wrong there.

diegomrsantos: `git submodule update --init --recursive` seems it's saying nim-libp2p uses git submodule and I don't think it's true anymore.

`git checkout v0.0.1` I don't think that's the `go-libp2p-daemon` being used now.

AlejandroCabeza: Wouldn't it be better then to just stick to the `script` based installation?

diegomrsantos: I'm fine with removing the manual part.

lchenut (approved):

PR #1023, opened 2024-02-07
improvement: enhanced checkExpiring macro with custom timeout

diegomrsantos:
Changes:

**Customizable Timeout:** The macro now accepts a timeout parameter, allowing users to specify a custom duration for the condition check. This enhances flexibility, especially in scenarios where the default 10-second timeout might not be suitable.

**Improved Condition Handling:** The macro has been upgraded to support multiple conditions, each on a separate line. It behaves like the `check` macro and provides detailed diagnostics by indicating which specific condition was not satisfied and why.


AlejandroCabeza (dismissed):
lgtm, nice qol

lchenut (approved):

PR #1024, opened 2024-02-18
test refc in CI in Nim 2.0 and later

tersec:


tersec:
This should test ORC on 2.0+ too, to be consistent elsewhere, but `refc` is the priority, so it does that.

tersec:
`linux-i386 (Nim version-2-0)` is failling because of https://github.com/nim-lang/Nim/issues/23319 due to
```nim
  let dst: Rect = (x.int, y.int, src.w * scale, src.h * scale)
```
in https://github.com/ftsf/nico/blob/60dc2f34f6d57e1f4bc4a8565a43950dfdc1cfd5/nico.nim#L2220-L2227 which is used as one of the example tests.

diegomrsantos:
we are testing Nim 2.0 on the daily job

tersec:
ok

tersec:
Specifically Nim 2.0 with `refc`? The default is ORC.

PR #1027, opened 2024-02-19
Limited Forwarding (delayed elimination of peers from which message/idontwant is received)

ufarooqstatus:
Remove peers from the relay (ToSendPeers) list, from which message or idontwant is received. We delay this action till sendMsg to increase such peer count.


arnetheduck:
this is a good idea - in fact, it should be extended so that it also skips messages that were sent by the other node "in full" while queued.

Basically, a peer can tell us they have the message already in two ways: by IDONTWANT and by sending the message in its entirety.

The latter is quite possible: because low-prio messages are the ones being forwarded across the network in bursts, it's possible we receive (a duplicate of) it from the peer while sending some other message.

PR #1037, opened 2024-03-01
move `allFutureThrowing` helper to tests

etan-status:
We use a generically seeming helper to help us track three `Future` exceptions throughout the entire codebase. Move the helper to tests where it finds more use. The helper has several design flaws that make it incompatible with `{.async: (raises).}` usage, and has the potential to swallow one of the three exceptions that we track with it.

The semantics should not change in this PR, this is a pure refactoring.

### DIFF
--- a/tests/hole-punching-interop/hole_punching.nim
+++ b/tests/hole-punching-interop/hole_punching.nim
@@ -12,7 +12,7 @@ import ../../libp2p/[builders,
                   protocols/connectivity/autonat/service,
                   protocols/ping]
 import ../stubs/autonatclientstub
-import ../helpers
+import ../errorhelpers
 
 proc createSwitch(r: Relay = nil, hpService: Service = nil): Switch =
   let rng = newRng()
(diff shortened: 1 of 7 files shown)


arnetheduck:
this PR I guess is fine:ish but it does introduce a bit of unnecessary churn - it would be more clear if first, one annotated all `close` functions with `raises` - this would likely obsolete it entirely as most exceptions it protects against actually can't happen.

arnetheduck (dismissed):

tests/hole-punching-interop/hole_punching.nim
```diff
@@ -47,7 +48,7 @@ proc main() {.async.} =
       isListener = getEnv("MODE") == "listen"
       switch = createSwitch(relayClient, hpservice)
       auxSwitch = createSwitch()
-      redisClient = open("redis", 6379.Port)
+      redisClient = open("localhost", 6379.Port)
```
etan-status, line 51:
```suggestion
      redisClient = open("redis", 6379.Port)
```

arnetheduck (approved):

PR #1039, opened 2024-03-02
document known `--mm:orc` crash

etan-status:
To avoid others stumbling across this one, document a known crash when running `nimble test` on macOS with Nim 2.0.2.


etan-status:
- https://github.com/status-im/nim-metrics/issues/79

PR #1040, opened 2024-03-02
define proper parent error type for `YamuxError`

etan-status:
All errors raised by `nim-libp2p` should descend from `LPError`. `YamuxError` was not correctly classified. Proposing it to be a specialized `MuxerError`, in line with `mplex` and `muxer` errors.

### DIFF
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -34,7 +34,7 @@ when defined(libp2p_yamux_metrics):
     buckets = [0.0, 100.0, 250.0, 1000.0, 2000.0, 3200.0, 6400.0, 25600.0, 256000.0]
 
 type
-  YamuxError* = object of CatchableError
+  YamuxError* = object of MuxerError
 
   MsgType = enum
     Data = 0x0


dryajov (approved):

dryajov:
> @etan-status Could you please explain what `LPError` means and why errors should descend from it?

`LPError` is the root object of libp2p's error hierarchy - this has always been it's original intent.

diegomrsantos:
Thanks, I meant literally, what does the name mean?

dryajov:
> Thanks, I meant literally, what does the name mean?

Oh, heh - the `LP` prefix was chosen as a shorthand for libp2p, otherwise we'd be writing something like `Libp2pError` all over the place :)

diegomrsantos:
I see, maybe `P2PError` would be more descriptive?

dryajov:
> I see, maybe `P2PError` would be more descriptive?

well, imo p2p is quite generic and LP is a well established convention in nim libp2p, I personally don't see a need to change it, but I'm also don't feel very strongly about it.

diegomrsantos:
It never crossed my mind that LP means libp2p. I believe a comment or docs would be beneficial.

etan-status:
> @etan-status Could you please explain what `LPError` means and why errors should descend from it?

It's a followup from #582. `LPError` is the root of all `nim-libp2p` originated errors. That convention is mostly followed, with very few individual exceptions that are being addressed. Rationale is so that `{.async: (raises).}` can be restricted to the parent for the outermost layer.

PR #1045, opened 2024-03-02
avoid triple lookup in `m.flushed` yamux table

etan-status:
Using `notin` and then `[header.streamId]` twice results in the table `m.flushed` to be searched three times. The code can be optimized to only a single lookup using `tables.withValue`.

### DIFF
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -513,16 +513,20 @@ method handle*(m: Yamux) {.async.} =
             await newStream.open()
             asyncSpawn m.handleStream(newStream)
         elif header.streamId notin m.channels:
-          if header.streamId notin m.flushed:
-            raise newException(YamuxError, "Unknown stream ID: " & $header.streamId)
-          elif header.msgType == Data:
-            # Flush the data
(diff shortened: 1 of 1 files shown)


dryajov (approved):

libp2p/muxers/yamux/yamux.nim
```diff
@@ -513,16 +513,20 @@ method handle*(m: Yamux) {.async.} =
...
+          m.flushed.withValue(header.streamId, flushed):
+            if header.msgType == Data:
+              flushed[].dec(int(header.length))
+              if flushed[] < 0:
+                raise newException(YamuxError,
+                  "Peer exhausted the recvWindow after reset")
+          do:
+            raise newException(YamuxError,
+              "Unknown stream ID: " & $header.streamId)
+          if header.length > 0:
```
diegomrsantos, line 526: @etan-status this `if` should be inside the `if` at line 518, shouldn't it?

etan-status: You are right, thanks for the thorough review!

- https://github.com/vacp2p/nim-libp2p/pull/1069

Fixed here.

PR #1049, opened 2024-03-02
catch `WebSocketError` in `wstransport`

etan-status:
Ensure all errors raised by `nim-websock` are caught. That library is not yet updated regarding `{.async: (raises).}`, so an extra entry is needed for `CatchableError` to avoid polluting the callstack with it.

### DIFF
--- a/libp2p/transports/wstransport.nim
+++ b/libp2p/transports/wstransport.nim
@@ -63,13 +63,17 @@ template mapExceptions(body: untyped) =
   try:
     body
   except AsyncStreamIncompleteError:
-    raise newLPStreamEOFError()
+    raise newLPStreamIncompleteError()
+  except AsyncStreamLimitError:
+    raise newLPStreamLimitError()
   except AsyncStreamUseClosedError:
     raise newLPStreamEOFError()
   except WSClosedError:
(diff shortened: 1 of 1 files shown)


etan-status:
Also aligned with `chronosstream` exception converter template.

dryajov (approved):

libp2p/transports/wstransport.nim
```diff
@@ -63,13 +63,17 @@ template mapExceptions(body: untyped) =
...
   except AsyncStreamUseClosedError:
     raise newLPStreamEOFError()
   except WSClosedError:
     raise newLPStreamEOFError()
-  except AsyncStreamLimitError:
-    raise newLPStreamLimitError()
+  except WebSocketError:
+    raise newLPStreamEOFError()
+  except CatchableError:
+    raise newLPStreamEOFError()
```
diegomrsantos, line 76: If they all raise the same error, why do we need an individual `except` for each one instead of only one `except CatchableError`?

etan-status: `CatchableError` should eventually go away, once `{.async: (raises).}` annotations are applied to `nim-websock`.

PR #1051, opened 2024-03-04
Send priority with queue fix

arnetheduck:
This is a smaller version of #1028 that focuses on clearing the priority queue at both ends since the completion order of send futures may be reversed.

The PR also introduces a workaround for redundant msg copies when sending encoded messages and ensures that `sendMsg` always returns a future that can be used for backpressure.

### DIFF
--- a/libp2p/protocols/pubsub/pubsub.nim
+++ b/libp2p/protocols/pubsub/pubsub.nim
@@ -138,18 +138,34 @@ method unsubscribePeer*(p: PubSub, peerId: PeerId) {.base, gcsafe.} =
 
   libp2p_pubsub_peers.set(p.peers.len.int64)
 
-proc send*(p: PubSub, peer: PubSubPeer, msg: RPCMsg) {.raises: [].} =
-  ## Attempt to send `msg` to remote peer
+proc send*(p: PubSub, peer: PubSubPeer, msg: RPCMsg, isHighPriority: bool) {.raises: [].} =
+  ## This procedure attempts to send a `msg` (of type `RPCMsg`) to the specified remote peer in the PubSub network.
   ##
+  ## Parameters:
+  ## - `p`: The `PubSub` instance.
+  ## - `peer`: An instance of `PubSubPeer` representing the peer to whom the message should be sent.
+  ## - `msg`: The `RPCMsg` instance that contains the message to be sent.
+  ## - `isHighPriority`: A boolean indicating whether the message should be treated as high priority.
+  ## High priority messages are sent immediately, while low priority messages are queued and sent only after all high
+  ## priority messages have been sent.
 
   trace "sending pubsub message to peer", peer, msg = shortLog(msg)
-  peer.send(msg, p.anonymize)
+  asyncSpawn peer.send(msg, p.anonymize, isHighPriority)
 
 proc broadcast*(
   p: PubSub,
   sendPeers: auto, # Iteratble[PubSubPeer]
-  msg: RPCMsg) {.raises: [].} =
-  ## Attempt to send `msg` to the given peers
+  msg: RPCMsg,
+  isHighPriority: bool) {.raises: [].} =
+  ## This procedure attempts to send a `msg` (of type `RPCMsg`) to a specified group of peers in the PubSub network.
+  ##
+  ## Parameters:
+  ## - `p`: The `PubSub` instance.
+  ## - `sendPeers`: An iterable of `PubSubPeer` instances representing the peers to whom the message should be sent.
+  ## - `msg`: The `RPCMsg` instance that contains the message to be broadcast.
+  ## - `isHighPriority`: A boolean indicating whether the message should be treated as high priority.
+  ## High priority messages are sent immediately, while low priority messages are queued and sent only after all high
+  ## priority messages have been sent.
 
   let npeers = sendPeers.len.int64
   for sub in msg.subscriptions:
@@ -195,19 +211,19 @@ proc broadcast*(
 
   if anyIt(sendPeers, it.hasObservers):
     for peer in sendPeers:
-      p.send(peer, msg)
+      p.send(peer, msg, isHighPriority)
   else:
     # Fast path that only encodes message once
     let encoded = encodeRpcMsg(msg, p.anonymize)
     for peer in sendPeers:
-      asyncSpawn peer.sendEncoded(encoded)
+      asyncSpawn peer.sendEncoded(encoded, isHighPriority)
 
 proc sendSubs*(p: PubSub,
                peer: PubSubPeer,
                topics: openArray[string],
                subscribe: bool) =
   ## send subscriptions to remote peer
-  p.send(peer, RPCMsg.withSubs(topics, subscribe))
+  p.send(peer, RPCMsg.withSubs(topics, subscribe), isHighPriority = true)
 
   for topic in topics:
     if subscribe:
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -251,26 +251,24 @@ template sendMetrics(msg: RPCMsg): untyped =
         libp2p_pubsub_sent_messages.inc(labelValues = [$p.peerId, t])
 
 proc clearSendPriorityQueue(p: PubSubPeer) =
-  while p.rpcmessagequeue.sendPriorityQueue.len > 0 and p.rpcmessagequeue.sendPriorityQueue[0].finished:
-    when defined(libp2p_expensive_metrics):
-      libp2p_gossipsub_priority_queue_size.dec(labelValues = [$p.peerId])
+  if p.rpcmessagequeue.sendPriorityQueue.len == 0:
+    return # fast path
+
+  while p.rpcmessagequeue.sendPriorityQueue.len > 0 and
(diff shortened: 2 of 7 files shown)


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -248,19 +292,55 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte]) {.async.} =
...
 
-    await conn.close() # This will clean up the send connection
+    trace "sending encoded msg to peer", conn, encoded = shortLog(msg)
+    let f = conn.writeLp(msg)
+    if not f.completed():
+      sendMsgContinue(conn, f)
+    else:
+      f
+  else:
+    sendMsgSlow(p, msg)
```
diegomrsantos, line 310: What's the point `sendMsgSlow` and it being called here if `p.sendConn` at this point is nil or closed?

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -248,19 +292,53 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte]) {.async.} =
     debug "No send connection", p, msg = shortLog(msg)
     return
 
-  trace "sending encoded msgs to peer", conn, encoded = shortLog(msg)
+  trace "sending encoded msg to peer", conn, encoded = shortLog(msg)
+  await sendMsgContinue(conn, conn.writeLp(msg))
```
diegomrsantos, line 296: should we use this same pattern here?
```

    let f = conn.writeLp(msg)
    if not f.completed():
      sendMsgContinue(conn, f)
    else:
      f
```

we could also move this if/else inside `sendMsgContinue`.

arnetheduck: the reason for doing the f.completed check is to avoid an allocation that happens inside `sendMsgContinue` - the effect here would be negligible because only a small subset of writes actually go to the slow version (I hesitated doing it for the hot path even - it doesn't _greatly_ matter but on the other hand, there's a lot of sendMsg calls).

arnetheduck: > move this if/else inside

inside, it doesn't matter any more - the code generated by chronos is reasonably efficient once you get past the initial bump of calling an `{.async.}` function

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -322,13 +322,11 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte], isHighPriority: bool): Future[v
 
   if msg.len <= 0:
     debug "empty message, skipping", p, msg = shortLog(msg)
-    return
-
-  if msg.len > p.maxMessageSize:
+    Future[void].completed()
```
diegomrsantos, line 325: why is this necessary?

arnetheduck: the function must return a Future on all return paths (it's no longer `async`, so there's no "implicit" creation of the future)

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -330,6 +417,43 @@ proc canAskIWant*(p: PubSubPeer, msgId: MessageId): bool =
...
+  debug "starting sendNonPriorityTask", p
+  if p.rpcmessagequeue.sendNonPriorityTask.isNil:
+    p.rpcmessagequeue.sendNonPriorityTask = p.sendNonPriorityTask()
+
+proc stopSendNonPriorityTask*(p: PubSubPeer) =
+  if not p.rpcmessagequeue.sendNonPriorityTask.isNil:
+    debug "stopping sendNonPriorityTask", p
+    p.rpcmessagequeue.sendNonPriorityTask.cancel()
+    p.rpcmessagequeue.sendNonPriorityTask = nil
+    p.rpcmessagequeue.sendPriorityQueue.clear()
```
diegomrsantos, line 445: Is it necessary to do this before the `clear`?

```
for f in p.rpcmessagequeue.sendPriorityQueue:
      f.cancel()
```

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -330,6 +417,43 @@ proc canAskIWant*(p: PubSubPeer, msgId: MessageId): bool =
       return true
   return false
 
+proc sendNonPriorityTask(p: PubSubPeer) {.async.} =
+  while true:
+     # we send non-priority messages only if there are no pending priority messages
+     let msg = await p.rpcmessagequeue.nonPriorityQueue.popFirst()
+     while p.rpcmessagequeue.sendPriorityQueue.len > 0:
+       p.clearSendPriorityQueue()
+       # this minimizes the number of times we have to wait for something (each wait = performance cost)
```
diegomrsantos, line 426: I guess we need to remove this comment

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -330,6 +417,43 @@ proc canAskIWant*(p: PubSubPeer, msgId: MessageId): bool =
...
+       # we will never wait for a finished future and by waiting for the last one, all that come before it are guaranteed
```
diegomrsantos, line 427: I guess we need to remove or change this comment

libp2p/protocols/pubsub/pubsubpeer.nim
arnetheduck: `fail` would be appropriate - `cancel` should generally only be called by the caller that monitors the future.

arnetheduck: actually, scratch that - neither cancel nor fail are correct here, in a "middle" layer - it's the responsibility of the creator of the future to complete/fail it and the responsibility of the "outermost" owner to cancel it - here, we're just a middleman observing its progress

libp2p/protocols/pubsub/pubsubpeer.nim
arnetheduck: 43d41f8c84cefc89d45f533f10be0cced1dd554a

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -227,17 +250,38 @@ template sendMetrics(msg: RPCMsg): untyped =
...
 
-  if msg.len > p.maxMessageSize:
-    info "trying to send a msg too big for pubsub", maxSize=p.maxMessageSize, msgSize=msg.len
-    return
+  while p.rpcmessagequeue.sendPriorityQueue.len > 0 and
+      p.rpcmessagequeue.sendPriorityQueue[^1].finished:
+    discard p.rpcmessagequeue.sendPriorityQueue.popLast()
+
+  when defined(libp2p_expensive_metrics):
+    libp2p_gossipsub_priority_queue_size.set(labelValues = [$p.peerId])
```
diegomrsantos, line 267: This doesnt compile when the flag is enabled

libp2p/protocols/pubsub/pubsubpeer.nim
diegomrsantos: yeah makes sense.

libp2p/protocols/pubsub/pubsubpeer.nim
arnetheduck: 26a37e389f3a6ed1f4e22b4934331ca47ba40960 - might be a good idea to enable these flags in CI builds

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -317,11 +407,11 @@ proc send*(p: PubSubPeer, msg: RPCMsg, anonymize: bool) {.raises: [].} =
 
   if encoded.len > p.maxMessageSize and msg.messages.len > 1:
     for encodedSplitMsg in splitRPCMsg(p, msg, p.maxMessageSize, anonymize):
-      asyncSpawn p.sendEncoded(encodedSplitMsg)
+      await p.sendEncoded(encodedSplitMsg, isHighPriority)
```
diegomrsantos, line 410: do we want to change it to `await`?

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -138,18 +138,34 @@ method unsubscribePeer*(p: PubSub, peerId: PeerId) {.base, gcsafe.} =
...
+  ## - `p`: The `PubSub` instance.
+  ## - `peer`: An instance of `PubSubPeer` representing the peer to whom the message should be sent.
+  ## - `msg`: The `RPCMsg` instance that contains the message to be sent.
+  ## - `isHighPriority`: A boolean indicating whether the message should be treated as high priority.
+  ## High priority messages are sent immediately, while low priority messages are queued and sent only after all high
+  ## priority messages have been sent.
 
   trace "sending pubsub message to peer", peer, msg = shortLog(msg)
-  peer.send(msg, p.anonymize)
+  asyncSpawn peer.send(msg, p.anonymize, isHighPriority)
```
diegomrsantos, line 153: do we want to add it here? I reverted these changes here https://github.com/status-im/nim-libp2p/pull/1028/commits/d1f38c5a1126ab97deae3d2499b6f8658d573c42

diegomrsantos (approved):

Issue #962, opened 2023-10-12
Async exception tracking


Menduist:
Once https://github.com/status-im/nim-chronos/pull/251 is merged, and some time has passed, we should start to integrate it in libp2p API

This will require to think long and hard about how we handle exceptions, and will probably catch a lot of exception bugs

While exception tracking can mostly be added in a backward-compatible manner, that is not the case for async callbacks (for instance, gossipsub validators). We can either breack backward compatibility (requires a 2.0), or just keep the callbacks unchecked (seems reasonable)

Also keep in mind that once exception tracking is in, adding new exceptions will break compatbilitly

arnetheduck:
> Also keep in mind that once exception tracking is in, adding new exceptions will break compatbilitly

Already breaks compatibility if raising a new type (`except` changes meaning in calling code), albeit at runtime instead of compile-time.

arnetheduck:
async raises have been merged in chronos, so this is now unblocked

etan-status:
Have started annotating `{.async: (raises).}`.

For callbacks, there's a way to allow `handler =` and `new` to still accept old-style callbacks while also accept ones that are tagged with `(raises)` by wrapping the new ones with the old one. That doesn't need a 2.0.

vladopajic:
all `async` have specified raised exceptions. multiple PRs in this effort have mentioned this issue (as shown above).

PR #1058, opened 2024-03-06
`{.async: (raises).}` for `relay/utils.nim`

etan-status:
Annotate `relay/utils` module with `{.async: (raises).}`. These build on top of `writeLp` and can be updated independently of the rest of libp2p.


arnetheduck (approved):

etan-status:

PR #1059, opened 2024-03-06
`{.async: (raises).}` annotations for `protocols/secure`

etan-status:
Extend files in `libp2p/protocols/secure` with `{.async: (raises).}` annotations and also apply some minor cleanups such as removing unnecessary `result` usage, using `min` template and reordering `secio` variables to be closer to their usage, changing `var` to `let` etc.

This is somewhat intertwined due to `method handshake*` which spans across all three files and must be updated in all files simultaneously.

### DIFF
--- a/libp2p/protocols/secure/noise.nim
+++ b/libp2p/protocols/secure/noise.nim
@@ -181,8 +181,7 @@ proc init(_: type[SymmetricState]): SymmetricState =
   result.cs = CipherState(k: EmptyKey)
 
 proc mixKey(ss: var SymmetricState, ikm: ChaChaPolyKey) =
-  var
-    temp_keys: array[2, ChaChaPolyKey]
+  var temp_keys: array[2, ChaChaPolyKey]
   sha256.hkdf(ss.ck, ikm, [], temp_keys)
   ss.ck = temp_keys[0]
   ss.cs = CipherState(k: temp_keys[1])
@@ -198,8 +197,7 @@ proc mixHash(ss: var SymmetricState, data: openArray[byte]) =
 
 # We might use this for other handshake patterns/tokens
 proc mixKeyAndHash(ss: var SymmetricState, ikm: openArray[byte]) {.used.} =
-  var
-    temp_keys: array[3, ChaChaPolyKey]
+  var temp_keys: array[3, ChaChaPolyKey]
   sha256.hkdf(ss.ck, ikm, [], temp_keys)
   ss.ck = temp_keys[0]
   ss.mixHash(temp_keys[1])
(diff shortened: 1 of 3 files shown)


libp2p/protocols/secure/noise.nim
```diff
@@ -509,21 +530,28 @@ method write*(
...
 
   # https://github.com/libp2p/specs/tree/master/noise#libp2p-data-in-handshake-messages
-  let
-    signedPayload = p.localPrivateKey.sign(
-      PayloadString & p.noiseKeys.publicKey.getBytes).tryGet()
+  let signedPayload = p.localPrivateKey.sign(
+    PayloadString & p.noiseKeys.publicKey.getBytes)
+  if signedPayload.isErr():
+    raise newException(NoiseHandshakeError,
+      "Failed to sign public key: " & $signedPayload.error())
```
etan-status, line 549: The `ResultError[T]` from `tryGet` is now wrapped into `NoiseHandshakeError`. Similarly for other `tryGet` in `secio`. Conceptually, it makes sense to restrict `handshake` to `LPStreamError` (and descendants).

libp2p/protocols/secure/secio.nim
```diff
@@ -259,17 +260,16 @@ method write*(
     await sconn.stream.write(msg)
     sconn.activity = true
 
-proc newSecioConn(conn: Connection,
-                  hash: string,
-                  cipher: string,
-                  secrets: Secret,
-                  order: int,
-                  remotePubKey: PublicKey): SecioConn
-                  {.raises: [LPError].} =
```
etan-status, line 268: it doesn't actually raise, so use default `{.raises: [].}` from top level `{.push.}`.

arnetheduck (approved):

libp2p/protocols/secure/noise.nim
arnetheduck: technically, there exists `toException` to make `tryGet` raise something else, but it's not a widely used mechanism.. https://github.com/arnetheduck/nim-results?tab=readme-ov-file#example

etan-status: Like, one of the motivations of annotating libp2p with `{.async: (raises).}` is to have a chance to get rid of `toException` converting every single `cstring/string` of whoever pulls in libp2p as a dependency. Once the annotations are done, `toException` can be replaced with explicit converter `mapErr(lpErr)` at the call sites that still need it. Doing it blindly is not possible though, as there are 400 callers and not all of them need it.

etan-status: It's triggered by @tersec request of me adding the `{.push raises: [].}` to test_keymanager_api so that he doesn't get hit by random linter error when bumping references / doing mechanical work on it. I noticed that in `test_keymanager_api` there are random functions that start requiring `{.raises: [LPError].}` because the implicit converter gets pulled in.

In `nimbus-eth2` we also have `tryGet` for downloading genesis states, but luckily they have a `except: CatchableError` that would also catch the `LPError` instead of `ResultError[cstring]`.

etan-status: https://status-im.github.io/nim-style-guide/language.converters.html

> Surprising conversions lead to ambiguous calls

arnetheduck (approved):

etan-status:

Issue #581, opened 2021-06-01
"Error: 't' is of type <lent string> which cannot be captured as it would violate memory safety" (Nim-1.4)


stefantalpalaru:
[amd] 14 Tue Jun 01 15:32:23 |/mnt/sde1/storage/nimbus-eth2|
stefan$ rm -rf nimcache; make -j8 QUICK_AND_DIRTY_COMPILER=1 NIM_COMMIT=version-1-4 nimbus_beacon_node 

Warning: to ensure you are building in a supported repo state, please always run "make update" after "git pull"
(it looks like you've forgotten to do this).
This also applies whenever you switch to a new branch or commit, e.g.: whenever you run "git checkout ...".

Building: build/nimbus_beacon_node
[using Nim version version-1-4]
/mnt/sde1/storage/nimbus-eth2/beacon_chain/spec/presets.nim(134, 19) Warning: Deprecated since v1.4; it was more confusing than useful, use `[]=`; add is deprecated [Deprecated]
/mnt/sde1/storage/nimbus-eth2/vendor/nim-libp2p/libp2p/protocols/pubsub/gossipsub/scoring.nim(311, 34) Error: 't' is of type <lent string> which cannot be captured as it would violate memory safety, declared here: /mnt/sde1/storage/nimbus-eth2/vendor/nim-libp2p/libp2p/protocols/pubsub/gossipsub/scoring.nim(304, 7); using '-d:nimWorkaround14447' helps in some cases
make: *** [Makefile:311: nimbus_beacon_node] Error 1

---

Command executed in this branch: https://github.com/status-im/nimbus-eth2/pull/2594

PR #1061, opened 2024-03-07
fix(issue-1052): Single topic for RPC Message

AlejandroCabeza:
## Description
Rename RPC's Message `topicIds` to `topicId`.
Please, do look carefully into this one as I'm not sure how a well covered by tests is the repo, and my changes could be breaking stuff.

# Issue
Closes (hopefully): https://github.com/status-im/nim-libp2p/issues/1052

### DIFF
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -148,12 +148,13 @@ method rpcHandler*(f: FloodSub,
       discard
 
     var toSendPeers = initHashSet[PubSubPeer]()
-    for t in msg.topicIds:                     # for every topic in the message
-      if t notin f.topics:
-        continue
-      f.floodsub.withValue(t, peers): toSendPeers.incl(peers[])
+    let topic = msg.topicId
+    if topic notin f.topics:
+      continue
+    f.floodsub.withValue(topic, peers):
+      toSendPeers.incl(peers[])
 
-      await handleData(f, t, msg.data)
+    await handleData(f, topic, msg.data)
 
     # In theory, if topics are the same in all messages, we could batch - we'd
     # also have to be careful to only include validated messages
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -148,7 +148,7 @@ method rpcHandler*(f: FloodSub,
       discard
 
     var toSendPeers = initHashSet[PubSubPeer]()
-    let topic = msg.topicId
+    let topic = msg.topic
     if topic notin f.topics:
       continue
     f.floodsub.withValue(topic, peers):
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -150,7 +150,9 @@ method rpcHandler*(f: FloodSub,
     var toSendPeers = initHashSet[PubSubPeer]()
     let topic = msg.topic
     if topic notin f.topics:
+      debug "Dropping message due to topic not in floodsub topics", topic, msgId, peer
       continue
+
     f.floodsub.withValue(topic, peers):
       toSendPeers.incl(peers[])
 
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -292,8 +292,8 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
 
     if isPruneNotEmpty:
       for prune in respControl.prune:
-        if g.knownTopics.contains(prune.topic):
-          libp2p_pubsub_broadcast_prune.inc(labelValues = [prune.topic])
+        if g.knownTopics.contains(prune.topicID):
+          libp2p_pubsub_broadcast_prune.inc(labelValues = [prune.topicID])
         else:
           libp2p_pubsub_broadcast_prune.inc(labelValues = ["generic"])
 
@@ -557,7 +557,7 @@ method onTopicSubscription*(g: GossipSub, topic: string, subscribed: bool) =
     # in the topic
     let msg = RPCMsg(control: some(ControlMessage(
           prune: @[ControlPrune(
-            topic: topic,
+            topicID: topic,
             peers: g.peerExchangeList(topic),
             backoff: g.parameters.unsubscribeBackoff.seconds.uint64)])))
     g.broadcast(mpeers, msg, isHighPriority = true)
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -14,19 +14,20 @@
 import std/[sets, sequtils]
 import chronos, chronicles, metrics
 import chronos/ratelimit
-import ./pubsub,
-       ./floodsub,
-       ./pubsubpeer,
-       ./peertable,
-       ./mcache,
-       ./timedcache,
-       ./rpc/[messages, message, protobuf],
-       ../protocol,
-       ../../stream/connection,
-       ../../peerinfo,
-       ../../peerid,
-       ../../utility,
-       ../../switch
+import
+  ./pubsub,
+  ./floodsub,
+  ./pubsubpeer,
+  ./peertable,
+  ./mcache,
+  ./timedcache,
+  ./rpc/[messages, message, protobuf],
+  ../protocol,
+  ../../stream/connection,
+  ../../peerinfo,
+  ../../peerid,
+  ../../utility,
+  ../../switch
 
 import stew/results
 export results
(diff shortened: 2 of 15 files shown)


libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -27,52 +27,52 @@ proc expectedFields[T](t: typedesc[T], existingFieldNames: seq[string]) {.raises
...
+    subscribe*: bool
+    topic*: string
+
+  MessageId* = seq[byte]
+
+  Message* = object
+    fromPeer*: PeerId
+    data*: seq[byte]
+    seqno*: seq[byte]
+    topicId*: string
```
arnetheduck, line 44: good opportunity to line up the field name with the spec protobuf name

libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -113,7 +113,7 @@ func shortLog*(msg: Message): auto =
     fromPeer: msg.fromPeer.shortLog,
     data: msg.data.shortLog,
     seqno: msg.seqno.shortLog,
-    topicIds: $msg.topicIds,
+    topicId: $msg.topicId,
```
lchenut, line 116: I'm nitpicking, but `$` isn't necessary anymore, topicId is already a string.

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -148,12 +148,13 @@ method rpcHandler*(f: FloodSub,
...
     var toSendPeers = initHashSet[PubSubPeer]()
-    for t in msg.topicIds:                     # for every topic in the message
-      if t notin f.topics:
-        continue
-      f.floodsub.withValue(t, peers): toSendPeers.incl(peers[])
+    let topic = msg.topicId
+    if topic notin f.topics:
+      continue
+    f.floodsub.withValue(topic, peers):
+      toSendPeers.incl(peers[])
```
lchenut, line 155: Hum... it looks wrong to me. The `continue` was for the loop you removed. Something like that should fix it.
```suggestion
    if topic in f.topics:
      f.floodsub.withValue(topic, peers):
        toSendPeers.incl(peers[])
      await handleData(f, topic, msg.data)
```

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -344,18 +344,18 @@ proc validateAndRelay(g: GossipSub,
     # store in cache only after validation
     g.mcache.put(msgId, msg)
 
-    g.rewardDelivered(peer, msg.topicIds, true)
+    let topic = msg.topicId
+    g.rewardDelivered(peer, [topic], true)
```
lchenut, line 348: It might be interesting to change `rewardDelivered` to only take one topics instead of an openarray. Since it's used only on this file.

libp2p/protocols/pubsub/floodsub.nim
lchenut: Nevermind, `f.broadcast` will do nothing if `topic` isn't in `f.topics`

libp2p/protocols/pubsub/rpc/messages.nim
AlejandroCabeza: Do nitpick, that's what we're here for :)

libp2p/protocols/pubsub/gossipsub.nim
AlejandroCabeza: I did think on changing this, but for some reason didn't check whether it was called somewhere else and then I forgot 🤦‍♀️

If you guys think this is not going to be called with multiple topics then I'm all for updating it. Do confirm, please :)

arnetheduck: +1 to change

arnetheduck:
one thing missing here is that the code must check that there is (at least) one topic when decoding the message from bytes

libp2p/protocols/pubsub/gossipsub/scoring.nim
```diff
@@ -21,16 +21,56 @@ import ../pubsub
...
-declareCounter(libp2p_gossipsub_bad_score_disconnection, "the number of peers disconnected by gossipsub", labels = ["agent"])
-declareGauge(libp2p_gossipsub_peers_score_firstMessageDeliveries, "Detailed gossipsub scoring metric", labels = ["agent"])
-declareGauge(libp2p_gossipsub_peers_score_meshMessageDeliveries, "Detailed gossipsub scoring metric", labels = ["agent"])
-declareGauge(libp2p_gossipsub_peers_score_meshFailurePenalty, "Detailed gossipsub scoring metric", labels = ["agent"])
-declareGauge(libp2p_gossipsub_peers_score_invalidMessageDeliveries, "Detailed gossipsub scoring metric", labels = ["agent"])
-declareGauge(libp2p_gossipsub_peers_score_appScore, "Detailed gossipsub scoring metric", labels = ["agent"])
-declareGauge(libp2p_gossipsub_peers_score_behaviourPenalty, "Detailed gossipsub scoring metric", labels = ["agent"])
-declareGauge(libp2p_gossipsub_peers_score_colocationFactor, "Detailed gossipsub scoring metric", labels = ["agent"])
-declarePublicCounter(libp2p_gossipsub_peers_rate_limit_hits, "The number of times peers were above their rate limit", labels = ["agent"])
+declareGauge(
```
arnetheduck, line 24: these (nph-like) changes should probably not be part of this PR

libp2p/protocols/pubsub/rpc/messages.nim
AlejandroCabeza: On the one hand, we can't change `fromPeer` to `from` as the second one is a reserved keyword in nim.

On the other, `go-lib-p2p` does use `Message.topic` ([ref](https://github.com/libp2p/go-libp2p-pubsub/blob/048a4d30d0c3c00829181fd81aa697eb60497448/pb/rpc.proto#L21)), but uses `ControlIHave.topicID` ([ref](https://github.com/libp2p/go-libp2p-pubsub/blob/048a4d30d0c3c00829181fd81aa697eb60497448/pb/rpc.proto#L34)). I don't know which one I like more, but I do know I'd stick to using one and the same for both cases (as we're already doing).

libp2p/protocols/pubsub/gossipsub/scoring.nim
AlejandroCabeza: Yeah, my bad, styling tool in vscode running when it pleases 🤦‍♀️ Reverting

libp2p/protocols/pubsub/rpc/messages.nim
arnetheduck: The canonical names for each field are in the spec: 

https://github.com/libp2p/specs/blob/master/pubsub/README.md#the-message

It doesn't greatly matter what go-libp2p does, it's not authoritative. 

Technically, we can use \`from\` (in backticks), though it's a bit of a bother in nim, so for that field in particular, we have reason to not go with the canonical name).

AlejandroCabeza: I did check the spec, but also double checked what other libraries were doing given the `from` situation (given I wasn't aware of the backtick trick).

I'm fing changing it to `topic` tbh, doing it.

arnetheduck: We will likely move to a model where we generate fields from the `.proto` directly so this is a step in the right direction. It's also useful when using the [protobuf JSON mapping](https://protobuf.dev/programming-guides/proto3/#json) - again, not for today, but good habit in general.

libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -27,52 +27,52 @@ proc expectedFields[T](t: typedesc[T], existingFieldNames: seq[string]) {.raises
...
+
+  ControlMessage* = object
+    ihave*: seq[ControlIHave]
+    iwant*: seq[ControlIWant]
+    graft*: seq[ControlGraft]
+    prune*: seq[ControlPrune]
+    idontwant*: seq[ControlIWant]
+
+  ControlIHave* = object
+    topic*: string
```
arnetheduck, line 56: The gossipsub fields are named different from floodsub (https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md#protobuf)

arnetheduck:
> What's the expected behaviour if there isn't?

penalize the peer for sending an invalid message (same as if signature was bad or it wasn't protobuf at all)

libp2p/protocols/pubsub/rpc/messages.nim
AlejandroCabeza: Oh, alright. That I did not check, thanks for the heads up.

libp2p/protocols/pubsub/gossipsub/scoring.nim
```diff
@@ -250,43 +250,46 @@ proc punishInvalidMessage*(g: GossipSub, peer: PubSubPeer, msg: Message) {.async
...
+    peer: PubSubPeer,
+    topic: string,
+    first: bool,
+    delay = ZeroDuration,
+) =
+  let t = topic
+  if t notin g.topics:
+    return
+
+  let topic = t
```
lchenut, line 276: These two assignations `let t = topic` and then `let topic = t` confuse me so much, do you have any idea why it was this way?

AlejandroCabeza: No, and when I simplified it it stopped working 😂😂

libp2p/protocols/pubsub/rpc/protobuf.nim
```diff
@@ -288,7 +288,8 @@ proc decodeMessage*(pb: ProtoBuffer): ProtoResult[Message] {.inline.} =
   if ?pb.getField(4, msg.topic):
     trace "decodeMessage: read topic", topic = msg.topic
   else:
-    trace "decodeMessage: topic is missing"
+    error "decodeMessage: topic is required but missing"
```
arnetheduck, line 291: broadly we don't log errors for something that is under the control of a remote peer as this would allow spamming users (who are normally assumed to run on `INF` level) - trace is the right level here (or possibly debug, since it should be uncommon)

AlejandroCabeza: Oh, alright, will get it updated.

libp2p/protocols/pubsub/gossipsub/scoring.nim
lchenut: So I dug a bit into it and there's two things related:
- `let t = topic` : https://github.com/vacp2p/nim-libp2p/issues/581
- `let topic = t` : https://github.com/vacp2p/nim-libp2p/commit/ce42674d80222e5dc6f9c779cef128005d696bf9
Both were issues raised with nim1.4, I can understand why the first hasn't been fixed (or even if it's fixable), but the second seems a bit odd. What happens if you just remove the second assignation?

AlejandroCabeza: I ended up removing the two "unnecessary" assigments and tests seemed to pass locally. I did that once previously and tests failed, so... I don't know. Let's see what the CI says.

AlejandroCabeza: Passed, amazing. All good then.

libp2p/protocols/pubsub/gossipsub/scoring.nim
```diff
@@ -250,43 +250,43 @@ proc punishInvalidMessage*(g: GossipSub, peer: PubSubPeer, msg: Message) {.async
         await g.disconnectPeer(peer)
         raise newException(PeerRateLimitError, "Peer disconnected because it's above rate limit.")
 
+  let topic = msg.topic
+  # TODO: REVIEW
+  if topic notin g.topics:
```
arnetheduck, line 255: even if the topic is not in topics, `invalidMessageDeliveries` needs to be incremented

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -170,7 +200,9 @@ method onNewPeer*(g: GossipSub, peer: PubSubPeer) =
   peer.iHaveBudget = IHavePeerBudget
   peer.pingBudget = PingsPeerBudget
 
-method onPubSubPeerEvent*(p: GossipSub, peer: PubSubPeer, event: PubSubPeerEvent) {.gcsafe.} =
+method onPubSubPeerEvent*(
```
arnetheduck, line 203: these are unrelated reformattings that should not be part of this PR

diegomrsantos: @AlejandroCabeza could you please revert all formatting changes?

libp2p/protocols/pubsub/floodsub.nim
diegomrsantos: Could you please share why `f.broadcast` won't do anything? Even if it's true, IMO we shouldn't rely on that and change the previous behavior without a good reason for doing so.

lchenut: Because `broadcast` [will update some metrics](https://github.com/vacp2p/nim-libp2p/blob/unstable/libp2p/protocols/pubsub/pubsub.nim#L170-L210) by incrementing by the number of `toSendPeers` which is zero.
Then it iterates through `toSendPeers` [to send data](https://github.com/vacp2p/nim-libp2p/blob/unstable/libp2p/protocols/pubsub/pubsub.nim#L212-L219) (encoded or not) but it is empty so it won't do anything.

diegomrsantos: Thanks, you are right. But it'd still log. Maybe adding a log when `topic` isn't in `f.topics` is enough.

AlejandroCabeza: Added

libp2p/protocols/pubsub/gossipsub/scoring.nim
AlejandroCabeza: But that was not what was happening before:
```nim
for tt in msg.topicIds:
    let t = tt
    if t notin g.topics:
      continue

    let tt = t
    # update stats
    g.withPeerStats(peer.peerId) do (stats: var PeerStats):
      stats.topicInfos.mgetOrPut(tt, TopicInfo()).invalidMessageDeliveries += 1
```
Only topics in `g.topics` are counted in `invalidMessageDeliveries`. 
Was the previous code wrong?

libp2p/protocols/pubsub/gossipsub.nim
AlejandroCabeza: Sure, I'll try to undo the stylings.

diegomrsantos: I think the formatting when the code also changed is fine, I, at least, I was referring to the changes that are purely formatting related.

libp2p/protocols/pubsub/gossipsub/scoring.nim
arnetheduck: you're right - I misread the indent level

arnetheduck: let's get rid of this TODO then

AlejandroCabeza: Done.

lchenut (approved):
LGTM

arnetheduck (approved):

PR #1064, opened 2024-03-11
add support for setting protocol handlers with `{.raises.}` annotation

etan-status:
All of the internal protocol handlers are restricted to `{.raises.}` of `[CancelledError]`. However, the `LPProtoHandler` type is part of the public interface, and example / test code raises more errors than just `[CancelledError]`. The caller is aware of that and `CatchableError` is caught.

To allow annotating the internal handlers with the proper `{.raises`.} annotation, support for an extra `LPProtoHandler2` is added. This is backward compatible as the old `LPProtoHandler` is still allowed. Examples still compile fine. There is one test that needs a slight adjustment as it accesses the internal `handler` field directly. That field needs to be renamed so that the `template` is used instead.

Eventually, `LPProtoHandler` may be phased out, with appropriate notices to users who define custom handlers and the migration requiring errors to be handled inside the handler instead of raising them. At this time, such a deprecation is not yet applied, especially while the internal logic is still relying on the previous handler flow.


libp2p/protocols/protocol.nim
```diff
@@ -19,14 +19,16 @@ const
...
 type
   LPProtoHandler* = proc (
-    conn: Connection,
-    proto: string):
-    Future[void]
-    {.gcsafe, raises: [].}
+      conn: Connection,
+      proto: string): Future[void] {.async.}
+
+  LPProtoHandler2* = proc (
```
arnetheduck, line 25: hm.. 

not that I necessarily disagree with this change, but this becomes an odd case which should have been solved by covariance - ie technically, because we allow _any_ exception to be raised, we should not constrain here to `[CancelledError]`..

I wonder if there's a way to solve this with some generics, ie create a `new` that takes an `new[E](handler: proc(): InternalRaisesFuture[void, E]` - if that's possible, it's probably something we should formalize in chronos somehow, to allow expressing it without referencing `Internal..`

etan-status: sure, it would still involve going through the `new` or `handler=` instead of constructing the object manually. Agree that making it more generic would also serve the role of allowing `{.async: raises.}` annotation, but for now would have to go with `Internal`. I guess that's alright to get started.

libp2p/protocols/protocol.nim
```diff
@@ -22,9 +22,9 @@ type
       conn: Connection,
       proto: string): Future[void] {.async.}
 
-  LPProtoHandler2* = proc (
+  LPProtoHandler2*[E] = proc (
```
arnetheduck, line 25: does `lp2[E] = proc ...: Future[void].Raising(E)` work?

etan-status: the `handler=` implementation worked for me.

I tested it by changing `protocols/connectivity/relay/client.nim` to

```nim
proc handleStream(conn: Connection, proto: string) {.async: (raises: [CancelledError]).} =
  ...
cl.handler = handleStream
```

and `NIMFLAGS="--mm:refc" nimble test` still worked fine (refc because on orc it segfaults).

etan-status: the `proc` takes parameters and is passed around, hose parts would have to be continuously repeated. putting the `[E]` directly on the `proc` doesn't work, only the result type can be done that way.

Exposing `InternalRaisesFuture` in a single place seems like the lesser evil to me, it's trivial to adjust if necessary

libp2p/protocols/protocol.nim
```diff
@@ -19,14 +19,16 @@ const
...
-    conn: Connection,
-    proto: string):
-    Future[void]
-    {.gcsafe, raises: [].}
+      conn: Connection,
+      proto: string): Future[void] {.async.}
+
+  LPProtoHandler2*[E] = proc (
+      conn: Connection,
+      proto: string): InternalRaisesFuture[void, E]
```
arnetheduck, line 27:
```suggestion
  LPProtoHandler2*[E] = proc (
      conn: Connection,
      proto: string): Future[void].Raising(E)
```

this does not work?

libp2p/protocols/protocol.nim
diegomrsantos: IMO adding any `*2` to the codebase isn't a good software engineering practice as it isn't sufficiently descriptive. Also, why is an `Ineternal*` type from another project referenced here?

diegomrsantos: how about `LPProtoHandlerRaising`?

diegomrsantos: if `InternalRaisesFuture` is internal indeed, it shouldn't be referenced outside the original project. If it isn't internal, it should be renamed IMO.

diegomrsantos: I don't know much about the underlying issue, but it feels like a design issue.

etan-status: Agree, hence why I think having the workaround in 1 place is better than spreading it further.

@arnetheduck do you know the background of the `2` naming in chronos? Do you like `...Raising` suffix instead (despite it actually being less raising than the version without `2` which implicitly raises `[CatchableError]`)? Do you think chronos could rename `InternalRaisesFuture` to `RaisesFuture`?

arnetheduck: I don't have any issues with `2` _really_ in this case - it's a new version of the same underlying concept and we're changing an implementation detail more than changing concept so inventing names for it is just convoluted (just like you v2 REST interfaces if you didn't account for something in v1 but haven't changed the _intent_ of the thing). A number signals a progression from worse to better, hopefully, so in new code you know which to use.

In this case, the old name is there just to not break old programs unnecessarily - in private code, we'd just deal with whatever breakage without changing name - adding `Raising` is a path-dependent outcome, not a sensible name that any reasonable programmer would come up with on a green field.

> Do you like ...Raising suffix 

No, it's long, ugly and misleading, specially when `Raising([])`. We should not penalise use of the "better" version of this type in such a way. A higher number at least has a positive connotation in much of society.

> Do you think chronos could rename InternalRaisesFuture to RaisesFuture?

No, we need to pass E through a translation layer every time and future versions might use a different translation that encodes the errors in a different way which would break anything that uses a "naked" `RaisesFuture`. It is deliberately kept this way.

Using `Future[void].Raising([E])` like I suggested _should_ work but there's regression in 2.0 unfortunately: https://github.com/nim-lang/Nim/issues/23432.

I think it's fine to use `InternalRaisesFuture` for the time being until that bug is fixed, annotating the code with a TODO.

libp2p/protocols/protocol.nim
```diff
@@ -19,14 +19,16 @@ const
...
 
   LPProtocol* = ref object of RootObj
     codecs*: seq[string]
-    handler*: LPProtoHandler ## this handler gets invoked by the protocol negotiator
+    handlerImpl: LPProtoHandler  ## invoked by the protocol negotiator
```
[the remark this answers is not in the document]
etan-status: Yes, that's what I mentioned above. It has to be renamed otherwise `handler=` doesn't take precedence, and `handler=` is used a lot, while using `LPProtocol(...)` over `LPProtocol.new` was only used in 1 example and 1 test, where it seems to be not intentional.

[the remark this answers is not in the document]
arnetheduck: breaks nimbus-eth2 though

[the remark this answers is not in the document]
etan-status: 1 liner to fix

libp2p/protocols/protocol.nim
arnetheduck: If we're dead set on a non-numbered name, `LPProtocolHandler` is free

etan-status: I think having different naming schemes is worse than a single consistent one. But up to @diegomrsantos , fine with `LPProtocolHandler` as well (with `LPProtoHandler` being the old one).

diegomrsantos: Maybe the suggestion wasn't good, but if I see `LPProtoHandler` and `LPProtoHandler2`, I have no idea when I should use one or the other without reading the code and inferring what `2` was supposed to mean. Sorry for the argument from authority, but I, particularly, have never seen such naming used in mainstream libraries.

etan-status: Do you like `LPProtocolHandler`? Can switch to that one if you prefer it.

etan-status: Have bumped to `LPProtocolHandler` to see how it looks.

arnetheduck: > I have no idea when I should use one or the other

I can't say that `LPProtocolHandler` vs `LPProtoHandler` solves this problem though. 2 at least tells you which one to use in new code.

diegomrsantos: Could you please add a small comment describing the practical difference from the other handler and/or when it should be used? Btw thank you for working on that.

diegomrsantos: > > I have no idea when I should use one or the other
> 
> I can't say that `LPProtocolHandler` vs `LPProtoHandler` solves this problem though. 2 at least tells you which one to use in new code.

There are only two hard things in Computer Science: cache invalidation and naming things.

-- Phil Karlton

:)

arnetheduck: > have never seen such naming used in mainstream libraries.

The must used and popular api on earth does this, more or less: https://learn.microsoft.com/en-us/windows/win32/api/winver/nf-winver-getfileversioninfoexa (though they call it `Ex` instead of `2`)

I understand your concern, but sometimes, there just isn't any good _naming_ solution and indeed, a comment might be the best option.

etan-status: They also have `Ex2`: https://learn.microsoft.com/en-us/windows/win32/api/vds/nf-vds-ivdsvolumemf3-formatex2

arnetheduck: with a cheery on top for 3 in the interface name ;) in fact, com interfaces / all of .NET _often_ versions itself this way

diegomrsantos: I knew I'd regret the comment, but never worked with MS tho.

libp2p/protocols/protocol.nim
```diff
@@ -19,14 +19,21 @@ const
...
 type
   LPProtoHandler* = proc (
-    conn: Connection,
-    proto: string):
-    Future[void]
-    {.gcsafe, raises: [].}
+      conn: Connection,
+      proto: string): Future[void] {.async.}
+
+  # Callbacks that are annotated with `{.async: (raises).}` explicitly
```
diegomrsantos, line 25: looks great, thanks.

libp2p/protocols/protocol.nim
arnetheduck: So .. thinking a bit more about this, the answer is actually simple: we can remove `LPProtocolHandler`/`LPProtoHandler2` altogether and just use `Future[...].Raising(E)` explicitly in the wrapper, which ensures user code is not polluted with a name that we don't necessarily want to support.

This makes sense also because there as limitations to how you can use aliases together with `` ie `var f: LPProtocolHandler[[ValueError]]` doesn't work because of macro instantiation order issues.

arnetheduck (approved):

PR #1065, opened 2024-03-11
Handle graceful shutdown on TransportUseClosedError.

AlejandroCabeza:
fixes: https://github.com/status-im/nim-libp2p/issues/1007

### DIFF
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -234,6 +234,11 @@ proc upgradeMonitor(
   finally:
     upgrades.release()
 
+proc accept_cleanup(upgrades: AsyncSemaphore, connection: Connection) {.async.} = 
+  upgrades.release() # always release the slot
+  if not isNil(connection):
+    await connection.close()
+
 proc accept(s: Switch, transport: Transport) {.async.} = # noraises
   ## switch accept loop, ran for every transport
   ##
@@ -273,11 +278,12 @@ proc accept(s: Switch, transport: Transport) {.async.} = # noraises
     except CancelledError as exc:
       trace "releasing semaphore on cancellation"
       upgrades.release() # always release the slot
+    except TransportUseClosedError:
+      await accept_cleanup(upgrades, conn)
+      return
     except CatchableError as exc:
       error "Exception in accept loop, exiting", exc = exc.msg
-      upgrades.release() # always release the slot
-      if not isNil(conn):
-        await conn.close()
+      await accept_cleanup(upgrades, conn)
       return
 
 proc stop*(s: Switch) {.async, public.} =


lchenut (dismissed):
LGTM

libp2p/switch.nim
```diff
@@ -234,6 +234,11 @@ proc upgradeMonitor(
   finally:
     upgrades.release()
 
+proc accept_cleanup(upgrades: AsyncSemaphore, connection: Connection) {.async.} =
```
diegomrsantos, line 237: We don't use this pattern, but rather `acceptCleanup`.

diegomrsantos (changes requested):
LGTM, just a minor request. It'd be great if we could add a test too.

libp2p/switch.nim
```diff
@@ -273,11 +278,12 @@ proc accept(s: Switch, transport: Transport) {.async.} = # noraises
     except CancelledError as exc:
       trace "releasing semaphore on cancellation"
       upgrades.release() # always release the slot
+    except TransportUseClosedError:
+      await accept_cleanup(upgrades, conn)
```
diegomrsantos, line 282: maybe we can add a trace log here too

AlejandroCabeza: Done

AlejandroCabeza:
> LGTM, just a minor request. It'd be great if we could add a test too.

That's a good idea, but it might be a tad too hard for me:
Can you point me to an already existing test I can use as template where `accept` is being tested/called? Also, I don't know how to raise the `TransportUseClosedError` which is necessary for the test, or how to mock it, for that matter. 

Otherwise, this might be too time-consuming (for me).

libp2p/switch.nim
```diff
@@ -273,11 +278,13 @@ proc accept(s: Switch, transport: Transport) {.async.} = # noraises
     except CancelledError as exc:
       trace "releasing semaphore on cancellation"
       upgrades.release() # always release the slot
+    except TransportUseClosedError:
+      trace "Graceful shutdown in accept loop, exiting"
```
diegomrsantos, line 282: we can use the same as below `"Exception in accept loop, exiting", exc = exc.msg`

AlejandroCabeza: Don't you think that might trigger the same reactions, from users, that @cheatfate was mentioning in the issue?

diegomrsantos: I think the issue was about the `error` log level. We can keep the one you used, my comment was about the rest.

diegomrsantos:
If testing it is too time consuming, or not even possible, that's fine.

libp2p/switch.nim
diegomrsantos: Users aren't supposed to use `trace` unless they are trying to debug an issue. That was precisely my intention suggesting `"Exception in accept loop, exiting", exc = exc.msg`, adding more useful info to the log.

diegomrsantos: It doesn't need to be exactly this msg, but something more specific about what happened than the one currently used, plus the exception msg.

AlejandroCabeza: Okay, let's see if I understood correctly:
1. When a `TransportUseClosedError` is raised, the whole server should stop?
2. Should the same behaviour happen with other `CatchableError`s as well?
3. What are you referring to with `acceptLoop` the indefinite calls to the `accept` proc? When reading your point 3 it doesn't seem that way, and it confuses me.

diegomrsantos: > So to properly fix this issue you should first break/cancel/stop your loop which perform connection acceptance, and only after that you should close chronos server.

Switch accept loop uses the transport `running` flag, so it's possible that the current accept will run when the transport has been stopped and this will be raised https://github.com/vacp2p/nim-libp2p/blob/9059a8aced87741de98813bfe0ef2ed63e0217ab/libp2p/transports/tcptransport.nim#L203.

AlejandroCabeza: @cheatfate

diegomrsantos: The issue might be fixed by https://github.com/vacp2p/nim-libp2p/pull/1095#discussion_r1593835102. I'm tempted to close this PR.

diegomrsantos: At the moment, I don't know how to reproduce it, if it hasn't been fixed yet.

PR #1066, opened 2024-03-12
`{.async: (raises).}` for `MultistreamSelect`

etan-status:
Annotate `MultistreamSelect` with `{.async: (raises).}` and ensure that `handle` returns a `MultiStreamError` instead of `CatchableError` in the "invalid first message" case.

### DIFF
--- a/libp2p/multistream.nim
+++ b/libp2p/multistream.nim
@@ -260,8 +260,33 @@ proc addHandler*(m: MultistreamSelect,
                                protocol: protocol,
                                match: matcher))
 
-proc start*(m: MultistreamSelect) {.async: (raises: []).} =
-  await noCancel allFutures(m.handlers.mapIt(it.protocol.start()))
+proc start*(m: MultistreamSelect) {.async: (raises: [CancelledError]).} =
+  let
+    handlers = m.handlers
+    futs = handlers.mapIt(it.protocol.start())
+  try:
+    await allFutures(futs)
+    for fut in futs:
+      await fut
+  except CancelledError as exc:
+    var pending: seq[Future[void].Raising([])]
+    for i, fut in futs:
+      if not fut.finished:
+        pending.add noCancel fut.cancelAndWait()
+      elif fut.completed:
+        pending.add handlers[i].protocol.stop()
+      else:
+        static: doAssert typeof(fut).E is (CancelledError,)
+        doAssert fut.cancelled
+    await noCancel allFutures(pending)
+    raise exc
+
 
 proc stop*(m: MultistreamSelect) {.async: (raises: []).} =
-  await noCancel allFutures(m.handlers.mapIt(it.protocol.stop()))
+  # Nim 1.6.18: Using `mapIt` results in a seq of `.Raising([CancelledError])`
+  var futs = newSeqOfCap[Future[void].Raising([])](m.handlers.len)
+  for it in m.handlers:
+    futs.add it.protocol.stop()
+  await noCancel allFutures(futs)
+  for fut in futs:
+    await fut
--- a/libp2p/multistream.nim
+++ b/libp2p/multistream.nim
@@ -277,7 +277,6 @@ proc start*(m: MultistreamSelect) {.async: (raises: [CancelledError]).} =
         pending.add handlers[i].protocol.stop()
       else:
         static: doAssert typeof(fut).E is (CancelledError,)
-        doAssert fut.cancelled
     await noCancel allFutures(pending)
     raise exc
 
--- a/libp2p/multistream.nim
+++ b/libp2p/multistream.nim
@@ -1,5 +1,5 @@
 # Nim-LibP2P
-# Copyright (c) 2023 Status Research & Development GmbH
+# Copyright (c) 2023-2024 Status Research & Development GmbH
 # Licensed under either of
 #  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
 #  * MIT license ([LICENSE-MIT](LICENSE-MIT))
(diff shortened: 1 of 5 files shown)


libp2p/multistream.nim
```diff
@@ -98,19 +101,29 @@ proc select*(_: MultistreamSelect | type MultistreamSelect,
...
 
-proc select*(_: MultistreamSelect | type MultistreamSelect,
-             conn: Connection,
-             proto: string): Future[bool] {.async.} =
+proc select*(
+    _: MultistreamSelect | type MultistreamSelect,
+    conn: Connection,
+    proto: string
+): Future[bool] {.async: (raises: [
+    CancelledError, LPStreamError, MultiStreamError]).} =
```
arnetheduck, line 109: not sure this is the right PR for it, but reraising underlying stream errors represents an abstraction leak - ie they should be caught and rebranded as multistreamerror _really_

etan-status: Agree, but the pattern of re-raising `LPStreamError` applies to quite large parts of the codebase, and the goal for now is to just document the reality with the `{.async: (raises).}`, while fixing the obvious offenders of `raise (ref CatchableError)()`. Semantic changes should be separate.

etan-status: Another precedent we have in `muxer.nim` where the procs currently raise `{.async: (raises: [CancelledError, LPStreamError, MuxerError].}`, it's just how it's done today.

libp2p/multistream.nim
```diff
@@ -242,8 +260,8 @@ proc addHandler*(m: MultistreamSelect,
                                protocol: protocol,
                                match: matcher))
 
-proc start*(m: MultistreamSelect) {.async.} =
-  await allFutures(m.handlers.mapIt(it.protocol.start()))
+proc start*(m: MultistreamSelect) {.async: (raises: []).} =
+  await noCancel allFutures(m.handlers.mapIt(it.protocol.start()))
```
arnetheduck, line 264: hm, this is somewhat unexpected, that protocol start functions shouldn't support cancellation - I'm thinking about the case where for example a protocol might need to perform a dns lookup or the like and we want to exit the process - timeouts in such protocols can be longer than the "cancellation tolerance".

etan-status: Once more it is simply reflecting current behaviour. But agree that having cancellation on `start` feels more intuitive. Have updated the logic to support that, and also to properly process cancellation to avoid inconsistent state.

arnetheduck: When deciding how to constrain raises, it's important to take into account an important aspect of exception-based API: when they raise `CatchableError` it means the calling code must be prepared to handle any error _even if_ de facto there's no raise today - this is an important "feature" of designing exception-based api: you can "future-proof" them by signalling in the signature that you're raising more than you actually are - this is why the strategy of simply "describing current behavior" is .. incomplete: it doesn't take into account the non-code intent that went into the design.

Of course, in this particular case, there was probably no intent or thinking about cancellation at all since the code didn't actually do cancellation correctly until your fix, but when "narrowing" an API to raise less, it's a good question to ask.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -701,30 +701,40 @@ proc maintainDirectPeers(g: GossipSub) {.async.} =
     for id, addrs in g.parameters.directPeers:
       await g.addDirectPeer(id, addrs)
 
-method start*(g: GossipSub) {.async.} =
+method start*(
+    g: GossipSub
+): Future[void] {.async: (raises: [CancelledError], raw: true).} =
```
arnetheduck, line 706: why the raw here? can't see how it matters.

libp2p/multistream.nim
etan-status: Yes, widening it to match the intended design space makes sense.

libp2p/protocols/pubsub/gossipsub.nim
etan-status: not familiar enough to know _if_ it matters, but it makes the flow more explicit. there is no async involved in this method, so if the transformation introduces more complex logic or increases compile time, rather opting out of it

arnetheduck: makes it hard to read though - complexity is negligible in simple cases like this, specially since it's not part of any hot path

etan-status: yeah, that's true, the non-raw version is a bit simpler to read, but is a bit less clear about what's going on / as in, there is no async involved at all in the implementation, it's  just a regular synchronous function.

side note: one of the other PRs actually was triggered by switching some of the redirectors to `raw`, it changed test timing just slightly to make it fail, revealing a small logic issue.

anyway, will keep it in mind going forward and transform to `raw` a bit less agressively. I think one could argue that anytime `newFuture` can be avoided, the readability benefit generally outweighs the optimization aspect; while for redirections / wrappers where `newFuture` is not necessary, the `raw` version looks similar enough to the transformed version that there is no downside to using it?

can do another cleanup pass for readability after the raises annotations are done, depending on the outcome of the discussion. personally I'm also fine with `newFuture`.

arnetheduck: the thing with newFuture is that it opens up two avenues for "classic" bugs: returning nil on some code path _and_ returning an accidentally unfinished future on some code path - it's risky shit, for a dumb start function. ie when reviewing, I had to spend time ascertaining this point (it would have been more clear btw had you put the newfuture call at the end of the function where it gets returned, instead of at the beginning because as is, there's too much unnecessary distance between construction and use).

The most simple things are the most pleasurable to debate ;)

etan-status: True, haven't been bitten by those personally, but can see them happening relatively easily. It's a good point, generally, especially as it seems that the async transformation is not too costly for purely synchronous functions.

etan-status:

PR #1067, opened 2024-03-12
Avoid unnecessary rate limit message copy

arnetheduck:
The way the code is written, a new Opt instance is created for every RPC message processed by gossipsub - this involves copying the full message even though only the size is needed for rate limiting purposes - in Nimbus, this becomes a performance/memory allocation hotspot.


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -398,46 +398,37 @@ proc validateAndRelay(g: GossipSub,
...
-  var rmsg = rpcMsgOpt.valueOr:
-    peer.overheadRateLimitOpt.withValue(overheadRateLimit):
-      if not overheadRateLimit.tryConsume(msgSize):
-        libp2p_gossipsub_peers_rate_limit_hits.inc(labelValues = [peer.getAgent()]) # let's just measure at the beginning for test purposes.
-        debug "Peer sent a msg that couldn't be decoded and it's above rate limit.", peer, uselessAppBytesNum = msgSize
-        if g.parameters.disconnectPeerAboveRateLimit:
-          await g.disconnectPeer(peer)
-          raise newException(PeerRateLimitError, "Peer disconnected because it's above rate limit.")
-
-    raise newException(CatchableError, "Peer msg couldn't be decoded")
```
lchenut, line 413: I'm confused, isn't this part of the code useful? Why remove it?
Other than that, it looks good to me.

arnetheduck: This snippet was for the case where the message is invalid - if you look carefully, the code in this snippet simply uses the `msgSize` in this case, as does the new code here: https://github.com/vacp2p/nim-libp2p/pull/1067/files#diff-abbf987f19a24c9940cd01bde79a1b1e0819f1c3b6d7ba2efa04ad680d022c78R431

arnetheduck: oh, now I see what you mean, there's an exception raised here.. hm, good catch

arnetheduck: e469c14ef6e23be3f2ddf501c4704c945948b17a

lchenut (approved):
LGTM

diegomrsantos:
Thanks for that, looks much better now anyway IMO.

PR #1068, opened 2024-03-12
annotate `upgrademngrs` with `{.async: (raises).}`

etan-status:
The next layer of modules that can get `{.async: (raises).}` annotations are the `upgrademngrs`. This is the first layer that mixes so many error types that it makes sense to start with `{.async: (raises: [LPError]).}` especially because the callers add raw `LPError` as well. It may be worthwhile to try and wrap inner exceptions in a proper outer type, but as this pattern spans large parts of the library, such wrapping should be investigated separately from initial annotation work.

### DIFF
--- a/libp2p/multistream.nim
+++ b/libp2p/multistream.nim
@@ -261,20 +261,22 @@ proc addHandler*(m: MultistreamSelect,
                                match: matcher))
 
 proc start*(m: MultistreamSelect) {.async: (raises: [CancelledError]).} =
-  let
-    handlers = m.handlers
-    futs = handlers.mapIt(it.protocol.start())
+  # Nim 1.6.18: Using `mapIt` results in a seq of `.Raising([])`
+  var futs = newSeqOfCap[Future[void].Raising([CancelledError])](m.handlers.len)
+  for it in m.handlers:
+    futs.add it.protocol.start()
   try:
     await allFutures(futs)
     for fut in futs:
       await fut
   except CancelledError as exc:
     var pending: seq[Future[void].Raising([])]
+    doAssert m.handlers.len == futs.len, "Handlers modified while starting"
     for i, fut in futs:
       if not fut.finished:
         pending.add noCancel fut.cancelAndWait()
       elif fut.completed:
-        pending.add handlers[i].protocol.stop()
+        pending.add m.handlers[i].protocol.stop()
       else:
         static: doAssert typeof(fut).E is (CancelledError,)
     await noCancel allFutures(pending)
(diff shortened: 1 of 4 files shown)


libp2p/multistream.nim
```diff
@@ -261,20 +261,22 @@ proc addHandler*(m: MultistreamSelect,
                                match: matcher))
 
 proc start*(m: MultistreamSelect) {.async: (raises: [CancelledError]).} =
-  let
-    handlers = m.handlers
-    futs = handlers.mapIt(it.protocol.start())
+  # Nim 1.6.18: Using `mapIt` results in a seq of `.Raising([])`
```
arnetheduck, line 264: er .. that doesn't make sense at all - should result in `.Raising([CancelledError])` since that's what start does now..
```nim
import chronos, sequtils

proc x(v: int) {.async: (raises: [CancelledError]).} = discard

var s = mapIt([1, 2, 3], x(it))

static: echo type(s)
```
nim prints:
```
seq[InternalRaisesFuture[system.void, (CancelledError,)]]
```

diegomrsantos:
@etan-status it'd be good to track what has been done regarding this topic in this issue https://github.com/vacp2p/nim-libp2p/issues/962

libp2p/multistream.nim
etan-status: Also, interestingly, in `stop` there is the reverse problem where `mapIt` invents a random `.Raising([CancelledError])`

etan-status:
Thanks, have put comments linking to that issue to the relevant PRs so far, so that they appear in #962

libp2p/multistream.nim
arnetheduck:
```nim
import chronos, sequtils

proc start(v: int) {.async: (raises: [CancelledError]).} = discard
proc stop(v: int) {.async: (raises: []).} = discard

var a = mapIt([1, 2, 3], start(it))

static: echo type(a)
var b = mapIt([1, 2, 3], stop(it))

static: echo type(b)
```
funny bunny.

```
seq[InternalRaisesFuture[system.void, (CancelledError,)]]
seq[InternalRaisesFuture[system.void, (CancelledError,)]]
```

etan-status: oh, nice that you could find a more minimal version that's actually postable to nim-lang :) less painful than expected. have run into these separately, in different PRs. 

`stop` was fixed here: 

- https://github.com/vacp2p/nim-libp2p/pull/1066

While `start` gets fixed in the current PR.

Funny that they interfere with each other, didn't expect there to be a link between `start` and `stop`.

Could it be a name clash thing with `chronos`?

arnetheduck: https://github.com/nim-lang/Nim/issues/23418

arnetheduck: https://github.com/status-im/nim-chronos/pull/522

PR #1073, opened 2024-03-19
Improve memory efficiency of seen cache

arnetheduck:
The `seen` cache currently is a significant memory usage hotspot due to its inefficient implementation: for every entry, two copies of the message id + timing data + `seq` overhead causes it to use much more memory than it has to.

In addition, each check involves several layers of allocations as the computed message id gets salted.

This PR improves on the situation by:

* using a hash of the message id with the salt instead of joining strings
* computing the salted id only once per message
* storing one digest instead of two message id:s


arnetheduck:
On holesky, this PR reduces memory usage of the seen cache by ~100mb

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -32,20 +33,27 @@ const FloodSubCodec* = "/floodsub/1.0.0"
...
-    seen*: TimedCache[MessageId] # message id:s already seen on the network
-    seenSalt*: seq[byte]
+    seen*: TimedCache[SaltedId] # message id:s already seen on the network
+    seenSalt*: sha256
+      # The salt in this case is a partially updated SHA256 context pre-seeded
+      # with some random data
 
-proc hasSeen*(f: FloodSub, msgId: MessageId): bool =
-  f.seenSalt & msgId in f.seen
+proc salt*(f: FloodSub, msgId: MessageId): SaltedId =
```
diegomrsantos, line 41: should we update `msgId` to `saltedMsgId`?

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -32,20 +33,27 @@ const FloodSubCodec* = "/floodsub/1.0.0"
...
+  var tmp = f.seenSalt
+  tmp.update(msgId)
+  SaltedId(data: tmp.finish())
 
-proc addSeen*(f: FloodSub, msgId: MessageId): bool =
+proc hasSeen*(f: FloodSub, msgId: SaltedId): bool =
```
diegomrsantos, line 46: Should we update msgId to saltedMsgId? Same for the other places where the Type has changed but the corresponding param hasn't been renamed.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -318,7 +318,7 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
 
 proc validateAndRelay(g: GossipSub,
                       msg: Message,
-                      msgId, msgIdSalted: MessageId,
+                      msgId: MessageId, msgIdSalted: SaltedId,
```
diegomrsantos, line 321: why does `validationSeen` use `msgIdSalted`, but `mcache` uses `msgId`?

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -32,20 +33,27 @@ const FloodSubCodec* = "/floodsub/1.0.0"
 type
   FloodSub* {.public.} = ref object of PubSub
     floodsub*: PeerTable      # topic to remote peer map
-    seen*: TimedCache[MessageId] # message id:s already seen on the network
-    seenSalt*: seq[byte]
+    seen*: TimedCache[SaltedId] # message id:s already seen on the network
```
diegomrsantos, line 36: What about we add the salting to the Data Structures?

arnetheduck: Fixing `nim` to salt its `Table` is perhaps a fine long-term option but it would take too long for the benefit to show here for it to be practical - it also introduces a tiny bit of overhead so it is not a given that upstream would accept it.

There are some ideas around reimplementing `Table` completely, fixing this and many other things but I don't think there's enough support to warrant the significant effort that would entail (including updating all code to use it)

libp2p/protocols/pubsub/gossipsub.nim
arnetheduck: the latter only holds validated data, therefore doesn't need to be salted - the former is controlled by a (potential) remote attacker (until it has passed validation).

libp2p/protocols/pubsub/floodsub.nim
diegomrsantos: We can create a SecureTable. Would the effort be too big to replace Table where we use salted keys? We could change it incrementally.

diegomrsantos: It seems there should be two implementations or adding a param to the Table constructor instructing it to salt the keys. Salting isn't necessary for all use cases.

libp2p/protocols/pubsub/gossipsub.nim
arnetheduck: 513ee0bba5b7713745632a6609c90693baf8357e

libp2p/protocols/pubsub/floodsub.nim
arnetheduck: 513ee0bba5b7713745632a6609c90693baf8357e

libp2p/protocols/pubsub/floodsub.nim
arnetheduck: we could, but it feels out of scope for this PR - ie it already increases safety by using a separate type for the salted id - a rewrite with a special collection is a different task entirely, perhaps best done in isolation

tests/pubsub/testtimedcache.nim
```diff
@@ -33,6 +35,23 @@ suite "TimedCache":
...
+    check:
+      cache.del(4).isSome()
+      4 notin cache
+
     check:
       not cache.put(100, now + 100.seconds) # expires everything
       100 in cache
+      2 notin cache
+
+  test "spam":
```
diegomrsantos, line 47: Maybe sth like "can handle a large number of entries"? If that's the purpose of this test. Is there any reason why it starts at 101?

libp2p/protocols/pubsub/timedcache.nim
```diff
@@ -83,16 +102,24 @@ func put*[K](t: var TimedCache[K], k: K, now = Moment.now()): bool =
...
 func contains*[K](t: TimedCache[K], k: K): bool =
-  k in t.entries
-
-func addedAt*[K](t: TimedCache[K], k: K): Moment =
-  t.entries.getOrDefault(k).addedAt
-
+  let tmp = TimedEntry[K](key: k)
+  tmp in t.entries
+
+func addedAt*[K](t: var TimedCache[K], k: K): Moment =
```
diegomrsantos, line 113: What about returning Opt[Moment] and letting the caller decide what to do?

libp2p/protocols/pubsub/timedcache.nim
```diff
@@ -55,14 +74,14 @@ func put*[K](t: var TimedCache[K], k: K, now = Moment.now()): bool =
...
   t.expire(now)
 
-  var previous = t.del(k) # Refresh existing item
-
-  var addedAt = now
-  previous.withValue(previous):
-    addedAt = previous.addedAt
+  let
+    previous = t.del(k) # Refresh existing item
+    addedAt = if previous.isSome():
```
diegomrsantos, line 79: We had a long PR in the past to remove this pattern from the codebase and decrease the risk of raising defects. You can use https://github.com/vacp2p/nim-libp2p/blob/2725be64bacb46997c0959560d833ff946c8cb3b/libp2p/utility.nim#L124.

libp2p/protocols/pubsub/timedcache.nim
arnetheduck: this code matches the previous behavior of returning a default-initialized Moment - I don't see the need to change this in this PR.

libp2p/protocols/pubsub/timedcache.nim
arnetheduck: `valueOr` is not applicable in this case because we're accessing a field of `previous[]`, not `previous` itself

tests/pubsub/testtimedcache.nim
arnetheduck: no particular reason except to stress the hashset a bit more (by not starting at 0)

arnetheduck: 6c42af5f0108b16e85f876ff8c13b7d5f03703db

libp2p/protocols/pubsub/timedcache.nim
diegomrsantos: True, but withValue can be used in this case.

diegomrsantos (approved):
LGTM

libp2p/protocols/pubsub/timedcache.nim
arnetheduck: doesn't work in generic code, due to similar problems as https://github.com/arnetheduck/nim-results/issues/34

diegomrsantos: this seems to work fine:

```
    addedAt = block:
      previous.withValue(p):
        p[].addedAt
      else:
        now
```

PR #1074, opened 2024-03-19
fix: sendMsg must not raise errors

diegomrsantos:
This error making nimbus crash has been reported:

```
/home/seamonkey/Downloads/holesky/nimbus-eth2/vendor/nim-testutils/testutils/moduletests.nim(21) moduletests
/home/seamonkey/Downloads/holesky/nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(2376) main
/home/seamonkey/Downloads/holesky/nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(2299) handleStartUpCmd
/home/seamonkey/Downloads/holesky/nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(2200) doRunBeaconNode
/home/seamonkey/Downloads/holesky/nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(1957) start
/home/seamonkey/Downloads/holesky/nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(1904) run
/home/seamonkey/Downloads/holesky/nimbus-eth2/vendor/nim-chronos/chronos/internal/asyncengine.nim(150) poll
Error: unhandled exception: Asynchronous task [sendMsgSlow() at pubsubpeer.nim:288] was cancelled! [FutureDefect]
```

### DIFF
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -288,14 +288,17 @@ proc sendMsgSlow(p: PubSubPeer, msg: seq[byte]) {.async.} =
   if p.sendConn == nil:
     # Wait for a send conn to be setup. `connectOnce` will
     # complete this even if the sendConn setup failed
-    await p.connectedFut
+    try:
+      await p.connectedFut
+    except CatchableError as exc:
+      debug "Error when waiting for a send conn to be setup", msg = exc.msg, p
 
   var conn = p.sendConn
   if conn == nil or conn.closed():
-    debug "No send connection", p, msg = shortLog(msg)
+    debug "No send connection", msg = shortLog(msg), p
     return
 
-  trace "sending encoded msg to peer", conn, encoded = shortLog(msg)
+  trace "sending encoded msg to peer", conn, encoded = shortLog(msg), p
   await sendMsgContinue(conn, conn.writeLp(msg))
 
 proc sendMsg(p: PubSubPeer, msg: seq[byte]): Future[void] =
@@ -303,7 +306,7 @@ proc sendMsg(p: PubSubPeer, msg: seq[byte]): Future[void] =
     # Fast path that avoids copying msg (which happens for {.async.})
     let conn = p.sendConn
 
-    trace "sending encoded msg to peer", conn, encoded = shortLog(msg)
+    trace "sending encoded msg to peer", conn, encoded = shortLog(msg), p
     let f = conn.writeLp(msg)
     if not f.completed():
       sendMsgContinue(conn, f)
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -299,7 +299,10 @@ proc sendMsgSlow(p: PubSubPeer, msg: seq[byte]) {.async.} =
     return
 
   trace "sending encoded msg to peer", conn, encoded = shortLog(msg), p
-  await sendMsgContinue(conn, conn.writeLp(msg))
+  try:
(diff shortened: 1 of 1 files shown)


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -288,22 +288,25 @@ proc sendMsgSlow(p: PubSubPeer, msg: seq[byte]) {.async.} =
...
 
   var conn = p.sendConn
   if conn == nil or conn.closed():
-    debug "No send connection", p, msg = shortLog(msg)
+    debug "No send connection", msg = shortLog(msg), p
     return
 
-  trace "sending encoded msg to peer", conn, encoded = shortLog(msg)
+  trace "sending encoded msg to peer", conn, encoded = shortLog(msg), p
   await sendMsgContinue(conn, conn.writeLp(msg))
```
cheatfate, line 302: This place could also be source of `CancelledError`.
Every `await` could be source.

diegomrsantos: even if `sendMsgContinue` doesn't raise any error?

cheatfate: of course.

arnetheduck:
see https://github.com/vacp2p/nim-libp2p/pull/1075 - the problem is that stopping the non-priority send task cancels the sendMsg future even though the non-priority task is not the - catching exceptions will not help with this problem

diegomrsantos:
@arnetheduck why canceling the task would cancel the futures inside the queue if they aren't related?

arnetheduck:
when you `await` a task in an `async` function (such as the low prio loop) and cancel the call to that `async` function, the cancellation propagates to the thing currently being awaited - consider what happens when you call  - in this case the queue item - `race` and `allFutures` "break" this chain of cancellations.

consider:

```nim
proc f() {.async.} = await sleepAsync(10.minutes)
proc g() {.async.} = await f()

let fut = g()
fut.cancelSoon()
```

here, you can clearly see that if you cancel `fut`, you want the cancellation to propagate all the way to `sleepAsync` so that the sleep is cancelled.

in the case of `sendMsg`, we store the future in a sequence and there are _two_ things waiting for it - the asyncSpawn task and the non-priority loop - `asyncSpawn` is the "owner" of the future chain and therefore the only one that should cancel the future - the non-prio loop should not cancel it because it is not the owner.

diegomrsantos:
Can we use https://github.com/status-im/nim-chronos/blob/e296ae30c84bdd1f0b12c50ab551ed080f8a815c/chronos/internal/asyncfutures.nim#L1029?

diegomrsantos:
> catching exceptions will not help with this problem

This PR won't prevent the Future from being canceled but will fix the crash.

arnetheduck:
> Can we use https://github.com/status-im/nim-chronos/blob/e296ae30c84bdd1f0b12c50ab551ed080f8a815c/chronos/internal/asyncfutures.nim#L1029?

no - `noCancel` blocks cancellation until the given future is finished "naturally" - in the sleep example, it would mean that the cancellation takes 10 minutes to complete.

> This PR won't prevent the Future from being canceled but will fix the crash.

There's no mechanism right now that is able to prevent futures from finishing - the work @etan-status is doing allows for better discovery of where error and cancellation handling is needed however - as soon as the "initial" introduction is done, it will be easy to find where we need to add cancellation handling and where not

diegomrsantos:
so should we close this PR?

arnetheduck:
yes, I think it's not needed at the moment - `raises`, when introduced, will reveal the exact error handling needs in this function

diegomrsantos:
closing this for now

PR #1075, opened 2024-03-20
avoid cancelling send future

arnetheduck:
otherwise, the use of `asyncSpawn` results in breakage


diegomrsantos:
supersedes https://github.com/vacp2p/nim-libp2p/pull/1074

diegomrsantos (approved):

PR #1077, opened 2024-03-21
feat: add max number of elements to non-prio queue

diegomrsantos:
This PR adds a default limit to the number of elements in the non-priority queue. When this limit has been reached, the peer will be disconnected. This limit can also be configured through a new GossipSub param `maxNumElementsInNonPriorityQueue`.

### DIFF
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -83,6 +83,7 @@ type
     overheadRateLimitOpt*: Opt[TokenBucket]
 
     rpcmessagequeue: RpcMessageQueue
+    maxNumElementInNonPriorityQueue*: int # The max number of elements allowed in the non-priority queue.
 
   RPCHandler* = proc(peer: PubSubPeer, data: seq[byte]): Future[void]
     {.gcsafe, raises: [].}
@@ -181,6 +182,24 @@ proc handle*(p: PubSubPeer, conn: Connection) {.async.} =
     debug "exiting pubsub read loop",
       conn, peer = p, closed = conn.closed
 
+proc disconnect(p: PubSubPeer) {.async.} =
+  if p.sendConn != nil:
+    trace "Removing send connection", p, conn = p.sendConn
+    await p.sendConn.close()
+    p.sendConn = nil
+
+  if not p.connectedFut.finished:
+    p.connectedFut.complete()
+
+  try:
+    if p.onEvent != nil:
+      p.onEvent(p, PubSubPeerEvent(kind: PubSubPeerEventKind.Disconnected))
+  except CancelledError as exc:
+    raise exc
+  except CatchableError as exc:
+    debug "Errors during diconnection events", error = exc.msg
+  # don't cleanup p.address else we leak some gossip stat table
+
 proc connectOnce(p: PubSubPeer): Future[void] {.async.} =
   try:
     if p.connectedFut.finished:
@@ -207,23 +226,7 @@ proc connectOnce(p: PubSubPeer): Future[void] {.async.} =
 
     await handle(p, newConn)
   finally:
-    if p.sendConn != nil:
-      trace "Removing send connection", p, conn = p.sendConn
-      await p.sendConn.close()
-      p.sendConn = nil
-
-    if not p.connectedFut.finished:
-      p.connectedFut.complete()
-
-    try:
-      if p.onEvent != nil:
-        p.onEvent(p, PubSubPeerEvent(kind: PubSubPeerEventKind.Disconnected))
-    except CancelledError as exc:
-      raise exc
-    except CatchableError as exc:
-      debug "Errors during diconnection events", error = exc.msg
-
-    # don't cleanup p.address else we leak some gossip stat table
+    await p.disconnect()
 
 proc connectImpl(p: PubSubPeer) {.async.} =
   try:
@@ -337,10 +340,13 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte], isHighPriority: bool): Future[v
         libp2p_gossipsub_priority_queue_size.inc(labelValues = [$p.peerId])
     f
   else:
-    let f = p.rpcmessagequeue.nonPriorityQueue.addLast(msg)
-    when defined(pubsubpeer_queue_metrics):
-      libp2p_gossipsub_non_priority_queue_size.inc(labelValues = [$p.peerId])
-    f
+    if len(p.rpcmessagequeue.nonPriorityQueue) == p.maxNumElementInNonPriorityQueue:
+      p.disconnect()
+    else:
+      let f = p.rpcmessagequeue.nonPriorityQueue.addLast(msg)
+      when defined(pubsubpeer_queue_metrics):
+        libp2p_gossipsub_non_priority_queue_size.inc(labelValues = [$p.peerId])
+      f
 
 iterator splitRPCMsg(peer: PubSubPeer, rpcMsg: RPCMsg, maxSize: int, anonymize: bool): seq[byte] =
   ## This iterator takes an `RPCMsg` and sequentially repackages its Messages into new `RPCMsg` instances.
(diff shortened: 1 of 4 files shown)


diegomrsantos:
@arnetheduck how about replacing https://github.com/vacp2p/nim-libp2p/pull/1077/files#diff-abbf987f19a24c9940cd01bde79a1b1e0819f1c3b6d7ba2efa04ad680d022c78R182-R186 by `unsubscribePeer`?

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -181,6 +182,24 @@ proc handle*(p: PubSubPeer, conn: Connection) {.async.} =
     debug "exiting pubsub read loop",
       conn, peer = p, closed = conn.closed
 
+proc disconnect(p: PubSubPeer) {.async.} =
```
arnetheduck, line 185: this doesn't disconnect the peer - it just closes the send connection

arnetheduck:
> @arnetheduck how about replacing https://github.com/vacp2p/nim-libp2p/pull/1077/files#diff-abbf987f19a24c9940cd01bde79a1b1e0819f1c3b6d7ba2efa04ad680d022c78R182-R186 by unsubscribePeer?

let's start with peer disconnection - when we reach a stable point with that working, we can consider other strategies (so that we can get a release out).

libp2p/protocols/pubsub/pubsubpeer.nim
diegomrsantos: I can't disconnect the peer from `PubSubPeer`, it's available only in GossipSub and I don't have a reference to it here cause of circular dependency limitation.

diegomrsantos: It seems we emit a `Disconnected` event that doesn't mean a disconnection.

diegomrsantos: Let me know if I goit it right.

1. We reset the stream and close the underlying connection here, but don't raise any error so at the time no error happens https://github.com/vacp2p/nim-libp2p/blob/28609597d104a9be880ed5e1648e1ce18ca9dc38/libp2p/muxers/mplex/lpchannel.nim#L220
2. Next time we try to write to the stream this error happens https://github.com/vacp2p/nim-libp2p/blob/e0bcc5fed20ea1b6c8eea605a4ad4caa275b3fb3/libp2p/protocols/pubsub/pubsubpeer.nim#L181
3. Then this is execute https://github.com/vacp2p/nim-libp2p/blob/e0bcc5fed20ea1b6c8eea605a4ad4caa275b3fb3/libp2p/protocols/pubsub/pubsubpeer.nim#L229

So the only difference now is that the underlying connection hasn't been closed?

arnetheduck: > So the only difference now is that the underlying connection hasn't been closed?

yes - closing the underlying ensures that all ongoing concurrent reads, writes etc are aborted and we get rid of the peer - this is important because we don't know what state gossipsub is in if we just abort here - they might have received control messages or not so closing the connection entirely is one of the safest and least bad things you can do.

arnetheduck: there exists an ugly-hack solution: drain the non-prio queue and send all non-prio messages as priority messages - this will _likely_ cause mplex to disconnect

diegomrsantos: But how can I close the underlying connection here? The conn here is actually a stream (btw what about calling it a stream when it is one?) and `Connection` doesn't expose the underlying transport connection.

arnetheduck: @dryajov knows more about the stream vs connection naming, maybe

diegomrsantos: I could rename `PubSubPeerEventKind.Disconnected` to `PubSubPeerEventKind.StreamClosed` and create a `PubSubPeerEventKind.Disconnect`. Then `GossipSub` would disconnect the peer when receiving the latter. This would be the closest to the current behavior, I guess.  Another option would be to disconnect the peer here https://github.com/vacp2p/nim-libp2p/blob/1734160f8a3e759ea03145878b3e7f19f2addbad/libp2p/protocols/pubsub/gossipsub.nim#L185, but it would change the current behavior.

arnetheduck: > But how can I close the underlying connection here? 

see suggested send-all-non-prio-messages trick - it should work because if there's at least 1 in-flight future, it will cause mplex to do the disconnection.

diegomrsantos: Wouldn't emit an event that tells `GossipSub` to disconnect the peer be simpler and easier to understand?

diegomrsantos: but I think just writing the whole queue might not trigger the disconnection if the futures complete fast enough.

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -338,10 +345,18 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte], isHighPriority: bool): Future[v
         libp2p_gossipsub_priority_queue_size.inc(labelValues = [$p.peerId])
     f
   else:
-    let f = p.rpcmessagequeue.nonPriorityQueue.addLast(msg)
-    when defined(pubsubpeer_queue_metrics):
-      libp2p_gossipsub_non_priority_queue_size.inc(labelValues = [$p.peerId])
-    f
+    if len(p.rpcmessagequeue.nonPriorityQueue) == p.maxNumElementInNonPriorityQueue:
+      if not p.disconnected:
```
arnetheduck, line 349: if this flag is set, it seems prudent to check in `connectOnce` that we don't re-attempt connection, ie `DisconnectionRequested` should lead to the send conn never being opened again

diegomrsantos: thanks, I saw that peers kept being disconnected multiple times and that was probably the reason. But I had to add the check in `connectImpl`.

arnetheduck: well, the non-prio queue should also be cleared (ie all sends in it should be failed) at this point since it will no longer be sent - and if there are any new sends after the disconnection, they should fail

arnetheduck: also, it _maybe_ makes sense to balance the streamOpened event with a streamclosed in this case, such events typically benefit from being neutralised this way

diegomrsantos: Canceling the future will cause this problem again https://github.com/vacp2p/nim-libp2p/pull/1075, won't it?

diegomrsantos: > maybe makes sense to balance the streamOpened event with a streamclosed in this case

Could you please elaborate more on what you mean by that?

diegomrsantos: > Canceling the futures will cause this problem again #1075, won't it?

Not sure if you meant canceling the futures. `stopSendNonPriorityTask` clears the queue and it's called in `unsubscribePeer` which should be called when the peer is disconnected.

arnetheduck: > Not sure if you meant canceling the futures. 

I meant failing them, not cancelling, but now I realise we always return a completed future instead of storing pending futures in the queue, so there's nothing to fail:

https://github.com/vacp2p/nim-libp2p/blob/unstable/libp2p/protocols/pubsub/pubsubpeer.nim#L341

`addLast` always returns a completed future when the async queue is unbounded (as it is in this case).

arnetheduck: > Should I also emit a Streamclosed?

this is an idea for a future improvement - it's maybe not needed right now, however

arnetheduck: There's basically a lifecycle here:

`created -> [stream opened -> stream closed]* -> disconnected` where the middle two states repeat, so when designing an event api around it, these are the relevant events.

arnetheduck (approved):

PR #1078, opened 2024-03-25
improvement: create a new gossipsub constructor

diegomrsantos:
Before this PR, there were two ways of creating a GossipSubParam instance: using `GossipSubParam.init`or the `GossipSubParams(propertyName: propertyValue)` pattern. In the first case, there was no way to specify only a few values and use the rest as the default defined by nim-libp2p, as the `init` proc doesn't accept params. It was the same for the second case, the values not specified would be initialized as the nim default value for their types. This PR creates a new constructor with default values defined for all the params. The user can choose which ones to override or use the default value defined by the library.

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -87,6 +87,82 @@ proc init*(_: type[GossipSubParams]): GossipSubParams =
       maxNumElementsInNonPriorityQueue: DefaultMaxNumElementsInNonPriorityQueue
     )
 
+proc new*(
+  _: type[GossipSubParams],
+  explicit = true,
+  pruneBackoff = 1.minutes,
+  unsubscribeBackoff = 5.seconds,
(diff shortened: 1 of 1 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -87,6 +87,82 @@ proc init*(_: type[GossipSubParams]): GossipSubParams =
       maxNumElementsInNonPriorityQueue: DefaultMaxNumElementsInNonPriorityQueue
     )
 
+proc new*(
```
arnetheduck, line 90: the name `new` is for functions that introduce a `ref` in the return type - `GossipSubParams` - this should be called `init`: https://status-im.github.io/nim-style-guide/language.objconstr.html

diegomrsantos: Done.

arnetheduck (approved):

PR #1079, opened 2024-04-03
fix: valueOr and withValue utilities

lchenut:
This PR fixes a double call when using the `valueOr`, `withValue` and `toOpt` templates.

The issue was found by @chaitanyaprem with a double call of a custom protocol here: https://github.com/waku-org/nwaku/blob/feat--nwaku-sync/waku/waku_sync/protocol.nim#L125-L126

### DIFF
--- a/libp2p/utility.nim
+++ b/libp2p/utility.nim
@@ -115,21 +115,24 @@ template withValue*[T](self: Opt[T] | Option[T], value, body: untyped): untyped 
 macro withValue*[T](self: Opt[T] | Option[T], value, body, body2: untyped): untyped =
   let elseBody = body2[0]
   quote do:
-    if `self`.isSome:
-      let `value` {.inject.} = `self`.get()
+    let temp = `self`
+    if temp.isSome:
+      let `value` {.inject.} = temp.get()
       `body`
     else:
       `elseBody`
 
 template valueOr*[T](self: Option[T], body: untyped): untyped =
-  if self.isSome:
-    self.get()
+  let temp = (self)
+  if temp.isSome:
+    temp.get()
   else:
     body
 
 template toOpt*[T, E](self: Result[T, E]): Opt[T] =
-  if self.isOk:
+  let temp = (self)
+  if temp.isOk:
     when T is void: Result[void, void].ok()
-    else: Opt.some(self.unsafeGet())
+    else: Opt.some(temp.unsafeGet())
   else:
     Opt.none(type(T))
--- a/tests/testutility.nim
+++ b/tests/testutility.nim
@@ -9,6 +9,7 @@
 # This file may not be copied, modified, or distributed except according to
 # those terms.
 
+import options
 import ./helpers
 import ../libp2p/utility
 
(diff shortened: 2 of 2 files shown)


libp2p/utility.nim
```diff
@@ -115,21 +115,24 @@ template withValue*[T](self: Opt[T] | Option[T], value, body: untyped): untyped
 macro withValue*[T](self: Opt[T] | Option[T], value, body, body2: untyped): untyped =
   let elseBody = body2[0]
   quote do:
-    if `self`.isSome:
-      let `value` {.inject.} = `self`.get()
+    let temp = `self`
```
diegomrsantos, line 118: We fixed the template but didn't notice the macro. Should we use `()` as in the template? I don't remember why it's necessary. Could you please rename `body2` to a more descriptive name and add docs similar to the template?

tests/testutility.nim
```diff
@@ -71,3 +72,75 @@ suite "Utility":
   test "unsuccessful safeConvert from uint to int":
     check not (compiles do:
       result: uint = safeConvert[int, uint](11.uint))
+
+  test "withValue":
```
diegomrsantos, line 76: This test would benefit from being split into smaller tests with descriptive names that test only one thing like "withValue calls right branch when Opt is some", "withValue calls right branch when Opt is none", "withValue is called only once when Opt is some", etc.

libp2p/utility.nim
lchenut: The `()` are necessary in C macro when you're doing something like that for example
```C
#include <stdio.h>
#define Mult_by_two(x) x * 2
#define Mult_by_two_parenthesis(x) (x) * 2

int main(void) {
  printf("%i\n", Mult_by_two(2+2)); // will print 6 (2 + 2 * 2);
  printf("%i\n", Mult_by_two_parenthesis(2+2)); // will print 8 ((2+2)*2);
  return (0);
}
```
And I usually tend to use this in the other languages, even if the language fix this "issue".

tests/testutility.nim
lchenut: Done :+1:

tests/testutility.nim
```diff
@@ -71,3 +72,88 @@ suite "Utility":
...
+
+  proc objIncAndOption(self: TestObj): Option[TestObj] =
+    self.x.inc()
+    return some(self)
+
+  test "withValue calls right branch when Opt/Option is none":
+    var counter = 0
+    # check Opt/Option withValue with else
+    Opt.none(TestObj).withValue(v):
+      check false
```
diegomrsantos, line 93: how about `fail()` instead?

tests/testutility.nim
```diff
@@ -71,3 +72,88 @@ suite "Utility":
...
+    else:
+      counter.inc()
```
diegomrsantos, line 95: we can probably extract it to a proc inside the test and avoid the repetition

tests/testutility.nim
```diff
@@ -71,3 +72,88 @@ suite "Utility":
...
+    check counter == 2
+
+    # check Opt/Option withValue without else
+    Opt.none(TestObj).withValue(v):
+      check false
+    none(TestObj).withValue(v):
+      check false
+
+  test "withValue calls right branch when Opt/Option is some":
+    var obj = TestObj(x: 0)
```
diegomrsantos, line 109: Do you know if we need this object? Could we create an Option of an int and use a counter like you did in the previous test?

tests/testutility.nim
lchenut: You mean something like that?
```nim
proc testProc[T](val: Opt[T] | Option[T], counter: var int) =
  val.withValue(v):
    fail()
  else:
    counter.inc()
testProc(Opt.none(TestObj), counter)
testProc(none(TestObj), counter)
```
If there were more than two occurrences of this test I would do it, but I don't think it really brings something in this specific case.

diegomrsantos: Yes, the benefit is avoiding code duplication, which I believe is never desirable.

diegomrsantos (approved):
Looks great. Thanks a lot for taking care of it and addressing the comments!

PR #1082, opened 2024-04-08
fix: reset accept fut in stop

Ivansete-status:
Changes:

- Reset `acceptFuts` seq when `stop` is invoked: this is needed to properly perform connection restarts between nodes and this issue was detected in `nwaku` tests, as explained in https://github.com/waku-org/nwaku/issues/2145

- Add a debug log when an exception is caught. This change is not directly needed by the "reconnect" issue mentioned in the previous point.

### DIFF
--- a/libp2p/transports/tcptransport.nim
+++ b/libp2p/transports/tcptransport.nim
@@ -190,7 +190,9 @@ method stop*(self: TcpTransport) {.async.} =
     await allFutures(toWait)
 
     self.servers = @[]
-    self.acceptFuts = @[]
+
+    if self.acceptFuts.allIt(it.finished()):
+      self.acceptFuts = @[]
 
     trace "Transport stopped"
     untrackCounter(TcpTransportTrackerName)
--- a/libp2p/transports/tcptransport.nim
+++ b/libp2p/transports/tcptransport.nim
@@ -222,6 +222,7 @@ method accept*(self: TcpTransport): Future[Connection] {.async.} =
       let observedAddr = MultiAddress.init(transp.remoteAddress).tryGet()
       return await self.connHandler(transp, Opt.some(observedAddr), Direction.In)
     except CancelledError as exc:
+      debug "CancelledError", exc = exc.msg
       transp.close()
       raise exc
     except CatchableError as exc:
--- a/libp2p/transports/tcptransport.nim
+++ b/libp2p/transports/tcptransport.nim
@@ -190,6 +190,7 @@ method stop*(self: TcpTransport) {.async.} =
     await allFutures(toWait)
 
     self.servers = @[]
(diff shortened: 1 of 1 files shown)


libp2p/transports/tcptransport.nim
```diff
@@ -221,6 +222,7 @@ method accept*(self: TcpTransport): Future[Connection] {.async.} =
       let observedAddr = MultiAddress.init(transp.remoteAddress).tryGet()
       return await self.connHandler(transp, Opt.some(observedAddr), Direction.In)
     except CancelledError as exc:
+      debug "CancelledError", exc = exc.msg
```
diegomrsantos, line 225: I believe this is logged by the caller as it is re-raised.

libp2p/transports/tcptransport.nim
```diff
@@ -190,6 +190,7 @@ method stop*(self: TcpTransport) {.async.} =
     await allFutures(toWait)
 
     self.servers = @[]
+    self.acceptFuts = @[]
```
diegomrsantos, line 193: I don't know if this is the right approach as I don't have much knowledge about this code. @dryajov could you please advise?

dryajov: We need to make sure that all the accept futs are properly canceled, before resetting.

libp2p/transports/tcptransport.nim
```diff
@@ -191,6 +191,9 @@ method stop*(self: TcpTransport) {.async.} =
 
     self.servers = @[]
 
+    if self.acceptFuts.allIt(it.finished()):
```
diegomrsantos, line 194: I interpreted @dryajov's comment as: for each future that hasn't been finished, it must be canceled.

diegomrsantos (dismissed):
LGTM, many thanks.

libp2p/transports/tcptransport.nim
dryajov: Yes, that should be enough. So this new check isn't necessary, the previous version without this change should be enough.

Ivansete-status: > Yes, that should be enough. So this new check isn't necessary, the previous version without this change should be enough.

Okay perfect! I just reverted the previous commit.

@diegomrsantos - Kindly re-approve and merge if you consider it appropriate :)
Thanks!

diegomrsantos: Sure, sorry for the trouble.

diegomrsantos (approved):

libp2p/transports/tcptransport.nim
Ivansete-status: > Sure, sorry for the trouble.

haha no worries at all! Thanks for the patience! :raised_hands:

PR #1083, opened 2024-04-09
feat: behaviour penalty when non-priority queue reaches maxNumElementsInNonPriorityQueue

diegomrsantos:
This PR applies a behavior penalty to peers whose non-prio queue reaches the max limit configured, instead of the previous strategy of disconnecting the peer. A conservative penalty of `0.0001` is added to behaviourPenalty for each message tried to be sent when the queue is over the limit, and the message is discarded. This usually results in a `behaviourPenalty` around [0.1, 0.2] when the score is updated and its value around [-0.4, -0.1]. 

It causes the peer to be pruned due to its negative score. This PR also clears the non-prio queue at this moment.

### DIFF
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -35,8 +35,6 @@ when defined(pubsubpeer_queue_metrics):
   declareGauge(libp2p_gossipsub_priority_queue_size, "the number of messages in the priority queue", labels = ["id"])
   declareGauge(libp2p_gossipsub_non_priority_queue_size, "the number of messages in the non-priority queue", labels = ["id"])
 
-declareCounter(libp2p_pubsub_disconnects_over_non_priority_queue_limit, "number of peers disconnected due to over non-prio queue capacity")
-
 const
   DefaultMaxNumElementsInNonPriorityQueue* = 1024
 
@@ -90,7 +88,6 @@ type
 
     rpcmessagequeue: RpcMessageQueue
     maxNumElementsInNonPriorityQueue*: int # The max number of elements allowed in the non-priority queue.
-    disconnected: bool
 
   RPCHandler* = proc(peer: PubSubPeer, data: seq[byte]): Future[void]
     {.gcsafe, raises: [].}
@@ -191,7 +188,7 @@ proc handle*(p: PubSubPeer, conn: Connection) {.async.} =
 
 proc closeSendConn(p: PubSubPeer, event: PubSubPeerEventKind) {.async.} =
   if p.sendConn != nil:
-    trace "Removing send connection", p, conn = p.sendConn
+    debug "Removing send connection", p, conn = p.sendConn
     await p.sendConn.close()
     p.sendConn = nil
 
@@ -219,7 +216,7 @@ proc connectOnce(p: PubSubPeer): Future[void] {.async.} =
     # remote peer - if we had multiple channels up and one goes down, all
     # stop working so we make an effort to only keep a single channel alive
 
-    trace "Get new send connection", p, newConn
+    debug "Get new send connection", p, newConn
 
     # Careful to race conditions here.
     # Topic subscription relies on either connectedFut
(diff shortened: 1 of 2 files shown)


diegomrsantos:
[nimbus-log.txt](https://github.com/vacp2p/nim-libp2p/files/14934174/nimbus-log.txt)

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -352,12 +352,10 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte], isHighPriority: bool): Future[v
     f
   else:
     if len(p.rpcmessagequeue.nonPriorityQueue) >= p.maxNumElementsInNonPriorityQueue:
-      if not p.disconnected:
-        p.disconnected = true
-        libp2p_pubsub_disconnects_over_non_priority_queue_limit.inc()
-        p.closeSendConn(PubSubPeerEventKind.DisconnectionRequested)
-      else:
-        Future[void].completed()
+      p.behaviourPenalty += 0.0001
```
AlejandroCabeza, line 355: I'd suggest moving the value used here to a constant.

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -352,12 +352,10 @@ proc sendEncoded*(p: PubSubPeer, msg: seq[byte], isHighPriority: bool): Future[v
...
+      trace "Peer has reached maxNumElementsInNonPriorityQueue. Discarding message and applying behaviour penalty.", peer = p, score = p.score,
+        behaviourPenalty = p.behaviourPenalty, agent = p.getAgent()
+      Future[void].completed()
```
AlejandroCabeza, line 358: Given most return paths are `Future[void].completed()` how about moving it outside the `if` statement and doing an explicit `return` on each arm for the cases that are not `Future[void].completed()`.

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -281,6 +274,13 @@ proc clearSendPriorityQueue(p: PubSubPeer) =
       value = p.rpcmessagequeue.sendPriorityQueue.len.int64,
       labelValues = [$p.peerId])
 
+proc clearNonPriorityQueue*(p: PubSubPeer) =
+  if len(p.rpcmessagequeue.nonPriorityQueue) > 0:
+    p.rpcmessagequeue.nonPriorityQueue.clear()
+
+    when defined(pubsubpeer_queue_metrics):
+      libp2p_gossipsub_non_priority_queue_size.set(labelValues = [$p.peerId], value = 0)
+
```
AlejandroCabeza, line 283: How asynchronous is this operation? That is: How feasible it is that, between the `clear` call and the metrics update, some value is added to the `nonPriorityQueue`? 🤔

diegomrsantos: This proc isn't async and won't be suspended.

libp2p/protocols/pubsub/pubsubpeer.nim
diegomrsantos: done

libp2p/protocols/pubsub/pubsubpeer.nim
AlejandroCabeza: Oh, that's right. nvm then

libp2p/protocols/pubsub/pubsubpeer.nim
AlejandroCabeza: There's nothing wrong per se. What I was hinting at was, when you have code akin to this:
```nim
if <cond>:
  foo()
  return A  

elif <cond>:
  bar()
  return A

elif <cond>:
  baz()
  return B

...

else:
  foobar()
  return A
```

Sometimes it's clearer to group/extract the returns:
```nim
if <cond>:
  foo()

elif <cond>:
  bar()


elif <cond>:
  baz()
  return B

...

else:
  foobar()

return A
```

But I just realised here it's almost 50/50, so not a lot of improvement can be made.

strinnityk (approved):

richard-ramos:
Closing PR as it is stale

PR #1085, opened 2024-04-18
use a mock rng in tests

diegomrsantos:


Ben-PH:
I haven't had the chance to check the changes in any real detail. Here are some checklist items I would like to see:

- The form of the data needs to not change much. That is to say, if you're testing where a 512 bit secure hash is called for in production, and you're just incrementing a 512bit integer from 0 to do the tests, then you're testing with data that is completely different to real use cases.
- Ideally, the mock-RNG for testing needs to be deterministic and cross platform. That is to say, no matter which OS, test contex, environment, etc. the data you get will be deterministic. This can be achieved by using a specifically chosen seed each time. It could be randomly selected each run, but the seed would have to be logged so that a test can be re-run with the exact same data.
- Is performance during testing due to RNG an issue that can be referenced to (e.g. an issue or discussion in GH)? If so, please reference it in your PR. If not, please such a PR would merit having that issue raised as a discussion point.

A couple of other things might come up, but these are the items at the top of my head.

I would encourage a discussion being started (or link to one already in progress).

PR #1086, opened 2024-04-22
fix(yamux): set EoF when remote peer half closes the stream in yamux

diegomrsantos:
- remove unnecessary `returnedEoF` field
- don't set isEoF when the local peer half closes the stream
- add more logs
- add test that closes the stream for writing but reads from it


libp2p/muxers/yamux/yamux.nim
```diff
@@ -442,7 +441,7 @@ proc cleanupChannel(m: Yamux, channel: YamuxChannel) {.async: (raises: []).} =
   when defined(libp2p_yamux_metrics):
     libp2p_yamux_channels.set(
       m.lenBySrc(channel.isSrc).int64, [$channel.isSrc, $channel.peerId])
-  if channel.isReset and channel.recvWindow > 0:
+  if (channel.isReset or not channel.closedRemotely.completed()) and channel.recvWindow > 0:
```
diegomrsantos, line 444: I'm not sure about this new condition.

libp2p/muxers/yamux/yamux.nim
```diff
@@ -273,17 +273,16 @@ method readOnce*(
         newLPStreamClosedError()
       else:
         newLPStreamConnDownError()
-  if channel.returnedEof:
+  if channel.isEof:
```
lchenut, line 276: As it's the only place where returnedEof was used, you can safely remove it from the `YamuxChannel` definition aswell.

libp2p/muxers/yamux/yamux.nim
lchenut: `channel.closedRemotely` is always completed after `channel.join()`.
In theory, the only possible case is if we try to cancel, but `cleanupChannel is async spawned, so we cannot cancel it.

libp2p/muxers/yamux/yamux.nim
diegomrsantos: done

libp2p/muxers/yamux/yamux.nim
lchenut: `channel.join()` is completed after `LPstream.closeImpl()` is called.
`LPStream.closeImpl()` is called by `actuallyClosed` [here](https://github.com/vacp2p/nim-libp2p/pull/1086/files#diff-fa0addf2f141e6d44cffe29109c7b4fa96b608c52425616af4ff1cf763976d8fR194-R197), and it's called after checking if `closedRemotely` is completed.

diegomrsantos: True, reverting.

tests/transport-interop/Dockerfile
```diff
@@ -11,6 +11,6 @@ COPY . nim-libp2p/
 
 RUN \
   cd nim-libp2p && \
-  nim c --skipProjCfg --skipParentCfg --NimblePath:./nimbledeps/pkgs -p:nim-libp2p -d:chronicles_log_level=WARN --threads:off ./tests/transport-interop/main.nim
+  nim c --skipProjCfg --skipParentCfg --NimblePath:./nimbledeps/pkgs -p:nim-libp2p -d:chronicles_log_level=WARN -d:chronicles_default_output_device=stderr --threads:off ./tests/transport-interop/main.nim
```
diegomrsantos, line 14: This is how it should be according to the test spec in `test-plans`. This is a bit unrelated to the current scope of the PR, but the PR itself was born from debugging this test and this change was necessary, otherwise, the test fails when increasing the log level. Ideally, it should be done in a different PR, but it is fine in this case.

lchenut (approved):
LGTM

PR #1093, opened 2024-05-06
staggered_sending

ufarooqstatus:
we use semaphores to limit simultaneous transmissions. can speedup message reception for large messages


ufarooqstatus:
[PR-1100](https://github.com/vacp2p/nim-libp2p/pull/1100)

PR #1095, opened 2024-05-07
fix(transport): various tcp transport races

arnetheduck:
This PR is more or less a rewrite of the TCP transport to fix several race conditions and leaks that can happen during startup / shutdown as well as spurious exceptions leaking out of it

* simplify close monitoring since `ChronosStream` already closes the underlying chronos client of a `Connection`
* fix dialling cancellation
* stop accepting incoming connections before closing existing ones (race)
* close in-flight connections during `TcpTransport.stop`
* make sure only Transport-related exceptions leak out of it


libp2p/dialer.nim
```diff
@@ -81,16 +81,17 @@ proc dialAndUpgrade(
           if dialed.dir != dir:
             dialed.dir = dir
           await transport.upgrade(dialed, peerId)
+        except CancelledError as exc:
```
diegomrsantos, line 84: Could you please elaborate on why this is necessary and why we don't need to call `close`?

arnetheduck: oops, close should stay (ed6be85d6e3d72008070625c0fb1637ceed37095) - the point was to not return `nil` which makes the dialler try the next address instead of aborting the dialling

diegomrsantos: Could we catch only `LPError` instead? https://github.com/vacp2p/nim-libp2p/blob/c5db35d9b006b597886440293ff5cbd86162452c/libp2p/transports/transport.nim#L88

libp2p/switch.nim
```diff
@@ -288,6 +289,12 @@ proc stop*(s: Switch) {.async, public.} =
 
   s.started = false
 
+  try:
+    # Stop accepting incoming connections
+    await allFutures(s.acceptFuts.mapIt(it.cancelAndWait())).wait(1.seconds)
+  except CatchableError as exc:
+    debug "Cannot cancel accepts", error = exc.msg
+
```
diegomrsantos, line 297: It seems to be a bit related https://github.com/vacp2p/nim-libp2p/pull/1065#discussion_r1561355153

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
-proc connHandler*(self: TcpTransport,
-                  client: StreamTransport,
-                  observedAddr: Opt[MultiAddress],
-                  dir: Direction): Future[Connection] {.async.} =
-
-  trace "Handling tcp connection", address = $observedAddr,
-                                   dir = $dir,
-                                   clients = self.clients[Direction.In].len +
-                                   self.clients[Direction.Out].len
+proc connHandler*(
```
diegomrsantos, line 46: Do you think this should be public? Could you please add a doc to this proc? I don't understand what is going on here.

diegomrsantos: Also not sure that's a descriptive name.

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
+      initialized = true
+    finally:
+      if not initialized:
+        # Clean up partial success on exception
+        await noCancel allFutures(self.servers.mapIt(it.closeWait()))
+        reset(self.servers)
 
-    trace "Listening on", address = ma
+    try:
+      await procCall Transport(self).start(supported)
```
diegomrsantos, line 159: You're probably trying to set "running" only after all the code above ran without issues, but it seems strange to call a parent method only at the end. It will be confusing to see the start log only after the start happened and the associated logs.

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
+  ): Future[void] {.async: (raises: [transport.TransportError, CancelledError]).} =
+    if self.running:
+      warn "TCP transport already running"
+      return
 
-  for i, ma in addrs:
-    if not self.handles(ma):
-      trace "Invalid address detected, skipping!", address = ma
-      continue
+    trace "Starting TCP transport"
```
diegomrsantos, line 122: I just realized there's a double log here as the parent start also logs this. I also believe this kind of long should be info, wdyt?

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
+        if not self.handles(ma):
+          trace "Invalid address detected, skipping!", address = ma
+          continue
 
-    self.servers &= server
+        let
+          ta = initTAddress(ma).expect("valid address per handles check above")
+          server =
+            try:
+              createStreamServer(ta, flags = self.flags)
```
diegomrsantos, line 138: Could you please elaborate on why have you changed the previous code? The `handles` proc above only checks if the address is TCP, more checks are done in the `createStreamServer` used previoulsy. This change also seems unrelated to the objective stated in the PR description.

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
+    var initialized = false
+    try:
+      for i, ma in addrs:
+        if not self.handles(ma):
+          trace "Invalid address detected, skipping!", address = ma
+          continue
 
-    self.servers &= server
+        let
+          ta = initTAddress(ma).expect("valid address per handles check above")
```
diegomrsantos, line 135: If I understand correctly, this is going to raise a `Defect`. What's the motivation for that? I understood the objective was to eliminate all the potential sources of Defects unless it's explicitly clear we want to do that due to an unrecoverable error. This doesn't seem to be the case.

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
+            "Can init from local address"
+          )
+        )
+
+      initialized = true
+    finally:
+      if not initialized:
+        # Clean up partial success on exception
+        await noCancel allFutures(self.servers.mapIt(it.closeWait()))
+        reset(self.servers)
```
diegomrsantos, line 156: Would it be better to use `except CatchableError` instead and reraise at the end?

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
 
-    trace "Listening on", address = ma
+    try:
+      await procCall Transport(self).start(supported)
+    except CatchableError:
+      raiseAssert "Base method does not raise"
```
diegomrsantos, line 161: Could you please elaborate on what we're trying to achieve here?

diegomrsantos:
What do you think about splitting this PR into smaller ones, perhaps one for each TCP proc? There are too many changes in the TCP transport file, GitHub doesn't even load them by default. I'd argue is too risky to make all those modifications at once. If there's an issue, it might be difficult to understand the cause.

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
+      await noCancel allFutures(
+        self.servers.mapIt(it.closeWait()) &
+          self.clients[Direction.In].mapIt(it.closeWait()) &
+          self.clients[Direction.Out].mapIt(it.closeWait())
+      )
+
+      self.servers = @[]
+
+      for acceptFut in self.acceptFuts:
+        if acceptFut.completed():
```
diegomrsantos, line 194: Don't we want to cancel it when it's not completed?

libp2p/dialer.nim
arnetheduck: we could, but for that to be safe, we would need to annotate dialAndUpgrade with raises annotations as well, so that changes in `upgrate` would be caught by the compiler - this would significantly increase the scope of this PR

diegomrsantos: The code catches and raises `CancelledError`, I thought this would be equivalent to only catching `LPError`. Upgrade - the code in the `try/except` only raises those two errors. But thinking more about it, you said we still want to call `close`when catching `CancelledError`, so my suggestion probably doesn't improve anything. Btw, did you forget to call `close`?

arnetheduck: ah, highlighted wrong commit: 047019ae0fdd82630b148193c6b5284d042bc210 - close should be there

arnetheduck: > only raises those two errors.

if there is no raises annotation and the function being called starts raising a new exception, there's no way to tell which code is affected - therefore, when writing exception _handling_ code, it is reasonable to assume the worse case ("any exception can happen") even  if today there is an annotation. when we add `raises` here, we can catch more specific exceptions because changes in transport will be made visible

libp2p/transports/tcptransport.nim
arnetheduck: this is a function that didn't really change - it's handling the interface between chronos and Connection, hence its name which I believe dates back to @dryajov's participation.

libp2p/transports/tcptransport.nim
arnetheduck: this is a weakness of the inheritance / method structure of the code - ie the semantics of `running` are such that we should not mark the transport as "running" until it has been fully started (all transports are running), but the way transport in general is implemented is that calling the base method is the way to "set" running. 

The previous way of setting `running` to true before actually starting the listeners is  part of the buggy behavior that led to race conditions.

refactoring this flow would require refactoring _all_ transport implementations which is beyond the scope of this PR.

libp2p/transports/tcptransport.nim
arnetheduck: broadly, we use any logging above trace level to provide a "side channel" when the API is too limited to transport the information to the caller.

the more a library sits at the core infrastructure level of an application, the less it should log because logging is part of the application responsibilities broadly - instead of logging, it's better for libraries to provide API where the application can _choose_ to log certain events or not.

thus, any non-trace log in libp2p is really to be seen as a TODO to provide a better error reporting API to the consumer of libp2p - whether it is "correct" to log depends on use case and libp2p can provide greater flexibility to applications by providing the data needed to write good logs instead of doing the logging

libp2p/transports/tcptransport.nim
arnetheduck: the `createStreamServer` in `wire` raises all kinds of exceptions and fixing that would require significant changes elsewhere in the code. it is also using an out-of-date version of chronos.createStreamServer - making the call here directly allows improving both error reporting and avoiding deprecated api without changing all of libp2p at once

libp2p/transports/tcptransport.nim
arnetheduck: catching `CatchableError` and re-raising it means that the compiler is unable to trace the precise type of exception being raised which breaks the `raises` annotations. this way, the `raises` type annotations survive.

libp2p/transports/tcptransport.nim
arnetheduck: in a future version of the code, the base methods should be refactored to add `raises` annotations so that this try/catch can be removed - until that is done, this except is here just to silence the compiler whic.

libp2p/transports/tcptransport.nim
arnetheduck: no - only the owner of a future should cancel it - in this case, `accept` is the owner of the futures in `acceptFut` so only accept should cancel them.. 

this is something of  a grey zone in the code, ie the act of closing the transport will cause all accept futures to fail _except_ those that have already been completed, so `accept` does not strictly need to clean up `acceptFuts` even if it "owns" them.

cleaning this up would require changing the api which would require much more extensive changes, thus better left for another PR

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
+          await acceptFut.value().closeWait()
+      self.acceptFuts = @[]
+
+      # All clients should have cleaned themself up..
+      doAssert self.clients[Direction.In].len == 0
+      doAssert self.clients[Direction.Out].len == 0
```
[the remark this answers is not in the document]
arnetheduck: yes - it would indicate a logic faiult in the flow (indeed, the fault that this PR seeks to solve, namely that of lingering connection leaks).

in an ideal world, we'd actually remove `clients` entirely - broadly, it's not needed but its removal is a much more significant change

libp2p/switch.nim
arnetheduck: yes - though relying on `transport.running` runs into synchronization problems. 

The way to think about futures and indeed connections is to be clear about who owns them - in the case of `accept`, the owner is the switch so the most natural approach to "stop" accepting connections is for the switch to be in control of doing so instead of having the transport pull the rug from under the switch and have the switch recover.

there's also a simple strategy to follow when doing startup and shutdown: you shut down in the reverse order that you start things - in the switch, we create the transport, then start it, then accept connections from it, so during shutdown, the natural order is to stop accepting connections, stop the transport then get rid of it.

ditto connections: it doesn't make sense to keep accepting connections _after_ stopping the transport and yet that's what the code was doing, leading to the race condition that this PR is fixing.

libp2p/dialer.nim
diegomrsantos: Isn't it annotated in the code I mentioned?

diegomrsantos: Ah, I see what you mean now, didn't think about it. So even if the called proc is annotated, it isn't safe to trust it if the caller isn't annotated as well.

diegomrsantos: But thinking more about it, all errors in this project should inherit from `LPError`, shouldn't they? In theory, it should be fine then.

diegomrsantos: Maybe it's even more correct as `LPError` should represent all the expected errors we thought about and are fine ignoring in this case. All the others shouldn't probably be swallowed here.

libp2p/switch.nim
diegomrsantos: Makes sense, thanks for explaining.

libp2p/transports/tcptransport.nim
diegomrsantos: But couldn't the lib do both? The lib user can choose what level to use depending on their needs. Something like described here https://www.baeldung.com/hibernate-logging-levels.

libp2p/transports/tcptransport.nim
diegomrsantos: If nothing important is being done in the parent and we don't plan to change it, perhaps we can stop calling it and set the flag manually.

libp2p/transports/tcptransport.nim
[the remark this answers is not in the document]
diegomrsantos: But the application has been used so far and this condition isn't always true, i.e., this doesn't put the application in an unusable state. I wonder if it'd be better to just log a warning or even an error, but let it continue running until it's fixed.

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
+): Future[Connection] =
+  ## dial a peer
+  proc impl(
+      self: TcpTransport, hostname: string, address: MultiAddress, peerId: Opt[PeerId]
+  ): Future[Connection] {.async: (raises: [transport.TransportError, CancelledError]).} =
+    if self.stopping:
+      raise newTransportClosedError()
+
+    let ta = initTAddress(address).valueOr:
+      raise (ref TcpTransportError)(msg: "Unsupported address: " & $address)
```
diegomrsantos, line 294: why not https://nim-lang.org/docs/system.html#newException.t%2Ctypedesc%2Cstring%2Cref.Exception ?

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
+          if self.networkReachability == NetworkReachability.NotReachable and
+              self.addrs.len > 0:
+            let local = initTAddress(self.addrs[0]).expect("self address is valid")
+            self.clientFlags.incl(SocketFlags.ReusePort)
+            connect(ta, flags = self.clientFlags, localAddress = local)
+          else:
+            connect(ta, flags = self.clientFlags)
+        )
+      except CancelledError as exc:
+        raise exc
```
diegomrsantos, line 309: `await transp.closeWait()`?

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
+      except CatchableError as exc:
+        raise (ref TcpTransportError)(msg: exc.msg, parent: exc)
+
+    # If `stop` is called after `connect` but before `await` returns, we might
+    # end up with a race condition where `stop` returns but not all connections
+    # have been closed - we drop connections in this case in order not to leak
+    # them
+    if self.stopping:
```
diegomrsantos, line 317: This pattern is quite tricky and it's potentially unreliable to delegate to devs the responsibility to remember about those cases. Is there a way to automate it? Something happens when `stop` is called that cancels all pending awaited futures and the necessary cleanups happen in an `except` block.

diegomrsantos: We could add those futures in a seq in the transport when they are created, and make them remove themselves when they complete. When we stop the transport we cancel all the futures in the seq.

libp2p/transports/tcptransport.nim
```diff
@@ -12,262 +12,325 @@
...
-    info "Unexpected error accepting connection", exc = exc.msg
-    raise exc
+    let remote =
+      try:
+        transp.remoteAddress
+      except TransportOsError as exc:
+        # The connection had errors / was closed before `await` returned control
+        await transp.closeWait()
+        debug "Cannot read remote address", exc = exc.msg
+        return nil
```
diegomrsantos, line 272: why do we start returning `nil` here?

diegomrsantos (approved):

libp2p/dialer.nim
arnetheduck: Broadly, it's best when every abstraction has its own hierarchy - ie failure to decode a multihash is not related to a transport being closed in any meaningful way so having both derived from lperror doesn't really have any underlying motivation except that they happen to be implemented in the same codebase.

As such, it's usually best if exceptions are mapped to the abstraction layer that they're operating at, and each layer translates the exceptions coming from other layers to their own level - ie "socket closed" becomes "transport closed" becomes "peer disconnected" as it travels through the layers.

arnetheduck: as noted though, changing exception types is _usually_ a breaking change, so it needs to be done carefully - in this particular case, it needs to be done across the entire transport hierarchy at the same time.

libp2p/transports/tcptransport.nim
arnetheduck: that requires refactoring all transports (ws/tor too) - something beyond the scope here

libp2p/transports/tcptransport.nim
arnetheduck: doing both is usually not needed - ie if an API has a rich error handling system "built in", there's not so much value in logs because the information that you would log is available to the consumer of the library. this is not a hard rule though, only a note that different applications will have different logging needs.

it is of course possible to make everything dynamic and configurable, but such options come at the cost of performance and complexity in usage and implementation

libp2p/transports/tcptransport.nim
arnetheduck: more typing and you can't set custom fields, so `(ref Xxx)` works in more cases when exceptions have rich error information

libp2p/transports/tcptransport.nim
arnetheduck: broadly, when we transition to 2.0 and the new gc, it will be possible to make this more ergonomic - until then, it's simply a pattern to learn and repeat

libp2p/transports/tcptransport.nim
arnetheduck: it was always returning nil, by not returning anything (implicitly)

libp2p/transports/tcptransport.nim
diegomrsantos: why refactoring? it's stop calling the parent and setting the flags in the child.

arnetheduck: because to do it properly, one would remove the code from the base call (or really, change the structure so that the base contains functionality for this that doesn't involve inheritance) - then the other transports need to be updated - notably though, the base method needs refactoring anyway for `raises`, so it can be done together then.

it would be confusing to have some implementations call the base and some not

diegomrsantos:
This is flaky on mac https://github.com/vacp2p/nim-libp2p/actions/runs/9029804098/job/24812857646?pr=1095.

It says an unhandled error was raised by `testdaemon`, but all tests passed. Do you know the proper way to fix that?

arnetheduck:
> It says an unhandled error was raised by testdaemon, but all tests passed. Do you know the proper way to fix that?


no, but it would be good if this was fixed.

diegomrsantos (approved):

PR #1096, opened 2024-05-08
chore(gossipsub): cleanups

kaiserd:
factored out from: https://github.com/vacp2p/nim-libp2p/pull/1092

* add some docs
* add missing penalty for invalid messages
* simplify code
* remove unused gossipsub fields

### DIFF
--- a/.github/workflows/bumper.yml
+++ b/.github/workflows/bumper.yml
@@ -2,8 +2,7 @@ name: Bumper
 on:
   push:
     branches:
-      - unstable
-      - bumper
+      - master
   workflow_dispatch:
 
 jobs:
--- a/libp2p/dialer.nim
+++ b/libp2p/dialer.nim
@@ -81,16 +81,18 @@ proc dialAndUpgrade(
           if dialed.dir != dir:
             dialed.dir = dir
           await transport.upgrade(dialed, peerId)
+        except CancelledError as exc:
+          await dialed.close()
+          raise exc
         except CatchableError as exc:
           # If we failed to establish the connection through one transport,
           # we won't succeeded through another - no use in trying again
           await dialed.close()
           debug "Connection upgrade failed", err = exc.msg, peerId = peerId.get(default(PeerId))
-          if exc isnot CancelledError:
-            if dialed.dir == Direction.Out:
-              libp2p_failed_upgrades_outgoing.inc()
-            else:
-              libp2p_failed_upgrades_incoming.inc()
+          if dialed.dir == Direction.Out:
+            libp2p_failed_upgrades_outgoing.inc()
+          else:
+            libp2p_failed_upgrades_incoming.inc()
 
           # Try other address
           return nil
--- a/libp2p/errors.nim
+++ b/libp2p/errors.nim
@@ -44,12 +44,3 @@ macro checkFutures*[F](futs: seq[F], exclude: untyped = []): untyped =
             # We still don't abort but warn
             debug "A future has failed, enable trace logging for details", error=exc.name
             trace "Exception details", msg=exc.msg
-
-template tryAndWarn*(message: static[string]; body: untyped): untyped =
-  try:
-    body
-  except CancelledError as exc:
-    raise exc
-  except CatchableError as exc:
-    debug "An exception has ocurred, enable trace logging for details", name = exc.name, msg = message
-    trace "Exception details", exc = exc.msg
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -273,6 +273,7 @@ proc accept(s: Switch, transport: Transport) {.async.} = # noraises
     except CancelledError as exc:
       trace "releasing semaphore on cancellation"
       upgrades.release() # always release the slot
+      return
     except CatchableError as exc:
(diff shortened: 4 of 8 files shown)


diegomrsantos:
Should we use the conventional commits style for the PR's names?

diegomrsantos:
I believe only the PR title is committed into master when we squash and merge the PR.

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -57,10 +57,10 @@ proc addSeen*(f: FloodSub, saltedId: SaltedId): bool =
...
   f.seen.addedAt(saltedId)
 
-proc handleSubscribe*(f: FloodSub,
-                      peer: PubSubPeer,
-                      topic: string,
-                      subscribe: bool) =
+proc handleSubscribe(f: FloodSub,
+                     peer: PubSubPeer,
+                     topic: string,
+                     subscribe: bool) =
```
diegomrsantos, line 63: it might be better to follow this pattern https://github.com/vacp2p/nim-libp2p/pull/1095/files#diff-94a6536a89bc744b23b27b6d4577eee57ed6e4ed0b1f13f077b9db22c16ae8b8R280-R284. @arnetheduck how are you formatting it? Let's add it to the readme so we all share the same formatting rules.

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -106,10 +106,9 @@ method unsubscribePeer*(f: FloodSub, peer: PeerId) =
 method rpcHandler*(f: FloodSub,
                    peer: PubSubPeer,
                    data: seq[byte]) {.async.} =
-
   var rpcMsg = decodeRpcMsg(data).valueOr:
     debug "failed to decode msg from peer", peer, err = error
-    raise newException(CatchableError, "")
+    raise newException(CatchableError, "Peer msg couldn't be decoded")
```
diegomrsantos, line 111: I believe we should either log and discard the exception or create a sequence Parent/Child that gets logged only once. Unfortunately, this seems not possible in Nim. I created this issue some time ago, but it was closed without a solution https://github.com/nim-lang/Nim/issues/20677. @arnetheduck thought?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -630,38 +631,39 @@ method publish*(g: GossipSub,
...
-    var fanoutPeers = g.fanout.getOrDefault(topic).toSeq()
-    if fanoutPeers.len < g.parameters.dLow:
-      g.replenishFanout(topic)
-      fanoutPeers = g.fanout.getOrDefault(topic).toSeq()
+  elif peers.len < g.parameters.dLow:
+    # not subscribed or bad mesh, send to fanout peers
+    # when flood-publishing, fanout won't help since all potential peers have
+    # already been added
 
+    g.replenishFanout(topic) # Make sure fanout is populated
```
diegomrsantos, line 656: Why has the `replenishFanout` condition been modified?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -630,38 +631,39 @@ method publish*(g: GossipSub,
...
 
-  if peers.len < g.parameters.dLow:
-    # not subscribed, or bad mesh, send to fanout peers
-    var fanoutPeers = g.fanout.getOrDefault(topic).toSeq()
-    if fanoutPeers.len < g.parameters.dLow:
-      g.replenishFanout(topic)
-      fanoutPeers = g.fanout.getOrDefault(topic).toSeq()
+  elif peers.len < g.parameters.dLow:
+    # not subscribed or bad mesh, send to fanout peers
+    # when flood-publishing, fanout won't help since all potential peers have
```
diegomrsantos, line 653: not sure this is true, I guess only if we assume that `peers.len < g.parameters.dLow` is always false when flood publishing is enabled

libp2p/protocols/pubsub/floodsub.nim
arnetheduck: my preference is to format code with https://github.com/arnetheduck/nph which broadly follows https://status-im.github.io/nim-style-guide/formatting.style.html

libp2p/protocols/pubsub/floodsub.nim
arnetheduck: you can: `(ref CatchableError)(mgs: "test", parent: otherException)` - when catching, you can use the parent field to print all parents as well - this latter thing does not exist but the core support in nim is there.

libp2p/protocols/pubsub/gossipsub.nim
arnetheduck: the same condition is checked inside of `replenishFanout` already

libp2p/protocols/pubsub/gossipsub.nim
arnetheduck: when flood publishing is turned on, all possible peers will already have been added to the possible peers to send to - ie everyone that is subscribed gets a message, so there's no point to try again with fanout peers (which are a subset)

diegomrsantos (approved):

PR #1100, opened 2024-05-20
Staggered message sending with elimination of peers during transmission based on idontwants

ufarooqstatus:
1) We use semaphores to limit simultaneous transmissions to two 
2) During message transmission, we check received idontwants to cancel unnecessary transmissions
3) we relay to outbound peers first to minimize chances of two peers sending to each other

closes [PR1093](https://github.com/vacp2p/nim-libp2p/pull/1093)

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -418,10 +418,19 @@ proc validateAndRelay(g: GossipSub,
           break
     toSendPeers.excl(seenPeers)
 
+    #We first send to the outbound peers to avoid peers sending same message to each other
+    var outboundPeers: seq[PubSubPeer]
+    for mpeer in toSendPeers:
+      if mpeer.outbound():
+        outboundPeers.add(mpeer)
+    if outboundPeers.len > 0:
+      g.broadcast(outboundPeers, RPCMsg(messages: @[msg]), isHighPriority = false, some(saltedId))
+      toSendPeers.excl(outboundPeers.toHashSet)
+
     # In theory, if topics are the same in all messages, we could batch - we'd
     # also have to be careful to only include validated messages
-    g.broadcast(toSendPeers, RPCMsg(messages: @[msg]), isHighPriority = false)
-    trace "forwarded message to peers", peers = toSendPeers.len, msgId, peer
+    g.broadcast(toSendPeers, RPCMsg(messages: @[msg]), isHighPriority = false, some(saltedId))
+    trace "forwarded message to peers", peers = toSendPeers.len + outboundPeers.len, msgId, peer
 
     if g.knownTopics.contains(topic):
       libp2p_pubsub_messages_rebroadcasted.inc(toSendPeers.len.int64, labelValues = [topic])
@@ -528,6 +537,11 @@ method rpcHandler*(g: GossipSub,
         g.rewardDelivered(peer, topic, false, delay)
 
       libp2p_gossipsub_duplicate.inc()
+
+      #Dont relay to the peers from which we already received 
+      #We do it for large messages only
+      if msg.data.len > msgId.len * 10:     
+        peer.heDontWants[^1].incl(msgIdSalted)
 
       # onto the next message
       continue
@@ -700,15 +714,23 @@ method publish*(g: GossipSub,
 
   g.mcache.put(msgId, msg)
 
+  #We first send to the outbound peers
+  var outboundPeers: seq[PubSubPeer]
+  for mpeer in peers:
+    if mpeer.outbound():
+      outboundPeers.add(mpeer)
+  if outboundPeers.len > 0:
+    g.broadcast(outboundPeers, RPCMsg(messages: @[msg]), isHighPriority = true)
+    peers.excl(outboundPeers.toHashSet)
   g.broadcast(peers, RPCMsg(messages: @[msg]), isHighPriority = true)
 
   if g.knownTopics.contains(topic):
-    libp2p_pubsub_messages_published.inc(peers.len.int64, labelValues = [topic])
+    libp2p_pubsub_messages_published.inc( (peers.len + outboundPeers.len).int64, labelValues = [topic])
   else:
-    libp2p_pubsub_messages_published.inc(peers.len.int64, labelValues = ["generic"])
+    libp2p_pubsub_messages_published.inc( (peers.len + outboundPeers.len).int64, labelValues = ["generic"])
 
-  trace "Published message to peers", peers=peers.len
-  return peers.len
+  trace "Published message to peers", peers=peers.len + outboundPeers.len
+  return (peers.len + outboundPeers.len)
 
 proc maintainDirectPeer(g: GossipSub, id: PeerId, addrs: seq[MultiAddress]) {.async.} =
   if id notin g.peers:
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -19,7 +19,9 @@ import rpc/[messages, message, protobuf],
        ../../stream/connection,
        ../../crypto/crypto,
        ../../protobuf/minprotobuf,
-       ../../utility
+       ../../utility,
+       ../../utils/semaphore
+import atomics
 
 export peerid, connection, deques
 
@@ -36,6 +38,12 @@ when defined(pubsubpeer_queue_metrics):
   declareGauge(libp2p_gossipsub_non_priority_queue_size, "the number of messages in the non-priority queue", labels = ["id"])
 
 declareCounter(libp2p_pubsub_disconnects_over_non_priority_queue_limit, "number of peers disconnected due to over non-prio queue capacity")
+
+var 
+  libp2p_gossipsub_staggerDontWantSave2: Atomic[int]
+  libp2p_gossipsub_staggerDontWantSave3: Atomic[int]
+
+export libp2p_gossipsub_staggerDontWantSave2, libp2p_gossipsub_staggerDontWantSave3
 
 const
   DefaultMaxNumElementsInNonPriorityQueue* = 1024
@@ -59,11 +67,15 @@ type
   DropConn* = proc(peer: PubSubPeer) {.gcsafe, raises: [].} # have to pass peer as it's unknown during init
   OnEvent* = proc(peer: PubSubPeer, event: PubSubPeerEvent) {.gcsafe, raises: [].}
 
+  MessageWithSaltedId = object
+    message: seq[byte]
+    sid: SaltedId
+
   RpcMessageQueue* = ref object
     # Tracks async tasks for sending high-priority peer-published messages.
     sendPriorityQueue: Deque[Future[void]]
     # Queue for lower-priority messages, like "IWANT" replies and relay messages.
-    nonPriorityQueue: AsyncQueue[seq[byte]]
+    nonPriorityQueue: AsyncQueue[MessageWithSaltedId]
     # Task for processing non-priority message queue.
     sendNonPriorityTask: Future[void]
 
@@ -91,6 +103,7 @@ type
     behaviourPenalty*: float64 # the eventual penalty score
     overheadRateLimitOpt*: Opt[TokenBucket]
 
+    semTxLimit: ptr AsyncSemaphore             #Control Max simultaneous transmissions to speed up indivisual receptions
     rpcmessagequeue: RpcMessageQueue
     maxNumElementsInNonPriorityQueue*: int # The max number of elements allowed in the non-priority queue.
     disconnected: bool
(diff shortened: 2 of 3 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -529,6 +538,11 @@ method rpcHandler*(g: GossipSub,
 
       libp2p_gossipsub_duplicate.inc()
 
+      #Dont relay to the peers from which we already received 
+      #We do it for large messages only
+      if msg.data.len > msgId.len * 10:
```
arnetheduck, line 543: if the message is in `seen`, it should not be sent to anyone meaning we should not need to track it individually (doing so costs significant amounts of memory)

arnetheduck: actually, right - the seen check has been done and we need to stop the message after the seen check . hm . it would likely be better that we keep track of in-flight / queued messages and cancel these sends specifically on idontwant rather than storing all idontwant hashes

arnetheduck:
broadly, limiting concurrent sends this way means two peers can hold up the queue for everyone else, including the fast peers - this is a key reason we haven't gone down this route.

the limit would have to be a lot higher - ie somewhere on the order of dmax probably - we would also need to consider a timeout policy of individual sends stronger than what we have today.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -418,10 +418,19 @@ proc validateAndRelay(g: GossipSub,
           break
     toSendPeers.excl(seenPeers)
 
+    #We first send to the outbound peers to avoid peers sending same message to each other
```
diegomrsantos, line 421: Could you please elaborate on why that is the case?

libp2p/protocols/pubsub/gossipsub.nim
diegomrsantos: Don't we already have this here https://github.com/vacp2p/nim-libp2p/blob/0911cb20f4a0b880451938988f24cb077cfd8b1b/libp2p/protocols/pubsub/gossipsub.nim#L319 ?

diegomrsantos: > actually, right - the seen check has been done and we need to stop the message after the seen check . hm . it would likely be better that we keep track of in-flight / queued messages and cancel these sends specifically on idontwant rather than storing all idontwant hashes

Haven't this been already tried here? https://github.com/vacp2p/nim-libp2p/pull/851#issuecomment-1586818191

diegomrsantos: I believe `heDontWants` is meant to handle `idontwant` msgs received and we already have the seen data structure to avoid relaying duplicates.

diegomrsantos: We could check the `idowant` again when removing a msg from the non-prio queue. We would need to:

1.  Store the msg ID in the queue.
2. Have a `ref` to the structure in the `Pubsubpeer` object.

ufarooqstatus: > actually, right - the seen check has been done and we need to stop the message after the seen check . hm . it would likely be better that we keep track of in-flight / queued messages and cancel these sends specifically on idontwant rather than storing all idontwant hashes

Yes, for 12 (1MB) message transmissions in a 1000 peer network, we cancel roughly 8k-10k messages [here](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub.nim#L412-L419). But rechecking for received idontwants before [actual transmission](https://github.com/vacp2p/nim-libp2p/pull/1100/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R348-R353) eliminates additional 1K-2K messages. [In-flight message elimination](https://github.com/vacp2p/nim-libp2p/pull/1100/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R355-R365) also filters 6K-8K messages, but benefits of in-flight message cancellation are questionable IMO

ufarooqstatus: > I believe `heDontWants` is meant to handle `idontwant` msgs received and we already have the `seen` data structure to avoid relaying duplicates.

IMO we can use heDontWants just before forwarding message to the peer to check if we have already received IDontWant [here](https://github.com/vacp2p/nim-libp2p/pull/1100/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R346-R353)

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -311,24 +328,49 @@ proc sendMsgSlow(p: PubSubPeer, msg: seq[byte]) {.async.} =
...
+        if not f.completed:
+          await f.cancelAndWait()
+          #asyncSpawn sendMsgContinue(conn, f)
+        p.semTxLimit[].release()
+      
+      except LPStreamError as exc:
+        p.semTxLimit[].release()
+        await conn.close()
+    
+proc sendEncoded*(p: PubSubPeer, msg: seq[byte], isHighPriority: bool, saltedId: Option[SaltedId] = none(SaltedId)): Future[void] =
```
diegomrsantos, line 373: why is saltedId optional?

libp2p/protocols/pubsub/gossipsub.nim
ufarooqstatus: > We could check the `idontwant` again when removing a msg from the non-prio queue. We would need to:
> 
> 1. Store the msg ID in the queue.

Yes, its done [here](https://github.com/vacp2p/nim-libp2p/pull/1100/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R70-R78)

> 2. Have a `ref` to the structure in the `Pubsubpeer` object.

I'm [passing message ID in this case](https://github.com/vacp2p/nim-libp2p/pull/1100/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R331)

libp2p/protocols/pubsub/gossipsub.nim
ufarooqstatus: It was intended to avoid A sending to B and B sending to A at the same time,,, but yes, its not making much impact,,, due to the preceding IDontWants

libp2p/protocols/pubsub/gossipsub.nim
diegomrsantos: > > We could check the `idontwant` again when removing a msg from the non-prio queue. We would need to:
> > 
> > 1. Store the msg ID in the queue.
> 
> Yes, its done [here](https://github.com/vacp2p/nim-libp2p/pull/1100/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R70-R78)
> 
> > 2. Have a `ref` to the structure in the `Pubsubpeer` object.
> 
> I'm [passing message ID in this case](https://github.com/vacp2p/nim-libp2p/pull/1100/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R331)

Yes, I commented before reading the whole PR. I deleted the comment but you were still able to reply to it.

libp2p/protocols/pubsub/pubsubpeer.nim
ufarooqstatus: [heDontWants is storing saltedId](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub/behavior.nim#L271)

ufarooqstatus: > why is saltedId optional?

same SendMsg for control and published messages, so for some messages we did not need saltedID

diegomrsantos: Currently, we can improve this design as I believe high-prio are only control and msgs published by the local peer. It means no other peer can send an `idontwant` for them.  It means we can move the `idontwant` to the non-prio branch in this proc.

ufarooqstatus:
> broadly, limiting concurrent sends this way means two peers can hold up the queue for everyone else, including the fast peers - this is a key reason we haven't gone down this route.

This is [configurable](https://github.com/vacp2p/nim-libp2p/pull/1100/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R355-R367), but i could find very similar results for 2-5 simultaneous sends

> the limit would have to be a lot higher - ie somewhere on the order of dmax probably - we would also need to consider a timeout policy of individual sends stronger than what we have today.

I have tried [this way to time-limit](https://github.com/vacp2p/nim-libp2p/pull/1100/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R355-R367)

libp2p/protocols/pubsub/pubsubpeer.nim
diegomrsantos: In fact, If we were enqueueing those msgs and we didn't dequeue them fast enough, peers could receive them via peers we sent first. But as we always send them immediately, it seems this can't happen. Makes sense?

libp2p/protocols/pubsub/gossipsub.nim
ufarooqstatus: > > > I believe `heDontWants` is meant to handle `idontwant` msgs received and we already have the `seen` data structure to avoid relaying duplicates.
> > 
> > 
> > IMO we can use heDontWants just before forwarding message to the peer to check if we have already received IDontWant [here](https://github.com/vacp2p/nim-libp2p/pull/1100/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R346-R353)
> 
> Yes, that's what I meant. But here it seems to me that you are adding a message id to the peer list even though the peer didn't send an `idontwant` for this id. That's not the purpose of the data structure.

If we received a message from a peer, it implies that peer does not want that message (we can kind of reuse hedontwants data structure for faster elimination of messages). otherwise we'll have to check both, seen cache and hedontwants

libp2p/protocols/pubsub/pubsubpeer.nim
ufarooqstatus: > Currently, we can improve this design as I believe high-prio are only control and msgs published by the local peer. It means no other peer can send an `idontwant` for them. It means we can move the `idontwant` to the non-prio branch in this proc. Wdyt?

yes, we dont need to check hedontwants for the messages published by the local peer,,, but i guess IDontWant should be transmitted with the highest priority

ufarooqstatus: In fact, i believe, IDontWant should be issued before [this step](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub.nim#L366). Because even if a message is not valid, we can react early to stop the spread of such message. And in case of valid messages, we get to learn early that our peers have received this message

ufarooqstatus:
> I have tried [this way to time-limit](https://github.com/vacp2p/nim-libp2p/pull/1100/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R355-R367)

In fact (at-least) a slow peer can identify itself as a slow peer if it can not finish all sends within a strict time-limit. Also the peers for which we always miss the timebound. Thinking to try different locks for fast/slow peers

libp2p/protocols/pubsub/gossipsub.nim
diegomrsantos: I believe we already don't send a message from a peer we've just received it and also don't relay the same msg twice. Isn't this equivalent?

diegomrsantos: This data structure is probably to store ID of msgs other peers in the network have already seen but we haven't yet. Then when we do see it, we don't send it to them.

ufarooqstatus: IMO our send inserts the message in non-priority queue, and may consume some time before its aired. if we receive the message during this interval, we can still cancel transmit

libp2p/protocols/pubsub/pubsubpeer.nim
diegomrsantos: I meant that checking `idontwant` before sending a msg could be done only for msgs relayed at line 417 of this file.

libp2p/protocols/pubsub/gossipsub.nim
diegomrsantos: True, but https://github.com/vacp2p/nim-libp2p/blob/0911cb20f4a0b880451938988f24cb077cfd8b1b/libp2p/protocols/pubsub/gossipsub.nim#L319 and checking `idontwant` when dequeuing from non-prio are enough. We don't seem to need this code for that.

diegomrsantos: Maybe we could even end up not sending the msg to the current peer as it is enqueued and before sending we check `idontwant`.

ufarooqstatus: yes, this part does not have much impact!

libp2p/protocols/pubsub/pubsubpeer.nim
ufarooqstatus: yes for relayed messages only

PR #1103, opened 2024-05-25
Send IDONTWANT before validating message

arnetheduck:
Even before we've done validation, we already know we don't want additional copies of the message.

Also, increase the minimum threshold for sending IDONTWANT to 512 to prevent spam in case of very small message id:s

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -363,6 +363,28 @@ proc validateAndRelay(g: GossipSub,
                       msgId: MessageId, saltedId: SaltedId,
                       peer: PubSubPeer) {.async.} =
   try:
+    template topic: string = msg.topic
+
+    if msg.data.len > max(512, msgId.len * 10):
+      # If the message is "large enough", let the mesh know that we do not want
+      # any more copies of it, regardless if it is valid or not.
+      #
+      # In the case that it is not valid, this leads to some redundancy
+      # (since the other peer should not send us an invalid message regardless),
+      # but the expectation is that this is rare (due to such peers getting
+      # descorded) and that the savings from honest peers are greater than the
+      # cost a dishonest peer can incur in short time (since the IDONTWANT is
+      # small).
+      var toSendPeers = HashSet[PubSubPeer]()
+      g.floodsub.withValue(topic, peers): toSendPeers.incl(peers[])
+      g.mesh.withValue(topic, peers): toSendPeers.incl(peers[])
+      g.subscribedDirectPeers.withValue(topic, peers): toSendPeers.incl(peers[])
+      toSendPeers.excl(peer)
+
+      g.broadcast(toSendPeers, RPCMsg(control: some(ControlMessage(
+          idontwant: @[ControlIWant(messageIDs: @[msgId])]
+        ))), isHighPriority = true)
+
     let validation = await g.validate(msg)
 
     var seenPeers: HashSet[PubSubPeer]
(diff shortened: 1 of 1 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -363,6 +363,28 @@ proc validateAndRelay(g: GossipSub,
...
+    template topic: string = msg.topic
+
+    if msg.data.len > max(512, msgId.len * 10):
+      # If the message is "large enough", let the mesh know that we do not want
+      # any more copies of it, regardless if it is valid or not.
+      #
+      # In the case that it is not valid, this leads to some redundancy
+      # (since the other peer should not send us an invalid message regardless),
+      # but the expectation is that this is rare (due to such peers getting
+      # descorded) and that the savings from honest peers are greater than the
```
diegomrsantos, line 375:
```suggestion
      # descored) and that the savings from honest peers are greater than the
```

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -363,6 +363,28 @@ proc validateAndRelay(g: GossipSub,
                       msgId: MessageId, saltedId: SaltedId,
                       peer: PubSubPeer) {.async.} =
   try:
+    template topic: string = msg.topic
+
+    if msg.data.len > max(512, msgId.len * 10):
+      # If the message is "large enough", let the mesh know that we do not want
```
diegomrsantos, line 369: But could a malicious peer flood the network with invalid msgs and thus `idontwant` faster than it gets punished? It might require some testing.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -402,21 +425,13 @@ proc validateAndRelay(g: GossipSub,
...
-    if msg.data.len > msgId.len * 10:
-      g.broadcast(toSendPeers, RPCMsg(control: some(ControlMessage(
-          idontwant: @[ControlIWant(messageIDs: @[msgId])]
-        ))), isHighPriority = true)
-
     for peer in toSendPeers:
       for heDontWant in peer.heDontWants:
         if saltedId in heDontWant:
-          seenPeers.incl(peer)
+          toSendPeers.excl(peer)
```
diegomrsantos, line 431: This causes `len(s) == length` the length of the HashSet changed while iterating over it [AssertionDefect]`. Pushing fix.

diegomrsantos (approved):

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -363,6 +363,28 @@ proc validateAndRelay(g: GossipSub,
...
+      # (since the other peer should not send us an invalid message regardless),
+      # but the expectation is that this is rare (due to such peers getting
+      # descored) and that the savings from honest peers are greater than the
+      # cost a dishonest peer can incur in short time (since the IDONTWANT is
+      # small).
+      var toSendPeers = HashSet[PubSubPeer]()
+      g.floodsub.withValue(topic, peers): toSendPeers.incl(peers[])
+      g.mesh.withValue(topic, peers): toSendPeers.incl(peers[])
+      g.subscribedDirectPeers.withValue(topic, peers): toSendPeers.incl(peers[])
+      toSendPeers.excl(peer)
```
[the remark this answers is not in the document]
diegomrsantos: I find the version I pushed slightly more readable wdyt? I fine with it being merged. 

Btw, isn't there a way to prevent Nim from making a copy when doing assignments of seq, string, HashSet, etc?

Issue #1102, opened 2024-05-24
Support gossipsub 1.2.0


arnetheduck:
https://github.com/libp2p/specs/pull/548 - this has been merged and IDONTWANT is officially part of the spec - the feature is formally only enabled in gossipsub 1.2.0, meaning we must start supporting that version as a protocol in order to receive IDONTWANT messages from other clients (putting the fields in 1.1 was a hack).

diegomrsantos:
Is it enough to create a new version here https://github.com/vacp2p/nim-libp2p/blob/0911cb20f4a0b880451938988f24cb077cfd8b1b/libp2p/protocols/pubsub/gossipsub/types.nim#L22 and add it here https://github.com/vacp2p/nim-libp2p/blob/0911cb20f4a0b880451938988f24cb077cfd8b1b/libp2p/protocols/pubsub/gossipsub.nim#L197 ?

ufarooqstatus:
[Gossipsub-v1.2 specification]( https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md?plain=1#L81-L82) suggest that we MAY cancel outstanding IWANT if we receive the message from some other peer 

[This specification]( https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md?plain=1#L83) also suggests that we SHOULD cancel any queued IWANT reply if we receive IDONTWANT from the requesting peer. IMO, we have this available in this [draft PR]( https://github.com/vacp2p/nim-libp2p/pull/1100/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R349-R352)

PR #1106, opened 2024-05-27
feat(gossipsub): support version 1.2.0

diegomrsantos:
- Support GossipSub 1.2 in the code
- Add utility template `template exclIfIt*[T](set: var HashSet[T], condition: untyped)`
- Renaming

closes https://github.com/vacp2p/nim-libp2p/issues/1102

### DIFF
--- a/tests/pubsub/utils.nim
+++ b/tests/pubsub/utils.nim
@@ -28,9 +28,9 @@ type TestGossipSub* = ref object of GossipSub
 
 proc getPubSubPeer*(p: TestGossipSub, peerId: PeerId): PubSubPeer =
   proc getConn(): Future[Connection] =
-    p.switch.dial(peerId, GossipSubCodec)
+    p.switch.dial(peerId, GossipSubCodec_12)
 
-  let pubSubPeer = PubSubPeer.new(peerId, getConn, nil, GossipSubCodec, 1024 * 1024)
+  let pubSubPeer = PubSubPeer.new(peerId, getConn, nil, GossipSubCodec_12, 1024 * 1024)
   debug "created new pubsub peer", peerId
 
   p.peers[peerId] = pubSubPeer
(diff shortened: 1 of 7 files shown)


diegomrsantos:
This PR depends on this being fixed https://github.com/vacp2p/nim-libp2p/issues/1115

diegomrsantos:
- Send `idontwant` only to peers that support GossipSub 1.2
- Add utility template `template exclIfIt*[T](set: var HashSet[T], condition: untyped)`

Those are gonna be moved to a new PR cause of https://github.com/vacp2p/nim-libp2p/issues/1115.

diegomrsantos:
> I'm not an expert on GS by any means, but I'm expecting something like `if GossipSubCodec_12 in g.codecs: # do something GS1.2 related`. From what I understand, we just add the codecs (which needed to be done) but there's no difference between GossipSubCodec_10, GossipSubCodec_11 and GossipSubCodec_12.

I believe `g.codecs` are the versions supported by the current code, instead of the one negotiated with the peer - `pubsubpeer.codec`. Behaving like 1.2 - sending `idontwant` - will be implemented in a new PR as mentioned here: https://github.com/vacp2p/nim-libp2p/pull/1106#issuecomment-2149411500. 

It seems that our current code can negotiate 1.0 but will still behave like 1.1, as there's no check on how the protocol should behave depending on the version negotiated with a remote peer.

tests/pubsub/utils.nim
```diff
@@ -26,9 +26,9 @@ type
 
 proc getPubSubPeer*(p: TestGossipSub, peerId: PeerId): PubSubPeer =
   proc getConn(): Future[Connection] =
-    p.switch.dial(peerId, GossipSubCodec)
+    p.switch.dial(peerId, GossipSubCodec_12)
```
AlejandroCabeza, line 29: What happens if dialing 1.2.0 fails? Should we attempt to dial with a lower version?

diegomrsantos: This is only for tests. Production code tries all versions.

lchenut (approved):

lchenut (approved):

Issue #1101, opened 2024-05-24
IHave / IWant message behavior seriously impacts large message transmissions.


ufarooqstatus:
IHave / IWant message behavior seriously impacts large message transmissions.

In a realistic network scenario (with peers having dissimilar bandwidth/latencies), some peers receive messages much earlier than many other peers in the network (The time difference increases with the message size). 

The early receivers [spread Ihave messages](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub/behavior.nim#L683-L691). On getting an ihave all remaining peers immidiately [send iwant message](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub.nim#L320).

This behavior has some inherent problems (resulting in significantly higher network-wide message dissemination time/fluctuations)

1) On receiving an ihave, [a peer only checks for seen cache, peer score, and ihave budget](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub/behavior.nim#L243-L257). So potentially, a peer can issue multiple iwant messages against the same message ID to different peers. 
2) Large message reception may consume significant time, so a peer can issue an iwant message even if it is already receiving the same message
3) The peer issuing ihave message will have to service many peers in addition to its mesh ([as we only check peer score, msgID_Validation, and single transmission to any peer](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub/behavior.nim#L279-L286)). 

The following results compare gossipsub (current arrangement) with and without IWant messages in a 1000-node network. The bandwidth/latency are uniformly distributed between 50-150Mbps and 40-160ms in 4 stages. The results are averaged for 10 message transmissions (with two initial warm-up messages to raise Cwnd at peers) 

![image](https://github.com/vacp2p/nim-libp2p/assets/137903752/ffee32c5-3855-48ea-8eaa-1497c4f22014)

ufarooqstatus:
Obviously, disabling IWant messages is not the right choice. However, some possible work arounds can be to

1) Send only 1 IWant message against any msg ID
2) For large messages, a peer should send IHave only after it has relayed the message to all (or most of its mesh members)
3) A peer should not issue IWant message, against a message it is already receiving (will require knowledge of in-flight receives)

[...]

PR #1109, opened 2024-06-03
Skip sending IWants based on received IDontWants

ufarooqstatus:
We use IDontWant followed by the actual message transmission to minimize redundant large message transmissions. 

We can also skip sending IWant for a msgId if we have already received N number of IDontWants for the same msgId


ufarooqstatus:
https://github.com/vacp2p/nim-libp2p/issues/1101

PR #1111, opened 2024-06-04
Update go-libp2p-daemon build version.

AlejandroCabeza:
Fix build errors by updating the used version to a working one. Also, it's the one used in the CI for testing.


diegomrsantos:
@AlejandroCabeza could please explain what build errors this is fixing?

@kaiserd If I remember correctly, I tried to use a newer version and it didn't work. It's not clear to me why this daemon is even necessary. I'd suggest removing it altogether.

AlejandroCabeza:
> Thank you! Is there a reason why we do not used the tagged versions? v0.35.0 is only 14 days old. Seems to be recent enough. We could update the CI to that, too.

@kaiserd I'm fine having a look to see whether we can update the daemon to a more recent version. In any case I'd merge this PR regardless, and take that as a follow-up task.

> could please explain what build errors this is fixing?

@diegomrsantos Just build errors, I'm not a go expert so I can't elaborate. But it's as simple as running the build script as it is, you'll get those errors.

> It's not clear to me why this daemon is even necessary. I'd suggest removing it altogether.

@diegomrsantos The `p2pd` binary is used for running tests. I'm positive that's mentioned in the README.

diegomrsantos:
@AlejandroCabeza I meant when those build errors occur. As I've done this a long time ago, I don't remember. Using the current script to install it locally doesn't work?

 I meant it's not clear to me why we need `p2pd` to run tests as the vast majority of them don't need it. I think it's required only for 4 or so. I suggest investigating it and either removing all those tests or removing the dependency on the daemon.

AlejandroCabeza:
> I meant when those build errors occur. As I've done this a long time ago, I don't remember. Using the current script to install it locally doesn't work?
@diegomrsantos Correct, running it fails due to some `xerrors` library (or something like that).

> I meant it's not clear to me why we need p2pd to run tests as the vast majority of them don't need it. I think it's required only for 4 or so. I suggest investigating it and either removing all those tests or removing the dependency on the
@diegomrsantos No clue, but I think that's a different job altogether. If you think it's worth considering, would you mind writing an issue for it so we can discuss it during next weekly?

diegomrsantos (approved):

Issue #1115, opened 2024-06-05
Pubsubpeer has wrong protocol codecs


diegomrsantos:
When [subscribePeer](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/pubsub.nim#L397) runs, a remote peer `PubSubPeer` is created, but with the local peer `pubsub.codecs` instead of those from the remote one. The remote peer is added to the local peer's `peers` table. When [handleConn](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/pubsub.nim#L379) executes - now with the right codec - the remote peer continues with the wrong codecs, as it's already present in the table and it isn't created again or modified.

PR #1116, opened 2024-06-05
fix(gossipsub): pubsubpeer is created with wrong gossipsub version

diegomrsantos:
closes https://github.com/vacp2p/nim-libp2p/issues/1115

### DIFF
--- a/libp2p/protocols/pubsub/pubsub.nim
+++ b/libp2p/protocols/pubsub/pubsub.nim
@@ -348,19 +348,22 @@ method onPubSubPeerEvent*(
     discard
 
 method getOrCreatePeer*(
-    p: PubSub, peerId: PeerId, protos: seq[string]
-): PubSubPeer {.base, gcsafe.} =
+    p: PubSub, peerId: PeerId, protosToDial: seq[string]
+,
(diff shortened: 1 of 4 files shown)


libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -298,19 +298,22 @@ method onPubSubPeerEvent*(p: PubSub, peer: PubSubPeer, event: PubSubPeerEvent) {
 method getOrCreatePeer*(
     p: PubSub,
     peerId: PeerId,
-    protos: seq[string]): PubSubPeer {.base, gcsafe.} =
+    protosToDial: seq[string],
+    protoNegotiated: string = ""): PubSubPeer {.base, gcsafe.} =
 
   p.peers.withValue(peerId, peer):
+    if peer[].codec == "" and protoNegotiated != "":
```
diegomrsantos, line 305: This solution is ugly, but couldn't think of anything better. Ideas are welcome.

lchenut: `if peer[].codec == "":` will have the same result in the end. And it's a bit better, I guess?

diegomrsantos: True

PR #1117, opened 2024-06-05
chore: Cleanup CI

AlejandroCabeza:


AlejandroCabeza:
> Thank you 🙌 Can you please elaborate a bit on the proposed changes?

@kaiserd Still working on it, but the idea is threefold: Understand precisely how the workflows behave (looking to properly implement the maxver and minver ones we talked about months ago), reduce duplicated workflows, and improve naming and descriptions (so everyone can understand them better).

.github/workflows/coverage.yml
```diff
@@ -1,20 +1,23 @@
...
+name: Coverage
 
 on:
-  #On push to common branches, this computes the "bases stats" for PRs
+  # On push to common branches, this computes the coverage PRs will compared against
   push:
     branches:
       - master
-      - unstable
   pull_request:
```
diegomrsantos, line 8: I believe it should be as before, i.e., run for every PR.

.github/workflows/coverage.yml
```diff
@@ -1,20 +1,23 @@
...
+    branches:
```
diegomrsantos, line 9: I believe it should be as before, i.e., run for every PR.

.github/workflows/ci.yml
```diff
@@ -1,92 +0,0 @@
-name: CI
```
diegomrsantos, line 1: Why has this file been deleted?

.github/workflows/base_tests.yml
```diff
@@ -71,14 +88,15 @@ jobs:
...
+      
+      - name: Install deps
+        run: |
+          nimble install -y --depsOnly
 
       - name: Run tests
         run: |
           nim --version
           nimble --version
-          nimble install -y --depsOnly
```
diegomrsantos, line 79: why has `nimble install -y --depsOnly` been deleted?

.github/workflows/ci.yml
AlejandroCabeza: Answered in the other comment.

.github/workflows/base_tests.yml
AlejandroCabeza: Not deleted, moved. I'm trying to remove one of the (almost) duplicate files. Currently in the process of checking whether everything behaves the same, and tests pass.

.github/workflows/coverage.yml
AlejandroCabeza: I'd suggest not taking the PR too seriously yet. It's on `Draft` and I haven't assigned reviewers yet for a reason :)

In any case, answering your question: The reason is there's a limit on the amount of codecov uploads with a free account. Maybe I'm getting too ahead of the problem, but we can discuss that later, when I finish making everything work :)

.github/workflows/dependencies.yml
```diff
@@ -34,6 +36,7 @@ jobs:
           git checkout $GITHUB_SHA
 
       - name: Commit this bump
+        # Is it mandatory to create a new commit? Or does it exist as a way to communicate why the version was updated.
```
diegomrsantos, line 39: Did you forget to remove this comment?

.github/workflows/dependencies.yml
```diff
@@ -1,13 +1,15 @@
...
 on:
   push:
     branches:
       - master
   workflow_dispatch:
 
 jobs:
-  bumpProjects:
+  update_version:
+    # Is this job necessary?
```
diegomrsantos, line 10: There might be a misunderstanding regarding comments. They aren't for asking questions about the code.

.github/workflows/coverage.yml
```diff
@@ -1,24 +1,28 @@
...
+name: Coverage
 
 on:
-  #On push to common branches, this computes the "bases stats" for PRs
+  # On push to common branches, this computes the coverage PRs will compared against
   push:
     branches:
       - master
   pull_request:
+    branches:
```
diegomrsantos, line 9: I don't believe this is necessary.

.github/workflows/base_tests.yml
```diff
@@ -0,0 +1,130 @@
+name: Base Tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_call:
```
diegomrsantos, line 8: Could you please elaborate more on why this is necessary.

.github/workflows/base_tests.yml
```diff
@@ -0,0 +1,130 @@
...
+          - os: windows
+            builder: windows-2022
+            cpu: amd64
+            shell: msys2 {0}
+        nim_branch:
+          - version: version-1-6
+            memory_management: refc
+          - version: version-2-0
+            memory_management: refc
+          # - version: devel
```
diegomrsantos, line 60: Could you please explain why this comment is necessary?

.github/workflows/dependencies.yml
diegomrsantos: We can use the PR to ask questions.

.github/workflows/base_tests.yml
```diff
@@ -0,0 +1,130 @@
...
+        required: false
+        type: string
+        default: "[]"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  delete_cache:
```
diegomrsantos, line 26: Could you please elaborate more on why this is necessary.

.github/workflows/formatter.yml
```diff
@@ -0,0 +1,34 @@
...
+  formatter:
+    name: "Format Codebase"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2  # In PR, has extra merge commit: ^1 = PR, ^2 = base
+
+      - name: Format with NPH
```
diegomrsantos, line 23: This isn't to format the code, it should already be formatted. This is to make sure the code has been formatted.

.github/workflows/formatter.yml
```diff
@@ -0,0 +1,34 @@
+name: Code Formatter
+
+on:
+  push:
```
diegomrsantos, line 4: I guess this one can run only for PRs.

.github/workflows/dependencies.yml
diegomrsantos: Answering your question, yes, it is necessary for Nimbus. But I think it's not necessary for other projects. You might remove Codex and Waku. Could you guys advise? @dryajov @jm-clius

.github/workflows/base_tests.yml
diegomrsantos: Ok, it seems it was necessary cause now the daily reuses the base tests. There was indeed duplication in those files. But one difference is that daily didn't use i`nstall_pinned`. This should continue as before.

.github/workflows/dependencies.yml
jm-clius: I think this is used for the regular autobump CI run: https://github.com/waku-org/nwaku/pull/1153? A draft PR is created against latest nim-libp2p and based on CI run we can detect compatibility issues early. Although currently broken and not used/checked by Waku very often, I do think it's a helpful gauge to have available.

.github/workflows/daily.yml
```diff
@@ -1,12 +1,25 @@
-name: Daily
+name: Daily Tests
 on:
   schedule:
     - cron: "30 6 * * *"
   workflow_dispatch:
```
diegomrsantos, line 5: please add a temporary "pull_request:" here to see if this workflow works.

.github/workflows/dependencies.yml
AlejandroCabeza: I did not. This comment serves as a reminder that no one is completely certain how this exactly functions. If somebody does know how it does, they should document it as this is a quite obscure way of handling dependency updates.

.github/workflows/dependencies.yml
AlejandroCabeza: > There might be a misunderstanding regarding comments. They aren't for asking questions about the code.

@diegomrsantos 
I disagree, as elaborated in the other comment on this file. This is a cue to remind myself (and others) that there's a piece of code we don't understand. Certainly, there are multiple ways to handle this, such as stating `# Please, document the purposes if this` or asking in discord. 

Having said this, I already knew we (the libp2p team) didn't know exactly how it worked; and I also knew it was broken-ish/unused-ish as I had asked nwaku already.

In any case, thank you for pinging them as I wanted to double check and had forgotten to do it.

> I do think it's a helpful gauge to have available

@jm-clius 
What part is broken?

.github/workflows/coverage.yml
AlejandroCabeza: As I answered to a comment you made while this PR was in draft, the `Upload coverage to codecov` request has a monthly limit as we're not using a paid tier. Thus, this two lines attempt to reduce the number of calls to `codecov` in order not to exceed that limit. 

I do agree, though, that maybe I'm providing a solution a (yet) non-existing problem, as we probably still haven't reached that limit. Also, the only affected step is `Upload coverage to codecov`, so running that job conditionally might be enough, as we probably don't need all coverages uploaded to codecov, but the `master`-targeted PR ones.

What do you think about the second approach?

.github/workflows/base_tests.yml
AlejandroCabeza: The reason I removed the `install_pinned` was because we were using both dependency install methods in the CI, and I tried to reduce them to one, as having multiple provides confusion and it might even lead to errors due to version differences.

Because of the reasons above, I still think it's a good idea to stick to just one method, but I'm certainly fine with either. Can you remind me why should it be the `install_pinned` one? Is that the default method for installing dependencies?

.github/workflows/dependencies.yml
diegomrsantos: I don't see what value this comment adds.

.github/workflows/base_tests.yml
AlejandroCabeza: The `devel` version is not used in the default matrix (`base_tests.yml`), but it's used in one of the daily runs. To accomplish that we use the `include` parameter, and because that's used in a different file I thought it'd be useful keeping it as comment in the original matrix. That way it serves as documentation of what's in use and a guideline for "nim environment" configurations for other workflows.

Would you prefer if I added a clarifying comment mentioning its purpose?

.github/workflows/dependencies.yml
diegomrsantos: As I mentioned in another comment, I don't see what value such comment adds.

.github/workflows/base_tests.yml
AlejandroCabeza: That's something that exists in the current `daily_common.yml` workflow.

What it does is `Deletes GitHub Actions caches for a branch.`. I think its utility is better described with the example the repository provides: `Caches created in the HEAD branch of a pull request are no longer used, so delete them immediately to save repository cache space.`.

I think it still provides some utility and, although it isn't essential, it might be a good idea to keep it on a different workflow. That way it could be run only on a `PR Closed` event, as they show in their example.

What do you think?

REF: https://github.com/snnaplab/delete-branch-cache-action

.github/workflows/formatter.yml
AlejandroCabeza: I think you might have misread the repo's documentation. In order to do that you need to provide the `--check <file>` flag, otherwise it does indeed format the files.

Besides this, are you aware if we are actually enforcing `nph` formatting before committing? Either "manually" or via a pre-commit hook?

.github/workflows/formatter.yml
AlejandroCabeza: Fair point.

Out of curiosity, though, do you know if somebody can push into `master` directly?

.github/workflows/daily.yml
AlejandroCabeza: Given the `workflow_dispatch` trigger is enabled, we should be able to manually trigger the workflow without having to include the `pull_request` one. 

I suggest setting everything in place, and trying these out afterwards.

.github/workflows/base_tests.yml
diegomrsantos: I'd prefer if you could please remove all commented code.

.github/workflows/base_tests.yml
diegomrsantos: The CI for the PRs use fixed versions in the .pinned files for stability purposes. The daily job uses the nimble resolution. Please, keep it like before.

.github/workflows/formatter.yml
diegomrsantos: This is a job that is gonna fail if the code hasn't been formatted and therefore the PR can't be merged. It won't commit any change it might produce. Maybe it's possible to use check indeed, but it is already working as intended and doesn't require changes.

.github/workflows/formatter.yml
diegomrsantos: No, it's not allowed.

.github/workflows/daily.yml
diegomrsantos: I couldn't find a way to do that. Probably it's available only after the PR is merged.

.github/workflows/base_tests.yml
diegomrsantos: But comments explaing why include and exclude were added and how they are used could be helpful.

.github/workflows/dependencies.yml
diegomrsantos: I don't know about the other projects but for Nimbus, the `Checkout this ref` step update `nim-libp2p` - which is a submodule dependency. This step is committing this update.

.github/workflows/dependencies.yml
diegomrsantos: > This is a cue to remind myself (and others) that there's a piece of code we don't understand
I explained why this is necessary. If you feel something in the codebase is hard to understand and needs documentation, please create an issue. Using comments like this one, todos, etc, is fine while one is working on the task. Once it's ready to be reviewed and merged, it should be production-grade and all of those things should be removed and addressed with issues - that have proper visibility and can be prioritized accordingly.

diegomrsantos: Feel free to add a concise comment about what has been discussed in this PR about this workflow.

AlejandroCabeza: There's a very interesting discussion regarding comments which I'll gladly answer to if you ever feel like talking about it. 

In any case, going back to the point: Yes, I'll update the comment with the newly acquired information, as intended from the start.

.github/workflows/dependencies.yml
AlejandroCabeza: I'm aware of how it functions. I'm simply not aware of its true purpose and/or whether that's the only way to handle it.

I've asked Nimbus and they confirmed they're still using this as a mechanism to test integrations of every libp2p trunk commit with nimbus-eth2.

Knowing this, there's no need for this comment anymore.

.github/workflows/base_tests.yml
AlejandroCabeza: I am aware of the behaviour of those methods, but my question is about why are we using two different method for dependency management.

.github/workflows/base_tests.yml
AlejandroCabeza: I wouldn't mind removing this one, although I would argue it provides context (this combination is used somewhere else as stated above).
But I would definitely avoid adding a comment explaining an already-documented Github's CI's feature.

.github/workflows/formatter.yml
AlejandroCabeza: Ah, I hadn't noticed the `git diff --exit-code` line. That explains it.

.github/workflows/daily.yml
AlejandroCabeza: I've double checked the docs, and apparently that's allowed only when the workflow is in the `default` branch. I'll do so when we agree on the rest of the PR.

.github/workflows/dependencies.yml
arnetheduck: @jm-clius the intent is that the draft PR also gets deployed to some fleet nodes, so that that the in-flight libp2p changes get some real-world exposure - this is an invaluable part of libp2p's testing regime based on nimbus experience: https://metrics.status.im/d/pgeNfj2Wz23/nimbus-fleet-testnets?orgId=1&from=now-30d&to=now&var-instance=linux-01.ih-eu-mda1.nimbus.sepolia&var-container=beacon-node-sepolia-libp2p&refresh=15m

having waku participate is valuable due to how the two networks differ in topology, payload behavior etc - would be good to get it up and running if it's not working presently.

.github/workflows/coverage.yml
diegomrsantos: We don't need to spend time doing something if there's no evidence it's an issue.

.github/workflows/base_tests.yml
```diff
@@ -0,0 +1,132 @@
...
+          nimble install -y --depsOnly
+
+      - name: Run tests
+        run: |
+          nim --version
+          nimble --version
+          NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim_branch.memory_management }}"
+          nimble test
+
+  # lint:
```
diegomrsantos, line 112: Did you forget to remove this code?

AlejandroCabeza: Ah, yes. Thanks for pointing that out.

.github/workflows/dependencies.yml
```diff
@@ -1,27 +1,29 @@
...
     runs-on: ubuntu-latest
+    name: Update libp2p's version for ${{ matrix.target.repo }} from ${{ matrix.target.source_ref }}
     strategy:
       fail-fast: false
       matrix:
         target: [
-          { repo: status-im/nimbus-eth2, branch: unstable },
-          { repo: waku-org/nwaku, branch: master },
-          { repo: codex-storage/nim-codex, branch: master }
+          { repo: status-im/nimbus-eth2, source_ref: master },
```
diegomrsantos, line 17: this should be unstable

.github/workflows/coverage.yml
AlejandroCabeza: Right, as long as you remember to fix that when it happens.

AlejandroCabeza: Right, as long as you remember to fix that when it happens I'm fine with that.

.github/workflows/dependencies.yml
AlejandroCabeza: Isn't `unstable` deprecated? I asked on discord but I got no answer.

AlejandroCabeza:
> What is the status here? Is is the PR ready for re-review?

I've fixed answered threads but I'm waiting for answers on a couple or three others.

.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
+  test-all:
```
diegomrsantos, line 14: What's the current purpose of this file?

.github/workflows/daily.yml
diegomrsantos: We need to be sure this workflow is working properly.

.github/workflows/ci.yml
AlejandroCabeza: This might've been a `rebase` misshap. This file was migrated to `base_tests.yml`.

.github/workflows/dependencies.yml
diegomrsantos: nimbus unstable isn't deprecated afaik

.github/workflows/base_tests.yml
diegomrsantos: The `.pinned` file behaves like a lock file - that wasn't supported before Nimble 0.14 - fixing the dependencies versions.

.github/workflows/dependencies.yml
diegomrsantos: I'm also not sure there's a need to rename "branch"

AlejandroCabeza: Ah! Now I understand, I got confused with the names. Thanks for pointing that out.

.github/workflows/base_tests.yml
AlejandroCabeza: Right. Does that mean it's supported now?

.github/workflows/coverage.yml
```diff
@@ -40,25 +42,35 @@ jobs:
...
+      
+      - name: Upload coverage to codecov
+        run: |
           bash <(curl -s https://codecov.io/bash) -f coverage/coverage.f.info || echo "Codecov did not collect coverage reports"
 
-          #- uses: actions/upload-artifact@master
-          #  with:
-          #    name: coverage
-          #    path: coverage
+        # Alternatively, use the upload artifact action
```
diegomrsantos, line 72: Commented code can be removed.

.github/workflows/base_tests.yml
diegomrsantos: Only for Nim 2, I guess. Not sure how well it works.

diegomrsantos: We can try to use it instead of this .pinned file when we don't need to support Nim 1.6.x anymore.

AlejandroCabeza: Alright.
Would it be alright to use this `.pinned` method for dependencies for the daily tests? Or is there some reason to keep using nimble's dep resolver?

diegomrsantos: The .pinned file is used only by us to have a deterministic version of the dependencies. The libp2p.nimble file uses ranges or no dependency at all. This can be disruptive with newer deps versions being released. The daily job is like that because that's the likely context libp2p will be used as a dep.

diegomrsantos: Feel free to document it with comments.

diegomrsantos: In the daily workflow we test libp2p with newer deps versions without disrupting developers' PRs.

AlejandroCabeza: I'm aware of the `.pinned` usages. 

But, as far as I know, wasn't the `nimble install` unreliable?

.github/workflows/dependencies.yml
```diff
@@ -1,27 +1,29 @@
...
 on:
   push:
     branches:
       - master
   workflow_dispatch:
 
 jobs:
-  bumpProjects:
+  update_version:
+    # Pushes new refs to interested external repositorysitories, so they can do early testing against them.
```
diegomrsantos, line 10:
```suggestion
    # Pushes new refs to interested external repositories, so they can do early testing against them.
```

.github/workflows/dependencies.yml
```diff
@@ -1,27 +1,29 @@
...
     runs-on: ubuntu-latest
+    name: Update libp2p's version for ${{ matrix.target.repository }} from ${{ matrix.target.ref }}
     strategy:
       fail-fast: false
       matrix:
         target: [
-          { repo: status-im/nimbus-eth2, branch: unstable },
-          { repo: waku-org/nwaku, branch: master },
-          { repo: codex-storage/nim-codex, branch: master }
+          { repository: status-im/nimbus-eth2, ref: unstable },
```
diegomrsantos, line 17: why does it need to be renamed from branch to ref?

.github/workflows/dependencies.yml
```diff
@@ -1,27 +1,29 @@
...
         ]
     steps:
-      - name: Clone repo
-        uses: actions/checkout@v2
+      - name: Clone repository
+        uses: actions/checkout@v4
         with:
-          repository: ${{ matrix.target.repo }}
-          ref: ${{ matrix.target.branch }}
+          repositorysitory: ${{ matrix.target.repository }}
```
diegomrsantos, line 25: there's a typo here

.github/workflows/dependencies.yml
```diff
@@ -33,7 +35,7 @@ jobs:
           cd vendor/nim-libp2p
           git checkout $GITHUB_SHA
 
-      - name: Commit this bump
+      - name: Commit this ref
```
diegomrsantos, line 38:
```suggestion
      - name: Commit updated libp2p dependency version
```

.github/workflows/dependencies.yml
AlejandroCabeza: I'd argue it's a better choice since it can be any kind of ref, but most importantly it matches the `ref` parameter in `checkout` action, which is where it's used.

.github/workflows/dependencies.yml
AlejandroCabeza: I changed the name slightly, but didn't add the "updated libp2p" as that's already mentioned a couple times in the names and descriptions of the workflow and jobs.

I think it looks good as is, but if you insist on the naming I don't mind changing it.

diegomrsantos: I don't have a strong opinion, but some of those changes are subjective. This one is less clear to me.

AlejandroCabeza: Can you explain what you mean?

diegomrsantos: It's ok. We can resolve it.

lchenut (approved):
LGTM

PR #1119, opened 2024-06-06
chore(version): update libp2p.nimble to 1.3.0

kaiserd:
Update libp2p.nimble to version to 1.3.0.
This commit will be tagged as 1.3.0.

This super-seeds and closes
https://github.com/vacp2p/nim-libp2p/pull/1114
because the commits since 1.2.0 contain updates to gossipsub that go beyond a fix / chore and justify a new minor version instead of a mere patch version.

> *Note*: There is a jump from 1.1.0 to 1.3.0 because I forgot to update `libp2p.nimble` for the 1.2.0 release.
The `1.3.0` release also acts as a patch solving this issue, which we unfortunately did not detect earlier.

### DIFF
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -349,13 +349,11 @@ proc start*(s: Switch) {.async, public.} =
       s.acceptFuts.add(s.accept(t))
       s.peerInfo.listenAddrs &= t.addrs
 
-  await s.peerInfo.update()
-
-  await s.ms.start()
-
   for service in s.services:
     discard await service.setup(s)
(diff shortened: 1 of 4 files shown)


diegomrsantos (approved):

AlbertoSoutullo (approved):
From DST we were able to test this version with both Yamux and Mplex in a mesh with 1000 nodes.
Note that there is still a lot of room for testing improvement.
Metric looks good so far and no major issues were found.

lchenut (approved):

PR #1120, opened 2024-06-06
fix(services): setup services before peerinfo is updated

diegomrsantos:
- Setup services before peerinfo is updated - this allows services to set address mappers.
- Fix HP interop tests - improve condition related to relay addr being published.
- Disable wildcard resolver in `testautonatservice` and test `testautorelay`.


diegomrsantos:
> Thank you. Why do disable testautonatservice and test testautorelay fail with the wildcard resolver being active?

Cause now `peerInfo.addrs` is populated with the resolved addresses right after the `switch` is started and the tests are probably not expecting that.

PR #156, opened 2020-04-30
Add PubSub observer+ hooks (they can modify as well)

sinkingsugar:
I think this PR covers both
https://github.com/status-im/nim-libp2p/issues/153
and
https://github.com/status-im/nim-libp2p/issues/155

PR #1128, opened 2024-06-17
feat: adding onValidated observer

gabrielmer:
### Description

Adding an `onValidated` observer which will run every time a message is received and validated. This comes from the necessity of precisely track message deliveries and network activity.

`onRecv` observers run before any check is performed on the received message, which means that it runs every time a duplicate or invalid message arrives, which is inefficient and inaccurate for our purpose of tracking only received, unique and valid messages. Therefore, adding this extra option of running an observer for every message after all validation checks pass.

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -454,7 +454,7 @@ proc validateAndRelay(
     g.rewardDelivered(peer, topic, true)
 
     # trigger hooks
-    peer.recvAndValidatedObservers(msg, msgId)
+    peer.validatedObservers(msg, msgId)
 
     # The send list typically matches the idontwant list from above, but
     # might differ if validation takes time
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -453,6 +453,9 @@ proc validateAndRelay(
 
     g.rewardDelivered(peer, topic, true)
 
+    # trigger hooks
+    peer.recvAndValidatedObservers(msg, msgId)
+
     # The send list typically matches the idontwant list from above, but
     # might differ if validation takes time
     var toSendPeers = HashSet[PubSubPeer]()
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -175,13 +175,12 @@ proc recvObservers*(p: PubSubPeer, msg: var RPCMsg) =
         if not (isNil(obs.onRecv)):
           obs.onRecv(p, msg)
 
-proc recvAndValidatedObservers*(p: PubSubPeer, msg: Message, msgId: MessageId) =
+proc validatedObservers*(p: PubSubPeer, msg: Message, msgId: MessageId) =
   # trigger hooks
   if not (isNil(p.observers)) and p.observers[].len > 0:
     for obs in p.observers[]:
-      if not (isNil(obs)): # TODO: should never be nil, but...
-        if not (isNil(obs.onValidated)):
-          obs.onValidated(p, msg, msgId)
+      if not (isNil(obs.onValidated)):
+        obs.onValidated(p, msg, msgId)
 
 proc sendObservers(p: PubSubPeer, msg: var RPCMsg) =
   # trigger hooks
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -67,6 +67,8 @@ type
   PubSubObserver* = ref object
     onRecv*: proc(peer: PubSubPeer, msgs: var RPCMsg) {.gcsafe, raises: [].}
     onSend*: proc(peer: PubSubPeer, msgs: var RPCMsg) {.gcsafe, raises: [].}
+    onRecvAndValidated*:
+      proc(peer: PubSubPeer, msg: Message, msgId: MessageId) {.gcsafe, raises: [].}
 
   PubSubPeerEventKind* {.pure.} = enum
     StreamOpened
@@ -171,6 +173,13 @@ proc recvObservers*(p: PubSubPeer, msg: var RPCMsg) =
     for obs in p.observers[]:
       if not (isNil(obs)): # TODO: should never be nil, but...
         obs.onRecv(p, msg)
+
+proc recvAndValidatedObservers*(p: PubSubPeer, msg: Message, msgId: MessageId) =
+  # trigger hooks
+  if not (isNil(p.observers)) and p.observers[].len > 0:
+    for obs in p.observers[]:
+      if not (isNil(obs)): # TODO: should never be nil, but...
+        obs.onRecvAndValidated(p, msg, msgId)
 
 proc sendObservers(p: PubSubPeer, msg: var RPCMsg) =
   # trigger hooks
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -172,21 +172,24 @@ proc recvObservers*(p: PubSubPeer, msg: var RPCMsg) =
   if not (isNil(p.observers)) and p.observers[].len > 0:
     for obs in p.observers[]:
       if not (isNil(obs)): # TODO: should never be nil, but...
-        obs.onRecv(p, msg)
+        if not (isNil(obs.onRecv)):
+          obs.onRecv(p, msg)
 
 proc recvAndValidatedObservers*(p: PubSubPeer, msg: Message, msgId: MessageId) =
   # trigger hooks
   if not (isNil(p.observers)) and p.observers[].len > 0:
     for obs in p.observers[]:
       if not (isNil(obs)): # TODO: should never be nil, but...
-        obs.onRecvAndValidated(p, msg, msgId)
+        if not (isNil(obs.onRecvAndValidated)):
+          obs.onRecvAndValidated(p, msg, msgId)
 
 proc sendObservers(p: PubSubPeer, msg: var RPCMsg) =
   # trigger hooks
   if not (isNil(p.observers)) and p.observers[].len > 0:
     for obs in p.observers[]:
       if not (isNil(obs)): # TODO: should never be nil, but...
-        obs.onSend(p, msg)
+        if not (isNil(obs.onSend)):
+          obs.onSend(p, msg)
 
 proc handle*(p: PubSubPeer, conn: Connection) {.async.} =
   debug "starting pubsub read loop", conn, peer = p, closed = conn.closed
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -67,7 +67,7 @@ type
   PubSubObserver* = ref object
     onRecv*: proc(peer: PubSubPeer, msgs: var RPCMsg) {.gcsafe, raises: [].}
     onSend*: proc(peer: PubSubPeer, msgs: var RPCMsg) {.gcsafe, raises: [].}
-    onRecvAndValidated*:
+    onValidated*:
       proc(peer: PubSubPeer, msg: Message, msgId: MessageId) {.gcsafe, raises: [].}
 
   PubSubPeerEventKind* {.pure.} = enum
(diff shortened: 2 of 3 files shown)


diegomrsantos:
Are you using the latest master? It seems there are some formatting changes that should be addressed in the latest master.

diegomrsantos:
I see. It's cause a new param was added exactly when the param list stopped being in one line and it was difficult to see the change https://github.com/vacp2p/nim-libp2p/pull/1128/files#diff-abbf987f19a24c9940cd01bde79a1b1e0819f1c3b6d7ba2efa04ad680d022c78R385

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -776,6 +818,9 @@ method publish*(g: GossipSub, topic: string, data: seq[byte]): Future[int] {.asy
 
   g.mcache.put(msgId, msg)
 
+  trace "publish message to peers",
```
diegomrsantos, line 821: You said "we can't have them in TRACE as running at that level would pollute too much our simulations' logs", but here it is `trace`. Could you please elaborate?

gabrielmer:
> 1. Would it be possible to express what you need in a app layer validator call back that gets handed to`validateAndRelay` ?
>    That seems to be a generally useful approach.
>    It could be called after the standard nim-libp2p validation, and add app specific validation.
>    This might allow moving the additional logging to Waku / applications.

Thanks so much! I think I can do exactly the same I'm doing in this PR but adding a generic construct to add app specific logic.

I think it can be hard to accurately understand what we exactly need without actually coding it and finding constraints we might hadn't think about.

If it sounds good to you, I can implement the same thing by doing this generic extension, get feedback, iterate and go from there.

lmk what you think :)) @kaiserd @Ivansete-status

diegomrsantos:
I believe you can experiment here until you find what you need, but eventually it's better to find a way to add the changes to master for two reasons:
1) it can benefit other projects if they need the same.
2) it's going to be bothersome to keep this branch up-to-date with master.

gabrielmer:
> I believe you can experiment here until you find what you need, but eventually it's better to find a way to add the changes to master for two reasons:
> 
> 1. it can benefit other projects if they need the same.
> 2. it's going to be bothersome to keep this branch up-to-date with master.

I agree 100%, the idea is to now change this PR with a more generic approach with the target to actually get merged to master

gabrielmer:
It does seem that we can actually get the information that we want without changing anything in `nim-libp2p` by using `onSend` observers instead of `onRecv` which was our first approach.

We will merge https://github.com/waku-org/nwaku/pull/2851 for it.

If after using it for simulations we conclude that this approach logs all the necessary information, then I will close this PR. Otherwise, will implement over here a solution for it that could be merged to `nim-libp2p`

diegomrsantos:
Thanks for the update.

gabrielmer:
We started having the original issue and were in need of info only available at the `nim-libp2p` layer, so pushed again changes to this PR

Currently having it over an old `nim-libp2p` commit so it matches the one used in our latest release, but soon will rebase it.

I think this new approach is simpler and pretty neat, please let me know if a solution like this is something that could be eventually merged into master :)

gabrielmer:
@diegomrsantos @kaiserd WDYT about this new approach? is it something that could be merged?

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -67,6 +67,8 @@ type
   PubSubObserver* = ref object
     onRecv*: proc(peer: PubSubPeer, msgs: var RPCMsg) {.gcsafe, raises: [].}
     onSend*: proc(peer: PubSubPeer, msgs: var RPCMsg) {.gcsafe, raises: [].}
+    onRecvAndValidated*:
```
diegomrsantos, line 70: How about `onValidated`?

tests/pubsub/testgossipsub.nim
```diff
@@ -360,12 +360,17 @@ suite "GossipSub":
           inc observed
       )
       obs2 = PubSubObserver(
+        onRecvAndValidated: proc(peer: PubSubPeer, msg: Message, msgId: MessageId) =
```
diegomrsantos, line 363: Could this be added to the existing observer?

tests/pubsub/testgossipsub.nim
```diff
@@ -382,7 +387,7 @@ suite "GossipSub":
     await allFuturesThrowing(nodes[0].switch.stop(), nodes[1].switch.stop())
 
     await allFuturesThrowing(nodesFut.concat())
-    check observed == 2
+    check observed == 3
```
diegomrsantos, line 390: It's probably better to create a new test for this.

libp2p/protocols/pubsub/pubsubpeer.nim
gabrielmer: Perfect! Renamed it. Had the `onRecv` part to be more explicit that it was only when messages are received (not sure if we run validations before sending messages). But if `onValidated` is clear enough then even better :)

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -170,14 +172,24 @@ proc recvObservers*(p: PubSubPeer, msg: var RPCMsg) =
...
       if not (isNil(obs)): # TODO: should never be nil, but...
-        obs.onRecv(p, msg)
+        if not (isNil(obs.onRecv)):
+          obs.onRecv(p, msg)
+
+proc recvAndValidatedObservers*(p: PubSubPeer, msg: Message, msgId: MessageId) =
+  # trigger hooks
+  if not (isNil(p.observers)) and p.observers[].len > 0:
+    for obs in p.observers[]:
+      if not (isNil(obs)): # TODO: should never be nil, but...
```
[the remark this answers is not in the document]
gabrielmer: Which part? I added `if not (isNil(obs))` because it is run for all the other observers and if I understand correctly this case should be the same for all observers.

The `if not (isNil(obs.onValidated))` I added it here and for the other observers,  because if we register an observer and define only one of them (for example setting a proc for `onRecv` but leaving `onSend` empty), it would crash when trying to read from `nil` when attempting to run the undefined observer.

So added this to avoid this crash, and not have to define an empty proc for an observer that is not used.

Not sure if it makes sense

tests/pubsub/testgossipsub.nim
gabrielmer: Yes, definitely! But I guess it depends if we keep testing it here (and I add it to the existing observer) or if it will go on a separate test. LMK

libp2p/protocols/pubsub/pubsubpeer.nim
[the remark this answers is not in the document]
diegomrsantos: I was talking about `if not (isNil(obs))`, there shouldn't be any `nil` inside this seq. Also, in the future please avoid using TODOs and create issues instead.

I'm also not sure about `observers*: ref seq[PubSubObserver] # ref as in smart_ptr`, it's the first time I see a ref being used like that in the codebase. @dryajov could you please advise here, Is it needed? I found this, but I'm not sure about it https://github.com/vacp2p/nim-libp2p/pull/156#discussion_r420042988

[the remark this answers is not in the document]
diegomrsantos: `if not (isNil(p.observers))` shouldn't be necessary either.

tests/pubsub/testgossipsub.nim
diegomrsantos: This test seems to be related to sending messages over fanout, however, it's not clear to me what it's testing. A proper test for this validator would check it is being called how it's expected and not being called otherwise.

libp2p/protocols/pubsub/pubsubpeer.nim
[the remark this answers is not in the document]
gabrielmer: > I was talking about `if not (isNil(obs))`, there shouldn't be any `nil` inside this seq. Also, in the future please avoid using TODOs and create issues instead.

I just copy pasted how the observers are currently called (including the TODO comment) and added the `if not (isNil(obs.onValidated))` after I had crashes running tests when leaving one empty observer
https://github.com/vacp2p/nim-libp2p/blob/f8d4da6421b767727dd1501cb31b27636af87755/libp2p/protocols/pubsub/pubsubpeer.nim#L168-L180

Should I delete that check and the TODO from all 3 of them?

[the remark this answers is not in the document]
diegomrsantos: You can leave the others for now, but remove them from yours. Thanks.

tests/pubsub/testgossipsub.nim
diegomrsantos: Yes, there are improvements to be made. Please add a new one if that's ok.

gabrielmer: Done :)

libp2p/protocols/pubsub/pubsubpeer.nim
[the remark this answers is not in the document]
gabrielmer: Done, I did have to leave the
```
if not (isNil(obs.onValidated))
```

because otherwise I did get a
```
```
on a test in which other observers are registered but that one was left empty

tests/pubsub/testgossipsub.nim
```diff
@@ -220,6 +220,70 @@ suite "GossipSub":
 
     await allFuturesThrowing(nodesFut.concat())
 
+  asyncTest "GossipSub's observers should run after message is sent, received and validated":
```
diegomrsantos, line 223: Looks much better, many thanks!

tests/pubsub/testgossipsub.nim
```diff
@@ -220,6 +220,70 @@ suite "GossipSub":
...
+    proc onValidated(peer: PubSubPeer, msg: Message, msgId: MessageId) =
+      inc validatedCounter
+
+    let obs0 = PubSubObserver(onSend: onSend)
+    let obs1 = PubSubObserver(onRecv: onRecv, onValidated: onValidated)
+
+    let
+      nodes = generateNodes(2, gossip = true)
+      # start switches
+      nodesFut = await allFinished(nodes[0].switch.start(), nodes[1].switch.start())
```
diegomrsantos, line 249: is nodesFut necessary?

tests/pubsub/testgossipsub.nim
```diff
@@ -220,6 +220,70 @@ suite "GossipSub":
...
+  asyncTest "GossipSub's observers should run after message is sent, received and validated":
+    var
+      handlerFut = newFuture[bool]()
+      recvCounter = 0
+      sendCounter = 0
+      validatedCounter = 0
+
+    proc handler(topic: string, data: seq[byte]) {.async.} =
+      check topic == "foo"
+      handlerFut.complete(true)
```
diegomrsantos, line 232: Is this necessary?

libp2p/protocols/pubsub/pubsubpeer.nim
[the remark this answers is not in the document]
diegomrsantos: `if not (isNil(obs.onValidated))` is fine, although it could be an Option. But it's not necessary to change it now. 

I guess the `ref` purpose in `ref seq[PubSubObserver]` is to set `observers` in `PubSub` and have a reference to it in the `PubSubPeer` instances.

tests/pubsub/testgossipsub.nim
diegomrsantos: Thanks a lot.

tests/pubsub/testgossipsub.nim
gabrielmer: It is for the last line, for `await allFuturesThrowing(nodesFut.concat())`, to raise an exception in case one switch failed starting.

Basically all tests in that file do it this way, so just followed the existing pattern. I would personally do the check right after attempting to start the switches and not at the end of the test, but wanted to be consistent with all the other tests. And maybe I was missing something.

LMK how you prefer

tests/pubsub/testgossipsub.nim
gabrielmer: Not really, it was just an extra sanity check. But I can get rid of it if you want

gabrielmer: done

tests/pubsub/testgossipsub.nim
diegomrsantos: I don't believe this is necessary. Don't always take the existing code as a good solution.

diegomrsantos:
Please update the PR's title and description to reflect its current state.

tests/pubsub/testgossipsub.nim
gabrielmer: Yes for sure, but for someone who is not very familiar with the codebase, sometimes the safest and most natural approach is to follow the patterns done by the people who wrote it :)

gabrielmer: Removed `nodesFut`, please take a look now

gabrielmer:
> Please update the PR's title and description to reflect its current state.

Yup, ready 🫡

tests/pubsub/testgossipsub.nim
diegomrsantos: Agreed, it was just an observation ;)

diegomrsantos (approved):
Amazing job, thanks a lot!

gabrielmer:
> Amazing job, thanks a lot!

Thanks to you for the time and all the help! 🫶

diegomrsantos:
It seems the test failed here https://github.com/vacp2p/nim-libp2p/actions/runs/10679844554/job/29599995590

PR #1140, opened 2024-06-26
chore(yamux): remove asyncSpawn

### LINKED ISSUE #1134: Yamux: Remove trySend AsyncSpawn
AsyncSpawn, in this specific case is clearly overkill. AsyncSpawn is not recommended as it is uncontrollable, and, here, it could easily be replaced by another method which is safer such as an AsyncQueue or just storing the Future in the Yamux object.

lchenut:
closes #1134

### DIFF
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -150,7 +150,10 @@ type
     conn: Connection
     isSrc: bool
     opened: bool
-    isSending: bool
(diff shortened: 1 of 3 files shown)


lchenut:
Closing this PR for two reasons: 
- it's not urgent at all
- it raises some interopability issues that are probably related to https://github.com/libp2p/js-libp2p/issues/2505... so this needs to be clear first.

PR #1143, opened 2024-07-02
chore(version): update libp2p.nimble to 1.4.0

kaiserd:
Update libp2p.nimble to version to 1.4.0.
This commit will be tagged as 1.4.0.


AlejandroCabeza (approved):

AlbertoSoutullo (approved):
DST node compiled and is working without problems with the commit hash.

PR #1145, opened 2024-07-04
fix(multiaddress): Raise `MaError` instead of `LPError` on `&`

etan-status:
`LPError` is the top level error type of libp2p, it makes more sense to raise a multi address specific subtype to avoid requiring callers to catch errors too broadly. As `MaError` inherits from `LPError`, existing error handlers will still work fine.


etan-status:
Nim 2.0 is working fine on the latest release (2.0.8) - but current `version-2-0` branch seems to fail (4856beae70ba058612a07ab787c9f283f885b453)

diegomrsantos:
> Nim 2.0 is working fine on the latest release (2.0.8) - but current `version-2-0` branch seems to fail (4856beae70ba058612a07ab787c9f283f885b453); and also does so on other PRs.

https://discord.com/channels/613988663034118151/740631022080622745/1279118505819508857

diegomrsantos (approved):

PR #1155, opened 2024-07-28
remove valueOr template from utility module

Ivansete-status:
The `valueOr` template is already defined in
https://github.com/arnetheduck/nim-results, which is the one that we are using in `nwaku`.

Therefore, we are willing to prevent conflicts between the two 'valueOr' implementations.

### DIFF
--- a/libp2p/utility.nim
+++ b/libp2p/utility.nim
@@ -10,8 +10,9 @@
 {.push raises: [].}
 
 import std/[sets, options, macros]
-import stew/[byteutils, results]
+import stew/byteutils
 
+import results
 export results
 
 template public*() {.pragma.}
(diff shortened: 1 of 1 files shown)


Ivansete-status:
I see that unfortunately, everything fails :S 

I guess we'd need to add the dependency that I mentioned in the description.

Ivansete-status:
I'm closing after having a conversation with @lchenut given that is a nice to have and not urgent.
I paste Ludovic's comments for future reference:
> The problem with this PR is that we use Option (which is a nim native object) and not a Result (or Opt) which are from Jacek's nim-results
And to be honest, I'm not 100% sure why we use three different objects to do the same thing.
But this is the reason why we need  template valueOr*[T](self: Option[T], body: untyped): untyped. Because it's not on nim-results. And it's not native either.
One thing to investigate instead might be to remove the native Option. But there might be some other issues here.

arnetheduck:
Most other code has been ported to `Opt` - there is a small semantic difference where `Option[ref Xxx]` doesn't allow `nil` and `Opt` does - other than that, it's simply easier to combine `Opt` with `Result`.

PR #1159, opened 2024-07-31
fix(test): interop transport

AlejandroCabeza:
Free disk space before running the steps of interop's transport test.
The original job has enough space but in our repo it crashes  midway due to missing disk space.

### DIFF
--- a/.github/workflows/interop.yml
+++ b/.github/workflows/interop.yml
@@ -15,6 +15,12 @@ jobs:
     name: Run transport interoperability tests
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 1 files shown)


diegomrsantos (changes requested):
As we discussed on discord, this is not the right solution.

The job now logs much more things and takes longer to run. It might be better to find a way to reduce the logs like it was before. it's taking 41 min to run now. Previously, it took 10 min.

diegomrsantos (approved):
We agreed to merge it and work on fixing the caching issue that is making the docker images to be built on every run.

PR #1168, opened 2024-08-07
test(gossipsub): peer discovery

AlejandroCabeza:


tests/testdiscovery.nim
```diff
@@ -18,7 +18,7 @@ import
     discovery/discoverymngr,
     discovery/rendezvousinterface,
   ]
-import ./helpers
+import ./helpers, ./asyncunit, ./utils/[async, assertions, futures]
```
diegomrsantos, line 21: Those tests need to be moved to a new PR with a different scope as they are unrelated to GossipSub.

AlejandroCabeza: I don't think I follow. Why are you linking the comment to the imports?

diegomrsantos: Sorry, I mean the tests in this file.

tests/utils/tests.nim
```diff
@@ -0,0 +1,29 @@
...
+
+template testx*(name: string, body: untyped) =
+  test name:
+    skip()
+
+template xasyncTest*(name: string, body: untyped) =
+  test name:
+    skip()
+
+template asyncTestx*(name: string, body: untyped) =
```
diegomrsantos, line 27: This isn't necessary, if a test should not run, just remove it.

tests/utils/tests.nim
```diff
@@ -0,0 +1,29 @@
+import testutils/unittests
+
+template xsuite*(name: string, body: untyped) =
```
diegomrsantos, line 3: Same for everything in this file.

tests/utils/async.nim
```diff
@@ -0,0 +1,30 @@
...
+  else:
+    return results.err("Future still not finished.")
+
+proc waitForResult*[T](
+    future: Future[T], timeout = DURATION_TIMEOUT
+): Future[Result[T, string]] {.async.} =
+  discard await future.withTimeout(timeout)
+  return future.toResult()
+
+proc reset*[T](future: Future[T]): void =
```
diegomrsantos, line 27: We shouldn't access internal fields.

tests/testdiscovery.nim
AlejandroCabeza: I would say they are related to Gossipsub given they are specified in Gossipsub Testplan's Peer Discovery block, as linked in the description.

Do you perhaps mean to move them to a different file?

tests/utils/tests.nim
AlejandroCabeza: The usage of `x` to specify skipped test is not my invention. This module has been sourced from Waku, and it's also used in some other testing libraries in other languages. An example of this would be pytest, IIRC.

There's a lot of reasons why you would want tests to not be run. Specifically, I used it to skip a test which is required, but can't be enabled due to not being able to fully assert the case.

tests/utils/async.nim
AlejandroCabeza: This function is to be used only during testing, thus being inside `tests/utils/async.nim`. It's a not-perfect-but-good-enough shortcut to reuse futures in handlers, for instance.

tests/testdiscovery.nim
diegomrsantos: Peer discovery isn't related to GossipSub.

tests/utils/tests.nim
diegomrsantos: That's not a pattern in the libp2p codebase, please remove it.

diegomrsantos: Also, we shouldn't have commented or any code that's unnecessary/not being used.

tests/testdiscovery.nim
diegomrsantos: I recommend moving those tests to Rendezvous test plans and moving them to a separate PR for clarity.

AlejandroCabeza: I'd argue they are related, since Peer Discovery is a dependecy of GossipSub, as clearly stated in the [official documentation](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md).
I don't consider I have any say over the test plans so, if you feel like it's inapropriate, I'd suggest raising a question in our related channel.
Having said this, I don't see the benefit of splitting this small amount of code into a different PR.

tests/utils/tests.nim
AlejandroCabeza: Well, ideally we would enable a way for that to be tested as that would provide better coverage. If you don't want to do that then I'll happily remove it.

diegomrsantos: I would suggest that you create an issue and it can get prioritized. The test can be posted there.

tests/utils/async.nim
AlejandroCabeza: As I previously stated, it's just a test-only workaround. Regardless, I'm happy to delete it and add it back when I need it.

diegomrsantos: We shouldn't be introducing workarounds in the codebase unless they are strictly necessary. Is there a `Future` feature that is missing and this workaround is fixing?

AlejandroCabeza: Just a handier way of resetting a future.  During testing they need to be reset one way or another, even if it doesn't see, that way at first.

But don't worry, I'll happily remove until I make clear use of it.

tests/testdiscovery.nim
diegomrsantos: > Whenever a new peer is connected, the gossipsub implementation checks to see if the peer implements floodsub and/or gossipsub, and if so, it sends it a hello packet that announces the topics that it is currently subscribing to.

I don't know if we are doing this part _checks to see if the peer implements floodsub and/or gossipsub_.

tests/pubsub/testgossipsub.nim
```diff
@@ -277,54 +278,6 @@ suite "GossipSub":
 
     await allFuturesThrowing(nodes[0].switch.stop(), nodes[1].switch.stop())
 
-  asyncTest "GossipSub unsub - resub faster than backoff":
```
diegomrsantos, line 280: Why has this been removed?

diegomrsantos: Why is the test wrong?

diegomrsantos (changes requested):
Rendezvous tests are unrelated to GossipSub, they shound't be in GossipSub test cases. Please create a new PR with a different description.

tests/pubsub/testgossipsub.nim
```diff
@@ -839,18 +792,40 @@ suite "GossipSub":
...
     gossip0.routingRecordsHandler.add(
       proc(peer: PeerId, tag: string, peers: seq[RoutingRecordsPair]) =
         check:
           tag == "foobar"
           peers.len == 2
           peers[0].record.isSome() xor peers[1].record.isSome()
-        passed.complete()
+        passed0.complete()
     )
+    gossip1.routingRecordsHandler.add(
```
diegomrsantos, line 808: First time I see `RoutingRecordsHandler`. Would you happen to know what is the purpose of this? Is it something mentioned in the spec?

tests/utils/async.nim
```diff
@@ -0,0 +1,25 @@
...
+    else:
+      return results.err("Future finished but failed.")
+  else:
+    return results.err("Future still not finished.")
+
+proc waitForResult*[T](
+    future: Future[T], timeout = DURATION_TIMEOUT
+): Future[Result[T, string]] {.async.} =
+  discard await future.withTimeout(timeout)
+  return future.toResult()
```
diegomrsantos, line 25: This complexity isn't necessary. `Result` is a substitute for aa `Exception`, not a `Future`. Our focus should be to make the codebase simpler and easier to understand, not the other way around.

tests/pubsub/testgossipsub.nim
```diff
@@ -839,18 +792,40 @@ suite "GossipSub":
...
+          tag == "foobar"
+          peers.len == 2
+          peers[0].record.isSome() xor peers[1].record.isSome()
+        passed2.complete()
+    )
+
+    # Unsubscribe from the topic
     nodes[1].unsubscribe("foobar", handler)
 
-    await passed.wait(5.seconds)
```
diegomrsantos, line 853: why has this been removed?

tests/utils/futures.nim
```diff
@@ -0,0 +1,5 @@
+import chronos
+
+const
+  DURATION_TIMEOUT* = 1.seconds
+  DURATION_TIMEOUT_EXTENDED* = 1500.milliseconds
```
diegomrsantos, line 5: This shouldn't be defined in an utils file. The constant names don't have any meaningful description and we don't have any requirement or consensus about the use of those values in the test suite. If you want to avoid magic numbers and repetition in your tests, it's better to define them for a particular test file.

tests/pubsub/utils.nim
```diff
@@ -201,3 +207,18 @@ proc waitSubGraph*(nodes: seq[PubSub], key: string) {.async.} =
 
     await sleepAsync(5.milliseconds)
     doAssert Moment.now() < timeout, "waitSubGraph timeout!"
+
+proc waitForMesh*(
```
diegomrsantos, line 211: This doesn't seem to be used.

tests/testdiscovery.nim
```diff
@@ -32,32 +32,119 @@ proc createSwitch(rdv: RendezVous = RendezVous.new()): Switch =
...
+      query1 = dmA.request(rdvNamespace)
+      res1 = await query1.getPeer().waitForResult(1.seconds)
+
+    res1.assertIsOk()
+    check res1.value{PeerId}.get() == clientB.peerInfo.peerId
+
+    await rdvB.unsubscribe(namespace)
+    var
+      query2 = dmA.request(rdvNamespace)
+      res2 = await query2.getPeer().waitForResult(1.seconds)
```
diegomrsantos, line 94: You can use `withTimeout`and check the result is `false`.

tests/pubsub/testgossipsub.nim
AlejandroCabeza: Because it doesn't take into account the fanout mechanism. The message is sent via fanout, it doesn't care about backoff period.

tests/pubsub/testgossipsub.nim
AlejandroCabeza: It's the callback for peer exchange.

tests/pubsub/testgossipsub.nim
AlejandroCabeza: Because it's been replaced by the `waitForResult` call. Conceptually similar, but handier for the checks below.

tests/pubsub/utils.nim
AlejandroCabeza: It was used in a test that was removed by your suggestion. Removing it.

tests/testdiscovery.nim
AlejandroCabeza: Those are two different things, I believe.
What you mention is the callback that lets you run code when a peer joins or leaves. 
This is a discovery mechanism that allows you to find peers available for connection.

tests/testdiscovery.nim
```diff
@@ -32,32 +32,119 @@ proc createSwitch(rdv: RendezVous = RendezVous.new()): Switch =
...
       res[PeerId] == clientB.peerInfo.peerId
       res.getAll(PeerId) == @[clientB.peerInfo.peerId]
       toHashSet(res.getAll(MultiAddress)) == toHashSet(clientB.peerInfo.addrs)
-    await allFutures(clientA.stop(), clientB.stop(), remoteNode.stop())
+
+  asyncTest "Subscribe and unsubscribe":
+    dmB.advertise(rdvNamespace)
+    let
+      query1 = dmA.request(rdvNamespace)
+      res1 = await query1.getPeer().waitForResult(1.seconds)
```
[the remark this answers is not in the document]
AlejandroCabeza: Using `await` means there's no timeout, which is something quite handy for this situation. This operation should take way lower than 1 second, which means if the timeout is hit, there's something very wrong with this and the test should quite. If, by any chance, the reason the timeout is hit is the task is hanging, this makes the test exit and not leave the test suite running forever.

Regarding the name, if you have a look at the tests you will see it's kind of standard practice. If you have a suggestion I'm all ears :)

tests/testdiscovery.nim
AlejandroCabeza: Yes, I could. That being said, I consider casting the `Future` to `Result` a much cleaner approach which also allows for value checking if need be (not in this case, ofc).

tests/utils/async.nim
AlejandroCabeza: `Result` is not a substitute for `Exception`, but a more general any failed computation.
In this case, this simplifies and makes more robust a lot of checks in tests. When you run `withTimeout` you are effectively collapsing the "uncertainty of the future", that is, making it either have a value, or an error. 
Casting that "collapsed future" into a result allows for a more visual and easier handling of that situation.

tests/utils/futures.nim
AlejandroCabeza: I would say `DURATION_TIMEOUT` is not a horrible name. It specifies the type (duration) and the context (timeout). Given that's the default measurement I'm using, I decided to leave the name as is. I believe the `EXTENDED` suffix can be easily understood from that.

The reason I'm specifying them in a file is precisely because I'm using those numbers widely across different files, so it felt a bit repetitive having to repeat them all over the test codebase.

tests/testdiscovery.nim
AlejandroCabeza: Also, I believe that having removed the `xasyncTest`, this PR's scope is just peer discovery.

tests/pubsub/testgossipsub.nim
diegomrsantos: Please, add the test back.

AlejandroCabeza: Even if it's not working?

tests/testdiscovery.nim
```diff
@@ -32,32 +32,119 @@ proc createSwitch(rdv: RendezVous = RendezVous.new()): Switch =
...
+      res2 = await query2.getPeer().waitForResult(1.seconds)
+
+    res2.assertIsErr()
+
+  asyncTest "Frequent sub/unsub":
+    for i in 0 ..< 10:
+      dmB.advertise(rdvNamespace)
+      let
+        query1 = dmA.request(rdvNamespace)
+        res1 = await query1.getPeer().waitForResult(1.seconds)
```
lchenut, line 103: A `query` is a tool to re-use. Recreating one every loop is not the way to do.
It should work just by writing
```nim
asyncTest "Frequent sub/unsub":
  let
    query1 = dmA.request(rdvNamespace)
    query2 = dmA.request(rdvNamespace)
  for i in 0 ..< 10:
    (...)
```

tests/testdiscovery.nim
```diff
@@ -32,32 +32,119 @@ proc createSwitch(rdv: RendezVous = RendezVous.new()): Switch =
...
+
+      res1.assertIsOk()
+      check res1.value{PeerId}.get() == clientB.peerInfo.peerId
+
+      await rdvB.unsubscribe(namespace)
+      await sleepAsync(DURATION_TIMEOUT_EXTENDED)
+      var
+        query2 = dmA.request(rdvNamespace)
```
lchenut, line 111: Shouldn't this be `dmB.request(...)`?

tests/testdiscovery.nim
```diff
@@ -32,32 +32,119 @@ proc createSwitch(rdv: RendezVous = RendezVous.new()): Switch =
...
+        res2 = await query2.getPeer().waitForResult(1.seconds)
+
+      res2.assertIsErr()
+
+  asyncTest "Frequent sub/unsub with multiple clients":
```
lchenut, line 116: Same thing in this test, you should create only one query per discovery manager.

tests/testdiscovery.nim
AlejandroCabeza: Just did that and it doesn't work, it breaks at the 3rd iteration. I don't know what might the correlation be.

lchenut: Then there's a problem in rendezvous or the discovery manager :confused:

AlejandroCabeza: What would you rather I do with this? Leave it as is and open an issue or would you like to investigate it before merging?

lchenut: I would like to investigate, it's a real problem. I'll do this next week.

Issue #1165, opened 2024-08-06
bug: yamux Futures leak


gabrielmer:
When running tests on nwaku and activating [dumpPendingFutures](https://github.com/status-im/nim-chronos/blob/dc3847e4d6733dfc3811454c2a9c384b87343e26/chronos/debugutils.nim#L26) from `nim-chronos`, we noticed that the pending futures grow indefinitely in `yamux`.

Most of them come from the following line https://github.com/vacp2p/nim-libp2p/blob/a60f0c553264bad21bf0af5d71f5dbffa6d147a7/libp2p/muxers/yamux/yamux.nim#L282-L283

For example

```
Future[4433822480] with name "AsyncEvent.wait" created at <asyncsync.nim:171> and state = Pending
```

I see these Futures get created at a really fast pace, and not seeing them getting completed. Even if they do eventually get completed, if we create futures at a faster rate than what it takes to complete them, the pending futures would grow indefinitely.

When using `mplex` instead, the amount of pending Futures is stable and doesn't grow indefinitely

Version: nim-libp2p https://github.com/vacp2p/nim-libp2p/commit/94d93cbf2531296f0e636aa702b500b242eb4158

Attaching logs with prints of the pending Futures for both `yamux` and `mplex`

[logs_yamux.txt](https://github.com/user-attachments/files/16510137/logs_yamux.txt)
[logs_mplex.txt](https://github.com/user-attachments/files/16510136/logs_mplex.txt)

lchenut:
I'll summarize here what we discussed in private about this issue:
Both `channel.closedRemotely.wait()` and `channel.receivedData.wait()` are Future created just for this specific line.
`race` await for one Future to be finished. By design all the others unfinished Futures are not cancelled when the first Future is completed. Therefore, each time `readOnce` is called (and it's frequent) two Futures are created and only one is finished for sure by the end of the procedure.

We can reproduce the problem with the following code (compiling with `-d:chronosFutureTracking`)
```nim
import chronos, random

when defined(chronosFutureTracking):
  proc dumpChronosInfo() {.async.} =
    while true:
      echo "\n----------- Dumping Chronos Info -----------"
      echo dumpPendingFutures(OnlyPending)
      await sleepAsync(1.seconds)

  asyncSpawn dumpChronosInfo()

proc main() {.async.} =
  while true:
    discard await race(sleepAsync(50.millis()), sleepAsync(10.seconds()))

waitFor(main())
```
[...]

PR #1171, opened 2024-08-09
fix(yamux): future leak

lchenut:
At first glance, this fixes https://github.com/vacp2p/nim-libp2p/issues/1165.

It raises another issue: I always thought that Future unfinished caught in the GC were cancelled. But that doesn't always seem to be the case.

### DIFF
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -282,11 +282,12 @@ method readOnce*(
     let
       closedRemotelyFut = channel.closedRemotely.wait()
       receivedDataFut = channel.receivedData.wait()
+    defer:
+      if not closedRemotelyFut.finished():
+        await closedRemotelyFut.cancelAndWait()
+      if not receivedDataFut.finished():
+        await receivedDataFut.cancelAndWait()
     await closedRemotelyFut or receivedDataFut
-    if not closedRemotelyFut.finished():
-      await closedRemotelyFut.cancelAndWait()
-    if not receivedDataFut.finished():
-      await receivedDataFut.cancelAndWait()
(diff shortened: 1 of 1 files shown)


libp2p/muxers/yamux/yamux.nim
```diff
@@ -279,10 +279,15 @@ method readOnce*(
...
-      raiseAssert("Futures list is not empty")
+    let
+      closedRemotelyFut = channel.closedRemotely.wait()
+      receivedDataFut = channel.receivedData.wait()
+    defer:
+      if not closedRemotelyFut.finished():
+        await closedRemotelyFut.cancelAndWait()
+      if not receivedDataFut.finished():
+        await receivedDataFut.cancelAndWait()
+    await closedRemotelyFut or receivedDataFut
```
diegomrsantos, line 290: It is probably better to use
```nim
proc raceCancel*(
    futs: seq[InternalRaisesFuture]
): Future[FutureBase] {.async: (raises: [ValueError, CancelledError]).} =
  try:
    race(futs)
  finally:
    await noCancel allFutures(futs.filterIt(not it.finished).mapIt(it.cancelAndWait))
```

diegomrsantos: It's also going to handle cancelations.

lchenut: I'm not a huge fan of using `InternalRaisesFuture` outside of Chronos tbh.

diegomrsantos: Me neither, but that's the only way for now, unless SomeFuture is changed to public.

diegomrsantos (approved):

PR #1172, opened 2024-08-09
chore(async): `{.async: (raises).}` for `connmanager`

lchenut:
Add `{.async: (raises).}` to connmanager.nim and to semaphore.nim (which is needed by connmanager.nim)

### DIFF
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -67,7 +67,8 @@ type
     muxed: Table[PeerId, seq[Muxer]]
     connEvents: array[ConnEventKind, OrderedSet[ConnEventHandler]]
     peerEvents: array[PeerEventKind, OrderedSet[PeerEventHandler]]
-    expectedConnectionsOverLimit*: Table[(PeerId, Direction), Future[Muxer]]
+    expectedConnectionsOverLimit*:
+      Table[(PeerId, Direction), Future[Muxer].Raising([CancelledError])]
     peerStore*: PeerStore
 
   ConnectionSlot* = object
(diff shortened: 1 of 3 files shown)


libp2p/connmanager.nim
```diff
@@ -67,7 +67,8 @@ type
     muxed: Table[PeerId, seq[Muxer]]
     connEvents: array[ConnEventKind, OrderedSet[ConnEventHandler]]
     peerEvents: array[PeerEventKind, OrderedSet[PeerEventHandler]]
-    expectedConnectionsOverLimit*: Table[(PeerId, Direction), Future[Muxer]]
+    expectedConnectionsOverLimit*:
+      Table[(PeerId, Direction), Future[Muxer].Raising([CancelledError])]
```
AlejandroCabeza, line 71: What would be the difference between this and using a `Future[Result[Muxer, ...]]`.

Actually, I guess the broader question would be: Is there a reason we're going for exceptions over handling errors with `Result`?

diegomrsantos: I'm not an expert, but Nim natively supports exceptions and there's also compiler support for pros signatures https://nim-lang.org/docs/tut2.html#exceptions-annotating-procs-with-raised-exceptions. `Result`was created in some status library. It might be painful to use `Result`if Nim std wasn't designed for that. I can't say much as in general, this project uses more exceptions than `Result`.

libp2p/utils/semaphore.nim
```diff
@@ -38,14 +38,14 @@ proc tryAcquire*(s: AsyncSemaphore): bool =
...
-proc acquire*(s: AsyncSemaphore): Future[void] =
+proc acquire*(s: AsyncSemaphore): Future[void].Raising([CancelledError]) =
   ## Acquire a resource and decrement the resource
   ## counter. If no more resources are available,
   ## the returned future will not complete until
   ## the resource count goes above 0.
   ##
 
-  let fut = newFuture[void]("AsyncSemaphore.acquire")
+  let fut = Future[void].Raising([CancelledError]).init("AsyncSemaphore.acquire")
```
diegomrsantos, line 48: we can create a new template in utils/future.nim:
```nim
 template newFuture*[T](fromProc: static[string] = ""): auto =
  Future[T].Raising([CancelledError]).init(fromProc)
 ```

vladopajic:
this was fixed in https://github.com/vacp2p/nim-libp2p/pull/1263

PR #1175, opened 2024-08-11
create raceCancell

diegomrsantos:


AlejandroCabeza (approved):

libp2p/utils/future.nim
```diff
@@ -31,3 +32,19 @@ proc anyCompleted*[T](futs: seq[Future[T]]): Future[Future[T]] {.async.} =
 
     let index = requests.find(raceFut)
     requests.del(index)
+
+proc raceAndCancelPending*(
+    futs: seq[SomeFuture]
+): Future[void] {.async: (raises: [ValueError, CancelledError]).} =
```
vladopajic, line 38: `raceAndCancelPending` should return future that finished the race, aka result from `race`, because one might need this future.

libp2p/utils/future.nim
```diff
@@ -31,3 +32,19 @@ proc anyCompleted*[T](futs: seq[Future[T]]): Future[Future[T]] {.async.} =
 
     let index = requests.find(raceFut)
     requests.del(index)
+
+proc raceAndCancelPending*(
```
vladopajic, line 36: how about `raceAndCancelPending` -> `raceAndCancel` ?

vladopajic:
> This looks useful, and perhaps should exist in chronos 🤔

seems better suited for cronos. 

@arnetheduck wdyt?

libp2p/utils/future.nim
arnetheduck: it should indeed, mostly because race is declared that way - that said, returning the future is actually of limited use since the value type is lost - most uses of `race` end up discarding it and using the "original" instance.

The other risk is that one forgets that more than one future might be finished by the time `race` resumes.

libp2p/utils/future.nim
```diff
@@ -31,3 +32,19 @@ proc anyCompleted*[T](futs: seq[Future[T]]): Future[Future[T]] {.async.} =
...
+  ## Cancels any remaining futures that have not yet completed.
+  ##
+  ## - `futs`: A sequence of futures to race.
+  ##
+  ## Raises:
+  ## - `ValueError` if the sequence of futures is empty.
+  ## - `CancelledError` if the operation is canceled.
+  try:
+    discard await race(futs)
+  finally:
```
arnetheduck, line 49: there is no need for finally .. in particular, if `race` is cancelled, it should not cancel `futs` to maintain the `race` contract in general - ValueError means there are no futures, so nothing to cancel either

vladopajic:
> > This looks useful
> 
> It looks useful, but the question is whether it _is_ useful ;) It would be good to see a motivating example where it's actually put to use before adding to chronos proper - ie where in libp2p would you use it, for example?

closest usecase:
https://github.com/vacp2p/nim-quic/blob/e5bcee4b909a9924fa6fbea305884888ce466330/quic/connection.nim#L173-L202

arnetheduck:
> closest usecase:

This is a classic example of the trap I mentioned - even if the handshake completes successfully by the time the race finishes, the code might prefer to treat it as a timeout which often isn't what you want - but more than that, it's worth rewriting the code to use `raceOrCancel` in a few different situations to see that it truly works (with all the generics, error conditions etc) and that the outcome really looks better - it's a lot easier to add things to chronos (and libp2p btw) than it is to remove them, so we want to make sure the ergonomics have been practically stressed before adding too much

Issue #1174, opened 2024-08-09
Add separate CI job using SAT solver


AlejandroCabeza:
Add a daily job that runs with the sat solver so we can verify how reliable it is over a few weeks.

PR #1177, opened 2024-08-13
ci: Enable conditional SAT solving

AlejandroCabeza:
Use only on a daily job so we test out if it behaves correctly.

closes: https://github.com/vacp2p/nim-libp2p/issues/1174

### DIFF
--- a/.github/workflows/daily_common.yml
+++ b/.github/workflows/daily_common.yml
@@ -91,6 +91,12 @@ jobs:
         run: |
           nim --version
           nimble --version
-          [[ ${{ inputs.use_sat_solver }} ]] && dependency_solver="sat" || dependency_solver="legacy"
-          NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }} --solver:$dependency_solver"
+
+          if [[ "${{ inputs.use_sat_solver }}" == "true" ]]; then
+            dependency_solver="sat"
+          else
+            dependency_solver="legacy"
+          fi
+
+          NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }} --solver:${dependency_solver}"
           nimble test
--- a/.github/workflows/daily_common.yml
+++ b/.github/workflows/daily_common.yml
@@ -92,6 +92,5 @@ jobs:
           nim --version
           nimble --version
           [[ ${{ inputs.use_sat_solver }} ]] && dependency_solver="sat" || dependency_solver="legacy"
-          NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }} --solver:$dependency_solver" 
-          echo "NIMFLAGS: $NIMFLAGS"
+          NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }} --solver:$dependency_solver"
           nimble test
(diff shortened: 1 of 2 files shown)


AlejandroCabeza:
> Please make sure the changes are working as expected.

Verified.

.github/workflows/daily_common.yml
```diff
@@ -86,4 +91,7 @@ jobs:
         run: |
           nim --version
           nimble --version
-          NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }}" nimble test
+          [[ ${{ inputs.use_sat_solver }} ]] && dependency_solver="sat" || dependency_solver="legacy"
```
diegomrsantos, line 94:
```suggestion
    if [[ "${{ inputs.use_sat_solver }}" == "true" ]]; then
      dependency_solver="sat"
    else
      dependency_solver="legacy"
    fi
```

.github/workflows/daily_common.yml
```diff
@@ -86,4 +91,7 @@ jobs:
...
+          NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }} --solver:$dependency_solver"
```
diegomrsantos, line 95:
```suggestion
NIMFLAGS="--mm:${{ matrix.nim.memory_management }} --solver:${dependency_solver}"
```

AlejandroCabeza: This is written as it was before: Instead of assuming `NIMFLAGS` is empty we just append the values to it, as it's common practice.

.github/workflows/daily_common.yml
AlejandroCabeza: Any reason you oppose to ternary?

.github/workflows/daily_common.yml
diegomrsantos: Right, I updated it.

.github/workflows/daily_common.yml
diegomrsantos: It's more readable in this way.

diegomrsantos: But just a suggestion, up to you

.github/workflows/daily_common.yml
diegomrsantos: It's about the missing `{}`

diegomrsantos: It's probably better to change the name to upper case too.

.github/workflows/daily_common.yml
AlejandroCabeza: Actually, you're right; that ternary is a bit convoluted.

.github/workflows/daily_common.yml
AlejandroCabeza: Oh, I see, I misread it.

AlejandroCabeza: Ah, I see, I misunderstood.
Right, it would work regardless as `$variable` can be used interchangeably with `${variable}`; not the other way around, though.

AlejandroCabeza: I'd kinda keep the lowercase, though, as uppercase is used for env variables while lowercase for local ones. IIRC.

diegomrsantos (approved):
Thank you

Issue #1176, opened 2024-08-12
Stop creating duplicate msg keys in logs


jakubgs:
We have identified an example of a log that contains the same `msg` key twice while reviewing Nimbus logs:
```json
{"msg":"No send connection","topics":"libp2p pubsubpeer","p":"16U*Y1w1AH","msg":"12a30212ef01...736e61707079"}
```
Which is there:
https://github.com/vacp2p/nim-libp2p/blob/cde5ed7e8ccc3b452878db4c82c6f2f2e70d28f4/libp2p/protocols/pubsub/pubsubpeer.nim#L344
Which led me to check if you have any other examples of this:
https://github.com/search?q=repo%3Avacp2p%2Fnim-libp2p+%22%2C+msg+%3D+%22&type=code
And you do, about 22 instances of this mistake. Which is not great because it makes parsing the logs with Logstash impossible.

Would appreciate a fix.

PR #1180, opened 2024-08-13
chore(logs): remove duplicate msg key

AlejandroCabeza:
Chronicles uses `msg` for the string that describes the log, so this PR removes the duplicate `msg` usage.

closes: https://github.com/vacp2p/nim-libp2p/issues/1176

### DIFF
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -136,7 +136,8 @@ proc triggerConnEvent*(c: ConnManager, peerId: PeerId, event: ConnEvent) {.async
   except CancelledError as exc:
     raise exc
   except CatchableError as exc:
-    warn "Exception in triggerConnEvents", msg = exc.msg, peer = peerId, event = $event
+    warn "Exception in triggerConnEvents",
+      errMsg = exc.msg, peer = peerId, event = $event
 
 proc addPeerEventHandler*(
     c: ConnManager, handler: PeerEventHandler, kind: PeerEventKind
@@ -169,7 +170,7 @@ proc triggerPeerEvents*(c: ConnManager, peerId: PeerId, event: PeerEvent) {.asyn
   except CancelledError as exc:
     raise exc
   except CatchableError as exc: # handlers should not raise!
-    warn "Exception in triggerPeerEvents", exc = exc.msg, peer = peerId
+    warn "Exception in triggerPeerEvents", errMsg = exc.msg, peer = peerId
 
 proc expectConnection*(
     c: ConnManager, p: PeerId, dir: Direction
@@ -212,7 +213,7 @@ proc closeMuxer(muxer: Muxer) {.async.} =
     try:
       await muxer.handler # TODO noraises?
     except CatchableError as exc:
-      trace "Exception in close muxer handler", exc = exc.msg
+      trace "Exception in close muxer handler", errMsg = exc.msg
   trace "Cleaned up muxer", m = muxer
 
 proc muxCleanup(c: ConnManager, mux: Muxer) {.async.} =
@@ -235,7 +236,7 @@ proc muxCleanup(c: ConnManager, mux: Muxer) {.async.} =
   except CatchableError as exc:
     # This is top-level procedure which will work as separate task, so it
     # do not need to propagate CancelledError and should handle other errors
-    warn "Unexpected exception peer cleanup handler", mux, msg = exc.msg
+    warn "Unexpected exception peer cleanup handler", mux, errMsg = exc.msg
 
 proc onClose(c: ConnManager, mux: Muxer) {.async.} =
   ## connection close even handler
@@ -358,7 +359,7 @@ proc trackConnection*(cs: ConnectionSlot, conn: Connection) =
     try:
       await conn.join()
     except CatchableError as exc:
-      trace "Exception in semaphore monitor, ignoring", exc = exc.msg
+      trace "Exception in semaphore monitor, ignoring", errMsg = exc.msg
 
     cs.release()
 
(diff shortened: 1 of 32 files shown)


libp2p/connmanager.nim
```diff
@@ -169,7 +170,7 @@ proc triggerPeerEvents*(c: ConnManager, peerId: PeerId, event: PeerEvent) {.asyn
   except CancelledError as exc:
     raise exc
   except CatchableError as exc: # handlers should not raise!
-    warn "Exception in triggerPeerEvents", exc = exc.msg, peer = peerId
+    warn "Exception in triggerPeerEvents", errMsg = exc.msg, peer = peerId
```
diegomrsantos, line 173: What about `errorMsg`?

AlejandroCabeza: I decided to go with that one because `errMsg` is already widely used across the codebase.

diegomrsantos: I see, I didn't realize it. Thanks

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -106,7 +106,7 @@ method rpcHandler*(f: FloodSub, peer: PubSubPeer, data: seq[byte]) {.async.} =
     debug "failed to decode msg from peer", peer, err = error
     raise newException(CatchableError, "Peer msg couldn't be decoded")
 
-  trace "decoded msg from peer", peer, msg = rpcMsg.shortLog
+  trace "decoded msg from peer", peer, payload = rpcMsg.shortLog
```
diegomrsantos, line 109: msgShortLog?

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -219,7 +219,7 @@ method publish*(f: FloodSub, topic: string, data: seq[byte]): Future[int] {.asyn
       trace "Error generating message id, skipping publish", error = error
       return 0
 
-  trace "Created new message", msg = shortLog(msg), peers = peers.len, topic, msgId
+  trace "Created new message", payload = shortLog(msg), peers = peers.len, topic, msgId
```
diegomrsantos, line 222: msgShortLog?

libp2p/protocols/pubsub/floodsub.nim
AlejandroCabeza: Kind of a similar reasoning: `payload` is a widely accepted word for this situations.

AlejandroCabeza: Kind of the same reasoning, `payload` is a widely accepted word for this situations.

diegomrsantos: This is not logging the whole content https://github.com/vacp2p/nim-libp2p/blob/96bfefc928ee06924a4e97cd8910dd541c0cf328/libp2p/utility.nim#L24

libp2p/connmanager.nim
```diff
@@ -212,7 +213,7 @@ proc closeMuxer(muxer: Muxer) {.async.} =
     try:
       await muxer.handler # TODO noraises?
     except CatchableError as exc:
-      trace "Exception in close muxer handler", exc = exc.msg
+      trace "Exception in close muxer handler", errMsg = exc.msg
```
jakubgs, line 216: Are we sure we want to use the name `errMsg` for logs that are `trace` or `debug` level? Seems kinda weird to me. Also would make filtering logs with `grep` using `err` show not just actual errors but more like low level warning messages about poor connections with peers.

Thanks for the fix, but seems to me like using the keyword `err` in this kinda misses the point of having log levels.

diegomrsantos: I agree, how about `description`?

jakubgs: Sounds good to me.

libp2p/protocols/pubsub/floodsub.nim
AlejandroCabeza: Would that be an issue?

libp2p/connmanager.nim
AlejandroCabeza: Fair point, didn't think about the `| grep ERR` possibility. Fine by me, will change it.

libp2p/protocols/pubsub/floodsub.nim
diegomrsantos: `shortLog` creates a shortened representation of the data, it's not the payload. What about `summary`?

jakubgs (approved):

libp2p/protocols/pubsub/floodsub.nim
AlejandroCabeza: `shortLog` is actually a bit of an abused name, as it vastly depends on the object being shortlogged: Some are actual short representations, some others are obfuscated representations, and others are the full object representations. 

What I mean is: There's not a unique key that perfectly handles all cases; unless we go for a very generic `context` or the likes, which doesn't really provide meaning. Ideally it should be handled on a case by case basis, but I think that's a different task altogether.

Personally, I'd stick to `payload` as it's already there and we can get it merged already, and create a chore to keep track of this naming issue and spend on time whenever we feel appropriate. That is, unless you strongly oppose it.

diegomrsantos: Ok, could you please create an issue?

diegomrsantos (approved):

libp2p/protocols/pubsub/floodsub.nim
AlejandroCabeza: Done: https://github.com/vacp2p/nim-libp2p/issues/1181

PR #1182, opened 2024-08-19
fix(ci): Daily workflow parent's name

strinnityk:


diegomrsantos:
Have you tested it to see if it fixes the issue?

PR #1183, opened 2024-08-28
feat: rendezvous refactor

SionoiS:
Hello!

This PR aim to refactor rendezvous code so that it is easier to impl. Waku rdv strategy. The hardcoded min and max TTL were out of range with what we needed and specifying which peers to interact with is also needed since Waku deals with peers on multiple separate shards.

I tried to keep the changes to a minimum, specifically I did not change the name of any public procs which result in less than descriptive names in some cases. I also wanted to return results instead of raising exceptions but didn't. Would it be acceptable to do so?

Please advise on best practices, thank you.

### DIFF
--- a/libp2p/protocols/rendezvous.nim
+++ b/libp2p/protocols/rendezvous.nim
@@ -515,7 +515,7 @@ proc batchAdvertise*(
     raise newException(RendezVousError, "Invalid namespace")
 
   if ttl notin rdv.minDuration .. rdv.maxDuration:
-    raise newException(RendezVousError, "Invalid time to live")
+    raise newException(RendezVousError, "Invalid time to live: " & $ttl)
 
   let sprBuff = rdv.switch.peerInfo.signedPeerRecord.encode().valueOr:
     raise newException(RendezVousError, "Wrong Signed Peer Record")
@@ -690,6 +690,15 @@ proc new*(
     minDuration = MinimumDuration,
     maxDuration = MaximumDuration,
 ): T =
+  if minDuration < 1.minutes:
+    raiseAssert("TTL too short: 1 minute minimum")
+
+  if maxDuration > 72.hours:
+    raiseAssert("TTL too long: 72 hours maximum")
+
+  if minDuration >= maxDuration:
+    raiseAssert("Minimum TTL longer than maximum")
+
   let
     minTTL = minDuration.seconds.uint64
     maxTTL = maxDuration.seconds.uint64
--- a/libp2p/protocols/rendezvous.nim
+++ b/libp2p/protocols/rendezvous.nim
@@ -33,10 +33,6 @@ declareGauge(libp2p_rendezvous_namespaces, "number of registered namespaces")
 
 const
   RendezVousCodec* = "/rendezvous/1.0.0"
-  MinimumDuration* = 2.hours
-  MaximumDuration = 72.hours
-  MinimumTTL = MinimumDuration.seconds.uint64
-  MaximumTTL = MaximumDuration.seconds.uint64
   RegistrationLimitPerPeer = 1000
   DiscoverLimit = 1000'u64
   SemaphoreDefaultSize = 5
@@ -320,6 +316,10 @@ type
     peers: seq[PeerId]
     cookiesSaved: Table[PeerId, Table[string, seq[byte]]]
     switch: Switch
+    minimumDuration: Duration
+    maximumDuration: Duration
+    minimumTTL: uint64
+    maximumTTL: uint64
 
 proc checkPeerRecord(spr: seq[byte], peerId: PeerId): Result[void, string] =
   if spr.len == 0:
@@ -395,7 +395,7 @@ proc save(
     rdv.registered.add(
       RegisteredData(
         peerId: peerId,
-        expiration: Moment.now() + r.ttl.get(MinimumTTL).int64.seconds,
+        expiration: Moment.now() + r.ttl.get(rdv.minimumTTL).int64.seconds,
         data: r,
       )
     )
@@ -409,8 +409,8 @@ proc register(rdv: RendezVous, conn: Connection, r: Register): Future[void] =
   libp2p_rendezvous_register.inc()
   if r.ns.len notin 1 .. 255:
     return conn.sendRegisterResponseError(InvalidNamespace)
-  let ttl = r.ttl.get(MinimumTTL)
-  if ttl notin MinimumTTL .. MaximumTTL:
+  let ttl = r.ttl.get(rdv.minimumTTL)
+  if ttl notin rdv.minimumTTL .. rdv.maximumTTL:
     return conn.sendRegisterResponseError(InvalidTTL)
   let pr = checkPeerRecord(r.signedPeerRecord, conn.peerId)
   if pr.isErr():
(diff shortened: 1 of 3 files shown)


diegomrsantos:
Could you please add to the PR description a high-level overview of the changes and if possible how they make the protocol easier to use?

lchenut:
The reason why min and max TTL are hardcoded is because those limits are set by the specifications: https://github.com/libp2p/specs/blob/master/rendezvous/README.md?plain=1#L270-L273.
In theory, you shouldn't have to change those.

libp2p/protocols/rendezvous.nim
```diff
@@ -662,14 +682,27 @@ proc setup*(rdv: RendezVous, switch: Switch) =
   rdv.switch.addPeerEventHandler(handlePeer, Joined)
   rdv.switch.addPeerEventHandler(handlePeer, Left)
 
-proc new*(T: typedesc[RendezVous], rng: ref HmacDrbgContext = newRng()): T =
+proc new*(
+    T: typedesc[RendezVous],
+    rng: ref HmacDrbgContext = newRng(),
+    minimumDuration = 2.hours,
+    maximumDuration = 72.hours,
```
diegomrsantos, line 689: What about making those values constants to avoid duplication?

libp2p/protocols/rendezvous.nim
```diff
@@ -701,9 +734,13 @@ proc new*(T: typedesc[RendezVous], rng: ref HmacDrbgContext = newRng()): T =
   return rdv
 
 proc new*(
-    T: typedesc[RendezVous], switch: Switch, rng: ref HmacDrbgContext = newRng()
+    T: typedesc[RendezVous],
+    switch: Switch,
+    rng: ref HmacDrbgContext = newRng(),
+    minimumDuration = 2.hours,
+    maximumDuration = 72.hours,
```
diegomrsantos, line 741: What about just `minDuration` and `maxDuration`?

diegomrsantos:
@SionoiS is the PR's purpose to make those hardcoded values configurable?

SionoiS:
Anything else I can do or suggestions for this PR @diegomrsantos @lchenut ?

Waku's rendezvous work is blocked by this.

cc @jm-clius

diegomrsantos:
>  I also wanted to return results instead of raising exceptions but didn't. Would it be acceptable to do so?

The use of exceptions and results isn't consistent in the project, we still need to define a new strategy for new code. I would stick with what is implemented instead of changing it when it comes to existing code.

libp2p/protocols/rendezvous.nim
```diff
@@ -506,24 +508,35 @@ proc advertisePeer(rdv: RendezVous, peer: PeerId, msg: seq[byte]) {.async.} =
...
-    raise newException(RendezVousError, "Wrong Signed Peer Record")
+proc batchAdvertise*(
+    rdv: RendezVous, ns: string, ttl: Duration, peers: seq[PeerId]
+) {.async.} =
   if ns.len notin 1 .. 255:
     raise newException(RendezVousError, "Invalid namespace")
-  if ttl notin MinimumDuration .. MaximumDuration:
+
+  if ttl notin rdv.minDuration .. rdv.maxDuration:
     raise newException(RendezVousError, "Invalid time to live")
```
diegomrsantos, line 518: I see the code was already like that, but given we are touching it, I'd add what the wrong value was.

lchenut:
@SionoiS I rechecked and the only thing that annoys me with this PR is that you do not verify anything about the minimum and the maximum ttl. 
- Users can write a minimum bigger than the maximum, which will cause a weird behavior.
- While I get why you could use a shorter TTL, using one bigger than 72 hours seems a bit excessive, I would limit the maximum.

But overall, this PR looks great.

libp2p/protocols/rendezvous.nim
```diff
@@ -701,9 +736,13 @@ proc new*(T: typedesc[RendezVous], rng: ref HmacDrbgContext = newRng()): T =
...
 proc new*(
-    T: typedesc[RendezVous], switch: Switch, rng: ref HmacDrbgContext = newRng()
+    T: typedesc[RendezVous],
+    switch: Switch,
+    rng: ref HmacDrbgContext = newRng(),
+    minDuration = MinimumDuration,
+    maxDuration = MaximumDuration,
 ): T =
-  let rdv = T.new(rng)
+  let rdv = T.new(rng, minDuration, maxDuration)
```
diegomrsantos, line 745: Good catch from @lchenut. We should check invalid values from min and max. I see that it is possible to create a negative `Duration` instance. Does it make sense? If so, we need to validate it here too.

SionoiS:
Thanks for the comments. I will make the changes.

lchenut (approved):
LGTM!

libp2p/protocols/rendezvous.nim
```diff
@@ -690,6 +690,15 @@ proc new*(
     minDuration = MinimumDuration,
     maxDuration = MaximumDuration,
 ): T =
+  if minDuration < 1.minutes:
+    raiseAssert("TTL too short: 1 minute minimum")
```
diegomrsantos, line 694: I'm not sure raising a Defect here is appropriate.

diegomrsantos (changes requested):
Please address the defect concern and this PR should also include tests for the features added.

libp2p/protocols/rendezvous.nim
lchenut: Good catch, you're right, raiseAssert is not the good call here... I did my last review a bit too fast (or too early).

libp2p/protocols/rendezvous.nim
```diff
@@ -662,14 +684,36 @@ proc setup*(rdv: RendezVous, switch: Switch) =
...
+  let
+    minTTL = minDuration.seconds.uint64
+    maxTTL = maxDuration.seconds.uint64
+
   let rdv = T(
     rng: rng,
     salt: string.fromBytes(generateBytes(rng[], 8)),
     registered: initOffsettedSeq[RegisteredData](1),
     defaultDT: Moment.now() - 1.days,
     #registerEvent: newAsyncEvent(),
```
diegomrsantos, line 711: @lchenut just realized this. Do you remember why is this commented?

lchenut: I don't. I guess it's an artifact of an old implementation. Because it's not really useful afaik

SionoiS:
The first thing I did was raise exceptions but I forgot about the pragma and was confused it didn't work so I tried something else. We good now.

I also add a test for misconfig.

diegomrsantos:
> specifying which peers to interact with is also needed since Waku deals with peers on multiple separate shards.

Thank you for the changes. Could you please elaborate more on this feature?

SionoiS:
> > specifying which peers to interact with is also needed since Waku deals with peers on multiple separate shards.
> 
> Thank you for the changes. Could you please elaborate more on this feature?

Waku rendezvous strategy will be to register at rdv points for multiple namespaces specific to GossipSub topics that a node uses. Without a way to choose specific peers to register at, it was not possible to impl. this strategy.

libp2p/protocols/rendezvous.nim
```diff
@@ -506,24 +508,35 @@ proc advertisePeer(rdv: RendezVous, peer: PeerId, msg: seq[byte]) {.async.} =
   await rdv.sema.acquire()
   discard await advertiseWrap().withTimeout(5.seconds)
 
-method advertise*(
-    rdv: RendezVous, ns: string, ttl: Duration = MinimumDuration
-) {.async, base.} =
-  let sprBuff = rdv.switch.peerInfo.signedPeerRecord.encode().valueOr:
-    raise newException(RendezVousError, "Wrong Signed Peer Record")
+proc batchAdvertise*(
```
diegomrsantos, line 511: I have just realized this has been changed from a method to a proc, could you please explain why?

SionoiS: The `advertise` method now call a public `batchAdvertise` proc so that the interface still work like before but you can also use the proc with the new parameters.

diegomrsantos: @lchenut does it make sense to pass the peers as a parameter here instead of using the seq inside the rdv instance?

lchenut: There's really nothing against it. In theory, the rendezvous protocol do it with the peer event handler on joined/left, but if you want to use your own, why not. It wasn't something we needed at the beginning (because we don't really use it outside the test), but if waku needs it, it's a good thing to implement.

diegomrsantos: Ideally, it would be good to add docs explaining the use of those procs and testing covering them. I don't know if different names for them are a good idea either.

SionoiS: > Ideally, it would be good to add docs explaining the use of those procs and testing covering them. I don't know if different names for them are a good idea either.

How would I update the docs? Is there a different repo?

Tests already call the new procs via the methods so it is covered as for names I'm open to suggestions.

diegomrsantos (approved):
Thank you, it looks good.

libp2p/protocols/rendezvous.nim
diegomrsantos: @lchenut could you please create an issue to add docs to the procs?

diegomrsantos: @SionoiS the docs are added right below the proc definition. For the name, I'd just use `advertise` as both procs do the same thing.

SionoiS:
I changed the names but I'll let someone else do the docs since there's none yet.

diegomrsantos (approved):

libp2p/protocols/rendezvous.nim
diegomrsantos: @lchenut it'd be good to add to the issue an evaluation of the use of methods and procs.

lchenut: I'm not sure I understand your last point

diegomrsantos: Some things are implemented as procs and some as methods, why?

lchenut: I did a little checking. The Rendezvous protocol inherits from LPProtocol, so it should have a `start` and `stop` method, which it does. But `advertise` is also a method, which is wrong, it should be proc.

diegomrsantos: This explains the difference and when to use one or the other https://matthiashager.com/proc-method-nim

PR #1185, opened 2024-08-30
chore(connmanager): also show peerID and dir when too many conns

Ivansete-status:
Bring a bit more detail when the "Too many connections for peer" message is being logged. Particularly, we are adding the offending `peerId` and the stream direction


diegomrsantos:
Please add a description explaining why this is necessary. The reason is to prevent people from removing this info in the future if they don't know why it was added.

AlejandroCabeza (approved):

Ivansete-status:
> Please add a description explaining why this is necessary. The reason is to prevent people from removing this info in the future if they don't know why it was added.

Sure! Sorry for the delay in answering. I just added a description now. Thanks!

PR #1189, opened 2024-09-06
fix(ci): Fix dependency solver flag not being parsed.

strinnityk:


lchenut (approved):
LGTM! :+1:

AlejandroCabeza:
> There's a mistake in the PR. The `pull_request` trigger hasn't been removed. Moreover, this PR needs to make sure the changes are working as expected.

I don't think I follow. The `pull_request` trigger hasn't been removed because it remains as proof that it's indeed working as expected. Could you elaborate what would you need me to do next?

PR #1190, opened 2024-09-08
chore: move mm:refc config and remove skipParentCfg 

diegomrsantos:
- Moving the `mm:refc` configuration to the main `config.nims` file makes it be used in tests and tasks that build examples.
- It's not clear why `skipParentCfg`was being used and it doesn't seem to be necessary.


lchenut (approved):
LGTM

Issue #1197, opened 2024-09-19
Daily jobs report success even when workflows failed


diegomrsantos:
One example https://github.com/vacp2p/nim-libp2p/actions/runs/10935762205. Jobs must report failure when there are failed jobs.

PR #1200, opened 2024-09-24
fix(ci): Daily workflows report

AlejandroCabeza:
Fix daily workflows' green tick when jobs failed.

Based of off: https://stackoverflow.com/a/58859404

Closes:  https://github.com/vacp2p/nim-libp2p/issues/1197

### DIFF
--- a/.github/workflows/daily_common.yml
+++ b/.github/workflows/daily_common.yml
@@ -60,7 +60,7 @@ jobs:
 
     name: '${{ matrix.platform.os }}-${{ matrix.cpu }} (Nim ${{ matrix.nim.branch }})'
     runs-on: ${{ matrix.platform.builder }}
-    continue-on-error: ${{ matrix.nim.branch == 'devel' || matrix.nim.branch == 'version-2-0' }}
+    # continue-on-error: ${{ matrix.nim.branch == 'devel' || matrix.nim.branch == 'version-2-0' }}
     # if: success() || failure() # Continue jobs even if previous jobs failed, but not if cancelled.
     steps:
       - name: Checkout
--- a/.github/workflows/daily_common.yml
+++ b/.github/workflows/daily_common.yml
@@ -60,8 +60,6 @@ jobs:
 
     name: '${{ matrix.platform.os }}-${{ matrix.cpu }} (Nim ${{ matrix.nim.branch }})'
     runs-on: ${{ matrix.platform.builder }}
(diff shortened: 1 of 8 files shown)


.github/workflows/daily_common.yml
```diff
@@ -60,7 +60,7 @@ jobs:
 
     name: '${{ matrix.platform.os }}-${{ matrix.cpu }} (Nim ${{ matrix.nim.branch }})'
     runs-on: ${{ matrix.platform.builder }}
-    continue-on-error: ${{ matrix.nim.branch == 'devel' || matrix.nim.branch == 'version-2-0' }}
+    if: success() || failure() # Continue jobs even if previous jobs failed, but not if cancelled.
```
lchenut, line 63: I'm not 100% sure this change will fix the problem. I just see that as a way to write `continue-on-error` differently. But I dig into this a bit too much and I think removing this line will solve the issue.
With `fail-fast: false` defined in the strategy up above, the jobs will continue on success and failure, even without this line (or the `continue-on-error`).

AlejandroCabeza: I'm also not completely sure. But according to the post I mentioned in the description it apparently will do what it suggests. 

I could set the daily jobs to run on PR so we can get the results here, now that I think about it.

AlejandroCabeza: It seems it works, as the jobs have been marked as failed? I can't say for sure, though.

lchenut: Could you try without this line, just for my own sanity? :smile:

lchenut: Besides that, like you said, it seems it worked. Just remove the `on pull_request` and I approve it.

AlejandroCabeza: Here the checks show as failed, but if you go check the `Actions` tab you'll see a difference between them. The workflows run with `continue-on-error` show as green tick (even if they failed), while the ones with the `if: success() || failure()` show as failed; which is the intended behaviour.

AlejandroCabeza: Oh, I see. Doing that :)

lchenut: It seems to work without it.

AlejandroCabeza: True. Shall we go for this?

lchenut (approved):

PR #1201, opened 2024-09-26
test(gossipsub): Topic Membership Tests

shashankshampi:
Test Plan: https://www.notion.so/Gossipsub-651e02d4d7894bb2ac1e4edb55f3192d

Topic Membership Tests:

Test added: (TC6.6: Verify the handling of subscription changes when the underlying pubsub framework sends SUBSCRIBE and UNSUBSCRIBE events.)

1. Simulate the `SUBSCRIBE` event and check proper handling in the mesh and gossipsub structures
2. This test will simulate an UNSUBSCRIBE event and check if the topic is removed from the relevant data structures but remains in gossipsub
3. This test ensures that multiple topics can be subscribed to and unsubscribed from, with proper initialization of the topic structures.
4. This test ensures that the number of subscriptions does not exceed the limit set in the GossipSub parameters

Part 2:

TC6.1: Verify that peers can correctly join a topic using JOIN(topic).
TC6.2: Ensure that peers can correctly leave a topic using LEAVE(topic).
TC6.5: Multiple peers join and leave topic simultaneously


diegomrsantos:
how about `testgossiptopicmembership.nim`?

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,184 @@
...
+const MsgIdSuccess = "msg id gen success"
+
+suite "GossipSub Topic Membership Tests":
+  teardown:
+    checkTrackers()
+
+  # Addition of Designed Test cases for 6. Topic Membership Tests: https://www.notion.so/Gossipsub-651e02d4d7894bb2ac1e4edb55f3192d
+
+  # Simulate the `SUBSCRIBE` event and check proper handling in the mesh and gossipsub structures
+  asyncTest "handle SUBSCRIBE event":
```
diegomrsantos, line 40: This test is very similar to https://github.com/vacp2p/nim-libp2p/blob/68306cf1f172c70cf14df6b31426e4b99140a708/tests/pubsub/testgossipinternal.nim#L36

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,184 @@
...
+      gossipSub.gossipsub[topic].incl(peer)
+        # Ensure peers are added to gossipsub for the topic
+
+    # Subscribe to the topic first
+    gossipSub.PubSub.subscribe(
+      topic,
+      proc(topic: string, data: seq[byte]): Future[void] {.async.} =
+        discard
+      ,
+    )
```
diegomrsantos, line 102: There's some code repetition in these tests

diegomrsantos:
We follow conventional commits for the PR title. In this case, it will start with `test(gossipsub):`

diegomrsantos:
> > how about `testgossiptopicmembership.nim`?
> 
> I am adding other tests in the same file related to Topic Membership Tests for Join topic, leave topic, and mesh updated on peer subscribing and leaving topic.
> 
> Earlier suggestion looks fine but else will modify as per your suggestion

I'm not sure if gossip membership makes sense.

tests/pubsub/testgossipmembership.nim
diegomrsantos: > Uses an anonymous function when subscribing to the topic, and it does not test for the dynamic dispatch mechanism used to handle topic messages.

This doesn't need to be tested in this test.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,169 @@
...
+        ,
+      )
+
+  # Helper function to unsubscribe to topics
+  proc unsubscribeFromTopics(gossipSub: TestGossipSub, topics: seq[string]) =
+    for topic in topics:
+      gossipSub.PubSub.unsubscribeAll(topic)
+
+  # Simulate the `SUBSCRIBE` event and check proper handling in the mesh and gossipsub structures
+  asyncTest "handle SUBSCRIBE event":
```
diegomrsantos, line 82: Not sure if using `event` is a good choice here. It has a technical meaning and I don't think that's what is happening in the spec and in the code.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,169 @@
...
+    let topic = "test-topic"
+    let (gossipSub, conns) = setupGossipSub(topic, 5)
+
+    # Subscribe to the topic (ensure `@[topic]` is passed as a sequence)
+    subscribeToTopics(gossipSub, @[topic].toSeq()) # Pass topic as seq[string]
+
+    check gossipSub.topics.contains(topic) # Check if the topic is in topics
+    check gossipSub.gossipsub[topic].len() > 0 # Check if topic added to gossipsub
+
```
diegomrsantos, line 91: The existing text also checks `gossipSub.mesh[topic].len() > 0`. Is there a place defining what the tests should ensure?

tests/pubsub/testgossipmembership.nim
diegomrsantos: But we could replace the existing test with the first two you created.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,184 @@
...
+    # Check if the topic is present in the list of subscribed topics
+    check gossipSub.topics.contains(topic)
+
+    # Check if the topic is added to gossipsub and the peers list is not empty
+    check gossipSub.gossipsub[topic].len() > 0
+
+    # Iterate over the peers and check connection handling
+    for peer in gossipSub.peers.values:
+      check peer.sendConn != nil
+      check not peer.sendConn.closed()
```
diegomrsantos, line 98: I don't think this is necessary.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,184 @@
...
+    for topic in topics:
+      gossipSub.PubSub.unsubscribeAll(topic)
+
+  # Simulate the `SUBSCRIBE` to the topic and check proper handling in the mesh and gossipsub structures
+  asyncTest "handle SUBSCRIBE to the topic and verify connection handling":
+    let topic = "test-topic"
+    let (gossipSub, conns) = setupGossipSub(topic, 5)
+
+    # Subscribe to the topic
+    subscribeToTopics(gossipSub, @[topic].toSeq()) # Pass topic as seq[string]
```
diegomrsantos, line 87: `.toSeq()` isn't necessary

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,184 @@
...
+
+    # Subscribe to the topic first
+    subscribeToTopics(gossipSub, @[topic]) # Pass topic as seq[string]
+
+    # Now unsubscribe from the topic
+    unsubscribeFromTopics(gossipSub, @[topic]) # Pass topic as seq[string]
+
+    # Verify the topic is removed from relevant structures
+    check topic notin gossipSub.topics # The topic should not be in topics
+    check topic notin gossipSub.mesh # The topic should be removed from the mesh
```
diegomrsantos, line 123: I think those comments aren't necessary as the code is clear.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,184 @@
...
+    for peer in gossipSub.peers.values:
+      check peer.sendConn == nil or peer.sendConn.closed()
+
+  # Simulate an UNSUBSCRIBE to the topic and check if the topic is removed from the relevant data structures but remains in gossipsub
+  asyncTest "handle UNSUBSCRIBE to the topic":
+    let topic = "test-topic"
+    let (gossipSub, conns) = setupGossipSub(topic, 5)
+
+    # Subscribe to the topic first
+    subscribeToTopics(gossipSub, @[topic]) # Pass topic as seq[string]
```
diegomrsantos, line 116: Comment here also isn't necessary.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,184 @@
...
+
+    # Now unsubscribe from the topic
+    unsubscribeFromTopics(gossipSub, @[topic]) # Pass topic as seq[string]
```
diegomrsantos, line 119: Comment here also isn't necessary.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,184 @@
...
+        ,
+      )
+
+  # Helper function to unsubscribe to topics
+  proc unsubscribeFromTopics(gossipSub: TestGossipSub, topics: seq[string]) =
+    for topic in topics:
+      gossipSub.PubSub.unsubscribeAll(topic)
+
+  # Simulate the `SUBSCRIBE` to the topic and check proper handling in the mesh and gossipsub structures
+  asyncTest "handle SUBSCRIBE to the topic and verify connection handling":
```
diegomrsantos, line 82: "handle SUBSCRIBE to the topic" is enough for the name.

tests/pubsub/testgossipmembership.nim
diegomrsantos: Please remove `asyncTest "subscribe/unsubscribeAll":`i n testgossipinternal.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,180 @@
...
+    let (gossipSub, conns) = setupGossipSub(topic, 5)
+
+    # Subscribe to the topic
+    subscribeToTopics(gossipSub, @[topic])
+
+    # Check if the topic is present in the list of subscribed topics
+    check gossipSub.topics.contains(topic)
+
+    # Check if the topic is added to gossipsub and the peers list is not empty
+    check gossipSub.gossipsub[topic].len() > 0
```
diegomrsantos, line 93: If we move this line before `subscribeToTopics(gossipSub, @[topic]) ` it's already true, so we aren't testing anything here.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,176 @@
...
```
AYAHASSAN287, line 93: since the numbers of peers is set to 5 so to ensure the length is set correctly  we could change to 
check gossipSub.gossipsub[topic].len() == 5

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,176 @@
...
+
+    var conns = newSeq[Connection]()
+    for i in 0 .. gossipSub.topicsHigh + 5:
+      let topic = "topic" & $i
+      # Ensure all topics are properly initialized before subscribing
+      gossipSub.mesh[topic] = initHashSet[PubSubPeer]()
+      gossipSub.topicParams[topic] = TopicParams.init()
+      gossipSub.gossipsub[topic] = initHashSet[PubSubPeer]()
+
+      if gossipSub.topics.len < gossipSub.topicsHigh:
```
AYAHASSAN287, line 160: since we want to test the reaction of the logic if we subscribed to more than the limit then we need to try to add more subscribtions than the limit and check the reaction

tests/pubsub/testgossipmembership.nim
diegomrsantos: what was exactly done?

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,176 @@
...
+        ,
+      )
+
+  # Helper function to unsubscribe to topics
+  proc unsubscribeFromTopics(gossipSub: TestGossipSub, topics: seq[string]) =
+    for topic in topics:
+      gossipSub.PubSub.unsubscribeAll(topic)
+
+  # Simulate the `SUBSCRIBE` to the topic and check proper handling in the mesh and gossipsub structures
+  asyncTest "handle SUBSCRIBE to the topic":
```
diegomrsantos, line 82: This test is still not checking changes to the mesh.

diegomrsantos (changes requested):
There are things to be addressed

tests/pubsub/testgossipmembership.nim
diegomrsantos: Im not sure this separation is necessary tbh.

tests/pubsub/testgossipmembership.nim
```diff
@@ -146,17 +146,16 @@ suite "GossipSub Topic Membership Tests":
...
-      gossipSub.mesh[topic] = initHashSet[PubSubPeer]()
-      gossipSub.topicParams[topic] = TopicParams.init()
-      gossipSub.gossipsub[topic] = initHashSet[PubSubPeer]()
+    # Use setupGossipSub to initialize the GossipSub system with connections
+    let (gossipSub, conns) = setupGossipSub(topicNames, 0) # No peers for now
+
+    gossipSub.topicsHigh = gossipSubParams # Set the subscription limit
 
+    for topic in topicNames:
       if gossipSub.topics.len < gossipSub.topicsHigh:
```
AlejandroCabeza, line 159: Now that I think about this, this doesn't make a lot of sense. The limit shouldn't be handled by the test, not manually by us. This is what the test should be asserting, actually: That regardless of the amount of subscriptions done, it shouldn't exceed the limit.

P.S.: Please, remove all the `PubSub` usages you added as these are `GossipSub` cases.

tests/pubsub/testgossipmembership.nim
```diff
@@ -214,39 +237,76 @@ suite "GossipSub Topic Membership Tests":
...
+      doAssert currentGossip.gossipsub.hasKey(topic),
+        "Node is not subscribed to the topic"
+
+    # Print mesh status before connecting nodes
+    echo "Initial mesh size:"
+    for i in 0 ..< numberOfNodes:
+      let currentGossip = GossipSub(nodes[i]) # Rename to 'currentGossip'
+      echo "Node ", i, " mesh size: ", currentGossip.mesh.getOrDefault(topic).len
+
+    # Connect all nodes to each other and ensure subscription propagation
```
AlejandroCabeza, line 266: This is already achieved by the `await subscribedNodes(nodes)` run before.

tests/pubsub/testgossipmembership.nim
```diff
@@ -214,39 +237,76 @@ suite "GossipSub Topic Membership Tests":
...
+    for x in 0 ..< numberOfNodes:
+      for y in 0 ..< numberOfNodes:
+        if x != y:
+          await waitSub(nodes[x], nodes[y], topic)
+
+    # Allow time for mesh stabilization
+    await sleepAsync(2 * DURATION_TIMEOUT)
+
+    # Print mesh status after stabilization
```
AlejandroCabeza, line 275: Same here regarding printing.

tests/pubsub/testgossipmembership.nim
```diff
@@ -214,39 +237,76 @@ suite "GossipSub Topic Membership Tests":
...
+      check:
+        currentGossip.gossipsub[topic].len == expectedNumberOfPeers
+        currentGossip.mesh[topic].len == expectedNumberOfPeers
+        currentGossip.fanout.len == 0
+
+    # Simulate unsubscription of 3 peers
+    let firstNodeGossip = GossipSub(nodes[0]) # Initialize for the first node
+    let peersToUnsubscribe = firstNodeGossip.mesh[topic].toSeq()[0 .. 2]
     for peer in peersToUnsubscribe:
+      firstNodeGossip.unsubscribe(topic, voidTopicHandler)
```
AlejandroCabeza, line 294: You are using `peer` as the iteration variable but the unsubscribed node is `firstNodeGossip`. 

The test should fail due to this and it doesn't in the CI/CD. That being said, I ran it locally and I get a failure for it. Can you double check it?

AlejandroCabeza (changes requested):

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,284 @@
...
+
+    # Allow time for heartbeat to adjust the mesh
+    await sleepAsync(3 * DURATION_TIMEOUT)
+
+    # Check the mesh size for the first node after unsubscription
+    let firstNodeGossip = nodes[0]
+    echo "Mesh size after unsubscription: ",
+      firstNodeGossip.mesh.getOrDefault(topic).len
+    doAssert firstNodeGossip.mesh.getOrDefault(topic).len == 3,
+      "Expected 3 peers to remain in the mesh"
```
AlejandroCabeza, line 276: Just as you did before, I'd check the `gossipsub` and `mesh` (and `fanout`, if you want) are in the expected state, not just the `mesh`. As those two (`gossipsub` and `mesh`) are directly involved in the subscription/unsubscription you're testing).

AlejandroCabeza (changes requested):

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,246 @@
...
+        currentGossip.mesh[topic].len == expectedNumberOfPeers
+        currentGossip.fanout.len == 0
+
+    let firstNodeGossip = GossipSub(nodes[0])
+    let peersToUnsubscribe = firstNodeGossip.mesh[topic].toSeq()[0 .. 2]
+    let peerIdsToUnsubscribe = peersToUnsubscribe.mapIt(it.peerId)
+    for node in nodes:
+      for peerId in peerIdsToUnsubscribe:
+        node.unsubscribe(topic, voidTopicHandler)
+        node.peers.del(peerId)
```
AlejandroCabeza, line 237: You can't call `unsubscribe` from a `PubSubPeer`:
- `PubSubPeer` is a node's representation of each of its peers. Because it's just a representation/handler it can't call methods such as `unsubscribe` on the node's behalf.
- `unsubscribe` can only be called on `PubSub` instances, or its children. E.g.: `FloodSub`, `GossipSub`.

Due to reasons above, to `unsubscribe` a node you need to use the node itself:
```suggestion
    let 
      firstNodeGossip = GossipSub(nodes[0])
      peersToUnsubscribe = nodes[1 ..< 3]
    for peer in peersToUnsubscribe:
        peer.unsubscribe(topic, voidTopicHandler)
```

AlejandroCabeza (changes requested):

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,240 @@
...
+let DURATION_TIMEOUT = 500.milliseconds
+
+suite "GossipSub Topic Membership Tests":
+  teardown:
+    checkTrackers()
+
+  # Addition of Designed Test cases for 6. Topic Membership Tests: https://www.notion.so/Gossipsub-651e02d4d7894bb2ac1e4edb55f3192d
+
+  # Generalized setup function to initialize one or more topics
+  proc setupGossipSub(
```
AlejandroCabeza, line 44: I'd suggest removing the usages of `setupGossipSub` and following the `generateNodes` for all remaining tests.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,240 @@
...
+      nodes: seq[TestGossipSub],
+      topic: string,
+      handler: proc(topic: string, data: seq[byte]) {.async.},
+  ) =
+    for node in nodes:
+      node.unsubscribe(topic, handler)
+    echo "Unsubscribed all nodes from the topic: ", topic
+
+  # Simulate the `SUBSCRIBE` to the topic and check proper handling in the mesh and gossipsub structures
+  asyncTest "handle SUBSCRIBE to the topic":
```
AlejandroCabeza, line 104: Please, add a `mesh` check as well, so we have the state properly mapped.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,240 @@
...
+  # Simulate an UNSUBSCRIBE to the topic and check if the topic is removed from the relevant data structures but remains in gossipsub
+  asyncTest "handle UNSUBSCRIBE to the topic":
+    let topic = "test-topic"
+    let (gossipSub, conns) = setupGossipSub(topic, 5)
+
+    subscribeToTopics(gossipSub, @[topic])
+
+    unsubscribeFromTopics(gossipSub, @[topic])
+
+    check topic notin gossipSub.topics
```
AlejandroCabeza, line 126: For the cases where the value exists (`topic in gossipSub.gossipsub`), do make the check in a more exhaustive way. That is, checking for length as you did before.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,240 @@
...
+  # Test subscribing and unsubscribing multiple topics
+  asyncTest "handle SUBSCRIBE and UNSUBSCRIBE multiple topics":
+    let topics = ["topic1", "topic2", "topic3"].toSeq()
+    let (gossipSub, conns) = setupGossipSub(topics, 5)
+
+    subscribeToTopics(gossipSub, topics)
+
+    check gossipSub.topics.len == 3
+    for topic in topics:
+      check gossipSub.gossipsub[topic].len() == 5
```
AlejandroCabeza, line 142: Add a check for `mesh` as well, please

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,240 @@
...
+    let topicCount = 15
+    let gossipSubParams = 10
+    let topicNames = toSeq(mapIt(0 .. topicCount - 1, "topic" & $it))
+
+    let (gossipSub, conns) = setupGossipSub(topicNames, 0)
+
+    gossipSub.topicsHigh = gossipSubParams
+
+    for topic in topicNames:
+      if gossipSub.topics.len < gossipSub.topicsHigh:
```
AlejandroCabeza, line 165: You don't need to handle this case. This is something that `subscribe` does internally, and it's precisely what we are trying to verify.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,240 @@
...
+          topic,
+          proc(topic: string, data: seq[byte]): Future[void] {.async.} =
+            discard
+          ,
+        )
+      else:
+        check gossipSub.topics.len == gossipSub.topicsHigh
+
+    check gossipSub.topics.len <= gossipSub.topicsHigh
+    check gossipSub.topics.len == gossipSub.topicsHigh
```
AlejandroCabeza, line 176: You don't need both checks, just one is enough. Please, keep the more exhaustive one.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,240 @@
...
+
+    await allFuturesThrowing(conns.mapIt(it.close()))
+    await gossipSub.switch.stop()
+
+  # Test for verifying peers joining a topic using `JOIN(topic)`
+  asyncTest "handle JOIN topic and mesh is updated":
```
AlejandroCabeza, line 182: What's the difference between this case and the `handle SUBSCRIBE to the topic` one?

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,240 @@
...
+      node.subscribe(topic, voidTopicHandler)
+
+    await sleepAsync(2 * DURATION_TIMEOUT)
+
+    for i in 0 ..< numberOfNodes:
+      let currentGossip = GossipSub(nodes[i])
+      check currentGossip.gossipsub.hasKey(topic)
+
+    for i in 0 ..< numberOfNodes:
+      let currentGossip = GossipSub(nodes[i])
```
AlejandroCabeza, line 214: This is doing nothing. Please, review the code carefully before uploading.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,240 @@
...
+
+    await subscribeNodes(nodes)
+    for node in nodes:
+      node.subscribe(topic, voidTopicHandler)
+
+    await sleepAsync(2 * DURATION_TIMEOUT)
+
+    for i in 0 ..< numberOfNodes:
+      let currentGossip = GossipSub(nodes[i])
+      check currentGossip.gossipsub.hasKey(topic)
```
AlejandroCabeza, line 211: If you're checking the state, which is a great idea, I'd suggest checking it completely. I mean, checking all relevant attributes to this test: `gossipsub`, `mesh`. Maybe `topics` as well, but this last one is not that important.

AlejandroCabeza (changes requested):

tests/pubsub/testgossipmembership.nim
AlejandroCabeza: I think you might be able to condense the two cases into one.

Are they two different cases in the test plans?

AlejandroCabeza (commented):
I think it'd be a good addition to this PR to also follow the [Given-When-Then](https://martinfowler.com/bliki/GivenWhenThen.html) formula used in the rest of the codebase. It documents the case by high-level blocks, making it easier to understand for external eyes.

You can check other PRs and or test files for examples. 
Also, here's an example using one of your implemented cases:
```nim
  asyncTest "handle SUBSCRIBE to the topic":
    # Given 5 gossipsub nodes
    let
      numberOfNodes = 5
      topic = "test-topic"
      nodes = generateNodes(numberOfNodes, gossip = true)

    await allFuturesThrowing(nodes.mapIt(it.switch.start()))

    # When all of them are connected and subscribed to the same topic
    await subscribeNodes(nodes)
    for node in nodes:
      node.subscribe(topic, voidTopicHandler)

    await sleepAsync(2 * DURATION_TIMEOUT)

    # Then their related attributes should reflect that
    for node in nodes:
      let currentGossip = GossipSub(node)

      check currentGossip.topics.contains(topic)
      check currentGossip.gossipsub[topic].len() == numberOfNodes - 1
      check currentGossip.mesh[topic].len() == numberOfNodes - 1

    await allFuturesThrowing(nodes.mapIt(allFutures(it.switch.stop())))
```

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,255 @@
...
+  asyncTest "handle SUBSCRIBE and UNSUBSCRIBE multiple topics":
+    # Given 3 nodes and multiple topics
+    let
+      numberOfNodes = 3
+      topics = ["topic1", "topic2", "topic3"].toSeq()
+      nodes = generateNodes(numberOfNodes, gossip = true)
+
+    await allFuturesThrowing(nodes.mapIt(it.switch.start()))
+
+    # When nodes subscribe and then unsubscribe to multiple topics
```
AlejandroCabeza, line 115: I'd remove the second part here. Leave only: `When nodes subscribe to multiple topics`, as that's what you're doing here, and later you're handling the `unsubscribe` part.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,255 @@
...
+      if gossipSub.topics.len < gossipSub.topicsHigh:
+        gossipSub.subscribe(
+          topic,
+          proc(topic: string, data: seq[byte]): Future[void] {.async.} =
+            discard
+          ,
+        )
+      else:
+        # Then the subscription count should not exceed the limit
+        check gossipSub.topics.len == gossipSub.topicsHigh
```
AlejandroCabeza, line 173: Here you don't need to handle the `if gossipSub.topics.len < gossipSub.topicsHigh:`. That's precisely the situation that we want to test, and that `subscribe` should be handling:
```suggestion
    for topic in topicNames:
      gossipSub.subscribe(
        topic,
        proc(topic: string, data: seq[byte]): Future[void] {.async.} =
          discard
        ,
      )
```

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,255 @@
...
+
+    check gossipSub.topics.len == gossipSub.topicsHigh
```
AlejandroCabeza, line 175: I would check the `gossipsub.topics.len` against `gossipSubParams`, which is the limit you defined here. That would make the check accurate.

Also, maybe it'd be interesting to connect the nodes and check for `mesh` and `gossipsub` population.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,255 @@
...
+  asyncTest "handle JOIN topic and mesh is updated":
+    # Given 5 nodes and a join request to a topic
+    let
+      topic = "test-join-topic"
+      numberOfNodes = 5
+      nodes = generateNodes(numberOfNodes, gossip = true)
+
+    await allFuturesThrowing(nodes.mapIt(it.switch.start()))
+
+    # When nodes join and subscribe to the topic
```
AlejandroCabeza, line 189: Isn't join and subscribe the same thing?

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,255 @@
...
+    await subscribeNodes(nodes)
+    for node in nodes:
+      node.subscribe(topic, voidTopicHandler)
+
+    await sleepAsync(2 * DURATION_TIMEOUT)
+
+    # Then each node's mesh should reflect this update
+    for node in nodes:
+      let currentGossip = GossipSub(node)
+      check currentGossip.mesh[topic].len == numberOfNodes - 1
```
AlejandroCabeza, line 199: Add a gossipsub check as well, for completeness sake :)

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,255 @@
...
+      nodes = generateNodes(numberOfNodes, gossip = true)
+      nodesFut = await allFinished(nodes.mapIt(it.switch.start()))
+
+    await subscribeNodes(nodes)
+    for node in nodes:
+      node.subscribe(topic, voidTopicHandler)
+
+    await sleepAsync(2 * DURATION_TIMEOUT)
+
+    # When they all join the topic, their attributes should reflect this
```
AlejandroCabeza, line 219: The first half of this should be a bit up, because the connection and subscription is done in the lines above.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,255 @@
...
+    for i in 0 ..< numberOfNodes:
+      let currentGossip = GossipSub(nodes[i])
+      check currentGossip.gossipsub.hasKey(topic)
+      check currentGossip.mesh.hasKey(topic)
+      check currentGossip.topics.contains(topic)
+
+    # When peers subscribe to each other's topics
```
AlejandroCabeza, line 226: The subscription is done above, on the `subscribe`. This is just a synchronous check to turn the test deterministic: Make sure all nodes are connected between themselves.

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,254 @@
...
+      node.subscribe(topic, voidTopicHandler)
+
+    await sleepAsync(2 * DURATION_TIMEOUT)
+
+    # Then each node's mesh should reflect this update
+    for node in nodes:
+      let currentGossip = GossipSub(node)
+      check currentGossip.mesh[topic].len == numberOfNodes - 1
+      check currentGossip.gossipsub.hasKey(topic)
+      check currentGossip.topics.contains(topic)
```
AlejandroCabeza, line 198: Please, do these checks as exhaustive as possible :)

tests/pubsub/testgossipmembership.nim
```diff
@@ -0,0 +1,254 @@
...
+      node.subscribe(topic, voidTopicHandler)
+
+    await sleepAsync(2 * DURATION_TIMEOUT)
+
+    # Their attributes should reflect in this loop
+    for i in 0 ..< numberOfNodes:
+      let currentGossip = GossipSub(nodes[i])
+      check currentGossip.gossipsub.hasKey(topic)
+      check currentGossip.mesh.hasKey(topic)
+      check currentGossip.topics.contains(topic)
```
AlejandroCabeza, line 223: Same here (re: exhaustiveness).

PR #1184, opened 2024-08-29
test(gossipsub): message propagation

strinnityk:
Implement Gossipsub's [test plan](https://www.notion.so/Gossipsub-651e02d4d7894bb2ac1e4edb55f3192d)'s Message Propagation block.

Edit:
@rlve: I'm trying to revive this PR. Actions taken:
- rebased with `master`
- changed base branch to `master`
- added changes related to `test(gossipSub)` from https://github.com/vacp2p/nim-libp2p/pull/1168
- reverted changes related to `test(peerDiscovery)` from https://github.com/vacp2p/nim-libp2p/pull/1168
- refactor `generateNodes` in `tests/pubsub/utils.nim`

PR #1205, opened 2024-10-01
test(gossipsub): Part 2 Test cases covering JOIN and LEAVE Events

shashankshampi:
Test Plan: https://www.notion.so/Gossipsub-651e02d4d7894bb2ac1e4edb55f3192d

Test Scenario: Block 6
Topic Membership Tests:

**Part 2:**

1.  **TC6.1**: Verify that peers can correctly join a topic using `JOIN(topic)`.
2. **TC6.2**: Ensure that peers can correctly leave a topic using `LEAVE(topic)`.
3. **TC6.5**:  Multiple peers join and leave topic simultaneously

### DIFF
--- a/tests/pubsub/testgossipmembership.nim
+++ b/tests/pubsub/testgossipmembership.nim
@@ -237,7 +237,7 @@ suite "GossipSub Topic Membership Tests":
     var peersToUnsubscribe = gossipSub.mesh[topic].toSeq()[0 .. 2]
     for peer in peersToUnsubscribe:
       echo "Unsubscribing peer: ", peer.peerId
-      gossipSub.PubSub.unsubscribe(topic, dummyHandler)
+      gossipSub.unsubscribe(topic, dummyHandler)
 
     # Now assert that 6 peers still remain in the mesh because the mesh retains peers
     doAssert gossipSub.mesh[topic].len == 6,


AlejandroCabeza (changes requested):
There's one issue in some of the tests, regarding SUBSCRIBING and UNSUBSCRIBING, that I feel would benefit from some clarification, as it's more complex than it may seem at first.

There's a few key things to understand those operations. For that, let's define `node` as the local instance and `peer` as a remote instance:

When `node` subscribes to a topic it begins listening to the messages that arrive for that topic. In order to do this, `node` will add that topic to `node.topics`. That's a PubSub feature.

When a `peer` connects to `node`, `node` will keeps track of `peer`'s subscribed topics in the `GossipSub.gossipsub` table. If no topics are shared, it will remain as a `metadata` connection.

When both `peer` and `node` are connected and subscribed to the same topic, there will be an attempt to `GRAFT` (`PRUNE` being the opposite operation) them, which means converting the connecting to full-message. Esentially, both GossipSub instances will add an entry for the connected peer into `mesh`. When GossipSub receives a message it will send it to all peers in that topic's mesh, if any.

With all this context:
- I didn't realise this until now (so my bad 😅) but it'd be great to have tests for both situations:
    - Node without connections
    - Node connected with some peer/s

- Some tests are making improper use of `conns`. 
    - In some tests it's declared but not used, or even initialised.
    - It does make sense to have it and use it (when testing the "connected" cases), but in some of those cases you're not actually initialising the variable, or triggering the connection as you may have expected.

AYAHASSAN287 (approved):

tests/pubsub/testgossipmembership.nim
```diff
@@ -175,3 +175,79 @@ suite "GossipSub Topic Membership Tests":
...
+
+    # Define a simple handler for unsubscribing the peers
+    proc dummyHandler(topic: string, data: seq[byte]): Future[void] {.async.} =
+      discard
+
+    # Simulate 3 peers leaving the topic by unsubscribing them
+    var peersToUnsubscribe = gossipSub.mesh[topic].toSeq()[0 .. 2]
+    for peer in peersToUnsubscribe:
+      echo "Unsubscribing peer: ", peer.peerId
+      gossipSub.PubSub.unsubscribe(topic, dummyHandler)
```
AlejandroCabeza, line 241: Same here.

AlejandroCabeza (changes requested):
A couple of importante notes:

1. 
I would recommend not using the subscribe system you're using here. The reason being it's misleading, makes you think stuff works differently than it actually does. That behaves more like a mock of sorts, than the actual expected behaviour.
I may have said that too late because I didn't wanna bother you too much, but I've realised it's quite misleading. I suggest you follow the approach I use in my tests. You can check them out here: https://github.com/vacp2p/nim-libp2p/pull/1184/files, for example in `messages are not sent back to source or forwarding peer`. You might need to check several test cases, as some of them are slightly more complex than others. Checking some of them might be key to get the idea.

2.
I believe you are making some assumptions in how information flows which are not correct. 
E.g.: You are calling `unsubscribe` multiple times on the same node on the same topic on the same unsubscribed-handler, which means nothing's going to be unsubscribed.
I suggest checking the provided docs and/or code and/or requesting a deeper explanation if that's not clear.

3. 
Following from 2. Be careful not to conform tests to the results, but the other way around. If results are not as expected you should double check your test does what it intends. If it does, then that's a signal there might be something wrong in the codebase. But absolutely never adapt the assertions just because.

PR #1206, opened 2024-10-02
fix(CI): free disk space on interop transport job

AlejandroCabeza:
Readd the free disk space job, as space issues still happen when building images.

### DIFF
--- a/.github/workflows/interop.yml
+++ b/.github/workflows/interop.yml
@@ -17,6 +17,12 @@ jobs:
     name: Run transport interoperability tests
     runs-on: ubuntu-22.04
     steps:
+      - name: Free Disk Space
+        # For some reason the original job (libp2p/test-plans) has enough disk space, but this one doesn't.
+        uses: jlumbroso/free-disk-space@v1.3.1
+        with:
+          tool-cache: true
+
       - uses: actions/checkout@v4
       - uses: docker/setup-buildx-action@v3
       - name: Build image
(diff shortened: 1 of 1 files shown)


diegomrsantos:
It'd be good to identify why this problem is happening even using the cache. It didn't happen before.

.github/workflows/interop.yml
```diff
@@ -35,6 +41,12 @@ jobs:
     name: Run hole-punching interoperability tests
     runs-on: ubuntu-22.04
     steps:
+      - name: Free Disk Space
+        # For some reason the original job (libp2p/test-plans) has enough disk space, but this one doesn't.
+        uses: jlumbroso/free-disk-space@v1.3.1
+        with:
+          tool-cache: true
+
```
diegomrsantos, line 49: Is this currently failing?

.github/workflows/interop.yml
```diff
@@ -17,6 +17,12 @@ jobs:
     name: Run transport interoperability tests
     runs-on: ubuntu-22.04
     steps:
+      - name: Free Disk Space
+        # For some reason the original job (libp2p/test-plans) has enough disk space, but this one doesn't.
```
diegomrsantos, line 21: The disk space isn't related to the action on libp2p/test-plans. This comment is misleading, please remove it.

diegomrsantos: We can say this is a workaround for a lack of smaller images for GitHub runners https://github.com/actions/runner-images/issues/10414.

.github/workflows/interop.yml
AlejandroCabeza: Yes, we are getting a not enough space error, as reported by Shashank on discord.

diegomrsantos: If I understood correctly, only the transport test is failing.

.github/workflows/interop.yml
AlejandroCabeza: I find hard to believe that we're the only repo having issues with HDD space if all of us have access to the same kinds of runners, hence my quite generic comment.

I've updated it to be even more generic, removing the reference to `libp2p/test-plans`. If you feel like it doesn't fit I'd appreciate if you could provide one that does and I'll update it.

lchenut (approved):

PR #1215, opened 2024-10-16
fix(ci): nim 2.0 dependency failure

AlejandroCabeza:
Broken due to an update to Nimble 0.16.2 for the `version-2-0` branch. This PR sets the ref to the previous commit. Also, updates the key naming so it fits better.

Nim's commit list: https://github.com/nim-lang/Nim/commits/version-2-0/


AlejandroCabeza:
Reopening. Cause: Github actions is expecting jobs that don't exist anymore.

PR #1226, opened 2024-11-25
Fix PeerStore missing remote endpoints of established connection.

cheatfate:


etan-status (approved):
Not too familiar with this part, but it looks alright and @cheatfate confirmed it was tested and working consistently with LH.

PR #1230, opened 2024-12-17
feat: idontwant on publish

### LINKED ISSUE #1224: Implement IDONTWANT API (or verify that it works)
https://github.com/libp2p/specs/pull/642

richard-ramos:
Closes #1224 

The following changes were done:
1. Once a message is broadcasted, if it exceeds the threeshold to consider it a large message it will send an IDONTWANT message before publishing the message
https://github.com/vacp2p/nim-libp2p/blob/4f9c21f6999a58dff0fde2d4f031086bc846c9f7/libp2p/protocols/pubsub/gossipsub.nim#L800-L801
2. Extracts the logic to broadcast an IDONTWANT message and to verify the size of a message into separate functions
3. Adds a template to filter values of a hashset without modifying the original

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -102,6 +102,7 @@ proc init*(
     overheadRateLimit = Opt.none(tuple[bytes: int, interval: Duration]),
     disconnectPeerAboveRateLimit = false,
     maxNumElementsInNonPriorityQueue = DefaultMaxNumElementsInNonPriorityQueue,
+    sendIDontWantOnPublish = false,
 ): GossipSubParams =
   GossipSubParams(
     explicit: true,
@@ -139,6 +140,7 @@ proc init*(
     overheadRateLimit: overheadRateLimit,
     disconnectPeerAboveRateLimit: disconnectPeerAboveRateLimit,
     maxNumElementsInNonPriorityQueue: maxNumElementsInNonPriorityQueue,
+    sendIDontWantOnPublish: sendIDontWantOnPublish,
   )
 
 proc validateParameters*(parameters: GossipSubParams): Result[void, cstring] =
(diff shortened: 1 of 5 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -784,6 +797,9 @@ method publish*(g: GossipSub, topic: string, data: seq[byte]): Future[int] {.asy
 
   g.broadcast(peers, RPCMsg(messages: @[msg]), isHighPriority = true)
 
+  if isLargeMessage(msg, msgId):
+    g.sendIDontWant(msg, msgId, peers)
```
arnetheduck, line 801: ... why should we send IDONTWANT _after_ broadcasting the message? If the other peers receive the message, they already know that I don't want it.

There exists a case where sending it before broadcasting makes sense, though I'm not sure how well it has been studied, ie it needs research: when streaming a large message to a peer, this takes time - in the meantime, the peer might start streaming the same message to _us_ - an IDONTWANT before sending the message tells the other peer not to do that.

arnetheduck:
I think there's some confusion about #1224 here - what that issue addresses is the case where a message exists "out of band" _and_ is published on gossipsub - the aim is to tell my neighbours that I have received the message out of band already, and they should not waste bandwidth on sending me an extra copy.

In this particular case, the message is a "blob" from a blob transaction - most of the time, blobs are in the mempool before they are added in a block so if I have the blob in the mempool, there's no point receiving it over gossipsub also.

richard-ramos:
>  the aim is to tell my neighbours that I have received the message out of band already

Would you say that in this case I should instead provide a way for someone using gossipsub to indicate that they don't want to receive some message? (i.e. something like     `pubsub.IDontWant(someMessageId)`)?

If so, do you think it makes sense to also keep the changes i'm proposing (and changing the order so the IDONTWANT is sent *before* the broadcast?

richard-ramos:
Other implementations do seem to send it before it's published unless i understood wrong:
- sigp/lighthouse/pull/6513/
- libp2p/jvm-libp2p/pull/386

arnetheduck:
> Would you say that in this case I should instead provide a way for someone using gossipsub to indicate that they don't want to receive some message? (i.e. something like pubsub.IDontWant(someMessageId))?

I think a function like that is broadly useful, yes - though the docs should make clear that it's a "hint".

> If so, do you think it makes sense to also keep the changes i'm proposing (and changing the order so the IDONTWANT is sent before the broadcast? or should that be researched first?

they certainly have to be reordered - the policy for _when_ to send a pre-emptive idontwant is however not clear to me - the only time it is useful is during the aforementioned case where the other peer makes the decision to start sending the message _while_ we are also sending it .. it happens, but ... only when streaming time is large enough - all other times, it's a waste of time and bandwidth, and I'm not sure the threshold is well understood. Perhaps this has been answered already _somewhere_ but doing a bit of digging might be the right course of action here.

richard-ramos:
@arnetheduck, I changed the PR now to make publishing an IDONTWANT on broadcast to be optional depending on whether you set a gossipsub parameter to true or no. By default it will not, making this feature opt-in.

PR #1233, opened 2025-01-09
feat(pubsub): add `{.async: (raises).}` annotations

richard-ramos:
This PR adds `{.async: (raises).}` annotations to the pubsub package.
The cases in which a `raises:[CatchableError]` was added were due to not being part of the package and should probably be changed in a separate PR

### DIFF
--- a/examples/tutorial_4_gossipsub.nim
+++ b/examples/tutorial_4_gossipsub.nim
@@ -95,11 +95,8 @@ proc oneNode(node: Node, rng: ref HmacDrbgContext) {.async.} =
     node.gossip.subscribe(
       "metrics",
       proc(topic: string, data: seq[byte]) {.async: (raises: []).} =
-        let m = MetricList.decode(data)
-        if m.isErr:
-          raiseAssert "failed to decode metric"
-        else:
-          echo m
+        let m = MetricList.decode(data).expect("metric can be decoded")
+        echo m
       ,
     )
   else:
--- a/examples/tutorial_4_gossipsub.nim
+++ b/examples/tutorial_4_gossipsub.nim
@@ -75,7 +75,9 @@ proc oneNode(node: Node, rng: ref HmacDrbgContext) {.async.} =
   # This procedure will handle one of the node of the network
   node.gossip.addValidator(
     ["metrics"],
-    proc(topic: string, message: Message): Future[ValidationResult] {.async.} =
+    proc(
+        topic: string, message: Message
+    ): Future[ValidationResult] {.async: (raises: []).} =
       let decoded = MetricList.decode(message.data)
       if decoded.isErr:
         return ValidationResult.Reject
@@ -92,8 +94,12 @@ proc oneNode(node: Node, rng: ref HmacDrbgContext) {.async.} =
   if node.hostname == "John":
     node.gossip.subscribe(
       "metrics",
-      proc(topic: string, data: seq[byte]) {.async.} =
-        echo MetricList.decode(data).tryGet()
+      proc(topic: string, data: seq[byte]) {.async: (raises: []).} =
+        let m = MetricList.decode(data)
+        if m.isErr:
+          raiseAssert "failed to decode metric"
+        else:
+          echo m
       ,
     )
   else:
--- a/libp2p/daemon/daemonapi.nim
+++ b/libp2p/daemon/daemonapi.nim
@@ -162,7 +162,7 @@ type
   .}
   P2PPubSubCallback* = proc(
     api: DaemonAPI, ticket: PubsubTicket, message: PubSubMessage
-  ): Future[bool] {.gcsafe, raises: [CatchableError].}
+  ): Future[bool] {.gcsafe, async: (raises: [CatchableError]).}
 
   DaemonError* = object of LPError
   DaemonRemoteError* = object of DaemonError
@@ -1296,7 +1296,7 @@ proc pubsubLoop(api: DaemonAPI, ticket: PubsubTicket) {.async.} =
 
 proc pubsubSubscribe*(
     api: DaemonAPI, topic: string, handler: P2PPubSubCallback
-): Future[PubsubTicket] {.async.} =
+): Future[PubsubTicket] {.async: (raises: [CatchableError]).} =
   ## Subscribe to topic ``topic``.
   var transp = await api.newConnection()
   try:
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -219,8 +219,6 @@ method init*(g: GossipSub) =
       # This is top-level procedure which will work as separate task, so it
       # do not need to propogate CancelledError.
       trace "Unexpected cancellation in gossipsub handler", conn
-    except CatchableError as exc:
-      trace "GossipSub handler leaks an error", description = exc.msg, conn
 
   g.handler = handler
   g.codecs &= GossipSubCodec_12
@@ -490,7 +488,9 @@ proc validateAndRelay(
       )
 
     await handleData(g, topic, msg.data)
-  except CatchableError as exc:
+  except CancelledError as exc:
+    info "validateAndRelay failed", description = exc.msg
+  except PeerRateLimitError as exc:
     info "validateAndRelay failed", description = exc.msg
 
 proc dataAndTopicsIdSize(msgs: seq[Message]): int =
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -385,7 +385,7 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
 
 proc validateAndRelay(
     g: GossipSub, msg: Message, msgId: MessageId, saltedId: SaltedId, peer: PubSubPeer
-) {.async.} =
+) {.async: (raises: []).} =
   try:
     template topic(): string =
       msg.topic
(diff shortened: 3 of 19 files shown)


libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -581,37 +586,46 @@ method removeValidator*(
...
-  case result
+    try:
+      let res = fut.read()
+      if res != ValidationResult.Accept:
+        valResult = res
+        if res == ValidationResult.Reject:
+          break
+    except CatchableError as e:
+      trace "validator for message could not be executed, ignoring", topic = topic, err = e.msg
+      valResult = ValidationResult.Ignore
```
richard-ramos, line 617: To avoid returning a `CatchableError` I capture the exception here, and ignore the message.

tersec (approved):

tests/pubsub/testgossipsub2.nim
```diff
@@ -65,10 +65,15 @@ suite "GossipSub":
       var handler: TopicHandler
       closureScope:
         var peerName = $dialer.peerInfo.peerId
-        handler = proc(topic: string, data: seq[byte]) {.async, closure.} =
-          if peerName notin seen:
-            seen[peerName] = 0
-          seen[peerName].inc
+        handler = proc(
+            topic: string, data: seq[byte]
+        ) {.async: (raises: []), closure.} =
```
arnetheduck, line 70:
```suggestion
        ) {.async: (raises: []).} =
```

tests/pubsub/testgossipsub2.nim
```diff
@@ -65,10 +65,15 @@ suite "GossipSub":
...
+          try:
+            if peerName notin seen:
+              seen[peerName] = 0
+            seen[peerName].inc
+          except KeyError:
+            raiseAssert "seen checked before"
           info "seen up", count = seen.len
```
arnetheduck, line 77:
```suggestion
          seen.mGetOrPut(peerName, 0).inc()
          info "seen up", count = seen.len
```

should work..

examples/tutorial_4_gossipsub.nim
```diff
@@ -92,8 +94,12 @@ proc oneNode(node: Node, rng: ref HmacDrbgContext) {.async.} =
...
     node.gossip.subscribe(
       "metrics",
-      proc(topic: string, data: seq[byte]) {.async.} =
-        echo MetricList.decode(data).tryGet()
+      proc(topic: string, data: seq[byte]) {.async: (raises: []).} =
+        let m = MetricList.decode(data)
+        if m.isErr:
+          raiseAssert "failed to decode metric"
+        else:
+          echo m
```
arnetheduck, line 102:
```suggestion
        let m = MetricList.decode(data).expect("metric can be decoded")
        echo m
```

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -808,11 +816,13 @@ proc maintainDirectPeer(g: GossipSub, id: PeerId, addrs: seq[MultiAddress]) {.as
...
 
-proc addDirectPeer*(g: GossipSub, id: PeerId, addrs: seq[MultiAddress]) {.async.} =
+proc addDirectPeer*(
+    g: GossipSub, id: PeerId, addrs: seq[MultiAddress]
+) {.async: (raises: [CancelledError]).} =
   g.parameters.directPeers[id] = addrs
   await g.maintainDirectPeer(id, addrs)
 
-proc maintainDirectPeers(g: GossipSub) {.async.} =
+proc maintainDirectPeers(g: GossipSub) {.async: (raises: [CancelledError]).} =
```
arnetheduck, line 825: often when introducing `raises`, it also makes sense to make a pass over callers to functions to ensure that they no longer contain unnecessary try/except - not sure if that's the case, but it's worth investigating. In particular, we strive to avoid `except CatchableError` in code except where it deliberately is the right thing to do ("catch exceptions from any abstraction level").

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -581,37 +594,47 @@ method removeValidator*(
...
     topic = topic, registered = toSeq(p.validators.keys)
   if topic in p.validators:
     trace "running validators for topic", topic = topic
-    for validator in p.validators[topic]:
-      pending.add(validator(topic, message))
+    try:
+      for validator in p.validators[topic]:
+        pending.add(validator(topic, message))
+    except KeyError:
+      raiseAssert "topic checked before"
```
arnetheduck, line 609: the alternate form that can be used is this:

```suggestion
    p.validators.withValue(topic, validators):
      for validator in validators[][topic]:
        pending.add(validator(topic, message))
```

This avoids having to do two lookups (one for `in` and one for the actual lookup - it also avoids the extra except KeyError. 

It only works when you have `var` access to the table however - here, it should work afair since `PubSub` is `ref`.

libp2p/daemon/daemonapi.nim
```diff
@@ -162,7 +162,7 @@ type
   .}
   P2PPubSubCallback* = proc(
     api: DaemonAPI, ticket: PubsubTicket, message: PubSubMessage
-  ): Future[bool] {.gcsafe, raises: [CatchableError].}
+  ): Future[bool] {.gcsafe, async: (raises: [CatchableError]).}
```
arnetheduck, line 165: btw, to maintain "callback compatibility" in public api, what we do is to introduce a `2` overload, like so: https://github.com/status-im/nim-chronos/blob/7c5cbf04a6bb2258654f16cb1d51b4304986cfd1/chronos/transports/stream.nim#L121

then the overload using the old callback includes a "translation wrapper" and/or is deprecated

tests/pubsub/testgossipsub.nim
```diff
@@ -771,10 +774,15 @@ suite "GossipSub":
       var handler: TopicHandler
       capture dialer, i:
         var peerName = $dialer.peerInfo.peerId
-        handler = proc(topic: string, data: seq[byte]) {.async, closure.} =
-          if peerName notin seen:
-            seen[peerName] = 0
-          seen[peerName].inc
+        handler = proc(
+            topic: string, data: seq[byte]
+        ) {.async: (raises: []), closure.} =
```
arnetheduck, line 779: `closure` is a leftover from old nim versions, should not be needed any more

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -271,7 +276,7 @@ proc connectOnce(p: PubSubPeer): Future[void] {.async.} =
   finally:
     await p.closeSendConn(PubSubPeerEventKind.StreamClosed)
 
-proc connectImpl(p: PubSubPeer) {.async.} =
+proc connectImpl(p: PubSubPeer) {.async: (raises: []).} =
```
arnetheduck, line 279: this _looks like_ it should propagate cancellederror  - "most" api should do this to support cancellation - is there a specific reason here not to?

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -79,7 +81,7 @@ type
   PubSubPeerEvent* = object
     kind*: PubSubPeerEventKind
 
-  GetConn* = proc(): Future[Connection] {.gcsafe, raises: [].}
+  GetConn* = proc(): Future[Connection] {.gcsafe, async: (raises: [GetConnDialError]).}
```
arnetheduck, line 84: `gcsafe` is not needed either, with `async`

for some callbacks, it's worth considering allowing `CancelledError` - again, depending on the context, I don't remember this one in particular, but "cancelling a connection attempt" seems reasonable logically at least

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -542,7 +555,7 @@ proc subscribe*(p: PubSub, topic: string, handler: TopicHandler) {.public.} =
 
 method publish*(
     p: PubSub, topic: string, data: seq[byte]
-): Future[int] {.base, async, public.} =
+): Future[int] {.base, async: (raises: [LPError]), public.} =
```
arnetheduck, line 558: a function _like_ publish should probably not be raising - ie it's a "fire-and-forget" kind of function that attempts to publish things to as many peers as possible but just like udp, there are no guarantees that it will reach anybody - as such, it doesn't _really_ make sense that it raises, *I think*.

libp2p/protocols/pubsub/pubsubpeer.nim
richard-ramos: Good question. The prev version had an `except CatchableError` that captures the `CancelledError` and also interestingly enoug a comment that says `never cancelled` https://github.com/vacp2p/nim-libp2p/blob/1fa30f07e81e9ef415a83c5d3846e86e99a7857f/libp2p/protocols/pubsub/pubsubpeer.nim#L285-L286 

Looking at the history, you introduced this change in https://github.com/vacp2p/nim-libp2p/commit/70deac9e0d16f7d00a0d4404ed191042dbf079be#diff-6f99d21f9b0d636313f9633209e5b125903440de8a134b40821e70c9b9ae4591 as it previously raised the `CancelledError`. Do you remember why?
image

libp2p/daemon/daemonapi.nim
richard-ramos: Thanks for the suggestion! Will implement this in a subsequent PR

libp2p/protocols/pubsub/pubsub.nim
richard-ramos: `publish` can raise a `LPError` if `anonimize` is `false` and we enter this branch of code:
https://github.com/vacp2p/nim-libp2p/blob/c5aa3736f96e4d66f6aa653a2351ded74b7d21a9/libp2p/protocols/pubsub/rpc/message.nim#L85-L86

This can happen if passing a nil `peerInfo` when calling `newSwitch`, (although if `SwitchBuilder` is used, it will avoid that scenario, since `build` will populate the `peerInfo` anyway)

PR #1234, opened 2025-01-10
Large message handling combined

ufarooqstatus:
We use message preamble to detect incoming messages. This information is used to:
-  Announce the messages we are receiving to full-message mesh using IMReceiving messages.
-  Make IWANT requests only if we are not already receiving the same message
-  Limit outstanding IWANT requests for a message to 1
-  Stagger (only add small delay) mesh transmissions

PoC implementation for shadow simulation. Not intended for merging
Reduces duplicates to less than 1.5 for large messages

[More details here](https://github.com/libp2p/specs/pull/654)

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -36,6 +36,25 @@ import ./gossipsub/[types, scoring, behavior], ../../utils/heartbeat
 
 export types, scoring, behavior, pubsub
 
+import std/atomics
+var 
+  lma_idontwant_saves: Atomic[uint32]
+  lma_duplicate_count: Atomic[uint32]
+  lma_dup_during_validation: Atomic[uint32]
+  lma_iwants_sent: Atomic[uint32]
+  lma_iwants_replied: Atomic[uint32] 
+  lma_unique_receives: Atomic[uint32]
+
+lma_idontwant_saves.store(0)
+lma_duplicate_count.store(0)
+lma_dup_during_validation.store(0)
+lma_iwants_sent.store(0)
+lma_iwants_replied.store(0)
+lma_unique_receives.store(0)
+
+export lma_idontwant_saves, lma_duplicate_count, lma_dup_during_validation, 
+        lma_iwants_sent, lma_iwants_replied, lma_unique_receives
+
 logScope:
   topics = "libp2p gossipsub"
 
@@ -360,6 +379,7 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
   if isPruneNotEmpty or isIWantNotEmpty:
     if isIWantNotEmpty:
       libp2p_pubsub_broadcast_iwant.inc(respControl.iwant.len.int64)
+      lma_iwants_sent.atomicInc(respControl.iwant.len.uint32)
 
     if isPruneNotEmpty:
       for prune in respControl.prune:
@@ -381,6 +401,7 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
 
     # iwant replies have lower priority
     trace "sending iwant reply messages", peer
+    lma_iwants_replied.atomicInc(messages.len.uint32)
     g.send(peer, RPCMsg(messages: messages), isHighPriority = false)
 
 proc validateAndRelay(
@@ -428,6 +449,7 @@ proc validateAndRelay(
     var seenPeers: HashSet[PubSubPeer]
     discard g.validationSeen.pop(saltedId, seenPeers)
     libp2p_gossipsub_duplicate_during_validation.inc(seenPeers.len.int64)
+    lma_dup_during_validation.atomicInc()
     libp2p_gossipsub_saved_bytes.inc(
       (msg.data.len * seenPeers.len).int64, labelValues = ["validation_duplicate"]
     )
@@ -470,6 +492,7 @@ proc validateAndRelay(
           libp2p_gossipsub_saved_bytes.inc(
             msg.data.len.int64, labelValues = ["idontwant"]
           )
+          lma_idontwant_saves.atomicInc()
           return true
       return false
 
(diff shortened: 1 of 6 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -782,7 +873,25 @@ method publish*(g: GossipSub, topic: string, data: seq[byte]): Future[int] {.asy
...
+      ))), isHighPriority = true)
+
+    await sem.acquire()
+    defer: sem.release()
+
+    g.broadcast(@[p], RPCMsg(messages: @[msg]), isHighPriority = false)
+
+  for p in staggerPeers:
+    asyncSpawn sendToOne(p)
+
```
richard-ramos, line 894: Using the AsyncSemaphore(1) means that the preambles are being sent sequentially to each peer, right? Why is this the case? Shouldn't we try to get the preamble messages being sent concurrently?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -782,7 +873,25 @@ method publish*(g: GossipSub, topic: string, data: seq[byte]): Future[int] {.asy
...
+  #We send message immediately after sending preamble to each peer
+  proc sendToOne(p: PubSubPeer) {.async.} =
+    g.broadcast(@[p], RPCMsg(control: some(ControlMessage(
+        preamble: @[ControlIHave(topicID: topic, messageIDs: @[msgId])]
+      ))), isHighPriority = true)
+
+    await sem.acquire()
+    defer: sem.release()
+
+    g.broadcast(@[p], RPCMsg(messages: @[msg]), isHighPriority = false)
```
richard-ramos, line 890: No `isHighPriority = true`?

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -309,12 +326,49 @@ proc handleIDontWant*(g: GossipSub, peer: PubSubPeer, iDontWants: seq[ControlIWa
...
+      peer.heIsReceivings[^1].incl(messageId)
 
 proc handleIWant*(
     g: GossipSub, peer: PubSubPeer, iwants: seq[ControlIWant]
-): seq[Message] =
+): tuple[messages: seq[Message], ids: seq[MessageId]] =
   var
-    messages: seq[Message]
+    #ids: seq[MessageId]
+    #messages: seq[Message]
```
richard-ramos, line 371: :scissors: ?
```suggestion
    #ids: seq[MessageId]
    #messages: seq[Message]
```

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -329,14 +383,15 @@ proc handleIWant*(
...
+            return result
           continue
         let msg = g.mcache.get(mid).valueOr:
           libp2p_gossipsub_received_iwants.inc(1, labelValues = ["unknown"])
           continue
         libp2p_gossipsub_received_iwants.inc(1, labelValues = ["correct"])
-        messages.add(msg)
-  return messages
+        result.messages.add(msg)
+        result.ids.add(mid)
```
richard-ramos, line 393: It's probably better to define a `messages` and `ids` variables, instead of using `result`. See https://status-im.github.io/nim-style-guide/language.result.html

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -309,12 +326,49 @@ proc handleIDontWant*(g: GossipSub, peer: PubSubPeer, iDontWants: seq[ControlIWa
...
+        ))), isHighPriority = true)
+
+
+proc handleIMReceiving*(g: GossipSub,
+                      peer: PubSubPeer,
+                      imreceivings: seq[ControlIWant]) =
+  for imreceiving in imreceivings:
+    for messageId in imreceiving.messageIDs:
+      if peer.heIsReceivings[^1].len > 1000: break
+      if messageId.len > 100: continue
```
richard-ramos, line 363: When do messageIds exceed 100? I'm wondering if this condition should be added in other places as i haven't seen length checking for messageIds using `100`. (Probably a good idea to extract this into a constant too)

libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -63,6 +63,8 @@ type
     graft*: seq[ControlGraft]
     prune*: seq[ControlPrune]
     idontwant*: seq[ControlIWant]
+    preamble*: seq[ControlIHave]
+    imreceiving*: seq[ControlIWant]
```
richard-ramos, line 67: Shouldn't these have their own type? `ControlPreamble` and `ControlImReceiving`?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -782,7 +873,25 @@ method publish*(g: GossipSub, topic: string, data: seq[byte]): Future[int] {.asy
...
-  g.broadcast(peers, RPCMsg(messages: @[msg]), isHighPriority = true)
+  #g.broadcast(peers, RPCMsg(messages: @[msg]), isHighPriority = true)
+  let sem = newAsyncSemaphore(1)
+  var staggerPeers = toSeq(peers)
+  g.rng.shuffle(staggerPeers)
+
+  #We send message immediately after sending preamble to each peer
+  proc sendToOne(p: PubSubPeer) {.async.} =
+    g.broadcast(@[p], RPCMsg(control: some(ControlMessage(
+        preamble: @[ControlIHave(topicID: topic, messageIDs: @[msgId])]
```
richard-ramos, line 884: Sending the message length is still pending, right?

libp2p/protocols/pubsub/gossipsub.nim
richard-ramos, line 890: Reading in the specs, I see the following:
> Adding a message preamble may increase control overhead for small messages. 
> Therefore, it is preferable to use it only for messages that exceed the preamble_threshold.

The way it is coded right now, it will send the preamble always, right? Is this because this PR is for experiments?

richard-ramos (commented):
Very interesting PR!


Is this implementing too this part of the spec?:
> if the message ID is found in the ongoing_receives list, 
the peer should postpone sending the IWANT request for a defer_interval. 
The defer_interval may be based on the message download time.

PR #1241, opened 2025-02-03
fix(docs): readme

bravesasha:
corrected the link to the license


bravesasha:
> Thank you for your contribution. However, I prefer to avoid updating the documentation unless it's absolutely necessary. The current issue is minor and does not meaningfully affect the reader's comprehension.

I understand you, thank you for your reply

PR #1242, opened 2025-02-05
add isClosedRemotely public bool attribute in lpstream

Ivansete-status:
Also:
* add the gcsafe to getStreams method so that we can invoke it from async procs


Ivansete-status:
To add more context, this was needed to submit this PR: https://github.com/waku-org/nwaku/pull/3281

PR #1244, opened 2025-02-06
refactor(pubsub): do not raise exceptions on publish

richard-ramos:
As commented on https://github.com/vacp2p/nim-libp2p/pull/1243#issuecomment-2639659840 the only scenario in which is it possible to have this exception being raised is due to a misconfig when initializing a message in 
- https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/floodsub.nim#L215-L219
- https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub.nim#L784-L788

Since `Message.init` is used only in these places and it's not expected for a consumer of nim-libp2p to create a `Message` themselves, adding this assertion is safe as it will immediatly indicate that the message is being initialized incorrectly.

### DIFF
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -194,7 +194,7 @@ method init*(f: FloodSub) =
 
 method publish*(
     f: FloodSub, topic: string, data: seq[byte]
-): Future[int] {.async: (raises: [LPError]).} =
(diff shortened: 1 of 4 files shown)


NagyZoltanPeter (approved):
Yes, it's is clear now! Thank you.

PR #1250, opened 2025-02-12
chore(protocol): list raised exceptions in handlers

vladopajic:

### DIFF
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -177,7 +177,7 @@ method rpcHandler*(
   f.updateMetrics(rpcMsg)
 
 method init*(f: FloodSub) =
-  proc handler(conn: Connection, proto: string) {.async.} =
+  proc handler(conn: Connection, proto: string) {.async: (raises: [CancelledError]).} =
     ## main protocol handler that gets triggered on every
     ## connection for a protocol string
     ## e.g. ``/floodsub/1.0.0``, etc...
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -210,7 +210,7 @@ proc validateParameters*(parameters: TopicParams): Result[void, cstring] =
     ok()
 
 method init*(g: GossipSub) =
-  proc handler(conn: Connection, proto: string) {.async.} =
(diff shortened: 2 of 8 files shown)


libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -177,7 +177,7 @@ method rpcHandler*(
   f.updateMetrics(rpcMsg)
 
 method init*(f: FloodSub) =
-  proc handler(conn: Connection, proto: string) {.async.} =
+  proc handler(conn: Connection, proto: string) {.async: (raises: [CancelledError]).} =
     ## main protocol handler that gets triggered on every
     ## connection for a protocol string
     ## e.g. ``/floodsub/1.0.0``, etc...
```
vladopajic, line 183: here `CancelledError` has to be raised because it is handler, but this handler is catching this exception.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -210,7 +210,7 @@ proc validateParameters*(parameters: TopicParams): Result[void, cstring] =
     ok()
 
 method init*(g: GossipSub) =
-  proc handler(conn: Connection, proto: string) {.async.} =
+  proc handler(conn: Connection, proto: string) {.async: (raises: [CancelledError]).} =
     ## main protocol handler that gets triggered on every
     ## connection for a protocol string
     ## e.g. ``/floodsub/1.0.0``, etc...
```
vladopajic, line 216: here `CancelledError` has to be raised because it is handler, but this handler is catching this exception.

richard-ramos:
Looks good so far. CI seems to be complaining tho.

libp2p/protocols/pubsub/floodsub.nim
arnetheduck: it might be a good idea to review whether the handler _should_ catch the cancellation or let it propagate - much of this code was written as cancellation was being developed and it might not always be following best practice .. catching the cancellation and translating it to some other outcome is sometimes the right thing, but often not (it "hides" the fact that a cancellation is ongoing from the calling code which deprives it of the means to make the distinction).

vladopajic:
closing this PR in favor of  #1260

libp2p/protocols/pubsub/floodsub.nim
vladopajic: hey @arnetheduck, yes, it makes much more sense to catch cancellation. this effort is in #1260

PR #1251, opened 2025-02-12
chore: list raised exceptions in `switch` services

vladopajic:
this PR adds raising exceptions to switch service methods `setup`, `run` and `stop`.


libp2p/services/autorelayservice.nim
```diff
@@ -115,19 +122,23 @@ proc innerRun(self: AutoRelayService, switch: Switch) {.async.} =
         break
       self.relayPeers[relayPid] = self.reserveAndUpdate(relayPid, switch)
 
-    if self.relayPeers.len() > 0:
+    try:
```
richard-ramos, line 125: I'm not sure we should replace this condition by a try/except, but instead add to it with something like

```nim
if self.relayPeers.len() > 0:
   try:
       await one(toSeq(self.relayPeers.values())) or self.peerAvailable.wait()
   except ValueError:
       raiseAssert "checked with relayPeers.len()" 
else:
   await self.peerAvailable.wait()
```

This is because raising an exception is expensive compared to checking the condition and acting based on the condition (since `relayPeers.len() is != 0`, then `one(` will not raise the exception).

libp2p/services/hpservice.nim
```diff
@@ -89,13 +89,18 @@ proc newConnectedPeerHandler(
...
   hasBeenSetup = hasBeenSetup and await self.autonatService.setup(switch)
 
   if hasBeenSetup:
-    let dcutrProto = Dcutr.new(switch)
-    switch.mount(dcutrProto)
+    try:
+      let dcutrProto = Dcutr.new(switch)
+      switch.mount(dcutrProto)
+    except LPError as err:
+      trace "Failed to mount Dcutr", err = err.msg
```
richard-ramos, line 103: Should probably be an error. Trace's are not logged unless you explicitly set that that log level.
```suggestion
      error "Failed to mount Dcutr", err = err.msg
```

Also, does it make sense for the setup to continue Dcutr wasn't mounted? I guess it should be okay since this is used for direct connections when behind a nat :thinking:

vladopajic: well, exceptions raised in `switch.mount` doesn't really help here, 
because `Dcutr.new(switch)` was created here, so we know that protocol is "mountable". 

what we can do is add `raiseAssert` here?

richard-ramos (approved):
LGTM!
cc-ing: @arnetheduck for additional insights about this PR.

PR #1253, opened 2025-02-13
chore: add trace log to analyse remote stream reset

Ivansete-status:
Add more debug logs. Interesting to analyse an issue discovered by @AlbertoSoutullo


richard-ramos:
Shouldn't these log lines be `trace`?  a trace level should be used when need extremely detailed information, usually for diagnosing complex issues or profiling performance, with example of these being: entering and exiting functions/methods (especially low-level operations), detailed execution flow,  or temporary logs added for investigating an issue.

Also, please use conventional commits format for both the PR title and commits (This policy is similar to nwaku)

richard-ramos (approved):

richard-ramos (approved):

PR #1256, opened 2025-02-14
ci: remove commit message check

vladopajic:


.github/workflows/commit_lint.yml
```diff
@@ -1,77 +0,0 @@
...
-      - name: Check commit message
-        id: check_commit_message
-        if: always()
-        run: |
-          set +e
-          
-          base_sha=${{ github.event.pull_request.base.sha }} 
-          head_sha=${{ github.event.pull_request.head.sha }}
-
-          output=$(.github/scripts/commit_check.sh "${base_sha}" "${head_sha}" 2>&1)
```
richard-ramos, line 32: Probably makes sense to also remove .github/scripts/commit_check.sh, since it's not going to be used anymore after this change.

richard-ramos (approved):

PR #1260, opened 2025-02-18
wip: chore(protocol): list raised exceptions

vladopajic:


richard-ramos (approved):
Good work!

richard-ramos (approved):

examples/circuitrelay.nim
```diff
@@ -26,15 +26,18 @@ proc main() {.async.} =
...
+    try:
+      var msg = string.fromBytes(await conn.readLp(1024))
+      echo "1 - Dst Received: ", msg
+      assert "test1" == msg
+      await conn.writeLp("test2")
+      msg = string.fromBytes(await conn.readLp(1024))
+      echo "2 - Dst Received: ", msg
+      assert "test3" == msg
+      await conn.writeLp("test4")
+    except:
```
arnetheduck, line 39: we avoid `except:` as it hides defect and is undefined behavior - always `CatchableError` or a more specific exception

libp2p/protocols/perf/server.nim
```diff
@@ -24,7 +24,7 @@ type Perf* = ref object of LPProtocol
 
 proc new*(T: typedesc[Perf]): T {.public.} =
   var p = T()
-  proc handle(conn: Connection, proto: string) {.async.} =
+  proc handle(conn: Connection, proto: string) {.async: (raises: []).} =
```
arnetheduck, line 27: I'm not convinced having the handlers swallow cancellations is a good idea - when disconnecting a peer, the owner of the future should cancel any handlers and that should propagate all the way. What's the reasoning here?

libp2p/protocols/rendezvous.nim
```diff
@@ -734,8 +734,8 @@ proc new*(
         await rdv.discover(conn, msg.discover.tryGet())
       of MessageType.DiscoverResponse:
         trace "Got an unexpected Discover Response", response = msg.discoverResponse
-    except CancelledError as exc:
-      raise exc
+    except CancelledError:
+      trace "cancelled rendezvous handler"
     except CatchableError as exc:
```
arnetheduck, line 739: after specifying raises, it's good to check if these broad handlers can be removed - removing broad exception handlers or commenting why they are broad is one way to ensure that the conversion is complete

examples/circuitrelay.nim
vladopajic: `except:` and `except CatchableError as e:` should be the same thing, right? 

i have used `except:` when i don't need exception variable `e`, for example: 
- in tests, if exceptions happens, test should fail no matter what
- in examples, if exceptions happens, should print to stdout (it is way for handling errors in examples)

libp2p/protocols/rendezvous.nim
vladopajic: i am doing that, but not at every place. this task has already exceed planned time for it. sometimes going in and taking care of those details, requires more time to understand logic and implement correct behavior. not to mention that sometime behavior is not specified for differed cases nor it is covered with tests.

i will suggest to have another round of refactoring effort where `except CatchableError` and raising `CatchableError` should be specified with specific error or commented with explanation why `CatchableError` is used.

arnetheduck: > i will suggest to have another round of refactoring

in that case, one thing _not_ to change is which exceptions get caught and which are raised - ie in cases like this, leave the CancelledError to propagate - this should be done before proceeding any further, ie otherwise there is risk to introduc subtle issues.

examples/circuitrelay.nim
arnetheduck: > should be the same thing, right?

arnetheduck: `except:` == `except Exception:` which is different - it also catches `Defect` _sometimes_ which leads to problems.

arnetheduck: `except CatchableError:` is the canonical way to catch everything, when you don't need `e` (rather unfortunate, but a backwards-compatibility issue in nim)

libp2p/protocols/perf/server.nim
vladopajic: owner of `handlers` will not know how to handle exceptions from `handlers` (should they raise). and it's up to `handlers` to handle them as best as possible. this was the reason of not propagating exceptions.

cancellation: owner of `handle` futures can cancel them, and `handler` will gracefully handle cancellation of it's logic inside. futures will finish (as early as possible) giving control to owner.

arnetheduck: it's fine not to raise other errors, but propagating cancellations is _probably_ the better thing to do in this case - when you are the owner and perform a cancellation, the exception gets handled automatically by chronos at the right abstraction layer.

The rare exception in which we don't propagate cancellations is for "top-level" tasks that are meant to be spawned (a bit like the main procedure of a thread).

libp2p/protocols/rendezvous.nim
vladopajic: got it! 

i was not thinking about `CancelledError` here (this discussion is in comment above). i wanted to mention that there are places where `CatchableError` is used but it might have been more suitable for `ValueError` (narrowing down exceptions).

libp2p/protocols/perf/server.nim
vladopajic: i see handlers as "top-level" tasks as well... 

okay, i realize point you are tiring to make. let me investigate bit into this matter.

libp2p/protocols/rendezvous.nim
arnetheduck: >  more suitable for ValueError (narrowing down exceptions).


Yeah, as a general approach, we _tend to_ remap random exceptions like this to something appropriate for the abstraction, when working with abstractions like libp2p tends to do - this is above all important in order not to leak implementation details into the "exception ABI", ie the set of exceptions that the consumer of the library is expected to handle - changing raised exceptions to something outside of the given raises hierachy is an ABI break which (in theory) requires a new major release.

libp2p/protocols/perf/server.nim
arnetheduck: yeah, handlers tend to be somewhere in-between top-level and not top-level - I'm not too fond of the handlers design in general because they "invert" the async abstraction a bit.. but you can imagine a web server where each request is a "handler", in which case you have something like
```
proc myserver {.async ....}
  var handlers: seq[Handler]
  try:
    while true: 
      handlers.add makehandler(await nextrequest())
  except CancelledError as e:
    await nocancel allFutures(handlers.mapIt(it.cancelAndWait())
    raise e
```

here, the handlers are not _really_ top-level the way you imagine them - the server is and you might have multiple servers which now become top-level.

It's not _wrong_ per se to have the handler catch cancellation and report "successful completion" to its owner, but it's also the more "unusual" choice, and this is maybe the key reason we let the cancellederror propagate - it makes for a recogniseable pattern.

Notably, in the above case we de facto ignore the difference between completed and cancelled handlers because we never actually _read_ the handler outcome - we just await long enough to know that there was _some_ form of completion.

vladopajic: > It's not wrong per se to have the handler catch cancellation and report "successful completion" to its owner, 

yes :+1:

> it makes for a recogniseable pattern

yes, handling cancellation the same way across code-base simplifies making decision what to do. if in some places you propagate, but in another you dont then it's harder for devs to make correct decision. in any case it is required to have firm principles for handling this.

> it's also the more "unusual" choice

it only meters if caller cares about outcome to make next decision. if next decision doesn't depend, then handler (in this case) can swallow cancelled exception and it will not have any impact (as we agree based on first statement here).

vladopajic: anyway i will proceed with raising `CancelledError`, everywhere. real top level handlers can always handle this exception and _not do anything_ or just log it.

PR #1261, opened 2025-02-18
fix: only send IDONT to mesh peers

richard-ramos:
According to the [specs](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md), IDONTWANT messages should be sent to mesh peers, and does not specify that  floodsub peers and direct peers should also receive this type of message.

cc: @AkshayaMani

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -424,18 +424,9 @@ proc validateAndRelay(
     template topic(): string =
       msg.topic
 
-    proc addToSendPeers(toSendPeers: var HashSet[PubSubPeer]) =
-      g.floodsub.withValue(topic, peers):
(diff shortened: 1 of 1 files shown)


vladopajic (approved):
lgtm

arnetheduck:
hm, interesting conundrum .. if _they_ have us as a direct peer, we want to send IDONTWANT to them to avoid getting a redundant copy - I don't remember if we announce the direct-ness to the other party the same way we understand their participation in the mesh.

Ditto fanout, could we somehow anticipate if they are about to send us a message as part of their fanout? I guess not, but ...

gmelodie (approved):

richard-ramos:
@kaiserd @ufarooqstatus.
I'm closing this PR in the meantime since it's stale. 
Please let me know if it should be reopened.

PR #1265, opened 2025-02-20
feat: quic

richard-ramos:
This PR enables quic transport with a proper TLS backend, and tests connection and comms between two nodes. 
Do note that one of the commits changes the `not implemented` assertion to something that gives a bit more context, so it's easier to know which abstract method a dev forgot to implement.

### DIFF
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -26,6 +26,8 @@ type
   QuicTransportError* = object of transport.TransportError
   QuicTransportDialError* = object of transport.TransportDialError
 
+const alpn = "libp2p"
+
 # Stream
 type QuicStream* = ref object of P2PConnection
   stream: Stream
@@ -137,7 +139,7 @@ method handle*(m: QuicMuxer): Future[void] {.async: (raises: []).} =
 method close*(m: QuicMuxer) {.async: (raises: []).} =
   try:
     await m.quicSession.close()
-    m.handleFut.cancel()
+    m.handleFut.cancelSoon()
   except CatchableError as exc:
     discard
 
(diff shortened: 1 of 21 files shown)


richard-ramos:
Looks like there are some issues with building mbed-tls in windows, macos(arm64) and linux(i386).
Will try to fix this in https://github.com/vacp2p/nim-libp2p/pull/1278 as it seems that certificate generation has trouble on those OSs

libp2p/transports/quictransport.nim
```diff
@@ -155,12 +171,27 @@ method start*(
...
   doAssert self.listener.isNil, "start() already called"
   #TODO handle multiple addr
+
+  let pubkey = self.privateKey.getPublicKey().valueOr:
+    doAssert false, "could not obtain public key"
+    return
+
+  let keypair = KeyPair(seckey: self.privateKey, pubkey: pubkey)
+
+  let certPair = generate(keypair, EncodingFormat.PEM)
```
vladopajic, line 181: - `generate` what? it's deductible from context that certificate is created but better naming would be appreciated. 
- also `generate` returns tuple, but why not named tuple? or why not simple type? usage like `certPair[0]`, `certPair[1]` is not really readable and understandable unless you go into the function.
 
 
i know it's not related to this pr, but if you agree we can create issue for this, ideal for  "good first issues" lable.

vladopajic (approved):
lgtm

libp2p/transports/quictransport.nim
```diff
@@ -155,12 +164,35 @@ method start*(
...
   doAssert self.listener.isNil, "start() already called"
   #TODO handle multiple addr
+
+  try:
+    let tlsConfig = TLSConfig.init()
+    if self.rng.isNil:
+      self.rng = newRng()
+    self.client = QuicClient.init(tlsConfig, rng = self.rng)
+  except QuicConfigError as exc:
+    doAssert false, "invalid quic setup: " & $exc.msg
```
vladopajic, line 174: so client will not have certificate?

vladopajic:
🚀🚀🚀

PR #1266, opened 2025-02-24
chore(transports): specify raised exceptions

vladopajic:


libp2p/transports/tortransport.nim
```diff
@@ -147,13 +157,13 @@ proc readServerReply(transp: StreamTransport) {.async.} =
 
 proc parseOnion3(
     address: MultiAddress
-): (byte, seq[byte], seq[byte]) {.raises: [LPError, ValueError].} =
+): (byte, seq[byte], seq[byte]) {.raises: [LPError].} =
   var addressArray = ($address).split('/')
   if addressArray.len < 2:
-    raise newException(LPError, fmt"Onion address not supported {address}")
+    raise newException(LPError, "Onion address not supported")
```
richard-ramos, line 163: Curious about why remove the {address} from this and other errors

libp2p/transports/transport.nim
```diff
@@ -96,3 +105,17 @@ method handles*(self: Transport, address: MultiAddress): bool {.base, gcsafe.} =
...
+      await noCancel stream.closeWait()
+    except CatchableError as e:
+      trace "Error closing", description = e.msg
+
+template safeClose*(stream: untyped) =
+  if not isNil(stream):
+    try:
+      await noCancel stream.close()
+    except CatchableError as e:
+      trace "Error closing", description = e.msg
```
richard-ramos, line 121: Wdyt of having these safe methods return a result, similar to this but with `Result[void, cstring]`:

```
  proc safeToLowerAscii*(s: string): Result[string, cstring] =
    try:
      ok(s.toLowerAscii())
    except CatchableError:
      err("toLowerAscii failed")
```


I guess it would look something like
```nim
proc safeClose*(stream: Stream): Result[void, cstring] =
  if isNil(stream):
       return ok()  # Using early return here

  return try:
      await noCancel stream.close()
      ok()
   except CatchableError as e:
     err(e.msg)
```

Then wherever you use safeClose / safeCloseWait:
```nim
safeClose(stream).isOkOr:
   trace "error closing stream", description = error
```
thus removing the responsability for logging from the safeClose/safeCloseWait functions and leaving that to the caller.

libp2p/transports/tcptransport.nim
```diff
@@ -206,130 +208,121 @@ method accept*(self: TcpTransport): Future[Connection] =
...
-      except common.TransportError as exc: # Needed for chronos 4.0.0 support
-        raise (ref TcpTransportError)(msg: exc.msg, parent: exc)
-      except CancelledError as exc:
-        cancelAcceptFuts()
-        raise exc
 
-    if not self.running: # Stopped while waiting
-      await transp.closeWait()
-      raise newTransportClosedError()
+  proc cancelAcceptFuts() =
```
richard-ramos, line 212: I wonder what was the reason for having this logic in a proc within the `accept` method. The code is clearer now

richard-ramos (approved):

libp2p/transports/tortransport.nim
vladopajic: `fmt` can cause `ValueError`, and i didn't want to propagate this all the way up. also it seems unnecessary to include address here because caller knows what address it passed:

pseudo code:
```
res, err := parseOnion3(address)
if err != nil { 
   // here caller knows what address it passed and he can choose to log it, 
   // no need to included it in err
  printf("could not parse adddress %v, reason: %v", address, err)
  // stdout: "could not parse {addr}, reason: Onion address not supported"
}

```

libp2p/transports/transport.nim
vladopajic: the idea of `safeXxx` was to have one liner to handle common use case.

```
safeClose(stream).isOkOr:
   trace "error closing stream", description = error
```
^ why not include `trace "error closing stream"` in method, if it will be repeated every time? if message needs to be adjusted it can be optional parameter (i intentionally didn't went in that direction, as i want to keep it simple for now)

vladopajic: however, this is something that we need to discuss. because this is common use case that is handled differently all over the place (guess it's handled differently because of lack of focus on details not because it needed different use case).

PR #1268, opened 2025-02-25
chore(protocols):  specify raised exceptions (part 2)

vladopajic:
this PR specifies raised exceptions in `protocols` module. it is part 2, continuation to  #1260.

### DIFF
--- a/libp2p/protocols/connectivity/autonat/client.nim
+++ b/libp2p/protocols/connectivity/autonat/client.nim
@@ -69,8 +69,9 @@ method dialMe*(
     if incomingConnection.completed():
       try:
         await (await incomingConnection).connection.close()
-      except:
-        discard
+      except AlreadyExpectingConnectionError as e:
+        # this err is already handled above and could not happen later
+        error "Unexpected error", description = e.msg
 
   try:
     trace "sending Dial", addrs = switch.peerInfo.addrs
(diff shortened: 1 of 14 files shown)


libp2p/protocols/connectivity/autonat/client.nim
```diff
@@ -61,14 +67,36 @@ method dialMe*(
...
       # Safer to always try to cancel cause we aren't sure if the peer dialled us or not
     if incomingConnection.completed():
-      await (await incomingConnection).connection.close()
-  trace "sending Dial", addrs = switch.peerInfo.addrs
-  await conn.sendDial(switch.peerInfo.peerId, switch.peerInfo.addrs)
-  let response = getResponseOrRaise(AutonatMsg.decode(await conn.readLp(1024)))
+      try:
+        await (await incomingConnection).connection.close()
+      except:
+        discard
```
richard-ramos, line 73: Curious about the exception being generated here. Do you think it's worth it to add a trace log?

richard-ramos (approved):

libp2p/protocols/connectivity/autonat/client.nim
vladopajic: i internationally didn't wanted to add log there, as compiler tells that `AlreadyExpectingConnectionError` can happen here, but, if this error happens it will happen only right after `switch.connManager.expectConnection(pid, In)` is called and that case is already handled. however i have added comment and `error` log. it should never happen tho

richard-ramos (approved):

PR #1278, opened 2025-03-05
fix: compile error due to mbedtls

richard-ramos:


richard-ramos:
Closing PR as we'll use openssl instead of mbedtls for certificate generation as described in https://github.com/vacp2p/nim-libp2p/pull/1298

PR #1287, opened 2025-03-14
chore: remove python dependency

richard-ramos:


vladopajic (commented):
there should be python in `transport-interop/Dockerfile` as well.

PR #974, opened 2023-11-02
feat: WIP Adds support for poseidon2 multihashes

markspanbroek:
This is probably going to be work in progress for a while...

PR #1291, opened 2025-03-17
Replace nimcrypto SHA256 with BLS

munna0908:
While running benchmarks in nim-codex, we identified that the nimcrypto SHA256 hash function is slow. Therefore, we are replacing it with blst_sha256, which outperforms the nimcrypto implementation in most scenarios.


dryajov (approved):

dryajov:
going to push directly to our running branch #974

PR #1295, opened 2025-03-20
feat(quic): tls certificate verify

vladopajic:


libp2p/transports/quictransport.nim
```diff
@@ -150,9 +150,28 @@ type QuicTransport* = ref object of Transport
...
+      return false
+
+    let cert =
+      try:
+        parse(certificatesDer[0])
+      except CertificateParsingError as e:
+        trace "CertificateVerifier: failed to parse certificate", msg = e.msg
+        return false
+
+    return cert.verify()
```
richard-ramos, line 167: I missed this in the PR that introduces `verify*`, but we should also validate the expiration date of the certificate:
```
The certificate MUST have NotBefore and NotAfter fields set such that the certificate is valid at the time it is received by the peer. When receiving the certificate chain, an endpoint MUST check these conditions and abort the connection attempt if (a) the presented certificate is not yet valid, OR (b) if it is expired. 
```

richard-ramos:
Seems like it got confused with merge/rebase but logic related to certificate verification lgtm

PR #1300, opened 2025-03-27
chore: update quic and expect exception in test

richard-ramos:


tests/testquic.nim
```diff
@@ -102,8 +102,12 @@ suite "Quic transport":
 
     proc runClient() {.async.} =
       let client = await createTransport(true)
-      let connFut = client.dial("", server.addrs[0])
-      # conn should be refused by server
-      check not await connFut.withTimeout(1.seconds)
+      try:
+        let conn = await client.dial("", server.addrs[0])
+        await conn.close()
```
vladopajic, line 107:
```suggestion
        let conn = await client.dial("", server.addrs[0])
        check false # conn should be refused by client
```

connection should not be established with invalid cert? so if dial is successfully `check false` will ensure to fail test

vladopajic (approved):

PR #1303, opened 2025-03-30
feat: interop

richard-ramos:


richard-ramos:
image

@vladopajic, @kaiserd ! :) good news:
Only zig failed interop with an ERR_CRYPTO :thinking: meaning that there might be an issue with the certificate (either what they send or how we validate it?)

richard-ramos:
Looking good now:

image

I was missing ED25519 from the list of supported algos for certificates

richard-ramos:
Now the only thing left is to make adding security managers / muxers optional, depending on the transports you add

vladopajic (approved):

gmelodie (approved):

PR #1304, opened 2025-04-01
feat(transport): add memory transport

vladopajic:
this pr add `memory` transports that can be used as drop in replacement for other transports in case if users don't need to go into network stack. memory transport only works in scope of same process.

### DIFF
--- a/libp2p/stream/bridgestream.nim
+++ b/libp2p/stream/bridgestream.nim
@@ -1,0 +1,60 @@
+# Nim-LibP2P
+# Copyright (c) 2025 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
+import pkg/chronos
+import connection, bufferstream
+
+export connection
+
+type
+  WriteHandler = proc(data: seq[byte]): Future[void] {.
+    async: (raises: [CancelledError, LPStreamError])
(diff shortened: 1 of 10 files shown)


libp2p/stream/bridgestream.nim
```diff
@@ -0,0 +1,60 @@
...
+  procCall BufferStream(s).closeImpl()
+
+method getWrapped*(s: BridgeStream): Connection =
+  nil
+
+proc bridgedConnections*(): (BridgeStream, BridgeStream) =
+  let connA = BridgeStream()
+  let connB = BridgeStream()
+  connA.dir = Direction.In
+  connB.dir = Direction.In
```
richard-ramos, line 42: should both connections have same direction?

libp2p/transports/memorytransport.nim
```diff
@@ -0,0 +1,126 @@
...
+
+proc new*(
+    T: typedesc[MemoryTransport],
+    upgrade: Upgrade = Upgrade(),
+    rng: ref HmacDrbgContext = newRng(),
+): T =
+  T(upgrader: upgrade, rng: rng)
+
+proc listenAddress(self: MemoryTransport, ma: MultiAddress): MultiAddress =
+  if $ma != "/memory/*":
```
richard-ramos, line 42: Wdyt about making `/memory/*` a constant. You could later reuse this constant when instantiating your switch

richard-ramos (approved):
good work!

libp2p/stream/bridgestream.nim
vladopajic: it doesn't really have any impact on memory transport. in reality both should be bidirectional (In + Out), but that option is not possible to set.

PR #1317, opened 2025-04-08
fix: do not use while loop for quic transport errors

richard-ramos:
While reviewing code in more detail I saw that instead of having a `while true` and `continue` in the quic transport, we should instead return nil. This is gonna be later handled in here https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/switch.nim#L273-L283 where we check whether the returned connection is nil or not. Since in this PR, when there's a QuicError, no connection will be returned, then we'll enter this condition https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/switch.nim#L281-L283 . Do note that this code is already inside a `while transport.isRunning`. https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/switch.nim#L264 so it will try to `accept` again in the next iteration

This should be refactored later in all transports to not return nil, but an option instead since it's semantically better

### DIFF
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -255,7 +255,7 @@ method stop*(transport: QuicTransport) {.async: (raises: []).} =
 
 proc wrapConnection(
     transport: QuicTransport, connection: QuicConnection
-): QuicSession {.raises: [TransportOsError, LPError].} =
+): QuicSession {.raises: [TransportOsError, MaError].} =
   let
     remoteAddr = connection.remoteAddress()
     observedAddr =
@@ -290,12 +290,16 @@ method accept*(
   try:
     let connection = await self.listener.accept()
     return self.wrapConnection(connection)
-  except CancelledError as e:
-    raise e
-  except QuicError as e:
-    raise (ref QuicTransportError)(msg: e.msg, parent: e)
-  except CatchableError as e:
-    raise (ref QuicTransportError)(msg: e.msg, parent: e)
+  except CancelledError as exc:
+    raise exc
+  except QuicError as exc:
+    debug "Quic Error", description = exc.msg
+  except MaError as exc:
+    debug "Multiaddr Error", description = exc.msg
+  except CatchableError as exc: # TODO: removing this requires async/raises in nim-quic
+    info "Unexpected error accepting quic connection", description = exc.msg
+  except TransportOsError as exc:
+    debug "OS Error", description = exc.msg
 
 method dial*(
     self: QuicTransport,
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -283,22 +283,19 @@ method accept*(
 .} =
   doAssert not self.listener.isNil, "call start() before calling accept()"
 
-  while true:
-    if not self.running:
-      # stop accept only when transport is stopped (not when error occurs)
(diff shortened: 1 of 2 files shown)


tests/testquic.nim
```diff
@@ -12,6 +12,7 @@ import
     errors,
     wire,
   ]
+import quic
```
vladopajic, line 15: what is this import?

vladopajic (approved):

vladopajic:
it seams like code from PR is not returning nil connection but raising exception.
which will get propagated [here](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/switch.nim#L275-L278), then it will brake accept loop [here](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/switch.nim#L304)

vladopajic:
also this code in switch seems bit inefficient when connection nil is returned with no exception.
more specifically, these statements are repeated unnecessarily  https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/switch.nim#L271:L272

libp2p/transports/quictransport.nim
```diff
@@ -283,22 +283,19 @@ method accept*(
 .} =
   doAssert not self.listener.isNil, "call start() before calling accept()"
 
-  while true:
-    if not self.running:
-      # stop accept only when transport is stopped (not when error occurs)
-      raise newException(QuicTransportAcceptStopped, "Quic transport stopped")
+  if not self.running:
+    # stop accept only when transport is stopped (not when error occurs)
+    raise newException(QuicTransportAcceptStopped, "Quic transport stopped")
```
vladopajic, line 288: is this exception ever going to be raised?

vladopajic: ` doAssert not self.listener.isNil` will happen before `if not self.running:`

technically, both will not be satisfied simultaneously, unless it is a race condition (which is vaaaary low probability). but code anyway has a lot of those.

tests/testquic.nim
```diff
@@ -24,13 +24,18 @@ proc createServerAcceptConn(
...
+
       let stream = await getStream(QuicSession(conn), Direction.In)
       var resp: array[6, byte]
       await stream.readExactly(addr resp, 6)
       check string.fromBytes(resp) == "client"
 
       await stream.write("server")
       await stream.close()
+    except QuicTransportError:
+      discard # Transport failed (maybe dial error / handshake err, etc)
```
vladopajic, line 38: i am not sure about this. if `QuicTransportError` is raised then it implies that `CatchableError` has happened, if it do happen then `discard` doesn't seem right because we don't expect that to happen so we need to brake the tests.

aslo, last time i added this `try` block i was lazy, we should only guard `let conn = await server.accept()` this line with try block.

libp2p/transports/quictransport.nim
```diff
@@ -283,22 +283,19 @@ method accept*(
...
-      raise (ref QuicTransportError)(msg: e.msg, parent: e)
+  try:
+    let connection = await self.listener.accept()
+    return self.wrapConnection(connection)
+  except CancelledError as e:
+    raise e
+  except QuicError as e:
+    debug "Quic Error", description = e.msg
+  except CatchableError as e:
+    raise (ref QuicTransportError)(msg: e.msg, parent: e)
```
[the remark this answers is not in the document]
richard-ramos: will require adding `async/raises` in nim-quic first!

Issue #1307, opened 2025-04-03
rendezvous: code improvements


richard-ramos:
- [ ] Some functions contain timeouts that should probably not be their responsibility, but instead should be specified instead by whoever calls these functions
- [ ] Handling of optional fields in the protobuffer seem incorrect (specially with the namespace field). It currently assumes the default value "" means undefined, while it should check for presence on the wire. i.e.: if the namespace is not set, nothing will be set in wire, if set to "", wire will contain the field with value "".
- [ ] `readLP` calls have a max size that can likely trigger an error if the number of records returned by discovery exceeds that max size. (go-libp2p uses `network.MessageSizeMax`  and rust `MAX_MESSAGE_LEN_BYTES`,  instead)
- [ ]  I have my doubts regarding using an offseted sequence to store peers. I wonder how performant is it when having lots of peers registered in a namespace, and filtering peers or deleting them

PR #1319, opened 2025-04-09
chore: rendezvous improvements

gmelodie:
Fixes (most of) #1307 

1. Uses `Opt[string]` for `namespace` fields on `Cookie` and `request` (which does `Discover`s). You can now do `rdv.request()` or `rdv.request(Opt.none(string))` to get all namespaces.
2. Functions that timeout now get `timeout` as a parameter instead of hardcoded values
3.  Add `MaximumMessageLen` (4MB) as a safe max read size for `readLp`, similar to [go-libp2p's implementation](https://github.com/libp2p/go-libp2p/blob/v0.41.1/core/network/network.go#L19C1-L23C39)

### DIFF
--- a/libp2p/protocols/rendezvous.nim
+++ b/libp2p/protocols/rendezvous.nim
@@ -516,7 +516,7 @@ proc advertisePeer(
       rdv.sema.release()
 
   await rdv.sema.acquire()
-  discard await advertiseWrap().withTimeout(5.seconds)
+  await advertiseWrap()
 
 proc advertise*(
     rdv: RendezVous, ns: string, ttl: Duration, peers: seq[PeerId]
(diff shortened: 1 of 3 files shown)


richard-ramos (changes requested):
Some minor comments.
Also this is failing due to nph formatting: https://github.com/arnetheduck/nph

Also regarding the PR title: https://www.conventionalcommits.org/en/v1.0.0/

libp2p/protocols/rendezvous.nim
```diff
@@ -690,14 +710,14 @@ proc unsubscribe*(
...
-  discard await allFutures(futs).withTimeout(5.seconds)
+  await allFutures(futs)
 
 proc unsubscribe*(
     rdv: RendezVous, ns: string
 ) {.async: (raises: [RendezVousError, CancelledError]).} =
   rdv.unsubscribeLocally(ns)
 
-  await rdv.unsubscribe(ns, rdv.peers)
+  discard await rdv.unsubscribe(ns, rdv.peers).withTimeout(5.seconds)
```
richard-ramos, line 720:
```suggestion
  discard await rdv.unsubscribe(ns, rdv.peers)
```

richard-ramos (approved):
Good work!

PR #1324, opened 2025-04-13
feat(kad-dht)

richard-ramos:


richard-ramos:
It is not necessary to have a -d:kad flag for this code as it is not used anywhere else. 
And yes, this is only missing testing to ensure code is correct.

Ben-PH:
@gmelodie @vladopajic 

I've been looking into doing the tests for this PR, and richard suggested we ping you in on this discussion.

In the interest of keeping the sizes of PRs small, and being explicit about work delegations I thought it would be worth asking to have this PR merged +- as-is, putting this feature behind a `-d:kad` flag until I get the tests sorted out.

In addition to that, I think a productive workflow for me would be as follows:

- recreate the rendesvousinterface.nim file in the discovery module for kad
- recreate the rdv tutorial example using kad
- From there, implement the test suite for kad

Richard thought having the interface and e.g. would be good (I'll leave it to him to explain: don't want to put words in his mouth).

I feel this approach is warranted, as the process of putting together the interface/e.g. will not only make kad a more complete feature on arrival, but will give me a more practical and meaningful understanding of nim-libp2p ins-and-outs.

I could probably get _some_ tests done without this PR merged +- as-is, but I they'll probably be implemented more naively than we would prefer.

Happy to hear your thoughts.

libp2p/protocols/kademlia/consts.nim
```diff
@@ -1,6 +1,10 @@
+import chronos
+
 const
   IdLength* = 32 # 256-bit IDs
   k* = 20 # replication parameter
+  alpha* = 10 # concurrency parameter
+  ttl* = 24.hours
```
Ben-PH, line 7: Though it goes against the reference style guide, having constants in the same style as a variable causes confusion and cognitive overhead.

In addition, alpha and ttl are common identifiers, and with nims way of handling namespace imports, I would suggest a `kad` prefix:

Suggest rename to `KAD_ALPHA` and `KAD_TTL`

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,133 @@ type KadDHT* = ref object of LPProtocol
   rtable*: RoutingTable
   maintenanceLoop: Future[void]
 
+proc sendFindNode(
+    kad: KadDHT, peerId: PeerId, targetId: Key
+): Future[Message] {.
+    async: (raises: [CancelledError, DialFailedError, ValueError, LPStreamError])
```
Ben-PH, line 32: can any of these errors be handled as a `Result`?

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,133 @@ type KadDHT* = ref object of LPProtocol
...
+      state.markFailed(timedOut)
+
+    state.done = state.checkConvergence()
+
+  return state.selectClosestK()
+
+proc bootstrap*(
+    kad: KadDHT, bootstrapNodes: seq[PeerInfo]
+) {.async: (raises: [CancelledError]).} =
+  # TODO: every 10 minutes configurable, run once on start
```
Ben-PH: convert to issue before merging

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -37,12 +163,26 @@ proc new*(
           buf = await conn.readLp(4096)
           msg = Message.decode(buf).tryGet()
 
-        # TODO: handle msg.msgType
+        case msg.msgType
+        of MessageType.findNode:
+          let targetIdBytes = msg.key.get()
+          let targetId = PeerId.init(targetIdBytes).tryGet()
+          let closerPeers = kad.rtable.findClosest(targetId.toKey(), k)
```
Ben-PH: This might not be the best place for it, but some sort of logging for network state might be useful.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -37,12 +163,26 @@ proc new*(
...
+          let responsePb = encodeFindNodeReply(closerPeers, switch)
+          await conn.writeLp(responsePb.buffer)
+
+          # Peer is useful. adding to rtable
+          discard kad.rtable.insert(conn.peerId)
+        else:
+          raise newException(LPError, "unhandled kad-dht message type")
+        # TODO: implement other types
```
Ben-PH: Convert to GH issue before merging

libp2p/protocols/kademlia/keys.nim
```diff
@@ -26,6 +26,11 @@ proc toKey*(s: seq[byte]): Key =
 proc toKey*(p: PeerId): Key =
   return Key(kind: KeyType.PeerId, peerId: p)
 
+proc toPeerId*(k: Key): PeerId {.raises: [ValueError].} =
+  if k.kind != KeyType.PeerId:
+    raise newException(ValueError, "not a peerId")
+  k.peerId
```
Ben-PH, line 32: We lost a significant portion of a day because this line was lost in a rebase or something, and nims use of the magic `result` variable returned a default value.

I say we got _very_ lucky, in that we just wasted time. It's not fanciful to anticipate something much more insideous, e.g. a default variable is the same as a return variable in 99.9 % of cases, doesn't get caught in testing, and becomes a super costly bug that takes an unacceptably long time to nail down.

Out of scope for this PR, but it needs to be brought up.

In my opinion, we should have linters, style guides, CI, etc. that safeguards against all use of the magic `result` variable, so as to mitigate the risk of such events.

libp2p/protocols/kademlia/lookupstate.nim
```diff
@@ -0,0 +1,120 @@
...
+    distance: XorDistance
+    queried: bool # have we already queried this node?
+    pending: bool # is there an active request rn?
+    failed: bool # did the query timeout or error?
+
+  LookupState* = object
+    targetId: Key
+    shortlist: seq[LookupNode] # current known closest node
+    activeQueries*: int # how many queries in flight
+    alpha: int # parallelism level
```
Ben-PH, line 22: From what I saw, this is only ever `alpha` the constant, so does it make sense to have this field?

libp2p/protocols/kademlia/lookupstate.nim
```diff
@@ -0,0 +1,120 @@
...
+    k: int # number of closest nodes to find
```
Ben-PH: I don't know if it's in the style guide, but under the "strong opinions, weakly heald", I am of the strong opinion that single letter variables are +- verbotten.

Request rename to something like `targetClosestN`

libp2p/protocols/kademlia/lookupstate.nim
```diff
@@ -0,0 +1,120 @@
...
+    done*: bool # has lookup converged
+
+proc alreadyInShortlist(state: LookupState, peer: Peer): bool =
```
Ben-PH, line 26: can this be a `func` instead?

Come to think of it, are there linters that detect `proc` when `func` is possible?

libp2p/protocols/kademlia/lookupstate.nim
```diff
@@ -0,0 +1,120 @@
...
+          LookupNode(
+            peerId: peerInfo.peerId,
+            distance: xorDistance(peerInfo.peerId, state.targetId),
+            queried: false,
+            pending: false,
+            failed: false,
+          )
+        )
+      except Exception as exc:
+        debug "could not update shortlist", err = exc.msg
```
Ben-PH: Can this be bubbled up as an error? I don't assert that it needs to be done for this PR, but it might be prudent to add it to the list: The caller might be able to handle failed update.

libp2p/protocols/kademlia/lookupstate.nim
```diff
@@ -0,0 +1,120 @@
...
+    if p.peerId == peerId:
+      p.pending = true
+      p.queried = true
+      break
+
+proc selectAlphaPeers*(state: LookupState): seq[PeerId] =
+  var selected: seq[PeerId] = @[]
+  for p in state.shortlist:
+    if not p.queried and not p.failed and not p.pending:
+      selected.add(p.peerId)
```
Ben-PH, line 75: `PeerId` is of type `seq[byte]`, and as i understand it, 64 bytes.

is it possible to have a sequence of references to p.peerId to avoid copying? I don't see this being worth the effort or consideration under "avoid premature optimisation", as well as the more complicated memory model, but if we are serving highly resource-constrained use-cases, and we are concerned with "death by a thousand cuts", this, or where this sort of pattern is used at higher scale, might merit some attention at some point.

libp2p/protocols/kademlia/lookupstate.nim
```diff
@@ -0,0 +1,120 @@
...
+        queried: false,
+        pending: false,
+        failed: false,
+      )
+    )
+
+  result.shortlist.sort(
+    proc(a, b: LookupNode): int =
+      cmp(a.distance, b.distance)
+  )
```
Ben-PH: if shortlist must always be sorted, consider using a min-heap

libp2p/protocols/kademlia/requests.nim
```diff
@@ -0,0 +1,34 @@
+import ../../peerid
+import ../../switch
+import ../../peerstore
+import ./protobuf
+import ../../protobuf/minprotobuf
+import ./keys
+
+proc encodeFindNodeReply*(
+    closerPeers: seq[Key], switch: Switch
+): ProtoBuffer {.raises: [].} =
```
Ben-PH, line 10: Is it ever possible that the returned protobuffer hasn't had anything added to it? consider returning an `Option` if so.

If not, consider adding an assert that only runs during debug builds to assert it's non-empty

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,133 @@ type KadDHT* = ref object of LPProtocol
...
+    kad: KadDHT, bootstrapNodes: seq[PeerInfo]
+) {.async: (raises: [CancelledError]).} =
+  # TODO: every 10 minutes configurable, run once on start
+  for b in bootstrapNodes:
+    try:
+      await kad.switch.connect(b.peerId, b.addrs)
+      debug "connected to bootstrap peer", peerId = b.peerId
+
+      try:
+        let msg = await kad.sendFindNode(b.peerId, kad.rtable.selfId).wait(5.seconds)
```
Ben-PH: Document a follow-up to draw this wait from a configuration (which defaults to 5)

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,133 @@ type KadDHT* = ref object of LPProtocol
...
+      failedPeers.add(peerId)
+      error "could not send find_node to peer", peerId, err = getCurrentExceptionMsg()
+
+  return (receivedReplies, failedPeers)
+
+proc findNode(
+    kad: KadDHT, targetId: Key
+): Future[seq[PeerId]] {.async: (raises: [CancelledError]).} =
+  #debug "findNode", target = target
+  # TODO: check if already exist in rtable
```
Ben-PH: convert to documented follow-up item

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,133 @@ type KadDHT* = ref object of LPProtocol
...
+
+  var receivedReplies: seq[Message] = @[]
+  var failedPeers: seq[PeerId] = @[]
+
+  for (peerId, replyFut) in pendingFutures.pairs:
+    try:
+      let reply = await replyFut
+      receivedReplies.add(reply)
+    except CatchableError:
+      failedPeers.add(peerId)
```
Ben-PH, line 66: Try to avoid using exceptions for control flow.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: replace `targetId.toKey()` with `msg.key.get()`?

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -37,12 +163,26 @@ proc new*(
...
+          let targetIdBytes = msg.key.get()
+          let targetId = PeerId.init(targetIdBytes).tryGet()
+          let closerPeers = kad.rtable.findClosest(targetId.toKey(), k)
+          let responsePb = encodeFindNodeReply(closerPeers, switch)
+          await conn.writeLp(responsePb.buffer)
+
+          # Peer is useful. adding to rtable
+          discard kad.rtable.insert(conn.peerId)
+        else:
+          raise newException(LPError, "unhandled kad-dht message type")
```
Ben-PH, line 186: This gets immediately caught and discarded, from what I can tell, and is used as a mechanism to break out of the `while not conn.atEoF` loop

consider non-exception based whay to do this. Perhaps a helper function to run the loop, and instead of `raise`, return a `error(...)`?

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,133 @@ type KadDHT* = ref object of LPProtocol
...
+    for timedOut in timedOutPeers:
+      state.markFailed(timedOut)
+
+    state.done = state.checkConvergence()
+
+  return state.selectClosestK()
+
+proc bootstrap*(
+    kad: KadDHT, bootstrapNodes: seq[PeerInfo]
+) {.async: (raises: [CancelledError]).} =
```
Ben-PH, line 121: multiple errors are caught and logged. Consider returning `KadResult[void]`, either in this PR, or document for follow-up before merging.

Ben-PH (commented):
I hope my comments are helpful. For the most part, my comments come from "strong opinions, weakly held". I defer to the interests of moving forward, but where invited to do so, I would be glad to advocate for them. 

I feel the need to more strongly insist with respect to the use of exceptions as control flow, so here is my take on the matter, though I recognise my naivity on this topic:

By its very nature, exceptions introduce ambiguity into the execution flow of a program. Control flow using a mechanism that has execution flow abiguity built into it is (probably) a fast-path to code-rot, and should be corrected before it has a chance to take root.

libp2p/protocols/kademlia/consts.nim
richard-ramos: These constants are to be invoked only within kad-dht, so adding the prefix would make it redundant. Wrt styling, i dont think we should deviate from nim style guide unless absolutely necessary. Related discussion: https://github.com/vacp2p/nim-libp2p/issues/1531

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: Yeah, if it adds value go for it!, although keep in mind it will probably be very noisy. Ideally it should be `trace`

libp2p/protocols/kademlia/lookupstate.nim
richard-ramos: At the time i was not sure whether it would make sense to have alpha as configurable. Might have been a premature optimization. Do feel free to remove this attribute

libp2p/protocols/kademlia/lookupstate.nim
richard-ramos: Go for it! :) , Not sure about the naming tho, as in the spec it says: `Replication parameter (k)`

libp2p/protocols/kademlia/lookupstate.nim
richard-ramos: sure, seems to be side effect free. 
No idea about linters tho :thinking:

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: This `else / raise` shouldnt exist in the final version. I just added it because I had at the time not included all of the message types. But once those cases are added (including deprecated ping), we should get rid of it, as unhandled message types will be handled anyway by the protobuffer decoding:
```
  if not checkedEnumAssign(msgType, msgTypeVal):
    return err(ProtoError.BadWireType)
```

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: yes

libp2p/protocols/kademlia/requests.nim
richard-ramos: Returning an Option does not add value here as we take the response of encodeFindNodeReply and immediatly write it on the stream:
```
          let responsePb = encodeFindNodeReply(closerPeers, switch)
          await conn.writeLp(responsePb.buffer)
```

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: These exceptions are generated from failures interacting with other nodes, with failures ranging from timeout, streams abruptly closed, dial failures, etc. There's no guarantee that a node will reply to a FindNode request

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: resolved via #1545

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: resolved via #1545

libp2p/protocols/kademlia/lookupstate.nim
Ben-PH: resolved via #1545

libp2p/protocols/kademlia/lookupstate.nim
Ben-PH: resolved via #1545

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: resolved via #1545

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: resolved via #1545

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: resolved via #1545

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: resolved via #1545

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: resolved via #1545

libp2p/protocols/kademlia/lookupstate.nim
Ben-PH: resolved via #1545

libp2p/protocols/kademlia/lookupstate.nim
Ben-PH: I'll use `replParam` and doc-comment the spec equivilent (`k`)

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: resolved via #1545

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: This was less trivial than I anticipated. I was going by an assumption that the type system would make it 100% trivial, but some things were unaccounted for.

My thinking now is to leave as-is, and let our work with the type system govern things like this.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,142 @@ type KadDHT* = ref object of LPProtocol
...
+proc sendFindNode(
+    kad: KadDHT, peerId: PeerId, addrs: seq[MultiAddress], targetId: Key
+): Future[Message] {.
+    async: (raises: [CancelledError, DialFailedError, ValueError, LPStreamError])
+.} =
+  var conn: Connection
+  if addrs.len == 0:
+    conn = await kad.switch.dial(peerId, KadCodec)
+  else:
+    conn = await kad.switch.dial(peerId, addrs, KadCodec)
```
gmelodie: I think this check should be done inside `kad.switch.dial`, or at least 
```nim
let conn =
  if addrs.len == 0:
    ...
  else:
    ...

File: libp2p/protocols/kademlia/kademlia.nim [hunk 20]
```diff
@@ -17,14 +24,142 @@ type KadDHT* = ref object of LPProtocol
...
+
+  var receivedReplies: seq[Message] = @[]
+  var failedPeers: seq[PeerId] = @[]
+
+  for (peerId, replyFut) in pendingFutures.pairs:
+    try:
+      let reply = await replyFut
+      receivedReplies.add(reply)
+    except CatchableError:
+      failedPeers.add(peerId)
```
gmelodie: Nit: maybe use `filterIt`

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,142 @@ type KadDHT* = ref object of LPProtocol
...
+
+  var initialPeers = kad.rtable.findClosestPeers(targetId, DefaultReplic)
+  var state = LookupState.init(targetId, initialPeers)
+  var addrTable: Table[PeerId, seq[MultiAddress]] =
+    initTable[PeerId, seq[MultiAddress]]()
+
+  while not state.done:
+    let toQuery = state.selectAlphaPeers()
+
+    var pendingFutures = initTable[PeerId, Future[Message]]()
```
gmelodie, line 84: Nitity nit: coud we maybe have a `var` that we initialize once and clear in every iteration instead? There's maybe not much gain though

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,142 @@ type KadDHT* = ref object of LPProtocol
...
+          let p = PeerId.init(peer.id).tryGet()
+          discard kad.rtable.insert(p)
+          kad.switch.peerStore[AddressBook][p] = peer.addrs
+
+        # bootstrap node replied succesfully. Adding to routing table
+        discard kad.rtable.insert(b.peerId)
+      except CatchableError as e:
+        error "bootstrap failed for peer", peerId = b.peerId, exc = e.msg
+    except CatchableError as e:
+      error "failed to connect to bootstrap peer", peerId = b.peerId, error = e.msg
```
gmelodie: Could we do this not to require nested `try...except`s instead?
```suggestion
    try:
      await kad.switch.connect(b.peerId, b.addrs)
      debug "connected to bootstrap peer", peerId = b.peerId
    except CatchableError as e:
      error "failed to connect to bootstrap peer", peerId = b.peerId, error = e.msg
    try:
      let msg =
        await kad.sendFindNode(b.peerId, b.addrs, kad.rtable.selfId).wait(5.seconds)
      for peer in msg.closerPeers:
        let p = PeerId.init(peer.id).tryGet()
        discard kad.rtable.insert(p)
        kad.switch.peerStore[AddressBook][p] = peer.addrs

        # bootstrap node replied succesfully. Adding to routing table
        discard kad.rtable.insert(b.peerId)
    except CatchableError as e:
      error "bootstrap failed for peer", peerId = b.peerId, exc = e.msg
   
```

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -37,12 +172,25 @@ proc new*(
...
+          raise newException(LPError, "unhandled kad-dht message type")
     except CancelledError as exc:
       raise exc
     except CatchableError:
-      error "could not handle request",
-        peerId = conn.PeerId, err = getCurrentExceptionMsg()
+      discard
+      # TODO: figure out why this fails:
+      # error "could not handle request",
+      #   peerId = conn.PeerId, err = getCurrentExceptionMsg()
```
gmelodie, line 193: Why not leave the error log?

libp2p/protocols/kademlia/lookupstate.nim
```diff
@@ -0,0 +1,120 @@
...
+    activeQueries*: int # how many queries in flight
+    alpha: int # parallelism level
+    repliCount: int ## aka `k` in the spec: number of closest nodes to find
+    done*: bool # has lookup converged
+
+proc alreadyInShortlist(state: LookupState, peer: Peer): bool =
+  for p in state.shortlist:
+    if p.peerId.getBytes() == peer.id:
+      return true
+  return false
```
gmelodie: We could just return a `filterIt(...).len == 0`, no?

libp2p/protocols/kademlia/lookupstate.nim
```diff
@@ -0,0 +1,120 @@
...
+
+proc updateShortlist*(
+    state: var LookupState, msg: Message, onInsert: proc(p: PeerInfo) {.gcsafe.}
+) =
+  for newPeer in msg.closerPeers:
```
gmelodie: On top of the comment above, could we maybe do

```
for newPeer in msg.closerPeers.filterIt(state, it):
```

libp2p/protocols/kademlia/lookupstate.nim
```diff
@@ -0,0 +1,120 @@
...
+
+proc checkConvergence*(state: LookupState): bool =
+  let ready = state.activeQueries == 0
+  let noNew = selectAlphaPeers(state).len == 0
+  return ready and noNew
+
+proc selectClosestK*(state: LookupState): seq[PeerId] =
+  result = @[]
+  for p in state.shortlist:
+    if not p.failed:
```
gmelodie: Same as above, maybe use `filterIt`

gmelodie (approved):
Looks good, just needs CI fixes.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I like the proposed alternative.

With respect to how to go about calling `dial`, I see it as needing a higher-level discussion (which I started here: #1552)

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,142 @@ type KadDHT* = ref object of LPProtocol
...
+    conn = await kad.switch.dial(peerId, addrs, KadCodec)
+
+  defer:
+    await conn.close()
+
+  let msg = Message(msgType: MessageType.findNode, key: some(targetId.getBytes()))
+
+  await conn.writeLp(msg.encode().buffer)
+
+  let reply = Message.decode(await conn.readLp(4096)).tryGet()
```
vladopajic: 4096 could be const

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,142 @@ type KadDHT* = ref object of LPProtocol
...
+): Future[(seq[Message], seq[PeerId])] {.async: (raises: [CancelledError]).} =
+  await allFutures(toSeq(pendingFutures.values))
+
+  var receivedReplies: seq[Message] = @[]
+  var failedPeers: seq[PeerId] = @[]
+
+  for (peerId, replyFut) in pendingFutures.pairs:
+    try:
+      let reply = await replyFut
+      receivedReplies.add(reply)
```
vladopajic:
```suggestion
      receivedReplies.add(await replyFut)
```

kills one line, unnecessary naming variable (whose scope expires in next line)...

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,142 @@ type KadDHT* = ref object of LPProtocol
...
+
+    for peer in toQuery:
+      if pendingFutures.hasKey(peer):
+        continue
+
+      state.markPending(peer)
+
+      pendingFutures[peer] = kad
+        .sendFindNode(peer, addrTable.getOrDefault(peer, @[]), targetId)
+        .wait(5.seconds)
```
vladopajic, line 94: why `.wait(5.seconds)`?

libp2p/protocols/kademlia/kademlia.nim
vladopajic: +1 if nesting could be avoided

libp2p/protocols/kademlia/lookupstate.nim
```diff
@@ -0,0 +1,120 @@
...
+      p.queried = true
+      state.activeQueries.dec
+      break
+
+proc markPending*(state: var LookupState, peerId: PeerId) =
+  for p in mitems(state.shortlist):
+    if p.peerId == peerId:
+      p.pending = true
+      p.queried = true
+      break
```
vladopajic, line 69:
```
  for p in mitems(state.shortlist):
    if p.peerId == peerId:
```

fyi, iterating and searching for a pear could be a template here it looks like something that can be repeated

libp2p/protocols/kademlia/lookupstate.nim
```diff
@@ -0,0 +1,120 @@
...
+
+  result.shortlist.sort(
+    proc(a, b: LookupNode): int =
+      cmp(a.distance, b.distance)
+  )
+
+proc checkConvergence*(state: LookupState): bool =
+  let ready = state.activeQueries == 0
+  let noNew = selectAlphaPeers(state).len == 0
+  return ready and noNew
```
vladopajic, line 109: with variable assignment  `noNew` will always be evaluated, even in case when `ready=false` and there is no need to.
also `selectAlphaPeers` in this case would be a bit more efficient if it was `hasAlphaPeers` returning true if there is any alpha peer.

vladopajic (approved):
lgtm, only nitpicks

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: for added context, this is an item in #1545

I can't answer as to why `5` was chosen.

libp2p/protocols/kademlia/lookupstate.nim
Ben-PH: Looking at the code in more detail, the list requires ordered iteration without consuming, which isn't really the fortee of heaps.

I can possibly think of btrees.

The motivation here is to encode within the type system that `shortlist` is always sorted. BTreeMap is one that comes to mind, but really: it's kinda splitting hairs and bikeshedding: the use is very simple, it doesn't appear to be calling for more complexity, so I say just leave it for now.

hit "resolve" if agree.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -17,14 +24,142 @@ type KadDHT* = ref object of LPProtocol
...
+        error "bootstrap failed for peer", peerId = b.peerId, exc = e.msg
+    except CatchableError as e:
+      error "failed to connect to bootstrap peer", peerId = b.peerId, error = e.msg
+
+  try:
+    # Adding some random node to prepopulate the table
+    discard await kad.findNode(PeerId.random(kad.rng).tryGet().toKey())
+    info "bootstrap lookup complete"
+  except CatchableError as e:
+    error "bootstrap lookup failed", error = e.msg
```
richard-ramos, line 147: Continuing our conversation about your find about code not following specs:
> On every run, we generate a random peer ID for every non-empty routing table's k-bucket 

We can probably fix this by replacing this code by a call to `refreshBuckets` which is supposed to do what spec suggests.
```suggestion
await refreshBuckets()
```

And remove `maintainBuckets` proc,

Also, `refreshBuckets` will likely not need that `if kad.rtable.buckets[i].isStale():` that checks if the bucket contains stale peers or not, and can just execute findNode directly.

libp2p/protocols/kademlia/lookupstate.nim
Ben-PH:
```nim
proc alreadyInShortlist(state: LookupState, peer: Peer): bool =
  return state.shortlist.anyIt(it.peerId.getBytes() == peer.id)
```
Will use this.

libp2p/protocols/kademlia/lookupstate.nim
Ben-PH: You're correct going by the code as written, however this would be probably be optimised away by the compiler.

 If the compiler inlined `ready` and `noNew`, that would mean that `selectAlphaPeers` would _only_ be run if `ready` is true. By my understanding, such inlining is trivially easy. 
 
 Thoughts?

libp2p/protocols/kademlia/lookupstate.nim
Ben-PH: I've replaced them with:

```nim
  for p in mitems(state.shortlist.filterIt(it.peerId == peerId)):
```

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I'm using filterIt elsewhere, but I'm not sure how it can apply here. Maybe it's something to do with exceptions that I'm yet to come across?

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I highly doubt anything would be gained performance wise; even if this is not optimal at face value, the compiler is likely to to optimise away anyway.

For me, this layout makes the code more readable and maintainable, as it means that there is less potential for confusion about what scope the variable is relevant to.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I think a follow up fix PR is not an inappropriate approach.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I'm not familiar with the context behind this. @richard-ramos would you be happy to chime in?

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: sounds good

tests/testnative.nim
```diff
@@ -34,7 +34,7 @@ import
   testdiscovery, testyamux, testautonat, testautonatservice, testautorelay, testdcutr,
   testhpservice, testutility, testhelpers, testwildcardresolverservice, testperf
 
-import kademlia/[testencoding, testroutingtable]
+import kademlia/[testencoding, testroutingtable, testfindnode]
```
gmelodie, line 37: Try changing the order here, does autotls still fail?
```suggestion
import kademlia/[testencoding, testfindnode, testroutingtable]
```

libp2p/protocols/kademlia/lookupstate.nim
vladopajic: yeah could be, but not really sure if it does that

tests/testnative.nim
Ben-PH: yeah, still fails:

```
=======================
  /home/ben/IFT/nim-libp2p/tests/testnative 'AutoTLS ACME API::register to acme server'
-----------------------
    Opened LPChannel: 1283
Closed LPChannel: 1267
    Opened SecureConn: 453
Closed SecureConn: 437
    Opened libp2p.tcptransport: 330
Closed libp2p.tcptransport: 323
    Opened stream.transport: 650
Closed stream.transport: 634
    Opened stream.server: 366
Closed stream.server: 359
    Opened ChronosStream: 591
Closed ChronosStream: 575

  [FAILED ] (  0.49s) register to acme server

==============================
  /home/ben/IFT/nim-libp2p/tests/testnative 'AutoTLS ACME API::request challenge for a domain'
------------------------------
    Opened LPChannel: 1283
Closed LPChannel: 1267
    Opened SecureConn: 453
Closed SecureConn: 437
    Opened libp2p.tcptransport: 330
Closed libp2p.tcptransport: 323
    Opened stream.transport: 650
Closed stream.transport: 634
    Opened stream.server: 366
Closed stream.server: 359
    Opened ChronosStream: 591
Closed ChronosStream: 575

  [FAILED ] (  0.39s) request challenge for a domain

==============================
  /home/ben/IFT/nim-libp2p/tests/testnative 'AutoTLS ACME API::register with unsupported keys'
------------------------------
    Opened LPChannel: 1283
Closed LPChannel: 1267
    Opened SecureConn: 453
Closed SecureConn: 437
    Opened libp2p.tcptransport: 330
Closed libp2p.tcptransport: 323
    Opened stream.transport: 650
Closed stream.transport: 634
    Opened stream.server: 366
Closed stream.server: 359
    Opened ChronosStream: 591
Closed ChronosStream: 575

  [FAILED ] (  0.16s) register with unsupported keys

==============================
  /home/ben/IFT/nim-libp2p/tests/testnative 'AutoTLS ACME API::challenge completed successful'
------------------------------
    Opened LPChannel: 1283
Closed LPChannel: 1267
    Opened SecureConn: 453
Closed SecureConn: 437
    Opened libp2p.tcptransport: 330
Closed libp2p.tcptransport: 323
    Opened stream.transport: 650
Closed stream.transport: 634
    Opened stream.server: 366
Closed stream.server: 359
    Opened ChronosStream: 591
Closed ChronosStream: 575

  [FAILED ] (  0.30s) challenge completed successful

=======================================
  /home/ben/IFT/nim-libp2p/tests/testnative 'AutoTLS ACME API::challenge completed max retries reached'
---------------------------------------
    Opened LPChannel: 1283
Closed LPChannel: 1267
    Opened SecureConn: 453
Closed SecureConn: 437
    Opened libp2p.tcptransport: 330
Closed libp2p.tcptransport: 323
    Opened stream.transport: 650
Closed stream.transport: 634
    Opened stream.server: 366
Closed stream.server: 359
    Opened ChronosStream: 591
Closed ChronosStream: 575

  [FAILED ] (  0.83s) challenge completed max retries reached

===========================
  /home/ben/IFT/nim-libp2p/tests/testnative 'AutoTLS ACME API::challenge completed invalid'
---------------------------
    Opened LPChannel: 1283
Closed LPChannel: 1267
    Opened SecureConn: 453
Closed SecureConn: 437
    Opened libp2p.tcptransport: 330
Closed libp2p.tcptransport: 323
    Opened stream.transport: 650
Closed stream.transport: 634
    Opened stream.server: 366
Closed stream.server: 359
    Opened ChronosStream: 591
Closed ChronosStream: 575

  [FAILED ] (  0.34s) challenge completed invalid

===============================
  /home/ben/IFT/nim-libp2p/tests/testnative 'AutoTLS ACME API::finalize certificate successful'
-------------------------------
    Opened LPChannel: 1283
Closed LPChannel: 1267
    Opened SecureConn: 453
Closed SecureConn: 437
    Opened libp2p.tcptransport: 330
Closed libp2p.tcptransport: 323
    Opened stream.transport: 650
Closed stream.transport: 634
    Opened stream.server: 366
Closed stream.server: 359
    Opened ChronosStream: 591
Closed ChronosStream: 575

  [FAILED ] (  0.41s) finalize certificate successful

========================================
  /home/ben/IFT/nim-libp2p/tests/testnative 'AutoTLS ACME API::finalize certificate max retries reached'
----------------------------------------
    Opened LPChannel: 1283
Closed LPChannel: 1268
    Opened SecureConn: 453
Closed SecureConn: 437
    Opened libp2p.tcptransport: 330
Closed libp2p.tcptransport: 323
    Opened stream.transport: 650
Closed stream.transport: 634
    Opened stream.server: 366
Closed stream.server: 359
    Opened ChronosStream: 591
Closed ChronosStream: 575

  [FAILED ] (  1.60s) finalize certificate max retries reached

============================
  /home/ben/IFT/nim-libp2p/tests/testnative 'AutoTLS ACME API::finalize certificate invalid'
----------------------------
    Opened LPChannel: 1283
Closed LPChannel: 1268
    Opened SecureConn: 453
Closed SecureConn: 437
    Opened libp2p.tcptransport: 330
Closed libp2p.tcptransport: 323
    Opened stream.transport: 650
Closed stream.transport: 634
    Opened stream.server: 366
Closed stream.server: 359
    Opened ChronosStream: 591
Closed ChronosStream: 575

  [FAILED ] (  1.83s) finalize certificate invalid

=====================================
  /home/ben/IFT/nim-libp2p/tests/testnative 'AutoTLS ACME API::expect error on invalid JSON response'
-------------------------------------
    Opened LPChannel: 1283
Closed LPChannel: 1268
    Opened SecureConn: 453
Closed SecureConn: 437
    Opened libp2p.tcptransport: 330
Closed libp2p.tcptransport: 323
    Opened stream.transport: 650
Closed stream.transport: 634
    Opened stream.server: 366
Closed stream.server: 359
    Opened ChronosStream: 591
Closed ChronosStream: 575

  [FAILED ] (  1.46s) expect error on invalid JSON response

(7.81s)   AutoTLS ACME API

AutoTLS ACME Client FF (1.68s)
==============================================
  /home/ben/IFT/nim-libp2p/tests/testnative 'AutoTLS ACME Client::client registers new account when instantiated'
----------------------------------------------
    Opened LPChannel: 1283
Closed LPChannel: 1268
    Opened SecureConn: 453
Closed SecureConn: 437
    Opened libp2p.tcptransport: 330
Closed libp2p.tcptransport: 323
    Opened stream.transport: 650
Closed stream.transport: 634
    Opened stream.server: 366
Closed stream.server: 359
    Opened ChronosStream: 591
Closed ChronosStream: 575

  [FAILED ] (  0.58s) client registers new account when instantiated

==============================================================================
  /home/ben/IFT/nim-libp2p/tests/testnative 'AutoTLS ACME Client::getCertificate succeeds on sendChallengeCompleted but fails on requestFinalize'
------------------------------------------------------------------------------
    Opened LPChannel: 1283
Closed LPChannel: 1268
    Opened SecureConn: 453
Closed SecureConn: 437
    Opened libp2p.tcptransport: 330
Closed libp2p.tcptransport: 323
    Opened stream.transport: 650
Closed stream.transport: 634
    Opened stream.server: 366
Closed stream.server: 359
    Opened ChronosStream: 591
Closed ChronosStream: 575
```

Ben-PH: Thanks for the help in discord. When adding this to `testfindnode.nim`

```diff
 suite "KadDHT - FindNode":
+  asyncTeardown:
+    checkTrackers()
```

...we get this:

```
================
  /home/ben/IFT/nim-libp2p/tests/testnative 'KadDHT - FindNode::Simple find peer'
----------------
    Opened LPChannel: 24
Closed LPChannel: 17
    Opened SecureConn: 6
Closed SecureConn: 0
    Opened libp2p.tcptransport: 3
Closed libp2p.tcptransport: 0
    Opened stream.transport: 6
Closed stream.transport: 0
    Opened stream.server: 3
Closed stream.server: 0
    Opened ChronosStream: 6
Closed ChronosStream: 0

  [FAILED ] (  1.14s) Simple find peer

===============
  /home/ben/IFT/nim-libp2p/tests/testnative 'KadDHT - FindNode::Relay find peer'
---------------
    Opened LPChannel: 62
Closed LPChannel: 46
    Opened SecureConn: 16
Closed SecureConn: 0
    Opened libp2p.tcptransport: 7
Closed libp2p.tcptransport: 0
    Opened stream.transport: 16
Closed stream.transport: 0
    Opened stream.server: 7
Closed stream.server: 0
    Opened ChronosStream: 16
Closed ChronosStream: 0
```

...so the source of the error is unclosed connections, as opened must == closed.

tests/kademlia/testfindnode.nim
```diff
@@ -0,0 +1,136 @@
...
+
+    await broKad.bootstrap(@[parentSwitch.peerInfo])
+    # Bro and parent know each other
+    doAssert(countBucketEntries(broKad.rtable.buckets, parentKad.rtable.selfId) == 1)
+    doAssert(countBucketEntries(parentKad.rtable.buckets, broKad.rtable.selfId) == 1)
+
+    await sisKad.bootstrap(@[parentSwitch.peerInfo])
+
+    # Sis and parent know each other...
+    doAssert(countBucketEntries(sisKad.rtable.buckets, parentKad.rtable.selfId) == 1)
```
vladopajic, line 111: would be better if  `check` is used then `doAssert`. as we currently use that in tests.

```
check countBucketEntries(sisKad.rtable.buckets, parentKad.rtable.selfId) == 1
```

tests/testnative.nim
Ben-PH: Been looking at the unclosed connections issue.

Seems to be something to do with how the connections are handled within the switch/kad/mounting system

If I mount everything up, but don't hit `start, no errors

with 3 nodes in the swarm:

0: bootstrap target
1: bootstraps off 0
2: dootstraps off 0

error with `start`:

```
    Opened libp2p.tcptransport: 3
Closed libp2p.tcptransport: 0
    Opened stream.server: 3
Closed stream.server: 0
```

error following bootstraps:

```
    Opened LPChannel: 16
Closed LPChannel: 11
    Opened SecureConn: 4
Closed SecureConn: 0
    Opened libp2p.tcptransport: 3
Closed libp2p.tcptransport: 0
    Opened stream.transport: 4
Closed stream.transport: 0
    Opened stream.server: 3
Closed stream.server: 0
    Opened ChronosStream: 4
Closed ChronosStream: 0
```

following node 1 finding node 2:

```
    Opened LPChannel: 24
Closed LPChannel: 17
    Opened SecureConn: 6
Closed SecureConn: 0
    Opened libp2p.tcptransport: 3
Closed libp2p.tcptransport: 0
    Opened stream.transport: 6
Closed stream.transport: 0
    Opened stream.server: 3
Closed stream.server: 0
    Opened ChronosStream: 6
Closed ChronosStream: 0
```
stop a switch:

```
/home/ben/IFT/nim-libp2p/libp2p/switch.nim(339) stop
/home/ben/IFT/nim-libp2p/libp2p/multistream.nim(273) stop
/home/ben/.nimble/pkgs2/chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a/chronos/futures.nim(208) finished
/home/ben/.nimble/pkgs2/chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a/chronos/futures.nim(199) state
Error: execution of an external program failed: '/home/ben/IFT/nim-libp2p/tests/testnative'
```

stop a kad:

```
================
  /home/ben/IFT/nim-libp2p/tests/testnative 'KadDHT - FindNode::Simple find peer'
----------------
/home/ben/.nimble/pkgs2/unittest2-0.2.4-746106a4dfefffce497f1693733f1c1513b5c62c/unittest2.nim(1170) testfindnode
/home/ben/.nimble/pkgs2/unittest2-0.2.4-746106a4dfefffce497f1693733f1c1513b5c62c/unittest2.nim(1097) runDirect
/home/ben/IFT/nim-libp2p/tests/utils/async_tests.nim(25) runTest`gensym8
/home/ben/.nimble/pkgs2/chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a/chronos/internal/asyncfutures.nim(649) waitFor
/home/ben/.nimble/pkgs2/chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a/chronos/internal/asyncfutures.nim(624) pollFor
/home/ben/.nimble/pkgs2/chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a/chronos/internal/asyncengine.nim(150) poll
/nix/store/gdxcfavh457rc7hlwn4479bjrmgagkfv-nim-unwrapped-2.2.4/nim/lib/std/assertions.nim(36) raiseAssert
/nix/store/gdxcfavh457rc7hlwn4479bjrmgagkfv-nim-unwrapped-2.2.4/nim/lib/system/fatal.nim(53) sysFatal

    Unhandled defect: Async procedure (testfindnode.nim(37)    anonymous) yielded `nil`, are you await'ing a `nil` Future? [AssertionDefect]
    Opened LPChannel: 24
Closed LPChannel: 17
    Opened SecureConn: 6
Closed SecureConn: 0
    Opened libp2p.tcptransport: 3
Closed libp2p.tcptransport: 0
    Opened stream.transport: 6
Closed stream.transport: 0
    Opened stream.server: 3
Closed stream.server: 0
    Opened ChronosStream: 6
Closed ChronosStream: 0
```

Ben-PH: resolved with [3d8b39a](https://github.com/vacp2p/nim-libp2p/pull/1324/commits/3d8b39ae25622417d70ebf08e710d8f6766dfb18)

thanks @richard-ramos

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: added to follow-ups

libp2p/protocols/kademlia/lookupstate.nim
Ben-PH: ...unfortunately, that just doesn't work. 

I'm not a fan of templates. I personally prefer the code to be a bit more verbose if that's the price to pay to avoid introducing implicitness.

tests/kademlia/testfindnode.nim
Ben-PH: I looked up `check`, and it doesn't seem to come with the ability to provide failure context? I don't use it extensively in this PR, but I do use them, and they can be quite useful for quickly getting to the bottom of where, why and how of a failure fore the next person to come along.

libp2p/protocols/kademlia/lookupstate.nim
Ben-PH: My position is that we should trust the compiler. If we cannot afford to trust the compiler because it's massively performance critical, then the job should be much more involved (benchmarking, profiling, CI to detect regressions, etc). If the quality of the compiler is that we cannot trust it, then it's an upstream issue: the nim-language itself needs to do better.

The inlining you suggest makes it more readable to the compiler, but if we are micro-optimising everywhere, it risks a death by a thousand cuts in terms of readability and bitrot.

I don't hold this position strongly though. Having shared my position, I'm happy to make the change if you feel it's necessary.

Ben-PH:
Tests are passing, and CI is happy. I've addressed the feedback, and unilaterally closed those that I feel allow for it, after giving a few days.

The tests are somewhat minimal: This is not suitable for production, but enough for downstream users to play around with.

Discovery interface and put-value are in initial, limited working stages.

This protocol is a bit of a chicken-egg problem: the various components (put/get value, discovery, testing of find-node, etc) have elements of mutual dependency when it comes to each coming up to deployment-worthy standard.

For merging, I will do a squash rebase, specifying richard as an author

tests/kademlia/testfindnode.nim
vladopajic: what exactly are you considering under failure context?
if `doAssert` is better then `check` then probably we should do that across all other tests?

libp2p/protocols/kademlia/lookupstate.nim
vladopajic: no i don't really feel it's necessary. 

but i also wouldn't trust that compiler would do such optimization, because this code kinda tells compiler evaluate this and assign it to variable.

tests/kademlia/testfindnode.nim
Ben-PH: So for example, if the assert is in a loop, it can help to output the index.

You can also contextualise the failure in useful ways.

> then probably we should do that across all other tests?

Based on my reading of `check` documentation/examples, and what code I've seen using it, I think using `doAssert` in the future, and incrementally moving from `check`, is the to go.

Ideally, an assertion like so:

```nim
let foo = 5
let bar = 6
doAssertEq(foo, bar, "some string")
```
would output something like:

```
equality assertion failure: LHS: 'foo' = 5, RHS: 'bar' = 6; "some string"
```

or if its `doAssert(foo == bar, "..."`

```
assertion failure: 'foo == bar'; "..."
```

Key elements:

 - surfaces the syntax that assertion failed on
 - provides the means to contextualise the failure
 - can assert with common checks such as equality, ordering, inequality, etc, and report failures in more meaningful way.
 
 That's all outside the scope of this PR and thread, but I think it contextualises my thoughts on moving on from `check`. Maybe `check` does these things but I'm not aware of it, for example.

gmelodie: For simplicity, I think we should go with `check` since almost the entire codebase uses this. We can discuss `check` vs `doAssert` at another time.

PR #1341, opened 2025-04-22
chore: fix warnings

narimiran:
Two warnings are "left as an exercise to the reader":

- _Warning: Please use cancelSoon() or cancelAndWait() instead; cancel is deprecated [Deprecated]_
- not using `stew/shims/net`, which is removed from the latest version of `stew`

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -4,7 +4,6 @@ if dirExists("nimbledeps/pkgs"):
 if dirExists("nimbledeps/pkgs2"):
   switch("NimblePath", "nimbledeps/pkgs2")
 
-switch("warningAsError", "UnusedImport:on")
 switch("warning", "CaseTransition:off")
 switch("warning", "ObservableStores:off")
 switch("warning", "LockLevel:off")
--- a/config.nims
+++ b/config.nims
@@ -4,6 +4,7 @@ if dirExists("nimbledeps/pkgs"):
 if dirExists("nimbledeps/pkgs2"):
   switch("NimblePath", "nimbledeps/pkgs2")
 
+switch("warningAsError", "UnusedImport:on")
 switch("warning", "CaseTransition:off")
 switch("warning", "ObservableStores:off")
 switch("warning", "LockLevel:off")
--- a/libp2p.nimble
+++ b/libp2p.nimble
@@ -10,7 +10,7 @@ skipDirs = @["tests", "examples", "Nim", "tools", "scripts", "docs"]
 requires "nim >= 1.6.0",
   "nimcrypto >= 0.6.0 & < 0.7.0", "dnsclient >= 0.3.0 & < 0.4.0", "bearssl >= 0.2.5",
   "chronicles >= 0.10.2", "chronos >= 4.0.3", "metrics", "secp256k1", "stew#head",
-  "websock", "unittest2",
+  "websock", "unittest2", "results",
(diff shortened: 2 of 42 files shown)


tersec:
nph apparently still wants
```nim
diff --git a/tests/commontransport.nim b/tests/commontransport.nim
index 45cf17b..cd02b18 100644
--- a/tests/commontransport.nim
+++ b/tests/commontransport.nim
@@ -2,7 +2,8 @@
 
 import chronos, results, stew/byteutils
 import
-  ../libp2p/[stream/connection, transports/transport, upgrademngrs/upgrade, multiaddress]
+  ../libp2p/
+    [stream/connection, transports/transport, upgrademngrs/upgrade, multiaddress]
 
 import ./helpers
 
```

tersec (approved):

narimiran:
> also it would be nice to reduce scope of PRs to address each group of warnings separately.

That's why I made separate commits for separate things, so you can review each separately if you wish.

Moving each commit to a different PR means I'll have to wait for 4 approvals (ok, no big deal) and the CIs will run 4 times, wasting resources.  
But if that's a general consensus (I'll wait for somebody else to chime in), I'll do it that way....

vladopajic:
> > also it would be nice to reduce scope of PRs to address each group of warnings separately.
> 
> That's why I made separate commits for separate things, so you can review each separately if you wish.
> 
> Moving each commit to a different PR means I'll have to wait for 4 approvals (ok, no big deal) and the CIs will run 4 times, wasting resources. But if that's a general consensus (I'll wait for somebody else to chime in), I'll do it that way....

i was thinking, if you are planning to fix warnings, then better better choose which warnings are you fixing and fix those to fullest. fixing some warning here and there in bulk can causing conflict with other efforts - you wouldn't believe but i started to  work on #1344 simultaneously with your PR without knowing you started to work on this.

since your work is mainly fixing unused import i would suggest to confine changes only to this effort. if there are some changes more then this, leave it as is, but confine changes to only unused import and fix them to the fullest. 

hint: run `nimble test` to check if compile is passing without errors.

narimiran:
> Error: imported and not used: 'chronicles' [UnusedImport]

Hello darkness my old friend.

This was also shown in `nwaku`, so I removed the import, the tests were green, and then few days later someone from their team noticed that in fact those chronicles imports should be kept.

TL;DR: I learned my lesson and won't be removing it.

narimiran:
> since your work is mainly fixing unused import 

It isn't.
My main focus was to make a switch from `stew/results` to `results`, the other fixes were "while I'm in there".

vladopajic:
ah okay...

could you revert [address the review comment](https://github.com/vacp2p/nim-libp2p/pull/1341/commits/f433e0e2acb91624160fde126ced34ada2761f1f)`, so we can merge this PR?

vladopajic (approved):

narimiran:
> i started to work on https://github.com/vacp2p/nim-libp2p/pull/1344 simultaneously with your PR without knowing you started to work on this.

If you want something else to work on (which I won't do in parallel :)), see the initial comment:

> Two warnings are "left as an exercise to the reader":
> 
> * _Warning: Please use cancelSoon() or cancelAndWait() instead; cancel is deprecated [Deprecated]_
> * not using `stew/shims/net`, which is removed from the latest version of `stew`

PR #1355, opened 2025-05-01
feat(cert): add certificate signing request (CSR) generation

gmelodie:
Add certificate signing request generation in `transports/tls/certificate.c`

### DIFF
--- a/libp2p/transports/tls/certificate.c
+++ b/libp2p/transports/tls/certificate.c
@@ -960,3 +960,132 @@ void cert_free_key(cert_key_t key) {
   EVP_PKEY_free(k->pkey);
   free(k);
 }
+
+// Function to check if a Common Name is correct
+// each label should have <= 63 characters
+// the whole CN should have <= 253 characters
+cert_error_t check_cn(const char *cn) {
+  cert_error_t ret_code = CERT_SUCCESS;
+
+  if (!cn || strlen(cn) == 0) {
+    return CERT_ERROR_CN_EMPTY;
+  }
+  if (strlen(cn) > 253) {
+    return CERT_ERROR_CN_TOO_LONG;
+  }
+
+  char *cn_copy = strdup(cn);
+  char *cn_copy_orig = cn_copy;
+
+  // trim trailing dot if any before checking
+  size_t len = strlen(cn_copy);
+  if (len > 0 && cn_copy[len - 1] == '.') {
+    cn_copy[len - 1] = '\0';
+  }
+
+  char *label;
+  char *last = NULL;
+  char *ptr = cn_copy;
+
+  while ((label = strtok(ptr, ".")) != NULL) {
+    if (last && last + strlen(last) + 1 != label) {
+      // empty label (e.g., "example..com")
+      ret_code = CERT_ERROR_CN_EMPTY_LABEL;
+      break;
+    }
+    if (strlen(label) > 63) {
+      ret_code = CERT_ERROR_CN_LABEL_TOO_LONG;
+      break;
+    }
+    last = label;
+    ptr = NULL;
+  }
+
+  free(cn_copy_orig);
+  return ret_code;
+}
+
+cert_error_t cert_signing_req(const char *cn, cert_key_t key, cert_buffer **csr_buffer) {
+  cert_error_t ret_code = CERT_SUCCESS;
+  X509_REQ *x509_req = NULL;
+  X509_NAME *name = NULL;
+  X509_EXTENSION *ext = NULL;
+  X509V3_CTX ctx;
+  STACK_OF(X509_EXTENSION) *exts = NULL;
+  unsigned char *der = NULL;
+  size_t der_len = 0;
+
+  ret_code = check_cn(cn);
+  if (ret_code != CERT_SUCCESS) {
+    goto cleanup;
+  }
+
+  if (!key || !(key->pkey)) {
+    ret_code = CERT_ERROR_NO_PUBKEY;
+    goto cleanup;
+  }
+  EVP_PKEY *pkey = key->pkey;
+
+  x509_req = X509_REQ_new();
+  if (!x509_req) {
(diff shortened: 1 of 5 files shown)


libp2p/transports/tls/certificate.c
```diff
@@ -959,4 +959,87 @@ void cert_free_key(cert_key_t key) {
...
+
+  der_len = i2d_X509_REQ(x509_req, &der);
+  if (der_len < 0) {
+    ret_code = CERT_ERROR_X509_REQ_DER;
+    goto cleanup;
+  }
+
+  ret_code = init_cert_buffer(csr_buffer, der, der_len);
+  if (ret_code < 0) {
+    goto cleanup;
```
vladopajic, line 1027: there is no specific `ret_code` for this case?

libp2p/transports/tls/certificate_ffi.nim
```diff
@@ -79,3 +79,7 @@ proc cert_free_buffer*(
 proc cert_free_parsed*(
   cert: ptr cert_parsed
 ): void {.cdecl, importc: "cert_free_parsed".}
+
+proc cert_signing_req*(
```
vladopajic, line 83: this logic is lacking unit tests (we can't really know that it works)

vladopajic (approved):

libp2p/transports/tls/certificate.c
gmelodie: Yes there is, but it comes from `init_cert_buffer`. Overwriting it would make the error less nuanced since `init_cert_buffer` can fail for many reasons, and if we instead were to setup a new error it'd always look like "init_cert_buffer failed"

vladopajic: got it...

it is also bit strange that this c function now returns error codes from two spaces: 1) our `nim-libp2p` values and 2) `init_cert_buffer` values. and if second space overlaps with first space of error code values, then we can only wish luck for anyone down the road if they are debugging with this case.

gmelodie: `init_cert_buffer` returns the same types of errors that our function does (the ones defined in `certificate.h`)

vladopajic: oh right, then we are all good here :)

vladopajic (approved):

libp2p/transports/tls/certificate.c
```diff
@@ -959,4 +959,87 @@ void cert_free_key(cert_key_t key) {
...
+  }
+
+  if (!X509_REQ_set_pubkey(x509_req, pkey)) {
+    ret_code = CERT_ERROR_PUBKEY_SET;
+    goto cleanup;
+  }
+
+  // Build SAN extension
+  X509V3_set_ctx(&ctx, NULL, NULL, x509_req, NULL, 0);
+  char san_str[256];
```
richard-ramos, line 994: DNS spec (RFC 1035) limits domain names to 255 bytes in wire format, including label length prefixes and the terminating null byte. in textual form (like in a SAN), this means the domain name can be at most 253 characters. so for a SAN entry like "DNS:domain", the domain can be up to 253 bytes, and the full SAN string including "DNS:" and the null terminator needs a buffer of at least 258 bytes,  so it's probably a good idea to increase the buffer size to 258 and add a condition here to validate the length, even tho we might never run into this error, but at least will make the function correct:
Maybe something like this?
```c
if (strlen(cn) > 253) {   // "DNS:" + domain + null terminator = ~258 max
  ret_code = CERT_ERROR_CN_TOO_LONG;
  goto cleanup;
}
```

libp2p/transports/tls/certificate.c
```diff
@@ -959,4 +959,87 @@ void cert_free_key(cert_key_t key) {
...
+    goto cleanup;
+  }
+  EVP_PKEY *pkey = key->pkey;
+
+  x509_req = X509_REQ_new();
+  if (!x509_req) {
+    ret_code = CERT_ERROR_X509_REQ_GEN;
+    goto cleanup;
+  }
+
```
richard-ramos, line 986: Quickly reading the RFC 6125, even tho the CN is not required **for validation**, it still should exist. Interestingly enough, you can just truncate the content for cases in which the cn exceeds 64 bytes.

Maybe we can  add something like this?
```cpp
char truncated_cn[65]; // 64 + null terminator
strncpy(truncated_cn, cn, 64);
truncated_cn[64] = '\0'; // ensure null termination

if (!X509_NAME_add_entry_by_txt(name, "CN", MBSTRING_ASC,
        (unsigned char *)truncated_cn, -1, -1, 0)) {
  ret_code = CERT_ERROR_CN;
  goto cleanup;
}
```

libp2p/transports/tls/certificate.c
```diff
@@ -959,4 +959,87 @@ void cert_free_key(cert_key_t key) {
...
+    sk_X509_EXTENSION_pop_free(exts, X509_EXTENSION_free);
+  if (x509_req)
+    X509_REQ_free(x509_req);
+  if (der)
+    OPENSSL_free(der);
+  if (ret_code != CERT_SUCCESS && csr_buffer && *csr_buffer) {
+    if ((*csr_buffer)->data)
+      free((*csr_buffer)->data);
+    free(*csr_buffer);
+    *csr_buffer = NULL;
```
richard-ramos, line 1041: There are some other instances of code similar to the one you used here, (i gotta fix that), but ideally we should use `cert_free_buffer`:
```c
if (ret_code != CERT_SUCCESS)) {
  cert_free_buffer(csr_buffer);
  csr_buffer = NULL;
}
```

libp2p/transports/tls/certificate.c
gmelodie: I found that Let's Encrypt misbehaves when CN is set.

richard-ramos (approved):

Issue #1325, opened 2025-04-14
AutoTLS - `vac:p2p:ift:2025q2-nimlibp2p-autotls`


richard-ramos:
Implement a process that automates the issuance of Let’s Encrypt wilcard TLS certificates for nim-libp2p nodes. This would enable reliable browser-based clients to connect to libp2p nodes, regardless of the network topology which improves resilience and opens up a window to more capable browser clients. This task is limited to the client functionality; Server functionality could be implemented on a separate commitment. [AutoTLS](https://blog.libp2p.io/autotls/)

https://roadmap.vac.dev/p2p/ift/2025q2-nimlibp2p-autotls

richard-ramos:
Task's been succesfully implemented.

PR #1358, opened 2025-05-07
feat(autotls): add autotls support

gmelodie:
Adding [AutoTLS](https://blog.libp2p.io/autotls/) support (addresses #1325)
- `AutoTLSManager` is responsible for always keeping a valid certificate signed by [Let's Encrypt](https://letsencrypt.org/)
- ACME functionality (`libp2p/autotls/acme.nim`): send signed requests to ACME server to request and finalize challenges
- [PeerID Authentication](https://github.com/libp2p/specs/blob/master/http/peer-id-auth.md) (`libp2p/autotls/peeridauth.nim`): authenticate with AutoTLS broker and send challenge
- DNS: check for RRs (`TXT _acme-challenge.{peerID}.libp2p.direct` and `A {dashed-ip-addr}.{peerID}.libp2p.direct`) to appear
- Certificates (`libp2p/transports/tls/challenge{.c, .h, .nim, _ffi.nim}`): generate certificate singing requests (CSR) and encode them for sending to ACME server as `finalize` payload (this has been merged in https://github.com/vacp2p/nim-libp2p/pull/1355)

### DIFF
--- a/libp2p/autotls/acme.nim
+++ b/libp2p/autotls/acme.nim
@@ -10,7 +10,8 @@ import ../transports/tls/certificate
 
 # TODO: change staging to actual url
 const
-  LetsEncryptURL = "https://acme-staging-v02.api.letsencrypt.org"
+  LetsEncryptURL = "https://acme-v02.api.letsencrypt.org"
+  LetsEncryptURLStaging = "https://acme-staging-v02.api.letsencrypt.org"
   Alg = "RS256"
 
 type ACMEAccount* = object
@@ -20,6 +21,7 @@ type ACMEAccount* = object
   session*: HttpSessionRef
   kid*: Opt[string]
   directory: JsonNode
+  acmeServerURL: string
 
 proc new*(
     T: typedesc[ACMEAccount],
@@ -27,13 +29,12 @@ proc new*(
     status: Opt[string] = Opt.none(string),
     contact: Opt[seq[string]] = Opt.none(seq[string]),
     kid: Opt[string] = Opt.none(string),
+    acmeServerURL: string = LetsEncryptURL,
 ): Future[ref ACMEAccount] {.async: (raises: [ACMEError, CancelledError]).} =
   try:
     let session = HttpSessionRef.new()
-    let directoryResponse = await HttpClientRequestRef
-    .get(session, LetsEncryptURL & "/directory")
-    .get()
-    .send()
+    let directoryResponse =
+      await HttpClientRequestRef.get(session, acmeServerURL & "/directory").get().send()
     let directory = bytesToString(await directoryResponse.getBodyBytes()).parseJson()
 
     let acc = new(ACMEAccount)
@@ -43,6 +44,7 @@ proc new*(
     acc.key = key
     acc.session = session
     acc.directory = directory
+    acc.acmeServerURL = acmeServerURL
     return acc
   except HttpError:
     raise newException(ACMEError, "Failed to connect to ACME server")
--- a/libp2p/autotls/acme.nim
+++ b/libp2p/autotls/acme.nim
@@ -16,6 +16,7 @@ type ACMEAccount* = object
   status*: Opt[string]
   contact*: Opt[seq[string]]
   key*: KeyPair
+  session*: HttpSessionRef
   kid*: Opt[string]
   directory: JsonNode
 
@@ -27,8 +28,9 @@ proc new*(
     kid: Opt[string] = Opt.none(string),
 ): Future[ref ACMEAccount] {.async: (raises: [ACMEError, CancelledError]).} =
   try:
+    let session = HttpSessionRef.new()
     let directoryResponse = await HttpClientRequestRef
-    .get(HttpSessionRef.new(), LetsEncryptURL & "/directory")
+    .get(session, LetsEncryptURL & "/directory")
     .get()
     .send()
     let directory = bytesToString(await directoryResponse.getBodyBytes()).parseJson()
@@ -38,6 +40,7 @@ proc new*(
     acc.contact = contact
     acc.kid = kid
     acc.key = key
+    acc.session = session
     acc.directory = directory
     return acc
   except HttpError:
@@ -59,8 +62,7 @@ proc newNonce(
 ): Future[string] {.async: (raises: [ACMEError, CancelledError]).} =
   try:
     let nonceURL = self.directory.getJSONField("newNonce").getStr
-    let resp =
-      await HttpClientRequestRef.get(HttpSessionRef.new(), nonceURL).get().send()
+    let resp = await HttpClientRequestRef.get(self.session, nonceURL).get().send()
     return resp.headers.getString("replay-nonce")
   except HttpError:
     raise newException(ACMEError, "Failed to request new nonce from ACME server")
@@ -103,7 +105,7 @@ proc signedAcmeRequest(
   try:
     let response = await HttpClientRequestRef
     .post(
-      HttpSessionRef.new(),
+      self.session,
       url,
       body = $body,
       headers = [("Content-Type", "application/jose+json")],
@@ -155,7 +157,7 @@ proc requestChallenge*(
   let authzResponseBody =
     try:
       let authzResponse =
-        await HttpClientRequestRef.get(HttpSessionRef.new(), authzURL).get().send()
+        await HttpClientRequestRef.get(self.session, authzURL).get().send()
       await authzResponse.getParsedResponseBody()
     except Exception as e:
       raise newException(ACMEError, "failed to request challenge: " & e.msg)
@@ -196,7 +198,7 @@ proc notifyChallengeCompleted*(
     var checkResponseBody: JsonNode
     try:
       let checkResponse =
-        await HttpClientRequestRef.get(HttpSessionRef.new(), checkURL).get().send()
+        await HttpClientRequestRef.get(self.session, checkURL).get().send()
       checkResponseBody = bytesToString(await checkResponse.getBodyBytes()).parseJson()
     except HttpError:
       raise newException(ACMEError, "Failed to connect to ACME server")
@@ -247,7 +249,7 @@ proc finalizeCertificate*(
     var finalizedResponseBody: JsonNode
     try:
       let finalizedResponse =
-        await HttpClientRequestRef.get(HttpSessionRef.new(), finalizeURL).get().send()
+        await HttpClientRequestRef.get(self.session, finalizeURL).get().send()
       finalizedResponseBody =
         bytesToString(await finalizedResponse.getBodyBytes()).parseJson()
     except HttpError:
@@ -272,17 +274,15 @@ proc downloadCertificate*(
 ): Future[string] {.async: (raises: [ACMEError, CancelledError]).} =
   try:
     let downloadResponse =
-      await HttpClientRequestRef.get(HttpSessionRef.new(), orderURL).get().send()
+      await HttpClientRequestRef.get(self.session, orderURL).get().send()
     if downloadResponse.status != 200:
       raise newException(ACMEError, "Failed to download certificate")
     let certificateDownloadURL = bytesToString(await downloadResponse.getBodyBytes())
       .parseJson()
       .getJSONField("certificate").getStr
 
-    let certificateResponse = await HttpClientRequestRef
-    .get(HttpSessionRef.new(), certificateDownloadURL)
-    .get()
-    .send()
+    let certificateResponse =
+      await HttpClientRequestRef.get(self.session, certificateDownloadURL).get().send()
     let rawCertificate = bytesToString(await certificateResponse.getBodyBytes())
     return rawCertificate
   except HttpError:
--- a/libp2p/autotls/acme.nim
+++ b/libp2p/autotls/acme.nim
@@ -8,11 +8,15 @@ import ../crypto/rsa
 import ../transports/tls/certificate_ffi
 import ../transports/tls/certificate
 
-# TODO: change staging to actual url
 const
   LetsEncryptURL* = "https://acme-v02.api.letsencrypt.org"
   LetsEncryptURLStaging* = "https://acme-staging-v02.api.letsencrypt.org"
   Alg = "RS256"
+  DefaultChalCompletedRetries = 10
+  DefaultChalCompletedRetryTime = 1.seconds
+  DefaultFinalizeRetries = 10
+  DefaultFinalizeRetryTimes = 1.seconds
+  DefaultRandStringSize = 256
 
 type ACMEAccount* = object
   status*: Opt[string]
@@ -103,7 +107,7 @@ proc signedAcmeRequest(
     token.sign(pemPrivKey)
     body = token.toFlattenedJson()
   except CatchableError as e:
-    raise newException(ACMEError, "could not create JWT: " & e.msg)
+    raise newException(ACMEError, "Failed to create JWT: " & e.msg)
   try:
     let response = await HttpClientRequestRef
     .post(
@@ -128,7 +132,7 @@ proc register*(self: ref ACMEAccount) {.async: (raises: [ACMEError, CancelledErr
   let newAccountURL = self.directory.getJSONField("newAccount").getStr
   let response = await self.signedAcmeRequest(newAccountURL, payload, needsJwk = true)
   let jsonResponseBody = await response.getParsedResponseBody()
-  if response.status != 201:
+  if response.status != HttpCreated:
     raise newException(
       ACMEError, "Unable to register with ACME server: " & $jsonResponseBody
     )
@@ -162,7 +166,7 @@ proc requestChallenge*(
         await HttpClientRequestRef.get(self.session, authzURL).get().send()
       await authzResponse.getParsedResponseBody()
     except CatchableError as e:
-      raise newException(ACMEError, "failed to request challenge: " & e.msg)
+      raise newException(ACMEError, "Failed to request challenge: " & e.msg)
 
   let challenges = authzResponseBody.getJSONField("challenges")
   var dns01: JsonNode = nil
@@ -176,11 +180,11 @@ proc requestChallenge*(
   return (dns01, finalizeURL, orderURL)
 
 proc notifyChallengeCompleted*(
-    self: ref ACMEAccount, chalURL: string, retries: int = 5
+    self: ref ACMEAccount, chalURL: string, retries: int = DefaultChalCompletedRetries
 ): Future[bool] {.async: (raises: [ACMEError, CancelledError]).} =
   let emptyPayload = newJObject()
   let completedResponse = await self.signedAcmeRequest(chalURL, emptyPayload)
-  if completedResponse.status != 200:
+  if completedResponse.status != HttpOk:
     return false
 
   var completedResponseBody: JsonNode
@@ -201,21 +205,22 @@ proc notifyChallengeCompleted*(
     try:
       let checkResponse =
         await HttpClientRequestRef.get(self.session, checkURL).get().send()
+      if checkResponse.status != HttpOk:
+        raise newException(ACMEError, "Failed to check challenge completion")
       checkResponseBody = bytesToString(await checkResponse.getBodyBytes()).parseJson()
     except HttpError:
       raise newException(ACMEError, "Failed to connect to ACME server")
-    except CatchableError as e:
+    except CatchableError as exc:
       raise newException(
-        ACMEError, "Unexpected error while signaling challenge completion: " & e.msg
+        ACMEError, "Unexpected error while signaling challenge completion", exc
       )
     case checkResponseBody.getJSONField("status").getStr
     of "pending":
-      discard # try again
+      await sleepAsync(DefaultChalCompletedRetryTime) # try again after some delay
     of "valid":
       return true
     else:
       return false
-    await sleepAsync(1.seconds)
 
   return false
 
@@ -224,20 +229,16 @@ proc finalizeCertificate*(
     domain: string,
     finalizeURL: string,
     orderURL: string,
-    retries: int = 5,
+    retries: int = DefaultFinalizeRetries,
 ): Future[bool] {.async: (raises: [ACMEError, CancelledError]).} =
   var certKey: cert_key_t
   var certCtx: cert_context_t
   var derCSR: ptr cert_buffer = nil
 
-  # TODO: use secure method
-  var randomCstring: cstring
-  try:
-    randomCstring = urandomToCString(256)
-  except:
-    raise newException(ACMEError, "Failed to generate random string")
-
-  if cert_init_drbg(randomCstring, 256, certCtx.addr) != CERT_SUCCESS:
+  let personalizationStr = "libp2p_autotls"
+  if cert_init_drbg(
+    personalizationStr.cstring, personalizationStr.len.csize_t, certCtx.addr
+  ) != CERT_SUCCESS:
     raise newException(ACMEError, "Failed to initialize certCtx")
   if cert_generate_key(certCtx, certKey.addr) != CERT_SUCCESS:
     raise newException(ACMEError, "Failed to generate cert key")
@@ -250,7 +251,7 @@ proc finalizeCertificate*(
 
   # send finalize request
   let finalizedResponse = await self.signedAcmeRequest(finalizeURL, payload)
-  if finalizedResponse.status != 200:
+  if finalizedResponse.status != HttpOk:
     raise newException(ACMEError, "Failed to request cert finalization")
 
   # keep checking order until it's finalized
@@ -271,10 +272,9 @@ proc finalizeCertificate*(
     of "valid":
       return true
     of "processing":
-      discard # keep trying
+      await sleepAsync(DefaultFinalizeRetryTimes) # try again after some delay
     else:
       return false
-    await sleepAsync(1.seconds)
 
   return false
 
@@ -285,7 +285,7 @@ proc downloadCertificate*(
     let downloadResponse =
       await HttpClientRequestRef.get(self.session, orderURL).get().send()
 
-    if downloadResponse.status != 200:
+    if downloadResponse.status != HttpOk:
       raise newException(ACMEError, "Failed to download certificate")
 
     let certificateInfoBody =
--- a/libp2p/autotls/acme.nim
+++ b/libp2p/autotls/acme.nim
@@ -91,19 +91,15 @@ proc signedAcmeRequest(
 ): Future[HttpClientResponseRef] {.async: (raises: [ACMEError, CancelledError]).} =
   let acmeHeader = await self.acmeHeader(url, needsJwk)
   var token: JWT
+  var body: JsonNode
   try:
     token = toJWT(%*{"header": acmeHeader, "claims": payload})
     let derPrivKey = self.key.seckey.rsakey.getBytes.get
     let pemPrivKey: string = pemEncode(derPrivKey, "PRIVATE KEY")
     token.sign(pemPrivKey)
+    body = token.toFlattenedJson()
   except Exception as e:
     raise newException(ACMEError, "could not create JWT: " & e.msg)
-  let body =
-    %*{
-      "payload": token.claims.toBase64,
-      "protected": token.header.toBase64,
-      "signature": base64UrlEncode(token.signature),
-    }
   try:
     let response = await HttpClientRequestRef
     .post(
--- a/libp2p/autotls/acme.nim
+++ b/libp2p/autotls/acme.nim
@@ -1,4 +1,4 @@
-import options, json, base64
+import options, json, base64, sequtils
 from times import DateTime, parse
 import chronos/apps/http/httpclient, jwt, results, bearssl/pem
 
@@ -24,7 +24,9 @@ type ACMEAccount* = object
   key*: KeyPair
   session*: HttpSessionRef
   kid*: Opt[string]
-  directory: JsonNode
+  newNonceURL: string
+  newAccountURL: string
+  newOrderURL: string
   acmeServerURL: string
 
 proc new*(
@@ -56,7 +58,9 @@ proc new*(
   acc.kid = kid
   acc.key = key
   acc.session = session
-  acc.directory = directory
+  acc.newNonceURL = directory.getJSONField("newNonce").getStr
+  acc.newAccountURL = directory.getJSONField("newAccount").getStr
+  acc.newOrderURL = directory.getJSONField("newOrder").getStr
   acc.acmeServerURL = acmeServerURL
   return acc
 
@@ -64,8 +68,8 @@ proc newNonce(
     self: ref ACMEAccount
 ): Future[string] {.async: (raises: [ACMEError, CancelledError]).} =
   try:
-    let nonceURL = self.directory.getJSONField("newNonce").getStr
-    let resp = await HttpClientRequestRef.get(self.session, nonceURL).get().send()
+    let resp =
+      await HttpClientRequestRef.get(self.session, self.newNonceURL).get().send()
     return resp.headers.getString("Replay-Nonce")
   except HttpError as exc:
     raise newException(ACMEError, "Failed to request new nonce from ACME server", exc)
@@ -125,8 +129,8 @@ proc register*(self: ref ACMEAccount) {.async: (raises: [ACMEError, CancelledErr
 
   let payload = %*{"termsOfServiceAgreed": true}
 
-  let newAccountURL = self.directory.getJSONField("newAccount").getStr
-  let response = await self.signedAcmeRequest(newAccountURL, payload, needsJwk = true)
+  let response =
+    await self.signedAcmeRequest(self.newAccountURL, payload, needsJwk = true)
   let jsonResponseBody = await response.getParsedResponseBody()
   if response.status != HttpCreated:
     raise newException(
@@ -139,15 +143,11 @@ proc requestChallenge*(
     self: ref ACMEAccount, domains: seq[string]
 ): Future[(JsonNode, string, string)] {.async: (raises: [ACMEError, CancelledError]).} =
   # request challenge from ACME server
-  var identifiers: seq[JsonNode]
-
-  for domain in domains:
-    identifiers.add(%*{"type": "dns", "value": domain})
+  var identifiers: seq[JsonNode] = domains.mapIt(%*{"type": "dns", "value": it})
 
   let orderPayload = %*{"identifiers": identifiers}
-  let newOrderURL = self.directory.getJSONField("newOrder").getStr
 
-  let challengeResponse = await self.signedAcmeRequest(newOrderURL, orderPayload)
+  let challengeResponse = await self.signedAcmeRequest(self.newOrderURL, orderPayload)
   let challengeResponseBody = await challengeResponse.getParsedResponseBody()
   let orderURL = challengeResponse.headers.getString("location")
   if orderURL == "":
@@ -177,11 +177,15 @@ proc requestChallenge*(
 
 proc notifyChallengeCompleted*(
     self: ref ACMEAccount, chalURL: string, retries: int = DefaultChalCompletedRetries
-): Future[bool] {.async: (raises: [ACMEError, CancelledError]).} =
+): Future[void] {.async: (raises: [ACMEError, CancelledError]).} =
   let emptyPayload = newJObject()
   let completedResponse = await self.signedAcmeRequest(chalURL, emptyPayload)
   if completedResponse.status != HttpOk:
-    return false
+    raise newException(
+      ACMEError,
+      "Failed got HTTP status code " & $completedResponse.status &
+        " while sending completed message to ACME server",
+    )
 
   var completedResponseBody: JsonNode
   try:
@@ -210,7 +214,8 @@ proc notifyChallengeCompleted*(
       raise newException(
         ACMEError, "Unexpected error while signaling challenge completion", exc
       )
-    case checkResponseBody.getJSONField("status").getStr
+    let status = checkResponseBody.getJSONField("status").getStr
+    case status
     of "pending":
       var retryAfter: Duration
       try:
@@ -219,11 +224,11 @@ proc notifyChallengeCompleted*(
         retryAfter = DefaultChalCompletedRetryTime
       await sleepAsync(retryAfter) # try again after some delay
     of "valid":
-      return true
+      return
     else:
-      return false
-
-  return false
+      raise newException(
+        ACMEError, "Failed challenge completion: expected 'valid', got '" & status & "'"
+      )
 
 proc finalizeCertificate*(
     self: ref ACMEAccount,
--- a/libp2p/autotls/acme.nim
+++ b/libp2p/autotls/acme.nim
@@ -1,4 +1,5 @@
 import options, json, base64
+from times import DateTime, parse
 import chronos/apps/http/httpclient, jwt, results, bearssl/pem
 
 import ./utils
@@ -278,20 +279,26 @@ proc finalizeCertificate*(
 
 proc downloadCertificate*(
     self: ref ACMEAccount, orderURL: string
-): Future[string] {.async: (raises: [ACMEError, CancelledError]).} =
+): Future[(string, DateTime)] {.async: (raises: [ACMEError, CancelledError]).} =
   try:
     let downloadResponse =
       await HttpClientRequestRef.get(self.session, orderURL).get().send()
+
     if downloadResponse.status != 200:
       raise newException(ACMEError, "Failed to download certificate")
-    let certificateDownloadURL = bytesToString(await downloadResponse.getBodyBytes())
-      .parseJson()
-      .getJSONField("certificate").getStr
+
+    let certificateInfoBody =
+      bytesToString(await downloadResponse.getBodyBytes()).parseJson()
+
+    let certificateDownloadURL = certificateInfoBody.getJSONField("certificate").getStr
+    let certificateExpiry = parse(
+      certificateInfoBody.getJSONField("expires").getStr, "yyyy-MM-dd'T'HH:mm:ss'Z'"
+    )
 
     let certificateResponse =
       await HttpClientRequestRef.get(self.session, certificateDownloadURL).get().send()
     let rawCertificate = bytesToString(await certificateResponse.getBodyBytes())
-    return rawCertificate
+    return (rawCertificate, certificateExpiry)
   except HttpError:
     raise newException(ACMEError, "Failed to connect to ACME server")
   except Exception as e:
--- a/libp2p/autotls/acme.nim
+++ b/libp2p/autotls/acme.nim
@@ -10,8 +10,8 @@ import ../transports/tls/certificate
 
 # TODO: change staging to actual url
 const
-  LetsEncryptURL = "https://acme-v02.api.letsencrypt.org"
-  LetsEncryptURLStaging = "https://acme-staging-v02.api.letsencrypt.org"
+  LetsEncryptURL* = "https://acme-v02.api.letsencrypt.org"
+  LetsEncryptURLStaging* = "https://acme-staging-v02.api.letsencrypt.org"
   Alg = "RS256"
 
 type ACMEAccount* = object
--- a/libp2p/autotls/acme.nim
+++ b/libp2p/autotls/acme.nim
@@ -244,25 +244,32 @@ proc finalizeCertificate*(
     raise newException(ACMEError, "Failed to create CSR")
 
   let b64CSR = base64.encode(derCSR.toSeq, safe = true)
+  let payload = %*{"csr": b64CSR}
 
+  # send finalize request
+  let finalizedResponse = await self.signedAcmeRequest(finalizeURL, payload)
+  if finalizedResponse.status != 200:
+    raise newException(ACMEError, "Failed to request cert finalization")
+
+  # keep checking order until it's finalized
+  var checkResponseBody: JsonNode
   for _ in 0 .. retries:
-    var finalizedResponseBody: JsonNode
+    let finalizedResponse = await self.signedAcmeRequest(finalizeURL, payload)
     try:
-      let finalizedResponse =
-        await HttpClientRequestRef.get(self.session, finalizeURL).get().send()
-      finalizedResponseBody =
-        bytesToString(await finalizedResponse.getBodyBytes()).parseJson()
-    except HttpError:
-      raise newException(ACMEError, "Failed to connect to ACME server")
+      let checkResponse =
+        await HttpClientRequestRef.get(self.session, orderURL).get().send()
+      checkResponseBody = bytesToString(await checkResponse.getBodyBytes()).parseJson()
     except Exception as e:
       raise newException(
         ACMEError, "Unexpected error while finalizing certificate: " & e.msg
       )
-    case finalizedResponseBody.getJSONField("status").getStr
-    of "processing":
-      discard # try again
+
+    let status = checkResponseBody.getJSONField("status").getStr
+    case status
     of "valid":
       return true
+    of "processing":
+      discard # keep trying
     else:
       return false
     await sleepAsync(1.seconds)
--- a/libp2p/autotls/acme.nim
+++ b/libp2p/autotls/acme.nim
@@ -54,7 +54,7 @@ proc new*(
     raise newException(ACMEError, "Unable to parse JSON")
   except IOError:
     raise newException(ACMEError, "Unable to parse JSON")
-  except Exception as e:
+  except CatchableError as e:
     raise newException(
       ACMEError,
       "Unexpected error occurred while getting ACME server directory: " & e.msg,
@@ -103,7 +103,7 @@ proc signedAcmeRequest(
     let pemPrivKey: string = pemEncode(derPrivKey, "PRIVATE KEY")
     token.sign(pemPrivKey)
     body = token.toFlattenedJson()
-  except Exception as e:
+  except CatchableError as e:
     raise newException(ACMEError, "could not create JWT: " & e.msg)
   try:
     let response = await HttpClientRequestRef
@@ -162,7 +162,7 @@ proc requestChallenge*(
       let authzResponse =
         await HttpClientRequestRef.get(self.session, authzURL).get().send()
       await authzResponse.getParsedResponseBody()
-    except Exception as e:
+    except CatchableError as e:
       raise newException(ACMEError, "failed to request challenge: " & e.msg)
 
   let challenges = authzResponseBody.getJSONField("challenges")
@@ -190,7 +190,7 @@ proc notifyChallengeCompleted*(
       bytesToString(await completedResponse.getBodyBytes()).parseJson()
   except HttpError:
     raise newException(ACMEError, "Failed to connect to ACME server")
-  except Exception as e:
+  except CatchableError as e:
     raise newException(
       ACMEError, "Unexpected error while signaling challenge completion: " & e.msg
     )
@@ -205,7 +205,7 @@ proc notifyChallengeCompleted*(
       checkResponseBody = bytesToString(await checkResponse.getBodyBytes()).parseJson()
     except HttpError:
       raise newException(ACMEError, "Failed to connect to ACME server")
-    except Exception as e:
+    except CatchableError as e:
       raise newException(
         ACMEError, "Unexpected error while signaling challenge completion: " & e.msg
       )
@@ -262,7 +262,7 @@ proc finalizeCertificate*(
       let checkResponse =
         await HttpClientRequestRef.get(self.session, orderURL).get().send()
       checkResponseBody = bytesToString(await checkResponse.getBodyBytes()).parseJson()
-    except Exception as e:
+    except CatchableError as e:
       raise newException(
         ACMEError, "Unexpected error while finalizing certificate: " & e.msg
       )
@@ -303,7 +303,7 @@ proc downloadCertificate*(
     return (rawCertificate, certificateExpiry)
   except HttpError:
     raise newException(ACMEError, "Failed to connect to ACME server")
-  except Exception as e:
+  except CatchableError as e:
     raise newException(
       ACMEError, "Unexpected error while downloading certificate: " & e.msg
     )
--- a/libp2p/autotls/acme.nim
+++ b/libp2p/autotls/acme.nim
@@ -1,6 +1,8 @@
-import options, json, base64, sequtils
+import options, base64, sequtils, serialization, json_serialization
 from times import DateTime, parse
 import chronos/apps/http/httpclient, jwt, results, bearssl/pem
+from std/json import
+  JsonNode, `%*`, `%`, `[]`, `[]=`, `$`, parseJson, getStr, items, newJObject
 
 import ./utils
 import ../crypto/crypto
@@ -18,15 +20,18 @@ const
   DefaultFinalizeRetryTime = 1.seconds
   DefaultRandStringSize = 256
 
+type ACMEDirectory = object
+  newNonce: string
+  newOrder: string
+  newAccount: string
+
 type ACMEAccount* = object
   status*: Opt[string]
   contact*: Opt[seq[string]]
   key*: KeyPair
   session*: HttpSessionRef
   kid*: Opt[string]
-  newNonceURL: string
-  newAccountURL: string
-  newOrderURL: string
+  directory: ACMEDirectory
   acmeServerURL: string
 
 proc new*(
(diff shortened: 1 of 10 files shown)


libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+    clientSender: bool = true,
+): string =
+  let prefix = "libp2p-PeerID"
+  let parts =
+    # parts need to be sorted alphabetically by key
+    if clientSender:
+      @[
+        SigParam(k: "challenge-client", v: challenge.toByteSeq()),
+        SigParam(k: "hostname", v: hostname.toByteSeq()),
+        SigParam(k: "server-public-key", v: publicKey.getBytes().get()),
```
richard-ramos, line 59: The usage of `get` in here and line 64 and 68 assume that `getBytes()` and ``sign` cannot fail. Is this a safe assumption to make?

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+      ]
+    else:
+      @[
+        SigParam(k: "challenge-server", v: challenge.toByteSeq()),
+        SigParam(k: "client-public-key", v: publicKey.getBytes().get()),
+        SigParam(k: "hostname", v: hostname.toByteSeq),
```
richard-ramos, line 65:
```suggestion
        SigParam(k: "hostname", v: hostname.toByteSeq()),
```

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+logScope:
+  topics = "libp2p peerid auth"
+
+type SigParam = object
+  k: string
+  v: seq[byte]
+
+proc randomChallenge(): string =
+  randomize()
+  let charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
```
richard-ramos, line 14: const

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+
+logScope:
+  topics = "libp2p peerid auth"
+
+type SigParam = object
+  k: string
+  v: seq[byte]
+
+proc randomChallenge(): string =
+  randomize()
```
richard-ramos, line 13: Can we use libp2p rng instead of `random`?

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+    if x != 0:
+      byteVal = byteVal or 0x80
+    varInt.add(byteVal)
+  return varInt
+
+proc genDataToSign(prefix: string, parts: seq[SigParam]): seq[byte] =
+  var buf: seq[byte] = prefix.toByteSeq()
+  for p in parts:
+    buf.add encodeVarint(p.k.len + p.v.len + 1)
+    buf.add (p.k & "=").toByteSeq
```
richard-ramos, line 41: () for consistency
```suggestion
    buf.add (p.k & "=").toByteSeq()
```

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+    hostname: string,
+): bool {.raises: [PeerIDAuthError].} =
+  let prefix = "libp2p-PeerID"
+  let parts: seq[SigParam] =
+    @[
+      SigParam(k: "challenge-server", v: challengeServer.toByteSeq()),
+      SigParam(k: "client-public-key", v: clientPublicKey.getBytes().get()),
+      SigParam(k: "hostname", v: hostname.toByteSeq),
+    ]
+  let signedBytes = genDataToSign(prefix, parts)
```
richard-ramos, line 84: This is very similar to code prev function. Could we extract this into a separate function?

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+  var serverSignature: Signature
+  try:
+    if not serverSignature.init(base64.decode(serverSig).toByteSeq):
```
richard-ramos, line 87:
```suggestion
    if not serverSignature.init(base64.decode(serverSig).toByteSeq()):
```

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+        ("User-Agent", "nim-libp2p"),
+        ("authorization", authHeader),
+      ],
+    )
+    .get()
+    .send()
+  except HttpError:
+    raise newException(
+      PeerIDAuthError, "Failed to send request with bearer token for PeerID Auth"
+    )
```
richard-ramos, line 201:
```suggestion
  except HttpError as exc:
    raise newException(
      PeerIDAuthError, "Failed to send request with bearer token for PeerID Auth", exc
    )
```

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+    peerInfo: PeerInfo,
+    payload: JsonNode,
+    bearerToken: Opt[string] = Opt.none(string),
+): Future[(string, HttpClientResponseRef)] {.
+    async: (raises: [AutoTLSError, PeerIDAuthError, CancelledError])
+.} =
+  if bearerToken.isNone:
+    return await peerIdAuthenticate(url, session, peerInfo, payload)
+
+  let authHeader = "libp2p-PeerID bearer=\"" & bearerToken.get & "\""
```
richard-ramos, line 182: Can probably reuse a constant here

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+  try:
+    response = await HttpClientRequestRef
+    .post(
+      session,
+      url,
+      body = $payload,
+      headers = [
+        ("Content-Type", "application/json"),
+        ("User-Agent", "nim-libp2p"),
+        ("authorization", authHeader),
```
richard-ramos, line 193: I know that header keys are case insensitive, but probably a good idea to capitalize the `a` in `authorization` here for consistency. Of course, this nitpick would depend whether the server assumes that header keys can be case insensitive

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+      ],
+    )
+    .get()
```
richard-ramos, line 196: It's possible for `post` to return an error, right?

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+    ]
+  let signedBytes = genDataToSign(prefix, parts)
+  var serverSignature: Signature
+  try:
+    if not serverSignature.init(base64.decode(serverSig).toByteSeq):
+      raise newException(
+        PeerIDAuthError, "Failed to initialize Signature from base64 encoded sig"
+      )
+  except ValueError:
+    raise newException(PeerIDAuthError, "Failed to decode server's signature")
```
richard-ramos, line 92:
```suggestion
  except ValueError as exc:
    raise newException(PeerIDAuthError, "Failed to decode server's signature", exc)
```

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+    async: (raises: [AutoTLSError, PeerIDAuthError, CancelledError])
+.} =
+  # Authenticate in three ways as per the PeerID Auth spec
+  # https://github.com/libp2p/specs/blob/master/http/peer-id-auth.md
+
+  # request authentication
+  let base36PeerId = encodePeerId(peerInfo.peerId)
+  var authStartResponse: HttpClientResponseRef
+  try:
+    authStartResponse = await HttpClientRequestRef.get(session, url).get().send()
```
richard-ramos, line 109: Under which conditions could HttpClientRequestRef.get(session, url).get() return an error? This assumes it wont, right?

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+  except HttpError:
+    raise newException(PeerIDAuthError, "Failed to start PeerID Auth")
+
```
richard-ramos, line 112:
```suggestion
  except HttpError as exc:
    raise newException(PeerIDAuthError, "Failed to start PeerID Auth", exc)

```

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+  # www-authenticate
+  let wwwAuthenticate = authStartResponse.headers.getString("www-authenticate")
+  if wwwAuthenticate == "":
+    raise newException(PeerIDAuthError, "www-authenticate not present in response")
+  let challengeClient = extractField(wwwAuthenticate, "challenge-client")
+  var serverPublicKey: PublicKey
+  try:
+    serverPublicKey = PublicKey
+      .init(decode(extractField(wwwAuthenticate, "public-key")).toByteSeq)
+      .get()
```
richard-ramos, line 122: Could `init` fail? I think the call to `get()` assumes it does not (not sure what exception it raises tho).

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+
+  # www-authenticate
+  let wwwAuthenticate = authStartResponse.headers.getString("www-authenticate")
+  if wwwAuthenticate == "":
+    raise newException(PeerIDAuthError, "www-authenticate not present in response")
+  let challengeClient = extractField(wwwAuthenticate, "challenge-client")
+  var serverPublicKey: PublicKey
+  try:
+    serverPublicKey = PublicKey
+      .init(decode(extractField(wwwAuthenticate, "public-key")).toByteSeq)
```
richard-ramos, line 121:
```suggestion
      .init(decode(extractField(wwwAuthenticate, "public-key")).toByteSeq())
```

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+  # request authentication
+  let base36PeerId = encodePeerId(peerInfo.peerId)
+  var authStartResponse: HttpClientResponseRef
+  try:
+    authStartResponse = await HttpClientRequestRef.get(session, url).get().send()
+  except HttpError:
+    raise newException(PeerIDAuthError, "Failed to start PeerID Auth")
+
+  # www-authenticate
+  let wwwAuthenticate = authStartResponse.headers.getString("www-authenticate")
```
richard-ramos, line 114: Usually capitalized like `WWW-Authenticate`

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+  try:
+    serverPublicKey = PublicKey
+      .init(decode(extractField(wwwAuthenticate, "public-key")).toByteSeq)
+      .get()
+  except ValueError:
+    raise newException(PeerIDAuthError, "Could not decode public-key")
+  let opaque = extractField(wwwAuthenticate, "opaque")
+
+  let hostname = "registration." & AutoTLSDNSServer # registration.libp2p.direct
+  let clientPubKeyB64 = base64.encode(peerInfo.publicKey.getBytes().get(), safe = true)
```
richard-ramos, line 128: could `publicKey.getBytes()` fail?

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+  let challengeServer = randomChallenge()
+  let sig = peerIdSign(peerInfo.privateKey, challengeClient, serverPublicKey, hostname)
+  let authHeader =
+    "libp2p-PeerID public-key=\"" & clientPubKeyB64 & "\"" & ", opaque=\"" & opaque &
```
richard-ramos, line 132: Probably can reuse constant

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,245 @@
...
+
+logScope:
+  topics = "libp2p autotls"
+
+const
+  DefaultDnsServers =
+    @[
+      initTAddress("1.1.1.1:53"),
+      initTAddress("1.0.0.1:53"),
+      initTAddress("[2606:4700:4700::1111]:53"),
```
richard-ramos, line 41: Interestingly enough, `dnsclient` uses 8.8.8.8. I kinda prefer cloudflare like you added here, but wonder if we should try to unify both?

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,245 @@
...
+  var txt: seq[string]
+  var ip4: seq[TransportAddress]
+
+  for _ in 0 .. retries:
+    txt = await self.dnsResolver.resolveTxt(acmeChalDomain)
+    try:
+      ip4 = await self.dnsResolver.resolveIp(ip4Domain, 0.Port)
+    except:
+      raise newException(AutoTLSError, "Failed to resolve IP")
+    if txt.len > 0 and txt[0] != "not set yet" and ip4.len > 0:
```
richard-ramos, line 76: Where is this `not set yet` coming from? What can happen is this text changes?

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,245 @@
...
+      return true
+    await sleepAsync(1.seconds)
```
richard-ramos, line 78: Isn't 1s too much time? Probably makes sense to extract into a constant too

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,245 @@
...
+
+method issueCertificate(
+    self: AutoTLSManager
+): Future[void] {.base, async: (raises: [AutoTLSError, CancelledError]).} =
+  trace "Issuing new certificate"
+  var peerInfo: PeerInfo
+  try:
+    peerInfo = self.peerInfo.get
+  except:
+    raise newException(AutoTLSError, "Cannot issue new certificate: peerInfo not set")
```
richard-ramos, line 111: I think you can do:
```suggestion
let peerInfo = self.peerInfo.valueOr:
    raise newException(AutoTLSError, "Cannot issue new certificate: peerInfo not set")
```

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,245 @@
...
+
+  # generate autotls domain string: "*.{peerID}.libp2p.direct"
+  let base36PeerId = encodePeerId(peerInfo.peerId)
+  let baseDomain = base36PeerId & "." & AutoTLSDNSServer
+  let domain = "*." & baseDomain
+
+  debug "Requesting ACME challenge"
```
richard-ramos, line 118: `debug` or `trace`?

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,245 @@
...
+  var bearerToken: string
+  if self.bearerToken.isSome:
+    (bearerToken, response) = await peerIdAuthSend(
+      registrationURL,
+      self.httpSession,
+      peerInfo,
+      payload,
+      bearerToken = self.bearerToken,
+    )
+    discard bearerToken
```
richard-ramos, line 151: if you have no need for `bearerToken` i think you can use this. Also why the discard?
```suggestion
    (_, response) = await peerIdAuthSend(
      registrationURL,
      self.httpSession,
      peerInfo,
      payload,
      bearerToken = self.bearerToken,
    )
```

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,245 @@
...
+  )
+
+  var strMultiaddresses: seq[string] = @[]
+  for ma in peerInfo.addrs:
+    strMultiaddresses.add($ma)
+
+  let payload = %*{"value": keyAuthorization, "addresses": strMultiaddresses}
+  let registrationURL = AutoTLSBroker & "/v1/_acme-challenge"
+
+  debug "Sending challenge to AutoTLS broker"
```
richard-ramos, line 140: `debug` or `trace`

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,245 @@
...
+    raise newException(AutoTLSError, "ACME challenge completion notification failed")
+
+  debug "Finalize cert request with CSR"
+  if not await self.acmeAccount.finalizeCertificate(domain, finalizeURL, orderURL):
+    raise newException(AutoTLSError, "ACME certificate finalization request failed")
+
+  debug "Downloading certificate"
+  let (rawCert, expiry) = await self.acmeAccount.downloadCertificate(orderURL)
+
+  debug "Installing certificate"
```
richard-ramos, line 191: `debug` or `trace` ?

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,245 @@
...
+  if self.managerLoop.isNil:
+    warn "Stopping AutoTLS without starting it"
+    return
+
+  await self.managerLoop.cancelAndWait()
+  self.managerLoop = nil
+  if not self.acmeAccount.isNil:
+    await self.acmeAccount.session.closeWait()
+  if not self.httpSession.isNil:
+    await self.httpSession.closeWait()
```
richard-ramos, line 249: Is it necessary to wait? or could using cancelSoon and close be fine?

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,307 @@
...
+  if self.kid.isSome:
+    # already registered
+    return
+
+  let payload = %*{"termsOfServiceAgreed": true}
+
+  let newAccountURL = self.directory.getJSONField("newAccount").getStr
+  let response = await self.signedAcmeRequest(newAccountURL, payload, needsJwk = true)
+  let jsonResponseBody = await response.getParsedResponseBody()
+  if response.status != 201:
```
richard-ramos, line 130: Probably a good idea to have a constant HttpCreated for the 201

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,307 @@
...
+
+  let challengeResponse = await self.signedAcmeRequest(newOrderURL, orderPayload)
+  let challengeResponseBody = await challengeResponse.getParsedResponseBody()
+  let orderURL = challengeResponse.headers.getString("location")
+  if orderURL == "":
+    raise newException(ACMEError, "'location' header not found in ACME response")
+  let finalizeURL = challengeResponseBody.getJSONField("finalize").getStr
+
+  # get challenges
+  let authzURL = challengeResponseBody.getJSONField("authorizations")[0].getStr
```
richard-ramos, line 157: `authorizations`  will always contain at least 1 authorization?

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,307 @@
...
+    raise newException(ACMEError, "DNS01 challenge not found in ACME response")
+
+  return (dns01, finalizeURL, orderURL)
+
+proc notifyChallengeCompleted*(
+    self: ref ACMEAccount, chalURL: string, retries: int = 5
+): Future[bool] {.async: (raises: [ACMEError, CancelledError]).} =
+  let emptyPayload = newJObject()
+  let completedResponse = await self.signedAcmeRequest(chalURL, emptyPayload)
+  if completedResponse.status != 200:
```
richard-ramos, line 182: Maybe an HttpOk constant?

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,307 @@
...
+      ACMEError, "Unexpected error while signaling challenge completion: " & e.msg
+    )
+
+  let checkURL = completedResponseBody.getJSONField("url").getStr
+  # check until acme server is done (poll validation)
+  for _ in 0 .. retries:
+    var checkResponseBody: JsonNode
+    try:
+      let checkResponse =
+        await HttpClientRequestRef.get(self.session, checkURL).get().send()
```
richard-ramos, line 202: can .get(self.session, checkURL) return an error?

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,307 @@
...
+    domain: string,
+    finalizeURL: string,
+    orderURL: string,
+    retries: int = 5,
+): Future[bool] {.async: (raises: [ACMEError, CancelledError]).} =
+  var certKey: cert_key_t
+  var certCtx: cert_context_t
+  var derCSR: ptr cert_buffer = nil
+
+  # TODO: use secure method
```
richard-ramos, line 232: What does this refer to?

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,307 @@
...
+  var randomCstring: cstring
+  try:
+    randomCstring = urandomToCString(256)
```
richard-ramos, line 235: You can probably use libp2p rng

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+  var serverPublicKey: PublicKey
+  try:
+    serverPublicKey = PublicKey
+      .init(decode(extractField(wwwAuthenticate, "public-key")).toByteSeq)
+      .get()
+  except ValueError:
+    raise newException(PeerIDAuthError, "Could not decode public-key")
+  let opaque = extractField(wwwAuthenticate, "opaque")
+
+  let hostname = "registration." & AutoTLSDNSServer # registration.libp2p.direct
```
richard-ramos, line 127: Should it be `let hostname = AutoTLSBroker` ?

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,249 @@
...
+    self.bearerToken = Opt.some(bearerToken)
+
+  # no need to do anything from this point forward if there are not public ip addresses on host
+  var hostPrimaryIP: IpAddress
+  try:
+    hostPrimaryIP = getPrimaryIPAddr()
+    if not isPublicIPv4(hostPrimaryIP):
+      raise newException(AutoTLSError, "Host does not have a public IPv4 address")
+  except Exception:
+    raise newException(AutoTLSError, "Failed to get primary IP address for host")
```
vladopajic, line 165: code should not catch `Exception` because it is root of `CatchableError` and `Defect`. 
`Defect` is catchable, but code shouldn't catch it unless it is deliberately done for some reason.
therefore in all our code we catch only `CatchableError`.

```
Exception (root of all exceptions)
├── CatchableError (recoverable, can be caught with `except`)
│  
└── Defect (programming bugs, NOT catchable with `except`)
    ├── AssertionDefect
    ├── NilAccessDefect
    └── IndexDefect, etc.
```

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,249 @@
...
+  if self.acmeAccount.isNil:
+    let accountKey = KeyPair.random(PKScheme.RSA, self.rng[]).get()
+    self.acmeAccount = (await ACMEAccount.new(accountKey))
+  await self.acmeAccount.register()
+
+  heartbeat "Certificate Management", self.renewCheckTime:
+    if self.cert.isNone or self.certExpiry.isNone:
+      try:
+        await self.issueCertificate()
+      except Exception as e:
```
vladopajic, line 214: should be `CatchableError`

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,249 @@
...
+        error "Failed to issue certificate", err = e.msg
+        break
+
+    # AutoTLSManager will renew the cert 1h before it expires
+    let expiry = self.certExpiry.get
+    let waitTime: Duration = expiry - Moment.now - self.renewBufferTime
+    if waitTime <= self.renewBufferTime:
+      try:
+        await self.issueCertificate()
+      except Exception as e:
```
vladopajic, line 224: should be `CatchableError`

libp2p/autotls/utils.nim
```diff
@@ -0,0 +1,93 @@
...
+
+proc getParsedResponseBody*(
+    response: HttpClientResponseRef
+): Future[JsonNode] {.async: (raises: [ACMEError]).} =
+  try:
+    let responseBody = bytesToString(await response.getBodyBytes()).parseJson()
+    return responseBody
+  except ValueError, OSError, IOError:
+    raise newException(ACMEError, "Unable to parse JSON body")
+  except Exception as e:
```
vladopajic, line 65: should be `CatchableError`

tests/testautotls.nim
```diff
@@ -0,0 +1,141 @@
...
+
+    asyncTeardown:
+      await switch.stop()
+      checkTrackers()
+
+    asyncTest "test AutoTLSManager start":
+      var hostPrimaryIP: IpAddress
+      try:
+        hostPrimaryIP = getPrimaryIPAddr()
+      except Exception:
```
vladopajic, line 83: should be `CatchableError`

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,309 @@
...
+    return acc
+  except HttpError:
+    raise newException(ACMEError, "Failed to connect to ACME server")
+  except ValueError:
+    raise newException(ACMEError, "Unable to parse JSON")
+  except OSError:
+    raise newException(ACMEError, "Unable to parse JSON")
+  except IOError:
+    raise newException(ACMEError, "Unable to parse JSON")
+  except Exception as e:
```
vladopajic, line 57: should be `CatchableError`

libp2p/autotls/autotls.nim
gmelodie: Thx for the tip!

tests/testautotls.nim
```diff
@@ -0,0 +1,141 @@
...
+      check acc.kid.isSome
+
+    asyncTest "test ACME challange request":
+      await acc.register()
+      # challenge requested
+      let (dns01Challenge, finalizeURL, orderURL) =
+        await acc.requestChallenge(@["some.dummy.domain.com"])
+      check dns01Challenge.isNil == false
+      check finalizeURL.len > 0
+      check orderURL.len > 0
```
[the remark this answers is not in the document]
gmelodie: `asyncSetup` runs once per test in that suite, not once per suite. Though in this case that's more of an argument against than for it haha

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+  let charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
+  result = ""
+  for _ in 0 ..< 48:
+    result.add sample(charset)
+
+proc extractField(data, key: string): string {.raises: [PeerIDAuthError].} =
+  # Helper to extract quoted value from key
+  for segment in data.split(","):
+    if key in segment:
+      return segment.split("=", 1)[1].strip(chars = {' ', '"'})
```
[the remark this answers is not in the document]
gmelodie: it is not unless the autotls broker is very poorly configured

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,202 @@
...
+  return buf
+
+proc peerIdSign(
+    privateKey: PrivateKey,
+    challenge: string,
+    publicKey: PublicKey,
+    hostname: string,
+    clientSender: bool = true,
+): string =
+  let prefix = "libp2p-PeerID"
```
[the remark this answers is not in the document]
gmelodie: I won't do it to the SigParam keys since  I think it takes more jumping around to find them while reading the code aftewards

libp2p/autotls/peeridauth.nim
gmelodie: Yes since the keys come from `peerInfo`, and `peerInfo` comes from the switch itself.

libp2p/autotls/peeridauth.nim
gmelodie: Yes! Added a check for status codes different than `200`

gmelodie: Apparently it'll never return an error https://github.com/status-im/nim-chronos/blob/b55e2816eb45f698ddaca8d8473e401502562db2/chronos/apps/http/httpclient.nim#L1050

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,308 @@
...
+  kid*: Opt[string]
+  directory: JsonNode
+  acmeServerURL: string
+
+proc new*(
+    T: typedesc[ACMEAccount],
+    key: KeyPair,
+    status: Opt[string] = Opt.none(string),
+    contact: Opt[seq[string]] = Opt.none(seq[string]),
+    kid: Opt[string] = Opt.none(string),
```
richard-ramos, line 31: Does it make sense for these fields to be setup when instantiating ACMEAccount? It seems that `status` and `kid` are setup later. Is `contact` being used at all?

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,258 @@
...
+
+method start*(
+    self: AutoTLSManager, peerInfo: PeerInfo
+): Future[void] {.base, async: (raises: [CancelledError]).} =
+  if not self.managerLoop.isNil:
+    warn "Starting AutoTLS twice"
+    return
+
+  self.peerInfo = Opt.some(peerInfo)
+  self.managerLoop = self.manageCertificate()
```
richard-ramos, line 245: Should it really be called `managerLoop`? or `managerFut`?

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,308 @@
...
+
+    let status = checkResponseBody.getJSONField("status").getStr
+    case status
+    of "valid":
+      return true
+    of "processing":
+      discard # keep trying
+    else:
+      return false
+    await sleepAsync(1.seconds)
```
richard-ramos, line 277: See prev comment about RFC8555 `Retry-after`

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,258 @@
...
+    raise newException(
+      AutoTLSError, "Failed to authenticate with AutoTLS Broker at " & AutoTLSBroker
+    )
+
+  # no need to do anything from this point forward if there are not public ip addresses on host
+  var hostPrimaryIP: IpAddress
+  try:
+    hostPrimaryIP = getPrimaryIPAddr()
+    if not isPublicIPv4(hostPrimaryIP):
+      raise newException(AutoTLSError, "Host does not have a public IPv4 address")
```
richard-ramos, line 171: Some projects like nwaku obtain their IP address via NAT(upmp/pmp) (look for `setupNAT` in nwaku repo), or even allow the user to hardcode an IPAddress.  It probably makes sense to have an optional attribute in AutoTLSManager so a dev can setup an IP (this should be also a parameter passed to withAutoTLSManager()) option.
If the dev using nim-libp2p sets up the optional IP, it should use that as the hostPrimaryIP, otherwise it can determine the IP just like you're doing here

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,258 @@
...
+): Future[bool] {.async: (raises: [AutoTLSError, CancelledError]).} =
+  var txt: seq[string]
+  var ip4: seq[TransportAddress]
+
+  for _ in 0 .. retries:
+    txt = await self.dnsResolver.resolveTxt(acmeChalDomain)
+    try:
+      ip4 = await self.dnsResolver.resolveIp(ip4Domain, 0.Port)
+    except:
+      raise newException(AutoTLSError, "Failed to resolve IP")
```
richard-ramos, line 75: Should we also use the retries for dns resolution failures? they do tend to happen (not frequently tho)

libp2p/autotls/acme.nim
gmelodie: If the request succeeded (which, if we're getting to this point it did), it should ([link](https://www.rfc-editor.org/rfc/rfc8555#section-7.1.3)):
> authorizations (required, array of string):  For pending orders, the
      authorizations that the client needs to complete before the
      requested certificate can be issued (see [Section 7.5](https://www.rfc-editor.org/rfc/rfc8555#section-7.5)), including
      unexpired authorizations that the client has completed in the past
      for identifiers specified in the order.  The authorizations
      required are dictated by server policy; there may not be a 1:1
      relationship between the order identifiers and the authorizations
      required.  For final orders (in the "valid" or "invalid" state),
      the authorizations that were completed.  Each entry is a URL from
      which an authorization can be fetched with a POST-as-GET request.

libp2p/autotls/acme.nim
gmelodie: leftovers from tests

libp2p/autotls/acme.nim
gmelodie: `contact` is there for future use, we're not using it as of now

libp2p/autotls/autotls.nim
gmelodie: Changing this makes nim 1.6 fail

libp2p/autotls/autotls.nim
gmelodie: Changing this makes nim 1.6 fail

libp2p/autotls/autotls.nim
gmelodie: Changing this makes nim 1.6 fail

tests/testautotls.nim
gmelodie: Changing this makes nim 1.6 fail

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,309 @@
...
+
+  # keep checking order until it's finalized
+  var checkResponseBody: JsonNode
+  for _ in 0 .. retries:
+    let finalizedResponse = await self.signedAcmeRequest(finalizeURL, payload)
+    try:
+      let checkResponse =
+        await HttpClientRequestRef.get(self.session, orderURL).get().send()
+      checkResponseBody = bytesToString(await checkResponse.getBodyBytes()).parseJson()
+    except Exception as e:
```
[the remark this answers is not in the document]
gmelodie: Changing this makes nim 1.6 fail

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,309 @@
...
+      certificateInfoBody.getJSONField("expires").getStr, "yyyy-MM-dd'T'HH:mm:ss'Z'"
+    )
+
+    let certificateResponse =
+      await HttpClientRequestRef.get(self.session, certificateDownloadURL).get().send()
+    let rawCertificate = bytesToString(await certificateResponse.getBodyBytes())
+    return (rawCertificate, certificateExpiry)
+  except HttpError:
+    raise newException(ACMEError, "Failed to connect to ACME server")
+  except Exception as e:
```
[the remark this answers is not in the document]
gmelodie: Changing this makes nim 1.6 fail

libp2p/autotls/acme.nim
gmelodie: Changing this makes nim 1.6 fail

libp2p/autotls/utils.nim
```diff
@@ -0,0 +1,87 @@
...
+
+proc getParsedResponseBody*(
+    response: HttpClientResponseRef
+): Future[JsonNode] {.async: (raises: [ACMEError]).} =
+  try:
+    let responseBody = bytesToString(await response.getBodyBytes()).parseJson()
+    return responseBody
+  except ValueError, OSError, IOError:
+    raise newException(ACMEError, "Unable to parse JSON body")
+  except Exception as exc:
```
vladopajic, line 67: why is `Exception` rally cached here?
code enumerates all possible exceptions and then add this as final `Exception` catch. it doesn't seem right. 
ether code lists all specific exceptions (probably the best choice), or it catches `CatchableError` .

libp2p/autotls/utils.nim
```diff
@@ -0,0 +1,87 @@
...
+  return Base36.encode(cidResult.get().data.buffer)
+
+proc getParsedResponseBody*(
+    response: HttpClientResponseRef
+): Future[JsonNode] {.async: (raises: [ACMEError]).} =
+  try:
+    let responseBody = bytesToString(await response.getBodyBytes()).parseJson()
+    return responseBody
+  except ValueError, OSError, IOError:
+    raise newException(ACMEError, "Unable to parse JSON body")
```
vladopajic, line 66: could we add error message from root error here? it can be better in understanding if for any reason this fails,

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,317 @@
...
+
+proc newNonce(
+    self: ref ACMEAccount
+): Future[string] {.async: (raises: [ACMEError, CancelledError]).} =
+  try:
+    let nonceURL = self.directory.getJSONField("newNonce").getStr
+    let resp = await HttpClientRequestRef.get(self.session, nonceURL).get().send()
+    return resp.headers.getString("Replay-Nonce")
+  except HttpError:
+    raise newException(ACMEError, "Failed to request new nonce from ACME server")
```
vladopajic, line 74: Could we propagate parent error message here?

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,317 @@
...
+      self.session,
+      url,
+      body = $body,
+      headers = [("Content-Type", "application/jose+json")],
+    )
+    .get()
+    .send()
+    return response
+  except HttpError:
+    raise newException(ACMEError, "Failed to send HTTP request to the ACME server")
```
vladopajic, line 123: Could we propagate parent error message here?

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,317 @@
...
+  let completedResponse = await self.signedAcmeRequest(chalURL, emptyPayload)
+  if completedResponse.status != HttpOk:
+    return false
+
+  var completedResponseBody: JsonNode
+  try:
+    completedResponseBody =
+      bytesToString(await completedResponse.getBodyBytes()).parseJson()
+  except HttpError:
+    raise newException(ACMEError, "Failed to connect to ACME server")
```
vladopajic, line 195: Could we propagate parent error message here?

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,317 @@
...
+  # check until acme server is done (poll validation)
+  for _ in 0 .. retries:
+    var checkResponse: HttpClientResponseRef
+    var checkResponseBody: JsonNode
+    try:
+      checkResponse =
+        await HttpClientRequestRef.get(self.session, checkURL).get().send()
+      checkResponseBody = bytesToString(await checkResponse.getBodyBytes()).parseJson()
+    except HttpError:
+      raise newException(ACMEError, "Failed to connect to ACME server")
```
vladopajic, line 211: Could we propagate parent error message here?

vladopajic (changes requested):
exceptions does not seem correctly handled
- hiding root cause message - we should always propagate it, as it can be beneficial in debugging 
- catching Exceptions everywhere - seems unnecessary or it looks like it catches error in inappropriately handled code.  if really necessary let's document why. why this codebase didn't really had to do this until now?

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,317 @@
...
+
+    let acc = new(ACMEAccount)
+    acc.status = status
+    acc.contact = contact
+    acc.kid = kid
+    acc.key = key
+    acc.session = session
+    acc.directory = directory
+    acc.acmeServerURL = acmeServerURL
+    return acc
```
vladopajic, line 52: nitpick: consider not having practice of having big `try-except` statements. 
when developers add code, the just add code in place, extending `try-except` even more, then they have `try-except` in `try-except`... 

in this case `ACMEAccount` can be initialized after  `try-except

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,317 @@
...
+  try:
+    token = toJWT(%*{"header": acmeHeader, "claims": payload})
+    let derPrivKey = self.key.seckey.rsakey.getBytes.get
+    let pemPrivKey: string = pemEncode(derPrivKey, "PRIVATE KEY")
+    token.sign(pemPrivKey)
+    body = token.toFlattenedJson()
+  except Exception as exc:
+    raise newException(ACMEError, "Failed to create JWT", exc)
+  try:
+    let response = await HttpClientRequestRef
```
vladopajic, line 112: nitpick: since variable `response` is not really used we can directly return here `return await ...`

libp2p/autotls/utils.nim
```diff
@@ -0,0 +1,108 @@
...
+  return
+    not (
+      ip.startsWith("10.") or
+      (ip.startsWith("172.") and parseInt(ip.split(".")[1]) in 16 .. 31) or
+      ip.startsWith("192.168.") or ip.startsWith("127.") or ip.startsWith("169.254.")
+    )
+
+proc asMoment*(dt: DateTime): Moment =
+  let unixTime: int64 = dt.toTime.toUnix
+  return Moment.init(unixTime, Second)
```
richard-ramos, line 45: This is nice!  Wondering if we should later send a pr to nim-chronos to include this helper

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,271 @@
...
+    self.bearerToken = Opt.some(bearerToken)
+  if response.status != HttpOk:
+    raise newException(
+      AutoTLSError, "Failed to authenticate with AutoTLS Broker at " & AutoTLSBroker
+    )
+
+  # no need to do anything from this point forward if there are not public ip addresses on host
+  var hostPrimaryIP: IpAddress
+  try:
+    hostPrimaryIP = self.ipAddress.get(checkedGetPrimaryIPAddr())
```
richard-ramos, line 178: I wonder if this `get` gets lazily evaluated? because if not, maybe something like this can be used to avoid callign checkedGetPrimaryIPAddr if not needed:
```
let hostPrimaryIP: IpAddress = self.ipAddress.valueOr:
   try:
      let ip = checkedGetPrimaryIPAddr()
      if not isPublicIPV(ip):
         raise newException(AutoTLSError, "Host does not have a public IPv4 address")
      ip
   except GetPrimaryIPError as exc:
     raise newException(AutoTLSError, "Failed to get primary IP address for host", exc)
   except CatchableError as exc:
     raise newException(
      AutoTLSError, "Unexpected error while getting primary IP address for host", exc
    )
```

libp2p/autotls/utils.nim
```diff
@@ -0,0 +1,108 @@
...
+  try:
+    return getPrimaryIPAddr()
+  except Exception as exc:
+    raise newException(GetPrimaryIPError, "Error while getting primary IP address", exc)
+
+proc getJSONField*(node: JsonNode, field: string): JsonNode {.raises: [ACMEError].} =
+  try:
+    return node[field]
+  except CatchableError:
+    raise newException(ACMEError, "'" & field & "' field not found in JSON")
```
richard-ramos, line 80: You can avoid having a try catch if you do something like this
```
proc getJSONField*(node: JsonNode, field: string): JsonNode {.raises: [ACMEError].} =
  if not node.hasKey(field):
    raise newException(ACMEError, "'" & field & "' field not found in JSON")
  node[field]
```

libp2p/autotls/utils.nim
```diff
@@ -0,0 +1,108 @@
...
+  except OSError as exc:
+    raise newException(ACMEError, "Unable to parse JSON body", exc)
+  except IOError as exc:
+    raise newException(ACMEError, "Unable to parse JSON body", exc)
+  except CatchableError as exc:
+    raise
+      newException(ACMEError, "Unexpected error occurred while getting body bytes", exc)
+
+proc thumbprint*(key: KeyPair): string =
+  # TODO: check if scheme is RSA
```
richard-ramos, line 99: Probably safer to add an add an assertion for keypair not being RSA. Might save us some debugging later on the road
```
doAssert key.seckey.scheme == PKScheme.RSA, "unsupported keytype"
```

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,271 @@
...
+            thumbprint(self.acmeAccount.key)
+          ).toByteSeq
+        ).data
+      )
+    )
+  )
+
+  var strMultiaddresses: seq[string] = @[]
+  for ma in peerInfo.addrs:
+    strMultiaddresses.add($ma)
```
richard-ramos, line 147: nitpick, feel free to ignore:

```suggestion
  let strMultiaddresses: seq[string] = peerInfo.addrs.mapIt($it)
```

vladopajic, line 147: var strMultiaddresses = peerInfo.addrs.mapIt($it)

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,316 @@
...
+  self.status = Opt.some(jsonResponseBody.getJSONField("status").getStr)
+
+proc requestChallenge*(
+    self: ref ACMEAccount, domains: seq[string]
+): Future[(JsonNode, string, string)] {.async: (raises: [ACMEError, CancelledError]).} =
+  # request challenge from ACME server
+  var identifiers: seq[JsonNode]
+
+  for domain in domains:
+    identifiers.add(%*{"type": "dns", "value": domain})
```
vladopajic, line 145: maybe mapIt?

richard-ramos, line 145: nitpick, feel free to ignore
```suggestion
  var identifiers: seq[JsonNode] = domains.mapIt(%*{"type": "dns", "value": it})
```

richard-ramos (commented):
Except for quic being added twice in nimble, and the missing ip parameter for withAutoTLSManager option, all of my comments might be either random comments or nitpicks that can be ignored.

libp2p/autotls/autotls.nim
```diff
@@ -0,0 +1,271 @@
...
+  let dashedIpAddr = ($hostPrimaryIP).replace(".", "-")
+  let acmeChalDomain = "_acme-challenge." & baseDomain
+  let ip4Domain = dashedIpAddr & "." & baseDomain
+  if not await self.checkDNSRecords(ip4Domain, acmeChalDomain):
+    raise newException(AutoTLSError, "DNS records not set")
+
+  debug "Notifying challenge completion to ACME server"
+  let chalURL = dns01Challenge.getJSONField("url").getStr
+  if not await self.acmeAccount.notifyChallengeCompleted(chalURL):
+    raise newException(AutoTLSError, "ACME challenge completion notification failed")
```
vladopajic, line 203: there is no information why it failed.

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,316 @@
...
+proc new*(
+    T: typedesc[ACMEAccount],
+    key: KeyPair,
+    status: Opt[string] = Opt.none(string),
+    contact: Opt[seq[string]] = Opt.none(seq[string]),
+    kid: Opt[string] = Opt.none(string),
+    acmeServerURL: string = LetsEncryptURL,
+): Future[ref ACMEAccount] {.async: (raises: [ACMEError, CancelledError]).} =
+  let session = HttpSessionRef.new()
+  var directory: JsonNode
```
vladopajic, line 39: why `JsonNode` are used? why not parse it to appropriate model right away and use model. it gives better type safety then accessing individual properties using `getJSONField`, etc..

tests/testautotls.nim
```diff
@@ -0,0 +1,140 @@
...
+    let base36PeerId = encodePeerId(switch.peerInfo.peerId)
+    let dnsTXTRecord = (
+      await dnsResolver.resolveTxt(
+        fmt"_acme-challenge.{base36PeerId}.{AutoTLSDNSServer}"
+      )
+    )[0]
+
+    # DNS TXT record is set
+    let keyAuthorization = switch.autoTLSMgr.keyAuthorization.valueOr:
+      check false
```
vladopajic, line 107: don't use `check false` use `raiseAssert "this should not happen message"`

vladopajic: other places as well...

tests/testautotls.nim
```diff
@@ -0,0 +1,140 @@
...
+      .withRng(newRng())
+      .withAddress(MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet())
+      .withTcpTransport()
+      .withAutoTLSManager(acmeServerURL = LetsEncryptURLStaging)
+      .withYamux()
+      .withNoise()
+      .build()
+
+    defer:
+      checkTrackers()
```
vladopajic, line 67: convention in code is currently is to have `checkTrackers` in `teardown`. it would be nice to follow the same convention.

libp2p/autotls/utils.nim
```diff
@@ -0,0 +1,108 @@
...
+type GetPrimaryIPError* = object of LPError
+type AutoTLSError* = object of LPError
+type ACMEError* = object of AutoTLSError
+type PeerIDAuthError* = object of AutoTLSError
+
+const
+  AutoTLSBroker* = "registration.libp2p.direct"
+  AutoTLSDNSServer* = "libp2p.direct"
+  HttpOk* = 200
+  HttpCreated* = 201
```
vladopajic, line 15: nipick: you can use single `type` here like `const` was used below.

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,316 @@
...
+  DefaultFinalizeRetryTime = 1.seconds
+  DefaultRandStringSize = 256
+
+type ACMEAccount* = object
+  status*: Opt[string]
+  contact*: Opt[seq[string]]
+  key*: KeyPair
+  session*: HttpSessionRef
+  kid*: Opt[string]
+  directory: JsonNode
```
vladopajic, line 27: `directory` should not be generic type. please use user defined type.

tests/testautotls.nim
gmelodie: Did you not want me to remove `setup` and `teardown` in https://github.com/vacp2p/nim-libp2p/pull/1358#discussion_r2108735434?

libp2p/autotls/acme.nim
gmelodie: Discussion: https://github.com/vacp2p/nim-libp2p/pull/1358#discussion_r2114587975

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,321 @@
...
+    let resp =
+      await HttpClientRequestRef.get(self.session, self.newNonceURL).get().send()
+    return resp.headers.getString("Replay-Nonce")
+  except HttpError as exc:
+    raise newException(ACMEError, "Failed to request new nonce from ACME server", exc)
+
+# TODO: save n and e in account so we don't have to recalculate every time
+proc acmeHeader(
+    self: ref ACMEAccount, url: string, needsJwk: bool, kid: string = ""
+): Future[JsonNode] {.async: (raises: [ACMEError, CancelledError]).} =
```
vladopajic, line 80: should return some user specified type, instead of `JsonNode`.

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,321 @@
...
+  if response.status != HttpCreated:
+    raise newException(
+      ACMEError, "Unable to register with ACME server: " & $jsonResponseBody
+    )
+  self.kid = Opt.some(response.headers.getString("location"))
+  self.status = Opt.some(jsonResponseBody.getJSONField("status").getStr)
+
+proc requestChallenge*(
+    self: ref ACMEAccount, domains: seq[string]
+): Future[(JsonNode, string, string)] {.async: (raises: [ACMEError, CancelledError]).} =
```
vladopajic, line 144: should return some user specified type, instead of JsonNode.

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,229 @@
...
+    signedBytes.toOpenArray(0, signedBytes.len - 1), serverPublicKey
+  )
+
+proc peerIdAuthenticate(
+    url: string,
+    session: HttpSessionRef,
+    peerInfo: PeerInfo,
+    payload: JsonNode,
+    rng: ref HmacDrbgContext,
+): Future[(string, HttpClientResponseRef)] {.
```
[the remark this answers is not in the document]
vladopajic: same type can hold all values (string + everything needed from HttpClientResponseRef)

libp2p/autotls/acme.nim
vladopajic: same type can have everyting from JsonNode + other two strings

libp2p/autotls/peeridauth.nim
```diff
@@ -0,0 +1,229 @@
...
+    raise newException(PeerIDAuthError, "Failed to decode server's signature", exc)
+  return serverSignature.verify(
+    signedBytes.toOpenArray(0, signedBytes.len - 1), serverPublicKey
+  )
+
+proc peerIdAuthenticate(
+    url: string,
+    session: HttpSessionRef,
+    peerInfo: PeerInfo,
+    payload: JsonNode,
```
vladopajic, line 104: should not use `JsonNode` as payload, pass exactly values (with types) that you need, and create payload inside.

vladopajic (commented):
dont use `HttpClientResponseRef` and `JsonNode` in point of interface (return values, input values).

libp2p/autotls/utils.nim
```diff
@@ -0,0 +1,109 @@
...
+  try:
+    return getPrimaryIPAddr()
+  except Exception as exc:
+    raise newException(GetPrimaryIPError, "Error while getting primary IP address", exc)
+
+proc getJSONField*(node: JsonNode, field: string): JsonNode {.raises: [ACMEError].} =
+  try:
+    return node[field]
+  except CatchableError:
+    raise newException(ACMEError, "'" & field & "' field not found in JSON")
```
vladopajic, line 81: for decoding end encoding json objects, please use [json-serialization](https://github.com/status-im/nim-json-serialization). 
library is already use in this repo [here](https://github.com/vacp2p/nim-libp2p/blob/1f98c1eed7d28bc93ccef9f29a3e4239d693be74/interop/transport/main.nim#L96).

vladopajic (changes requested):
there is no clear separation in interaction with API
- parsing of response data overflows to different methods, outside of where http request was made
- creating api requests is made in different places
- methods that deal with API are not clearly isolated, there is no separation of concern

---- 

for guidance, please create structures that clearly separate interaction with API only. for example there are 2 types of api interaction:
- ACME
- AutoTLSBroker

create 2 types that will encapsulate API interaction inside.

```
type AcmeAPI ...
proc register*(self: AcmeAPI): AcmeRegisterResponse ....
proc requestChallenge*(self: AcmeAPI,  domains: seq[string]): AcmeChallengeResponse ....


type BrokerAPI ...
proc authPeerId*(...): BrokerAuthResponse ....

```

these two types must have clear interface, it's methods must return strongly typed structures and inputs must be strongly typed.

other logic will use these two types, without knowing if these methods under the hood are requesting from internet or these methods are mocked (they will be in unit tests later).

---

as per contribution rules: 
> Small PRs. Try to keep PRs atomic and digestible. This makes the review process and pinpointing bugs easier.

PLEASE, create these two types in new PR branching of latest master. we are going to review and merge that PR independently of this PR. and this PR will rebase to that PR.

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,321 @@
...
+    header["jwk"] = %*{"kty": "RSA", "n": n, "e": e}
+  else:
+    if self.kid.isNone:
+      raise newException(ACMEError, "no kid registered for account")
+    header["kid"] = %*(self.kid.get)
+  return header
+
+proc signedAcmeRequest(
+    self: ref ACMEAccount, url: string, payload: JsonNode, needsJwk: bool = false
+): Future[HttpClientResponseRef] {.async: (raises: [ACMEError, CancelledError]).} =
```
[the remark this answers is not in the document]
gmelodie: This is a helper function to send http requests. It can receive a plethora of different responses. Parsing should not be done here.

libp2p/autotls/acme.nim
gmelodie: This is a helper function to build an http header. The header will be sent as a json object. It makes no sense for the response of this function not to be json

libp2p/autotls/peeridauth.nim
gmelodie: Peeridauth is an authentication scheme. At the end of the exchange the client will send data to the server. What data that client will send can greatly vary as this is not an autotls-specific protocol, autotls just uses it to authenticate node and broker. This function should not handle parsing. Parsing should be handled outside of it

libp2p/autotls/utils.nim
gmelodie: So if an object has 10 fields and I just need 1, do you want me to parse the entire object and get that one field?

libp2p/autotls/peeridauth.nim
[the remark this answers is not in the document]
gmelodie: As previously stated, `peerIdAuthenticate` is NOT an AutoTLS function, it's a PeerID Authentication helper function, thus if it is to be reusable it can't implement parsing on the response since the response will greatly vary depending on the application using it.

libp2p/autotls/acme.nim
vladopajic: > The header will be sent as a json object.

work with type in code, real user defined types.  header should be be some user defined type with some fields `AcmeHeader`. 

the code that needs this header in json format will encode it in place where it needs it.

richard-ramos:
@vladopajic, @gmelodie
this PR has gone through several iterations and is functionally solid rn. To avoid fatigue and keep momentum, i suggest we merge this and move remaining suggestions to a follow-up. proposed plan:
1 -  open an issue summarizing pending feedback — this can block final completion of the AutoTLS commitment
2 -  follow up with another PR for improvements/refactoring
Optionally, comment out withAutoTLSManager temporarily so no one uses the feature until cert installation is handled
since this code isn’t in production yet and lives in an isolated package, current changes shouldn’t block merge. We can still review the changes in architecture in detail. We can create an issue for this. (it's a lot of code, but not that much tbh: 1000loc). 
Further PRs will of course need to be smaller in scope to make reviewing easy.

PR #1364, opened 2025-05-13
chore(interop): use the same redis dependency 

vladopajic:
hole punching to use same redis as interop tests.


richard-ramos (approved):

gmelodie (approved):
I personally would choose having both interop Dockerfiles with `redis`, but I'm fine with whatever [as discussed here](https://discord.com/channels/864066763682218004/1371857954533412934/1372193821294723092)

vladopajic:
> I personally would choose having both interop Dockerfiles with `redis`, but I'm fine with whatever [as discussed here](https://discord.com/channels/864066763682218004/1371857954533412934/1372193821294723092)

`redis` should not be used because it can introduce braking change, so version of redis should be pinned to version or commit.

we are not using version since the last version released is from 2022 and commit used in interop tests (one we updated hp tests) is from 2023 (newer).

PR #1370, opened 2025-05-14
chore: fix import warnings

narimiran:
- change from deprecated `stew/results` to `results`
- remove unused imports, as reported by the compiler

### DIFF
--- a/libp2p.nim
+++ b/libp2p.nim
@@ -52,7 +52,6 @@ else:
       stream/connection,
       transports/transport,
       transports/tcptransport,
-      transports/quictransport,
       protocols/secure/noise,
       cid,
       multihash,
--- a/libp2p.nim
+++ b/libp2p.nim
@@ -52,6 +52,7 @@ else:
       stream/connection,
       transports/transport,
       transports/tcptransport,
+      transports/quictransport,
       protocols/secure/noise,
       cid,
       multihash,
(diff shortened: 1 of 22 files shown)


tests/testnative.nim
```diff
@@ -24,8 +24,7 @@ import
   testtortransport,
   testwstransport,
   testquic,
-  testmemorytransport,
-  transports/tls/testcertificate
```
vladopajic, line 28: `transports/tls/testcertificate` should be kept here.

add `{.used.}` to top of testcertificate.nim

libp2p.nim
```diff
@@ -52,7 +52,6 @@ else:
       stream/connection,
       transports/transport,
       transports/tcptransport,
-      transports/quictransport,
```
vladopajic, line 55: add export of `quictransport` below (after pubsub)...

vladopajic (changes requested):

richard-ramos:
Also, linting is required. We use [`nph`](https://github.com/arnetheduck/nph)

richard-ramos (approved):

vladopajic (approved):

arnetheduck:
>  linting is required

btw, https://github.com/arnetheduck/nph-action will post the required code changes as review suggestions so you can just accept them from the PR review screen - just added it to a few projects, ie see https://github.com/status-im/nim-rocksdb/pull/86/

PR #1371, opened 2025-05-14
chore: bump quic

richard-ramos:
Requires:
- https://github.com/vacp2p/nim-quic/pull/71


vladopajic:
@richard-ramos could we use quic as `"quic>= 0.3.0 & < 0.4.0",` so we don't have to bump for each patch version ?

gmelodie (approved):

PR #1376, opened 2025-05-16
ci(test): reduce timeout

vladopajic:
timeout is way to big after all recent ci improvements. usually completing in 15min. on rare occasion it takes 30-35min. 
so 45min looks enough, even 35min looks good, but we might sometimes need to manually retry if ci fails due to timeout?


rlve (approved):
We should be able to do even 40 👀

gmelodie (changes requested):
We should also update the same timeout in `daily-common.yml`

gmelodie (approved):

Issue #1377, opened 2025-05-16
unit test ocassionaly hang on macos


vladopajic:
some ci examples:

https://github.com/vacp2p/nim-libp2p/actions/runs/15045554663/attempts/1?pr=1204
https://github.com/vacp2p/nim-libp2p/actions/runs/15054157891/job/42315960400

vladopajic:
issue was in quic (read operation did not had proper timeout implemented). this issue was fixed and new quic lib version is used https://github.com/vacp2p/nim-libp2p/pull/1387.

Issue #1367, opened 2025-05-13
ci: hole punching stopped working


vladopajic:
this issue tracks effort for fixing hole punching tests.

after issue is resolved we should make interoperability tests mandatory in ci again (currently temporarily disabled).

vladopajic:
- no other libp2p implementation has this implemented and working
  -  rust have these tests, but are failing
  -  other implementations don’t have these tests at all
- notified libp2p discord community [message](https://discord.com/channels/1204447718093750272/1204448960773623818/1379437592772349964)

PR #1378, opened 2025-05-16
chore(hp): add timeout on dial

vladopajic:
this pr adds timeout on `dial` to test theory if hole punching intepor test are failing due to dial hanging indefinitely.

reasons to support this theory:
- unit tests are hanging  #1377
- hp hanging clues (#1367):
  -  the last message in logs is `Dialing peer`
  - test ends after bit after 4min (there is timeout on total execution time to 4min)

-----------

second change, hp and transport interop tests have not correctly processed main future:
- should quit(0) if program is successful 
- after timeout there is no error message

### DIFF
--- a/tests/hole-punching-interop/hole_punching.nim
+++ b/tests/hole-punching-interop/hole_punching.nim
@@ -43,7 +43,7 @@ proc createSwitch(r: Relay = nil, hpService: Service = nil): Switch =
   s.mount(Ping.new(rng = rng))
   return s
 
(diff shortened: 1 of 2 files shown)


vladopajic:
[here](https://github.com/vacp2p/nim-libp2p/actions/runs/15072197225/job/42371071923?pr=1378) is an error message when ci fails after current improvement

richard-ramos (approved):

PR #1391, opened 2025-05-21
test(gossipsub): heartbeat tests

rlve:
Add tests related to GossipSub Heartbeat according to section 4 of [Test Plan ](https://www.notion.so/Gossipsub-651e02d4d7894bb2ac1e4edb55f3192d).

New tests:
- `"Mesh is rebalanced during heartbeat - pruning peers"`
- `"Mesh is rebalanced during heartbeat - grafting new peers"`
- `"Mesh is rebalanced during heartbeat - opportunistic grafting"`
- `"Fanout maintenance during heartbeat - expired peers are dropped"`
- `"Fanout maintenance during heartbeat - fanout peers are replenished"`
- `"iDontWants history - last element is pruned during heartbeat"`
- `"sentIHaves history - last element is pruned during heartbeat"`


tests/pubsub/testgossipsubheartbeat.nim
```diff
@@ -0,0 +1,305 @@
...
+      )
+      nodes[i].addObserver(pubsubObserver)
+
+    # Nodes are connected to Node0
+    for i in 1 ..< numberOfNodes:
+      await connectNodes(node0, nodes[i])
+    subscribeAllNodes(nodes, topic, voidTopicHandler)
+    await waitForHeartbeat()
+    check:
+      node0.mesh[topic].len == 9
```
vladopajic, line 38: `node0.mesh[topic].len == numberOfNodes -1`

tests/pubsub/testgossipsubheartbeat.nim
```diff
@@ -0,0 +1,305 @@
...
+
+    await waitForHeartbeat()
+    let after2ndHearbeat = lastPrune
+
+    let measuredHeartbeat = after2ndHearbeat - after1stHearbeat
+    let heartbeatDiff = measuredHeartbeat - TEST_GOSSIPSUB_HEARTBEAT_INTERVAL
+
+    # Then mesh of Node0 is rebalanced and peers are pruned to adapt to new values
+    check:
+      node0.mesh[topic].len >= 2 and node0.mesh[topic].len <= 4
```
vladopajic, line 66: `node0.mesh[topic].len >= 2 and node0.mesh[topic].len <= 4`

what are 2 and 4 here, `newDVales.dLow` and `newDVales.dHigh` ?

tests/pubsub/testgossipsubheartbeat.nim
```diff
@@ -0,0 +1,305 @@
...
+      if peer in node0.mesh[topic]:
+        # Assign scores in starting Mesh
+        peer.score = score
+        score += 100.0
+      else:
+        # Assign scores higher than median to Peers not in starting Mesh and expect them to be grafted
+        peer.score = 800.0
+        expectedGrafts &= peer
+
+    # Then during heartbeat Peers with lower than median scores are pruned and max 2 Peers are grafted
```
vladopajic, line 155: `and max 2 Peers are grafted` why max 2?

tests/pubsub/testgossipsubheartbeat.nim
```diff
@@ -0,0 +1,305 @@
...
+    await waitForHeartbeat()
+
+    let actualGrafts = node0.mesh[topic].toSeq().filterIt(it notin startingMesh)
+    check:
+      actualGrafts.len == 2
```
vladopajic, line 160: `actualGrafts.len == 2` should not be hardcoded

tests/pubsub/testgossipsubheartbeat.nim
```diff
@@ -0,0 +1,305 @@
...
+    # When Node0 sends a message to the topic
+    tryPublish await node0.publish(topic, newSeq[byte](10000)), 1
+
+    # Then Node0 fanout peers are populated
+    await waitForPeersInTable(node0, topic, 6, PeerTableType.Fanout)
+    check:
+      node0.fanout.hasKey(topic) and node0.fanout[topic].len == 6
+
+    # When peers of Node0 fanout are disconnected
+    let peersToDisconnect = node0.fanout[topic].toSeq()[1 .. ^1].mapIt(it.peerId)
```
vladopajic, line 216: ` let peersToDisconnect = node0.fanout[topic].toSeq()[1 .. ^1].mapIt(it.peerId)`

->

`
let peersToDisconnect =  pickAtRandomSkip(node0.fanout[topic].toSeq(), 2)
`

tests/pubsub/testgossipsubheartbeat.nim
```diff
@@ -0,0 +1,305 @@
...
+    await findAndStopPeers(nodes, peersToDisconnect, topic, voidTopicHandler)
+    await waitForPeersInTable(node0, topic, 1, PeerTableType.Fanout)
+
+    # Then Node0 fanout peers are replenished during heartbeat
+    await waitForHeartbeat()
+    check:
+      node0.fanout[topic].len == 4
+      node0.fanout[topic].toSeq().allIt(it.peerId notin peersToDisconnect)
```
vladopajic, line 224: `node0.fanout[topic].len == 4` might be unnecessary as condition below will imply this

redundant check is okay, but magic value `4` is not okay -> should be `node0.fanout[topic].toSeq().len - peersToDisconnect.len`

----

more bullet proof check would be:

`node0.fanout[topic]` == `old node0.fanout[topic] removing peersToDisconnect`

tests/pubsub/testgossipsubheartbeat.nim
```diff
@@ -0,0 +1,305 @@
...
+      tryPublish await nodes[0].publish(topic, newSeq[byte](1000)), 1
+
+    # Find Peer outside of mesh to which Node 0 will send IHave
+    let peer =
+      nodes[0].gossipsub[topic].toSeq().filterIt(it notin nodes[0].mesh[topic])[0]
+
+    # When next heartbeat occurs
+    let timeout = Moment.now() + 100.milliseconds
+    while peer.sentIHaves[^1].len == 0:
+      await activeWait(1.milliseconds, timeout, "wait for sentIHaves timeout")
```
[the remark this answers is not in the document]
vladopajic: also, `1 ms` sleep interval seems vary small, depending on language, os, machine, and other factors, it's vary arguable if code will sleep (it may just skip). it would be better to use bit larger timeout like `10 ms`

vladopajic, line 291: 1ms sleep interval seems vary small, depending on language, os, machine, and other factors, it's vary arguable if code will sleep (it may just skip). it would be better to use bit larger timeout like 10 ms.

tests/pubsub/testgossipsubheartbeat.nim
rlve: It's according to the [spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#opportunistic-grafting) 

https://github.com/vacp2p/nim-libp2p/blob/ad5eae9adf593e98fdc9db90cdfcf8dccd8d32cc/libp2p/protocols/pubsub/gossipsub/behavior.nim#L541

tests/pubsub/testgossipsubheartbeat.nim
rlve: I choose here to disconnect all peers but the first one

vladopajic: can we pick peers at random? i guess similar utility might be useful in other tests when we need to pick in arbitrary way,

tests/pubsub/testgossipsubheartbeat.nim
vladopajic: got it, then code is missing this constant, and that this constant needs to point here... i'll create issue for this

vladopajic: https://github.com/vacp2p/nim-libp2p/issues/1395

tests/pubsub/testgossipsubheartbeat.nim
rlve: I will add a utility like this when there is a need next time. Let's keep it as it is and don't increase the scope of the PR if it's not necessary.

tests/pubsub/testgossipsubheartbeat.nim
rlve: 👌

tests/pubsub/utils.nim
```diff
@@ -512,6 +516,21 @@ proc addIDontWantObservers*[T: PubSub](
...
   return allMessages
 
+proc findAndStopPeers*[T: PubSub](
+    nodes: seq[T], peers: seq[PeerId], topic: string, handler: TopicHandler
+) {.async.} =
+  for i in 0 ..< nodes.len:
+    let node = nodes[i]
+    if peers.anyIt(it == node.peerInfo.peerId):
+      node.unsubscribe(topic, voidTopicHandler)
+      await node.stop()
```
vladopajic, line 526: i am not really sure about this func, maybe it should only do 1 thing (unsubscribe). 
if node is stopped then it will also trigger mesh to rebalanced?

vladopajic (approved):

tests/pubsub/testgossipsubheartbeat.nim
```diff
@@ -0,0 +1,309 @@
...
+import ../helpers
+
+suite "GossipSub Heartbeat":
+  teardown:
+    checkTrackers()
+
+  asyncTest "Mesh is rebalanced during heartbeat - pruning peers":
+    let
+      numberOfNodes = 10
+      topic = "foobar"
```
gmelodie, line 15: Could this be a `const`?

tests/pubsub/testgossipsubheartbeat.nim
```diff
@@ -0,0 +1,309 @@
...
+        .toGossipSub()
+      node0 = nodes[0]
+
+    startNodesAndDeferStop(nodes)
+
+    # Nodes are connected to Node0
+    for i in 1 ..< numberOfNodes:
+      await connectNodes(node0, nodes[i])
+    subscribeAllNodes(nodes, topic, voidTopicHandler)
+    await waitForHeartbeat()
```
gmelodie, line 94: Could this logic be moved into an `asyncSetup` block?

tests/pubsub/testgossipsubheartbeat.nim
```diff
@@ -0,0 +1,309 @@
...
```
gmelodie, line 139: Again, maybe move this into `asyncSetup`

tests/pubsub/testgossipsubheartbeat.nim
```diff
@@ -0,0 +1,309 @@
...
+
+    # Waiting 2 heartbeats to populate sentIHaves history
+    await waitForHeartbeat(2)
+
+    # When Node0 sends a messages to the topic
+    tryPublish await nodes[0].publish(topic, newSeq[byte](1000)), 1
+
+    # Find Peer outside of mesh to which Node 0 will send IHave
+    let peer =
+      nodes[0].gossipsub[topic].toSeq().filterIt(it notin nodes[0].mesh[topic])[0]
```
gmelodie, line 290: nit
```suggestion
      nodes[0].gossipsub[topic].toSeq().filterIt(peer notin nodes[0].mesh[topic])[0]
```

gmelodie (approved):

tests/pubsub/testgossipsubheartbeat.nim
rlve: It can't. `asyncSetup` works like `beforeEach`, which means it's common for all the tests in the suite. In my case, setup of each test is a little bit different.

tests/pubsub/testgossipsubheartbeat.nim
rlve: https://github.com/vacp2p/nim-libp2p/pull/1391#discussion_r2108608085

tests/pubsub/testgossipsubheartbeat.nim
rlve: https://nim-lang.org/docs/sequtils.html#filterIt.t%2Cuntyped%2Cuntyped:~:text=Unlike%20the%20filter%20proc%20and%20filter%20iterator%2C%20the%20predicate%20needs%20to%20be%20an%20expression%20using%20the%20it%20variable%20for%20testing%2C%20like%3A%20filterIt(%22abcxyz%22%2C%20it%20%3D%3D%20%27x%27).

PR #1393, opened 2025-05-22
fix(gossipsub): save iDontWants messages in the first element of history

rlve:
https://github.com/vacp2p/nim-libp2p/issues/1394


tests/pubsub/testgossipsubcontrolmessages.nim
```diff
@@ -444,7 +444,7 @@ suite "GossipSub Control Messages":
 
     # Then B doesn't relay the message to C.
     checkUntilTimeout:
-      nodes[1].mesh.getOrDefault(topic).anyIt(it.iDontWants[^1].len == 1)
+      nodes[1].mesh.getOrDefault(topic).anyIt(it.iDontWants.anyIt(it.len == 1))
```
richard-ramos, line 447: Curious to understand why the change from `[^1]` to `anyIt`, is this because there's no guarantee for the messages to arrive in the same heartbeat?

richard-ramos (approved):
Nice catch and fix

vladopajic (approved):

PR #1398, opened 2025-05-25
feat: `withWsTransport`

richard-ramos:
cc: @Ivansete-status

### DIFF
--- a/interop/transport/main.nim
+++ b/interop/transport/main.nim
@@ -47,12 +47,9 @@ proc main() {.async.} =
         MultiAddress.init("/ip4/" & ip & "/udp/0/quic-v1").tryGet()
       )
   of "ws":
-    discard switchBuilder
-      .withTransport(
-        proc(upgr: Upgrade, privateKey: PrivateKey): Transport =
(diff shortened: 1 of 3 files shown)


Ivansete-status (commented):
LGTM!
I'd suggest making `proc withTransport*(` private , if possible , so that we notice this change and adapt `nwaku` accordingly when we consume this commit :)

Also, notice we distinguish between `ws` and `wss` in nwaku.
See the two implementations within `nwaku`:

https://github.com/waku-org/nwaku/blob/72d1fa0d349960280563a1d738efdf6a33ab6d7d/waku/node/waku_switch.nim#L45-L58

https://github.com/waku-org/nwaku/blob/72d1fa0d349960280563a1d738efdf6a33ab6d7d/waku/node/waku_switch.nim#L21-L25

... which are used in:
https://github.com/waku-org/nwaku/blob/72d1fa0d349960280563a1d738efdf6a33ab6d7d/waku/node/waku_switch.nim#L111-L114

vladopajic (approved):

Issue #1331, opened 2025-04-17
test: ensure all test have `checkTrackers()`


vladopajic:
let's ensure that all test follow the same convention to have `checkTrackers()`. this will help us ensure that test will not fail if test ordering has changed.

Issue #1330, opened 2025-04-17
test(interop): some test here might no be clossing conections


vladopajic:
we should check interop test and update them if they are not closing all connections. we should also update test and add ` teardown: checkTrackers()` on test sute to ensure that it wont happen in future.

PR #1399, opened 2025-05-26
chore: unused imports are now errors

### LINKED ISSUE #1343: remove unused imports
imported but unused dependencies can slightly increase compile time, let's 1) remove all unused 2) ensure with CI to never allow commits with those.

narimiran:
This closes #1343.


vladopajic:
> > pay attention to error messages next to "FAILED"
> 
> Thanks for the tip!
> 
> There are lots of messages with FAILED, but all seem to fall into the same category: the number of opened stream.transport is **always** for **exactly 6** larger than the number of closed stream.transport.

those issues are not really related to your, changes but are manifested with your changes.

some tests have 
```
  teardown:
    checkTrackers()
```
which ensure that after test is run all resources are closed. 

some tests don't have these checks https://github.com/vacp2p/nim-libp2p/issues/1331. 

then imaging following scenario:
- run test A and finally checkTrackers() -> passes
- run test B and do not run checkTrackers() -> passes (but some resources are not closed)
-> test suite completes successfully 

now, when test suite order changed:
- run test B and do not run checkTrackers() -> passes (but some resources are not closed)
- run test A and finally checkTrackers() -> now fails because it accounted some unclosed resources from previous run)
-> test suite completes unsuccessfully 

this is usually this issue.

vladopajic:
> Is the solution to add `teardown: checkTrackers()` to all places that are missing it? If so, I can do that.

well... it's not really solution for your problem.
because once you add `checkTrackers` if test fails (which some certainly will) you'll have to investigate why it failed and what you need to close. so it's not as trivial as just add `checkTrackers()`. 

but if you want to go down this road then i guess fixing #1331 first makes more sense (in separate pr). or you can investigate what is not closed in `tests/testdaemon.nim` since it looks like these test are causing most issues.

vladopajic:
> Eh, based on the comment in this PR: #1335, and this issue: #1330, I think I found (one of) the problem(s), and it seems it has been known for three years: https://github.com/vacp2p/nim-libp2p/blob/master/tests/commoninterop.nim#L167-L172
> 
> And if y'all who are familiar with the codebase couldn't figure it out, my chances as an outsider are basically zero.

will https://github.com/vacp2p/nim-libp2p/pull/1402 this help?

narimiran:
> copy the change on your branch, it's 1 line...

Heh, yeah, that's definitely the faster way to do it :D 

It has worked. Ready for review/merge.

tests/testinterop.nim
```diff
@@ -2,6 +2,8 @@ import helpers, commoninterop
 import ../libp2p
 import ../libp2p/crypto/crypto, ../libp2p/protocols/connectivity/relay/relay
 
+{.used.}
```
vladopajic, line 5: other `testXXX` files have `{.used.}` at the top (first line). can we keep the same convention?

vladopajic (approved):
lgtm, minor comment

richard-ramos (approved):

Issue #1394, opened 2025-05-22
GossipSub: iDontWants messages saved only in last element of history


rlve:
### Context:

Messages are added to a peer's iDontWants list when the local node receives an IDONWANT control message from that peer. 
When an IDONWANT message is received for a particular messageId from a peer, the salted version of this messageId is added to a set.

https://github.com/vacp2p/nim-libp2p/blob/ad5eae9adf593e98fdc9db90cdfcf8dccd8d32cc/libp2p/protocols/pubsub/gossipsub/behavior.nim#L305-L310

During each heartbeat:
- A new empty set is added to the beginning of the equence. This new set should store IDONWANT messages received during the next heartbeat interval.
- If the length of `peer.iDontWants` exceeds `g.parameters.historyLength`, the oldest set (the last element in the sequence) is removed.

https://github.com/vacp2p/nim-libp2p/blob/ad5eae9adf593e98fdc9db90cdfcf8dccd8d32cc/libp2p/protocols/pubsub/gossipsub/behavior.nim#L699-L709

### Issue:

Currently new IDONWANT messages are always being added to the last element of the sequence, which is also deleted during the heartbeat, and the rest of sets is always empty.

See the log below that should help visualize current behaviour. 

![Image](https://github.com/user-attachments/assets/23cfa846-8cdb-43e6-800e-1e66ecada0b7)

Scenario above can be reproduced using following PR: https://github.com/vacp2p/nim-libp2p/pull/1393
`nim c -r tests/pubsub/testgossipsubheartbeat.nim`

### Expected:

I expect that new IDONWANT messages should be added to the first element of the sequence when received and the full history is utilised.

PR #1405, opened 2025-05-26
fix(GossipSub): save sent iHave in first element

rlve:
`sentIHaves` has the same issue as `iDontWants` history (https://github.com/vacp2p/nim-libp2p/issues/1394) - new elements are being saved in the last element of the queue and they're being pruned only after one heartbeat. 

Behaviour of `sentIHaves` during hearbeat is analogical to `iDontWants`:
- A new empty set is added to the beginning of the queue. This new set should store `MessageIDs` of messages included in `IHave` sent during the current heartbeat interval.
- If the length of `peer.sentIHaves` exceeds `g.parameters.historyLength`, the oldest set (the last element in the queue) is removed.

https://github.com/vacp2p/nim-libp2p/blob/3a7745f920d11fb29305070cd9991874529dfc28/libp2p/protocols/pubsub/gossipsub/behavior.nim#L699-L711

Changes:
- fix: save new elements in the first position of `sentIHaves` to utilise full history
- update test


vladopajic (approved):

tests/pubsub/testgossipsubheartbeat.nim
```diff
@@ -289,6 +289,7 @@ suite "GossipSub Heartbeat":
         dValues =
           some(DValues(dLow: some(1), dHigh: some(1), d: some(1), dOut: some(0))),
         heartbeatInterval = heartbeatInterval,
+        gossipThreshold = -100.0,
```
vladopajic: make this threshold a const. then use that threshold in code below p.socre = 2 * threshold....

richard-ramos (approved):

PR #1420, opened 2025-05-28
feat(gossipsub): Add support for custom connection handling (Mix protocol integration)

AkshayaMani:
This PR implements support in GossipSub for protocols like the [Mix protocol](https://rfc.vac.dev/vac/raw/mix), which require custom peer selection and connection handling when publishing messages.

### Summary of changes
- Adds a new `useCustomConn` flag to `GossipSub.publish` (defaults to `false`).
- When `useCustomConn` is `true`, `publish` uses callbacks registered during GossipSub initialization:
  - `peerSelectionCB`: returns the peers to publish to.
  - `customConnCreationCB`: returns a custom connection for sending the message.
- If these callbacks are not provided and `useCustomConn` is `true`, `publish` returns an error.

### Motivation
This is part of ongoing integration work for the Mix protocol — a custom, stateless libp2p protocol designed for anonymous message routing in peer-to-peer networks.
The callback mechanism is designed to be generic and may be reused by other protocols requiring fine-grained control over peer selection and connection setup.

### Notes
- All changes are backward-compatible; default `publish` behavior is unaffected.
- Code has been formatted using `nph`.

### Related
[VAC RAW RFC: Mix Protocol](https://rfc.vac.dev/vac/raw/mix)

> **Note:** This PR also includes a small unrelated change needed to support the custom connection logic. It does not impact GossipSub behavior directly.

### DIFF
--- a/libp2p/stream/chronosstream.nim
+++ b/libp2p/stream/chronosstream.nim
@@ -34,8 +34,6 @@ when defined(libp2p_agents_metrics):
   declareCounter libp2p_peers_traffic_read, "incoming traffic", labels = ["agent"]
   declareCounter libp2p_peers_traffic_write, "outgoing traffic", labels = ["agent"]
 
-declareCounter libp2p_network_bytes, "total traffic", labels = ["direction"]
-
 func shortLog*(conn: ChronosStream): auto =
   try:
     if conn == nil:
--- a/libp2p/stream/connection.nim
+++ b/libp2p/stream/connection.nim
@@ -52,6 +52,8 @@ func shortLog*(conn: Connection): string =
 chronicles.formatIt(Connection):
   shortLog(it)
 
+declarePublicCounter libp2p_network_bytes, "total traffic", labels = ["direction"]
+
 method initStream*(s: Connection) =
   if s.objName.len == 0:
     s.objName = ConnectionTrackerName
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -1,7 +1,8 @@
 import std/sequtils
-import pkg/chronos
-import pkg/chronicles
-import pkg/quic
+import chronos
+import chronicles
+import metrics
+import quic
 import results
 import ../multiaddress
 import ../multicodec
@@ -58,6 +59,7 @@ method readOnce*(
     result = min(nbytes, stream.cached.len)
     copyMem(pbytes, addr stream.cached[0], result)
     stream.cached = stream.cached[result ..^ 1]
+    libp2p_network_bytes.inc(result.int64, labelValues = ["in"])
   except CatchableError as exc:
     raise newLPStreamEOFError()
 
@@ -66,6 +68,7 @@ method write*(
     stream: QuicStream, bytes: seq[byte]
 ) {.async: (raises: [CancelledError, LPStreamError]).} =
   mapExceptions(await stream.stream.write(bytes))
+  libp2p_network_bytes.inc(bytes.len.int64, labelValues = ["out"])
 
 {.pop.}
 
--- a/tests/pubsub/testgossipsubgossip.nim
+++ b/tests/pubsub/testgossipsubgossip.nim
@@ -306,15 +306,13 @@ suite "GossipSub Gossip Protocol":
 
     # When node 0 sends a large message
     let largeMsg = newSeq[byte](1000)
-    check (await nodes[0].publish(topic, largeMsg)) == 1
-    await waitForHeartbeat()
+    tryPublish await nodes[0].publish(topic, largeMsg), 1
 
     # Only node 2 should have received the iDontWant message
-    let receivedIDontWants = messages[].mapIt(it[].len)
-    check:
-      receivedIDontWants[0] == 0
-      receivedIDontWants[1] == 0
-      receivedIDontWants[2] == 1
+    checkUntilTimeout:
+      messages[].mapIt(it[].len)[2] == 1
+      messages[].mapIt(it[].len)[1] == 0
+      messages[].mapIt(it[].len)[0] == 0
 
   asyncTest "e2e - GossipSub peer exchange":
     # A, B & C are subscribed to something
--- a/tests/pubsub/testgossipsubheartbeat.nim
+++ b/tests/pubsub/testgossipsubheartbeat.nim
@@ -169,14 +169,17 @@ suite "GossipSub Heartbeat":
       node.subscribe(topic, voidTopicHandler)
     await waitForHeartbeat(heartbeatInterval)
 
-    # When Node0 sends a message to the topic
     let node0 = nodes[0]
+    checkUntilTimeout:
+      node0.gossipsub.hasKey(topic)
+
+    # When Node0 sends a message to the topic
     tryPublish await node0.publish(topic, newSeq[byte](10000)), 3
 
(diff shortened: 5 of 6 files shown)


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -95,6 +95,17 @@ type
...
+  CustomConnectionCallbacks* = object
+    customConnCreationCB*: proc(
+      destAddr: Option[MultiAddress], destPeerId: PeerId, codec: string
+    ): Connection {.gcsafe, raises: [].}
+    peerSelectionCB*: proc(
+      allPeers: HashSet[PubSubPeer],
+      directPeers: HashSet[PubSubPeer],
+      meshPeers: HashSet[PubSubPeer],
+      fanoutPeers: HashSet[PubSubPeer],
+    ): HashSet[PubSubPeer] {.gcsafe, raises: [].}
```
richard-ramos: To make this `CustomConnectionCallbacks` object easier to read, could you extract the procs to a type? That way The object would look like:
```
CustomConnectionCallbacks* = object
    customConnCreationCB*: someCustomProcTypeHere
    peerSelectionCB*: anotherCustomProcTypeHere
```

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -356,21 +368,42 @@ proc sendMsgSlow(p: PubSubPeer, msg: seq[byte]) {.async: (raises: [CancelledErro
...
-    let conn = p.sendConn
+proc sendMsg(
+    p: PubSubPeer, msg: seq[byte], useCustomConn: bool = false
+): Future[void] {.async: (raises: []).} =
+  var slowPath = false
+  let (conn, connType) =
+    if useCustomConn and p.customConnCallbacks.isSome:
+      let address =
+        if p.address.isSome:
+          some(p.address.get)
```
richard-ramos: since p.address is already `some`,  couldnt you return p.address directly?

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -356,21 +368,42 @@ proc sendMsgSlow(p: PubSubPeer, msg: seq[byte]) {.async: (raises: [CancelledErro
...
+        else:
+          none(MultiAddress)
+      (
+        p.customConnCallbacks.get().customConnCreationCB(address, p.peerId, p.codec),
+        "custom connection",
```
richard-ramos: Probably makes sense to use a enum for `custom connection`, `send connection` and `slow path`, instead of relying on strings

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -719,55 +719,70 @@ method publish*(
...
-        let
-          bandwidth = (g.parameters.bandwidthEstimatebps) div 8 div 1000
-            # Divisions are to convert it to Bytes per ms TODO replace with bandwidth estimate
-          msToTransmit = max(data.len div bandwidth, 1)
-        max(
-          g.parameters.heartbeatInterval.milliseconds div msToTransmit,
-          g.parameters.dLow,
+  if useCustomConn:
+    if g.customConnCallbacks.isSome:
+      peers.incl(
```
richard-ramos: Might make sense to add a trace log here to know we're using a customConnCallback (similar to how it's done on line 7333)

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -719,55 +719,70 @@ method publish*(
...
+          g.subscribedDirectPeers.getOrDefault(topic),
+          g.mesh.getOrDefault(topic),
+          g.fanout.getOrDefault(topic),
         )
-      else:
-        int.high() # unlimited
+      )
+    else:
+      trace "No custom connection callbacks provided, skipping publish"
+      libp2p_gossipsub_failed_publish.inc()
```
richard-ramos: This condition should not happen, as it would indicate a bad configuration from the developer. Should we use an assertion here and intentionally crash the program?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -719,55 +719,70 @@ method publish*(
 
   var peers: HashSet[PubSubPeer]
```
vladopajic: in `publish` method it looks like `peers` are created using two ways based on `useCustomConn`.

```
var peers: HashSet[PubSubPeer]

if useCustomConn
  .... populates peers
else 
  ... populates peers
 
# use peers here

```
because this code is hard to follow could we extract these two ways of creating peers in functions so that we get code like this:

```
var peers: HashSet[PubSubPeer]

if useCustomConn
   peers = g.makePeersForPublishUsingCustomCon(topic)
else 
   peers = g.makePeersForPublishDefault(topic)
 
# use peers here

```

it will make `publish` method much easier to read in future.

vladopajic (approved):
lgtm in general just needs some tidying up

libp2p/protocols/pubsub/pubsubpeer.nim
AkshayaMani: Refactored `CustomConnectionCallbacks` to use named proc types: `CustomConnCreationProc` and `CustomPeerSelectionProc`.

libp2p/protocols/pubsub/pubsubpeer.nim
AkshayaMani: Simplified this by returning `p.address` directly

libp2p/protocols/pubsub/pubsubpeer.nim
AkshayaMani: Replaced connection type strings with a `ConnectionType` enum (`ctCustom`, `ctSend`, `ctSlow`).

libp2p/protocols/pubsub/gossipsub.nim
AkshayaMani: Added a `trace` log in `makePeersForPublishUsingCustomConn` function to indicate when a custom peer selection callback is being used.

libp2p/protocols/pubsub/gossipsub.nim
AkshayaMani: Added an assert to fail when `useCustomConn` is `true` but `customConnCallbacks` is not set.

libp2p/protocols/pubsub/gossipsub.nim
AkshayaMani: Refactored peer selection logic in `publish` into two helper functions: `makePeersForPublishUsingCustomConn` and `makePeersForPublishDefault`.

vladopajic:
hmm, just realized it would be good to have some tests for this code

Ben-PH:
Relevant context; this PR breaks some things in the [mix]() repo. Don't let that stop any progress on this PR. This context might help paint a picture of the bigger picture.

AkshayaMani:
> hmm, just realized it would be good to have some tests for this code

Good point! We can probably add tests for the peer selection logic (custom path) and also verify behavior when `customConnCallbacks` is missing.
Curious if you had any specific edge cases in mind?

> Relevant context; this PR breaks some things in the mix repo. Don't let that stop any progress on this PR. This context might help paint a picture of the bigger picture.

Yep, aware of that — the changes in the mix repo are minor and can be addressed once this PR is merged. Thanks for flagging it!

PR #1421, opened 2025-05-28
ci(dependencies): fix access to tokens

vladopajic:
we are having this error currently:
```
Invalid workflow file: .github/workflows/dependencies.yml#L20
The workflow is not valid. .github/workflows/dependencies.yml (Line: 20, Col: 20): Unrecognized named-value: 'secrets'. Located at position 1 within expression: secrets.ACTIONS_GITHUB_TOKEN_NIMBUS_ETH2 .github/workflows/dependencies.yml (Line: 23, Col: 20): Unrecognized named-value: 'secrets'. Located at position 1 within expression: secrets.ACTIONS_GITHUB_TOKEN_NWAKU
```

secrets could not be passed in matrix, and must be used from within job.

ci fixed proof: https://github.com/vacp2p/nim-libp2p/actions/workflows/dependencies.yml

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -15,16 +15,26 @@ jobs:
       fail-fast: false
       matrix:
         target:
-          - repository: status-im/nimbus-eth2
+          - id: nimbus
+            repository: status-im/nimbus-eth2
             ref: unstable
-            token: ${{ secrets.ACTIONS_GITHUB_TOKEN_NIMBUS_ETH2 }}
-          - repository: waku-org/nwaku
+          - id: nwaku
+            repository: waku-org/nwaku
             ref: master
-            token: ${{ secrets.ACTIONS_GITHUB_TOKEN_NWAKU }}
-          - repository: codex-storage/nim-codex
+          - id: codex
+            repository: codex-storage/nim-codex
             ref: master
-            token: ${{ secrets.ACTIONS_GITHUB_TOKEN_NIM_CODEX }}
     steps:
+      - name: Set secret based on matrix target
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/dependencies.yml
```diff
@@ -15,24 +15,34 @@ jobs:
...
     steps:
+      - name: Set secret based on matrix target
+        run: |
+          if [ "${{ matrix.target.id }}" = "nimbus" ]; then
+            echo "REPO_TOKEN=${{ secrets.ACTIONS_GITHUB_TOKEN_NIMBUS_ETH2 }}" >> $GITHUB_ENV
+          elif [ "${{ matrix.target.id }}" = "nwaku" ]; then
+            echo "REPO_TOKEN=${{ secrets.ACTIONS_GITHUB_TOKEN_NWAKU }}" >> $GITHUB_ENV
+          elif [ "${{ matrix.target.id }}" = "codex" ]; then
+            echo "REPO_TOKEN=${{ secrets.ACTIONS_GITHUB_TOKEN_NIM_CODEX }}" >> $GITHUB_ENV
+          fi
```
richard-ramos, line 36: I think it should be possible to simplify this a bit:

In the targets you can add the name of the secret variable like this
```
- id: nimbus
  repository: status-im/nimbus-eth2
  ref: unstable
  secret: ACTIONS_GITHUB_TOKEN_NIMBUS_ETH2      <----------
  ```
  
  and then in the `run`:
  ```
echo "REPO_TOKEN=${{ secrets[matrix.target.secret] }} >> $GITHUB_ENV
```

vladopajic: yep, beautifully suggestion! thnx

PR #1422, opened 2025-05-28
chore: bump quic to see if it CI downloads Chronicles v0.10.3

richard-ramos:


narimiran:
Change  https://github.com/vacp2p/nim-libp2p/blob/master/libp2p.nimble#L12

to be `"chronicles >= 0.10.3 & < 0.11.0"`.

`chronicles` 0.11.0 is NOT compatible with Nim 1.6. 
And neither is the latest 0.10.3 commit (`a8fb38a`) - that's why in https://github.com/vacp2p/nim-libp2p/blob/master/.pinned I chose the `81a4a7a` commit hash when I updated that file. That's the latest one that still works with Nim 1.6.

richard-ramos:
Closing PR as it was just to test. Thanks for the solution, @narimiran,  it was introduced in this PR: https://github.com/vacp2p/nim-libp2p/pull/1423

PR #1432, opened 2025-05-30
chore: Add some more context when an exception is caught

Ivansete-status:
Instead of re-raising any caught exception, this adds some more context.
For example:

```nim
 except Exception as e:
      raise newException(CatchableError, e.msg)
```

is replaced with:
```nim
except Exception as e:
      raise newException(CatchableError, "Exception calling bLPop: " & e.msg, e)
```

### DIFF
--- a/interop/hole-punching/hole_punching.nim
+++ b/interop/hole-punching/hole_punching.nim
@@ -84,10 +84,8 @@ proc main() {.async.} =
     debug "Dialing relay...", relayMA
     let relayId = await switch.connect(relayMA).wait(30.seconds)
     debug "Connected to relay", relayId
-  except AsyncTimeoutError:
-    raise newException(
-      CatchableError, "Connection to relay timed out: " & getCurrentExceptionMsg()
-    )
+  except AsyncTimeoutError as e:
+    raise newException(CatchableError, "Connection to relay timed out: " & e.msg, e)
 
   # Wait for our relay address to be published
   while not switch.peerInfo.addrs.anyIt(it.contains(multiCodec("p2p-circuit")).tryGet()):
@@ -132,8 +130,8 @@ try:
     return "done"
 
   discard waitFor(mainAsync().wait(4.minutes))
-except AsyncTimeoutError:
-  error "Program execution timed out: " & getCurrentExceptionMsg()
+except AsyncTimeoutError as e:
+  error "Program execution timed out", description = e.msg
   quit(-1)
 except CatchableError as e:
   error "Unexpected error", description = e.msg
--- a/interop/hole-punching/hole_punching.nim
+++ b/interop/hole-punching/hole_punching.nim
@@ -65,7 +65,7 @@ proc main() {.async.} =
     try:
       redisClient.bLPop(@["RELAY_TCP_ADDRESS"], 0)
     except Exception as e:
-      raise newException(CatchableError, "Exception calling bLPop: " & e.msg, e)
+      raise newException(CatchableError, e.msg)
 
   debug "All relay addresses", relayAddr
 
--- a/interop/hole-punching/hole_punching.nim
+++ b/interop/hole-punching/hole_punching.nim
@@ -65,7 +65,7 @@ proc main() {.async.} =
     try:
       redisClient.bLPop(@["RELAY_TCP_ADDRESS"], 0)
     except Exception as e:
-      raise newException(CatchableError, e.msg)
+      raise newException(CatchableError, "Exception calling bLPop: " & e.msg, e)
 
   debug "All relay addresses", relayAddr
 
@@ -85,7 +85,9 @@ proc main() {.async.} =
     let relayId = await switch.connect(relayMA).wait(30.seconds)
     debug "Connected to relay", relayId
   except AsyncTimeoutError:
-    raise newException(CatchableError, "Connection to relay timed out")
+    raise newException(
+      CatchableError, "Connection to relay timed out: " & getCurrentExceptionMsg()
+    )
 
   # Wait for our relay address to be published
   while not switch.peerInfo.addrs.anyIt(it.contains(multiCodec("p2p-circuit")).tryGet()):
@@ -103,7 +105,7 @@ proc main() {.async.} =
       try:
         PeerId.init(redisClient.bLPop(@["LISTEN_CLIENT_PEER_ID"], 0)[1]).tryGet()
       except Exception as e:
-        raise newException(CatchableError, e.msg)
+        raise newException(CatchableError, "Exception init peer: " & e.msg, e)
 
     debug "Got listener peer id", listenerId
     let listenerRelayAddr = MultiAddress.init($relayMA & "/p2p-circuit").tryGet()
@@ -131,7 +133,7 @@ try:
 
   discard waitFor(mainAsync().wait(4.minutes))
 except AsyncTimeoutError:
-  error "Program execution timed out."
+  error "Program execution timed out: " & getCurrentExceptionMsg()
   quit(-1)
 except CatchableError as e:
   error "Unexpected error", description = e.msg
--- a/interop/transport/main.nim
+++ b/interop/transport/main.nim
@@ -47,12 +47,9 @@ proc main() {.async.} =
         MultiAddress.init("/ip4/" & ip & "/udp/0/quic-v1").tryGet()
       )
   of "ws":
-    discard switchBuilder
-      .withTransport(
-        proc(upgr: Upgrade, privateKey: PrivateKey): Transport =
-          WsTransport.new(upgr)
+    discard switchBuilder.withWsTransport().withAddress(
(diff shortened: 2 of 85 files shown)


vladopajic:
additional context is always nice.

however, i am not sure about the two usecases:
- in case of any timeout, a message from that error will not add any value.

- when handling a cancelled error, we don't really care at which point code got cancelled. when handling this error we just stop, do any cleanup, and return. but we may add logging if necessary.

vladopajic (commented):
additional context is always nice.

libp2p/protocols/connectivity/relay/relay.nim
```diff
@@ -403,7 +406,7 @@ proc deletesReservation(r: Relay) {.async: (raises: [CancelledError]).} =
         if n > r.rsvp[k]:
           r.rsvp.del(k)
     except KeyError:
-      raiseAssert "checked with in"
+      raiseAssert "checked with in: " & getCurrentExceptionMsg()
```
vladopajic, line 409: `raiseAssert "checked with"` constructs must never happen, they are there just to make compiler happy. there is no need in adding message here.

interop/hole-punching/hole_punching.nim
```diff
@@ -65,7 +65,7 @@ proc main() {.async.} =
     try:
       redisClient.bLPop(@["RELAY_TCP_ADDRESS"], 0)
     except Exception as e:
```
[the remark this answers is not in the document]
richard-ramos: I like this idea. We can probably use something like this (copying part of the logic from newException), and maybe even add a rule for the linting to make sure we use this instead of newException (something similar is is/was done for status-go to make sure go routines always call `defer common.LogOnPanic()`)

```nim
template newExceptionWithContext(exceptn: typedesc, message: string;
                       parentException: ref Exception = nil): untyped =
  ## Creates an exception object of type `exceptn` and sets its `msg` field
  ## to `message`, concatenating the parent exception message. Returns the 
  ## new exception object.
  let msg = if parentException.isNil:
      message
   else:
      message & ": " & parentException.msg
  (ref exceptn)(msg: msg, parent: parentException)  
```

tests/commoninterop.nim
```diff
@@ -260,7 +260,7 @@ proc commonInteropTests*(name: string, swCreator: SwitchCreator) =
           check line == test
           testFuture.complete(line)
         except CancelledError as e:
-          raise e
+          raise newException(CancelledError, "Native handler cancelled: " & e.msg, e)
```
richard-ramos, line 263: Cancel errors should probably be re-raised without modification to maintain the call stack.

libp2p/wire.nim
```diff
@@ -146,7 +148,8 @@ proc createStreamServer*[T](
       initTAddress(ma).tryGet(), flags, udata, sock, backlog, bufferSize, child, init
     )
   except CatchableError as exc:
-    raise newException(LPError, exc.msg)
+    raise
+      newException(LPError, "Exception in createStreamServer simpler: " & exc.msg, exc)
```
richard-ramos, line 152: Adding `Exception in` seems redundant. The type already conveys this is an exception.
(This comment applies to other messages that contain `Exception in `

libp2p/utils/future.nim
```diff
@@ -29,9 +29,12 @@ proc anyCompleted*[T](
...
     except ValueError:
       raise newException(
-        AllFuturesFailedError, "None of the futures completed successfully"
+        AllFuturesFailedError,
+        "None of the futures completed successfully: " & getCurrentExceptionMsg(),
       )
     except CancelledError as exc:
-      raise exc
+      raise newException(
+        CancelledError, "One of the futures was cancelled: " & exc.msg, exc
```
richard-ramos, line 37: Should probably continue to be `raise exc` to maintain the call stack

libp2p/transports/wstransport.nim
```diff
@@ -304,12 +306,14 @@ method accept*(
...
   except CancelledError as exc:
-    raise exc
+    raise newExpception(CancelledError, "WsTransport accept cancelled: " & exc.msg, exc)
   except TransportOsError as exc:
     debug "OS Error", description = exc.msg
   except CatchableError as exc:
     info "Unexpected error accepting connection", description = exc.msg
-    raise newException(transport.TransportError, exc.msg, exc)
+    raise newException(
+      transport.TransportError, "caught error in WsTransport accept: " & exc.msg, exc
```
richard-ramos, line 315: can probably just be:
```suggestion
      transport.TransportError, "Error in WsTransport accept: " & exc.msg, exc
```

libp2p/transports/wstransport.nim
```diff
@@ -304,12 +306,14 @@ method accept*(
     debug "Server was closed", description = exc.msg
     raise newTransportClosedError(exc)
   except CancelledError as exc:
-    raise exc
+    raise newExpception(CancelledError, "WsTransport accept cancelled: " & exc.msg, exc)
```
richard-ramos, line 309: Should probably continue to be `raise exc` to maintain the call stack

libp2p/transports/wstransport.nim
```diff
@@ -269,9 +271,9 @@ method accept*(
     try:
       await one(self.acceptFuts)
     except ValueError:
-      raiseAssert("already checked with if")
+      raiseAssert("already checked with if: " & getCurrentExceptionMsg())
```
richard-ramos, line 274:
```suggestion
      raiseAssert("already checked with if")
```
No need to add the exception here. This is an error that wont happen and just added to make the compiler happy

libp2p/transports/wstransport.nim
```diff
@@ -269,9 +271,9 @@ method accept*(
...
     except CancelledError as e:
-      raise e
+      raise newException(CancelledError, "error in WsTransport accept: " & e.msg, e)
```
richard-ramos, line 276: Should probably continue to be `raise e` to maintain the call stack

richard-ramos (commented):
Was going to comment each line of CancelledError with same comment, but saw that there are too many :) . Here's a comment that applies to all: They do not need to be modified, and can be kept as `raise exc`.  

Wrapping the exception vs re-raising it has some differences: with the former, you're creating a new error. The stacktrace is going to be from where the new exception was raised and not where the original cancelled error happened. If you keep the original exception, the trace is preserved.

interop/hole-punching/hole_punching.nim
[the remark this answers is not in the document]
Ivansete-status: That's a great idea indeed, but I was aiming to suggest a simpler "refactor" without adding more "complexity."

The "prepending" context makes more sense when using `Result`, e.g., `return err("Could not run: " & $error)`

-----

> then developers when they copy-paste code will copy-past without thinking of propagating error message?

That won't happen because reviewers will notice in the upcoming PRs that the context is not properly added :)
The important thing is to keep the idea in the minds of all reviewers ( if you like the idea of adding context, of course ;P .)
We can create new procs, but they might not be used in the future either—there's no way to enforce the use of the new template.

richard-ramos (approved):
LGTM

interop/hole-punching/hole_punching.nim
```diff
@@ -85,7 +85,9 @@ proc main() {.async.} =
     let relayId = await switch.connect(relayMA).wait(30.seconds)
     debug "Connected to relay", relayId
   except AsyncTimeoutError:
-    raise newException(CatchableError, "Connection to relay timed out")
+    raise newException(
+      CatchableError, "Connection to relay timed out: " & getCurrentExceptionMsg()
+    )
```
gmelodie: Is there a difference between this and 
```nim
except AsyncTimeoutError as exc:
  error "Program execution timed out: " & exc.msg
```
Same for all occurrences of `getCurrentExceptionmsg()`

Ivansete-status: @gmelodie - I'm afraid you refer to the use of `getCurrentExceptionMsg()`?

Both options have the same result.
The benefit of `getCurrentExceptionMsg()` is that it works in an catch-all `except:` clauses too.

Is the approach we use to use at nwaku, when handling exceptions.

Let me know if that is suitable to you or need me to use the `as exc` approach instead.

arnetheduck: > works in an catch-all except:

we avoid catch-all since they are undefined behaviour for Defect.

> as exc

it's better to use this style consistently as getCurrentExceptionMsg can be problematic in nested exception handlers.

interop/transport/main.nim
```diff
@@ -83,7 +83,7 @@ proc main() {.async.} =
       try:
         redisClient.bLPop(@["listenerAddr"], testTimeout.seconds.int)[1]
       except Exception as e:
```
arnetheduck, line 82: broadly, we should not using `except:` and `except Exception:` since this is undefined behavior with respect to Defect

gmelodie (approved):

PR #1436, opened 2025-06-02
feat(autotls): add acme client

gmelodie:
Adds ACME functionality that AutoTLS needs (`libp2p/autotls/acme.nim`): send signed requests to ACME server to request and finalize challenges

Reference: https://github.com/vacp2p/nim-libp2p/pull/1358

### DIFF
--- a/libp2p/autotls/acme/api.nim
+++ b/libp2p/autotls/acme/api.nim
@@ -1,4 +1,4 @@
-import options, base64, sequtils, json
+import options, base64, sequtils, strutils, json
 from times import DateTime, parse
 import chronos/apps/http/httpclient, jwt, results, bearssl/pem
 
@@ -21,7 +21,6 @@ const
 
 type Nonce = string
 type Kid = string
-type AccountStatus = string
 
 type ACMEDirectory = object
   newNonce: string
@@ -58,17 +57,28 @@ type ACMERegisterRequest* = object
   termsOfServiceAgreed: bool
   contact: seq[string]
 
+type ACMEAccountStatus = enum
+  valid
+  deactivated
+  revoked
+
 type ACMERegisterResponseBody = object
-  status*: AccountStatus
+  status*: ACMEAccountStatus
 
 type ACMERegisterResponse* = object
   kid*: Kid
-  status*: AccountStatus
+  status*: ACMEAccountStatus
+
+type ACMEChallengeStatus* {.pure.} = enum
+  pending = "pending"
+  processing = "processing"
+  valid = "valid"
+  invalid = "invalid"
 
 type ACMEChallenge = object
   url*: string
   `type`*: string
-  status*: string
+  status*: ACMEChallengeStatus
   token*: string
 
 type ACMEChallengeIdentifier = object
@@ -79,12 +89,12 @@ type ACMEChallengeRequest = object
   identifiers: seq[ACMEChallengeIdentifier]
 
 type ACMEChallengeResponseBody = object
-  status: string
+  status: ACMEChallengeStatus
   authorizations: seq[string]
   finalize: string
 
 type ACMEChallengeResponse* = object
-  status: string
+  status: ACMEChallengeStatus
   authorizations: seq[string]
   finalize: string
   orderURL: string
@@ -100,8 +110,23 @@ type ACMEAuthorizationsResponse* = object
 type ACMECompletedResponse* = object
   checkURL: string
 
+type ACMEOrderStatus* {.pure.} = enum
+  pending = "pending"
+  ready = "ready"
+  processing = "processing"
+  valid = "valid"
+  invalid = "invalid"
+
+type ACMECheckKind = enum
+  ACMEOrderCheck
+  ACMEChallengeCheck
+
 type ACMECheckResponse* = object
-  status: string
+  case kind: ACMECheckKind
+  of ACMEOrderCheck:
+    orderStatus: ACMEOrderStatus
+  of ACMEChallengeCheck:
+    chalStatus: ACMEChallengeStatus
   retryAfter: Duration
 
 type ACMEFinalizedResponse* = object
@@ -159,9 +184,11 @@ proc acmeHeader(
   if not needsJwk and kid.isNone:
     raise newException(ACMEError, "kid not set")
 
+  if key.pubkey.scheme != PKScheme.RSA or key.seckey.scheme != PKScheme.RSA:
+    raise newException(ACMEError, "Unsupported signing key type")
+
   let newNonce = await self.requestNonce()
   if needsJwk:
-    # TODO: check if scheme is RSA
     let pubkey = key.pubkey.rsakey
     let nArray = @(getArray(pubkey.buffer, pubkey.key.n, pubkey.key.nlen))
     let eArray = @(getArray(pubkey.buffer, pubkey.key.e, pubkey.key.elen))
@@ -279,7 +306,7 @@ proc requestChallenge*(
   )
 
 proc requestCheck(
-    self: ACMEApi, checkURL: string, key: KeyPair, kid: Kid
+    self: ACMEApi, checkURL: string, checkKind: ACMECheckKind, key: KeyPair, kid: Kid
 ): Future[ACMECheckResponse] {.async: (raises: [ACMEError, CancelledError]).} =
   handleError("requestCheck"):
     let rawResponse =
@@ -292,7 +319,25 @@ proc requestCheck(
       except ValueError:
         DefaultChalCompletedRetryTime
 
-    ACMECheckResponse(status: body["status"].getStr, retryAfter: retryAfter)
+    case checkKind
+    of ACMEOrderCheck:
+      try:
+        ACMECheckResponse(
+          kind: checkKind,
+          orderStatus: parseEnum[ACMEOrderStatus](body["status"].getStr),
+          retryAfter: retryAfter,
+        )
+      except ValueError:
+        raise newException(ACMEError, "Invalid order status: " & body["status"].getStr)
+    of ACMEChallengeCheck:
+      try:
+        ACMECheckResponse(
+          kind: checkKind,
+          chalStatus: parseEnum[ACMEChallengeStatus](body["status"].getStr),
+          retryAfter: retryAfter,
+        )
+      except ValueError:
+        raise newException(ACMEError, "Invalid order status: " & body["status"].getStr)
 
 proc requestCompleted(
     self: ACMEApi, chalURL: string, key: KeyPair, kid: Kid
@@ -314,17 +359,18 @@ proc challengeCompleted*(
   let completedResponse = await self.requestCompleted(chalURL, key, kid)
   # check until acme server is done (poll validation)
   for i in 0 .. retries:
-    let checkResponse = await self.requestCheck(completedResponse.checkURL, key, kid)
-    case checkResponse.status
-    of "pending":
+    let checkResponse =
+      await self.requestCheck(completedResponse.checkURL, ACMEChallengeCheck, key, kid)
+    case checkResponse.chalStatus
+    of ACMEChallengeStatus.pending:
       await sleepAsync(checkResponse.retryAfter) # try again after some delay
-    of "valid":
+    of ACMEChallengeStatus.valid:
       return
     else:
       raise newException(
         ACMEError,
-        "Failed challenge completion: expected 'valid', got '" & checkResponse.status &
-          "'",
+        "Failed challenge completion: expected 'valid', got '" &
+          $checkResponse.chalStatus & "'",
       )
 
 proc requestFinalize*(
@@ -356,11 +402,11 @@ proc finalizeCertificate*(
   handleError("finalizeCertificate (check finalized)"):
     var checkResponse: ACMECheckResponse
     for i in 0 .. retries:
-      let checkResponse = await self.requestCheck(orderURL, key, kid)
-      case checkResponse.status
-      of "valid":
+      let checkResponse = await self.requestCheck(orderURL, ACMEOrderCheck, key, kid)
+      case checkResponse.orderStatus
+      of ACMEOrderStatus.valid:
         return true
-      of "processing":
+      of ACMEOrderStatus.processing:
         await sleepAsync(checkResponse.retryAfter) # try again after some delay
       else:
         return false
--- a/tests/testautotls_integration.nim
+++ b/tests/testautotls_integration.nim
@@ -45,7 +45,7 @@ suite "AutoTLS Integration":
 
     check challenge.dns01.url.len() > 0
     check challenge.dns01.`type`.len() > 0
-    check challenge.dns01.status.len() > 0
+    check challenge.dns01.status == ACMEChallengeStatus.pending
     check challenge.dns01.token.len() > 0
 
   asyncTest "test register with unsupported keys":


libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,329 @@
...
+  let challengeResponse =
+    await self.signedAcmeRequest(self.directory.newOrder, orderPayload)
+  let challengeResponseBody = await challengeResponse.getParsedResponseBody()
+  let orderURL = challengeResponse.headers.getString("location")
+  if orderURL == "":
+    raise newException(ACMEError, "'location' header not found in ACME response")
+  let finalizeURL = challengeResponseBody.getJSONField("finalize").getStr
+
+  # get challenges
+  let authzURL = challengeResponseBody.getJSONField("authorizations")[0].getStr
```
vladopajic: could we create type for `challengeResponse` then parse using that type as schema?

```
   type
    ChallengeResponse = object
      authorizations: ...
      finalize: ...
      
  var resp = Json.decode(rawJson, ChallengeResponse)
```

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,329 @@
...
+    await self.signedAcmeRequest(self.directory.newAccount, payload, needsJwk = true)
+  let jsonResponseBody = await response.getParsedResponseBody()
+  if response.status != HttpCreated:
+    raise newException(
+      ACMEError, "Unable to register with ACME server: " & $jsonResponseBody
+    )
+  self.kid = Opt.some(response.headers.getString("location"))
+  self.status = Opt.some(jsonResponseBody.getJSONField("status").getStr)
+
+proc requestChallenge*(
```
vladopajic: by reading signature of this method is unclear what it returns.

could `requestChallenge` return user defined type, that will hold any data that it needs to return? instead of using `JsonNode` - generic type, string, string..

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,329 @@
...
+    if downloadResponse.status != HttpOk:
+      raise newException(ACMEError, "Failed to download certificate")
+
+    let certificateInfoBody =
+      bytesToString(await downloadResponse.getBodyBytes()).checkedParseJson()
+
+    let certificateDownloadURL = certificateInfoBody.getJSONField("certificate").getStr
+    let certificateExpiry = parse(
+      certificateInfoBody.getJSONField("expires").getStr, "yyyy-MM-dd'T'HH:mm:ss'Z'"
+    )
```
vladopajic: could we create type that represents response object and use that type as schema for deserialization?

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,329 @@
...
+  DefaultFinalizeRetryTime = 1.seconds
+  DefaultRandStringSize = 256
+
+type ACMEDirectory = object
+  newNonce: string
+  newOrder: string
+  newAccount: string
+
+type ACMEAccount* = object
+  status*: Opt[string]
```
vladopajic: `status` - access to API should not hold any state.

libp2p/autotls/acme.nim
```diff
@@ -0,0 +1,329 @@
...
+  contact*: Opt[seq[string]]
```
vladopajic: `contact` - access to API should not hold any state.

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,354 @@
...
+    ).body,
+    ACMERegisterResponse,
+  )
+
+proc requestChallenge*(
+    self: ref ACMEApi, domains: seq[string], key: KeyPair, kid = Kid
+): Future[ACMEChallengeResponse] {.async: (raises: [ACMEError, CancelledError]).} =
+  # request challenge from ACME server
+  var identifiers: seq[JsonNode] = domains.mapIt(%*{"type": "dns", "value": it})
+  let orderPayload = %*{"identifiers": identifiers}
```
vladopajic: create appropriate type `ACMEChallengeRequest`

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,354 @@
...
+  return challengeResponse
+
+proc notifyChallengeCompleted*(
+    self: ref ACMEApi,
+    chalURL: string,
+    key: KeyPair,
+    kid: Kid,
+    retries: int = DefaultChalCompletedRetries,
+): Future[void] {.async: (raises: [ACMEError, SerializationError, CancelledError]).} =
+  let emptyPayload = newJObject()
```
vladopajic: can you pass nil or something uninitialized?

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,354 @@
...
+    let rawResponse =
+      try:
+        await HttpClientRequestRef
+        .get(self.session, completedResponse.checkURL)
+        .get()
+        .send()
+      except HttpError as exc:
+        raise newException(ACMEError, "Failed to connect to ACME server", exc)
+      except ValueError as exc:
+        raise newException(ACMEError, "Failed to parse JSON", exc)
```
vladopajic: no parsing is happening here. check if value error is needed or change to correct message.

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,354 @@
...
+    raise newException(ACMEError, "Failed to create JWT", exc)
+  except HttpError as exc:
+    raise newException(ACMEError, "Failed to send HTTP request to the ACME server", exc)
+
+proc acmeRegister*(
+    self: ref ACMEApi, key: KeyPair
+): Future[ACMERegisterResponse] {.
+    async: (raises: [ACMEError, SerializationError, CancelledError])
+.} =
+  let payload = %*{"termsOfServiceAgreed": true}
```
vladopajic: create appropriate type for payload

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,354 @@
...
+    self: ref ACMEApi,
+    url: string,
+    payload: auto,
+    key: KeyPair,
+    needsJwk: bool = false,
+    kid: Opt[Kid] = Opt.none(Kid),
+): Future[ACMEResponse] {.async: (raises: [ACMEError, CancelledError]).} =
+  try:
+    let acmeHeader = await self.acmeHeader(url, key, needsJwk, kid)
+    var token = toJWT(%*{"header": acmeHeader, "claims": payload})
```
vladopajic, line 236: create type 

```
AcmeRequest type = 
   header: AcmeHeader
   claims: auto
```

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,354 @@
...
+
+  # TODO: check if scheme is RSA
+  let pubkey = key.pubkey.rsakey
+  let nArray = @(getArray(pubkey.buffer, pubkey.key.n, pubkey.key.nlen))
+  let eArray = @(getArray(pubkey.buffer, pubkey.key.e, pubkey.key.elen))
+  let n = base64UrlEncode(nArray)
+  let e = base64UrlEncode(eArray)
+
+  let newNonce = await self.newNonce()
+  var header = %*{"alg": Alg, "typ": "JWT", "nonce": newNonce, "url": url}
```
vladopajic: convert this to type as well.

libp2p/autotls/acme/account.nim
```diff
@@ -0,0 +1,33 @@
+import options, serialization, json_serialization, tables
+import chronos/apps/http/httpclient, results
+from std/json import
+  JsonNode, `%*`, `%`, `[]`, `[]=`, `$`, parseJson, getStr, items, newJObject, to,
+  JsonKindError
```
vladopajic: simplify imports

libp2p/autotls/utils.nim
```diff
@@ -0,0 +1,89 @@
+import
```
vladopajic: drop things form this file that are not needed with this pr. we'll introduce those once they are needed or we might not if we find different solution.

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,354 @@
...
+  let authorizationsResponse =
+    try:
+      Json.decode(
+        (
+          await HttpClientRequestRef
+          .get(self.session, challengeResponse.authzURL)
+          .get()
+          .send()
+          .getResponseBody()
+        ),
```
vladopajic, line 216: `getResponseBody()` and in request above we use `.body`, could we unify to better approach?

libp2p/autotls/utils.nim
vladopajic: also if these utilities are needed by `acme` move them to that dir

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,342 @@
...
+  try:
+    body
+  except JsonKindError as exc:
+    raise newException(ACMEError, "Failed to decode JSON", exc)
+  except exceptions.KeyError as exc:
+    raise newException(ACMEError, "Failed to decode JSON", exc)
+  except ValueError as exc:
+    raise newException(ACMEError, "Failed to decode JSON", exc)
+  except HttpError as exc:
+    raise newException(ACMEError, "Failed to connect to ACME server", exc)
```
vladopajic: interesting!

- name is not ideal, perhaps "handle error"?
- when errors are raised using this, it will be impossible to tell which place it raised as there are may places with same error message
- why is it used on some places and on other it isn't? can we have have unified usage?

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,342 @@
...
+    let finalizedResponse = wrapSerialization:
+      (await self.signedAcmeRequest(finalizeURL, payload, key, kid = Opt.some(kid))).body.to(
+        ACMEFinalizedResponse
+      )
+    let rawResponse =
+      try:
+        await HttpClientRequestRef.get(self.session, orderURL).get().send()
+      except CatchableError as exc:
+        raise
+          newException(ACMEError, "Unexpected error while finalizing certificate", exc)
```
vladopajic: error here  is probably not correctly handled -> if `CancelledError` happens here it will be convered to `ACMEError`, which is not what we really want.
whenever there is `await` within `try` you need to handle `CancelledError` separately and re-raise this exception without converting it. additional you may need to preform cleanups related current method call (if any).

vladopajic: the same applies to other places in this pr

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,342 @@
...
+): Future[ACMERegisterResponse] {.async: (raises: [ACMEError, CancelledError]).} =
+  let payload = %*{"termsOfServiceAgreed": true}
+  wrapSerialization:
+    let acmeResponse = (
+      await self.signedAcmeRequest(
+        self.directory.newAccount, payload, key, needsJwk = true
+      )
+    )
+    ACMERegisterResponse(
+      status: acmeResponse.body["status"].getStr,
```
vladopajic: it would be better to follow through the same convention to decoded response using type. and afterwords to manually set fields from header.

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,342 @@
...
+): Future[ACMEChallengeResponse] {.async: (raises: [ACMEError, CancelledError]).} =
+  # request challenge from ACME server
+  var identifiers: seq[JsonNode] = domains.mapIt(%*{"type": "dns", "value": it})
+  let orderPayload = %*{"identifiers": identifiers}
+
+  var challengeResponse = wrapSerialization:
+    (
+      await self.signedAcmeRequest(
+        self.directory.newOrder, orderPayload, key, kid = Opt.some(kid)
+      )
```
vladopajic: instead of doing request via `signedAcmeRequest` 
 ```
 await self.signedAcmeRequest(self.directory.newOrder, orderPayload, key, kid = Opt.some(kid))
 ```
 
 consider refactoring this in following way:
 
 change `signedAcmeRequest` to `createSignedAcmeRequest` which will take in necessery parameters and return signed payload - basically copy-past everything from `signedAcmeRequest` with return at `$token.toFlattenedJson()` line.
 
 this way you can uniformly handle http request in the code:
 
 ```
let payload = createSignedAcmeRequest(...)
await HttpClientRequestRef.post(url, payload, ....)
 ```

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+
+type ACMEOrderResponse* = object
+  certificate: string
+  expires: string
+
+type ACMECertificateResponse* = object
+  rawCertificate: string
+  certificateExpiry: DateTime
+
+template handleError(msg: string, body: untyped): untyped =
```
vladopajic, line 142: `handleError` must propagate CancelledError, otherwise it will be caught by `CatchableError`

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+  let personalizationStr = "libp2p_autotls"
+  if cert_init_drbg(
+    personalizationStr.cstring, personalizationStr.len.csize_t, certCtx.addr
+  ) != CERT_SUCCESS:
+    raise newException(ACMEError, "Failed to initialize certCtx")
+  if cert_generate_key(certCtx, certKey.addr) != CERT_SUCCESS:
+    raise newException(ACMEError, "Failed to generate cert key")
+
+  if cert_signing_req(domain.cstring, certKey, derCSR.addr) != CERT_SUCCESS:
+    raise newException(ACMEError, "Failed to create CSR")
```
vladopajic: could this code be extracted to separate function? 

why?
- it would make `finalizeCertificate` shorter, thus easier to read
- that utility function could be tested individually
- this utility function could be reused or maybe it already should be?
- maybe it was supposed to be implemented in `tls` package...
- ....

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+      )
+    let (retryAfterHeader, checkResponse) = handleError(
+      "notifyChallengeCompleted (check " & $i & ")"
+    ):
+      let rawResponse =
+        await HttpClientRequestRef.get(self.session, orderURL).get().send()
+      (
+        rawResponse.headers.getString("Retry-After"),
+        (await rawResponse.getResponseBody()).to(ACMECheckResponse),
+      )
```
vladopajic: i have no idea what's happening here... unless i use all my brain cells

libp2p/autotls/acme/utils.nim
```diff
@@ -0,0 +1,24 @@
...
+  encoded.removeSuffix("=")
+  encoded.removeSuffix("=")
+  return encoded
+
+proc getResponseBody*(
+    response: HttpClientResponseRef
+): Future[JsonNode] {.async: (raises: [ACMEError]).} =
+  try:
+    let responseBody = bytesToString(await response.getBodyBytes()).parseJson()
+    return responseBody
```
vladopajic, line 19: because  of `await`, code must raise `CatchableError` without converting it to `ACMEError`

gmelodie: Do you mean `CancelledError`?

libp2p/autotls/acme/utils.nim
```diff
@@ -0,0 +1,24 @@
...
+type ACMEError* = object of LPError
+
+proc base64UrlEncode*(data: seq[byte]): string =
+  ## Encodes data using base64url (RFC 4648 §5) — no padding, URL-safe
+  var encoded = base64.encode(data, safe = true)
+  encoded.removeSuffix("=")
+  encoded.removeSuffix("=")
+  return encoded
+
+proc getResponseBody*(
```
vladopajic, line 14: if we utilize `nim-json-serialization` lib, we will not need this method at all.

code would look like:
```
let rawJson = bytesToString(await response.getBodyBytes())
Json.decode(rawJson, <type>)
```

can we try with this?

libp2p/autotls/acme/utils.nim
vladopajic: yes :)

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+): Future[ACMERequestHeader] {.async: (raises: [ACMEError, CancelledError]).} =
+  if not needsJwk and kid.isNone:
+    raise newException(ACMEError, "kid not set")
+
+  # TODO: check if scheme is RSA
+  let pubkey = key.pubkey.rsakey
+  let nArray = @(getArray(pubkey.buffer, pubkey.key.n, pubkey.key.nlen))
+  let eArray = @(getArray(pubkey.buffer, pubkey.key.e, pubkey.key.elen))
+  let n = base64UrlEncode(nArray)
+  let e = base64UrlEncode(eArray)
```
vladopajic: nit: consider moving this to part of if condition where it is used

libp2p/autotls/acme/utils.nim
vladopajic: afaik, this try/catch was because of parseJson? so if we can avoid using we will not need this code at all

gmelodie: `nim-json-serialization` versions are clashing with other packages that use it as a dependency. I don't want to use external libs when there's `std` native support

tests/testautotls.nim
```diff
@@ -0,0 +1,45 @@
...
+import ./helpers
+
+suite "AutoTLS":
+  teardown:
+    checkTrackers()
+
+  asyncTest "test ACME":
+    let api = await ACMEApi.new(acmeServerURL = LetsEncryptURLStaging)
+    defer:
+      await api.session.closeWait()
```
vladopajic: `await api.session.closeWait()`
could this be 
`await api.close()`

usually code that creates something (in this case ACMEApi instance) should not access individual properties to close, but rather type should expose method to close all resources

tests/testautotls.nim
```diff
@@ -0,0 +1,45 @@
...
+    let key = KeyPair.random(PKScheme.RSA, newRng()[]).get()
+    let registerResponse = await api.acmeRegister(key)
+    # account was registered (kid set)
+    check registerResponse.kid != ""
+    if registerResponse.kid == "":
```
vladopajic: is it really necessary to have `check` and `if`?

tests/testautotls.nim
```diff
@@ -0,0 +1,45 @@
...
+import chronos/apps/http/httpclient
+import ../libp2p/[stream/connection, upgrademngrs/upgrade, autotls/acme/api, wire]
+
+import ./helpers
+
+suite "AutoTLS":
+  teardown:
+    checkTrackers()
+
+  asyncTest "test ACME":
```
vladopajic: consider adding more tests, ones where error happen.
- register with unsupported key
- requestChallenge with unsupported key + invalid kid

consider testing other endpoints:
- finalizeCertificate
- downloadCertificate
- notifyChallengeCompleted

tests/testautotls.nim
```diff
@@ -0,0 +1,45 @@
+{.used.}
```
vladopajic: - what is request rate limit for `LetsEncryptURLStaging` api?
- if we are doing these request for each commit in CI for each developer are we going to overdo it?
   - consider 18 jobs (os x nim version x architecture)  x number of commits daily say 100 
  - if so, can we run daily tests with additional environment variable (eg. `DAILY_TEST='yes'`) and run api tests only if `let isDaily = getEnv("DAILY_TEST", "no") == "yes"`

libp2p/autotls/acme/utils.nim
vladopajic: clashing in what way? 
can we fix it?

we are already using this lib in `nim-libp2p`, [here](https://github.com/vacp2p/nim-libp2p/blob/5407b18b63ace861659612f84e534e748d90dd62/.pinned#L7). dependency is already there we just need to import it?

tests/testautotls.nim
gmelodie: testing other endpoints requires the autotls broker e.g. you can't finalize a certificate that has not passed the challenge test (passing the challenge test requires autotls broker)

tests/testautotls.nim
gmelodie: I wanted to discuss this but did not get many replies: https://discord.com/channels/864066763682218004/1113778107359436821/1375151097135501364

gmelodie: The bigger issue is account creation which has a limit of 10 accounts every 3 hours. We've never hit this limit on github actions. Maybe each test gets interpreted as a different session/machine/ip

tests/testautotls.nim
vladopajic: got it! we can still test and assert that methods are raising exceptions.

tests/testautotls.nim
vladopajic: well, in this case we not only need to run this in daily builds, but we also need to run them only for one variant of (os, nim version, architecture).

vladopajic: definitely we need to start with `isDaily()` check, since if we are running tests locally we shouldn't run these test. and also in CI we might get rate limited if it is deployed to master (when all CIs are running tests). currently only this branch was running tests so i guess we were below threshold because of that.

if we get issues with daily CI, we will need to narrow down the execution. maybe we can address this later, if issues arises? since if CI is passing for this branch it might pass for daily CI.

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+      ).getResponseBody()
+    ).to(ACMEAuthorizationsResponse)
+
+  let dns01 = authorizationsResponse.challenges.filterIt(it.`type` == "dns-01")[0]
+
+  return ACMEChallengeResponseWrapper(
+    finalizeURL: challengeResponse.finalize, orderURL: orderURL, dns01: dns01
+  )
+
+proc notifyChallengeCompleted*(
```
vladopajic: nit:  consider different naming for `notifyChallengeCompleted` method

- usually `foo.notifySomething()` or simply `foo.notify()`  is method that implements information flow in direction from `foo` to `Something` (or anything else implied by context) 
   - foo is notifying (sending notification) to something
   - foo has information and is sending it 
- this method's logic is not really notifying (it's not sending information), but rather it's pooling to get new information - it doesn't have the information

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+  except JsonKindError as exc:
+    raise newException(ACMEError, msg & ": Failed to decode JSON", exc)
+  except ValueError as exc:
+    raise newException(ACMEError, msg & ": Failed to decode JSON", exc)
+  except HttpError as exc:
+    raise newException(ACMEError, msg & ": Failed to connect to ACME server", exc)
+  except CatchableError as exc:
+    raise newException(ACMEError, msg & ": Unexpected error", exc)
+
+proc new*(
```
vladopajic, line 158: consider creating fundamental building blocks (first class citizens) from each endpoint - every endpoint is individual function.
 
- `requestDirectory` (currently this request is embedded in `new` proc, but should be new functions that is called in `new` proc)
- `requestNonce` -> is already implemented
- `requestRegister` -> is already implemented
- `requestNewOrder` -> used in `requestChallenge`
- `requestAuthorizations` -> used in  `requestChallenge` 
- `requestCheck` -> used in `notifyChallengeCompleted`  returns `ACMECheckResponse`
- `requestFinalized -> returns `ACMEFinalizedResponse`
-  `requestGetOrder` -> returns`ACMEOrderResponse`
- ... and others if missing in this list

note: naming of functions here is just to get an idea - feel free to name them

why?
- allows better reuse of code (observe code changes after implementing suggestion)
- allows better testing (even if we can't test success path of method, we can assert unsuccessful path)
- introducing convention (patterns for making requests in this case) makes code easier to maintain 
  - following same convention by itself have consequences, like identifying bugs (if something follows convention it's harder to introduce issues (as this patter is already tested in other places), and easier to spot bugs ("whys is this code <pattern> like this? should it be that??? oh this must be a bug!")
- new functions make code of composed functions easier to read (observe implications on `notifyChallengeCompleted`)
- smaller building blocks, makes code easier to maintain
  - hypothetically, consider that logic in `notifyChallengeCompleted` has to be changed in differrent way but it requires the same building blocks (`requestXxxxx` methods). if building blocks are defined then new implementation will just used building block in different way it would not need to repeat re-implementing the same building blocks - potentially introducing issues in those spots (even if copy-paste'ed). 
- ...

tests/testautotls.nim
vladopajic: also consider implementing this as the last piece, or disabling it if implemented earlier, so we can have CI check while developing. :)

libp2p/autotls/acme/utils.nim
```diff
@@ -0,0 +1,24 @@
+import base64, strutils, chronos/apps/http/httpclient, json
+import ../../errors
+
+type ACMEError* = object of LPError
```
vladopajic, line 5: nit: maybe move this type to api.nim file?

.pinned
```diff
@@ -17,3 +17,6 @@ testutils;https://github.com/status-im/nim-testutils@#9e842bd58420d23044bc55e160
 unittest2;https://github.com/status-im/nim-unittest2@#8b51e99b4a57fcfb31689230e75595f024543024
 websock;https://github.com/status-im/nim-websock@#d5cd89062cd2d168ef35193c7d29d2102921d97e
 zlib;https://github.com/status-im/nim-zlib@#daa8723fd32299d4ca621c837430c29a5a11e19a
+jwt;https://github.com/vacp2p/nim-jwt@#18f8378de52b241f321c1f9ea905456e89b95c6f
+bearssl_pkey_decoder;https://github.com/vacp2p/bearssl_pkey_decoder@#21dd3710df9345ed2ad8bf8f882761e07863b8e0
+bio;https://github.com/xzeshen/bio@#0f5ed58b31c678920b6b4f7c1783984e6660be97
```
vladopajic, line 22: let's not import dependencies that are not needed by this pr.  is `bio` needed here? is `bearssl_pkey_decoder` needed here?

libp2p.nimble
```diff
@@ -10,7 +10,8 @@ skipDirs = @["tests", "examples", "Nim", "tools", "scripts", "docs"]
 requires "nim >= 1.6.0",
   "nimcrypto >= 0.6.0 & < 0.7.0", "dnsclient >= 0.3.0 & < 0.4.0", "bearssl >= 0.2.5",
   "chronicles >= 0.10.3 & < 0.11.0", "chronos >= 4.0.4", "metrics", "secp256k1",
-  "stew >= 0.4.0", "websock >= 0.2.0", "unittest2", "results", "quic >= 0.2.7"
+  "stew >= 0.4.0", "websock >= 0.2.0", "unittest2", "results", "quic >= 0.2.7", "bio",
```
vladopajic, line 13: https://github.com/vacp2p/nim-libp2p/pull/1436/files#r2126431655

libp2p/autotls/acme/api.nim
gmelodie: Following the same convention would require us to have Kid as an optional field, but the option would have to be an `std` `Option` instead of what we normally use (`Opt`). I think it's best to leave it as is

.pinned
gmelodie: yes

libp2p/autotls/acme/utils.nim
gmelodie: dependency cycle:
`utils.nim` would have to import `api`, but `api` imports `utils`

tests/testautotls.nim
gmelodie: it doesn't make sense to continue if account was not registered. All other checks will fail

libp2p/autotls/acme/utils.nim
vladopajic: got it, please consider

- if we leave it here, `ACMEError` will have to be exported via `api.nim` as it doesn't make sense to import it separately
- could we move this utility (that requires  `ACMEError`) to `api.nim` since it's only one small utility? 
- can utility be change to throw some other error ?
- maybe this utility doesnt really need to exist? if nim-json-serializaiton is used?

gmelodie: in that case we might as well move every function in `utils` to `api`, since the whole file is 43 lines of code and only 9 of them would not require ACMEError. Having another error does not make a lot of sense to me

libp2p/autotls/acme/utils.nim
gmelodie: This is what I'm getting when trying to use `Json.decode` instead of `to()`:
```sh
.nimble/pkgs/serialization-0.2.7/serialization.nim(49, 27) Error: undeclared identifier: 'Reader'
candidates (edit distance, scope distance); see '--spellSuggest': 
 (2, 12): 'ReadEof' [enumField declared in /home/gmelodie/.nimble/pkgs/chronos-4.0.4/chronos/transports/common.nim(129, 5)]
 (2, 12): 'ReadEof' [enumField declared in /home/gmelodie/.nimble/pkgs/chronos-4.0.4/chronos/transports/common.nim(129, 5)]
 (2, 12): 'ReadEof' [enumField declared in /home/gmelodie/.nimble/pkgs/chronos-4.0.4/chronos/transports/common.nim(129, 5)]
 (2, 12): 'ReadEof' [enumField declared in /home/gmelodie/.nimble/pkgs/chronos-4.0.4/chronos/transports/common.nim(129, 5)]
 (2, 12): 'headers' [iterator declared in /home/gmelodie/.nimble/pkgs/httputils-0.3.0/httputils.nim(1307, 10)]
 (2, 12): 'headers' [iterator declared in /home/gmelodie/.nimble/pkgs/httputils-0.3.0/httputils.nim(1332, 10)]
```

In libp2p we're only currently using `Json.encode` from `json_serialization`, not `Json.decode`, in `interop/transport/main.nim`. I think we best avoid that for now, I spent some time looking at it but seems like an issue with versions that need to be updated.

tests/testautotls.nim
gmelodie: > since if we are running tests locally we shouldn't run these test

why not?

tests/testautotls.nim
```diff
@@ -0,0 +1,69 @@
...
+# import std/[strformat, net] # uncomment after re-enabling AutoTLSManager
+import chronos
+import chronos/apps/http/httpclient
+import
+  ../libp2p/
+    [
+      stream/connection,
+      upgrademngrs/upgrade,
+      autotls/acme/api,
+      autotls/acme/utils,
```
vladopajic: `import autotls/acme/utils`
it doesn't really make sense to import internal utils in test...

.github/workflows/daily_common.yml
```diff
@@ -96,3 +96,4 @@ jobs:
 
           export NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }}"
           nimble test
+          nimble testautotls
```
vladopajic: we need to separate `testautotls` in part that tests `api` and anything else that may come later.
could this test be `testautotls_api`?

tests/testautotls.nim
```diff
@@ -0,0 +1,69 @@
...
+      wire,
+    ]
+
+import ./helpers
+
+when defined(linux) and defined(amd64):
+  {.used.}
+
+suite "AutoTLS":
```
vladopajic: "AutoTLS" -> "AutoTLSApi"

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+
+template handleError(msg: string, body: untyped): untyped =
+  try:
+    body
+  # re-raise ACMEError
+  except ACMEError as exc:
+    raise (ref ACMEError)(exc)
+  # re-raise CancelledError
+  except CancelledError as exc:
+    raise (ref CancelledError)(exc)
```
vladopajic: `raise (ref CancelledError)(exc)` -> `raise exc` ?

vladopajic: comment above does not really add any value

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+  handleError("acmeRegister"):
+    let payload = await self.createSignedAcmeRequest(
+      self.directory.newAccount, registerRequest, key, needsJwk = true
+    )
+    let rawResponse = await HttpClientRequestRef
+    .post(
+      self.session, self.directory.newAccount, body = payload, headers = ACMEHttpHeaders
+    )
+    .get()
+    .send()
```
vladopajic: bonus:  consider having function for this these 6 lines become one `let resp = await self.post(url, payload)`

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+    let body = await rawResponse.getResponseBody()
+    let headers = rawResponse.headers
+    ACMERegisterResponse(
+      status: body["status"].getStr, kid: headers.getString("location")
+    )
+
+proc requestNewOrder(
+    self: ACMEApi, domains: seq[string], key: KeyPair, kid: Kid
+): Future[(string, ACMEChallengeResponse)] {.
```
vladopajic: it would be much better to return only `ACMEChallengeResponse` type here (we are doing similar in `requestChallenge`).

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+    .get()
+    .send()
+    let body = await rawResponse.getResponseBody()
+    let headers = rawResponse.headers
+    (headers.getString("location"), body.to(ACMEChallengeResponse))
+
+proc requestAuthorizations(
+    self: ACMEApi, authorizations: seq[string], key: KeyPair, kid: Kid
+): Future[ACMEAuthorizationsResponse] {.async: (raises: [ACMEError, CancelledError]).} =
+  handleError("requestAuthorizations"):
```
vladopajic, line 287: consider adding `doAssert` to ensure  `authorizations.len > 0`

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+    let certificateDownloadURL = orderResponse.certificate
+    let certificateExpiry = parse(orderResponse.expires, "yyyy-MM-dd'T'HH:mm:ss'Z'")
+
+    let rawResponse =
+      await HttpClientRequestRef.get(self.session, certificateDownloadURL).get().send()
+    let rawBody = await rawResponse.getBodyBytes()
+
+    ACMECertificateResponse(
+      rawCertificate: bytesToString(rawBody), certificateExpiry: certificateExpiry
+    )
```
vladopajic:
```suggestion
    let rawResponse =
      await HttpClientRequestRef.get(self.session, orderResponse.certificate).get().send()

    ACMECertificateResponse(
      rawCertificate: bytesToString(await rawResponse.getBodyBytes()), 
      certificateExpiry: parse(orderResponse.expires, "yyyy-MM-dd'T'HH:mm:ss'Z'")
    )
```

libp2p/autotls/acme/utils.nim
```diff
@@ -0,0 +1,43 @@
...
+  return encoded
+
+proc getResponseBody*(
+    response: HttpClientResponseRef
+): Future[JsonNode] {.async: (raises: [ACMEError, CancelledError]).} =
+  try:
+    let responseBody = bytesToString(await response.getBodyBytes()).parseJson()
+    return responseBody
+  except CancelledError as exc:
+    raise newException(CancelledError, "Future cancelled", exc)
```
vladopajic: `raise newException(CancelledError, "Future cancelled", exc)` 
we don't really need to add any more context to `CancelledError`. when they happen we just need to return propagating back the same exception (`raise exc`). if this operation is of point of interest we can add logs. but exception will not be handled so adding any infromation will not help (as exception is just discarded).

tests/testautotls.nim
vladopajic: if test is run locally, by default, it will definitely reach limit, i was ruining tests more then this limit

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+    let body = await rawResponse.getResponseBody()
+    body.to(ACMEAuthorizationsResponse)
+
+proc requestChallenge*(
+    self: ACMEApi, domains: seq[string], key: KeyPair, kid: Kid
+): Future[ACMEChallengeResponseWrapper] {.async: (raises: [ACMEError, CancelledError]).} =
+  let (orderURL, challengeResponse) = await self.requestNewOrder(domains, key, kid)
+
+  if challengeResponse.authorizations.len() == 0:
+    raise newException(ACMEError, "Authorizations field is empty")
```
vladopajic: is this code better suited it `self.requestNewOrder` func? since if this property is not set there then it means that something was wrong with requesting that information from api. 
because if we need to reuse `requestNewOrder` we would have to check it multiple times, on each spot where it's being reused, but if we do it in `requestNewOrder` then we dont really need to worry.

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,387 @@
...
+
+  let authorizationsResponse =
+    await self.requestAuthorizations(challengeResponse.authorizations, key, kid)
+
+  let dns01 = authorizationsResponse.challenges.filterIt(it.`type` == "dns-01")[0]
+
+  return ACMEChallengeResponseWrapper(
+    finalizeURL: challengeResponse.finalize, orderURL: orderURL, dns01: dns01
+  )
```
vladopajic:
```suggestion
  let authorizationsResponse =
    await self.requestAuthorizations(challengeResponse.authorizations, key, kid)

  return ACMEChallengeResponseWrapper(
    finalizeURL: challengeResponse.finalize, 
    orderURL: orderURL, 
    dns01: authorizationsResponse.challenges.filterIt(it.`type` == "dns-01")[0]
  )
```

^ might be opinionated suggestion but 
if `var dns01` can be initialized in one line
then better directly initialize struct field

.github/workflows/daily_common.yml
vladopajic: or `testautotls_integration`

tests/testautotls.nim
gmelodie: I'd say unlikely but how it is set right now the only way to run the autotls tests is intentionally through `nimble testautotls` (`nimble testall` or `nimble test` will not trigger autotls`

https://letsencrypt.org/docs/staging-environment/#rate-limits

tests/testautotls.nim
gmelodie: For ACMEError

.github/workflows/daily_common.yml
gmelodie: Let's cross that bridge when we get to it. Testing autotls broker will need the api. I'll leave it like this for now.

tests/testautotls.nim
```diff
@@ -0,0 +1,69 @@
...
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
+{.push raises: [].}
+
+# import std/[strformat, net] # uncomment after re-enabling AutoTLSManager
```
vladopajic: this comment might be obsolete but that time -> let's remove it.

tests/testautotls.nim
vladopajic: https://github.com/vacp2p/nim-libp2p/pull/1436#discussion_r2126614721

tests/testautotls.nim
gmelodie: This is not testing autotlsapi, this is testing the autotls part of the code. Autotlsapi doesn't even exist yet. Autotlsapi will likely be on the same level as acmeapi like so:
- `AutoTLS`
  - `AutoTLSApi`
  - `ACMEApi`

tests/testautotls.nim
vladopajic: please, consider those options.

- if you dont want to change code just export `ACMEError` (first option) 
   - add one line to api file `export ACMEError`
- or consider other options

tests/testautotls.nim
vladopajic: you have the point. these are "Auto TLS Integration" tests, could we re-frame them like that?

vladopajic: please update the files name and nimble task to match the spirit

.github/workflows/daily_common.yml
vladopajic: since these tests that are currently under `testautotls` are integration tests and since they test integration parts, we should name these appropriately now.
name was `testautotls.nim` was copy pasted from original pr, but in this context it would be more appropriate to set clean distinction with additional `integration` postfix.

gmelodie: Disagree. All AutoTLS tests are integration tests. All AutoTLS does is integrate (with acme, and with broker).

tests/testautotls.nim
gmelodie: https://github.com/vacp2p/nim-libp2p/pull/1436#discussion_r2127289706

.github/workflows/daily_common.yml
vladopajic: > AutoTLS tests are integration tests.

super! lets name them in this spirit.

tests/testautotls.nim
vladopajic: i see `ACMEError` exported, but do we need import of `autotls/acme/utils`?

vladopajic (approved):
famous! 

--- 

notes:
- composite functions (like `finalizeCertificate`, and others) might not be ideally suited for this type. we can refactor that later possibly. because it will be better to mock interface that only has first order citizen endpoints function. for example, if we mock  this type with current interface, we will have to mock `finalizeCertificate` (as it is part of the interface) which will avoid testing internals of this function (like checking response and sleeping etc..). but if this method is defined in some other type, we will test these internals with tests that cover that type. 
in general things that we are going to mock, should have the smallest possible interface. because mocks should test composite behavior of mocked interfaces.

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,395 @@
...
+type ACMERequestType = enum
+  ACMEJwkRequest
+  ACMEKidRequest
+
+type ACMERequestHeader = object
+  alg: string
+  typ: string
+  nonce: string
+  url: string
+  case kind: ACMERequestType
```
richard-ramos, line 50: Very cool, reminds me of a C union

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,395 @@
...
+type ACMERegisterResponseBody = object
+  status*: AccountStatus
+
+type ACMERegisterResponse* = object
+  kid*: Kid
+  status*: AccountStatus
+
+type ACMEChallenge = object
+  url*: string
+  `type`*: string
```
richard-ramos, line 80: If we ever move to nim-json-serialization, we can save ourserlves from having to write ``type`` with the single quotes, as we can rename this to something different and use `{.serializedFieldName("type").}`

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,395 @@
...
+# TODO: save n and e in account so we don't have to recalculate every time
+proc acmeHeader(
+    self: ACMEApi, url: string, key: KeyPair, needsJwk: bool, kid: Opt[Kid]
+): Future[ACMERequestHeader] {.async: (raises: [ACMEError, CancelledError]).} =
+  if not needsJwk and kid.isNone:
+    raise newException(ACMEError, "kid not set")
+
+  let newNonce = await self.requestNonce()
+  if needsJwk:
+    # TODO: check if scheme is RSA
```
richard-ramos: Instead of having this TODO, we should probably add an assertion or throw an exception, checking if `key.scheme == PKScheme.RSA` (like in line 204)

libp2p/autotls/acme/api.nim
```diff
@@ -0,0 +1,395 @@
...
+    key: KeyPair,
+    kid: Kid,
+    retries: int = DefaultChalCompletedRetries,
+): Future[void] {.async: (raises: [ACMEError, CancelledError]).} =
+  let completedResponse = await self.requestCompleted(chalURL, key, kid)
+  # check until acme server is done (poll validation)
+  for i in 0 .. retries:
+    let checkResponse = await self.requestCheck(completedResponse.checkURL, key, kid)
+    case checkResponse.status
+    of "pending":
```
richard-ramos: Since "valid" and "pending" are used too in line 361, consider having an enum, with an utils function like

```nim
type
  ResponseStatus = enum
    pending,
    valid,

proc toResponseStatus(s: string): Opt[ResponseStatus] =
  try:
    Opt.some(parseEnum[ResponseStatus](s))
  except ValueError:
    Opt.none(ResponseStatus)
```
And then the code would look like
```nim
let checkResponseStatus = checkResponse.status.toResponseStatus.valueOr:
  raise newException(
        ACMEError,
        "Failed challenge completion, got '" & checkResponse.status &
          "'",
      )

case checkResponseStatus:
  of ResponseStatus.pending:
       ........
  of ResponseStatus.valid:
       ........
```

richard-ramos (approved):
LGTM,  just left some nitpicks for your consideration

libp2p/autotls/acme/api.nim
richard-ramos: If it's an overkill, i guess having some constants should be fine too :)

vladopajic: parsing to enum adds safety because if the value is not recognized when parsing it will throw an error. 

if for any reason api changes (status field name changes), or code changes (in a way that's not deserializing correctly) current code will always process this case with the final `else` branch. which may not be incorrect, based on the code usecase, but it is a swallowing an error.

vladopajic: +1 for enums ^

libp2p/autotls/acme/api.nim
gmelodie: Yup totally

Issue #1429, opened 2025-05-29
flaky issue (GossipSub): adaptive gossip dissemination, dLazy and gossipFactor to 0


vladopajic:
https://github.com/vacp2p/nim-libp2p/actions/runs/15326321345/job/43121668636?pr=1418

```
 ==========================================================
  /Users/runner/work/nim-libp2p/nim-libp2p/tests/testall 'GossipSub Gossip Protocol::adaptive gossip dissemination, dLazy and gossipFactor to 0'
----------------------------------------------------------
    /Users/runner/work/nim-libp2p/nim-libp2p/tests/pubsub/testgossipsubgossip.nim(196, 62): Check failed: (await nodes[0].publish(topic, "Hello!".toBytes())) == 3
    (await nodes[0].publish(topic, "Hello!".toBytes())) was 2

  [FAILED ] (  0.83s) adaptive gossip dissemination, dLazy and gossipFactor to 0
```

Issue #1426, opened 2025-05-29
Unify checkUntilTimeout and checkUntilCustomTimeout


gmelodie:
`checkUntilCustomTimeout` allows for the user to specify the timeout, while `checkUntilTimeout` uses hardcoded times. These two should be unified as `checkUntilTimeout`, which should allow specifying the times.

https://github.com/vacp2p/nim-libp2p/pull/1416#discussion_r2112946066

gmelodie:
https://github.com/vacp2p/nim-libp2p/issues/1425

Issue #1425, opened 2025-05-29
test: should we drop entierlly usage of `waitForPeersInTable` in favor of `checkUntilCustomTimeout`


vladopajic:
discussion from https://github.com/vacp2p/nim-libp2p/pull/1416#discussion_r2113604256

Issue #1424, opened 2025-05-29
test: add tests for `checkUntilCustomTimeout`


vladopajic:
from review https://github.com/vacp2p/nim-libp2p/pull/1416#discussion_r2113613643:

`checkUntilCustomTimeout`  is missing tests when: 

- condition is never true
- condition is true after few intervals

PR #1437, opened 2025-06-02
test: checkUntilTimeout refactor

rlve:
Some refactors around recent issues related to `checkUntilTimeout` helper and flaky `GossipSub` tests.

- https://github.com/vacp2p/nim-libp2p/issues/1424 - added more unit tests for `checkUntilCustomTimeout` and `checkUntilTimeout`
- https://github.com/vacp2p/nim-libp2p/issues/1425 - removed `waitForPeersInTable` and used `checkUntilTimeout` instead
- https://github.com/vacp2p/nim-libp2p/issues/1426 - here again I don't agree on unifying `checkUntilTimeout` with `checkUntilCustomTimeout` - I believe current implementation is elegant, `checkUntilTimeout` with default timeout and interval and `checkUntilCustomTimeout` if more flexibility is needed. `checkUntilTimeout` looks way cleaner in the code, when parameters don't need to be passed. What I did was to change the default values in  `checkUntilTimeout` to `timeout=2s` and `interval=50ms`, which work for majority of the tests, and only use  `checkUntilCustomTimeout` where it's really needed
- https://github.com/vacp2p/nim-libp2p/issues/1429 - fix flaky tests by using `tryPublish` when publishing messages

### DIFF
--- a/tests/helpers.nim
+++ b/tests/helpers.nim
@@ -94,7 +94,7 @@ proc new*(T: typedesc[TestBufferStream], writeHandler: WriteHandler): T =
   testBufferStream.initStream()
   testBufferStream
 
-macro checkUntilCustomTimeout*(
+macro checkUntilTimeoutCustom*(
     timeout: Duration, sleepInterval: Duration, code: untyped
 ): untyped =
   ## Periodically checks a given condition until it is true or a timeout occurs.
@@ -105,17 +105,17 @@ macro checkUntilCustomTimeout*(
   ## Examples:
   ##   ```nim
   ##   # Example 1:
-  ##   asyncTest "checkUntilCustomTimeout should pass if the condition is true":
+  ##   asyncTest "checkUntilTimeoutCustom should pass if the condition is true":
   ##     let a = 2
   ##     let b = 2
-  ##     checkUntilCustomTimeout(2.seconds):
+  ##     checkUntilTimeoutCustom(2.seconds):
   ##       a == b
   ##
   ##   # Example 2: Multiple conditions
-  ##   asyncTest "checkUntilCustomTimeout should pass if the conditions are true":
+  ##   asyncTest "checkUntilTimeoutCustom should pass if the conditions are true":
   ##     let a = 2
   ##     let b = 2
-  ##     checkUntilCustomTimeout(5.seconds)::
+  ##     checkUntilTimeoutCustom(5.seconds)::
   ##       a == b
   ##       a == 2
   ##       b == 1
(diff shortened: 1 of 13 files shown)


tests/testhelpers.nim
```diff
@@ -39,3 +52,32 @@ suite "Helpers":
...
+  teardown:
+    require testStatusIMPL == TestStatus.Failed
+    testStatusIMPL = TestStatus.OK
+    exitProcs.setProgramResult(QuitSuccess)
+
+  asyncTest "checkUntilTimeout should timeout if condition is never true":
+    let a = 1
+    let b = 2
+    checkUntilTimeout:
+      a == b
```
vladopajic: could we simplify this to just be:
```
 checkUntilTimeout:
      false
```      
or 
```
 checkUntilTimeout:
      1 == 2
```      

--- 
i think first test cases are just there to test if check can support multiple conditions, here we dont need that.

tests/testhelpers.nim
```diff
@@ -39,3 +52,32 @@ suite "Helpers":
...
+
+  asyncTest "checkUntilCustomTimeout should timeout if condition is never true":
+    let a = 1
+    let b = 2
+    checkUntilCustomTimeout(100.milliseconds, 10.milliseconds):
+      a == b
```
vladopajic: if possible we could simplify this (as above)

tests/testhelpers.nim
```diff
@@ -10,8 +10,10 @@
 # those terms.
 
 import ./helpers
+import unittest2
+from std/exitprocs import nil
```
vladopajic, line 14: `from std/exitprocs import nil` why is this?

tests/helpers.nim
```diff
@@ -175,7 +175,7 @@ macro checkUntilTimeout*(code: untyped): untyped =
   ##       b == 1
   ##   ```
   result = quote:
-    checkUntilCustomTimeout(10.seconds, 100.milliseconds, `code`)
+    checkUntilCustomTimeout(2.seconds, 50.milliseconds, `code`)
```
vladopajic: can we rename `checkUntilCustomTimeout` to `checkUntilTimeoutCustom`? 
usually it makes sense for same functions to have the same prefix. and suffix is added to distinguish them.

vladopajic: it should be simple change and can easily go with this PR (since is doing most work around this)

tests/testhelpers.nim
rlve: `exitprocs` is needed for: https://github.com/vacp2p/nim-libp2p/blob/b64f20385b987d8d00a4aa3655030e0cf0f06bdb/tests/testhelpers.nim#L71

I copy-pasted the assertion and import from https://github.com/status-im/nim-unittest2/blob/master/tests/tunittest.nim 😅

vladopajic: aha, got it!

richard-ramos (approved):

PR #1438, opened 2025-06-02
fix(gossipsub): fail `connectedFut`

richard-ramos:


richard-ramos:
Closing the PR as it is not necessary because `connectedFut` will get completed in the finally at https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/pubsubpeer.nim#L277

PR #1442, opened 2025-06-04
test(gossipsub): paremeters

rlve:


tests/pubsub/utils.nim
```diff
@@ -129,7 +129,6 @@ proc setupGossipSubWithPeers*(
 
 proc teardownGossipSub*(gossipSub: TestGossipSub, conns: seq[Connection]) {.async.} =
   await allFuturesThrowing(conns.mapIt(it.close()))
-  await gossipSub.switch.stop()
```
rlve, line 132: Seems like it wasn't needed. With it we can observe logs:

```
WRN 2025-06-05 12:34:58.136+01:00 service is already stopped                 topics="libp2p switch" tid=15516792
WRN 2025-06-05 12:34:58.136+01:00 TCP transport already stopped              topics="libp2p tcptransport" tid=15516792
```

tests/pubsub/testgossipsubcontrolmessages.nim
```diff
@@ -511,3 +512,34 @@ suite "GossipSub Control Messages":
...
+        ControlIWant(messageIDs: generateMessageIds(600)),
+        ControlIWant(messageIDs: generateMessageIds(600)),
+      ]
+
+    # When node handles iDontWants
+    gossipSub.handleIDontWant(peer, iDontWants)
+
+    # Then it saves max 1001 messages in the history and the rest is dropped
+    check:
+      peer.iDontWants[0].len == 1001
```
rlve, line 543: Compared to the magic number, as the condition in code is done in the following way:

https://github.com/vacp2p/nim-libp2p/blob/3a7745f920d11fb29305070cd9991874529dfc28/libp2p/protocols/pubsub/gossipsub/behavior.nim#L308-L309

[Spec for context](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md#parameters)

vladopajic (approved):

tests/pubsub/testgossipsubcontrolmessages.nim
vladopajic: #1446 

i'll make this pr tomorrow

tests/pubsub/testgossipsubcontrolmessages.nim
```diff
@@ -511,3 +512,34 @@ suite "GossipSub Control Messages":
...
+    defer:
+      await teardownGossipSub(gossipSub, conns)
+
+    let peer = peers[0]
+
+    # And sequence of iDontWants with more messages than max number (1200)
+    proc generateMessageIds(count: int): seq[MessageId] =
+      result = newSeq[MessageId](count)
+      for i in 0 ..< count:
+        result[i] = ("msg_id_" & $i & $Moment.now()).toBytes()
```
richard-ramos: More of a nitpick, but it's recommended on the style guide to not use a result but instead do an explicit return: https://status-im.github.io/nim-style-guide/language.result.html , so probably this whole function can be something like this:

```
proc generateMessageIds(count: int): seq[MessageId] =
  return (0..<count).mapIt(("msg_id_" & $it & $Moment.now()).toBytes())
  ```

richard-ramos (approved):

tests/pubsub/testgossipsubcontrolmessages.nim
rlve: Good to know 👍

PR #1445, opened 2025-06-05
feat(peeridauth): add peeridauth

gmelodie:
Add `peeridauth.nim`: implements https://github.com/libp2p/specs/blob/master/http/peer-id-auth.md

### DIFF
--- a/libp2p/peeridauth.nim
+++ b/libp2p/peeridauth.nim
@@ -1,0 +1,262 @@
+import base64, json, strutils
+import chronos/apps/http/httpclient, results, chronicles, bio
+import ./peerinfo, ./crypto/crypto
+
+logScope:
+  topics = "libp2p peerid auth"
+
+const NimLibp2pUserAgent = "nim-libp2p"
+const PeerIDAuthPrefix = "libp2p-PeerID"
+const ChallengeCharset =
+  "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
+const ChallengeDefaultLen = 48
+
+type PeerIDAuthError* = object of LPError
+
+type BearerToken* = string
+type PeerIDAuthChallenge = string
+type PeerIDAuthOpaque = string
+type PeerIDAuthSignature = string
+
+type PeerIDAuthApi* = object
+  bearer*: Opt[BearerToken]
+  session: HttpSessionRef
+  rng: ref HmacDrbgContext
+
+type PeerIDAuthAuthenticationResponse* = object
+  challengeClient: PeerIDAuthChallenge
+  opaque: PeerIDAuthOpaque
+  serverPubkey: PublicKey
+
+type PeerIDAuthAuthorizationResponse* = object
+  sig: PeerIDAuthSignature
+  bearer: BearerToken
+
+type SigParam = object
+  k: string
+  v: seq[byte]
+
+proc new*(T: typedesc[PeerIDAuthApi], rng: ref HmacDrbgContext): PeerIDAuthApi =
+  PeerIDAuthApi(session: HttpSessionRef.new(), rng: rng)
+
+proc randomChallenge(
+    rng: ref HmacDrbgContext, challengeLen: int = ChallengeDefaultLen
+): PeerIDAuthChallenge {.raises: [PeerIDAuthError].} =
+  var rng = rng[]
+  var challenge = ""
+  try:
+    for _ in 0 ..< challengeLen:
+      challenge.add(rng.sampleChar(ChallengeCharset))
+  except ValueError as exc:
+    raise newException(PeerIDAuthError, "Failed to generate challenge", exc)
+  PeerIDAuthChallenge(challenge)
+
+proc extractField(data, key: string): string {.raises: [PeerIDAuthError].} =
+  # Helper to extract quoted value from key
+  for segment in data.split(","):
+    if key in segment:
+      return segment.split("=", 1)[1].strip(chars = {' ', '"'})
+  raise newException(PeerIDAuthError, "Failed to find " & key & " in " & data)
+
+proc encodeVarint(n: int): seq[byte] =
+  var varInt: seq[byte] = @[]
+  var x = uint64(n)
+  while x != 0:
+    var byteVal = byte(x and 0x7F)
+    x = x shr 7
+    if x != 0:
+      byteVal = byteVal or 0x80
+    varInt.add(byteVal)
+  return varInt
+
+proc genDataToSign(parts: seq[SigParam], prefix: string = PeerIDAuthPrefix): seq[byte] =
+  var buf: seq[byte] = prefix.toByteSeq()
+  for p in parts:
+    buf.add encodeVarint(p.k.len + p.v.len + 1)
+    buf.add (p.k & "=").toByteSeq()
+    buf.add p.v
+  return buf
+
+proc getSigParams(
+    clientSender: bool, hostname: string, challenge: string, publicKey: PublicKey
+): seq[SigParam] =
+  if clientSender:
+    @[
+      SigParam(k: "challenge-client", v: challenge.toByteSeq()),
+      SigParam(k: "hostname", v: hostname.toByteSeq()),
+      SigParam(k: "server-public-key", v: publicKey.getBytes().get()),
+    ]
+  else:
+    @[
+      SigParam(k: "challenge-server", v: challenge.toByteSeq()),
+      SigParam(k: "client-public-key", v: publicKey.getBytes().get()),
+      SigParam(k: "hostname", v: hostname.toByteSeq()),
+    ]
+
+proc sign(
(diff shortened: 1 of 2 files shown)


tests/testpeeridauth.nim
```diff
@@ -0,0 +1,57 @@
...
+import json, uri
+import chronos
+import chronos/apps/http/httpclient
+import ../libp2p/[stream/connection, upgrademngrs/upgrade, peeridauth, wire]
+
+import ./helpers
+
+const
+  AuthPeerURL = "https://registration.libp2p.direct/v1/_acme-challenge"
+  HttpPeerAuthFailed = 401
```
richard-ramos: Shouldnt these variables be part of a file withing libp2p/ folder?

libp2p/peeridauth.nim
```diff
@@ -0,0 +1,304 @@
...
+      byteVal = byteVal or 0x80
+    varInt.add(byteVal)
+  return varInt
+
+proc genDataToSign(parts: seq[SigParam], prefix: string = PeerIDAuthPrefix): seq[byte] =
+  var buf: seq[byte] = prefix.toByteSeq()
+  for p in parts:
+    buf.add encodeVarint(p.k.len + p.v.len + 1)
+    buf.add (p.k & "=").toByteSeq()
+    buf.add p.v
```
richard-ramos: Couldnt this and encodeVarint be replaced by libp2p varint.nim?
with something like 
```nim
proc genDataToSign(parts: seq[SigParam], prefix: string = PeerIDAuthPrefix): seq[byte] =
  var buf: seq[byte] = prefix.toByteSeq()
  for p in parts:
    buf.add LP.encodeVarint(p.k.len + p.v.len + 1)
    buf.add (p.k & "=").toByteSeq()
    buf.add p.v
```

libp2p/peeridauth.nim
```diff
@@ -0,0 +1,304 @@
...
+): Future[HttpClientResponseRef] {.async: (raises: [HttpError, CancelledError]).} =
+  await HttpClientRequestRef
+  .post(
+    self.session,
+    uri,
+    body = payload,
+    headers = [
+      ("Content-Type", "application/json"),
+      ("User-Agent", NimLibp2pUserAgent),
+      ("authorization", authHeader),
```
richard-ramos:
```suggestion
      ("Authorization", authHeader),
```

libp2p/peeridauth.nim
```diff
@@ -0,0 +1,304 @@
...
+        PeerIDAuthError, "Failed to send Authorization for PeerID Auth", exc
+      )
+
+  let authenticationInfo = rawResponse.headers.getString("authentication-info")
+
+  let bearerExpires =
+    try:
+      Opt.some(parse3339DateTime(extractField(authenticationInfo, "expires")))
+    except ValueError, PeerIDAuthError:
+      Opt.none(DateTime)
```
richard-ramos: Could be interesting to add a trace log here to understand why the exception happened.

richard-ramos (approved):

tests/testpeeridauth.nim
```diff
@@ -0,0 +1,57 @@
+{.used.}
```
vladopajic: tests that do interact with external api, should be part for integration tests group.

libp2p/peeridauth.nim
```diff
@@ -0,0 +1,304 @@
...
+  let rawResponse =
+    try:
+      await self.post($uri, $payload, authHeader)
+    except HttpError as exc:
+      raise newException(
+        PeerIDAuthError, "Failed to send request with bearer token for PeerID Auth", exc
+      )
+  return (bearer, rawResponse)
+
+proc send*(
```
vladopajic, line 297: `send` is vary vague name. can we use name that describe what method will do (like authenticate,... )?

tests/testpeeridauth.nim
```diff
@@ -0,0 +1,57 @@
...
+      }
+
+    let (bearer, responseWithoutBearer) =
+      await api.send(parseUri(AuthPeerURL), peerInfo, payload)
+    check responseWithoutBearer.status != HttpPeerAuthFailed
+    doAssert bearer.token.len > 0
+
+    let (_, responseWithBearer) =
+      await api.send(parseUri(AuthPeerURL), peerInfo, payload, bearer)
+    check responseWithBearer.status != HttpPeerAuthFailed
```
vladopajic: why is user forced to use `responseWithBearer.status` to check if send, unfortunately named, is successful? this is causing user to process errors in two ways: 1st) one that he must - exceptions 2nd) checking `responseWithBearer.status`
can send throw exception in case when action is successful?

libp2p/peeridauth.nim
```diff
@@ -0,0 +1,304 @@
...
+    self: PeerIDAuthApi,
+    uri: Uri,
+    peerInfo: PeerInfo,
+    payload: auto,
+    bearer: BearerToken = BearerToken(),
+): Future[(BearerToken, HttpClientResponseRef)] {.
```
vladopajic, line 303: why does it return `HttpClientResponseRef`? exposed function should not return types of internal mechanism (http). for example how would this function be mocked, why would mock be forced to mock `HttpClientResponseRef` if we only care about simple outcome (just bearer token)?`

libp2p/peeridauth.nim
```diff
@@ -0,0 +1,304 @@
...
+proc sendWithBearer(
+    self: PeerIDAuthApi,
+    uri: Uri,
+    peerInfo: PeerInfo,
+    payload: auto,
+    bearer: BearerToken,
+): Future[(BearerToken, HttpClientResponseRef)] {.
+    async: (raises: [PeerIDAuthError, CancelledError])
+.} =
+  if bearer.expires.isSome and DateTime(bearer.expires.get) <= now():
```
vladopajic, line 285: what if not bearer.expires.isSome??

tests/testpeeridauth.nim
```diff
@@ -0,0 +1,57 @@
...
+
+import json, uri
+import chronos
+import chronos/apps/http/httpclient
+import ../libp2p/[stream/connection, upgrademngrs/upgrade, peeridauth, wire]
+
+import ./helpers
+
+const
+  AuthPeerURL = "https://registration.libp2p.direct/v1/_acme-challenge"
```
vladopajic: `AuthPeerURL` why is this not default value used in `send`? would it be required to redefined this variable in other places: 1st) in test 2nd) component that uses `PeerIDAuthApi` (could be many

tests/testpeeridauth.nim
gmelodie: `AuthPeerURL` is the URL of a peer to which we can test, could be any one in particular. Remember: this PR is for peeridauth, it should work with all peers as long as they're using peerid auth. Same for `HttpPeerAuthFailed`: we now it failed for this specific peer, this is not some http standard code for unsuccessful peerid auths

libp2p/peeridauth.nim
gmelodie: The exception happens when there's no expiry field on response. That's fine. Maybe we could add a debug log message but I don't think that's required (too verbose for something that happens regularly)

libp2p/peeridauth.nim
gmelodie: I had that exact name before! (`authenticate` i mean). The problem with that, I think, is that it seems to suggest that it _only_ authenticates, when in reality it _also_ authenticates (meaning it authenticates and sends data). What do you think of having two functions then: `authenticateAndSend` and `authenticatedSend` (or maybe just `send` for when user already has `bearer`)?

tests/testpeeridauth.nim
gmelodie: `peeridauth` is an authentication scheme for HTTP requests, and because of that the meaning of a particular status code largely depends on the application behind it. It is up to the user to figure out what the error means or if that's an error at all! (in this case the user is the test)

here I used the autotls broker to test peeridauth because i know more or less how that works, but any other peer could be used too

libp2p/peeridauth.nim
gmelodie: \*vilain laughter\* you activated my trap card sir! This is an authentication scheme for HTTP. It's job is to do authentication and send authenticated HTTP requests, thus returning HTTP responses.

> we only care about simple outcome (just bearer token)?

We most certainly do *not* only care about bearer token! On the particular set of interop tests I added here you're right, since the response from the server contains no payload, but it very well could! Again, it's an HTTP authentication scheme.

libp2p/peeridauth.nim
gmelodie: it doesn't even evaluate right side of an `and` statement

tests/testpeeridauth.nim
gmelodie: > AuthPeerURL why is this not default value used in send? 

Because one can use `peeridauth` with several different applications, this is not something exclusive to autotls. I'm just using the autotls broker to test it here.

PR #1447, opened 2025-06-06
add const

vladopajic:


libp2p/protocols/pubsub/gossipsub/types.nim
```diff
@@ -50,6 +50,8 @@ const
   # rust sigp: https://github.com/sigp/rust-libp2p/blob/f53d02bc873fef2bf52cd31e3d5ce366a41d8a8c/protocols/gossipsub/src/config.rs#L572
   # go: https://github.com/libp2p/go-libp2p-pubsub/blob/08c17398fb11b2ab06ca141dddc8ec97272eb772/gossipsub.go#L155
   IHaveMaxLength* = 5000
+  IDontWantMaxCount* = 1000 # maximum number of IDontWant messages in history
+  MaxOpportunisticGraftPeers* = 2
```
rlve: Should both of the parameters be configurable and also added to `GossipSubParams`?

libp2p/protocols/pubsub/gossipsub/types.nim
```diff
@@ -50,6 +50,8 @@ const
   # rust sigp: https://github.com/sigp/rust-libp2p/blob/f53d02bc873fef2bf52cd31e3d5ce366a41d8a8c/protocols/gossipsub/src/config.rs#L572
   # go: https://github.com/libp2p/go-libp2p-pubsub/blob/08c17398fb11b2ab06ca141dddc8ec97272eb772/gossipsub.go#L155
   IHaveMaxLength* = 5000
+  IDontWantMaxCount* = 1000 # maximum number of IDontWant messages in history
```
rlve:
```suggestion
  IDontWantMaxCount* = 1000 # maximum number of IDontWant messages in one slot of the history
```

libp2p/protocols/pubsub/gossipsub/types.nim
vladopajic: currently there was no usecase for adjusting these values?

if anyone will need different values then they can add it to `GossipSubParams` and using these consts as default values.

richard-ramos (approved):

PR #1448, opened 2025-06-06
feat: GossipSub v1.4

ufarooqstatus:
Added GossipSub v1.4 implementation. Specs proposal [here](https://github.com/libp2p/specs/pull/654).

Still need to complete the safety strategy (in next PR)

Requires the compile flag: `-d:libp2p_gossipsub_1_4`, although running `nimble testpubsub` does add the flag automatically.

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -458,10 +458,8 @@ when defined(libp2p_gossipsub_1_4):
     if msg.data.len < preambleMessageSizeThreshold:
       return
 
-    let preamblePeers = toSendPeers.filterIt(it.codec == GossipSubCodec_14)
-
     g.broadcast(
-      preamblePeers,
+      toSendPeers.filterIt(it.codec == GossipSubCodec_14),
       RPCMsg(
         control: some(
           ControlMessage(
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -376,7 +376,6 @@ when defined(libp2p_gossipsub_1_4):
     else:
       vals[mid]
 
-when defined(libp2p_gossipsub_1_4):
   proc handlePreamble*(
       g: GossipSub, peer: PubSubPeer, preambles: seq[ControlPreamble]
   ) =
@@ -437,7 +436,6 @@ when defined(libp2p_gossipsub_1_4):
             isHighPriority = true,
           )
 
-when defined(libp2p_gossipsub_1_4):
   proc handleIMReceiving*(
       g: GossipSub, peer: PubSubPeer, imreceivings: seq[ControlIMReceiving]
   ) =
--- a/tests/pubsub/integration/testgossipsubcontrolmessages.nim
+++ b/tests/pubsub/integration/testgossipsubcontrolmessages.nim
@@ -390,6 +390,7 @@ suite "GossipSub Integration - Control Messages":
     # Then IDONTWANT is sent to B on publish
     checkUntilTimeout:
       nodes[1].mesh.getOrDefault(topic).anyIt(it.iDontWants.anyIt(it.len == 1))
+
   when defined(libp2p_gossipsub_1_4):
     asyncTest "emit IMReceiving while handling preamble control msg":
       # Given GossipSub node with 1 peer


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -880,6 +890,8 @@ method publish*(
   if g.parameters.sendIDontWantOnPublish and isLargeMessage(msg, msgId):
     g.sendIDontWant(msg, msgId, peers)
 
+  g.sendPreamble(msg, msgId, peers)
```
ufarooqstatus: This is only useful if multiple sources introduce the same message to the network. If no one else is sending, we can skip sending the preamble. So, we can put it under sendIDontWantOnPublish check above

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -860,3 +868,9 @@ proc heartbeat*(g: GossipSub) {.async: (raises: [CancelledError]).} =
     for trigger in g.heartbeatEvents:
       trace "firing heartbeat event", instance = cast[int](g)
       trigger.fire()
+
+proc bandwithHeartbeat*(g: GossipSub) {.async: (raises: [CancelledError]).} =
+  heartbeat "Gossipsub Peer Bandwidth Tracking", 1.seconds:
+    for (topic, peers) in g.gossipsub.pairs:
+      for peer in peers.items:
+        peer.bandwidthTracking.update()
```
ufarooqstatus: If there's no traffic for a few seconds, bandwidth estimate moves to 0. 

Actually, we are interested in estimating how fast a peer will transfer to us. Maybe we can use $\frac{messageSize}{downloadTime - preambleTime}$, and use EMA to maintain it
We won't need bandwithHeartbeat in that case

libp2p/protocols/pubsub/bandwidth.nim
```diff
@@ -0,0 +1,37 @@
...
+
+type
+  ExponentialMovingAverage* = object
+    alpha: float
+    value: float
+    accum: Atomic[int]
+
+  BandwidthTracking* = ref object
+    upload*: ExponentialMovingAverage
+    download*: ExponentialMovingAverage
```
ufarooqstatus: We just need to know how fast a remote peer can transfer to us, so upload can suffice

libp2p/protocols/pubsub/gossipsub/behavior.nim
richard-ramos: Maybe we can just skip updating the EMA if the sample is 0, altho that will leave you with a bandwidth estimation that is even less precise if the node is not chatty.

But, if going with this formula you suggest, what does the `downloadTime` and `preambleTime` refer to?

ufarooqstatus: yes, sounds good. we just need a rough estimate.
[Preamble time](https://github.com/vacp2p/nim-libp2p/pull/1448/files#diff-4a8279b47cb86bc36bbdcfb9c0f999a7de0e62fb4ced0c51e1a8fd3aa6cf5a50R387) is whe we received preamble. download time is when we received complete message from the same peer

ufarooqstatus: We just want to be able to configure that we only issue IMRECEIVING on receiving preamble from top X % of peers

libp2p/protocols/pubsub/gossipsub/receivestore.nim
```diff
@@ -0,0 +1,61 @@
...
+  a.expiresAt < b.expiresAt
+
+proc initPreambleStore*(): PreambleStore =
+  result.byId = initTable[MessageId, PreambleInfo]()
+  result.heap = initHeapQueue[PreambleInfo]()
+
+proc insert*(ps: var PreambleStore, msgId: MessageId, info: PreambleInfo) =
+  try:
+    if ps.byId.hasKey(msgId):
+      ps.byId[msgId].deleted = true
```
ufarooqstatus: This replaces the preamble info we are watching (for expiry with the new one). Potentially mesh may keep sending us preambles and we will keep updating this. 

Although this never happens because we check for existing entries [here](https://github.com/vacp2p/nim-libp2p/pull/1448/commits/eff192499aa4f1f9661ccad2a6c453d4311d0652#diff-4a8279b47cb86bc36bbdcfb9c0f999a7de0e62fb4ced0c51e1a8fd3aa6cf5a50R369-R370)

richard-ramos: Yes, this is to keep the functionaliyu of `PreambleStore` to behave the same as a Table, (altho it internally also has a minheap). Except for the `popExpired` function, you'll notice that the API and behavior matches that of a Table.

libp2p/protocols/pubsub/gossipsub/receivestore.nim
```diff
@@ -0,0 +1,61 @@
...
+    return false
+  except KeyError:
+    assert false, "checked with hasKey"
+
+proc len*(ps: var PreambleStore): int =
+  return ps.byId.len
+
+proc popExpired*(ps: var PreambleStore, now: Moment): Option[PreambleInfo] =
+  while ps.heap.len > 0:
+    if not ps.heap[0].deleted:
```
ufarooqstatus:
```suggestion
    if ps.heap[0].deleted:
```

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -884,12 +886,47 @@ proc preambleExpirationHeartbeat*(g: GossipSub) {.async: (raises: [CancelledErro
...
+        @[peers[0]],
+        RPCMsg(
+          control: some(
+            ControlMessage(
+              iwant: @[ControlIWant(messageIDs: @[expiredOngoingReceive.messageId])]
+            )
+          )
+        ),
+        isHighPriority = true,
+      )
```
ufarooqstatus: we still dont know if peers[0] has the required message. what if we store 2-3 mesh members [from here](https://github.com/vacp2p/nim-libp2p/blob/f6ce1dfcfbadaeb1be872731c5a55311ee6d84ee/libp2p/protocols/pubsub/gossipsub/behavior.nim#L317) and 2-3 non-memsh members [from here](https://github.com/vacp2p/nim-libp2p/blob/f6ce1dfcfbadaeb1be872731c5a55311ee6d84ee/libp2p/protocols/pubsub/gossipsub/behavior.nim#L302) to preambleInfo (maybe as meshList and IHaveList). That can be more efficient, as we won't need to fetch mesh members, and we also won't need topicID in preambleInfo

gmelodie (commented):
I think this PR should be split into smaller ones as it was with AutoTLS

richard-ramos:
@gmelodie PR is now small enough for review after extracting part of it in #1513 and #1515

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -371,13 +388,34 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
...
+        # So receiver will send IMReciving only for preambles received from mesh members
+        preambles.add(
+          ControlPreamble(
+            topicID: smsg.topic,
+            messageID: msgIDs[i],
+            messageLength: smsg.data.len.uint32,
+          )
+        )
+
+    when defined(libp2p_gossipsub_1_4):
```
gmelodie, line 412:
```suggestion
```

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -328,14 +350,103 @@ proc handleIWant*(
...
           continue
         let msg = g.mcache.get(mid).valueOr:
           libp2p_gossipsub_received_iwants.inc(1, labelValues = ["unknown"])
           continue
         libp2p_gossipsub_received_iwants.inc(1, labelValues = ["correct"])
-        messages.add(msg)
-  return messages
+        result.messages.add(msg)
+        result.ids.add(mid)
+  return result
```
gmelodie: Do we need `return result` or should `return` be enough? Also we can probably remove the second `return` statement entirely

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -328,14 +350,103 @@ proc handleIWant*(
...
+
+    let vals = p.toSeq().mapIt(it.bandwidthTracking.download.value()).sorted()
+    echo vals
+    let mid = vals.len div 2
+    if vals.len mod 2 == 0:
+      (vals[mid - 1] + vals[mid]) / 2
+    else:
+      vals[mid]
+
+when defined(libp2p_gossipsub_1_4):
```
gmelodie:
```suggestion
```

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -328,14 +350,103 @@ proc handleIWant*(
...
+
+    for preamble in preambles:
+      dec peer.preambleBudget
+      if peer.preambleBudget <= 0:
+        return
+      if g.hasSeen(g.salt(preamble.messageID)):
+        continue
+      if peer.heIsSendings.hasKey(preamble.messageID):
+        continue
+      if g.ongoingReceives.hasKey(preamble.messageID):
```
gmelodie: Maybe swap `if`s for `elif`s?

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -328,14 +350,103 @@ proc handleIWant*(
...
+            ),
+            isHighPriority = true,
+          )
+
+when defined(libp2p_gossipsub_1_4):
+  proc handleIMReceiving*(
+      g: GossipSub, peer: PubSubPeer, imreceivings: seq[ControlIMReceiving]
+  ) =
+    for imreceiving in imreceivings:
+      if peer.heIsReceivings.len > 50:
```
gmelodie: Set 50 to a `const`

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -777,3 +891,72 @@ proc heartbeat*(g: GossipSub) {.async: (raises: [CancelledError]).} =
     for trigger in g.heartbeatEvents:
       trace "firing heartbeat event", instance = cast[int](g)
       trigger.fire()
+
+when defined(libp2p_gossipsub_1_4):
+  proc preambleExpirationHeartbeat*(
```
gmelodie, line 897: Maybe we could extract the `heartbeat` call to when this function is used and have the function just be a single run?

gmelodie (approved):

tests/pubsub/integration/testgossipsubcontrolmessages.nim
```diff
@@ -386,3 +390,50 @@ suite "GossipSub Integration - Control Messages":
     # Then IDONTWANT is sent to B on publish
     checkUntilTimeout:
       nodes[1].mesh.getOrDefault(topic).anyIt(it.iDontWants.anyIt(it.len == 1))
+  when defined(libp2p_gossipsub_1_4):
```
vladopajic: add some space here to separate tests visually

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -411,6 +449,36 @@ proc sendIDontWant(
...
+      g: GossipSub, msg: Message, msgId: MessageId, toSendPeers: var HashSet[PubSubPeer]
+  ) =
+    if msg.data.len < preambleMessageSizeThreshold:
+      return
+
+    let preamblePeers = toSendPeers.filterIt(it.codec == GossipSubCodec_14)
+
+    g.broadcast(
+      preamblePeers,
+      RPCMsg(
```
vladopajic:
```suggestion
    g.broadcast(
      toSendPeers.filterIt(it.codec == GossipSubCodec_14),
      RPCMsg(
```

no need to assign to variable, we can assign to param in place

libp2p/protocols/pubsub/gossipsub.nim
richard-ramos: Why remove this?

libp2p/protocols/pubsub/gossipsub/behavior.nim
richard-ramos: Why remove this?

libp2p/protocols/pubsub/gossipsub/behavior.nim
richard-ramos: I think ideally we shouldnt use the result variable as recommended by the style guide. I'll change that

libp2p/protocols/pubsub/gossipsub.nim
gmelodie: There's another identical `when defined` right above this in line 401

libp2p/protocols/pubsub/gossipsub/behavior.nim
gmelodie: There's another when defined right above this in line 363

PR #1452, opened 2025-06-09
feat(perf): add stats

vladopajic:
- adding stats to Perf client, needed by perf benchmarking implementation (https://github.com/libp2p/test-plans/pull/658)
- adding tests for Perf protocol
- discovered bug #1462

### DIFF
--- a/libp2p/protocols/perf/client.nim
+++ b/libp2p/protocols/perf/client.nim
@@ -16,25 +16,37 @@ import ./core, ../../stream/connection
 logScope:
   topics = "libp2p perf"
 
-type PerfClient* = ref object of RootObj
+type Stats* = object
+  isFinal*: bool
+  uploadBytes*: uint
+  downloadBytes*: uint
+  duration*: Duration
+
+type PerfClient* = ref object
+  stats*: Stats
+
+proc new*(T: typedesc[PerfClient]): T =
+  return T()
 
 proc perf*(
-    _: typedesc[PerfClient],
-    conn: Connection,
-    sizeToWrite: uint64 = 0,
-    sizeToRead: uint64 = 0,
+    p: PerfClient, conn: Connection, sizeToWrite: uint64 = 0, sizeToRead: uint64 = 0
 ): Future[Duration] {.public, async: (raises: [CancelledError, LPStreamError]).} =
+  trace "starting performance benchmark", conn, sizeToWrite, sizeToRead
+
   var
     size = sizeToWrite
     buf: array[PerfSize, byte]
+
+  p.stats = Stats()
   let start = Moment.now()
-  trace "starting performance benchmark", conn, sizeToWrite, sizeToRead
 
   await conn.write(toSeq(toBytesBE(sizeToRead)))
   while size > 0:
     let toWrite = min(size, PerfSize)
     await conn.write(buf[0 ..< toWrite])
     size -= toWrite
+    p.stats.duration = Moment.now() - start
+    p.stats.uploadBytes += toWrite
 
   await conn.close()
 
@@ -44,6 +56,10 @@ proc perf*(
     let toRead = min(size, PerfSize)
     await conn.readExactly(addr buf[0], toRead.int)
     size = size - toRead
+    p.stats.duration = Moment.now() - start
+    p.stats.downloadBytes += toRead
+
+  p.stats.isFinal = true
 
   let duration = Moment.now() - start
   trace "finishing performance benchmark", duration
--- a/libp2p/protocols/perf/client.nim
+++ b/libp2p/protocols/perf/client.nim
@@ -29,7 +29,7 @@ proc new*(T: typedesc[PerfClient]): T =
   return T()
 
 proc currentStats*(p: PerfClient): Stats =
-  p.stats
+  return p.stats
 
 proc perf*(
     p: PerfClient, conn: Connection, sizeToWrite: uint64 = 0, sizeToRead: uint64 = 0
--- a/libp2p/protocols/perf/client.nim
+++ b/libp2p/protocols/perf/client.nim
@@ -75,6 +75,6 @@ proc perf*(
 
   p.stats.isFinal = true
 
-  trace "finishing performance benchmark", p.stats.duration
+  trace "finishing performance benchmark", duration = p.stats.duration
 
   return p.stats.duration
--- a/libp2p/protocols/perf/client.nim
+++ b/libp2p/protocols/perf/client.nim
@@ -54,7 +54,10 @@ proc perf*(
 
   while size > 0:
     let toRead = min(size, PerfSize)
-    await conn.readExactly(addr buf[0], toRead.int)
+    try:
+      await conn.readExactly(addr buf[0], toRead.int)
+    except LPStreamClosedError:
+      break
     size = size - toRead
     p.stats.duration = Moment.now() - start
     p.stats.downloadBytes += toRead
--- a/libp2p/protocols/perf/client.nim
+++ b/libp2p/protocols/perf/client.nim
@@ -72,6 +72,6 @@ proc perf*(
 
   p.stats.isFinal = true
 
-  let duration = Moment.now() - start
-  trace "finishing performance benchmark", duration
-  return duration
+  trace "finishing performance benchmark", p.stats.duration
+  
+  return p.stats.duration
--- a/libp2p/protocols/perf/client.nim
+++ b/libp2p/protocols/perf/client.nim
@@ -36,41 +36,47 @@ proc perf*(
 ): Future[Duration] {.public, async: (raises: [CancelledError, LPStreamError]).} =
   trace "starting performance benchmark", conn, sizeToWrite, sizeToRead
 
-  var
-    size = sizeToWrite
-    buf: array[PerfSize, byte]
-
   p.stats = Stats()
-  let start = Moment.now()
 
-  await conn.write(toSeq(toBytesBE(sizeToRead)))
-  while size > 0:
-    let toWrite = min(size, PerfSize)
-    await conn.write(buf[0 ..< toWrite])
-    size -= toWrite.uint
+  try:
+    var
+      size = sizeToWrite
+      buf: array[PerfSize, byte]
 
-    # set stats using copy value to avoid race condition
-    var statsCopy = p.stats
-    statsCopy.duration = Moment.now() - start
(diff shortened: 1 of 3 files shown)


libp2p/protocols/perf/client.nim
```diff
@@ -16,25 +16,37 @@ import ./core, ../../stream/connection
...
+
+proc new*(T: typedesc[PerfClient]): T =
+  return T()
 
 proc perf*(
-    _: typedesc[PerfClient],
-    conn: Connection,
-    sizeToWrite: uint64 = 0,
-    sizeToRead: uint64 = 0,
+    p: PerfClient, conn: Connection, sizeToWrite: uint64 = 0, sizeToRead: uint64 = 0
```
richard-ramos, line 35: According to the readme:
```
Some of libp2p procedures are marked as .public., they will remain compatible during each MAJOR version
```
With this change, we're breaking this rule. Do we know if `perf` is being used somewhere in nwaku / codex / nimbus?

vladopajic: if braking change is introduced then release must follow rules of braking change (so no rule is broken).

i haven't checked if anyone else is using this, and regardless of that we should consider following the rules.
however, `Perf` protocol only makes sense for benchmarking, which is very unlikely to be used anywhere else except in `libp2p/test-plans/perf`.

richard-ramos: Alright, i guess in this case we can bend the rule!

libp2p/protocols/perf/client.nim
```diff
@@ -16,25 +16,44 @@ import ./core, ../../stream/connection
...
+  duration*: Duration
+
+type PerfClient* = ref object
+  stats: Stats
+
+proc new*(T: typedesc[PerfClient]): T =
+  return T()
+
+proc currentStats*(p: PerfClient): Stats =
+  p.stats
```
richard-ramos: for consistency either both use `return` or none!

richard-ramos (approved):

libp2p/protocols/perf/client.nim
```diff
@@ -16,25 +16,44 @@ import ./core, ../../stream/connection
...
+  isFinal*: bool
+  uploadBytes*: uint
+  downloadBytes*: uint
+  duration*: Duration
+
+type PerfClient* = ref object
+  stats: Stats
+
+proc new*(T: typedesc[PerfClient]): T =
+  return T()
```
gmelodie, line 29: Why is this needed?

libp2p/protocols/perf/client.nim
```diff
@@ -16,25 +16,44 @@ import ./core, ../../stream/connection
...
+  p.stats
 
 proc perf*(
-    _: typedesc[PerfClient],
-    conn: Connection,
-    sizeToWrite: uint64 = 0,
-    sizeToRead: uint64 = 0,
+    p: PerfClient, conn: Connection, sizeToWrite: uint64 = 0, sizeToRead: uint64 = 0
 ): Future[Duration] {.public, async: (raises: [CancelledError, LPStreamError]).} =
+  trace "starting performance benchmark", conn, sizeToWrite, sizeToRead
```
gmelodie, line 37: Why move this up?

libp2p/protocols/perf/client.nim
```diff
@@ -16,25 +16,44 @@ import ./core, ../../stream/connection
...
     let toWrite = min(size, PerfSize)
     await conn.write(buf[0 ..< toWrite])
-    size -= toWrite
+    size -= toWrite.uint64
+
+    # set stats using copy value to avoid race condition
+    var statsCopy = p.stats
+    statsCopy.duration = Moment.now() - start
+    statsCopy.uploadBytes += toWrite.uint64
+    p.stats = statsCopy
```
gmelodie: Avoids race condition but might result in incorrect values being set?
1. I don't think you need that for `statsCopy.duration` since it's just setting, not reading.
2. I don't have much context on `perf` here but it seems that `p.stats.uploadBytes` might be used by other threads/futures? If that's the case we need a lock, or else what happens if it's updated by some other actor after line `55` ran?

libp2p/protocols/perf/client.nim
```diff
@@ -43,8 +62,16 @@ proc perf*(
...
     let toRead = min(size, PerfSize)
     await conn.readExactly(addr buf[0], toRead.int)
-    size = size - toRead
+    size = size - toRead.uint64
+
+    # set stats using copy value to avoid race condition
+    var statsCopy = p.stats
+    statsCopy.duration = Moment.now() - start
+    statsCopy.downloadBytes += toRead.uint64
+    p.stats = statsCopy
```
gmelodie: Same as above

libp2p/protocols/perf/client.nim
```diff
@@ -43,8 +62,16 @@ proc perf*(
...
+
+  p.stats.isFinal = true
+
+  trace "finishing performance benchmark", duration = p.stats.duration
 
-  let duration = Moment.now() - start
-  trace "finishing performance benchmark", duration
-  return duration
+  return p.stats.duration
```
gmelodie, line 83: Save `p.stats.duration` in a variable and use that for both `trace` and `return` to avoid it having different values (debugging in this case would be hell)

tests/testperf.nim
```diff
@@ -0,0 +1,74 @@
...
+  var builder = SwitchBuilder
+    .new()
+    .withRng(newRng())
+    .withAddresses(@[MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet()])
+    .withTcpTransport()
+    .withNoise()
+  if useMplex:
+    builder = builder.withMplex()
+  if useYamux:
+    builder = builder.withYamux()
```
gmelodie, line 30: This should be 1 variable, or (I think) it could lead to behavior you don't want (i.e. no `mplex` nor `yamux`). Ignore this comment if there's a case where you're using neither, but by the looks of it that doesn't seem to be the case.

gmelodie (changes requested):
Blocking only because of the race condition question.

libp2p/protocols/perf/client.nim
vladopajic: image

dont want logging to affect measuring time

tests/testperf.nim
vladopajic: that's fine. if you do something like that test will not work. this is simple utility local to this file

libp2p/protocols/perf/client.nim
vladopajic: >I don't think you need that for statsCopy.duration since it's just setting, not reading

it's needed in pr from description


> I don't have much context on perf here but it seems that p.stats.uploadBytes might be used by other threads/futures?

no they could not. only way to access `uploadBytes` by other threads is by `currentStats` proc which will return copy of stats. anything return at that point is fine. since stats is modified using copy value by only one thread.

--- 

race condition is following:

```
p.stats.duration = ...
p.stats.uploadBytes = ...
```

if one gets stats in the middle of setting these two variables they would get duration increased but not uploadbytes, which can be misleading metrics at that point.

---

locks are not used for reason - not to affect performance.

libp2p/protocols/perf/client.nim
vladopajic: no, duration is needed in stats. stats contain all infromation of perf. why have separate variable?

libp2p/protocols/perf/client.nim
gmelodie: Alright but this does not look like a very elegan solution. We should use an [atomic](https://nim-lang.org/docs/atomics.html) tuple for these two (or maybe for the entire struct if at all possible). Something else you can do is use locks but calculate the `duration` before acquiring the `lock`

libp2p/protocols/perf/client.nim
gmelodie: I mean this:
```nim
let duration = p.stats.duration
trace "finishing performance benchmark", duration = p.stats.duration
return duration
```

Ignore if nobody else is writing to `p.stats.duration`

libp2p/protocols/perf/client.nim
vladopajic: yeah could be done with one atomic

libp2p/protocols/perf/client.nim
richard-ramos: I was curious about this, so, I went over to libp2p/test-plans repo and saw that they create the perf protocol as part of the test so it's not really used anywhere else. Based on this i think it would make sense to move libp2p/protocols/perf  inside the interop folder.

vladopajic: > Alright, i guess in this case we can bend the rule!

👽

vladopajic: > I was curious about this, so, I went over to libp2p/test-plans repo and saw that they create the perf protocol as part of the test so it's not really used anywhere else. Based on this i think it would make sense to move libp2p/protocols/perf  inside the interop folder.

hmmm..  [pref](https://github.com/libp2p/specs/blob/master/perf%2Fperf.md) is just a protocol like any other (identity for example). regardless of how and where protocol is used all protocols should be under the protocol directory. if one is searching for protocol they would know where to look.

also tomorrow someone might use protocol in a different way then it might not be ideally suited to be in the interop folder.

now, perf client-server implementations form test-plans repo could go in the interop folder, but i dont see strong point for that.

libp2p/protocols/perf/client.nim
vladopajic: however, is there an issue with the current code that iam missing?

libp2p/protocols/perf/client.nim
vladopajic: to ignore what?

libp2p/protocols/perf/client.nim
gmelodie: No issue no :)

libp2p/protocols/perf/client.nim
gmelodie: Ignore the comment if nobody else is writing to `p.stats.duration` (if they're only reading from it the code is fine)

gmelodie (approved):

libp2p/protocols/perf/client.nim
vladopajic: not really needed. but reasoning was this:

```
Perf.new()              # creates protocol/server
PerfClient.new()   # creates client
```

PR #1453, opened 2025-06-09
feat(kad-dht): protobuffers

richard-ramos:


libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -0,0 +1,161 @@
...
+    recPb: seq[byte]
+    closerPbs: seq[seq[byte]]
+    providerPbs: seq[seq[byte]]
+
+  var pb = initProtoBuffer(buf)
+
+  var msgTypeVal: uint32
+  if not ?pb.getField(1, msgTypeVal).toOpt():
+    return Opt.none(T)
+  else:
```
gmelodie: I don't think you need an `else` here

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -0,0 +1,161 @@
...
+  discard ?pb.getRepeatedField(8, closerPbs).toOpt()
+  for ppb in closerPbs:
+    let peer = Peer.decode(initProtoBuffer(ppb))
+    if peer.isSome:
+      m.closerPeers.add(peer.unsafeGet())
+
+  discard ?pb.getRepeatedField(9, providerPbs).toOpt()
+  for ppb in providerPbs:
+    let peer = Peer.decode(initProtoBuffer(ppb))
+    if peer.isSome:
```
gmelodie: Can you use `withValue` here?

tests/kademlia/testencoding.nim
```diff
@@ -0,0 +1,157 @@
...
+      decoded.key.get() == rec.key.get()
+      decoded.value.get() == rec.value.get()
+      decoded.timeReceived.get() == rec.timeReceived.get()
+
+  test "peer encode/decode":
+    let maddr = MultiAddress.init("/ip4/127.0.0.1/tcp/9000").tryGet()
+    let peer = Peer(id: @[1'u8, 2, 3], addrs: @[maddr], connection: connected)
+    let encoded = peer.encode()
+    let decodedRes = Peer.decode(initProtoBuffer(encoded.buffer))
+    check decodedRes.isSome()
```
gmelodie: Maybe add a `doAssert` here so you don't continue in case `decodedRes.isNone`

tests/kademlia/testencoding.nim
```diff
@@ -0,0 +1,157 @@
...
+    let decoded = decodedRes.get()
+    check:
+      decoded.id == peer.id
+      decoded.addrs.len == 1
```
gmelodie: Maybe add a check for `peer.addrs.len == 1` too to show that they're both len of 1

tests/kademlia/testencoding.nim
```diff
@@ -0,0 +1,157 @@
...
+      decoded.addrs[0].data == maddr.data
```
gmelodie: Maybe add `doAssert` or an if statement so you dont do this in case `decoded.addrs.len == 0`

tests/kademlia/testencoding.nim
```diff
@@ -0,0 +1,157 @@
...
+      decoded.connection == connected
+
+  test "message encode/decode roundtrip":
+    let maddr = MultiAddress.init("/ip4/10.0.0.1/tcp/4001").tryGet()
+    let peer = Peer(id: @[9'u8], addrs: @[maddr], connection: canConnect)
+    let msg = Message(
+      msgType: Opt.some(findNode),
```
gmelodie: I prefer `msgType: Opt.some(Message.findNode),` (isn't that why you added `{.pure.}`?)

tests/kademlia/testencoding.nim
```diff
@@ -0,0 +1,157 @@
...
+      ),
+      closerPeers: @[peer],
+      providerPeers: @[peer],
+    )
+    let encoded = msg.encode()
+    let decodedRes = Message.decode(encoded.buffer)
+    check decodedRes.isSome()
+    let decoded = decodedRes.get()
+    check:
+      decoded.msgType.get() == findNode
```
gmelodie: Same `findNode` comment as before: `Message.findNode`

tests/kademlia/testencoding.nim
```diff
@@ -0,0 +1,157 @@
...
+      decoded.key.get() == @[7'u8]
+      decoded.record.get().key.get() == @[1'u8]
+      decoded.closerPeers.len == 1
+      decoded.providerPeers.len == 1
+
+  test "decode record with missing fields":
+    var pb = initProtoBuffer()
+    # no fields written
+    let res = Record.decode(pb)
+    check res.isSome()
```
gmelodie: Again, consider adding a `doAssert` here

tests/kademlia/testencoding.nim
```diff
@@ -0,0 +1,157 @@
...
+    var pb = initProtoBuffer()
+    let res = Peer.decode(pb)
+    check res.isNone()
+
+  test "decode peer with invalid connection type":
+    var pb = initProtoBuffer()
+    pb.write(1, @[1'u8, 2, 3]) # id field
+    pb.write(3, uint32(999)) # bogus connection type
+    let res = Peer.decode(pb)
+    check res.isNone()
```
gmelodie: I wonder if `Peer.decode` should be throwing an error when `pb` is invalid.

tests/kademlia/testencoding.nim
```diff
@@ -0,0 +1,157 @@
...
+    check res.isSome()
+    let rec = res.get()
+    check rec.key.isNone()
+    check rec.value.isNone()
+    check rec.timeReceived.isNone()
+
+  test "decode peer with missing id (invalid)":
+    var pb = initProtoBuffer()
+    let res = Peer.decode(pb)
+    check res.isNone()
```
gmelodie: I wonder if `Peer.decode` should be throwing an error when `pb` is invalid.

tests/kademlia/testencoding.nim
```diff
@@ -0,0 +1,157 @@
...
+    pb.write(1, @[1'u8, 2, 3]) # id field
+    pb.write(3, uint32(999)) # bogus connection type
+    let res = Peer.decode(pb)
+    check res.isNone()
+
+  test "decode message with invalid msgType":
+    var pb = initProtoBuffer()
+    pb.write(1, uint32(999)) # invalid MessageType
+    let res = Message.decode(pb.buffer)
+    check res.isNone()
```
gmelodie: I wonder if `Message.decode` should be throwing an error when `pb` is invalid.

tests/kademlia/testencoding.nim
```diff
@@ -0,0 +1,157 @@
...
+
+  test "decode message with invalid peer in closerPeers":
+    let badPeerBuf = @[0'u8, 1, 2] # junk
+    var pb = initProtoBuffer()
+    pb.write(8, badPeerBuf) # closerPeers field
+    let res = Message.decode(pb.buffer)
+    check res.isNone()
```
gmelodie: I wonder if `Message.decode` should be throwing an error when pb is invalid.

tests/kademlia/testencoding.nim
gmelodie: Also maybe make `999` a variable (maybe `const`?) named `invalidMsgType`.

tests/kademlia/testencoding.nim
```diff
@@ -0,0 +1,157 @@
...
+
+  test "decode message with invalid embedded record":
+    # encode junk data into field 3 (record)
+    var pb = initProtoBuffer()
+    pb.write(1, uint32(putValue)) # valid msgType
+    pb.write(3, @[0x00'u8, 0xFF, 0xAB]) # broken protobuf for record
+    let res = Message.decode(pb.buffer)
+    check res.isSome()
+    check res.unsafeGet().record.isNone()
```
gmelodie: I wonder if `Message.decode` should be throwing an error when pb is invalid.

again: `putValue` -> `Message.putValue`

gmelodie (approved):
Only nitpicks

tests/kademlia/testencoding.nim
richard-ramos: It will fail anyway with something like this:
```
        ...     /home/r/vacp2p/nim-libp2p/tests/kademlia/testencoding.nim(39, 27): Check failed: decodedRes.isSome()
        ...  ... ...
        ...     Unhandled defect: Trying to access value with err Result [ResultDefect]
        ...   [FAILED ] (  0.00s) peer encode/decode
```

tests/kademlia/testencoding.nim
richard-ramos: Not necessary as you can see on line 34 that the peer is being manually built with a single multiaddress and the test case is not testing input value construction, but that encoding / decoding is correct (the focus is on the result of the *code functions)

tests/kademlia/testencoding.nim
richard-ramos: The {.pure.} does let you use the enum without the `Message` prefix, but i agree that having `Message.findNode` is aesthetically more pleasant than just `findNode`!

tests/kademlia/testencoding.nim
richard-ramos: Assuming that len == 0, the test would immediatly fail and show a stacktrace indicating where it failed.

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -0,0 +1,157 @@
...
+proc encode*(record: Record): ProtoBuffer {.raises: [].} =
+  var pb = initProtoBuffer()
+  pb.writeOpt(1, record.key)
+  pb.writeOpt(2, record.value)
+  pb.writeOpt(5, record.timeReceived)
+  pb.finish()
+  return pb
+
+proc write*(pb: var ProtoBuffer, field: int, value: Record) {.raises: [].} =
+  pb.write(field, value.encode())
```
vladopajic: nit: move this closer to other write procs

tests/kademlia/testencoding.nim
```diff
@@ -0,0 +1,153 @@
...
+      key: some(@[1'u8, 2, 3]),
+      value: some(@[4'u8, 5, 6]),
+      timeReceived: some("2025-05-12T12:00:00Z"),
+    )
+    let encoded = rec.encode()
+    let decoded = Record.decode(encoded).valFromResultOption
+    check:
+      decoded.key.get() == rec.key.get()
+      decoded.value.get() == rec.value.get()
+      decoded.timeReceived.get() == rec.timeReceived.get()
```
vladopajic, line 38: is there method to do deep equal for two objects? here and in other places.

PR #1455, opened 2025-06-09
feat(kad-dht): handler

richard-ramos:


gmelodie (approved):

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -0,0 +1,74 @@
...
+        peerId = conn.PeerId, err = getCurrentExceptionMsg()
+    finally:
+      await conn.close()
+
+  return kad
+
+method start*(
+    kad: KadDHT
+): Future[void] {.async: (raises: [CancelledError], raw: true).} =
+  let fut = newFuture[void]()
```
vladopajic, line 54: why is this future needed?

richard-ramos: It seems because of the `raw: true` on the method definition.
If not returning a future, then we end up with this err:
```
/home/r/vacp2p/nim-libp2p/libp2p/switch.nim(371) start
/home/r/vacp2p/nim-libp2p/libp2p/multistream.nim(258) start
/home/r/.nimble/pkgs2/chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a/chronos/futures.nim(208) finished
/home/r/.nimble/pkgs2/chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a/chronos/futures.nim(199) state
```

the `mount` method of switch fails doing an `await` on the start method.

gmelodie: This is done in other `start` methods as well

vladopajic: yeah, got it. 

it looked unnecessary, unless `raw` is needed to comply with the method sigantue.

vladopajic (approved):

libp2p/protocols/kademlia/kademlia.nim
gmelodie: I think summoning the wizard (cc @arnetheduck) here is appropriate. Why do we need this?

PR #1457, opened 2025-06-10
chore(acme): add MockACMEApi for testing

gmelodie:
1. Add a mocked ACME API for unit testing
2. Slightly modify `libp2p/autotls/acme/api.nim`
3. Move some tests from `testautotls_integration.nim` to `testautotls.nim`

### DIFF
--- a/libp2p/autotls/acme/api.nim
+++ b/libp2p/autotls/acme/api.nim
@@ -19,18 +19,22 @@ const
   DefaultRandStringSize = 256
   ACMEHttpHeaders = [("Content-Type", "application/jose+json")]
 
-type Nonce = string
-type Kid = string
+type Nonce* = string
+type Kid* = string
 
-type ACMEDirectory = object
-  newNonce: string
-  newOrder: string
-  newAccount: string
+type ACMEDirectory* = object
+  newNonce*: string
+  newOrder*: string
+  newAccount*: string
 
-type ACMEApi* = object
+type ACMEApi* = ref object of RootObj
   directory: ACMEDirectory
   session: HttpSessionRef
-  acmeServerURL: string
+  acmeServerURL*: string
+
+type HTTPResponse* = object
+  body*: JsonNode
+  headers*: HttpTable
 
 type JWK = object
   kty: string
@@ -94,10 +98,10 @@ type ACMEChallengeResponseBody = object
   finalize: string
 
 type ACMEChallengeResponse* = object
-  status: ACMEChallengeStatus
-  authorizations: seq[string]
-  finalize: string
-  orderURL: string
+  status*: ACMEChallengeStatus
+  authorizations*: seq[string]
+  finalize*: string
+  orderURL*: string
 
 type ACMEChallengeResponseWrapper* = object
   finalizeURL*: string
@@ -105,7 +109,7 @@ type ACMEChallengeResponseWrapper* = object
   dns01*: ACMEChallenge
 
 type ACMEAuthorizationsResponse* = object
-  challenges: seq[ACMEChallenge]
+  challenges*: seq[ACMEChallenge]
 
 type ACMECompletedResponse* = object
   checkURL: string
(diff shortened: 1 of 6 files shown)


tests/testautotls.nim
```diff
@@ -0,0 +1,90 @@
...
+
+import sequtils, json
+import chronos, chronos/apps/http/httpclient
+import ../libp2p/[stream/connection, upgrademngrs/upgrade, autotls/acme/mockapi, wire]
+
+import ./helpers
+
+suite "AutoTLS ACME Client":
+  var api {.threadvar.}: MockACMEApi
+  var key {.threadvar.}: KeyPair
```
vladopajic, line 22: could we initialize  these variables here?

libp2p/autotls/acme/api.nim
```diff
@@ -210,15 +214,26 @@ proc acmeHeader(
...
+): Future[ACMEResponse] {.
+    async: (raises: [ACMEError, HttpError, CancelledError]), base
 .} =
-  await HttpClientRequestRef
+  let rawResponse = await HttpClientRequestRef
   .post(self.session, url, body = payload, headers = ACMEHttpHeaders)
   .get()
   .send()
+  let body = await rawResponse.getResponseBody()
+  ACMEResponse(body: body, headers: rawResponse.headers)
```
vladopajic: nit: `ACMEResponse` naming seems bit off. in this context it does represent response of ACME api. but it's type is not really tided to anything from ACME. what about just `HTTPResponse`?

libp2p/autotls/acme/mockapi.nim
```diff
@@ -0,0 +1,36 @@
...
+import ./api, ./utils
+
+export api
+
+type MockACMEApi* = ref object of ACMEApi
+  mockedHeaders*: HttpTable
+  mockedBody*: JsonNode
+
+proc new*(
+    T: typedesc[MockACMEApi], acmeServerURL: string = LetsEncryptURL
```
vladopajic: is for mock `acmeServerURL` really needed?

tests/testautotls.nim
```diff
@@ -0,0 +1,90 @@
...
+
+{.push raises: [].}
+
+import sequtils, json
+import chronos, chronos/apps/http/httpclient
+import ../libp2p/[stream/connection, upgrademngrs/upgrade, autotls/acme/mockapi, wire]
+
+import ./helpers
+
+suite "AutoTLS ACME Client":
```
vladopajic, line 20: would be interesting to test:
- `challengeCompleted` because it has loop
  -  case when loop condition is never satisfied (must return)
  - case when loop condition is satisfied
- `finalizeCertificate` same as above
- all methods can have simple test, where body is not matching schema (type) used for de-serialization - expect error.

vladopajic (approved):
famoso!

tests/testautotls.nim
gmelodie: not if we want to have `checkTrackers` on `teardown` because it'll check for unclosed futures and in this case `api` will have its `session` open

libp2p/autotls/acme/mockapi.nim
gmelodie: i think this way the footprint of the mock is smaller i.e. the user calls `new` on `MockACMEApi` the same way they would call it on `ACMEApi`, this way a person could just copy code from the tests and remove the `Mock` and it'd work

vladopajic: that usecase will probably never happen. and is probably not ideal approach.

what happens when someone adds new property to constructor to non-mocked struct? do they have to add property to mocked struct? will they remember to do that? what if they forgot, then why have rule if it is not enforced?
what happens when someone adds new property to mock, lets say mock has to be created with something additional to regular field. do you add same field to non-mock struct?

it's always better to remove anything unused. or if you don't do it now, other developer will clean it up after you ("oh this is unused, let me remove that").

if you want to have drop in replacment then it's solved via different approach.

PR #1461, opened 2025-06-11
test(gossipsub): refactor and unify scoring tests

rlve:
Changes:
- refactor `GossipSub Scoring` tests to latest conventions before adding new tests in this area
- extract common utilities

### DIFF
--- a/tests/pubsub/utils.nim
+++ b/tests/pubsub/utils.nim
@@ -20,6 +20,7 @@ import
   ]
 import ../helpers
 import chronicles
+import metrics
 
 export builders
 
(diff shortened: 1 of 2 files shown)


tests/pubsub/utils.nim
```diff
@@ -35,6 +36,12 @@ proc waitForHeartbeat*(multiplier: int = 1) {.async.} =
 proc waitForHeartbeat*(timeout: Duration) {.async.} =
   await sleepAsync(timeout)
 
+proc waitForHeartbeatByEvent*[T: PubSub](node: T, multiplier: int = 1) {.async.} =
+  for _ in 0 ..< multiplier:
+    let evnt = newAsyncEvent()
+    node.heartbeatEvents &= evnt
+    await evnt.wait()
```
rlve, line 43: Found this in one of old tests and it seems to be more predictable than waiting with `sleepAsync`. Will try to refactor tests to use it in future PR.

PR #1466, opened 2025-06-16
fix: force close of streams instead of reset when closing connection

### LINKED ISSUE #1462: bug(perf): client fails when used with yamux
when perf client and server are configured to use yamux, `client.perf()` proc will fail with `Stream Closed! [LPStreamClosedError]` error.

richard-ramos:
Fixes #1462

Even tho we do not handle half closed states correctly, i dived deeper into how [js-waku handles yamux closing a connection](https://github.com/ChainSafe/js-libp2p-yamux/blob/master/src/muxer.ts#L308) and noticed that streams are not reset there but closed instead, so implemented something similar while we refactor the connection abstractions.

### DIFF
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -512,7 +512,14 @@ method close*(m: Yamux) {.async: (raises: []).} =
   trace "Closing yamux"
   let channels = toSeq(m.channels.values())
   for channel in channels:
-    await channel.reset(isLocal = true)
+    for (d, s, fut) in channel.sendQueue:
+      fut.fail(newLPStreamEOFError())
+    channel.sendQueue = @[]
+    channel.sendWindow = 0
+    channel.closedLocally = true
+    channel.opened = false
+    await channel.remoteClosed()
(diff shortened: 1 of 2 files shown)


gabrielmer (approved):
LGTM, thanks so much!

NagyZoltanPeter (approved):
LTGM as much as I can judge!

libp2p/muxers/yamux/yamux.nim
```diff
@@ -512,7 +512,11 @@ method close*(m: Yamux) {.async: (raises: []).} =
   trace "Closing yamux"
   let channels = toSeq(m.channels.values())
   for channel in channels:
-    await channel.reset(isLocal = true)
+    for (d, s, fut) in channel.sendQueue:
+      fut.fail(newLPStreamEOFError())
+    channel.sendQueue = @[]
+    channel.sendWindow = 0
+    channel.closedLocally = true
```
Ivansete-status, line 519: Maybe?
```suggestion
    channel.closedLocally = true
    channel.opened = false
```

Ivansete-status (approved):

PR #1467, opened 2025-06-17
test(gossipsub): rpcHandler - rateLimit and punishInvalidMessage tests

rlve:
Changes:
- add unit tests for Gossipsub `rpcHandler` around `rateLimit` and penalising peers with `punishInvalidMessage`:
  - `"Peer is disconnected and rate limit is hit when overhead rate limit is exceeded"`
  -  `"Peer is punished when message contains invalid sequence number"`
  -  `"Peer is punished when message id generation fails"`
  -  `"Peer is punished when signature verification fails"`
  -  `"Peer is punished when message validation is rejected"`

### DIFF
--- a/tests/pubsub/testgossipsub.nim
+++ b/tests/pubsub/testgossipsub.nim
@@ -9,7 +9,6 @@
 
 {.used.}
 
-import chronicles
 import stew/byteutils
 import utils
 import ../../libp2p/protocols/pubsub/[gossipsub, mcache, peertable]
@@ -81,7 +80,6 @@ suite "GossipSub":
       peer.behaviourPenalty > 0.0
 
     await conn.close()
-    await gossipSub.switch.stop()
 
   asyncTest "invalid message bytes":
     let gossipSub = TestGossipSub.init(newStandardSwitch())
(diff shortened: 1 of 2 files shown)


tests/pubsub/testgossipsub.nim
```diff
@@ -92,4 +91,146 @@ suite "GossipSub":
     expect(CatchableError):
       await gossipSub.rpcHandler(peer, @[byte 1, 2, 3])
 
-    await gossipSub.switch.stop()
+  asyncTest "Peer is disconnected and rate limit is hit when overhead rate limit is exceeded":
+    # Given a GossipSub instance with one peer
+    const topic = "foobar"
```
gmelodie: Consider making this a module-scoped `const`, also fix places where `topic` is a mutable variable (`let topic`) above.

tests/pubsub/testgossipsub.nim
```diff
@@ -92,4 +91,146 @@ suite "GossipSub":
...
+    let
+      (gossipSub, conns, peers) = setupGossipSubWithPeers(1, topic)
+      peer = peers[0]
+      rateLimitHits = currentRateLimitHits("unknown")
+    defer:
+      await teardownGossipSub(gossipSub, conns)
```
gmelodie, line 101: I think we should move this to a `teardown` block and move all tests that need this to a separate `suite` with that `teardown` block

gmelodie (approved):

tests/pubsub/testgossipsub.nim
```diff
@@ -9,23 +9,26 @@
...
+import ../../libp2p/protocols/pubsub/[gossipsub, mcache, peertable, pubsubpeer]
 import ../../libp2p/protocols/pubsub/rpc/[message, protobuf]
 import ../helpers
 
 suite "GossipSub":
-  teardown:
+  const topic = "foobar"
+  var gossipSub {.threadvar.}: TestGossipSub
+  var conns {.threadvar.}: seq[Connection]
+  var peers {.threadvar.}: seq[PubSubPeer]
```
vladopajic: this will be a problem if we want to run tests in parallel

vladopajic: also it doesn't make sense to have it as shared variables when all tests are using different variables. 

what was actually the reason for doing this?

rlve: I was trying to implement this suggestion https://github.com/vacp2p/nim-libp2p/pull/1467#discussion_r2159041846

rlve: reverted

vladopajic: it's not really needed. all variables are only used locally to tests, so there is no need to assign them to global variable.

rlve: Also it would be nice to have a test runner capable of running tests from the same file in parallel, but I don't think such exists at the moment

vladopajic: all variables should have smallest scope possible. there is no justification to unnecessary increase scope of variables, at it only can lead to issues down the road.

---

unittest2 had parallel execution, so that feature can be re implemented once again.

vladopajic (approved):

PR #1468, opened 2025-06-18
test(gossipsub): organise Behavior tests 

rlve:
Changes:
- organise GossipSub Behavior tests - rename and group by method


gmelodie (approved):
I wonder if we should group these using suites `suite "handleIHave"` instead of naming  `asyncTest "handleIHave - ..."`

PR #1470, opened 2025-06-18
chore(acme): add ACMEClient to hold information about an ACME account

gmelodie:
This is what will be directly used by AutoTLS

### DIFF
--- a/libp2p/autotls/acme/api.nim
+++ b/libp2p/autotls/acme/api.nim
@@ -1,4 +1,4 @@
-import options, base64, sequtils, strutils, json
+import options, base64, sequtils, strutils, json, uri
 from times import DateTime, parse
 import chronos/apps/http/httpclient, jwt, results, bearssl/pem
 
@@ -19,8 +19,10 @@ const
   DefaultRandStringSize = 256
   ACMEHttpHeaders = [("Content-Type", "application/jose+json")]
 
-type Nonce* = string
+type Authorization* = string
+type Domain* = string
 type Kid* = string
+type Nonce* = string
 
 type ACMEDirectory* = object
   newNonce*: string
@@ -28,9 +30,9 @@ type ACMEDirectory* = object
   newAccount*: string
 
 type ACMEApi* = ref object of RootObj
-  directory: ACMEDirectory
+  directory: Opt[ACMEDirectory]
   session: HttpSessionRef
-  acmeServerURL*: string
+  acmeServerURL*: Uri
 
 type HTTPResponse* = object
   body*: JsonNode
@@ -49,7 +51,7 @@ type ACMERequestType = enum
 type ACMERequestHeader = object
   alg: string
   typ: string
-  nonce: string
+  nonce: Nonce
   url: string
   case kind: ACMERequestType
   of ACMEJwkRequest:
@@ -57,14 +59,16 @@ type ACMERequestHeader = object
   of ACMEKidRequest:
     kid: Kid
 
+type Email = string
+
 type ACMERegisterRequest* = object
   termsOfServiceAgreed: bool
-  contact: seq[string]
+  contact: seq[Email]
 
 type ACMEAccountStatus = enum
-  valid
-  deactivated
-  revoked
+  valid = "valid"
+  deactivated = "deactivated"
+  revoked = "revoked"
 
 type ACMERegisterResponseBody = object
   status*: ACMEAccountStatus
@@ -79,11 +83,18 @@ type ACMEChallengeStatus* {.pure.} = enum
   valid = "valid"
   invalid = "invalid"
 
-type ACMEChallenge = object
+type ACMEChallengeType* {.pure.} = enum
+  dns01 = "dns-01"
+  http01 = "http-01"
+  tlsalpn01 = "tls-alpn-01"
+
+type ACMEChallengeToken* = string
+
+type ACMEChallenge* = object
   url*: string
-  `type`*: string
+  `type`*: ACMEChallengeType
   status*: ACMEChallengeStatus
-  token*: string
+  token*: ACMEChallengeToken
 
 type ACMEChallengeIdentifier = object
   `type`: string
@@ -94,18 +105,18 @@ type ACMEChallengeRequest = object
 
 type ACMEChallengeResponseBody = object
   status: ACMEChallengeStatus
-  authorizations: seq[string]
+  authorizations: seq[Authorization]
   finalize: string
 
 type ACMEChallengeResponse* = object
   status*: ACMEChallengeStatus
-  authorizations*: seq[string]
+  authorizations*: seq[Authorization]
   finalize*: string
-  orderURL*: string
+  order*: string
 
 type ACMEChallengeResponseWrapper* = object
-  finalizeURL*: string
-  orderURL*: string
+  finalize*: string
+  order*: string
   dns01*: ACMEChallenge
 
 type ACMEAuthorizationsResponse* = object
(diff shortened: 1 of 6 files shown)


libp2p/autotls/acme/api.nim
```diff
@@ -144,6 +155,13 @@ type ACMECertificateResponse* = object
   rawCertificate: string
   certificateExpiry: DateTime
 
+proc toString*(status: ACMEChallengeStatus): string =
```
richard-ramos:
```suggestion
proc `$`*(status: ACMEChallengeStatus): string =
```

richard-ramos: Also, if the string value of the enum does not change, then you dont need to implement this. You can just do `$TheEnum`

libp2p/autotls/acme/client.nim
```diff
@@ -0,0 +1,71 @@
...
+
+proc getChallenge*(
+    self: ACMEClient, domains: seq[api.Domain]
+): Future[ACMEChallengeResponseWrapper] {.async: (raises: [ACMEError, CancelledError]).} =
+  await self.api.requestChallenge(domains, self.key, self.kid)
+
+proc getCertificate*(
+    self: ACMEClient, domain: api.Domain, challenge: ACMEChallengeResponseWrapper
+): Future[ACMECertificateResponse] {.async: (raises: [ACMEError, CancelledError]).} =
+  let chalURL = challenge.dns01.url
```
richard-ramos: Probably not needed?

richard-ramos (approved):

libp2p/autotls/acme/client.nim
gmelodie: You mean the function or `  let chalURL = challenge.dns01.url`?

tests/testautotls.nim
```diff
@@ -160,19 +190,48 @@ suite "AutoTLS ACME Client":
...
-      discard await api.requestGetOrder("some-order-url")
+      discard await api.requestGetOrder(parseUri("http://example.com/some-order-url"))
+
+suite "AutoTLS ACME Client":
+  var acmeApi {.threadvar.}: MockACMEApi
+  var acme {.threadvar.}: ACMEClient
+
+  asyncSetup:
+    acmeApi = await MockACMEApi.new()
+    acmeApi.mockedHeaders = HttpTable.init()
```
vladopajic: could headers be initiated inside `MovkACMApi.new` proc?

libp2p/autotls/acme/api.nim
```diff
@@ -161,24 +172,26 @@ template handleError*(msg: string, body: untyped): untyped =
...
+    T: typedesc[ACMEApi], acmeServerURL: Uri = parseUri(LetsEncryptURL)
 ): Future[ACMEApi] {.async: (raises: [ACMEError, CancelledError]).} =
   let session = HttpSessionRef.new()
   let directory = handleError("new API"):
-    let rawResponse =
-      await HttpClientRequestRef.get(session, acmeServerURL & "/directory").get().send()
+    let rawResponse = await HttpClientRequestRef
+    .get(session, $(acmeServerURL / "directory"))
+    .get()
+    .send()
```
vladopajic: directory request does is not separate method like other requests (`requestDirectory`), why? 
would it be possible to avoid http request in `new`? can this request be delayed to when needed?

libp2p/autotls/acme/api.nim
```diff
@@ -270,62 +283,69 @@ proc requestRegister*(
...
 
   return ACMEChallengeResponseWrapper(
-    finalizeURL: challengeResponse.finalize,
-    orderURL: challengeResponse.orderURL,
-    dns01: authorizationsResponse.challenges.filterIt(it.`type` == "dns-01")[0],
+    finalize: challengeResponse.finalize,
+    order: challengeResponse.order,
+    dns01: authorizationsResponse.challenges.filterIt(
+      it.`type` == ACMEChallengeType.dns01
+    )[0],
```
vladopajic, line 357: would be cool to add comment here that will explain why `filterIt(...)[0]` is safe and why

libp2p/autotls/acme/client.nim
```diff
@@ -0,0 +1,70 @@
...
+  let registerResponse = await api.requestRegister(key)
+  T(api: api, key: key, kid: registerResponse.kid)
+
+proc genKeyAuthorization(self: ACMEClient, token: string): KeyAuthorization =
+  base64UrlEncode(@(sha256.digest((token & "." & thumbprint(self.key)).toByteSeq).data))
+
+proc getChallenge*(
+    self: ACMEClient, domains: seq[api.Domain]
+): Future[ACMEChallengeResponseWrapper] {.async: (raises: [ACMEError, CancelledError]).} =
+  await self.api.requestChallenge(domains, self.key, self.kid)
```
vladopajic: does `getChallenge` needs `await`, or could it just return ?

libp2p/autotls/acme/client.nim
```diff
@@ -0,0 +1,70 @@
...
+
+proc getCertificate*(
+    self: ACMEClient, domain: api.Domain, challenge: ACMEChallengeResponseWrapper
+): Future[ACMECertificateResponse] {.async: (raises: [ACMEError, CancelledError]).} =
+  discard
+    await self.api.requestCompleted(parseUri(challenge.dns01.url), self.key, self.kid)
```
vladopajic: `requestCompleted` looks like weird name in this context (i had to see what this function actually does; after seeing "send notify" string it was)

libp2p/autotls/acme/client.nim
```diff
@@ -0,0 +1,70 @@
...
+
+  let completed = await self.api.checkChallengeCompleted(
+    parseUri(challenge.dns01.url), self.key, self.kid
```
vladopajic: `parseUri(challenge.dns01.url)` multiple times for same value

tests/testautotls.nim
```diff
@@ -160,19 +190,48 @@ suite "AutoTLS ACME Client":
...
-      discard await api.requestFinalize("some-domain", "some-finalize-url", key, "kid")
+      discard await api.requestFinalize(
+        "some-domain", parseUri("http://example.com/some-finalize-url"), key, "kid"
+      )
 
     expect(ACMEError):
-      discard await api.requestGetOrder("some-order-url")
+      discard await api.requestGetOrder(parseUri("http://example.com/some-order-url"))
+
+suite "AutoTLS ACME Client":
```
vladopajic, line 297: the key logic of this type is  `getCertificate` method. it would be nice to have test when everything is going on the happy path but `certificateFinalized` fails.

vladopajic (approved):

tests/testautotls.nim
gmelodie: nice catch!

libp2p/autotls/acme/client.nim
gmelodie: Suggestions? I agree, but I remember in past reviews you did not like the `signal` word either. What this method does is send an HTTP request to tell the ACME server that the challenge was completed and that it can start the challenge validation process

libp2p/autotls/acme/api.nim
gmelodie: Delaying the request will require us to check if directory is set in several functions and, if not, send a request for it. I could create an `init` function but that would require users to know that they need to call that. I get what you mean though, just out of better ideas. I didn't want to have `requestFoo` functions to do more than one HTTP call as that would complicate mocking.

libp2p/autotls/acme/client.nim
vladopajic: `sendChallengeCompleted` perhaps? ... it will violate convention but convention might needs update :)

libp2p/autotls/acme/api.nim
vladopajic: how about lazy initialization?

pseudo code:
```
proc getDirectory(): Directory
 if self.directory.isSome:
    return return self.directory.get()

 self.directory = requestDirectory(...)
 
 return  return self.directory.get()
    
```

then in all `requestXXXX` methods you would call `getDirectory` instead of accessing it as field directly.

gmelodie: > directory request does is not separate method like other requests (requestDirectory), why?

as per this, it's because the `self.get` calls you're referring to require an `ACMEApi`, which, to be constructed, needs the `directory`. Kind of a chicken an egg problem. Moving the call out of new would allow us to this though, but then we'd have the `init` issue I mentioned.

vladopajic: > as per this, it's because the `self.get` calls you're referring to require an `ACMEApi`, which, to be constructed, needs the `directory`. Kind of a chicken an egg problem. Moving the call out of new would allow us to this though, but then we'd have the `init` issue I mentioned.

what about suggestion from comment above?

gmelodie: Sorry async jitters, I didn't refresh the page so I didn't see your first response :p

that sounds good!

richard-ramos (approved):

PR #1472, opened 2025-06-19
feat(autotls): add AutoTLSManager

gmelodie:
This adds the `AutoTLSManager` for downloading certificates, which uses [`ACMEClient`](https://github.com/vacp2p/nim-libp2p/pull/1470) (a wrapper around [`ACMEApi`](https://github.com/vacp2p/nim-libp2p/pull/1436)) and [`PeerIDAuthClient`](https://github.com/vacp2p/nim-libp2p/pull/1445)

~~**Obs**: this should only be merged after ACMEClient is merged: https://github.com/vacp2p/nim-libp2p/pull/1470 (will rebase when that is done).~~ This is ready for merging.

### DIFF
--- a/libp2p/autotls/acme/api.nim
+++ b/libp2p/autotls/acme/api.nim
@@ -1,4 +1,4 @@
-import options, base64, sequtils, strutils, json, uri
+import options, sequtils, strutils, json, uri
 from times import DateTime, parse
 import chronos/apps/http/httpclient, jwt, results, bearssl/pem
 
@@ -83,6 +83,13 @@ type ACMEChallengeStatus* {.pure.} = enum
   valid = "valid"
   invalid = "invalid"
 
+type ACMEOrderStatus* {.pure.} = enum
+  pending = "pending"
+  ready = "ready"
+  processing = "processing"
+  valid = "valid"
+  invalid = "invalid"
+
 type ACMEChallengeType* {.pure.} = enum
   dns01 = "dns-01"
   http01 = "http-01"
@@ -104,12 +111,12 @@ type ACMEChallengeRequest = object
   identifiers: seq[ACMEChallengeIdentifier]
 
 type ACMEChallengeResponseBody = object
-  status: ACMEChallengeStatus
+  status: ACMEOrderStatus
   authorizations: seq[Authorization]
   finalize: string
 
 type ACMEChallengeResponse* = object
-  status*: ACMEChallengeStatus
+  status*: ACMEOrderStatus
   authorizations*: seq[Authorization]
   finalize*: string
   order*: string
@@ -123,14 +130,7 @@ type ACMEAuthorizationsResponse* = object
   challenges*: seq[ACMEChallenge]
 
 type ACMECompletedResponse* = object
-  checkURL: string
-
-type ACMEOrderStatus* {.pure.} = enum
-  pending = "pending"
-  ready = "ready"
-  processing = "processing"
-  valid = "valid"
-  invalid = "invalid"
+  url: string
 
 type ACMECheckKind* = enum
   ACMEOrderCheck
@@ -152,8 +152,8 @@ type ACMEOrderResponse* = object
   expires: string
 
 type ACMECertificateResponse* = object
-  rawCertificate: string
-  certificateExpiry: DateTime
+  rawCertificate*: string
+  certificateExpiry*: DateTime
 
 template handleError*(msg: string, body: untyped): untyped =
   try:
@@ -338,20 +338,20 @@ proc requestAuthorizations*(
 proc requestChallenge*(
     self: ACMEApi, domains: seq[Domain], key: KeyPair, kid: Kid
 ): Future[ACMEChallengeResponseWrapper] {.async: (raises: [ACMEError, CancelledError]).} =
-  let challengeResponse = await self.requestNewOrder(domains, key, kid)
-  if challengeResponse.status != ACMEChallengeStatus.pending:
-    raise newException(
-      ACMEError, "Invalid new challenge status: " & $challengeResponse.status
-    )
+  let orderResponse = await self.requestNewOrder(domains, key, kid)
+  if orderResponse.status != ACMEOrderStatus.pending and
+      orderResponse.status != ACMEOrderStatus.ready:
+    # ready is a valid status when renewing certs before expiry
+    raise newException(ACMEError, "Invalid new order status: " & $orderResponse.status)
 
   let authorizationsResponse =
-    await self.requestAuthorizations(challengeResponse.authorizations, key, kid)
+    await self.requestAuthorizations(orderResponse.authorizations, key, kid)
   if authorizationsResponse.challenges.len == 0:
     raise newException(ACMEError, "No challenges received")
 
   return ACMEChallengeResponseWrapper(
-    finalize: challengeResponse.finalize,
-    order: challengeResponse.order,
+    finalize: orderResponse.finalize,
+    order: orderResponse.order,
     dns01: authorizationsResponse.challenges.filterIt(
       it.`type` == ACMEChallengeType.dns01
     )[0],
@@ -396,7 +396,7 @@ proc requestCheck*(
 proc sendChallengeCompleted*(
     self: ACMEApi, chalURL: Uri, key: KeyPair, kid: Kid
 ): Future[ACMECompletedResponse] {.async: (raises: [ACMEError, CancelledError]).} =
-  handleError("sendChallengeCompleted (send notify)"):
+  handleError("sendChallengeCompleted"):
     let payload =
       await self.createSignedAcmeRequest(chalURL, %*{}, key, kid = Opt.some(kid))
     let acmeResponse = await self.post(chalURL, payload)
@@ -438,12 +438,9 @@ proc completeChallenge*(
 proc requestFinalize*(
     self: ACMEApi, domain: Domain, finalize: Uri, key: KeyPair, kid: Kid
 ): Future[ACMEFinalizeResponse] {.async: (raises: [ACMEError, CancelledError]).} =
-  let derCSR = createCSR(domain)
-  let b64CSR = base64.encode(derCSR.toSeq, safe = true)
-
   handleError("requestFinalize"):
     let payload = await self.createSignedAcmeRequest(
-      finalize, %*{"csr": b64CSR}, key, kid = Opt.some(kid)
+      finalize, %*{"csr": createCSR(domain)}, key, kid = Opt.some(kid)
     )
     let acmeResponse = await self.post(finalize, payload)
     # server responds with updated order response
--- a/libp2p/autotls/acme/api.nim
+++ b/libp2p/autotls/acme/api.nim
@@ -78,22 +78,22 @@ type ACMERegisterResponse* = object
   status*: ACMEAccountStatus
 
 type ACMEChallengeStatus* {.pure.} = enum
-  pending = "pending"
-  processing = "processing"
-  valid = "valid"
-  invalid = "invalid"
+  PENDING = "pending"
+  PROCESSING = "processing"
+  VALID = "valid"
+  INVALID = "invalid"
 
 type ACMEOrderStatus* {.pure.} = enum
-  pending = "pending"
-  ready = "ready"
-  processing = "processing"
-  valid = "valid"
-  invalid = "invalid"
+  PENDING = "pending"
+  READY = "ready"
+  PROCESSING = "processing"
+  VALID = "valid"
+  INVALID = "invalid"
 
 type ACMEChallengeType* {.pure.} = enum
-  dns01 = "dns-01"
-  http01 = "http-01"
-  tlsalpn01 = "tls-alpn-01"
+  DNS01 = "dns-01"
+  HTTP01 = "http-01"
+  TLSALPN01 = "tls-alpn-01"
 
 type ACMEChallengeToken* = string
 
@@ -213,7 +213,7 @@ method requestNonce*(
 proc acmeHeader(
     self: ACMEApi, uri: Uri, key: KeyPair, needsJwk: bool, kid: Opt[Kid]
 ): Future[ACMERequestHeader] {.async: (raises: [ACMEError, CancelledError]).} =
-  if not needsJwk and kid.isNone:
+  if not needsJwk and kid.isNone():
     raise newException(ACMEError, "kid not set")
 
   if key.pubkey.scheme != PKScheme.RSA or key.seckey.scheme != PKScheme.RSA:
@@ -318,7 +318,7 @@ proc requestNewOrder*(
     let acmeResponse =
       await self.post(parseUri((await self.getDirectory()).newOrder), payload)
     let challengeResponseBody = acmeResponse.body.to(ACMEChallengeResponseBody)
-    if challengeResponseBody.authorizations.len() == 0:
+    if challengeResponseBody.authorizations.len == 0:
       raise newException(ACMEError, "Authorizations field is empty")
     ACMEChallengeResponse(
       status: challengeResponseBody.status,
@@ -339,8 +339,8 @@ proc requestChallenge*(
     self: ACMEApi, domains: seq[Domain], key: KeyPair, kid: Kid
 ): Future[ACMEChallengeResponseWrapper] {.async: (raises: [ACMEError, CancelledError]).} =
   let orderResponse = await self.requestNewOrder(domains, key, kid)
-  if orderResponse.status != ACMEOrderStatus.pending and
-      orderResponse.status != ACMEOrderStatus.ready:
+  if orderResponse.status != ACMEOrderStatus.PENDING and
+      orderResponse.status != ACMEOrderStatus.READY:
     # ready is a valid status when renewing certs before expiry
     raise newException(ACMEError, "Invalid new order status: " & $orderResponse.status)
 
@@ -353,7 +353,7 @@ proc requestChallenge*(
     finalize: orderResponse.finalize,
     order: orderResponse.order,
     dns01: authorizationsResponse.challenges.filterIt(
-      it.`type` == ACMEChallengeType.dns01
+      it.`type` == ACMEChallengeType.DNS01
     )[0],
       # getting the first element is safe since we checked that authorizationsResponse.challenges.len != 0
   )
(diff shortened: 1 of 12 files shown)


libp2p/autotls/acme/api.nim
```diff
@@ -83,6 +83,13 @@ type ACMEChallengeStatus* {.pure.} = enum
   valid = "valid"
   invalid = "invalid"
 
+type ACMEOrderStatus* {.pure.} = enum
+  pending = "pending"
```
Ivansete-status: I think is not needed to give value. A pure enum should convert to its code representation when converting to string (`$`). Also, I think enum should be uppercase (nitpick comment.)
```suggestion
type ACMEOrderStatus* {.pure.} = enum
  PENDING
```
EDITED: ah! I see you moved code. The nitpick suggestion still applies ;P

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,292 @@
...
+      ip.startsWith("192.168.") or ip.startsWith("127.") or ip.startsWith("169.254.")
+    )
+  except ValueError as exc:
+    raise newException(AutoTLSError, "Failed to parse IP address", exc)
+
+proc asMoment*(dt: DateTime): Moment =
+  let unixTime: int64 = dt.toTime.toUnix
+  return Moment.init(unixTime, Second)
+
+proc encodePeerId*(peerId: PeerId): string {.raises: [AutoTLSError].} =
```
Ivansete-status: For brand new code, I'd suggest avoid raising exceptions. The suggestion applies elsewhere
```suggestion
proc encodePeerId*(peerId: PeerId): Result[string, string] {.raises: [].} =
```

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,292 @@
...
+      error "Failed to resolve IP", description = exc.msg # retry
+    if txt.len > 0 and txt[0] == keyAuth and ip4.len > 0:
+      return true
+    await sleepAsync(DefaultDnsRetryTime)
+
+  return false
+
+method issueCertificate(
+    self: AutoTLSManager
+): Future[void] {.
```
Ivansete-status: Nitcpick comment: `async` implicitly puts `Future[void]` as a returned value. 
```suggestion
) {.
```

libp2p/peeridauth/client.nim
```diff
@@ -174,9 +178,15 @@ method post*(
...
 
 method get*(
-    self: PeerIDAuthClient, uri: string
-): Future[PeerIDAuthResponse] {.async: (raises: [HttpError, CancelledError]), base.} =
-  let rawResponse = await HttpClientRequestRef.get(self.session, $uri).get().send()
+    self: PeerIDAuthClient, uri: Uri
+): Future[PeerIDAuthResponse] {.
+    async: (raises: [PeerIDAuthError, HttpError, CancelledError]), base
+.} =
+  if self.session.isNil:
```
Ivansete-status: Super nitpick, sorry: For "verbs" statements I always suggest using parenthesis and for "nouns" , withouth parenthesis (e.g., `len`)
Applies elsewhere :)
```suggestion
  if self.session.isNil():
```

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,292 @@
...
+      ip.startsWith("10.") or
+      (ip.startsWith("172.") and parseInt(ip.split(".")[1]) in 16 .. 31) or
+      ip.startsWith("192.168.") or ip.startsWith("127.") or ip.startsWith("169.254.")
+    )
+  except ValueError as exc:
+    raise newException(AutoTLSError, "Failed to parse IP address", exc)
+
+proc asMoment*(dt: DateTime): Moment =
+  let unixTime: int64 = dt.toTime.toUnix
+  return Moment.init(unixTime, Second)
```
Ivansete-status: I think the following is not needed.
I'd suggest using making `certExpiry` of type `DateTime`, which also contains a `now` and `Duration` arithmetic.
```suggestion
```

tests/testautotls_integration.nim
```diff
@@ -66,3 +71,73 @@ suite "AutoTLS Integration":
...
+
+    check certBefore != certAfter
+
+    let certExpiry = switch.autotls.certExpiry.valueOr:
+      raiseAssert "certificate expiry not found"
+
+    # cert is valid
+    check certExpiry > Moment.now
+
+    await switch.stop()
```
vladopajic: `await switch.stop()` - better have it just below switch.start() using defer

tests/testautotls_integration.nim
```diff
@@ -66,3 +71,73 @@ suite "AutoTLS Integration":
...
+    # certificate was downloaded and parsed
+    let cert = switch.autotls.cert.valueOr:
+      raiseAssert "certificate not found"
+    let certBefore = cert
+
+    # invalidate certificate
+    switch.autotls.certExpiry = Opt.some(Moment.now - 2.hours)
+
+    # cert was invalidated correctly
+    check switch.autotls.certExpiry.get < Moment.now
```
vladopajic: not sure if this check make sense at all, because this is asserting assignment (=) operation defined by the language.

tests/testautotls_integration.nim
```diff
@@ -66,3 +71,73 @@ suite "AutoTLS Integration":
     check challenge.dns01.`type` == ACMEChallengeType.dns01
     check challenge.dns01.status == ACMEChallengeStatus.pending
     check challenge.dns01.token.len() > 0
+  asyncTest "AutoTLSManager correctly downloads challenges":
```
vladopajic, line 75: add some new line space before test?

libp2p/peeridauth/client.nim
```diff
@@ -174,9 +178,15 @@ method post*(
...
+    raise newException(PeerIDAuthError, "Session is nil")
```
vladopajic, line 186: might be unnecessary since it is created in new and never modified and changed from outside nor inside the type

tests/testautotls_integration.nim
```diff
@@ -66,3 +71,73 @@ suite "AutoTLS Integration":
...
+    await switch.autotls.certReady.wait()
+    # clear since we'll use it again for renewal
+    switch.autotls.certReady.clear()
+
+    let dnsResolver = DnsResolver.new(
+      @[
+        initTAddress("1.1.1.1:53"),
+        initTAddress("1.0.0.1:53"),
+        initTAddress("[2606:4700:4700::1111]:53"),
+      ]
```
vladopajic: why is `DefaultDnsServers` not used here?

tests/testautotls.nim
```diff
@@ -63,7 +63,7 @@ suite "AutoTLS ACME API":
```
vladopajic, line 63: AutoTLSManager needs tests using mocked client

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,292 @@
...
+  let dashedIpAddr = ($ipAddress).replace(".", "-")
+  let acmeChalDomain = api.Domain("_acme-challenge." & baseDomain)
+  let ip4Domain = api.Domain(dashedIpAddr & "." & baseDomain)
+
+  var txt: seq[string]
+  var ip4: seq[TransportAddress]
+  for _ in 0 .. retries:
+    txt = await self.dnsResolver.resolveTxt(acmeChalDomain)
+    try:
+      ip4 = await self.dnsResolver.resolveIp(ip4Domain, 0.Port)
```
vladopajic: propagate cancelled error?

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,292 @@
...
+
+  debug "Registering ACME account"
+  if self.acmeClient.isNone:
+    self.acmeClient = Opt.some(await ACMEClient.new(acmeServerURL = self.acmeServerURL))
+
+  heartbeat "Certificate Management", self.renewCheckTime:
+    if self.cert.isNone or self.certExpiry.isNone:
+      try:
+        await self.issueCertificate()
+      except CatchableError as exc:
```
vladopajic, line 164: propagate cancelled error?

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,292 @@
...
+        error "Failed to issue certificate", err = exc.msg
+        break
+
+    # AutoTLSManager will renew the cert 1h before it expires
+    let expiry = self.certExpiry.get
+    let waitTime = expiry - Moment.now - self.renewBufferTime
+    if waitTime <= self.renewBufferTime:
+      try:
+        await self.issueCertificate()
+      except CatchableError as exc:
```
vladopajic, line 176: propagate cancelled error?

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,292 @@
...
+  renewCheckTime*: Duration
+  renewBufferTime*: Duration
+  peerInfo: Opt[PeerInfo]
+  acmeServerURL: Uri
+  ipAddress: Opt[IpAddress]
+
+proc new*(
+    T: typedesc[AutoTLSManager],
+    rng: ref HmacDrbgContext = newRng(),
+    acmeClient: Opt[ACMEClient] = Opt.none(ACMEClient),
```
vladopajic, line 70: this code can work even if `acmeClient` is not provided?

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,292 @@
...
+proc isPublic*(ip: IpAddress): bool {.raises: [AutoTLSError].} =
+  let ip = $ip
+  try:
+    not (
+      ip.startsWith("10.") or
+      (ip.startsWith("172.") and parseInt(ip.split(".")[1]) in 16 .. 31) or
+      ip.startsWith("192.168.") or ip.startsWith("127.") or ip.startsWith("169.254.")
+    )
+  except ValueError as exc:
+    raise newException(AutoTLSError, "Failed to parse IP address", exc)
```
vladopajic: move this to utils?

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,292 @@
...
+      if not ip.isPublic:
+        raise newException(AutoTLSError, "Host does not have a public IPv4 address")
+      ip
+    except AutoTLSError as exc:
+      raise exc
+    except CatchableError as exc:
+      raise newException(
+        AutoTLSError, "Unexpected error while getting primary IP address for host", exc
+      )
+  self.ipAddress = Opt.some(ipAddress)
```
vladopajic: code this code be moved to separate function?

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,292 @@
...
+
+  debug "Registering ACME account"
+  if self.acmeClient.isNone:
+    self.acmeClient = Opt.some(await ACMEClient.new(acmeServerURL = self.acmeServerURL))
```
vladopajic: why not initialize this in new?

libp2p/peeridauth/client.nim
gmelodie: I think of this as future proofing the code in case changes are made to session handling

libp2p/autotls/manager.nim
gmelodie: yes. `manageCertificate` will create a new client if one is not provided.

libp2p/peeridauth/client.nim
vladopajic: future proofing is tricky card to play.  it can be used almost everywhere but one don't really need to future proof every single line.  

in this case what is the worst think that can happen? nil ptr dereference should be solved easily (should it happen).

libp2p/autotls/manager.nim
vladopajic: since AutoTLSManager always requires some kind of `acmeClient` it is proabbly the best to initialize it in new, right away, and not have it as `Opt` property.

in general unnecessary `Opt` should be avoided.

libp2p/autotls/manager.nim
gmelodie: Because that would require `AutoTLSManager`'s `new` to be `async` and while building the switch it is a requirement that this should not be the case. `ACMEClient.new` is async because it registers a new ACME account and therefore sends HTTP requests. I didn't want to make `ACMEClient` synchronous because I wanted it to register a new account. I think it wouldn't make sense for the client not to have a registered account when created.

vladopajic: okay got it. 

if you don't like `AutoTLSManager.new` to be async why is it okay for `ACMEClient.new` ?

gmelodie: > you don't like AutoTLSManager.new to be async

I don't mind either way, but if `AutoTLSManager` were to be `async`, then so would `withAutotls`, and that breaks the switch building process. I see your point, but changing ACMEClient's conventions is not in the scope of this PR.

libp2p/autotls/manager.nim
gmelodie: There's two things to consider here:
1. Having the ability to pass an ACMEClient allows the users to be able to reuse ACME accounts, this is especially important for being able to control certificates for domains already owned by an account.
2.  Initializing ACMEClient requires `await`, which in turn would require `AutoTLSManager.new` to return a `Future`, which yields problems with `build` as stated in the `ACMEClient`. See below for why I think we should not be thinking of changing ACMEClient other than strictly necessary in this PR.

tests/testautotls.nim
gmelodie: I talked to @richard-ramos and he seems to agree that since AutoTLSManager is simply a wrapper for functions implemented elsewhere (`peeridauth` and `acme`), and since those are properly tested, then unit testing this too would be an overkill.

vladopajic (approved):

libp2p/autotls/manager.nim
richard-ramos: FWIW, having an async constructor looks like a code smell. Constructors should be lightweight and also follow the single responsibility principle, and be deterministic, while making it async you're smuggling some IO into instantiation. At that point, i'd just have a factory.

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,200 @@
...
+    managerFut: nil,
+    cert: Opt.none(TLSCertificate),
+    certExpiry: Opt.none(Moment),
+    certReady: newAsyncEvent(),
+    acmeClient: acmeClient,
+    brokerClient: brokerClient,
+    dnsResolver: dnsResolver,
+    bearer: Opt.none(BearerToken),
+    renewCheckTime: DefaultRenewCheckTime,
+    renewBufferTime: DefaultRenewBufferTime,
```
richard-ramos: Is it necessary to have these attributes? They are not part of the constructor params. Perhaps they should, and use the `Default___` constants as default values?

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,200 @@
...
+proc getIpAddress(self: AutoTLSManager): IpAddress {.raises: [AutoTLSError].} =
+  return self.ipAddress.valueOr:
+    getPublicIPADdress()
+
+method issueCertificate(
+    self: AutoTLSManager
+) {.base, async: (raises: [AutoTLSError, ACMEError, PeerIDAuthError, CancelledError]).} =
+  trace "Issuing new certificate"
+  let peerInfo = self.peerInfo.valueOr:
+    raise newException(AutoTLSError, "Cannot issue new certificate: peerInfo not set")
```
richard-ramos: Should this be an assertion or an exception? Assertions are meant to be used for invariants, like peerInfo being none indicating a bad configuration, while exceptions are for runtime errors.

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,200 @@
...
+
+  # generate autotls domain string: "*.{peerID}.libp2p.direct"
+  let base36PeerId = encodePeerId(peerInfo.peerId)
+  let baseDomain = api.Domain(base36PeerId & "." & AutoTLSDNSServer)
+  let domain = api.Domain("*." & baseDomain)
+
+  let acmeClient = self.acmeClient.valueOr:
```
richard-ramos, line 113: Not necessary to change it into this PR, but reading the code, I think we have a code smell, similar to prev comment about checking for peerInfo being setup:

The problem with managecertificate/issuecertificate it is that you're reliying in `start` function being called. In other words, you basically made a state machine :) , and hidden state in AutoTLSManager with no clear clues on what is valid when unless you read the code.

A solution for a future refactor would be to model this as a state machine by explicitly stating the state, or use something similar to phantom types in rust (more or less like it was done here https://github.com/waku-org/waku-rust-bindings/blob/57505a1c06e933b912dc657ada94965395ca30e7/waku-bindings/src/node/mod.rs#L37-L41) or the state machine implemented in nim-quic for connection states

libp2p/autotls/manager.nim
```diff
@@ -0,0 +1,200 @@
...
+  except TLSStreamProtocolError:
+    raise newException(AutoTLSError, "Could not parse downloaded certificates")
+  self.certReady.fire()
+
+proc manageCertificate(
+    self: AutoTLSManager
+) {.async: (raises: [AutoTLSError, ACMEError, CancelledError]).} =
+  trace "Starting AutoTLS manager"
+
+  debug "Registering ACME account"
```
richard-ramos: Do we need two log lines one after the other? or could we survive just with the debug line

libp2p/autotls/utils.nim
```diff
@@ -0,0 +1,116 @@
...
+  try:
+    not (
+      ip.startsWith("10.") or
+      (ip.startsWith("172.") and parseInt(ip.split(".")[1]) in 16 .. 31) or
+      ip.startsWith("192.168.") or ip.startsWith("127.") or ip.startsWith("169.254.")
+    )
+  except ValueError as exc:
+    raise newException(AutoTLSError, "Failed to parse IP address", exc)
+
+proc getPublicIPADdress*(): IpAddress {.raises: [AutoTLSError].} =
```
richard-ramos:
```suggestion
proc getPublicIPAddress*(): IpAddress {.raises: [AutoTLSError].} =
```

libp2p/switch.nim
```diff
@@ -332,6 +334,9 @@ proc stop*(s: Switch) {.public, async: (raises: [CancelledError]).} =
     except CatchableError as exc:
       warn "error cleaning up transports", description = exc.msg
 
+  if not s.autotls.isNil:
+    await s.autotls.stop()
```
richard-ramos, line 338: Wdyt of modeling AutoTLSManager as a service, so it's setup in line 374? The advantage of it is that Switch wouldnt need to get an additional attribute, the peerInfo can be extracted during the call to `setup`, and you wouldnt have to do this s.autotls.isNil check here and in the stop. 

Can of course be done on a separate PR, but for sure it would look more correct. Then withAutoTLS what it'd do is append autoTLS service to b.services in the SwitchBuilder

richard-ramos (approved):

libp2p/switch.nim
gmelodie: Neat! Will merge this and do this in a next PR

Issue #1417, opened 2025-05-27
make `nimble test` gracefull if there is no go deamon


vladopajic:
for the sake of simplicity, especially having in mind  new contributors, we should improve `nimble test` command so that it can run without crashing if go daemon is not installed.

we need to come up with solution what to do in CI, because there same `nimble test` must fail if go daemon is not installed. perhaps some new flag `--force-daemon` may be implemented and used only in CI?

gmelodie:
Fixed in #1705

PR #1477, opened 2025-06-23
chore(tests): move testdaemon out of testall

gmelodie:
Moving `testdaemon` out of `testall` allows `nimble test` to succeed without needing to install the go-daemon. CI still runs `testdaemon`.

Obs: I put this together last friday because I had some free time, but I'm not sure this is how we want to approach #1417.


vladopajic:
>  I just like this way better

you like having many commands to run tests vs one command to run tests? you want to force users to do the same? and to expect them to remember which type of tests they need to run?

---

user should have in mind only one command `nimble test`. if command fails due to daemon not being installed it should output message:

"failed to run daemon tests - go daemon is not installed. skip running daemon tests by including "--skip-daemon" flag to test command."

this approach will have all the benefits:
- single command == simplicity; simpler the better
- single compile - faster build
- compile errors in all files are checked with same command
- future proof, in case of similar patterns the same approach could be used

richard-ramos:
My 2c here is that we should not spend much time dealing with go-daemon, and even consider getting rid of it entirely, specially considering the following text on their readme: 
image

The usage of go-daemon probably made sense at the beginning of the project (7y ago) as gossipsub was probably being developed and we wanted to ensure interop, however at this we can probably replace this native code from nim-libp2p and proceed also to work on adding gossipsub interop tests once available in libp2p/test-plans repository

richard-ramos:
image

PR #1485, opened 2025-06-26
feat(gossipsub): Add support for skipping message insertion into the message cache (Mix protocol integration)

AkshayaMani:
This PR implements support in GossipSub for protocols like the [Mix protocol](https://rfc.vac.dev/vac/raw/mix), which require skipping message insertion into the message cache (mcache)

### Summary of changes
- Introduces a new `PublishParams` object, which bundles optional parameters (_e.g.,_ `useCustomConn`) for the `GossipSub.publish` function.
- Adds a `skipMCache` flag (defaults to false) to the `PublishParams` object.
- When `skipMCache` is true, the published message is not added to the mcache.
- This prevents the message from being advertised via `IHAVE` control messages.

### Motivation
This is part of ongoing integration work for the Mix protocol — a custom, stateless libp2p protocol designed for anonymous message routing in peer-to-peer networks.
The `skipMCache` flag is designed to be generic and may be reused by other protocols that require fine-grained control over whether messages are advertised via `IHAVE`. This is particularly useful in protocols aiming to preserve anonymity.

### Notes
- All changes are backward-compatible; default `publish` behavior is unaffected.
- A new integration test verifies that messages published with `skipMCache = true` do not appear in the local message cache.
- Code has been formatted using `nph`.

### Related
- VAC RAW RFC: [Mix Protocol](https://rfc.vac.dev/vac/raw/mix)
- Follow-up to: #1420

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -775,7 +775,10 @@ proc makePeersForPublishDefault(
   return peers
 
 method publish*(
-    g: GossipSub, topic: string, data: seq[byte], useCustomConn: bool = false
+    g: GossipSub,
+    topic: string,
+    data: seq[byte],
+    publishParams: Option[PublishParams] = none(PublishParams),
 ): Future[int] {.async: (raises: []).} =
   logScope:
     topic
@@ -789,8 +792,12 @@ method publish*(
 
   trace "Publishing message on topic", data = data.shortLog
 
+  var pubParams = publishParams.get(PublishParams())
+  if publishParams.isSome:
+    pubParams = publishParams.get()
+
   let peers =
-    if useCustomConn:
+    if pubParams.useCustomConn:
       g.makePeersForPublishUsingCustomConn(topic)
     else:
       g.makePeersForPublishDefault(topic, data)
@@ -828,7 +835,8 @@ method publish*(
     trace "Dropping already-seen message"
     return 0
 
-  g.mcache.put(msgId, msg)
+  if not pubParams.skipMCache:
+    g.mcache.put(msgId, msg)
 
   if g.parameters.sendIDontWantOnPublish and isLargeMessage(msg, msgId):
     g.sendIDontWant(msg, msgId, peers)
@@ -837,7 +845,7 @@ method publish*(
     peers,
     RPCMsg(messages: @[msg]),
     isHighPriority = true,
-    useCustomConn = useCustomConn,
+    useCustomConn = pubParams.useCustomConn,
   )
 
   if g.knownTopics.contains(topic):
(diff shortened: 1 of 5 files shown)


tests/pubsub/integration/testgossipsubskipmcache.nim
```diff
@@ -0,0 +1,66 @@
...
+
+    nodes[1].subscribe(topic, handler)
+    await waitSub(nodes[0], nodes[1], topic)
+
+    let publishData = "hello".toBytes()
+
+    let msgId = GossipSub(nodes[0]).getMsgId(topic, publishData)
+
+    tryPublish await nodes[0].publish(
+      topic, publishData, publishParams = some(PublishParams(useCustomConn: true))
```
vladopajic: should it be `skipMCache: true` ?

vladopajic (changes requested):

tests/pubsub/integration/testgossipsubskipmcache.nim
```diff
@@ -0,0 +1,66 @@
...
+
+suite "GossipSub Integration - Skip MCache Support":
+  teardown:
+    checkTrackers()
+
+  asyncTest "publish with skipMCache prevents message from being added to mcache":
+    let
+      topic = "test"
+      handler = proc(topic: string, data: seq[byte]) {.async.} =
+        discard
```
rlve: You could reuse `voidTopicHandler` from `tests/pubsub/utils` instead

tests/pubsub/integration/testgossipsubskipmcache.nim
```diff
@@ -0,0 +1,66 @@
...
+      nodes = generateNodes(2, gossip = true)
```
rlve:
```suggestion
      nodes = generateNodes(2, gossip = true).toGossipSub()
```

And then no need to convert the node to `GossipSub` in lines 38 and 44.

tests/pubsub/integration/testgossipsubskipmcache.nim
AkshayaMani: Thanks, updated to use `voidTopicHandler`.

tests/pubsub/integration/testgossipsubskipmcache.nim
AkshayaMani: Good catch — thanks! Fixed it.

tests/pubsub/integration/testgossipsubskipmcache.nim
AkshayaMani: Done, updated to convert nodes directly to `GossipSub `.

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -789,8 +792,12 @@ method publish*(
 
   trace "Publishing message on topic", data = data.shortLog
 
+  var pubParams = publishParams.get(PublishParams())
+  if publishParams.isSome:
+    pubParams = publishParams.get()
```
richard-ramos: Arent these 3 lines equivalent to:
```
let pubParams = publishParams.get(PublishParams())
```

AkshayaMani: Yes, and much cleaner — thanks!

richard-ramos (approved):

vladopajic (approved):

PR #1486, opened 2025-06-26
chore(autotls): make autotls manager into service

gmelodie:


vladopajic (approved):

libp2p/autotls/acme/client.nim
```diff
@@ -20,52 +20,65 @@ export api
...
+    key: Opt[KeyPair] = Opt.none(KeyPair),
+    kid: Kid = Kid(""),
+): T {.raises: [].} =
   let key = key.valueOr:
     KeyPair.random(PKScheme.RSA, rng[]).get()
-  let registerResponse = await api.requestRegister(key)
-  T(api: api, key: key, kid: registerResponse.kid)
+  T(api: api, key: key, kid: kid)
+
+proc getKid*(
```
richard-ramos: The name of this proc is not semantically correct as it does more than returning `self.kid`. Maybe something like `getOrInitKid` or anything that indicates that there's a conditional fetching happening if `kid` aint available?

libp2p/autotls/acme/client.nim
```diff
@@ -20,52 +20,65 @@ export api
...
 ): Future[ACMECertificateResponse] {.async: (raises: [ACMEError, CancelledError]).} =
   let chalURL = parseUri(challenge.dns01.url)
   let orderURL = parseUri(challenge.order)
   let finalizeURL = parseUri(challenge.finalize)
-  discard await self.api.sendChallengeCompleted(chalURL, self.key, self.kid)
+  debug "sending challenge completed notification"
+  discard await self.api.sendChallengeCompleted(chalURL, self.key, await self.getKid())
 
-  let completed = await self.api.checkChallengeCompleted(chalURL, self.key, self.kid)
+  debug "checking for completed challenge"
```
richard-ramos: Should these lines be a trace or debug? because they're just narrating the steps in the process but no debug information per se is being logged. (same in line 74 / 81)

I think a good criteria for knowing when to use debug/trace could be:
-    trace = "i am doing a thing", basically every breath the code took
-    debug = "here's what happened when i did the thing", or what devs care about when diagnosing

libp2p/autotls/utils.nim
```diff
@@ -53,19 +53,10 @@ proc isPublic*(ip: IpAddress): bool {.raises: [AutoTLSError].} =
...
-    return ip
-  except AutoTLSError as exc:
-    raise exc
-  except CatchableError as exc:
-    raise newException(
-      AutoTLSError, "Unexpected error while getting primary IP address for host", exc
-    )
+  let ip = checkedGetPrimaryIPAddr()
+  assert ip.isIPv4(), "Host does not have an IPv4 address"
+  assert ip.isPublic(), "Host does not have a public IPv4 address"
```
richard-ramos: I'm not sure of these assertions, as these could be runtime errors.
It could happen for a device to move to diff network and end up then in one in which the assigned IP is not public. Then in that case we'd be crashing the whole app.

libp2p/builders.nim
```diff
@@ -315,10 +314,13 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
     ms = MultistreamSelect.new()
     muxedUpgrade = MuxedUpgrade.new(b.muxers, secureManagerInstances, ms)
 
+  if not b.autotls.isNil():
+    b.services.insert(b.autotls, 0)
```
richard-ramos, line 318: Why insert it into 0 pos? (curious because the wilcard service is inserted there too and that's what's used to resolve 0.0.0.0 IPs)

tests/testautotls_integration.nim
```diff
@@ -32,105 +32,120 @@ import ./helpers
...
+        await acme.close()
+
+      let challenge = await acme.getChallenge(@["some.dummy.domain.com"])
+
+      check challenge.finalize.len > 0
+      check challenge.order.len > 0
+      check challenge.dns01.url.len > 0
+      check challenge.dns01.`type` == ACMEChallengeType.DNS01
+      check challenge.dns01.status == ACMEChallengeStatus.PENDING
+      check challenge.dns01.token.len > 0
```
richard-ramos: you can also do something like this to avoid repeating `check`

```suggestion
      check:
        challenge.finalize.len > 0
        challenge.order.len > 0
        challenge.dns01.url.len > 0
        challenge.dns01.`type` == ACMEChallengeType.DNS01
        challenge.dns01.status == ACMEChallengeStatus.PENDING
        challenge.dns01.token.len > 0
```

libp2p/builders.nim
gmelodie: It's just inserting into the very beginning of the sequence. It could be in the end too, it shouldn't make any difference.

PR #1488, opened 2025-06-26
chore(yamux): improve performance with zero allocation queue

vladopajic:
this pr improves yamux performance by 20-40% as observed when running perf test. 

previously data was constantly appended to `recvQueue` which caused memory trashing due to resizing. 
change from this pr introduced `ZeroQueue` that will allow data to be appended and consumed with zero allocations.

### DIFF
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -12,6 +12,7 @@
 import sequtils, std/[tables]
 import chronos, chronicles, metrics, stew/[endians2, byteutils, objects]
 import ../muxer, ../../stream/connection
+import ../../utils/zeroqueue
 
 export muxer
 
@@ -151,7 +152,7 @@ type
     opened: bool
     isSending: bool
     sendQueue: seq[ToSend]
-    recvQueue: seq[byte]
+    recvQueue: ZeroQueue
     isReset: bool
     remoteReset: bool
     closedRemotely: AsyncEvent
@@ -229,7 +230,6 @@ proc reset(channel: YamuxChannel, isLocal: bool = false) {.async: (raises: []).}
   for (d, s, fut) in channel.sendQueue:
     fut.fail(newLPStreamEOFError())
   channel.sendQueue = @[]
-  channel.recvQueue = @[]
   channel.sendWindow = 0
   if not channel.closedLocally:
     if isLocal and not channel.isSending:
@@ -252,7 +252,7 @@ proc updateRecvWindow(
   ##
   # In order to avoid spamming a window update everytime a byte is read,
   # we send it everytime half of the maxRecvWindow is read.
-  let inWindow = channel.recvWindow + channel.recvQueue.len
+  let inWindow = channel.recvWindow + channel.recvQueue.len()
   if inWindow > channel.maxRecvWindow div 2:
     return
 
@@ -279,7 +279,7 @@ method readOnce*(
         newLPStreamConnDownError()
   if channel.isEof:
     raise newLPStreamRemoteClosedError()
-  if channel.recvQueue.len == 0:
+  if channel.recvQueue.isEmpty():
     channel.receivedData.clear()
     let
       closedRemotelyFut = channel.closedRemotely.wait()
@@ -290,31 +290,26 @@ method readOnce*(
       if not receivedDataFut.finished():
         await receivedDataFut.cancelAndWait()
     await closedRemotelyFut or receivedDataFut
-    if channel.closedRemotely.isSet() and channel.recvQueue.len == 0:
+    if channel.closedRemotely.isSet() and channel.recvQueue.isEmpty():
       channel.isEof = true
       return
         0 # we return 0 to indicate that the channel is closed for reading from now on
 
-  let toRead = min(channel.recvQueue.len, nbytes)
-
-  var p = cast[ptr UncheckedArray[byte]](pbytes)
-  toOpenArray(p, 0, nbytes - 1)[0 ..< toRead] =
-    channel.recvQueue.toOpenArray(0, toRead - 1)
-  channel.recvQueue = channel.recvQueue[toRead ..^ 1]
+  let consumed = channel.recvQueue.consumeTo(pbytes, nbytes)
 
   # We made some room in the recv buffer let the peer know
   await channel.updateRecvWindow()
   channel.activity = true
-  return toRead
+  return consumed
 
 proc gotDataFromRemote(
     channel: YamuxChannel, b: seq[byte]
 ) {.async: (raises: [CancelledError, LPStreamError]).} =
   channel.recvWindow -= b.len
-  channel.recvQueue = channel.recvQueue.concat(b)
+  channel.recvQueue.push(b)
   channel.receivedData.fire()
   when defined(libp2p_yamux_metrics):
-    libp2p_yamux_recv_queue.observe(channel.recvQueue.len.int64)
+    libp2p_yamux_recv_queue.observe(channel.recvQueue.len().int64)
   await channel.updateRecvWindow()
 
 proc setMaxRecvWindow*(channel: YamuxChannel, maxRecvWindow: int) =
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -252,7 +252,7 @@ proc updateRecvWindow(
   ##
   # In order to avoid spamming a window update everytime a byte is read,
   # we send it everytime half of the maxRecvWindow is read.
-  let inWindow = channel.recvWindow + channel.recvQueue.len()
+  let inWindow = channel.recvWindow + channel.recvQueue.len
   if inWindow > channel.maxRecvWindow div 2:
     return
 
@@ -309,7 +309,7 @@ proc gotDataFromRemote(
   channel.recvQueue.push(b)
   channel.receivedData.fire()
   when defined(libp2p_yamux_metrics):
-    libp2p_yamux_recv_queue.observe(channel.recvQueue.len().int64)
+    libp2p_yamux_recv_queue.observe(channel.recvQueue.len.int64)
   await channel.updateRecvWindow()
 
 proc setMaxRecvWindow*(channel: YamuxChannel, maxRecvWindow: int) =
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -257,7 +257,7 @@ proc updateRecvWindow(
     return
 
   let delta = channel.maxRecvWindow - inWindow
-  channel.recvWindow.inc(delta)
+  channel.recvWindow.inc(delta.int)
   await channel.conn.write(YamuxHeader.windowUpdate(channel.id, delta.uint32))
   trace "increasing the recvWindow", delta
 
--- a/libp2p/utils/zeroqueue.nim
+++ b/libp2p/utils/zeroqueue.nim
@@ -1,4 +1,16 @@
+# Nim-Libp2p
+# Copyright (c) 2025 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
 type ZeroQueue* = object
+  # ZeroQueue is queue structure optimized for efficient pushing and popping of 
+  # byte sequences `seq[byte]`. This type is useful for streaming or buffering 
+  # scenarios where chunks of binary data are accumulated and consumed incrementally.
   data: seq[seq[byte]] = @[]
 
 proc clear*(q: var ZeroQueue) =
@@ -22,13 +34,18 @@ proc pop*(q: var ZeroQueue, count: int): seq[byte] =
     return @[]
 
   let first = q.data[0]
+
+  # first frame has up to requested count elements,
+  # queue will return this frame (frame might have less then requested)
   if first.len <= count:
     q.data = q.data[1 ..^ 1]
     return first
-  else:
-    let ret = first[0 ..< count]
-    q.data[0] = first[count ..^ 1]
-    return ret
+
+  # first frame has more elements then requested count, 
+  # queue will return view of first count elements, leaving the rest in the queue
+  let ret = first[0 ..< count]
+  q.data[0] = first[count ..^ 1]
+  return ret
 
 proc consumeTo*(q: var ZeroQueue, pbytes: pointer, nbytes: int): int =
   var consumed = 0
--- a/libp2p/utils/zeroqueue.nim
+++ b/libp2p/utils/zeroqueue.nim
@@ -1,0 +1,41 @@
+type ZeroQueue* = object
+  data: seq[seq[byte]] = @[]
+
+proc clear*(q: var ZeroQueue) =
+  q.data = @[]
+
+proc isEmpty*(q: ZeroQueue): bool =
+  return q.data.len == 0
+
+proc len*(q: ZeroQueue): int64 =
+  var l: int64
+  for b in q.data:
+    l += b.len
+  return l
+
+proc push*(q: var ZeroQueue, b: seq[byte]) =
+  if b.len > 0:
+    q.data.add(b)
+
+proc pop*(q: var ZeroQueue, count: int): seq[byte] =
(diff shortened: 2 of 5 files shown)


libp2p/utils/zeroqueue.nim
```diff
@@ -0,0 +1,58 @@
...
+proc pop*(q: var ZeroQueue, count: int): seq[byte] =
+  if q.data.len == 0 or count == 0:
+    return @[]
+
+  let first = q.data[0]
+
+  # first frame has up to requested count elements,
+  # queue will return this frame (frame might have less then requested)
+  if first.len <= count:
+    q.data = q.data[1 ..^ 1]
```
arnetheduck: fyi, this will copy all the little queues in the big queue - `[]` allocates a new `seq`, then copies the range one by one and so on - ditto all the other places where it gets used, ie `push` will copy `b`.

In general, it just so happens that https://github.com/status-im/nim-faststreams/pull/71 was pushed just now, exposing the queue from faststreams in a semi-consumable way - it does something similar that you want to do here, but with fewer allocations and copies all around. for example the trick to not make massive copies is to reuse seqs and use a `deque` instead of a seq to manage the forward motion.

vladopajic: > fyi, this will copy all the little queues in the big queue - [] allocates a new seq, then copies the range one by one and so on - ditto all the other places where it gets used, ie push will copy b.

if i understand it correctly there is some trashing when `q.data` is changed with push and pop. but 2nd dimension elements are reused, right?

arnetheduck: no reuse whatsoever. `[]` always creates a full deep copy of everything but the zero-eth element that you're discarding. if you had 1000 seqs in q.data, and each had 1000 bytes, this will allocate 999000 bytes (plus rounding) and then throw them away after the assignment.

With a `seq`, you can never remove anything from the front without _at least_ moving data around - it's the equivalent of a C++ `vector`. Sometimes you can avoid copying the "inner" elements, but the way you've written it, they are really allocated anew.

[`q.data.delete(0)`](https://nim-lang.org/docs/system.html#delete,seq[T],Natural) would move the elements in the seq. ie in the above example, we would not allocate any additional data, but we'd still have to perform 999 moves, no copies and no allocations. With a [`Deque[seq[byte]]`](https://nim-lang.org/docs/deques.html) you don't perform any moves or copies apart from that first item you're popping out of there.

arnetheduck: fwiw, `PageBuffers` in faststreams were created for the same problem that you're working on here. I'd recommend taking a look and maybe trying to improve PB so we can reuse the same general buffer for multiple projects. I think libp2p already depends on faststreams - or it will when using `nim-protobuf-serialization` which uses faststreams under the hood.

arnetheduck: (and as a second note, we've outlawed the use of `[a..b]` in everything but tests - if you want a read-only view, `q.data.toOpenArray(a, b)` is the way but that's not what you want to use here (since using it would imply a copy instead of a move)

libp2p/utils/zeroqueue.nim
```diff
@@ -0,0 +1,76 @@
...
+
+proc len*(q: ZeroQueue): int64 =
+  var l: int64
+  for b in q.chunks:
+    l += b.len()
+  return l
+
+proc push*(q: var ZeroQueue, b: seq[byte]) =
+  if b.len > 0:
+    q.chunks.add(Chunk(address: addr b[0], size: b.len, start: 0))
```
arnetheduck: this idea unfortunately will lead to a crash or invalid data. if `b` is garbage-collected before the chunk is removed, the program will crash (or write some random data). 

Generally, `ptr` is used when the lifetime of the pointee is managed by "something", such as a `ref`, and that something outlives the `ptr` itself.

vladopajic: yes, could be fixed by adding `ref` to the `Chunk`  👽

libp2p/utils/zeroqueue.nim
vladopajic: got it

i'll have a look at `PageBuffers` tomorrow, i overdid my screen time for today.

on the side note, i  don't really like this approach although it does improve performance here. each layer will introduces performance loss where data has to be copied. it would be much better if ownership of data (starting from origin) is transferred up the layers.  iam trying to say it still doesn't feel like it's right way to do..

arnetheduck: indeed, this has long been an issue in nim - there is some work done to improve the situation where the "last use" of a seq automatically makes it eligible for moving instead of copying the data - this would be useful in a buffer like this, though it needs an additional `sink` annotation - see https://nim-lang.org/docs/destructors.html#sink-parameters

libp2p/muxers/yamux/yamux.nim
```diff
@@ -252,12 +252,12 @@ proc updateRecvWindow(
   ##
   # In order to avoid spamming a window update everytime a byte is read,
   # we send it everytime half of the maxRecvWindow is read.
-  let inWindow = channel.recvWindow + channel.recvQueue.len
+  let inWindow = channel.recvWindow + channel.recvQueue.len()
```
gmelodie: nit: maybe we should adopt the same style as other teams here?
verbs: put `()`
nouns (like `len`): no parenthesis
https://github.com/vacp2p/nim-libp2p/pull/1472#discussion_r2162507398

libp2p/muxers/yamux/yamux.nim
```diff
@@ -290,31 +290,26 @@ method readOnce*(
...
 proc gotDataFromRemote(
     channel: YamuxChannel, b: seq[byte]
 ) {.async: (raises: [CancelledError, LPStreamError]).} =
   channel.recvWindow -= b.len
-  channel.recvQueue = channel.recvQueue.concat(b)
+  channel.recvQueue.push(b)
   channel.receivedData.fire()
   when defined(libp2p_yamux_metrics):
-    libp2p_yamux_recv_queue.observe(channel.recvQueue.len.int64)
+    libp2p_yamux_recv_queue.observe(channel.recvQueue.len().int64)
```
gmelodie: as above

libp2p/utils/zeroqueue.nim
```diff
@@ -0,0 +1,81 @@
...
+
+proc len*(q: ZeroQueue): int64 =
+  var l: int64
+  for b in q.chunks.items():
+    l += b.len()
+  return l
+
+proc push*(q: var ZeroQueue, b: sink seq[byte]) =
+  if b.len > 0:
+    q.chunks.addLast(newChunk(b))
```
gmelodie, line 46: What if `b.len == 0`?

tests/testzeroqueue.nim
```diff
@@ -0,0 +1,86 @@
...
+# those terms.
+
+import unittest2
+import ../libp2p/utils/zeroqueue
+
+proc toSeq(p: pointer, length: int): seq[byte] =
+  let b = cast[ptr UncheckedArray[byte]](p)
+  result = newSeq[byte](length)
+  for i in 0 ..< length:
+    result[i] = b[i]
```
gmelodie: Maybe this instead? You have to guarantee that `p` has valid data in it though
```suggestion
result = newSeq[byte](length)
copyMem(result[0].addr, p, length)
```

tests/testzeroqueue.nim
```diff
@@ -0,0 +1,86 @@
...
+
+suite "ZeroQueue":
+  test "push-pop":
+    var q: ZeroQueue
+    check q.len() == 0
+    check q.isEmpty()
```
gmelodie, line 25: nit
```suggestion
    check q.len() == 0 and q.isEmpty()
```

tests/testzeroqueue.nim
```diff
@@ -0,0 +1,86 @@
...
+    check q.popChunkSeq(1).len == 0 # pop empty seq when queue is empty
+
+    q.push(@[1'u8, 2, 3])
+    q.push(@[4'u8, 5])
+    check q.len() == 5
+    check not q.isEmpty()
```
gmelodie, line 31: nit
```suggestion
    check  not q.isEmpty() and q.len() == 5
```

tests/testzeroqueue.nim
```diff
@@ -0,0 +1,86 @@
...
+    defer:
+      dealloc(pbytes)
+
+    # consumeTo should fill up to nbytes (queue is emptied)
+    q.push(@[1'u8, 2, 3])
+    q.push(@[4'u8, 5])
+    q.push(@[6'u8, 7])
+    check q.consumeTo(pbytes, nbytes) == 7
+    check toSeq(pbytes, 7) == @[1'u8, 2, 3, 4, 5, 6, 7]
+    check q.isEmpty()
```
gmelodie: this could be a separate test

tests/testzeroqueue.nim
```diff
@@ -0,0 +1,86 @@
...
+
+    # consumeTo should fill only 1 element, leaving 2 elements in the queue
+    q.push(@[1'u8, 2, 3])
+    check q.consumeTo(pbytes, 1) == 1
+    check toSeq(pbytes, 1) == @[1'u8]
+    check q.len() == 2
+    # assert remaning elements
+    check q.consumeTo(pbytes, nbytes) == 2
+    check toSeq(pbytes, 2) == @[2'u8, 3]
+    check q.isEmpty()
```
gmelodie: this could be a separate test

tests/testzeroqueue.nim
```diff
@@ -0,0 +1,86 @@
...
+    q.clear()
+    q.push(@[4'u8, 5])
+    q.push(@[1'u8, 2, 3])
+    check q.consumeTo(pbytes, 3) == 3
+    check toSeq(pbytes, 3) == @[4'u8, 5, 1]
+    check q.len() == 2
+    # assert remaning elements
+    check q.consumeTo(pbytes, nbytes) == 2
+    check toSeq(pbytes, 2) == @[2'u8, 3]
+    check q.isEmpty()
```
gmelodie: this could be a separate test

gmelodie (approved):

tests/testzeroqueue.nim
vladopajic: not in favor of this, if check fails you will not know which statements where false

tests/testzeroqueue.nim
vladopajic: not in favor of this, if check fails you will not know which statements where false

libp2p/utils/zeroqueue.nim
vladopajic: there is no need in appending sequence with zero size

tests/testzeroqueue.nim
vladopajic: could be, but there is no need. this tests tests everything `consumeTo` related

tests/testzeroqueue.nim
vladopajic: could be, but there is no need. this tests tests everything `consumeTo` related

tests/testzeroqueue.nim
vladopajic: could be, but there is no need. this tests tests everything `consumeTo` related

libp2p/utils/zeroqueue.nim
```diff
@@ -0,0 +1,81 @@
...
+  c.size - c.start
+
+type ZeroQueue* = object
+  # ZeroQueue is queue structure optimized for efficient pushing and popping of 
+  # byte sequences `seq[byte]` (called chunks). This type is useful for streaming or buffering 
+  # scenarios where chunks of binary data are accumulated and consumed incrementally.
+  chunks: Deque[Chunk]
+
+proc clear*(q: var ZeroQueue) =
+  q.chunks.clear()
```
richard-ramos, line 33: Maybe to keep consistency with other procs?
```suggestion
  return q.chunks.clear()
```

libp2p/utils/zeroqueue.nim
```diff
@@ -0,0 +1,81 @@
...
+
+proc isEmpty*(q: ZeroQueue): bool =
```
richard-ramos, line 35: Can probably be a template so this code gets inlined

libp2p/utils/zeroqueue.nim
```diff
@@ -0,0 +1,81 @@
...
+  ret.size = ret.start + count
+  first.start += count
+  q.chunks.addFirst(first)
+  return ret
+
+proc consumeTo*(q: var ZeroQueue, pbytes: pointer, nbytes: int): int =
+  var consumed = 0
+  while consumed < nbytes and not q.isEmpty():
+    let chunk = q.popChunk(nbytes - consumed)
+    let dest = cast[pointer](cast[ByteAddress](pbytes) + consumed)
```
richard-ramos, line 68: Byteaddress are deprecated, gotta use `uint` instead:
https://nim-lang.github.io/Nim/ctypes.html#ByteAddress

tests/testzeroqueue.nim
```diff
@@ -0,0 +1,85 @@
...
+
+# Nim-Libp2p
+# Copyright (c) 2025 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
```
richard-ramos, line 11: Edge cases:
  1. A large chunk is partially consumed multiple times
  2. Consuming from a queue with multiple partially consumed chunks (i.e. push, partially consume, push, partially consume) to make sure that doing a consume, next push is not messed up
  3. Empty pushes are being tested in isolation, what about pushing something, partial consume, empty push, and consume.
  4. Behavior of consume on an empty queue

libp2p/utils/zeroqueue.nim
```diff
@@ -0,0 +1,81 @@
...
+  var l: int64
+  for b in q.chunks.items():
+    l += b.len()
+  return l
+
+proc push*(q: var ZeroQueue, b: sink seq[byte]) =
+  if b.len > 0:
+    q.chunks.addLast(newChunk(b))
+
+proc popChunk(q: var ZeroQueue, count: int): Chunk {.inline.} =
```
richard-ramos, line 48: Should `Chunk` be a ref object? because otherwise popFirst/addFirst would copy, right?

richard-ramos (approved):

libp2p/utils/zeroqueue.nim
tersec: As https://status-im.github.io/nim-style-guide/language.result.html notes,
> "expression" returns guarantee that all code branches produce one (and only one) value to be returned

this is something `return foo` doesn't do, so all else equal, expression return is slightly nicer.

libp2p/utils/zeroqueue.nim
tersec: Nim can often use RVO/NRVO in such situations, but it depends.

tests/testzeroqueue.nim
```diff
@@ -0,0 +1,85 @@
...
+import unittest2
+import ../libp2p/utils/zeroqueue
+
+proc toSeq(p: pointer, length: int): seq[byte] =
+  let b = cast[ptr UncheckedArray[byte]](p)
+  result = newSeq[byte](length)
```
tersec: This unnecessarily zero-initializes `result`.

libp2p/utils/zeroqueue.nim
vladopajic: @richard-ramos clear does not return any type?

libp2p/utils/zeroqueue.nim
vladopajic: there is no probably no benefit on inlining this, its not use that often

libp2p/utils/zeroqueue.nim
vladopajic: getting error in that case:
```
Error: type mismatch
Expression: cast[uint](pbytes) + consumed
```

libp2p/utils/zeroqueue.nim
vladopajic: > Should Chunk be a ref object? 

non ref type would copy, but it would be sallow copy, which is negotiable in this case since memory footprint of this type is small (8 bytes). and total number of copies should be vary small, as these operations are not really that much frequent.

however i have made `Chunk` ref object - it goes more in favor for "zero allocation".

libp2p/utils/zeroqueue.nim
richard-ramos: :man_facepalming:  oops

PR #1497, opened 2025-07-01
refactor(noise): reduce memory usage

vladopajic:
- reducing memory usage in `handshakeXXOutbound` and `handshakeXXInbound` methods
  - removed usage of `StreamSeq` as it copies memory with `add`, `assign` ...
  - instead copying seq[bytes] to single container bytes are directly transfers to message frame as parts
- `sendHSMessage` now sends message frame using variable parts that concatenated represent full message 
  - unnecessary allocation in `writeFrame` is avoided 
 - as result this pr reduces total memory usage when sending frame (message) from close to 3x of message size to close to 1x of message size; consequently reducing unnecessary memory trashing


libp2p/protocols/secure/noise.nim
```diff
@@ -316,28 +319,25 @@ proc readFrame(
...
-): Future[void] {.async: (raises: [CancelledError, LPStreamError], raw: true).} =
-  doAssert buf.len <= uint16.high.int
-  var
-    lesize = buf.len.uint16
-    besize = lesize.toBytesBE
-    outbuf = newSeqOfCap[byte](besize.len + buf.len)
-  trace "writeFrame", sconn, size = lesize, data = shortLog(buf)
-  outbuf &= besize
-  outbuf &= buf
-  sconn.write(outbuf)
```
vladopajic, line 330: this allocation of `outbuf` is avoided; this code required 2x of memory to send message + time for copying data

vladopajic:
> How did you measure this?

i haven't measured, that's observation from analysing the code

PR #1499, opened 2025-07-02
fix(ci): move testintegration to new step

gmelodie:
CI run: https://github.com/vacp2p/nim-libp2p/actions/runs/16032349564/job/45235784271

### DIFF
--- a/.github/workflows/daily_common.yml
+++ b/.github/workflows/daily_common.yml
@@ -85,7 +85,7 @@ jobs:
           nimble install_pinned
 
       - name: Install dependencies (latest)
-        if: ${{ inputs.pinned_deps != 'true' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/daily_common.yml
```diff
@@ -96,4 +96,3 @@ jobs:
 
           export NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }}"
           nimble test
-          nimble testintegration
```
vladopajic: instead of copying 99% the same workflow for integration.yml, could additional input be added that will determine if `nimble test` should be run, or `nimble testintegration` ?

vladopajic (approved):

richard-ramos (approved):

PR #1501, opened 2025-07-03
test(gossipsub): behavior unit tests 1

rlve:
Changes:
- add unit tests for `GossipSub` `behaviour`:
  - `"handleIHave - do not handle IHave when message already in cache"`
  - `"handleIWant - message not handled without sent IHave"`
  - `"handleIWant - message not handled when not in cache"`
  - `"handleIWant - stops processing after 21 invalid requests"`
  - `"handlePrune - peer is pruned and backoff is set"`
  - `"handleGraft - peer joins mesh for subscribed topic"`
  - `"handleGraft - do not graft if peer already in mesh"`
  - `"handleGraft - reject peer when mesh at dHigh threshold"` 
  - `"handleGraft - accept outbound peer when mesh at dHigh but below dOut threshold"`
  - `"handleGraft - reject outbound peer when mesh at dHigh and dOut threshold met"`

### DIFF
--- a/.github/workflows/daily_common.yml
+++ b/.github/workflows/daily_common.yml
@@ -96,4 +96,12 @@ jobs:
 
           export NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }}"
           nimble test
+
+      - name: Run integration tests
+        if: ${{ matrix.platform.os == 'linux' && matrix.cpu == 'amd64' }}
+        run: |
+          nim --version
+          nimble --version
+
+          export NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }}"
           nimble testintegration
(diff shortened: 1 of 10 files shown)


vladopajic (approved):

tests/pubsub/testbehavior.nim
```diff
@@ -152,6 +158,88 @@ suite "GossipSub Behavior":
...
+
+    # And an IWANT message 
+    let msg = ControlIWant(messageIDs: @[id])
+
+    # When IWANT is handled
+    let messages = gossipSub.handleIWant(peer, @[msg])
+
+    # Then IWant is ignored
+    check:
+      messages.len == 0
```
gmelodie, line 183: question: why does this mean that iwant was ignored and not handled? I assume that after iwant is handled the messages queue would be empty again, no?

gmelodie (approved):

tests/pubsub/testbehavior.nim
rlve: `handleIWant` processes `IWANT` requests from peers and returns the actual messages they're asking for. It takes a list of message IDs and returns a sequence of Message objects.

The function only adds messages to the result if both conditions are met:
We previously sent an `IHAVE` for that message ID to the peer and the message still exists in the cache.

In this test: `sentIHaves` doesn't contain the message id and therefore the message is not returned.

PR #1504, opened 2025-07-04
refactor(yamux): memory usage improvements

vladopajic:


tests/testrelayv2.nim
```diff
@@ -224,6 +224,9 @@ suite "Circuit Relay V2":
         await allFutures(src.stop(), dst.stop(), rel.stop())
 
       asyncTest "Connection duration exceeded":
+        if useYamux:
+          return
```
vladopajic, line 228: this test case fails when yamux header is not sent together with message in the same bytes seq.

tests/testrelayv2.nim
```diff
@@ -275,6 +275,9 @@ suite "Circuit Relay V2":
         await allFutures(src.stop(), dst.stop(), rel.stop())
 
       asyncTest "Connection data exceeded":
+        if useYamux:
+          return
```
vladopajic, line 279: this test case fails when yamux header is not sent together with message in the same bytes seq.

libp2p/muxers/yamux/yamux.nim
```diff
@@ -88,20 +88,21 @@ proc `$`(header: YamuxHeader): string =
     ) & "}, " & "streamId: " & $header.streamId & ", " & "length: " & $header.length &
     "}"
 
-proc encode(header: YamuxHeader): array[12, byte] =
-  result[0] = header.version
-  result[1] = uint8(header.msgType)
-  result[2 .. 3] = toBytesBE(uint16(cast[uint8](header.flags)))
+proc encode(header: YamuxHeader): seq[byte] =
```
arnetheduck, line 91: not sure what this is about, but an array is stack-allocated and fixed-size while a seq is heap-allocated and variable-size - ie all in all, this is a 10x memory usage regression + associated allocation and gc costs resulting in 100x cpu overhead at least ;) 

the c/assembly code inspector and the profiler is your friend here..

vladopajic: the header was converted to seq later on, when sending, so i thought better do it right away

PR #1507, opened 2025-07-06
chore: removing unused type StreamSeq

vladopajic:


arnetheduck:
something to keep in mind about the streamseq: it's sometimes used to collect many smaller writes into a single larger write - ie it's _sometimes_ faster to copy the data to a single larger write than to perform many smaller writes with an await in between - it's probably worth checking the code for any "small-writes" phenomena and buffer those.

vladopajic:
> something to keep in mind about the streamseq: it's sometimes used to collect many smaller writes into a single larger write - ie it's _sometimes_ faster to copy the data to a single larger write than to perform many smaller writes with an await in between - it's probably worth checking the code for any "small-writes" phenomena and buffer those.

thanks for the heads up.

there could be potential performance gains in this case, but intuitively if message is small performance gains would likely be minimal. also in this case it would be worth thinking about throughput, and what happens in high throughput case. 

if it becomes a measurable issue down the line, we can revisit, but for now it may be more efficient to leave those performance gains.

richard-ramos (approved):

PR #1520, opened 2025-07-11
perf(noise): encryptWithAd improvement

vladopajic:
some performance improvement for `encryptWithAd`
- better allocation (exactly as needed)
- utilize `newSeqUninit`
- memory copied instead of adding with  `.add`

### DIFF
--- a/libp2p/protocols/secure/noise.nim
+++ b/libp2p/protocols/secure/noise.nim
@@ -132,7 +132,7 @@ proc hasKey(cs: CipherState): bool =
   cs.k != EmptyKey
 
 proc encrypt(
-    state: var CipherState, data: var openArray[byte], ad: openArray[byte]
+    state: var CipherState, data: openArray[byte], ad: openArray[byte]
 ): ChaChaPolyTag {.noinit, raises: [NoiseNonceMaxError].} =
   var nonce: ChaChaPolyNonce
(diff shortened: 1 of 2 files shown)


richard-ramos (approved):

libp2p/protocols/secure/noise.nim
```diff
@@ -146,15 +146,16 @@ proc encrypt(
...
 ): seq[byte] {.raises: [NoiseNonceMaxError].} =
-  result = newSeqOfCap[byte](data.len + sizeof(ChaChaPolyTag))
-  result.add(data)
+  let tag = encrypt(state, data, ad)
 
-  let tag = encrypt(state, result, ad)
-
-  result.add(tag)
+  var res = newSeqUninit[byte](data.len + tag.len)
+  copyMem(addr res[0], unsafeAddr data[0], data.len)
```
arnetheduck, line 152: something to be careful with is that when `data.len == 0`, `[0]` will raise a defect. not sure if we can end with that here in particular, but it's a simple mistake to make.

arnetheduck: also, if `add` does not do the right thing here (ie avoid zeroing and uses copyMem), ie if you can spot any actual differences in the generated assembly, that's a bug in `add` that should be reported upstream - there's no reason add would not do the exact same thing here.

PR #1521, opened 2025-07-11
toByte return direct seq

NikhilBisht2:


libp2p/crypto/ecnist.nim
```diff
@@ -396,20 +392,12 @@ proc toBytes*(seckey: EcPrivateKey, data: var openArray[byte]): EcResult[int] =
     p.finish()
     b.write(p)
     b.finish()
-    var blen = len(b)
-    if len(data) >= blen:
-      copyMem(addr data[0], addr b.buffer[0], blen)
-    # ok anyway, since it might have been a query...
-    ok(blen)
+    ok(b.buffer[0 ..< b.len])
```
vladopajic, line 395: `b.buffer[0 ..< b.len]` will create new copy of bytes, why is it needed? 
and other instances...

vladopajic (commented):
make sure that tests are passing:

`nim c -r ./tests/testecnist.nim` to run tests on this file, if they are passing
`nimble runtestnative` then run tests an all files

libp2p/crypto/ecnist.nim
NikhilBisht2: sorry , i thought to resize it, ill make the changes

vladopajic:
is this still being worked on? if no we should leave issue unassigned

RWSINC (approved):

Issue #1482, opened 2025-06-25
bug(quic): does not work with perf protocol


vladopajic:
test `Perf protocol::quic` fails with error

```
/home/vlado/Workspace/ift/nim-libp2p/libp2p/protocols/perf/client.nim(66) perf
/home/vlado/Workspace/ift/nim-libp2p/libp2p/stream/lpstream.nim(159) readExactly
/home/vlado/Workspace/ift/nim-libp2p/libp2p/transports/quictransport.nim(64) readOnce
```

underlying error is "stream is closed".

--- 

apparently perf client closes connection after upload part finishes (with intention to close for write). after this perf attempts reading from the same connection but it is closed.

PR #1524, opened 2025-07-11
fix(perf): add QUIC transport compatibility

MorganaFuture:
**What**

  Fixed QUIC transport compatibility with perf protocol.

  - Move connection close after read phase for TCP
  - Add timeout-based read for QUIC streams
  - Fix infinite read loop with proper EOF detection
  - Make QUIC calls conditional on support flag

  **Why**
  QUIC doesn't support half-close like TCP transports. Perf was closing connection
  after write but before read, breaking QUIC streams with "stream is closed" errors.
  Now both QUIC and TCP transports work correctly with perf protocol.
  
  **Links**
Issue [1482](https://github.com/vacp2p/nim-libp2p/issues/1482)

### DIFF
--- a/libp2p/protocols/perf/client.nim
+++ b/libp2p/protocols/perf/client.nim
@@ -12,6 +12,8 @@
 import chronos, chronicles, sequtils
 import stew/endians2
 import ./core, ../../stream/connection
+when defined(libp2p_quic_support):
+  import ../../transports/quictransport
 
 logScope:
   topics = "libp2p perf"
@@ -57,7 +59,13 @@ proc perf*(
       statsCopy.uploadBytes += toWrite.uint
       p.stats = statsCopy
 
-    await conn.close()
+    # Close connection after writing for TCP, but not for QUIC
+    when defined(libp2p_quic_support):
+      if not (conn of QuicStream):
+        await conn.close()
+      # For QUIC streams, don't close yet - let server manage lifecycle
+    else:
+      await conn.close()
 
     size = sizeToRead
 
@@ -71,6 +79,11 @@ proc perf*(
       statsCopy.duration = Moment.now() - start
       statsCopy.downloadBytes += toRead.uint
       p.stats = statsCopy
+
+    # Close QUIC connections after read phase
+    when defined(libp2p_quic_support):
+      if conn of QuicStream:
+        await conn.close()
   except CancelledError as e:
     raise e
   except LPStreamError as e:
--- a/libp2p/protocols/perf/server.nim
+++ b/libp2p/protocols/perf/server.nim
@@ -14,6 +14,8 @@
 import chronos, chronicles
 import stew/endians2
 import ./core, ../protocol, ../../stream/connection, ../../utility
+when defined(libp2p_quic_support):
+  import ../../transports/quictransport
 
 export chronicles, connection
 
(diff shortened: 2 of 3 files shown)


tests/testperf.nim
```diff
@@ -109,10 +112,15 @@ suite "Perf protocol":
...
-    let server = createSwitch(isServer = true, useQuic = true)
-    let client = createSwitch(useQuic = true)
-    await runTest(server, client)
+    when defined(libp2p_quic_support):
+      try:
+        let server = createSwitch(isServer = true, useQuic = true)
+        let client = createSwitch(useQuic = true)
+        await runTest(server, client)
+      except CatchableError as e:
+        raise e
```
vladopajic: since this `except` is not doing anything other then raising, it is not needed. code that raises naturally raise.

libp2p/protocols/perf/client.nim
```diff
@@ -57,7 +59,13 @@ proc perf*(
...
       p.stats = statsCopy
 
-    await conn.close()
+    # Close connection after writing for TCP, but not for QUIC
+    when defined(libp2p_quic_support):
+      if not (conn of QuicStream):
+        await conn.close()
+      # For QUIC streams, don't close yet - let server manage lifecycle
+    else:
+      await conn.close()
```
vladopajic, line 68: this would be nicer if could be written like this
```
 if not (conn of QuicStream):
        await conn.close()
```

without  `when defined(libp2p_quic_support):` because if it is quic conn then we can assume that quic is supported. and in other places...

libp2p/protocols/perf/server.nim
```diff
@@ -36,8 +38,32 @@ proc new*(T: typedesc[Perf]): T {.public.} =
...
       try:
-        while true:
-          bytesRead += await conn.readOnce(addr toReadBuffer[0], PerfSize)
+        # Different handling for QUIC vs TCP streams
+        when defined(libp2p_quic_support):
+          if conn of QuicStream:
+            # QUIC needs timeout-based approach to detect end of upload
+            while not conn.atEof:
+              let readFut = conn.readOnce(addr toReadBuffer[0], PerfSize)
+              if not await readFut.withTimeout(100.milliseconds):
```
vladopajic, line 47: if this is the case then we should probably fix that, as this can be vary flaky

libp2p/protocols/perf/client.nim
vladopajic: is `when defined(libp2p_quic_support)` necessary in these cases?

libp2p/protocols/perf/server.nim
vladopajic: we can merge pr with this timeout, and open issue to to track effort on fixing this.

libp2p/protocols/perf/client.nim
MorganaFuture: @vladopajic 
Thanks for your feedback! I considered simplifying the
```
 when
   defined(libp2p_quic_support)
```  
blocks, but they're actually necessary for compile-time safety.

Without the conditional compilation, the code would fail to compile when QUIC support is disabled because:
  1. `QuicStream` type wouldn't be declared
  2. `conn of QuicStream` would cause a compilation error

Alternative approaches like string-based type checking would be less efficient and less type-safe than the
current compile-time approach.

The `when defined` pattern is commonly used throughout nim-libp2p for optional features to ensure the code compiles correctly regardless of which transports are enabled/disabled.

libp2p/protocols/perf/server.nim
MorganaFuture: @vladopajic Agreed, the timeout approach is indeed flaky.
Let's merge this PR with the current timeout implementation to fix the immediate QUIC compatibility issue.
Could you open a separate issue to track implementing a more robust EOF detection mechanism for QUIC streams? Feel free to assign it to me - I'll research proper QUIC stream lifecycle management and explore better alternatives to the timeout-based approach.

tests/testperf.nim
MorganaFuture: Fixed

libp2p/protocols/perf/client.nim
vladopajic: thanks! we should also explore improving this. it would be much nicer code if we can write it without `when defined(libp2p_quic_support):` in examples like these. ideally library should only forbid to instantiate quic transport but types should be there for simplifying code

vladopajic: added issue https://github.com/vacp2p/nim-libp2p/issues/1529 for this

libp2p/protocols/perf/server.nim
vladopajic: added issue https://github.com/vacp2p/nim-libp2p/issues/1530 for this

gmelodie (approved):

Ben-PH:
@MorganaFuture I'm seeing this, is this what you're seeing? https://github.com/vacp2p/nim-libp2p/actions/runs/16267405031/job/45929192700?pr=1524

Could you please link to what you're seeing if different.

I might be seeing some flakey tests on my end, and I'm on the lookout for other occurrences that can help me see a pattern emerge.

Could you please describe the flakiness of it, and if there's anything on your end that might help understand what's making it flakey, that might help as well.

MorganaFuture:
@Ben-PH  I have fixed CI/CD issue in this [PR](https://github.com/vacp2p/nim-libp2p/pull/1541)

richard-ramos:
:rocket:

Issue #1522, opened 2025-07-11
create `toBytesLE` and `toBytesBE` utilites that fill already allocated array


vladopajic:
currently code uses  `toBytesLE` and `toBytesBE`  (as defied [here](https://github.com/status-im/nim-stew/blob/e5740014961438610d336cd81706582dbf2c96f0/stew/endians2.nim#L107C6-L107C15)) that return `array[sizeof(x), byte]`.

since all code that uses these utilities already have allocated sequence where bytes should be written it would be nice to create another set of utilities that will write directly to allocated buffer.

```nim
# instead of doing this
result[4 .. 7] = toBytesBE(header.streamId)

# should be:
toBytesBESeq(header.streamId, result, 4)
```

PR #1527, opened 2025-07-11
feat: create toBytesLE and toBytesBE

MorganaFuture:
**What**

  Adds new in-place byte conversion utilities that write directly to allocated
  buffers:
  - toBytesLE*(value: SomeEndianInt, dest: var openArray[byte], offset: int = 0)
  - toBytesBE*(value: SomeEndianInt, dest: var openArray[byte], offset: int = 0)

  Updates Yamux and Protobuf modules to use the new utilities.

  **Why**

  Current toBytesLE/toBytesBE functions create temporary arrays that are immediately
  copied to pre-allocated buffers. This causes unnecessary allocations and memory
  copies in hot paths like network protocol encoding.

**Links**
Issue: [1522](https://github.com/vacp2p/nim-libp2p/issues/1522)

### DIFF
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -12,7 +12,7 @@
 import sequtils, std/[tables]
 import chronos, chronicles, metrics, stew/[endians2, byteutils, objects]
 import ../muxer, ../../stream/connection
-import ../../utils/[zeroqueue, sequninit]
+import ../../utils/[zeroqueue, sequninit], ../../utility
 
 export muxer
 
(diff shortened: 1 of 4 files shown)


gmelodie (approved):

arnetheduck:
arrays are allocated on the stack meaning there's no heap allocation involved. further, nim already performs [RVO](https://en.wikipedia.org/wiki/Copy_elision#Return_value_optimization) which rewrites return values to become parameters when it makes sense to do so (ie when they're larger than a register or two, which these byte arrays are not) - small values like this are transferred in registers.

Before attempting microoptimizations like this which tend to make the code harder to read and harder to work with,  it's important to do two things:
* run a profiler to check if it's a problem and to verify that the "optimization" indeed makes the situation better (with LTO enabled)
* examine the generated C and [assembly code](https://godbolt.org/z/jz474ehzP) to verify that a problem exists to begin with, and what that problem is (more often than not, it can be for example array range checking that interferes with compiler optimizations)

MorganaFuture:
@arnetheduck You were absolutely right! Thank you for the valuable feedback.

I've conducted the performance analysis you suggested, and the results confirm your intuition:

### Benchmark Results:
- **Array version**: 175-188ms per 1M operations
- **In-place version**: 180-208ms per 1M operations
- **Real-world scenario (yamux encoding)**: 10.7% slower with in-place version

### C Code Analysis:
The in-place version still creates the array internally and then adds a manual copy loop with bounds checking:
```
toBytesLE(value, bytes);  // Still creates the array!
for (i = 0; i < 4; i++) {
         // bounds checks, overflow checks...
         dest[offset + i] = bytes[i];
     }
```
You were correct that arrays are stack-allocated, so there's no heap allocation to avoid. The "optimization" actually adds overhead instead of removing it. I'll close this PR as the changes degrade performance rather than improve it. Thank you!

PR #1533, opened 2025-07-14
chore: hide autotls under compile flag

richard-ramos:
Adds a `-d:libp2p_autotls_support` flag to enable the usage of AutoTLS
As suggested by @arnetheduck, we're disabling this feature, as any additional dependency is a potential security issue.
cc: @tersec

### DIFF
--- a/libp2p/autotls/acme/api.nim
+++ b/libp2p/autotls/acme/api.nim
@@ -1,509 +1,519 @@
-import options, sequtils, strutils, json, uri
-from times import DateTime, parse
-import chronos/apps/http/httpclient, jwt, results, bearssl/pem, chronicles
+when defined(libp2p_autotls_support):
+  import options, sequtils, strutils, json, uri
+  from times import DateTime, parse
+  import chronos/apps/http/httpclient, jwt, results, bearssl/pem, chronicles
 
-import ./utils
-import ../../crypto/crypto
-import ../../crypto/rsa
+  import ./utils
+  import ../../crypto/crypto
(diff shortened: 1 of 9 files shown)


gmelodie (approved):

libp2p/autotls/acme/api.nim
```diff
@@ -1,509 +1,519 @@
-import options, sequtils, strutils, json, uri
```
vladopajic, line 1: curious if what's the better way to do this.

- could we leave types and types definitions but just remove `withAutotls` method in switch - then we technically disabled the feature?
- adding `when defined(libp2p_autotls_support):` in front of whole file removes line history (sometimes vary interesting info)

vladopajic (approved):

libp2p/autotls/acme/api.nim
arnetheduck: the long-term plan is to integrate `-d` support with [nimble feature flags](https://github.com/nim-lang/nimble/pull/1346) using some suitable convention that has yet to be set in stone - it might be worth to experiment with newer nimble versions to understand what this could look like.

richard-ramos: > could we leave types and types definitions but just remove withAutotls method in switch - then we technically disabled the feature?

I can try this.

>  it might be worth to experiment with newer nimble versions.

This sounds good, but lets do it on a separate PR whose scope is experimenting with that feature, while getting this PR merged ASAP to fix the error found by @jakubgs

PR #1535, opened 2025-07-14
feat(wstransport): add autotls support

gmelodie:
Use autotls installed certificates in `wstransport`.

### DIFF
--- a/libp2p/autotls/mockservice.nim
+++ b/libp2p/autotls/mockservice.nim
@@ -1,0 +1,32 @@
+import ./service, ./acme/client, ../peeridauth/client
+
+import ../crypto/crypto, ../crypto/rsa, websock/websock
+
+type MockAutotlsService* = ref object of AutotlsService
+  mockedCert*: TLSCertificate
+  mockedKey*: TLSPrivateKey
+
+proc new*(
+    T: typedesc[MockAutotlsService],
+    rng: ref HmacDrbgContext = newRng(),
+    config: AutotlsConfig = AutotlsConfig.new(),
+): T =
+  T(
+    acmeClient: ACMEClient.new(api = ACMEApi.new(acmeServerURL = config.acmeServerURL)),
+    brokerClient: PeerIDAuthClient.new(),
+    bearer: Opt.none(BearerToken),
+    cert: Opt.none(AutotlsCert),
+    certReady: newAsyncEvent(),
+    running: newAsyncEvent(),
+    config: config,
+    rng: rng,
+  )
+
+method getCertWhenReady*(
+    self: MockAutotlsService
+): Future[AutotlsCert] {.async: (raises: [AutoTLSError, CancelledError]).} =
+  AutotlsCert.new(self.mockedCert, self.mockedKey, Moment.now)
+
+method setup*(self: MockAutotlsService) {.base, async.} =
+  self.running.fire()
+  discard
--- a/libp2p/autotls/service.nim
+++ b/libp2p/autotls/service.nim
@@ -10,19 +10,16 @@
 {.push raises: [].}
 {.push public.}
 
-import net, results, json, sequtils
-
-import chronos/apps/http/httpclient, chronos, chronicles, bearssl/rand
-
+import chronos, chronicles, net, results
+import chronos/apps/http/httpclient, bearssl/rand
 import
   ./acme/client,
   ./utils,
   ../crypto/crypto,
   ../nameresolving/dnsresolver,
   ../peeridauth/client,
-  ../peerinfo,
   ../switch,
-  ../utils/heartbeat,
+  ../peerinfo,
   ../wire
 
 logScope:
@@ -40,6 +37,9 @@ const
   DefaultRenewCheckTime* = 1.hours
   DefaultRenewBufferTime = 1.hours
 
+  DefaultIssueRetries = 3
+  DefaultIssueRetryTime = 1.seconds
+
   AutoTLSBroker* = "registration.libp2p.direct"
   AutoTLSDNSServer* = "libp2p.direct"
   HttpOk* = 200
@@ -53,6 +53,7 @@ type SigParam = object
 
 type AutotlsCert* = ref object
   cert*: TLSCertificate
+  privkey*: TLSPrivateKey
   expiry*: Moment
 
 type AutotlsConfig* = ref object
@@ -61,27 +62,43 @@ type AutotlsConfig* = ref object
   ipAddress: Opt[IpAddress]
   renewCheckTime*: Duration
   renewBufferTime*: Duration
+  issueRetries*: int
+  issueRetryTime*: Duration
 
 type AutotlsService* = ref object of Service
-  acmeClient: ACMEClient
+  acmeClient*: ACMEClient
+  brokerClient*: PeerIDAuthClient
   bearer*: Opt[BearerToken]
-  brokerClient: PeerIDAuthClient
   cert*: Opt[AutotlsCert]
   certReady*: AsyncEvent
-  config: AutotlsConfig
+  running*: AsyncEvent
+  config*: AutotlsConfig
   managerFut: Future[void]
   peerInfo: PeerInfo
-  rng: ref HmacDrbgContext
+  rng*: ref HmacDrbgContext
 
 when defined(libp2p_autotls_support):
-  proc new*(T: typedesc[AutotlsCert], cert: TLSCertificate, expiry: Moment): T =
-    T(cert: cert, expiry: expiry)
+  import json, sequtils, bearssl/pem
 
-  proc getCertWhenReady*(
+  import
+    ../crypto/rsa,
+    ../utils/heartbeat,
+    ../transports/tcptransport,
+    ../transports/transport
+
+  proc new*(
+      T: typedesc[AutotlsCert],
+      cert: TLSCertificate,
+      privkey: TLSPrivateKey,
+      expiry: Moment,
+  ): T =
+    T(cert: cert, privkey: privkey, expiry: expiry)
+
+  method getCertWhenReady*(
       self: AutotlsService
-  ): Future[TLSCertificate] {.async: (raises: [AutoTLSError, CancelledError]).} =
+  ): Future[AutotlsCert] {.base, async: (raises: [AutoTLSError, CancelledError]).} =
     await self.certReady.wait()
-    return self.cert.get.cert
+    return self.cert.get
 
   proc new*(
       T: typedesc[AutotlsConfig],
@@ -90,6 +107,8 @@ when defined(libp2p_autotls_support):
       acmeServerURL: Uri = parseUri(LetsEncryptURL),
       renewCheckTime: Duration = DefaultRenewCheckTime,
       renewBufferTime: Duration = DefaultRenewBufferTime,
+      issueRetries: int = DefaultIssueRetries,
+      issueRetryTime: Duration = DefaultIssueRetryTime,
   ): T =
     T(
       dnsResolver: DnsResolver.new(nameServers),
@@ -97,6 +116,8 @@ when defined(libp2p_autotls_support):
       ipAddress: ipAddress,
       renewCheckTime: renewCheckTime,
       renewBufferTime: renewBufferTime,
+      issueRetries: issueRetries,
+      issueRetryTime: issueRetryTime,
     )
 
   proc new*(
@@ -111,6 +132,7 @@ when defined(libp2p_autotls_support):
       bearer: Opt.none(BearerToken),
       cert: Opt.none(AutotlsCert),
       certReady: newAsyncEvent(),
+      running: newAsyncEvent(),
       config: config,
       managerFut: nil,
       peerInfo: nil,
@@ -123,7 +145,6 @@ when defined(libp2p_autotls_support):
     trace "Setting up AutotlsService"
     let hasBeenSetup = await procCall Service(self).setup(switch)
     if hasBeenSetup:
-      self.peerInfo = switch.peerInfo
       if self.config.ipAddress.isNone():
         try:
           self.config.ipAddress = Opt.some(getPublicIPAddress())
@@ -135,12 +156,14 @@ when defined(libp2p_autotls_support):
 
   method issueCertificate(
       self: AutotlsService
-  ) {.
+  ): Future[bool] {.
       base, async: (raises: [AutoTLSError, ACMEError, PeerIDAuthError, CancelledError])
   .} =
     trace "Issuing certificate"
 
-    assert not self.peerInfo.isNil(), "Cannot issue new certificate: peerInfo not set"
+    if self.peerInfo.isNil():
+      error "Cannot issue new certificate: peerInfo not set"
+      return false
 
     # generate autotls domain string: "*.{peerID}.libp2p.direct"
     let baseDomain =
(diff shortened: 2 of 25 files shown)


tests/testwstransport.nim
```diff
@@ -136,3 +233,73 @@ suite "WebSocket transport":
...
+    autotls.mockedCert = TLSCertificate.init(AutoTLSCertificate)
+    await autotls.setup()
+
+    let wstransport = WsTransport.new(
+      Upgrade(),
+      nil, # TLSPrivateKey
+      nil, # TLSCertificate
+      autotls,
+      {TLSFlags.NoVerifyHost, TLSFlags.NoVerifyServerName},
+    )
```
richard-ramos: Are these flags needed?, also on line 278

libp2p/autotls/mockservice.nim
```diff
@@ -0,0 +1,32 @@
...
+  )
+
+method getCertWhenReady*(
+    self: MockAutotlsService
+): Future[AutotlsCert] {.async: (raises: [AutoTLSError, CancelledError]).} =
+  AutotlsCert.new(self.mockedCert, self.mockedKey, Moment.now)
+
+method setup*(self: MockAutotlsService) {.base, async.} =
+  self.running.fire()
+  discard
```
richard-ramos:
```suggestion

```

libp2p/transports/wstransport.nim
```diff
@@ -181,8 +202,6 @@ method start*(
 
   trace "Listening on", addresses = self.addrs
 
-  self.running = true
```
richard-ramos, line 184: when is `self.running = true` happening?

libp2p/transports/wstransport.nim
```diff
@@ -255,8 +274,14 @@ proc connHandler(
...
-  ## accept a new WS connection
-  ##
+  trace "WsTransport accept"
+
+  # wstransport can only start accepting connections after autotls is done
+  # if autotls is not present, self.running will be true right after start is called
+  var retries = 0
+  while not self.running and retries < DefaultAutotlsRetries:
+    retries += 1
+    await sleepAsync(DefaultAutotlsWaitTimeout)
```
richard-ramos, line 284: This loop is a bit ugly.

Wdyt of having an async event that is fired when running is set to true, and then here the code could look like

```
await self.onRunning.wait().withTimeout(someTimeout)
```

libp2p/transports/wstransport.nim
```diff
@@ -349,13 +372,15 @@ method dial*(
 method handles*(t: WsTransport, address: MultiAddress): bool {.gcsafe, raises: [].} =
   if procCall Transport(t).handles(address):
     if address.protocols.isOk:
+      debug "WS matching", address = address, match = WebSockets.match(address)
```
richard-ramos:
```suggestion
```

libp2p/switch.nim
```diff
@@ -350,8 +351,20 @@ proc start*(s: Switch) {.public, async: (raises: [CancelledError, LPError]).} =
...
+    debug "addresses", addrs = addrs
+
     if addrs.len > 0 or t.running:
-      startFuts.add(t.start(addrs))
+      let fut = t.start(addrs)
+      startFuts.add(fut)
+      if t of TcpTransport:
+        await fut
+        s.acceptFuts.add(s.accept(t))
+        s.peerInfo.listenAddrs &= t.addrs
```
richard-ramos, line 363: Instead of having this special condition for TcpTransport here and in line 380, wouldn't it be possible for AutoTLSService to asynsspawn a routine that will check the switch for the listenAddrs it has, and request the certificate when the switch has an available TCP addrs?

libp2p/autotls/service.nim
```diff
@@ -135,12 +156,14 @@ when defined(libp2p_autotls_support):
 
   method issueCertificate(
       self: AutotlsService
-  ) {.
+  ): Future[bool] {.
```
vladopajic, line 160: hmm with this change code needs to handle errors and check result when doing `issueCertificate`? return value kind of looks  redundant here. what was reasoning for this change?

libp2p/autotls/service.nim
```diff
@@ -163,57 +193,87 @@ when defined(libp2p_autotls_support):
...
     self.cert = Opt.some(newCert)
     self.certReady.fire()
-    debug "Certificate installed"
+    trace "Certificate installed"
+    true
+
+  proc tryIssueCertificate(self: AutotlsService) {.async: (raises: [CancelledError]).} =
+    for _ in 0 ..< self.config.issueRetries:
+      try:
+        if await self.issueCertificate():
```
vladopajic, line 246: does return value of `issueCertificate` needs to be checked here?

libp2p/autotls/service.nim
```diff
@@ -163,57 +193,87 @@ when defined(libp2p_autotls_support):
...
+    let tcpTransports = switch.transports
+      .filterIt(it of TcpTransport and it.running)
+      .mapIt(TcpTransport(it))
+
+    if tcpTransports.len == 0:
+      error "Could not find a running TcpTransport in switch"
+      return
+
+    let tcpTransport = tcpTransports[0]
+
```
vladopajic: why is this code needed? why is running tcp transport needed here? what happens with `tcpTransport` variable?

libp2p/switch.nim
```diff
@@ -364,12 +377,11 @@ proc start*(s: Switch) {.public, async: (raises: [CancelledError, LPError]).} =
 
   for t in s.transports: # for each transport
     if t.addrs.len > 0 or t.running:
+      if t of TcpTransport:
+        continue # already added previously
```
vladopajic, line 384: please explain

tests/testwstransport.nim
```diff
@@ -23,6 +25,98 @@ import
 import ./helpers, ./commontransport
 
 const
+  AutoTLSCertificate =
```
vladopajic: why are these hardcoded values needed now?

gmelodie: They allow us to check which cert was set in each case (if autotls is enabled but SecureCert was manually set, we prefer SecureCert). Does that make sense?

libp2p/switch.nim
gmelodie: The problem here runs a bit deeper. In line 369 we await all transports' futures, which include `wstransport`. For wstransport to be finished it needs certs from autotls, but autotls needs tcptransport to be running *and* accepting connections, which only happens in 382 if we don't have this condition here.

libp2p/switch.nim
gmelodie: Sure! There's a dependency train that requires `tcptransport`, `wstransport` and `autotls service` to be started in the following order:
1. `tcptransport` (needs to be started *and* be accepting connections)
2. `autotls`
3. `wstransport`

So what I did was to make sure `tcptransport` was started and started accepting connections in lines 359-361, thus this `continue` will keep us from doing it again here.

tests/testwstransport.nim
vladopajic: can't those be generated instead?

libp2p/transports/wstransport.nim
gmelodie: [Here!](https://github.com/vacp2p/nim-libp2p/blob/5f256049abefbc688fc9c15752f5e6a4052203a4/libp2p/transports/transport.nim#L43) When we do [`await procCall Transport(self).start(addrs)`](https://github.com/vacp2p/nim-libp2p/pull/1535/files/d80a77749c946e9c4cd08354b1305f01eda338e7#diff-32742335c1fc84ce5a63289b969f76e9bfb25c3185d7567ad47a5144599763f4R154)

libp2p/transports/wstransport.nim
gmelodie: I think this could be a change introduced in a separate PR like so:
1. add `onRunning` asyncEvent field to base `Transport` type
2. make `Transport.start` also `fire` `onRunning`
3. change other transports to benefit from this too

In a note, I saw that many transports are redundant in the way they set `self.running` (i.e. they set `self.running = true` and also call the base `Transport.start` method, which does the same thing). We could clean that up as well in this PR.

libp2p/autotls/service.nim
gmelodie: > what happens with tcpTransport variable?

Oops! My bad there. 

> why is this code needed? why is running tcp transport needed here? 

We need a tcpTransport to receive a dial from the autotls broker. I remember we talked about having autotls spawn its own tcp server on DMs but I spoke to @richard-ramos and we decided to go with the other approach (i.e. using a tcpTransport from the switch). Edited the code to be more straightforward.

libp2p/switch.nim
vladopajic: hmm.. could this code be changed:

```
      s.acceptFuts.add(s.accept(t))
      s.peerInfo.listenAddrs &= t.addrs
```

so that it only happens once? 
if this means we need to sort items in specific order, lets sort items then process them in that order in one place rather then two places.

vladopajic: aha i get it now....

libp2p/autotls/service.nim
vladopajic: so point of this code is to return error if tcp transport is not started ?
if so please consider improving understanding what's happening here and why. because looking at the code is not obvious.

tests/testwstransport.nim
gmelodie: Nope!

libp2p/switch.nim
```diff
@@ -350,8 +351,20 @@ proc start*(s: Switch) {.public, async: (raises: [CancelledError, LPError]).} =
 
     s.peerInfo.listenAddrs.keepItIf(it notin addrs)
 
+    debug "addresses", addrs = addrs
```
vladopajic:
```suggestion
    debug "listen addresses", addrs = addrs
```

libp2p/autotls/service.nim
gmelodie: Agreed! Can you give me some pointers on how you'd do that?

tests/testwstransport.nim
gmelodie: I believe we discussed this a couple weeks back. The problem with generating the certs is that we'd either have to do that in an integration test to get certs from Let's Encrypt (we couldn't do that in this test here) or we'd generate certs that wouldn't really match what we'd see in real scenarios. @richard-ramos thoughts?

libp2p/switch.nim
vladopajic: ws transports -> accept blocked until auto tls is ready -> blocked until tcp transport is accepting
tcp transport -> no dependencies for accepting 

does this graph represent what start dependency should be? if so, then these are separate processes that can happen in separate futures?

tests/testwstransport.nim
vladopajic: yeah we discussed.. but now question is are ever going to use generated certs or we will always have to use hardcoded certs? 

what happens someone adds new issuer, then new issuer creates different certificate. do we have to hardcode values of those issuer here as well?

and isn't responsibility of  testing actual cert value in another place? ws transport is integration of many components, one of whom is `TlsHttpServer`. if we really wanted to ensure that `TlsHttpServer` can accept some type of certificate (actual values) we need to specify that test in file that tests `TlsHttpServer` functionality - and not here where we just integrate different components.

in general these hardcoded values here are missing the scope of what needs to be tested here. because if we want to test `TlsHttpServer` then those test cases need to go to different file. and for purpose of testing `wstransport` any valid certificate will work.

libp2p/autotls/service.nim
vladopajic: create proc `hasTcpTransportStarted` that returns `bool`

call proc in this place:


```
# starting auto tls requires tcp transport to be running
if not hasTcpStarted():
 error "Could not find a running TcpTransport in switch"
 return
 ```

File: libp2p/switch.nim (hunk 6 above)
Comment (gmelodie) [reply libp2p/switch.nim:363]: Pretty much yes. The only thing missing there is the `start` functions:

ws transports -> start blocked until auto tls is ready -> blocked until tcp transport is accepting
tcp transport -> accept blocked until trasport started -> no dependencies to start

Comment (vladopajic) [reply libp2p/switch.nim:363]: > ws transports -> start blocked until auto tls is ready -> blocked until tcp transport is accepting
> tcp transport -> accept blocked until trasport started -> no dependencies to start

we just need to collect all futures to single wait:
wait allFutures = transport.start() + transport.accept()

ws started  (gets blocked on auto tls) -> tcp started -> ws accept started (gets blocked because not started) -> tcp accept started -> ws start gets unblocked auto tls complets -> ws accept completes

Comment (vladopajic) [reply libp2p/switch.nim:363]: having transports [ws, tc], code would call in this order:
ws start
tcp start
ws accept
tcp accept

since tcp is not depended on anything it is guaranteed to complete first (tcp start future completes -> then tcp accept completes).
ws will complete with delay after tcp completes.

Comment (gmelodie) [reply libp2p/switch.nim:363]: Hmmm I don't know if that's not going to break anything. Accept futures require `self.running == true` otherwise they fail (e.g. tcpTransport [here](https://github.com/vacp2p/nim-libp2p/blob/5f256049abefbc688fc9c15752f5e6a4052203a4/libp2p/transports/tcptransport.nim#L219)). I think we should add a separate task to create `asyncEvent`s that fire when `running == true`, making async dependencies cleaner and easier to work with (i suggested this [here](https://github.com/vacp2p/nim-libp2p/pull/1535#discussion_r2213301899)).

File: libp2p/autotls/service.nim (hunk 9 above)
Comment (gmelodie) [reply libp2p/autotls/service.nim]: Better now?

File: tests/testwstransport.nim (hunk 11 above)
Comment (gmelodie) [reply tests/testwstransport.nim]: Better now?

File: libp2p/transports/wstransport.nim (hunk 4 above)
Comment (richard-ramos) [reply libp2p/transports/wstransport.nim:284]: This seems like a good thing to do in that separate PR, yes!

Comment (gmelodie) [reply libp2p/transports/wstransport.nim:284]: https://github.com/vacp2p/nim-libp2p/issues/1553

File: libp2p/switch.nim (hunk 6 above)
Comment (vladopajic) [reply libp2p/switch.nim:363]: it definitely would, but that what's coming should be fixable for wizards :)

File: tests/testwstransport.nim (hunk 11 above)
Comment (vladopajic) [reply tests/testwstransport.nim]: this is _pleasure_ for my eyes and mind

File: libp2p/switch.nim (hunk 12 above)
Comment (gmelodie) [reply libp2p/switch.nim]: I don't quite understand the difference between `addrs` and `listenAddrs`, but apparently there is one. So `addrs` is _technically_ not `listenAddrs`. I'm fine with whatever though. lmk what you think

Comment (vladopajic) [reply libp2p/switch.nim]: i wanted to be more descriptive, what are these addresses used for? if you see log in in terminal:

`DEBUG [TIME] addresses   addresses={0x111, 0x222,  0x333}`

 what do we know here?  
 addresses for what? 
 
current log message only looks usefully if you really know what are you look at...
 
 --- 
 
if there is better log text then "listen addresses" consider using it

File: libp2p/autotls/service.nim (hunk 9 above)
Comment (vladopajic) [reply libp2p/autotls/service.nim]: i dont see any change

Comment (vladopajic) [reply libp2p/autotls/service.nim]: image

File: libp2p/switch.nim [hunk 13]
```diff
@@ -354,7 +355,19 @@ proc start*(s: Switch) {.public, async: (raises: [CancelledError, LPError]).} =
...
 
     if addrs.len > 0 or t.running:
-      startFuts.add(t.start(addrs))
+      let fut = t.start(addrs)
+      startFuts.add(fut)
+      if t of TcpTransport:
+        await fut
+        s.acceptFuts.add(s.accept(t))
+        s.peerInfo.listenAddrs &= t.addrs
+  debug "switch addresses",
```
richard-ramos, line 364: This is probably unneeded as later in the start() we have 
```
  debug "Started libp2p node", peer = s.peerInfo
```
which will log the same information

PR #1536, opened 2025-07-14
test: initial sniff tests for find node implementation

Ben-PH:
First test for find node.

### DIFF
--- a/tests/kademlia/testfindnode.nim
+++ b/tests/kademlia/testfindnode.nim
@@ -1,5 +1,5 @@
 {.used.}
-import strformat
+import chronicles, strformat, sequtils
 import std/enumerate
 import chronos
 import ../../libp2p/[switch, builders]
@@ -39,12 +39,10 @@ suite "KadDHT - FindNode":
       switches[i].mount(kads[i])
 
     # Once the the creation/mounting of switches are done, we can start
-    # TODO: instead of awaiting sequentially, do it concurrently
-    for i in 0 ..< swarmSize:
-      await switches[i].start()
+    await switches.mapIt(it.start()).allFutures()
 
     # Now we can activate the network
-    # TODO: instead of awaiting sequentially, do it concurrently
+    # Needs to be done sequentially, hence the deterministic ordering of completion
     for i in 1 ..< swarmSize:
       await kads[i].bootstrap(@[switches[0].peerInfo])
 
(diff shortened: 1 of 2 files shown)


libp2p/protocols/kademlia/routingtable.nim
```diff
@@ -23,6 +23,9 @@ type
     selfId*: Key
     buckets*: seq[Bucket]
 
+proc `$`*(rt: RoutingTable): string =
+  "selfId(" & $rt.selfId & ") buckets(" & $rt.buckets & ")"
+
```
Ben-PH, line 28: can be useful for debugging/logging

tests/kademlia/testfindnode.nim
```diff
@@ -14,27 +19,60 @@ proc createSwitch(): Switch =
...
+      for id in entries:
+        let count = countBucketEntries(kad.rtable.buckets, id)
+        assert(
+          count == 1,
+          fmt"bootstrap state broken - count: {count}|entries: {entries}|i: {i}|key: {id}|buckets: {kad.rtable.buckets}",
+        )
+      entries.add(kad.rtable.selfId)
 
-    await sleepAsync(2.seconds)
+    echo "finding"
```
richard-ramos: These echo's should not be part of the unit test. If they're absolutely required, do make them a `trace` log.

tests/kademlia/testfindnode.nim
```diff
@@ -14,27 +19,60 @@ proc createSwitch(): Switch =
...
+    # every node needs a switch, and an assosciated kad mounted to it
+    for i in 0 ..< swarmSize:
+      switches.add(createSwitch())
+      kads.add(KadDHT.new(switches[i]))
+      switches[i].mount(kads[i])
+
+    # Once the the creation/mounting of switches are done, we can start
+    # TODO: instead of awaiting sequentially, do it concurrently
+    for i in 0 ..< swarmSize:
+      await switches[i].start()
```
richard-ramos: you can use `await allFutures(switches)` to not do the for, same in line 48 (altho there you'll have to filter out node 0)

tests/kademlia/testfindnode.nim
```diff
@@ -14,27 +19,60 @@ proc createSwitch(): Switch =
...
 
-    # TODO: test
+    echo "starting"
+    # assert that every node has exactly one entry for the id of every other node
+    for id in entries:
+      for k in kads:
+        if k.rtable.selfId == id:
+          continue
+        let count = countBucketEntries(k.rtable.buckets, id)
+        assert(
```
richard-ramos: In tests we do use `doAssert` instead of assert. For consistence i'm thinking we should use the same

tests/kademlia/testfindnode.nim
```diff
@@ -14,27 +19,60 @@ proc createSwitch(): Switch =
...
+        let count = countBucketEntries(kad.rtable.buckets, id)
+        assert(
+          count == 1,
+          fmt"bootstrap state broken - count: {count}|entries: {entries}|i: {i}|key: {id}|buckets: {kad.rtable.buckets}",
+        )
+      entries.add(kad.rtable.selfId)
 
-    await sleepAsync(2.seconds)
+    echo "finding"
+    discard await kads[1].findNode(kads[2].switch.peerInfo.peerId.toKey())
```
vladopajic: why not `kads[2].switch.peerInfo` -> `switches[2].peerInfo`? then you could keep  `Switch.switch` field private.

vladopajic (approved):

tests/kademlia/testfindnode.nim
Ben-PH: oops, was meant to remove them.

tests/kademlia/testfindnode.nim
Ben-PH: That's actually something I noticed writing some more tests: I'm going to change to `kads[2].rtable.self`, which means the switch can be kept private.

Though the switch can be kept private, is there a motivation to keep it that way? I would imagine there would be some merit to being able to look into the switch on which a protocol is mounted on, but I'm not up to speed on such ins/outs of this library.

vladopajic: not really sure what future brings, but in general all _unnecessary_ relaxations should be avoided as they potentially allow some form of mess to be introduced.

Ben-PH: hard-agree with that.

tests/kademlia/testfindnode.nim
Ben-PH: The second loop needs to be done with deterministic ordering. The tests assume `1` is bootstrapped before `2`, but using `allFutures` could non-deterministically change the ordering, causing flaky false-fails

Issue #1534, opened 2025-07-14
failed to accept connection: Future operation cancelled!


gmelodie:
The following error appears fairly often when running `nimble test`.
```
ERR 2025-07-14 14:03:37.870+02:00 Exception in accept loop, exiting          topics="libp2p switch" tid=405703 description="failed to accept connection: Future operation cancelled!"
```

Ben-PH:
To clarify: this is a logging output reporting an error, and not a "test failed" error.

Having this error on blast during `nimble testnative` can raise some uncertainty for those without a certain degree of familiarity with the code-base such as new core contributors (e.g. me), and community contributors:

- when a test is failing/bug is being dealt with, the log-output errors can be a source of friction in resolving the issue: Relevant errors could be lost in the crowd, or the relevancy of common log-errors can be uncertain.
- when running tests, the error logs could be interpreted as "soft-failures", errornously or otherwise.
- to understand the significance of these errors, whether they be signs of something that needs fixing, or part of normal ops, takes a certain intuition that usually comes from time and experience working with the code base, adding to the barrier-to-entry for community contributors.

Largely, that's just a few ways of saying the same thing. Breaking from that theme:

- error logs are "normalised". When an error log comes around that _is_ significant, this normalisation of error logs elevates the risk of such a log being overlooked.

Thank you @MorganaFuture for rolling up your sleeves to address this. I don't just speak for myself when I say we appreciate your involvement.

PR #1537, opened 2025-07-14
fix: properly handle CancelledError in switch accept loop

MorganaFuture:
**What**
Separate handling of CancelledError from other CatchableError
  exceptions in the switch accept loop

**Why**

 - When switch.stop() is called during tests, it cancels all active accept futures
 - Previously, CancelledError was caught as CatchableError and wrapped with "failed to accept connection" message
- This caused spurious ERROR logs during normal test teardown: "Exception   in accept loop, exiting: Future operation cancelled!"
- Cancellation is expected behavior during shutdown and should not be logged as an error

**Links**
Issue: [1534](https://github.com/vacp2p/nim-libp2p/issues/1534)


vladopajic:
@MorganaFuture hey man could you push changes again with signed signature so we ca merge this one? thanks

vladopajic:
@MorganaFuture get ready for merged notification

RWSINC (approved):

Issue #1532, opened 2025-07-14
Switch builder using nil-checks instead of `Option[<type>]`


Ben-PH:
The builder type here:

https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/builders.nim#L53-L75

for example, the `rdv` field is of type `RendezVous`, as if it must always exist, but its use contradicts this:

https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/builders.nim#L364-L366

the `rdv` field should be `Option[RendezVous]`, and its use should be more like

```nim
b.rdv.withValue(rdvLpp):
  b.rdvLpp.setup(switch)
  switch.mount(rdvLpp)
```

This might be happening with other builders. 

The issues with this are as follows:

- adds null-dereference as a potential issue to the code base
- Not having a type of a field as `Option` requires precious cognitive overhead to understand that the field is optional
- If a code base is peppered with potentially nil values, any non-primitive value could potentially be nil, increasing developer friction.

A follow up to this issue could be to audit the code base for nil checks, and replace them with the use of `Option`.

PR #1539, opened 2025-07-14
refactor: switch SwitchBuilder to use Option types instead of nil-checks

MorganaFuture:
**What**
Replace nil-checks with Option[T] types in SwitchBuilder for autotls, circuitRelay, and rdv fields.

**Why**

Eliminates nil-dereference risks and provides compile-time safety. Makes optional vs required fields explicit.

**Links**
Issue: [1532](https://github.com/vacp2p/nim-libp2p/issues/1532)


gmelodie (commented):
Use [`nim-results`](https://github.com/arnetheduck/nim-results)

MorganaFuture:
> Use [`nim-results`](https://github.com/arnetheduck/nim-results)

Accepted

vladopajic (approved):

arnetheduck:
doesn't this break public api? if yes, the old api needs to be maintained with `{.deprecated.}`

Ben-PH:
@MorganaFuture would you like me to take over for a bit, if only to resolve conflicts with master? or perhaps bring this home so that something you started gets finished? You're in control, this is just a prompt so things don't go _too_ stale.

If I notice no response after some time, I'll go ahead and get it finished for you, making sure you're on the author list for the relevant commits.

gmelodie:
merged in #1732

PR #1540, opened 2025-07-14
fix: avoid raw async methods

Ivansete-status:
## Description

Simple PR to encourage minimize the use of `raw: true` methods/procs as much as possible.

According to [this](https://status-im.github.io/nim-chronos/async_procs.html#raw-async-procedures), the `raw: true` technique bypasses the `async` transformation performed by the the compiler. Then, when using `raw: true`, the returned `Future[T]` should be _manually_ completed or failed. Making the code prone to errors because is very easy to overlook any detail.

Also, notice that the `raw: true` bypasses the usual `Future` exception handling, leading to unattended exceptions. Therefore, the following `await msgFut` ( await writeLp(...) ) can throw an exception that may not be caught properly: https://github.com/vacp2p/nim-libp2p/blob/eee8341ad2fdf9cb2b4258eefd282026eeb56297/libp2p/protocols/pubsub/pubsubpeer.nim#L357

..which in turn I think is the first candidate to lead to the memory issues described by @NagyZoltanPeter in
- https://github.com/waku-org/nwaku/issues/3496

-------

Besides, and according to [nim-chronos docs](https://status-im.github.io/nim-chronos/async_procs.html#raw-async-procedures), the `raw: true` procs/methods should not raise any exception for the reason exposed above.

-------

Summarizing, let's avoid using `raw: true` as much as possible :)

### DIFF
--- a/libp2p/stream/lpstream.nim
+++ b/libp2p/stream/lpstream.nim
@@ -246,23 +246,20 @@ method write*(
   raiseAssert("[LPStream.write] abstract method not implemented!")
 
 method writeLp*(
-    s: LPStream, msg: openArray[byte]
-): Future[void] {.
-    base, async: (raises: [CancelledError, LPStreamError], raw: true), public
-.} =
+    s: LPStream, msg: seq[byte]
+) {.base, async: (raises: [LPStreamError, CancelledError]), public.} =
   ## Write `msg` with a varint-encoded length prefix
-  let vbytes = PB.toBytes(msg.len().uint64)
-  var buf = newSeqUninit[byte](msg.len() + vbytes.len)
-  buf[0 ..< vbytes.len] = vbytes.toOpenArray()
-  buf[vbytes.len ..< buf.len] = msg
-  s.write(buf)
+  let lenPrefix = PB.toBytes(msg.len.uint64)
+  let totalLen = lenPrefix.len + msg.len
+  var buf = newSeq[byte](totalLen)
+  buf[0 ..< lenPrefix.len] = lenPrefix.toOpenArray() # Copy varint length prefix
+  buf[lenPrefix.len ..< totalLen] = msg # Copy actual message
+  await s.write(buf)
 
 method writeLp*(
     s: LPStream, msg: string
-): Future[void] {.
-    base, async: (raises: [CancelledError, LPStreamError], raw: true), public
-.} =
-  writeLp(s, msg.toOpenArrayByte(0, msg.high))
+) {.base, async: (raises: [CancelledError, LPStreamError]), public.} =
+  await writeLp(s, msg.toBytes())
 
 proc write*(
     s: LPStream, msg: string
--- a/libp2p/protocols/connectivity/relay/utils.nim
+++ b/libp2p/protocols/connectivity/relay/utils.nim
@@ -36,21 +36,21 @@ proc sendStatus*(
 
 proc sendHopStatus*(
     conn: Connection, code: StatusV2
-) {.async: (raises: [CancelledError, LPStreamError], raw: true).} =
+) {.async: (raises: [CancelledError, LPStreamError]).} =
   trace "send hop relay/v2 status", status = $code & "(" & $ord(code) & ")"
   let
     msg = HopMessage(msgType: HopMessageType.Status, status: Opt.some(code))
     pb = encode(msg)
-  conn.writeLp(pb.buffer)
+  await conn.writeLp(pb.buffer)
 
 proc sendStopStatus*(
     conn: Connection, code: StatusV2
-) {.async: (raises: [CancelledError, LPStreamError], raw: true).} =
+) {.async: (raises: [CancelledError, LPStreamError]).} =
   trace "send stop relay/v2 status", status = $code & " (" & $ord(code) & ")"
   let
     msg = StopMessage(msgType: StopMessageType.Status, status: Opt.some(code))
     pb = encode(msg)
-  conn.writeLp(pb.buffer)
+  await conn.writeLp(pb.buffer)
 
 proc bridge*(
     connSrc: Connection, connDst: Connection


libp2p/stream/lpstream.nim
```diff
@@ -246,23 +246,20 @@ method write*(
...
+) {.base, async: (raises: [LPStreamError, CancelledError]), public.} =
   ## Write `msg` with a varint-encoded length prefix
-  let vbytes = PB.toBytes(msg.len().uint64)
-  var buf = newSeqUninit[byte](msg.len() + vbytes.len)
-  buf[0 ..< vbytes.len] = vbytes.toOpenArray()
-  buf[vbytes.len ..< buf.len] = msg
-  s.write(buf)
+  let lenPrefix = PB.toBytes(msg.len.uint64)
+  let totalLen = lenPrefix.len + msg.len
+  var buf = newSeq[byte](totalLen)
```
richard-ramos, line 254: There was a change done recently to avoid unnecesarily initializing a sequence. It should probably be kept.

libp2p/stream/lpstream.nim
```diff
@@ -246,23 +246,20 @@ method write*(
...
+  await s.write(buf)
 
 method writeLp*(
     s: LPStream, msg: string
-): Future[void] {.
-    base, async: (raises: [CancelledError, LPStreamError], raw: true), public
-.} =
-  writeLp(s, msg.toOpenArrayByte(0, msg.high))
+) {.base, async: (raises: [CancelledError, LPStreamError]), public.} =
+  await writeLp(s, msg.toBytes())
```
richard-ramos, line 262: Why this change from `toOpenArrayByte` to `toBytes`?

NagyZoltanPeter:
> ## Description
> Simple PR to encourage minimize the use of `raw: true` methods/procs as much as possible.
> 
> According to [this](https://status-im.github.io/nim-chronos/async_procs.html#raw-async-procedures), the `raw: true` technique bypasses the `async` transformation performed by the the compiler. Then, when using `raw: true`, the returned `Future[T]` should be _manually_ completed or failed. Making the code prone to errors because is very easy to overlook any detail.
> 
> Also, notice that the `raw: true` bypasses the usual `Future` exception handling, leading to unattended exceptions. Therefore, the following `await msgFut` ( await writeLp(...) ) can throw an exception that may not be caught properly:
> 
> https://github.com/vacp2p/nim-libp2p/blob/eee8341ad2fdf9cb2b4258eefd282026eeb56297/libp2p/protocols/pubsub/pubsubpeer.nim#L357
> 
> ..which in turn I think is the first candidate to lead to the memory issues described by @NagyZoltanPeter in
> 
> * [bug: extensive memory consumption during publishing messages waku-org/nwaku#3496](https://github.com/waku-org/nwaku/issues/3496)
> 
> Besides, and according to [nim-chronos docs](https://status-im.github.io/nim-chronos/async_procs.html#raw-async-procedures), the `raw: true` procs/methods should not raise any exception for the reason exposed above.
> 
> Summarizing, let's avoid using `raw: true` as much as possible :)

After reading through discussion in nim-language channel and reviewing this part of yamux.writeLp/write I have the feeling that here `raw = true` is used judicously.
The end of the call is [yamu.write](https://github.com/vacp2p/nim-libp2p/blob/cd60b254a0700b0daac7a6cb2c0c48860b57c539/libp2p/muxers/yamux/yamux.nim#L411) where the data finally copied into the sendQueue, so to avoid async context copies, it is one possible option.
The last write creates the future manually as required, although I don't see why the raises pragma is used here...
So to say I see option to leak the payload if thee sendQueue is not emptied correctly. But stream reset will correctly empty it....

libp2p/stream/lpstream.nim
```diff
@@ -246,23 +246,20 @@ method write*(
   raiseAssert("[LPStream.write] abstract method not implemented!")
 
 method writeLp*(
-    s: LPStream, msg: openArray[byte]
-): Future[void] {.
-    base, async: (raises: [CancelledError, LPStreamError], raw: true), public
-.} =
+    s: LPStream, msg: seq[byte]
+) {.base, async: (raises: [LPStreamError, CancelledError]), public.} =
```
arnetheduck, line 250: this function should not be changed since doing so introduces an additional copy of all data that flows through libp2p, effectively - it would a a significant regression on the hot path.

libp2p/stream/lpstream.nim
Ivansete-status: Good question!
This proposal is a bit aside from what I really wanted to suggest in the PR. I will clean this noise asap.
At first I just wanted to keep consistency in the code with other functions.
`openArray` is in general more efficient to be passed as a parameter but in the `nim-chronos` context, given that the underlying data is copied, it has the same "copy" effect as `seq[byte]` ( @arnetheduck correct me if I'm wrong .)

libp2p/stream/lpstream.nim
Ivansete-status: You mean that we should keep the `raw: true` in that case, isn't it?
Would it help the use of `sink openArray[byte]` + ORC + raw:false?

libp2p/stream/lpstream.nim
arnetheduck: Generally, `openArray` is a view type - there are never any copies involved with it but it's limited in that the underlying data must remain in scope which is enforced by allowing it only as a function parameter (for now, upcoming Nim versions relax this).

`toOpenArrayByte` is used to convert a string (whose corresponding view type is `openArray[char]`) to `openArray[byte]` - this operation does not make a copy of the data, instead it just changes the type.

`toBytes` converts `string` to `seq[byte]` making a copy of the data in the process, since `seq` is not a view type - seq owns the copied payload.

this PR proposes two changes to `writeLp` for bytes: `msg` goes from `openArray[byte]` to `seq[byte]` and you add `{.async.}` to it.

In the old version, there is one copy of the string payload: that which is written to `buf` after the length prefix.

In the new version, there are 3 copies and one full zeroing done:
* one copy in `toBytes`
* one copy due to {.async.} which copies `msg` into the async environment
* one copy from async environment to buf
* one zeroing of the full buffer when `newSeq` is called.

We can therefore expect the new version to be about 3x slower than the old version, for sufficiently large buffers (assuming zeroing and copying is about the same - it's not quite that but similar enough).

A corollary here is that when you use `{.async.}`, you cannot use `openArray` parameters - this is tied to the above constraint, that the _underlying data_ might go out of scope - this is why `seq` and/or a copy is needed when performing the async transformation.

libp2p/stream/lpstream.nim
arnetheduck: `openArray[byte]` cannot be `sink` because it does not own the underlying data and therefore the compiler is not allowed to move it.

A version that uses `sink seq[byte]` would reduce the number of copies in this PR from 3 to 2 (it would avoid copying the data from the msg parameter to the async environment).

Ivansete-status:
Closing.
I don't consider appropriate merging it, considering all the comments. Thanks for them btw! :raised_hands: :partying_face: 
It's been a great learning exercise regarding `raw` futures.
The most relevant take-aways from my PoV:
- `raw` coroutines should not raise exceptions in the traditional way. Instead, they should `fut.fail(newException)` the returned future.
- `raw` coroutines are needed for more performant processes as they avoid copying extra data to the chronos async space.
- `raw` coroutines should be await'ed too but the future lifetime is controlled manually. It is crucial to either `complete(...)` or `fail(..)` any future created within a `raw` coroutine to avoid memory leaks.

PR #1541, opened 2025-07-15
fix: build failures caused by missing import

MorganaFuture:
## What
Fixes CI/CD build failures caused by missing import also update deprecated Chronos API calls.

 ## Why

CI/CD was failing because:
Missing import caused compilation error. Additionally fixed deprecated API warnings that would become errors after fixing the main issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ The libp2p implementation in Nim is a work in progress. We welcome contributors 
 - **Add tests**. Help nim-libp2p to be more robust by adding more tests to the [tests folder](tests/).
 - **Small PRs**. Try to keep PRs atomic and digestible. This makes the review process and pinpointing bugs easier.
 - **Code format**. Code should be formatted with [nph](https://github.com/arnetheduck/nph) and follow the [Status Nim Style Guide](https://status-im.github.io/nim-style-guide/).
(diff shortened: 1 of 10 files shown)


tersec:
> Fixes CI/CD build failures due to --styleCheck:error treating deprecation warnings as errors.

`--styleCheck:error` should not be doing this. It has nothing to do with deprecation warnings (or errors).

richard-ramos (commented):
In the lines in which you change .cancel() to .cancelAndWait(), can you describe the intention behind that change?  `cancel` currently calls `cancelSoon`, and with changing some of them to `await .cancelAndWait` you're changing the behavior of the code.

gmelodie:
@MorganaFuture I think we should keep all of them as `cancelSoon`, given that that's [how `cancel`s are implemented](https://github.com/status-im/nim-chronos/blob/b55e2816eb45f698ddaca8d8473e401502562db2/chronos/internal/asyncfutures.nim#L987). This way we'd get rid of the warnings without introducing behavior changes (and bugs possibly).

MorganaFuture:
> lgtm
> 
> change adding conditional import in `libp2p/protocols/pubsub/pubsubpeer.nim` was causing failure? if so, how `cancelSoon` is connected to this?

Just fixed the deprecated api

PR #1544, opened 2025-07-15
test(gossipsub): performance tests PoC

rlve:
This PR is a proof of concept of an automated performance testing workflow for `nim-libp2p`. It contains one basic performance scenario. The solution is based on code from DST: https://github.com/vacp2p/dst-gossipsub-test-node/tree/dockerized. The purpose of the PR is to review the approach.

**What’s included?**

- **Dockerfile** ([performance/Dockerfile](https://github.com/vacp2p/nim-libp2p/blob/test-gossipsub-performance/performance/Dockerfile)):
  - Installs dependencies and compiles the node test logic.
- **Test runner** ([runner.sh](https://github.com/vacp2p/nim-libp2p/blob/9243f47de94fac0bc80f5a9c323578fe04adb4bf/performance/runner.sh)):
  - Spins up multiple Docker containers (each running a `nim-libp2p` node with the test script).
  - Cleans up containers and output files before and after each run.
- **Node test logic** ([main.nim](https://github.com/vacp2p/nim-libp2p/blob/9243f47de94fac0bc80f5a9c323578fe04adb4bf/performance/main.nim)):
  - Configures each node with custom network parameters and metrics.
  - Handles peer discovery, connection, and message publishing.
  - Measures message delivery latency (in milliseconds, with microsecond resolution) for each node.
  - Collects per-node stats: messages sent/received, latency (min/max/avg).
  - Writes results to a JSON file for aggregation after the test run.
- **Aggregation script** ([aggregate_stats.nim](https://github.com/vacp2p/nim-libp2p/blob/13da92104f8abf36b9ce5520773d773ef95b37f8/performance/aggregate_stats.nim)):
  - Parses all per-node JSON stats files from the output directory.
  - Aggregates metrics: total messages sent/received, latency (min/max/avg).
  - Outputs a markdown-formatted summary for GitHub Actions, including a latency table and key stats.
- **CI workflow** ([.github/workflows/performance.yml](https://github.com/vacp2p/nim-libp2p/blob/9243f47de94fac0bc80f5a9c323578fe04adb4bf/.github/workflows/performance.yml)):
  - Builds the Docker image for the test nodes (with caching).
  - Executes the test runner.
  - Sets up Python and runs the aggregation script.
  - Publishes the markdown GitHub Actions job summary.
  - Adds or edits comment in the PR with the results.

### Example output in CI

https://github.com/vacp2p/nim-libp2p/actions/runs/16416412792?pr=1544

image

### How to run locally

```bash
# Build docker image
docker build -f performance/Dockerfile -t dst-test-node .

# Run the test
./performance/runner.sh 

# Run the aggregation
nim c -r performance/aggregate_stats.nim
```


performance/main.nim
```diff
@@ -0,0 +1,235 @@
...
+    return err("metrics HTTP server start failed: " & $metricsServerRes.error)
+  let server = metricsServerRes.value
+  try:
+    waitFor server.start()
+  except CatchableError:
+    return err("metrics HTTP server start failed: " & getCurrentExceptionMsg())
+  info "Metrics HTTP server started", serverIp = $serverIp, serverPort = $serverPort
+  ok(metricsServerRes.value)
+
+proc main() {.async.} =
```
vladopajic, line 31: this main proc starts here and keeps doing it's thing till there is no file left. would be nicer if some things could be broken into functions, to make it easier to read.

performance/aggregate_stats.py
```diff
@@ -0,0 +1,55 @@
...
+    total_received = 0
+    min_latency = float("inf")
+    max_latency = 0.0
+    sum_avg_latency = 0.0
+    valid_nodes = 0
+
+    for file_path in glob(os.path.join(output_dir, "*.json")):
+        with open(file_path) as f:
+            stats = json.load(f)
+        sent = int(stats.get("totalSent", 0))
```
vladopajic: python is okay choice here, but so it can be nim :)

- python does not give anything better for integration with github then any other language because you are writing to the file.
- code should be the same complexity in eater case
- python has fast setup / nim has 20-30sec cost (might be better with setup-nim github action)
- using nim keeps repo monolingual (it's just nice if we can keep it)

tld, i'am slightly more biased to use nim here

performance/runner.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+custom_network_name="my_custom_network"
+PEERS=10
+CONNECT_TO=5
+MSG_COUNT=50
+MSG_INTERVAL=100 # ms
+MSG_SIZE=100 # bytes
```
vladopajic, line 8: not really sure if we really need to pass these as arguments. and it doesn't appear that those will ever need to be parametrized, because `main.nim` will probably never be used at two places simultaneously with different params. params could change over time but only one `main.nim` instance should ever use it.

ether way, it would be nicer to have it code simplified - use `const`s in the `main.nim`. if, there ever would be reason to have arguments it would be easily extended, and for now less code the better.

performance/main.nim
```diff
@@ -0,0 +1,235 @@
...
+  var sentMessages: Table[uint64, int64] # messageId -> send time (nanoseconds)
+  var receivedMessages: Table[uint64, seq[int64]]
+    # messageId -> seq of receive times (nanoseconds)
+  var latencies: seq[int64] # ms latency per message
+  proc messageHandler(topic: string, data: seq[byte]) {.async.} =
+    # Skip warm-up messages by checking their data
+    if data == "warmup".toBytes():
+      return # warm-up phase
+    let msgId = uint64.fromBytesLE(data)
+    let sentNs = int64(msgId)
```
vladopajic, line 105: this code will be error for measuring. better get current time to variable first thing then return or process bytes?

performance/main.nim
```diff
@@ -0,0 +1,235 @@
...
+      let nowMoment = Moment.now()
+      let nowNs = epochNanoSeconds(nowMoment)
+      var nowBytes = @(toBytesLE(uint64(nowNs))) & newSeq[byte](msgSize)
+      sentMessages[uint64(nowNs)] = nowNs
+      echo "Sending message ", nowNs, " at: ", $nowMoment
+      doAssert((await gossipSub.publish(topic, nowBytes)) > 0)
+
+  # Wait for all messages to be delivered
+  echo "Waiting 2 seconds for message delivery..."
+  await sleepAsync(2.seconds)
```
vladopajic, line 194: could we wait until all messages are deliver? we know total number of messages so we can spin lock here until condition?

performance/runner.sh
vladopajic, line 8: MSG_COUNT _double it or give it to the next person_ - seems to low
MSG_SIZE not sure, but seems low

.github/workflows/performance.yml
```diff
@@ -0,0 +1,57 @@
+name: Performance
```
vladopajic, line 1: looks fancy, but thing is no one will ever look at these numbers :/

performance/main.nim
rlve: This is just a prototype of the whole workflow to review the approach. It will be improved.

performance/aggregate_stats.py
rlve: I'm only writing to the file in this PoC. The idea was to add comments to the PRs which will require using GitHub API and whatever is needed to maintain the history (which still need to be figured out).

performance/runner.sh
rlve: Sure, makes sense

performance/main.nim
rlve: Not sure what do you mean exactly, it reads the timestamp from the received message when it was sent

performance/main.nim
rlve: Sure, will be improved

performance/runner.sh
rlve: The scenario is pretty random at the moment, as it's just a prototype to showcase the workflow. Will be improved

.github/workflows/performance.yml
rlve: If the results are easier accessible then hopefully it will be useful. That's why I want to review the approach to see if it makes sense.

performance/aggregate_stats.py
rlve: Changed to Nim

performance/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Create the build image
+FROM nimlang/nim:1.6.18 AS build
```
richard-ramos, line 2: I'm somewhat confused about the versions, because in performance.yml, version 2.x is installed, and here it's 1.6.18.

.github/workflows/performance.yml
```diff
@@ -0,0 +1,90 @@
...
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+      - name: Run
+        run: |
+          ./performance/runner.sh
+
+      - name: Set up Nim
+        uses: jiro4989/setup-nim-action@v2
+        with:
```
richard-ramos, line 52: Shouldn't we use the same nim install as we do in https://github.com/vacp2p/nim-libp2p/blob/master/.github/workflows/ci.yml#L75

performance/aggregate_stats.nim
```diff
@@ -0,0 +1,73 @@
+import os, strutils, json, strformat
+
+proc getEnvOrDefault(key, default: string): string =
```
richard-ramos, line 3: getEnv already supports a default:
`getEnv(key, "the default")`

performance/aggregate_stats.nim
```diff
@@ -0,0 +1,73 @@
...
+        if stats.hasKey(key):
+          stats[key].getInt
+        else:
+          default
+
+      proc getFloatOr(stats: JsonNode, key: string, default: float): float =
+        if stats.hasKey(key):
+          stats[key].getStr.parseFloat
+        else:
+          default
```
richard-ramos, line 31: getInt and getFloat support setting a default value (see https://nim-lang.org/docs/json.html), their signature look like this:

```
proc getFloat(n: JsonNode; default: float = 0.0): float {.raises: [], tags: [],     forbids: [].}

proc getInt(n: JsonNode; default: int = 0): int {.raises: [], tags: [],
    forbids: [].}
```

If for some reason those do not work and you still need these,  they should be moved outside of main func.

performance/runner.sh
```diff
@@ -0,0 +1,54 @@
...
+
+if ! docker network inspect "$custom_network_name" >/dev/null 2>&1; then
+  docker network create "$custom_network_name"
+  docker network create --attachable --driver bridge "$custom_network_name"
```
richard-ramos, line 12: Is this creating the network twice?

performance/runner.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
```
richard-ramos, line 2:
```suggestion
set -euo pipefail
```
to make the script less forgiving

performance/runner.sh
```diff
@@ -0,0 +1,54 @@
...
+# Wait for all containers to finish
+for cname in "${container_names[@]}"; do
+    docker wait "$cname"
+done
+
+# Clean up all containers (force remove)
+for cname in "${container_names[@]}"; do
+    docker rm -f "$cname"
+done
```
richard-ramos, line 54: Should we cleanup the network we created?
```
docker network rm "$custom_network_name" || true
```

performance/main.nim
```diff
@@ -0,0 +1,235 @@
...
+  let metricsServerRes = MetricsHttpServerRef.new($serverIp, serverPort)
+  if metricsServerRes.isErr():
+    return err("metrics HTTP server start failed: " & $metricsServerRes.error)
+  let server = metricsServerRes.value
+  try:
+    waitFor server.start()
+  except CatchableError:
+    return err("metrics HTTP server start failed: " & getCurrentExceptionMsg())
+  info "Metrics HTTP server started", serverIp = $serverIp, serverPort = $serverPort
+  ok(metricsServerRes.value)
```
richard-ramos, line 29:
```suggestion
  let server = MetricsHttpServerRef.new($serverIp, serverPort).valueOr:
    return err("metrics HTTP server start failed: " & error)
  try:
    waitFor server.start()
  except CatchableError as exc:
    return err("metrics HTTP server start failed: " & $e.msg)
  info "Metrics HTTP server started", serverIp = $serverIp, serverPort = $serverPort
  ok(server)
```

performance/main.nim
```diff
@@ -0,0 +1,235 @@
...
+    myId = parseInt(getEnv("PEER_NUMBER"))
+    publisherCount = parseInt(getEnv("PEERS"))
+    connectTo = parseInt(getEnv("CONNECT_TO"))
+    msgCount = parseInt(getEnv("MSG_COUNT"))
+    msgInterval = parseInt(getEnv("MSG_INTERVAL"))
+    msgSize = parseInt(getEnv("MSG_SIZE"))
+    isPublisher = myId <= publisherCount
+    rng = libp2p.newRng()
+
+  echo "Hostname: ", hostname
```
richard-ramos, line 45: Ideally instead of `echo` we should use chronicles  (having a timestamp for a log line might be helpful if we evr have to debug delays)

.github/workflows/performance.yml
rlve: Nim in the workflow is set up only to run aggregation script, nothing else, and this action is way faster.

performance/Dockerfile
rlve: Test node is built in Dockerfile, so it's built with `nim:1.6.18`. It's a copy/paste from DST solution - can be changed if we agree to run these tests with a different version.

richard-ramos (approved):

performance/Dockerfile
richard-ramos: We have a PR to drop support for 1.6.  Pls update to at least 2.x

AlbertoSoutullo: Yup, this is what I thought as well. It should be updated to 2.x.

AlbertoSoutullo (commented):
I think the general idea is good. But from my point of view, the scope is a bit large. 
I would limit this to `d_low` number of nodes so you can publish, and test it with different and very specific setups, like multiplexer and using `perf` (that I don't see it being used here (?)).
The more concrete the better, because for a generalist performance check like this one, I have the feeling we are overlapping work with DST. And running big meshes of nodes with `perf` is something that doesn't suite DST scope, but it does suit QA in my opinion.

PR #1549, opened 2025-07-16
test(kad): Test chain-find

Ben-PH:
This PR fixes a bug where finding a node of "a friend of a friend" would try to dial to a peer it did not have a streaming connection to, without using a multiaddress, and adds a test for it.

### DIFF
--- a/tests/kademlia/testfindnode.nim
+++ b/tests/kademlia/testfindnode.nim
@@ -97,13 +97,30 @@ suite "KadDHT - FindNode":
     await neiceSwitch.start()
 
     await broKad.bootstrap(@[parentSwitch.peerInfo])
-    # TODO: assert parent only has broKad and visa versa
-    await sisKad.bootstrap(@[parentSwitch.peerInfo])
-    # TODO: assert same again, but sisKad has parent and br, and sis has been added to parent
-    await neiceKad.bootstrap(@[sisSwitch.peerInfo])
-    # TODO: assert same again, but sisKad has neice added, and neice has the same content as sis
+    # Bro and parent know each other
+    doAssert(countBucketEntries(broKad.rtable.buckets, parentKad.rtable.selfId) == 1)
+    doAssert(countBucketEntries(parentKad.rtable.buckets, broKad.rtable.selfId) == 1)
 
-    # Bro should only know parent
+    await sisKad.bootstrap(@[parentSwitch.peerInfo])
+
+    # Sis and parent know each other...
+    doAssert(countBucketEntries(sisKad.rtable.buckets, parentKad.rtable.selfId) == 1)
+    doAssert(countBucketEntries(parentKad.rtable.buckets, sisKad.rtable.selfId) == 1)
+
+    # But has been informed of bro by parent during bootstrap
+    doAssert(countBucketEntries(sisKad.rtable.buckets, broKad.rtable.selfId) == 1)
+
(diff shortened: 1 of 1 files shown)


tests/kademlia/testfindnode.nim
```diff
@@ -74,3 +74,46 @@ suite "KadDHT - FindNode":
...
+    sisSwitch.mount(sisKad)
+    await sisSwitch.start()
+
+    let neiceSwitch = createSwitch()
+    let neiceKad = KadDHT.new(neiceSwitch)
+    neiceSwitch.mount(neiceKad)
+    await neiceSwitch.start()
+
+    await broKad.bootstrap(@[parentSwitch.peerInfo])
+    # TODO: assert parent only has broKad and visa versa
```
richard-ramos: Are these TODOs going to be implemented in this PR?

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -25,11 +25,16 @@ type KadDHT* = ref object of LPProtocol
...
+    kad: KadDHT, peerId: PeerId, addrs: seq[MultiAddress], targetId: Key
 ): Future[Message] {.
     async: (raises: [CancelledError, DialFailedError, ValueError, LPStreamError])
 .} =
-  let conn = await kad.switch.dial(peerId, KadCodec)
+  var conn: Connection
+  if addrs.len == 0:
+    conn = await kad.switch.dial(peerId, KadCodec)
+  else:
+    conn = await kad.switch.dial(peerId, addrs, KadCodec)
```
richard-ramos, line 36: Is it possible for findNode to return a peer that contains no addresses? I wonder if we can just use the dial from line 34 and drop 36

Ben-PH: the first call to `sendFindNode` from `findNode` doesn't provide any addresses. just the peer id.

this snippet is a bit of a hack, admittedly, to account for this.

I'm not 100% sure on what the "clean" way to do this is, but I'm not yet "seeing the matrix" of libp2p connections enough to grok the with/without multiaddress patterns going on.

tests/kademlia/testfindnode.nim
Ben-PH: Thinking about it: "if a job ain't worth doing properly, the job ain't worth doing" applies here.

I'll get that done.

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: Ah i see. This is in `bootstrap`, right? I see that function receives a sequence of peerInfos, so you should have the multiaddresses available there. This was an oversight of my part to use the peerId only.
```
  for b in bootstrapNodes:
    try:
      await kad.switch.connect(b.peerId, b.addrs)
      debug "connected to bootstrap peer", peerId = b.peerId

      try:
        let msg = await kad.sendFindNode(b.peerId, b.addrs, kad.rtable.selfId).wait(5.seconds)
```

richard-ramos (approved):

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: it's not the bootstrap that is the issue (though I was getting errors there, but I haven't looked into it), it's that for the part of `find-node` where you have to follow up with a new shortlist member, the peerId isn't enough.

Issue #1554, opened 2025-07-17
Create onRunning


gmelodie:
Add an `asyncEvent` (`onRunning`) to the base `Transport` type that is fired when `running` is set to `true` on the `start` method

PR #1558, opened 2025-07-17
feat: add onRunning asyncEvent to Transport base type

MorganaFuture:
**What**
This PR implements the onRunning asyncEvent for the Transport base type, providing an event-driven mechanism to detect when a transport becomes ready.

**Why**
Current transport startup requires ugly polling loops to detect readiness:

```
  // Before: polling approach
  while not transport.running:
    await sleepAsync(10.milliseconds)
```

```
  // After: event-driven approach
  await transport.onRunning.wait()
```
  
  **Links**:
  Issue: [1554](https://github.com/vacp2p/nim-libp2p/issues/1554)


libp2p/transports/transport.nim
```diff
@@ -49,6 +55,8 @@ method start*(
   trace "starting transport on addrs", address = $addrs
   self.addrs = addrs
   self.running = true
+  if not isNil(self.onRunning):
+    self.onRunning.fire()
```
vladopajic: isn't `self.onRunning` always going to be initialized?

tests/testtransportonrunning.nim
```diff
@@ -0,0 +1,172 @@
...
+    # Give the event handler time to run
+    await sleepAsync(10.milliseconds)
+    
+    check:
+      eventFired == true
+      transport.running == true
+    
+    await transport.stop()
+
+  asyncTest "onRunning works with quic transport":
```
vladopajic: test kind reuse same code, could we create proc that will reuse common logic in tests?

for example could test be, like this:


```
asyncTest "onRunning with websocket":
   # setup transport here ...
   runTest(transport)
   
asyncTest "onRunning with tcptransport":
   # setup transport here ...
   runTest(transport)

... etc
```

libp2p/transports/transport.nim
gmelodie: yes, thought the same thing. if for some reason it's not we don't want to silently ignore it.

libp2p/transports/transport.nim
```diff
@@ -36,10 +36,16 @@ type
...
+    onRunning*: AsyncEvent
 
 proc newTransportClosedError*(parent: ref Exception = nil): ref TransportError =
   newException(TransportClosedError, "Transport closed, no more connections!", parent)
 
+method initTransport*(self: Transport) {.base, gcsafe, raises: [].} =
+  ## Initialize transport fields - called by derived transports
+  ##
+  self.onRunning = newAsyncEvent()
+
```
gmelodie: Instead of having a new method that is called every time `start` is called, just have `start` do `onRunning.fire()`

tests/testtransportonrunning.nim
```diff
@@ -0,0 +1,172 @@
+{.used.}
```
gmelodie: It would be best to make this test be `testtransport`, and test for:
- start sets `running` to `true` and `fire`s `onRunning`
so we could later do
- `stop` sets `running` to `false` and `fire`s `onStop` 

actually, you could go ahead and implement `onStop` on this PR too

gmelodie (dismissed):
Note: when firing `onRunning` and `onStop`, always do so after `running` has been set to `true` or `false` respectively, never before that

libp2p/transports/transport.nim
MorganaFuture: Agreed. Removed the nil checks since they're unnecessary and would mask programming errors.

tests/testtransportonrunning.nim
MorganaFuture: Accepted

libp2p/transports/transport.nim
MorganaFuture: Accepted

tests/testtransportonrunning.nim
MorganaFuture: Accepted

libp2p/transports/transport.nim
gmelodie: What I meant was: get rid of `initTransport` entirely

MorganaFuture: @gmelodie  Could you please re-loook?

tests/testtortransport.nim
```diff
@@ -162,3 +167,119 @@ suite "Tor transport":
...
+    onStopFired == false
+    transport.running == true
+
+  await transport.stop()
+
+  # Give the event handler time to run  
+  await sleepAsync(10.milliseconds)
+
+  check:
+    onRunningFired == true
```
gmelodie, line 208: I don't think we need this check

gmelodie (approved):
Very cool! Thanks for working on this!

gmelodie:
Closing in favor of #1731

Issue #1480, opened 2025-06-24
flaky test: GossipSub Integration - Control Messages::Received GRAFT for non-subscribed topic


richard-ramos:
2025-06-24T15:04:56.1203130Z =======================================
2025-06-24T15:04:56.1226060Z   /Users/runner/work/nim-libp2p/nim-libp2p/tests/testall 'GossipSub Integration - Control Messages::Received GRAFT for non-subscribed topic'
2025-06-24T15:04:56.1263270Z ---------------------------------------
2025-06-24T15:04:56.1276180Z     /Users/runner/work/nim-libp2p/nim-libp2p/tests/pubsub/integration/testgossipsubcontrolmessages.nim(95, 28): Check failed: n1.gossipsub.hasPeerId(topic, n0.peerInfo.peerId)
2025-06-24T15:04:56.1340640Z 
2025-06-24T15:04:56.1385080Z   [FAILED ] (  0.22s) Received GRAFT for non-subscribed topic[0m

rlve:
```
============================================================
  /Users/runner/work/nim-libp2p/nim-libp2p/tests/testall 'GossipSub Integration - Heartbeat::iDontWants history - last element is pruned during heartbeat'
------------------------------------------------------------
    [TIMEOUT] Timeout was reached and the conditions were not true. Check if the code is working as expected or consider increasing the timeout param.
    /Users/runner/work/nim-libp2p/nim-libp2p/tests/helpers.nim(137, 11): Check failed: peer.iDontWants[0].len == msgCount
    peer.iDontWants[0].len was 0
    msgCount was 5
    [TIMEOUT] Timeout was reached and the conditions were not true. Check if the code is working as expected or consider increasing the timeout param.
    /Users/runner/work/nim-libp2p/nim-libp2p/tests/helpers.nim(137, 11): Check failed: peer.iDontWants.mapIt(it.len) == expectedHistory
    peer.iDontWants.mapIt(it.len) was @[0, 0, 0]
    expectedHistory was @[0, 5, 0]
    [TIMEOUT] Timeout was reached and the conditions were not true. Check if the code is working as expected or consider increasing the timeout param.
    /Users/runner/work/nim-libp2p/nim-libp2p/tests/helpers.nim(137, 11): Check failed: peer.iDontWants.mapIt(it.len) == expectedHistory
    peer.iDontWants.mapIt(it.len) was @[0, 0, 0]
    expectedHistory was @[0, 0, 5]

[...]

PR #1559, opened 2025-07-18
chore(ci): initial flake.nix file with dev shell

Ben-PH:
Not much: only adds the openssl dependencies needed.

I have my nim and nimble requirements setup in my system configuration, hence why it's not in here.

For now, this is only really useful for _me_, but it can eventually becoming a powerful tool.

For example, with tho go daemon dependency, if that is put into the flake for a testing shell, anyone with a `nix` package manager (not neccisarily nixos), would be able to run `nix develop testing`, or something to that effect, and everything is automagically ready to run `nimble test`: all the dependencies (openssl, nim, nimble, go deamon, etc) are _just there_

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ git clone https://github.com/vacp2p/nim-libp2p
 cd nim-libp2p
 nimble install -dy
 ```
+You can use `nix develop` to start a shell with Nim and Nimble.
 ### Testing
 Run unit tests:
 ```sh
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1752620740,
+        "narHash": "sha256-f3pO+9lg66mV7IMmmIqG4PL3223TYMlnlw+pnpelbss=",
+        "owner": "NixOS",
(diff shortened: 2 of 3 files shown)


jakubgs (approved):

Ben-PH:
As far as I can tell, there's no reason not to merge, but I've brought it up in flaky tests issue #1480 just in case.

Thanks for putting a proper flake together.

I think it's pretty low prio for now, but support for testing under the go-daemon would be the next step, I think, though it will become priority when I need it for testing. When that time comes around, I'll give it a shot myself, and I'll ping you again for a much needed professional eye (I'm just a code-monkey when it comes to writing `.nim` files).

Don't let anyone stop you setting that up in the meantime, if that's something you feel like doing; would be cool to have around in any case.

jakubgs:
Windows tests are failing... lets try one more time.

gmelodie:
@jakubgs the coverage test failing is a typo recently introduced by https://github.com/vacp2p/nim-libp2p/pull/1448, will be fixed in https://github.com/vacp2p/nim-libp2p/pull/1562

jakubgs:
Thanks for the explanation.

I personally prefer using `rebase` instead of `merge` to avoid pointless merge commits. I'll rebase and merge.

PR #1561, opened 2025-07-18
chore: add TransportConfig

gmelodie:
This should be merged before https://github.com/vacp2p/nim-libp2p/pull/1539, which will then build on top of this to introduce `Opt`s in order to avoid `nil` checking.

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -44,8 +44,11 @@ const MemoryAutoAddress* = memorytransport.MemoryAutoAddress
 
 type
   TransportProvider* {.public.} = proc(
-    upgr: Upgrade, privateKey: PrivateKey, autotls: AutotlsService
+    upgr: Upgrade, privateKey: PrivateKey, config: TransportConfig
   ): Transport {.gcsafe, raises: [].}
+
+  TransportConfig* = ref object
+    autotls*: AutotlsService
 
   SecureProtocol* {.pure.} = enum
     Noise
@@ -160,7 +163,7 @@ proc withTransport*(
       .new()
       .withTransport(
         proc(
-            upgr: Upgrade, privateKey: PrivateKey, autotls: AutotlsService
+            upgr: Upgrade, privateKey: PrivateKey, config: TransportConfig
         ): Transport =
           TcpTransport.new(flags, upgr)
       )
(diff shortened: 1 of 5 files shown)


gmelodie:
I does modify the public API, but @vladopajic and I were discussing that users should not be calling `withTransport` directly, so we wanted to discus the possibility of changing public API in this case. @richard-ramos what's the reasoning behind `withTransport` and `TransportProvider` being public in the first place?

richard-ramos:
> what's the reasoning behind withTransport and TransportProvider being public in the first place?

Anyone could implement their own transport and not be limited by what nim-libp2p provides.

richard-ramos:
not against it, as public api stability is mostly a meme when you've got so few direct consumers. In this case, I'd just get a :+1: from @tersec and @Ivansete-status whether they'd agree to this breaking change. AFAIK, Waku is the only project that uses `withTransport` (and even then, they could switch to `withWsTransport` which is available in latest release)


But in general wrt breaking changes in the Readme says
```
## Stability
...
The versioning follows [semver](https://semver.org/), with some additions:
- Some of libp2p procedures are marked as `.public.`, they will remain compatible during each `MAJOR` version
- The rest of the procedures are considered internal, and can change at any `MINOR` version (but remain compatible for each new `PATCH`)
```
So, if it's `{.public.}` we should support both the existing version of a proc and mark it as `{.deprecated.}`, and the new version, specially if it's trivial to do so.

richard-ramos (approved):

libp2p/builders.nim
```diff
@@ -43,10 +43,16 @@ export
 const MemoryAutoAddress* = memorytransport.MemoryAutoAddress
 
 type
-  TransportProvider* {.public.} = proc(
-    upgr: Upgrade, privateKey: PrivateKey, autotls: AutotlsService
+  TransportProvider* {.deprecated: "Use TransportBuilder instead".} =
+    proc(upgr: Upgrade, privateKey: PrivateKey): Transport {.gcsafe, raises: [].}
+
+  TransportBuilder* {.public.} = proc(
+    upgr: Upgrade, privateKey: PrivateKey, config: TransportConfig
```
vladopajic: why `Upgrade` and `PrivateKey` are not part of `TransportConfig`?

gmelodie: Nice catch! I'll add them

PR #1570, opened 2025-07-22
fix: memory consumption in yamux and pubsubpeer

Ivansete-status:
## Description
This PR aims to avoid endlessly keep allocating memory when publishing. See [this](https://github.com/waku-org/nwaku/issues/3496#issuecomment-3102867775) for details.

There are some style/nitpick changes but the ones that give real impact to memory consumption are:
- Addition of `channel.clearQueues()` in yamux.nim `readOnce` methods
-  Replace `channel.sendQueue.keepItIf(not (it.fut.cancelled() and it.sent == 0))` with `channel.sendQueue.keepItIf(not it.fut.finished())`

This PR is fundamentally inspired by the following @NagyZoltanPeter 's comment: https://github.com/vacp2p/nim-libp2p/pull/1540#issuecomment-3071666756. So kudos to you my friend if we manage to merge it :partying_face: :raised_hands: 

---

## How to replicate

Follow these instructions: https://github.com/waku-org/nwaku/issues/3496#issuecomment-3102252599
( there is another approach, using waku simulator, Zoltán can give better insight on that one .)

### DIFF
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -274,7 +274,7 @@ method readOnce*(
     raise
       if channel.remoteReset:
         trace "stream is remote reset when readOnce", channel = $channel
-        newLPStreamResetError()()
+        newLPStreamResetError()
       elif channel.closedLocally:
         trace "stream is closed locally when readOnce", channel = $channel
         newLPStreamClosedError()
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -377,9 +377,9 @@ proc sendLoop(channel: YamuxChannel) {.async.} =
       inBuffer.inc(bufferToSend)
 
     trace "try to send the buffer", h = $header
-    channel.sendWindow.dec(inBuffer)
     try:
       await channel.conn.write(sendBuffer)
+      channel.sendWindow.dec(inBuffer)
     except LPStreamError as exc:
       error "failed to send the buffer", description = exc.msg
       let connDown = newLPStreamConnDownError(exc)
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -217,9 +217,12 @@ method closeImpl*(channel: YamuxChannel) {.async: (raises: []).} =
         discard
     await channel.actuallyClose()
 
-proc clearQueues(channel: YamuxChannel) =
+proc clearQueues(channel: YamuxChannel, error: ref CatchableError = nil) =
   for toSend in channel.sendQueue:
-    toSend.fut.complete()
+    if error.isNil():
+      toSend.fut.complete()
+    else:
+      toSend.fut.fail(error)
   channel.sendQueue = @[]
   channel.recvQueue.clear()
 
@@ -232,7 +235,7 @@ proc reset(channel: YamuxChannel, isLocal: bool = false) {.async: (raises: []).}
   trace "Reset channel"
   channel.isReset = true
   channel.remoteReset = not isLocal
-  channel.clearQueues()
+  channel.clearQueues(newLPStreamEOFError())
 
   channel.sendWindow = 0
   if not channel.closedLocally:
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -274,7 +274,7 @@ method readOnce*(
     raise
       if channel.remoteReset:
         trace "stream is remote reset when readOnce", channel = $channel
-        newLPStreamRemoteClosedError()
+        newLPStreamResetError()()
       elif channel.closedLocally:
         trace "stream is closed locally when readOnce", channel = $channel
         newLPStreamClosedError()
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -526,7 +526,10 @@ method close*(m: Yamux) {.async: (raises: []).} =
   trace "Closing yamux"
   let channels = toSeq(m.channels.values())
   for channel in channels:
-    channel.clearQueues(newLPStreamEOFError())
+    for toSend in channel.sendQueue:
+      toSend.fut.fail(newLPStreamEOFError())
+      channel.sendQueue = @[]
+
     channel.sendWindow = 0
     channel.closedLocally = true
     channel.isReset = true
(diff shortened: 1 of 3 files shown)


Ivansete-status:
It is in draft now until some performance analysis/comparison is made

vladopajic:
@Ivansete-status i'll measure performance, just ping me once CI is green all tests should be passing

Ivansete-status:
I just set it to draft until the CI becomes green

NagyZoltanPeter:
I have made several stress test measurements with and without the fix, to compare and see the effect of this fix on nim-waku memory consumption pattern under stress conditions and performance of publishing messages.

https://github.com/waku-org/nwaku/issues/3496#issuecomment-3104832294

libp2p/muxers/yamux/yamux.nim
```diff
@@ -337,76 +347,84 @@ proc trySend(
...
   ## Write to yamux channel
   ##
-  result = newFuture[void]("Yamux Send")
+  var resFut = newFuture[void]("Yamux Send")
+
   if channel.remoteReset:
     trace "stream is reset when write", channel = $channel
-    result.fail(newLPStreamResetError())
-    return result
+    resFut.fail(newLPStreamRemoteClosedError())
```
richard-ramos, line 409: Why change the type of error here? 

I find it curious because i do still see a mention of newLPStreamResetError() 
in when `channel.remoteReset` is `true` https://github.com/vacp2p/nim-libp2p/blob/1b6e2fffe6ad974af7f87253ae5df5cead7f0182/libp2p/muxers/yamux/yamux.nim#L280 it still continues to be LPStreamResetError

Ivansete-status: ah yes, I think this is a leftover that is not strictly needed

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -360,13 +360,18 @@ proc clearSendPriorityQueue(p: PubSubPeer) =
...
+
   try:
     await msgFut
     trace "sent pubsub message to remote", conn
-  except CatchableError as exc: # never cancelled
+  except CatchableError as exc:
+    trace "Unexpected exception", conn, description = exc.msg
+
+  ## possible error should be handled in the following manner because we use raw future
+  if msgFut.failed:
```
arnetheduck, line 371: `await` already raises in case `msgFut` holds an error - what's this code for? the use of raw futures is not relevant here, ie there's no difference between manually setting an error and having the async transformation do it.

libp2p/muxers/yamux/yamux.nim
```diff
@@ -315,17 +324,18 @@ proc gotDataFromRemote(
 proc setMaxRecvWindow*(channel: YamuxChannel, maxRecvWindow: int) =
   channel.maxRecvWindow = maxRecvWindow
 
-proc trySend(
-    channel: YamuxChannel
-) {.async: (raises: [CancelledError, LPStreamError]).} =
+proc sendLoop(channel: YamuxChannel) {.async.} =
```
arnetheduck, line 327: send loops of this type must always be declared `{.async: (raises: [].}` - if an exception escapes from them, it will crash the application.

libp2p/muxers/yamux/yamux.nim
```diff
@@ -337,76 +347,84 @@ proc trySend(
...
+    resFut.complete()
+    return resFut
+
+  channel.sendQueue.add(ToSend(data: msg, sent: 0, fut: resFut))
+
   when defined(libp2p_yamux_metrics):
     libp2p_yamux_send_queue.observe(channel.lengthSendQueue().int64)
-  asyncSpawn channel.trySend()
+
+  asyncSpawn channel.sendLoop()
```
arnetheduck, line 425: doesn't this start a new send loop for every write? is that the intent? should there be multiple send loops, each popping data from `sendQueue`?

Ivansete-status: At first, I tried to use `AsyncEvent` with one permanent `while true` loop that is continuously waiting for new messages to be sent but it didn't work very well and therefore I decided to carry on with the existing approach of spawning an async task every time.

The key point to avoid memory allocations with this approach is to make sure that every `sendLoop` ends as soon as possible.

But yes, I think the best would be having a permanent async task and not creating more on every write.

arnetheduck: what I mean is that this code looks broken the way it's implemented. If you have two send loops processing the same send queue, they will interleave their sends under pressure and send garbage to the other end.

The reason you're not really seeing this effect now in tests is that _practically_, `channel.conn.write` will queue the full buffer for sending synchronously meaning that `sendQueue` is not really needed _at all_ - this quirk is not a guarantee however and should `channel.conn.write` stop queueing everything in future releases, this code will break also.

libp2p/muxers/yamux/yamux.nim
Ivansete-status: Thanks! addressed in https://github.com/vacp2p/nim-libp2p/pull/1579

libp2p/protocols/pubsub/pubsubpeer.nim
Ivansete-status: Thanks! Addressed in https://github.com/vacp2p/nim-libp2p/pull/1579

libp2p/muxers/yamux/yamux.nim
richard-ramos: @arnetheduck , just to confirm, isn't `sendLoop` doing mostly the same what `trySend` did before? except for some small code reorg / variable renames, the function is practically the same, right?

arnetheduck: > mostly the same

didn't check tbh, just read the code - the new name kind of makes it kind of obvious though that it might be a bad idea .. think of `asyncSpawn` as launching a new thread - if two threads were to loop over the same send queue unsynchronized, you can easily understand how things can go wrong.

there are also differences wrt threads - in particular, when you call an async function, all the code _up to_ the first await always runs synchronously - if the await results in a future that is completed on the spot, it continues to run synchronously after that as well - this is good for efficiency, but it sometimes masks bugs that would happen if `await` were to actually schedule a wait - writes in particular are susceptible to this effect because you have to do a fair bit of writing to get to the point where the buffers in the lower layers are exhausted and a wait actually happens.

arnetheduck: the other notable point is that `asyncSpawn` doesn't actually do anything - it just discards the future that is returned from an async function after installing a panic handler for escaping exceptions. 

If there are indeed any memory savings in this PR, my gut feeling is that they're coming from the clearing of the buffers, and if this is true, it _likely_ means there are some circular references somewhere that keep the data alive instead of allowing it to be collected - the GC takes much longer to release circles so that's probably why any buffers in the send queue also linger - anything referenced from a circle of references remains in memory until the circle is collected and one way of optimizing for memory is hunting down said references and clearing them manually (by setting things to `nil`) or structuring the code such that is doesn't happen.

PR #1571, opened 2025-07-22
ci: add nimbus compilation daily test

gmelodie:
Daily CI run [here](https://github.com/vacp2p/nim-libp2p/actions/runs/16650249139/job/47120777049?pr=1571)

### DIFF
--- a/.github/workflows/daily_common.yml
+++ b/.github/workflows/daily_common.yml
@@ -31,7 +31,7 @@ jobs:
     steps:
       - uses: snnaplab/delete-branch-cache-action@v1
 
-  test:
+  compile_nim_libp2p:
     needs: delete_cache
     timeout-minutes: 40
     strategy:
@@ -50,7 +50,6 @@ jobs:
         nim: ${{ fromJSON(inputs.nim) }}
         cpu: ${{ fromJSON(inputs.cpu) }}
         exclude: ${{ fromJSON(inputs.exclude) }}
-
     defaults:
       run:
         shell: ${{ matrix.platform.shell }}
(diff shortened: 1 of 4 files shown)


.github/workflows/daily_common.yml
```diff
@@ -89,6 +89,21 @@ jobs:
...
+          nim --version
+          nimble --version
+
+          export NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }}"
+
+          git clone --branch unstable --single-branch https://github.com/status-im/nimbus-eth2.git
+          cd nimbus-eth2
+          git submodule set-branch --branch ${{ github.sha }} vendor/nim-libp2p
+
+          make -j"$(nproc)"
```
richard-ramos: Isn't it necessary to execute `make update`?
nimbus-build-system README says:
> Instruct your users to run make update after cloning your project, after a git pull or after switching branches.

.github/workflows/daily_common.yml
```diff
@@ -89,6 +89,21 @@ jobs:
         run: |
           nimble install -y --depsOnly
 
+      - name: Compile nimbus using nim-libp2p
+        if: ${{ matrix.platform.os == 'linux' && matrix.cpu == 'amd64' && matrix.nim.ref == 'version-2-2'}}
```
vladopajic: not sure if we should put everything into single daily job. we currently have daily tests, but we can have other daily jobs doing other things. similarly how on commit we have different jobs doing different things. 
 
- code organization gets worse: if we start putting everything into single workflow
- dependencies of jobs are not clear: if tests fails, with same specs as compile nimbus, compile nimbus will not happen
- parameters of daily test workflow as specifically designed for nim-libp2p testing
- generic failure notifications: when something fails in workflow we have to see what exactly has failed (unit tests or nimbus compile)  
- what benefits does current approach brings to outweigh split workflows?

gmelodie: I did this mostly to avoid compiling nim-libp2p twice, and thought that we wouldn't want to try and compile it against nimbus unless tests passed, thus why it's where it is

.github/workflows/daily_common.yml
gmelodie: Weird, [the build guide doesn't mention `update`](https://nimbus.guide/build.html)

.github/workflows/daily_common.yml
vladopajic: 5min a day doesn't look like a big waste of energy for this, but we get split notifications!

gmelodie: Roger! Will do that then

vladopajic (approved):

.github/workflows/daily_nimbus.yml
```diff
@@ -0,0 +1,94 @@
...
+        if: ${{ matrix.pinned_deps }}
+        run: |
+          nimble install_pinned
+
+      - name: Install dependencies (latest)
+        if: ${{ matrix.pinned_deps == false }}
+        run: |
+          nimble install -y --depsOnly
+
+
```
richard-ramos: These steps shouldnt be needed. Nimbus build system uses its own nim that it builds during `make` step. The same with dependencies, it will not use those from nimble but whatever is available in vendor/ folder of nimbus

.github/workflows/daily_nimbus.yml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Compile nimbus using nim-libp2p
+        run: |
+          nim --version
+          nimble --version
+
+          export NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }}"
```
richard-ramos: Probably unneeded because of https://github.com/status-im/nimbus-eth2/blob/stable/config.nims#L138

richard-ramos (approved):

PR #1572, opened 2025-07-22
chore: remove support for nim 1.6

gmelodie:
Since nimbus, waku and codex (which are our most important clients) no longer support nim 1.6, as, we hope, most of our users also have, we've decided to drop support for nim 1.6.

### DIFF
--- a/interop/hole-punching/Dockerfile
+++ b/interop/hole-punching/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1.5-labs
-FROM nimlang/nim:2.0.14 as builder
+FROM nimlang/nim:latest as builder
 
 WORKDIR /workspace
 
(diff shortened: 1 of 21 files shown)


interop/hole-punching/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1.5-labs
-FROM nimlang/nim:1.6.16 as builder
+FROM nimlang/nim:2.0.0 as builder
```
arnetheduck: something to keep in mind is that `x.y.0` is often more buggy than higher patch versions in the previous release series as bugfixes get backported - we generally assume "one of" the latest patch releases for any given `x.y`

gmelodie: should we use a higher 2.0 minor patch then? maybe 2.0.16?

vladopajic: we can use latest for docker images?

PR #1573, opened 2025-07-23
test(gossipsub): Performance tests - base scenario and runner

rlve:
This PR is a follow up to the https://github.com/vacp2p/nim-libp2p/pull/1544 - it contains refactored base scenario with a test runner.

**What’s included?**
- **Dockerfile**:
  - Installs dependencies and compiles the node test logic
- **Test runner**:
  - Spins up multiple Docker containers (each running a `nim-libp2p` node with the test script).
  - Cleans up containers and output files before and after each run.
- **Node test logic**:
  - Configures each GossipSub node.
  - Handles peer discovery, connection, and message publishing.
  - Measures message delivery latency (in milliseconds, with microsecond resolution) for each node.
  - Collects per-node stats: messages sent/received, latency (min/max/avg).
  - Writes results to a JSON file for aggregation after the test run.

### How to run locally
```bash
# Build docker image
docker build -f performance/Dockerfile -t test-node .

# Run the test
./performance/runner.sh 
```

### DIFF
--- a/performance/runner.sh
+++ b/performance/runner.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# Create Docker network
+network="performance-test-network"
+if ! docker network inspect "$network" > /dev/null 2>&1; then
+  docker network create --attachable --driver bridge "$network" > /dev/null
+fi
+
+# Clean up output
+output_dir="$(pwd)/performance/output"
+mkdir -p "$output_dir"
+rm -f "$output_dir"/*.json
+
+# Run Test Nodes
+container_names=()
+PEERS=10
+for ((i = 0; i < $PEERS; i++)); do
+    hostname_prefix="node-" 
+    hostname="$hostname_prefix$i"
+
+    docker run -d \
+      --name "$hostname" \
+      -e NODE_ID="$i" \
+      -e HOSTNAME_PREFIX="$hostname_prefix" \
+      -v "$output_dir:/output" \
+      --hostname="$hostname" \
+      --network="$network" \
+      test-node > /dev/null
+
+    container_names+=("$hostname")
+done
+
+# Show logs in real time for all containers
+for container_name in "${container_names[@]}"; do
+    docker logs -f "$container_name" &
+done
+
+# Wait for all containers to finish
+for container_name in "${container_names[@]}"; do
+    docker wait "$container_name" > /dev/null
+done
+
+# Clean up all containers
+for container_name in "${container_names[@]}"; do
+    docker rm -f "$container_name" > /dev/null
+done
+
+# Remove the custom Docker network
+docker network rm "$network" > /dev/null
+exit 0
--- a/performance/scenarios.nim
+++ b/performance/scenarios.nim
@@ -14,10 +14,10 @@ proc test1*() {.async.} =
     nodeCount = 10
     publisherCount = 10
     peerLimit = 5
-    msgCount = 50
-    msgInterval = 100
-    msgSize = 100
-    warmupCount = 5
+    msgCount = 200
+    msgInterval = 20 # ms
+    msgSize = 500 # bytes
+    warmupCount = 20
 
   # --- Node Setup ---
   let
@@ -50,7 +50,6 @@ proc test1*() {.async.} =
   await sleepAsync(5.seconds)
 
   # --- Peer Discovery & Connection ---
-
   var peersAddresses = resolvePeersAddresses(nodeCount, hostnamePrefix, nodeId)
   rng.shuffle(peersAddresses)
 
@@ -65,14 +64,12 @@ proc test1*() {.async.} =
     gossipSub, warmupCount, msgCount, msgInterval, msgSize, publisherCount, nodeId
   )
 
-  # Wait for all messages to be delivered
-  info "Waiting 2 seconds for message delivery..."
+  info "Waiting 2 seconds for message delivery"
   await sleepAsync(2.seconds)
 
-  # Performance summary
+  # --- Performance summary  ---
   let stats = getStats(receivedMessages[], sentMessages)
   info "Performance summary", nodeId, stats = $stats
 
-  # Write stats to JSON file in shared volume
   let outputPath = "/output/" & hostname & ".json"
   writeResultsToJson(outputPath, scenario, stats)
--- a/performance/scenarios.nim
+++ b/performance/scenarios.nim
@@ -1,0 +1,78 @@
+import metrics
+import metrics/chronos_httpserver
+import os
+import strutils
+import ../libp2p
+import ../libp2p/protocols/ping
+import ./utils
+from nativesockets import getHostname
+
+proc test1*() {.async.} =
+  const
+    # --- Scenario ---
+    scenario = "Base test"
+    nodeCount = 10
+    publisherCount = 10
+    peerLimit = 5
+    msgCount = 50
+    msgInterval = 100
+    msgSize = 100
+    warmupCount = 5
+
+  # --- Node Setup ---
+  let
+    hostnamePrefix = getEnv("HOSTNAME_PREFIX", "unknown")
+    nodeId = parseInt(getEnv("NODE_ID", "0"))
+    hostname = getHostname()
+    rng = libp2p.newRng()
+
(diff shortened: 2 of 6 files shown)


performance/scenarios.nim
```diff
@@ -0,0 +1,75 @@
+import metrics
+import metrics/chronos_httpserver
+import os
+import strutils
+import ../libp2p
+import ../libp2p/protocols/ping
+import ./utils
+from nativesockets import getHostname
+
+proc test1*() {.async.} =
```
gmelodie: Maybe use a more descriptive name than `test1`

performance/utils.nim
```diff
@@ -0,0 +1,238 @@
...
+import metrics/chronos_httpserver
+import chronos
+import json
+import ../libp2p
+import ../libp2p/protocols/pubsub/rpc/messages
+import ../libp2p/muxers/mplex/lpchannel
+import ../libp2p/protocols/ping
+
+const
+  topic* = "test"
```
gmelodie, line 18: How about `"perf test"` or `"performance test"`?

performance/utils.nim
```diff
@@ -0,0 +1,238 @@
...
+  warmupData* = "warmup".toBytes()
+
+proc msgIdProvider*(m: Message): Result[MessageId, ValidationResult] =
+  ok(($m.data.hash).toBytes())
+
+proc setupNode*(nodeId: int, rng: ref HmacDrbgContext): (Switch, GossipSub, Ping) =
+  let
+    myPort = 5000 + nodeId
```
gmelodie, line 26: Do we want to make sure that port is available (much like `listen`ing on `127.0.0.1:0` (port zero))?

performance/utils.nim
```diff
@@ -0,0 +1,238 @@
...
+    myAddress = "0.0.0.0:" & $myPort
+    address = initTAddress(myAddress)
+
+    switch = SwitchBuilder
+      .new()
+      .withAddress(MultiAddress.init(address).tryGet())
```
gmelodie, line 32: Building up on the comment above, can we do this instead?
```suggestion
    switch = SwitchBuilder
      .new()
      .withAddress(MultiAddress.init(/ip4/0.0.0.0/tcp/0).tryGet())
```

performance/utils.nim
```diff
@@ -0,0 +1,238 @@
...
+      .withRng(rng)
+      .withYamux()
+      .withMaxConnections(250)
```
gmelodie: What's the reasoning behind the max connections here?

performance/utils.nim
```diff
@@ -0,0 +1,238 @@
...
+  gossipSub.parameters.dHigh = 8
+  gossipSub.parameters.dScore = 6
+  gossipSub.parameters.dOut = 3
+  gossipSub.parameters.dLazy = 6
+  gossipSub.topicParams[topic] = TopicParams(
+    topicWeight: 1,
+    firstMessageDeliveriesWeight: 1,
+    firstMessageDeliveriesCap: 30,
+    firstMessageDeliveriesDecay: 0.9,
+  )
```
gmelodie, line 67: Can you add a comment explaining why we want these parameters specifically?

performance/utils.nim
```diff
@@ -0,0 +1,238 @@
...
+    connected = 0
+    index = 0
+  while connected < peerLimit:
+    while true:
+      let address = peersAddresses[index]
+      try:
+        let peerId =
+          await switch.connect(address, allowUnknownPeerId = true).wait(5.seconds)
+        connected.inc()
+        index.inc()
```
gmelodie, line 137: Do we want to update the index even when there's an error? If so, consider putting this on a `finally` clause

performance/scenarios.nim
```diff
@@ -0,0 +1,75 @@
...
+    nodeId,
+    address = switch.peerInfo.addrs,
+    peerId = switch.peerInfo.peerId,
+    isPublisher = nodeId <= publisherCount,
+    hostname = hostname
+  await sleepAsync(5.seconds)
+
+  # --- Peer Discovery & Connection ---
+  var peersAddresses = resolvePeersAddresses(nodeCount, hostnamePrefix, nodeId)
+  rng.shuffle(peersAddresses)
```
gmelodie, line 54: Why shuffle addresses?

performance/utils.nim
```diff
@@ -0,0 +1,238 @@
...
+    msgInterval: int,
+    msgSize: int,
+    publisherCount: int,
+    nodeId: int,
+): Future[seq[uint64]] {.async.} =
+  # Warm-up phase
+  info "Sending warmup messages", nodeId
+  for msg in 0 ..< warmupCount:
+    await sleepAsync(msgInterval)
+    discard await gossipSub.publish(topic, warmupData)
```
gmelodie, line 157: Why not publish before sleeping and not the other way around? Would this make sense on the first run?

performance/utils.nim
```diff
@@ -0,0 +1,238 @@
...
+    minLatencyMs = latencies.min
+    maxLatencyMs = latencies.max
+    var sumLatency: float = 0.0
+    for l in latencies:
+      sumLatency += l
+    avgLatencyMs = sumLatency / float(latencies.len)
+  else:
+    minLatencyMs = 0.0
+    maxLatencyMs = 0.0
+    avgLatencyMs = 0.0
```
gmelodie: Avoid else like so:

```suggestion
  var minLatencyMs: float = 0
  var maxLatencyMs: float = 0
  var avgLatencyMs: float = 0
  if latencies.len > 0:
    minLatencyMs = latencies.min
    maxLatencyMs = latencies.max
    var sumLatency: float = 0.0
    for l in latencies:
      sumLatency += l
    avgLatencyMs = sumLatency / float(latencies.len)
```

performance/utils.nim
rlve: The test run in controlled Docker network and ports from 5000 to 5009 are available.

performance/utils.nim
rlve: Can you elaborate? I'm not sure what do you mean here

performance/utils.nim
rlve: It's a copy-paste from DST code, probably doesn't apply here

performance/utils.nim
rlve: It's a copy-paste from DST code, but the values look mostly default.

performance/utils.nim
rlve: Index needs to be updated only when connection succeeded.

performance/scenarios.nim
rlve: The scenario runs in 10 containers concurrently, and the list of peer addresses to connect to is composed in the same way. The addresses are shuffled to make sure the nodes have different peers in their meshes.

performance/utils.nim
rlve: It could be the other way around, but considering the parameters of the scenario it doesn't really make a difference.

performance/utils.nim
rlve: Removed

performance/runner.sh
```diff
@@ -0,0 +1,52 @@
...
+fi
+
+# Clean up output
+output_dir="$(pwd)/performance/output"
+mkdir -p "$output_dir"
+rm -f "$output_dir"/*.json
+
+# Run Test Nodes
+container_names=()
+PEERS=10
```
fbarbu15, line 18: I see we have num perrs both here and here https://github.com/vacp2p/nim-libp2p/pull/1573/files#diff-ff964bd89ab96fc3539a4177bfe188b80b69dd4e4d1935779dc0b4b28179bf0aR14
would it be possible to have them in a common place like a config file like json/yaml/.env ?

performance/utils.nim
gmelodie: If we don't need the switch to listen on a specific port, then we don't need `myport` and consequently `myaddress`. However if we do need it to be `5000 + nodeId`, then ignore this comment

performance/utils.nim
gmelodie: I trust your judgement, but maybe we should consider specifying where these values come from. If they are "mostly default", then we should say where "default" is defined?

richard-ramos (approved):

gmelodie (approved):

PR #1579, opened 2025-07-24
fix: more secure send loop in yamux and exception-handling-code cleanup in pubsubpeer

Ivansete-status:
This PR is motivated by: 

https://github.com/vacp2p/nim-libp2p/pull/1570#discussion_r2228640214 --> Cleanup `libp2p/protocols/pubsub/pubsubpeer.nim`. There's no need to check the `failed` field in a `raw` future because `await` already extracts the exception in `try: await rawFut except`. 

https://github.com/vacp2p/nim-libp2p/pull/1570#discussion_r2228646674 --> Make `sendLoop` more secure and make sure it doesn't throw any exception

### DIFF
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -324,7 +324,7 @@ proc gotDataFromRemote(
 proc setMaxRecvWindow*(channel: YamuxChannel, maxRecvWindow: int) =
   channel.maxRecvWindow = maxRecvWindow
 
-proc sendLoop(channel: YamuxChannel) {.async.} =
+proc sendLoop(channel: YamuxChannel) {.async: (raises: []).} =
   if channel.isSending:
     return
   channel.isSending = true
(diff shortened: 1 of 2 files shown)


libp2p/muxers/yamux/yamux.nim
```diff
@@ -383,12 +383,17 @@ proc sendLoop(channel: YamuxChannel) {.async.} =
     try:
       await channel.conn.write(sendBuffer)
       channel.sendWindow.dec(inBuffer)
+    except CancelledError as exc:
+      trace "cancelled sending the buffer", description = exc.msg
+      for fut in futures.items():
+        fut.cancelSoon()
```
arnetheduck: generally, we do not cancel futures we [do not own](https://status-im.github.io/nim-chronos/concepts.html#ownership) - if a send fails because of a cancellation, we use `fail` instead if we're on the providing side of a future value - however, a cancellation can never happen here. why? because the future that `sendLoop` returned was passed to `asyncSpawn` which discards it - there is therefore no owner that can _initiate_ a cancellation, and as a consequence, this write can never be cancelled either and `except CancelledError` will never be entered.

Ivansete-status: Thanks! Addressed in https://github.com/vacp2p/nim-libp2p/pull/1579/commits/0b548feec10ea411ea886392f2ffcc6fcd41cfc9

richard-ramos (approved):

PR #1582, opened 2025-07-24
feat(kad): Initial unstable putval api

Ben-PH:


libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,46 @@
...
+type ValidatedEntry* = object
+  key*: EntryKey
+  val*: EntryVal
+  time*: TimeStamp
+
+type EntryValidator* = ref object of RootObj
+method validate*(
+    self: EntryValidator, key: EntryKey, val: EntryVal
+): bool {.base, raises: [], gcsafe.} =
+  discard
```
richard-ramos: Consider adding an assertion here, so it forces the dev to implement this method

libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,46 @@
...
+  time: TimeStamp
+
+type LocalTable* = object
+  entries*: Table[EntryKey, (EntryVal, TimeStamp)]
+
+proc init*(self: typedesc[LocalTable]): LocalTable {.raises: [].} =
+  LocalTable(entries: initTable[EntryKey, (EntryVal, TimeStamp)]())
+
+proc insert*(self: var LocalTable, val: sink ValidatedEntry) =
+  info "inserting", key = val.key.data, val = val.val.data
```
richard-ramos: This might be a bit noisy, consider changing to debug

libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,46 @@
...
+
+type RecordVal* = object
+  value: EntryVal
+  time: TimeStamp
+
+type LocalTable* = object
+  entries*: Table[EntryKey, (EntryVal, TimeStamp)]
+
+proc init*(self: typedesc[LocalTable]): LocalTable {.raises: [].} =
+  LocalTable(entries: initTable[EntryKey, (EntryVal, TimeStamp)]())
```
richard-ramos: Since we dropped 1.6 support, it is no longer necessary to call initTable from 2.0 onwards according to the docs.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +74,46 @@ proc waitRepliesOrTimeouts(
...
+
+  defer:
+    await conn.close()
+
+  let msg = Message(
+    msgType: MessageType.putValue,
+    record: some(Record(key: some(key.getBytes()), value: some(val))),
+  )
+
+  info "writing"
```
richard-ramos: This should be a trace log

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +74,46 @@ proc waitRepliesOrTimeouts(
...
+  await conn.writeLp(msg.encode().buffer)
+
+  info "waiting reply"
```
richard-ramos: Trace log

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +74,46 @@ proc waitRepliesOrTimeouts(
 
   return (receivedReplies, failedPeers)
 
+proc dispatchPutVal(
+    kad: KadDHT, peer: PeerId, key: Key, val: seq[byte]
+): Future[void] {.
+    async: (raises: [CancelledError, DialFailedError, ValueError, LPStreamError])
+.} =
+  info "dialing", me = kad.rtable.selfId, peer = peer, key = key, val = val[0 .. 3]
```
richard-ramos: Debug log or trace.
Also, might run into an index out of bounds with `val[0..3]` depending on the length of val

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +74,46 @@ proc waitRepliesOrTimeouts(
...
+    raise newException(ValueError, "unexpected message type in reply: " & $reply)
+
+proc putVal*(
+    kad: KadDHT, val: seq[byte], replic: int
+): Future[seq[PeerId]] {.async: (raises: [CancelledError]).} =
+  let keyDat: array[IdLength, byte] = sha256.digest(val).data
+  let key = Key(kind: KeyType.Hashed, data: keyDat)
+
+  let closests = kad.rtable.findClosestPeers(key, replic)
+  info "closests", closests = closests
```
richard-ramos: debug / trace

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +74,46 @@ proc waitRepliesOrTimeouts(
...
+
+  if reply.msgType != MessageType.putValue:
+    raise newException(ValueError, "unexpected message type in reply: " & $reply)
+
+proc putVal*(
+    kad: KadDHT, val: seq[byte], replic: int
+): Future[seq[PeerId]] {.async: (raises: [CancelledError]).} =
+  let keyDat: array[IdLength, byte] = sha256.digest(val).data
+  let key = Key(kind: KeyType.Hashed, data: keyDat)
+
```
richard-ramos, line 109: In kademlia specs (the OG kademlia), something that is mentioned wrt "STORE" rpc method: "to ensure the persistence of the information, nodes periodically republish files". "STORE" is the equivalent to "PUT_VALUE" in libp2p kademlia. We probably have to add this behavior too because due to network churn, nodes in which we published a value might end up dissapearing.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -182,6 +238,21 @@ proc new*(
 
           # Peer is useful. adding to rtable
           discard kad.rtable.insert(conn.peerId)
+        of MessageType.putValue:
+          var record = msg.record.get()
```
richard-ramos: Let's use valueOr, unless you're 100% sure a msg will contain a record. Same in lines 243 and 244

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -182,6 +238,21 @@ proc new*(
...
+          let key = EntryKey(data: record.key.get())
+          let val = EntryVal(data: record.value.get())
+          let ts = TimeStamp(ts: $times.now().utc)
+          info "ts", ts = ts
```
richard-ramos: debug/trace

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +74,46 @@ proc waitRepliesOrTimeouts(
...
+  let key = Key(kind: KeyType.Hashed, data: keyDat)
+
+  let closests = kad.rtable.findClosestPeers(key, replic)
+  info "closests", closests = closests
+  try:
+    for peer in closests:
+      info "dispatching", peer = peer, key = key, val = val
+      await kad.dispatchPutVal(peer, key, val)
+  except CatchableError as e:
+    error "todo: think about handling error properly", e = e.msg
```
richard-ramos: Something to think about:
What is the criteria to use to consider a value to be published succesfully? i.e. if you have 10 peers, and only 1 accepted the put value, should we consider this a success? 9 other peers could have failed due to either them not accepting the connection (they went offline / no conn slots available / etc), or they just simply did not accept the value (not sure if they'd do a stream reset in this case as i did not see a response message in specs).

libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,46 @@
...
+): bool {.base, raises: [], gcsafe.} =
+  discard
+
+# get a validated entry from a app defined validator and 
+
+type RecordVal* = object
+  value: EntryVal
+  time: TimeStamp
+
+type LocalTable* = object
```
richard-ramos, line 52: Will values be stored permanently in the local table? or will there be a TTL associated to them? (maybe the timestamp is intended to be used to determine which values to remove? In that case maybe a structure similar to the PreambleStore can be used (i.e. table + minheap)? (https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub/preamblestore.nim#L31C24-L74)

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: Left over from dev. I think it can go, or would you prefer it to stay, with more info detail, as a trace?

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: same as https://github.com/vacp2p/nim-libp2p/pull/1582#discussion_r2242572456

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: OG kademlia specs also talk about replication, in which nodes which have stored values, also publish the values to closest nodes.  Both republishing and replicating are done at 1h intervals. Worth taking a look into this

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: My hunch is that this is an application-specific thing. one app could be "well, our replic value is 10, but I'd be happy with 4 if I have high confidence in the nodes that succesfully stored"

This implies that `putVal` would have to expert telemetry. It's not up to nim-libp2p to define success, but it must provide the information the user needs to make that decision.

Different applications, different policies. In my mind, the specification is limited to what it will/won't/might provide in order to design/enforce policy at the application level.

libp2p/protocols/kademlia/dhttypes.nim
Ben-PH: For now: indefinately.

I'll have to look into the spec to see what the standard is, however my feeling is that TTL is the domain of application policy.

I envisage some sort of standard way to provide a kad instance with a policy, and there would be some sort of mechanism to either trigger a specific policy enforcement operation, or a way to batch-process policy enforcement.

I believe that's thinking too far ahead at this stage, though. For now, we have TS to track data-livliness, and as we assemble the different peices of kad, we will grok what's needed. As various flavors of policy setting/enforcement comes into play, we can adapt to these needs as they surface.

libp2p/protocols/kademlia/dhttypes.nim
Ben-PH: > Starting from Nim v0.20, tables are initialized by default and it is not necessary to call this function explicitly.

> initialized by default

> default

I don't like implicit behavior. Silent default initialization is one of the less insidious forms of implicit behavior, but it's still silent and implicit behavior.

The concept of "normalising undesired behavior" is given a lot of attention in high-stakes industries, such as aviation, medicine, civil engineering, etc. Any normalisation step that takes the direction of an increasing the risk of undesired state, is a precedent for the next step. The earlier that chain is broken, the better.

I would like to push back  on the basis that being explicit about initialization is a way of drawing a line in the sand, and not allowing this PR to become a precedent for another PR down the line that further extends the acceptability of silent, implicit behavior.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I've made a TODO to format the data in a log-friendly form. As this PR moves towards non-draft, the various TODOs will be address/shifted to follow-up PRs, etc.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: getting rid of.

libp2p/protocols/kademlia/dhttypes.nim
Ben-PH: resolved: [08c994b](https://github.com/vacp2p/nim-libp2p/pull/1582/commits/08c994b32af9064ffdcdb19fff8c0fbce7dc8ed7)

(let me know if there's a standard format to fail with)

libp2p/protocols/kademlia/dhttypes.nim
Ben-PH: [08c994b](https://github.com/vacp2p/nim-libp2p/pull/1582/commits/08c994b32af9064ffdcdb19fff8c0fbce7dc8ed7)

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: [08c994b](https://github.com/vacp2p/nim-libp2p/pull/1582/commits/08c994b32af9064ffdcdb19fff8c0fbce7dc8ed7)

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: [08c994b](https://github.com/vacp2p/nim-libp2p/pull/1582/commits/08c994b32af9064ffdcdb19fff8c0fbce7dc8ed7)

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: [08c994b](https://github.com/vacp2p/nim-libp2p/pull/1582/commits/08c994b32af9064ffdcdb19fff8c0fbce7dc8ed7)

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: same as https://github.com/vacp2p/nim-libp2p/pull/1582#discussion_r2242685948 (set up a TODO)

[08c994b](https://github.com/vacp2p/nim-libp2p/pull/1582/commits/08c994b32af9064ffdcdb19fff8c0fbce7dc8ed7)

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: [08c994b](https://github.com/vacp2p/nim-libp2p/pull/1582/commits/08c994b32af9064ffdcdb19fff8c0fbce7dc8ed7)

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: >  I think it can go, or would you prefer it to stay, with more info detail, as a trace?

if we can drop the log line because it is not essential, let's nuke it. Less log spam is always better.

libp2p/protocols/kademlia/kademlia.nim
arnetheduck: In general, when you reason about distributed pubsub systems, you assume success as soon as a value has been accepted locally, without exposing too much information about "how many peers" and similar metrics.

Broadcast, and by extension kad, just like UDP, is fundamentally modelled as lossy without taking into account whether an update was sent to any specific number of peers - instead, one relies on the local database being up to date and let randomness and recovery protocols deal with failures.

As such, put should validate that the key/value fits the local acceptance criteria and _attempt_ to update remotes per spec, but not tell the application about it.

Applications that wish to measure success should be guided towards implementing their own acknowledgement schemes which is the only real way of knowing whether a message reached a certain number of peers - without an ACK, you simply don't know and providing api of the "successfully tried to send to N peers" kind tends to lead to flawed assumptions in the app consuming the api and not much else.

richard-ramos: interesting perspective, i think it makes sense.

A bit offtopic: based on this criteria you shared, do you think we should move towards changing the API of the publish method of pubsub? Currently looks like this: 
```
method publish*(  
    topic: string,
    data: seq[byte],
    publishParams: Option[PublishParams] = none(PublishParams),
): Future[int] {.async: (raises: []).}
```
returning the number of peers it chose to publish the message, but maybe it should not?

arnetheduck: Indeed - I think it was a mistake to put it there to begin with as it's fairly meaningless and can be misleading for application developers as seen in #1243.

See also https://github.com/vacp2p/nim-libp2p/pull/1243#discussion_r1963127988 and https://github.com/vacp2p/nim-libp2p/pull/1243#issuecomment-2639508010 - _perhaps_ the best thing we can do with it now that it's there is to redefine what it mean:

* 0 if the message was already in the seen cache, 1 otherwise (to indicate that it was accepted as new and will likely be synced with peers _eventually_ through any of the available means like mesh/ihave/etc) - this distinction is mostly interesting for a subset of applications really that include provenance in their message id function, since in the more general case of pubsub, two peers can publish the same message independently and their relative ordering or source tends to be meaningless. One consequence of this point is that when you publish, the message should usually be "propagated" to the local node itself iff it is subscribed to the topic
* number of peers in the mesh (I don't like this option really, because it's not much more meaningful than "number of peers I tried to send it to")
* remove it entirely (messy backwards compat)

The reasoning is similar to what is presented in that PR - ie `publish` cannot fail in any meaningful way except by programmer error (ie publishing a message for which we cannot generate a message id) - all other cases are "benign", including re-publishing the same message twice and so on.

For statistics, Q/A, monitoring and so on, these are best done out-of-band with respect to an operation like "putkey" or indeed "publish".

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: We could  add this to the `KadDHT` obj:

```nim
  ## Not part of libp2p spec, but part af the original kad spec: republish and replicating locally stored values at a specified interval (default 1h)
  valReplicationLoop: Option[Future[void]]
```

richard-ramos: Went over libp2p specs, and it seems that replication / republishing is part of the responsability of providers according to https://github.com/libp2p/specs/blob/master/kad-dht/README.md#content-provider-advertisement-and-discovery, so this behavior I wrote about does not need to be part of PUT_VALUE.

libp2p/protocols/kademlia/dhttypes.nim
richard-ramos: Went over the specs and found out that stale record GC is done as part of provider responsabilities https://github.com/libp2p/specs/blob/master/kad-dht/README.md#content-provider-advertisement-and-discovery so no need to take this into account for this PR

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +76,48 @@ proc waitRepliesOrTimeouts(
...
+
+  if reply.msgType != MessageType.putValue:
+    raise newException(ValueError, "unexpected message type in reply: " & $reply)
+
+proc putVal*(
+    kad: KadDHT, val: seq[byte], replic: int
+): Future[seq[PeerId]] {.async: (raises: [CancelledError]).} =
+  let keyDat: array[IdLength, byte] = sha256.digest(val).data
+  let key = Key(kind: KeyType.Hashed, data: keyDat)
+
```
richard-ramos, line 109: Something to consider is doing here is checking in the local table    if there is an old value already, and it’s not equal to the new one, use the `Select` from the validator (as defined in the spec) to pick the determine whether the old value is prefered to the new one. If the new value is not selected because older is "better", then we can return an error indicating that the value cannot be replaced.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +76,48 @@ proc waitRepliesOrTimeouts(
...
+  await conn.writeLp(msg.encode().buffer)
+
+  let reply = Message.decode(await conn.readLp(MaxMsgSize)).tryGet()
+
+  if reply.msgType != MessageType.putValue:
+    raise newException(ValueError, "unexpected message type in reply: " & $reply)
+
+proc putVal*(
+    kad: KadDHT, val: seq[byte], replic: int
+): Future[seq[PeerId]] {.async: (raises: [CancelledError]).} =
```
richard-ramos: putVal does not need to return a list of peerIds. It should be mostly a fire and forget function

libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,60 @@
...
+  key*: EntryKey
+  val*: EntryVal
+
+type ValidatedEntry* = object
+  key*: EntryKey
+  val*: EntryVal
+
+type RecordVal* = object
+  val*: EntryVal
+  time*: TimeStamp
```
Ben-PH: All this is public for dev purposes. will need to make things private as appropriate

the separation of candidate vand valid entry is to do with the typestate pattern. the idea being is that the type being used encapsulates the invariants being upheld.

im still depating how to approach enforcing valid entries

libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,60 @@
...
+
+## Top tip: add chronicles logs to your implementation
```
Ben-PH, line 34: I feel like I don't have as good of a big picture of how this library is used, and the other factors that go into adding high quality logging events. Level decisions, scope decisions, what is useful and why, etc. I can guess, but I fear I risk compromising the signal to noise ratio in logs

libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,60 @@
...
+): bool {.base, raises: [], gcsafe.} =
+  doAssert(false, "unimplimented base method")
+
+type EntrySelector* = ref object of RootObj
+method select*(
+    self: EntrySelector, cand: RecordVal, others: seq[RecordVal]
+): RecordVal {.base, raises: [], gcsafe.} =
+  doAssert(false, "EntrySelection base not implemented")
+
+# TODO: make library public, but hidden to users of library
```
Ben-PH: I'm not well versed in the idiomatic nim way of scoping items. Feedback from those familiar with this in nim is needed to do this properly.

libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,60 @@
...
+): ValidatedEntry {.raises: [].} =
+  ValidatedEntry(key: entry.key, val: entry.val)
+
+type LocalTable* = object
+  entries*: Table[EntryKey, RecordVal]
+
+proc init*(self: typedesc[LocalTable]): LocalTable {.raises: [].} =
+  LocalTable(entries: initTable[EntryKey, RecordVal]())
+
+# TODO: make library public, but hidden to users of library
```
Ben-PH: Same as previous comment regarding module scoping

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -1,19 +1,24 @@
...
 import ../protocol
-import ../../switch
 import ./protobuf
+import ../../switch
+import ../../multihash
 import ../../utils/heartbeat
-import std/[options, tables]
+import std/[times, options, tables]
 import results
+# TODO: Lots of `.get()` that don't consider the `none` possibility, as reviewed here:
```
Ben-PH: resolve before RFR

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -23,6 +28,9 @@ type KadDHT* = ref object of LPProtocol
   rng: ref HmacDrbgContext
   rtable*: RoutingTable
   maintenanceLoop: Future[void]
+  dataTable*: LocalTable
+  entryValidator*: EntryValidator
+  entrySelector*: EntrySelector
```
Ben-PH: make private as appropriate

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +76,61 @@ proc waitRepliesOrTimeouts(
 
   return (receivedReplies, failedPeers)
 
+proc dispatchPutVal(
+    kad: KadDHT, peer: PeerId, entry: ValidatedEntry
+): Future[void] {.
+    async: (raises: [CancelledError, DialFailedError, ValueError, LPStreamError])
```
Ben-PH: get rid of ValueError before RFR

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +76,61 @@ proc waitRepliesOrTimeouts(
...
+
+proc putVal*(
+    kad: KadDHT, key: keys.Key, val: EntryVal, replic: int
+): Future[Result[void, string]] {.async: (raises: [CancelledError]), gcsafe.} =
+  let cand = EntryCandidate(key: EntryKey(data: key.getBytes()), val: val)
+  if not kad.entryValidator.validate(cand):
+    return err("invalid key/val pair")
+  let ts = TimeStamp(ts: $times.now().utc)
+  try:
+    let others: seq[RecordVal] =
```
Ben-PH, line 111: This still logs an error in tests: `ERR 2025-08-05 11:12:03.195+02:00 could not send find_node to peer           topics="kad-dht" tid=214810 peerId=12D*W9z11w err="failed new dial: internalConnect can't dial self!"`

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +76,61 @@ proc waitRepliesOrTimeouts(
...
+      if cand.key in kad.dataTable.entries:
+        @[kad.dataTable.entries[cand.key]]
+      else:
+        @[]
+
+    let candAsRec = RecordVal(val: cand.val, time: ts)
+    let confirmedRec = kad.entrySelector.select(candAsRec, others)
+    # let confirmedCand = EntryCandidate(key: cand.key, val: confirmedRec.val)
```
Ben-PH: clear before RFR

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -91,7 +154,7 @@ proc findNode*(
 
       pendingFutures[peer] = kad
         .sendFindNode(peer, addrTable.getOrDefault(peer, @[]), targetId)
-        .wait(5.seconds)
+        .wait(chronos.seconds(5))
```
Ben-PH, line 172: Required as both times and chronos has seconds imported. either import peicemeal, or keep as is

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -182,6 +257,28 @@ proc new*(
...
+          let val = EntryVal(data: sval)
+          let ts = TimeStamp(ts: $times.now().utc)
+          record.timeReceived = some(ts.ts)
+
+          # TODO: thunderdome the entries
+          if kad.entryValidator.validate(EntryCandidate(key: key, val: val)):
+            let validated = ValidatedEntry(key: key, val: val)
+            kad.dataTable.insert(validated, ts)
+          let resp = Message(record: some(record))
+          await conn.writeLp(resp.encode().buffer)
```
Ben-PH: probably want this in helper function

libp2p/protocols/kademlia/keys.nim
```diff
@@ -8,14 +9,25 @@ type
...
+proc sha256Hash*(k: Key): array[IdLength, byte] =
+  return
+    case k.kind
+    of KeyType.PeerId:
+      sha256.digest(k.peerId.getBytes()).data
+    of KeyType.Raw, KeyType.Unhashed:
+      sha256.digest(k.data).data
+    of Hashed:
+      k.data
+
```
Ben-PH: probably not necisary,

libp2p/protocols/kademlia/keys.nim
```diff
@@ -36,7 +48,7 @@ proc getBytes*(k: Key): seq[byte] =
     case k.kind
     of KeyType.PeerId:
       k.peerId.getBytes()
-    of KeyType.Raw, KeyType.Unhashed:
+    of KeyType.Raw, KeyType.Unhashed, KeyType.Hashed:
```
Ben-PH: possibly not neccisary. if it is, probably best for another pr, as what a hashed key looks like should probably be defined by the application, making this feature non-trivial enough to be its own efforts

libp2p/protocols/kademlia/xordistance.nim
```diff
@@ -31,15 +31,15 @@ proc `<`*(a, b: XorDistance): bool =
 proc `<=`*(a, b: XorDistance): bool =
   cmp(a, b) <= 0
 
-proc hashFor(k: Key): seq[byte] =
+proc hashFor*(k: Key): seq[byte] =
```
Ben-PH: identify why this was made public, then update this comment accordingly,

tests/kademlia/testputval.nim
```diff
@@ -0,0 +1,147 @@
...
+
+proc countBucketEntries(buckets: seq[Bucket], key: Key): uint32 =
+  var res: uint32 = 0
+  for b in buckets:
+    for ent in b.peers:
+      if ent.nodeId == key:
+        res += 1
+  return res
+
+type PermissiveValidator = ref object of EntryValidator
```
Ben-PH: this, and selector could be moved to utilities. doc comment to indicate that usually its the responsibility of the application to define the validator and selecto

tests/kademlia/testputval.nim
```diff
@@ -0,0 +1,147 @@
...
+method select(
+    self: ApatheticSelector, cand: RecordVal, others: seq[RecordVal]
+): RecordVal =
+  return cand
+
+method validate(self: RestrictiveValidator, cand: EntryCandidate): bool =
+  info "validating false"
+  false
+
+suite "KadDHT - PutVal":
```
Ben-PH, line 38: this passes 100%* due to replication factor of 2, meaning that all nodes will be getting the entry. with a rep factor of 1, the test is flakey. Can't say for sure, but current working hypothesis is that peer ids are rng'd, and selection of putval targets is based on peer id in this test. this selection doesn't always make the choice we expect.

Subject te review feedback, make GH issue to follow up before merging. 

This probably relates to the k-buckets divergence from spec that was rased in initial kad-dht PR

Ben-PH (commented):
The main body of work is complete. it's not optimised for code quality at time of writing, and there are a couple of architectural dev-conveniences that need to be corrected, but for the most part, it's in a state that represents the merge-goal.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: https://github.com/libp2p/specs/issues/691#issuecomment-3154082301

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: ...putting it in a helper couses problems with the record. it needs to be changed, but to pass it in as var causes compiler errors

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -94,15 +153,18 @@ proc findNode*(
...
       state.activeQueries.inc
 
     let (successfulReplies, timedOutPeers) = await waitRepliesOrTimeouts(pendingFutures)
 
     for msg in successfulReplies:
       for peer in msg.closerPeers:
-        addrTable[PeerId.init(peer.id).get()] = peer.addrs
+        let pid = PeerId.init(peer.id)
+        if not pid.isOk:
+          continue
```
richard-ramos, line 183: This shouldnt really happen. Probably makes sense to log the err

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +73,60 @@ proc waitRepliesOrTimeouts(
...
+  )
+
+  await conn.writeLp(msg.encode().buffer)
+
+  let reply = Message.decode(await conn.readLp(MaxMsgSize)).valueOr:
+    # todo log this more meaningfully
+    error "putval reply decode fail", error = error
+    return
+
+  if reply.msgType != MessageType.putValue:
```
richard-ramos: Since the reply to putValue is an echo of the request, maybe you can compare the response:  `if msg == reply:` instead of comparing individual values

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +73,60 @@ proc waitRepliesOrTimeouts(
...
+    error "unexpected message type in reply: ", msg = msg, reply = reply
+
+proc putVal*(
+    kad: KadDHT, key: keys.Key, val: EntryVal, replic: int
+): Future[Result[void, string]] {.async: (raises: [CancelledError]), gcsafe.} =
+  let cand = EntryCandidate(key: EntryKey(data: key.getBytes()), val: val)
+  if not kad.entryValidator.validate(cand):
+    return err("invalid key/val pair")
```
richard-ramos: Not needed for now, but looking at this it makes me wonder if `validate` should return a `Result[void, string]` so the error we return here could contain the actual reason why validation failed?

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +73,60 @@ proc waitRepliesOrTimeouts(
...
+    error "putval reply decode fail", error = error
+    return
+
+  if reply.msgType != MessageType.putValue:
+    error "unexpected message type in reply: ", msg = msg, reply = reply
+
+proc putVal*(
+    kad: KadDHT, key: keys.Key, val: EntryVal, replic: int
+): Future[Result[void, string]] {.async: (raises: [CancelledError]), gcsafe.} =
+  let cand = EntryCandidate(key: EntryKey(data: key.getBytes()), val: val)
```
richard-ramos: Is there an advantage in using `EntryKey` here, compared to EntryCandidate use a `keys.Key` directly? This line would look like this:
```
let cand = EntryCandidate(key: key, val: val)
```

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +73,60 @@ proc waitRepliesOrTimeouts(
...
+  let reply = Message.decode(await conn.readLp(MaxMsgSize)).valueOr:
+    # todo log this more meaningfully
+    error "putval reply decode fail", error = error
+    return
+
+  if reply.msgType != MessageType.putValue:
+    error "unexpected message type in reply: ", msg = msg, reply = reply
+
+proc putVal*(
+    kad: KadDHT, key: keys.Key, val: EntryVal, replic: int
```
richard-ramos: Consider renaming val to `value`.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +73,60 @@ proc waitRepliesOrTimeouts(
...
+): Future[Result[void, string]] {.async: (raises: [CancelledError]), gcsafe.} =
+  let cand = EntryCandidate(key: EntryKey(data: key.getBytes()), val: val)
+  if not kad.entryValidator.validate(cand):
+    return err("invalid key/val pair")
+  let ts = TimeStamp(ts: $times.now().utc)
```
richard-ramos, line 108: `times.now().utc` already returns a DateTime, couldnt that type be used directly without requiring to convert it to string at this point?  the actual string format seems to be a concern at the moment of encoding/decoding the protobuffer, and having it as a string here seems like the wrong place

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: let's use a default for `replic` as most of the time users from the lib should not deviate from the defaults, and also lets rename it into something easier to read? Maybe something like this:
```
replicationFactor: int = DefaultReplic
```
(`DefaultReplic` comes from consts.nim)

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +73,60 @@ proc waitRepliesOrTimeouts(
...
+    let confirmedRec = kad.entrySelector.select(candAsRec, others)
+
+    let confirmedEnt = EntryCandidate(key: cand.key, val: confirmedRec.val)
+    if not kad.entryValidator.validate(confirmedEnt):
+      return err("invalid overruled entry")
+    let valEnt = ValidatedEntry.take(confirmedEnt)
+
+    let peers = kad.rtable.findClosestPeers(key, replic)
+    let putClip = peers.mapIt(kad.dispatchPutVal(it, valEnt))
+    kad.dataTable.insert(valEnt, ts)
```
richard-ramos: Consider moving `kad.dataTable.insert(valEnt, ts)` before line 121.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +73,60 @@ proc waitRepliesOrTimeouts(
...
+
+  let reply = Message.decode(await conn.readLp(MaxMsgSize)).valueOr:
+    # todo log this more meaningfully
+    error "putval reply decode fail", error = error
+    return
+
+  if reply.msgType != MessageType.putValue:
+    error "unexpected message type in reply: ", msg = msg, reply = reply
+
+proc putVal*(
```
richard-ramos: Consider renaming putVal to putValue, so it matches the protobuffer enum name

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +73,60 @@ proc waitRepliesOrTimeouts(
...
+
+    let candAsRec = RecordVal(val: cand.val, time: ts)
+    let confirmedRec = kad.entrySelector.select(candAsRec, others)
+
+    let confirmedEnt = EntryCandidate(key: cand.key, val: confirmedRec.val)
+    if not kad.entryValidator.validate(confirmedEnt):
+      return err("invalid overruled entry")
+    let valEnt = ValidatedEntry.take(confirmedEnt)
+
+    let peers = kad.rtable.findClosestPeers(key, replic)
```
richard-ramos: In here we should not use `findClosestPeers`. That will use only the records that you have on the routing table. We should use instead `kad.findNode` so a lookup is performed.
See spec:
> To put a value the DHT finds k or less closest peers to the key of the value using the FIND_NODE RPC (see [peer routing section](https://github.com/libp2p/specs/blob/master/kad-dht/README.md#peer-routing)), and then sends a PUT_VALUE RPC message with the record value to each of the peers.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +73,60 @@ proc waitRepliesOrTimeouts(
...
+  let ts = TimeStamp(ts: $times.now().utc)
+  try:
+    let others: seq[RecordVal] =
+      if cand.key in kad.dataTable.entries:
+        @[kad.dataTable.entries[cand.key]]
+      else:
+        @[]
+
+    let candAsRec = RecordVal(val: cand.val, time: ts)
+    let confirmedRec = kad.entrySelector.select(candAsRec, others)
```
richard-ramos: `select` is just returning a record here, but you do not have an indicator whether the value you're returning is actually better. That means that it's possible that `select` here returns an older value, because for some reason the newer is not better. If an old value is returned then it is not necessary to proceed with adding the value to the local table nor doing a PUT_VALUE broadcast. 

(In the spec, it has this signature `Select(key string, values [][]byte) (int, error)`, returns an `int`, and that would let you determine whether your value is the best (`i` would point to the best value in the sequence of possible values to select) )

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -162,35 +225,75 @@ proc refreshBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
+              (record.key.unsafeGet(), record.value.unsafeGet())
+            else:
+              raise newException(CatchableError, "no key or no value in rpc buffer")
+          let key = EntryKey(data: skey)
+          let val = EntryVal(data: sval)
+          let ts = TimeStamp(ts: $times.now().utc)
+          record.timeReceived = some(ts.ts)
+
+          # TODO: thunderdome the entries
+          if kad.entryValidator.validate(EntryCandidate(key: key, val: val)):
```
richard-ramos: Also missing the `Select`

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -162,35 +225,75 @@ proc refreshBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
-  heartbeat "refresh buckets", 10.minutes:
+  heartbeat "refresh buckets", chronos.minutes(10):
     await kad.refreshBuckets()
 
 proc new*(
-    T: typedesc[KadDHT], switch: Switch, rng: ref HmacDrbgContext = newRng()
+    T: typedesc[KadDHT],
+    switch: Switch,
+    validator: EntryValidator,
+    entrySelector: EntrySelector,
```
richard-ramos, line 252: This is fine for now, but we need to come up with some defaults for validator and entrySelector that would let you use kad-dht out of the box.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: the reply can be different. ive added the time received (could probably do to remove that?), and if the entry value changes at destination, the reply changes to reflect this change.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: the canonical representation of the entry key is just a raw sequence of bytes. the use of EntryKey is more along the lines of distinct type to constrain byte sequences intended as entry key to only be used as entry key. the `Key` type is more general in use and representation. it can be changed between peerid, has various types, etc.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I considered that approach. The reason why I went with this one, is that I'm not such a fan of default function arguments. It creates hidden variability in what the function does, and becomes a bit of an easter egg, in a way, though i do agree that a default is called for.

How do you feel about making it an Option, and if none is passed in, then it will be interpreted as DefaultReplic?

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: dispatching a put values should be borrowing the values, then copying them into the data buffer for the transport socket. insert takes the value as `sink`, meaning any future uses involve needless copies.

it's a bit of a micro-optimisation, which in itself doesn't justify this pattern on its own. It does, however, more faithfully represents the flow of data and the use of memory. Slightly less readable at face value, however if the intent is succesful, it makes it more readable in terms of understanding the behavior of the data.

That's the idea at least. What matters is how it lands with other readers. Do you think that with a comment it might come across as intended?

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: Select is to be defined by the library user, similar to validation. The value that is returned is the value to be placed, and if the wrong value is placed, that is a bug in the definition of the selector.

That can easily be changed so that the definition of select is in-library, though.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: oops. thanks for catching that. will have to add tests.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: righto. can mark this as resolved when this msg is edited to point to a follow up issue.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: why isn't the thinking emoji part of the reaction choices?

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: DateTime is a pretty bloated data structure last i checked. a u64 in miliseconds since the 1970 epoch is the ideal, I think. If you look at dhttypes.nim where our timestamp is defined, there is a comment to upgrade.

as the inner format is black boxed away, it shouldn't be consequential to the behavior, and we can find the ideal storage form in a follow up PR. how do you feel about that?

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: given that the validator is provided at the application level, it kinda makes sense to set the return type to be T, so long as T can somehow be generically used in a boolean check.

so the user cold just have the type be Foo, and so long as Foo can be used in an `if` check,' that's all good. it would be up to the user to define how to handle extra errors for example.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: with that said, the current type design is pretty bad. I do think there needs to be some sort of typestate/system that constrains code such that only correct code is valid code.

libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,60 @@
...
+  key*: EntryKey
+  value*: EntryVal
+
+type ValidatedEntry* = object
+  key*: EntryKey
+  value*: EntryVal
+
+type RecordVal* = object
+  value*: EntryVal
+  time*: TimeStamp
```
Ben-PH, line 32: This is just overall bad for now. the idea is to use the typestate pattern such that the only valid code that can compile is code that is logically correct.

Still have to think about what form the pattern should take, though.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I still use findClosestPeers, but i preceed it with a FindNode. Find node to update the rtable specific to the putval to be done, and findClosestPeers to use the rtable to get all possible peers, and to limit the count to the replication value.

Is this appropriate, or do you interpret the spec that the target peers must be a function of the find node RPC. This approach defines the targets as a set of nodes that is a superset to the RPC, then truncated to a list of the closest nodes, size of replication value.

libp2p/protocols/kademlia/kademlia.nim
gmelodie: Maybe we could have a type `PublishResponse` with more information (peers sent to, peers that received the message, etc), much like [`ValidatorResult`](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/errors.nim). I don't see much use in not returning something that could be valuable for someone. We can just have a type and, in case someone needs more information, we add it to the type.

libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,60 @@
...
+
+type EntryKey* = object
+  data*: seq[byte]
+
+type EntryVal* = object
+  data*: seq[byte]
+
+type TimeStamp* = object
+  # Currently a string, because for some reason, that's what is chosen at the protobuf level
+  # TODO: convert between RFC3339 strings and use of integers (i.e. the _correct_ way)
```
gmelodie, line 19: Maybe we could fork https://github.com/Skrylar/rfc3339 into `vacp2p` and give the code a bit of brushing and maintain it. At first I thought this would be a lot of work but looking at the code base it's just one file and since the format appears in other places that could benefit us I think we could do that.

libp2p/protocols/kademlia/kademlia.nim
gmelodie: I'll +1 @richard-ramos's comment. I think it's best not having to create (and maintain/deal with) a new `TimeStamp` type. Let's just use what everyone else is used to in Nim. The "bloating" here is not of great impact.

libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,60 @@
...
+  ts*: string
```
gmelodie, line 20: See [comment below](https://github.com/vacp2p/nim-libp2p/pull/1582/files#r2257309895)

libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,60 @@
...
+  key*: EntryKey
+  value*: EntryVal
+
+type RecordVal* = object
+  value*: EntryVal
+  time*: TimeStamp
+
+## Top tip: add chronicles logs to your implementation
+type EntryValidator* = ref object of RootObj
+method validate*(
```
gmelodie: maybe
```suggestion
method isValid*(
```

so we could call it in `if` statements like
```nim
if ...isValid:
# or
if not ...isValid:
```

libp2p/protocols/kademlia/kademlia.nim
gmelodie: Import peicemeal (idk what that means but guessing it means we get `5.seconds` in the end which is what I'd like to see rofl)

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -162,35 +231,83 @@ proc refreshBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
           await conn.writeLp(responsePb.buffer)
 
           # Peer is useful. adding to rtable
           discard kad.rtable.insert(conn.peerId)
+        of MessageType.putValue:
+          var record =
+            if msg.record.isSome():
+              msg.record.unsafeGet()
+            else:
+              raise newException(CatchableError, "no record in message buffer")
```
gmelodie: Can we use `valueOr` here?

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -162,35 +231,83 @@ proc refreshBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
+          let ts = TimeStamp(ts: $times.now().utc)
+
+          let cand = EntryCandidate(key: key, value: value)
+          if not kad.entryValidator.validate(cand):
+            return
+          let others =
+            if kad.dataTable.entries.contains(key):
+              @[kad.dataTable.entries[key]]
+            else:
+              @[]
```
gmelodie, line 314: There's a [`getOrDefault`](https://nim-lang.org/docs/tables.html#getOrDefault%2CTableRef%5BA%2CB%5D%2CA) method that I think could be used here instead of `if...else`

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -200,11 +317,14 @@ proc new*(
...
-      await conn.close()
-
   return kad
 
+proc setSelector*(kad: KadDHT, selector: EntrySelector) =
+  kad.entrySelector = selector
+
+proc setValidator*(kad: KadDHT, validator: EntryValidator) =
+  kad.entryValidator = validator
+
```
gmelodie:
```suggestion
```
✂️ 
I don't like dumb getters and setters (dumb == ones that do no extra work for you)

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: problem there, is that get gives back a `Record` not a `seq[Record]` as needed. if you can suggest a way to ergonomacally do "get, and make it the only item in a sequence, otherwise the empty sequence", I'm all ears.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: me neither, but a) there isn't a usecase I can anticipate that calls for a get, and (more importantly)  b) removing the setter means that the field needs to be public, which means that it can be set to nil, thus causing a regression on the "if it's valid code, it's correct code" objective.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I got started with it, and it seems to be non-trivial. I've imported the obvious ones, but I get met with a wall of non-sensical errors, and so I have to sift through the file and errors slowly and painstakingly.

Will mark this resolved when a follow up issue is created, and will link to it.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I like it, I also changed the exception raising for a err log.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: in the interest of mitigating the risk of bike shedding, can this be defered to a follow-up issue, or do we consider finding and implementing an answer to this question a blocker?

libp2p/protocols/kademlia/dhttypes.nim
richard-ramos: nim's stdlib already norms silent default init. calling initTable without a size adds noise (similar to calling newSeq() with no size). Skipping it aligns with idiomatic nim >= 0.20.

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: If we ever need more info from putValue, we can introduce that in a future revision. Returning `Future[Result[void, string]]` matches the design described before for broadcast like functions.

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: Marking it as resolved as `putValue` signature's been updated

libp2p/protocols/kademlia/dhttypes.nim
richard-ramos: Fair. I think adding logs should follow from real debugging needs, with information that actually helps when tests break or behavior is weird. Lets just add logs where they prove useful, not just in case.

libp2p/protocols/kademlia/dhttypes.nim
richard-ramos: Maybe worth creating a `libp2p/protocols/kademlia.nim` and list there the API for kademlia? Similar to how it is done in libp2p/protocols/pubsub.nim in which only pubsub/floodsub/gossipsub are re-exported there, but internals like memcache, peertables, pubsubpeers, etc are not.

I think this would help surface most of the API devs actually care about

libp2p/protocols/kademlia/kademlia.nim
gmelodie: Ah yes. Well I'm nitpicking at this point but you could do this to get rid of the `else`:
```nim
var others = @[]
if kad.dataTable.entries.contains(key):
  others = @[kad.dataTable.entries[key]]
```

libp2p/protocols/kademlia/kademlia.nim
gmelodie: Can you not do `kad.setValidator(nil)`?

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: The specs imply that the result of FIND_NODE is the peerset for the PUT_VALUE, not just those coming from a rtable refresh. Implementing it this way deviates from the spec.

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: Reading both the specs and code for other implementations, `k` is defined as a node level constant, not per RPC. So this `replic` parameter is not needed. Especially if we use `FindNode` instead of `FindClosestPeers`

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: sure, it's user-defined, but you still need a way to tell if your value was selected. Right now, if the selector returns an older value (the value you sent isnt the "better" value), you’ll rebroadcast stale data. Instead you should’ve returned an error instead.

the spec describes select as `Select(key string, values [][]byte) (int, error)` so it's easy to tell if your value lost. Returning just the record loses that info.

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: Do note as described in https://github.com/libp2p/specs/issues/691 and in the code, the response will not be different if the record is valid and if it's selected as the best record. In other cases you'll simply get no answer. Your code would likely fail at line 95.
https://github.com/vacp2p/nim-libp2p/blob/c4590b2e41aa43dee43efa25177a38415c735fbe/libp2p/protocols/kademlia/kademlia.nim#L95

Look how `nil` is returned here for validation/select failures
- https://github.com/libp2p/go-libp2p-kad-dht/blob/944883ea5a55102c8950478645d89183901859b4/handlers.go#L169
- https://github.com/libp2p/go-libp2p-kad-dht/blob/944883ea5a55102c8950478645d89183901859b4/handlers.go#L198
- https://github.com/libp2p/go-libp2p-kad-dht/blob/944883ea5a55102c8950478645d89183901859b4/handlers.go#L202

And how no response is written in these cases: https://github.com/libp2p/go-libp2p-kad-dht/blob/944883ea5a55102c8950478645d89183901859b4/dht_net.go#L111-L132

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: My fallable memory tells me that I considered the var approach. On the one hand, I like that it's visually cleaner. on the other hand, it's making the variable mutable, which misrepresents the purpose of it being there. Self documenting code is readable code, and something that exists as var, but should never be changed following initialisation, makes the self-documentation here misinform the reader.

By the same token, nim has a pretty loose approach with mutability due to `type XXX = ref object...` being universally mutable.

as it's a nit, and resolving this is risking bikeshed, I'll go ahead and mark this comment as resolved, though I don't think there isn't room for improvement. This snippet lacks good taste, which can often be a symptom/indicator that subtle but valuable improvements can be made at a bigger picture level.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: ...good catch. I'll correct the setter to do a nil check and crash the program.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: self is resolved elsewhere

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: https://github.com/vacp2p/nim-libp2p/issues/1605

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: https://github.com/vacp2p/nim-libp2p/issues/1605

tests/kademlia/testputval.nim
Ben-PH: https://github.com/vacp2p/nim-libp2p/issues/1605

libp2p/protocols/kademlia/dhttypes.nim
Ben-PH: looks like it's either 100% public, or in-file only. Will have to make kademlia.nim much bigger, and just eat the risk of future work not inadvertently abusing items that are accesible when they shouldn't be :crossed_fingers:

libp2p/protocols/kademlia/dhttypes.nim
Ben-PH: due to nims limited functionality when it comes to accesibility control/scoping, I'll leave as is, and it can be refined in the future if it's problematic.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: #1605

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: #1605

libp2p/protocols/kademlia/dhttypes.nim
Ben-PH: #1605

libp2p/protocols/kademlia/dhttypes.nim
Ben-PH: #1605

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: #1605

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: Left this unresolved as I might have it right at this point, but I'm not yet confident enough of that yet. need to parse other ref impls first.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -162,35 +236,84 @@ proc refreshBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
         let
           buf = await conn.readLp(MaxMsgSize)
           msg = Message.decode(buf).tryGet()
 
         case msg.msgType
         of MessageType.findNode:
-          let targetIdBytes = msg.key.get()
-          let targetId = PeerId.init(targetIdBytes).tryGet()
+          let targetIdBytes = msg.key.valueOr:
+            error "findNode message without key data present"
```
richard-ramos: It's probably useful for debugging to log the peerId from the conn to know who sent an invalid message (comment applies to line 277, 287 and 293 too)

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -162,35 +236,84 @@ proc refreshBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
+              @[kad.dataTable.entries[key]]
+            else:
+              @[]
+          let selectedRec =
+            kad.entrySelector.select(RecordVal(value: value, time: ts), others)
+          let validated = ValidatedEntry(key: key, value: selectedRec.value)
+
+          # Untill the spec/convention is confirmed otherwise: always return unmodified message
+          # if an insertion is done
+          kad.dataTable.insert(validated, ts)
```
richard-ramos: Same comment as https://github.com/vacp2p/nim-libp2p/pull/1582#discussion_r2255010921 applies here. You do not know if the valor you are selecting matches what you're receiving, so it's possible to be reinserting the same data you had if the older value is better.

The TODO in line 311 is related to this, right? if so, This should have a corresponding issue so we dont forget as this here is deviating from what other implementations do

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +77,66 @@ proc waitRepliesOrTimeouts(
...
+      if cand.key in kad.dataTable.entries:
+        @[kad.dataTable.entries[cand.key]]
+      else:
+        @[]
+
+    let candAsRec = RecordVal(value: cand.value, time: ts)
+    let confirmedRec = kad.entrySelector.select(candAsRec, others)
+
+    let confirmedEnt = EntryCandidate(key: cand.key, value: confirmedRec.value)
+    if not kad.entryValidator.isValid(confirmedEnt):
```
richard-ramos: Validation of record should be done before selection, so line 119 should go before 115.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +77,66 @@ proc waitRepliesOrTimeouts(
...
+      return err("invalid overruled entry")
+    let validEnt = ValidatedEntry.take(confirmedEnt)
+
+    let peers = await kad.findNode(key)
+    var ammo: seq[PeerId]
```
richard-ramos: why `ammo` and `putClip`? Wouldnt it be better to use naming that matches the spec instead of gun related analogies?

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +77,66 @@ proc waitRepliesOrTimeouts(
...
+    for p in peers.filterIt(it.data != kad.rtable.selfId.getBytes()):
```
richard-ramos: Filtering the node's own peerID should not be done here. Consider modifying how LookupState is initialized so it is aware of the node's own peerID, and that way when `selectClosestK` is executed, it does not return your own peerID among the list of candidates.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +77,66 @@ proc waitRepliesOrTimeouts(
...
+      ammo.add(p)
+      if ammo.len >= DefaultReplic:
+        break
+    let putClip = ammo.mapIt(kad.dispatchPutVal(it, validEnt))
+    kad.dataTable.insert(validEnt, ts)
```
richard-ramos: It's not necessary to deal with the replication factor here, as FindNode already did that for you. The code, taking in account prev comment about doing the filtering of own peer id in the lookup state, could be simplified to look like this (from lines 123 to 129)

```
let peers = await kad.findNode(key)
let putClip = peers.mapIt(kad.dispatchPutVal(it, validEnt)
```

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +77,66 @@ proc waitRepliesOrTimeouts(
...
+    try:
+      await putClip.allFutures().wait(chronos.seconds(timeout.get(5)))
+    except AsyncTimeoutError:
+      discard
+
+    return results.ok()
+  except:
+    return err("todo: refine exceptions")
```
richard-ramos: Since this will catch all the exceptions, could we in the meantime add the exception message to the `err` so at during development we notice the type of exceptions happening here?

libp2p/protocols/kademlia/dhttypes.nim
```diff
@@ -0,0 +1,62 @@
+#TODO: This file was mostly put together with the intent of using the typestate pattern
+# unfortunately, due to various design decisions, my assumption that the accessibility
+# controls needed for this to be a sane direction is not upheld. This file should be
+# considered a monument to my naivity, and to be dismantled
```
richard-ramos: This comment should be rephrased. It is fine for PR thread, but in code, self-flagelation does look poorly in a public codebase as it is just leaking vibe/mood. Replace with a TODO with specific actions

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: there are two validations. the first is on the rpc provided value, the second on the result of the selection. On the one hand, you can assume that something that's already in there is valid. On the other hand, a version update of the distributed application might update the validator, with the procedure being to abandon that assumption.

Thinking about it with fresh eyes, that's thinking ahead, and we should run with the assumytion that existing entries are valid, and cross this bridge when we come to it.

ETA: This is inconsistent with the handling code, which emphasises that the code base is not ready for such complexities. Consistency within the code base should be ensured through the use of good architecture, and not depend on the developer getting it right each and every time.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: >  you’ll rebroadcast stale data

if the selector selects an older value, then the rebroadcast that follows is rebroadcasting a "better" value. the kad spec doesn't specify what makes a selection correct, or what the interface _must_ be:

> Similarly, Select() is a pure function that returns the best record out of 2 or more candidates. It may use a sequence number, a timestamp, or other heuristic of the value to make the decision.

My interpretation of this, is that it prohibits any given implementation from defining the decision process.

> stale data

if stale data is rebroadcasts due to malformed selection, then that is up to the application writer to debug. At the protocol level, our responsibility is to provide the documentation, tooling, logs, etc. to make the debug process as friction-less as possible.

That's my take. 

With that in mind, i've made a change to return a Result[Record, string], added some logs, and returning the error string

richard-ramos (approved):

PR #1585, opened 2025-07-28
test(gossipsub): Performance tests - more scenarios

rlve:


performance/scenarios.nim
```diff
@@ -64,12 +65,137 @@ proc baseTest*() {.async.} =
...
+
+proc latencyTest*() {.async.} =
+  const
+    latency = 100
+    jitter = 20
+
+  let enable = execShellCommand(
+    fmt"{enableTcCommand} netem delay {latency}ms {jitter}ms distribution normal"
+  )
+  echo "TC Enable ", enable
```
richard-ramos: Let's use chronicles instead of echo

performance/utils.nim
```diff
@@ -217,17 +229,70 @@ proc `$`*(stats: Stats): string =
...
+      )
+      .strip()
+    debug "Shell command executed", cmd, output
+    return output
+  except OSError as e:
+    raise newException(OSError, "Shell command failed")
+
+const syncDir = "/output/sync"
+
+proc syncNodes*(stage: string, nodeId, nodeCount: int, maxRetries = 50) {.async.} =
```
vladopajic: - why do nodes need to sync?
- cant they use different syncing mechanism? for example `started` stage can spinlock until there are X connected peers.
- `maxRetries` here is in function of time `50 * 100ms = 5sec` in general it would be better if it is duration
- this logic gives 100ms precision for sync. better precision could be achieved if last node writes when to continue, in different file. for example last node will write `time.Now() + 5.seconds` to `sync_continue` file. all nodes will block until this file exists, and when it exists they will read value from file (representing time in future), and block until that time.

performance/scenarios.nim
```diff
@@ -26,6 +27,9 @@ proc baseTest*() {.async.} =
     hostname = getHostname()
     rng = libp2p.newRng()
 
+  if nodeId == 0:
+    clearSyncFiles()
```
vladopajic, line 43: hmm could it be the case that this node clears file of some other node that was started earlier?

performance/main.nim
```diff
@@ -2,3 +2,12 @@ import chronos
 import ./scenarios
```
vladopajic, line 1: this now looks more like  "network performance" tests (to be more descriptive)

performance/main.nim
```diff
@@ -2,3 +2,12 @@ import chronos
...
```
vladopajic: would it make sense to block ci if not all messages are received?

performance/utils.nim
rlve: 1. Test scripts are running concurrently in 10 separate docker containers where nodes need to be started, then establish connection between each other, spread messages and calculate the latencies.
The previous approach with hardcoded delays worked when there was 1 scenario. However, with 10 scenarios running one after another, without any sync mechanism between tests, the delays started to accumulate and nodes started to fail due to race conditions, like when in one node some scenario already finished, and other node was still sending messages.
Syncing nodes solved that issue.
2. Nodes probably could use a different sync mechanism, but using sync files in the shared volume is just the simplest approach and I don't see a reason to complicate it. Shared volume already existed for stats aggregation. This approach is also flexible and doesn't rely on the GossipSub logic, therefore I could reuse it at every stage where nodes had to be in sync.
3. Sure, can change to duration based timeout.
4. Agree, that would provide better sync precision, however, I don't think such precision is needed, the tests seem to be doing fine, so I'm not sure it makes sense to make the sync logic more complex.

performance/scenarios.nim
rlve: Yes, it could, however the initial wait in `syncNodes()` helps with that.

performance/main.nim
rlve: What exactly you suggest to rename?

performance/main.nim
rlve: I'm not sure, missing a few messages doesn't really impact latency statistics.

performance/main.nim
vladopajic: i posted this as comment because general message when approving doesn't have replays.

---

> What exactly you suggest to rename?

i don't have hard opinion on this, but generally effort, since this pr, has shifted a bit from just performance  more into the territory or testing resilience (chaos testing). now there are metrics for which we could say is part of performance, but functionally these appear more then performance.

just wanted to post this comment, not really suggesting any rename. but if this sparks some ideas even better

performance/utils.nim
vladopajic: when measuring something, tools used for measuring should be as precise as possible. now here we don't really need precise measurement (which arguably implies we don't really need measurements). but it depends how much these 100ms can impact latency results...

rlve: I'm not saying we don't need precise measurement, I'm saying 100ms precision in sync mechanism is enough. As results are showing it doesn't impact latency results at all. Also, there is a warmup phase when sending messages where latency is not measured, which helps guard against such situations.

PR #1586, opened 2025-07-30
doc: Document nimble 0.20.1 dev req and instructions

Ben-PH:
From the problems I had yesterday, the resolution required using nimble 0.20.1, which isn't available out of the box (yet) using nix package repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,9 @@ cd nim-libp2p
 nimble install -dy
 ```
 You can use `nix develop` to start a shell with Nim and Nimble.
+
+nimble 0.20.1 is required for running `testnative`. At time of writing, this is not available in nixpkgs: If using `nix develop`, follow up with `nimble install nimble`, and use that (typically `~/.nimble/bin/nimble`).
+
 ### Testing
 Run unit tests:
 ```sh


gmelodie (approved):

README.md
```diff
@@ -72,6 +72,9 @@ cd nim-libp2p
 nimble install -dy
 ```
 You can use `nix develop` to start a shell with Nim and Nimble.
+
```
richard-ramos, line 75: should we specify the version?
 `nimble install nimble@0.20.1`

Ben-PH: both work, with the `@` is more explicit:

the "normal" is that something like `sudo apt install nimble` will just give you the latest nimble in the package. for nixpkgs, that's 0.18.???.

the equivilent to apt in nix is to add `nimble` to the list of packages in your system specification (layed out in a .nix file). There are ways to auto-magicallly set things up so this install instruction is redundant, but I'm a nix-noob, and `0.20.1` will be in nixpkgs soon enough to make all that redundant.

`nimble install nimble` is good enough for now. if things escalate to where things need to be more specific, we can update the project `flake.nix` to do this automagically.

(also, if there's anyone out there that's actually good with nix and want's to contribute, they might take the initiative on the flake.nix ;) )

Issue #1545, opened 2025-07-15
kad-dht follow up issue.


Ben-PH:
In preparation of #1324 there are some items that shouldn't block merging, but should be GH documented for follow up once that PR is merged.

- [ ] (multiple) [Result use](https://github.com/vacp2p/nim-libp2p/pull/1324#discussion_r2205782050) (this can fall under a general "improve the code base so it follows the style guide, so doesn't need addressing before close)
- [x] bootstrap configuration [TODO](https://github.com/vacp2p/nim-libp2p/pull/1324#discussion_r2205786450)
- [x] Strategically add [trace logging](https://github.com/vacp2p/nim-libp2p/pull/1324#discussion_r2205791497) (done: forget where mentioned, but the comment was made that where the logs should go will emerge over time)
- [x] implement more [message types](https://github.com/vacp2p/nim-libp2p/pull/1324#discussion_r2205792423) (crossed off as "this is the main big-picture", and we are reliably on-coarse)
  - [x] revert use of exceptions for [control flow](https://github.com/vacp2p/nim-libp2p/pull/1324#discussion_r2207785199) (#1627)
- [x] check for side-effect free procs that can be [funcs](https://github.com/vacp2p/nim-libp2p/pull/1324#discussion_r2205814038) (Decided to cancel this follow up)
- [x] configurable wait time when sending the [find-node](https://github.com/vacp2p/nim-libp2p/pull/1324#discussion_r2205841897)
- [x] check contents of [rtable](https://github.com/vacp2p/nim-libp2p/pull/1324#discussion_r2205844682) (not needed)
- digression with spec: [ktables](https://github.com/vacp2p/nim-libp2p/pull/1324#discussion_r2219806429)
- [x] skip dialing self in find-node

PR #1587, opened 2025-07-30
feat(kad): Make some async wait-limits configurable

Ben-PH:
Resolves follow-up #1545 for configuring bootstrap and findnode times.

question for reviewers: current default is 5seconds (dispatching message) and 10 minutes (heartbeat). Are these sane defaults? Does it make sense to think about some sort of default configuration? my thinking is that somewhere between this code base, and an end user running a node, there might be a desire to say "no: I want my node/application/network/etc. to have a heartbeat of $DURATION"

### DIFF
--- a/libp2p/protocols/kademlia/kademlia.nim
+++ b/libp2p/protocols/kademlia/kademlia.nim
@@ -1,3 +1,4 @@
+import results
 import chronos
 import chronicles
 import sequtils
@@ -23,6 +24,8 @@ type KadDHT* = ref object of LPProtocol
   rng: ref HmacDrbgContext
   rtable*: RoutingTable
   maintenanceLoop: Future[void]
+  findNodeTimeout: Opt[Duration]
+  bucketHeartbeat: Opt[Duration]
 
 const MaxMsgSize = 4096
 
@@ -91,7 +94,7 @@ proc findNode*(
 
       pendingFutures[peer] = kad
         .sendFindNode(peer, addrTable.getOrDefault(peer, @[]), targetId)
-        .wait(5.seconds)
+        .wait(kad.findNodeTimeout.get(5.seconds))
 
       state.activeQueries.inc
 
@@ -127,8 +130,9 @@ proc bootstrap*(
       error "failed to connect to bootstrap peer", peerId = b.peerId, error = e.msg
 
     try:
-      let msg =
-        await kad.sendFindNode(b.peerId, b.addrs, kad.rtable.selfId).wait(5.seconds)
+      let msg = await kad.sendFindNode(b.peerId, b.addrs, kad.rtable.selfId).wait(
+        kad.findNodeTimeout.get(5.seconds)
+      )
       for peer in msg.closerPeers:
         let p = PeerId.init(peer.id).tryGet()
         discard kad.rtable.insert(p)
@@ -153,8 +157,20 @@ proc refreshBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
       discard await kad.findNode(randomKey)
 
 proc maintainBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
-  heartbeat "refresh buckets", 10.minutes:
+  heartbeat "refresh buckets", kad.bucketHeartbeat.get(10.minutes):
     await kad.refreshBuckets()
+
+proc setFindNodeTimeout*(self: var KadDHT, duration: Duration) =
+  self.findNodeTimeout = Opt.some(duration)
+
+proc setBucketHeartbeat*(self: var KadDHT, duration: Duration) =
+  self.bucketHeartbeat = Opt.some(duration)
+
+proc unSetFindNodeTimeout*(self: var KadDHT) =
+  self.findNodeTimeout = Opt.none(Duration)
+
+proc unSetBucketHeartbeat*(self: var KadDHT) =
+  self.bucketHeartbeat = Opt.none(Duration)
 
 proc new*(
     T: typedesc[KadDHT], switch: Switch, rng: ref HmacDrbgContext = newRng()


libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -91,7 +93,7 @@ proc findNode*(
 
       pendingFutures[peer] = kad
         .sendFindNode(peer, addrTable.getOrDefault(peer, @[]), targetId)
-        .wait(5.seconds)
+        .wait(kad.findNodeTimeout.get(chronos.seconds(5)))
```
gmelodie: i think you can get rid of `chronos.`
```suggestion
        .wait(kad.findNodeTimeout.get(5.seconds))
```

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -127,8 +129,9 @@ proc bootstrap*(
       error "failed to connect to bootstrap peer", peerId = b.peerId, error = e.msg
 
     try:
-      let msg =
-        await kad.sendFindNode(b.peerId, b.addrs, kad.rtable.selfId).wait(5.seconds)
+      let msg = await kad.sendFindNode(b.peerId, b.addrs, kad.rtable.selfId).wait(
+        kad.findNodeTimeout.get(chronos.seconds(5))
```
gmelodie: same as above
```suggestion
        kad.findNodeTimeout.get(5.seconds)
```

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -23,6 +23,8 @@ type KadDHT* = ref object of LPProtocol
   rng: ref HmacDrbgContext
   rtable*: RoutingTable
   maintenanceLoop: Future[void]
+  findNodeTimeout: Option[chronos.Duration]
+  bucketHeartbeat: Option[chronos.Duration]
```
gmelodie: I _think_ we should be using `Opt` instead of `Option` everywhere (@richard-ramos ?)

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -153,9 +156,21 @@ proc refreshBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
       discard await kad.findNode(randomKey)
 
 proc maintainBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
-  heartbeat "refresh buckets", 10.minutes:
+  heartbeat "refresh buckets", kad.bucketHeartbeat.get(chronos.minutes(10)):
     await kad.refreshBuckets()
 
+proc setFindNodeTimeout*(self: var KadDHT, duration: chronos.Duration) =
```
gmelodie: Do the same for other functions
```suggestion
proc setFindNodeTimeout*(self: var KadDHT, duration: Duration) =
```

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -153,9 +156,21 @@ proc refreshBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
+  self.findNodeTimeout = some(duration)
+
+proc setBucketHeartbeat*(self: var KadDHT, duration: chronos.Duration) =
+  self.bucketHeartbeat = some(duration)
+
+proc unSetFindNodeTimeout*(self: var KadDHT, duration: chronos.Duration) =
```
gmelodie: Unsets don't need to receive `duration` (also change the other unset proc)
```suggestion
proc unSetFindNodeTimeout*(self: var KadDHT) =
```

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -153,9 +156,21 @@ proc refreshBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
+  self.findNodeTimeout = none(chronos.Duration)
```
gmelodie: Same for other function
```suggestion
  self.findNodeTimeout = none(Duration)
```

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I'm importing time for for sampling times, and and going to/from `RFC????` stringified time specification.

Given that time sampling and stringification is something that interacts with other languages, this implimentation will have to interact with `time` implimentations of said other languages. In this context, standard libraries interacting with standard libraries is preferable, so we should avoid using 3rd party libs where it makes sense to do so.

`chronos` is primarily for in-binary async management, so makes no sense to me to use it for sampling, formating, and parsing time.

As for `from xxx import xxx`, that can be problematic: some items from some libraries, peicemeal importing doesn't work. On the other hand, glob importing is problematic. unfortunately, nim doesn't leave you with a clean way of doing this, from what I gather.

Fact is, we need std-time for some things, and chronos for others, and they both provide `seconds`. as such, the namespace needs to be specified from what I understand.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I've been using std-opt out of habit more than anything else. +1 on needing an explination.

libp2p/protocols/kademlia/kademlia.nim
gmelodie: You're not importing `time` in this file, so it's safe to remove the `chronos.` everywhere.

Ben-PH: ah yes. I've been working on `put-val`, which requires what I was describing, and still had that cached in my mind.

I'll revert as you suggest, and we can revisit this in the context of `kad-putval` PR

richard-ramos:
For Opt vs Option, we should use Opt because
- Suggested by the status nim styleguide: https://status-im.github.io/nim-style-guide/libraries.results.html
- Has better integration with Result

Actually this could be a good first issue, to look for instances of Option and replace those by Opt (as long as it is not part of public API)

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: I've made the change. The other std-options in the file are out of scope, so I'll leave them to the issue suggested: https://github.com/vacp2p/nim-libp2p/pull/1587#issuecomment-3139952792

gmelodie (approved):

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -153,9 +157,21 @@ proc refreshBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
+
+proc setBucketHeartbeat*(self: var KadDHT, duration: Duration) =
+  self.bucketHeartbeat = Opt.some(duration)
+
+proc unSetFindNodeTimeout*(self: var KadDHT) =
+  self.findNodeTimeout = Opt.none(Duration)
+
+proc unSetBucketHeartbeat*(self: var KadDHT) =
+  self.bucketHeartbeat = Opt.none(Duration)
+
```
richard-ramos, line 174: I wonder if in practice we will use these methods on runtime, instead of passing the node timeout and heartbeat value on the constructor as part of a confg object optionally pased as a  parameter in the constructor.

richard-ramos, line 174: Once line 160 is executed, the heartbeat duration can't be updated so the procs for setting / unsetting the bucketHeartbeat wont do anything unless `maintenanceLoop` is cancelled and recreated.

Ben-PH: oooohhhhhhh. I see that now.

That kinda blocks this PR in that case. the thinking being is that each time it runs, it sets the sleep based on the current `KadDHT` field. Didn't think to check.

It makes sense to me for this to be changeable. One contrived e.g.: want to find a problem node, so up the timeout to 10s. Want to spend 10m doing rapid-fire recovery/repair actions before falling back to the usual once per hour, or so.

Sorting that out Is probably out of scope for what I had in mind. This was supposed to be a simple follow up. Do this, and it requires a touch of architecture change.

Thoughts?

gmelodie:
As per @richard-ramos's comment
> If we ever need to be able to change timeouts during runtime, we can look into it then, but just trying to add support for it without a clear need would be wasting time

PR #1589, opened 2025-07-30
feat(kad|getval): Provide `GET_VALUE` RPC

Ben-PH:
at time of writing, an initial sketch:

- adds public `getValue` proc
- uses private `dispatchGetVal` proc

Once `put-val` is merged, the work of this PR will be able to help with the iterative improvement of put-val, as getting values involves putting values (to update out-dated entries), thus assisting with surfacing defects, use-cases, etc.


libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -162,6 +162,52 @@ proc findNode*(
 
   return state.selectClosestK()
 
+proc dispatchGetValue(
+    self: KadDHT, peer: PeerId, key: Key
+): Future[(closests: seq[PeerId], val: EntryVal)] {.async: (raises: [CancelledError]).} =
+  discard
```
richard-ramos, line 172: TODO: retrieved entries must be validated

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -162,6 +162,52 @@ proc findNode*(
...
+    self: KadDHT, key: Key, quarum: uint8
+): Future[EntryVal] {.async: (raises: [CancelledError]).} =
+  var fetched = 0
+  var bestVal: Option[EntryVal]
+  var queriedPeers: HashSet[PeerId]
+  var toQuery: (Key, seq[PeerId]) = (key, kad.rtable.findClosestPeers(key, quorum))
+  var outdatedHolders: HashSet[PeerId]
+  var activeQueries: seq[string] = @["todo: sort out tracking of active queries"]
+
+  while fetched < quorum:
```
richard-ramos, line 184: It should be possible to specify the mechanism to select the best record out of a list of retrieved values. In the spec it is suggested something like this:
```
// Validator is an interface that should be implemented by record
// validators.
type Validator interface {
	// Select selects the best record from the set of records (e.g., the
	// newest).
	//
	// Decisions made by select should be stable.
	Select(key string, values [][]byte) (int, error)
}

// Select() is a pure function that returns the best record out of 2 or more candidates. It may use a sequence number, a timestamp, or other heuristic of the value to make the decision.
```

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -162,6 +162,52 @@ proc findNode*(
...
 
+proc dispatchGetValue(
+    self: KadDHT, peer: PeerId, key: Key
+): Future[(closests: seq[PeerId], val: EntryVal)] {.async: (raises: [CancelledError]).} =
+  discard
+
+proc getValue*(
+    self: KadDHT, key: Key, quarum: uint8
+): Future[EntryVal] {.async: (raises: [CancelledError]).} =
+  var fetched = 0
```
richard-ramos, line 177: getValue should attempt to retrieve the value from the local store, and depending on the quorim mechanism used, it might not be necessary to do any further queries to other nodes

PR #1594, opened 2025-08-04
fix(kad): Skip self when iterating through findNode dialouts

Ben-PH:
As per title.

I've also inlined the checking of convergence, as it is particular to finding node, and used only once. I felt the inlining + commenting makes the threads of logic more parsable for humans.

Still remaining is the following error:

```
ERR 2025-08-04 11:53:57.305+02:00 could not send find_node to peer           topics="kad-dht" tid=96273 peerId=12D*Dyne2W err="failed dial existing: Couldn't get muxed stream in dial for peer_id: 12D*Dyne2W"
```

I've left in some `debug` lines that will aid in resolving that, and any other issues that might crop up relating to fiding.

### DIFF
--- a/libp2p/protocols/kademlia/kademlia.nim
+++ b/libp2p/protocols/kademlia/kademlia.nim
@@ -68,6 +68,8 @@ proc waitRepliesOrTimeouts(
 
   return (receivedReplies, failedPeers)
 
+# Helper function forward declaration
+proc checkConvergence(state: LookupState, me: PeerId): bool {.raises: [], gcsafe.}
 proc findNode*(
     kad: KadDHT, targetId: Key
 ): Future[seq[PeerId]] {.async: (raises: [CancelledError]).} =
@@ -79,7 +81,6 @@ proc findNode*(
   var addrTable: Table[PeerId, seq[MultiAddress]] =
     initTable[PeerId, seq[MultiAddress]]()
 
-  debug "self", self = kad.switch.peerInfo.peerId
   while not state.done:
     let toQuery = state.selectAlphaPeers()
     debug "queries", list = toQuery.mapIt(it.shortLog()), addrTab = addrTable
@@ -115,12 +116,14 @@ proc findNode*(
       state.markFailed(timedOut)
 
     # Check for covergence: no active queries, and no other peers to be selected
-    let ready = state.activeQueries == 0
-    let noNew =
-      selectAlphaPeers(state).filterIt(kad.switch.peerInfo.peerId != it).len == 0
-    state.done = ready and noNew
+    state.done = checkConvergence(state, kad.switch.peerInfo.peerId)
 
   return state.selectClosestK()
+
+proc checkConvergence(state: LookupState, me: PeerId): bool {.raises: [], gcsafe.} =
+  let ready = state.activeQueries == 0
+  let noNew = selectAlphaPeers(state).filterIt(me != it).len == 0
+  return ready and noNew
 
 proc bootstrap*(
     kad: KadDHT, bootstrapNodes: seq[PeerInfo]


libp2p/protocols/kademlia/lookupstate.nim
```diff
@@ -103,11 +103,6 @@ proc init*(T: type LookupState, targetId: Key, initialPeers: seq[PeerId]): T =
   )
   return res
 
-proc checkConvergence*(state: LookupState): bool =
```
vladopajic, line 106: why remove this `proc` ?

vladopajic (approved):

libp2p/protocols/kademlia/lookupstate.nim
gmelodie: +1

gmelodie (approved):

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -79,14 +79,16 @@ proc findNode*(
   var addrTable: Table[PeerId, seq[MultiAddress]] =
     initTable[PeerId, seq[MultiAddress]]()
 
+  debug "self", self = kad.switch.peerInfo.peerId
```
richard-ramos: This is kinda noisy and not necessary as the switch should print the info on startup
```suggestion
```

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -79,14 +79,16 @@ proc findNode*(
...
   while not state.done:
     let toQuery = state.selectAlphaPeers()
+    debug "queries", list = toQuery.mapIt(it.shortLog()), addrTab = addrTable
     var pendingFutures = initTable[PeerId, Future[Message]]()
 
-    for peer in toQuery:
-      if pendingFutures.hasKey(peer):
-        continue
-
+    # TODO: pending futures always empty here, no?
```
richard-ramos, line 89: pendingFutures should likely be defined before `while not.state.done:`

libp2p/protocols/kademlia/lookupstate.nim
richard-ramos: +2

Ben-PH: I mentioned in the text of the PR, but to expand:

this proc is only used once, in kademlia.nim. it is public, so it adds noise to the public interface, and it being in a different file, means that you have to navigate the code base to understand what's happening. On top of that, having a helper function indicates the possibility of repeated use, and the public + different file adds to this.

The purpose of helper functions is to serve the objective of making the code base more maintainable and readable. in this case, the opposite effect has taken place based on my experience working with the findNode function

instead of inlining, ive moved the helper to just below the callsite.

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: due to performance reasons, code style, or correctness?

If for style, I think having it within the loop clearly communicates the fresh table each time. 

if for performance, it would be trivially easy for the compiler to recognise that the newly initted table is taking over the old one, and so automatically do the refreshing that would need to be done manually. 

If for correctness, then i can fix and test in this PR, or a follow up if it misses this one.

libp2p/protocols/kademlia/lookupstate.nim
vladopajic: > instead of inlining, ive moved the helper to just below the callsite.

:+1:

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -68,6 +68,8 @@ proc waitRepliesOrTimeouts(
 
   return (receivedReplies, failedPeers)
 
+# Helper function forward declaration
+proc checkConvergence(state: LookupState, me: PeerId): bool {.raises: [], gcsafe.}
```
vladopajic, line 72: not sure about this as general practice. i would suggest just having proc definition here as well.

Ben-PH: I decided on this approach as the implemention is very tightly coupled to call site in terms of readability. its unfortunate that nim needs forward declaration, but moving up there would probably be more of a distraction than anything else. I triod this approach in the hope of finding a middle ground.

this has gotten a bit too bike-sheddy, and this seems to be in the domain of "nit", so I'll go ahead and merge the changes as is. Not getting this just right is worth moving on to the more important things.

RWSINC (approved):

PR #1599, opened 2025-08-06
chore(quic): add length prefixed test

vladopajic:


tests/testquic.nim
```diff
@@ -174,3 +174,36 @@ suite "Quic transport":
...
+          await stream.writeLp(fromHex("5678"))
+          await stream.close()
+
+      proc runClient(server: QuicTransport) {.async.} =
+        let client = await createTransport()
+        let conn = await client.dial("", server.addrs[0])
+        let stream = await getStream(QuicSession(conn), Direction.Out)
+        await stream.writeLp(fromHex("1234"))
+        check (await stream.readLp(100)) == fromHex("5678")
+        await client.stop()
```
gmelodie, line 202: Nits:
1. unless you're planning to use this proc on other tests, i'd say unpack it (not have it).
2. `await client.stop()` could be `defer: await client.stop()` right after creating the transport, no?

gmelodie (approved):

tests/testquic.nim
vladopajic: 1. this will be only usecase for this proc
2. since this is 5 lines of code and code wont be modified, i dont see much value with defer

gmelodie: Sure thing man, they're nits after all :)

tests/testquic.nim
```diff
@@ -174,3 +174,36 @@ suite "Quic transport":
...
+          server: QuicTransport
+      ) {.async: (raises: [transport.TransportError, LPStreamError, CancelledError]).} =
+        while true:
+          let conn =
+            try:
+              await server.accept()
+            except QuicTransportAcceptStopped:
+              return # Transport is stopped
+          if conn == nil:
+            continue
```
Ben-PH, line 189: Perhaps out of scope for this PR, but the only place time I've felt that `nil`/`null` makes sense from an engineering perspective only when `Option` is not possible, such as the lowest of OS components, core logic for `Option` construction, etc. anything that behaves in terms of "it is valid that this may or may not exist", should be expressed in terms of "this might be a `some`, it might be a `none"

Use Result to encapsulate "Getting something implies a complete traversal of the happy-path. getting an error implies the happy path was taken at some point, and the error is recoverable." (not being recoverable implies that the program and/or thread that is expecting a result should be terminated)

note, that in the results documentation, i recall seeing something along the lines of "use Option when an error is simple/trivial". this is just _wrong_, and it's incredibly misguiding. fetching a value from a hash map using a key that isn't in the map is not an error. it is perfectly normal to do a get that returns a `none`. a safe division wouldn't return an `error` when when safe-dividing by zero: raw division returns an int, except when the divisor is zero, in which case the result is undefined, and by convention, it causes a crash/exception/undefined behavior/etc. . safe-div returns `Option[int]` _always_ (not considering the size-limit of numbers in real-word), if the divisor is zero, the returned value is well-defined, valid, and not indicitive of a state from which recovery is needed: `none`

to illustrate: a database read should probably return `Result[Option[DbEntry], DbReadError]` without straying from the happy path, an entry will be returned, if it exists. if a read error occurs, you are now handling the error-wrapped type, whicth is typically some kind of context about how the code execution diverged from the happy-path.

tests/testquic.nim
```diff
@@ -174,3 +174,36 @@ suite "Quic transport":
 
     # run multiple clients simultainiously
     await allFutures(runClient(), runClient(), runClient())
+
+    asyncTest "read/write Lp":
+      proc serverHandler(
+          server: QuicTransport
+      ) {.async: (raises: [transport.TransportError, LPStreamError, CancelledError]).} =
```
Ben-PH, line 181: normally I would suggest removing all but the CancelledError, and catching the other exceptions from a single-line try-block. This might infer the need to return a `Result[void, ???]` (initially, string to make things easier, but that should evolve to an informative `enum` context)

however, this is test code, so the benefit of "towards 100% `raises: []`" is less self-evident.

tests/testquic.nim
```diff
@@ -174,3 +174,36 @@ suite "Quic transport":
...
+              await server.accept()
+            except QuicTransportAcceptStopped:
+              return # Transport is stopped
+          if conn == nil:
+            continue
+
+          let stream = await getStream(QuicSession(conn), Direction.In)
+          check (await stream.readLp(100)) == fromHex("1234")
+          await stream.writeLp(fromHex("5678"))
+          await stream.close()
```
Ben-PH, line 194: move up with:

```nim
defer:
  await stream.close()
```

Ben-PH (approved):
LGTM. nits and ~~rants~~ sharing of opinion.

tests/testquic.nim
vladopajic: it had auto-merge on. i'll revisit this next time iam around

PR #1600, opened 2025-08-06
feat(perf): implement proper half-close semantics

MorganaFuture:
- Add closeWrite method to Connection base class
  - Implement closeWrite for ChronosStream (TCP), QuicStream, LPChannel (mplex), YamuxChannel
  - Fix PERF client to use closeWrite instead of full close
  - Fix PERF server EOF detection to work with muxed connections
  - Ensures PERF protocol compliance

### DIFF
--- a/libp2p/protocols/perf/client.nim
+++ b/libp2p/protocols/perf/client.nim
@@ -12,8 +12,6 @@
 import chronos, chronicles, sequtils
 import stew/endians2
 import ./core, ../../stream/connection
-when defined(libp2p_quic_support):
-  import ../../transports/quictransport
 
 logScope:
   topics = "libp2p perf"
@@ -59,13 +57,8 @@ proc perf*(
       statsCopy.uploadBytes += toWrite.uint
       p.stats = statsCopy
 
-    # Close connection after writing for TCP, but not for QUIC
-    when defined(libp2p_quic_support):
-      if not (conn of QuicStream):
-        await conn.close()
-      # For QUIC streams, don't close yet - let server manage lifecycle
-    else:
-      await conn.close()
+    # Close write side of the stream (half-close) to signal EOF to server
+    await conn.closeWrite()
 
     size = sizeToRead
 
@@ -80,10 +73,8 @@ proc perf*(
       statsCopy.downloadBytes += toRead.uint
       p.stats = statsCopy
 
-    # Close QUIC connections after read phase
-    when defined(libp2p_quic_support):
-      if conn of QuicStream:
-        await conn.close()
+    # Close the connection after reading
+    await conn.close()
   except CancelledError as e:
     raise e
   except LPStreamError as e:
--- a/libp2p/stream/connection.nim
+++ b/libp2p/stream/connection.nim
@@ -40,6 +40,12 @@ type
 
 proc timeoutMonitor(s: Connection) {.async: (raises: []).}
 
+method closeWrite*(s: Connection): Future[void] {.base, async: (raises: []).} =
+  ## Close the write side of the connection
+  ## Subclasses should implement this for their specific transport
+  ## Default implementation just closes the entire connection
+  await s.close()
+
 func shortLog*(conn: Connection): string =
   try:
     if conn == nil:
--- a/.pinned
+++ b/.pinned
@@ -8,7 +8,7 @@ json_serialization;https://github.com/status-im/nim-json-serialization@#2b1c5eb1
 metrics;https://github.com/status-im/nim-metrics@#6142e433fc8ea9b73379770a788017ac528d46ff
 ngtcp2;https://github.com/status-im/nim-ngtcp2@#9456daa178c655bccd4a3c78ad3b8cce1f0add73
 nimcrypto;https://github.com/cheatfate/nimcrypto@#19c41d6be4c00b4a2c8000583bd30cf8ceb5f4b1
-quic;https://github.com/status-im/nim-quic.git@#d9a4cbccd509f7a3ee835f75b01dec29d27a0f14
+quic;https://github.com/status-im/nim-quic.git@#a5bd3dd
 results;https://github.com/arnetheduck/nim-results@#df8113dda4c2d74d460a8fa98252b0b771bf1f27
 secp256k1;https://github.com/status-im/nim-secp256k1@#f808ed5e7a7bfc42204ec7830f14b7a42b63c284
 serialization;https://github.com/status-im/nim-serialization@#548d0adc9797a10b2db7f788b804330306293088
--- a/libp2p.nimble
+++ b/libp2p.nimble
@@ -10,7 +10,7 @@ skipDirs = @["tests", "examples", "Nim", "tools", "scripts", "docs"]
 requires "nim >= 2.0.0",
   "nimcrypto >= 0.6.0 & < 0.7.0", "dnsclient >= 0.3.0 & < 0.4.0", "bearssl >= 0.2.5",
   "chronicles >= 0.11.0 & < 0.12.0", "chronos >= 4.0.4", "metrics", "secp256k1",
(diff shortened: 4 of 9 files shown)


MorganaFuture:
@vladopajic Could you please re-look?   It practically fully implements the standard perf protocol, I only added uploadSize to get rid of the flaky quic test, which sometimes has a deadlock. Actually, I spent a lot of time trying to understand why it occurs, it only reproduces on the libp2p protocol side, here's an example of a test that repeats the perf logic at the quic protocol level, where I can't catch the deadlock.

```

import std/sequtils
import pkg/chronos
import pkg/chronos/unittest2/asynctests
import pkg/quic/errors
import pkg/quic/transport/stream
import pkg/quic/transport/quicconnection
import pkg/quic/transport/ngtcp2/native
import pkg/quic/udp/datagram
import tests/helpers/simulation

suite "perf protocol like test - half-close hang":
  setup:
    var (client, server) = waitFor performHandshake()

  teardown:
    waitFor client.drop()
    waitFor server.drop()

  asyncTest "perf protocol simulation hangs on read after closeWrite":
    # This test simulates the exact perf protocol flow:
    # 1. Client sends 8 bytes (download size)
    # 2. Client sends upload data (100KB)
    # 3. Client calls closeWrite() 
    # 4. Server reads all data including the closeWrite signal (should get EOF)
    # 5. Server sends download data back
    
    let simulation = simulateNetwork(client, server)
    let clientStream = await client.openStream()
    
    const 
      uploadSize = 100000  # 100KB like in perf test
      downloadSize = 10000000  # 10MB like in perf test
      chunkSize = 65536  # 64KB chunks like perf
    
    proc serverHandler() {.async.} =
      let serverStream = await server.incomingStream()
      echo "SERVER: Got incoming stream"
      
      # Step 1: Read download size (8 bytes) 
      echo "SERVER: Reading download size (8 bytes)"
      let sizeData = await serverStream.read()
      echo "SERVER: Read download size: ", sizeData.len, " bytes"
      # In real perf this would be parsed as uint64, but we skip that
      
      # Step 2: Read upload data until EOF
      echo "SERVER: Starting to read upload data"
      var totalBytesRead = 0
      while true:
        echo "SERVER: Waiting for next chunk..."
        let chunk = await serverStream.read()  # THIS WILL HANG after closeWrite!
        echo "SERVER: Read chunk: ", chunk.len, " bytes"
        
        if chunk.len == 0:
          echo "SERVER: Got EOF, total read: ", totalBytesRead
          break
        
        totalBytesRead += chunk.len
        echo "SERVER: Total bytes read so far: ", totalBytesRead
        
        if totalBytesRead >= uploadSize:
          echo "SERVER: Read all expected upload data"
          break  # Exit the reading loop once we have all expected data
      
      echo "SERVER: Starting to send download data"
      # Step 3: Send download data back
      var remainingToSend = downloadSize
      while remainingToSend > 0:
        let toSend = min(remainingToSend, chunkSize)
        let dummyData = newSeq[byte](toSend)
        await serverStream.write(dummyData)
        remainingToSend -= toSend
        echo "SERVER: Sent ", toSend, " bytes, remaining: ", remainingToSend
      
      await serverStream.close()
      echo "SERVER: Done"
    
    # Start server handler
    let serverFut = serverHandler()
    
    # Step 1: Send download size (8 bytes) - activate stream first
    echo "CLIENT: Sending download size"
    await clientStream.write(@[0'u8, 0'u8, 0'u8, 0'u8, 0'u8, 152'u8, 150'u8, 128'u8]) # 10MB in big endian
    
    # Step 2: Send upload data in chunks
    echo "CLIENT: Starting upload of ", uploadSize, " bytes"
    var remainingToSend = uploadSize
    while remainingToSend > 0:
      let toSend = min(remainingToSend, chunkSize)
      let dummyData = newSeq[byte](toSend)
      await clientStream.write(dummyData)
      remainingToSend -= toSend
      echo "CLIENT: Sent ", toSend, " bytes, remaining: ", remainingToSend
    
    # Step 3: Close write side (this is where the problem happens)
    echo "CLIENT: Upload complete, closing write side"
    await clientStream.closeWrite()
    echo "CLIENT: Write side closed"
    
    # Step 4: Start reading download data
    echo "CLIENT: Starting to read download data"
    var totalDownloaded = 0
    while totalDownloaded < downloadSize:
      let chunk = await clientStream.read()
      totalDownloaded += chunk.len
      echo "CLIENT: Downloaded ", chunk.len, " bytes, total: ", totalDownloaded
    
    echo "CLIENT: Download complete"
    await clientStream.close()
    
    # Server should complete, but it hangs on the read after closeWrite
    await serverFut
    
    await simulation.cancelAndWait()
```

libp2p/protocols/perf/client.nim
```diff
@@ -48,6 +46,7 @@ proc perf*(
     let start = Moment.now()
 
     await conn.write(toSeq(toBytesBE(sizeToRead)))
+    await conn.write(toSeq(toBytesBE(sizeToWrite)))
```
vladopajic, line 49: this additional write will brake comparability with other perf implementations

vladopajic: we do not necessarily have to have comparability, for sake of banchmarking, but not sure if we should go down that route.

MorganaFuture: Yea, I have added it for fix flaky quic test. Need to figure out why quic sometimes catches a deadlock in this scenario. I have wrote about this case above

vladopajic: yeah just got that. it good that we found deadlock, i think we need to solve this because it might be happening in other usecases.

vladopajic (commented):
there is defiantly value in this pr. but since perf protocol is changed, and since quic has that deadlock we should postpone a bit merge of this pr.

libp2p/protocols/perf/client.nim
vladopajic: hey @MorganaFuture 

https://github.com/vacp2p/nim-quic/pull/91/files
this test is not causing deadlock in ci

MorganaFuture: @vladopajic  Sorry remove this. I  forgot to remove this when testing. 

```
        if totalBytesRead >= uploadSize:
          echo "SERVER: Read all expected upload data"
          break  # Exit the reading loop once we have all expected data
```
And its really flaky on my machine i catch it once per 10-20 runs

vladopajic: thanks man, found the issue and fixed it in the same pr (# 91).

could you use last commit of pr#91 and change this pr so that perf protocol is unchanged?

vladopajic: @MorganaFuture issue was fixed in `v0.2.12` quic. 

could you please update this pr, so that the perf protocol is not changed?

libp2p/stream/connection.nim
```diff
@@ -40,6 +40,12 @@ type
 
 proc timeoutMonitor(s: Connection) {.async: (raises: []).}
 
+method closeWrite*(s: Connection): Future[void] {.base, async: (raises: []).} =
+  ## Close the write side of the connection
+  ## Subclasses should implement this for their specific transport
+  ## Default implementation just closes the entire connection
```
richard-ramos, line 46: I wonder if we should add a compile time warning:
```
    warning("using default connection closeWrite")
```
or something similar so a developer notices that they did not implement their own closeWrite, yet, this aint that bad enough to guarantee an assert.

vladopajic: if connections are not closed correctly they should be catched by `checkTrackers`

vladopajic: so we probably shouldn't care much about warning

libp2p/protocols/perf/client.nim
MorganaFuture: @vladopajic @richard-ramos 
I guess this was merged accidentally? I’ll prepare a patch with the fix this evening when I have some free time.

PR #1618, opened 2025-08-12
fix: pubsub signature verification

gmelodie:
Fixing pubsub signature verification as per [the spec](https://github.com/libp2p/specs/blob/0caf71bb29a525ab0ef14ea4511d68e873d4a2d9/pubsub/README.md?plain=1#L215-L216)


richard-ramos (approved):
Very nice catch!  :-)

vladopajic (approved):

PR #1623, opened 2025-08-13
refactor: Removal of "Hashed" key variant

Ben-PH:
The kad spect specifies that hashing of keys is only used for creating the arguments that go into xordistance calculation. as such, having a "hashed" enum variant of a Key is inconsistent with a typesystem that aligns with the spec.

This would usually mean that all xordistance calculations use a valid sha256 hash of keys. This makes testing problematic, as it is non trivial to use easily reasoned testing data: 256 bits of aparant random value is very noise.

As such, the execution of hash defaults to a chosen implementation, but that can be overriden by specifying an alternative function. for example, noop.

### DIFF
--- a/libp2p/protocols/kademlia/xordistance.nim
+++ b/libp2p/protocols/kademlia/xordistance.nim
@@ -1,4 +1,5 @@
 import ./consts
+import stew/arrayOps
 import ./keys
 import nimcrypto/sha2
 import ../../peerid
@@ -19,9 +20,7 @@ proc noOpHasher*(
     input: seq[byte]
 ): array[IdLength, byte] {.raises: [], nimcall, noSideEffect, gcsafe.} =
   var data: array[IdLength, byte]
-  for i in 0 ..< min(IdLength, input.len):
-    # memcopy instead?
-    data[i] = input[i]
+  discard data.copyFrom(input)
   return data
 
 proc countLeadingZeroBits*(b: byte): int =


libp2p/protocols/kademlia/xordistance.nim
```diff
@@ -31,25 +49,23 @@ proc `<`*(a, b: XorDistance): bool =
 proc `<=`*(a, b: XorDistance): bool =
   cmp(a, b) <= 0
 
-proc hashFor(k: Key): seq[byte] =
+proc hashFor(k: Key, hasher: Opt[XorDHasher]): seq[byte] =
```
Ben-PH, line 52: opt was chosen instead of default values due to my observation that this approach alerts the developer to unintentional malformed function calls. Requiring an explicit choice forces the developer to make a valid decision. Default values leaves the developer vulnerable to overlooking something that requires their attention

richard-ramos (approved):
LGTM

libp2p/protocols/kademlia/xordistance.nim
```diff
@@ -2,8 +2,27 @@ import ./consts
...
+): array[IdLength, byte] {.raises: [], nimcall, noSideEffect, gcsafe.} =
+  return sha256.digest(input).data
+
+# useful for testing purposes
+proc noOpHasher*(
+    input: seq[byte]
+): array[IdLength, byte] {.raises: [], nimcall, noSideEffect, gcsafe.} =
+  var data: array[IdLength, byte]
+  for i in 0 ..< min(IdLength, input.len):
+    # memcopy instead?
```
vladopajic: yes ☺️

vladopajic (approved):

libp2p/protocols/kademlia/xordistance.nim
Ben-PH: ahhh, yes. remember putting in that comment. got lazy about the follow up. I'll change to use memcpy, then run the merge. cheers.

PR #1624, opened 2025-08-13
feat(kad-dht): findPeer

richard-ramos:
Adds an utils function to find specific peers. Also does some filtering around the node's own peerId and append addresses to the peer store instead of replacing them

### DIFF
--- a/tests/testnative.nim
+++ b/tests/testnative.nim
@@ -31,8 +31,9 @@ import
   testnameresolve, testmultistream, testbufferstream, testidentify,
   testobservedaddrmanager, testconnmngr, testswitch, testnoise, testpeerinfo,
   testpeerstore, testping, testmplex, testrelayv1, testrelayv2, testrendezvous,
-  testdiscovery, testyamux, testautonat, testautonatservice, testautorelay, testdcutr,
-  testhpservice, testutility, testhelpers, testwildcardresolverservice, testperf
+  testdiscovery, testyamux, testyamuxheader, testautonat, testautonatservice,
+  testautorelay, testdcutr, testhpservice, testutility, testhelpers,
+  testwildcardresolverservice, testperf
 
 import kademlia/[testencoding, testroutingtable, testfindnode, testputval]
 
--- a/tests/testyamuxheader.nim
+++ b/tests/testyamuxheader.nim
@@ -1,0 +1,321 @@
+{.used.}
+
+# Nim-Libp2p
+# Copyright (c) 2025 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
+import ../libp2p/stream/[bufferstream, lpstream]
+import ./helpers
+
+include ../libp2p/muxers/yamux/yamux
+
+proc readBytes(bytes: array[12, byte]): Future[YamuxHeader] {.async.} =
+  let bs = BufferStream.new()
+  defer:
+    await bs.close()
+
+  await bs.pushData(@bytes)
+
+  return await readHeader(bs)
+
+suite "Yamux Header Tests":
+  teardown:
+    checkTrackers()
+
+  asyncTest "Data header":
+    const
+      streamId = 1
+      length = 100
+      flags = {Syn}
+    let header = YamuxHeader.data(streamId = streamId, length = length, flags)
+    let dataEncoded = header.encode()
+
+    # [version == 0, msgType, flags_high, flags_low, 4x streamId_bytes, 4x length_bytes]
+    const expected = [byte 0, 0, 0, 1, 0, 0, 0, streamId.byte, 0, 0, 0, length.byte]
+    check:
+      dataEncoded == expected
+
+    let headerDecoded = await readBytes(dataEncoded)
+    check:
+      headerDecoded.version == 0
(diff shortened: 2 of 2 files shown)


libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -186,7 +186,12 @@ proc findNode*(
         msg,
         proc(p: PeerInfo) =
           discard kad.rtable.insert(p.peerId)
-          kad.switch.peerStore[AddressBook][p.peerId] = p.addrs
+          # Nodes might return different addresses for a peer, so we append instead of replacing
+          var existingAddresses =
+            kad.switch.peerStore[AddressBook][p.peerId].toHashSet()
```
Ben-PH, line 252: Is there any reason to keep it in a seq instead of hash set? Wolud a BTreeMap/Set be appropriate? it's an ordered, sparse collection, but as discussed earlier, scaling isn't a concern: I'm more thinking that the _nature_ of this collection is "indexible by key, sparse in nature, keys are inherently ordered" which fits the bill for a BTree of some sort quite nicely

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -186,7 +186,12 @@ proc findNode*(
...
+          for a in p.addrs:
```
Ben-PH, line 253: consider using an iterator, such as `mapIt`

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -199,6 +204,24 @@ proc findNode*(
 
   return state.selectClosestK()
 
+proc findPeer*(
+    kad: KadDHT, peer: PeerId
+): Future[Result[PeerInfo, string]] {.async: (raises: [CancelledError]).} =
+  if kad.switch.peerInfo.peerId == peer:
+    return err("can't look for node itself")
```
Ben-PH: observation: if you're looking to get peer info on an id, it's legitimate to get that for yourself. So why is findingPeer returning details that are valid to fetch about yourself, but the rules of this function is that you're not allowed to do that?

Maybe this observation prompts you to see something about the api that deserves adjusting. maybe net. in any case, don't let this observation trigger a bikeshed

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -199,6 +204,24 @@ proc findNode*(
...
+
+  let peerKey = peer.toKey()
+
+  if kad.switch.isConnected(peer):
+    # Return known info about already connected peer
+    return ok(PeerInfo(peerId: peer, addrs: kad.switch.peerStore[AddressBook][peer]))
+
+  let foundNodes = await kad.findNode(peerKey)
```
Ben-PH: inline peerKey. if not inline, at least move to line 218. as read, it signals that it's somewhere in 215-217

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -199,6 +204,24 @@ proc findNode*(
...
+  if foundNodes.len == 0 or not foundNodes.contains(peer):
```
Ben-PH: can it just be the contains call? if it is found, len is not 0, if it's not found, len doesn't matter.

tests/kademlia/testfindnode.nim
```diff
@@ -144,3 +144,34 @@ suite "KadDHT - FindNode":
     await broSwitch.stop()
     await sisSwitch.stop()
     await neiceSwitch.stop()
+
+
+  asyncTest "Find peer":
```
Ben-PH, line 149: If this is meaningfully different to "simple find peer", put in a # comment, as it's not immediately obvious. if it isn't, and this is chosen because "simple find peer" isn't as readable as putting it in this form (understandable), we can drop "simple find peer" (or is it "node"?)

For context, I wrote that test with scalability in mind. we'll probably want to scale up the scenarios being tested, but that's probably a "premature abstraction" thing.

Ben-PH (approved):
LGTM. just some things that you might like to change if they align with your style.

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: Right, i guess i could just return the peerInfo from the switch :thinking:

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: Not sure of the reason why the AddressBook uses a sequence to store its addresses instead of a different data structure. Sadly it's marked as `{.public.}` which makes refactoring harder as it's already part of the Public API 
```
  AddressBook* {.public.} = ref object of SeqPeerBook[MultiAddress]
```

tests/kademlia/testfindnode.nim
richard-ramos: Fixed in last commit

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: Fixed in last commit

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: Fixed in last commit

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: Fixed in last commit

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: I could not find an iterator that would let me do a `forEach` :thinking: .

Issue #1622, opened 2025-08-13
flaky test: IHAVE for message not held by peer triggers IWANT response to sender


vladopajic:
```
  /Users/runner/work/nim-libp2p/nim-libp2p/tests/testall 'GossipSub Integration - Control Messages::IHAVE for message not held by peer triggers IWANT response to sender'
--------------------------------------------------------------------
/Users/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/unittest2-0.2.4-746106a4dfefffce497f1693733f1c1513b5c62c/unittest2.nim(1170) testgossipsubcontrolmessages
/Users/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/unittest2-0.2.4-746106a4dfefffce497f1693733f1c1513b5c62c/unittest2.nim(1097) runDirect
/Users/runner/work/nim-libp2p/nim-libp2p/tests/utils/async_tests.nim(25) runTest`gensym668
/Users/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a/chronos/internal/asyncfutures.nim(649) waitFor
/Users/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a/chronos/internal/asyncfutures.nim(624) pollFor
/Users/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a/chronos/internal/asyncengine.nim(150) poll
/Users/runner/work/nim-libp2p/nim-libp2p/tests/pubsub/integration/testgossipsubcontrolmessages.nim(322) anonymous
/Users/runner/work/nim-libp2p/nim-libp2p/nim/lib/system/fatal.nim(53) sysFatal
[[reraised from:
/Users/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/unittest2-0.2.4-746106a4dfefffce497f1693733f1c1513b5c62c/unittest2.nim(1170) testgossipsubcontrolmessages
[...]

PR #1626, opened 2025-08-14
fix(tests): prevent race condition in testgossipsubcontrolmessages

MorganaFuture:
Wait for actual message receipt instead of fixed timeout.
Links: [Issue:1626](https://github.com/vacp2p/nim-libp2p/issues/1622)


tests/pubsub/integration/testgossipsubcontrolmessages.nim
```diff
@@ -243,7 +243,10 @@ suite "GossipSub Integration - Control Messages":
     # When an IHAVE message is sent
     let p1 = n0.getOrCreatePeer(n1.peerInfo.peerId, @[GossipSubCodec_12])
     n0.broadcast(@[p1], RPCMsg(control: some(ihaveMessage)), isHighPriority = false)
-    await waitForHeartbeat()
+
+    # Wait until IHAVE response is received
+    checkUntilTimeout:
+      receivedIHaves[].len > 0
```
Ben-PH: I'm not familiar with the nature of what's being tested. If the number of receiveds is deterministic, use an ==, if it's non deterministic, and it makes sense that being outside of a certain range is highly unlikely to happen unless there's a pontential issue, use that range, and couple the failure with a message that there is an unlikely chance that it's a false-positive.

the rare false positive is acceptable, in my opinion, if it allows for a true positive to be surfaced. (ETA: I do not hold this opinion strongly, especially in this case, as I do not have an informed understanding of the nature of this check)

MorganaFuture: @Ben-PH Accepted. Now use ==

PR #1627, opened 2025-08-14
refactor: Refine, and reduce, exception scope

Ben-PH:
It is recommended to review this PR with whitespace ignored, as much of it is just changing indentation due to removal of a try scope.

This PR goes through kademlia.nim and attempts to move use of exceptions towards zero. There are still some left, but they are either CancelledError (which is super non-trivial to get rid of), or the dispatchputvalue function. It's late, and I'm not too far from going on holiday, so I thought it appropriate to at least get te content of this PR done, and can come back to that later.

### DIFF
--- a/libp2p/protocols/kademlia/kademlia.nim
+++ b/libp2p/protocols/kademlia/kademlia.nim
@@ -165,41 +165,35 @@ proc putValue*(
   if not kad.entryValidator.isValid(entKey, value):
     return err("invalid key/value pair")
 
+  let others: seq[EntryRecord] =
+    if entKey in kad.dataTable.entries:
+      @[kad.dataTable.entries.getOrDefault(entKey)]
+    else:
+      @[]
+
+  let candAsRec = EntryRecord.init(value, none(TimeStamp))
+  let confirmedRec = kad.entrySelector.select(candAsRec, others).valueOr:
+    error "application provided selector error (local)", msg = error
+    return err(error)
+  trace "local putval", candidate = candAsRec, others = others, selected = confirmedRec
+  let validEnt = ValidatedEntry.init(entKey, confirmedRec.value)
+
+  let peers = await kad.findNode(entKey.data.toKey())
+  # We first prime the sends so the data is ready to go
+  let rpcBatch = peers.mapIt(kad.dispatchPutVal(it, validEnt))
+  # then we do the `move`, as insert takes the data as `sink`
+  kad.dataTable.insert(validEnt, confirmedRec.time)
   try:
-    let others: seq[EntryRecord] =
-      if entKey in kad.dataTable.entries:
-        @[kad.dataTable.entries[entKey]]
-      else:
-        @[]
-
-    let candAsRec = EntryRecord.init(value, none(TimeStamp))
-    let confirmedRec = kad.entrySelector.select(candAsRec, others).valueOr:
-      error "application provided selector error (local)", msg = error
-      return err(error)
-    trace "local putval",
-      candidate = candAsRec, others = others, selected = confirmedRec
-
-    let validEnt = ValidatedEntry.init(entKey, confirmedRec.value)
-
-    let peers = await kad.findNode(entKey.data.toKey())
-    # We first prime the sends so the data is ready to go
-    let rpcBatch = peers.mapIt(kad.dispatchPutVal(it, validEnt))
-    # then we do the `move`, as insert takes the data as `sink`
-    kad.dataTable.insert(validEnt, confirmedRec.time)
-    try:
-      # now that the all the data is where it needs to be in memory, we can dispatch the
-      # RPCs
-      await rpcBatch.allFutures().wait(chronos.seconds(timeout.get(5)))
+    # now that the all the data is where it needs to be in memory, we can dispatch the
+    # RPCs
+    await rpcBatch.allFutures().wait(chronos.seconds(timeout.get(5)))
 
     # It's quite normal for the dispatch to timeout, as it would require all calls to get
     # their response. Downstream users may desire some sort of functionality in the 
     # future to get rpc telemetry, but in the meantime, we just move on...
-    except AsyncTimeoutError:
-      discard
-
-    return results.ok()
-  except CatchableError as e:
-    return err("todo: refine exceptions - " & e.msg)
+  except AsyncTimeoutError:
+    discard
+  return results.ok()
 
 # Helper function forward declaration
 proc checkConvergence(state: LookupState, me: PeerId): bool {.raises: [], gcsafe.}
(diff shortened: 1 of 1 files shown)


libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -212,29 +207,40 @@ proc bootstrap*(
...
-      let msg = await kad.sendFindNode(b.peerId, b.addrs, kad.rtable.selfId).wait(
-        chronos.seconds(5)
-      )
-      for peer in msg.closerPeers:
-        let p = PeerId.init(peer.id).tryGet()
-        discard kad.rtable.insert(p)
+    except DialFailedError as e:
+      # at some point will want to bubble up a Result[void, SomeErrorEnum]
+      error "failed to dial to bootstrap peer", peerId = b.peerId, error = e.msg
+      return
```
richard-ramos: if a bootstrap node is unreachable, the client should skip it and try the next one. It should be expected to include multiple bootstrap addresses so the failure of any single one doesn’t block the bootstrapping process

```suggestion
      continue
```

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -212,29 +207,40 @@ proc bootstrap*(
...
+          chronos.seconds(5)
+        )
+      except CatchableError as e:
+        error "In the future, this will be a Result instead of exception", err = e.msg
+        return
+    for peer in msg.closerPeers:
+      let p = PeerId.init(peer.id).valueOr:
+        debug "error", error = error
+        doAssert(false, "this should never happen")
+        return
```
richard-ramos:
```suggestion
        debug "invalid peer id received", error = error
        continue
```
A node might 'lie' and return garbage for the `id`. In this case  we should skip the peer. Having an assert would nuke the app when this is more of a runtime error
(There's a case here for even considering blacklisting peers that returns invalid data, but that's out of scope for this PR)

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -212,29 +207,40 @@ proc bootstrap*(
...
+        error "this is here because an ergonomic means of keying into a table without exceptions is unknown"
 
-      # bootstrap node replied succesfully. Adding to routing table
-      discard kad.rtable.insert(b.peerId)
-    except CatchableError as e:
-      error "bootstrap failed for peer", peerId = b.peerId, exc = e.msg
+    # bootstrap node replied succesfully. Adding to routing table
+    discard kad.rtable.insert(b.peerId)
+    # except CatchableError as e:
+    #   error "bootstrap failed for peer", peerId = b.peerId, exc = e.msg
```
richard-ramos: Is this no longer needed? if so, maybe we can remove it

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -212,29 +207,40 @@ proc bootstrap*(
...
 
-  try:
-    # Adding some random node to prepopulate the table
-    discard await kad.findNode(PeerId.random(kad.rng).tryGet().toKey())
-    info "bootstrap lookup complete"
-  except CatchableError as e:
-    error "bootstrap lookup failed", error = e.msg
+  let key = PeerId.random(kad.rng).valueOr:
+    doAssert(false, "this should never happen")
```
richard-ramos, line 322: I wonder under what conditions would the rng fail :thinking:

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -267,75 +273,85 @@ proc new*(
...
-          let validated = ValidatedEntry(key: key, value: selectedRec.value)
-
-          kad.dataTable.insert(validated, ts)
-          # consistent with following link, echo message without change
-          # https://github.com/libp2p/js-libp2p/blob/cf9aab5c841ec08bc023b9f49083c95ad78a7a07/packages/kad-dht/src/rpc/handlers/put-value.ts#L22
+        let key = EntryKey(data: skey)
+        let value = EntryVal(data: svalue)
+        let ts = TimeStamp(ts: $times.now().utc)
+
+        # Value sanatisation done. Start insertion process
```
richard-ramos:
```suggestion
        # Value sanitisation done. Start insertion process
```

libp2p/protocols/kademlia/kademlia.nim
Ben-PH:
```nim
func init*(t: typedesc[PeerId], pubkey: PublicKey): Result[PeerId, cstring] =
  ## Create new peer id from public key ``pubkey``.
  var pubraw =
    ?pubkey.getBytes().orError(cstring("peerid: failed to get bytes from given key"))
  var mh: MultiHash
  if len(pubraw) <= maxInlineKeyLength:
    mh = ?MultiHash.digest("identity", pubraw)
  else:
    mh = ?MultiHash.digest("sha2-256", pubraw)
  ok(PeerId(data: mh.data.buffer))
```

this seems to be the only part of the call stack thet can fail. Can't get bytes (could an rng fail to provide gettable bytes?), or cannot digest the gotten bytes. 

this is the potential get-bytes error source:

```nim
proc getRawBytes*(key: PrivateKey | PublicKey): CryptoResult[seq[byte]] =
  ## Return private key ``key`` in binary form (using scheme's own
  ## serialization).
  case key.scheme
  of PKScheme.RSA:
    when supported(PKScheme.RSA):
      key.rsakey.getBytes().orError(KeyError)
    else:
      err(SchemeError)
  of PKScheme.Ed25519:
    when supported(PKScheme.Ed25519):
      ok(key.edkey.getBytes())
    else:
      err(SchemeError)
  of PKScheme.ECDSA:
    when supported(PKScheme.ECDSA):
      key.eckey.getBytes().orError(KeyError)
    else:
      err(SchemeError)
  of PKScheme.Secp256k1:
    when supported(PKScheme.Secp256k1):
      ok(key.skkey.getBytes())
    else:
      err(SchemeError)
```

only considering valid use of compile time generated code, the get bytes can error like so:

```nim
proc getBytes*(key: RsaPrivateKey): RsaResult[seq[byte]] =
  ## Serialize RSA private key ``key`` to ASN.1 DER binary form and
  ## return it.
  if isNil(key):
    return err(RsaKeyIncorrectError)
  var res = newSeqUninit[byte](4096)
  let length = ?key.toBytes(res)
  if length > 0:
    res.setLen(length)
    ok(res)
  else:
    err(RsaKeyIncorrectError)
```

is nil? not our problem. if nil is in your code base, when `Option` is available, then that's on you, the solution is "you shouldn't have used nil in the first place", by my reackoning. if you use `Option[nil]`, same again, but more so.

that leaves us with empty length rng. and my understanding is that won't ever return an empty array, no?

libp2p/protocols/kademlia/kademlia.nim
Ben-PH: image


Yeah, I feel i would neglect my duty to represent my country of origin not to push back on this (mostly for the memes. say the word and I'll change it)

(EDIT: the original suggestion was `sanitization`. With a `z`. Yes, I too, was unable to prevent myself from dropping my monacle...)

Joking aside, for future reference, the standard to use under umbrella organisation is British English: https://guide.logos.co/voice/#style

richard-ramos: Will update the suggestion :)

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -269,29 +262,38 @@ proc bootstrap*(
...
+
+    let msg =
+      try:
+        await kad.sendFindNode(b.peerId, b.addrs, kad.rtable.selfId).wait(
+          chronos.seconds(5)
+        )
+      except CatchableError as e:
+        debug "send find node exception during bootstrap",
+          target = b.peerId, addrs = b.addrs, err = e.msg
+        continue
```
gmelodie, line 307: Do you need this continue?

Ben-PH: it's either continue or return. it makes no sense to progress for this particular bootstrap node, which in itself implies a return is needed, but there may be other bootstrap nodes to try, which means instead of return, a continue is needed.

the rest of the loop is dependant on the `msg` reply being valid.

Issue #1619, opened 2025-08-13
flaky test: NativeSrc -> NativeRelay -> DaemonDst


vladopajic:
````
   D:\a\nim-libp2p\nim-libp2p\tests\testall.exe "Interop relay using yamux::NativeSrc -> NativeRelay -> DaemonDst"
-------------------------------------
D:\a\nim-libp2p\nim-libp2p\tests\testinterop.nim(58) testinterop
D:\a\nim-libp2p\nim-libp2p\nimbledeps\pkgs2\unittest2-0.2.4-746106a4dfefffce497f1693733f1c1513b5c62c\unittest2.nim(1170) relayInteropTests
D:\a\nim-libp2p\nim-libp2p\nimbledeps\pkgs2\unittest2-0.2.4-746106a4dfefffce497f1693733f1c1513b5c62c\unittest2.nim(1097) runDirect
D:\a\nim-libp2p\nim-libp2p\tests\utils\async_tests.nim(25) runTest`gensym1467
D:\a\nim-libp2p\nim-libp2p\nimbledeps\pkgs2\chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a\chronos\internal\asyncfutures.nim(649) waitFor
D:\a\nim-libp2p\nim-libp2p\nimbledeps\pkgs2\chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a\chronos\internal\asyncfutures.nim(624) pollFor
D:\a\nim-libp2p\nim-libp2p\nimbledeps\pkgs2\chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a\chronos\internal\asyncengine.nim(150) poll
D:\a\nim-libp2p\nim-libp2p\nimbledeps\pkgs2\chronos-4.0.4-455802a90204d8ad6b31d53f2efff8ebfe4c834a\chronos\internal\asyncmacro.nim(2536) readOnce
[[reraised from:
D:\a\nim-libp2p\nim-libp2p\tests\testinterop.nim(58) testinterop
D:\a\nim-libp2p\nim-libp2p\nimbledeps\pkgs2\unittest2-0.2.4-746106a4dfefffce497f1693733f1c1513b5c62c\unittest2.nim(1170) relayInteropTests
[...]

PR #1628, opened 2025-08-15
fix(test): Race condition on Windows-specific daemon close

MorganaFuture:
…in relay tests

The "NativeSrc -> NativeRelay -> DaemonDst" relay interop test was failing on Windows due to a race condition during daemon shutdown. The issue stems from platform differences in process termination:

Linux: uses terminate() for graceful SIGTERM shutdown
Windows: uses kill() for forceful process termination

This causes TransportUseClosedError on Windows when the daemon's accept loop attempts to read from an already-closed transport during shutdown.

Links [Issue](https://github.com/vacp2p/nim-libp2p/issues/1619)

### DIFF
--- a/tests/pubsub/integration/testgossipsubcontrolmessages.nim
+++ b/tests/pubsub/integration/testgossipsubcontrolmessages.nim
@@ -243,10 +243,14 @@ suite "GossipSub Integration - Control Messages":
     # When an IHAVE message is sent
     let p1 = n0.getOrCreatePeer(n1.peerInfo.peerId, @[GossipSubCodec_12])
     n0.broadcast(@[p1], RPCMsg(control: some(ihaveMessage)), isHighPriority = false)
-    await waitForHeartbeat()
 
-    # Then the peer has the message ID
+    # Wait until IHAVE response is received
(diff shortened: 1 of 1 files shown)


Ben-PH (approved):
LGTM. 

**Personal** thoughts: Such long branch names are not needed. In trying to share how I try to approach it, I lean towards this style: The branch name is there to help ease the cognitive load while things are being worked on (if "foobar" and "fizzbuzz" tells you and the team the relevant context, then fine!). So long as active branches don't cause confusion in names, you're good. Structured branch naming/conventions are redundant to tools such as conversations on github PRs/issues, commit logs, etc.

With that said, we appreciate the enthusiasm, as well as what I perceive as "err on the side of doing it the right way". That foobar/fizzbuzz example shouldn't come from "err"ing the other way, but from arriving there after a healthy evolution of convention, often starting from places  such as yours

Ben-PH:
@MorganaFuture I invite you to take the honor of hitting auto-merge with a squash if you're ready (if the system allows, I assume the approvals means that anyone can hit that button).

PR #1634, opened 2025-08-20
test(yamux): Add unit tests - frame handling and stream initiation

rlve:
Changes:
- add new test suite for Yamux: `Frame handling and stream initiation`
- new tests:
  - `"Ping Syn responds Ping Ack"`
  - `"Go Away Status responds with Go Away"`
  - 2x `"Syn opens stream and sends Ack"`
    - Data header
    - WindowUpdate header
  - 5x `"Reject invalid/unknown header"`
    - Reserved parity on Data+Syn (even id against responder)
    - Reserved stream id 0
    - First frame missing Syn on unopened stream
    - Reserved parity on WindowUpdate+Syn (even id against responder)
    - Unknown stream WindowUpdate without Syn
  - `"Flush unknown-stream Data up to budget then ProtocolError when exceeded"`

### DIFF
--- a/tests/testyamux.nim
+++ b/tests/testyamux.nim
@@ -430,7 +430,7 @@ suite "Yamux":
 
   suite "Frame handling and stream initiation":
     asyncTest "Ping Syn responds Ping Ack":
-      mSetup(startHandlerA = false)
+      mSetup(startHandlera = false)
 
       let payload: uint32 = 0x12345678'u32
       await conna.write(YamuxHeader.ping(MsgFlags.Syn, payload))
@@ -442,7 +442,7 @@ suite "Yamux":
         header.length == payload
 
     asyncTest "Go Away Status responds with Go Away":
-      mSetup(startHandlerA = false)
+      mSetup(startHandlera = false)
 
       await conna.write(YamuxHeader.goAway(GoAwayStatus.ProtocolError))
 
@@ -457,7 +457,7 @@ suite "Yamux":
       YamuxHeader.windowUpdate(streamId = 5'u32, delta = 0, {Syn}),
     ]:
       asyncTest "Syn opens stream and sends Ack - " & $testCase:
-        mSetup(startHandlerA = false)
+        mSetup(startHandlera = false)
 
         yamuxb.streamHandler = proc(conn: Connection) {.async: (raises: []).} =
           try:
@@ -496,7 +496,7 @@ suite "Yamux":
       YamuxHeader.windowUpdate(streamId = 13'u32, delta = 0),
     ]:
       asyncTest "Reject invalid/unknown header - " & $badHeader:
-        mSetup(startHandlerA = false)
+        mSetup(startHandlera = false)
 
         await conna.write(badHeader)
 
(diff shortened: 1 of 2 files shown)


tests/testyamux.nim
```diff
@@ -13,9 +13,16 @@ import sugar
...
+include ../libp2p/muxers/yamux/yamux
+
 proc newBlockerFut(): Future[void] {.async: (raises: [], raw: true).} =
   newFuture[void]()
 
+proc completedFuture(): Future[void] =
+  let f = newFuture[void]()
+  f.complete()
+  f
+
```
gmelodie: What's the reasoning behind this?

tests/testyamux.nim
```diff
@@ -34,7 +43,14 @@ suite "Yamux":
...
         Yamux.new(connb, windowSize = ws, inTimeout = inTo, outTimeout = outTo)
-      (handlera, handlerb) = (yamuxa.handle(), yamuxb.handle())
+    var
+      handlera = completedFuture()
+      handlerb = completedFuture()
+
+    if startHandlerA:
+      handlera = yamuxa.handle()
+    if startHandlerB:
+      handlerb = yamuxb.handle()
```
gmelodie: Why do this way instead of the old one?

tests/testyamux.nim
```diff
@@ -183,12 +199,12 @@ suite "Yamux":
 
       # Need to exhaust initial window first
       await wait(streamA.write(newSeq[byte](256000)), 1.seconds) # shouldn't block
-      await streamA.write(newSeq[byte](142))
+      await streamA.write(newSeq[byte](160))
```
gmelodie: Maybe we could have a `const` to avoid hardcoding `160`?

gmelodie (approved):

tests/testyamux.nim
vladopajic: @rlve i think we need this on utility level, this is not the first time we need this (in tests and source)

tests/testyamux.nim
```diff
@@ -13,9 +13,16 @@ import sugar
 import chronos
 import ../libp2p/[stream/connection, stream/bridgestream, muxers/yamux/yamux], ./helpers
```
vladopajic, line 17: this line brake annoys me

tests/testyamux.nim
```diff
@@ -24,6 +31,8 @@ suite "Yamux":
       ws: int = YamuxDefaultWindowSize,
       inTo: Duration = 5.minutes,
       outTo: Duration = 5.minutes,
+      startHandlerA = true,
+      startHandlerB = true,
```
vladopajic: usually default value should be `false` (or zero/default value for type) and naming should match.

rlve: I changed the naming to match the variables. I'm going to keep these defaults as true, I don't want to break the backward compatibility and do unnecessary changes.

tests/testyamux.nim
rlve: Moved to `tests/utils/futures`

tests/testyamux.nim
rlve: I init handlerA/handlerB with completedFuture so I can always await them in defer. The tests I added don’t start handler A because they read headers directly, running it would race and cause “only one concurrent read” on BufferStream.

PR #1637, opened 2025-08-22
chore(autonat-v2): add message types

gmelodie:
Adds the following message types (as per [the autonat-v2 spec](https://github.com/libp2p/specs/blob/master/autonat/autonat-v2.md#rpc-messages)), as well as their respective `encode`/`decode` methods:
- `DialRequest`
- `DialResponse`
- `DialDataRequest`
- `DialDataResponse`
- `DialBack`
- `DialBackResponse`
- `AutonatV2Msg` (encapsulates the first four)


libp2p/protocols/connectivity/autonatv2/core.nim
```diff
@@ -0,0 +1,223 @@
...
+    return Opt.none(T)
+  Opt.some(T(data: data))
+
+# AutonatV2Msg
+proc encode*(msg: AutonatV2Msg): ProtoBuffer =
+  result = initProtoBuffer()
+  case msg.msgType
+  of MsgType.Unused:
+    # TODO: error log
+    discard
```
richard-ramos: Use an assertion here

libp2p/protocols/connectivity/autonatv2/core.nim
```diff
@@ -0,0 +1,223 @@
...
+    of MsgType.DialResponse:
+      dialResp*: DialResponse
+    of MsgType.DialDataRequest:
+      dialDataReq*: DialDataRequest
+    of MsgType.DialDataResponse:
+      dialDataResp*: DialDataResponse
+
+# DialRequest
+proc encode*(dialReq: DialRequest): ProtoBuffer =
+  result = initProtoBuffer()
```
richard-ramos: Avoid using result for returning values. (this applies to all code)
https://status-im.github.io/nim-style-guide/language.result.html

libp2p/protocols/connectivity/autonatv2/core.nim
```diff
@@ -0,0 +1,223 @@
...
+  if not ?pb.getRepeatedField(1, addrs).toOpt():
+    return Opt.none(T)
+  if not ?pb.getField(2, nonce).toOpt():
+    return Opt.none(T)
+  Opt.some(T(addrs: addrs, nonce: nonce))
+
+# DialResponse
+proc encode*(dialResp: DialResponse): ProtoBuffer =
+  result = initProtoBuffer()
+  result.write(1, cast[float64](dialResp.status))
```
richard-ramos: Why is the status casted to float64. Enums are encoded as int32 https://protobuf.dev/programming-guides/encoding/#bools-and-enums

libp2p/protocols/connectivity/autonatv2/core.nim
```diff
@@ -0,0 +1,223 @@
...
+  var
+    status: float64
+    addrIdx: uint32
+    dialStatus: uint
+  if not ?pb.getField(1, status).toOpt():
+    return Opt.none(T)
+  if not ?pb.getField(2, addrIdx).toOpt():
+    return Opt.none(T)
+  if not ?pb.getField(3, dialStatus).toOpt():
+    return Opt.none(T)
```
richard-ramos: Only when `status` is `OK`, you'll have data in these fields. In other scenarios addrIdx and dialStatus will be nil or default  (they use proto3) (see: https://github.com/libp2p/go-libp2p/blob/83d77757a2b73d2303724b5f556b817e3ff31d4e/p2p/protocol/autonatv2/server.go#L138-L144) . So if one of those status were being received, you'd be returning Opt.none(DialResponse), when you should really return a Opt.some with only the status.

(Maybe addrIdx and dialStatus should be an Opt?)

libp2p/protocols/connectivity/autonatv2/core.nim
```diff
@@ -0,0 +1,223 @@
...
+
+  DialStatus* {.pure.} = enum
+    Unused = 0
+    EDialError = 100
+    EDialBackError = 101
+    Ok = 200
+
+  DialRequest* = object
+    addrs*: seq[MultiAddress]
+    nonce*: uint64
```
richard-ramos: Woud it make sense to create a fixed64 type in minprotobuf.nim (i.e. type fixed64* = distinct uint64, similar to zint / hint), modify minprotobuf to be aware of this type and use it here? (to avoid having to do any float64 castings)

libp2p/protocols/connectivity/autonatv2/core.nim
```diff
@@ -0,0 +1,223 @@
...
+  result = initProtoBuffer()
+  result.write(1, cast[float64](dialResp.status))
+    # minprotobuf casts uses float64 for fixed64 fields
+  result.write(2, dialResp.addrIdx)
+  result.write(3, dialResp.dialStatus.uint)
+  result.finish()
+
+proc decode*(T: typedesc[DialResponse], pb: ProtoBuffer): Opt[T] =
+  var
+    status: float64
```
richard-ramos: See comment about enum == int32

tests/testautonatv2.nim
```diff
@@ -0,0 +1,120 @@
...
+    let dialReq = DialRequest(
+      addrs:
+        @[
+          MultiAddress.init("/ip4/127.0.0.1/tcp/4040").get(),
+          MultiAddress.init("/ip4/127.0.0.1/tcp/4041").get(),
+        ],
+      nonce: 42,
+    )
+    let encodedReq = dialReq.encode()
+    let decodedReq = DialRequest.decode(encodedReq).get()
```
vladopajic:
```suggestion
    let decodedReq = DialRequest.decode(dialReq.encode()).get()
```

nitpicking this here and other places in this file, because variable is not really needed

tests/testautonatv2.nim
```diff
@@ -0,0 +1,120 @@
...
+    check:
+      decodedReq.addrs == dialReq.addrs
+      decodedReq.nonce == dialReq.nonce
```
vladopajic:
```suggestion
    check decodedReq == dialReq
```

this is not really nitpick but the way you should check whole variables because it makes code more robust because changing variables in types you don't have to update tests. if property is removed then test update will be necessary, but if property is added will you really remember to check if that property is decoded/encoded?

also makes code much more simpler (smaller)

libp2p/protocols/connectivity/autonatv2/core.nim
```diff
@@ -0,0 +1,223 @@
+# Nim-LibP2P
```
vladopajic: this file is usually named "types".

vladopajic (changes requested):
in tests, type assertions should be done on the entire type, not on individual properties.

tests/testautonatv2.nim
vladopajic: it is also possible to avoid `decodedReq`:

```nim
check dialReq == DialRequest.decode(dialReq.encode()).get()
```

vladopajic: it might be sexy to check encode -> decode using template or helper func:
```checkEncodeDecode(dialReq)``` (assuming type parameter can be inferred from value) 

this helper would do everything you do in check but would work for all types

File: libp2p/protocols/connectivity/autonatv2/core.nim [hunk 10]
```diff
@@ -0,0 +1,223 @@
...
+
+type
+  # DialBack and DialBackResponse are not defined as AutonatV2Msg as per the spec
+  # likely because they are expected in response to some other message
+  MsgType* {.pure.} = enum
+    Unused = 0 # nim requires the first variant to be zero
+    DialRequest = 1
+    DialResponse = 2
+    DialDataRequest = 3
+    DialDataResponse = 4
```
vladopajic: do you really have to assign value to enums?
why `DialRequest` does not have value 0?

libp2p/protocols/connectivity/autonatv2/core.nim
```diff
@@ -0,0 +1,223 @@
...
+
+  ResponseStatus* {.pure.} = enum
+    EInternalError = 0
+    ERequestRejected = 100
+    EDialRefused = 101
+    Ok = 200
+
+  DialBackStatus* {.pure.} = enum
+    Ok = 0
```
vladopajic: why does `Ok` has to have value of `0`?

libp2p/protocols/connectivity/autonatv2/core.nim
```diff
@@ -0,0 +1,223 @@
...
+    status*: DialBackStatus
+
+  DialDataRequest* = object
+    addrIdx*: uint32
+    numBytes*: uint64
+
+  DialDataResponse* = object
+    data*: seq[byte]
+
+  AutonatV2Msg* = object # TODO: case...
```
vladopajic: comment ` # TODO: case... ` not really usefully for other devs....

libp2p/protocols/connectivity/autonatv2/core.nim
gmelodie: I'm just following the names in autonat v1. Do you still want me to change it?

vladopajic: yes. we should rename that file as well

libp2p/protocols/connectivity/autonatv2/core.nim
gmelodie: This looks weird indeed, but apparently our [minprotobuf impl converts `fixed64` fields to `float64`?](https://github.com/vacp2p/nim-libp2p/blob/25a8ed4d07fc7e9a80a2794d6e78ced9642a7acc/libp2p/protobuf/minprotobuf.nim#L52). I wasn't sure if that was a problem or not, buf if it is I can open an issue and fix it.

obs: same happens for `fixed32` and `float32`

libp2p/protocols/connectivity/autonatv2/core.nim
gmelodie: Oh because [the spec](https://github.com/libp2p/specs/blob/master/autonat/autonat-v2.md#rpc-messages) dictates that each type of message uses a particular protobuf field. So if the message is a `DialRequest`, it'll use field number `1`. Defining it this way allows us to do:
```nim
proc encode*(msg: AutonatV2Msg): ProtoBuffer =
  result = initProtoBuffer()
  case msg.msgType
  of MsgType.Unused:
    doAssert false
  of MsgType.DialRequest:
    result.write(MsgType.DialRequest.int, msg.dialReq.encode())
  of MsgType.DialResponse:
    result.write(MsgType.DialResponse.int, msg.dialResp.encode())
  of MsgType.DialDataRequest:
    result.write(MsgType.DialDataRequest.int, msg.dialDataReq.encode())
  of MsgType.DialDataResponse:
    result.write(MsgType.DialDataResponse.int, msg.dialDataResp.encode())
  result.finish()
```
Instead of 
```nim
proc encode*(msg: AutonatV2Msg): ProtoBuffer =
  result = initProtoBuffer()
  case msg.msgType
  of MsgType.Unused:
    doAssert false
  of MsgType.DialRequest:
    result.write(1, msg.dialReq.encode()) # requires hardcoded '1', or a const
  of MsgType.DialResponse:
    result.write(2, msg.dialResp.encode())
  of MsgType.DialDataRequest:
    result.write(3, msg.dialDataReq.encode())
  of MsgType.DialDataResponse:
    result.write(4, msg.dialDataResp.encode())
  result.finish()
```
It is dumb that enums are required to start with 0.

libp2p/protocols/connectivity/autonatv2/core.nim
gmelodie: This is from [the spec](https://github.com/libp2p/specs/blob/master/autonat/autonat-v2.md#rpc-messages)

libp2p/protocols/connectivity/autonatv2/core.nim
gmelodie: My bad, this was reminiscent of a time where there was no `case` inside that type (now there is :)

tests/testautonatv2.nim
gmelodie: 100%

libp2p/protocols/connectivity/autonatv2/core.nim
gmelodie: I see what you mean, but I should be noted that the [spec](https://github.com/libp2p/specs/blob/master/autonat/autonat-v2.md#autonat-v2-protocol) doesn't really say "Only when status is OK, you'll have data in these fields.":
> Upon completion of the dial back, the server sends a DialResponse message to the client node on the /libp2p/autonat/2/dial-request stream. The response contains addrIdx, the index of the address the server selected to dial and DialStatus, a dial status indicating the outcome of the dial back.

tests/testautonatv2.nim
```diff
@@ -0,0 +1,86 @@
...
+    # DialResponse
+    let dialResp = DialResponse(
+      status: ResponseStatus.Ok,
+      addrIdx: Opt.some(1.uint32),
+      dialStatus: Opt.some(DialStatus.Ok),
+    )
+    checkEncodeDecode(dialResp)
+
+    # DialDataRequest
+    let dataReq = DialDataRequest(addrIdx: 0, numBytes: 128)
```
vladopajic: `addrIdx: 0` do not use default value (aka zero value).

tests/testautonatv2.nim
```diff
@@ -0,0 +1,86 @@
...
+      AutonatV2Msg(msgType: MsgType.DialDataResponse, dialDataResp: dataResp)
+    checkEncodeDecode(msgDataResp)
+
+    # DialBack
+    let dialBack = DialBack(nonce: 123456)
+    checkEncodeDecode(dialBack)
+
+    # DialBackResponse
+    let dialBackResp = DialBackResponse(status: DialBackStatus.Ok)
+    checkEncodeDecode(dialBackResp)
```
vladopajic: variable `dialBackResp` is not needed...

tests/testautonatv2.nim
```diff
@@ -0,0 +1,86 @@
...
+    checkEncodeDecode(msgDataReq)
+
+    # AutonatV2Msg - DialDataResponse
+    let msgDataResp =
+      AutonatV2Msg(msgType: MsgType.DialDataResponse, dialDataResp: dataResp)
+    checkEncodeDecode(msgDataResp)
+
+    # DialBack
+    let dialBack = DialBack(nonce: 123456)
+    checkEncodeDecode(dialBack)
```
vladopajic: variable `dialBack` is not needed....

tests/testautonatv2.nim
```diff
@@ -0,0 +1,86 @@
...
+
+    # AutonatV2Msg - DialDataRequest
+    let msgDataReq =
+      AutonatV2Msg(msgType: MsgType.DialDataRequest, dialDataReq: dataReq)
+    checkEncodeDecode(msgDataReq)
+
+    # AutonatV2Msg - DialDataResponse
+    let msgDataResp =
+      AutonatV2Msg(msgType: MsgType.DialDataResponse, dialDataResp: dataResp)
+    checkEncodeDecode(msgDataResp)
```
vladopajic: variable `msgDataResp` is not needed...

libp2p/protocols/connectivity/autonatv2/types.nim
```diff
@@ -0,0 +1,240 @@
...
+  if not ?pb.getField(1, data).toOpt():
+    return Opt.none(T)
+  Opt.some(T(data: data))
+
+# AutonatV2Msg
+proc encode*(msg: AutonatV2Msg): ProtoBuffer =
+  var encoded = initProtoBuffer()
+  case msg.msgType
+  of MsgType.Unused:
+    doAssert false
```
vladopajic: add message to `doAssert`

vladopajic (approved):
lgtm, some nitpicks left

libp2p/protocols/connectivity/autonatv2/core.nim
richard-ramos: Indeed :thinking: . 

This seems to be one of the cases in which the specs is ambiguous when compared to the implementations.  
Went over to the rust impl to see how they handle this, and looks like they also have their own interpretation:
https://github.com/libp2p/rust-libp2p/blob/1755bca475928ad0e603c42b98c29107c88816cb/protocols/autonat/src/v2/server/handler/dial_request.rs#L161-L184

With the following cases:
- Status = Internal Error - AddrIdx = (some parameter), DialStatus = Unused
- Status = Rejected - AddrIdx = 0, DialStatus = Unused
- Status = Refused - AddrIdx = 0, DialStatus = Unused
- Status = Ok - AddrIdx = (some parameter), DialStatus = (some status)

libp2p/protocols/connectivity/autonatv2/types.nim
vladopajic: or use `raiseAssert`, not really sure about difference from top of the head

tests/testautonatv2.nim
vladopajic: and other places... would be nicer to kill all those variables that expire on the next line

tests/testautonatv2.nim
vladopajic: because then we are not really sure if property is decoded...

libp2p/protocols/connectivity/autonatv2/core.nim
richard-ramos: Go:  https://github.com/libp2p/go-libp2p/blob/83d77757a2b73d2303724b5f556b817e3ff31d4e/p2p/protocol/autonatv2/server.go

Status: Rejected,   Other fields: unspecified  (probably will be protobuf defaults)
Status: Refused,   Other fields: unspecified  (probably will be protobuf defaults)
Status: Internal Error is not used at all  :thinking: 
Status: Ok, AddrIdx = (some parameter), DialStatus = (some status)

richard-ramos: JS: https://github.com/libp2p/js-libp2p/blob/main/packages/protocol-autonat-v2/src/server.ts

-  Status = Refused - AddrIdx = some parameter, DialStatus = Unused
-  Status = Ok - AddrIdx = (some parameter), DialStatus = (some status)
-  Status = Internal Error  - Not used at all? :thinking: 
-  Status = Rejected   - Not used at all? :thinking:

richard-ramos: I guess the TLDR is we shouldnt reject if fields other than status aint present

gmelodie: Yes, ~~I'll make~~ I made them optional

tests/testautonatv2.nim
gmelodie: done!

libp2p/protocols/connectivity/autonatv2/core.nim
richard-ramos: Based on comments here: https://github.com/vacp2p/nim-libp2p/pull/1637#discussion_r2296039129 it seems some implementations do use UNUSED, so this comment will make sense only if you decide to use unused too :grimacing: !

libp2p/protocols/connectivity/autonatv2/core.nim
richard-ramos: looking at the specs, looks like fixed64 is only used on autonat spec, so doing this might be more of a nice to have and not really required

PR #1640, opened 2025-08-26
fix(gossipsub): split IWANT replies when size exceeds maxMessageSize

MorganaFuture:
Individual messages below maxSize but combined exceeding maxSize were causing "trying to send a msg too big for pubsub" errors. This fix splits IWANT reply messages into batches before sending to avoid transport size limits.

Links: [Issue](https://github.com/vacp2p/nim-libp2p/issues/1603)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -355,6 +355,51 @@ proc handleSubscribe(g: GossipSub, peer: PubSubPeer, topic: string, subscribe: b
...
+  ## to avoid exceeding maxMessageSize limits
+
+  if messages.len == 0:
+    return
+
+  trace "sending iwant reply messages", peer, count = messages.len
+
+  # Split messages into batches to avoid exceeding maxMessageSize
+  # Use 90% of maxMessageSize to account for protobuf overhead
+  let maxSafeSize = (g.maxMessageSize * 90) div 100
```
gmelodie, line 369: Where does 90% come from? Is this in a spec somewhere? Could we get away with 95% or more?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -355,6 +355,51 @@ proc handleSubscribe(g: GossipSub, peer: PubSubPeer, topic: string, subscribe: b
...
+
+  for msg in messages:
+    let msgSize = byteSize(msg)
+    trace "processing message",
+      msgSize = msgSize, currentSize = currentSize, maxSafeSize = maxSafeSize
+
+    # Skip messages that are individually too large
+    if msgSize > maxSafeSize:
+      trace "skipping oversized message", msgSize = msgSize, maxSafeSize = maxSafeSize
+      continue
```
gmelodie: Should there be a TODO here to split these into smaller messages?

libp2p/protocols/pubsub/gossipsub.nim
MorganaFuture: The 90% is a conservative safety buffer, existing code in pubsubpeer already uses 10% overhead. But actual protobuf overhead is low, and i guest we can safe increase it to 95%

libp2p/protocols/pubsub/gossipsub.nim
MorganaFuture: Message fragmentation would break signatures since they cover the entire message including the data field. It would also need protocol changes for reassembly and other libp2p implementations don't do this.
Am I missing something, or would this approach face fundamental issues that make it impractical?

MorganaFuture: P.S. I don't claim to be an expert, I'm just trying to figure it out.

libp2p/protocols/pubsub/gossipsub.nim
gmelodie: The question here is how this affects performance and what's the right number. I believe @richard-ramos and @ufarooqstatus can provide more information.

libp2p/protocols/pubsub/gossipsub.nim
gmelodie: Makes sense! Just pinging @richard-ramos to be sure.

gmelodie (approved):

libp2p/protocols/pubsub/gossipsub.nim
vladopajic, line 369: nit:

```suggestion
  let maxSafeSize = g.maxMessageSize * 0.9
```

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -355,6 +355,51 @@ proc handleSubscribe(g: GossipSub, peer: PubSubPeer, topic: string, subscribe: b
...
+  var currentSize = 0
+
+  for msg in messages:
+    let msgSize = byteSize(msg)
+    trace "processing message",
+      msgSize = msgSize, currentSize = currentSize, maxSafeSize = maxSafeSize
+
+    # Skip messages that are individually too large
+    if msgSize > maxSafeSize:
+      trace "skipping oversized message", msgSize = msgSize, maxSafeSize = maxSafeSize
```
vladopajic: trace -> warn  
since this should catch someones eyes

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -355,6 +355,51 @@ proc handleSubscribe(g: GossipSub, peer: PubSubPeer, topic: string, subscribe: b
 
   trace "gossip peers", peers = g.gossipsub.peers(topic), topic
 
+proc sendIWantMessages(g: GossipSub, peer: PubSubPeer, messages: seq[Message]) =
```
vladopajic, line 358: not sure if we need `trace` log in all those branches seems too much

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -355,6 +355,51 @@ proc handleSubscribe(g: GossipSub, peer: PubSubPeer, topic: string, subscribe: b
...
+      trace "sending iwant reply batch",
+        peer, batchSize = currentBatch.len, totalSize = currentSize
+      g.send(peer, RPCMsg(messages: currentBatch), isHighPriority = false)
+      currentBatch = @[msg]
+      currentSize = msgSize
+    else:
+      currentBatch.add(msg)
+      currentSize += msgSize
+      trace "added message to batch",
+        batchSize = currentBatch.len, totalSize = currentSize
```
vladopajic: nit:

following if statement suits more to the comment; and we kinda always what to do thing from else block right?
```suggestion
    if currentSize + msgSize > maxSafeSize:
      g.send(peer, RPCMsg(messages: currentBatch), isHighPriority = false)
      currentBatch = @[]
      currentSize = 0

    currentBatch.add(msg)
    currentSize += msgSize
 ```

libp2p/protocols/pubsub/gossipsub.nim
vladopajic: @vacp2p/p2p please confirm

gmelodie: Agree

libp2p/protocols/pubsub/gossipsub.nim
MorganaFuture: I catch type error so i should write something like this: 
`int(float(g.maxMessageSize) * 0.9)`  seems strange to me convert int->float->int just get 90%. Am I doing this correctly?

libp2p/protocols/pubsub/gossipsub.nim
MorganaFuture: Done

libp2p/protocols/pubsub/gossipsub.nim
MorganaFuture: Accepted

libp2p/protocols/pubsub/gossipsub.nim
MorganaFuture: Accepted

libp2p/protocols/pubsub/gossipsub.nim
vladopajic: yes you are..

if you care about conversions then you don't really need a final int conversion - having value in float is fine.
and if you do `0.9 * maxMessageSize`, having float `0.9` first then you will not need float conversion.

vladopajic: didn't really knew, actually type conversions are need in nim...

PR #1646, opened 2025-08-28
fix(rendezvous): pagination offset

rlve:
Fix cookie offset handling in discover pagination that caused skipped entries across consecutive requests.

Fix:
- Define the cookie offset as “next index to read.”
- Reset cookies to the current window: underflow → low; overflow → high + 1 (empty page).
- Reset on namespace mismatch to that namespace’s current low.
- Treat high + 1 as “end of window,” so new entries appear on the next request.

New tests:
- `"Request with namespace pagination with multiple namespaces"`
- `"Request with namespace with expired peers"`
- `"Cookie offset is reset to end (returns empty) then new peers are discoverable"`
- `"Cookie offset is reset to low after flush (returns current entries)"`
- `"Cookie namespace mismatch resets to low (returns peers despite offset)"`

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -1,11 +1,6 @@
 name: Performance
 
 on:
-  push:
-    branches:
-      - master
-  pull_request:
-  merge_group:
   workflow_dispatch:
 
 concurrency:
--- a/libp2p/muxers/yamux/yamux.nim
+++ b/libp2p/muxers/yamux/yamux.nim
@@ -533,13 +533,13 @@ method close*(m: Yamux) {.async: (raises: []).} =
   trace "Closing yamux"
   let channels = toSeq(m.channels.values())
   for channel in channels:
-    for toSend in channel.sendQueue:
-      toSend.fut.fail(newLPStreamEOFError())
-    channel.sendQueue = @[]
+    channel.clearQueues(newLPStreamEOFError())
+    channel.recvWindow = 0
     channel.sendWindow = 0
     channel.closedLocally = true
     channel.isReset = true
     channel.opened = false
+    channel.isClosed = true
     await channel.remoteClosed()
     channel.receivedData.fire()
   try:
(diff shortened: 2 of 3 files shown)


libp2p/protocols/rendezvous.nim
```diff
@@ -841,3 +847,29 @@ method stop*(rdv: RendezVous): Future[void] {.async: (raises: [], raw: true).} =
   rdv.registerDeletionLoop.cancelSoon()
   rdv.registerDeletionLoop = nil
   fut
+
+# Test-only accessors to internal fields
+# Enable with: -d:libp2p_tests_private
+when defined(libp2p_tests_private):
+  type TestRendezVous* = ref object
+    rdv*: RendezVous
```
vladopajic: i don't think we should have this as practice. currently we expose field/proc if we need it.
this approach seems hard to use and maintain if applied regularly.

rlve: I wasn't sure what would be the best thing to do and if changing the public interface is acceptable. But in that case I'll just expose what I need for the tests.

libp2p/protocols/rendezvous.nim
```diff
@@ -468,11 +468,17 @@ proc discover(
...
         return
     else:
-      Cookie(offset: rdv.registered.low().uint64 - 1)
-  if d.ns.isSome() and cookie.ns.isSome() and cookie.ns.get() != d.ns.get() or
-      cookie.offset < rdv.registered.low().uint64 or
-      cookie.offset > rdv.registered.high().uint64:
-    cookie = Cookie(offset: rdv.registered.low().uint64 - 1)
+      # Start from the current lowest index (inclusive)
+      Cookie(offset: rdv.registered.low().uint64)
+  if d.ns.isSome() and cookie.ns.isSome() and cookie.ns.get() != d.ns.get():
```
richard-ramos, line 473: Thanks for making this condition more readable

richard-ramos (approved):

PR #1649, opened 2025-08-29
feat(autonatv2): add AutonatV2 client and server

gmelodie:


libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,117 @@
...
+  )
+  let messagesToSend =
+    (req.numBytes + MaxDialDataResponsePayload - 1) div MaxDialDataResponsePayload
+  for i in 0 ..< messagesToSend:
+    await conn.send(msg)
+    debug "sending dialdataresponse", i = i, messagesToSend = messagesToSend
+
+  # get DialResponse
+  msg = AutonatV2Msg.decode(initProtoBuffer(await conn.readLp(1024))).valueOr:
+    raise newException(AutonatV2Error, "Unable to decode message")
```
vladopajic: > Unable to decode message

okay, but will developer know what message is that based on logs?

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,117 @@
...
+
+type AutonatV2Client* = ref object of RootObj
+  rng: ref HmacDrbgContext
+  nonces: Table[PeerId, uint64]
+
+proc new*(T: typedesc[AutonatV2Client], rng: ref HmacDrbgContext, switch: Switch): T =
+  let client = T(rng: rng, nonces: initTable[PeerId, uint64]())
+  client
+
+proc handleDialDataRequest*(
```
vladopajic, line 94: `handleDialDataRequest` anyone calling this method has to do two things to handle error:
 - check if `Opt` values is set
 - `try/except` to handle errors 
 
 how come `handleDialDataRequest` can return `Opt.none(...)` and that is not error case?

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,117 @@
...
+
+  return Opt.some(msg.dialResp)
+
+proc sendDialRequest*(
+    self: AutonatV2Client,
+    switch: Switch,
+    pid: PeerId,
+    peerAddrs: seq[MultiAddress],
+    testAddrs: seq[MultiAddress],
+): Future[Opt[DialResponse]] {.
```
vladopajic: similar comment with `Opt` as in `handleDialDataRequest`

tests/testautonatv2.nim
```diff
@@ -107,3 +119,32 @@ suite "AutonatV2":
...
+    defer:
+      await allFutures(src.stop(), dst.stop())
+
+    await src.connect(dst.peerInfo.peerId, dst.peerInfo.addrs)
+    let client = AutonatV2Client.new(newRng(), src)
+    let dialResp = (
+      await client.sendDialRequest(
+        src, dst.peerInfo.peerId, dst.peerInfo.addrs, src.peerInfo.addrs
+      )
+    ).get()
```
vladopajic: issues pointed with comments related to `Opt` in return are shown here, code gets result like it's given.

tests/testautonatv2.nim
```diff
@@ -107,3 +119,32 @@ suite "AutonatV2":
 
     # DialBackResponse
     checkEncodeDecode(DialBackResponse(status: DialBackStatus.Ok))
+
+  asyncTest "Successful DialRequest":
+    let
+      src = newStandardSwitch()
```
vladopajic: not really sure if we should use `newStandardSwitch` here (because of comment from above).  
- changes in newStandardSwitch can easily break this code
- sure there is comment; but we should avoid it if possible 

could we have `newStandardSwitchBulder()`; that is called in `newStandardSwitch` and `createAutonatV2Switch` should use it and just add autonav2??

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,231 @@
...
+  for addr in reqAddrs:
+    let maFirst = addr[0].valueOr:
+      continue
+    if maFirst == observedIPAddr:
+      return true
+  return false
+
+proc chooseDialAddr(
+    self: AutonatV2, pid: PeerId, addrs: seq[MultiAddress]
+): Future[Opt[uint32]] {.
```
vladopajic: why not return `MultiAddress`?

libp2p/protocols/connectivity/autonatv2/client.nim
gmelodie: Hummm indeed. However the problem is that message could be total garbage, or only partial garbage. In both cases decoding would fail. Should we print the raw bytes in a debug log?

libp2p/protocols/connectivity/autonatv2/types.nim
```diff
@@ -239,3 +249,47 @@ proc decode*(T: typedesc[AutonatV2Msg], pb: ProtoBuffer): Opt[T] =
 # Custom `==` is needed to compare since AutonatV2Msg is a case object
 proc `==`*(a, b: AutonatV2Msg): bool =
   a.msgType == b.msgType and a.encode() == b.encode()
+
+proc connect*(
```
richard-ramos: Is types.nim the right place for code from 253 to 295?

libp2p/protocols/connectivity/autonatv2/types.nim
```diff
@@ -239,3 +249,47 @@ proc decode*(T: typedesc[AutonatV2Msg], pb: ProtoBuffer): Opt[T] =
...
+    switch: Switch, pid: PeerId, addrs: seq[MultiAddress], codec: AutonatV2Codec
```
richard-ramos: What's the advantage of having these connect/send functions compared to calling the underliying switch and conn functions directly?  Is it because of wrapping errors with AutonatV2Error? because if so, an alternative to having to define these functions would be to create a utils template like this:

```nim
template rewrap[E](outer: typedesc[E], prefix: string, body: untyped): untyped =
  try:
    body
  except CancelledError as exc:
    raise exc
  except DialFailedError as exc:
    raise newException(outer, prefix & exc.msg, exc)
```

and you would use it like this
```nim
let conn = rewrap(AutonatV2Error, "unexpected error when dialling: "):
  await switch.dial(pid, addrs, $codec)
...


rewrap(AutonatV2Error, "Sending DialBack failed"):
    trace "Sending DialBack", dialBack = dialBack
    await conn.writeLp(dialBack.encode().buffer)
```

libp2p/protocols/connectivity/autonatv2/client.nim
vladopajic: ` "Unable to decode AutonatV2Msg message"` should do it

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,117 @@
...
+const
+  MaxAcceptedDialDataRequest* = 100 * 1024 # 100 KB
+  MaxDialDataResponsePayload* = 1024
+
+type AutonatV2Client* = ref object of RootObj
+  rng: ref HmacDrbgContext
+  nonces: Table[PeerId, uint64]
+
+proc new*(T: typedesc[AutonatV2Client], rng: ref HmacDrbgContext, switch: Switch): T =
+  let client = T(rng: rng, nonces: initTable[PeerId, uint64]())
```
richard-ramos: In nim tables no longer require to be initialized since version 0.2

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,117 @@
...
+
+const
+  MaxAcceptedDialDataRequest* = 100 * 1024 # 100 KB
+  MaxDialDataResponsePayload* = 1024
+
+type AutonatV2Client* = ref object of RootObj
+  rng: ref HmacDrbgContext
+  nonces: Table[PeerId, uint64]
+
+proc new*(T: typedesc[AutonatV2Client], rng: ref HmacDrbgContext, switch: Switch): T =
```
richard-ramos: Looks like the switch parameter is not used at all. Should it be an attribute of AutonatV2Client?

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,117 @@
...
+  let client = T(rng: rng, nonces: initTable[PeerId, uint64]())
+  client
+
+proc handleDialDataRequest*(
+    conn: Connection, req: DialDataRequest
+): Future[Opt[DialResponse]] {.
+    async: (raises: [CancelledError, AutonatV2Error, LPStreamError])
+.} =
+  debug "handleDialDataRequest"
```
richard-ramos: trace vs debug

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,117 @@
...
+  debug "Receive message", msgType = $msg.msgType
+  if msg.msgType != MsgType.DialResponse:
+    raise
+      newException(AutonatV2Error, "Expecting DialResponse, but got " & $msg.msgType)
+
+  return Opt.some(msg.dialResp)
+
+proc sendDialRequest*(
+    self: AutonatV2Client,
+    switch: Switch,
```
richard-ramos: Should this recevie the switch parameter or should it use the switch from `new`

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,231 @@
...
+
+  return Opt.some(hostIp)
+
+proc dialBack(
+    self: AutonatV2, pid: PeerId, addrs: MultiAddress, nonce: uint64
+): Future[DialStatus] {.
+    async: (raises: [CancelledError, AutonatV2Error, LPStreamError])
+.} =
+  # dial peer on dial-back codec/stream
+  let conn = await self.switch.connect(pid, @[addrs], AutonatV2Codec.DialBack)
```
richard-ramos: This will likely reuse the same connection, and just open a new stream. Do look into autonatV1 how it uses `tryDial` instead, which bypasses the connection manager entirely:
```
futs = addrs.mapIt(autonat.switch.dialer.tryDial(conn.peerId, @[it]))
```
tryDial however closes the connection as soon as it was succesful. So probably we need something like tryDial but that doenst close the connection immediatly, but returns it instead.

libp2p/protocols/connectivity/autonatv2/server.nim
gmelodie: Oh this is from the spec, the DialResponse will include a `uint32` called `addrIdx`, the index for address dialed in the original address list offered by the client

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,231 @@
...
+  topics = "libp2p autonat v2 server"
+
+const
+  DefaultDialTimeout: Duration = 15.seconds
+  DefaultAmplificationAttackDialTimeout: Duration = 3.seconds
+  DefaultDialDataSize: uint64 = 50 * 1024 # 50 KiB > 50 KB
+
+type AutonatV2* = ref object of LPProtocol
+  switch*: Switch
+  dialTimeout: Duration
```
richard-ramos, line 35: This doesnt seem to be used and probably should!

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,231 @@
...
+  return false
+
+proc chooseDialAddr(
+    self: AutonatV2, pid: PeerId, addrs: seq[MultiAddress]
+): Future[Opt[uint32]] {.
+    async: (raises: [CancelledError, AutonatV2Error, LPStreamError])
+.} =
+  for i, ma in addrs:
+    try:
+      if (await self.switch.dialer.tryDial(pid, @[ma])).isSome():
```
richard-ramos: I think this shouldnt be done here. You'd use a new connection for doing the dialback and send the `DialBack` protobuf, yet here you're connecting and if succesful, disconnecting, for later connecting again to send the DialBack.

What chooseDialAddr should do is return an address from the list to dial, but not dial itself to the address:

> the server selects an address from the list to dial. The server SHOULD use the first address it is willing to dial. The server MUST NOT dial any address other than this one.

You have to make sure that the chosen address matches this:
> Client SHOULD NOT send any private address as defined in [RFC 1918](https://datatracker.ietf.org/doc/html/rfc1918#section-3) in the list. The Server SHOULD NOT dial any private address.

(The spec says this wrt client side, but server must ensure also that client followed this rule or disconnect otherwise since it's a protocol violation error)

For testing this might be a bit annoying, so probably we need a setting that allows us to bypass this rule

richard-ramos (commented):
I still havent reviewed client. This only takes into acount types.nim and server.nim :)

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,165 @@
...
+
+  # handler for DialBack messages
+  proc handleStream(
+      conn: Connection, proto: string
+  ) {.async: (raises: [CancelledError]).} =
+    try:
+      let dialBack = DialBack.decode(initProtoBuffer(await conn.readLp(1024))).valueOr:
+        raise newException(AutonatV2Error, "Unable to decode DialBack")
+      debug "Received DialBack", nonce = $dialBack.nonce
+      await client.handleDialBack(conn, dialBack)
```
richard-ramos: Probably makes sense to have a timeout here

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,165 @@
...
+
+  return msg.dialResp
+
+proc sendDialRequest*(
+    self: AutonatV2Client,
+    switch: Switch,
+    pid: PeerId,
+    peerAddrs: seq[MultiAddress],
+    testAddrs: seq[MultiAddress],
+): Future[DialResponse] {.
```
richard-ramos: Probably makes sense to return an object that contains the multiaddress (instead of the index), the dial status, and the reachability status similar to AutoNatV1. We also gotta be sure that we received a Dialback on the same address the server is informing us it dialed us back in the DialResponse (this should likely be done in line 154 and 156).   
```
NetworkReachability* {.pure.} = enum
    Unknown
    NotReachable
    Reachable
```

This is because we can't trust what the server is telling us. Server might say it dialed some address, but we didnt really received a dial back in such address.

Wrt Reachability, the values we should use are: `Reachable` on the happy path,  `NotReachable` when we receive a DIAL_ERROR, and unknown for DIAL_BACK error. Unknown for anythign else

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,165 @@
...
+    async: (raises: [CancelledError, AutonatV2Error, LPStreamError])
+.} =
+  debug "Received DialDataRequest",
+    numBytes = req.numBytes, maxAcceptedNumBytes = MaxAcceptedDialDataRequest
+
+  if req.numBytes > MaxAcceptedDialDataRequest:
+    raise newException(
+      AutonatV2Error, "Rejecting DialDataRequest: numBytes is greater than the maximum"
+    )
+
```
richard-ramos, line 106: We should probably validate too that the AddrIndex received here is valid (maybe it'a negative number or a number greater than the number of addresses we sent)

libp2p/protocols/connectivity/autonatv2/server.nim
vladopajic: fu*k the specs 


return and use what you need

vladopajic: aha, haven't seen it in DialResponse on the first glance

libp2p/protocols/connectivity/autonatv2/client.nim
gmelodie: Using it now

libp2p/protocols/connectivity/autonatv2/types.nim
gmelodie: Removed it entirely

libp2p/protocols/connectivity/autonatv2/types.nim
gmelodie: I think we can use defaul methods and get away with just `try...except`ing whenever we need (avoiding to have new obscure template)

libp2p/protocols/connectivity/autonatv2/client.nim
gmelodie: I see what you mean, but I don't think we should be not trusting the server. If we don't trust the server, what's the point of requesting a dial in the first place? Imagine the server dials the _second_ address it's wiling to dial, as opposed to the _first_ one, as it should. There's no way for the client to know that the dialed address is not the first one the server could dial.

gmelodie: I'll implement the check, I just think it's something to think about (maybe the spec could be improved?)

richard-ramos: It's all part of "Trust but verify":  With autonat your local perspective is incomplete, so you ask servers to give you their perspective about you, but you cannot trust them blindly to tell you the truth that you are 'reachable' on some specific address. 

For example, if you are using a circuit relay addressm and via autonat some server tells you you are reachable, you will then stop using a circuit relay reservation because you will think that real peers can reach you, when in reality they cannot. Once you no longer maintain the circuit relay reservation, other peers would not be able to reach you, so the server effectively isolated you.

PR #1650, opened 2025-09-01
remove macos-13 from the matrix

narimiran:
It is no longer supported, starting from September 1st: https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,8 +25,6 @@ jobs:
             cpu: i386
           - os: linux-gcc-14
             cpu: amd64
-          - os: macos
-            cpu: amd64
           - os: macos-14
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -45,10 +43,6 @@ jobs:
               os: linux-gcc-14
             builder: ubuntu-24.04
             shell: bash
-          - platform:
-              os: macos
-            builder: macos-13
-            shell: bash
```
gmelodie, line 51: I wonder if we want to point `os: macos` to `builder: macos-14`

PR #1653, opened 2025-09-01
fix(quic): add commontransport tests

gmelodie:
Tentative PR. I wonder why we don't have commontransport tests in quic.

https://github.com/vacp2p/nim-libp2p/issues/1652

### DIFF
--- a/tests/testquic.nim
+++ b/tests/testquic.nim
@@ -12,7 +12,7 @@ import
     errors,
     wire,
   ]
-import ./helpers
+import ./helpers, ./commontransport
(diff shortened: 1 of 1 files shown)


richard-ramos (approved):

vladopajic:
quic = transport + muxer
commontransport tests only work for transport 

so they could not be used for quic without major changes in tests or quic.

---

in https://github.com/vacp2p/nim-libp2p/pull/1662 some tests that make sense are copied from commontransport to quic tests.

PR #1655, opened 2025-09-02
test(rendezvous): Registration TTL tests

rlve:
Changes:
- small refactor of `rendezvous`:
  - use constants in the TTL validation
  - add `MinimumAcceptedDuration` constant and comments explaining the difference between `MinimumDuration`
  - rename `defaultDT` into `expiredDT` to reflect better it's function in marking records as expired
- new tests:
  - `"Peer default TTL is saved when advertised"`
  - `"Peer TTL is saved when advertised with TTL"`
  - `"Peer can reregister to update its TTL before previous TTL expires"`

### DIFF
--- a/tests/discovery/testrendezvous.nim
+++ b/tests/discovery/testrendezvous.nim
@@ -390,6 +390,40 @@ suite "RendezVous":
       rendezvousRdv.registered.s[0].expiration >= timeBefore + ttl
       rendezvousRdv.registered.s[0].expiration <= timeAfter + ttl
 
+  asyncTest "Peer can reregister to update its TTL before previous TTL expires":
+    let (rendezvousNode, peerNodes, peerRdvs, rendezvousRdv) =
+      setupRendezvousNodeWithPeerNodes(1)
+    (rendezvousNode & peerNodes).startAndDeferStop()
+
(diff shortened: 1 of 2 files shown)


tests/discovery/testrendezvous.nim
```diff
@@ -342,6 +342,88 @@ suite "RendezVous":
...
+    let (rendezvousNode, peerNodes, peerRdvs, rendezvousRdv) =
+      setupRendezvousNodeWithPeerNodes(1)
+    (rendezvousNode & peerNodes).startAndDeferStop()
+
+    await connectNodes(peerNodes[0], rendezvousNode)
+
+    const namespace = "foo"
+    let timeBefore = Moment.now()
+    await peerRdvs[0].advertise(namespace)
+    let timeAfter = Moment.now()
```
gmelodie, line 355: Can you use `timeBefore.elapsed()` here?

rlve: Can you elaborate? I can't find method `elapsed` anywhere

gmelodie: My bad, must've thought of rust

PR #1656, opened 2025-09-02
fix(rendezvous): peer registration limit

rlve:
### Problem
The [spec](https://github.com/libp2p/specs/tree/master/rendezvous) recommends a limit of 1000 registrations for each peer:

> Rendezvous points are also recommended to allow:
> - a maximum of 1000 registration for each peer
>   - defend against trivial DoS attacks

There is a bug in the current implementation of this limit: the `countRegister()` method is used to count peer registrations during `register()`, however it always returns 1.

https://github.com/vacp2p/nim-libp2p/blob/d2eaf07960f363ed75183e6187ce9471d4639f0b/libp2p/protocols/rendezvous.nim#L399-L403

https://github.com/vacp2p/nim-libp2p/blob/d2eaf07960f363ed75183e6187ce9471d4639f0b/libp2p/protocols/rendezvous.nim#L439-L440

This happens because when a peer record already exists and the peer tries to register again, the original record’s expiration is moved into the past (so it’s marked for deletion on the next deletion heartbeat). Because of this and the expiration-time check in `countRegister()`, the method only counts the last record and doesn’t enforce the limit.

### Changes

Fix:
- Remove the expiration-time check in `countRegister()` and count all records with the same `peerId`.

New test:
- `"Peer registration is ignored if limit of 1000 registrations is reached"`


tests/discovery/testrendezvous.nim
```diff
@@ -424,6 +424,32 @@ suite "RendezVous":
...
+    await peerRdv.advertise(namespace)
+
+    # Simulate large number of registrations for the same Peer
+    let record = rendezvousRdv.registered.s[0]
+    for i in 0 ..< RegistrationLimitPerPeer - 2:
+      rendezvousRdv.registered.s.add(record)
+
+    # 1000th registration allowed
+    await peerRdv.advertise(namespace)
+    check rendezvousRdv.registered.s.len == RegistrationLimitPerPeer
```
vladopajic, line 447:
```suggestion
    # Simulate large number of registrations for the same Peer
    let record = rendezvousRdv.registered.s[0]
    for i in 0 ..< RegistrationLimitPerPeer:
      rendezvousRdv.registered.s.add(record)
   check rendezvousRdv.registered.s.len == RegistrationLimitPerPeer
```

vladopajic: why not like this?

```
for i in 0 ..< RegistrationLimitPerPeer:
 futs.add(peerRdv.advertise(namespace))
await allFutures(futs)
check rendezvousRdv.registered.s.len == RegistrationLimitPerPeer
```

is it much slower or ?

rlve: With the proposed way I'm getting the following error:
```
    Unhandled defect: ../libp2p/stream/bufferstream.nim(72, 11) `not s.pushing` Only one concurrent push allowed for stream 12D*RfnGU6:68b830e9c3ae3376b62584f8 [AssertionDefect]
```

And registering sequentially is too slow.

vladopajic: in this case could you add the method `fastAdvertise` locally to this test file and document why it was added (please explain that sequential code is slow and running multiple causes defect) . 

anyone reading code should understand right away intentions.

vladopajic: because next week i will forget intentions and try to tidy this up...

PR #1657, opened 2025-09-02
chore(autonat-v2): add utils

gmelodie:
Add util functions needed for autonatv2


libp2p/protocols/connectivity/autonatv2/types.nim
```diff
@@ -10,25 +10,38 @@
...
 
 import results, chronos, chronicles
-import ../../../multiaddress, ../../../peerid #, ../../../errors
-import ../../../protobuf/minprotobuf
+import
+  ../../../multiaddress,
+  ../../../peerid,
+  ../../../protobuf/minprotobuf,
+  ../../../switch,
+  ../autonat/service
```
richard-ramos: Probably makes more sense to extract from service the things that are common to autonat v1 and v2 into some file that's shared by both that file and this one

libp2p/protocols/connectivity/autonatv2/types.nim
```diff
@@ -80,6 +91,12 @@ type
     of MsgType.DialDataResponse:
       dialDataResp*: DialDataResponse
 
+# codec to string
+proc `$`*(self: AutonatV2Codec): string =
+  case self
+  of AutonatV2Codec.DialRequest: "/libp2p/autonat/2/dial-request"
+  of AutonatV2Codec.DialBack: "/libp2p/autonat/2/dial-back"
+
```
richard-ramos: No need to define this. If you print an AutonatV2Codec enum, it will print the string value you setup in the enum definition
```suggestion
```

libp2p/protocols/connectivity/autonatv2/utils.nim
```diff
@@ -0,0 +1,110 @@
...
+  ../../../peerid,
+  ../../../protobuf/minprotobuf,
+  ../autonat/service,
+  ./types
+
+proc ipAddrMatches*(observedIPAddr: MultiAddress, reqAddrs: seq[MultiAddress]): bool =
+  ## Checks if at least one of the requests' addresses
+  ## has the same IP address of observedIPAddr
+
+  let observedIPAddr = observedIPAddr[0].valueOr:
```
richard-ramos: There's no guarantee that the first element of the address is gonna be an ip (i.e. `/p2p/somePeerId` is a valid multiaddress). Look into how observedaddrmanager extracts IP for inspiration. https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/observedaddrmanager.nim#L30-L67

libp2p/protocols/connectivity/autonatv2/utils.nim
```diff
@@ -0,0 +1,110 @@
...
+  for addrs in s.peerInfo.listenAddrs:
+    addrs[0].withValue(addrIp):
+      if IP4.match(addrIp):
+        ipv4 = true
+      elif IP6.match(addrIp):
+        ipv6 = true
+
+  (ipv4, ipv6)
+
+proc isPrivateIP*(ip: string): bool {.raises: [ValueError].} =
```
richard-ramos: AutoTLS has a `isPublic`. Could we maybe put both that function and this one in some utils file?

libp2p/protocols/connectivity/autonatv2/utils.nim
```diff
@@ -0,0 +1,110 @@
...
+      dialResp: self,
+      addrs: Opt.none(MultiAddress),
+    )
+  AutonatV2Response(
+    reachability: self.asNetworkReachability(),
+    dialResp: self,
+    addrs: Opt.some(testAddrs[addrIdx]),
+  )
+
+proc areAddrsConsistent*(addrA, addrB: MultiAddress): bool =
```
vladopajic: should this be somewhere where `MultiAddress` type is defined so all can benefit?

libp2p/protocols/connectivity/autonatv2/types.nim
gmelodie: I wonder where that file would live
1. in `libp2p/protocols/connectivity/file.nim`
2. in `libp2p/protocols/connectivity/autonat/file.nim` (in this case `autonatv2/` should live inside `autonat/`, and `autonatv1` should also be inside `autonat/`)
3. Same as 2 but without creating a new `autonatv1` folder. Which is the same as just renaming `core.nim` to `types.nim` and importing it in v2.

I'll go with 3: change `core.nim` to `types.nim` in `autonat/` and import `autonat/types` in v2.

libp2p/utils/ipaddr.nim
```diff
@@ -0,0 +1,66 @@
...
+  let lookup = lookup.getPart(ipType).valueOr:
+    return false
+
+  for ma in addrs:
+    ma[0].withValue(ipAddr):
+      if ipAddr == lookup:
+        return true
+  false
+
+proc ipSupport*(s: Switch): (bool, bool) =
```
vladopajic: `proc ipSupport*(s: Switch): (bool, bool) =`
why not
`proc ipSupport*[addrs: seq[MultiAddress])): (bool, bool) `
?

tests/testipaddr.nim
```diff
@@ -0,0 +1,75 @@
...
+# those terms.
+
+import std/options
+import chronos
+import ../libp2p/[utils/ipaddr], ./helpers
+
+proc checkEncodeDecode[T](msg: T) =
+  # this would be equivalent of doing the following (e.g. for DialBack)
+  # check msg == DialBack.decode(msg.encode()).get()
+  check msg == T.decode(msg.encode()).get()
```
vladopajic: this is unused

vladopajic (approved):

richard-ramos (approved):

PR #1658, opened 2025-09-02
feat(autonat-v2): add server

gmelodie:
Add AutoNATv2 server


libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+      conn: Connection, proto: string
+  ) {.async: (raises: [CancelledError]).} =
+    try:
+      let msg = AutonatV2Msg.decode(initProtoBuffer(await conn.readLp(1024))).valueOr:
+        raise newException(AutonatV2Error, "Unable to decode AutonatV2Msg")
+      debug "Received message", msgType = $msg.msgType
+      if msg.msgType == MsgType.DialRequest:
+        await autonatV2.handleDialRequest(conn, msg.dialReq)
+      else:
+        raise newException(AutonatV2Error, "Expecting DialRequest, got " & $msg.msgType)
```
vladopajic: consider always first handling error case; this is like early return; like in go code
that is invert if with else in this block

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+    except CancelledError as exc:
+      trace "handler cancelled"
```
vladopajic: there's no need for logging in any `CancelledError` branches

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+      raise exc
+    except CatchableError as exc:
+      debug "exception in handler", description = exc.msg, conn
+    except LPStreamRemoteClosedError as exc:
+      error "connection closed by peer", description = exc.msg, peer = conn.peerId
```
vladopajic: since `CatchableError` is listed before `LPStreamRemoteClosedError`; will `CatchableError` swallow all errors - in that case with `LPStreamRemoteClosedError` will never be called?

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+    finally:
+      trace "closing connection", conn
```
vladopajic: `trace "closing connection", conn`
there is no value in this log

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+    switch: switch,
+    dialTimeout: dialTimeout,
+    dialDataSize: dialDataSize,
+    amplificationAttackTimeout: amplificationAttackTimeout,
+  )
+  proc handleStream(
+      conn: Connection, proto: string
+  ) {.async: (raises: [CancelledError]).} =
+    try:
+      let msg = AutonatV2Msg.decode(initProtoBuffer(await conn.readLp(1024))).valueOr:
```
vladopajic:
```suggestion
     let msg = AutonatV2Msg.decode(initProtoBuffer(await conn.readLp(AutonatV2MsgLpSize))).valueOr:
```

^ would be interesting to have this in our practice

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+  return Opt.none((Connection, AddrIdx))
+
+proc handleDialRequest(
+    self: AutonatV2, conn: Connection, req: DialRequest
+) {.async: (raises: [CancelledError, DialFailedError, LPStreamError]).} =
+  let observedIPAddr = (await conn.findObservedIPAddr(req)).valueOr:
+    error "Could not find observed IP address"
+    return
+
+  let (dialBackConn, addrIdx) = (await self.chooseDialAddr(conn.peerId, req.addrs)).valueOr:
```
vladopajic, line 205: `dialBackConn` do we need to close it?

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+.} =
+  for i, ma in addrs:
+    if self.canDial(ma):
+      try:
+        debug "Trying to dial", chosenAddrs = ma, addrIdx = i
+        return Opt.some(
+          (await self.switch.dial(pid, @[ma], $AutonatV2Codec.DialBack), i.AddrIdx)
+        )
+      except DialFailedError:
+        return Opt.none((Connection, AddrIdx))
```
vladopajic, line 194: `return Opt.none((Connection, AddrIdx))`
should this line be 
`continue`

?

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+      return false
+  for t in self.switch.transports:
+    if t.handles(addrs):
+      return true
+  false
+
+proc chooseDialAddr(
+    self: AutonatV2, pid: PeerId, addrs: seq[MultiAddress]
+): Future[Opt[(Connection, AddrIdx)]] {.
+    async: (raises: [CancelledError, LPStreamError])
```
vladopajic: hmmm why does it propagate LPStreamError?

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+    try:
+      if isPrivateIP($addrIp):
+        return false
+    except ValueError:
+      warn "Unable to parse IP address, skipping", addrs = $addrIp
+      return false
+  for t in self.switch.transports:
+    if t.handles(addrs):
+      return true
+  false
```
vladopajic: `return false`

libp2p/protocols/connectivity/autonatv2/server.nim
vladopajic: what do you guys think?

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+  return DialStatus.Ok
+
+proc handleDialDataResponses(
+    self: AutonatV2, conn: Connection
+) {.async: (raises: [CancelledError, AutonatV2Error, LPStreamError]).} =
+  var dataReceived: uint64 = 0
+
+  while dataReceived < self.dialDataSize:
+    # readLp needs to receive more than 4096 bytes (since it's a DialDataResponse) + overhead
+    let msg = AutonatV2Msg.decode(initProtoBuffer(await conn.readLp(5000))).valueOr:
```
vladopajic: adding up to suggestion from below, this kind of comments are much more suitable if it is added to `const` and not to line where it is used.

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+import results
+import chronos, chronicles
+import
+  ../../protocol,
+  ../../../switch,
+  ../../../multiaddress,
+  ../../../transports/transport,
+  ../../../multicodec,
+  ../../../peerid,
+  ../../../protobuf/minprotobuf,
```
vladopajic: `import  ../../../[protocol, multiaddress, transports/transport, multicodec, peerid ...]`

vladopajic (commented):
in general LGTM, some nitpicks; but would like to see updated code before approve

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
+# Nim-LibP2P
```
vladopajic, line 1: would be nice to have some tests for AutonatV2 server. it was not possible in this stage or?

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+    # receive DialBackResponse
+    let dialBackResp = DialBackResponse.decode(initProtoBuffer(await conn.readLp(1024))).valueOr:
+      error "DialBack failed, could not decode DialBackResponse"
+      return DialStatus.EDialBackError
+  except LPStreamRemoteClosedError as exc:
+    # failed because of nonce error (remote reset the stream): EDialBackError
+    error "DialBack failed, remote closed the connection", description = exc.msg
+    return DialStatus.EDialBackError
+
+  # TODO: failed because of client or server resources: EDialError
```
richard-ramos, line 103: Is this going to be done in a diff PR?

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+    self: AutonatV2, pid: PeerId, addrs: seq[MultiAddress]
+): Future[Opt[(Connection, AddrIdx)]] {.
+    async: (raises: [CancelledError, LPStreamError])
+.} =
+  for i, ma in addrs:
+    if self.canDial(ma):
+      try:
+        debug "Trying to dial", chosenAddrs = ma, addrIdx = i
+        return Opt.some(
+          (await self.switch.dial(pid, @[ma], $AutonatV2Codec.DialBack), i.AddrIdx)
```
richard-ramos: https://github.com/vacp2p/nim-libp2p/pull/1649#discussion_r2310408421

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,241 @@
...
+  return Opt.some(hostIp)
+
+proc dialBack(
+    conn: Connection, nonce: Nonce
+): Future[DialStatus] {.
+    async: (raises: [CancelledError, DialFailedError, LPStreamError])
+.} =
+  try:
+    # send dial back
+    await conn.writeLp(DialBack(nonce: nonce).encode().buffer)
```
richard-ramos, line 90: https://github.com/vacp2p/nim-libp2p/pull/1649#discussion_r2310319354

libp2p/protocols/connectivity/autonatv2/server.nim
gmelodie: i put it as a todo so we didn't forget, but I couldn't find a place in the code where resources would "run out" so I didn't know exactly where this error should happen

libp2p/protocols/connectivity/autonatv2/server.nim
gmelodie: We're [returning the connection](https://github.com/vacp2p/nim-libp2p/pull/1658/files#diff-1d4f8d222bb91a92236006e449307e073f1010505e8bd28c3a03ecf501a49063R181) so it can be used to [send the DialBack](https://github.com/vacp2p/nim-libp2p/pull/1658/files#diff-1d4f8d222bb91a92236006e449307e073f1010505e8bd28c3a03ecf501a49063R198). Not sure what you want done here.

libp2p/protocols/connectivity/autonatv2/server.nim
gmelodie: It most certainly should not! Rather, it MUST NOT.
[spec](https://github.com/libp2p/specs/blob/master/autonat/autonat-v2.md#overview)
> As the server dials exactly one address from the list, autonat v2 allows nodes to determine reachability for individual addresses.

libp2p/protocols/connectivity/autonatv2/server.nim
richard-ramos: The connection that you are returning here is likely the same connection on which you received the original DialRequest

https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/dialer.nim#L328

in here dial will call getStream to obtain a stream from the connection manager (even the trace log on line 325 indicates that you're reusing an existing connection). When dialing back you should not reuse a connection, you want to use a completely new one that uses the specific ip/port/transport specified in the multiaddress.

libp2p/protocols/connectivity/autonatv2/server.nim
gmelodie: Out of ideas in terms of checking, but added a somewhat dummy test to at least make sure the code compiles and gets mounted in a switch

vladopajic: coolio!

can you send a protocol message? for example you could send invalid protocol message (cannot be serializes, missing value, etc...) and a valid protocol message?

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,248 @@
...
+      let msg = AutonatV2Msg.decode(
+        initProtoBuffer(await conn.readLp(AutonatV2MsgLpSize))
+      ).valueOr:
+        raise newException(AutonatV2Error, "Unable to decode AutonatV2Msg")
+      debug "Received message", msgType = $msg.msgType
+      if msg.msgType != MsgType.DialRequest:
+        raise newException(AutonatV2Error, "Expecting DialRequest, got " & $msg.msgType)
+      await autonatV2.handleDialRequest(conn, msg.dialReq)
+    except LPStreamRemoteClosedError as exc:
+      error "connection closed by peer", description = exc.msg, peer = conn.peerId
```
vladopajic: `CancelledError` not propagated...

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,248 @@
...
+    try:
+      let msg = AutonatV2Msg.decode(
+        initProtoBuffer(await conn.readLp(AutonatV2MsgLpSize))
+      ).valueOr:
+        raise newException(AutonatV2Error, "Unable to decode AutonatV2Msg")
+      debug "Received message", msgType = $msg.msgType
+      if msg.msgType != MsgType.DialRequest:
+        raise newException(AutonatV2Error, "Expecting DialRequest, got " & $msg.msgType)
+      await autonatV2.handleDialRequest(conn, msg.dialReq)
+    except LPStreamRemoteClosedError as exc:
```
vladopajic: would be better if this `try` block is not catching error that are thrown from this code here

libp2p/protocols/connectivity/autonatv2/server.nim
gmelodie: Wouldn't it be propagated since we don't catch it?

libp2p/protocols/connectivity/autonatv2/server.nim
gmelodie: Not 100% sure how to rewrite this better. Any tips?

vladopajic: move msg deserialization and msg type check above try?

vladopajic: initializes rawMsg first and handling only connection errors here (readLp call)

libp2p/protocols/connectivity/autonatv2/server.nim
vladopajic: no it won't get propagated because it will get caught by CatchableError

gmelodie: my was looking at a different place in the code hehe

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,273 @@
...
+    try:
+      await autonatV2.handleDialRequest(conn, msg.dialReq)
+    except CancelledError as exc:
+      raise exc
+    except LPStreamRemoteClosedError as exc:
+      error "connection closed by peer", description = exc.msg, peer = conn.peerId
+    except CatchableError as exc:
+      error "exception in handler", description = exc.msg, conn
+    finally:
+      await conn.close()
```
vladopajic: should we close conn if error happens (Unable to decode AutonatV2Msg, Error receiving AutonatV2Msg, etc... )?

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -0,0 +1,273 @@
...
+    dialDataSize: uint64 = DefaultDialDataSize,
+    amplificationAttackTimeout: Duration = DefaultAmplificationAttackDialTimeout,
+): T =
+  let autonatV2 = T(
+    switch: switch,
+    dialTimeout: dialTimeout,
+    dialDataSize: dialDataSize,
+    amplificationAttackTimeout: amplificationAttackTimeout,
+  )
+  proc handleStream(
```
vladopajic, line 243: not sure if any of these logs in `handleStream` should be `error`...

vladopajic (approved):

libp2p/protocols/connectivity/autonatv2/server.nim
gmelodie: I think so. Is there anything else we could do?

vladopajic: defer: conn.close() as first line of this proc?

PR #1659, opened 2025-09-02
feat(autonat-v2): add client and tests

gmelodie:
Add AutonatV2 client and tests

Needs #1657 and #1658


libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,208 @@
...
+): T =
+  let client = T(switch: switch, rng: rng, dialBackTimeout: dialBackTimeout)
+
+  # handler for DialBack messages
+  proc handleStream(
+      conn: Connection, proto: string
+  ) {.async: (raises: [CancelledError]).} =
+    try:
+      let dialBack = DialBack.decode(initProtoBuffer(await conn.readLp(1024))).valueOr:
+        raise newException(AutonatV2Error, "Unable to decode DialBack")
```
vladopajic: no need to raise here just log error; as you are doing that in `CatchableError` case

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,208 @@
...
+    dialBackTimeout: Duration = DefaultDialBackTimeout,
+): T =
+  let client = T(switch: switch, rng: rng, dialBackTimeout: dialBackTimeout)
+
+  # handler for DialBack messages
+  proc handleStream(
+      conn: Connection, proto: string
+  ) {.async: (raises: [CancelledError]).} =
+    try:
+      let dialBack = DialBack.decode(initProtoBuffer(await conn.readLp(1024))).valueOr:
```
vladopajic: `1024` -> `DialBackLpSize`

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,208 @@
...
+        raise newException(AutonatV2Error, "Unable to decode DialBack")
+      if not await client.handleDialBack(conn, dialBack).withTimeout(
+        client.dialBackTimeout
+      ):
+        error "Sending DialBackResponse timeout"
+        return
+    except CancelledError as exc:
+      error "Handler cancelled", description = exc.msg
```
vladopajic: should not be error as `CancelledError` can appear at any place if node is stopped. also we generally should never log cancelled error.

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,208 @@
...
+      raise exc
+    except LPStreamRemoteClosedError as exc:
+      error "Connection closed by peer", description = exc.msg, peer = conn.peerId
```
vladopajic: not sure if it should be error, it valid case if other pear just drops connection

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,208 @@
...
+    dialDataResp: DialDataResponse(data: newSeq[byte](MaxDialDataResponsePayload)),
+  )
+  let messagesToSend =
+    (req.numBytes + MaxDialDataResponsePayload - 1) div MaxDialDataResponsePayload
+  for i in 0 ..< messagesToSend:
+    try:
+      await conn.writeLp(msg.encode().buffer)
+      debug "Sending DialDataResponse", i = i, messagesToSend = messagesToSend
+    except CatchableError as exc:
+      raise newException(AutonatV2Error, "Failed to send DialDataResponse", exc)
```
vladopajic: CancelledError not raised here? Could we list specific error instead of `CatchableError`

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,208 @@
...
+
+  # get DialResponse
+  msg = AutonatV2Msg.decode(initProtoBuffer(await conn.readLp(1024))).valueOr:
```
vladopajic: `1024` as cosnt

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,208 @@
...
+
+proc checkAddrIdx(
+    self: AutonatV2Client, addrIdx: AddrIdx, testAddrs: seq[MultiAddress], nonce: Nonce
+): bool {.raises: [AutonatV2Error].} =
+  debug "checking addrs", addrIdx = addrIdx, testAddrs = testAddrs, nonce = nonce
+  let dialBackAddrs =
+    try:
+      self.expectedNonces[nonce].get()
+    except KeyError:
+      error "Not expecting this nonce", nonce = nonce
```
vladopajic: not sure about error log here

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,208 @@
...
+      return false
+
+  if addrIdx >= testAddrs.len.AddrIdx:
+    error "addrIdx outside of testAddrs range", addrIdx = addrIdx
```
vladopajic: not sure about error log here

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,208 @@
...
+    return false
+
+  let dialRespAddrs = testAddrs[addrIdx]
+  if not areAddrsConsistent(dialRespAddrs, dialBackAddrs):
+    error "Invalid addrIdx: got DialBack in another address",
```
vladopajic: not sure about error log here

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,208 @@
...
+logScope:
+  topics = "libp2p autonat v2 client"
+
+const
+  MaxAcceptedDialDataRequest* = 100 * 1024 # 100 KB
+  MaxDialDataResponsePayload* = 1024
+  DefaultDialBackTimeout* = 5.seconds
+
+type AutonatV2Client* = ref object of LPProtocol
+  switch*: Switch
```
vladopajic: since `AutonatV2Client` only uses `switch.dialer` why not use `dialer: Dialer` as property? 
if you would mock this you would probably don't want to mock whole `Switch` here. or in any case minimum interface the better.

tests/testautonatv2.nim
```diff
@@ -156,13 +153,150 @@ suite "AutonatV2":
...
+
+    check autonatV2Resp ==
+      AutonatV2Response(
+        reachability: NotReachable,
+        dialResp: DialResponse(
+          status: ResponseStatus.Ok,
+          dialStatus: Opt.some(DialStatus.EDialError),
+          addrIdx: Opt.some(0.AddrIdx),
+        ),
+        addrs: Opt.some(MultiAddress.init("/ip4/1.1.1.1/tcp/4040").get()),
```
vladopajic:
```suggestion
        addrs: Opt.some(reqAddrs[0])
```

tests/testautonatv2.nim
```diff
@@ -156,13 +153,150 @@ suite "AutonatV2":
...
+    let autonatV2Resp =
+      await client.sendDialRequest(dst.peerInfo.peerId, dst.peerInfo.addrs, reqAddrs)
+
+    check autonatV2Resp ==
+      AutonatV2Response(
+        reachability: NotReachable,
+        dialResp: DialResponse(
+          status: ResponseStatus.Ok,
+          dialStatus: Opt.some(DialStatus.EDialError),
+          addrIdx: Opt.some(0.AddrIdx),
```
vladopajic, line 284: if `dialStatus` is `DialStatus.EDialError` why should it be that `addrIdx:` is `Opt.some` instead of `Opt,none`?

tests/testautonatv2.nim
```diff
@@ -156,13 +153,150 @@ suite "AutonatV2":
...
+        .withAutonatV2(config = AutonatV2Config.new(allowPrivateAddresses = true))
+        .build()
+
+    let client = AutonatV2Client.new(src, newRng())
+    src.mount(client)
+    await src.start()
+    await dst.start()
+
+    defer:
+      await allFutures(src.stop(), dst.stop())
```
vladopajic: this code seems like boilerplate here

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,208 @@
...
+    return false
+
+  let dialRespAddrs = testAddrs[addrIdx]
+  if not areAddrsConsistent(dialRespAddrs, dialBackAddrs):
+    error "Invalid addrIdx: got DialBack in another address",
+      dialBackAddrs = dialBackAddrs, dialRespAddrs = dialRespAddrs
+    return false
+  true
+
+proc sendDialRequest*(
```
vladopajic, line 146: `sendDialRequest` is never tested when listed errors are expected

tests/testautonatv2.nim
gmelodie: Because we want to return which address we tried to dial and failed

tests/testautonatv2.nim
gmelodie: Not sure what you mean. It's needed for the test

libp2p/builders.nim
```diff
@@ -290,6 +297,10 @@ proc withAutonatV2*(
   b.autonatV2Config = config
   b
 
+proc withAutonatV2Client*(b: SwitchBuilder): SwitchBuilder =
```
richard-ramos, line 300: This is interesting, because now we have:
- `withAutonatV2Client` which is clear what it does
- `withAutonatV2` seems like it lacks something :thinking: 

So i wonder if we should rename withAutonatV2 to withAutonatV2Server,  or maybe have this:
```
withAutonatV2  -- client fucntionality + service
withAutonatV2Server -- server
```
wdyt

tests/testautonatv2.nim
vladopajic: is that by specific. if dialing failed its always going to last index, right? otherwise it's a success.

gmelodie: We only try to dial exactly one address. If that address fails, we return a failure.

tests/testautonatv2.nim
vladopajic: its repeated, it could be template...

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,214 @@
...
+    dialDataResp: DialDataResponse(data: newSeq[byte](MaxDialDataResponsePayload)),
+  )
+  let messagesToSend =
+    (req.numBytes + MaxDialDataResponsePayload - 1) div MaxDialDataResponsePayload
+  for i in 0 ..< messagesToSend:
+    # try:
+    #   await conn.writeLp(msg.encode().buffer)
+    #   debug "Sending DialDataResponse", i = i, messagesToSend = messagesToSend
+    # except CatchableError as exc:
+    #   raise newException(AutonatV2Error, "Failed to send DialDataResponse", exc)
```
richard-ramos: If not needed, lets get rid of it
```suggestion
```

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,214 @@
...
+    await conn.writeLp(msg.encode().buffer)
+    debug "Sending DialDataResponse", i = i, messagesToSend = messagesToSend
+
+  # get DialResponse
+  msg = AutonatV2Msg.decode(initProtoBuffer(await conn.readLp(AutonatV2MsgLpSize))).valueOr:
+    raise newException(AutonatV2Error, "Unable to decode AutonatV2Msg")
+
+  debug "Receive message", msgType = $msg.msgType
```
richard-ramos: I think the `$` is not needed. Might be wrong
```suggestion
  debug "Receive message", msgType = msg.msgType
```

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,214 @@
...
+    self: AutonatV2Client, addrIdx: AddrIdx, testAddrs: seq[MultiAddress], nonce: Nonce
+): bool {.raises: [AutonatV2Error].} =
+  debug "checking addrs", addrIdx = addrIdx, testAddrs = testAddrs, nonce = nonce
+  let dialBackAddrs =
+    try:
+      self.expectedNonces[nonce].get()
+    except KeyError:
+      debug "Not expecting this nonce",
+        nonce = nonce, expectedNonces = self.expectedNonces
+      return false
```
richard-ramos: This should be slightly more efficient as it will not raise a exception

```suggestion
  if not self.expectedNonces.hasKey[nonce]:
    debug "Not expecting this nonce", nonce = nonce, expectedNonces = self.expectedNonces
    return false
    
  let dialBackAddrs =
    try:
      self.expectedNonces[nonce].get()
    except KeyError:
       doAssert false, "checked with hasKey"
```

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,214 @@
...
+
+  if addrIdx >= testAddrs.len.AddrIdx:
```
richard-ramos: This might be a bit more readable compared to casting the length to an AddrIdx
```suggestion
  if addrIdx.int >= testAddrs.len:
```

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,209 @@
...
+    self.expectedNonces[dialBack.nonce] = Opt.some(localAddr)
+  else:
+    error "Unable to get localAddr from connection"
+    return
+
+  try:
+    trace "Sending DialBackResponse"
+    await conn.writeLp(DialBackResponse(status: DialBackStatus.Ok).encode().buffer)
+  except CatchableError as exc:
+    raise newException(AutonatV2Error, "Sending DialBackResponse failed", exc)
```
vladopajic: CancelledError should not be wrapped to AutonatV2Error?

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,209 @@
...
+  ) {.async: (raises: [CancelledError]).} =
+    try:
+      let dialBack = DialBack.decode(initProtoBuffer(await conn.readLp(DialBackLpSize))).valueOr:
+        trace "Unable to decode DialBack"
+        return
+      if not await client.handleDialBack(conn, dialBack).withTimeout(
+        client.dialBackTimeout
+      ):
+        trace "Sending DialBackResponse timeout"
+        return
```
vladopajic: this return is usless

libp2p/builders.nim
gmelodie: I thought so too but i was following name conventions on autonatv1. Either:
1. autonatV2Server + autonatV2Client
2. autonatV2Client + autonatV2 <--------------------- (current)

autonatV2 + autonatV2Server is counterintuitive coming from v1's implementation

tests/testautonatv2.nim
```diff
@@ -20,18 +20,38 @@ import
     protocols/connectivity/autonatv2/types,
     protocols/connectivity/autonatv2/server,
     protocols/connectivity/autonatv2/utils,
+    protocols/connectivity/autonatv2/client,
+    protocols/connectivity/autonatv2/mock,
   ],
   ./helpers
 
+proc newSwitches(
```
vladopajic: nit: `newSwitches` -> `setupAutonat` ?

vladopajic: if this was `template` you cold have had:

```
   defer:
      await allFutures(src.stop(), dst.stop())
```

inside

tests/testautonatv2.nim
```diff
@@ -156,13 +176,156 @@ suite "AutonatV2":
...
+        addrs: Opt.some(reqAddrs[0]),
+      )
+
+  asyncTest "Server responding with invalid messages":
+    let
+      src = newStandardSwitchBuilder().build()
+      dst = newStandardSwitchBuilder().build()
+      client = AutonatV2Client.new(src, newRng())
+      autonatV2Mock = AutonatV2Mock.new(dst)
+      reqAddrs = @[MultiAddress.init("/ip4/127.0.0.1/tcp/4040").get()]
```
vladopajic, line 295: ```MultiAddress.init("/ip4/127.0.0.1/tcp/4040").get()``` is repeated often

libp2p/protocols/connectivity/autonatv2/mock.nim
```diff
@@ -0,0 +1,68 @@
...
+      multicodec,
+      peerid,
+      protobuf/minprotobuf,
+    ],
+  ../../protocol,
+  ./types,
+  ./server
+
+type AutonatV2Mock* = ref object of LPProtocol
+  switch*: Switch
```
vladopajic: why is Switch needed here?

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -0,0 +1,202 @@
...
+  proc handleStream(
+      conn: Connection, proto: string
+  ) {.async: (raises: [CancelledError]).} =
+    try:
+      let dialBack = DialBack.decode(initProtoBuffer(await conn.readLp(DialBackLpSize))).valueOr:
+        trace "Unable to decode DialBack"
+        return
+      if not await client.handleDialBack(conn, dialBack).withTimeout(
+        client.dialBackTimeout
+      ):
```
vladopajic, line 77: nit: if `wait` was used then timeout is handled as `except` statement might be more cleaner

vladopajic (approved):
lgtm; nitpicks only

libp2p/builders.nim
richard-ramos: I'm happy to go in this case with clarity > convention!
:thinking: 


Looking for inspiration, in other impls we have the following:
- go-libp2p: `libp2p.EnableNATService()`  (this is the server side), and `libp2p.EnableAutoNATv2()`
- rust-libp2p: They do register behaviors so no direct comparison possible: https://github.com/umgefahren/rust-libp2p/blob/ec56cbc645e2caaa083162a00c8843c0e859ecdd/examples/autonatv2/src/bin/autonatv2_server.rs#L55
- js-libp2p: In this case the service starts both client and server functionality 
```js
const node = await createLibp2p({
  services: {
    autoNAT: autoNATv2()
  }
})
```

gmelodie: I'll do this on a separate PR then

PR #1665, opened 2025-09-05
chore: `splitRPCMsg` improvements

vladopajic:
- warn when message is not included
- remove else block for remaining send
- cosmetics changes

### DIFF
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -478,19 +478,17 @@ iterator splitRPCMsg(
   ## exceeds the `maxSize` when trying to fit into an empty `RPCMsg`, the latter is skipped as too large to send.
   ## Every constructed `RPCMsg` is then encoded, optionally anonymized, and yielded as a sequence of bytes.
 
-  var currentRPCMsg = rpcMsg
-  currentRPCMsg.messages = newSeq[Message]()
-
-  var currentSize = byteSize(currentRPCMsg)
+  var currentRPCMsg = RPCMsg()
+  var currentSize = 0
+  const messageOverhead = 1.1 # Guessing 10% protobuf overhead
 
   for msg in rpcMsg.messages:
     let msgSize = byteSize(msg)
 
     # Check if adding the next message will exceed maxSize
-    if float(currentSize + msgSize) * 1.1 > float(maxSize):
-      # Guessing 10% protobuf overhead
-      if currentRPCMsg.messages.len == 0:
-        trace "message too big to sent", peer, rpcMsg = shortLog(currentRPCMsg)
+    if float(currentSize + msgSize) * messageOverhead > float(maxSize):
+      if msgSize > maxSize:
+        warn "message too big to sent", peer, rpcMsg = shortLog(msg)
         continue # Skip this message
 
       trace "sending msg to peer", peer, rpcMsg = shortLog(currentRPCMsg)
@@ -502,11 +500,9 @@ iterator splitRPCMsg(
     currentSize += msgSize
 
   # Check if there is a non-empty currentRPCMsg left to be added
-  if currentSize > 0 and currentRPCMsg.messages.len > 0:
+  if currentRPCMsg.messages.len > 0:
     trace "sending msg to peer", peer, rpcMsg = shortLog(currentRPCMsg)
     yield encodeRpcMsg(currentRPCMsg, anonymize)
-  else:
-    trace "message too big to sent", peer, rpcMsg = shortLog(currentRPCMsg)
 
 proc send*(
     p: PubSubPeer,
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -480,14 +480,15 @@ iterator splitRPCMsg(
 
   var currentRPCMsg = RPCMsg()
   var currentSize = 0
-  const messageOverhead = 1.1 # Guessing 10% protobuf overhead
+  let maxSizeOverheadExcluded = (maxSize * 90) div 100
+    # Guessing 10% protobuf overhead; that is messages should not exceed 90% of maxSize
 
   for msg in rpcMsg.messages:
     let msgSize = byteSize(msg)
 
     # Check if adding the next message will exceed maxSize
-    if float(currentSize + msgSize) * messageOverhead > float(maxSize):
-      if msgSize > maxSize:
+    if currentSize + msgSize > maxSizeOverheadExcluded:
+      if msgSize > maxSizeOverheadExcluded:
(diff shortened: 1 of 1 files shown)


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -478,19 +478,16 @@ iterator splitRPCMsg(
   ## exceeds the `maxSize` when trying to fit into an empty `RPCMsg`, the latter is skipped as too large to send.
   ## Every constructed `RPCMsg` is then encoded, optionally anonymized, and yielded as a sequence of bytes.
 
-  var currentRPCMsg = rpcMsg
-  currentRPCMsg.messages = newSeq[Message]()
-
-  var currentSize = byteSize(currentRPCMsg)
+  var currentRPCMsg = RPCMsg()
```
richard-ramos, line 481: Don't you lose the `subscriptions`, `control`, `ping` and `pong` messages that the original message might have? because according to specs, a message might have control messages piggybacked in it https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md#control-message-piggybacking

vladopajic: on line [495](https://github.com/vacp2p/nim-libp2p/pull/1665/files/31e233ed96e1a35a83c3e4aa661c04c6554f0b3f#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3R495) after batch is sent we are already loosing that (this is already on master).

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -542,8 +537,11 @@ proc send*(
       sendMetrics(msg)
       encodeRpcMsg(msg, anonymize)
 
-  if encoded.len > p.maxMessageSize and msg.messages.len > 1:
-    for encodedSplitMsg in splitRPCMsg(p, msg, p.maxMessageSize, anonymize):
+  # Messages should not exceed 90% of maxMessageSize. Guessing 10% protobuf overhead.
+  let maxEncodedMsgSize = (p.maxMessageSize * 90) div 100
```
gmelodie, line 541: nit: I think `maxEncodedMsgSize` should be part of `PubSubPeer`, initialized when it's created using a `const` that says how much of the messageSize should be the `maxEncodedMsgSize` (here, that const would be `0.9`)

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -542,8 +537,11 @@ proc send*(
       sendMetrics(msg)
       encodeRpcMsg(msg, anonymize)
 
-  if encoded.len > p.maxMessageSize and msg.messages.len > 1:
-    for encodedSplitMsg in splitRPCMsg(p, msg, p.maxMessageSize, anonymize):
+  # Messages should not exceed 90% of maxMessageSize. Guessing 10% protobuf overhead.
+  let maxEncodedMsgSize = (p.maxMessageSize * 90) div 100 
+
+  if encoded.len > maxEncodedMsgSize and msg.messages.len > 1:
+    for encodedSplitMsg in splitRPCMsg(p, msg, maxEncodedMsgSize, anonymize):
```
gmelodie, line 544: nit: if we apply my suggestion from above, this could be written as
```suggestion
    for encodedSplitMsg in p.splitRPCMsg(msg, anonymize):
```

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -478,19 +478,16 @@ iterator splitRPCMsg(
...
 
     # Check if adding the next message will exceed maxSize
-    if float(currentSize + msgSize) * 1.1 > float(maxSize):
-      # Guessing 10% protobuf overhead
-      if currentRPCMsg.messages.len == 0:
-        trace "message too big to sent", peer, rpcMsg = shortLog(currentRPCMsg)
+    if currentSize + msgSize > maxSize:
+      if msgSize > maxSize:
+        warn "message too big to sent", peer, rpcMsg = shortLog(msg)
         continue # Skip this message
```
gmelodie, line 491: Should this `continue` not be triggered in both cases (i.e. when either `currentSize + msgSize > maxSize` or `msgSize > maxSize` are true)? Right now it's only triggered when `msgSize > maxSize` is true.

gmelodie (changes requested):

libp2p/protocols/pubsub/pubsubpeer.nim
vladopajic: the message is not included if it's size exceeds maxSize. conditions are correct.

arnetheduck:
The best thing I can think of for ``splitRPCMsg` is to remove it entirely - the idea behind it is odd, to say the least. 

The user of libp2p controls the "maximum" allowed message size - this can deterministically be translated into a "maximum payload size" depending on which options are used (signature etc) - or the other way around, that an application that wants to send payloads of a certain size can compute a maximum message size. Therefore, it's kind of odd that libp2p would go about internally splitting rpc messages - the only time those can happen is when the application violates its own limits which it should not be doing.

libp2p/protocols/pubsub/pubsubpeer.nim
vladopajic: it's probably not ideal to have two properties that are directly related to each other - as this can be unnecessary if you can easily derive other properties from main.

const for percentage, sure it can be used but since this is only used in one place i don't see a big urge for this.

libp2p/protocols/pubsub/pubsubpeer.nim
vladopajic: i'd rather not use two properties here.

libp2p/protocols/pubsub/pubsubpeer.nim
gmelodie: But if `currentSize + msgSize > maxSize`, should msg be included then? If so, why?

vladopajic: if there is 0 messages currently in batch, and message size exceeds limits then this message is not included.
this means that this message size exceeds max size.

gmelodie (approved):

vladopajic:
yeah, i was just making sure to see if we should still keep maxMessageSize. i'll make this change in new pr, this one got automerged.

arnetheduck:
Yeah, this is a key feature in gossipsub in general, ie that there exists a maximum message size - further, the message size must be coordinated between nodes, ie they must all use the same max message size or the network will split (since nodes will consider it a protocol violation if too-large messages are being propagated / sent.

The tricky part about rpcmsg is to make sure that when packing a message and piggypacking other information on top, ie when subscribing to topics, the topic subscription list (and similar control messages) don't grow beyond the max message size - this implies that when subscribing and performing other control operations, you have to be mindful of this size as well.

PR #1669, opened 2025-09-08
chore(autonat-v2): add server config

gmelodie:
- Add config struct
- Add `allowPrivateAddresses` to config (for testing purposes mainly)

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -19,7 +19,7 @@ jobs:
           fetch-depth: 2  # In PR, has extra merge commit: ^1 = PR, ^2 = base
 
       - name: Check `nph` formatting
-        uses: arnetheduck/nph-action@v1
+        uses: arnetheduck/nph-action@ef5e9fae6dbaf88ec4308cbede780a0ba45f845d
         with:
           version: 0.6.1
           options: "examples libp2p tests interop tools *.nim*"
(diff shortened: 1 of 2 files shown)


libp2p/builders.nim
```diff
@@ -282,8 +283,11 @@ proc withAutonat*(b: SwitchBuilder): SwitchBuilder =
   b.autonat = true
   b
 
-proc withAutonatV2*(b: SwitchBuilder): SwitchBuilder =
+proc withAutonatV2*(
+    b: SwitchBuilder, config: AutonatV2Config = AutonatV2Config.new()
```
richard-ramos, line 287: Consider making the config optional, so autonatV2 can work with sane defaults and dev does not have to worry about passing a config

gmelodie: So we either have it inside an `Opt` or not but with defaults. I prefer the latter to avoid the opt checks every time.

libp2p/builders.nim
```diff
@@ -282,8 +283,11 @@ proc withAutonat*(b: SwitchBuilder): SwitchBuilder =
...
+): SwitchBuilder =
   b.autonatV2 = true
+  b.autonatV2Config = config
```
vladopajic, line 290: if `autonatV2Config` is set then it can mean that `autonatV2` is used (autonatV2 is derived property).

`if autonatV2Config is set then autonatV2 = true`

PR #1670, opened 2025-09-08
chore: remove `splitRPCMsg`

vladopajic:
as per comment https://github.com/vacp2p/nim-libp2p/pull/1665#issuecomment-3266425444; `splitRPCMsg` is removed.
in addition to comment provided here `splitRPCMsg` introduces issues where piggybacked control messages are removed (and other properties send in RPCMsg). 

overall better let application control messages it sends.

### DIFF
--- a/tests/pubsub/integration/testgossipsubmessagehandling.nim
+++ b/tests/pubsub/integration/testgossipsubmessagehandling.nim
@@ -76,33 +76,6 @@ suite "GossipSub Integration - Message Handling":
   teardown:
     checkTrackers()
 
-  asyncTest "Split IWANT replies when individual messages are below maxSize but combined exceed maxSize":
-    # This test checks if two messages, each below the maxSize, are correctly split when their combined size exceeds maxSize.
-    # Expected: Both messages should be received.
-    let (gossip0, gossip1, receivedMessages) = await setupTest()
-
-    let messageSize = gossip1.maxMessageSize div 2 + 1
-    let (iwantMessageIds, sentMessages) =
-      createMessages(gossip0, gossip1, messageSize, messageSize)
-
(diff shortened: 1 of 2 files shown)


gmelodie (approved):

richard-ramos (approved):
Deleted code is the best kind of code

tests/pubsub/integration/testgossipsubmessagehandling.nim
```diff
@@ -157,41 +130,6 @@ suite "GossipSub Integration - Message Handling":
 
     await teardownTest(gossip0, gossip1)
 
-  asyncTest "Split IWANT replies when one message is below maxSize and the other exceeds maxSize":
```
arnetheduck, line 160: this test is still relevant though: when _creating_ the IWANT response, you must ensure that you don't exceed the max message size (ie when you're responding with 2 message payloads for example) - that means that the response must be split up, and you have to do the splitting _before_ calling send which implies you have to keep adding up the encoded message size as you're creating the response.

this applies to IWANT responses, but also to subscriptions and other control messages - imagine that the max message size is 1024 bytes - then, you would only fit a few topic subscriptions in a single message..

vladopajic: this won't be trivial change then... iam gonna close this pr

arnetheduck: there are few of the trivially fixed workarounds and hacks left in the codebase ;) 

that said, there exists a mostly trivial and correct way to do it, which is to send a separate rpcmsg for every iwant response payload - since each payload resulted from a message that was smaller than the max message size, this would be safe

PR #1679, opened 2025-09-10
fix(pubsub): use custom conn when message is sent as lower-priority

vladopajic:


gmelodie (approved):
Maybe we should add a test too? Not sure if that already exists

richard-ramos (approved):

PR #1680, opened 2025-09-10
chore: rename withAutonatV2 to withAutonatV2Server

gmelodie:
From [this discussion](https://github.com/vacp2p/nim-libp2p/pull/1659#discussion_r2331399433) from #1659

### DIFF
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -92,11 +92,17 @@ type
     # have to pass peer as it's unknown during init
   OnEvent* = proc(peer: PubSubPeer, event: PubSubPeerEvent) {.gcsafe, raises: [].}
 
+  QueuedMessage = object
+    # Messages sent as lower-priority are queued and sent in other location
+    # in which we need to keep arguments like `useCustomConn`.
+    data: seq[byte]
+    useCustomConn: bool
+
   RpcMessageQueue* = ref object
     # Tracks async tasks for sending high-priority peer-published messages.
     sendPriorityQueue: Deque[Future[void]]
     # Queue for lower-priority messages, like "IWANT" replies and relay messages.
-    nonPriorityQueue: AsyncQueue[seq[byte]]
+    nonPriorityQueue: AsyncQueue[QueuedMessage]
     # Task for processing non-priority message queue.
     sendNonPriorityTask: Future[void]
 
@@ -465,7 +471,9 @@ proc sendEncoded*(
       else:
         Future[void].completed()
     else:
-      let f = p.rpcmessagequeue.nonPriorityQueue.addLast(msg)
+      let f = p.rpcmessagequeue.nonPriorityQueue.addLast(
+        QueuedMessage(data: msg, useCustomConn: useCustomConn)
+      )
       when defined(pubsubpeer_queue_metrics):
         libp2p_gossipsub_non_priority_queue_size.inc(labelValues = [$p.peerId])
       f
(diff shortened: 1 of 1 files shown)


libp2p/builders.nim
```diff
@@ -403,13 +402,15 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
...
-    let autonatV2 = AutonatV2.new(switch, config = b.autonatV2Config)
-    switch.mount(autonatV2)
-  elif b.autonat:
-    let autonat = Autonat.new(switch)
-    switch.mount(autonat)
+    # TODO: start service
+
+  b.autonatV2ServerConfig.withValue(config):
+    let autonatV2Server = AutonatV2.new(switch, config = config)
+    switch.mount(autonatV2Server)
```
vladopajic:
```suggestion
    switch.mount(AutonatV2.new(switch, config = config))
```

unnecessary variable

libp2p/builders.nim
```diff
@@ -403,13 +402,15 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
...
+  else:
+    if b.autonat: # withValue doesn't accept elif
+      let autonat = Autonat.new(switch)
+      switch.mount(autonat)
```
vladopajic:
```suggestion
     switch.mount(Autonat.new(switch))
```

libp2p/builders.nim
```diff
@@ -403,13 +402,15 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
...
-  elif b.autonat:
-    let autonat = Autonat.new(switch)
-    switch.mount(autonat)
+    # TODO: start service
+
+  b.autonatV2ServerConfig.withValue(config):
+    let autonatV2Server = AutonatV2.new(switch, config = config)
+    switch.mount(autonatV2Server)
+  else:
+    if b.autonat: # withValue doesn't accept elif
```
vladopajic: then use `if b.autonatV2ServerConfig.isSet()`

libp2p/builders.nim
```diff
@@ -290,15 +289,15 @@ proc withAutonat*(b: SwitchBuilder): SwitchBuilder =
...
 ): SwitchBuilder =
-  b.autonatV2 = true
-  b.autonatV2Config = config
+  b.autonatV2ServerConfig = Opt.some(config)
   b
 
-proc withAutonatV2Client*(b: SwitchBuilder): SwitchBuilder =
+proc withAutonatV2*(b: SwitchBuilder): SwitchBuilder =
   b.autonatV2Client = true
+  # TODO: start service here too
```
vladopajic: it probably should not start here but in start() proc

libp2p/builders.nim
```diff
@@ -403,13 +402,15 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
...
     let autonatV2Client = AutonatV2Client.new(switch.dialer, b.rng)
     switch.mount(autonatV2Client)
-
-  if b.autonatV2:
-    let autonatV2 = AutonatV2.new(switch, config = b.autonatV2Config)
-    switch.mount(autonatV2)
-  elif b.autonat:
-    let autonat = Autonat.new(switch)
-    switch.mount(autonat)
+    # TODO: start service
```
vladopajic: it probably should not start here but in start() proc

libp2p/builders.nim
gmelodie: Using isSet or isSome or any other such options will require us to catch `KeyError` when doing `autonatV2ServerConfig.get()`

libp2p/builders.nim
```diff
@@ -403,13 +402,15 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
...
+
+  b.autonatV2ServerConfig.withValue(config):
+    let autonatV2Server = AutonatV2.new(switch, config = config)
+    switch.mount(autonatV2Server)
+  else:
```
richard-ramos: This is something i missed earlier, but as a server, it should be possible to run both autonatV1 and autonatV2 at same time, since there are nodes that might be running autonatv1 client side, and ideally we can serve them both.

richard-ramos (approved):

PR #1683, opened 2025-09-10
test(rendezvous): error cases

rlve:
Changes:
- add tests to `RendezVous` suite:
  - `"Peer is not registered when peer record validation fails"`
  - `"Unsubscribe for not registered namespace is ignored"`
- add `RendezVous Errors` suite with new tests:
  - `"Node returns ERROR_CODE for invalid message"` - 5 cases
  - `"Node returns NotAuthorized when Register exceeding peer limit"`
- move 2 scenarios from `RendezVous`  to `RendezVous Errors` 
- add new utils


gmelodie (approved):

tests/discovery/testrendezvouserrors.nim
```diff
@@ -0,0 +1,150 @@
...
+      # Discover - Invalid Cookie (malformed/undecodable)
+      (
+        (
+          proc(node: RendezVous): Message =
+            # Empty buffer will fail Cookie.decode().tryGet() and yield InvalidCookie
+            prepareDiscoverMessage(cookie = Opt.some(newSeq[byte]()))
+        ),
+        ResponseStatus.InvalidCookie,
+      ),
+    ]
```
vladopajic, line 106: should be type with `Name: string`, `WatheverThisCallbackIs: proc(node: RendezVous): Messag`, `ExpectedStatus: XXXX` properties.

```
type
  TestCase = (string, proc(node: RendezVous): Message, ResponseStatus)

... or

type TestCase = object
   Name: string,
   WatheverThisCallbackIs:  proc(node: RendezVous): Messag
   ExpectedStatus: XXXX
    


let testCases: seq[TestCase] =
````


use `Name` from comment and append it to `asyncTest`  for identification.

PR #1687, opened 2025-09-12
feat(mix): crypto

richard-ramos:
Moves the crypto logic from mix repo to nim-libp2p

cc: @AkshayaMani

### DIFF
--- a/libp2p/protocols/mix/crypto.nim
+++ b/libp2p/protocols/mix/crypto.nim
@@ -1,12 +1,10 @@
-import endians, nimcrypto, results
+import endians, nimcrypto
 
-# This function processes 'data' using AES in CTR mode.
-# For CTR mode, the same function handles both encryption and decryption.
-proc aes_ctr*(key, iv, data: openArray[byte]): Result[seq[byte], string] =
-  if key.len != 16:
-    return err("Key must be 16 bytes for AES-128")
-  if iv.len != 16:
-    return err("IV must be 16 bytes for AES-128")
+proc aes_ctr*(key, iv, data: openArray[byte]): seq[byte] =
+  ## Processes 'data' using AES in CTR mode.
+  ## For CTR mode, the same function handles both encryption and decryption.
+  doAssert key.len == 16, "Key must be 16 bytes for AES-128"
+  doAssert iv.len == 16, "IV must be 16 bytes for AES-128"
 
   var
     ctx: CTR[aes128]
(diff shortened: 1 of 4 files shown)


libp2p/protocols/mix/crypto.nim
```diff
@@ -0,0 +1,60 @@
+import endians, nimcrypto, results
+
+# This function processes 'data' using AES in CTR mode.
+# For CTR mode, the same function handles both encryption and decryption.
+proc aes_ctr*(key, iv, data: openArray[byte]): Result[seq[byte], string] =
+  if key.len != 16:
+    return err("Key must be 16 bytes for AES-128")
+  if iv.len != 16:
+    return err("IV must be 16 bytes for AES-128")
```
vladopajic: what about `doAssert key.len != 16, "Key must be 16 bytes for AES-128"`

tests/mix/testcrypto.nim
```diff
@@ -0,0 +1,210 @@
...
+    let encryptedResult = aes_ctr(key, iv, data)
+    if encryptedResult.isErr:
+      error "Encryption failed", err = encryptedResult.error
+      fail()
+    let encrypted = encryptedResult.get()
+
+    let decryptedResult = aes_ctr(key, iv, encrypted)
+    if decryptedResult.isErr:
+      error "Decryption failed", err = decryptedResult.error
+      fail()
```
vladopajic: hmm we could have this as utility-template, it actually seems vary interesting to be used in general. 
`require {condition}, {message}` 

in this case:
`require decryptedResult.isErr, "Decryption failed", decryptedResult.error`
`require {condition} {msg1} {msg2} .. {msgN}`

these tests here would greatly benefit as code would be much more smaller and readable. also we should adopt that in other tests in this repo.

tests/mix/testcurve25519.nim
```diff
@@ -0,0 +1,74 @@
+{.used.}
+
+import chronicles, results, unittest
+import ../../libp2p/crypto/curve25519
+import ../../libp2p/protocols/mix/curve25519
+
+# Helper function for non-zero check
```
gmelodie:
```suggestion
```

tests/mix/testcurve25519.nim
```diff
@@ -0,0 +1,74 @@
...
+
+    var res1Result = multiplyBasePointWithScalars(@[x1, x2])
+    if res1Result.isErr:
+      error "Multiply base point with scalar errors", err = res1Result.error
+      fail()
+    let res1 = res1Result.get()
+
+    let
+      intermediate = public(x2)
+      res2 = multiplyPointWithScalars(intermediate, @[x1])
```
gmelodie: Why not do the following?
```suggestion
    let
      res2 = multiplyBasePointWithScalars(@[x2, x1])
```

If you're testing the commutative property, one would expect to see the exact same operation but with swapped inputs.

tests/mix/testcrypto.nim
```diff
@@ -0,0 +1,210 @@
...
+  test "aes_ctr_encrypt_decrypt":
+    let
+      key = cast[array[16, byte]]("thisis16byteskey")
+      iv = cast[array[16, byte]]("thisis16bytesiv!")
+      data: seq[byte] = cast[seq[byte]]("thisisdata")
+
+    let encryptedResult = aes_ctr(key, iv, data)
+    if encryptedResult.isErr:
+      error "Encryption failed", err = encryptedResult.error
+      fail()
```
gmelodie: Why not use `doAssert` here? Same for all other parts where this pattern happens.
```suggestion
      doAssert false, "Encryption failed: " & $encryptedResult.error
```

tests/mix/testcrypto.nim
```diff
@@ -0,0 +1,210 @@
...
+      error "Decrypted data does not match the original data",
+        original = data, decrypted = decrypted
+      fail()
+
+    if data == encrypted:
+      error "Encrypted data should not match the original data",
+        original = data, encrypted = encrypted
+      fail()
+
+  test "sha256_hash_computation":
```
gmelodie, line 20: Nit: use a template/macro/proc to avoid repeating `..._computation` logic

tests/mix/testcrypto.nim
```diff
@@ -0,0 +1,210 @@
...
+      expectedHmacHex = "b075dd302655e085d35e8cef5dfdf101"
+      expectedHmac = fromHex(expectedHmacHex)
+      hmacResult = hmac(key, data)
+
+    if hmacResult != expectedHmac:
+      error "HMAC does not match the expected HMAC",
+        hmac = hmacResult, expected = expectedHmac
+      fail()
+
+  test "aes_ctr_empty_data":
```
gmelodie, line 49: Same comment as above: use template/macro/proc to avoid repeating `..._empty_data` logic

gmelodie, line 49: Same comment as above: use template/macro/proc to avoid repeating `..._empty_data` logic

libp2p/protocols/mix/crypto.nim
```diff
@@ -0,0 +1,60 @@
...
+
+# This function advances the counter in the AES-CTR IV by a specified number of blocks.
+proc advance_ctr*(iv: var openArray[byte], blocks: int) =
+  var counter: uint64
+  bigEndian64(addr counter, addr iv[8])
+  counter += uint64(blocks)
+  bigEndian64(addr iv[8], addr counter)
+
+# This function encrypting processes 'data' using AES in CTR mode from startIndex, without processing all preceding data.
+# For CTR mode, the same function handles both encryption and decryption.
```
gmelodie: These should be documentation comments (i.e. `## this function...` inside the function). Also trim `This function`, no need for that. Same for similar comments in this file

libp2p/protocols/mix/crypto.nim
```diff
@@ -0,0 +1,60 @@
...
+  ctx.clear()
+
+  return ok(output)
+
+# This function advances the counter in the AES-CTR IV by a specified number of blocks.
+proc advance_ctr*(iv: var openArray[byte], blocks: int) =
+  var counter: uint64
+  bigEndian64(addr counter, addr iv[8])
+  counter += uint64(blocks)
+  bigEndian64(addr iv[8], addr counter)
```
gmelodie: Could you maybe have `counter` converted do `iv` only when you need it to be that way? I don't know which format of `counter`/`iv` is used more often, but if it's `uint64` (aka `counter`) then we should use that and use sth like `toOpenArray` when needed

tests/mix/testcurve25519.nim
richard-ramos: Actually not sure. I guess with using `multiplyPointWithScalars` it's clear that it will be `intermediate` the value that is going to be multiplied against the sequence of FieldElements received in second parameter :thinking:

richard-ramos: Ended up adding your suggestion to the existing test unit to check for commutivity

tests/mix/testcrypto.nim
richard-ramos: Ended up using `.expect()` and `check` since the former will concatenate the error message to the exception, and the latter since some of the messages are really not necessary, like:

```
    if emptyData != decrypted:
      error "Decrypted empty data does not match the original empty data",
        original = emptyData, decrypted = decrypted
      fail()
```

vs

```
check emptyData == decrypted
```

The test runner will inform us of that these variables dont match anyway.
This ended up reducing a lot of the test unit code

PR #1688, opened 2025-09-12
feat(mix): sequence number generator and tag manager

richard-ramos:

### DIFF
--- a/libp2p/protocols/mix/seqno_generator.nim
+++ b/libp2p/protocols/mix/seqno_generator.nim
@@ -1,0 +1,31 @@
+import std/endians, times
+import ../../peerid
+import ./crypto
+
+type SeqNo* = object
+  counter: uint32
+
+proc init*(T: typedesc[SeqNo], peerId: PeerId): T =
+  var seqNo = T()
+  let peerIdHash = sha256_hash(peerId.data)
+  for i in 0 .. 3:
+    seqNo.counter = seqNo.counter or (uint32(peerIdHash[i]) shl (8 * (3 - i)))
+  return seqNo
+
+proc generateSeqNo*(seqNo: var SeqNo, messageBytes: seq[byte]) =
+  let
+    currentTime = getTime().toUnix() * 1000
+    currentTimeBytes = newSeq[byte](8)
+  bigEndian64(unsafeAddr currentTimeBytes[0], unsafeAddr currentTime)
+  let messageHash = sha256_hash(messageBytes & currentTimeBytes)
+  var cnt: uint32
+  for i in 0 .. 3:
+    cnt = cnt or (uint32(messageHash[i]) shl (8 * (3 - i)))
+  seqNo.counter = (seqNo.counter + cnt) mod high(uint32)
+
+proc incSeqNo*(seqNo: var SeqNo) =
+  seqNo.counter = (seqNo.counter + 1) mod high(uint32)
+    # ToDo: Manage sequence no. overflow in a way that it does not affect re-assembly
+
+proc getSeqNo*(seqNo: SeqNo): uint32 =
+  return seqNo.counter
--- a/libp2p/protocols/mix/seqno_generator.nim
+++ b/libp2p/protocols/mix/seqno_generator.nim
@@ -1,31 +1,29 @@
 import std/endians, times
 import ../../peerid
 import ./crypto
+import ../../utils/sequninit
 
-type SeqNo* = object
-  counter: uint32
+type SeqNo* = uint32
 
-proc init*(T: typedesc[SeqNo], peerId: PeerId): T =
-  var seqNo = T()
-  let peerIdHash = sha256_hash(peerId.data)
+proc init*(T: typedesc[SeqNo], data: seq[byte]): T =
+  var seqNo: SeqNo = 0
+  let hash = sha256_hash(data)
   for i in 0 .. 3:
-    seqNo.counter = seqNo.counter or (uint32(peerIdHash[i]) shl (8 * (3 - i)))
+    seqNo = seqNo or (uint32(hash[i]) shl (8 * (3 - i)))
   return seqNo
 
-proc generateSeqNo*(seqNo: var SeqNo, messageBytes: seq[byte]) =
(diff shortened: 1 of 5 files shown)


libp2p/protocols/mix/seqno_generator.nim
```diff
@@ -0,0 +1,31 @@
...
+  var seqNo = T()
+  let peerIdHash = sha256_hash(peerId.data)
+  for i in 0 .. 3:
+    seqNo.counter = seqNo.counter or (uint32(peerIdHash[i]) shl (8 * (3 - i)))
+  return seqNo
+
+proc generateSeqNo*(seqNo: var SeqNo, messageBytes: seq[byte]) =
+  let
+    currentTime = getTime().toUnix() * 1000
+    currentTimeBytes = newSeq[byte](8)
```
vladopajic:
```suggestion
    currentTimeBytes = newSeqUninit[byte](8)
```

i know it's just small size but let's be consistent

tests/mix/testtagmanager.nim
```diff
@@ -0,0 +1,100 @@
...
+    tm = TagManager.new()
+
+  teardown:
+    clearTags(tm)
+
+  test "add_and_check_tag":
+    let tagRes = generateRandomFieldElement()
+    if tagRes.isErr:
+      error "Generate random field element error", err = tagRes.error
+      fail()
```
gmelodie: As mentioned in the other PR, we should get rid of these patterns, either creating a template/macro/proc or using `doAssert`

tests/mix/testtagmanager.nim
```diff
@@ -0,0 +1,100 @@
...
+    let tag = tagRes.get()
+
+    addTag(tm, tag)
+    if not isTagSeen(tm, tag):
+      error "Tag should be seen after adding", tag = tag
+      fail()
+
+    removeTag(tm, tag)
+    if isTagSeen(tm, tag):
+      error "Tag should not be seen after removal", tag = tag
```
gmelodie: `error` on tests? Why not:

```suggestion
      check not isTagSeen(tm, tag), "Tag should not be seen after removal"
```

Same for other places where this happens

libp2p/protocols/mix/seqno_generator.nim
```diff
@@ -0,0 +1,31 @@
+import std/endians, times
+import ../../peerid
+import ./crypto
+
+type SeqNo* = object
+  counter: uint32
```
gmelodie:
```suggestion
type SeqNo* = uint32
```

libp2p/protocols/mix/seqno_generator.nim
```diff
@@ -0,0 +1,31 @@
...
+
+proc generateSeqNo*(seqNo: var SeqNo, messageBytes: seq[byte]) =
+  let
+    currentTime = getTime().toUnix() * 1000
+    currentTimeBytes = newSeq[byte](8)
+  bigEndian64(unsafeAddr currentTimeBytes[0], unsafeAddr currentTime)
+  let messageHash = sha256_hash(messageBytes & currentTimeBytes)
+  var cnt: uint32
+  for i in 0 .. 3:
+    cnt = cnt or (uint32(messageHash[i]) shl (8 * (3 - i)))
```
gmelodie: This is the same logic as `init`, there should only be one function for that. The rest should be wrappers that call the function

libp2p/protocols/mix/seqno_generator.nim
```diff
@@ -0,0 +1,31 @@
...
+  seqNo.counter = (seqNo.counter + cnt) mod high(uint32)
+
+proc incSeqNo*(seqNo: var SeqNo) =
```
gmelodie:
```suggestion
proc inc*(seqNo: var SeqNo) =
```

libp2p/protocols/mix/seqno_generator.nim
```diff
@@ -0,0 +1,31 @@
...
+  seqNo.counter = (seqNo.counter + 1) mod high(uint32)
+    # ToDo: Manage sequence no. overflow in a way that it does not affect re-assembly
+
+proc getSeqNo*(seqNo: SeqNo): uint32 =
```
gmelodie:
```suggestion
proc get*(seqNo: SeqNo): uint32 =
```

libp2p/protocols/mix/seqno_generator.nim
```diff
@@ -0,0 +1,31 @@
...
+  counter: uint32
+
+proc init*(T: typedesc[SeqNo], peerId: PeerId): T =
+  var seqNo = T()
+  let peerIdHash = sha256_hash(peerId.data)
+  for i in 0 .. 3:
+    seqNo.counter = seqNo.counter or (uint32(peerIdHash[i]) shl (8 * (3 - i)))
+  return seqNo
+
+proc generateSeqNo*(seqNo: var SeqNo, messageBytes: seq[byte]) =
```
gmelodie:
```suggestion
proc generate*(seqNo: var SeqNo, messageBytes: seq[byte]) =
```

tests/mix/testseqnogenerator.nim
```diff
@@ -0,0 +1,114 @@
...
+    let
+      peerId =
+        PeerId.init("16Uiu2HAmFkwLVsVh6gGPmSm9R3X4scJ5thVdKfWYeJsKeVrbcgVC").get()
+      msg = @[byte 1, 2, 3]
+    var seqNo = SeqNo.init(peerId)
+
+    generateSeqNo(seqNo, msg)
+    let seqNo1 = seqNo.counter
+
+    sleep(1000) # Wait for 1 second
```
gmelodie: can you do sth like this using `std/times`?
```suggestion
    sleep(1.seconds)
```

richard-ramos: No, but will use a constant for the 1000 so it looks less ugly

tests/mix/testtagmanager.nim
richard-ramos: Got rid of those by using `expect`

gmelodie (approved):

PR #1689, opened 2025-09-12
feat(mix): serialization

richard-ramos:

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -83,7 +83,8 @@ type
     peerStoreCapacity: Opt[int]
     autonat: bool
     autonatV2ServerConfig: Opt[AutonatV2Config]
-    autonatV2Client: bool
+    autonatV2Client: AutonatV2Client
+    autonatV2ServiceConfig: AutonatV2ServiceConfig
     autotls: AutotlsService
     circuitRelay: Relay
     rdv: RendezVous
@@ -296,8 +297,11 @@ proc withAutonatV2Server*(
   b.autonatV2ServerConfig = Opt.some(config)
   b
 
-proc withAutonatV2*(b: SwitchBuilder): SwitchBuilder =
-  b.autonatV2Client = true
+proc withAutonatV2*(
+    b: SwitchBuilder, serviceConfig = AutonatV2ServiceConfig.new()
+): SwitchBuilder =
+  b.autonatV2Client = AutonatV2Client.new(b.rng)
+  b.autonatV2ServiceConfig = serviceConfig
   b
 
 when defined(libp2p_autotls_support):
@@ -386,6 +390,13 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
   if b.enableWildcardResolver:
     b.services.insert(WildcardAddressResolverService.new(), 0)
 
+  if not isNil(b.autonatV2Client):
+    b.services.add(
+      AutonatV2Service.new(
+        b.rng, client = b.autonatV2Client, config = b.autonatV2ServiceConfig
+      )
+    )
+
   let switch = newSwitch(
     peerInfo = peerInfo,
     transports = transports,
@@ -399,10 +410,9 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
 
   switch.mount(identify)
 
-  if b.autonatV2Client:
-    let autonatV2Client = AutonatV2Client.new(switch.dialer, b.rng)
-    b.services.add(AutonatV2Service.new(b.rng, client = autonatV2Client))
-    switch.mount(autonatV2Client)
+  if not isNil(b.autonatV2Client):
+    b.autonatV2Client.setup(switch)
+    switch.mount(b.autonatV2Client)
 
   b.autonatV2ServerConfig.withValue(config):
     switch.mount(AutonatV2.new(switch, config = config))
--- a/libp2p/protocols/connectivity/autonatv2/client.nim
+++ b/libp2p/protocols/connectivity/autonatv2/client.nim
@@ -58,11 +58,10 @@ proc handleDialBack(
 
 proc new*(
     T: typedesc[AutonatV2Client],
-    dialer: Dial,
     rng: ref HmacDrbgContext,
     dialBackTimeout: Duration = DefaultDialBackTimeout,
 ): T =
-  let client = T(dialer: dialer, rng: rng, dialBackTimeout: dialBackTimeout)
+  let client = T(rng: rng, dialBackTimeout: dialBackTimeout)
 
   # handler for DialBack messages
   proc handleStream(
@@ -86,6 +85,9 @@ proc new*(
   client.handler = handleStream
   client.codec = $AutonatV2Codec.DialBack
   client
+
+proc setup*(self: AutonatV2Client, switch: Switch) =
+  self.dialer = switch.dialer
 
 proc handleDialDataRequest*(
     conn: Connection, req: DialDataRequest
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -50,14 +50,6 @@ proc new(
 method getWrapped*(self: QuicStream): P2PConnection =
   self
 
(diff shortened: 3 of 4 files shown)


vladopajic (approved):

libp2p/protocols/mix/config.nim
```diff
@@ -0,0 +1,19 @@
...
+  delaySize* = 2 # Delay size
+  addrSize* = (t * k) - delaySize # Address size
+  packetSize* = 4608 # Total packet size (from spec)
+  messageSize* = packetSize - headerSize - k # Size of the message itself
+  payloadSize* = messageSize + k # Total payload size
+  surbSize* = headerSize + k + addrSize
+    # Size of a surb packet inside the message payload
+  surbLenSize* = 1 # Size of the field storing the number of surbs
+  surbIdLen* = k # Size of the identifier used when sending a message with surb
+  defaultSurbs* = uint8(4) # Default number of SURBs to send
```
gmelodie: This should be an object `MixConfig` inside `mix/serialization.nim` and have a `new` constructor with these values as default. If not all values are defaults but actual hard constants, then they should be constants, still on that file. Also, capitalize names e.g. `DefaultSurbs`

libp2p/protocols/mix/serialization.nim
```diff
@@ -0,0 +1,227 @@
...
+  let header = Header(
+    Alpha: serializedHeader[0 .. (alphaSize - 1)],
+    Beta: serializedHeader[alphaSize .. (alphaSize + betaSize - 1)],
+    Gamma: serializedHeader[(alphaSize + betaSize) .. (headerSize - 1)],
+  )
+
+  ok(header)
+
+type Message* = object
+  Content: seq[byte]
```
gmelodie:
```suggestion
type Message* = seq[byte]
```

libp2p/protocols/mix/serialization.nim
```diff
@@ -0,0 +1,227 @@
...
+
+proc init*(T: typedesc[Message], content: seq[byte]): T =
+  return T(Content: content)
+
+proc getContent*(message: Message): seq[byte] =
+  return message.Content
+
```
gmelodie: If we do`type Message* = seq[byte]` from suggestion above then we no longer need these
```suggestion
```

libp2p/protocols/mix/serialization.nim
```diff
@@ -0,0 +1,227 @@
...
+proc deserialize*(
+    T: typedesc[Message], serializedMessage: openArray[byte]
+): Result[T, string] =
+  if len(serializedMessage) != payloadSize:
+    return err("Serialized message must be exactly " & $payloadSize & " bytes")
+  return ok(T(Content: serializedMessage[k ..^ 1]))
+
+type Hop* = object
+  MultiAddress: seq[byte]
```
gmelodie, line 75:
```suggestion
type Hop* = seq[byte]
```

libp2p/protocols/mix/serialization.nim
```diff
@@ -0,0 +1,227 @@
...
+  T(
+    MultiAddress:
+      if multiAddress == newSeq[byte](addrSize):
+        @[]
+      else:
+        multiAddress
+  )
+
+proc getHop*(hop: Hop): seq[byte] =
+  return hop.MultiAddress
```
gmelodie: Same as above, we won't need this
```suggestion
```

libp2p/protocols/mix/serialization.nim
```diff
@@ -0,0 +1,227 @@
...
+    )
+  )
+
+type RoutingInfo* = object
+  Addr: Hop
+  Delay: seq[byte]
+  Gamma: seq[byte]
+  Beta: seq[byte]
+
+proc init*(
```
gmelodie, line 114: Change this to `new` and consider adding sane defaults here (maybe for `delay`, `gamma` and `beta`

libp2p/protocols/mix/serialization.nim
```diff
@@ -0,0 +1,227 @@
...
+      Beta:
+        data[(addrSize + delaySize + gammaSize) .. (((r * (t + 1)) + t + 2) * k) - 1],
+    )
+  )
+
+type SphinxPacket* = object
+  Hdr*: Header
+  Payload*: seq[byte]
+
+proc init*(T: typedesc[SphinxPacket], header: Header, payload: seq[byte]): T =
```
gmelodie, line 160: `new` instead of init, with defaults

libp2p/protocols/mix/serialization.nim
```diff
@@ -0,0 +1,227 @@
...
+  T(Hdr: header, Payload: payload)
+
+proc getSphinxPacket*(packet: SphinxPacket): (Header, seq[byte]) =
```
gmelodie:
```suggestion
proc get*(packet: SphinxPacket): (Header, seq[byte]) =
```

libp2p/protocols/mix/serialization.nim
```diff
@@ -0,0 +1,227 @@
...
+    return err("Sphinx packet size must be exactly " & $packetSize & " bytes")
+
+  let header = ?Header.deserialize(data)
+
+  return ok(SphinxPacket(Hdr: header, Payload: data[headerSize ..^ 1]))
+
+type
+  secret* = seq[seq[byte]]
+
+  key* = seq[byte]
```
gmelodie: Types should be capitalized

libp2p/protocols/mix/serialization.nim
```diff
@@ -0,0 +1,227 @@
...
+
+    let slice = data[offset ..< offset + dataLen]
+    offset += dataLen
+    return ok(slice)
+  elif data.len < offset:
+    return err("not enough data")
+
+  let slice = data[offset .. ^1]
+  offset = data.len
+  return ok(slice)
```
gmelodie: 1. use guards to avoid having the entire function inside an `if` statement
2. dataLen should be called `readSize` or `size` and should be an `Opt` instead of allowing negative values (which would save us an `if` statement)
3. check if `offset` is outside of `data`'s length

tests/mix/testserialization.nim
```diff
@@ -0,0 +1,145 @@
+{.used.}
+
+import chronicles, results, unittest
+import ../../libp2p/protocols/mix/[config, serialization]
+
+# Define test cases
+suite "serialization_tests":
+  test "serialize_and_deserialize_header":
+    let header = Header.init(
```
gmelodie, line 9: If we made this into a `new` function with defaults, we could do
```suggestion
    let header = Header.new()
```

tests/mix/testserialization.nim
```diff
@@ -0,0 +1,145 @@
...
+    let serializedRes = header.serialize()
+    if serializedRes.isErr:
+      error "Failed to serialize header", err = serializedRes.error
+      fail()
+    let serialized = serializedRes.get()
+
+    if len(serialized) != headerSize:
+      error "Serialized header size is incorrect",
+        expected = headerSize, actual = len(serialized)
+      fail()
```
gmelodie: This pattern again i swear to god

libp2p/protocols/mix/serialization.nim
richard-ramos: The convention in nim is to use `new` for things that live in the heap (i.e. ref objects), and `init` for things that live on the stack.

libp2p/protocols/mix/serialization.nim
richard-ramos: The convention in nim is to use `new` for things that live in the heap (i.e. ref objects), and `init` for things that live on the stack.

libp2p/protocols/mix/serialization.nim
richard-ramos: I'm gonna keep this as an object because It's possible that Hop will need additional fields.

tests/mix/testserialization.nim
richard-ramos: It's just that having default values for the header would make no sense, as those values are meant to be calculated

gmelodie (approved):

PR #1690, opened 2025-09-12
feat(mix): message

richard-ramos:

### DIFF
--- a/libp2p/protocols/mix/mix_message.nim
+++ b/libp2p/protocols/mix/mix_message.nim
@@ -1,0 +1,47 @@
+import chronicles, results
+import stew/[byteutils, leb128]
+import ../../protobuf/minprotobuf
+import ../../utils/sequninit
+
+type MixMessage* = object
+  message*: seq[byte]
+  codec*: string
+
+proc init*(T: typedesc[MixMessage], message: openArray[byte], codec: string): T =
+  return T(message: @message, codec: codec)
+
+proc serialize*(mixMsg: MixMessage): seq[byte] =
+  let vbytes = toBytes(mixMsg.codec.len.uint64, Leb128)
+  doAssert vbytes.len <= 2, "serialization failed: codec length exceeds 2 bytes"
+
+  var buf = newSeqUninit[byte](vbytes.len + mixMsg.codec.len + mixMsg.message.len)
+  buf[0 ..< vbytes.len] = vbytes.toOpenArray()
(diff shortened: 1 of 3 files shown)


libp2p/protocols/mix/mix_message.nim
```diff
@@ -0,0 +1,48 @@
...
+proc init*(T: typedesc[MixMessage], message: openArray[byte], codec: string): T =
+  return T(message: @message, codec: codec)
+
+proc serialize*(mixMsg: MixMessage): Result[seq[byte], string] =
+  let vbytes = toBytes(mixMsg.codec.len.uint64, Leb128)
+  if vbytes.len > 2:
+    return err("serialization failed: codec length exceeds 2 bytes")
+
+  var buf =
+    newSeqUninitialized[byte](vbytes.len + mixMsg.codec.len + mixMsg.message.len)
```
vladopajic: we use `newSeqUninit` template because `newSeqUninitialized` is only added in v2.2 nim

libp2p/protocols/mix/mix_message.nim
```diff
@@ -0,0 +1,48 @@
+import chronicles, results
+import stew/[byteutils, leb128]
+import ../../protobuf/minprotobuf
+
+type MixMessage* = object
+  message*: seq[byte]
+  codec*: string
+
+proc init*(T: typedesc[MixMessage], message: openArray[byte], codec: string): T =
```
gmelodie, line 10: `new` with defaults

tests/mix/testmixmessage.nim
```diff
@@ -0,0 +1,64 @@
...
+  test "serialize_and_deserialize_mix_message":
+    let
+      message = "Hello World!"
+      codec = "/test/codec/1.0.0"
+      mixMsg = MixMessage.init(message.toBytes(), codec)
+
+    let serializedResult = mixMsg.serialize()
+    if serializedResult.isErr:
+      error "Serialization failed", err = serializedResult.error
+      fail()
```
gmelodie: Again this pattern

libp2p/protocols/mix/mix_message.nim
richard-ramos: The convention in nim is to use new for things that live in the heap (i.e. ref objects), and init for things that live on the stack. Wrt defaults, there are no defaults here, as both fields are really required when creating an instance of MixMessage and setting an "" or a @[] for the message and codec would make no sense

gmelodie (approved):

PR #1695, opened 2025-09-16
chore(autonat-v2): add interop tests

gmelodie:
Simple interop tests to ensure autontav2 works with go-libp2p. For that we have to clone the go-libp2p repo and patch it to allow private IP addresses (for testing). They have the `allowPrivateAddrs` option but it's currently private, so there's no other way of accessing it.


interop/autonatv2/nim-peer/src/nim_peer.nim
```diff
@@ -0,0 +1,45 @@
+import os, chronos, libp2p
+import libp2p/protocols/connectivity/autonatv2/service
+import libp2p/protocols/connectivity/autonatv2/types
+
+proc main() {.async.} =
+  if paramCount() != 1:
+    quit("Usage: nim r src/nim_peer.nim <peerid>", 1)
+
+  echo paramStr(0)
+  echo paramStr(1)
```
vladopajic: how about info log from chronicles?

interop/autonatv2/nim-peer/src/nim_peer.nim
```diff
@@ -0,0 +1,45 @@
+import os, chronos, libp2p
+import libp2p/protocols/connectivity/autonatv2/service
+import libp2p/protocols/connectivity/autonatv2/types
+
+proc main() {.async.} =
+  if paramCount() != 1:
+    quit("Usage: nim r src/nim_peer.nim <peerid>", 1)
```
vladopajic, line 21: how about error log from chronicles?

interop/autonatv2/nim-peer/nim_peer.nimble
```diff
@@ -0,0 +1,12 @@
+# Package
+
+version = "0.1.0"
+author = "Gabriel Cruz"
+description = "A new awesome nimble package"
+license = "MIT"
```
vladopajic: author should be Status Research & Development GmbH, but not really sure

interop/autonatv2/nim-peer/nim_peer.nimble
```diff
@@ -0,0 +1,12 @@
+# Package
+
+version = "0.1.0"
+author = "Gabriel Cruz"
+description = "A new awesome nimble package"
```
vladopajic: description: "Auto NAT Peer" ...

interop/autonatv2/go-peer/testautonatv2.go
```diff
@@ -0,0 +1,100 @@
...
+
+	libp2p "github.com/libp2p/go-libp2p"
+	crypto "github.com/libp2p/go-libp2p/core/crypto"
+	peer "github.com/libp2p/go-libp2p/core/peer"
+	b58 "github.com/mr-tron/base58"
+)
+
+const (
+	privKeyFile = "peer.key"
+	peerIDFile  = "peer.id"
```
vladopajic, line 17: hardcode values to consts here, remove extra step where values are loaded from file

.github/workflows/interop.yml
```diff
@@ -60,3 +60,30 @@ jobs:
...
+        with:
+          nim-version: "stable"
+
+      - name: Run Go and Nim together
+        run: |
+          nimble install
+          cd interop/autonatv2/go-peer
+          git clone https://github.com/libp2p/go-libp2p
+          cd go-libp2p
+          git apply ../disable-filtering-of-private-ip-addresses.patch
```
vladopajic, line 85: what the hack is this hack

interop/autonatv2/go-peer/testautonatv2.go
vladopajic: i see... file is used as config for values, but still i think we can hardcode those values in both nim and go side

.github/workflows/interop.yml
vladopajic: what if this repo was fork with this patch?

vladopajic: anyway ` git clone https://github.com/libp2p/go-libp2p` has to be pinned to commit

interop/autonatv2/nim-peer/src/nim_peer.nim
gmelodie: Sorry leftover forgot to remove this altogether

interop/autonatv2/nim-peer/src/nim_peer.nim
gmelodie: Not sure. This is an usage error, not really a log. Imagine if you used `cat` wrong and got a log instead of just some usage text

.github/workflows/interop.yml
gmelodie: I thought about that too. I dont want to pin a commit on purpose, because we want the interop test to fail if the newest commit has changes that break interop

vladopajic: why is it needed to break?  whats is the purpose of braking ci this way?

gmelodie: Well the goal is to test whether our autonatv2 implementation can communicate with go-libp2p's implementation. So if go-libp2p's impl changed, this test should start to fail

interop/autonatv2/go-peer/disable-filtering-of-private-ip-addresses.patch
```diff
@@ -0,0 +1,119 @@
...
+-			continue
+-		}
+ 		if !as.dialerHost.Network().CanDial(p, a) {
+ 			continue
+ 		}
+@@ -206,6 +204,7 @@ func (as *server) serveDialRequest(s network.Stream) EventDialRequestCompleted {
+ 		addrIdx = i
+ 		break
+ 	}
++	log.Debug("here0")
```
richard-ramos: Are these `log.Debug("hereN")` intentional or leftover?

interop/autonatv2/nim-peer/src/nim_peer.nim
```diff
@@ -0,0 +1,46 @@
...
+
+  # ensure go peer is started
+  await sleepAsync(3.seconds)
```
richard-ramos, line 24: This might be flaky. Consider doing something like this:

```nim
import net, os

proc waitForService(host: string, port: Port, retries: int = 20, delayMs: Duration = 500.milliseconds): bool {.async.} =
  for i in 0..<retries:
    try:
      var s = newSocket()
      s.connect(host, port)
      s.close()
      return true
    except OSError:
      discard
    await asyncSleep(delay)
  return false

when isMainModule:
  if await waitForService("127.0.0.1", Port(4040)):
    echo "service is up"
  else:
    quit("timeout waiting for service", 1)

```

.github/workflows/interop.yml
vladopajic: that is not correct to expect, because changes in code doesn't mean changes in logic. 
if someone changes file in go-libp2p (formatting) we will brake code here.

interop/autonatv2/go-peer/disable-filtering-of-private-ip-addresses.patch
gmelodie: leftover :smiling_face_with_tear:

PR #1704, opened 2025-09-18
feat(mix): reply connection

richard-ramos:


libp2p/protocols/mix/entry_connection.nim
```diff
@@ -5,14 +5,27 @@ import ../../utils/sequninit
 import ./mix_protocol
 from fragmentation import DataSize
 
+const DefaultSurbs = uint8(4) # Default number of SURBs to send
```
gmelodie:
```suggestion
const DefaultSurbs = uint8(4)
```

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -5,14 +5,27 @@ import ../../utils/sequninit
...
+
 type MixDialer* = proc(
   msg: seq[byte], codec: string, destination: MixDestination
 ): Future[void] {.async: (raises: [CancelledError, LPStreamError], raw: true).}
 
+## Holds configuration parameters used in mix protocol operations.
```
gmelodie:
```suggestion
```

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -5,14 +5,27 @@ import ../../utils/sequninit
...
+type MixParameters* = object
+  expectReply*: Opt[bool]
+  numSurbs*: Opt[uint8]
+
 type MixEntryConnection* = ref object of Connection
   destination: MixDestination
   codec: string
   mixDialer: MixDialer
+  params: Opt[MixParameters]
+
```
gmelodie: Newline inside a type?
```suggestion
```

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -26,32 +39,126 @@ chronicles.formatIt(MixEntryConnection):
...
+  except CancelledError as exc:
+    raise exc
+  except LPStreamEOFError as exc:
+    raise exc
+  except CatchableError as exc:
+    raise (ref LPStreamError)(msg: "error in readOnce: " & exc.msg, parent: exc)
+
+  let toRead = min(nbytes, s.cached.len)
+  copyMem(pbytes, addr s.cached[0], toRead)
+  s.cached = s.cached[toRead ..^ 1]
```
gmelodie, line 57: I think there should be a better way to do this @vladopajic do you know?

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -26,32 +39,126 @@ chronicles.formatIt(MixEntryConnection):
...
+  return toRead
 
 method readExactly*(
     s: MixEntryConnection, pbytes: pointer, nbytes: int
 ): Future[void] {.async: (raises: [CancelledError, LPStreamError]), public.} =
-  # TODO: implement
-  raise newLPStreamEOFError()
+  ## Waits for `nbytes` to be available, then read
+  ## them and return them
```
gmelodie:
```suggestion
  ## Waits until `nbytes` are available, then returns them
```

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -26,32 +39,126 @@ chronicles.formatIt(MixEntryConnection):
...
     s: MixEntryConnection, limit = 0, sep = "\r\n"
 ): Future[string] {.async: (raises: [CancelledError, LPStreamError]), public.} =
-  # TODO: implement
-  raise newLPStreamEOFError()
+  ## Reads up to `limit` bytes are read, or a `sep` is found
+  # TODO replace with something that exploits buffering better
+  var lim = if limit <= 0: -1 else: limit
+  var state = 0
+
+  while true:
```
gmelodie: I think if this while loop looked a bit more like the one in the previous function we'd be able to get rid of some `if..else`s:
```nim
  var read = 0
  while read < nbytes and not (s.atEof()):
  ...
```

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -26,32 +39,126 @@ chronicles.formatIt(MixEntryConnection):
...
+    else:
+      state = 0
+      if limit > 0:
+        let missing = min(state, lim - len(result) - 1)
+        result.add(sep[0 ..< missing])
+      else:
+        result.add(sep[0 ..< state])
+
+      result.add(ch)
+      if len(result) == lim:
```
gmelodie: Let's not use `result`, if we implement the suggestion above we'll be able to get rid of it too

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -26,32 +39,126 @@ chronicles.formatIt(MixEntryConnection):
...
+        break
 
 method readVarint*(
     conn: MixEntryConnection
 ): Future[uint64] {.async: (raises: [CancelledError, LPStreamError]), public.} =
-  # TODO: implement
-  raise newLPStreamEOFError()
+  var buffer: array[10, byte]
```
gmelodie: What is `10`? This should be a `const`

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -26,32 +39,126 @@ chronicles.formatIt(MixEntryConnection):
...
+    await conn.readExactly(addr buffer[i], 1)
+
+    var
+      varint: uint64
+      length: int
+    let res = PB.getUVarint(buffer.toOpenArray(0, i), length, varint)
+    if res.isOk():
+      return varint
+    if res.error() != VarintError.Incomplete:
+      break
```
gmelodie: Maybe we could raise an error here and pass `res.error` to get a more accurate error message?

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -26,32 +39,126 @@ chronicles.formatIt(MixEntryConnection):
...
+  if true: # can't end with a raise apparently
+    raise (ref InvalidVarintError)(msg: "Cannot parse varint")
```
gmelodie: Image

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -26,32 +39,126 @@ chronicles.formatIt(MixEntryConnection):
...
 
 method readLp*(
     s: MixEntryConnection, maxSize: int
 ): Future[seq[byte]] {.async: (raises: [CancelledError, LPStreamError]), public.} =
-  # TODO: implement
-  raise newLPStreamEOFError()
+  ## read length prefixed msg, with the length encoded as a varint
+  let
+    length = await s.readVarint()
+    maxLen = uint64(if maxSize < 0: int.high else: maxSize)
```
gmelodie: So if we call `s.readLp(-1)` we want to read `int.high` bytes? This seems counter intuitive

obs: ignore this comment if this is default behavior for other `readLp`s

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -92,42 +199,64 @@ proc shortLog*(self: MixEntryConnection): string {.raises: [].} =
...
     codec: string,
+    params: Opt[MixParameters],
 ): T {.raises: [].} =
+  let params = params.get(MixParameters())
+  let expectReply = params.expectReply.get(false)
+  let numSurbs =
+    if expectReply:
+      params.numSurbs.get(DefaultSurbs)
+    else:
+      0
```
gmelodie, line 98: nit
```suggestion
  let expectReply = params.expectReply.get(false)
  var numSurbs = 0
  if expectReply:
    numSurbs = params.numSurbs.get(DefaultSurbs)
```

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -92,42 +199,64 @@ proc shortLog*(self: MixEntryConnection): string {.raises: [].} =
...
   instance.destination = destination
   instance.codec = codec
+
+  if expectReply:
+    instance.incoming = newAsyncQueue[seq[byte]]()
+    instance.replyReceivedFut = newFuture[void]()
+    let checkForIncoming = proc(): Future[void] {.async: (raises: [CancelledError]).} =
+      instance.cached = await instance.incoming.get()
+      instance.replyReceivedFut.complete()
+    instance.incomingFut = checkForIncoming()
```
gmelodie, line 110: nit: Group this if statement with the one above

libp2p/protocols/mix/exit_layer.nim
```diff
@@ -1,31 +1,93 @@
...
+  )
+
+proc replyDialerCbFactory(self: ExitLayer): MixReplyDialer =
+  return proc(
+      surbs: seq[SURB], msg: seq[byte]
+  ): Future[void] {.async: (raises: [CancelledError, LPStreamError]).} =
+    try:
+      var respFuts: seq[Future[void]] = @[]
+      for surb in surbs:
+        respFuts.add(self.onReplyDialer(surb, msg))
```
gmelodie: Could this be a `mapIt`? e.g.
```suggestion
      var respFuts: seq[Future[void]] = subs.mapIt(self.onReplyDialer(it, msg))
```

libp2p/protocols/mix/exit_layer.nim
```diff
@@ -1,31 +1,93 @@
...
+      await allFutures(respFuts)
+    except CancelledError as e:
+      raise e
+    except CatchableError as e:
+      error "Error during execution of reply: ", err = e.msg
```
gmelodie:
```suggestion
    except CancelledError as exc:
      raise exc
    except CatchableError as exc:
      error "Error during execution of reply: ", err = exc.msg
```

libp2p/protocols/mix/exit_layer.nim
```diff
@@ -1,31 +1,93 @@
...
+    if surbs.len != 0:
+      if not self.destReadBehavior.hasKey(codec):
+        error "No destReadBehavior for codec", codec
+        return
+
+      var behaviorCb: destReadBehaviorCb
+      try:
+        behaviorCb = self.destReadBehavior[codec]
+      except KeyError:
+        doAssert false, "checked with HasKey"
```
gmelodie, line 80: Can you not just to the `try` and put the `error "No destReadBehavior for codec"` inside the `except`?

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -24,6 +34,17 @@ type MixProtocol* = ref object of LPProtocol
   tagManager: TagManager
   exitLayer: ExitLayer
   rng: ref HmacDrbgContext
+  # TODO: verify if this requires cleanup for cases in which response never arrives (and connection is closed)
+  connCreds: Table[SURBIdentifier, ConnCreds]
+  destReadBehavior: TableRef[string, destReadBehaviorCb]
+
+proc hasDestReadBehavior*(mixProto: MixProtocol, codec: string): bool =
+  return mixProto.destReadBehavior.hasKey(codec)
```
gmelodie, line 42: Just do `hasKey` directly
```suggestion
```

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -24,6 +34,17 @@ type MixProtocol* = ref object of LPProtocol
...
+
+proc registerDestReadBehavior*(
+    mixProto: MixProtocol, codec: string, fwdBehavior: destReadBehaviorCb
+) =
+  mixProto.destReadBehavior[codec] = fwdBehavior
```
gmelodie, line 47: Just assign directly
```suggestion
```

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -110,13 +136,48 @@ proc handleMixNodeConnection(
...
+        return
+
+      let deserialized = MixMessage.deserialize(unpaddedMsg).valueOr:
+        error "Deserialization failed", err = error
+        mix_messages_error.inc(labelValues = ["Reply", "INVALID_SPHINX"])
+        return
+
+      await connCred.incoming.put(deserialized.message)
+    except KeyError:
+      doAssert false, "checked with hasKey"
```
gmelodie, line 154: We can remove all this indentation if we do sth like

```nim
let connCred =
  try:
    mixProto.connCreds[processedSP.id]
  except:
    mix_messages_error.inc(labelValues = ["Sender/Reply", "NO_CONN_FOUND"])
    return
```

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -156,39 +217,141 @@ proc getMaxMessageSizeForCodec*(
...
   ## Computes the maximum payload size (in bytes) available for a message when encoded  
-  ## with the given `codec` 
-  ## Returns an error if the codec length would cause it to exceeds the data capacity.  
+  ## with the given `codec`, optionally including space for the chosen number of surbs.  
+  ## Returns an error if the codec + surb overhead exceeds the data capacity.  
   let serializedMsg = MixMessage.init(@[], codec).serialize()
-  if serializedMsg.len > DataSize:
+  var totalLen = serializedMsg.len + SurbLenSize + (int(numberOfSurbs) * SurbSize)
+  if numberOfSurbs > 0:
+    totalLen += SurbIdLen
```
gmelodie: Should this not be
```suggestion
    totalLen += SurbIdLen*numberOfSurbs
```
?

If so, we could do
```nim
  let totalLen = serializedMsg.len + SurbLenSize + (int(numberOfSurbs) * SurbSize) + SurbIdLen*numberOfSurbs
```
directly

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -156,39 +217,141 @@ proc getMaxMessageSizeForCodec*(
...
+    mixProto: MixProtocol,
+    incoming: AsyncQueue[seq[byte]],
+    numSurbs: uint8,
+    destPeerId: PeerId,
+    exitPeerId: PeerId,
+): Result[seq[SURB], string] =
+  var response: seq[SURB]
+  var igroup = SURBIdentifierGroup(members: initHashSet[SURBIdentifier]())
+
+  for _ in 0.uint8 ..< numSurbs:
```
gmelodie, line 291: Having all function inside one for loop is cumbersome, let's break this down. Maybe the code inside the for loop could be moved to another function `buildSurb` that builds a single surb

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -156,39 +217,141 @@ proc getMaxMessageSizeForCodec*(
...
+    var
+      id: SURBIdentifier
+      multiAddrs: seq[MultiAddress] = @[]
+      publicKeys: seq[FieldElement] = @[]
+      hops: seq[Hop] = @[]
+      delay: seq[seq[byte]] = @[]
```
gmelodie: Do we need to recreate these at every for iteration?

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -156,39 +217,141 @@ proc getMaxMessageSizeForCodec*(
...
+      return err("No. of public mix nodes less than path length")
+
+    # Remove exit and dest node from nodes to consider for surbs
+    var pubNodeInfoKeys =
+      mixProto.pubNodeInfo.keys.toSeq().filterIt(it != exitPeerId and it != destPeerId)
+    var availableIndices = toSeq(0 ..< pubNodeInfoKeys.len)
+
+    # Select L mix nodes at random
+    var i = 0
+    while i < PathLength:
```
gmelodie:
```suggestion
    for i in 0..< PathLength:
```

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -156,39 +217,141 @@ proc getMaxMessageSizeForCodec*(
...
+          # Compute delay
+          let delayMillisec = cryptoRandomInt(mixProto.rng, 3).valueOr:
+            mix_messages_error.inc(labelValues = ["Entry/SURB", "NON_RECOVERABLE"])
+            return err("failed to generate random number: " & error)
+          (mixPubInfo.peerId, mixPubInfo.multiAddr, mixPubInfo.mixPubKey, delayMillisec)
+        else:
+          (
+            mixProto.mixNodeInfo.peerId, mixProto.mixNodeInfo.multiAddr,
+            mixProto.mixNodeInfo.mixPubKey, 0,
+          ) # No delay
```
gmelodie: This should be a function, in the same function you'd also put this from above
```nim
var pubNodeInfoKeys =
      mixProto.pubNodeInfo.keys.toSeq().filterIt(it != exitPeerId and it != destPeerId)
    var availableIndices = toSeq(0 ..< pubNodeInfoKeys.len)
```

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -156,39 +217,141 @@ proc getMaxMessageSizeForCodec*(
...
+    numSurbs: uint8 = 0,
+    destPeerId: PeerId,
+    exitPeerId: PeerId,
+): Result[seq[byte], string] =
+  let surbs = mixProto.buildSurbs(incoming, numSurbs, destPeerId, exitPeerId).valueOr:
+    return err(error)
+
+  let serialized = ?serializeMessageWithSURBs(msg, surbs)
+
+  ok(serialized)
```
gmelodie:
```suggestion
  ok(?serializeMessageWithSURBs(msg, surbs))
```

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -156,39 +217,141 @@ proc getMaxMessageSizeForCodec*(
...
+  let sphinxPacketBytes = sphinxPacket.serialize()
 
   try:
     let nextHopConn =
-      await mixProto.switch.dial(firstMixPeerId, @[firstMixAddr], @[MixProtocolID])
+      await mixProto.switch.dial(peerId, @[multiAddress], @[MixProtocolID])
     defer:
       await nextHopConn.close()
-    await nextHopConn.writeLp(sphinxPacket)
+    await nextHopConn.writeLp(sphinxPacketBytes)
```
gmelodie:
```suggestion
try:
let nextHopConn =
      await mixProto.switch.dial(peerId, @[multiAddress], @[MixProtocolID])
defer:
await nextHopConn.close()
    await nextHopConn.writeLp(sphinxPacket.serialize())
```

tests/mix/testconn.nim
```diff
@@ -0,0 +1,155 @@
...
+proc createSwitch(
+    multiAddr: MultiAddress, libp2pPrivKey: Opt[SkPrivateKey] = Opt.none(SkPrivateKey)
+): Switch =
+  var b = SwitchBuilder
+    .new()
+    .withAddress(multiAddr)
+    .withRng(crypto.newRng())
+    .withYamux()
+    .withTcpTransport()
+    .withNoise()
```
gmelodie: I think we can use [`newStandardSwitchBuilder`](https://github.com/vacp2p/nim-libp2p/blob/5f6b8e86a5c294cab9ed23afcaf496c18057d079/libp2p/builders.nim#L440)

gmelodie (commented):
Keep'em coming!

libp2p/protocols/mix/entry_connection.nim
vladopajic: agree comment does not really add clearity

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -5,14 +5,27 @@ import ../../utils/sequninit
...
+## Holds configuration parameters used in mix protocol operations.  
+type MixParameters* = object
+  expectReply*: Opt[bool]
+  numSurbs*: Opt[uint8]
+
 type MixEntryConnection* = ref object of Connection
   destination: MixDestination
   codec: string
   mixDialer: MixDialer
+  params: Opt[MixParameters]
```
vladopajic: if all values of `MixParameters` are optional do we really need to have `params` to be optional as well?

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -26,32 +39,126 @@ chronicles.formatIt(MixEntryConnection):
...
-  # TODO: implement
-  raise newLPStreamEOFError()
+  if s.isEof:
+    raise newLPStreamEOFError()
+
+  try:
+    await s.replyReceivedFut
+    s.isEof = true
+    if s.cached.len == 0:
+      raise newLPStreamEOFError()
```
vladopajic, line 47: i dont get this code, why us ` s.isEof = true` ?
it am confused here might be worth having a comment here.

libp2p/protocols/mix/entry_connection.nim
vladopajic: not really...

but this patter repeats so it could be template

`bytesMoved = moveTo(fromWhere, whereTo, maxBytes)`
->
`toRead = moveTo(s.cached, pbytes, nbytes) `

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -26,32 +39,126 @@ chronicles.formatIt(MixEntryConnection):
...
+    trace "couldn't read all bytes, stream EOF", s, nbytes, read
+    # Re-readOnce to raise a more specific error than EOF
+    # Raise EOF if it doesn't raise anything(shouldn't happen)
+    discard await s.readOnce(addr pbuffer[read], nbytes - read)
+
+    raise newLPStreamEOFError()
+
+  if read < nbytes:
+    trace "couldn't read all bytes, incomplete data", s, nbytes, read
+    raise newLPStreamIncompleteError()
```
vladopajic: could you explain this? every `MixEntryConnection.readLp(size)` must read `size` bytes? 
we usually had this `size` as maximum bytes, but not necessarily exactly.

libp2p/protocols/mix/entry_connection.nim
```diff
@@ -26,32 +39,126 @@ chronicles.formatIt(MixEntryConnection):
...
+  var state = 0
+
+  while true:
+    var ch: char
+    await readExactly(s, addr ch, 1)
+
+    if sep[state] == ch:
+      inc(state)
+      if state == len(sep):
+        break
```
vladopajic: brake or return?

libp2p/protocols/mix/entry_connection.nim
vladopajic: i dont think this is true (maybe it was in some earlier nims) - that is we can end with raise

libp2p/protocols/mix/entry_connection.nim
vladopajic: although this reduces number of lines, i like nim approach here (as already implemented) where we have expression in assignment

vladopajic: this code would be better if that is specified in params, 

```
proc getExpectedReplay(MixParameters): bool =
   get propery or default value
   
proc numSurbs(MixParameters): int =
   get property or default value
```

tests/mix/testconn.nim
```diff
@@ -0,0 +1,155 @@
...
+          "should have initialized mix protocol"
+        )
+      mixProto.add(proto)
+      switches[index].mount(proto)
+
+    let destNode = createSwitch(MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet())
+    defer:
+      await destNode.stop()
+
+    let fut = newFuture[seq[byte]]()
```
vladopajic: `fut` should describe the event `messageRecivedFut` or whathever

tests/mix/testconn.nim
```diff
@@ -0,0 +1,155 @@
...
+        NoReplyProtocolCodec,
+      )
+      .expect("could not build connection")
+
+    let data = @[1.byte, 2, 3, 4, 5]
+    await conn.writeLp(data)
+    await conn.close()
+
+    let receivedAtServer = await fut
+    check data == receivedAtServer
```
vladopajic:
```suggestion
    check data == (await fut)
```

tests/mix/testconn.nim
```diff
@@ -0,0 +1,155 @@
...
+    nodes.add(switch)
+
+  return nodes
+
+const NoReplyProtocolCodec = "/test/1.0.0"
+
+type NoReplyProtocol* = ref object of LPProtocol
+
+proc newNoReplyProtocol*(
+    switch: Switch, futToComplete: Future[seq[byte]]
```
vladopajic: futToComplete -> receivedMessageFut

vladopajic (approved):
lgtm. but would be nicer if comments are addressed.

libp2p/protocols/mix/entry_connection.nim
richard-ramos: It's so the attributes breathe :rofl:

libp2p/protocols/mix/entry_connection.nim
richard-ramos: Ended up realizing that a lot of the code was not really necessary, so I removed it.

libp2p/protocols/mix/entry_connection.nim
richard-ramos: Ended up removing this whole method

libp2p/protocols/mix/entry_connection.nim
richard-ramos: Removed the method

libp2p/protocols/mix/entry_connection.nim
richard-ramos: :rofl:

libp2p/protocols/mix/mix_protocol.nim
richard-ramos: Hm. Dunno. This `hasDestReadBehavior` is called from MixEntryConnection. Having it know the internal details of MixProtocol (like its attributes) seems like an unnecesary coupling. Wdyt?

libp2p/protocols/mix/mix_protocol.nim
richard-ramos: Hm, . `registerDestReadBehavior` is meant to be used by the users of mix protocol. My doubt with removing this and letting them have direct attr access is that it leaks internal state and makes the object less of a black box. if we expose destReadBehavior directly, we’re binding users to an implementation detail that might change later. 
wdyt?

libp2p/protocols/mix/entry_connection.nim
richard-ramos: Good question. I realized this was copied from LPStream, and it's not really necessary to have it since code will be inherited anyway, so this code in particular no longer exists here

libp2p/protocols/mix/entry_connection.nim
richard-ramos: Code no longer exist in last commit

libp2p/protocols/mix/entry_connection.nim
richard-ramos: Code no longer exist in last commit

libp2p/protocols/mix/exit_layer.nim
richard-ramos: Checking for `hasKey` first is slightly more efficient than raising an exception if key does not exist

libp2p/protocols/mix/mix_protocol.nim
richard-ramos: Hm. Good catch. This is actually not needed. The SurbID is already contained in the SurbSize

libp2p/protocols/mix/mix_protocol.nim
richard-ramos: Yes. These are individual per SURB

libp2p/protocols/mix/mix_protocol.nim
richard-ramos: image

libp2p/protocols/mix/entry_connection.nim
richard-ramos: No longer there are multiple instances of this code :)

libp2p/protocols/mix/entry_connection.nim
richard-ramos: Fixed. This should happen only when cached is 0

libp2p/protocols/mix/mix_protocol.nim
gmelodie: No strong feelings about it, do as you think is best

libp2p/protocols/mix/mix_protocol.nim
gmelodie: Same as above, i trust your judgement

gmelodie (approved):

PR #1711, opened 2025-09-22
yamux: fix warning

arnetheduck:
`fail` must have an exception of the right type or it prints a warning, even in code that doesn't use yamux


arnetheduck:
nevermind, I see this is fixed already

PR #1714, opened 2025-09-23
add a CI item to ensure libp2p is installable via nimble

fryorcraken:


.github/workflows/test-nimble-install.yml
```diff
@@ -0,0 +1,49 @@
+name: Test Nimble Installation
```
vladopajic, line 1: could you give reasoning why it wouldn't be installable via nimble so we can understand reasoning behind this pr?

.github/workflows/test-nimble-install.yml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
```
vladopajic, line 7: daily ci job might be more suitable here

examples/nimble/example.nimble
```diff
@@ -0,0 +1,15 @@
...
+license       = "MIT"
+srcDir        = "src"
+bin           = @["example"]
+
+
+# Dependencies
+
+requires "stew"
+requires "chronos"
+requires "libp2p#5f6b8e86a5c294cab9ed23afcaf496c18057d079"
```
vladopajic, line 15: why not using latest libp2p?

examples/nimble/example.nimble
```diff
@@ -0,0 +1,15 @@
+# Package
```
vladopajic, line 1: - this probably needs better naming than `examples/nimble`. 
- `nimble` dir might not be example because it it's utility is testing
- why can we reuse some code from current examples or why can't this check be indirectly checked when building examples?

.github/workflows/test-nimble-install.yml
fryorcraken: I was not able to install via waku and some of the error came from dependency conflicts within libp2p. I was testing it out and I was having the same issue when installing libp2p via nimble.

The lesson here is that there are still issues with nimble, and vac/nim team will take over so I can close this for now.

examples/nimble/example.nimble
fryorcraken: Because at the time of this PR this was the commit I needed :)

PR #1718, opened 2025-09-25
test(performance): chart generation refactor

rlve:
This PR intends to simplify chart generation for Performance tests data: Docker Resources and Latency history. We can utilise a [feature](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams) of GitHub Markdown that supports rendering of `Marmaid` charts from markdown. This way we remove a need to render and store the charts on our own.

Changes:
- Cleanup:
  - Remove GitHub actions:
    - `generate_plots` 
    - `publish_plots`
  - Remove scripts:
    - `add_plots_to_summary.nim‎`
    - `plot_docker_stats.py`
    - `plot_latency_history.py`
- Add scripts to prepare `Mermaid` markdown:
  - `docker_stats_charts.nim`
  - `latency_history_chart.nim`
  - `common.nim` - shared utility functions
- Refactor:
  - existing scripts to process docker and latency data
  - types
  - GitHub workflow


performance/scripts/common.nim
```diff
@@ -0,0 +1,227 @@
...
+  if not dirExists(dir):
+    raiseAssert "Directory not found: " & dir
+
+  var files: seq[string] = @[]
+  for entry in walkDir(dir):
+    if entry.kind == pcFile:
+      for pattern in patterns:
+        if pattern in entry.path or entry.path.endsWith(pattern):
+          files.add(entry.path)
+          break
```
gmelodie: nit: i think we could simplify this using `filterIt` and `anyIt`

performance/scripts/common.nim
```diff
@@ -0,0 +1,227 @@
...
+  return allLogFiles.filterIt(prefix in it)
+
+proc parseJsonFiles*(outputDir: string): seq[JsonNode] =
+  var jsons: seq[JsonNode]
+
+  for kind, path in walkDir(outputDir):
+    if kind == pcFile and path.endsWith(".json"):
+      let content = readFile(path)
+      let json = parseJson(content)
+      jsons.add(json)
```
gmelodie: nit: same as above

performance/scripts/common.nim
```diff
@@ -0,0 +1,227 @@
...
+) =
+  if toJobSummary:
+    let file = open(env.stepSummary, fmAppend)
+    file.write(content & "\n")
+    file.close()
+
+  if toComment:
+    let file = open(env.commentSummaryPath, fmAppend)
+    file.write(content & "\n")
+    file.close()
```
gmelodie: If `not` `toJobSummary` and `not` `toComment` then this function does nothing, seems odd.

performance/scripts/common.nim
```diff
@@ -0,0 +1,227 @@
...
+
+proc extractPrNumber*(filename: string): int =
+  let base = filename.extractFilename
+  if base.startsWith("pr") and base.endsWith("_latency.csv"):
```
gmelodie: Maybe invert this `if` statement to return early?
```suggestion
  if not base.startsWith("pr") or not base.endsWith("_latency.csv"):
    return 0
  # continue code
```

performance/scripts/common.nim
```diff
@@ -0,0 +1,227 @@
...
+): string =
+  if series.len == 0:
+    return ""
+
+  let values = series
+    .mapIt(
+      "    line \"" & it[0] & "\" [" &
+        it[1].mapIt(formatFloat(it, ffDecimal, precision)).join(", ") & "]"
+    )
+    .join("\n")
```
gmelodie: Nit: i know i suggested using iterators above, but I think here it'd be more readable to not use them here in favor of a regular `for` loop

performance/scripts/docker_stats_charts.nim
```diff
@@ -0,0 +1,107 @@
...
+  ResourceChartData(
+    title: "Total Network Data (MB)", yAxis: "MB", chartType: ResourceChartType.NetTotal
+  ),
+]
+
+proc readCsv(path: string): seq[ResourceChartsSample] =
+  let lines = readFile(path).splitLines()
+  var data: seq[ResourceChartsSample]
+  for i, line in lines:
+    if i == 0 or line.len == 0:
```
gmelodie: we should skip every line that has `len == 0` AND the first line with non-zero `len` (not necessarily `i == 0`)

performance/scripts/latency_history_chart.nim
```diff
@@ -0,0 +1,105 @@
+import algorithm, os, sequtils, strutils
+import ../types
+import ./common
+
+proc readLatencyCsv(path: string): seq[LatencyChartData] =
+  let lines = readFile(path).splitLines()
+  if lines.len < 2:
+    echo "Warning: CSV appears empty: " & path
+    return @[]
```
gmelodie: This check seems a bit weak (if the file has a bunch of newlines, it'd fail)

gmelodie (approved):
All nitpicks. Even the CSV checks I think are completely optional since this is not supposed to be a CSV parsing lib, but we should consider creating a repo to house an actual CSV lib in Nim (most of the code is already here). @richard-ramos this ties back to https://github.com/vacp2p/nim-libp2p/issues/1726. Nim needs more and better tooling, and we should consider investing time/money into that, given that it's our biggest time waster currently

performance/scripts/common.nim
rlve: I left `for` loop, but simplified it.

performance/scripts/common.nim
rlve: Simplified it with `findJsonFiles`.

performance/scripts/common.nim
rlve: Yeah, it was hard to read. I refactored the whole proc.

performance/scripts/docker_stats_charts.nim
rlve: As you said, it's not supposed to be CSV parsing library, so I'm not trying to handle every case. I control creation of these CSV files and I expect 1st line to always contain column names, that's why it's skipped.

performance/scripts/latency_history_chart.nim
rlve: I'm gonna keep it as it is, I don't expect any empty lines in these CSVs, and if there are, I want it to fail.

performance/scripts/docker_stats_charts.nim
```diff
@@ -0,0 +1,107 @@
...
+    title: "Network Throughput (MB/s)",
+    yAxis: "MB/s",
+    chartType: ResourceChartType.NetThroughput,
+  ),
+  ResourceChartData(
+    title: "Total Network Data (MB)", yAxis: "MB", chartType: ResourceChartType.NetTotal
+  ),
+]
+
+proc readCsv(path: string): seq[ResourceChartsSample] =
```
richard-ramos: Could we use https://nim-lang.org/docs/parsecsv.html ?

rlve: Changed in https://github.com/vacp2p/nim-libp2p/pull/1718/commits/5bf8948dc560dba73d26b760a94f0324debfe1de

richard-ramos (approved):

PR #1719, opened 2025-09-25
feat: modify rendezvous so that it can be extended 

chaitanyaprem:
Modify rendezvous code (using generics) to make it extendable and to support  app specific peerRecord e.g in Waku we want to publish Waku capabilities as part of SignedPeerRecord in rendezvous.

Note that the [SignedEnvelopes](https://github.com/libp2p/specs/blob/377f05a/RFC/0002-signed-envelopes.md) spec allows for CustomPeerRecords to be shared via [Rendezvous](https://github.com/libp2p/specs/blob/master/rendezvous/README.md#the-protocol) . This PR updates rendezvous code to support the same as Waku plans to use rendezvous for capability discovery especially for mix protocol.

Reference [research issue](https://github.com/waku-org/research/issues/118)

Changes

- [x] Update to use Generics instead of PeerRecord
- [x] Update tests
- [x] Add tests to validate by passing a custom PeerRecord 

Thanks a lot @Ivansete-status for all the help in debugging and fixing errors.

Co-authored-by: @Ivansete-status

### DIFF
--- a/libp2p/protocols/kademlia/consts.nim
+++ b/libp2p/protocols/kademlia/consts.nim
@@ -2,12 +2,13 @@ import chronos
 
 const
   IdLength* = 32 # 256-bit IDs
-  DefaultReplic* = 20 ## replication parameter, aka `k` in the spec
-  alpha* = 10 # concurrency parameter
-  ttl* = 24.hours
-  maxBuckets* = 256
-  QuorumResponses* = 5 # number of GetValue responses needed to decide
+
+  DefaultMaxBuckets* = 256
   DefaultTimeout* = 5.seconds
-  GetValRetries* = 5
+  DefaultRetries* = 5
+  DefaultReplication* = 20 ## aka `k` in the spec
+  DefaultAlpha* = 10 # concurrency parameter
+  DefaultTTL* = 24.hours
+  DefaultQuorum* = 5 # number of GetValue responses needed to decide
 
 const KadCodec* = "/ipfs/kad/1.0.0"
--- a/libp2p/protocols/kademlia/kademlia.nim
+++ b/libp2p/protocols/kademlia/kademlia.nim
@@ -73,19 +73,50 @@ method select*(
 
   return ok(bestIdx)
 
+type KadDHTConfig* = ref object
+  validator*: EntryValidator
+  selector*: EntrySelector
+  timeout: chronos.Duration
+  retries: int
+  replication: int
+  alpha: int
+  ttl: chronos.Duration
+  quorum: int
+
+proc new*(
+    T: typedesc[KadDHTConfig],
+    validator: EntryValidator = DefaultEntryValidator(),
+    selector: EntrySelector = DefaultEntrySelector(),
+    timeout: chronos.Duration = DefaultTimeout,
+    retries: int = DefaultRetries,
+    replication: int = DefaultReplication,
+    alpha: int = DefaultAlpha,
+    ttl: chronos.Duration = DefaultTTL,
+    quorum: int = DefaultQuorum,
+): T {.raises: [].} =
+  KadDHTConfig(
+    validator: validator,
+    selector: selector,
+    timeout: timeout,
+    retries: retries,
+    replication: replication,
+    alpha: alpha,
+    ttl: ttl,
+    quorum: quorum,
+  )
+
 type KadDHT* = ref object of LPProtocol
   switch: Switch
   rng: ref HmacDrbgContext
   rtable*: RoutingTable
   maintenanceLoop: Future[void]
   dataTable*: LocalTable
-  entryValidator: EntryValidator
-  entrySelector: EntrySelector
+  config*: KadDHTConfig
 
 proc insert*(
     self: var LocalTable, key: Key, value: sink seq[byte], time: TimeStamp
 ) {.raises: [].} =
-  debug "Local table insertion", key, value
+  debug "Local table insertion", key = key, value = value
   self[key] = EntryRecord(value: value, time: time)
 
 proc get*(self: LocalTable, key: Key): Opt[EntryRecord] {.raises: [].} =
@@ -142,14 +173,14 @@ proc isBestValue(kad: KadDHT, key: Key, record: EntryRecord): bool =
   ## Always returns `true` if we don't have the value locally
 
   kad.dataTable.get(key).withValue(existing):
-    kad.entrySelector.select(key, @[record, existing]).withValue(selectedIdx):
+    kad.config.selector.select(key, @[record, existing]).withValue(selectedIdx):
       return selectedIdx == 0
 
   true
 
 proc checkConvergence(state: LookupState, me: PeerId): bool {.raises: [], gcsafe.} =
   let ready = state.activeQueries == 0
-  let noNew = selectAlphaPeers(state).filterIt(me != it).len == 0
+  let noNew = state.selectAlphaPeers().filterIt(me != it).len == 0
   return ready and noNew
 
 proc findNode*(
@@ -157,8 +188,11 @@ proc findNode*(
 ): Future[seq[PeerId]] {.async: (raises: [CancelledError]).} =
   ## Iteratively search for the k closest peers to a target ID.
 
-  var initialPeers = kad.rtable.findClosestPeers(targetId, DefaultReplic)
-  var state = LookupState.init(targetId, initialPeers, kad.rtable.hasher)
+  var initialPeers = kad.rtable.findClosestPeers(targetId, kad.config.replication)
+  var state = LookupState.init(
+    targetId, initialPeers, kad.config.alpha, kad.config.replication,
+    kad.rtable.config.hasher,
+  )
   var addrTable: Table[PeerId, seq[MultiAddress]]
   for p in initialPeers:
     addrTable[p] = kad.switch.peerStore[AddressBook][p]
@@ -203,7 +237,7 @@ proc findNode*(
             debug "Could not update shortlist", err = exc.msg
           # TODO: add TTL to peerstore, otherwise we can spam it with junk
         ,
-        kad.rtable.hasher,
+        kad.rtable.config.hasher,
       )
 
     for timedOut in timedOutPeers:
@@ -270,11 +304,11 @@ proc dispatchPutVal(
       msg = msg, reply = reply, conn = conn
 
 proc putValue*(
-    kad: KadDHT, key: Key, value: seq[byte], timeout: Opt[int]
+    kad: KadDHT, key: Key, value: seq[byte]
 ): Future[Result[void, string]] {.async: (raises: [CancelledError]), gcsafe.} =
   let record = EntryRecord(value: value, time: $times.now().utc)
 
-  if not kad.entryValidator.isValid(key, record):
+  if not kad.config.validator.isValid(key, record):
     return err("invalid key/value pair")
 
   if not kad.isBestValue(key, record):
@@ -287,7 +321,7 @@ proc putValue*(
 
   kad.dataTable.insert(key, value, $times.now().utc)
   try:
-    await rpcBatch.allFutures().wait(chronos.seconds(5))
+    await rpcBatch.allFutures().wait(kad.config.timeout)
   except AsyncTimeoutError:
     # Dispatch will timeout if any of the calls don't receive a response (which is normal)
     discard
@@ -324,20 +358,18 @@ proc dispatchGetVal(
   received[][peer] = EntryRecord(value: value, time: time)
 
 proc getValue*(
-    kad: KadDHT, key: Key, timeout: timer.Duration = DefaultTimeout
+    kad: KadDHT, key: Key
 ): Future[Result[EntryRecord, string]] {.async: (raises: [CancelledError]), gcsafe.} =
-  # TODO: use timeout, make it configurable
   var remainingPeers = await kad.findNode(key)
   var received = ReceivedTable(newTable[PeerId, EntryRecord]())
 
   # TODO: only query alpha handling (selectAlphaPeers)
-  # TODO: make retries configurable
   var curTry = 0
-  while received.len < QuorumResponses and remainingPeers.len > 0 and
-      curTry < GetValRetries:
+  while received.len < kad.config.quorum and remainingPeers.len > 0 and
+      curTry < kad.config.retries:
     let rpcBatch = remainingPeers.mapIt(kad.switch.dispatchGetVal(it, key, received))
     try:
-      await rpcBatch.allFutures().wait(chronos.seconds(5))
+      await rpcBatch.allFutures().wait(kad.config.timeout)
     except AsyncTimeoutError:
       # Dispatch will timeout if any of the calls don't receive a response (which is normal)
       discard
@@ -347,8 +379,8 @@ proc getValue*(
 
   let
     records = received.values().toSeq()
-    validRecords = records.filterIt(kad.entryValidator.isValid(key, it))
-    selectedIdx = kad.entrySelector.select(key, validRecords).valueOr:
+    validRecords = records.filterIt(kad.config.validator.isValid(key, it))
+    selectedIdx = kad.config.selector.select(key, validRecords).valueOr:
       return err("Could not select value")
     best = validRecords[selectedIdx]
 
@@ -426,7 +458,8 @@ proc handleFindNode(
   let targetId = PeerId.init(targetIdBytes).valueOr:
     error "FindNode message without valid key data", msg = msg, conn = conn
     return
-  let closerPeers = kad.rtable.findClosest(targetId.toKey(), DefaultReplic)
+  let closerPeers = kad.rtable
+    .findClosest(targetId.toKey(), kad.config.replication)
     # exclude the node requester because telling a peer about itself does not reduce the distance,
     .filterIt(it != conn.peerId.toKey())
 
@@ -489,7 +522,7 @@ proc handlePutValue(
       return
 
   # Value sanitisation done. Start insertion process
-  if not kad.entryValidator.isValid(key, entryRecord):
+  if not kad.config.validator.isValid(key, entryRecord):
     debug "Record is not valid", key, entryRecord
     return
 
@@ -509,18 +542,14 @@ proc handlePutValue(
 proc new*(
     T: typedesc[KadDHT],
     switch: Switch,
-    validator: EntryValidator = DefaultEntryValidator(),
-    entrySelector: EntrySelector = DefaultEntrySelector(),
+    config: KadDHTConfig = KadDHTConfig.new(),
     rng: ref HmacDrbgContext = newRng(),
 ): T {.raises: [].} =
-  var rtable = RoutingTable.new(switch.peerInfo.peerId.toKey(), Opt.none(XorDHasher))
-  let kad = T(
-    rng: rng,
-    switch: switch,
-    rtable: rtable,
-    entryValidator: validator,
-    entrySelector: entrySelector,
+  var rtable = RoutingTable.new(
+    switch.peerInfo.peerId.toKey(),
+    config = RoutingTableConfig.new(replication = config.replication),
   )
+  let kad = T(rng: rng, switch: switch, rtable: rtable, config: config)
 
   kad.codec = KadCodec
   kad.handler = proc(
@@ -558,14 +587,6 @@ proc new*(
         error "Unhandled kad-dht message type", msg = msg
         return
   return kad
-
-proc setSelector*(kad: KadDHT, selector: EntrySelector) =
-  doAssert(selector != nil)
-  kad.entrySelector = selector
-
-proc setValidator*(kad: KadDHT, validator: EntryValidator) =
-  doAssert(validator != nil)
-  kad.entryValidator = validator
 
 method start*(kad: KadDHT): Future[void] {.async: (raises: [CancelledError]).} =
   if kad.started:
--- a/libp2p/protocols/kademlia/lookupstate.nim
+++ b/libp2p/protocols/kademlia/lookupstate.nim
@@ -1,5 +1,4 @@
 import sequtils
-import ./consts
 import ./protobuf
 import ./xordistance
 import ./keys
@@ -20,7 +19,7 @@ type
     shortlist: seq[LookupNode] # current known closest node
     activeQueries*: int # how many queries in flight
     alpha: int # parallelism level
-    repliCount: int ## aka `k`: number of closest nodes to find
+    replication: int ## aka `k`: number of closest nodes to find
     done*: bool # has lookup converged
 
 proc alreadyInShortlist(state: LookupState, peer: Peer): bool =
@@ -82,6 +81,8 @@ proc init*(
     T: type LookupState,
     targetId: Key,
     initialPeers: seq[PeerId],
+    alpha: int,
+    replication: int,
     hasher: Opt[XorDHasher],
 ): T =
   var res = LookupState(
@@ -89,7 +90,7 @@ proc init*(
     shortlist: @[],
     activeQueries: 0,
     alpha: alpha,
-    repliCount: DefaultReplic,
+    replication: replication,
     done: false,
   )
   for p in initialPeers:
@@ -113,6 +114,6 @@ proc selectClosestK*(state: LookupState): seq[PeerId] =
   var res: seq[PeerId] = @[]
   for p in state.shortlist.filterIt(not it.failed):
     res.add(p.peerId)
-    if res.len >= state.repliCount:
+    if res.len >= state.replication:
       break
   return res
--- a/libp2p/protocols/kademlia/routingtable.nim
+++ b/libp2p/protocols/kademlia/routingtable.nim
@@ -13,6 +13,8 @@ import results
 logScope:
   topics = "kad-dht rtable"
 
+const NoneHasher = Opt.none(XorDHasher)
+
 type
   NodeEntry* = object
     nodeId*: Key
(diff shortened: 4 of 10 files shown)


libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -298,30 +303,33 @@ proc advertise*(
 
   await allFutures(futs)
 
-method advertise*(
-    rdv: RendezVous, ns: string, ttl: Duration = rdv.minDuration
+method advertise*[T](
```
jm-clius: Found this thread, indicating deprecation of generic `method`s, which may explain at least some of your headaches: https://forum.nim-lang.org/t/4921

chaitanyaprem: Thanks, but that did not solve the issue.

Also looks like eventhough they are deprecated, methods with generics seem to work.

libp2p/multicodec.nim
```diff
@@ -397,6 +397,7 @@ const MultiCodecList = [
   ("onion3", 0x01BD),
   ("p2p-circuit", 0x0122),
   ("libp2p-peer-record", 0x0301),
+  ("libp2p-custom-peer-record", 0x0505), #not in multicodec list
```
richard-ramos: For this, we should either use a code from the [private range](https://github.com/multiformats/multicodec/tree/master?tab=readme-ov-file#private-use-area) or open a PR to multicodec proposing the new multicodec

richard-ramos (commented):
If we use any of the operations that accept a generic `[T]`, and we don't set up a peerValidator before that handles this `T`, then peers are not gonna be validated successfully, since we would always be expecting a PeerRecord.

Ideally this type of misconfiguration should not be allowed, which makes me think that the direct usage of these functions should not be encouraged at all.

The same with using the rendezvous codec for custom peer records, since any node using the default codec would expect that a PeerRecord is used and not a custom PeerRecord.

My suggestion would be to extract the generic functionality into it's own `GenericRendezvous` type, and then override methods for your more specialized types, each one with its own custom validator and codec. Here's an example you can run in nim playground  https://play.nim-lang.org/#pasty=JhUDRTEV

```nim

type
  PeerRecordValidator* = proc(spr: seq[byte]): bool

type GenericRendezvous[E] = object
  codec*: string
  peerValidator: PeerRecordValidator

proc new[E](T: typedesc[GenericRendezvous[E]], codec: string, validator: PeerRecordValidator): T =
  GenericRendezvous[E](codec: codec, peerValidator: validator)

proc advertise[T](rdv: GenericRendezvous[T], ns: string, signedPeerRecord: seq[byte]) =
  echo "GenericRendezvous: advertising -- type:", T, " ns:", ns, " bytes: ", signedPeerRecord

proc request[T](rdv: GenericRendezvous[T], ns: string): seq[T] =
  echo "GenericRendezvous: request -- type:", T, " ns:", ns 
  @[]

  
type PeerRecord = object
type CustomType = object
  
type Rendezvous = GenericRendezvous[PeerRecord]
type CustomRendezvous = GenericRendezvous[CustomType]

let validatorRendezvous = proc(spr: seq[byte]): bool =
  return true

let validatorCustomRendezvous = proc(spr: seq[byte]): bool =
  return true

proc advertise(rdv: Rendezvous, ns: string) =
  echo "Calling advertise on Rendezvous"
  procCall GenericRendezvous(rdv).advertise(ns, @[1.byte, 2, 3])
  

let foo = PeerRecord()
let bar = CustomType()

let rdv = Rendezvous.new("/rendezvous/1.0.0", validatorRendezvous)
let customRdv = CustomRendezvous.new("/custom-rendezvous/1.0.0", validatorCustomRendezvous)

rdv.advertise("abc")
```

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -273,17 +281,36 @@ proc advertisePeer(
...
   if ns.len < MinimumNamespaceLen or ns.len > MaximumNamespaceLen:
     raise newException(AdvertiseError, "Invalid namespace")
 
   if ttl < rdv.minDuration or ttl > rdv.maxDuration:
     raise newException(AdvertiseError, "Invalid time to live: " & $ttl)
 
-  let sprBuff = rdv.switch.peerInfo.signedPeerRecord.encode().valueOr:
-    raise newException(AdvertiseError, "Wrong Signed Peer Record")
+  var sprBuff = signedPeerRecord
+  if signedPeerRecord.len == 0:
```
richard-ramos: Instead of checking for 0 length to know if a signedPeerRecord has been specified, define the parameter like this: `signedPeerRecord: Opt[seq[byte]]`

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -301,30 +328,34 @@ proc advertise*(
 method advertise*(
     rdv: RendezVous, ns: string, ttl: Duration = rdv.minDuration
 ) {.base, async: (raises: [CancelledError, AdvertiseError]).} =
-  await rdv.advertise(ns, ttl, rdv.peers)
+  await rdv.advertise(ns, ttl, rdv.peers, @[])
```
richard-ramos: Opt.none(seq[byte])

SionoiS:
Nim doesn't have any concept of generic bounds?

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -301,30 +328,34 @@ proc advertise*(
...
+    peersOpt: Opt[seq[PeerId]] = Opt.some(rdv.peers),
+): Future[seq[T]] {.async: (raises: [DiscoveryError, CancelledError]).} =
   var
-    s: Table[PeerId, (PeerRecord, Register)]
+    s: Table[PeerId, (T, Register)]
     limit: uint64
     d = Discover(ns: ns)
-
+  let l = lt.get()
+  let peers = peersOpt.get()
```
richard-ramos: This code can fail if someone passes a none for `lt` and `peersOpt`.  You can change the code so it looks like this (so a default is used): 
```suggestion
proc request*[T](
    rdv: RendezVous,
    ns: Opt[string],
    lt: Opt[int],
    peersOpt: Opt[seq[PeerId]],
): Future[seq[T]] {.async: (raises: [DiscoveryError, CancelledError]).} =
var
    s: Table[PeerId, (T, Register)]
limit: uint64
d = Discover(ns: ns)
  let l = lt.get(DiscoverLimit.int)
  let peers = peersOpt.get(rdv.peers)
```

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -342,7 +375,7 @@ proc request*(
       if ns.isSome():
         try:
           Opt.some(rdv.cookiesSaved[peer][ns.get()])
-        except KeyError, CatchableError:
+        except Exception:
```
richard-ramos: `Exception` is too generic. Can we be more specific on the type of exceptions being raised here?

richard-ramos:
> @SionoiS: Nim doesn't have any concept of generic bounds?

It has concepts, but they're marked as experimental https://nim-lang.org/docs/manual_experimental.html#concepts

libp2p/multicodec.nim
chaitanyaprem: ah, Thanks did not know there was a private range..makes sense to use that.

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Ok, compiler was complaining about KeyError being conflicting type and hence i think this was changed to Exception. 

cc @Ivansete-status

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: I was reading about whether to use Opt for sequences, AI replied that idiomatic way was to use seq and check for length :) . 
Will change this.

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Updated to switch back to KeyError and CatchableError.

libp2p/multicodec.nim
```diff
@@ -402,6 +402,7 @@ const MultiCodecList = [
   ("onion3", 0x01BD),
   ("p2p-circuit", 0x0122),
   ("libp2p-peer-record", 0x0301),
+  ("libp2p-custom-peer-record", 0x0302), #not in multicodec list
```
richard-ramos: Maybe 0x300000 so it is part of the private area? eventually we will replace this list with a csv from https://github.com/multiformats/multicodec/blob/master/table.csv and in that list 0x0302 is already reserved

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -159,40 +166,46 @@ proc save(
     )
     rdv.namespaces[nsSalted].add(rdv.registered.high)
   #    rdv.registerEvent.fire()
-  except KeyError as e:
+  except exceptions.KeyError as e:
```
richard-ramos, line 176: Is the `exceptions.` prefix required?

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -159,40 +166,46 @@ proc save(
...
-  let pr = checkPeerRecord(r.signedPeerRecord, conn.peerId)
-  if pr.isErr():
-    return conn.sendRegisterResponseError(InvalidSignedPeerRecord, pr.error())
+  try:
+    let pr = rdv.peerRecordValidator(
+      rdv.switch.peerInfo.signedPeerRecord.data, r.signedPeerRecord, conn.peerId
+    )
+    if pr.isErr():
+      return conn.sendRegisterResponseError(InvalidSignedPeerRecord, pr.error())
+  except Exception as e:
```
richard-ramos: Exception is too generic. What kind of exceptions are being emitted here?

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -159,40 +166,46 @@ proc save(
...
+proc unregister*[E](rdv: RendezVous[E], conn: Connection, u: Unregister) =
   trace "Received Unregister", peerId = conn.peerId, ns = u.ns
   let nsSalted = u.ns & rdv.salt
   try:
     for index in rdv.namespaces[nsSalted]:
       if rdv.registered[index].peerId == conn.peerId:
         rdv.registered[index].expiration = rdv.expiredDT
         libp2p_rendezvous_registered.dec()
-  except KeyError:
+  except exceptions.KeyError:
```
richard-ramos, line 208: Is this prefix required?

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -39,23 +40,26 @@ const
...
 type
   RendezVousError* = object of DiscoveryError
-  RegisteredData = object
+  RegisteredData* = object
     expiration*: Moment
     peerId*: PeerId
     data*: Register
 
-  RendezVous* = ref object of LPProtocol
+  RendezVous*[E] = ref object of LPProtocol
```
richard-ramos: Wdyt about renaming this to:
`GenericRendezvous` and add a type `Rendezvous = GenericRendezvous[PeerRecord]`... (or `AbstractRendezvous`. The reason why I ask is because by default you would expect `Rendezvous` to use a `PeerRecord` (which is what the spec says), while with the proposed change you're forcing the developer to by default make a decision over the type of record Rendezvous is going to use

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -273,17 +286,39 @@ proc advertisePeer(
...
     raise newException(AdvertiseError, "Invalid namespace")
 
   if ttl < rdv.minDuration or ttl > rdv.maxDuration:
     raise newException(AdvertiseError, "Invalid time to live: " & $ttl)
 
-  let sprBuff = rdv.switch.peerInfo.signedPeerRecord.encode().valueOr:
-    raise newException(AdvertiseError, "Wrong Signed Peer Record")
+  var sprBuff = signedPeerRecord.valueOr:
+    rdv.switch.peerInfo.signedPeerRecord.encode().valueOr:
+      raise newException(AdvertiseError, "Wrong Signed Peer Record")
```
richard-ramos: This can probably cause an issue if you have an `E` that has a different type of record, and pass a none to `signedPeerRecord`, as it will attempt to encode a signed peer record, instead of encoding whatever `E` is.

Assuming you follow the `GenericRendezvous` suggestion, you could then have your own advertise for the specific `Rendezvous` implementation, which then will use a correct signed peer record
```nim
method advertise*(
    rdv: RendezVous, ns: string, ttl: Opt[Duration] = Opt.none(Duration)
) {.base, async: (raises: [CancelledError, AdvertiseError]).} =
  let lttl = ttl.get(rdv.minDuration)
  let sprBuff = rdv.switch.peerInfo.signedPeerRecord.encode().valueOr:
      raise newException(AdvertiseError, "Wrong Signed Peer Record")
  await rdv.advertise(ns, lttl, rdv.peers, sprBuff)
```

Then on: `proc advertise*[E](rdv: RendezVous[E], ns: string,  ttl: Duration,  peers: seq[PeerId],   signedPeerRecord: Opt[seq[byte]],` you would make the `signedPeerRecord` parameter not be optional, and just instead publish whatever you pass to it, i.e.:

```nim

proc advertise*[E](
    rdv: RendezVous[E],
    ns: string,
    ttl: Duration,
    peers: seq[PeerId],
    sprBuff:  seq[byte],
) {.async: (raises: [CancelledError, AdvertiseError]).} =
  if ns.len < MinimumNamespaceLen or ns.len > MaximumNamespaceLen:
    raise newException(AdvertiseError, "Invalid namespace")

  if ttl < rdv.minDuration or ttl > rdv.maxDuration:
    raise newException(AdvertiseError, "Invalid time to live: " & $ttl)

  let
    r = Register(ns: ns, signedPeerRecord: sprBuff, ttl: Opt.some(ttl.seconds.uint64))
    msg = encode(Message(msgType: MessageType.Register, register: Opt.some(r)))

  rdv.save(ns, rdv.switch.peerInfo.peerId, r)

  let futs = collect(newSeq()):
    for peer in peers:
      trace "Send Advertise", peerId = peer, ns
      rdv.advertisePeer(peer, msg.buffer).withTimeout(5.seconds)

  await allFutures(futs)
```
that way you don't get random surprises while using this function.

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -298,105 +333,88 @@ proc advertise*(
...
 
-proc request*(
-    rdv: RendezVous, ns: Opt[string], l: int = DiscoverLimit.int, peers: seq[PeerId]
-): Future[seq[PeerRecord]] {.async: (raises: [DiscoveryError, CancelledError]).} =
+proc requestPeer[E](
+    rdv: RendezVous[E], limit: uint64, ns: Opt[string], peer: PeerId
+): Future[seq[Register]] {.
+    async: (raises: [CancelledError, DialFailedError, LPStreamError])
+.} =
+  let conn = await rdv.switch.dial(peer, RendezVousCodec)
```
richard-ramos: Should this use the codec attr? so if you implement your custom rendezvous, it will use the codec from there?
```suggestion
  let conn = await rdv.switch.dial(peer, rdv.codec)
```

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -542,18 +577,63 @@ proc new*(
...
     maxDuration = MaximumDuration,
 ): T {.raises: [RendezVousError].} =
-  let rdv = T.new(rng, minDuration, maxDuration)
+  let rdv = T.new(rng, minDuration, maxDuration, checkPeerRecord)
   rdv.setup(switch)
   return rdv
 
-proc deletesRegister*(
-    rdv: RendezVous, interval = 1.minutes
+#[ proc new*[E](
```
richard-ramos: ✂️

libp2p/protocols/rendezvous/rendezvous.nim
gmelodie: nit: why `E` over `T` for the generic type name

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -39,23 +40,26 @@ const
...
   MaximumMessageLen = 1 shl 22 # 4MB
   MinimumNamespaceLen = 1
   MaximumNamespaceLen = 255
   RegistrationLimitPerPeer* = 1000
   DiscoverLimit = 1000'u64
-  SemaphoreDefaultSize = 5
+  SemaphoreDefaultSize* = 5
+
+type PeerRecordValidator*[E] =
+  proc(_: E, spr: seq[byte], peerId: PeerId): Result[void, string] {.gcsafe.}
```
gmelodie, line 53: Why do we need the first parameter `_: E`?
```suggestion
  proc(spr: seq[byte], peerId: PeerId): Result[void, string] {.gcsafe.}
```

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -77,16 +81,19 @@ type
     maxDuration: Duration
     minTTL: uint64
     maxTTL: uint64
+    peerRecordValidator: PeerRecordValidator[E]
 
-proc checkPeerRecord(spr: seq[byte], peerId: PeerId): Result[void, string] =
+proc checkPeerRecord*(
+    _: PeerRecord, spr: seq[byte], peerId: PeerId
```
gmelodie, line 90: Why do we need the first parameter `_: PeerRecord`?
```suggestion
    spr: seq[byte], peerId: PeerId
```

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -298,105 +333,88 @@ proc advertise*(
...
-    rdv: RendezVous, ns: string, ttl: Duration = rdv.minDuration
+method advertise*[E](
+    rdv: RendezVous[E], ns: string, ttl: Opt[Duration] = Opt.none(Duration)
 ) {.base, async: (raises: [CancelledError, AdvertiseError]).} =
-  await rdv.advertise(ns, ttl, rdv.peers)
+  let lttl = ttl.get(rdv.minDuration)
+  await rdv.advertise(ns, lttl, rdv.peers, Opt.none(seq[byte]))
 
-proc requestLocally*(rdv: RendezVous, ns: string): seq[PeerRecord] =
+proc requestLocally*[T](rdv: RendezVous[T], ns: string): seq[T] =
```
gmelodie: Why here the generics are called `T` and everywhere else `E`? If there's no change in semantics, let's make them all `T`

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -298,105 +333,88 @@ proc advertise*(
...
   try:
     collect(newSeq()):
       for index in rdv.namespaces[nsSalted]:
         if rdv.registered[index].expiration > n:
-          let res = SignedPeerRecord.decode(rdv.registered[index].data.signedPeerRecord).valueOr:
+          let res = SignedPayload[T].decode(rdv.registered[index].data.signedPeerRecord).valueOr:
             continue
           res.data
-  except KeyError as exc:
+  except Exception as exc:
```
gmelodie: Let's not use generic exceptions such as `Exception` or `CatchableError`

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -298,105 +333,88 @@ proc advertise*(
...
+): Future[seq[Register]] {.
+    async: (raises: [CancelledError, DialFailedError, LPStreamError])
+.} =
+  let conn = await rdv.switch.dial(peer, RendezVousCodec)
+  defer:
+    await conn.close()
+
+  var d = Discover()
+  d.ns = ns
+  d.limit = Opt.some(limit)
```
gmelodie: Can we do this instead?
```suggestion
  var d = Discover(ns: ns, limit: Opt.some(limit))
```

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -298,105 +333,88 @@ proc advertise*(
...
+  d.cookie =
+    if ns.isSome():
+      try:
+        Opt.some(rdv.cookiesSaved[peer][ns.get()])
+      except exceptions.KeyError, CatchableError:
```
gmelodie: We should not use `CatchableError`, list the possible specific errors instead

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -298,105 +333,88 @@ proc advertise*(
...
+    trace "Cannot discover", ns, status = resp.status, text = resp.text
+    return @[]
+  resp.cookie.withValue(cookie):
+    if ns.isSome:
+      let namespace = ns.get()
+      if cookie.len() < 1000 and
+          rdv.cookiesSaved.hasKeyOrPut(peer, {namespace: cookie}.toTable()):
+        try:
+          rdv.cookiesSaved[peer][namespace] = cookie
+        except Exception:
```
gmelodie: No `Exception`s

libp2p/protocols/rendezvous/rendezvous.nim
gmelodie: +1

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -405,7 +423,32 @@ proc request*(
...
+          continue
+        let
+          spr = SignedPayload[T].decode(r.signedPeerRecord).valueOr:
+            continue
+          pr = spr.data
+        if s.hasKey(pr.peerId):
+          let (prSaved, rSaved) =
+            try:
+              s[pr.peerId]
+            except Exception:
```
gmelodie:
```suggestion
            except KeyError:
```

libp2p/protocols/rendezvous/rendezvous.nim
gmelodie: This was not introduced in this PR, but maybe we could change this to not raise exceptions by returning a `Result` or having `MinimumDuration` and `MaximumDuration` be types that are evaluated upon their creation.

This would make many parts of this PR better by not needing to handle exceptions thrown by `new`

tests/discovery/testrendezvouserrors.nim
```diff
@@ -31,34 +31,52 @@ suite "RendezVous Errors":
     checkTrackers()
 
   asyncTest "Various local error":
-    let rdv = RendezVous.new(minDuration = 1.minutes, maxDuration = 72.hours)
+    let rdv = RendezVous[PeerRecord].new(
+      minDuration = 1.minutes,
+      maxDuration = 72.hours,
```
gmelodie:
```suggestion
      minDuration = MinimumAcceptedDuration,
      maxDuration = MaximumAcceptedDuration,
```

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Yea, otherwise compiler is complaining asking to prefix with either `exceptions` or `CryptoError` i think.

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Refer https://github.com/vacp2p/nim-libp2p/pull/1719#discussion_r2428091921

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Ah, i took help of github copilot in between and i think it changed it to E partially in some places because the [constructor](https://github.com/vacp2p/nim-libp2p/blob/c78f0a7ad2f65893d8dbfcaf03fa7341b67cf7e1/libp2p/protocols/rendezvous/rendezvous.nim#L579-L580)'s already take T for Type Rendezvous[PeerRecord]. I makes sense as this helps avoid confusion for associating T.

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Thanks, will fix these. As i mentioned above i took help of copilot for fixing some issues and these are remnants.

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Sure, i will revert this..i reverted this in other places but looks like i missed here. Thanks for catching it!

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Agreed, i found it odd that constructors raising exceptions. 
Not sure what the norm in nim is.

Rather than returning a result, how about we switch to defaults if someone passes values that are out of range with a warning log?

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Good catch, looks like i missed it. 
Wondering how the test worked then..this should have been caught in the test.

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Like the idea, let me try to change and see what ripple effects it causes.

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Thanks! 
Will modify to follow this.

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Actually none, removing this. i think this is remnant of something i was trying in between.

libp2p/protocols/rendezvous/rendezvous.nim
gmelodie: I like that too :)

libp2p/protocols/rendezvous/rendezvous.nim
chaitanyaprem: Made an attempt to implement as per your suggestion, please re-review.

chaitanyaprem:
> No idea why my vscode nph is screwing up formatting for `tutorial_6` file alone.
> My nph version is `v0.5.1-0-gde5cd48`.

any idea what could be causing the formatting issue in this file? 
are you guys facing the issue as well or is it something i need to fix in my setup? 
For now i have modified the file through vim and fixed the lint issues.

PR #1722, opened 2025-09-25
feat(mix): experimental - exit is destination

richard-ramos:


libp2p/protocols/mix/exit_connection.nim
```diff
@@ -0,0 +1,68 @@
...
+    responseFut.complete(0) # Nothing else to read.
+    return responseFut
+  if self.message.len < nbytes:
+    responseFut.fail(
+      newException(
+        LPStreamError, "Not enough data in to read exactly " & $nbytes & " bytes."
+      )
+    )
+    return responseFut
+  copyMem(pbytes, addr self.message[0], nbytes)
```
vladopajic, line 18: what if self.message.len < nbytes ?

libp2p/protocols/mix/exit_connection.nim
```diff
@@ -0,0 +1,68 @@
...
+    let fut = newFuture[void]()
+    fut.fail(
+      newException(LPStreamError, "exceeds max msg size of " & $DataSize & " bytes")
+    )
+    return fut
+
+  self.response.add(msg)
+  let fut = newFuture[void]()
+  fut.complete()
+  return fut
```
vladopajic: why you `raw`? code would be more simpler if it is not, here and other places in mix.

vladopajic (approved):

libp2p/protocols/mix/exit_layer.nim
```diff
@@ -48,19 +52,54 @@ proc reply(
...
+          except CatchableError as e:
+            error "Error during execution of MixProtocol handler: ", err = e.msg
+
+      if not hasHandler:
+        error "Handler doesn't exist", codec = codec
+        return
+
+      if surbs.len != 0:
+        let response = exitConn.getResponse()
+        await self.reply(surbs, response)
```
gmelodie: Should all of this be inside `defer`, or only
```nim
 if not exitConn.isNil:
        await exitConn.close()
```
?

libp2p/protocols/mix/exit_layer.nim
```diff
@@ -48,19 +52,54 @@ proc reply(
...
 ) {.async: (raises: [CancelledError]).} =
   # If dialing destination fails, no response is returned to
   # the sender, so, flow can just end here. Only log errors
   # for now
   # https://github.com/vacp2p/mix/issues/86
 
+  if destination == Hop():
+    error "no destination available"
+    mix_messages_error.inc(labelValues = ["Exit", "NO_DESTINATION"])
+    return
```
gmelodie, line 95: Did we not have a template for this pattern?

libp2p/protocols/mix/exit_connection.nim
richard-ramos: I just tried without raw and it works. I did try on diff PR and had to use raw there. Probably did something dumb back then... Will update the PR

libp2p/protocols/mix/exit_connection.nim
richard-ramos: Then the condition from lines 16-21 is raised

libp2p/protocols/mix/exit_layer.nim
richard-ramos: nope!

PR #1728, opened 2025-09-30
fix: testhelper incorrectly setting program result

gmelodie:
- Fix testhelper incorrectly setting program result
- Avoid testpkifilter from being called twice
- Include testpkifilter in `testnative`
- Fixed failing tests previously concealed (multiaddress, nameresolve)

### DIFF
--- a/libp2p/protocols/connectivity/relay/rtransport.nim
+++ b/libp2p/protocols/connectivity/relay/rtransport.nim
@@ -134,6 +134,8 @@ method handles*(self: RelayTransport, ma: MultiAddress): bool {.gcsafe.} =
   trace "Handles return", ma, result
 
 proc new*(T: typedesc[RelayTransport], cl: RelayClient, upgrader: Upgrade): T =
-  result = T(client: cl, upgrader: upgrader)
-  result.running = true
-  result.queue = newAsyncQueue[Connection](0)
+  let self = T(client: cl, upgrader: upgrader)
+  self.running = true
+  self.queue = newAsyncQueue[Connection](0)
+  procCall Transport(self).initialize()
+  self
--- a/libp2p/transports/memorytransport.nim
+++ b/libp2p/transports/memorytransport.nim
@@ -37,13 +37,15 @@ proc new*(
     upgrade: Upgrade = Upgrade(),
     rng: ref HmacDrbgContext = newRng(),
 ): T =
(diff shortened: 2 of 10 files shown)


tests/testmultiaddress.nim
```diff
@@ -125,10 +125,7 @@ const
 
   PathVectors = ["/unix/a/b/c/d/e/f/g/h/i.sock"]
 
-  PathExpects = [
-    "90030E2F746D702F703270642E736F636B",
```
richard-ramos, line 129: Why remove this value? it's been there for a while

gmelodie: Because it relates to go daemon. I forgot to remove it with go daemon

libp2p.nimble
```diff
@@ -31,9 +31,9 @@ proc runTest(filename: string, moreoptions: string = "") =
   if getEnv("CICOV").len > 0:
     excstr &= " --nimcache:nimcache/" & filename & "-" & $excstr.hash
   exec excstr &
-    " -r -d:libp2p_quic_support -d:libp2p_autotls_support -d:libp2p_mix_experimental_exit_is_dest tests/" &
+    " -r -d:libp2p_quic_support -d:libp2p_autotls_support -d:libp2p_mix_experimental_exit_is_dest -d:libp2p_gossipsub_1_4 tests/" &
     filename
-  rmFile "tests/" & filename.toExe
+  #rmFile "tests/" & filename.toExe
```
vladopajic: why is this commented? remove if it needs to be removed.

libp2p.nimble
```diff
@@ -56,14 +56,12 @@ task testpubsub, "Runs pubsub tests":
 
 task testfilter, "Run PKI filter test":
   runTest("testpkifilter")
-  runTest("testpkifilter", moreoptions = "-d:libp2p_pki_schemes=")
```
vladopajic, line 59: why removed?

libp2p.nimble
gmelodie: My bad, leftovers from testing

libp2p.nimble
gmelodie: As discussed with @richard-ramos, not specifying pki schemes will lead the test to use the default schemes, which is the same behavior of not specifying this flag at all (i.e. lines 58 and 59 do the same thing)

PR #1729, opened 2025-10-01
test(performance): reliability workflow

rlve:
Update the `Performance` GitHub workflow to run `Network Reliability` scenarios (TCP and QUIC under various network conditions) on a daily schedule.

Changes:
- Modify `Performance`  workflow:
    - Behaviour is unchanged when triggered by PRs.
    - Scheduled or manual runs execute Reliability scenarios - they do not track latency history and do not generate Docker resource charts.
- Update `Performance Summary` markdown
    - Sort scenarios alphabetically
    - Add an emoji indicating the test result – a test is considered failed if not all expected messages are received
- Misc fixes

### Example
https://github.com/vacp2p/nim-libp2p/actions/runs/18286814126

image

> [!NOTE]
> Not sure how much current test setup affects the results, but QUIC looks more unstable than TCP and has some issues while connecting to other peers under different conditions.

### DIFF
--- a/libp2p/protocols/kademlia/kademlia.nim
+++ b/libp2p/protocols/kademlia/kademlia.nim
@@ -113,68 +113,25 @@ proc waitRepliesOrTimeouts(
 
   return (receivedReplies, failedPeers)
 
-proc dispatchPutVal(
-    kad: KadDHT, peer: PeerId, key: Key, value: seq[byte]
-): Future[void] {.async: (raises: [CancelledError, DialFailedError, LPStreamError]).} =
-  let conn = await kad.switch.dial(peer, KadCodec)
-  defer:
-    await conn.close()
-  let msg = Message(
-    msgType: MessageType.putValue,
-    record: Opt.some(Record(key: Opt.some(key), value: Opt.some(value))),
-  )
-  await conn.writeLp(msg.encode().buffer)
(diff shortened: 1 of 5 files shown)


performance/scripts/process_latency_stats.nim
```diff
@@ -88,9 +89,16 @@ proc getMarkdownReport*(
...
 
-  for scenarioName, stats in results.pairs:
+  var sortedScenarios = toSeq(results.keys)
+  sortedScenarios.sort()
+
+  for scenarioName in sortedScenarios:
+    let stats = results[scenarioName]
     let nodes = validNodes[scenarioName]
-    output.add fmt"| {stats.scenarioName} | {nodes} | {stats.totalSent} | {stats.totalReceived} | {stats.latency.minLatencyMs:.3f} | {stats.latency.maxLatencyMs:.3f} | {stats.latency.avgLatencyMs:.3f} |"
+    let emoji =
```
gmelodie, line 98: Nit: i think we could get rid of this `emoji` variable here

gmelodie (approved):

performance/scripts/process_latency_stats.nim
rlve: Why? I think it's more readable this way.

Issue #1556, opened 2025-07-17
Use onRunning and onStop


gmelodie:
In `switch.nim`, use the `onRunning` and `onStop` `asyncEvent`s to unify `acceptFuts`, `startFuts` and the starting of `services` in only one future sequence. 

Note: this will be particularly important for `TcpTransport`, `WsTransport` and `AutotlsService`. The interdependencies between those must be sorted out when implementing this. Particularly, they should be in the following order
1. `TcpTransport.start` must finish
2. `TcpTransport.accept` must finish
3. `AutotlsService.setup` must finish
4. `WsTransport.start` must finish 
5. `WsTransport.accept` must finish

gmelodie:
https://github.com/vacp2p/nim-libp2p/issues/1558

Issue #1555, opened 2025-07-17
Create onStop


gmelodie:
Add an `asyncEvent` (`onStop`) to the base `Transport` type that is fired when `running` is set to `false` on method `stop`

gmelodie:
https://github.com/vacp2p/nim-libp2p/pull/1558

PR #1731, opened 2025-10-01
feat: add onRunning asyncEvent to Transport base type

gmelodie:
Add an asyncEvent that fires when transport becomes running during start(). This provides an elegant alternative to polling transport.running status.

Changes:
- Add onRunning field to Transport base type
- Add initTransport() method for proper asyncEvent initialization
- Update all transport constructors to call initTransport()
- Modify Transport.start() to fire onRunning when running = true
- Fix MemoryTransport to fire onRunning event in custom start method
- Remove redundant running assignment in QuicTransport
- Add comprehensive tests for TCP, Memory, QUIC, WebSocket, and Tor transports

credits: @MorganaFuture for implementing it

fixes https://github.com/vacp2p/nim-libp2p/issues/1554, https://github.com/vacp2p/nim-libp2p/issues/1555, and https://github.com/vacp2p/nim-libp2p/issues/1556


libp2p/transports/transport.nim
```diff
@@ -57,6 +62,9 @@ method stop*(self: Transport) {.base, async: (raises: []).} =
 
   trace "stopping transport", address = $self.addrs
   self.running = false
+  if self.onStop.isNil:
+    self.onStop = newAsyncEvent()
```
vladopajic: why isn't `self.onStop` initialized when creating transport?

libp2p/transports/transport.nim
```diff
@@ -49,6 +51,9 @@ method start*(
   trace "starting transport on addrs", address = $addrs
   self.addrs = addrs
   self.running = true
+  if self.onRunning.isNil:
+    self.onRunning = newAsyncEvent()
+  self.onRunning.fire()
```
vladopajic: why isn't `self.onRunning` initialized when creating transport?

libp2p/transports/transport.nim
gmelodie: Because there isn't a `new` or `setup` method defined for `Transport`

libp2p/transports/transport.nim
gmelodie: Same as above

libp2p/transports/transport.nim
vladopajic: would it be possible to create that method? and use that method to initialized these async events?

---

there is additional problem with this approach and that is if someone wants, for example `onRunning` event before starting transport, which can be reasonable usecase would result in nil pointer dereference.

this usecase we already have in auto tls - that is auto tls can be started before tcp transport. if one wants to refactor auto tls to start it after tcp transport is started this code would not be useful.

tests/testtortransport.nim
```diff
@@ -162,3 +167,125 @@ suite "Tor transport":
...
+          MultiAddress
+          .init(
+            "/ip4/127.0.0.1/tcp/8080/onion3/a2mncbqsbullu7thgm4e6zxda2xccmcgzmaq44oayhdtm6rav5vovcad:80"
+          )
+          .tryGet()
+        ],
+      )
+    finally:
+      await stubStartFut.cancelAndWait()
+      await stub.stop()
```
vladopajic: - minimize try to smallest logic where try is needed 
- also please explain why is try needed here?

tests/testtortransport.nim
```diff
@@ -162,3 +167,125 @@ suite "Tor transport":
...
+
+  check:
+    onRunningFired == true
+    onStopFired == true
+    transport.running == false
+
+suite "Transport Events":
+  asyncTest "onRunning with TCP transport":
+    let upgrader = Upgrade()
+    let transport = TcpTransport.new(upgrade = upgrader)
```
vladopajic: nit: inline upgrader since variable is not needed and line length of next line is almost the same.
 `let transport = TcpTransport.new(upgrade = Upgrade())`
 
 do this in other tests as well.

tests/testtortransport.nim
```diff
@@ -162,3 +167,125 @@ suite "Tor transport":
...
+
+proc testTransportEvents(transport: Transport, addrs: seq[MultiAddress]) {.async.} =
+  ## Common test procedure for transport events
+  var onRunningFired = false
+  var onStopFired = false
+
+  proc onRunningHandler() {.async.} =
+    # onRunning only gets set during start() call
+    if transport.onRunning.isNil:
+      transport.onRunning = newAsyncEvent()
```
vladopajic: do not initialize variables of transport here

tests/testtortransport.nim
```diff
@@ -162,3 +167,125 @@ suite "Tor transport":
...
+    await transport.onRunning.wait()
+    onRunningFired = true
+
+  proc onStopHandler() {.async.} =
+    # onStop only gets set during stop() call
+    if transport.onStop.isNil:
+      transport.onStop = newAsyncEvent()
```
vladopajic: do not initialize variables of transport here

tests/testtortransport.nim
```diff
@@ -162,3 +167,125 @@ suite "Tor transport":
     "/ip4/127.0.0.1/tcp/8080/onion3/a2mncbqsbullu7thgm4e6zxda2xccmcgzmaq44oayhdtm6rav5vovcad:80",
     "/ip4/127.0.0.1/tcp/8081/onion3/a2mncbqsbullu7thgm4e6zxda2xccmcgzmaq44oayhdtm6rav5vovcae:81",
   )
+
+proc testTransportEvents(transport: Transport, addrs: seq[MultiAddress]) {.async.} =
```
vladopajic: how does this method assert any invariant? what's utility of this method?

vladopajic (changes requested):
onRunning and onStop events should be initialized when transport is created.

libp2p/transports/transport.nim
gmelodie: Yes! As most transports have their own impl of `new`, I added an `init` method and called it in all `new` procs

tests/testtortransport.nim
gmelodie: it starts a transport and checks if `onRunning` was fired, then stops it and check if `onStop` was fired

tests/testtortransport.nim
gmelodie: I don't think `try` was needed, removed

tests/testtortransport.nim
```diff
@@ -18,9 +18,14 @@ import
...
     transports/tcptransport,
     transports/tortransport,
+    transports/transport,
+    transports/memorytransport,
+    transports/quictransport,
+    transports/wstransport,
     upgrademngrs/upgrade,
     multiaddress,
     builders,
+    crypto/crypto,
```
vladopajic: why does tor transport have these imports?

tests/commontransport.nim
```diff
@@ -232,3 +232,47 @@ template commonTransportTest*(prov: TransportBuilder, ma1: string, ma2: string =
...
+        await transport.onRunning.wait()
+        onRunningFired = true
+
+      proc onStopHandler() {.async.} =
+        # onStop only gets set during stop() call
+        await transport.onStop.wait()
+        onStopFired = true
+
+      asyncSpawn onRunningHandler()
+      asyncSpawn onStopHandler()
```
vladopajic: why is `asyncSpawn` needed, is it really needed?

tests/commontransport.nim
```diff
@@ -232,3 +232,47 @@ template commonTransportTest*(prov: TransportBuilder, ma1: string, ma2: string =
...
+
+      check:
+        onRunningFired == false
+        onStopFired == false
+        transport.running == false
+
+      await transport.start(addrs)
+
+      # Give the event handler time to run
+      await sleepAsync(10.milliseconds)
```
vladopajic: please avoid unnecessary `sleepAsync`

tests/commontransport.nim
```diff
@@ -232,3 +232,47 @@ template commonTransportTest*(prov: TransportBuilder, ma1: string, ma2: string =
...
+        await transport.onStop.wait()
+        onStopFired = true
+
+      asyncSpawn onRunningHandler()
+      asyncSpawn onStopHandler()
+
+      check:
+        onRunningFired == false
+        onStopFired == false
+        transport.running == false
```
vladopajic: hint: `check transport.onRunning.wait()`  is sufficient here

`check transport.onRunning.wait().wait(1.seconds)` should you want to bulletproof test

tests/commontransport.nim
```diff
@@ -232,3 +232,47 @@ template commonTransportTest*(prov: TransportBuilder, ma1: string, ma2: string =
       await handlerWait.wait(1.seconds) # when no issues will not wait that long!
 
       await transport1.stop()
+
+    asyncTest "transport start/stop events":
```
vladopajic, line 236: some transports don't use common transport, like quic.

gmelodie: Maybe this is a problem for another ~day~ PR? Quic transport should be using commontransport tests.

richard-ramos: Should we open an issue for this so we do not forget?

vladopajic: quic transport cant use `commonTransportTest` template, out of the box, because quic transport is transport+muxer, while other transports are just transports.

just copy paste the same 10lines to quic test.

libp2p/transports/transport.nim
```diff
@@ -36,10 +36,16 @@ type
     running*: bool
     upgrader*: Upgrade
     networkReachability*: NetworkReachability
+    onRunning*: AsyncEvent
+    onStop*: AsyncEvent
 
 proc newTransportClosedError*(parent: ref Exception = nil): ref TransportError =
   newException(TransportClosedError, "Transport closed, no more connections!", parent)
 
+proc init*(self: Transport) =
```
richard-ramos: Naming this `init` is kinda misleading as we have `init` for constructors that return objects, and `new` for those that return ref objects. Perhaps it should be called `initialize` ?

vladopajic (approved):

PR #1733, opened 2025-10-01
refactor(kad-dht): find node and put val

richard-ramos:
1. Make the record selector match what's suggested in the spec
2. Use `check` instead of `doAssert` in the tests
3. Remove data types that are not really needed
4. Use `Opt`  instead of `Option`
5. Change the logic to insert records only if the selector determines that the record inserted is new
6. Add default validator and selector
7. Make the `Key` datatype just be an alias of `seq[byte]`


vladopajic (approved):

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -89,10 +59,18 @@ type KadDHT* = ref object of LPProtocol
...
+  debug "local table insertion", key, value
+  self[key] = EntryRecord(value: value, time: time)
+
+proc get*(self: LocalTable, key: Key): Opt[EntryRecord] {.raises: [].} =
+  if not self.hasKey(key):
+    return Opt.none(EntryRecord)
+  try:
+    return Opt.some(self[key])
+  except KeyError:
+    doAssert false, "checked with hasKey"
```
gmelodie, line 73: I think we had this discussion before, but I'd rather:
```suggestion
  try:
    Opt.some(self[key])
  except KeyError:
    Opt.none(EntryRecord)
```

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -160,29 +138,24 @@ proc dispatchPutVal(
...
-    return err(error)
-  trace "local putval", candidate = candAsRec, others = others, selected = confirmedRec
-  let validEnt = ValidatedEntry.init(entKey, confirmedRec.value)
+  let existingRecord = kad.dataTable.get(key)
+  existingRecord.withValue(existing):
+    let selectedIdx = kad.entrySelector.select(key, @[value, existing.value]).valueOr:
+      error "application provided selector error (local)", msg = error
+      return err(error)
+    if selectedIdx != 0:
+      return err("can't replace a newer value with an older value")
```
gmelodie: Maybe this could be a function?

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -395,36 +360,27 @@ proc new*(
...
-
-        # Assume that if selection goes with another value, that it is valid
-        let validated = ValidatedEntry(key: key, value: selectedRec.value)
+        let existingRecord = kad.dataTable.get(key)
+        existingRecord.withValue(existing):
+          let selectedIdx = kad.entrySelector.select(key, @[value, existing.value]).valueOr:
+            error "application provided selector error", msg = error
+            return
+          if selectedIdx != 0:
+            error "can't replace a newer value with an older value"
```
gmelodie: If we turn above into a function, reuse it here

tests/kademlia/testfindnode.nim
```diff
@@ -76,10 +70,7 @@ suite "KadDHT - FindNode":
         if k.rtable.selfId == id:
           continue
         let count = countBucketEntries(k.rtable.buckets, id)
-        doAssert(
-          count == 1,
-          fmt"findNode post-check broken - entries: {entries}|id: {id}|buckets: {k.rtable.buckets}",
-        )
+        check count == 1
```
gmelodie:
```suggestion
        countBucketEntries(k.rtable.buckets, id) == 1
```

tests/kademlia/utils.nim
```diff
@@ -1,27 +1,29 @@
...
+  if values.len == 0:
+    return err("no values were given")
   return
-    if others.len == 0:
-      ok(cand)
+    if values.len == 1:
+      ok(0)
     else:
-      ok(others[0])
+      ok(1)
```
gmelodie:
```suggestion
  if values.len == 1:
    return ok(0)
  ok(1)
```

gmelodie (approved):

PR #1735, opened 2025-10-02
test(performance): Refactor runner

rlve:
### Summary

Refactors performance and reliability test orchestration. We move from a single “all scenarios in one image + shell runner” setup to a base image plus scenario‑driven orchestration. Each scenario now starts and cleans up its own fresh containers, picks the transport (TCP or QUIC) at runtime, and can inject pre/post commands. This makes tests more isolated, easier to iterate on, and less flaky.

### Before

1. Built one Docker image containing every scenario.
2. A shell script runner launched 10 containers, streamed logs, and handled cleanup.
3. Scenarios ran sequentially inside reused containers.
4. Hidden/shared state could leak between tests.

### After

1. Build a base test image.
2. Scenarios live outside the image and orchestrate their own containers.
3. Transport (TCP or QUIC) is chosen at runtime.
4. Each scenario gets fresh containers, no cross-test state.
5. Entrypoint supports pre- and post-test hooks for setup/teardown actions.

### New Directory Structure

```
performance/
├── runner.nim                    # Orchestrates Docker containers, networking, log streaming
├── scenarios_performance.nim     # Performance test suite
├── scenarios_reliability.nim     # Network reliability test suite
├── types.nim                     # Shared type definitions (stats, charts, CI config)
│
├── test_node/                    # Distributed test node implementation
│   ├── main.nim                  # Test entry point
│   ├── test_node.nim             # Core scenario logic
│   ├── utils.nim                 # Test utilities
│   ├── Dockerfile
│   └── entrypoint.sh             # Supports pre/post test hooks
│
├── scripts/                      # Post-test data processing
│   ├── common.nim                # Shared utilities for chart generation
│   ├── process_docker_stats.nim  # Parse Docker resource usage logs
│   ├── process_latency_stats.nim # Aggregate latency metrics
│   ├── docker_stats_charts.nim   # Generate resource usage charts
│   └── latency_history_chart.nim # Generate latency history charts
```


.github/workflows/performance.yml
```diff
@@ -53,15 +53,23 @@ jobs:
...
+        with:
+          nim-version: "2.x"
+          repo-token: ${{ env.GITHUB_TOKEN }}
+
       - name: Run
         run: |
-          ./performance/runner.sh
+          nimble install chronicles
+          nimble install chronos
+          nim c -r -d:release -o:/tmp/scenarios_performance -d:chronicles_log_level:INFO ./performance/scenarios_performance.nim
```
gmelodie: Maybe we could have a nimble target for this so we could do `nimble performance` instead of manually installing the dependencies here

performance/base_test/Dockerfile
```diff
@@ -0,0 +1,35 @@
+# Create the build image
+FROM nimlang/nim:2.2.4-alpine-regular AS build
+
+WORKDIR /node
+
+COPY libp2p.nimble config.nims ./
+RUN git config --global http.sslVerify false
```
gmelodie: Can you explain why this is needed? we should not be disabling https verifications

performance/base_test/main.nim
```diff
@@ -0,0 +1,11 @@
+import ./base_test
```
gmelodie: Could we not call `base_test.nim` directly?

performance/tests/scenarios_base.nim
```diff
@@ -9,14 +9,12 @@
...
 
-import ../../tests/helpers
-import ./base_test
-import ./utils
+import ../tests/utils/async_tests
+import ./runner
 
 suite "Performance Tests":
-  teardown:
-    checkTrackers()
```
gmelodie, line 18: Why are we removing `checkTrackers` here? I think we should always `checkTrackers`

gmelodie, line 18: We should `checkTrackers`

performance/scenarios_performance.nim
```diff
@@ -9,14 +9,12 @@
...
-import ./base_test
-import ./utils
+import ../tests/utils/async_tests
+import ./runner
 
 suite "Performance Tests":
-  teardown:
-    checkTrackers()
+  asyncTest "Base Test TCP":
+    run("TCP")
```
gmelodie, line 19: We should keep checking trackers

.github/workflows/performance.yml
rlve: Changed to use nimble to install dependencies and execute task.

performance/base_test/Dockerfile
rlve: Doesn't seem to be needed. Removed.

performance/base_test/main.nim
rlve: We could, but I find calling it via `main.nim` more clear.

performance/tests/scenarios_base.nim
rlve: In this context I don't have an access to `nim-libp2p` dependencies and I can't import `checkTrackers` without installing whole bunch of packages that are not needed here.

rlve: Moved it to `performance/test_node/main.nim` though.

rlve: As above.

performance/scenarios_performance.nim
rlve: As above.

performance/test_node/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+WORKDIR /node
+
+COPY libp2p.nimble config.nims ./
+RUN apk add --no-cache openssl-dev pkgconfig
+RUN nimble install -y
+
+COPY . .
+RUN nimble c \
+    -d:chronicles_colors=None \
```
richard-ramos, line 13: We use `--mm:refc` as the memory management strategy, so consider adding that flag wherever you compile a nim program

performance/performance.nimble
```diff
@@ -0,0 +1,13 @@
+# Package
+
+version = "0.1.0"
+author = "Status Research & Development GmbH"
+description = "A new awesome nimble package"
```
richard-ramos: Pls add a small description of what is this program about

performance/performance.nimble
```diff
@@ -0,0 +1,13 @@
...
+license = "MIT"
+
+# Dependencies
+
+requires "nim >= 2.2.4", "chronicles", "chronos", "unittest2"
+
+task performance, "Run performance tests":
+  exec "nim c -r -d:release -o:/tmp/scenarios_performance -d:chronicles_log_level:DEBUG ./scenarios_performance.nim"
```
richard-ramos: Same comment as before wrt --mm:refc

performance/runner.nim
```diff
@@ -0,0 +1,113 @@
...
+
+proc setupDockerNetwork*(network: string) =
+  let inspectOutput = execShellCommand(fmt"docker network inspect {network}")
+  if "Error" in inspectOutput:
+    echo execShellCommand(
+      fmt"docker network create --attachable --driver bridge {network} > /dev/null"
+    )
+
+proc removeDockerNetwork*(network: string) =
+  echo execShellCommand(fmt"docker network rm {network} > /dev/null")
```
richard-ramos: Since you're outputting to > /dev/null probably the `echo` can be replaced by a `discard`, right?

performance/runner.nim
```diff
@@ -0,0 +1,113 @@
...
+  envVars &= fmt" -e HOSTNAME_PREFIX={hostnamePrefix}"
+  envVars &= fmt" -e TRANSPORT_TYPE='{transportType}'"
+
+  if preExecCmd != "":
+    envVars &= fmt" -e PRE_EXEC_CMD='{preExecCmd}'"
+  if postExecCmd != "":
+    envVars &= fmt" -e POST_EXEC_CMD='{postExecCmd}'"
+
+  let containerId = execShellCommand(
+    fmt"""docker run -d \
```
richard-ramos, line 71: Could it be possible that a previous failing run of this task leaves a docker container running? I wonder if we should do a `docker rm -f {hostname}` before, or add a `--rm` flag so after the node stops, it automatically removes itself

performance/runner.nim
```diff
@@ -0,0 +1,113 @@
...
+
+proc run*(transportType: string, preExecCmd: string = "", postExecCmd: string = "") =
+  let outputDir = getOutputDir()
+
+  const network = "performance-test-network"
+  setupDockerNetwork(network)
+
+  var containerIds: seq[string]
+  try:
+    for i in 0 ..< 10:
```
richard-ramos: Can we make the 10 a constant?

performance/runner.nim
```diff
@@ -0,0 +1,113 @@
...
+      let hostname_prefix = "node-"
```
richard-ramos: maybe `hostnamePrefix` to maintain the camelCase style?

performance/runner.nim
```diff
@@ -0,0 +1,113 @@
...
+proc execShellCommand*(cmd: string): string =
+  try:
+    let output = execProcess(
+        "/bin/sh", args = ["-c", cmd], options = {poUsePath, poStdErrToStdOut}
+      )
+      .strip()
+    debug "Shell command executed", cmd, output
+    return output
+  except OSError as e:
+    raise newException(OSError, "Shell command failed")
```
richard-ramos:
```suggestion
    raise newException(OSError, "Shell command failed: " & getCurrentExceptionMsg())
```

richard-ramos (approved):
Left few comments and nitpicks

performance/runner.nim
rlve: `startContainer()` is wrapped in `try..catch` inside `run()` and `removeContainers()` is in `finally` block, so that situation should be covered.

PR #1736, opened 2025-10-02
fix(autonatv2): close dialback connection

richard-ramos:
We were only closing the stream but not the actual connection


vladopajic (approved):
`Stream` is always a stream… unless it’s a `Connection`. 
and `Connection` is always a connection… unless it’s a `Muxer`

gmelodie (approved):

PR #1742, opened 2025-10-06
feat: reset

richard-ramos:


libp2p/muxers/yamux/yamux.nim
```diff
@@ -256,6 +257,9 @@ proc reset(channel: YamuxChannel, isLocal: bool = false) {.async: (raises: []).}
     # If the reset is remote, there's no reason to flush anything.
     channel.recvWindow = 0
 
+method resetImpl*(channel: YamuxChannel) {.async: (raises: []).} =
+  channel.doReset(isLocal = true)
+
```
vladopajic, line 262: i wonder if `doReset` can become `resetImpl`, because `doReset` is actually `resetImpl` and having yet another `resetImpl` seems bit convoluted

richard-ramos: Hm. The main annoyance is that doReset has diff behaviors depending on the isLocal parameter.
While it's a bit convoluted it should be fine, since we want to not dedicate dev time to yamux anyway...

PR #1753, opened 2025-10-13
feat(mix): reuse streams

richard-ramos:
`vac:p2p:ift:2025q4-nimlibp2p-mix:stream-reuse`

### DIFF
--- a/libp2p/protocols/mix/mix_protocol.nim
+++ b/libp2p/protocols/mix/mix_protocol.nim
@@ -73,6 +73,23 @@ proc cryptoRandomInt(rng: ref HmacDrbgContext, max: int): Result[int, string] =
   let res = rng[].generate(uint64) mod uint64(max)
   ok(res.int)
 
+proc removeClosedConnections(
+    mixProto: MixProtocol, pid: PeerId
+) {.async: (raises: []).} =
+  var peersToGC = mixProto.connPool
+    .keys()
+    .toSeq()
+    .filterIt(not mixProto.switch.isConnected(it))
+    .toHashSet()
+  peersToGC.incl(pid)
+
+  for p in peersToGC:
+    try:
+      await mixProto.connPool[pid].close()
+      mixProto.connPool.del(pid)
+    except KeyError:
+      raiseAssert "checked with hasKey"
+
 proc getConn(
     mixProto: MixProtocol,
     pid: PeerId,
@@ -80,18 +97,15 @@ proc getConn(
     codecs: seq[string],
     forceNewStream: bool = false,
 ): Future[Connection] {.async: (raises: [DialFailedError, CancelledError]).} =
-  if mixProto.connPool.hasKey(pid):
-    try:
-      if forceNewStream:
-        await mixProto.connPool[pid].close()
-        mixProto.connPool.del(pid)
-      else:
-        return mixProto.connPool[pid]
-    except KeyError:
-      raiseAssert "checked with hasKey"
-  let c = await mixProto.switch.dial(pid, addrs, codecs)
-  mixProto.connPool[pid] = c
-  return c
+  if forceNewStream:
+    # GC all expired connections including the one used for `pid`
+    await mixProto.removeClosedConnections(pid)
+  try:
+    return mixProto.connPool[pid]
+  except KeyError:
+    let c = await mixProto.switch.dial(pid, addrs, codecs)
+    mixProto.connPool[pid] = c
+    return c
 
 proc writeLp(
     mixProto: MixProtocol,


libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -45,6 +45,7 @@ type MixProtocol* = ref object of LPProtocol
   # TODO: verify if this requires cleanup for cases in which response never arrives (and connection is closed)
   connCreds: Table[SURBIdentifier, ConnCreds]
   destReadBehavior: TableRef[string, destReadBehaviorCb]
+  sendPool: Table[PeerId, Connection]
```
vladopajic: `sendPool` -> `connPool`

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -72,21 +73,46 @@ proc cryptoRandomInt(rng: ref HmacDrbgContext, max: int): Result[int, string] =
...
+proc getConn(
+    mixProto: MixProtocol,
+    pid: PeerId,
+    addrs: seq[MultiAddress],
+    codecs: seq[string],
+    forceNewStream: bool = false,
+): Future[Connection] {.async: (raises: [DialFailedError, CancelledError]).} =
+  if mixProto.sendPool.hasKey(pid):
+    if forceNewStream:
+      mixProto.sendPool.del(pid)
```
vladopajic: close old conn here if new connections is forced?

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -72,21 +73,46 @@ proc cryptoRandomInt(rng: ref HmacDrbgContext, max: int): Result[int, string] =
...
+    else:
+      try:
+        return mixProto.sendPool[pid]
+      except KeyError:
+        raiseAssert "checked with hasKey"
+  let c = await mixProto.switch.dial(pid, addrs, codecs)
+  mixProto.sendPool[pid] = c
+  return c
```
gmelodie:
```suggestion
  if forceNewStream:
    mixProto.sendPool.del(pid)
  return mixProto.sendPool.getOrPut(pid, await mixProto.switch.dial(pid, addrs, codecs))
```

libp2p/protocols/mix/mix_protocol.nim
vladopajic: even if new conn is created because old got some error, it wouldn't bother to ensure closed connection.

vladopajic: also what happens to forced connections who closes them?

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -249,6 +271,21 @@ proc handleMixNodeConnection(
   of InvalidMAC:
     mix_messages_error.inc(labelValues = ["Intermediate/Exit", "INVALID_MAC"])
 
+proc handleMixNodeConnection(
+    mixProto: MixProtocol, conn: Connection
+) {.async: (raises: [LPStreamError, CancelledError]).} =
+  defer:
+    await conn.close()
```
vladopajic, line 293: if this connections got error and we created new connection. then old connection is closed here, but new connection will never be closed?

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -72,21 +73,46 @@ proc cryptoRandomInt(rng: ref HmacDrbgContext, max: int): Result[int, string] =
...
+    codecs: seq[string],
+    payload: seq[byte],
+) {.async: (raises: [DialFailedError, LPStreamError, CancelledError]).} =
+  let c = await mixProto.getConn(pid, addrs, codecs)
+  try:
+    await c.writeLp(payload)
+  except LPStreamClosedError, LPStreamResetError, LPStreamRemoteClosedError,
+      LPStreamConnDownError:
+    let c = await mixProto.getConn(pid, addrs, codecs, forceNewStream = true)
+    await c.writeLp(payload)
```
gmelodie, line 123: Is there not a check we can do in `c` to avoid calling `getConn` and `c.writeLp` twice? e.g.

```nim
var c = await mixProto.getConn(pid, addrs, codecs)
if c.isClosed() or c.isReset() or c.isRemoteClosed():
  await mixProto.getConn(pid, addrs, codecs, forceNewStream = true)
await c.writeLp(payload)
```

vladopajic: what if error happens while doing await `c.writeLp(payload)`... not before

vladopajic: this `getConn`  and `writeLp` in error branch do seem nasty, but it's correct way to do

vladopajic: alternatively, `for` loop can be used but probably wound's be as simple as current code. there are new variables, `i`, `maxAttempts`...

libp2p/protocols/mix/mix_protocol.nim
richard-ramos: The idea is to keep the connection open 'forever' (i.e. as long as peers are connected). Connection would be closed either when you lose connection to the peer or when you stop the switch.

libp2p/protocols/mix/mix_protocol.nim
gmelodie: makes sense

libp2p/protocols/mix/mix_protocol.nim
vladopajic: smaller improvement would be to remove connection if it was lost. but probably case where pool can grow with with lost connections is not probable , right?

libp2p/protocols/mix/mix_protocol.nim
vladopajic: @gmelodie  would `getOrPut` create new connection every time when `getOrPut` was called?

richard-ramos: Looking at sourcecode of getOrPut, it seems that the second parameter is not going to be lazily evaluated

libp2p/protocols/mix/mix_protocol.nim
richard-ramos: Hm. Maybe i could gc dropped connections anytime getConn is called? wdyt? 
```nim
  let peersToGC = mixProto.connPool.keys().toSeq().filterIt(not mixProto.switch.isConnected(it))
  for p in peersToGC:
    try:
      await mixProto.connPool[pid].close()
      mixProto.connPool.del(pid)
    except KeyError:
      raiseAssert "checked with hasKey"
```
(maybe could be even improved to check if N seconds have passed since last run)

libp2p/protocols/mix/mix_protocol.nim
richard-ramos: I think it's okay as this is on the receiving side. The new connection will invoke the handler of the receiving side, thus `handleMixNodeConnection will be executed again and be used for as long as possible.

libp2p/protocols/mix/mix_protocol.nim
gmelodie: ffs

libp2p/protocols/mix/mix_protocol.nim
vladopajic: >  Maybe i could gc dropped connections anytime getConn is called? wdyt?

yes, make sense, even more sense it makes to do this when `forceNewStream` because we assume that at least one connection is closed at that point, so it make sense to check and gc all.

```
  if forceNewStream:
     # will remove this connection and all other connections
     mixProto.removeClosedConnections()
  
  if mixProto.connPool.hasKey(pid):
        return mixProto.connPool[pid]

  let c = await mixProto.switch.dial(pid, addrs, codecs)
  mixProto.connPool[pid] = c
  return c
  ```

Comment (vladopajic) [reply libp2p/protocols/mix/mix_protocol.nim]: thinking about this:


```
  if mixProto.connPool.hasKey(pid):
      try:
         return mixProto.connPool[pid]
      except KeyError:
        raiseAssert "checked with if"

```
avoid `if` when we already must have `try/except`:

```
try:
   return mixProto.connPool[pid]
except KeyError:
  discard
```

full if/else:

```
try:
   return mixProto.connPool[pid]
except KeyError:
    let c = await mixProto.switch.dial(pid, addrs, codecs)
    mixProto.connPool[pid] = c
    return c
```

seems vary logical to reason this code? 

----

finally:

```
 if forceNewStream:
     # will remove peers connection and all other connections that
     mixProto.removeClosedConnections()
     
 try:
   return mixProto.connPool[pid]
except KeyError:
    let c = await mixProto.switch.dial(pid, addrs, codecs)
    mixProto.connPool[pid] = c
    return c

```

vladopajic: `mixProto.removeClosedConnections()` assumes current peers connection is not connected so it will be removed. but alternatively `mixProto.removeClosedConnections(pid)` we can pass peerId to ensure connection is removed if has this peer or is not connected...

PR #1754, opened 2025-10-14
feat(multiformats): Expose digest size from MHash

emizzle:
- Expose `digestSize` for a given `MHash`, eg useful externally to declare array sizes
- Change `digest*` signature from `hashcode (int)` to `mcodec (MultiCodec)`. This overload was not actually compiling and was not covered in tests, therefore no deprecation notice + overload needed.

### DIFF
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -126,6 +126,9 @@ proc getStream*(
   when defined(libp2p_agents_metrics):
     qs.shortAgent = session.shortAgent
 
+  # Inherit transportDir from parent session for GossipSub outbound peer tracking
+  qs.transportDir = session.transportDir
+
   session.streams.add(qs)
   return qs
 
@@ -303,7 +306,7 @@ method stop*(transport: QuicTransport) {.async: (raises: []).} =
   await procCall Transport(transport).stop()
 
 proc wrapConnection(
-    transport: QuicTransport, connection: QuicConnection
+    transport: QuicTransport, connection: QuicConnection, transportDir: Direction
 ): QuicSession {.raises: [TransportOsError].} =
   var observedAddr: MultiAddress
   var localAddr: MultiAddress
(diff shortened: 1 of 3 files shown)


libp2p/multihash.nim
```diff
@@ -396,6 +396,20 @@ proc digestImplWithoutHash(hash: MHash, data: openArray[byte]): MultiHash =
   result.data.writeArray(data)
   result.data.finish()
 
+func digestSize*(codec: MultiCodec): MhResult[int] =
+  let hash = CodeHashes.getOrDefault(codec)
+  if isNil(hash.coder):
+    err(ErrNotSupported)
+  else:
+    ok(hash.size)
```
gmelodie, line 404: nit: maybe this?
```suggestion
  let hash = CodeHashes.valueOr:
    return err(ErrNotSupported)
  ok(hash.size)
```

emizzle:
> I'm fine with this, but sth on the back of my mind makes me think this should be a simple field rather than a function

@gmelodie Yes, i agree. There is an odd pattern here where there are two similarly named types used with overlapping property names: 
`MHash`: used for registering a mulithash codec, digest size, and encoding procedure (at compile time)
`MultiHash`: contains hashed data, digest size, buffer, and buffer data offset

The digest size in these two types are not necessarily the same. `MultiHash` will take the `MHash.size` only if `MHash.size != 0`. If it is 0, then it will take the length of the data. Therefore, we'd need to expose `MHash.size` to the application. However, it is already exported, with no way for an application to lookup registered `MHashes` due to `HashesList` not being exported. This leaves us with two options:

1. **Lookup digest size.** This PR, in its current form, allows the multihash module to lookup the `MHash` by codec, and return the digest size.
2. **Export `HashesList`**. Let the application do the lookup. 

In Codex's use case, we need to be able to get the digest size from the registered multihash (`MHash`) since we need the value prior to hashing data, so either one of these options will work.

Thoughts?

libp2p/multihash.nim
emizzle: I don't think this works here without a Result or Option type being returned from the table lookup.

emizzle:
Just rebased on top of master, but should be g2g after green lights 👍

PR #1734, opened 2025-10-02
feat(multiformats): allow application-level extension of multiformats

emizzle:
Fixes: #1682.

Allows an application to extend the hardcoded and predefined multiformats with application-specific multiformats. It is best practice to upstream multiformats to the corresponding multiformats repo, [eg multicodec](https://github.com/multiformats/multicodec/blob/master/table.csv), but there are situations where this is not practical. In the case of Codex, as an example, we are currently using Poseidon hashing to build block merkle trees due to its suitability for zk proofs. However, Poseidon hashing will not be needed in the final version of Codex, since the proof system will be changing, yet we need to run the codebase in its current form for a prototype delivery. As such, the current state of hardcoded multiformats requires that we maintain a branch of nim-libp2p (and continually updating with upstream changes).  

This PR aims to solve that situation by allowing applications to extend multiformats. The following multiformats can be extended:

| Multi format   | Compiler flag<br>(path to extensions file) | Expected definition in extension file |
|----------------|--------------------------------------------|-------------------------------|
| `MultiCodec`   | `-d:libp2p_multicodec_exts`                | `const CodecExts = []`        |
| `MultiHash`    | `-d:libp2p_multihash_exts`                 | `const HashExts = []`         |
| `MultiAddress` | `-d:libp2p_multiaddress_exts`              | `const AddressExts = []`      |
| `MultiBase`    | `-d:libp2p_multibase_exts`                 | `const BaseExts = []`         |

For example, a file called `multihash_exts.nim` could be created and contain `MultiHash` extensions:
```nim
proc coder1(data: openArray[byte], output: var openArray[byte]) =
  copyMem(addr output[0], unsafeAddr data[0], len(output))

proc coder2(data: openArray[byte], output: var openArray[byte]) =
  copyMem(addr output[0], unsafeAddr data[0], len(output))

proc sha2_256_override(data: openArray[byte], output: var openArray[byte]) =
  copyMem(addr output[0], unsafeAddr data[0], len(output))

const HashExts = [
  MHash(mcodec: multiCodec("codec_mc1"), size: 0, coder: coder1),
  MHash(mcodec: multiCodec("codec_mc2"), size: 6, coder: coder2),
  MHash(mcodec: multiCodec("sha2-256"), size: 6, coder: sha2_256_override),
]
```
These `MultiHashes` will be available at compile time when the binary is compiled with `-d:libp2p_multihash_exts=multihash_exts.nim`.

Co-authored-by: Arnaud <arnaud@status.im>

PR #1755, opened 2025-10-14
feat(multiformats): add Cid extensions

emizzle:
Allows Cid codecs to be added as extensions at compile time, such that Cids can be hashed with extended codecs (defined as codec extensions, see https://github.com/vacp2p/nim-libp2p/pull/1734).


richard-ramos (approved):

emizzle:
Just rebased on top of master, so should be g2g after green lights 👍

PR #1760, opened 2025-10-14
chore(kad): configuration options

gmelodie:
Fixes #1758 and #1757

PR #1762, opened 2025-10-15
chore(style): add decision about return statements

gmelodie:
As per https://github.com/vacp2p/nim-libp2p/pull/1760#pullrequestreview-3337299447

### DIFF
--- a/libp2p/protocols/kademlia/consts.nim
+++ b/libp2p/protocols/kademlia/consts.nim
@@ -2,12 +2,13 @@ import chronos
 
 const
   IdLength* = 32 # 256-bit IDs
-  DefaultReplic* = 20 ## replication parameter, aka `k` in the spec
-  alpha* = 10 # concurrency parameter
-  ttl* = 24.hours
-  maxBuckets* = 256
-  QuorumResponses* = 5 # number of GetValue responses needed to decide
+
+  DefaultMaxBuckets* = 256
   DefaultTimeout* = 5.seconds
-  GetValRetries* = 5
+  DefaultRetries* = 5
+  DefaultReplication* = 20 ## aka `k` in the spec
+  DefaultAlpha* = 10 # concurrency parameter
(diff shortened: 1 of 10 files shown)


CODING_STYLE.md
```diff
@@ -10,3 +10,10 @@ try:
 except KeyError as e:
   # exception handling
 ```
+
+2. Do not use `return` at the end of function calls
+```nim
+proc myProc(someVar: bool): bool =
+  if someVar:
+    return true
+  false
```
vladopajic: - are there any benefits of not using `return`?
- since almost all modern languages use `return` keyword, use of `return` would:
  - makes the code more familiar and approachable to developers coming from other languages
  - this kind of code style would be naturally produced by all developers (by default people would use it, and code reviews that are enforcing this would be just annoying).
- some people rely on visual scanning rather than deep reading to understand code flow. `return` keywords provide clear visual anchors that signal where functions exit. 
- especially in nim that doesn't use `{}` for block delimitation, last return acts as a semantic boundary marker.

---

- for one-liner procs, or proc that only produce that value (one value), like constructors, appears like only candidate where `return` could be optional.

----

i am against this convention, but will follow it if necessary.

arnetheduck:
> The style guide should enforce principles and not prohibitions:

Pursuant to this point, PR:s to the github.com/status-im/nim-style-guide/ that reduce its verbiage and make it easier to follow are always welcome. 

One reason we use the style guide is to avoid pointless "style-change-crusades" - often, for readability, having a single consistent style is often more important than any one particular choice (ie return-or-not) - back in the day, we picked "expression return" as the preferred way in part because compiler error messages are better and in part to just have fewer ways of doing the same thing. As python says, "there should be one way, and preferable only one way" - this applies internally to libp2p but also more globally for any project IFT produces (ie to help with mobility across projects) so the fewer "exceptions" we have across projects, the better.

PR #1764, opened 2025-10-15
chore(kad): add alpha handling

### LINKED ISSUE #1759: Use alpha concurrency parameter in putVal and getVal
https://github.com/vacp2p/nim-libp2p/pull/1750#discussion_r2427392943

gmelodie:
Fixes #1759

### DIFF
--- a/libp2p/protocols/kademlia/kademlia.nim
+++ b/libp2p/protocols/kademlia/kademlia.nim
@@ -318,8 +318,7 @@ proc putValue*(
 
   kad.dataTable.insert(key, value, $times.now().utc)
 
-  var chunk: seq[PeerId]
-  forChunks(peers, kad.config.alpha, chunk):
+  for chunk in peers.toChunks(kad.config.alpha):
     let rpcBatch = chunk.mapIt(kad.switch.dispatchPutVal(it, key, value))
     try:
       await rpcBatch.allFutures().wait(kad.config.timeout)
@@ -365,10 +364,9 @@ proc getValue*(
   var received = ReceivedTable(newTable[PeerId, EntryRecord]())
 
   var curTry = 0
-  var chunk: seq[PeerId]
   while received.len < kad.config.quorum and remainingPeers.len > 0 and
       curTry < kad.config.retries:
-    forChunks(remainingPeers, kad.config.alpha, chunk):
+    for chunk in remainingPeers.toChunks(kad.config.alpha):
       let rpcBatch = chunk.mapIt(kad.switch.dispatchGetVal(it, key, received))
       try:
         await rpcBatch.allFutures().wait(kad.config.timeout)
--- a/libp2p/utility.nim
+++ b/libp2p/utility.nim
@@ -154,11 +154,16 @@ macro includeFile*(file: static[string]): untyped =
 
   return res
 
-template forChunks*[T](seqData: seq[T], chunkSize: int, chunk: untyped, body: untyped) =
-  ## Iterate over `seqData` in chunks of size `chunkSize`.
-  ## Each chunk is available as `chunk` inside the body.
+proc toChunks*[T](data: seq[T], size: int): seq[seq[T]] {.raises: [].} =
+  ## Splits `data` into chunks of length `size`.
+  ## The last chunk may be smaller if data.len is not divisible by size.
+  if size <= 0:
+    return @[]
+
+  var result: seq[seq[T]] = @[]
   var i = 0
-  while i < seqData.len:
-    let chunk = seqData[i ..< min(i + chunkSize, seqData.len)]
-    body
-    i += chunkSize
+  while i < data.len:
+    let endIndex = min(i + size, data.len)
+    result.add(data[i ..< endIndex])
+    i = endIndex
+  return result


libp2p/utility.nim
```diff
@@ -153,3 +153,12 @@ macro includeFile*(file: static[string]): untyped =
     raiseAssert("Failed to include file: " & file & ", error: " & e.msg)
 
   return res
+
+template forChunks*[T](seqData: seq[T], chunkSize: int, chunk: untyped, body: untyped) =
```
gmelodie: Chunk cannot be defined by the template and has to be passed as argument because if not the compiler complains that chunk variable does not exist when calling template (probably because of how compiler handles nim-chronos async procs)

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -363,16 +364,18 @@ proc getValue*(
...
-    except AsyncTimeoutError:
-      # Dispatch will timeout if any of the calls don't receive a response (which is normal)
-      discard
+    forChunks(remainingPeers, kad.config.alpha, chunk):
+      let rpcBatch = chunk.mapIt(kad.switch.dispatchGetVal(it, key, received))
+      try:
+        await rpcBatch.allFutures().wait(kad.config.timeout)
+      except AsyncTimeoutError:
+        # Dispatch will timeout if any of the calls don't receive a response (which is normal)
+        discard
```
vladopajic: what about:

```
for chunk in remainingPeers.toChunks(chunkSize):
    ....
```

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -363,16 +364,18 @@ proc getValue*(
...
   while received.len < kad.config.quorum and remainingPeers.len > 0 and
       curTry < kad.config.retries:
-    let rpcBatch = remainingPeers.mapIt(kad.switch.dispatchGetVal(it, key, received))
-    try:
-      await rpcBatch.allFutures().wait(kad.config.timeout)
-    except AsyncTimeoutError:
-      # Dispatch will timeout if any of the calls don't receive a response (which is normal)
-      discard
+    forChunks(remainingPeers, kad.config.alpha, chunk):
+      let rpcBatch = chunk.mapIt(kad.switch.dispatchGetVal(it, key, received))
```
vladopajic: why exactly is this done in chunks? why can we `dispatchGetVal` to all at once?

libp2p/protocols/kademlia/kademlia.nim
vladopajic: `proc toChunks(data: seq[T], size: int): seq[seq[T]]`

libp2p/protocols/kademlia/kademlia.nim
vladopajic: all peers will get this request. so why code wait before it makes request to other peers? all requests are handled asynchronously by different peers.

if we request form peer A, B, C, D, E

we can send request to all in parallel
`request A, B, C, D, E`

and get response in parallel:
`<- response A, B, C, D, E`

instead of 

`request A, B`
`<- response A, B`
`request C, D`
`<- response C, D`
`request E`
`<- response E`

libp2p/utility.nim
vladopajic: seems vary dirty with that variable, especially because it pollutes variables space outside closure where it is used

libp2p/protocols/kademlia/kademlia.nim
vladopajic: latency is significantly increased in later scenario, while it appears not to give any advantage.

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -316,15 +316,16 @@ proc putValue*(
...
   kad.dataTable.insert(key, value, $times.now().utc)
-  try:
-    await rpcBatch.allFutures().wait(kad.config.timeout)
-  except AsyncTimeoutError:
-    # Dispatch will timeout if any of the calls don't receive a response (which is normal)
-    discard
+
+  var chunk: seq[PeerId]
+  forChunks(peers, kad.config.alpha, chunk):
+    let rpcBatch = chunk.mapIt(kad.switch.dispatchPutVal(it, key, value))
```
vladopajic, line 322: subject to same comment as here: https://github.com/vacp2p/nim-libp2p/pull/1764/files#r2433013327

libp2p/protocols/kademlia/kademlia.nim
gmelodie: Per [the spec](https://github.com/libp2p/specs/blob/master/kad-dht/README.md#value-retrieval) we can't do more than `alpha` requests at a time:
> We keep track of:
> 
> * the number of values we've fetched (`cnt`).
> * the best value we've found (`best`), and which peers returned it (`Pb`)
> * the set of peers we've already queried (`Pq`) and the set of next query
>   candidates sorted by distance from `Key` in ascending order (`Pn`).
> * the set of peers with outdated values (`Po`).
> 
> At initialization we seed `Pn` with the `α` peers from our routing table we know
> are closest to `Key`, based on the XOR distance function.
> 
> Then we loop:
> 
> 1. If we have collected `Q` or more answers, we cancel outstanding requests and
>    return `best`. If there are no outstanding requests and `Pn` is empty we
>    terminate early and return `best`. In either case we notify the peers holding
>    an outdated value (`Po`) of the best value we discovered, or holding no value
>    for the given key, even though being among the `k` closest peers to the key,
>    by sending `PUT_VALUE(Key, best)` messages.
> 2. Pick as many peers from the candidate peers (`Pn`) as the `α` concurrency
>    factor allows. Send each a `GET_VALUE(Key)` request, and mark it as _queried_
>    in `Pq`.
> 3. Upon a response:
> 	1. If successful, and we receive a value:
> 		1. If this is the first value we've seen, we store it in `best`, along
>            with the peer who sent it in `Pb`.
> 		2. Otherwise, we resolve the conflict by e.g. calling
>            `Validator.Select(best, new)`:
> 			1. If the new value wins, store it in `best`, and mark all formerly
>                “best" peers (`Pb`) as _outdated peers_ (`Po`). The current peer
>                becomes the new best peer (`Pb`).
> 			2. If the new value loses, we add the current peer to `Po`.
> 	2. If successful with or without a value, the response will contain the
>        closest nodes the peer knows to the `Key`. Add them to the candidate
>        list `Pn`, except for those that have already been queried.
> 	3. If an error or timeout occurs, discard it.
> 4. Go to 1.

vladopajic: `getValue` is not doing this:

> Then we loop:
    If we have collected Q or more answers, we cancel outstanding requests and
    return best. If there are no outstanding requests and Pn is empty we
    terminate early and return best. In either case we notify the peers holding
    an outdated value (Po) of the best value we discovered, or holding no value
    for the given key, even though being among the k closest peers to the key,
    by sending PUT_VALUE(Key, best) messages.
    Pick as many peers from the candidate peers (Pn) as the α concurrency
    factor allows. Send each a GET_VALUE(Key) request, and mark it as queried
    in Pq.
    Upon a response:
    If successful, and we receive a value:
        If this is the first value we've seen, we store it in best, along
        with the peer who sent it in Pb.
        Otherwise, we resolve the conflict by e.g. calling
        Validator.Select(best, new):
            If the new value wins, store it in best, and mark all formerly
            “best" peers (Pb) as outdated peers (Po). The current peer
            becomes the new best peer (Pb).
            If the new value loses, we add the current peer to Po.
    If successful with or without a value, the response will contain the
    closest nodes the peer knows to the Key. Add them to the candidate
    list Pn, except for those that have already been queried.
    If an error or timeout occurs, discard it.
    Go to 1.


there is ether wrong indentation or i am not understanding this code? but it appears like there is 
1) wrong indentation
2) lack of test that would assert correctness
3) algorithm is not matching specs

vladopajic (changes requested):
- getValue does not follow specs
- not convinced that proposed template is needed until logic matches getValue spec
- even if all above are correct, template solution does not seem to beat neatens of:
   ```for chunk in remainingPeers.toChunks(chunkSize):```

libp2p/protocols/kademlia/kademlia.nim
vladopajic: image

vladopajic: highlighted is the `loop` that is implemented

gmelodie: Yes it's incomplete, #1765 will complete it, along with this one

libp2p/protocols/kademlia/kademlia.nim
gmelodie: sure, but this is a nitpick

libp2p/protocols/kademlia/kademlia.nim
vladopajic: since this pr is just one small proc, then it makes sense to have it in complete pr and to review it as whole.
merging pr, like this one, doesn't make sense as solution that current pr wants to provide might be irrelevant in next pr. as example, previously we merge pr with `timeout` argument, although it wasn't used in that pr. and in next pr it was removed.

however leave it to you

libp2p/protocols/kademlia/kademlia.nim
vladopajic: nitpicks usually mean _non functional_ changes. 
these two have vary different functional implications and consequences of having both.

PR #1765, opened 2025-10-15
chore(kad): getVal loop improvements

### LINKED ISSUE #1763: Better getvalue loop
- Update key on nodes not kaving the key
- Add closerPeers received to list of candidate peers
- Send closerPeers when receiving a GetValue

gmelodie:
Fixes #1763
- Update key on nodes not kaving the key
- Add closerPeers received to list of candidate peers
- Send closerPeers when receiving a GetValue

### DIFF
--- a/libp2p/protocols/kademlia/kademlia.nim
+++ b/libp2p/protocols/kademlia/kademlia.nim
@@ -77,12 +77,12 @@ method select*(
 type KadDHTConfig* = ref object
   validator*: EntryValidator
   selector*: EntrySelector
-  timeout: chronos.Duration
-  retries: int
-  replication: int
-  alpha: int
-  ttl: chronos.Duration
-  quorum: int
+  timeout*: chronos.Duration
+  retries*: int
+  replication*: int
+  alpha*: int
+  ttl*: chronos.Duration
+  quorum*: int
 
 proc new*(
     T: typedesc[KadDHTConfig],
@@ -346,6 +346,8 @@ proc dispatchGetVal(
     error "GetValue reply decode fail", error = error, conn = conn
     return
 
+  received[peer] = Opt.none(EntryRecord)
+
   for peer in reply.closerPeers:
     let p = PeerId.init(peer.id).valueOr:
       debug "Invalid peer id received", error = error
@@ -353,20 +355,20 @@ proc dispatchGetVal(
     candidates[].incl(p)
 
   let record = reply.record.valueOr:
-    error "GetValue returned empty record", msg = msg, reply = reply, conn = conn
+    debug "GetValue returned empty record", msg = msg, reply = reply, conn = conn
     return
 
   let value = record.value.valueOr:
-    error "GetValue returned record with no value",
+    debug "GetValue returned record with no value",
       msg = msg, reply = reply, conn = conn
     return
 
   let time = record.timeReceived.valueOr:
-    error "GetValue returned record with no timeReceived",
+    debug "GetValue returned record with no timeReceived",
       msg = msg, reply = reply, conn = conn
     return
 
-  received[][peer] = Opt.some(EntryRecord(value: value, time: time))
+  received[peer] = Opt.some(EntryRecord(value: value, time: time))
 
 proc bestValidRecord(
     kad: KadDHT, key: Key, received: ReceivedTable
@@ -377,6 +379,9 @@ proc bestValidRecord(
       continue
     if kad.config.validator.isValid(key, record):
       validRecords.add(record)
+
+  if validRecords.len() < kad.config.quorum:
+    return err("Not enough valid records to achieve quorum")
 
   let selectedIdx = kad.config.selector.select(key, validRecords).valueOr:
     return err("Could not select best value")
--- a/tests/kademlia/testgetval.nim
+++ b/tests/kademlia/testgetval.nim
@@ -34,6 +34,7 @@ suite "KadDHT - GetVal":
       containsData(kad1, key, value)
       containsNoData(kad2, key)
 
+    kad2.config.quorum = 1
     discard await kad2.getValue(key)
 
     check:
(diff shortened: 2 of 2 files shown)


libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -445,6 +483,16 @@ proc refreshBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
       let randomKey = randomKeyInBucketRange(kad.rtable.selfId, i, kad.rng)
       discard await kad.findNode(randomKey)
 
+proc findClosestPeers*(kad: KadDHT, target: Key): seq[Peer] =
+  var closestPeers: seq[Peer]
+  for p in kad.rtable.findClosest(target, kad.config.replication).filterIt(
+    it != kad.switch.peerInfo.peerId.toKey()
+  ):
```
vladopajic: it's vary hard to read multi-line `for`

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -355,42 +366,69 @@ proc dispatchGetVal(
...
+
+  var validRecords: seq[EntryRecord]
+  for r in received.values():
+    let record = r.valueOr:
+      continue
+    if kad.config.validator.isValid(key, record):
+      validRecords.add(record)
+
+  if validRecords.len() == 0:
+    return err("No valid records not choose from")
```
vladopajic: should not have special case handling here, leave it for `selector.select`

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -341,6 +346,12 @@ proc dispatchGetVal(
     error "GetValue reply decode fail", error = error, conn = conn
     return
 
+  for peer in reply.closerPeers:
+    let p = PeerId.init(peer.id).valueOr:
+      debug "Invalid peer id received", error = error
+      continue
+    candidates[].incl(p)
```
richard-ramos, line 355: is the `[]` required here and in line 369 because these types are ref? why use ref types?

libp2p/protocols/kademlia/kademlia.nim
richard-ramos: Makes me wonder if the if  on line 374 is needed as well

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -23,9 +23,10 @@ proc init*(
 ): EntryRecord {.gcsafe, raises: [].} =
   EntryRecord(value: value, time: time.get(TimeStamp(ts: $times.now().utc)))
 
-type ReceivedTable = ref Table[PeerId, EntryRecord]
-
-type LocalTable* = Table[Key, EntryRecord]
+type
+  ReceivedTable = ref Table[PeerId, Opt[EntryRecord]]
```
richard-ramos: There is a TableRef type, that is created with newTable.  This line would be: `type ReceivedTable = TableRef[PeerId, Opt[EntryRecord]]` and would save you from using [] for any table op

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -355,42 +366,69 @@ proc dispatchGetVal(
...
 
 proc getValue*(
     kad: KadDHT, key: Key
 ): Future[Result[EntryRecord, string]] {.async: (raises: [CancelledError]), gcsafe.} =
-  var remainingPeers = await kad.findNode(key)
-  var received = ReceivedTable(newTable[PeerId, EntryRecord]())
+  let candidates = CandidatePeers()
+  for p in (await kad.findNode(key)):
+    candidates[].incl(p)
+  let received = ReceivedTable(newTable[PeerId, Opt[EntryRecord]]())
```
richard-ramos: If following suggestion of TableRef, this line would likely be:
```
let received = newTableRef[PeerId, Op[EntryRecord]]().
```
The question remains whether this should be heap or stack 🤔

libp2p/protocols/kademlia/kademlia.nim
gmelodie: > why use ref types?

to avoid copying when passing to dispatch functions

libp2p/protocols/kademlia/kademlia.nim
gmelodie: This is cool! Thanks :)

richard-ramos (approved):

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -23,9 +23,10 @@ proc init*(
 ): EntryRecord {.gcsafe, raises: [].} =
   EntryRecord(value: value, time: time.get(TimeStamp(ts: $times.now().utc)))
 
-type ReceivedTable = ref Table[PeerId, EntryRecord]
-
-type LocalTable* = Table[Key, EntryRecord]
+type
+  ReceivedTable = TableRef[PeerId, Opt[EntryRecord]]
```
vladopajic, line 27: not sure about this: table should have value as type; not opt value. 
when there is value in table it's not optional. table { gabe=human }, key "gabe" has value.
Opt.None() is same as having no key=value in table, that is if you look for "vlado" in table, there is no value. adding additional level of value presence seems unnecessary ?

libp2p/protocols/kademlia/kademlia.nim
```diff
@@ -341,57 +346,88 @@ proc dispatchGetVal(
...
-  received[][peer] = EntryRecord(value: value, time: time)
+  received[peer] = Opt.some(EntryRecord(value: value, time: time))
+
+proc bestValidRecord(
+    kad: KadDHT, key: Key, received: ReceivedTable
+): Result[EntryRecord, string] =
+  var validRecords: seq[EntryRecord]
+  for r in received.values():
+    let record = r.valueOr:
+      continue
```
vladopajic, line 379: would be better if `received.values()` actually returned values so check `let record = r.valueOr:` is not needed.

Issue #1633, opened 2025-08-19
quic: GossipSub mesh is not forming as well


vladopajic:
when quic is used in performance workflow, instead of TCP transport, it's kind of working, but GossipSub mesh is not forming as well as in case of TCP.

```
TCP
INF 2025-08-18 15:41:08.528+00:00 Mesh populated, synchronizing              tid=1 nodeId=9 meshSize=8 gossipsub=8 outbound=4
INF 2025-08-18 15:41:08.537+00:00 Mesh populated, synchronizing              tid=1 nodeId=2 meshSize=8 gossipsub=8 outbound=5
INF 2025-08-18 15:41:08.558+00:00 Mesh populated, synchronizing              tid=1 nodeId=7 meshSize=8 gossipsub=8 outbound=4
INF 2025-08-18 15:41:08.561+00:00 Mesh populated, synchronizing              tid=1 nodeId=0 meshSize=5 gossipsub=7 outbound=4
INF 2025-08-18 15:41:08.564+00:00 Mesh populated, synchronizing              tid=1 nodeId=5 meshSize=7 gossipsub=7 outbound=5
INF 2025-08-18 15:41:08.577+00:00 Mesh populated, synchronizing              tid=1 nodeId=3 meshSize=6 gossipsub=7 outbound=3
INF 2025-08-18 15:41:08.590+00:00 Mesh populated, synchronizing              tid=1 nodeId=8 meshSize=7 gossipsub=9 outbound=2
INF 2025-08-18 15:41:08.593+00:00 Mesh populated, synchronizing              tid=1 nodeId=1 meshSize=7 gossipsub=7 outbound=3
INF 2025-08-18 15:41:08.606+00:00 Mesh populated, synchronizing              tid=1 nodeId=6 meshSize=8 gossipsub=8 outbound=2
INF 2025-08-18 15:41:08.614+00:00 Mesh populated, synchronizing              tid=1 nodeId=4 meshSize=6 gossipsub=7 outbound=3
```
```
QUIC
INF 2025-08-18 15:41:40.007+00:00 Mesh populated, synchronizing              tid=1 nodeId=8 meshSize=2 gossipsub=5 outbound=0
INF 2025-08-18 15:41:40.032+00:00 Mesh populated, synchronizing              tid=1 nodeId=1 meshSize=3 gossipsub=6 outbound=0
[...]

PR #1768, opened 2025-10-16
fix(quic): GossipSub outbound peers

rlve:
While adding QUIC scenario to Performance Tests I noticed that GossipSub was not able to distinguish outbound peers, which could affect mesh formation [[spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#outbound-mesh-quotas)].

image

### Changes
This PR attempts to fix it. Also, I noticed there was no explicit test around this, so I added it as well.

### DIFF
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -354,7 +354,7 @@ method accept*(
 
   try:
     let connection = await self.listener.accept()
-    return self.wrapConnection(connection, transportDir = Direction.In)
+    return self.wrapConnection(connection, Direction.In)
   except QuicError as exc:
     debug "Quic Error", description = exc.msg
   except TransportOsError as exc:
@@ -382,7 +382,7 @@ method dial*(
 
     let client = self.client.get()
     let quicConnection = await client.dial(taAddress)
-    return self.wrapConnection(quicConnection, transportDir = Direction.Out)
+    return self.wrapConnection(quicConnection, Direction.Out)
   except QuicConfigError as e:
     raise newException(
       QuicTransportDialError, "error in quic dial: invalid tls config:" & e.msg, e


libp2p/transports/quictransport.nim
```diff
@@ -348,7 +354,7 @@ method accept*(
 
   try:
     let connection = await self.listener.accept()
-    return self.wrapConnection(connection)
+    return self.wrapConnection(connection, transportDir = Direction.In)
```
vladopajic: no need to name argument, just pass it

libp2p/transports/quictransport.nim
```diff
@@ -376,7 +382,7 @@ method dial*(
 
     let client = self.client.get()
     let quicConnection = await client.dial(taAddress)
-    return self.wrapConnection(quicConnection)
+    return self.wrapConnection(quicConnection, transportDir = Direction.Out)
```
vladopajic: no need to name argument, just pass it

vladopajic (approved):

vladopajic:
this pr should fixes [#1633](https://github.com/vacp2p/nim-libp2p/issues/1633)

libp2p/transports/quictransport.nim
rlve: I did that on purpose, to highlight it's transport direction and not confuse with QUIC Stream direction on a first glance.

vladopajic: understood. however i am referring to usage of `proc`s.

arguments should not be assigned as named unless there is reason to miss initialize different argument, when there are default values with same type. for example:
```
foo(a: string = "a", b: string = "b")

foo("aNew") -> a=aNew, b=b
foo("bNew") -> a=bNew, b=b 
foo(b = "bNew") -> a=a, b=bNew # we must use `b=` in order to assign value to b
```

that is this is just usage thing. and we should try to use it as default usage, without naming. because otherwise code gets cumbersome.

also note it would be better to avoid having code like `foo`.

--- 

now if you need to explain something, it should be comment on `wrapConnection` proc as documentation why transport direction is passed. because using named argument is not telling us anything.

rlve: fair enough, removed

PR #1770, opened 2025-10-18
chore: remove upper chronicles bound

arnetheduck:
This was added because of a nimble bug - instead of an upper bound, a newer nimble should have been used instead.

As is, it prevents anyone from using newer chronicles versions - in any case, 1.6 support has been dropped since, so it is no longer relevant.


arnetheduck:
build errors unrelated to this change

gmelodie (approved):

vladopajic (approved):

PR #1773, opened 2025-10-20
feat(kad): addProvider

gmelodie:
- Add `addProvider` RPC message
- Split RPC messages into modules (avoid bloating `kademlia.nim`)
- Include `testgetval.nim` and `testprovider.nim` in `testnative.nim`

### DIFF
--- a/libp2p/protocols/kademlia/provider.nim
+++ b/libp2p/protocols/kademlia/provider.nim
@@ -14,7 +14,7 @@ import ../protocol
 import ./[protobuf, types, find]
 
 proc dispatchAddProvider(
-    switch: Switch, peer: PeerId, key: Key
+    switch: Switch, peer: PeerId, cid: Cid
 ) {.async: (raises: [CancelledError, DialFailedError, LPStreamError]).} =
   let conn = await switch.dial(peer, KadCodec)
   defer:
@@ -22,18 +22,17 @@ proc dispatchAddProvider(
 
   let msg = Message(
     msgType: MessageType.addProvider,
-    key: key,
+    key: cid.toKey(),
     providerPeers: @[switch.peerInfo.toPeer()],
   )
   await conn.writeLp(msg.encode().buffer)
 
-proc addProvider*(kad: KadDHT, key: Key) {.async: (raises: [CancelledError]), gcsafe.} =
+proc addProvider*(kad: KadDHT, cid: Cid) {.async: (raises: [CancelledError]), gcsafe.} =
   ## Find the closest nodes to the key via FIND_NODE and send ADD_PROVIDER with self's peerInfo to each of them
 
-  let peers = await kad.findNode(key)
+  let peers = await kad.findNode(cid.toKey())
   for chunk in peers.toChunks(kad.config.alpha):
-    let rpcBatch =
-      chunk.mapIt(kad.switch.dispatchAddProvider(it, key.toCid().data.buffer))
+    let rpcBatch = chunk.mapIt(kad.switch.dispatchAddProvider(it, cid))
     try:
       await rpcBatch.allFutures().wait(kad.config.timeout)
     except AsyncTimeoutError:
--- a/tests/kademlia/testprovider.nim
+++ b/tests/kademlia/testprovider.nim
@@ -41,7 +41,7 @@ suite "KadDHT - AddProvider":
       switch3.peerInfo.addrs
 
     # kad1 has kad3 in its addressbook after adding provider
-    await kad3.addProvider(key)
+    await kad3.addProvider(key.toCid())
     await sleepAsync(10.milliseconds)
     check switch1.peerStore[AddressBook][switch3.peerInfo.peerId] ==
       switch3.peerInfo.addrs
--- a/libp2p/protocols/kademlia/types.nim
+++ b/libp2p/protocols/kademlia/types.nim
@@ -38,6 +38,9 @@ proc toCid*(k: Key): Cid =
     debug "Key is an invalid CID, encapsulating", key = k
     Cid.init(CIDv1, multiCodec("dag-pb"), MultiHash.digest("sha2-256", k).get()).get()
 
+proc toKey*(c: Cid): Key =
+  c.mhash().get().data.buffer
+
 proc toKey*(p: PeerId): Key =
   return Key(p.data)
 


libp2p/protocols/kademlia/provider.nim
```diff
@@ -0,0 +1,63 @@
...
+proc dispatchAddProvider(
+    switch: Switch, peer: PeerId, key: Key
+) {.async: (raises: [CancelledError, DialFailedError, LPStreamError]).} =
+  let conn = await switch.dial(peer, KadCodec)
+  defer:
+    await conn.close()
+
+  let msg = Message(
+    msgType: MessageType.addProvider,
+    key: Opt.some(key),
```
vladopajic: how can `key` be optional?

libp2p/protocols/kademlia/provider.nim
```diff
@@ -0,0 +1,63 @@
...
+    error "No key in message buffer", msg = msg, conn = conn
+    return
+
+  # Validate CID
+  if key.len > 80:
+    error "Key size too large", msg = msg, conn = conn, key = key
+    return
+  if key.len == 0:
+    error "Key is empty", msg = msg, conn = conn, key = key
+    return
```
vladopajic: should be implemented with:
`proc validateKey(key)`

libp2p/protocols/kademlia/provider.nim
```diff
@@ -0,0 +1,63 @@
...
+
+  # filter out infos that do not match sender's
+  for peer in msg.providerPeers.filterIt(it.id == conn.peerId.getBytes()):
```
vladopajic: `getBytes()` called for every peer -> memory trashing

libp2p/protocols/kademlia/provider.nim
vladopajic: proc always gives benefits:
- names block of code
- enables reuse (addProviders can use same proc in doAssert construct)
- enables testing

tests/kademlia/testprovider.nim
```diff
@@ -0,0 +1,65 @@
...
+      value = @[1.byte, 2, 3, 4, 5]
+
+    kad1.dataTable.insert(key, value, $times.now().utc)
+    kad2.dataTable.insert(key, value, $times.now().utc)
+    kad3.dataTable.insert(key, value, $times.now().utc)
+
+    check:
+      containsData(kad1, key, value)
+      containsData(kad2, key, value)
+      containsData(kad3, key, value)
```
vladopajic: not sure if needed to be checked here. `insert` should be tested elsewhere, do don't need to assert same things all over the place. assume correctness if tests covers it.

tests/kademlia/testprovider.nim
```diff
@@ -0,0 +1,65 @@
...
+
+    # repeat several times because address book can be updated by multiple sources
+
+    for _ in 0 ..< 10:
```
vladopajic: why repeat, if any of checks fail test fails?

libp2p/protocols/kademlia/provider.nim
vladopajic: maybe compiler knows how to optimize things in filerIt..

tests/kademlia/testprovider.nim
gmelodie: While testing I needed this to make sure it was working, adding the small `sleepAsync` fixed the issue (maybe addressbook takes a bit to update?), but I thought it couldn't hurt to leave the for there

libp2p/protocols/kademlia/provider.nim
gmelodie: Good question, I'll take a look and since this would touch all rpc methods, I'll implement this in another PR: https://github.com/vacp2p/nim-libp2p/issues/1756?issue=vacp2p%7Cnim-libp2p%7C1774

libp2p/protocols/kademlia/provider.nim
```diff
@@ -0,0 +1,60 @@
...
+    await conn.close()
+
+  let msg = Message(
+    msgType: MessageType.addProvider,
+    key: Opt.some(key),
+    providerPeers: @[switch.peerInfo.toPeer()],
+  )
+  await conn.writeLp(msg.encode().buffer)
+
+proc addProvider*(kad: KadDHT, key: Key) {.async: (raises: [CancelledError]), gcsafe.} =
```
richard-ramos: Is the provider record republish interval going to be handled as aprt of a separate PR?

richard-ramos: In the specs:
```
the request key is set to a CID. The target node verifies key is a valid CID, all providerPeers that match the RPC sender's PeerID are recorded as providers.
```

This would indicate that the parameter should not be a `Key` but a CID, and we should do the conversion ourselves to Key in this function

libp2p/protocols/kademlia/provider.nim
```diff
@@ -0,0 +1,60 @@
...
+    return
+
+  # filter out infos that do not match sender's
+  let peerBytes = conn.peerId.getBytes()
+  for peer in msg.providerPeers.filterIt(it.id == peerBytes):
+    let p = PeerId.init(peer.id).valueOr:
+      debug "Invalid peer id received", error = error
+      continue
+    # add provider to peerstore
+    kad.switch.peerStore[AddressBook][p] = peer.addrs
```
richard-ramos, line 58: Besides storing the address of the peer, where are we storing the providers for some key? If we implement some sort of structure for storing/retrieving providers, we should also take into account:
- provider records expire
- the  node calls ADD_PROVIDER would also be considered a provider, thus should be registered in the same provider structure

libp2p/protocols/kademlia/provider.nim
```diff
@@ -0,0 +1,60 @@
...
+
+proc handleAddProvider*(
+    kad: KadDHT, conn: Connection, msg: Message
+) {.async: (raises: [CancelledError]).} =
+  let key = msg.key.valueOr:
+    error "No key in message buffer", msg = msg, conn = conn
+    return
+
+  # Validate CID
+  if not key.isValid():
```
richard-ramos: Validation should likely be based on `Cid.init` being succesful

gmelodie: I wonder why go doesn't do that https://github.com/libp2p/go-libp2p-kad-dht/blob/095fe972d6b85f3481660d3e21978d7fdeb88c49/handlers.go#L323-L328

libp2p/protocols/kademlia/provider.nim
gmelodie: Let's implement a `ProviderManager` as Go does as a follow up (https://github.com/vacp2p/nim-libp2p/issues/1778)

libp2p/protocols/kademlia/provider.nim
gmelodie: https://github.com/vacp2p/nim-libp2p/issues/1778

libp2p/protocols/kademlia/provider.nim
richard-ramos: Good question, they seem to not be following the spec?
> ADD_PROVIDER: In the request key is set to a CID. The target node verifies key is a valid CID ...

libp2p/protocols/kademlia/provider.nim
richard-ramos: To implement this in a fashion similar to js-libp2p and go-libp2p (rust does not follow the spec wrt using CIDs), we would have to do the following:

1. Change addProvider to:
```
proc addProvider*(kad: KadDHT, cid: Cid) {.async: (raises: [CancelledError]), gcsafe.} =
```
2.  In findNode the `key` would be the bytes from `cid.mhash()`
3.  Change dispatchAddProvider to:
```
proc dispatchAddProvider(
    switch: Switch, peer: PeerId, mhash: MultiHash
) {.async: (raises: [CancelledError, DialFailedError, LPStreamError]).} =
```
4. The key on the `Message` to dispatch would be mhash.data.buffer`

In the handler `Cid.init` would not fail, it would just contain the multihash.

PR #1775, opened 2025-10-21
fix: init method for MixPubInfo

chaitanyaprem:
~~Tiniest fix ever i guess :)~~ 

Not so tiny anymore, no idea why examples kept failing only in this PR

Without this, we can't use the mix from libp2p in nwaku. 

@richard-ramos is there a way to give a patch release so that we can start using libp2p/mix instead of mix repo?

cc @fryorcraken

### DIFF
--- a/libp2p/protocols/identify.nim
+++ b/libp2p/protocols/identify.nim
@@ -152,6 +152,7 @@ method init*(p: Identify) =
       trace "handling identify request", conn
       var pb = encodeMsg(p.peerInfo, conn.observedAddr, p.sendSignedPeerRecord)
       await conn.writeLp(pb.buffer)
+      debug "identify: info sent", conn, info = p.peerInfo
     except CancelledError as exc:
       trace "cancelled identify handler"
       raise exc
@@ -180,7 +181,7 @@ proc identify*(
 
   var info = decodeMsg(message).valueOr:
     raise newException(IdentityInvalidMsgError, "Incorrect message received!")
-  debug "identify: decoded message", conn, info
+  debug "identify: info received", conn, info
   let
     pubkey = info.pubkey.valueOr:
       raise newException(IdentityInvalidMsgError, "No pubkey in identify")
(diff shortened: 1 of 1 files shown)


libp2p/protocols/mix/mix_node.nim
```diff
@@ -159,7 +159,7 @@ proc init*(
     libp2pPubKey: SkPublicKey,
 ): T =
   T(
-    peerId: PeerId,
+    peerId: peerId,
```
vladopajic, line 162: :O

vladopajic: how could this pass compile in our ci if type was used instead of case-insensitive value?

vladopajic: @chaitanyaprem are you using nim version >= 2.0?

libp2p/protocols/mix/exit_layer.nim
```diff
@@ -1,4 +1,4 @@
-import chronicles, chronos, metrics, std/sequtils
+import chronicles, chronos, metrics, std/sequtils, std/tables
```
richard-ramos, line 1: I'm very confused too on how is CI compiling when it was missing this import.

libp2p/protocols/mix/mix_node.nim
richard-ramos: wth

chaitanyaprem: Yep, in nwaku we use min version of 2.2.4

> @chaitanyaprem are you using nim version >= 2.0?

chaitanyaprem: > how could this pass compile in our ci if type was used instead of case-insensitive value?

well, i think since there was no code using the init method it went through

libp2p/protocols/mix/exit_layer.nim
chaitanyaprem: Yep, no idea how it was working before this change. 
It was failing in my PR though without these

chaitanyaprem: > I'm very confused too on how is CI compiling when it was missing this import.

nim magic unveiled

vladopajic (approved):

libp2p/protocols/mix/exit_layer.nim
fryorcraken: I think the ordering of files passed by nimble impacts this. I had cases where no issue with nbs but some ambiguous calls for nimble.

PR #1783, opened 2025-10-21
chore(kad): make key field mandatory

gmelodie:


libp2p/protocols/kademlia/put.nim
```diff
@@ -30,7 +30,7 @@ proc dispatchPutVal*(
     await conn.close()
   let msg = Message(
     msgType: MessageType.putValue,
-    record: Opt.some(Record(key: Opt.some(key), value: Opt.some(value))),
+    record: Opt.some(Record(key: key, value: Opt.some(value))),
```
richard-ramos, line 34:
```suggestion
    key: key,
    record: Opt.some(Record(key: key, value: Opt.some(value))),
```
according to specs:
> key on Message is set to equal key of the Record


Probably we need to validate this as well in `handlePutValue`

richard-ramos (approved):

libp2p/protocols/kademlia/put.nim
gmelodie: image

PR #1784, opened 2025-10-22
test(quic): Add Multiaddress Handling tests 

rlve:
Changes:
- add tests: 
  - `"multiaddress validation - accept valid addresses"`
  - `"multiaddress validation - reject invalid addresses"`
  - `"address normalization - port assignment"`
  - `"cannot bind second listener to same port"`
  - `"dial with malformed multiaddresses"`
  - `"observedAddr and localAddr are populated on connections"`
- move util functions to `utils.nim`
- clean up imports

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,11 +116,6 @@ The libp2p implementation in Nim is a work in progress. We welcome contributors 
 
 ### Compile time flags
 
-Enable quic transport support
-```bash
-nim c -d:libp2p_quic_support some_file.nim
-```
-
 Enable autotls support
 ```bash
 nim c -d:libp2p_autotls_support some_file.nim
--- a/flake.nix
+++ b/flake.nix
@@ -26,7 +26,7 @@
       devShells = forAllSystems (system: {
         default = pkgsFor.${system}.mkShell {
           nativeBuildInputs = with pkgsFor.${system}; [
-            nim-2_2 nimble openssl.dev
+            nim-2_2 nimble
           ];
         };
       });
--- a/interop/transport/Dockerfile
+++ b/interop/transport/Dockerfile
@@ -13,6 +13,6 @@ COPY . nim-libp2p/
 
 RUN \
   cd nim-libp2p && \
-  nim c --skipProjCfg --skipParentCfg --NimblePath:./nimbledeps/pkgs2 -p:nim-libp2p --mm:refc -d:libp2p_quic_support -d:chronicles_log_level=WARN -d:chronicles_default_output_device=stderr --threads:on ./interop/transport/main.nim
+  nim c --skipProjCfg --skipParentCfg --NimblePath:./nimbledeps/pkgs2 -p:nim-libp2p --mm:refc -d:chronicles_log_level=WARN -d:chronicles_default_output_device=stderr --threads:on ./interop/transport/main.nim
 
 ENTRYPOINT ["/app/nim-libp2p/interop/transport/main"]
--- a/libp2p.nim
+++ b/libp2p.nim
@@ -52,6 +52,7 @@ else:
       stream/connection,
       transports/transport,
       transports/tcptransport,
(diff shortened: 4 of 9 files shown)


tests/transports/testquic.nim
```diff
@@ -424,3 +349,97 @@ suite "Quic transport":
     await runClient(server)
     await serverFut
     await server.stop()
+
+  asyncTest "multiaddress validation - accept valid addresses":
+    let transport = await createTransport()
+    defer:
+      await transport.stop()
```
vladopajic: hint: `createTransport()` creates transport but does not start it. transport is started in `createTransport()` only if `isServer = true`. because transports used fro server sides do not need to be started.

https://github.com/vacp2p/nim-libp2p/pull/1784/files#diff-d96e6a34acc9f4de72802a718535853d536b6384bb901c3cbd6005575aa5cafdR79

rlve: oh right, removed

tests/transports/testquic.nim
```diff
@@ -424,3 +349,93 @@ suite "Quic transport":
...
+
+    # Try to bind client transport to same address
+    let client = await createTransport()
+    expect QuicTransportError:
+      await client.start(@[server.addrs[0]])
+
+  asyncTest "dial with malformed multiaddresses":
+    let server = await createTransport(isServer = true)
+    defer:
+      await server.stop()
```
vladopajic, line 410: we can use client here instead of server?
or maybe test with both?

tests/transports/testquic.nim
```diff
@@ -424,3 +349,93 @@ suite "Quic transport":
...
+
+  asyncTest "cannot bind second listener to same port":
+    let server = await createTransport(isServer = true)
+    defer:
+      await server.stop()
+
+    # Try to bind client transport to same address
+    let client = await createTransport()
+    expect QuicTransportError:
+      await client.start(@[server.addrs[0]])
```
vladopajic: `client` should be `server2` because transport here is attempting to become server by starting to accept

tests/transports/testquic.nim
```diff
@@ -424,3 +349,93 @@ suite "Quic transport":
...
+    for address in invalidAddresses:
+      check not transport.handles(MultiAddress.init(address).tryGet())
+
+  asyncTest "address normalization - port assignment":
+    # Start with port 0 and verify it gets assigned a real port
+    let transport = await createTransport()
+    let ma = MultiAddress.init("/ip4/127.0.0.1/udp/0/quic-v1").tryGet()
+    await transport.start(@[ma])
+    defer:
+      await transport.stop()
```
vladopajic, line 385: technically `createTransport(isServer = true)` does exactly this, however iam fine with repeating it

tests/transports/testquic.nim
```diff
@@ -424,3 +349,93 @@ suite "Quic transport":
...
+
+    check transport.addrs[0] != ma # Should assign non-zero port
```
vladopajic: would be nice to have other check(or test) here where
`check transport.addrs[0] == ma`
that is ip is used

tests/transports/testquic.nim
rlve: Changing to client would only mean that the transport is not started, right?

So you would like the same test for both: transport started and not?

vladopajic: yes, correct

gmelodie (approved):
Multiaddress validation tests should be shared across all transport tests. They can't be put into `commontransporttests` because quic doesn't use those, but maybe at least for now we could create a `reallycommontransporttests.nim` to hold only these, and import those in all transports. In the future, we'd use this structure to keep adding tests that can actually be run in all transports.

Obs: accepted, but I'd like to see other tests using the same code, preferrably in this PR.

PR #1789, opened 2025-10-23
feat(kad): add ProviderManager

### LINKED ISSUE #1778: Implement ProviderManager
https://github.com/vacp2p/nim-libp2p/pull/1773#discussion_r2448210104

https://github.com/vacp2p/nim-libp2p/pull/1773#discussion_r2448257494

gmelodie:
`ProviderManager`:
- keeps track of providers for each key
- cleans up stale provider records
- keeps track of the keys provided by us
- republishes keys provided by us

Fixes #1778

### DIFF
--- a/libp2p/protocols/kademlia/provider.nim
+++ b/libp2p/protocols/kademlia/provider.nim
@@ -34,7 +34,6 @@ proc addProviderRecord(pm: ProviderManager, record: ProviderRecord) =
     # remove old providerRecord if any
     let oldRecordIdx = pm.records.find(record)
     if oldRecordIdx != -1:
-      debug "updating provider"
       pm.records.del(oldRecordIdx)
 
     # push new providerRecord
@@ -99,10 +98,8 @@ proc hasExpiredRecords(pm: ProviderManager): bool =
 
 proc manageExpiredProviders*(kad: KadDHT) {.async: (raises: [CancelledError]).} =
   heartbeat "cleanup expired provider records", kad.config.cleanupProvidersInterval:
-    debug "cleaning up expired records", now = chronos.Moment.now()
     while kad.providerManager.hasExpiredRecords():
       let expired = kad.providerManager.records.pop()
-      debug "found expired record", now = chronos.Moment.now(), expired = expired
       kad.providerManager.rmProviderRecord(expired)
 
 proc handleAddProvider*(
@@ -115,10 +112,8 @@ proc handleAddProvider*(
   # filter out infos that do not match sender's
   let peerBytes = conn.peerId.getBytes()
 
-  debug "got addProvider", providerPeers = msg.providerPeers.len()
   for peer in msg.providerPeers.filterIt(it.id == peerBytes):
     let p = PeerId.init(peer.id).valueOr:
-      debug "Invalid peer id received", error = error
       continue
 
     # add provider to providerManager
@@ -129,7 +124,3 @@ proc handleAddProvider*(
         key: msg.key.toCid(),
       )
     )
-    debug "added ProviderRecord",
-      provider = peer.id,
-      key = msg.key.toCid(),
-      recordsLen = kad.providerManager.records.len()
--- a/tests/kademlia/testprovider.nim
+++ b/tests/kademlia/testprovider.nim
@@ -9,14 +9,11 @@
 
 {.used.}
 
-import std/[heapqueue, sets]
+import std/[heapqueue, sets, tables]
 from std/times import now, utc
 import chronos
-import unittest2
+import ../../libp2p/[protocols/kademlia, switch, builders]
 import ../tools/[unittest]
-import ../../libp2p/[switch, builders]
-import ../../libp2p/protocols/kademlia
-import ../utils/async_tests
 import ./utils.nim
 
 suite "KadDHT - ProviderManager":
@@ -79,13 +76,15 @@ suite "KadDHT - ProviderManager":
     check kad1.providerManager.records.len() == 2
 
     # wait less than expiration time
-    await sleepAsync(CleanupInterval)
+    await sleepAsync(kad1.config.cleanupProvidersInterval)
 
     # provider records have not yet expired
     check kad1.providerManager.records.len() == 2
 
     # wait expiration time
-    await sleepAsync(ExpirationInterval + CleanupInterval)
+    await sleepAsync(
+      kad1.config.providerExpirationInterval + 2 * kad1.config.cleanupProvidersInterval
+    )
 
     # provider records expired and evicted
     check kad1.providerManager.records.len() == 0
@@ -118,7 +117,7 @@ suite "KadDHT - ProviderManager":
     check kad1.providerManager.records.len() == 2
 
     # wait less than expiration time
-    await sleepAsync(CleanupInterval)
+    await sleepAsync(kad1.config.cleanupProvidersInterval)
 
     # provider records have not yet expired
     check kad1.providerManager.records.len() == 2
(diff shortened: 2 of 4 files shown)


libp2p/protocols/kademlia/types.nim
```diff
@@ -161,6 +167,21 @@ type
...
+
+  ProviderRecord* = object
+    provider*: Provider
+    expiresAt*: chronos.Moment
+    key*: Cid
+
+  ProviderManager* = ref object
+    records*: HeapQueue[ProviderRecord]
+    knownKeys*: Table[Cid, HashSet[Provider]]
+    providedKeys*: HashSet[Cid]
```
richard-ramos, line 180: Maybe we should call this knownCids and providedCids since the key for both the table and hashset is a Cid. (Same applies to ProviderRecord)

libp2p/protocols/kademlia/types.nim
```diff
@@ -161,6 +167,21 @@ type
...
+
+proc new*(T: typedesc[ProviderManager]): T =
+  T(providedKeys: initHashSet[Cid]())
```
richard-ramos, line 183: Not really necessary to call initHashSet:
https://nim-lang.org/docs/sets.html#init,HashSet[A]
> Starting from Nim v0.20, sets are initialized by default and it is not necessary to call this function explicitly.

libp2p/protocols/kademlia/provider.nim
```diff
@@ -7,12 +7,53 @@
...
+# for HeapQueue
+proc `<`*(a, b: ProviderRecord): bool =
+  a.expiresAt < b.expiresAt
+
+proc `<`*(a: ProviderRecord, b: chronos.Moment): bool =
+  a.expiresAt < b
+
+proc addProviderRecord(pm: ProviderManager, record: ProviderRecord) =
+  if not pm.knownKeys.hasKey(record.key):
+    pm.knownKeys[record.key] = initHashSet[Provider]()
```
richard-ramos, line 29: It shouldnt be necessary to call initHashSet 
https://nim-lang.org/docs/sets.html#init,HashSet[A]
> Starting from Nim v0.20, sets are initialized by default and it is not necessary to call this function explicitly.

libp2p/protocols/kademlia/provider.nim
```diff
@@ -39,20 +80,48 @@ proc addProvider*(kad: KadDHT, cid: Cid) {.async: (raises: [CancelledError]), gc
       # Dispatch will timeout if any of the calls don't receive a response (which is normal)
       discard
 
+proc startProviding*(kad: KadDHT, c: Cid) {.async: (raises: [CancelledError]).} =
+  kad.providerManager.providedKeys.incl(c)
+  await kad.addProvider(c)
+
+proc stopProviding*(kad: KadDHT, c: Cid) =
+  kad.providerManager.providedKeys.excl(c)
```
richard-ramos: Should this also remove the key from `knownKeys` and the `records` that know about this cid?

libp2p/protocols/kademlia/types.nim
```diff
@@ -161,6 +167,21 @@ type
     buckets*: seq[Bucket]
     config*: RoutingTableConfig
 
+  Provider* = Peer
+
+  ProviderRecord* = object
+    provider*: Provider
+    expiresAt*: chronos.Moment
+    key*: Cid
```
richard-ramos, line 175: Should this be called cid?

libp2p/protocols/kademlia/provider.nim
```diff
@@ -39,20 +80,48 @@ proc addProvider*(kad: KadDHT, cid: Cid) {.async: (raises: [CancelledError]), gc
       # Dispatch will timeout if any of the calls don't receive a response (which is normal)
       discard
 
+proc startProviding*(kad: KadDHT, c: Cid) {.async: (raises: [CancelledError]).} =
```
richard-ramos, line 83: Something to consider:
you might want to provide a CID for some Duration, or even provide it infinitely (, thus maybe you'd want to have a signature like:
```nim
proc startProviding*(kad: KadDHT, c: Cid, duration: Duration = InfiniteDuration)
```
would require `providedKeys` to be something like:
```nim
type ProvidedContent = object
   cid: Cid
   expiresAt*: chronos.Moment

type ProviderManager* = ref object
  ...
  providedKeys*: HashSet[ProvidedContent]
```
and the republish to check if the provided content isnt expired or GC it otherwise

libp2p/protocols/kademlia/provider.nim
```diff
@@ -7,12 +7,53 @@
...
+    let oldRecordIdx = pm.records.find(record)
+    if oldRecordIdx != -1:
+      pm.records.del(oldRecordIdx)
+
+    # push new providerRecord
+    pm.records.push(record)
+  except KeyError:
+    raiseAssert("checked with hasKey")
+
+proc rmProviderRecord(pm: ProviderManager, record: ProviderRecord) =
```
richard-ramos, line 44: Nit: Even if it's a mouthful... could it be:
```suggestion
proc removeProviderRecord(pm: ProviderManager, record: ProviderRecord) =
```
Havent seen the `rm` prefix being used in other places in libp2p

richard-ramos:
Another thing to consider. This provider manager has no limits. Should we consider adding a max number of records before dropping new records (or maybe adding a LRU?). Could be done of course as part of a diff PR if you consider adding this feature

tests/kademlia/testprovider.nim
```diff
@@ -1,57 +1,175 @@
...
+    await kad2.addProvider(key2.toCid())
+
+    # wait rest of expiration time
+    await sleepAsync(ExpirationInterval - 2 * CleanupInterval)
+
+    # provider records have not expired (refreshed)
+    check kad1.providerManager.records.len() == 2
+
+    # wait expiration time
+    await sleepAsync(ExpirationInterval + 2 * CleanupInterval)
```
vladopajic: it's more correct to use kad config for these intervals, kad config is source of truth. these const are unnecessary. they also make code more unreadable as one has to trace const to see how they end up in kad config.

libp2p/protocols/kademlia/provider.nim
```diff
@@ -7,12 +7,53 @@
...
+
+  if not pm.knownCids.hasKey(record.cid):
+    return
+
+  try:
+    pm.knownCids[record.cid].excl(record.provider)
+    if pm.knownCids[record.cid].len() == 0:
+      pm.knownCids.del(record.cid)
+  except KeyError:
+    raiseAssert("checked with hasKey")
```
vladopajic, line 55:
```nim
  try:
    pm.knownCids[record.cid].excl(record.provider)
    if pm.knownCids[record.cid].len() == 0:
      pm.knownCids.del(record.cid)
  except KeyError:
    discard # do nothing when knownCids don't have cid
```

consider that code don't really need `if not pm.knownCids.hasKey(record.cid): return`.  
this is not suggestion, just fyi.

tests/kademlia/utils.nim
```diff
@@ -65,7 +71,15 @@ proc containsNoData*(kad: KadDHT, key: Key): bool {.raises: [].} =
...
   let kad = KadDHT.new(
-    switch, config = KadDHTConfig.new(validator, selector, timeout = chronos.seconds(1))
+    switch,
+    config = KadDHTConfig.new(
+      validator,
+      selector,
+      timeout = Timeout,
+      cleanupProvidersInterval = CleanupInterval,
+      providerExpirationInterval = ExpirationInterval,
+      republishProvidedKeysInterval = RepublishInterval,
```
vladopajic: setting default values from const seem unnecessary here, unless there are more then one place where these const should be used (like before in sleeps).

PR #1790, opened 2025-10-23
chore(ci): add verbose output for test binaries

vladopajic:
when ci (test) gets stuck `--output-level=VERBOSE` gives more information where exactly it got stuck.

### DIFF
--- a/libp2p.nimble
+++ b/libp2p.nimble
@@ -10,7 +10,7 @@ skipDirs = @["tests", "examples", "Nim", "tools", "scripts", "docs"]
 requires "nim >= 2.0.0",
   "nimcrypto >= 0.6.0", "dnsclient >= 0.3.0 & < 0.4.0", "bearssl >= 0.2.5",
   "chronicles >= 0.11.0", "chronos >= 4.0.4", "metrics", "secp256k1", "stew >= 0.4.2",
-  "websock >= 0.2.1", "unittest2", "results", "quic >= 0.4.0",
+  "websock >= 0.2.1", "unittest2", "results", "quic >= 0.5.0",
   "https://github.com/vacp2p/nim-jwt.git#18f8378de52b241f321c1f9ea905456e89b95c6f"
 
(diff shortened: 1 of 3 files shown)


libp2p.nimble
```diff
@@ -55,7 +61,7 @@ task testmultiformatexts, "Run multiformat extensions tests":
     "-d:libp2p_multihash_exts=../tests/multiformat_exts/multihash_exts.nim " &
     "-d:libp2p_multibase_exts=../tests/multiformat_exts/multibase_exts.nim " &
     "-d:libp2p_contentids_exts=../tests/multiformat_exts/contentids_exts.nim "
-  runTest("multiformat_exts/testmultiformat_exts", opts)
+  runTest("multiformat_exts/testmultiformat_exts", opts, false)
```
vladopajic, line 64: this tests fails when `--output-level=VERBOSE` is used.

https://github.com/vacp2p/nim-libp2p/actions/runs/18779571389/job/53582120566
ci passes on some platfforms and fails on others

```
=======================================
  D:\a\nim-libp2p\nim-libp2p\tests\multiformat_exts\testmultiformat_exts.exe "ContentId extensions::extended contentids correctly hash data"
---------------------------------------
    D:\a\nim-libp2p\nim-libp2p\tests\multiformat_exts\testcontentids_exts.nim(27, 15): Check failed: cid0.hex == "0181FE0382FE030668656C6C6F00"
    cid0.hex was 0181FE0382FE030668656C6C6F30
    D:\a\nim-libp2p\nim-libp2p\tests\multiformat_exts\testcontentids_exts.nim(29, 12): Check failed: $cid0 == "zZ9hvSmq764xLht1dRH"
    $cid0 was zZ9hvSmq764xLht1dS7
```

vladopajic: if you know why please suggest

vladopajic: but it's not big deal if those test are not executed with detailed output as it is just small set of tests.
however it would be better not to introduce unnecessary workaround if there is proper approach.

PR #1791, opened 2025-10-24
chore(performance): increase msg size

vladopajic:
2KB seems reasonable size to test performance. 200 bytes is just to small message.

### DIFF
--- a/performance/test_node/test_node.nim
+++ b/performance/test_node/test_node.nim
@@ -31,7 +31,7 @@ proc baseTest*(scenarioName: string, transport: TransportType) {.async.} =
     peerLimit = 5
     msgCount = 100
     msgInterval = 300 # ms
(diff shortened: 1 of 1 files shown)


rlve (approved):
👌

performance/test_node/test_node.nim
```diff
@@ -30,8 +30,8 @@ proc baseTest*(scenarioName: string, transport: TransportType) {.async.} =
     publisherCount = 5
     peerLimit = 5
     msgCount = 100
-    msgInterval = 100 # ms
-    msgSize = 200 # bytes
+    msgInterval = 300 # ms
+    msgSize = 1024 # 1KB
```
gmelodie, line 34:
```suggestion
    msgSize = 1024 # 1KiB
```

PR #1796, opened 2025-10-26
chore: remove redundant `gcsafe` from `switch.disconnect`

arnetheduck:


arnetheduck:
right - I was using a slightly older commit .. fwiw, `async` already implies `gcsafe` meaning the latter can be removed

PR #1798, opened 2025-10-27
chore: prevent nimble tasks to run if nim version is not satisfied

vladopajic:
will prevent issues like https://github.com/libp2p/test-plans/issues/692 more easily


README.md
```diff
@@ -42,7 +42,7 @@ Learn more about libp2p at [**libp2p.io**](https://libp2p.io) and follow libp2p'
 
 ## Install
 
-> The currently supported Nim versions are 2.0 and 2.2.
+The currently **supported Nim versions** are **v2.0** and **v2.2**.
 
 ```
 nimble install libp2p
```
vladopajic, line 48: note: i have tried added compile time error https://github.com/vacp2p/nim-libp2p/commit/0af12054b6ecbbae324e93ccd9c5f65bbe7413ef

but it doesn't make sense because if  `nim` dependency is not satisfied then compile will not even work, as it is prerequisite to compile.

right?

vladopajic: okay above does not have much sense, but preventing nimble tasks from running make sense:
https://github.com/vacp2p/nim-libp2p/pull/1798/commits/5a2d8363337b54c4ba7e7968ec0e6ffab807f822

image

note error is shown early on. even before evaluating dependencies.

libp2p.nimble
```diff
@@ -7,10 +7,17 @@ description = "LibP2P implementation"
 license = "MIT"
 skipDirs = @["tests", "examples", "Nim", "tools", "scripts", "docs"]
 
-requires "nim >= 2.0.0",
-  "nimcrypto >= 0.6.0", "dnsclient >= 0.3.0 & < 0.4.0", "bearssl >= 0.2.5",
-  "chronicles >= 0.11.0", "chronos >= 4.0.4", "metrics", "secp256k1", "stew >= 0.4.2",
-  "websock >= 0.2.1", "unittest2", "results", "quic >= 0.5.0",
+# Nim version check - prevents doing any nimble task if 
+# version is not satisfied to show more cleaner error to user.
+when (NimMajor, NimMinor) < (2, 0):
```
gmelodie, line 12: Where do `NimMajor` and `NimMinor` come from? Do we have to update those when changing nim version requirements?

gmelodie (approved):

libp2p.nimble
vladopajic: both are system consts. they are create and set by language
https://nim-lang.org/docs/system.html#NimMajor

arnetheduck: this is probably not a great idea - ie what you're looking at is a nimble bug that should be reported as such.

The constants here reflect the nim version that nimble uses internally for discovering version requirements, not necessarily the nim version that's used to compile the code, or that nimble selects to run tasks, ie it may change during various phases of nimble execution, so it is not very meaningful to check.

PR #1804, opened 2025-10-28
fix: length prefix handling in exit layer for lenght-prefixed protocols

chaitanyaprem:
Note that this issue was identified and fixed with the help of Copilot and Claude Sonnet 4.5 :) 
Here is a summary generated of the issue and fix. 

# Fix: Length Prefix Handling in Mix Protocol for Length-Prefixed Protocols

## Problem

When using the mix protocol with protocols that use length-prefixed messages (like lightpush), the length prefix gets lost during the round-trip through the mix network, causing data corruption.

### Root Cause

1. The destination server (e.g., lightpush) writes responses with `writeLp()` which adds a varint length prefix
2. The mix exit layer reads the response with `readLp()` which strips the length prefix
3. The response flows back through the mix network **without** the length prefix
4. Application code calls `readLp()` on `MixEntryConnection` expecting a length-prefixed message
5. **Bug**: Since no length prefix exists, `readLp()` misinterprets the first data byte as the length, causing it to read incomplete or incorrect data

## Solution

Introduced a wrapper type `DestReadBehavior` that tracks whether a read callback uses length prefixing:

- When `usesLengthPrefix=true`, the exit layer restores the length prefix before sending the reply through the mix network
- When `usesLengthPrefix=false`, the exit layer sends the raw response as-is
- Factory functions (`readLp()`, `readExactly()`) automatically set the correct flag
- No manual parameters needed at call sites

## Benefits

✅ **Type-safe**: Wrapper type ensures metadata travels with the callback  
✅ **Zero boilerplate**: Factory functions handle the flag automatically  
✅ **Self-documenting**: Type and field names make intent clear  
✅ **Clean API**: No boolean parameters at call sites  
✅ **Extensible**: Easy to add more metadata fields in the future

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,9 +30,9 @@ jobs:
           - os: windows
             cpu: amd64
         nim:
-          - ref: version-2-0
+          - ref: v2.0.16
             memory_management: refc
-          - ref: version-2-2
+          - ref: v2.2.4
             memory_management: refc
         include:
           - platform:
--- a/tests/pubsub/integration/testgossipsubmeshmanagement.nim
+++ b/tests/pubsub/integration/testgossipsubmeshmanagement.nim
@@ -338,25 +338,21 @@ suite "GossipSub Integration - Mesh Management":
     checkUntilTimeout:
       node0.mesh.getOrDefault(topic).len == dValues.get.d.get
 
-  # TODO: Remove loop when GossipSub tests fully switched to QUIC
-  const transports = @[TransportType.TCP, TransportType.QUIC]
-  for transport in transports:
-    asyncTest "Outbound peers are marked correctly " & $transport:
-      let
-        numberOfNodes = 4
-        topic = "foobar"
-        nodes = generateNodes(numberOfNodes, gossip = true, transport = transport)
-          .toGossipSub()
+  asyncTest "Outbound peers are marked correctly":
+    let
+      numberOfNodes = 4
+      topic = "foobar"
+      nodes = generateNodes(numberOfNodes, gossip = true).toGossipSub()
 
-      startNodesAndDeferStop(nodes)
+    startNodesAndDeferStop(nodes)
 
-      await connectNodes(nodes[0], nodes[1]) # Out
-      await connectNodes(nodes[0], nodes[2]) # Out
-      await connectNodes(nodes[3], nodes[0]) # In
-      subscribeAllNodes(nodes, topic, voidTopicHandler)
+    await connectNodes(nodes[0], nodes[1]) # Out
+    await connectNodes(nodes[0], nodes[2]) # Out
+    await connectNodes(nodes[3], nodes[0]) # In
+    subscribeAllNodes(nodes, topic, voidTopicHandler)
 
-      checkUntilTimeout:
-        nodes[0].mesh.outboundPeers(topic) == 2
-        nodes[0].getPeerByPeerId(topic, nodes[1].peerInfo.peerId).outbound == true
-        nodes[0].getPeerByPeerId(topic, nodes[2].peerInfo.peerId).outbound == true
-        nodes[0].getPeerByPeerId(topic, nodes[3].peerInfo.peerId).outbound == false
+    checkUntilTimeout:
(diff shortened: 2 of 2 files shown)


tests/mix/testconn.nim
```diff
@@ -174,3 +175,82 @@ suite "Mix Protocol":
...
+    # Read response - this should work correctly with the length prefix fix
+    let response = await conn.readLp(1024)
+    await conn.close()
+
+    # Verify the response was read correctly
+    let expectedResponse = "Response: " & testMessage
+    check string.fromBytes(response) == expectedResponse
+
+    # Verify response length is correct (not truncated)
+    check response.len == expectedResponse.len
```
vladopajic: if `response == expectedResponse`
how can
`response.len != expectedResponse.len` ?

that is, why is this check necessary?

tests/mix/testconn.nim
```diff
@@ -174,3 +175,82 @@ suite "Mix Protocol":
...
+        try:
+          # Read the request with readLp
+          let request = await conn.readLp(1024)
+          receivedAtDest.complete(request)
+
+          # Send response with writeLp (adds length prefix)
+          let response = "Response: " & string.fromBytes(request)
+          await conn.writeLp(response.toBytes())
+        except CatchableError as e:
+          discard
```
vladopajic:
```suggestion
          raiseAssert "Unexpected error: " & e.msg
```

tests/mix/testconn.nim
```diff
@@ -174,3 +175,82 @@ suite "Mix Protocol":
...
+      proto.codec = TestCodec
+      return proto
+
+    var mixProto: seq[MixProtocol] = @[]
+    for index, _ in enumerate(switches):
+      let proto = MixProtocol.new(index, switches.len, switches[index]).expect(
+          "should have initialized mix protocol"
+        )
+      # Register with readLp behavior - this should preserve length prefix
+      proto.registerDestReadBehavior(TestCodec, readLp(1024))
```
vladopajic: `1024` can is const candiate

tests/mix/testconn.nim
```diff
@@ -174,3 +175,82 @@ suite "Mix Protocol":
...
+      mixProto.add(proto)
+      switches[index].mount(proto)
+
+    let destNode = switches[^1]
+    let testProto = newLengthPrefixTestProtocol()
+    destNode.mount(testProto)
+
+    # Start all nodes
+    await switches.mapIt(it.start()).allFutures()
```
vladopajic, line 233: do nodes need to be stopped?

tests/mix/testconn.nim
```diff
@@ -174,3 +175,82 @@ suite "Mix Protocol":
...
     check response != 0.seconds
+
+  asyncTest "length-prefixed protocol - verify readLp fix":
+    ## This test verifies the fix for the length prefix bug where responses
+    ## from protocols using readLp() were losing their length prefix when
+    ## flowing back through the mix network.
+    const TestCodec = "/lengthprefix/test/1.0.0"
+
+    # Test message that will be sent and received
+    let testMessage = "Hello from mix protocol with length prefix!"
```
vladopajic: testMessage - would be fancy to use some message that inspires Logos movement, network states, cypherpunk's manifesto or anything that comes to your mind in similar context... this can be your personal message or quote from the book (eg. farewell to westphalia) that will be be your _message in the code_.

vladopajic: https://github.com/vacp2p/nim-libp2p/blob/cf0d80947561a256b30068eb95bbd6e4c7012f9c/tests/transports/connection_tests.nim#L17

we already started this practice here

tests/mix/testconn.nim
chaitanyaprem: Not needed, removing it

tests/mix/testconn.nim
chaitanyaprem: They are stopped in asyncTearDown above defined in start of test suite.

tests/mix/testconn.nim
chaitanyaprem: Like the idea, put in my own message!

PR #1807, opened 2025-10-28
chore(test): reduce global vars (rng)

vladopajic:
-reduced global vars in test mainly by utilizing `rng` template instead of ` newRng()`.
-second order consequence is reduce number of new `rng`s created (see https://github.com/vacp2p/nim-libp2p/pull/1807#discussion_r2469448331).

### DIFF
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -134,7 +134,7 @@ proc isConnected*(s: Switch, peerId: PeerId): bool {.public.} =
 
 proc disconnect*(
     s: Switch, peerId: PeerId
-) {.gcsafe, public, async: (raises: [CancelledError]).} =
+) {.public, async: (raises: [CancelledError]).} =
   ## Disconnect from a peer, waiting for the connection(s) to be dropped
   await s.connManager.dropPeer(peerId)
 


tests/kademlia/testroutingtable.nim
```diff
@@ -20,10 +20,9 @@ proc testKey*(x: byte): Key =
   buf[31] = x
   return @buf
 
-let rng = crypto.newRng()
-
 suite "routing table":
   const TargetBucket = 6
+  let rng = crypto.newRng()
```
arnetheduck:
```suggestion
  setup:
    let rng = crypto.newRng()
```

without `setup`, it's still a global variable

tests/pubsub/testmcache.nim
```diff
@@ -6,9 +6,8 @@ import
   ../../libp2p/
     [peerid, crypto/crypto, protocols/pubsub/mcache, protocols/pubsub/rpc/message]
 
-var rng = newRng()
-
 proc randomPeerId(): PeerId =
+  var rng = newRng()
```
arnetheduck: depending on the os, creating a new rng for every peer id might exhaust the randomness supply and stall the test

richard-ramos (approved):

PR #1816, opened 2025-10-29
test: too many global vars workaround

rlve:
Changes:
- revert splitting tests to not extend execution time
- wrap `suite` from `unittest2` in a proc to avoid global vars
- use `tests/utils/unittests` abstraction instead of `unittest2` directly
- fix imports

### DIFF
--- a/tests/pubsub/integration/testfloodsub.nim
+++ b/tests/pubsub/integration/testfloodsub.nim
@@ -107,7 +107,7 @@ suite "FloodSub Integration":
 
   asyncTest "FloodSub validation should fail":
     proc handler(topic: string, data: seq[byte]) {.async.} =
-      check false # if we get here, it should fail
+      raiseAssert "Handler should not be called when validation fails"
 
     let nodes = generateNodes(2)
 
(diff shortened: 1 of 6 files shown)


tests/utils/unittests.nim
```diff
@@ -28,3 +29,11 @@ template asyncTest*(name: string, body: untyped): untyped =
           body
       )()
     )
+
+template suite*(name: string, body: untyped): untyped =
```
vladopajic: please add comment describing why was this introduced

performance/scenarios_reliability.nim
```diff
@@ -9,8 +9,9 @@
 
 {.used.}
 
+import chronos
 import strformat
-import ../tests/utils/async_tests
+import ../tests/utils/unittests
```
vladopajic, line 14: `../tests/utils/unittests` -> `../tests/tools/unittests` 
`utils` -> `tools`

vladopajic: ah, it was there before :(

PR #1819, opened 2025-10-29
test(tools): add tools package

vladopajic:
this pr has no functional changes, just code moved to `tools`  package. 
imports have change to only import from `tool` things that they really need. 
old `helpers` file contained anything and everything, also exported things that were not needed by most files.

instead of 
```
import helpers
```
now tests should pick and choose bits to import
```
import ./tools/[unittests, compare, futures]
```


richard-ramos (approved):

performance/scenarios_performance.nim
```diff
@@ -9,15 +9,14 @@
 
 {.used.}
 
-import chronos
-import ../tests/utils/unittests
```
vladopajic, line 13: unnecessary deps

performance/scenarios_reliability.nim
```diff
@@ -9,9 +9,8 @@
 
 {.used.}
 
-import chronos
 import strformat
-import ../tests/utils/unittests
```
vladopajic, line 14: unnecessary deps

PR #1820, opened 2025-10-29
feat(c-bindings): libp2p thread and request handling

richard-ramos:
Adds the foundation code to create c-bindings for libp2p, and an example calling an async "Hello World".
(This code however does not expose yet the libp2p functionality thru c-bindings)

### DIFF
--- a/cbind/ffi_types.nim
+++ b/cbind/ffi_types.nim
@@ -1,3 +1,12 @@
+# Nim-LibP2P
+# Copyright (c) 2023-2025 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0 ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
 # FFI Types and Utilities
 #
 # This file defines the core types and utilities for the library's foreign
@@ -10,9 +19,12 @@ type Libp2pCallback* = proc(
   callerRet: cint, msg: ptr cchar, len: csize_t, userData: pointer
 ) {.cdecl, gcsafe, raises: [].}
 
-const RET_OK*: cint = 0
-const RET_ERR*: cint = 1
-const RET_MISSING_CALLBACK*: cint = 2
+type
+  RetCode* {.size: sizeof(cint).}= enum
+    RET_OK = 0,
+    RET_ERR = 1,
+    RET_MISSING_CALLBACK = 2
+
 
 ### End of exported types
 ################################################################################
--- a/cbind/libp2p.h
+++ b/cbind/libp2p.h
@@ -3,7 +3,13 @@
  *
  * This header provides the public API for libp2p
  *
- * TODO: add description
+ * Copyright (c) 2025 Status Research & Development GmbH
+ * Licensed under either of
+ * -  Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+ * - MIT license ([LICENSE-MIT](LICENSE-MIT))
+ * at your option.
+ * This file may not be copied, modified, or distributed except according to
+ * those terms.
  */
 #ifndef __libp2p__
 #define __libp2p__
--- a/cbind/libp2p.nim
+++ b/cbind/libp2p.nim
@@ -1,3 +1,12 @@
+# Nim-LibP2P
+# Copyright (c) 2023-2025 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0 ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
 {.pragma: exported, exportc, cdecl, raises: [].}
 {.pragma: callback, cdecl, raises: [], gcsafe.}
 {.passc: "-fPIC".}
@@ -24,7 +33,7 @@ template checkLibParams*(
   ctx[].userData = userData
 
   if isNil(callback):
-    return RET_MISSING_CALLBACK
+    return RET_MISSING_CALLBACK.cint
 
 template callEventCallback(ctx: ptr LibP2PContext, eventName: string, body: untyped) =
   ## This template invokes the event callback for internal events
@@ -57,10 +66,10 @@ proc handleRequest(
     content: pointer,
     callback: Libp2pCallback,
     userData: pointer,
-): cint =
+): RetCode =
   libp2p_thread.sendRequestToLibP2PThread(ctx, requestType, content, callback, userData).isOkOr:
     let msg = "libp2p error: " & $error
-    callback(RET_ERR, addr msg[0], cast[csize_t](len(msg)), userData)
+    callback(RET_ERR.cint, addr msg[0], cast[csize_t](len(msg)), userData)
     return RET_ERR
 
   return RET_OK
@@ -118,7 +127,7 @@ proc libp2p_new(
   ## Create the Libp2p thread that will keep waiting for req from the Client thread.
   var ctx = libp2p_thread.createLibP2PThread().valueOr:
     let msg = "Error in createLibp2pThread: " & $error
-    callback(RET_ERR, addr msg[0], cast[csize_t](len(msg)), userData)
+    callback(RET_ERR.cint, addr msg[0], cast[csize_t](len(msg)), userData)
     return nil
 
   ctx.userData = userData
@@ -147,13 +156,13 @@ proc libp2p_destroy(
 
   libp2p_thread.destroyLibP2PThread(ctx).isOkOr:
     let msg = "libp2p error: " & $error
-    callback(RET_ERR, addr msg[0], cast[csize_t](len(msg)), userData)
-    return RET_ERR
+    callback(RET_ERR.cint, addr msg[0], cast[csize_t](len(msg)), userData)
+    return RET_ERR.cint
 
   ## always need to invoke the callback although we don't retrieve value to the caller
-  callback(RET_OK, nil, 0, userData)
+  callback(RET_OK.cint, nil, 0, userData)
 
-  return RET_OK
+  return RET_OK.cint
 
 proc libp2p_set_event_callback(
     ctx: ptr LibP2PContext, callback: Libp2pCallback, userData: pointer
@@ -177,7 +186,7 @@ proc libp2p_hello(
     HelloRequest.createShared(HelloMsgType.HELLO),
     callback,
     userData,
-  )
+  ).cint
 
 ### End of exported procs
 ################################################################################
--- a/cbind/libp2p_thread/inter_thread_communication/requests/libp2p_hello_requests.nim
+++ b/cbind/libp2p_thread/inter_thread_communication/requests/libp2p_hello_requests.nim
@@ -1,3 +1,12 @@
+# Nim-LibP2P
+# Copyright (c) 2023-2025 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0 ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
 # Thread Request Example Implementation
 #
 # This file defines the `HelloRequest` type and its lifecycle, including memory
--- a/cbind/libp2p_thread/libp2p_thread.nim
+++ b/cbind/libp2p_thread/libp2p_thread.nim
@@ -1,3 +1,12 @@
+# Nim-LibP2P
+# Copyright (c) 2023-2025 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0 ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
 # LibP2P Thread Manager
 #
 # This file defines the `LibP2PContext` and associated logic to manage the LibP2P thread.
--- a/examples/cbindings.c
+++ b/examples/cbindings.c
@@ -35,11 +35,11 @@ void signal_cond() {
 }
 
 void event_handler(int callerRet, const char *msg, size_t len, void *userData) {
-  if (callerRet == RET_ERR) {
-    printf("Error: %s\n", msg);
-    exit(1);
-  } else if (callerRet == RET_OK) {
+  if (callerRet == RET_OK) {
     printf("Receiving event: %s\n", msg);
+  } else {
+    printf("Error(%d): %s\n", callerRet, msg);
+    exit(1);
   }
   signal_cond();
 }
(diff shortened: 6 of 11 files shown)


libp2p.nimble
```diff
@@ -146,3 +163,20 @@ task install_pinned, "Reads the lockfile":
...
+task examples, "Build and run examples":
+  exec "nimble install -y nimpng"
+  exec "nimble install -y nico --passNim=--skipParentCfg"
+  buildSample("examples_build", false, "--styleCheck:off") # build only
+
+  buildSample("examples_run", true)
+
+  buildLibrary "static", ""
+  exec "g++ -o build/cbindings ./examples/cbindings.c ./build/libp2p.a -pthread"
+  exec "./build/cbindings"
```
vladopajic: why have it here?

`task examples` should not even be here it should be in `/examples` dir.
similarly to how `task performance` and `task reliability` are in `/peformance` dir.

if we go down that route, let's have a separate nimble file for `buildLibrary` and whatever that is need here.

libp2p.nimble
```diff
@@ -56,6 +56,30 @@ proc buildSample(filename: string, run = false, extraFlags = "") =
...
+        "dll"
+      elif defined(macosx):
+        "dylib"
+      else:
+        "so"
+  let app = if libType == "static": "staticlib" else: "lib"
+
+  exec nimc & " " & lang & " --out:build/libp2p." & ext & " --threads:on --app:" & app &
+    " --opt:size --noMain --mm:refc --header --undef:metrics --nimMainPrefix:libp2p --nimcache:nimcache -d:asyncTimer=system " &
+    cfg & " " & flags & " library/libp2p.nim"
```
vladopajic: should we consider having `libp2_library.nim` file where we add all bits and picies for library and include it in `libp2p.nimble`?

we can have `libp2p_test.nim` similarly. 

we end up with better organization of `libp2p.nimble`.

libp2p.nimble
```diff
@@ -10,7 +10,7 @@ skipDirs = @["tests", "examples", "Nim", "tools", "scripts", "docs"]
 requires "nim >= 2.0.0",
   "nimcrypto >= 0.6.0", "dnsclient >= 0.3.0 & < 0.4.0", "bearssl >= 0.2.5",
   "chronicles >= 0.11.0", "chronos >= 4.0.4", "metrics", "secp256k1", "stew >= 0.4.2",
-  "websock >= 0.2.1", "unittest2", "results", "quic >= 0.5.0",
+  "websock >= 0.2.1", "unittest2", "results", "quic >= 0.5.0", "taskpools >= 0.1.0",
```
vladopajic: no pinned file for `taskpools` ?

examples/cbindings.c
```diff
@@ -0,0 +1,62 @@
+#include "../library/libp2p.h"
```
vladopajic: would be nicer to have `examples/cbindings` dir and have all things here. it would be easier for user to see what things are needed. 

this dir should have files:
- `cbindings.nimble` with task `buildLibrary`
- `cbindings.c` (how about call this differently `main.c`).
- others..

maybe `examples/cbindings` ->  `examples/cbind`?

vladopajic: `#include "../library/libp2p.h"` -> `#include "../clib/libp2p.h"`

`nim-libp2p` is already a library, `clib` or even `cbind` would be more suitable name.  (+1 `cbind`)

cbind/libp2p.h
```diff
@@ -0,0 +1,39 @@
+/**
```
vladopajic, line 1:
```
# Nim-LibP2P
# Copyright (c) 2025 Status Research & Development GmbH
```
... in other files as well

cbind/libp2p.h
```diff
@@ -0,0 +1,39 @@
...
+ * libp2p.h - C Interface for nim-libp2p
```
vladopajic, line 2: how that we have `c` files, should we invest in some tooling: formatters etc..

library/events/json_base_event.nim
```diff
@@ -0,0 +1,14 @@
+# JSON Event definition
```
vladopajic: i would suggest not to add it unless we see how it is being used

library/alloc.nim
```diff
@@ -0,0 +1,78 @@
...
+  return ret
+
+proc allocSharedSeq*[T](s: seq[T]): SharedSeq[T] =
+  let data = allocShared(sizeof(T) * s.len)
+  if s.len != 0:
+    copyMem(data, unsafeAddr s[0], s.len)
+  return (cast[ptr UncheckedArray[T]](data), s.len)
+
+proc deallocSharedSeq*[T](s: var SharedSeq[T]) =
+  if not s.data.isNil():
```
vladopajic: early return perhaps?

library/alloc.nim
```diff
@@ -0,0 +1,78 @@
+# Shared Memory Management
+#
+# This file implements procedures for allocating and deallocating shared memory
+# used for communication between client and library threads.
+
+## Can be shared safely between threads
+type SharedSeq*[T] = tuple[data: ptr UncheckedArray[T], len: int]
```
vladopajic: why not use `struct`? `struct` its used by default in `nim-libp2p` , unless there is some reason why `tuple` is used?

library/alloc.nim
```diff
@@ -0,0 +1,78 @@
...
+
+    deallocShared(s.data)
+  s.len = 0
+
+proc toSeq*[T](s: SharedSeq[T]): seq[T] =
+  ## Creates a seq[T] from a SharedSeq[T]. No explicit dealloc is required
+  ## as req[T] is a GC managed type.
+  var ret = newSeq[T]()
+  for i in 0 ..< s.len:
+    ret.add(s.data[i])
```
vladopajic: `newSeq` -> `newSeqUninit`?

vladopajic: `copyMem` instead of `for` loop?

examples/cbindings.c
```diff
@@ -0,0 +1,62 @@
...
+
+void signal_cond() {
+  pthread_mutex_lock(&mutex);
+  callback_executed = 1;
+  pthread_cond_signal(&cond);
+  pthread_mutex_unlock(&mutex);
+}
+
+void event_handler(int callerRet, const char *msg, size_t len, void *userData) {
+  if (callerRet == RET_ERR) {
```
gmelodie: I have no context here, but I _think_ this should be `callerRet != RET_OK` to allow for different error types

examples/cbindings.c
```diff
@@ -0,0 +1,62 @@
...
+    printf("Error: %s\n", msg);
+    exit(1);
```
gmelodie: maybe:
```suggestion
    exit(callerRet);
```

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_hello_requests.nim
```diff
@@ -0,0 +1,37 @@
+# Thread Request Example Implementation
```
gmelodie, line 10: The name of this file's path is humongous

gmelodie, line 10: I think all the `hello` stuff should be under `examples/cbindings/`, no? like `examples/cbindings/hello`

examples/cbindings.c
gmelodie: +1

gmelodie: +1 cbind or cbindings

cbind/libp2p.h
gmelodie: +1

cbind/libp2p.h
gmelodie: https://github.com/vacp2p/nim-libp2p/issues/1836

libp2p.nimble
```diff
@@ -56,6 +56,30 @@ proc buildSample(filename: string, run = false, extraFlags = "") =
     exec "./examples/" & filename.toExe
   rmFile "examples/" & filename.toExe
 
+proc buildLibrary(libType: string, params = "") =
```
gmelodie: `buildCBindings`

libp2p.nimble
gmelodie: This is problem for another ~~day~~ PR, no? The code here is just being consistent with how we currently do things.

examples/cbindings.c
```diff
@@ -0,0 +1,62 @@
...
+  } else if (callerRet == RET_OK) {
+    printf("Receiving event: %s\n", msg);
+  }
```
gmelodie:
```suggestion
  printf("Receiving event: %s\n", msg);
```

library/libp2p_thread/libp2p_thread.nim
```diff
@@ -0,0 +1,149 @@
...
+
+  # This is the worker body. This runs the LibP2P instance
+  # and attends library user requests
+
+  var libp2p: LibP2P
+
+  while true:
+    await ctx.reqSignal.wait()
+
+    if ctx.running.load == false:
```
gmelodie:
```suggestion
    if not ctx.running.load:
```

cbind/ffi_types.nim
```diff
@@ -0,0 +1,35 @@
...
+################################################################################
+### Exported types
+
+type Libp2pCallback* = proc(
+  callerRet: cint, msg: ptr cchar, len: csize_t, userData: pointer
+) {.cdecl, gcsafe, raises: [].}
+
+const RET_OK*: cint = 0
+const RET_ERR*: cint = 1
+const RET_MISSING_CALLBACK*: cint = 2
```
gmelodie: Maybe `Ret` could be an `enum`

cbind/libp2p_thread/libp2p_thread.nim
```diff
@@ -0,0 +1,149 @@
...
+type LibP2PContext* = object
+  thread: Thread[(ptr LibP2PContext)] # The running thread that executes the LibP2P loop
+  lock: Lock # Used to serialize access to the SP channel
+  reqChannel: ChannelSPSCSingle[ptr LibP2PThreadRequest]
+  reqSignal: ThreadSignalPtr # To notify the LibP2P Thread that a request is ready
+  reqReceivedSignal: ThreadSignalPtr
+    # To notify the Client thread that the request was received
+  userData*: pointer
+  eventCallback*: pointer
+  eventUserData*: pointer
```
gmelodie, line 37: What if these were `Option`s so that we could use `valueOr` later instead of checking for nil?

cbind/libp2p.nim
```diff
@@ -0,0 +1,183 @@
...
+
+template callEventCallback(ctx: ptr LibP2PContext, eventName: string, body: untyped) =
+  ## This template invokes the event callback for internal events
+  if isNil(ctx[].eventCallback):
+    error eventName & " - eventCallback is nil"
+    return
+
+  if isNil(ctx[].eventUserData):
+    error eventName & " - eventUserData is nil"
+    return
```
gmelodie, line 46: If we follow my suggestion from before, we could use `valueOr` here

cbind/libp2p.nim
```diff
@@ -0,0 +1,183 @@
...
+      cast[Libp2pCallback](ctx[].eventCallback)(
+        RET_OK, addr event[0], cast[csize_t](len(event)), ctx[].eventUserData
+      )
+    except Exception, CatchableError:
+      let msg =
+        "Exception " & eventName & " when calling 'eventCallBack': " &
+        getCurrentExceptionMsg()
+      cast[Libp2pCallback](ctx[].eventCallback)(
+        RET_ERR, addr msg[0], cast[csize_t](len(msg)), ctx[].eventUserData
+      )
```
gmelodie, line 60: If we used `valueOr` before, this could be a bit more straightforward

```suggestion
  foreignThreadGc:
    try:
      let event = body
      cast[Libp2pCallback](eventCallback)(
        RET_OK, addr event[0], cast[csize_t](len(event)), eventUserData
      )
    except Exception, CatchableError:
      let msg =
        "Exception " & eventName & " when calling 'eventCallBack': " &
        getCurrentExceptionMsg()
      cast[Libp2pCallback](eventCallback)(
        RET_ERR, addr msg[0], cast[csize_t](len(msg)), eventUserData
      )
```

library/libp2p.nim
```diff
@@ -0,0 +1,183 @@
...
+
+  let retCode = handleRequest(
+    ctx,
+    RequestType.LIFECYCLE,
+    LifecycleRequest.createShared(LifecycleMsgType.CREATE_LIBP2P, appCallbacks),
+    callback,
+    userData,
+  )
+
+  if retCode == RET_ERR:
```
gmelodie:
```suggestion
  if retCode != RET_OK:
```

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_hello_requests.nim
richard-ramos: This `hello` i'm keeping it here temporarily as a guidance of what to do to expose a libp2p function. This could be for example a lifecycle request to start / stop node. In the next PR i'll remove it once i replace this by real libp2p functionality

libp2p.nimble
richard-ramos: I'll open an issue to track this.

richard-ramos: https://github.com/vacp2p/nim-libp2p/issues/1854

cbind/libp2p_thread/libp2p_thread.nim
richard-ramos: Does it make sense to have an Option[pointer]?  you could still do this:
```nim
var x pointer: = nil
var myOpt: Option[pointer] = Opt.some(x)
```
Then while `myOpt.isSome == true`, its value would still be `nil`

library/alloc.nim
richard-ramos: I copied from https://github.com/logos-co/nim-library-template  but I agree the struct is more elegant.

libp2p.nimble
richard-ramos: https://github.com/vacp2p/nim-libp2p/issues/1856

examples/cbindings.c
richard-ramos: Maybe can be done further along the development of this task. For now this cbindings example only requires that .c file and it's built as part of the nimble examples task. This could be done with https://github.com/vacp2p/nim-libp2p/issues/1854

cbind/libp2p_thread/libp2p_thread.nim
gmelodie: I think you're right, but we should use `pointer` only where there are no other options because of pointer issues (nil pointer dereference, double free, etc...). Besides creating a type to wrap the pointers and ensure safe operations through an API, I don't see any other option here (maybe @vladopajic has thoughts?). I think we can proceed as is, but be mindful in the future.

gmelodie (approved):

cbind/libp2p_thread/libp2p_thread.nim
vladopajic: `Opt.some(nil)` should be undefined behavior or it should `Opt.None(type of receiver)` -  if possible then later.

vladopajic (approved):

.pinned
```diff
@@ -19,3 +19,4 @@ websock;https://github.com/status-im/nim-websock@#35ae76f1559e835c80f9c1a3943bf9
 zlib;https://github.com/status-im/nim-zlib@#daa8723fd32299d4ca621c837430c29a5a11e19a
 jwt;https://github.com/vacp2p/nim-jwt@#18f8378de52b241f321c1f9ea905456e89b95c6f
 bearssl_pkey_decoder;https://github.com/vacp2p/bearssl_pkey_decoder@#21dd3710df9345ed2ad8bf8f882761e07863b8e0
+taskpools;https://github.com/status-im/nim-taskpools@#9e8ccc754631ac55ac2fd495e167e74e86293edb
```
vladopajic, line 22: hmm this `taskpools` should be in `./cbind/cbind.nimble`

cbind/libp2p_thread/libp2p_thread.nim
vladopajic: but not sure why library is not implemented in that way

Issue #1708, opened 2025-09-21
mix: determine what should happen if in `anonymizeLocalProtocolSend` an invalid multiaddress/peer is chosen


richard-ramos:
See: https://github.com/vacp2p/nim-libp2p/blob/9a45878c5d92efeef40e9f47b399ed9b263c9e4a/libp2p/protocols/mix/mix_protocol.nim#L289

PR #1821, opened 2025-10-30
fix: when an invalid multiaddr is encountered skip that node in mix path selection

chaitanyaprem:
Closes https://github.com/vacp2p/nim-libp2p/issues/1708

### DIFF
--- a/libp2p/protocols/mix/mix_protocol.nim
+++ b/libp2p/protocols/mix/mix_protocol.nim
@@ -552,8 +552,7 @@ proc anonymizeLocalProtocolSend*(
 
   var nextHopAddr: MultiAddress
   var nextHopPeerId: PeerId
-  var i = 0
-  while i < PathLength:
+  while hop.len < PathLength:
     if availableIndices.len == 0:
       mix_messages_error.inc(labelValues = ["Entry", "LOW_MIX_POOL"])
       return err("Ran out of available mix nodes while constructing path")
@@ -570,7 +569,7 @@ proc anonymizeLocalProtocolSend*(
       continue
 
     # Last hop will be the exit node that will forward the request
-    if i == PathLength - 1:
+    if hop.len == PathLength - 1:
       case destination.kind
       of ForwardAddr:
         # Last hop will be the exit node that will fwd the request
@@ -580,7 +579,7 @@ proc anonymizeLocalProtocolSend*(
         exitPeerId = destination.peerId
         randPeerId = destination.peerId
 
-    debug "Selected mix node: ", indexInPath = i, peerId = randPeerId
+    debug "Selected mix node: ", indexInPath = hop.len, peerId = randPeerId
 
     # Extract multiaddress, mix public key, and hop
     let (peerId, multiAddr, mixPubKey, _) =
@@ -600,15 +599,13 @@ proc anonymizeLocalProtocolSend*(
     # Only add to path after validation succeeds
     publicKeys.add(mixPubKey)
 
-    if i == 0:
+    if hop.len == 0:
       nextHopAddr = multiAddr
       nextHopPeerId = peerId
 
-    hop.add(Hop.init(multiAddrBytes))
-
     # Compute delay
     let delayMillisec =
-      if i != PathLength - 1:
+      if hop.len != PathLength - 1:
         cryptoRandomInt(mixProto.rng, 3).valueOr:
           mix_messages_error.inc(labelValues = ["Entry", "NON_RECOVERABLE"])
           return err(fmt"Failed to generate random number: {error}")
@@ -617,7 +614,7 @@ proc anonymizeLocalProtocolSend*(
 
     delay.add(@(delayMillisec.uint16.toBytesBE()))
 
-    i.inc() # Only increment when we successfully add a hop to the path
+    hop.add(Hop.init(multiAddrBytes))
 
   # Encode destination
   let destHop =
--- a/tests/mix/testconn.nim
+++ b/tests/mix/testconn.nim
@@ -9,12 +9,13 @@
 
 {.used.}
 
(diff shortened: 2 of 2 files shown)


libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -552,7 +552,12 @@ proc anonymizeLocalProtocolSend*(
 
   var nextHopAddr: MultiAddress
   var nextHopPeerId: PeerId
-  for i in 0 ..< PathLength:
+  var i = 0
+  while i < PathLength:
```
vladopajic: could condition change to:

pseudo code:
```
while path.len < PathLength:
```
to avoid `var i`, and especially to avoid `i.inc()` (for vary same reason it has comment) ?

`path.len` i assume is `hop.len` ?

tests/mix/testconn.nim
```diff
@@ -253,3 +253,86 @@ suite "Mix Protocol":
...
+        )
+      mixProto.add(proto)
+      switches[index].mount(proto)
+
+    # Now inject invalid node into sender's (node 0) pool
+    # Include enough valid nodes so that even after invalid node is removed,
+    # we still have sufficient nodes for PathLength = 3
+    # Using 6 valid nodes + 1 invalid = 7 total, which gives enough room
+    var nodePool = initTable[PeerId, MixPubInfo]()
+    for i in 1 ..< 7: # Load 6 valid nodes (indices 1, 2, 3, 4, 5, 6)
```
vladopajic: 6 should be const, `validNodesCount`

tests/mix/testconn.nim
```diff
@@ -253,3 +253,86 @@ suite "Mix Protocol":
...
+      let pubInfo = MixPubInfo.readFromFile(i).expect("could not read pub info")
+      nodePool[pubInfo.peerId] = pubInfo
+
+    nodePool[invalidPeerId] = invalidPubInfo
+    mixProto[0].setNodePool(nodePool)
+
+    # Verify pool has 6 valid + 1 invalid = 7 nodes
+    check mixProto[0].getNodePoolSize() == 7
```
vladopajic: `check mixProto[0].getNodePoolSize() == validNodesCount + 1`

tests/mix/testconn.nim
```diff
@@ -253,3 +253,86 @@ suite "Mix Protocol":
...
+
+    # Send messages in a loop until invalid node is encountered and removed
+    # With 6 valid + 1 invalid nodes and PathLength=3, we need multiple attempts
+    let testPayload = "test message".toBytes()
+    var initialPoolSize = mixProto[0].getNodePoolSize()
+
+    # Try up to 20 times to encounter the invalid node
+    # Stop if pool size goes below 3 (can't construct path anymore)
+    for attempt in 0 ..< 20:
+      if mixProto[0].getNodePoolSize() < 3:
```
vladopajic: why 3 where does 3 comes from?

tests/mix/testconn.nim
```diff
@@ -253,3 +253,86 @@ suite "Mix Protocol":
...
+      await conn.get().writeLp(testPayload)
+      discard await receivedMessage.wait(5.seconds)
+      await conn.get().close()
+
+      # Check if invalid node was removed
+      if mixProto[0].getNodePoolSize() < initialPoolSize:
+        break
+
+    # Verify invalid node was removed from pool (should be 6 valid nodes remaining)
+    check mixProto[0].getNodePoolSize() == 6
```
vladopajic: ` check mixProto[0].getNodePoolSize() == validNodesCount`

tests/mix/testconn.nim
vladopajic: if `switches` is changed to `3`, can `6` still be `6`?
please be mindful of how 6 value comes into being, if it depends on some other value please use it as deriving it from that other value.

vladopajic: `const validNodesCount = 6` is nice
but 
`const validNodesCount = switchies.len / 2 # use only half valid nodes`
is even better.

tests/mix/testconn.nim
```diff
@@ -253,3 +253,86 @@ suite "Mix Protocol":
...
+    # Stop if pool size goes below 3 (can't construct path anymore)
+    for attempt in 0 ..< 20:
+      if mixProto[0].getNodePoolSize() < 3:
+        break
+
+      let conn = mixProto[0].toConnection(
+        MixDestination.init(switches[1].peerInfo.peerId, switches[1].peerInfo.addrs[0]),
+        NoReplyProtocolCodec,
+        MixParameters(expectReply: Opt.none(bool), numSurbs: Opt.none(uint8)),
+      )
```
richard-ramos, line 347: If no specific parameters are to be used, it should be possible to skip the last parameter entirely
```suggestion
      let conn = mixProto[0].toConnection(
        MixDestination.init(switches[1].peerInfo.peerId, switches[1].peerInfo.addrs[0]),
        NoReplyProtocolCodec),
      )
```

tests/mix/testconn.nim
chaitanyaprem: 3 is min mix nodes required in the pool as that is the default path length and if it goes below the invocation fails with a different error, hence this protection. 
Let me see if there is already a const defined for it and will reuse it.

tests/mix/testconn.nim
chaitanyaprem: > `const validNodesCount = switchies.len / 2 # use only half valid nodes`

don't want to do that blindly, because what switches length get smaller, trying to take a different approach.

Thanks for this catch, as the changes here will make the test more robust.

libp2p/protocols/mix/mix_protocol.nim
chaitanyaprem: Thanks! Will include this change.

PR #1835, opened 2025-11-03
ci: linters names fix

vladopajic:
will requires to change repository pr checks from "Linters / nph" to "Linters / Linters"

### DIFF
--- a/tests/mix/testmixnode.nim
+++ b/tests/mix/testmixnode.nim
@@ -19,7 +19,7 @@ suite "Mix Node Tests":
 
   setup:
     const count = 5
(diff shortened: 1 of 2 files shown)


.github/workflows/linters.yml
```diff
@@ -10,15 +10,15 @@ concurrency:
 
 jobs:
   nph:
-    name: NPH
+    name: Linters
```
vladopajic, line 13: or any other suggestion?
but NPH doesn't make sense now that we are also running other checks here

richard-ramos (approved):

PR #1838, opened 2025-11-04
test(transports): add parallel stream tests

rlve:
Changes:
- add tests:
  - `"stream with multiple parallel writes"`
  - `"server with multiple parallel connections"`
- invert client and server is some tests to get rid of an initial write and QUIC disclaimer. **Note:** I left it on purpose in one scenario (`"client writes after EOF"`).
```
        # Quic does not signal the server about new stream
        # The peer can only accept the stream after data
        # has been sent to the stream, so we send a "hello"
        # just so the server handlers is triggered
```

### DIFF
--- a/tests/transports/stream_tests.nim
+++ b/tests/transports/stream_tests.nim
@@ -229,6 +229,7 @@ template streamTransportTest*(
     const chunkSize = 256 * 1024
     const parallelWrites = 10
     let message = newData(messageSize)
+    var serverHandlerDone = newFuture[void]()
 
     proc serverStreamHandler(stream: Connection) {.async: (raises: []).} =
       noExceptionWithStreamClose(stream):
@@ -249,6 +250,8 @@ template streamTransportTest*(
             if receivedData[offset + j] != expectedValue:
               break # stop on first mismatch (not to pollute stdout)
 
+        serverHandlerDone.complete()
+
     proc clientStreamHandler(stream: Connection) {.async: (raises: []).} =
       noExceptionWithStreamClose(stream):
         var writeFuts: seq[Future[void]] = @[]
@@ -258,6 +261,7 @@ template streamTransportTest*(
           let fut = stream.write(message)
           writeFuts.add(fut)
         await allFutures(writeFuts)
+        await serverHandlerDone
 
     await runSingleStreamScenario(
       ma, transportProvider, streamProvider, serverStreamHandler, clientStreamHandler


tests/transports/stream_tests.nim
```diff
@@ -220,25 +205,140 @@ template streamTransportTest*(
...
+              await stream.readExactly(addr buffer, 1)
+              await stream.write(@buffer)
+
+              # Signal that this stream handler is done
+              serverStreamHandlerFuts[handlerIndex].complete()
+
+          let startStreamHandlerAndCleanup = proc() {.async.} =
+            let muxerTask = muxer.handle()
+
+            await muxerTask
```
vladopajic:
```suggestion
            await muxer.handle()
```

tests/transports/stream_tests.nim
```diff
@@ -220,25 +205,140 @@ template streamTransportTest*(
...
+      var futs: seq[Future[void]]
+      for i in 0 ..< numConnections:
+        let conn = await server.accept()
+        let muxer = streamProvider(server, conn)
+
+        # Use a proc to properly capture loop index
+        proc setupConnection(conn: Connection, muxer: Muxer, handlerIndex: int) =
+          muxer.streamHandler = proc(stream: Connection) {.async: (raises: []).} =
+            noExceptionWithStreamClose(stream):
+              var buffer: array[1, byte]
```
vladopajic: 1 byte... better have larger data. because sending 1 byte is instant, so connections will never exist simultaneously. 

https://github.com/vacp2p/nim-quic/blob/main/tests/quic/testStreams.nim#L716-L717

vladopajic: also something that might be interesting to do is:

pseudocode:
```nim
while receivedData < dataSize
 receivedData += await stream.readExactly(...)
 away sleepAsync(50.milisecons)
```

that is simulate delay when reading data, because it should give control to other future. so that tests reads bits from different connection every time.

vladopajic: not sure how delay is going to behave, but would be nice to test

vladopajic: test can also assert that data is read from different connection:

pseudocode:
- whenever reading data in server hander, add `handlerIndex` (aka client ID) to some list
- finally, when tests completes assert that elements are not grouped:
  -  grouped: [1, 1, 1, 2, 2, 2, 3, 3, 3]; [1, 1, 1, 3, 3, 3, 2, 2, 2]
  - not grouped: : [1, 2, 3, 1, 2, 2, 1, 3, 3] <- means connection handler has happened in "parallel".

tests/transports/stream_tests.nim
```diff
@@ -220,25 +205,140 @@ template streamTransportTest*(
...
+      await allFutures(futs)
+
+    proc runClient(server: Transport, connectionId: int) {.async.} =
+      let client = transportProvider()
+      let conn = await client.dial(server.addrs[0])
+      let muxer = streamProvider(client, conn)
+      let muxerTask = muxer.handle()
+
+      let stream = await muxer.newStream()
+      await stream.write(@[byte(connectionId)])
```
vladopajic: note, comment from above can be inverted. that is server can read data without delay. and delay can be introduced on client side:

```nim
for i in chunksCount:
 await stream.write(newData(i))
 away asyncSleep(100.milliseconds)
```

the delay on client side should introduce blocking at server when chunk is read (server waits for next chunk), this will cause that future to get blocked which will give control to other future.

vladopajic:
aslo, would be interesting to have case with single connection and there are many streams.
and bonus, same test but many connections and many streams (not sure if necessary). 
can be other prs, just heads up.

tests/transports/stream_tests.nim
rlve: I added writing and reading in chunks with random delays and some attempt to assert that the reading order is not grouped.

Example:
```
  [Test   ]           server with multiple parallel connections
@[0, 1, 2, 3, 4, 0, 4, 2, 1, 3, 2, 0, 1, 4, 3, 2, 4, 3, 0, 1, 2, 4, 3, 1, 0, 1, 2, 3, 4, 4, 0, 1, 2, 3, 0, 1, 0, 4, 2, 1, 2, 4, 3, 0, 0,
1, 3, 4, 2, 1, 0, 3, 2, 1, 4, 2, 0, 3, 2, 3, 4, 1, 0, 2, 3, 0, 4, 1, 2, 4, 3, 0, 1, 2, 4, 0, 3, 1, 4, 2, 3, 0, 3, 1, 4, 0, 2, 4, 0, 2, 4,
3, 1, 0, 4, 1, 3, 0, 2, 2, 3, 1, 4, 1, 3, 0, 0, 2, 4, 0, 1, 3, 0, 4, 2, 3, 0, 4, 2, 1, 3, 0, 4, 1, 2, 3, 0, 1, 4, 0, 3, 1, 0, 2, 4, 3, 1,
4, 0, 2, 2, 3, 1, 4, 3, 0, 4, 2, 1, 3, 1, 1, 4, 3, 2, 4, 1, 3, 2, 2]
  [OK     ] (  2.12s) server with multiple parallel connections
```

https://github.com/vacp2p/nim-libp2p/pull/1838/commits/cd83c06101a835ca9cd1ff993ea48e26328ceef8

tests/transports/stream_tests.nim
```diff
@@ -220,25 +205,167 @@ template streamTransportTest*(
...
 
     await runSingleStreamScenario(
       ma, transportProvider, streamProvider, serverStreamHandler, clientStreamHandler
     )
+
+  asyncTest "server with multiple parallel connections":
+    let ma = @[MultiAddress.init(address).tryGet()]
+    const messageSize = 2 * 1024
+    const chunkSize = 64
+    const chunkCount = 32
```
vladopajic: `messageSize = chunkSize * chunkCount`

tests/transports/stream_tests.nim
```diff
@@ -220,25 +205,167 @@ template streamTransportTest*(
...
+              while receivedData.len < messageSize:
+                var chunk = newSeq[byte](chunkSize)
+                let bytesRead = await stream.readOnce(addr chunk[0], chunkSize)
+                if bytesRead == 0:
+                  break
+                receivedData.add(chunk[0 ..< bytesRead])
+                serverReadOrder.add(chunk[0])
+                # Random delay between reads (20-100ms)
+                await sleepAsync(rand(20 .. 100).milliseconds)
+
```
vladopajic, line 307: would be nice to assert whole received data

check receivedData == newData(messageSize, id)

tests/transports/stream_tests.nim
```diff
@@ -220,25 +205,167 @@ template streamTransportTest*(
...
+              # Read data in chunks with random delay
+              var receivedData: seq[byte] = @[]
+              while receivedData.len < messageSize:
+                var chunk = newSeq[byte](chunkSize)
+                let bytesRead = await stream.readOnce(addr chunk[0], chunkSize)
+                if bytesRead == 0:
+                  break
+                receivedData.add(chunk[0 ..< bytesRead])
+                serverReadOrder.add(chunk[0])
+                # Random delay between reads (20-100ms)
```
vladopajic, line 303: `# Random delay between reads (20-100ms)`
comments are better when they answer on **_why?_**
`# Random delay between reads (20-100ms), will block current future (stream handler) and gives opportunity to next future to execute. Doing this improves likelihood of parallel data transition on the muxer.`

rlve: https://github.com/vacp2p/nim-libp2p/pull/1838/commits/926878399a0f0d5e499e0fe3494c9f2028afb2d6

tests/transports/stream_tests.nim
rlve: https://github.com/vacp2p/nim-libp2p/pull/1838/commits/926878399a0f0d5e499e0fe3494c9f2028afb2d6

PR #1845, opened 2025-11-05
chore(pubusb): improvements

vladopajic:


libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -473,8 +473,10 @@ method handleConn*(
   proc handler(
       peer: PubSubPeer, data: seq[byte]
   ): Future[void] {.async: (raises: []).} =
-    # call pubsub rpc handler
-    p.rpcHandler(peer, data)
```
vladopajic, line 477:
```
method rpcHandler*(
    p: PubSub, peer: PubSubPeer, data: seq[byte]
): Future[void] {.
    base, async: (raises: [CancelledError, PeerMessageDecodeError, PeerRateLimitError])
.} =
```

how did this even compiled here?

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -185,7 +185,7 @@ method init*(f: FloodSub) =
     try:
       await f.handleConn(conn, proto)
     except CancelledError as exc:
-      trace "Unexpected cancellation in floodsub handler", conn, description = exc.msg
+      trace "floodsub handler cancelled", conn
```
vladopajic, line 188: description is not needed since it is CancelledError (there is nothing to describe here).

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -470,22 +470,28 @@ method handleConn*(
...
-    await peer.handle(conn) # spawn peer read loop
-    trace "pubsub peer handler ended", conn
+    peer.handler = peerHandler
+    await peer.runHandleLoop(conn)
   except CancelledError as exc:
     raise exc
-  except PeerMessageDecodeError as exc:
-    trace "exception ocurred in pubsub handle", description = exc.msg, conn
   finally:
     await conn.closeWithEOF()
```
vladopajic, line 496: not sure if this ever had any effect, because `await conn.close()` was closed [here](https://github.com/vacp2p/nim-libp2p/pull/1845/files#diff-cd23f9cd2a0f9c037ef0b4245ee7d06c09a24985c8ee5ed39fff496f705570c3L254)

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -231,34 +232,29 @@ proc sendObservers(p: PubSubPeer, msg: var RPCMsg) =
...
-        conn, peer = p, closed = conn.closed, description = exc.msg
-    finally:
-      await conn.close()
-  except CancelledError as e:
-    # This is top-level procedure which will work as separate task, so it
-    # do not need to propagate CancelledError.
-    trace "Unexpected cancellation in PubSubPeer.handle", description = e.msg
-  finally:
+  defer:
     debug "exiting pubsub read loop", conn, peer = p, closed = conn.closed
```
richard-ramos, line 241:
```suggestion
  defer:
debug "exiting pubsub read loop", conn, peer = p, closed = conn.closed
conn.close()
```

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -470,22 +470,28 @@ method handleConn*(
   ##    that we're interested in
   ##
```
vladopajic, line 472: `handleConn` is not really covered with test. cases:
- when `PeerRateLimitError` is recived assert that loop ends
- when `PeerMessageDecodeError` assert that loop is still running

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -231,34 +232,29 @@ proc sendObservers(p: PubSubPeer, msg: var RPCMsg) =
...
-    try:
-      while not conn.atEof:
-        trace "waiting for data", conn, peer = p, closed = conn.closed
-
-        var data = await conn.readLp(p.maxMessageSize)
-        trace "read data from peer",
-          conn, peer = p, closed = conn.closed, data = data.shortLog
-
-        await p.handler(p, data)
-        data = newSeqUninit[byte](0) # Release memory
```
richard-ramos, line 246: Hm. should we add this back?  The comment of this `Releasing memory` makes me worry a bit. I admit that i do not know if this line does have an effect, tho.

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -473,8 +473,10 @@ method handleConn*(
   proc handler(
       peer: PubSubPeer, data: seq[byte]
   ): Future[void] {.async: (raises: []).} =
-    # call pubsub rpc handler
-    p.rpcHandler(peer, data)
+    try:
+      await p.rpcHandler(peer, data)
+    except CatchableError as e: # must catch everything
```
gmelodie, line 478: This should be
```suggestion
    except PeerMessageDecodeError as e:
      ...
    except PeerRateLimitError as e:
      ...
```

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -473,8 +473,10 @@ method handleConn*(
   proc handler(
       peer: PubSubPeer, data: seq[byte]
   ): Future[void] {.async: (raises: []).} =
```
gmelodie, line 475:
```suggestion
  ): Future[void] {.async: (raises: [CancelledError]).} =
```

libp2p/protocols/pubsub/pubsubpeer.nim
vladopajic: not sure where and if to put that close, connection is already closed here:
https://github.com/vacp2p/nim-libp2p/pull/1845#discussion_r2494464036

libp2p/protocols/pubsub/pubsubpeer.nim
vladopajic: variable expires in next loop.. if we have issue here with this we have in all other places

vladopajic: in addition to this `handler` now has `sink` to data.

libp2p/protocols/pubsub/pubsub.nim
gmelodie: If this compiles, it makes me wonder if we should specify `raises` on base methods at all, since the compiler doesn't seem to enforce these rules.

libp2p/protocols/pubsub/pubsub.nim
vladopajic: this is old code

libp2p/protocols/pubsub/pubsubpeer.nim
richard-ramos: Yeah, as discussed on DM, i'm closing this comment as just having the closeWithEOF is enough

PR #1846, opened 2025-11-05
fix(pubsub): use negotiated protocol in getConn if available

richard-ramos:
Also bumps nim-quic

### DIFF
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -102,7 +102,7 @@ method unsubscribePeer*(f: FloodSub, peer: PeerId) =
   procCall PubSub(f).unsubscribePeer(peer)
 
 method rpcHandler*(
-    f: FloodSub, peer: PubSubPeer, data: seq[byte]
+    f: FloodSub, peer: PubSubPeer, data: sink seq[byte]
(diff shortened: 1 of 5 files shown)


vladopajic (approved):

tests/transports/stream_tests.nim
```diff
@@ -155,6 +161,9 @@ template streamTransportTest*(
 
     proc serverStreamHandler(stream: Connection) {.async: (raises: []).} =
       noException(stream):
+        var hello: array[1, byte]
+        await stream.readExactly(addr hello, 1)
```
vladopajic:
```suggestion
        var initData: array[1, byte]
        await stream.readExactly(addr initData, 1)
```

tests/transports/stream_tests.nim
```diff
@@ -111,6 +113,12 @@ template streamTransportTest*(
 
     proc clientStreamHandler(stream: Connection) {.async: (raises: []).} =
       noException(stream):
+        # Quic does not signal the server about new stream
+        # The peer can only accept the stream after data
+        # has been sent to the stream, so we send a "hello"
+        # just so the server handlers is triggered
+        await stream.write(newData(1))
```
gmelodie, line 128: what

PR #1850, opened 2025-11-06
test(transport): add ipv6 tcp tests

vladopajic:
`vac:p2p:ift:2025q4-nimlibp2p-ipv6`
- adding ipv6 tests for tcp
- overall cleanups of tcp specific tests

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -52,9 +52,7 @@ jobs:
       - name: Run test suite with coverage flags
         run: |
           export NIMFLAGS="--lineDir:on --passC:-fprofile-arcs --passC:-ftest-coverage --passL:-fprofile-arcs --passL:-ftest-coverage"
-          nimble testnative
-          nimble testpubsub
-          nimble testfilter
+          nimble test
 
       - name: Run coverage
         run: |
@@ -66,5 +64,8 @@ jobs:
           genhtml coverage/coverage.f.info --output-directory coverage/output
 
       - name: Upload coverage to codecov
-        run: |
-          bash <(curl -s https://codecov.io/bash) -f coverage/coverage.f.info || echo "Codecov did not collect coverage reports"
+        uses: codecov/codecov-action@v5
+        with:
+          files: coverage/coverage.f.info
+          fail_ci_if_error: true
+          token: ${{ secrets.CODECOV_TOKEN }}
--- a/.pinned
+++ b/.pinned
@@ -19,3 +19,4 @@ websock;https://github.com/status-im/nim-websock@#35ae76f1559e835c80f9c1a3943bf9
 zlib;https://github.com/status-im/nim-zlib@#daa8723fd32299d4ca621c837430c29a5a11e19a
 jwt;https://github.com/vacp2p/nim-jwt@#18f8378de52b241f321c1f9ea905456e89b95c6f
 bearssl_pkey_decoder;https://github.com/vacp2p/bearssl_pkey_decoder@#21dd3710df9345ed2ad8bf8f882761e07863b8e0
+taskpools;https://github.com/status-im/nim-taskpools@#9e8ccc754631ac55ac2fd495e167e74e86293edb
--- a/cbind/alloc.nim
+++ b/cbind/alloc.nim
@@ -1,0 +1,97 @@
+# Nim-LibP2P
+# Copyright (c) 2023-2025 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0 ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
+# Shared Memory Management
+#
+# This file implements procedures for allocating and deallocating shared memory
+# used for communication between client and library threads.
+
+import ../libp2p/utils/sequninit
+
+## Can be shared safely between threads
+type SharedSeq*[T] = object
+  data: ptr UncheckedArray[T]
+  len: int
+
+proc alloc*(str: cstring): cstring =
+  # Byte allocation from the given address.
+  # There should be the corresponding manual deallocation with deallocShared !
+  if str.isNil():
+    var ret = cast[cstring](allocShared(1)) # Allocate memory for the null terminator
+    ret[0] = '\0' # Set the null terminator
+    return ret
+
+  let ret = cast[cstring](allocShared(len(str) + 1))
+  copyMem(ret, str, len(str) + 1)
+  return ret
+
+proc alloc*(str: string): cstring =
+  ## Byte allocation from the given address.
+  ## There should be the corresponding manual deallocation with deallocShared !
+  var ret = cast[cstring](allocShared(str.len + 1))
+  let s = cast[seq[char]](str)
+  for i in 0 ..< str.len:
+    ret[i] = s[i]
+  ret[str.len] = '\0'
+  return ret
+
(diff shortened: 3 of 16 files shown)


tests/transports/tcp_tests.nim
rlve, line 1: I think I'd get rid of the `_suite` suffix

tests/transports/tcp_tests.nim
```diff
@@ -0,0 +1,192 @@
...
+    multicodec,
+    wire,
+  ]
+import ../tools/[unittest]
+import ./utils
+
+const
+  message = "No Backdoors. No Masters. No Silence."
+  zeroMultiaddressStrIP4 = "/ip4/0.0.0.0/tcp/0"
+  zeroMultiaddressStrIP6 = "/ip6/::/tcp/0"
```
vladopajic: open for better naming suggestions

tests/transports/tcp_tests.nim
```diff
@@ -0,0 +1,197 @@
...
+  zeroAddrIP4 = "0.0.0.0:0"
+  zeroAddrIP6 = "[::]:0"
+
+proc transportProvider(): TcpTransport =
+  TcpTransport.new(upgrade = Upgrade())
```
rlve: This already exists in `testtcp` and could be passed as an argument like in the other test templates.

rlve (approved):

tests/transports/tcp_tests.nim
```diff
@@ -0,0 +1,197 @@
...
+  ]
+import ../tools/[unittest]
+import ./utils
+
+const
+  message = "No Backdoors. No Masters. No Silence."
+  zeroMultiaddressStrIP4 = "/ip4/0.0.0.0/tcp/0"
+  zeroMultiaddressStrIP6 = "/ip6/::/tcp/0"
+  zeroAddrIP4 = "0.0.0.0:0"
+  zeroAddrIP6 = "[::]:0"
```
gmelodie: Nit: capitalized names for consts maybe?

tests/transports/tcp_tests.nim
gmelodie: This should be inline or template or, my favorite, not exist (just use `TcpTransport.new(upgrade = Upgrade())` directly).

Also, why is `Upgrade()` not the default for `TcpTransport.new`? There might be a reason but for anyone reading this code it just looks unnecessary.

tests/transports/tcp_tests.nim
```diff
@@ -0,0 +1,197 @@
...
+
+proc transportProvider(): TcpTransport =
+  TcpTransport.new(upgrade = Upgrade())
+
+template tcpListenerIPTests(suiteName: string, address: string) =
```
gmelodie: Why can't we have the tests listen on both ipv4 and ipv6 addresses? This way we'd only have one test and avoid this weird looking template

tests/transports/tcp_tests.nim
```diff
@@ -0,0 +1,197 @@
...
+  block:
+    let serverListenAddr = @[MultiAddress.init(address).tryGet()]
```
gmelodie: Why not put this directly inside the tests? Reading the tests you have to scroll up to find this

gmelodie: Also, can we not use `initTaddress`?

tests/transports/tcp_tests.nim
```diff
@@ -0,0 +1,197 @@
...
+      let client = transportProvider()
+      let conn = await client.dial(ma)
+      await conn.write(message)
+
+      await handlerFut.wait(1.seconds)
+      await conn.close()
+      await client.stop()
+      server.stop()
+      server.close()
+      await server.join()
```
gmelodie, line 128: Maybe these could be `defer`. Same for other places where this happens

tests/transports/tcp_tests.nim
vladopajic: - `tcp_tests` intentionally do not use this argument, as from their standpoint, they can never have any other variant that returns any other transport. also it seems annoying to pass that argument in templates where when it is not necessary. 
- these functions are not exactly the same, `transportProvider` here returns `TcpTransport`. 
- it was added as shorthand for `TcpTransport.new(upgrade = Upgrade())` and because it is familiar way of creating transports from other tests. 
- also this is just 2 lines of code repeated 

that was the reasoning why it was added. if it feels unnaturally to have this utility here the other approach would be not to have it at all, and use `TcpTransport.new(upgrade = Upgrade())` directly in the tests. i am equally fine with current and this approach, what would you suggest?

tests/transports/tcp_tests.nim
vladopajic: that should not be nit, if naming convention is to have capital lattes for `consts` that it should be mandatory change here. having that said, it doesn't seem that this is convention here.

tests/transports/tcp_tests.nim
vladopajic: changed to `TcpTransport.new()`. anyway, iam still curious about that opinion ^

tests/transports/tcp_tests.nim
vladopajic: > tests listen on both ipv4 and ipv6 addresses 

not sure how you are thinking on implementing that. feel free to open pr with rough draft so we can see that approach.

> this weird looking template

get used to that. this is how _most_ tests are implemented in `/tests/transports`

tests/transports/tcp_tests.nim
vladopajic: > Why not put this directly inside the tests

didn't wanted to repeat this, similar approach is in other tests

> you have to scroll up

get bigger screen :trollface: 
anyway, you would have to scroll up to find `address` in other case.

tests/transports/tcp_tests.nim
vladopajic: naah man, i am not even consider that. code would look way more unreadable / uglier.

imagine having 3 defer statements in ~20 lines of code. each defer adds 1 extra line and breaks reading flow of tests.

tests/transports/tcp_tests.nim
vladopajic: going with this:
- `nameMAStr` if variable is multiaddress in `string`
- `nameMA` if variable parsed `MultiAddress` ( `== MultiAddress.init(nameMAStr)`)
- `namaTAStr` if variable is transport address (ip) address in `string`
- `nameTA` if variable is parsed `TransportAddress` ( `== initTAddress(namaTAStr)` )

- all can end with `IP4`  or `IP6` indicating version

tests/transports/tcp_tests.nim
vladopajic: > _inline or template_ 

this is just a test. and this kind of change here has neglectable impact. also tests in general don't have to have as performant code as possible unlike code in the library `/libp2p/**`. 

> why is Upgrade() not the default

qood question. it's required, but that decision is not introduced now, so we have to follow it here.

rlve: I didn't notice that it is not exactly the same. So if it is intentional, then it's fine by me.

tests/transports/tcp_tests.nim
gmelodie:
```nim
defer:
  await conn.close()
  await client.stop()
  server.stop()
  server.close()
  await server.join()

vladopajic: why would this be in defer when it is already at the end?

tests/transports/tcp_tests.nim
gmelodie: image

gmelodie (approved):

PR #1852, opened 2025-11-06
feat(kad): get providers

gmelodie:
- _Getting_ the providers for a given key is done in the same way as _getting_ a value for a given key (see [getting values section](https://github.com/libp2p/specs/blob/master/kad-dht/README.md#getting-values)) except that instead of using the `GET_VALUE` RPC message the `GET_PROVIDERS` RPC message is used.
- When a node receives a `GET_PROVIDERS` RPC, it must look up the requested key in its datastore, and respond with any corresponding records in its datastore, plus a list of closer peers in its routing table.
- `GET_PROVIDERS`: In the request `key` is set to a CID. The target node returns the closest known `providerPeers` (if any) and the `k` closest known `closerPeers`.

### DIFF
--- a/tests/testutility.nim
+++ b/tests/testutility.nim
@@ -9,7 +9,7 @@
 
 {.used.}
 
-import options
+import options, chronos
 import ../libp2p/utility
 import ./tools/[unittest]
 
@@ -83,6 +83,23 @@ suite "Utility":
     check not (compiles do:
       result:
         uint = safeConvert[int](11.uint))
+
+  asyncTest "collectCompleted collects all futures that have finished":
+    proc futFinishes(): Future[int] {.async: (raises: [CancelledError]).} =
+      await sleepAsync(1.millis)
+      return 1
+
+    proc futStalls(): Future[int] {.async: (raises: [CancelledError]).} =
+      await sleepAsync(3.seconds)
+      return 2
+
+    let
+      f1 = futFinishes()
+      f2 = futFinishes()
+      f3 = futStalls()
+      futs = @[f1, f2, f3]
+
+    check (await futs.collectCompleted(10.millis)) == @[1, 1]
 
 suite "withValue and valueOr templates":
   type TestObj = ref object


libp2p/protocols/kademlia/provider.nim
```diff
@@ -172,3 +172,75 @@ proc handleAddProvider*(
...
+  defer:
+    await conn.close()
+  let msg = Message(msgType: MessageType.getProviders, key: key)
+  await conn.writeLp(msg.encode().buffer)
+
+  let reply = Message.decode(await conn.readLp(MaxMsgSize)).valueOr:
+    error "GetProviders reply decode fail", error = error, conn = conn
+    return
+
+  var providers: HashSet[Provider]
```
richard-ramos, line 191: Looks like we're not doing anything with the `closestPeers`, shouldnt we add those to the routing table and return them as well as part of the response of `getProviders`?

gmelodie: `closerPeers` is a weird field, the spec doesn't say what to do with it... ever. It just says "put the `k` closer peers there upon responding to `getProviders`". Apparently go and rust don't do anything with them either.

libp2p/protocols/kademlia/provider.nim
```diff
@@ -172,3 +172,75 @@ proc handleAddProvider*(
...
+  ## Get providers for a given `key` from the nodes closest to that `key`.
+
+  var allProviders: HashSet[Provider]
+
+  for chunk in (await kad.findNode(key)).toChunks(kad.config.alpha):
+    let rpcBatch = chunk.mapIt(kad.switch.dispatchGetProviders(it, key))
+    try:
+      await rpcBatch.allFutures().wait(kad.config.timeout)
+      for fut in rpcBatch:
+        if not fut.failed():
```
vladopajic:
```suggestion
        if fut.completed():
```

don't included cancelled, right?

libp2p/protocols/kademlia/provider.nim
```diff
@@ -172,3 +172,75 @@ proc handleAddProvider*(
...
+          try:
+            allProviders.incl(fut.read())
+          except CatchableError:
+            continue
+    except AsyncTimeoutError:
+      debug "Some GetProviders requests timed out", key = key
+      discard # Timeouts are expected; proceed with what we have
```
vladopajic: when timeout happens, this code will never use any value that is returned before timeout. in other words, whole batch is wasted.

libp2p/protocols/kademlia/provider.nim
```diff
@@ -172,3 +172,75 @@ proc handleAddProvider*(
...
+proc getProviders*(
+    kad: KadDHT, key: Key
+): Future[HashSet[Provider]] {.async: (raises: [CancelledError]), gcsafe.} =
+  ## Get providers for a given `key` from the nodes closest to that `key`.
+
+  var allProviders: HashSet[Provider]
+
+  for chunk in (await kad.findNode(key)).toChunks(kad.config.alpha):
+    let rpcBatch = chunk.mapIt(kad.switch.dispatchGetProviders(it, key))
+    try:
```
vladopajic: it will be interesting to have utility for this logic here.

```
collectCompleted(futs: Future[T], timeout ...): seq[T]
```

because it should be tested independently.

libp2p/protocols/kademlia/provider.nim
```diff
@@ -172,3 +172,75 @@ proc handleAddProvider*(
...
+      Message(
+        msgType: MessageType.getProviders,
+        key: msg.key,
+        closerPeers: kad.findClosestPeers(msg.key),
+        providerPeers: providers.toSeq(),
+      ).encode().buffer
+    )
+  except LPStreamError as exc:
+    debug "Failed to send get-providers RPC reply", conn = conn, err = exc.msg
+    return
```
vladopajic: `return` seems unnecessary

tests/kademlia/testprovider.nim
```diff
@@ -233,3 +233,39 @@ suite "KadDHT - ProviderManager":
       not kad1.providerManager.providedKeys.hasKey(cid1)
       kad2.providerManager.providerRecords.len() == 1
       keyBefore != keyAfter
+
+  asyncTest "Get providers":
```
vladopajic, line 237: needs test case when `getProviders` gets timeout.

richard-ramos (approved):

libp2p/utility.nim
```diff
@@ -167,3 +167,29 @@ proc toChunks*[T](data: seq[T], size: int): seq[seq[T]] {.raises: [].} =
     result.add(data[i ..< endIndex])
     i = endIndex
   return result
+
+proc collectCompleted*[T, E](
```
vladopajic, line 171: could `collectCompleted` be tested with unit test?

libp2p/utility.nim
```diff
@@ -167,3 +167,29 @@ proc toChunks*[T](data: seq[T], size: int): seq[seq[T]] {.raises: [].} =
...
+    await futs.allFutures().wait(timeout)
+  except AsyncTimeoutError:
+    # Some futures didn’t finish in time → ignore
+    discard
+
+  # Collect only successful results
+  for f in futs:
+    if f.completed():
+      try:
+        completed.add(f.read())
```
vladopajic: why not `f.value()`?

libp2p/utility.nim
```diff
@@ -167,3 +167,29 @@ proc toChunks*[T](data: seq[T], size: int): seq[seq[T]] {.raises: [].} =
...
+      except CancelledError as e:
+        raise e
+      except CatchableError:
+        discard
```
vladopajic: not sure why this errors, nothing should cause errors while reading completed futures

libp2p/protocols/kademlia/provider.nim
```diff
@@ -172,3 +172,67 @@ proc handleAddProvider*(
...
+    async: (raises: [LPStreamError, DialFailedError, CancelledError]), gcsafe
+.} =
+  ## Get providers for a given `key` from the nodes closest to that `key`.
+
+  var allProviders: HashSet[Provider]
+
+  for chunk in (await kad.findNode(key)).toChunks(kad.config.alpha):
+    let rpcBatch = chunk.mapIt(kad.switch.dispatchGetProviders(it, key))
+    for res in await rpcBatch.collectCompleted(kad.config.timeout):
+      allProviders.incl(res)
```
vladopajic, line 212:
```nim
   let res = await rpcBatch.collectCompleted(kad.config.timeout)
   allProviders.addAll(res)
```

not not sure if `addAll` exists on `HashSet`; point is, there is no need for `for`-loop

libp2p/utility.nim
gmelodie: This proc allows one to pass futures that are not `raises: []`, meaning they could raise anything. In the get providers usage example, commenting this `except CatchableError` out makes the compiler say:
```
/nim-libp2p/libp2p/utility.nim(189, 24) Error: read(f) can raise an unlisted exception: DialFailedError
``` 
Because the futures I'm passing can raise that

libp2p/protocols/kademlia/provider.nim
gmelodie: `collectCompleted` will return `seq[HashSet]`, so we iterate and for each `HashSet` we merge it with the `allProviders` one

vladopajic: `inclAll` or `+` 

```nim
allProviders += await rpcBatch.collectCompleted(kad.config.timeout)

# or 

allProviders.inclAll(await rpcBatch.collectCompleted(kad.config.timeout))
```

PR #1858, opened 2025-11-08
test: add utils dir

vladopajic:
effort in organizing tests in directories.

`utils` dir is anything `libp2p/utils` related


tests/testnative.nim
```diff
@@ -10,6 +10,7 @@
 {.used.}
 
 import ./tools/test_all
```
vladopajic, line 12: `tools` should become `testing` in future prs, to avoid confusions with other tools, utils, etc...

PR #1864, opened 2025-11-11
test: remove testnative bundle

vladopajic:
- `testnative` bundle does not really make much sense now
- all tests should be included in `/tests/test_all` top level `test_all` file
- and directories are more natural and useful bundle

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -22,7 +22,7 @@ jobs:
         uses: arnetheduck/nph-action@v1
         with:
           version: 0.6.1
-          options: "examples libp2p tests interop tools *.nim*"
+          options: "./. *.nim*"
           fail: true
           suggest: true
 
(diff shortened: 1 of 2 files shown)


tests/test_all.nim
```diff
@@ -9,4 +9,24 @@
...
+  testvarint, testminprotobuf, testmultibase, testmultihash, testmultiaddress, testcid,
+  testpeerid, testsigned_envelope, testrouting_record, testnameresolve, testmultistream,
+  testidentify, testobservedaddrmanager, testconnmngr, testswitch, testnoise,
+  testpeerinfo, testpeerstore, testping, testmplex, testrelayv1, testrelayv2, testyamux,
+  testyamuxheader, testautonat, testautonatservice, testautonatv2, testautonatv2service,
+  testautorelay, testdcutr, testhpservice, testutility, testwildcardresolverservice,
+  testperf
+
+when defined(libp2p_autotls_support):
+  import testautotls
```
vladopajic, line 32: note: we should aim to reorganize all of these into specific dirs. new test files should be added to it's dir (if it does not exists, pr should start/create one).

PR #1869, opened 2025-11-12
test: run final trackers check

vladopajic:
stdout example:
image

### DIFF
--- a/tests/tools/unittest.nim
+++ b/tests/tools/unittest.nim
@@ -141,6 +141,6 @@ template finalCheckTrackers*(): untyped =
 
   unittest2.suite "Final checkTrackers":
     test "test":
-      # checkTrackers must be executed within a suite or test; # otherwise, 
+      # checkTrackers must be executed within a suite or test. otherwise, 
       # its output won't appear on stdout.
       checkTrackers()


tests/test_all.nim
```diff
@@ -29,3 +29,9 @@ import
 
 when defined(libp2p_autotls_support):
   import testautotls
+
+# Run final trackers check. 
+# After all tests are executed final trackers check is prefromed to ensure that
```
rlve:
```suggestion
# After all tests are executed final trackers check is performed to ensure that 
```

performance/test_node/main.nim
```diff
@@ -22,4 +21,10 @@ let transportType =
 
 waitFor(baseTest(scenarioName, transportType))
 
-checkTrackers()
+
+# finalCheckTrackers is perfromned here instead in scenario tests to
```
rlve:
```suggestion
# finalCheckTrackers is performed here instead in scenario tests to 
```

😄

rlve (approved):

performance/test_node/main.nim
vladopajic: https://github.com/vacp2p/nim-libp2p/pull/1869/commits/3218910f731453c015c56197ec17f3af7531fa92 
i made two complete different typos on same word :D

richard-ramos (approved):

PR #1872, opened 2025-11-12
test: add wait group

vladopajic:
chronos pr has stalled, so i am adding WaitGroup here, to libp2p tests. 

it is useful in tests in may ways:
- when code needs to signal (unblock) after `count` things have passed
- as replacement for non raising future; this is case when `count=1` giving simplified constructor

### DIFF
--- a/.github/workflows/interop.yml
+++ b/.github/workflows/interop.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Build image
         run: docker buildx build --load -t nim-libp2p-transport-head -f interop/transport/Dockerfile .
       - name: Run tests
-        uses: libp2p/test-plans/.github/actions/run-transport-interop-test@master
+        uses: libp2p/test-plans/.github/actions/run-transport-interop-test@f58b7472d85aa053ce2ba8e9135bcc126a6fed3e
         with:
           test-filter: nim-libp2p-transport-head
           # without suffix action fails because "hole-punching-interop" artifacts have 
--- a/performance/test_node/main.nim
+++ b/performance/test_node/main.nim
@@ -7,7 +7,6 @@
 # This file may not be copied, modified, or distributed except according to
 # those terms.
 
-from ../../tests/tools/unittest import checkTrackers
 import ./test_node
 import ./utils
 import chronos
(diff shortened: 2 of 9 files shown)


gmelodie:
Linking the `chronos` PR here for reference: https://github.com/status-im/nim-chronos/pull/589

tests/tools/test_sync.nim
```diff
@@ -0,0 +1,67 @@
...
+  asyncTest "wait with interval":
+    let wg = newWaitGroup(1)
+    expect AsyncTimeoutError:
+      await wg.wait(10.millis)
+
+  asyncTest "canceling wait() does not cancel underlying future":
+    let wg = newWaitGroup(1)
+    discard await wg.wait().withTimeout(1.millis)
+
+    check not wg.wait().finished
```
gmelodie, line 67: Where is the cancelling happening here exactly? Can we have an assert to ensure that the `wait` was cancelled?

tests/tools/sync.nim
```diff
@@ -0,0 +1,38 @@
...
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
+import chronos
+
+type WaitGroup* = ref object of RootObj
+  ## A synchronization primitive that waits for a collection of 
+  ## asynchronous tasks to finish.
+  count: int
+  fut: Future[void].Raising([])
```
gmelodie, line 16: It would be nice if this could be generic over at least the return type of the future (but ideally over the raised errors too, but I think it might not be possible due to chronos things that I do not properly understand)

tests/tools/test_sync.nim
vladopajic: cancelling is happening by withTimeout - it will cancel that future if timeout is reached.

we can't and don't want to ever cancel the underlay future.

tests/tools/sync.nim
vladopajic: not sure.. that's missing the point of what WaitGroup is.

gmelodie: What is weird to me is that it says right there: `a collection of asynchronous tasks` yet there's only one future being awaited

PR #1876, opened 2025-11-13
feat(transports): lsquic

richard-ramos:


tests/libp2p/transports/test_quic.nim
```diff
@@ -10,6 +10,7 @@
 {.used.}
 
 import chronos, random, stew/byteutils
+import lsquic/lsquic_ffi
```
vladopajic: why do we need this here?

richard-ramos: Not needed at all!

performance/test_node/Dockerfile
```diff
@@ -1,10 +1,9 @@
 # Create the build image
-FROM nimlang/nim:2.2.4-alpine-regular AS build
```
gmelodie, line 2: Any reason why we're changing to ubuntu? otherwise we should stay on lighter images

libp2p/transports/quictransport.nim
```diff
@@ -65,74 +78,65 @@ method readOnce*(
...
+  if readLen == 0:
+    stream.isEof = true
+    return 0
 
-  let toRead = min(nbytes, stream.cached.len)
-  copyMem(pbytes, addr stream.cached[0], toRead)
-  stream.cached = stream.cached[toRead ..^ 1]
-  libp2p_network_bytes.inc(toRead.int64, labelValues = ["in"])
-  return toRead
+  stream.activity = true
```
gmelodie, line 91: nit: Are `isEof` and `activity` not always mutualy exclusive? If so we could only have one of those

performance/test_node/Dockerfile
vladopajic: we have build issues for alpine. https://github.com/vacp2p/nim-libp2p/actions/runs/20169037854/job/57899608086

vladopajic: alpine uses `musl` while ubuntu uses `libc`
we did this because it was easier to change image here, then to change code. but maybe we should do this in future.

vladopajic: https://github.com/vacp2p/nim-lsquic/issues/34

libp2p/transports/quictransport.nim
richard-ramos: Not necessarily,  you can get an EOF because the other end closed for writing, but you can still write in the meantime.

PR #1877, opened 2025-11-13
test(quic): QuicMuxer is created without upgrader

vladopajic:


tests/transports/testquic.nim
```diff
@@ -24,11 +24,11 @@ proc quicTransProvider(): Transport {.gcsafe, raises: [].} =
   except ResultError[crypto.CryptoError]:
     raiseAssert "should not happen"
 
-proc streamProvider(transport: Transport, conn: Connection): Muxer {.raises: [].} =
+proc streamProvider(_: Transport, conn: Connection): Muxer {.raises: [].} =
```
rlve: In that case `streamProvider` type doesn't need `Transport` argument.

vladopajic: good catch, will update

libp2p/transports/quictransport.nim
```diff
@@ -151,14 +151,27 @@ method getWrapped*(self: QuicSession): P2PConnection =
...
+    _: type QuicMuxer, conn: P2PConnection, peerId: Opt[PeerId] = Opt.none(PeerId)
+): QuicMuxer {.raises: [CertificateParsingError, LPError].} =
+  let session = QuicSession(conn)
+  session.peerId =
+    if peerId.isSome:
+      peerId.get()
+    else:
+      let certificates = session.connection.certificates()
+      let cert = parse(certificates[0])
+      cert.peerId()
```
gmelodie:
```suggestion
  session.peerId = peerId.valueOr:
    let certificates = session.connection.certificates()
    let cert = parse(certificates[0])
    cert.peerId()
```

Issue #1635, opened 2025-08-20
Remove / deprecate Yamux


arnetheduck:
Given that Quic does everything that Yamux does, but more standardised, there is little point in continuing work on Yamux.

Of the options that exist:

* continue sinking resources into a dead end
* leave it in the current buggy state
* mark it deprecated and hide it behind compile-time flag
* remove it entirely

What are blockers for the last item that would motivate going with the earlier options?

vladopajic:
i don't really see why we should remove it or even mark it as deprecated. or in other words i have all contra arguments against of doing this:
- we can just leave it be; in it's current state as it is
- core CC don't have to allocate time in it's improvements, while open source contributors are welcome 
- tcp+yamux is good to have as benchmark for comparing with quic 
- it should not be deprecated (since it's alternative when tcp transport is used)
- it should not hidden with compile time flag (since this does not give any benefits to end users, and it just complicates day-to-day life for nim-libp2p CC)

PR #1878, opened 2025-11-13
chore(discovery): simplify logic

gmelodie:


tests/discovery/utils.nim
```diff
@@ -43,7 +43,7 @@ proc createSwitch*(rdv: RendezVous): Switch =
   .withRng(newRng())
   .withAddresses(@[MultiAddress.init(MemoryAutoAddress).tryGet()])
   .withMemoryTransport()
-  .withMplex()
+  .withYamux()
```
vladopajic: why?

gmelodie: Apparently there was some error being raised in mplex when using `utils/heartbeat`, but I decided to just change it to yamux as we're not supposed to be spending time in mplex related stuff anymore

vladopajic: yamux is discontinued, mplex is what we want to continue supporting

vladopajic: fyi: https://github.com/vacp2p/nim-libp2p/issues/1635

gmelodie: Ah! My bad

libp2p/discovery/rendezvousinterface.nim
```diff
@@ -19,39 +22,35 @@ type
...
-    elif attr.ofType(DiscoveryService):
-      namespace = Opt.some(string attr.to(DiscoveryService))
     elif attr.ofType(PeerId):
       namespace = Opt.some($attr.to(PeerId))
     else:
       # unhandled type
       return
-  while true:
+
+  heartbeat "Requesting peer attributes", self.timeToRequest:
```
vladopajic, line 41: feels inconsistent to have heartbeat here, and below in similar function to have `while` loop.

libp2p/discovery/rendezvousinterface.nim
```diff
@@ -60,8 +59,6 @@ method advertise*(
     for attr in self.toAdvertise:
       if attr.ofType(RdvNamespace):
         toAdvertise.add string attr.to(RdvNamespace)
```
vladopajic, line 61: the whole file uses `()` and there comes this guy writing this line without them was like: now is the time!

PR #1885, opened 2025-11-17
fix(quic): muxer handling

vladopajic:
- opening Quic streams must be done through `QuicMuxer`, unified approach
- `QuicMuxer.handle` must continue until connection is closed. Errors getting single stream should not affect other streams.  

closes: #1748


libp2p/transports/quictransport.nim
```diff
@@ -121,20 +121,15 @@ method close*(session: QuicSession) {.async: (raises: []).} =
   safeClose(session.connection)
   await procCall P2PConnection(session).close()
 
-proc getStream*(
```
vladopajic, line 124: opening Quic streams must be done through `QuicMuxer`, unified approach

libp2p/transports/quictransport.nim
```diff
@@ -192,19 +187,30 @@ method handle*(m: QuicMuxer): Future[void] {.async: (raises: []).} =
...
+      except QuicError as exc:
+        if exc.msg == "connection closed":
+          # stop handling, connection was closed
+          break
+        else:
+          # keep handling, until connection is closed. 
+          # this stream failed but we need to keep handling for other streams.
+          trace "QuicMuxer.handler got error while opening stream", msg = exc.msg
+          continue
+    asyncSpawn handleStream(stream)
```
vladopajic, line 206: `QuicMuxer.handle` must continue until connection is closed. Errors getting single stream should not affect other streams.

libp2p/transports/quictransport.nim
```diff
@@ -192,19 +187,30 @@ method handle*(m: QuicMuxer): Future[void] {.async: (raises: []).} =
...
-    trace "Exception in quic handler", msg = exc.msg
+  while not m.session.atEof:
+    let stream =
+      try:
+        await m.session.getStream(Direction.In)
+      except CancelledError:
+        # keep handling, until connection is closed
+        continue
+      except QuicError as exc:
+        if exc.msg == "connection closed":
```
richard-ramos, line 198: Probably a good idea to raise a ConnectionError in lsquic, so we don't have to rely on the error msg, but instead have a separate `except` case

libp2p/transports/quictransport.nim
```diff
@@ -192,19 +187,30 @@ method handle*(m: QuicMuxer): Future[void] {.async: (raises: []).} =
...
-  except QuicTransportError as exc:
-    trace "Exception in quic handler", msg = exc.msg
+  while not m.session.atEof:
+    let stream =
+      try:
+        await m.session.getStream(Direction.In)
+      except CancelledError:
+        # keep handling, until connection is closed
+        continue
+      except QuicError as exc:
```
gmelodie, line 197: `exc`?

Image

libp2p/transports/quictransport.nim
```diff
@@ -192,19 +187,30 @@ method handle*(m: QuicMuxer): Future[void] {.async: (raises: []).} =
...
-      asyncSpawn handleStream(stream)
-  except QuicTransportError as exc:
-    trace "Exception in quic handler", msg = exc.msg
+  while not m.session.atEof:
+    let stream =
+      try:
+        await m.session.getStream(Direction.In)
+      except CancelledError:
+        # keep handling, until connection is closed
+        continue
```
gmelodie, line 196: Maybe `handleStream` and `handle` could raise `CancelledError`?

libp2p/transports/quictransport.nim
gmelodie, line 197: It feels weird to have the stop condition of a loop to be an exception. Maybe there's a better way of checking if the connection was closed? One that we could use to do 

`while not m.session.atEof and not m.connection.closed:` (pseudocode)

libp2p/transports/quictransport.nim
vladopajic: if they do, then `handle` is stopped.

libp2p/transports/quictransport.nim
vladopajic: all blocking code must throw error to stop (unblock), otherwise condition in while is useless. 

consider:

```nim
while not m.session.atEof:  #<-- connection is open, okay
  await m.session.getStream(Direction.In) #<-- opening stream. this is blocking call as we are waiting on future.  
  # how would we exit form this `getStream` call if we are not throwing error.
  # any condition in while is useless if `getStream` does not unblock itself
```

even better pseudo code is 
```nim 
while true:  # <--- why even bother making condition when it will never be useful and is not really needed
   await m.session.getStream(Direction.In)
```

vladopajic: haha, it was already there... didn't really paid attention to that

vladopajic: changed in downstream branch https://github.com/vacp2p/nim-libp2p/pull/1876/files#diff-e6732ffe4393737472ba25d1a6f85933cd5b5b2c85c7356a9684e05e4de8eabcR198

libp2p/transports/quictransport.nim
vladopajic: changed in downstream branch https://github.com/vacp2p/nim-libp2p/pull/1876/files#diff-e6732ffe4393737472ba25d1a6f85933cd5b5b2c85c7356a9684e05e4de8eabcR194

Issue #1840, opened 2025-11-04
testws: use gnerated keys instead of hardcoded


vladopajic:
file `testws` has hardcoded consts `SecureKey` and `SecureCert` https://github.com/vacp2p/nim-libp2p/blob/f5f1d958dc98564d883bec0d1657eb7de3766bff/tests/transports/testws.nim#L30:L68

instead of using these hardcoded values, generate key and cert in code inside `wsSecureTransProvider` proc.

PR #1894, opened 2025-11-18
test(transports): Generate cert for WSS

rlve:
Fixes [testws: use gnerated keys instead of hardcoded](https://github.com/vacp2p/nim-libp2p/issues/1840)

Changes:
- add `tlsCertGenerator`
- refactor `WSS` tests to use it
- rename test files in `tests/transports`

### DIFF
--- a/tests/tools/crypto.nim
+++ b/tests/tools/crypto.nim
@@ -7,8 +7,8 @@
 # This file may not be copied, modified, or distributed except according to
 # those terms.
 
-import stew/byteutils
-import ../../libp2p/[crypto/crypto, transports/wstransport, transports/tls/certificate]
+import chronos/streams/tlsstream, stew/byteutils
+import ../../libp2p/[crypto/crypto, transports/tls/certificate]
 
 type RngWrap = object
   rng: ref HmacDrbgContext


tests/transports/utils.nim
```diff
@@ -14,6 +14,7 @@ import
   ../../libp2p/[
     transports/quictransport,
     transports/transport,
+    transports/wstransport,
```
vladopajic: why?

tests/transports/test_ws.nim
```diff
@@ -130,10 +92,16 @@ suite "WebSocket transport":
 
   asyncTest "Hostname verification":
     let ma = @[MultiAddress.init("/ip4/0.0.0.0/tcp/0/wss").tryGet()]
+
+    # Generate cert with known keypair so we can derive the PeerId (used as CN in cert)
+    let testKeyPair = KeyPair.random(PKScheme.RSA, newRng()[]).get()
```
vladopajic, line 96: should use `rng` template import from `tools/[crypto]`.

tests/transports/utils.nim
```diff
@@ -39,6 +40,21 @@ proc isTorTransport*(ma: MultiAddress): bool =
 proc isWsTransport*(ma: MultiAddress): bool =
   WebSockets.match(ma)
 
+proc tlsCertGenerator*(
```
vladopajic: consider moving this to top level utilities, probably `tools/crypto`

vladopajic: note, we should aim to reduce and eliminate this transport/utils and similar files in other directories.

tests/transports/utils.nim
rlve: removed

tests/transports/utils.nim
rlve: moved

Issue #1899, opened 2025-11-21
Add kademlia to SwitchBuilder


gmelodie:
and call bootstrap when starting the switch

PR #1906, opened 2025-11-21
chore(kad): bootstrap on switch start

gmelodie:
Implements #1899

### DIFF
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -7,7 +7,7 @@
 # This file may not be copied, modified, or distributed except according to
 # those terms.
 
-import chronos, chronicles, results, sequtils
+import chronos, chronicles, results
 import ../utils/heartbeat
 import ../[peerid, switch, multihash]
 import ./protocol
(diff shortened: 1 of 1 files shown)


libp2p/protocols/kademlia.nim
```diff
@@ -18,51 +18,47 @@ export routingtable, protobuf, types, find, get, put, provider, ping
...
+    kad: KadDHT, peerId: PeerId, addrs: seq[MultiAddress]
 ) {.async: (raises: [CancelledError]).} =
-  for (peerId, addrs) in bootstrapNodes:
+  try:
+    await kad.switch.connect(peerId, addrs)
+    debug "Connected to bootstrap peer", peerId = peerId
+  except DialFailedError as exc:
+    # at some point will want to bubble up a Result[void, SomeErrorEnum]
+    error "failed to dial to bootstrap peer", peerId = peerId, error = exc.msg
+    return
```
richard-ramos, line 32: Shouldn't we use `continue` here? Because otherwise failing to dial one bootnode will stop the whole thing.

libp2p/protocols/kademlia.nim
```diff
@@ -18,51 +18,47 @@ export routingtable, protobuf, types, find, get, put, provider, ping
 logScope:
   topics = "kad-dht"
 
-proc bootstrap*(
-    kad: KadDHT, bootstrapNodes: seq[(PeerId, seq[MultiAddress])]
+proc bootstrapNode(
```
richard-ramos: Should this function be public in case you want to add more 'bootnodes' on runtime?

libp2p/protocols/kademlia.nim
richard-ramos, line 138: Probably for a subsequent PR, but we should look into exposing a way to start/stop these loops on demand.

libp2p/protocols/kademlia.nim
gmelodie: This function `bootstrapNode` only bootstraps a single node, it is called once for each bootstrap node in a loop inside `bootstrap`

libp2p/protocols/kademlia.nim
gmelodie: I don't see any reason why one would want to do that. Can you elaborate?

PR #1908, opened 2025-11-22
feat(kad): client mode

gmelodie:
Add client mode only to kademlia

### DIFF
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -7,7 +7,7 @@
 # This file may not be copied, modified, or distributed except according to
 # those terms.
 
-import chronos, chronicles, results, sequtils
+import chronos, chronicles, results
 import ../utils/heartbeat
 import ../[peerid, switch, multihash]
 import ./protocol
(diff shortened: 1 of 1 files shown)


libp2p/protocols/kademlia.nim
```diff
@@ -99,40 +100,41 @@ proc new*(
...
-  ) {.async: (raises: [CancelledError]).} =
-    defer:
-      await conn.close()
-    while not conn.atEof:
-      let buf =
-        try:
-          await conn.readLp(MaxMsgSize)
-        except LPStreamEOFError:
+  if not client:
+    kad.handler = proc(
```
vladopajic, line 99: code all over the place does this assigns like this. would be better if there was handle function so that code is 
`kad.handler = kad.handlerImp`, this makes constructor `new` much more concise and shorter. 

then if code goes down this route, why handlers are not just some proc/method on type?

gmelodie: Great point, but one I think is problem for another PR maybe? https://github.com/vacp2p/nim-libp2p/issues/1913

richard-ramos (approved):

PR #1910, opened 2025-11-24
feat(lsquic): readOnce

richard-ramos:


vladopajic (approved):

libp2p/transports/quictransport.nim
```diff
@@ -73,22 +78,19 @@ method readOnce*(
...
     raise newLPStreamRemoteClosedError()
 
-  if stream.cached.len == 0:
+  let readLen =
     try:
-      stream.cached = await stream.stream.read()
-      if stream.cached.len == 0:
-        stream.isEof = true
-        return 0
+      await stream.stream.readInto(cast[ptr byte](pbytes), nbytes)
```
arnetheduck: we usually call a function of this style `readOnce`/`readExactly` - why the different naming?

richard-ramos: This was just a brainfart from myself. Will rename it to one of those.

PR #1912, opened 2025-11-24
test: report

rlve:
Use `xml` test report in CI summary for easier browsing of failed tests.  

Changes:
- enable `xml` test results
- add `dorny/test-reporter@v2` action to the CI workflow
- add script to properly handle error messages in `xml` file (they were being truncated by the reporting action)

See: https://github.com/vacp2p/nim-libp2p/actions/runs/19671227627?pr=1912


libp2p.nimble
```diff
@@ -85,16 +88,16 @@ task testmultiformatexts, "Run multiformat extensions tests":
...
 task testpubsub, "Runs pubsub tests":
-  runTest("libp2p/pubsub/test_all", "-d:libp2p_gossipsub_1_4")
+  runTest("libp2p/pubsub/test_all", moreoptions = "-d:libp2p_gossipsub_1_4")
 
 task testintegration, "Runs integration tests":
   runTest("integration/test_all")
 
 task test, "Runs the test suite":
-  runTest("test_all")
+  runTest("test_all", xmlReport = true)
```
vladopajic: we need human readable report on ci stdout,  when running tests. 

image

^ what we currently have.

because outupt like these can help us know when some tests deadlocks (without report we don't know which test deadlocked) - we have this problems from time to time. also sometimes it's easier to see which logs belong to which test.

tools/fix_xml_newlines.nim
```diff
@@ -0,0 +1,34 @@
+# Nim-LibP2P
```
vladopajic, line 1: not really sure if this is good place for this tool. but iam fine if it stays here. here are some thoughts:
- we should strive to import things from `./tests/tools` only in `/tests/` dir.  as these are testing utilities. utilities that are defined here but not used here, beg to question why to defined it here?
  - maybe we should rename `./tests/tools` -> `./tests/testing`.  tools, helpers, utils are all anti-pattern names.
- better place for this could be `./.github/tools` ? (+1)
- or even `./tools`? - this this would make sense if we needed it across project

vladopajic (approved):

tools/fix_xml_newlines.nim
rlve: I added it here as it's a helper script related only to the tests. But okay, will move it to `./tools`.

gmelodie (approved):

PR #1918, opened 2025-11-26
test: dynamic imports

rlve:
Changes:
- add `importTests` macro to dynamically import test files instead of using `test_all` files with static imports
- `nimble test` - goes through all dirs and files inside `tests/libp2p/` and `tests/tools/` and imports all the tests (`test_*.nim`)
- `nimble testpath <path>` - can be used to selectively run tests that match `<path>` (see readme for examples)

- using dynamic imports uncovered an issue with `chronicles` in `kademlia` tests - when test files didn't import `chronicles` then we got compilation error `undeclared identifier: 'activeChroniclesStream'` but when they imported, then we got `chronicles not used` error - use chronicles in tests as a workaround

### DIFF
--- a/tests/libp2p/pubsub/component/test_gossipsub_message_cache.nim
+++ b/tests/libp2p/pubsub/component/test_gossipsub_message_cache.nim
@@ -73,9 +73,6 @@ suite "GossipSub Component - Message Cache":
     let messageId = nodes[1].mcache.window(topic).toSeq()[0]
 
     # When heartbeat happens, circular history shifts to the next position
-    # Waiting for 5(historyLength) heartbeats
-    await waitForHeartbeat(historyLength)
-
     # Then history is cleared when the position with the message is reached again
     # And message is removed
     checkUntilTimeout:
@@ -125,7 +122,6 @@ suite "GossipSub Component - Message Cache":
     # On heartbeat, Node0 mcache advances to the next position (rotating the message cache window)
     # Node0 will gossip about messages from the last few positions, depending on the mcache window size (historyGossip)
     # By waiting more than 'historyGossip' (2x3 = 6) heartbeats, we ensure Node0 does not send IHave messages for messages older than the window size
-    await waitForHeartbeat(2 * historyGossip)
 
     # Then nodeInsideMesh receives 3 (historyGossip) IHave messages
     checkUntilTimeout:
@@ -188,11 +184,13 @@ suite "GossipSub Component - Message Cache":
 
     # When Node0 handles the IWant message, it retrieves the message from its message cache using the MessageId
     # Then Node0 relays the original message to NodeOutsideMesh
-    checkUntilTimeout:
-      messageId in
-        receivedMessagesNodeOutsideMesh[].mapIt(
+    untilTimeout:
+      pre:
+        let messages = receivedMessagesNodeOutsideMesh[].mapIt(
           nodeOutsideMesh.msgIdProvider(it).value()
         )
+      check:
+        messageId in messages
 
   asyncTest "Published and received messages are added to the seen cache":
     const
@@ -259,7 +257,9 @@ suite "GossipSub Component - Message Cache":
     # When Node2 connects with Node0 and subscribes to the topic
     await connectNodes(nodes[0], nodes[2])
     nodes[2].subscribe(topic, voidTopicHandler)
-    await waitSub(nodes[0], nodes[2], topic)
+    checkUntilTimeout:
+      nodes[0].gossipsub.hasPeerId(topic, nodes[2].peerInfo.peerId)
+      nodes[2].gossipsub.hasPeerId(topic, nodes[0].peerInfo.peerId)
 
     # And messageIds are added to node0PeerNode2 sentIHaves to allow processing IWant
     let node0PeerNode2 = nodes[0].getPeerByPeerId(topic, nodes[2].peerInfo.peerId)
@@ -269,13 +269,17 @@ suite "GossipSub Component - Message Cache":
     # And messageId1 is added to seen messages cache of Node2
     check:
       not nodes[2].addSeen(nodes[2].salt(messageId1))
+      not nodes[2].hasSeen(nodes[2].salt(messageId2))
 
     # And Node2 sends IWant to Node0 requesting both messages
-    let iWantMessage =
-      ControlMessage(iwant: @[ControlIWant(messageIDs: @[messageId1, messageId2])])
-    let node2PeerNode0 = nodes[2].getPeerByPeerId(topic, nodes[0].peerInfo.peerId)
     nodes[2].broadcast(
-      @[node2PeerNode0], RPCMsg(control: some(iWantMessage)), isHighPriority = false
+      @[nodes[2].getPeerByPeerId(topic, nodes[0].peerInfo.peerId)],
+      RPCMsg(
+        control: some(
+          ControlMessage(iwant: @[ControlIWant(messageIDs: @[messageId1, messageId2])])
+        )
+      ),
+      isHighPriority = false,
     )
 
     # Then Node2 receives only messageId2 and messageId1 is dropped
--- a/tests/libp2p/pubsub/component/test_gossipsub_message_handling.nim
+++ b/tests/libp2p/pubsub/component/test_gossipsub_message_handling.nim
@@ -74,6 +74,8 @@ suite "GossipSub Component - Message Handling":
     # This test checks if two messages, each below the maxSize, are correctly split when their combined size exceeds maxSize.
     # Expected: Both messages should be received.
     let (gossip0, gossip1, receivedMessages) = await setupTest()
+    defer:
+      await teardownTest(gossip0, gossip1)
 
     let messageSize = gossip1.maxMessageSize div 2 + 1
     let (iwantMessageIds, sentMessages) =
@@ -93,14 +95,13 @@ suite "GossipSub Component - Message Handling":
 
     checkUntilTimeout:
       receivedMessages[] == sentMessages
-    check receivedMessages[].len == 2
-
-    await teardownTest(gossip0, gossip1)
 
   asyncTest "Discard IWANT replies when both messages individually exceed maxSize":
     # This test checks if two messages, each exceeding the maxSize, are discarded and not sent.
     # Expected: No messages should be received.
     let (gossip0, gossip1, receivedMessages) = await setupTest()
+    defer:
+      await teardownTest(gossip0, gossip1)
 
     let messageSize = gossip1.maxMessageSize + 10
     let (bigIWantMessageIds, sentMessages) =
@@ -121,12 +122,12 @@ suite "GossipSub Component - Message Handling":
     checkUntilTimeout:
       receivedMessages[].len == 0
 
-    await teardownTest(gossip0, gossip1)
-
   asyncTest "Process IWANT replies when both messages are below maxSize":
     # This test checks if two messages, both below the maxSize, are correctly processed and sent.
     # Expected: Both messages should be received.
     let (gossip0, gossip1, receivedMessages) = await setupTest()
+    defer:
+      await teardownTest(gossip0, gossip1)
     let size1 = gossip1.maxMessageSize div 2
     let size2 = gossip1.maxMessageSize div 3
     let (bigIWantMessageIds, sentMessages) =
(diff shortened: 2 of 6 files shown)


tests/test_all.nim
```diff
@@ -9,8 +9,11 @@
 
 {.used.}
 
-import ./libp2p/test_all
-import ./tools/test_all
```
vladopajic, line 13: consider to keep imports static here. as no tests should be added on root lvl, and if there is new dir including it here wont be hassle. then `libp2p/test_all` should be dynamically populated. as well as `tools/test_all`.

because, in current approach, if `tools/test_all` is not dynamically populated then one can add new test `tools/test_new` which would be dynamically included to root level `test_all` but it wont be added to `tools/test_all` - unless developers adds it manually which we want to avoid. 

in other words all `test_all` files  should have `importTests(currentSourcePath())` and code should import files, and `test_all` files if there is any directory in current source path.

tests/test_path.nim
```diff
@@ -0,0 +1,30 @@
...
+
+const test_path {.strdefine.} = ""
+
+when test_path == "":
+  {.
+    error:
+      "Please specify test_path via -d:test_path=\"path/to/test\" (e.g., -d:test_path=\"quic\")"
+  .}
+
+importTestsMatching(currentSourcePath().parentDir(), test_path)
```
vladopajic: consider writing to stdout all some information about what tests are imported. it is essentially for users to know this information because slight change in argument can have different imports. 

for example:

```
==================================
Dynamic test import
Imported 3 files.

nim-libp2p/tests/libp2p/test_multistream.nim
nim-libp2p/tests/libp2p/test_name_resolve.nim
nim-libp2p/tests/libp2p/test_observed_addr_manager.nim
==================================
```

README.md
```diff
@@ -84,14 +84,18 @@ Run unit tests:
...
 nimble test
 
-# run unit tests in specific package
-nim c -r ./tests/mix/test_all
-nim c -r ./tests/kademlia/test_all
-nim c -r ./tests/pubsub/test_all
+# run tests matching a specific path
+nimble testpath quic
+nimble testpath transports/test_ws
+nimble testpath mix
```
vladopajic, line 94: consider having paths with `libp2p` prefix like `libp2p/mix` to avoid any issues when novice users want to run something that may match multiple tests (autotls can run integration and unit test). 

or maybe just improve explanation here

libp2p.nimble
```diff
@@ -88,11 +88,9 @@ task testmultiformatexts, "Run multiformat extensions tests":
     "-d:libp2p_multiaddress_exts=../tests/libp2p/multiformat_exts/multiaddress_exts.nim " &
     "-d:libp2p_multihash_exts=../tests/libp2p/multiformat_exts/multihash_exts.nim " &
     "-d:libp2p_multibase_exts=../tests/libp2p/multiformat_exts/multibase_exts.nim " &
-    "-d:libp2p_contentids_exts=../tests/libp2p/multiformat_exts/contentids_exts.nim "
-  runTest("libp2p/multiformat_exts/test_all", opts)
-
-task testpubsub, "Runs pubsub tests":
-  runTest("libp2p/pubsub/test_all", "-d:libp2p_gossipsub_1_4")
+    "-d:libp2p_contentids_exts=../tests/libp2p/multiformat_exts/contentids_exts.nim " &
+    "-d:test_path=multiformat_exts"
```
vladopajic: `libp2p2/multiformat_exts` consider tightening test_path

libp2p/protocols/kademlia.nim
```diff
@@ -13,7 +13,7 @@ import ../[peerid, switch, multihash]
 import ./protocol
 import ./kademlia/[routingtable, protobuf, types, find, get, put, provider, ping]
 
-export routingtable, protobuf, types, find, get, put, provider, ping
+export routingtable, protobuf, types, find, get, put, provider, ping, chronicles
```
vladopajic: could you check if this is needed after change in `tests/libp2p/kademlia/test_encoding.nim`?
also would be better to avoid merging code without this hack. or if we merge we have to create TODO and open issue to resolve this.

tests/test_path.nim
```diff
@@ -0,0 +1,30 @@
+# Nim-LibP2P
```
vladopajic: technically this can be in `tests/test_all`. 
if there is path use it as filter
else run all tests

tests/test_path.nim
```diff
@@ -0,0 +1,30 @@
...
+
+import std/os
+import ./imports
+
+const test_path {.strdefine.} = ""
+
+when test_path == "":
+  {.
+    error:
+      "Please specify test_path via -d:test_path=\"path/to/test\" (e.g., -d:test_path=\"quic\")"
```
vladopajic: consider `-d:test_path` -> `-d:path` 

since there are no any other `path` arguments, consider using just that to make command bit shorter and bit easier remember.

vladopajic: consider single file `test_all` as entry point for running tests (with and without flag)


`nim c -r ./tests/test_all`
`nim c -r -d:path="mix" ./tests/test_all`

tests/imports.nim
```diff
@@ -0,0 +1,46 @@
...
+import std/strutils
+import std/sequtils
+
+proc buildImports(
+    dir: string, ignorePaths: seq[string] = @[], matchPath: string = ""
+): NimNode =
+  let imports = newStmtList()
+  for file in walkDirRec(dir):
+    let (path, name, ext) = splitFile(file)
+    let isTestFile = name.startsWith("test_") and not name.startsWith("test_all")
```
vladopajic: `name.startsWith("test_all")` -> can be buggy if file name is `"test_all_secure_channels"`.
perhaps `name != "test_all"` ?

vladopajic: or maybe we dont need this anymore?

tests/imports.nim
```diff
@@ -0,0 +1,46 @@
...
+  ## imports all files in the specified directory whose filename
+  ## starts with "test_" and ends in ".nim"
+  ## ignores test_all files and any paths listed in ignorePaths
+  buildImports(dir, ignorePaths)
+
+macro importTestsMatching*(dir: static string, singlePath: static string): untyped =
+  ## imports only files in the specified directory whose path contains singlePath
+  ## and whose filename starts with "test_" and ends in ".nim"
+  ## ignores test_all files
+  buildImports(dir, @[], singlePath)
```
vladopajic: why have 2 overloaded macros and one macro with different name? why can't it be single macro or just call proc directly ?

README.md
rlve: I'll improve the explanation

libp2p/protocols/kademlia.nim
rlve: Changes in `tests/libp2p/kademlia/test_encoding.nim` were caused by this one here. And exporting `chronicles` solved my issue, I'm not sure how else it can be resolved. 

Is it a a problem though? I found more places in a codebase were `chronicles` are exported as well:
```
libp2p/protocols/perf/server.nim
libp2p/protocols/rendezvous/rendezvous.nim
libp2p/protocols/ping.nim
```

tests/imports.nim
rlve: It's needed to not import itself 😄

tests/imports.nim
rlve: I did overloaded macro as I encountered some issues if when I wanted to set a default value for one of arguments. I'll simplify it.

tests/imports.nim
vladopajic: - `test_path.nim` is also running this logic, and this file has risk or including itself 
- `test_all.nim` does not have risk of including itself as search is limited in `libp2p` and `tools` dirs

--- 

to make tool generally useful:
- files that are calling `buildImports` should limit search to directories where anything with `test_*.nim` filename can be included.
- if anything should be ignored then `ignorePaths` should be used
- no special name handling should be done, as it can be error prone (in some cases here or other repositories)

all of above are just driving in direction where  `buildImports` could be copy-pasted to different projects, or event merged to `unittest2` as utility. and if `unittest2` does not want it there, we should have it here but also keep it as general purpose.

tests/imports.nim
rlve: changed in https://github.com/vacp2p/nim-libp2p/pull/1918/commits/4f320e778d782184161d92f1f09c0dcba707c3c7

tests/test_path.nim
rlve: done in https://github.com/vacp2p/nim-libp2p/pull/1918/commits/4f320e778d782184161d92f1f09c0dcba707c3c7

README.md
rlve: done in https://github.com/vacp2p/nim-libp2p/pull/1918/commits/4f320e778d782184161d92f1f09c0dcba707c3c7

tests/test_path.nim
rlve: done in https://github.com/vacp2p/nim-libp2p/pull/1918/commits/4f320e778d782184161d92f1f09c0dcba707c3c7

libp2p/protocols/kademlia.nim
vladopajic: it doesn't seem normal to me... we should not do that

tests/imports.nim
```diff
@@ -0,0 +1,58 @@
...
+  echo "Imported ", importedFiles.len, " files."
+  echo ""
+  for file in importedFiles:
+    # Compute relative path from the base directory
+    let relPath = file.replace(baseDir & DirSep, "")
+    echo relPath
+  echo "=================================="
+  echo "\n"
+
+macro importTests*(
```
vladopajic, line 29: @arnetheduck hey, does it make sense to have `importTests` in `unittest2` ?

libp2p/protocols/kademlia.nim
rlve: Any ideas how to solve it differently?

rlve: Ok, it's reverted. I just used chronicles in tests as a workaround. Feel free to suggest something better, but this does the trick for now.

vladopajic: > I just used chronicles in tests as a workaround

could you please create issue that addresses this "hack"?

>  I found more places in a codebase were chronicles are exported as well:

could you also create issue that addresses this? we should not export `chronicles` here.

tests/libp2p/kademlia/test_find_node.nim
```diff
@@ -14,6 +14,8 @@ import ../../../libp2p/[protocols/kademlia, switch, builders]
 import ../../tools/[unittest]
 import ./utils.nim
 
+trace "chronicles has to be imported to fix Error: undeclared identifier: 'activeChroniclesStream'"
```
richard-ramos, line 17: Why is this log necessary if we are already importing chronicles here?

tests/libp2p/pubsub/component/test_gossipsub_gossip.nim
```diff
@@ -11,7 +11,7 @@
 
 import chronos, std/[sequtils], stew/byteutils
 import ../../../../libp2p/protocols/pubsub/[gossipsub, mcache, peertable, rpc/message]
-import ../../../tools/[unittest, futures]
+import ../../../tools/[unittest]
```
richard-ramos:
```suggestion
import ../../../tools/unittest
```

tests/libp2p/kademlia/test_find_node.nim
richard-ramos: Nvm. Just saw the issue description

vladopajic: perhaps compiler thinks that `chronicles` is not used otherwise

rlve: Using dynamic imports uncovered an issue with `chronicles` in `kademlia` tests - when test files didn't import `chronicles` then we got compilation error `undeclared identifier: 'activeChroniclesStream'`, but when they imported, then we got `chronicles` not used error.

As I wasn't able to fix a root cause and I don't want to change the `switch("warningAsError", "UnusedImport:on")` setting, then I added these traces as a workaround, for `chronicles` to be used.

libp2p/protocols/kademlia.nim
rlve: Created issues:
https://github.com/vacp2p/nim-libp2p/issues/1940
https://github.com/vacp2p/nim-libp2p/issues/1941

PR #1919, opened 2025-11-27
feat(wstransport): support concurrent accept

fcecin:
NOTE: This is a draft PR; it depends on https://github.com/status-im/nim-websock/pull/180 and addresses bug https://github.com/waku-org/nwaku/issues/3634.

**Rationale**

nim-websock's `HttpServer.accept` proc, which is used by nim-libp2p's websocket client acceptor, does two things:
* Accepts a stream socket connection 
* Parses the HTTP request (with a headersTimeout)

If a client sends an incomplete HTTP request, `HttpServer.accept` will starve stream socket acceptance for the full duration of headersTimeout / handshakeTimeout (3 seconds as configured by nim-libp2p). That manifests as a buildup in the Recv-Q of the accept socket as reported in the nwaku bug.

**Proposed solution**

This PR is a follow-up and a replacement to this unmerged PR in nim-websock: https://github.com/status-im/nim-websock/pull/179. It's moving the proposed fix one level up from nim-websock and into nim-libp2p.

The proposed solution spawns a connection accept dispatcher task as part of WsTransport start. The accept dispatcher maintains a configured number of pending `HttpServer.accept` futures on each of the WsTransport's httpservers, handling handshakes asynchronously to prevent blocking the accept loop.

**Discussion**

As @arnetheduck [points out](https://github.com/status-im/nim-websock/pull/179#issuecomment-3556450768), this could be solved indirectly if we shift to a different architecture. But meanwhile, we can time out header parsing with some configured level of concurrency by spawning N tasks to do `HttpServer.accept` as previously [suggested](https://github.com/status-im/nim-websock/issues/79#issuecomment-985094790), which solves immediate service degradation due to the implicit serialization of websocket connections and header parsing timeouts.

### DIFF
--- a/tests/libp2p/transports/stream_tests.nim
+++ b/tests/libp2p/transports/stream_tests.nim
@@ -562,6 +562,7 @@ template streamTransportTest*(
     var clientFuts: seq[Future[void]]
     for i in 0 ..< numConnections:
       clientFuts.add(runClient(server, i))
+      await sleepAsync(300.millis)
 
     echo "[DEBUG] await allFutures(clientFuts)"
     await allFutures(clientFuts)


libp2p/transports/wstransport.nim
```diff
@@ -406,6 +487,11 @@ proc new*(
     factories: @factories,
     rng: rng,
     handshakeTimeout: handshakeTimeout,
+    concurrentAcceptsPerHttpServer:
+      if concurrentAcceptsPerHttpServer <= 0:
+        DefaultConcurrentAcceptsPerHttpServer
+      else:
+        concurrentAcceptsPerHttpServer,
```
vladopajic, line 514: if this logic is added there why is it not added to other `new` proc?
my main concern is if this logic, for fallback value, is really needed then it should be in other proc as well.

libp2p/transports/wstransport.nim
```diff
@@ -294,58 +388,44 @@ method accept*(
   if not self.running:
     raise newTransportClosedError()
 
-  if self.acceptFuts.len <= 0:
-    self.acceptFuts = self.httpservers.mapIt(it.accept())
-
-  if self.acceptFuts.len <= 0:
+  let res = await self.acceptResults.popFirst()
+
+  if res.isErr:
```
vladopajic, line 413: hmmm... i would still prefer to use same construct for handling error. that is using `try-expect` instead of `if-of`. i can understand why code was made like this, but i would prefer for all error handling across codebase to be `try-expect` instead of `if .. of ..`.

pseudocode:

```nim
# early return if there is value 
if not res.isErr: 
 return res.value 
 
 # handle errors
 
 try:
   raise res.error
 except WebSocketError:
 except:
 except:
 except:
```

would this be possible?

libp2p/transports/wstransport.nim
```diff
@@ -294,58 +388,44 @@ method accept*(
...
+    elif exc of TransportOsError:
+      debug "OS Error", description = exc.msg
+    elif exc of TransportUseClosedError:
+      debug "Server was closed", description = exc.msg
+      raise newTransportClosedError(exc)
+    elif exc of AcceptDispatcherFinishedError:
+      try:
+        self.acceptResults.addLastNoWait(res)
+      except AsyncQueueFullError:
+        raise newException(AssertionDefect, "wstransport handshakeResults queue full")
```
vladopajic, line 436:
```suggestion
      try:
        self.acceptResults.addLastNoWait(res)
      except AsyncQueueFullError:
        raiseAssert "acceptResults queue is full"
```
if accepted please search and updated other places...

reasoning: `raiseAssert` also raises `AssertionDefect`. libp2p across codebase mainly uses `raiseAssert`s

libp2p/transports/wstransport.nim
```diff
@@ -129,6 +137,123 @@ type WsTransport* = ref object of Transport
...
+) {.async: (raises: []).} =
+  try:
+    let req = await finished
+    try:
+      let wstransp = await self.wsserver.handleRequest(req).wait(self.handshakeTimeout)
+      let conn = await self.connHandler(wstransp, secure, Direction.In)
+      try:
+        self.acceptResults.addLastNoWait(AcceptResult.ok(conn))
+      except AsyncQueueFullError:
+        await noCancel req.stream.closeWait()
```
vladopajic, line 188: this never happens?

libp2p/transports/wstransport.nim
```diff
@@ -129,6 +137,123 @@ type WsTransport* = ref object of Transport
...
+    warn "WsTransport.acceptDispatcher has no work; exiting"
+    return
+
+  while self.running:
+    try:
+      var finished: Future[HttpRequest]
+      try:
+        finished = await one(self.acceptFuts)
+      except ValueError as exc:
+        raise newException(AssertionDefect, "wstransport accept error: " & exc.msg, exc)
```
vladopajic, line 221: should be `raiseAssert`

libp2p/transports/wstransport.nim
```diff
@@ -129,6 +137,123 @@ type WsTransport* = ref object of Transport
...
+      if serverIndex >= self.httpservers.len:
+        raise newException(AssertionDefect, "wstransport server index out of bounds")
+      let httpServer = self.httpservers[serverIndex]
+
+      # Replenish the completed accept() future for the same server
+      self.acceptFuts[futIndex] = httpServer.accept()
+
+      asyncSpawn self.handshakeWorker(finished, httpServer.secure)
+
+      await sleepAsync(0) # be nice
```
vladopajic, line 236: is this really necessary? would be interesting to here reasoning here

libp2p/transports/wstransport.nim
```diff
@@ -129,6 +137,123 @@ type WsTransport* = ref object of Transport
...
+      await noCancel req.stream.closeWait()
+      try:
+        self.acceptResults.addLastNoWait(AcceptResult.err(exc))
+      except AsyncQueueFullError:
+        discard
+  except CatchableError as exc:
+    try:
+      self.acceptResults.addLastNoWait(AcceptResult.err(exc))
+    except AsyncQueueFullError:
+      discard
```
vladopajic, line 199: please improve this logic:
- use specific errors instead of `CatchableError`
- do not use nasted error handling (`try-except` within `try-except`)

for example:

```
let wstransp = 
  try  
    await self.wsserver.handleRequest(req).wait(self.handshakeTimeout)
  except ....

 let conn = 
  try
     await self.connHandler(wstransp, secure, Direction.In)
  except ....
  
```

libp2p/transports/wstransport.nim
```diff
@@ -129,6 +137,123 @@ type WsTransport* = ref object of Transport
...
+
+  # Either httpservers is empty, or concurrentAcceptsPerHttpServer is zero (which is a defect)
+  if self.acceptFuts.len == 0:
+    warn "WsTransport.acceptDispatcher has no work; exiting"
+    return
+
+  while self.running:
+    try:
+      var finished: Future[HttpRequest]
+      try:
```
vladopajic, line 218: please avoid `try` within `try`

libp2p/transports/wstransport.nim
```diff
@@ -129,6 +137,123 @@ type WsTransport* = ref object of Transport
...
+          else:
+            MultiAddress.init("/ws")
+        remoteAddr = stream.stream.reader.tsource.remoteAddress
+        localAddr = stream.stream.reader.tsource.localAddress
+
+      (
+        MultiAddress.init(remoteAddr).tryGet() & codec.tryGet(),
+        MultiAddress.init(localAddr).tryGet() & codec.tryGet(),
+      )
+    except CatchableError as exc:
```
vladopajic, line 160: please use specific error instead of `CatchableError`.

libp2p/transports/wstransport.nim
```diff
@@ -37,6 +37,7 @@ const
   DefaultHeadersTimeout = 3.seconds
```
vladopajic, line 37: CI on windows: for some reason fails

fcecin: Will investigate. It's hanging non-deterministically in a `[Suite  ] WebSocket transport` test:

```
  [Test   ]           server with multiple parallel connections
Terminate batch job (Y/N)? 
^C
Error: The operation was canceled.
```

...that should take 2-3 seconds every time:

```
  [Test   ]           server with multiple parallel connections
@[0, 1, 2, 3, 4, 4, 2, 0, 3, 1, 4, 3, 0, 1, 2, 4, 3, 0, 1, 2, 4, 1, 0, 3, 2, 1, 4, 0, 3, 2, 1, 0, 4, 3, 4, 1, 2, 0, 2, 1, 3, 4, 1, 2, 3, 0, 4, 1, 2, 0, 1, 3, 4, 2, 0, 4, 2, 0, 3, 1, 4, 1, 2, 3, 0, 2, 4, 0, 2, 1, 0, 3, 4, 2, 1, 4, 0, 2, 3, 1, 3, 2, 0, 4, 1, 2, 3, 0, 4, 2, 1, 3, 0, 1, 4, 2, 1, 4, 0, 3, 4, 0, 1, 2, 4, 3, 3, 0, 1, 2, 4, 2, 3, 0, 1, 2, 4, 3, 4, 0, 2, 3, 1, 2, 3, 4, 1, 0, 2, 3, 4, 1, 3, 4, 2, 0, 3, 1, 4, 0, 2, 2, 3, 1, 4, 0, 3, 0, 2, 4, 1, 3, 0, 1, 4, 3, 1, 3, 0, 0]
  [OK     ] (  2.62s) server with multiple parallel connections
```

fcecin: If we add 300ms of delay between client connection attempts it doesn't hang, specifically on Windows and on Nim 2.2.6. There is a nonzero chance this is a Chronos + Nim 2.2.6 + Windows bug or some combination thereof. It doesn't seem to hang on the Nim 2.0.16 + Windows CI. I'll investigate further.

vladopajic: if we run client, then wait it to finish (with 300m delay) we will serialize all client connections, which avoids purpose of test which is to run may client connections in parallel. 

----

if you need to test via ci, you can run only ws tests in ci so you don't wait too much time for test to finish. 
to do this, change `test/test_all.nim` and just import ws tests here:

```
import ./libp2p/transports/test_ws
```

fcecin: > if we run client, then wait it to finish (with 300m delay) we will serialize all client connections, which avoids purpose of test which is to run may client connections in parallel.

Yes, I was just abusing the CI to quickly test some theories on Windows.

> if you need to test via ci, you can run only ws tests in ci so you don't wait too much time for test to run. to do this, change `test/test_all.nim` and just import ws tests :
> 
> ```
> import ./libp2p/transports/test_ws
> ```

Thanks!

If I need to do more Windows testing I'll set it up on a Windows machine. I didn't think it was something serious, but now I think it is not a quick fix.

Sorry for the noise. And there's a good chance that this nim-libp2p PR is [dead](https://github.com/status-im/nim-websock/pull/180#issuecomment-3598713863), in any case.

vladopajic: consider testing https://github.com/status-im/nim-websock/pull/180 - add test case that asserts concurrent accepts

vladopajic: fyi: similar test failed on  `macos-15-arm64 (Nim v2.2.6)` "master" (technically not master but code just adds logs)

https://github.com/vacp2p/nim-libp2p/actions/runs/19856566195/job/56895878119?pr=1924

image

fcecin: Thank you very much! Yes, I think this patch/PR is too risky. I'll submit an alternative one that doesn't have this problem.

PR #1921, opened 2025-11-27
feat(kad): extended Kademlia discovery

SionoiS:
I added;
- random walk proc that return signed peer records.
- periodic publishing of the node own self signed peer records.
- add/remove services
- extensible peer records

Implementation of [extended-kademlia-discovery](https://github.com/vacp2p/rfc-index/blob/master/docs/ift-ts/raw/extended-kad-disco.md) & [extensible-peer-records](https://github.com/vacp2p/rfc-index/blob/master/docs/ift-ts/raw/extensible-peer-records.md)

### DIFF
--- a/tests/libp2p/kademlia/mock_kademlia.nim
+++ b/tests/libp2p/kademlia/mock_kademlia.nim
@@ -15,7 +15,7 @@ type MockKadDHT* = ref object of KadDHT
   handleFindNodeCalls*: int
 
 method findNode*(
-    kad: MockKadDHT, target: Key
+    kad: MockKadDHT, target: Key, queue = newAsyncQueue[(PeerId, Opt[Message])]()
 ): Future[seq[PeerId]] {.async: (raises: [CancelledError]).} =
   kad.findNodeCalls.add(target)
   return kad.rtable.findClosestPeerIds(target, kad.config.replication)


libp2p/protocols/kademlia/types.nim
```diff
@@ -71,7 +72,8 @@ proc toPeer*(k: Key, switch: Switch): Result[Peer, string] =
         if switch.isConnected(peer):
           ConnectionType.connected
         else:
-          ConnectionType.notConnected,
+          ConnectionType.notConnected
+      ,
```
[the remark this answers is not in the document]
SionoiS: ah! my nph version must be out of date...

libp2p/protocols/kademlia.nim
```diff
@@ -66,6 +66,35 @@ proc maintainBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
+  let key = peerInfo.peerId.toKey()
+
+  let putRes = await kad.putValue(key, value)
+  if putRes.isErr:
+    error "Failed to put signed peer record", err = putRes.error
+
+proc maintainSelfSignedPeerRecord(kad: KadDHT) {.async: (raises: [CancelledError]).} =
+  heartbeat "refresh self signed peer record",
+    kad.config.selfSignedPeerRecordRefreshTime:
+    await kad.refreshSelfSignedPeerRecord()
```
jm-clius: Would be interesting to mimic as close as possible (at least in the first stages) the maintenance that discv5 does by default. Presumably "own" records are also refreshed periodically there? I'll take a look too to see if we can use those values by default and what other maintenance tasks (e.g. liveness pings) our discovery protocol could benefit from.

libp2p/protocols/kademlia.nim
```diff
@@ -66,6 +66,35 @@ proc maintainBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
   heartbeat "refresh buckets", kad.config.bucketRefreshTime:
     await kad.refreshBuckets()
 
+proc refreshSelfSignedPeerRecord(kad: KadDHT) {.async: (raises: [CancelledError]).} =
```
jm-clius: I'm wondering if we'll need a public `proc advertise(signedPeerRecord: SignedPeerRecord)*` method to allow higher layers to encode and advertise their own peer records (e.g. to encode app-level capabilities, such as waku protocol flags).

libp2p/protocols/kademlia/types.nim
```diff
@@ -250,6 +252,27 @@ method isValid*(
...
+type LogosEntryValidator* = ref object of EntryValidator
+method isValid*(
+    self: LogosEntryValidator, key: Key, record: EntryRecord
+): bool {.raises: [], gcsafe.} =
+  let sprRes = SignedPeerRecord.decode(record.value)
+
+  if sprRes.isErr:
+    return false
+
+  let spr = sprRes.get()
```
richard-ramos:
```suggestion
let spr = sprRes.valueOr:
  debug "some message here", error
  return false
```

libp2p/protocols/kademlia/types.nim
```diff
@@ -250,6 +252,27 @@ method isValid*(
...
+
+  let expectedPeerIdRes = key.toPeerId()
+
+  if expectedPeerIdRes.isErr:
+    return false
+
+  if spr.data.peerId != expectedPeerIdRes.get():
+    return false
```
richard-ramos:
```suggestion
  let expectedPeerId = key.toPeerId().valueOr:
     debug "some message here", error
     return false

  if spr.data.peerId != expectedPeerId:
    return false
```

libp2p/protocols/kademlia/find.nim
```diff
@@ -53,6 +53,41 @@ proc sendFindNode*(
 
   return reply
 
+proc findRandom*(
```
richard-ramos: This looks very interesting. You could probably generalize it with generics so callers can call find with arbitrary value types instead of only SignedPeerRecord.

The main thing that i see here: this behavior doesn’t belong inside libp2p kademlia. Assuming dht values are SignedPeerRecords is an application-level convention. Pulling it into the kad core makes the dht less generic. better to keep Kad dumb and let higher layers define whatever record type and advertising periodicity they need.

libp2p/protocols/kademlia/find.nim
```diff
@@ -53,6 +53,41 @@ proc sendFindNode*(
...
+
+  for fut in getFutures:
+    if fut.failed:
+      continue
+
+    let res = fut.read()
+    if res.isErr:
+      continue
+
+    let entry = res.get()
```
richard-ramos:
```suggestion
    let entry = fut.read().valueOr:
       debug "some message here", error
       continue
```

libp2p/protocols/kademlia/find.nim
```diff
@@ -53,6 +53,41 @@ proc sendFindNode*(
...
+
+    let decodeRes = SignedPeerRecord.decode(entry.value)
+    if decodeRes.isErr:
+      continue
+
+    records.add(decodeRes.get())
```
richard-ramos:
```suggestion
    let record = SignedPeerRecord.decode(entry.value).valueOr:
         debug "some message here", error
         continue

    records.add(record)
```

libp2p/protocols/kademlia/types.nim
```diff
@@ -250,6 +252,27 @@ method isValid*(
 ): bool {.raises: [], gcsafe.} =
   return true
 
+type LogosEntryValidator* = ref object of EntryValidator
```
richard-ramos: This too seems very app specific and should likely not be part of libp2p kad-dht

libp2p/protocols/kademlia/types.nim
```diff
@@ -278,11 +301,37 @@ method select*(
...
+
+  var maxSeqNo: uint64 = 0
+  var bestIdx: int = -1
+
+  for i, rec in records:
+    let sprRes = SignedPeerRecord.decode(rec.value)
+    if sprRes.isErr:
+      continue
+
+    let seqNo = sprRes.get().data.seqNo
```
richard-ramos:
```suggestion
    let spr = SignedPeerRecord.decode(rec.value).valueOr:
       debug "some message here", error
       continue
    let seqNo = spr.data.seqNo
```

libp2p/protocols/kademlia.nim
```diff
@@ -66,6 +66,35 @@ proc maintainBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
+  let peerInfo = kad.switch.peerInfo
+  let peerRecord = PeerRecord.init(peerInfo.peerId, peerInfo.addrs)
```
richard-ramos: Wdyt of having an additional init function for routing record that receives a PeerInfo, so you could do:
```
let peerRecord = PeerRecord.init(kad.switch.peerInfo)
```

libp2p/protocols/kademlia.nim
```diff
@@ -66,6 +66,35 @@ proc maintainBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
+  let privKey = kad.switch.privKey
+
+  let signedRecordRes = SignedPeerRecord.init(privKey, peerRecord)
+  if signedRecordRes.isErr:
+    error "Failed to sign peer record", err = $signedRecordRes.error
+    return
+
+  let signedRecord = signedRecordRes.get()
```
```suggestion
  let signerRecord = SignedPeerRecord.init(privKey, peerRecord).valueOr:
    error "Failed to sign peer record", err = error
    return
```

libp2p/protocols/kademlia.nim
```diff
@@ -66,6 +66,35 @@ proc maintainBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
+
+  let valueRes = signedRecord.encode()
+  if valueRes.isErr:
+    error "Failed to encode signed peer record", err = $valueRes.error
+    return
+
+  let value = valueRes.get()
```
```suggestion
  let value = signedRecord.encode().valueOr:
    error "Failed to encode signed peer record", err = error
    return
```

libp2p/protocols/kademlia.nim
richard-ramos: This seems to be app specific and should not be part of libp2p

libp2p/protocols/kademlia/find.nim
```diff
@@ -53,6 +53,41 @@ proc sendFindNode*(
...
+  var getFutures: seq[Future[Result[EntryRecord, string]]] = @[]
+  for pid in peerIds:
+    getFutures.add(kad.getValue(pid.toKey()))
+
+  await allFutures(getFutures)
+  var records: seq[SignedPeerRecord] = @[]
+
+  for fut in getFutures:
+    if fut.failed:
+      continue
```
richard-ramos: We should likely trace log why the future failed (might help with debugging)

libp2p/protocols/kademlia.nim
SionoiS: The problem I see is the versioning. Where does the version sits? At the app or libp2p lvl.

We could use some other key and a different version for app lvl records but can't use a single record for both.

edit: thinking about it some more, having signed peer records with keys that are not the peer id is not better than not signed at all since you can't be sure it comes from the same node...

libp2p/protocols/kademlia.nim
SionoiS: If your question is if the records are updated the answer is yes. Any KV in the DHT is refreshed when possible but it's the responsibility of the signer to `putValue` again periodically.

SionoiS:
I moved the logic out of kad dht.

I still pondering the best way to expose the functionalities and I will also add mix key to the peer records.

libp2p/routing_record.nim
```diff
@@ -25,6 +25,7 @@ type
     peerId*: PeerId
     seqNo*: uint64
     addresses*: seq[AddressInfo]
+    mixKey*: seq[byte]
```
jm-clius: Since we're doing something different here from what is specified (https://github.com/libp2p/specs/blob/377f05abe37fca3bfe392b5e07c134da94306c02/RFC/0003-routing-records.md), we have to create a separate type of routing record here with its own unique [payload type](https://github.com/libp2p/specs/blob/377f05abe37fca3bfe392b5e07c134da94306c02/RFC/0002-signed-envelopes.md#payload-type-information). We'll eventually merge this payload type to the "official" multicodec table [here](https://github.com/multiformats/multicodec).

It seems to me that what is necessary is something like an `ExtensibleRoutingRecord` that can have an arbitrary list of key-value pairs added by applications to encode more capabilities/keys (or other discoverable information) - similar to how ENRs did this.

Something like:

```protobuf
// ExtensiblePeerRecord contains the listen addresses and extensible list of services supported by a peer at a particular point in time.
message ExtensiblePeerRecord {
  // AddressInfo wraps a multiaddr. In the future, it may be extended to
  // contain additional metadata, such as "routability" (whether an address is
  // local or global, etc).
  message AddressInfo {
    bytes multiaddr = 1;
  }
  
  message ServiceInfo{
    string id = 1;
    optional bytes value = 2;
  }

  // the peer id of the subject of the record (who these addresses belong to).
  bytes peer_id = 1;

  // A monotonically increasing sequence number, used for record ordering.
  uint64 seq = 2;

  // All current listen addresses
  repeated AddressInfo addresses = 3;

  // Extensible list of advertised services
  repeated ServiceInfo services = 4;
}
```

This way applications can add their own supported services (as "service ids") and optionally add some "value" against that service entry. Mix, for example, would be `"waku.mix" : <mix-key>`, but nodes could add `"waku.store"`, `"waku.filter"`, etc. as supported services. It should then be possible to add general filtering by service ID in our library as well, which is one of the requirements.

(I think we need a basic spec for the wire protocol and API, even if it's just super raw with the protobuf and public methods. :) )

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -0,0 +1,114 @@
...
+import ./types
+
+proc findAllNode*(
+    disco: KademliaDiscovery, targetId: Key
+): Future[seq[PeerId]] {.async: (raises: [CancelledError]).} =
+  ## Return all peers on the way to a target ID.
+
+  var initialPeers = disco.rtable.findClosestPeerIds(targetId, disco.config.replication)
+  var state = LookupState.init(
+    targetId, initialPeers, disco.config.alpha, 100, disco.rtable.config.hasher
```
jm-clius: Why `100` and not `disco.config.replication`?

SionoiS: Maybe would have been more clear with `100,000,000`. I tried modifying a lot of the code but this was the simplest thing to do in the end.

libp2p/routing_record.nim
SionoiS: https://github.com/vacp2p/rfc-index/pull/220

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -0,0 +1,116 @@
...
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
+import std/[sets, tables, sequtils]
+import chronos, chronicles, results
+import ../../[peerid, switch, multihash]
+import ../kademlia/[types, routingtable, lookupstate, protobuf, find, get, protobuf]
+import ./types
+
+proc findAllNode*(
```
gmelodie: nit
```suggestion
proc findNodesBetween*(
```

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -0,0 +1,116 @@
...
+    disco: KademliaDiscovery, targetId: Key
+): Future[seq[PeerId]] {.async: (raises: [CancelledError]).} =
+  ## Return all peerIds on the way to a target ID.
+
+  var initialPeers = disco.rtable.findClosestPeerIds(targetId, disco.config.replication)
+  var state = LookupState.init(
+    targetId, initialPeers, disco.config.alpha, high(int), disco.rtable.config.hasher
```
gmelodie: The only difference between this function and `findNode` in [`find.nim`](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/kademlia/find.nim) is this line, where `kad.config.replication` was swapped for `high(int)`. Not sure why that is, but in any case it seems that `findNode` should be changed, not duplicated here.

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -0,0 +1,116 @@
...
+    debug "cannot generate random peer id", error
+    return @[]
+
+  let randomKey = randomPeerId.toKey()
+
+  let peerIds = await disco.findAllNode(randomKey)
+
+  var getFutures: seq[Future[Result[EntryRecord, string]]] = @[]
+  for pid in peerIds:
+    getFutures.add(disco.getValue(pid.toKey()))
```
gmelodie: nit: Maybe use `mapIt` instead here

libp2p/protocols/kademlia_discovery/types.nim
```diff
@@ -0,0 +1,155 @@
...
+
+proc toKey*(service: ServiceInfo): Key =
+  return MultiHash.digest("sha2-256", service.id).get().toKey()
+
+proc init*(
+    T: typedesc[LogosPeerRecord], peerInfo: PeerInfo, services: seq[ServiceInfo]
+): T =
+  LogosPeerRecord(
+    peerId: peerInfo.peerId,
+    seqNo: getTime().toUnix().uint64,
```
gmelodie: nit: Consider using a random sequence number between two values, like TCP, to avoid timing issues and conflicts

libp2p/protocols/kademlia_discovery/types.nim
```diff
@@ -0,0 +1,155 @@
...
+  let spr = SignedPeerRecord.decode(record.value).valueOr:
+    return false
+
+  let expectedPeerId = key.toPeerId().valueOr:
+    return false
+
+  if spr.data.peerId != expectedPeerId:
+    return false
+
+  return true
```
gmelodie: nit
```suggestion
  return spr.data.peerId == expectedPeerId
```

libp2p/protocols/kademlia_discovery/randomfind.nim
SionoiS: Exposing the replication factor for libp2p `find` does not make sense. The simplest thing to do was to duplicate the proc.

If you want to refactor `find` to suit my need I would appreciate, my goal was to minimize modify existing code.

libp2p/protocols/kademlia_discovery/types.nim
SionoiS: Peer records have only one authority, the signer, no conflict can occur. Using an ever increasing number (timestamp) for versioning is what the spec recommend.

libp2p/protocols/kademlia_discovery/randomfind.nim
gmelodie: I suggest adding a parameter `nodesBetween: bool = false` to `find` and have `kademlia_discovery` call find with that parameter set to true.

libp2p/protocols/kad_disco.nim
```diff
@@ -0,0 +1,98 @@
...
+  topics = "kad-disco"
+
+proc refreshSelfSignedPeerRecord(
+    disco: KademliaDiscovery
+) {.async: (raises: [CancelledError]).} =
+  await disco.switch.peerInfo.update()
+
+  let
+    peerInfo: PeerInfo = disco.switch.peerInfo
+    services: seq[ServiceInfo] = disco.services.toSeq()
```
gmelodie, line 22: Please correct me if I'm wrong, but `Service` here means the capabilities we're discovering, correct? If that's the case I think we should rename `Service` to `Cap` or `Capability`

libp2p/protocols/kad_disco.nim
```diff
@@ -0,0 +1,98 @@
...
+      addresses: peerInfo.addrs.mapIt(AddressInfo(address: it)),
+      services: services,
+    ),
+  ).valueOr:
+    error "Failed to create signed peer record", error
+    return
+
+  let encodedSR = extPeerRecord.encode().valueOr:
+    error "Failed to encode signed peer record", error
+    return
```
gmelodie, line 38: `putValue` should be in charge of doing the `encoding`

libp2p/protocols/kad_disco.nim
```diff
@@ -0,0 +1,98 @@
...
+proc maintainSelfSignedPeerRecord(
+    disco: KademliaDiscovery
+) {.async: (raises: [CancelledError]).} =
+  heartbeat "refresh self signed peer record", disco.config.bucketRefreshTime:
+    await disco.refreshSelfSignedPeerRecord()
+
+proc addService*(disco: KademliaDiscovery, service: ServiceInfo): bool =
+  ## Include this service in the set of services this node provides.
+
+  return disco.services.containsOrIncl(service)
```
gmelodie, line 55: Change `services` to a `HashSet`, maybe this will require implementing some methods for `ServiceInfo` (`==` and `hash` if I'm not mistaken), but that will make this more elegant and efficient. Then you'd just do `incl` and `excl`

libp2p/protocols/kad_disco.nim
```diff
@@ -0,0 +1,98 @@
...
+  disco.selfSignedLoop = nil
+
+  disco.maintenanceLoop.cancelSoon()
+  disco.maintenanceLoop = nil
+
+  disco.republishLoop.cancelSoon()
+  disco.republishLoop = nil
+
+  disco.expiredLoop.cancelSoon()
+  disco.expiredLoop = nil
```
gmelodie: This is pretty much duplicated from `kademlia`, needs to be removed entirely. If all you need is to run `disco.maintainSelfSignedPeerRecord()`, you can overload `start` and call its parent's `start` inside it, like:

```nim
method start*(disco: KademliaDiscovery) {.async: (raises: [CancelledError]).} =
  if disco.started:
    warn "Starting kad-disco twice"
    return

  disco.selfSignedLoop = disco.maintainSelfSignedPeerRecord()
  await procCall start(KadDHT(disco))
```

Same for `stop`

libp2p/protocols/kademlia_discovery/types.nim
```diff
@@ -0,0 +1,169 @@
...
+
+  for service in record.services:
+    var servPb = initProtoBuffer()
+    servPb.write(1, service.id)
+    if service.data.len > 0:
+      servPb.write(2, service.data)
+    pb.write(4, servPb)
+
+  pb.finish()
+  pb.buffer
```
gmelodie: This will be better suited for a file named `protobuf.nim`. Also, there should be a spec or RFC describing this new protocol. Is it written somewhere? Otherwise there's no way to tell if this is doing what's supposed to be doing.

tests/libp2p/kademlia_discovery/test_ext_peer_records.nim
```diff
@@ -0,0 +1,151 @@
...
+        @[
+          MultiAddress.init("/ip4/0.0.0.0/tcp/24").tryGet(),
+          MultiAddress.init("/ip4/0.0.0.0/tcp/25").tryGet(),
+        ]
+      services = @[ServiceInfo(id: "test_service", data: @[])]
+      routingRecord = ExtPeerRecord.init(peerId, multiAddresses, 42, services)
+
+      buffer = routingRecord.encode()
+
+      parsedRR = ExtPeerRecord.decode(buffer).tryGet()
```
gmelodie: nit: Why is this called `parsedRR` if it's a *peer* record? Should it not be `parsedPR`?

tests/libp2p/kademlia_discovery/test_ext_peer_records.nim
```diff
@@ -0,0 +1,151 @@
...
+
+    check:
+      parsedRR.peerId == peerId
+      parsedRR.seqNo == 42
+      parsedRR.addresses.len == 2
+      parsedRR.addresses[0].address == multiAddresses[0]
+      parsedRR.addresses[1].address == multiAddresses[1]
+      parsedRR.services[0].id == "test_service"
```
gmelodie: Instead of checking field by field, create a deterministic object and do `check parsedRR == expectedRR`, this is better than the current approach because in the current state if `parsedRR` has other fields that it shouldn't have or that are incorrectly set, the test will pass anyways (and it should fail)

tests/libp2p/kademlia_discovery/test_ext_peer_records.nim
```diff
@@ -0,0 +1,151 @@
...
+
+      parsedRR = SignedExtPeerRecord.decode(buffer).tryGet().data
+
+    check:
+      parsedRR.peerId == peerId
+      parsedRR.seqNo == 42
+      parsedRR.addresses.len == 2
+      parsedRR.addresses[0].address == multiAddresses[0]
+      parsedRR.addresses[1].address == multiAddresses[1]
+      parsedRR.services[0].id == "test_service"
```
gmelodie: Same as above

libp2p/protocols/kademlia_discovery/randomfind.nim
vladopajic: it never makes sense to duplicate code that has this much similarity, fixing issue or improving code in one place can leave other place vulnerable. and in code review process, reviewers can't really know that all duplicated code places, to guard and seek same changes everywhere.

that's why it is always the best approach to avoid any duplication right away.

tests/libp2p/kademlia_discovery/test_ext_peer_records.nim
```diff
@@ -0,0 +1,151 @@
+# Nim-LibP2P
```
gmelodie: For these tests, create a template like so
```nim
  template testEncodeDecode(toTest: auto) =
    check:
      toTest.encode().decode() == toTest
```

And all this template for all the types we're checking.

libp2p/protocols/kademlia_discovery/types.nim
```diff
@@ -0,0 +1,169 @@
...
+    if seqNo > maxSeqNo or bestIdx == -1:
+      maxSeqNo = seqNo
+      bestIdx = i
+
+  if bestIdx == -1:
+    return err("No valid records")
+
+  return ok(bestIdx)
+
+type KademliaDiscovery* = ref object of KadDHT
```
vladopajic: kind of the main type but burred here... would it be more elegant to list all types first, then their pros ?

libp2p/protocols/kademlia_discovery/types.nim
```diff
@@ -0,0 +1,169 @@
...
+  pb.finish()
+  pb.buffer
+
+type SignedExtPeerRecord* = SignedPayload[ExtPeerRecord]
+
+proc payloadDomain*(T: typedesc[ExtPeerRecord]): string =
+  $multiCodec("libp2p-peer-record")
+
+proc payloadType*(T: typedesc[ExtPeerRecord]): seq[byte] =
+  "/libp2p/logos-routing-record/".toBytes()
```
vladopajic: maybe const?

libp2p/protocols/kademlia_discovery/types.nim
```diff
@@ -0,0 +1,169 @@
...
+      servPb.write(2, service.data)
+    pb.write(4, servPb)
+
+  pb.finish()
+  pb.buffer
+
+type SignedExtPeerRecord* = SignedPayload[ExtPeerRecord]
+
+proc payloadDomain*(T: typedesc[ExtPeerRecord]): string =
+  $multiCodec("libp2p-peer-record")
```
vladopajic: maybe const?

libp2p/protocols/kademlia_discovery/types.nim
```diff
@@ -0,0 +1,169 @@
...
+  ?pb.getRequiredField(2, record.seqNo)
+
+  var addressInfos: seq[seq[byte]]
+  if ?pb.getRepeatedField(3, addressInfos):
+    for address in addressInfos:
+      var addressInfo = AddressInfo()
+      let subProto = initProtoBuffer(address)
+      let f = subProto.getField(1, addressInfo.address)
+      if f.get(false):
+        record.addresses &= addressInfo
```
vladopajic: `AddressInfo` type should be decoded using `AddressInfo.decode` proc. 
`record.addresses &= AddressInfo.decode(addressBytes)`

libp2p/protocols/kademlia_discovery/types.nim
```diff
@@ -0,0 +1,169 @@
...
+
+  var serviceInfos: seq[seq[byte]]
+  if ?pb.getRepeatedField(4, serviceInfos):
+    for service in serviceInfos:
+      var serviceInfo = ServiceInfo()
+      let subProto = initProtoBuffer(service)
+      let f = subProto.getField(1, serviceInfo.id)
+      discard ?subProto.getField(2, serviceInfo.data)
+      if f.get(false):
+        record.services &= serviceInfo
```
vladopajic: `ServiceInfo` type should be decoded using `ServiceInfo.decode` proc.

```record.services &= ServiceInfo.decode(serviceBytes)```

File: libp2p/protocols/kademlia_discovery/types.nim [hunk 36]
```diff
@@ -0,0 +1,169 @@
...
+proc encode*(record: ExtPeerRecord): seq[byte] =
+  var pb = initProtoBuffer()
+
+  pb.write(1, record.peerId)
+  pb.write(2, record.seqNo)
+
+  for address in record.addresses:
+    var addrPb = initProtoBuffer()
+    addrPb.write(1, address.address)
+    pb.write(3, addrPb)
```
vladopajic: ```pb.write(3, address.encode())```

libp2p/protocols/kademlia_discovery/types.nim
```diff
@@ -0,0 +1,169 @@
...
+
+  for service in record.services:
+    var servPb = initProtoBuffer()
+    servPb.write(1, service.id)
+    if service.data.len > 0:
+      servPb.write(2, service.data)
+    pb.write(4, servPb)
```
vladopajic:
```
pb.write(4, service.encode())
```

libp2p/protocols/kademlia_discovery/types.nim
SionoiS: > kind of the main type but burred here... would it be more elegant to list all types first, then their pros ?

I agree, types at the top would be better. To be honest I don't even like having a `types.nim` but I went with the same style.

libp2p/protocols/kademlia_discovery/randomfind.nim
SionoiS: If modifying existing code is fine then I will do that. I wanted to do other modifications anyway. :D

libp2p/protocols/kademlia_discovery/types.nim
SionoiS: I went with same style as existing code. All other record types work this way.

libp2p/protocols/kad_disco.nim
SionoiS: No objection on my side. @jm-clius @tinniaru3005 WDYT?

libp2p/protocols/kad_disco.nim
SionoiS: I disagree, since `putValue` is agnostic to the value stored. I think it's good that it only accept bytes.

But maybe you meant that I should add another abstraction in this case?  egg `refresh() -> putRecord() -> putValue()`

libp2p/protocols/kad_disco.nim
SionoiS: Yes that is how it works right now. Do you mean the parameter should be a hash set too?

libp2p/protocols/kad_disco.nim
SionoiS: Ah thanks I didn't know you could do that in Nim.

libp2p/protocols/kad_disco.nim
jm-clius: Although we do use this to discover "capabilities", the literature generally use "services" here when referring to the discovery protocol (see for example DISC-NG: https://ieeexplore.ieee.org/document/10629017). In other words, we allow capability discovery by mapping capabilities to service IDs in a service discovery protocol. I'm OK with switching to using either "capability" or "service" throughout, but I think we'll do better to stick to the one most idiomatic which seems to me to be "services".

libp2p/protocols/kademlia_discovery/types.nim
SionoiS: spec is WIP https://github.com/vacp2p/rfc-index/pull/220

I will add a `protobuf.nim`

libp2p/protocols/kad_disco.nim
gmelodie: I'm fine with whatever, but
- I find "disco" confusing
- either we use services or capabilities, as long as we're not mixing these two I'm fine

libp2p/protocols/kad_disco.nim
gmelodie: Let's leave it the way it is then, this is a nitpick anyway

libp2p/protocols/kad_disco.nim
gmelodie: My bad, was looking at another place in the code where you convert it to a `seq`. This is fine then.

tests/libp2p/kademlia_discovery/test_ext_peer_records.nim
SionoiS: yeah the name changed multiple times :rofl:  it was a routing record...

tests/libp2p/kademlia_discovery/test_ext_peer_records.nim
SionoiS: Wizardry... :rofl:

Very useful thanks!

SionoiS:
> @SionoiS I do think at this point it might make sense to open a parallel (basic) RFC to give the necessary context of what we're implementing here, especially as this is a precursor to actual capability (service?) discovery. It will also provide us with a good central place to settle on terms, such as service vs capability, etc.

https://github.com/vacp2p/rfc-index/pull/250

libp2p/protocols/kademlia_discovery/types.nim
```diff
@@ -0,0 +1,169 @@
...
+proc toKey*(service: ServiceInfo): Key =
+  return MultiHash.digest("sha2-256", service.id.toBytes()).get().toKey()
+
+proc init*(
+    T: typedesc[ExtPeerRecord],
+    peerInfo: PeerInfo,
+    seqNo: uint64 = getTime().toUnix().uint64,
+    services: seq[ServiceInfo] = @[],
+): T =
+  ExtPeerRecord(
```
richard-ramos: nit: Since T = ExtPeerRecord, you can use T directly
```suggestion
  T(
```

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -0,0 +1,80 @@
...
+
+      getValueFutures[peerId] = disco.switch
+        .dispatchGetVal(peerId, peerId.toKey(), received, candidates)
+        .wait(disco.config.timeout)
+
+    await allFutures(getValueFutures.values.toSeq())
+    for peerId, fut in getValueFutures.pairs:
+      try:
+        await fut
+      except CatchableError:
```
richard-ramos: This except should be more specific. Likely we should use a LPError here (otherwise we might end up capturing a CancelledError)

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -0,0 +1,80 @@
...
+  for entryOpt in received.values:
+    let entry = entryOpt.valueOr:
+      continue
+
+    let spr = ExtPeerRecord.decode(entry.value).valueOr:
+      debug "cannot decode signed peer record", error
+      continue
+
+    records.add(spr)
+
```
richard-ramos, line 59: The received values are not being validated. Should they be? or just decoding the ExtPeerRecord is enough to consider the value valid?

SionoiS: Yes, decoding also verify.

libp2p/protocols/kademlia_discovery/randomfind.nim
SionoiS: I'm not super well versed in how to handle exceptions. So far the only thing that compiles is `CatchableError`

libp2p/protocols/kad_disco.nim
tinniaru3005: Conceptually, what we’re discovering are capabilities, but I agree with @jm-clius that at the protocol level this is really a service discovery mechanism, and mapping capabilities → service IDs fits well with existing literature and prior art.

I don’t have a strong preference either way as long as we pick one term and use it consistently across code, APIs, and docs. Given that, I’m slightly in favor of sticking with “service” since it’s more idiomatic in discovery protocols and avoids re-inventing terminology.

SionoiS: alright service it is!

libp2p/protocols/kademlia_discovery/types.nim
```diff
@@ -0,0 +1,115 @@
...
+import ../kademlia/types
+
+const DefaultSelfSPRRereshTime* = 10.minutes
+
+type
+  ServiceInfo* = object
+    id*: string
+    data*: seq[byte]
+
+  ExtPeerRecord* = object
```
jm-clius: I know long type names are controversial, but here I'd suggest the more explicit `ExtensiblePeerRecord` to avoid ambiguity ("ext" sounds too much like "external"). Also in the spirit of related fully qualified type names, e.g. `SignedPeerRecord` here: https://github.com/vacp2p/nim-libp2p/blob/f6d9feb0c9de848cfb77e5b013f9e6fd76231c3a/libp2p/routing_record.nim#L78

jm-clius: Since `ExtensiblePeerRecord` is a separate protocol with a separate specification, I think the type definition and protobuf codec can all be extracted to a `extensible_peer_record.nim` module separate from `kademlia_discovery`. We also plan to reuse it for capability discovery, incorporate into `identify`, etc.
The model here is `routing_record.nim`.

libp2p/protocols/kademlia_discovery/types.nim
```diff
@@ -0,0 +1,115 @@
...
+    services: services,
+  )
+
+type SignedExtPeerRecord* = SignedPayload[ExtPeerRecord]
+
+proc payloadDomain*(T: typedesc[ExtPeerRecord]): string =
+  $multiCodec("libp2p-peer-record")
+
+proc payloadType*(T: typedesc[ExtPeerRecord]): seq[byte] =
+  "/libp2p/logos-routing-record/".toBytes()
```
jm-clius: Goes against what I suggested for the RFC, but I'm now thinking we should stick to using the exact same `payloadDomain` and `payloadType` as libp2p uses for signed peer records.
Reasoning: the protobufs are backwards compatible and if we plan to use XPRs in the `signedPeerRecord` field in `identify` we want to maintain compatibility (even if not all clients know how to parse the `serviceInfo` field, which they'll ignore by definition). WDYT?

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -0,0 +1,77 @@
...
+# those terms.
+
+import std/[tables, sequtils, sets]
+import chronos, chronicles, results
+import ../../[peerid, peerinfo, switch, multihash]
+import ../protocol
+import ../kademlia/[types, find, get, protobuf]
+import ./[types, protobuf]
+
+proc findRandom*(
```
jm-clius: Comment has been made before, but after a quick look at the underlying kademlia `findNode` method it seems we can reuse that with the slight modification that iterative `closerPeers` are streamed to us as consumers. I think it shouldn't be too complex to return these in an `AsyncQueue` (or alternatively implementing a callback).
Once the soon-to-be-merged RFC is published, we can discuss with the nim-libp2p team as to what the most idiomatic approach will be for this codebase.

libp2p/protocols/kad_disco.nim
```diff
@@ -0,0 +1,86 @@
...
+  let putRes = await disco.putValue(key, encodedSR)
+  if putRes.isErr:
+    error "Failed to put signed peer record", err = putRes.error
+
+proc maintainSelfSignedPeerRecord(
+    disco: KademliaDiscovery
+) {.async: (raises: [CancelledError]).} =
+  heartbeat "refresh self signed peer record", disco.config.bucketRefreshTime:
+    await disco.refreshSelfSignedPeerRecord()
+
```
jm-clius, line 51: If this module contains the consumer interface for Kademlia discovery, I think we need a public function to the tune of:

```suggestion
proc discover*(disco: KademliaDiscovery, seq[ServiceID]: serviceIds): Future[seq[PeerInfo]]
## Triggers random discovery process. Filters discovered peers by serviceID.
```

libp2p/protocols/kad_disco.nim
```diff
@@ -0,0 +1,86 @@
...
+) {.async: (raises: [CancelledError]).} =
+  await disco.switch.peerInfo.update()
+
+  let
+    peerInfo: PeerInfo = disco.switch.peerInfo
+    services: seq[ServiceInfo] = disco.services.toSeq()
+
+  let extPeerRecord = SignedExtPeerRecord.init(
+    peerInfo.privateKey,
+    ExtPeerRecord(
```
jm-clius: Wondering if it makes sense to compute the `ExtPeerRecord` within KademliaDiscovery or if it should be a record computed by the consumer of this module and provided here. This is because the XPR is planned to be used externally also in `identify` and hypothetically for other possible discovery methods (cf. the ENR being computed inside discv5 and the issues that caused with separability).

SionoiS: Services come from outside but not the record itself.

The only thing a consumer can do is add or remove a service. This abstraction prevent the consumer from messing with record internals.

libp2p/protocols/kademlia_discovery/randomfind.nim
SionoiS: Lets wait a bit, libp2p folks are changing this part of the code anyway.

libp2p/protocols/kad_disco.nim
SionoiS: I'll add a separate filter proc but same idea yeah.

libp2p/protocols/kad_disco.nim
jm-clius: Yes, perhaps too early to think of separation. Agree that we should certainly provide an abstraction where the consumer does not mess with the internals of the record, but I think that will be the "peer management" module that we plan for the future. This should bring together the management of discoverable records and capabilities/services for a node and the different discovery methods. To me it wouldn't make sense if the XPR belongs to one of these discovery methods. How would we for example distribute the XPR in `identify`? I don't think we want to create a coupling between identify and a specific discovery protocol. This is similar to how `enr` being created in discv5 provided abstraction, but eventually caused difficulty when a node's ENR had meaning outside of the discovery protocol.

.gitignore
```diff
@@ -1,5 +1,7 @@
 nimcache/
 nimbledeps/
+CLAUDE.md
+.claude/*
```
gmelodie: This is something to put inside `.git/info/exclude`, not `.gitignore`

SionoiS: I am S-P of clankers, resistance is futile, you will assimilated into an AI workflow. :rofl: 

Anyway... If it's not useful for you guys I'll remove it.

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -0,0 +1,72 @@
...
+      continue
+
+    records.add(sxpr.data)
+
+  return records
+
+proc filterByServices(
+    records: seq[ExtendedPeerRecord], services: HashSet[ServiceInfo]
+): seq[ExtendedPeerRecord] =
+  records.filterIt(it.services.anyIt(services.contains(it)))
```
gmelodie: This doesn't seem to be used anywhere.

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -0,0 +1,72 @@
...
+
+  var records: seq[ExtendedPeerRecord]
+  for buffer in buffers:
+    let sxpr = SignedExtendedPeerRecord.decode(buffer).valueOr:
+      debug "cannot decode signed extended peer record", error
+      continue
+
+    records.add(sxpr.data)
+
+  return records
```
gmelodie: This is the only real thing keeping me from giving the green lights on this PR, the change should be done to `findNode` directly, not duplicating code.

cc: @richard-ramos who can always overule me and push if he sees fit :)

tests/libp2p/kademlia_discovery/utils.nim
```diff
@@ -0,0 +1,38 @@
...
+{.used.}
+
+import results, chronos
+import ../../../libp2p/[switch, builders]
+import ../../../libp2p/protocols/kad_disco
+
+proc createSwitch*(): Switch =
+  SwitchBuilder
+  .new()
+  .withRng(newRng())
```
gmelodie: Use `rng` from `utils/crypto.nim`
```suggestion
  .withRng(rng())
```

libp2p/routing_record.nim
```diff
@@ -68,17 +159,49 @@ proc init*(
...
+proc payloadType*(T: typedesc[ExtendedPeerRecord]): seq[byte] =
+  @[(byte) 0x03, (byte) 0x01]
+
 proc checkValid*(spr: SignedPeerRecord): Result[void, EnvelopeError] =
   if not spr.data.peerId.match(spr.envelope.publicKey):
     err(EnvelopeInvalidSignature)
   else:
     ok()
+
+proc checkValid*(spr: SignedExtendedPeerRecord): Result[void, EnvelopeError] =
```
gmelodie: nit
```suggestion
proc valid*(spr: SignedExtendedPeerRecord): Result[void, EnvelopeError] =
```
or
```suggestion
proc isValid*(spr: SignedExtendedPeerRecord): Result[void, EnvelopeError] =
```

libp2p/protocols/kademlia_discovery/randomfind.nim
SionoiS: I forgot the access should be public

libp2p/protocols/kademlia_discovery/randomfind.nim
SionoiS: I don't understand what you mean, both `findNode` and `findRandom` have different logic.

Where is the duplication?

libp2p/routing_record.nim
SionoiS: Must be called `checkValid` because of the logic in `signed_envelope.nim` otherwise it's not compatible.

libp2p/protocols/kademlia_discovery/randomfind.nim
SionoiS: I talked with @jm-clius about it and now I understand, I think.

gmelodie (approved):

PR #1922, opened 2025-11-28
test(coverage): ignore tests

rlve:
Changes:
- Ignore `tests` in `Codecov` report.

### DIFF
--- a/tests/stublogger.nim
+++ b/tests/stublogger.nim
@@ -14,11 +14,13 @@ when not defined(nimscript):
   when defined(chronicles_runtime_filtering):
     setLogLevel(INFO)
 
-  when defaultChroniclesStream.outputs.type.arity == 1:
-    # Hide the json logs, they're just here to check if we compile
-    proc noOutput(logLevel: LogLevel, msg: LogOutputStr) =
-      discard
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -11,3 +11,7 @@ coverage:
         target: auto
         threshold: 5%
         base: auto
+ignore:
+  - "tests/**"
+  - "examples/**"
+  - "cbind/**"
```
vladopajic: i think we should have some cbind tests in future, wdyt @richard-ramos and @gmelodie?

rlve: Right, I'll remove cbind from the list

vladopajic: and should `/performance` be excluded from coverage? and what about `/interop`  and `/tools`?

richard-ramos: > i think we should have some cbind tests in future, wdyt @richard-ramos and @gmelodie?

Yes, currently there aint none is because bindings are on early development stage

rlve: Only `tests` got picked up by Codecov at some point, so there's not need to add any other directories to the ignore list. I guess in that case I'll remove `examples` as well.

Issue #1805, opened 2025-10-28
test: `nim c tests/testall` fails if compile time flags are not provided


vladopajic:
command `nim c tests/testall` fails if compile time flags are not provided.  probably `libp2p_mix_experimental_exit_is_dest` is causing issues.

1) command  `nim c tests/testall` must always compile successfully without specifying any additional options. code needs to be fixed.
2) should there be ci workflow that ensures that `nim c tests/testall` works? daily at least? this task will ensure that default configuration of `nim-libp2p` can be used. 
3) additionally binary of `nim c tests/testall` command can be run to ensure that tests are passing when compile flags are not used.

PR #1929, opened 2025-12-03
test: ensure tests run without flags

rlve:
Addresses: [test: nim c tests/testall fails if compile time flags are not provided](https://github.com/vacp2p/nim-libp2p/issues/1805)

### DIFF
--- a/.github/workflows/daily_tests_no_flags.yml
+++ b/.github/workflows/daily_tests_no_flags.yml
@@ -1,4 +1,4 @@
-name: Daily tests without feature flags
+name: Daily without feature flags
 
 on:
   schedule:


vladopajic (approved):

.github/workflows/tests_no_flags.yml
```diff
@@ -0,0 +1,62 @@
+name: Tests without feature flags
```
vladopajic: wdyt `Daily without feature flags` ?

vladopajic: image

vladopajic: ^ would be nicer for visual grouping in this list

rlve: Sure, renamed.

PR #1930, opened 2025-12-03
test(gossipsub): make test more robust

vladopajic:
various changes in gossip sub tests:
- reduced usage of `waitForHeartbeat` - since this can be flaky 
- adding delay after connecting nodes - it is needed for node to run all handlers, open transport streams etc..

in following prs i will make wait after subscribe to be more universal. currently code uses different ways to check this.

### DIFF
--- a/libp2p/protocols/pubsub/pubsub.nim
+++ b/libp2p/protocols/pubsub/pubsub.nim
@@ -398,7 +398,7 @@ method getOrCreatePeer*(
     except CancelledError as exc:
       raise exc
     except DialFailedError as e:
-      raise (ref GetConnDialError)(parent: e)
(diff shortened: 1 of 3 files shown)


tests/libp2p/pubsub/test_all.nim
```diff
@@ -14,6 +14,3 @@ import
   test_scoring, test_timed_cache, test_preamble_store
 
 import ./component/test_all
-
-when defined(libp2p_gossipsub_1_4):
-  import test_preamble_behavior
```
vladopajic, line 19: duplicate of `test_gossipsub_control_messages.nim`  -> `emit IMReceiving while handling preamble control msg`

richard-ramos (approved):

PR #1934, opened 2025-12-04
chore: allFuturesThrowing renamed to allFuturesRaising

vladopajic:
- rename `allFuturesThrowing` -> `allFuturesRaising`
  - initially PR's intention was to rename it to `allFuturesDiscarding`, but `Raising` has more sense based on https://github.com/vacp2p/nim-libp2p/pull/1934#discussion_r2592984943
- adding `Raising` to return type
- removed unnecessary call to `allFutures / allFinished` within `allFuturesThrowing`

### DIFF
--- a/tests/tools/futures.nim
+++ b/tests/tools/futures.nim
@@ -14,7 +14,9 @@ proc completedFuture*(): Future[void] =
   f.complete()
   f
 
-proc allFuturesThrowing*(args: varargs[FutureBase]): Future[void] =
+proc allFuturesRaising*(
+    args: varargs[FutureBase]
+): Future[void].Raising([CatchableError]) =
   # This proc is only meant for use in tests / not suitable for general use.
   # Swallowing errors arbitrarily instead of aggregating them is bad design
   # It raises `CatchableError` instead of the union of the `futs` errors,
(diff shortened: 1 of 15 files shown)


tests/tools/futures.nim
```diff
@@ -14,11 +14,11 @@ proc completedFuture*(): Future[void] =
   f.complete()
   f
 
-proc allFuturesThrowing*(args: varargs[FutureBase]): Future[void] =
+proc allFuturesDiscarding*(args: varargs[FutureBase]): Future[void] =
```
arnetheduck: "throwing" here refers to the fact that if any one of the futures given in `args` fails due to an exception, that exception will be raised by `await allFuturesThrowing(futs)` - "discarding" does not really represent this operation well.

semantically, this function performs the following operations:
```nim
var firstError: CatchableError
for f in futs:
  try:
    discard await fut
  except CatchableError as exc:
    if firstError != nil: firstError = exc
if firstError != nil:
  raise firstError
```

vladopajic: aha!

arnetheduck: (also, to nitpick, in nim we "raise", we don't "throw")

vladopajic: i have reverted renamed, leaving only other cosmetically changes.

vladopajic: > in nim we "raise"

somehow i totally missed this point. yes `Raising` makes sense more then `Throwing` i have renamed it accordingly.

tests/tools/futures.nim
```diff
@@ -14,11 +14,13 @@ proc completedFuture*(): Future[void] =
   f.complete()
   f
 
-proc allFuturesThrowing*(args: varargs[FutureBase]): Future[void] =
+proc allFuturesRaising*(
+    args: varargs[FutureBase]
+): Future[void].Raising([CatchableError]) =
```
vladopajic, line 19: also added Raising to return type

Issue #1856, opened 2025-11-07
chore: extract task logic from libp2p.nimble into their own files


richard-ramos:
should we consider having `libp2_library.nim` file where we add all bits and picies for library and include it in `libp2p.nimble`?

we can have `libp2p_test.nim` similarly. 

we end up with better organization of `libp2p.nimble`.

_Originally posted by @vladopajic in https://github.com/vacp2p/nim-libp2p/pull/1820#discussion_r2477645961_

gmelodie:
I don't think so, given that `libp2p.nimble` is less than 200 LoC long

Issue #1854, opened 2025-11-07
Move examples build code out of libp2p.nimble


richard-ramos:
why have it here?

`task examples` should not even be here it should be in `/examples` dir.
similarly to how `task performance` and `task reliability` are in `/peformance` dir.

if we go down that route, let's have a separate nimble file for `buildLibrary` and whatever that is need here.

_Originally posted by @vladopajic in https://github.com/vacp2p/nim-libp2p/pull/1820#discussion_r2477625160_

PR #1939, opened 2025-12-05
chore: organise nimble files

rlve:
Attempt to address: 
 - [Move examples build code out of libp2p.nimble](https://github.com/vacp2p/nim-libp2p/issues/1854)
 - [chore: extract task logic from libp2p.nimble into their own files](https://github.com/vacp2p/nim-libp2p/issues/1856)
 
Changes:
- add `cbind/cbind.nimble` and `examples/examples.nimble` and move tasks from `libp2p.nimble`
- update GH workflows: `documentation` and `examples`
- add GH workflow `cbindings`
- use local `libp2p` in `interop/autonatv2/nim-peer/nim_peer.nimble` instead of installing the package

### DIFF
--- a/tests/libp2p/pubsub/component/test_gossipsub_mesh_management.nim
+++ b/tests/libp2p/pubsub/component/test_gossipsub_mesh_management.nim
@@ -206,29 +206,29 @@ suite "GossipSub Component - Mesh Management":
 
     startNodesAndDeferStop(nodes)
 
-    # When nodes subscribe to multiple topics
     await connectNodesStar(nodes)
-    for topic in topics:
-      let t = topic
-      subscribeAllNodes(nodes, t, voidTopicHandler)
+
+    # When nodes subscribe to multiple topics
+    for t in topics:
+      let topic = t
+      subscribeAllNodes(nodes, topic, voidTopicHandler)
 
     # Then all nodes should be subscribed to the topics initially
-    for i in 0 ..< topics.len:
-      let topic = topics[i]
+    for t in topics:
+      let topic = t
       checkUntilTimeout:
         nodes.allIt(it.topics.contains(topic))
         nodes.allIt(it.gossipsub.getOrDefault(topic).len() == numberOfNodes - 1)
         nodes.allIt(it.mesh.getOrDefault(topic).len() == numberOfNodes - 1)
 
     # When they unsubscribe from all topics
-    for topic in topics:
-      let t = topic
-      unsubscribeAllNodes(nodes, t, voidTopicHandler)
+    for t in topics:
+      let topic = t
+      unsubscribeAllNodes(nodes, topic, voidTopicHandler)
 
     # Then topics should be removed from mesh and gossipsub
-    for i in 0 ..< topics.len:
-      let topic = topics[i]
-      checkUntilTimeout:
+    for topic in topics:
+      check:
         nodes.allIt(not it.topics.contains(topic))
         nodes.allIt(topic notin it.gossipsub)
         nodes.allIt(topic notin it.mesh)
--- a/tests/libp2p/pubsub/component/test_gossipsub_message_handling.nim
+++ b/tests/libp2p/pubsub/component/test_gossipsub_message_handling.nim
@@ -34,7 +34,7 @@ proc setupTest(): Future[
     receivedMessages[].incl(data)
 
   proc handlerB(topic: string, data: seq[byte]) {.async.} =
-    discard
+    raiseAssert "should not be called"
 
   nodes[0].subscribe("foobar", handlerA)
   nodes[1].subscribe("foobar", handlerB)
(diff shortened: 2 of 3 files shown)


examples/examples.nimble
```diff
@@ -0,0 +1,47 @@
...
+packageName = "examples"
+version = "1.0.0"
+author = "Status Research & Development GmbH"
+description = "Examples for LibP2P implementation"
+license = "MIT"
+
+# Dependencies are inherited from parent libp2p.nimble via nimble.paths
+# We don't need `requires` here since we run tasks, not build a package
+
+include "../common.nims"
```
vladopajic: this is probably not needed for examples. we should build examples with all defaults?

common.nims
```diff
@@ -0,0 +1,11 @@
...
+
+let nimc* = getEnv("NIMC", "nim") # Which nim compiler to use
+let lang* = getEnv("NIMLANG", "c") # Which backend (c/cpp/js)
+let flags* = getEnv("NIMFLAGS", "") # Extra flags for the compiler
+let verbose* = getEnv("V", "") notin ["", "0"]
+
+let cfg* =
+  " --styleCheck:usages --styleCheck:error" &
+  (if verbose: "" else: " --verbosity:0 --hints:off") &
+  " --skipUserCfg -f --threads:on --opt:speed"
```
vladopajic: would be better if we can keep this in `libp2p.nimble` only.

performance/test_node/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM nimlang/nim:2.2.4-alpine-regular AS build
 
 WORKDIR /node
 
-COPY libp2p.nimble config.nims ./
+COPY libp2p.nimble common.nims config.nims ./
```
vladopajic: for now it would be nicer if `.nimble` files are all self contained so we don't have to do this. we dont know all places where this can cause problems to downstream dependencies.

vladopajic: i would suggest duplicating code in all nimble files over doing this

cbind/cbind.nimble
```diff
@@ -0,0 +1,53 @@
...
+version = "0.1.0"
+author = "Status Research & Development GmbH"
+description = "C bindings for LibP2P implementation"
+license = "MIT"
+
+# Dependencies are inherited from parent libp2p.nimble via nimble.paths
+# We don't need `requires` here since we run tasks, not build a package
+
+include "../common.nims"
+
```
vladopajic: cbind has introduced dependency 
https://github.com/vacp2p/nim-libp2p/pull/1820#discussion_r2504073979
would it be possible to require it here, and remove it from libp2p-library

cbind/cbind.nimble
```diff
@@ -0,0 +1,53 @@
...
+packageName = "cbind"
+version = "0.1.0"
+author = "Status Research & Development GmbH"
+description = "C bindings for LibP2P implementation"
+license = "MIT"
+
+# Dependencies are inherited from parent libp2p.nimble via nimble.paths
+# We don't need `requires` here since we run tasks, not build a package
+
+include "../common.nims"
```
vladopajic: same as examples, i dont think we need it here

cbind/cbind.nimble
```diff
@@ -0,0 +1,53 @@
...
+    " --opt:size --noMain --mm:refc --header --undef:metrics" &
+    " --nimMainPrefix:libp2p --nimcache:nimcache" & " -d:asyncTimer=system " & cfg & " " &
+    flags & " libp2p.nim"
+
+task libDynamic, "Generate dynamic bindings":
+  buildCBindings "dynamic", ""
+
+task libStatic, "Generate static bindings":
+  buildCBindings "static", ""
+
```
richard-ramos, line 45: These rules don’t belong in cbind.nimble. They should stay in libp2p.nimble. If someone wants to integrate nim-libp2p via ffi, they shouldn’t have to get into a specific library directory just to figure out how to produce a static or dynamic library. that logic belongs at the project level.

Having said that, I guess if we document this in the README it should be fine, but not entirely sure :thinking:

richard-ramos: Hm... thinking it a bit more thoroughly, i think this is fine, we just need to document in the README how to generate a library. (That's of course out of scope for this PR).

rlve: Also, since this PR separates `cbind example` from `examples`, we need a workflow to build `c bindings`.

examples/examples.nimble
rlve: Done in https://github.com/vacp2p/nim-libp2p/pull/1939/commits/2deb91f39592083cf29475b9885bd7b34b1e4c42

common.nims
rlve: Done in https://github.com/vacp2p/nim-libp2p/pull/1939/commits/2deb91f39592083cf29475b9885bd7b34b1e4c42

performance/test_node/Dockerfile
rlve: Done in https://github.com/vacp2p/nim-libp2p/pull/1939/commits/2deb91f39592083cf29475b9885bd7b34b1e4c42

cbind/cbind.nimble
rlve: All right, moved `taskpools` dependency to `cbind.nimble` in https://github.com/vacp2p/nim-libp2p/pull/1939/commits/4bdbca58bb3c2937c2d5ec1bdcd812548c2aeaf7

libp2p.nimble
```diff
@@ -137,7 +92,6 @@ task pin, "Create a lockfile":
   exec nimc & " c -r tools/pinner.nim"
 
 import sequtils
```
vladopajic: nit: not sure about this import here. i think we should keep it on the top of file (maybe below `requires` block).

richard-ramos (approved):

PR #1942, opened 2025-12-06
test(gossipsub): various improvements

vladopajic:
- remove unused code
- refactor tests to use existing utility functions
- code tidy
- other improvements

### DIFF
--- a/.github/workflows/daily_tests_no_flags.yml
+++ b/.github/workflows/daily_tests_no_flags.yml
@@ -1,0 +1,62 @@
+name: Daily without feature flags
+
+on:
+  schedule:
+    - cron: "0 12 * * *"
+  workflow_dispatch:
+
+jobs:
+  tests-no-flags:
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+
+    defaults:
+      run:
+        shell: bash
+
+    name: "Build and run"
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


tests/libp2p/pubsub/component/test_gossipsub_message_handling.nim
```diff
@@ -66,6 +66,26 @@ proc createMessages(
...
+    nodes: seq[GossipSub], numPeersFirstMsg: int, numPeersSecondMsg: int
+) {.async.} =
+  block setup:
+    for i in 0 ..< 50:
+      if (await nodes[0].publish("foobar", ("Hello!" & $i).toBytes())) == nodes.len - 1:
+        break setup
+      await sleepAsync(100.milliseconds)
+    raiseAssert "Failed to publish message to peers"
+
+  check (await nodes[0].publish("foobar", newSeq[byte](2_500_000))) == numPeersFirstMsg
```
rlve, line 79: Should we make "foobar" a const and reuse across the file?

vladopajic: to me it looks better if all functions have it as argument. and then we pass whatever value from test. 
value in test can be shared const or local to test const. but since most test used later then, better have it that way.

i am not going down that rabbit hole atm. feel free to push pr.

PR #1936, opened 2025-12-04
test(gossipsub): making tests more robust pt3

vladopajic:
adds to previous prs: #1930,  #1933

in this pr, wait for subscription is improved

PR #1933, opened 2025-12-04
test(gossipsub): making tests more robust pt2

vladopajic:
adds to previous pr #1930

PR #1947, opened 2025-12-09
test(gossipsub): making tests more robust pt5

vladopajic:
adds to previous prs: #1930, #1933, #1936, #1938

fixes some flakiness, caused by:
- small heartbeat intervals 
- inadequate wait on subscribe condition

### DIFF
--- a/.github/workflows/cbindings.yml
+++ b/.github/workflows/cbindings.yml
@@ -1,4 +1,4 @@
-name: Examples
+name: C bindings examples
 
 on:
   push:
@@ -13,7 +13,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  examples:
+  c-bindings:
     timeout-minutes: 30
     strategy:
       fail-fast: false
@@ -22,7 +22,7 @@ jobs:
       run:
         shell: bash
 
-    name: "Build Examples"
+    name: "Build C bindings"
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout
@@ -50,11 +50,13 @@ jobs:
         run: |
           nimble install_pinned
 
-      - name: Build and run examples
+      - name: Build and run c bindings examples
         run: |
           nim --version
           nimble --version
           gcc --version
 
           nimble setup
-          cd examples && nimble examples
+          cd cbind
+          nimble setup
+          nimble examples
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -77,7 +77,7 @@ jobs:
           nim-version: 'stable'
 
       - name: Generate website
-        run: pip install mkdocs-material && nimble -y website
+        run: pip install mkdocs-material && cd examples && nimble -y website
 
       - name: Clone the gh-pages branch
         uses: actions/checkout@v4
--- a/.github/workflows/interop.yml
+++ b/.github/workflows/interop.yml
@@ -85,7 +85,7 @@ jobs:
 
       - name: Run Go and Nim together
         run: |
-          nimble install
+          nimble setup
           cd interop/autonatv2/go-peer
           git clone https://github.com/libp2p/go-libp2p
           cd go-libp2p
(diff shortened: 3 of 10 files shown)


tests/libp2p/pubsub/component/test_gossipsub_message_cache.nim
```diff
@@ -138,6 +138,9 @@ suite "GossipSub Component - Message Cache":
     let nodes = generateNodes(
         numberOfNodes,
         gossip = true,
+        heartbeatInterval = 300.milliseconds,
+          # A larger heartbeat interval is required because `checkUntilTimeout` may skip over
+          # the current heartbeat’s condition which is waited on
```
rlve: What do you mean by that?

vladopajic: `checkUntilTimeout` happens every 50ms while `heartbeatInterval` is 60ms. 
in these cases we need to assert what happens in  heartbeat following some action, for example:

action
heartbeat
now assert condition

but since `heartbeatInterval` is small, `checkUntilTimeout` sometimes skips heartbeat current (it get's on time for next heartbeat):


action
heartbeat no1
heartbeat no2
now assert condition.  it's never true because we need to assert what is happening in heartbeat no1 not heartbeat no2

rlve: ah, right, understood

but I would try to explain this better in the comment, that you're making sure that only 1 heartbeat happens before assertion

vladopajic: https://github.com/vacp2p/nim-libp2p/pull/1947/commits/cd8eeab0cea1a1813df06b14f9c5f3560b87f364

@rlve please review. 

1. removed custom `historyGossip` and `historyLength` - it does not seem needed here? with removing these we don't need custom heartbeart in this test
2. for other test we still need higher heartbeat. comment updated. if you can make comment better please ```suggest

tests/libp2p/pubsub/component/test_gossipsub_message_cache.nim
```diff
@@ -226,7 +224,16 @@ suite "GossipSub Component - Message Cache":
     const
       numberOfNodes = 3
       topic = "foobar"
-    let nodes = generateNodes(numberOfNodes, gossip = true).toGossipSub()
+    let nodes = generateNodes(
+        numberOfNodes,
+        gossip = true,
+        heartbeatInterval = 300.milliseconds,
+          # Becasue default heartbeat interval in tests is small (60ms) and vary close to `checkUntilTimeout`
```
rlve:
```suggestion
          # Becasue default heartbeat interval in tests is small (60ms) and very close to `checkUntilTimeout` 
```

tests/libp2p/pubsub/component/test_gossipsub_message_cache.nim
```diff
@@ -226,7 +224,16 @@ suite "GossipSub Component - Message Cache":
...
       numberOfNodes = 3
       topic = "foobar"
-    let nodes = generateNodes(numberOfNodes, gossip = true).toGossipSub()
+    let nodes = generateNodes(
+        numberOfNodes,
+        gossip = true,
+        heartbeatInterval = 300.milliseconds,
+          # Becasue default heartbeat interval in tests is small (60ms) and vary close to `checkUntilTimeout` 
+          # check interval (50ms). It can happen that two heartbeats happen before asserting.
+          # To prevent this from happeing, `heartbeatInterval` interval is increased here to ensure that only
```
rlve:
```suggestion
          # To prevent this from happening, `heartbeatInterval` interval is increased here to ensure that only
```

tests/libp2p/pubsub/component/test_gossipsub_message_cache.nim
```diff
@@ -226,7 +224,16 @@ suite "GossipSub Component - Message Cache":
...
+          # one heartbeat interval is happeing beofre assertion.
```
rlve:
```suggestion
          # one heartbeat interval is happening before assertion.
```

PR #1955, opened 2025-12-09
feat(cbind): connected peers

richard-ramos:


cbind/libp2p.nim
```diff
@@ -263,6 +263,41 @@ proc libp2p_peerinfo(
...
+): cint {.dynlib, exportc, cdecl.} =
+  initializeLibrary()
+  checkLibParams(ctx, callback, userData)
+
+  let direction =
+    case dir
+    of uint32(Direction.In):
+      Direction.In
+    of uint32(Direction.Out):
+      Direction.Out
```
vladopajic:
```nim
let direction = directionFromUint32(dir).valueOr:
      let msg = "invalid direction: " & $dir
      callback(RET_ERR.cint, nil, 0, msg[0].addr, cast[csize_t](len(msg)), userData)
      return RET_ERR.cint
```

create `directionFromUint32` utility

Issue #1944, opened 2025-12-09
refactor(test): create utilities that wait for nodes to subscribe


vladopajic:
currently tests cases have code like this:

```nim
    await connectNodesHub(node0, nodes[1 .. ^1])

    subscribeAllNodes(nodes, topic, voidTopicHandler)
    checkUntilTimeout:
      node0.gossipsub.getOrDefault(topic).len == numberOfNodes - 1
      nodes[1 .. ^1].allIt(it.gossipsub.getOrDefault(topic).len == 1)
```

where test case connects and subscribes all nodes to the topic then with `checkUntilTimeout` it waits for nodes to be subscribed (or it fails when condition is not meet).

it would be better to refactor this code to have following:
```nim
    await connectNodesHub(node0, nodes[1 .. ^1])
   
    subscribeAllNodes(nodes, topic, voidTopicHandler)
    waitSubscribeHub(node0, nodes[1 .. ^1])
```

there should `waitSubscribeXXX` for every `connectNodesXXX` proc:
- `waitSubscribeStar`
- `waitSubscribeHub`
- `waitSubscribeChain`
- `waitSubscribeStartRing`

for code that something special, other then above ways of connecting, we should leave it as is current custom `checkUntilTimeout` code.

PR #1958, opened 2025-12-10
test(gossipsub): utilities that wait for nodes to subscribe

rlve:
Addresses: [refactor(test): create utilities that wait for nodes to subscribe](https://github.com/vacp2p/nim-libp2p/issues/1944)


tests/libp2p/pubsub/utils.nim
```diff
@@ -358,6 +358,33 @@ proc connectNodesSparse*[T: PubSub](nodes: seq[T], degree: int = 2) {.async.} =
...
+  ## 
+  checkUntilTimeout:
+    hub.gossipsub.getOrDefault(topic).len == nodes.len
+    nodes.allIt(it.gossipsub.getOrDefault(topic).len == 1)
+
+proc waitSubscribeStar*[T: PubSub](nodes: seq[T], topic: string) {.async.} =
+  ## Star: 1-2; 1-3; 2-1; 2-3, 3-1, 3-2
+  ## 
+  checkUntilTimeout:
+    nodes.allIt(it.gossipsub.getOrDefault(topic).len == nodes.len - 1)
```
vladopajic: could we make these as `template` instead of `proc` ? 
we don't really need these as procs  and template results in code without `await` prefix:
`await waitSubscribeStar` -> `waitSubscribeStar`

rlve: sure

PR #1963, opened 2025-12-11
test(gossipsub): making tests more robust pt7

vladopajic:
adds to previous prs: https://github.com/vacp2p/nim-libp2p/pull/1930, https://github.com/vacp2p/nim-libp2p/pull/1933, https://github.com/vacp2p/nim-libp2p/pull/1936, https://github.com/vacp2p/nim-libp2p/pull/1938, https://github.com/vacp2p/nim-libp2p/pull/1947, #1950

in this pr: 
- `historyLength` is increased so that messages in cache are never clear while tests is running
- fix invalid check when messsages are never received

### DIFF
--- a/tests/libp2p/pubsub/component/test_gossipsub_message_handling.nim
+++ b/tests/libp2p/pubsub/component/test_gossipsub_message_handling.nim
@@ -18,7 +18,9 @@ import ../utils
 
 const MsgIdSuccess = "msg id gen success"
 
-proc setupTest(): Future[
+proc setupTest(
+    topic: string
+): Future[
     tuple[
       gossip0: GossipSub, gossip1: GossipSub, receivedMessages: ref HashSet[seq[byte]]
     ]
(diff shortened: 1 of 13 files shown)


tests/libp2p/pubsub/component/test_gossipsub_message_handling.nim
```diff
@@ -139,7 +147,10 @@ suite "GossipSub Component - Message Handling":
       isHighPriority = false,
     )
 
-    checkUntilTimeout:
```
vladopajic, line 145: `checkUntilTimeout` must not be used when asserting that messages are not received - as first check will make it satisfactory.

vladopajic: there aint ideal way to check this, but what we should do is wait some time then assert that messages are not received. still messages could be received after this timeout but still we are doing the best we can here

vladopajic: actually test can end when first node clears history ?

PR #1957, opened 2025-12-10
chore: utilize AsyncSemaphore from chronos

vladopajic:
closes: #1951

PR #1967, opened 2025-12-12
test: add import semaphore

vladopajic:
this pr is addition to https://github.com/vacp2p/nim-libp2p/pull/1957 pr.

semaphores needs to be imported still in case if chronos does not have it.

fixes compile issues in daily tests.

https://github.com/vacp2p/nim-libp2p/actions/runs/20164997664

### DIFF
--- a/libp2p/utils/semaphore.nim
+++ b/libp2p/utils/semaphore.nim
@@ -7,6 +7,8 @@
 # This file may not be copied, modified, or distributed except according to
 # those terms.
 
+{.used.}
+
 {.warning[UnusedImport]: off.}
 import chronos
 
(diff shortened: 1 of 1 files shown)


libp2p/connmanager.nim
```diff
@@ -344,7 +344,7 @@ proc getOutgoingSlot*(
     # still calling acquire to track this connection.
     discard c.outSema.acquire()
   elif not c.outSema.tryAcquire():
-    trace "Too many outgoing connections!", available = c.outSema.availableSlots
```
vladopajic, line 347: `if not c.outSema.tryAcquire()` then `c.outSema.availableSlots` is always `0`

gmelodie (approved):

PR #1969, opened 2025-12-12
test(gossipsub): making tests more robust pr10

vladopajic:
test `iDontWants history - last element is pruned during heartbeat` could still [occasionally fail](https://github.com/vacp2p/nim-libp2p/actions/runs/20173113200/job/57914053095), if messages are not published within same heartbeat. to combat this:
- heartbeat interval now scales with number of messages.
- message is reduced to something small (but still larger enough)
- publications will happen at the beginning of the heartbeat

documentation is also improved to ease reasoning .

### DIFF
--- a/tests/libp2p/pubsub/component/test_gossipsub_heartbeat.nim
+++ b/tests/libp2p/pubsub/component/test_gossipsub_heartbeat.nim
@@ -226,6 +226,10 @@ suite "GossipSub Component - Heartbeat":
       node0.fanout[topic].toSeq().allIt(it.peerId notin peersToDisconnect)
 
   asyncTest "iDontWants history - last element is pruned during heartbeat":
+    # When Node0 publish larger messages, it will also publish IDontWants, because 
+    # message is larger and sendIDontWantOnPublish is set to true.
+    # Then Node1 receives messages and IDontWants control messages.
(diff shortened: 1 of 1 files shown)


tests/libp2p/pubsub/component/test_gossipsub_heartbeat.nim
```diff
@@ -315,13 +330,13 @@ suite "GossipSub Component - Heartbeat":
     # When NodeInsideMesh sends a messages to the topic
     let peerInsideMesh = nodes[0].mesh[topic].toSeq()[0]
     let nodeInsideMesh = nodes.getNodeByPeerId(peerInsideMesh.peerId)
-    tryPublish await nodeInsideMesh.publish(topic, newSeq[byte](1000)), 1
+    tryPublish await nodeInsideMesh.publish(topic, toBytes("hellow")), 1
```
rlve, line 333: Why not some cypherpunk quote? 😄

vladopajic: just opened the notification and pr got merged. :(
i will set message to "exit the nation state" in the next occasion. or something similar vary short punchy text.

PR #1972, opened 2025-12-15
chore: reduce size of docker image

richard-ramos:


interop/transport/Dockerfile
```diff
@@ -1,18 +1,24 @@
 # syntax=docker/dockerfile:1.5-labs
-FROM nimlang/nim:latest as builder
+FROM nimlang/nim:2.2.4-ubuntu-regular as builder
```
vladopajic: better use 2.2.6, since we use that version all over the place

PR #1974, opened 2025-12-15
feat(cbind): dialing, opening and closing streams

richard-ramos:


gmelodie (approved):

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
```diff
@@ -0,0 +1,106 @@
...
+  if conn.isNil():
+    return err("unknown connection handle")
+
+  case self.operation
+  of CLOSE:
+    await conn.close()
+  of CLOSE_WITH_EOF:
+    await conn.closeWithEOF()
+  of DIAL, RELEASE:
+    raiseAssert "unsupported operation in processClose"
```
vladopajic: if code can return `err(...)` it might be more preferable then `raiseAssert`. 

this case is not something that should abort program execution, as `err(...)` could return. and on the other hand this is caller introduced issue so technically `raiseAssert` also makes sense for this case.... 

however if `err(...)` is possible then it might be tiny tiny bit more favorable:
- since it will guard from invalid execution but it wont crash it (theoretically program that has less changes for crash are better)
- also usage of `processClose` and it's result, in for example tests case, would be handled the same for all request types (not something that code currently have)
- if `StreamMsgType` has been constructed differently, for example if `processClose` has operation of `StreamCloseMsgType` with `CLOSE_WITH_EOF` and `CLOSE` then programmer will not be in chance of causing this assertions defect in this place. therefore `raiseAssert` seems unnecessary.

yeah, too much text for minor thing

vladopajic: image

this `raiseAssert` will never happen. which means whatever code dose here is avoided.

PR #1975, opened 2025-12-15
chore(kad): iterative get providers

gmelodie:
Make `GetProviders` iterative as `GetValue` until `k` providers are found or all peers are exhausted

### DIFF
--- a/libp2p/protocols/kademlia/provider.nim
+++ b/libp2p/protocols/kademlia/provider.nim
@@ -199,6 +199,15 @@ proc dispatchGetProviders(
 
   return (providers, reply.closerPeers)
 
+proc nextCandidates(
+    allCandidates: HashSet[PeerId], visitedCandidates: HashSet[PeerId], alpha: int
+): seq[PeerId] =
+  let unvisitedCandidates = allCandidates - visitedCandidates
+  if unvisitedCandidates.len() == 0:
+    return @[]
+
+  return unvisitedCandidates.toSeq()[0 .. min(alpha, unvisitedCandidates.len() - 1)]
+
 proc getProviders*(
     kad: KadDHT, key: Key
 ): Future[HashSet[Provider]] {.
@@ -207,24 +216,22 @@ proc getProviders*(
   ## Get providers for a given `key` from the nodes closest to that `key`.
 
   var
-    candidates: HashSet[PeerId]
+    allCandidates: HashSet[PeerId] = (await kad.findNode(key)).toHashSet()
+    visitedCandidates: HashSet[PeerId]
     allProviders: HashSet[Provider]
 
-  for peer in (await kad.findNode(key)):
-    candidates.incl(peer)
-
   # run until we find at least K providers
-  var curTry = 0
-  while allProviders.len() < kad.config.replication and curTry < kad.config.retries:
-    for chunk in candidates.toSeq().toChunks(kad.config.alpha):
-      let rpcBatch = chunk.mapIt(kad.switch.dispatchGetProviders(it, key))
-      for (providers, closerPeers) in await rpcBatch.collectCompleted(
-        kad.config.timeout
-      ):
-        allProviders.incl(providers)
-        for pid in closerPeers.toPeerIds():
-          candidates.incl(pid)
-    curTry.inc()
+  while allProviders.len() < kad.config.replication:
+    let candidates = nextCandidates(allCandidates, visitedCandidates, kad.config.alpha)
+    if candidates.len() == 0:
+      break
+
+    let rpcBatch = candidates.mapIt(kad.switch.dispatchGetProviders(it, key))
+    for (providers, closerPeers) in await rpcBatch.collectCompleted(kad.config.timeout):
+      allProviders.incl(providers)
+      allCandidates.incl(closerPeers.toPeerIds().toHashSet())
+
+    visitedCandidates.incl(candidates.toHashSet())
 
   return allProviders
 


libp2p/protocols/kademlia/provider.nim
```diff
@@ -206,12 +206,29 @@ proc getProviders*(
...
+      for (providers, closerPeers) in await rpcBatch.collectCompleted(
+        kad.config.timeout
+      ):
+        allProviders.incl(providers)
+        for peer in closerPeers:
+          let pid = PeerId.init(peer.id)
+          if not pid.isOk:
+            error "Invalid PeerId getProviders successful reply", peerId = peer.id
+            continue
+          nextCandidates.add(pid.get())
```
vladopajic: could there be overlap of `peerIds`? say in current candidates there is `peerA` and `peerB` both returning `peerC` then next loop will request from `peerC` two times?

libp2p/protocols/kademlia/provider.nim
```diff
@@ -206,12 +206,29 @@ proc getProviders*(
...
+      let rpcBatch = chunk.mapIt(kad.switch.dispatchGetProviders(it, key))
+      for (providers, closerPeers) in await rpcBatch.collectCompleted(
+        kad.config.timeout
+      ):
+        allProviders.incl(providers)
+        for peer in closerPeers:
+          let pid = PeerId.init(peer.id)
+          if not pid.isOk:
+            error "Invalid PeerId getProviders successful reply", peerId = peer.id
+            continue
```
vladopajic:
```
let ids = pluckPeerId(closerPeers)
...
```

create utility:
```proc pluckPeerId(peers: seq[Peer]): seq[PeerId]```

**Review by richard-ramos — APPROVED** [review]

File: libp2p/protocols/kademlia/provider.nim [hunk 3]
```diff
@@ -206,12 +206,29 @@ proc getProviders*(
...
+    nextCandidates: seq[PeerId] = await kad.findNode(key)
+    curCandidates: seq[PeerId]
+    allProviders: HashSet[Provider]
+
+  # run until we find at least K providers
+  var curTry = 0
+  while allProviders.len() < DefaultReplication and curTry < kad.config.retries:
+    curCandidates = nextCandidates
+    nextCandidates = @[]
+    for chunk in curCandidates.toChunks(kad.config.alpha):
```
vladopajic:
```suggestion
    for chunk in curCandidates.toChunks(kad.config.alpha):
      if allProviders.len() >= DefaultReplication:
        return allProviders
```
successful condition could be achieved before exhausting all chunks?

... if true then, ` allProviders.len() < DefaultReplication` condition in while is not needed

gmelodie: Yes, but `DefaultReplication` is a lower limit to ensure at least that amount is returned, but we want to return as much as possible within reasonable boundaries.

libp2p/protocols/kademlia/provider.nim
gmelodie: What the hell is pluck?

libp2p/protocols/kademlia/provider.nim
```diff
@@ -206,12 +206,29 @@ proc getProviders*(
...
-    for res in await rpcBatch.collectCompleted(kad.config.timeout):
-      allProviders.incl(res)
+  var
+    nextCandidates: seq[PeerId] = await kad.findNode(key)
+    curCandidates: seq[PeerId]
+    allProviders: HashSet[Provider]
+
+  # run until we find at least K providers
+  var curTry = 0
+  while allProviders.len() < DefaultReplication and curTry < kad.config.retries:
```
vladopajic: `DefaultReplication` here alludes to potential issue. if there is default value for replication then it suggest there there could be non default value - or value set by user. if later is true then code should use user configured value for replication - non default value?

if `DefaultReplication` could not be set by user, please consider renaming this constant to just `Replication`. also please taking a look at other part of code to see naming convention for consts - if i recall correctly, const start with lovercase

libp2p/protocols/kademlia/provider.nim
gmelodie: Added `toPeerIds`

libp2p/protocols/kademlia/provider.nim
gmelodie: My bad, should've used `kad.config.replication` instead

libp2p/protocols/kademlia/types.nim
```diff
@@ -82,6 +82,16 @@ proc toPeer*(peerInfo: PeerInfo): Peer =
     connection: ConnectionType.connected,
   )
 
+proc toPeerIds*(peers: seq[Peer]): seq[PeerId] =
```
vladopajic, line 85: would be better here if it is defined and imported from here ->  `import ../../[peerid]`

libp2p/protocols/kademlia/provider.nim
vladopajic: `pluckXXX`  naming is used in code where function returns single property from supplied input type.

search result from `pluckID`, but i assume there are also others like `pluckName`, etc..
https://github.com/search?q=pluckId&type=code

libp2p/protocols/kademlia/provider.nim
vladopajic: current code seems like an arbitrary decision. 

it keeps getting providers and it stops if condition is meet right at `DefaultReplication` in case if all chunks are excused in current iteration. but in all other iterations it will continue. why have such arbitrary behavior?

^ to explain,  consider we are here:
```
 while allProviders.len() < DefaultReplication
```
where `allProviders.len()` == 3 and `DefaultReplication` == 3, then code returns with 3 providers.

why return at 3? if we want to return as much as possible, then why not do one more chunk ?

libp2p/protocols/kademlia/provider.nim
```diff
@@ -206,12 +206,21 @@ proc getProviders*(
...
+  # run until we find at least K providers
+  var curTry = 0
+  while allProviders.len() < kad.config.replication and curTry < kad.config.retries:
+    for chunk in candidates.toChunks(kad.config.alpha):
+      let rpcBatch = chunk.mapIt(kad.switch.dispatchGetProviders(it, key))
+      for (providers, closerPeers) in await rpcBatch.collectCompleted(
+        kad.config.timeout
+      ):
+        allProviders.incl(providers)
+        candidates = closerPeers.toPeerIds()
```
vladopajic: should code append to candidates here?

libp2p/protocols/kademlia/provider.nim
vladopajic: this is why this const is anti pattern and should probably be removed. luckily naming was right

gmelodie: The const is used as a recommended default value, it comes from [the spec](https://github.com/libp2p/specs/blob/master/kad-dht/README.md#replication-parameter-k)

libp2p/protocols/kademlia/types.nim
gmelodie: `type Peer` is a kademlia protobufer thing

libp2p/protocols/kademlia/provider.nim
gmelodie: No, otherwise we'll send the requests to the same candidate twice, and we only want to try each candidate once

libp2p/protocols/kademlia/provider.nim
gmelodie: >  in case if all chunks are excused in current iteration

They're not. We check if we reach `DefaultReplication` providers, if not we ask all candidates (all chunks) and compile results again, if enough providers are achieved before going through all chunks, we keep going through them til the end

libp2p/protocols/kademlia/provider.nim
gmelodie: Indeed! Changed to use a HashSet

libp2p/protocols/kademlia/types.nim
vladopajic: got it

libp2p/protocols/kademlia/get.nim
```diff
@@ -76,16 +76,18 @@ proc bestValidRecord(
   ok(validRecords[selectedIdx])
 
 proc getValue*(
-    kad: KadDHT, key: Key
+    kad: KadDHT, key: Key, quorumOverride: Opt[int] = Opt.none(int)
```
vladopajic, line 79: `quorumOverride` not sure about naming.

PR #1976, opened 2025-12-16
feat(cbind): stream read / write

richard-ramos:


cbind/examples/cbindings.c
```diff
@@ -233,6 +253,23 @@ static void waitForCallback(void) {
...
+    exit(1);
+  }
+
+  printf("Read %zu bytes from ping stream\n", dataLen);
+  print_bytes("Read bytes", data, dataLen);
+
+  pthread_mutex_lock(&mutex);
+  callback_executed = 1;
+  pthread_cond_signal(&cond);
+  pthread_mutex_unlock(&mutex);
```
gmelodie, line 379: maybe this should be in a function

cbind/libp2p.nim
```diff
@@ -326,6 +326,136 @@ proc libp2p_dial(
...
+      StreamMsgType.WRITELP, conn = conn, data = data, dataLen = dataLen
+    ),
+    callback,
+    userData,
+  ).isOkOr:
+    let msg = "libp2p error: " & $error
+    callback(RET_ERR.cint, msg[0].addr, cast[csize_t](len(msg)), userData)
+    return RET_ERR.cint
+
+  return RET_OK.cint
```
gmelodie, line 525: This 
```
- if con is nil
- if data is nil
- sendRequestToLibp2pThread
```
pattern is super repetitive, we should definitely generalize it:

cbind/libp2p.nim
```diff
@@ -301,7 +303,239 @@ proc libp2p_connected_peers(
...
+    callback: Libp2pReadCallback,
+    userData: pointer,
+): cint {.dynlib, exportc, cdecl.} =
+  initializeLibrary()
+  checkLibParams(ctx, callback, userData)
+
+  if conn.isNil():
+    let msg = "connection is not set"
+    callback(RET_ERR.cint, nil, 0, msg[0].addr, cast[csize_t](len(msg)), userData)
+    return RET_ERR.cint
```
vladopajic: this could be one liner with template -> `checkConnParam(conn)`

cbind/libp2p.nim
```diff
@@ -301,7 +303,239 @@ proc libp2p_connected_peers(
...
+    RequestType.STREAM,
+    StreamRequest.createShared(
+      StreamMsgType.READEXACTLY, conn = conn, readLen = dataLen
+    ),
+    callback,
+    userData,
+  ).isOkOr:
+    let msg = "libp2p error: " & $error
+    callback(RET_ERR.cint, nil, 0, msg[0].addr, cast[csize_t](len(msg)), userData)
+    return RET_ERR.cint
```
vladopajic, line 439: this could be one liner with template, not sure how would i call that

PR #1978, opened 2025-12-16
feat(gossipsub): limit the maximum number of elements in the priority queue

richard-ramos:


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -458,6 +461,10 @@ proc sendEncoded*(
       maxSize = p.maxMessageSize, msgSize = msg.len
     Future[void].completed()
   elif isHighPriority or emptyQueues:
+    if p.rpcmessagequeue.sendPriorityQueue.len >= p.maxNumElementsInPriorityQueue:
+      warn "dropping pubsub message; priority queue limit exceeded",
```
arnetheduck, line 465: this is likely not a good idea, for two reasons:

* priority messages are those that concern protocol metadata - ie subscriptions, mesh maintenance etc - if these messages are dropped, the remote view of the mesh ends up undefined with respect to the local gossipsub state.
* there's not much benefit to surfacing a warning - ie there's not much the user can do when this mostly internal traffic "fails" - for tracing and debugging, a metric would be more appropriate - for the application using libp2p, a notification or counter, so that it can adapt and decrease the load or take some other action.

We could plausibly allow non-priority messages - broadcasts in particular - to be dropped instead of being kept in that big queue, though in that case, the queue should be  much smaller and again tracked with a metric - possibly it should also be a per-topic choice for the application - ie a policy.

Other, non-dropping options include removing the peer from the mesh which will decrease the amount of traffic directed at it which in turn might allow it to catch up.

PR #1980, opened 2025-12-17
chore(kad): GetProviders retries only failed peers

gmelodie:


libp2p/protocols/kademlia/provider.nim
```diff
@@ -218,20 +218,42 @@ proc getProviders*(
...
-      allProviders.incl(providers)
-      allCandidates.incl(closerPeers.toPeerIds().toHashSet())
+    # stop condition is not satisfied, but we can't retry (no peers have failed)
+    if failedCandidates.len == 0:
+      break
 
-    visitedCandidates.incl(candidates.toHashSet())
+    visitedCandidates = visitedCandidates - failedCandidates
+      # we need to visit failed peers again
+    failedCandidates.clear()
```
vladopajic, line 257: allCandidates = failedCandidates (search starts with allCanidates)
failedCandidates.clear()

gmelodie: if we don't remove failedCandidates from visitedCandidates then `nextCandidates` will not return correct values (but more importantly: i like this way better hehe)

vladopajic: `allCandidates` needs to be set to new value - `failedCandidates` because we need to request only those?

vladopajic:
```
allCandidates = failedCandidates 
visitedCandidates = visitedCandidates - failedCandidates
failedCandidates.clear()
```

libp2p/protocols/kademlia/provider.nim
```diff
@@ -218,20 +218,43 @@ proc getProviders*(
...
+      let candidates =
+        nextCandidates(allCandidates, visitedCandidates, kad.config.alpha)
+      if candidates.len() == 0:
+        break
+
+      let rpcBatch = candidates.mapIt(kad.switch.dispatchGetProviders(it, key))
+
+      for (providers, closerPeers) in await rpcBatch.collectCompleted(
+        kad.config.timeout
+      ):
```
richard-ramos: Nitpick: This is a bit hard to read. 
```suggestion
      let completedRCPBatch =  await rpcBatch.collectCompleted(kad.config.timeout)
      for (providers, closerPeers) in completedRPCBatch:
```

richard-ramos (approved):

libp2p/protocols/kademlia/provider.nim
```diff
@@ -218,20 +218,43 @@ proc getProviders*(
...
+        allCandidates.incl(closerPeers.toPeerIds().toHashSet())
+
+      visitedCandidates.incl(candidates.toHashSet())
 
-    let rpcBatch = candidates.mapIt(kad.switch.dispatchGetProviders(it, key))
-    for (providers, closerPeers) in await rpcBatch.collectCompleted(kad.config.timeout):
-      allProviders.incl(providers)
-      allCandidates.incl(closerPeers.toPeerIds().toHashSet())
+      for (cand, fut) in zip(candidates, rpcBatch):
+        if not fut.completed():
```
vladopajic, line 242:
```suggestion
        if fut.failed() or (not fut.completed()):
```

right? 

because future can:
- fail - request can fail for many reasons (connectivity errors, etc)...
- `not completed` if it get's cancelled ?

vladopajic: does `collectCompleted` needs to be changed accordingly ?

vladopajic (approved):

PR #1982, opened 2025-12-19
chore(kad): refactor find and lookupstate

gmelodie:


libp2p/protocols/kademlia.nim
```diff
@@ -25,7 +25,10 @@ proc bootstrapNode*(
 
   let msg =
     try:
-      await kad.sendFindNode(peerId, addrs, kad.rtable.selfId).wait(kad.config.timeout)
+      debug "trying to find self"
+      await kad.switch.dispatchFindNode(peerId, kad.rtable.selfId, Opt.some(addrs)).wait(
```
gmelodie, line 29: `addrs` is almost always not needed since `dial` finds it in the peerstore, but for bootstrapping we don't previously have addrs so we need to pass it

libp2p/protocols/kademlia/find.nim
```diff
@@ -141,12 +166,11 @@ proc findPeer*(
...
   let selfKey = kad.switch.peerInfo.peerId.toKey()
-  for p in kad.rtable.findClosest(target, kad.config.replication):
-    if p == selfKey: # do not return self as one of closest peers
-      continue
-    let peer = p.toPeer(kad.switch).valueOr:
-      continue
-    closestPeers.add(peer)
+  for p in kad.rtable.findClosest(target, kad.config.replication).filterIt(
+    it != selfKey
+  ):
```
vladopajic: would prefer not to have multi line `for`

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,115 +7,140 @@
...
+    .pairs()
+    .toSeq()
+    # sort by distance (closer first)
+    .sortedByIt(it[1])
+    # get pid
+    .mapIt(it[0])
+    # filter out already queried pids
+    .filterIt(not state.queried.contains(it))
+    # filter out our own pid
+    .filterIt(it != state.kad.switch.peerInfo.peerId)
```
vladopajic: code would be more optimal, if this is done first (removing pids), then doing everything else. as this can produce shorter result first.

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,115 +7,140 @@
...
+
+  # get alpha peers
+  for pid in sortedPids:
+    peers.add(pid)
+    if peers.len() >= state.kad.config.alpha:
+      break
+  peers
```
vladopajic: could be one line `return sortedPids[0 ... min(alpha, sortedPids.len - 1)]` ?

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,115 +7,140 @@
...
+
+  return res
+
+proc selectClosestK*(state: LookupState): seq[PeerId] =
+  var res: seq[PeerId] = @[]
+  for pid in state.shortlist.keys():
+    res.add(pid)
+    if res.len >= state.kad.config.replication:
+      break
+  return res
```
vladopajic: there is mismatch in naming and logic. proc does not guarantee that closest K peers are returned? 
getting closest peers is correctly implemented in `selectAlphaPeers` proc.

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,115 +7,140 @@
...
+
+proc done*(state: LookupState): bool {.raises: [], gcsafe.} =
+  let noNew = state.selectAlphaPeers().len() == 0
+
+  return noNew
+
+proc dispatchFindNode*(
+    switch: Switch,
+    peer: PeerId,
+    key: Key,
```
vladopajic: nit: `key` -> `targetKey`;  as searching for target

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,115 +7,140 @@
...
-    for peer in toQuery.filterIt(kad.switch.peerInfo.peerId != it):
-      state.markPending(peer)
-      let addrs = addrTable.getOrDefault(peer, @[])
-      if addrs.len == 0:
-        state.markFailed(peer)
-        continue
-      pendingFutures[peer] =
-        kad.sendFindNode(peer, addrs, targetId).wait(chronos.seconds(5))
+  while not state.done():
+    let toQuery = state.selectAlphaPeers()
```
vladopajic:
```suggestion 
  while true:
    let toQuery = state.selectAlphaPeers()
    if toQuery.len == 0: break
```

`state.done()` will call `selectAlphaPeers` that is needed right below, better reuse that.

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,115 +7,140 @@
...
+      if fut.completed():
+        state.queried.incl(peerId)
 
-    for msg in successfulReplies:
+    for msg in completedRPCBatch:
       for peer in msg.closerPeers:
         let pid = PeerId.init(peer.id)
         if not pid.isOk:
           error "Invalid PeerId in successful reply", peerId = peer.id
           continue
```
vladopajic:
```suggestion
    addNewPeerAddresses(kad.switch.peerStore[AddressBook], msg.closerPeers)
```

second `for` should not exist here. it adds to complexity that could be replaced by function

libp2p/protocols/kademlia/find.nim
vladopajic: looks like both functions should do the same but only difference is number of elements returned (alpha count vs replication count) ?

libp2p/protocols/kademlia/find.nim
vladopajic: - filter out 
- sort
- select K
- map

libp2p/protocols/kademlia/find.nim
vladopajic: looking at whole file some parts use `targetKey` some `key`. whenever `key` has `target` semantics better use that in name.

vladopajic: example `LookupState.init`

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,115 +7,140 @@
...
+    .filterIt(it != state.kad.switch.peerInfo.peerId)
+
+  # get alpha peers
+  for pid in sortedPids:
+    peers.add(pid)
+    if peers.len() >= state.kad.config.alpha:
+      break
+  peers
+
+proc init*(T: type LookupState, kad: KadDHT, targetId: Key): T =
```
vladopajic: `targetId` is of type `Key`....

shouldn't it then be `targetKey` or  `targetId: Id`?

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,117 +7,141 @@
...
+    .pairs()
+    .toSeq()
+    # filter out already queried pids
+    .filterIt(not state.queried.contains(it[0]))
+    # filter out our own pid
+    .filterIt(it[0] != state.kad.switch.peerInfo.peerId)
+    # sort by distance (closer first)
+    .sortedByIt(it[1])
+
+proc selectPeers*(state: LookupState, amount: int): seq[PeerId] =
```
vladopajic: selectPeers -> `selectClosesPeers`

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,117 +7,141 @@
...
+      newPeerInfos.add(peerInfo)
+
+  return newPeerInfos
+
+proc sortedShortlist(state: LookupState): seq[(PeerId, XorDistance)] =
+  ## Sort shortlist by closer distance first
+
+  return state.shortlist
+    .pairs()
+    .toSeq()
```
vladopajic: would be better if this is one operation (pairs + toSeq) , because memory is trashed unnecessary. and this is done quite often.

libp2p/protocols/kademlia/find.nim
vladopajic: bit more descriptive

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,117 +7,141 @@
...
+  var res = LookupState(kad: kad, targetKey: targetKey, queried: initHashSet[PeerId]())
+  for pid in kad.rtable.findClosestPeerIds(targetKey, kad.config.replication):
+    res.shortlist[pid] = xorDistance(pid, targetKey, kad.rtable.config.hasher)
+
+  return res
+
+proc done*(state: LookupState): bool {.raises: [], gcsafe.} =
+  let noNew = state.selectPeers(state.kad.config.alpha).len() == 0
+
+  return noNew
```
vladopajic: done is not needed?

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,117 +7,141 @@
...
-    state.done = checkConvergence(state, kad.switch.peerInfo.peerId)
-
-  return state.selectClosestK()
+  ## Iteratively search for the k closest peers to a key.
+
+  var state = LookupState.init(kad, key)
+
+  while true:
+    let toQuery = state.selectPeers(kad.config.alpha)
+
```
vladopajic, line 132:
```suggestion
```

libp2p/protocols/kademlia/find.nim
```diff
@@ -139,28 +163,30 @@ proc findPeer*(
   return ok(PeerInfo(peerId: peer, addrs: kad.switch.peerStore[AddressBook][peer]))
 
 proc findClosestPeers*(kad: KadDHT, target: Key): seq[Peer] =
+  let closerPeers = kad.rtable.findClosest(target, kad.config.replication).filterIt(
+      it != kad.switch.peerInfo.peerId.toKey()
+    )
+
   var closestPeers: seq[Peer]
```
vladopajic: naming of these two variables....

libp2p/protocols/kademlia/find.nim
```diff
@@ -139,28 +163,30 @@ proc findPeer*(
...
-      continue
-    let peer = p.toPeer(kad.switch).valueOr:
-      continue
-    closestPeers.add(peer)
+
+  for p in closerPeers:
+    p.toPeer(kad.switch).withValue(peer):
+      closestPeers.add(peer)
+
   return closestPeers
```
vladopajic: add utility `toPeer` (or however you would call this transformation in `for`):

```toPeer(peer seq[????]) seq[Peer]```

then 

`return toPeer(closestPeers)`

Comment (vladopajic) [reply libp2p/protocols/kademlia/find.nim]: avoid two variables of same name and explains what transformation is applied to them.

File: libp2p/protocols/kademlia/find.nim [hunk 16]
```diff
@@ -7,117 +7,141 @@
...
+
+proc addNewPeerAddresses(
+    addrsTable: SeqPeerBook[MultiAddress], closerPeers: seq[Peer]
+) =
+  for peer in closerPeers:
+    let pid = PeerId.init(peer.id)
+    if not pid.isOk:
+      error "Invalid PeerId in successful reply", peerId = peer.id
+      return
+    addrsTable[pid.get()] = peer.addrs
```
vladopajic: ``` addrsTable[pid.get()] = peer.addrs```
this overrides, but should it append ?

libp2p/protocols/kademlia/find.nim
vladopajic: would be even better if `shortlist` is heap. could `shortlist` already be without `queried and self ids`? 

with both getting K elements is constant.

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,117 +7,141 @@
...
+  return newPeerInfos
+
+proc sortedShortlist(state: LookupState): seq[(PeerId, XorDistance)] =
+  ## Sort shortlist by closer distance first
+
+  return state.shortlist
+    .pairs()
+    .toSeq()
+    # filter out already queried pids
+    .filterIt(not state.queried.contains(it[0]))
```
vladopajic: ah... queried should be included when algorithm is returning... it should return closest of all. 

but when algorithm is selecting next peers for while iteration, then it should not include queried.

libp2p/protocols/kademlia/find.nim
gmelodie: To quote the great Richard Ramos: newlines help code breathe (yes, I didn't know code had lungs either)

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,160 +7,186 @@
...
-      failedPeers.add(peerId)
-      error "Could not send find_node to peer", peerId, err = getCurrentExceptionMsg()
+      let peerInfo = PeerInfo(peerId: pid, addrs: newPeer.addrs)
+      newPeerInfos.add(peerInfo)
+
+  return newPeerInfos
+
+proc sortedShortlist(
+    state: LookupState, excludeQueried: bool = true
+): seq[(PeerId, XorDistance)] =
```
vladopajic, line 41: maybe return just `seq[PeerId]`?

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,160 +7,186 @@
...
+      continue
+    if excludeQueried and state.queried.contains(pid):
+      continue
+    sortedShortlist.add((pid, dist))
 
-  return (receivedReplies, failedPeers)
+  sortedShortlist.sort(
+    proc(a, b: (PeerId, XorDistance)): int =
+      cmp(a[1], b[1])
+  )
```
vladopajic, line 57: note, if `selectCloserPeers` is going to use small number for `amount` (default alpha is 10, replication 20), then searching through `shortlist` each time and picking closest might be more optimal (20 * N), then sorting (N logN).

vladopajic: ... but since shortlist is not going  to be that big, sort is fine

gmelodie: Yes this is primarily why I chose not to go with a heap as you suggested, because the gain was not a lot and the code would become more complicated in other parts.

libp2p/utility.nim
```diff
@@ -182,6 +182,10 @@ proc collectCompleted*[T, E](
   # Collect only successful results
   return futs.filterIt(it.completed()).mapIt(it.value())
 
+proc take*[T](s: seq[T], n: int): seq[T] =
+  ## Take first `n` elements of `s`, or `s.len()` if `n > s.len()`
+  return s[0 .. min(s.len() - 1, n)]
```
richard-ramos, line 187: This is inclusive, so if you have a sequence of 3 elements, and pass `n = 1`,  you would get 2 elements. Is this okay?
```nim
let x = @[1,2, 3]
let n = 1
echo x[0 .. n]
# @[1, 2]
```

libp2p/protocols/kademlia/find.nim
```diff
@@ -7,160 +7,186 @@
...
+    discard kad.rtable.insert(p.peerId)
+    # Nodes might return different addresses for a peer, so we append instead of replacing
+    try:
+      var existingAddresses = kad.switch.peerStore[AddressBook][p.peerId].toHashSet()
+      for a in p.addrs:
+        existingAddresses.incl(a)
+      kad.switch.peerStore[AddressBook][p.peerId] = existingAddresses.toSeq()
+    except KeyError as exc:
+      debug "Could not update shortlist", err = exc.msg
+    # TODO: add TTL to peerstore, otherwise we can spam it with junk
```
richard-ramos, line 111: Probably makes sense to add an issue for this otherwise we'll forget :(

PR #1990, opened 2025-12-26
test: add `kadInteropTest` template

### LINKED ISSUE #1987: add test for `kadInteropTest` proc
add test for `kadInteropTest` proc. this test case should be unit tests between two nim nodes.

hints:
- create file on path `/tests/interop/test_kad.nim`
- add test suite with following tests case:
  - test success path 
     - this test case should be vary trivial, create nim node with kademlia using switch. 
     - call `kadInteropTest` proc giving peer information from node created in step above
     - assert that `kadInteropTest` will return true
  - consider adding some some tests where `kadInteropTest` will return false.

Eshaan-byte:
Adds tests for `kadInteropTest` proc as requested in #1987

## Changes
- Created `tests/interop/test_kad.nim` with comprehensive test suite
- Added success path test with two nim nodes
- Added failure path tests for invalid peer ID and unreachable address
- Updated `tests/test_all.nim` to include interop tests

## Test Cases
1. Success path - two nim nodes communicate successfully
2. Failure path - invalid peer ID
3. Failure path - unreachable address

Closes #1987

### DIFF
--- a/tests/interop/test_kad.nim
+++ b/tests/interop/test_kad.nim
@@ -12,59 +12,18 @@
 import chronos, stew/byteutils
 import ../../libp2p/[switch, builders, peerid, protocols/kademlia, wire]
 import ../tools/[unittest]
-
-proc kadInteropTest(otherPeerId: PeerId, otherAddr: string, ourAddr: string): Future[bool] {.async.} =
-  ## Tests Kademlia DHT interoperability by:
-  ## 1. Creating a local switch with Kademlia
-  ## 2. Connecting to another peer
-  ## 3. Putting a value into the DHT
-  ## 4. Retrieving the value to verify it was stored correctly
-  var switch = SwitchBuilder
-    .new()
-    .withRng(newRng())
-    .withAddresses(@[MultiAddress.init(ourAddr).tryGet()])
-    .withTcpTransport()
-    .withMplex()
-    .withNoise()
-    .build()
-
-  let kad = KadDHT.new(
-    switch,
-    bootstrapNodes = @[(otherPeerId, @[MultiAddress.init(otherAddr).get()])],
-    config = KadDHTConfig.new(quorum = 1),
-  )
-
-  switch.mount(kad)
-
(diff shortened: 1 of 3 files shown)


tests/interop/test_kad.nim
```diff
@@ -0,0 +1,189 @@
+# Nim-LibP2P
```
vladopajic, line 1: - proc `kadInteropTest` in  `interop/kad/nim-peer/src/nim_peer.nim` is not tested at all with this code. technically same code should be reused. hints:
  - create `kadInteropTest` template similarly to `template streamTransportTest` (from the standpoint of code reuse), for example add it to file `tests/interop/kadTests.nim`
  - use `kadInteropTest` template in `tests/interop/test_kad.nim` - here other node is created using switch
  - use `kadInteropTest` template in `interop/kad/nim-peer/src/nim_peer.nim`. here `kadInteropTest` should use info from rust node.
  
  what is achieved this this? there is single `kadInteropTest` template used in `/interop` - real interop with other node implementation, and `/tests/interop` where same testing logic is used with node created using switch.

vladopajic: @Eshaan-byte, thanks for effort. please see comment above. this is main priority of this effort. reuse code in two different places. after main goal is achieved pr will be reviewed again - were we will seek higher coding standard.

vladopajic: aslo note, this `kadInteropTest` must be built in a way where other node is reused of each test. becasue in interop with rust node, we cant start and stop node that easily, or it would be to much effort. therfore logic in `kadInteropTest` should assume that node is started and same node is running until the end of test.

vladopajic: `kadInteropTest` does not have to be template, it can actually be `proc` or it chould actually be 100% exisitng code from `proc kadInteropTest*`. we just need reuse as a main goal.

Eshaan-byte:
PTAL @vladopajic

PR #1998, opened 2026-01-06
test(kad): Routing table and refresh bucket

rlve:
Changes:
- add tests in `test_routingtable`:
  - `"does not insert self"`
  - `"re-insert existing key updates lastSeen"`
  - `"findClosest returns all keys if less than n available"`
- add test in `test_find`:
  - `"Find node via refresh stale buckets"` => this test uncovered a bug where `handleFindNode` validation was failing because of the key format sent in `findNode` via `refreshBuckets` => fixed in https://github.com/vacp2p/nim-libp2p/pull/1998/changes/70316b63c92d61d606b0d0502433ec96d4a3c80e by @gmelodie

### DIFF
--- a/cbind/examples/cbindings.c
+++ b/cbind/examples/cbindings.c
@@ -5,13 +5,13 @@
 #include <string.h>
 #include <unistd.h>
 
+#define CID_BUF_SIZE 128
+
 // Shared synchronization variables
 pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
 pthread_cond_t cond = PTHREAD_COND_INITIALIZER;
 int callback_executed = 0;
 libp2p_stream_t *ping_stream = NULL;
-
-#define CID_BUF_SIZE 128
 
 typedef struct {
   char peerId[256];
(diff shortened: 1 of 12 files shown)


tests/libp2p/kademlia/test_find.nim
```diff
@@ -165,4 +165,4 @@ suite "KadDHT Find":
 
     # BUG: kad1 should discover kad3 via kad2, but doesn't
     # (refreshBuckets generates invalid keys that handleFindNode rejects)
-    check not kad1.hasKey(kad3.rtable.selfId)
+    check kad1.hasKey(kad3.rtable.selfId)
```
vladopajic, line 168: good catch.  :+1: 

this should now have been merged without associated issue (there is no issue in the comment).

tests/libp2p/kademlia/test_find.nim
```diff
@@ -165,4 +165,4 @@ suite "KadDHT Find":
 
     # BUG: kad1 should discover kad3 via kad2, but doesn't
```
vladopajic: comment is not needed anymore?

tests/libp2p/kademlia/test_find.nim
vladopajic: oh, somehow github ui didn't show this as new code (green highlight), but as old code (no highlight)... nvm me

PR #2000, opened 2026-01-06
feat(gossipsub): extensions control message

vladopajic:
`ift-ts:p2p:ift:2026q1-nimlibp2p-gossipsub-extensions:extensions`

this pr adds support for extensions control message:
https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.3.md

and "TestExtension" the experimental extensions
https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/extensions/experimental/test-extension.md

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -75,6 +75,9 @@ when defined(libp2p_expensive_metrics):
     labels = ["id", "topic"],
   )
 
+proc gossipExtensionsSupported(proto: string): bool =
+  return proto == GossipSubCodec_13 or proto == GossipSubCodec_14
+
 proc init*(
     _: type[GossipSubParams],
     pruneBackoff = 1.minutes,
@@ -112,6 +115,7 @@ proc init*(
     disconnectPeerAboveRateLimit = false,
     maxNumElementsInNonPriorityQueue = DefaultMaxNumElementsInNonPriorityQueue,
     sendIDontWantOnPublish = false,
+    extensionsDisabled = false,
     testExtensionConfig = none(TestExtensionConfig),
 ): GossipSubParams =
   GossipSubParams(
@@ -151,6 +155,7 @@ proc init*(
     disconnectPeerAboveRateLimit: disconnectPeerAboveRateLimit,
     maxNumElementsInNonPriorityQueue: maxNumElementsInNonPriorityQueue,
     sendIDontWantOnPublish: sendIDontWantOnPublish,
+    extensionsDisabled: extensionsDisabled,
     testExtensionConfig: testExtensionConfig,
   )
 
@@ -275,17 +280,18 @@ method onNewPeer*(g: GossipSub, peer: PubSubPeer) =
   # established then use protocol negotiated from connection. 
   # after peer codec is known, gossip sends extensions control message as first message
   # to this peer.
-  if peer.codec != "":
-    if peer.codec == GossipSubCodec_13:
-      sendExtensionsControl()
-  else:
-    proc addPeerAfterConnected(): Future[void] {.async.} =
-      await peer.connectedFut
-
-      if peer.sendConn.protocol == GossipSubCodec_13:
+  if not g.parameters.extensionsDisabled:
+    if peer.codec != "":
+      if gossipExtensionsSupported(peer.codec):
         sendExtensionsControl()
+    else:
+      proc addPeerAfterConnected(): Future[void] {.async.} =
+        await peer.connectedFut
 
-    asyncSpawn addPeerAfterConnected()
+        if gossipExtensionsSupported(peer.sendConn.protocol):
+          sendExtensionsControl()
+
+      asyncSpawn addPeerAfterConnected()
 
 method onPubSubPeerEvent*(
     p: GossipSub, peer: PubSubPeer, event: PubSubPeerEvent
@@ -1025,6 +1031,9 @@ proc maintainDirectPeers(g: GossipSub) {.async: (raises: [CancelledError]).} =
       await g.addDirectPeer(id, addrs)
 
 proc createExtensionsState(g: GossipSub): ExtensionsState =
+  if g.parameters.extensionsDisabled:
+    return ExtensionsState.new() # noop state
+
   proc onMissbehaveExtensions(id: PeerId) {.gcsafe, raises: [].} =
     g.peers.withValue(id, peer):
       peer[].behaviourPenalty += 0.1
--- a/libp2p/protocols/pubsub/gossipsub/extensions.nim
+++ b/libp2p/protocols/pubsub/gossipsub/extensions.nim
@@ -49,7 +49,7 @@ proc toPeerExtensions(ctrlExtensions: ControlExtensions): PeerExtensions =
     false
   PeerExtensions(testExtensionSupported: testExtensionSupported)
 
-proc isExtensionNegotiatedTestExtension(state: ExtensionsState, peerId: PeerId): bool =
+proc isExtensionNegotiated_TestExtension(state: ExtensionsState, peerId: PeerId): bool =
   # does both this node peer support "test extension"?
   state.testExtensionConfig.isSome() and
     state.peerExtensions.getOrDefault(peerId).testExtensionSupported
@@ -57,14 +57,14 @@ proc isExtensionNegotiatedTestExtension(state: ExtensionsState, peerId: PeerId):
 proc onHandleRPC(state: ExtensionsState, peerId: PeerId) =
   # extensions event called when node receives every RPC message
 
-  if state.isExtensionNegotiatedTestExtension(peerId):
+  if state.isExtensionNegotiated_TestExtension(peerId):
     state.testExtensionConfig.get().onHandleRPC(peerId)
 
 proc onNegotiated(state: ExtensionsState, peerId: PeerId) =
   # extension event called when both sides have negotiated (exchanged) extensions.
   # it will be called only once per connection session as soon as extensiosn are exchanged.
 
-  if state.isExtensionNegotiatedTestExtension(peerId):
+  if state.isExtensionNegotiated_TestExtension(peerId):
     state.testExtensionConfig.get().onNegotiated(peerId)
 
 proc addPeer*(state: ExtensionsState, peerId: PeerId) =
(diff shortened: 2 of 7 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -254,6 +259,34 @@ method onNewPeer*(g: GossipSub, peer: PubSubPeer) =
...
+    if peer.codec == GossipSubCodec_13:
+      sendExtensionsControl()
+  else:
+    proc addPeerAfterConnected(): Future[void] {.async.} =
+      await peer.connectedFut
+
+      if peer.sendConn.protocol == GossipSubCodec_13:
+        sendExtensionsControl()
+
+    asyncSpawn addPeerAfterConnected()
```
vladopajic: in future pr, i'll try to address this smell here with refactoring so that peer is created after connection is established in which case if/else like this will not be needed.

gmelodie (approved):

libp2p/protocols/pubsub/rpc/protobuf.nim
```diff
@@ -397,12 +440,19 @@ proc encodeRpcMsg*(msg: RPCMsg, anonymize: bool): seq[byte] =
...
   if msg.ping.len > 0:
     pb.write(60, msg.ping)
   if msg.pong.len > 0:
     pb.write(61, msg.pong)
+
+  # Canonical Extensions should register their messages here.
+  # They must use field numbers larger than 0x200000 to be encoded with at least 4 bytes.
+  if msg.testExtension.isSome():
+    # if set write empty bytes, this will set filed tag
+    pb.write(6492434, newSeq[byte](0))
```
richard-ramos: Can we have this in a constant or a proc (i.e. proc fieldID(T: typedesc[TestExtension]): int = 6492434` with the idea of doing
```
pb.write(TestExtension.fieldID, newSeq[byte](0))
```

libp2p/protocols/pubsub/gossipsub/extensions.nim
```diff
@@ -0,0 +1,108 @@
...
+import std/[options, sets, tables]
+import ../../../[peerid]
+import ../rpc/messages
+
+proc noopPeerCallback(peer: PeerId) {.gcsafe, raises: [].} =
+  discard
+
+type
+  PeerExtensions = object
+    testExtension: bool
```
richard-ramos: This is okay for now, but depending on how many extensions are created in the future we could end up with having to add more and more attributes, so we might have to refactor how we track the extensions.

I guess that's a bridge to cross in the future because for now the only extensions that come to mind are: TestExtension, PartialMessages, Gossip1_4, and PingPong

libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -216,32 +243,54 @@ proc byteSize(controlIMreceiving: ControlIMReceiving): int =
 proc byteSize*(imreceivings: seq[ControlIMReceiving]): int =
   imreceivings.foldl(a + b.byteSize, 0)
 
+static:
+  expectedFields(ControlExtensions, @["testExtension"])
+proc byteSize(controlExtensions: Option[ControlExtensions]): int =
+  if controlExtensions.isNone:
+    0
+  else:
+    4 # 4 byte for the bool aka uint32 - testExtension
```
richard-ramos: instead of returning a 4, can we do something like: this:
```
if controlExtensions.isNone:
  0
else:
  TestExtension.byteSize()
```

and have a 
```
proc byteSize(T: typedesc[TestExtension]): int =
    4
```
That way the `byteSize` proc of `ControlExtensions does not need to know exactly the size of TestExtension but obtain it via a proc call.

libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -216,32 +243,54 @@ proc byteSize(controlIMreceiving: ControlIMReceiving): int =
...
-      control.idontwant.foldl(a + b.byteSize, 0) +
+      control.idontwant.foldl(a + b.byteSize, 0) + byteSize(control.extensions) +
       control.preamble.foldl(a + b.byteSize, 0) +
       control.imreceiving.foldl(a + b.byteSize, 0)
 
 else:
   static:
-    expectedFields(ControlMessage, @["ihave", "iwant", "graft", "prune", "idontwant"])
+    expectedFields(
+      ControlMessage, @["ihave", "iwant", "graft", "prune", "idontwant", "extensions"]
```
richard-ramos, line 281: I am not sure about "extensions" being added here (maybe i'm misunderstanding the purpose of `expectedFields`). Because ControlExtensions is marked as optional in the protobuffer. So what would happen if the field is not present at al?  

Same question applies to line 287, since TestExtension is optional

libp2p/protocols/pubsub/rpc/protobuf.nim
```diff
@@ -416,4 +466,14 @@ proc decodeRpcMsg*(msg: seq[byte]): ProtoResult[RPCMsg] {.inline.} =
   assign(rpcMsg.control, ?pb.decodeControl())
   discard ?pb.getField(60, rpcMsg.ping)
   discard ?pb.getField(61, rpcMsg.pong)
+
+  # Canonical Extensions should register their messages here.
+  # They must use field numbers larger than 0x200000 to be encoded with at least 4 bytes.
+
+  var testExtension: seq[byte]
+  if ?pb.getField(6492434, testExtension):
```
richard-ramos: If using the idea of the fieldID proc,

```suggestion
  if ?pb.getField(TestExtension.fieldID(), testExtension):
```

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -254,6 +259,34 @@ method onNewPeer*(g: GossipSub, peer: PubSubPeer) =
...
+  proc sendExtensionsControl() =
+    g.extensionsState.addPeer(peer.peerId)
+    g.send(
+      peer,
+      RPCMsg(
+        control: some(
+          ControlMessage(extensions: some(g.extensionsState.makeControlExtensions()))
+        )
+      ),
+      false,
```
richard-ramos: Shouldnt this be high priority?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -983,6 +1024,23 @@ proc maintainDirectPeers(g: GossipSub) {.async: (raises: [CancelledError]).} =
     for id, addrs in g.parameters.directPeers:
       await g.addDirectPeer(id, addrs)
 
+proc createExtensionsState(g: GossipSub): ExtensionsState =
+  proc onMissbehaveExtensions(id: PeerId) {.gcsafe, raises: [].} =
+    g.peers.withValue(id, peer):
+      peer[].behaviourPenalty += 0.1
+
+  let testExtensionConfig = g.parameters.testExtensionConfig.valueOr:
```
richard-ramos, line 1041: Shouldnt the handling of testExtension be done in a separate proc and call that proc in here?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -983,6 +1024,23 @@ proc maintainDirectPeers(g: GossipSub) {.async: (raises: [CancelledError]).} =
...
+    # by default, parameters will not have this config. because param is mainly used
+    # in tests for testing purposes.
+
+    proc onNagotiated(id: PeerId) {.gcsafe, raises: [].} =
```
richard-ramos: typo

libp2p/protocols/pubsub/gossipsub/extensions.nim
```diff
@@ -0,0 +1,108 @@
...
+  state.testExtensionConfig.isSome() and
+    state.peerExtensions.getOrDefault(peerId).testExtension
+
+proc onHandleRPC(state: ExtensionsState, peerId: PeerId) =
+  # extensions event called when node receives every RPC message
+
+  if state.isExtensionNagotiatedTestExtensions(peerId):
+    state.testExtensionConfig.get().onHandleRPC(peerId)
+
+proc onNagotiated(state: ExtensionsState, peerId: PeerId) =
```
richard-ramos: typo

libp2p/protocols/pubsub/rpc/protobuf.nim
vladopajic: i don't like this approach, because the same type could have a different field id in different protobuf messages. 

TestExtension is used only here so for it it wouldn't be an issue, but if we use the same principle for other types it will not work.

richard-ramos: As long as the magic number is not hardcoded in here it's fine to keep the same logic it has right now.

vladopajic: in general i am not sure if we should invest in doing this any differently than currently, as these files should be auto generated like in other languages.

until that is possible, bare the smells

vladopajic: > As long as the magic number is not hardcoded in here it's fine to keep the same logic it has right now.

sure it can be const in file

richard-ramos: yeah, it's just that having the same number repeated in line 454 and 474 smells so bad it's unbearable while having at least a const would reduce the smelliness just a bit :rofl:

libp2p/protocols/pubsub/gossipsub/extensions.nim
vladopajic: yes, we will see...

libp2p/protocols/pubsub/rpc/messages.nim
vladopajic: > So what would happen if the field is not present at al?

field in structure is always there and it has type. type is holder for value that may be present or not.

there is no difference between fields (in this context):
```
foo: bool
bar: Option[bool]
```

libp2p/protocols/pubsub/gossipsub.nim
vladopajic: yes, it makes sense

libp2p/protocols/pubsub/gossipsub.nim
vladopajic: since this is one liner i wouldn't really bother now. maybe when new extensions are added there will be better idea where to place code like this. but, atm it suits perfectly here (subjectively).

PR #2008, opened 2026-01-07
chore(kad): bootstrap send findNode to all non-empty buckets

### LINKED ISSUE #2006: Kad: Lookup random PeerId for each non-empty bucket during bootstrap
[Kad DHT spec](https://github.com/libp2p/specs/tree/master/kad-dht#bootstrap-process) mentions:

> On every run, we generate a random peer ID for every non-empty routing table's k-bucket and we look it up via the process defined in peer routing. 

However, at the moment we do the lookup only for one random PeerId during initial bootstrap.

https://github.com/vacp2p/nim-libp2p/blob/c5e0614982b02b9afff733ba0bae3fa3e63b132a/libp2p/protocols/kademlia.nim#L60-L68

Consider expanding lookup during bootstrap.

gmelodie:
Fix #2006

### DIFF
--- a/libp2p/utils/heartbeat.nim
+++ b/libp2p/utils/heartbeat.nim
@@ -14,7 +14,7 @@ import chronos, chronicles
 export chronicles
 
 template heartbeat*(
-    name: string, interval: Duration, body: untyped, sleepFirst: bool = false
+    name: string, interval: Duration, sleepFirst: bool, body: untyped
 ): untyped =
   var nextHeartbeat = Moment.now()
   if sleepFirst:
@@ -35,3 +35,7 @@ template heartbeat*(
         debug "Missed heartbeat", heartbeat = name, delay = delay, hinterval = itv
       nextHeartbeat = now + itv
     await sleepAsync(nextHeartbeat - now)
+
+template heartbeat*(name: string, interval: Duration, body: untyped): untyped =
+  heartbeat(name, interval, sleepFirst = false):
+    body


libp2p/protocols/kademlia.nim
```diff
@@ -152,6 +144,8 @@ method start*(kad: KadDHT) {.async: (raises: [CancelledError]).} =
   kad.republishLoop = kad.manageRepublishProvidedKeys()
   kad.expiredLoop = kad.manageExpiredProviders()
 
+  for (peerId, addrs) in kad.bootstrapNodes:
+    await kad.addBootstrapNode(peerId, addrs)
```
rlve: Why did you move it out of bootstrap? 🤔

gmelodie: Because technically the bootstrapping procedure does not include this part as per the spec, so reading the spec and then reading the code makes more sense if this is not in there.

libp2p/protocols/kademlia.nim
```diff
@@ -18,66 +18,21 @@ export routingtable, protobuf, types, find, get, put, provider, ping
...
-      let randomKey = randomKeyInBucketRange(kad.rtable.selfId, i, kad.rng[])
+      let randomKey = randomKeyInBucket(kad.rtable.selfId, i, kad.rng[])
       discard await kad.findNode(randomKey)
 
+  info "Bootstrap complete"
+
 proc maintainBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
   heartbeat "refresh buckets", kad.config.bucketRefreshTime:
-    await kad.refreshBuckets()
+    await kad.bootstrap()
```
rlve, line 37: `refreshBuckets` has different logic than `bootstrap`, so why the change?

libp2p/protocols/kademlia.nim
```diff
@@ -18,66 +18,21 @@ export routingtable, protobuf, types, find, get, put, provider, ping
...
-  discard kad.rtable.insert(peerId)
-
 proc bootstrap*(kad: KadDHT) {.async: (raises: [CancelledError]).} =
-  for (peerId, addrs) in kad.bootstrapNodes:
-    await kad.bootstrapNode(peerId, addrs)
-
-  let key = PeerId.random(kad.rng).valueOr:
-    doAssert(false, "this should never happen")
-    return
+  ## Sends a findNode to a random peer ID for each non-empty k-bucket
```
rlve:
```suggestion
  ## Sends a findNode with selfId and then a random peer ID for each non-empty k-bucket
```

libp2p/protocols/kademlia.nim
gmelodie: Because it shouldn't!

> The bootstrap process is responsible for keeping the routing table filled and healthy throughout time.

Now that the bootstrap function behaves like it should as per the spec, we should just call it periodically

richard-ramos (approved):

PR #2017, opened 2026-01-09
refactor(kad): iterative search and cid handling

richard-ramos:
This PR extracts the logic for iterative search out of findNode, so it can be reused also when getting values and providers, since the logic is extremely similar among these three methods, with difference being what to do when you get a reply, and the stop condition.

Also, while testing the cbindings I could see that using the multihash as a key for registering providers lead to not finding providers, since the multihash on its own does not contain enough information to rebuild a cid. (it will be missing info like the version). On handleProvider this does not cause issues as we pretty much don't do anything with the cid itself, we just use the hash

Also ended up adding ourselves as providers if we already provide a key when getProviders is executed

And finally changed setupKadSwitch from template to proc, because when adding extra parameters, nim compiler complained :shrug:

Replaces #2016

### DIFF
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -183,19 +183,19 @@ method findNode*(
     discard
 
   let stop = proc(state: LookupState): bool {.raises: [], gcsafe.} =
+    # findNode does not have a stop condition per se except 
+    # for when there's no more nodes to query to traverse the DHT
     false
 
+  let dispatchFind = proc(
+      kad: KadDHT, peer: PeerId, target: Key
+  ): Future[Opt[Message]] {.
+      async: (raises: [CancelledError, DialFailedError, LPStreamError]), gcsafe
+  .} =
+    return await dispatchFindNode(kad, peer, target)
+
   let state = await kad.iterativeLookup(
-    target,
-    proc(
-        kad: KadDHT, peer: PeerId, target: Key
-    ): Future[Opt[Message]] {.
-        async: (raises: [CancelledError, DialFailedError, LPStreamError]), gcsafe
-    .} =
-      return await dispatchFindNode(kad, peer, target),
-    ignoreReply,
-    stop,
-    countFailedAsResponded = true,
+    target, dispatchFind, ignoreReply, stop, countFailedAsResponded = true
   )
 
   return state.selectCloserPeers(kad.config.replication, excludeResponded = false)
--- a/libp2p/protocols/kademlia/get.nim
+++ b/libp2p/protocols/kademlia/get.nim
@@ -30,7 +30,8 @@ proc dispatchGetVal(
     error "GetValue reply decode fail", error = error, conn = conn
     return Opt.none(Message)
 
-  kad.updatePeers(@[PeerInfo(peerId: conn.peerId, addrs: @[conn.observedAddr.get()])])
+  conn.observedAddr.withValue(observedAddr):
+    kad.updatePeers(@[PeerInfo(peerId: conn.peerId, addrs: @[observedAddr])])
 
   return Opt.some(reply)
 
--- a/libp2p/protocols/kademlia/provider.nim
+++ b/libp2p/protocols/kademlia/provider.nim
@@ -157,6 +157,10 @@ proc manageExpiredProviders*(kad: KadDHT) {.async: (raises: [CancelledError]).} 
 proc handleAddProvider*(
     kad: KadDHT, conn: Connection, msg: Message
 ) {.async: (raises: [CancelledError]).} =
+  if not MultiHash.validate(msg.key):
+    error "Received key is an invalid Multihash", msg = msg, conn = conn, key = msg.key
+    return
+
   # filter out infos that do not match sender's
   let peerBytes = conn.peerId.getBytes()
 
@@ -191,7 +195,8 @@ proc dispatchGetProviders*(
 
   debug "Received reply for GetProviders", peer = peer, reply = reply
 
-  kad.updatePeers(@[PeerInfo(peerId: conn.peerId, addrs: @[conn.observedAddr.get()])])
+  conn.observedAddr.withValue(observedAddr):
+    kad.updatePeers(@[PeerInfo(peerId: conn.peerId, addrs: @[observedAddr])])
 
   return Opt.some(reply)
 
--- a/tests/libp2p/kademlia/test_find.nim
+++ b/tests/libp2p/kademlia/test_find.nim
@@ -170,8 +170,8 @@ suite "KadDHT Find":
     check kad1.hasKey(kad3.rtable.selfId)
 
   asyncTest "Find node with empty key returns closest peers":
-    var (switch1, kad1) = setupKadSwitch(PermissiveValidator(), CandSelector())
-    var (switch2, kad2) = setupKadSwitch(
+    var (switch1, kad1) = await setupKadSwitch(PermissiveValidator(), CandSelector())
+    var (switch2, kad2) = await setupKadSwitch(
       PermissiveValidator(),
       CandSelector(),
       @[(switch1.peerInfo.peerId, switch1.peerInfo.addrs)],
@@ -181,7 +181,8 @@ suite "KadDHT Find":
 
     # Send FIND_NODE with empty key directly
     let emptyKey: Key = @[]
-    let response = await switch2.dispatchFindNode(switch1.peerInfo.peerId, emptyKey)
+    let response =
+      (await kad2.dispatchFindNode(switch1.peerInfo.peerId, emptyKey)).value()
 
     # Empty key is accepted and a valid response is returned
     check:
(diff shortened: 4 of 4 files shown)


libp2p/protocols/kademlia/get.nim
```diff
@@ -7,20 +7,19 @@
...
-    received: ReceivedTable,
-    candidates: CandidatePeers,
-) {.async: (raises: [CancelledError, DialFailedError, LPStreamError]).} =
-  let conn = await switch.dial(peer, KadCodec)
+    kad: KadDHT, peer: PeerId, key: Key
+): Future[Opt[Message]] {.
+    async: (raises: [CancelledError, DialFailedError, LPStreamError]), gcsafe
+.} =
+  let conn =
+    await kad.switch.dial(peer, kad.switch.peerStore[AddressBook][peer], KadCodec)
```
gmelodie, line 22: What happens when `peer` is not in the address book? Does it not raise a `KeyError`?

libp2p/protocols/kademlia/get.nim
```diff
@@ -91,19 +67,35 @@ proc getValue*(
...
+      return
+
+    let time = record.timeReceived.valueOr:
+      debug "GetValue returned record with no timeReceived, using current time instead",
+        reply = reply
+      TimeStamp($times.now().utc)
+
+    received[peer] = Opt.some(EntryRecord(value: value, time: time))
+
+  let stopCond = proc(state: LookupState): bool {.gcsafe.} =
```
gmelodie: nit
```suggestion
  let stop = proc(state: LookupState): bool {.gcsafe.} =
```

libp2p/protocols/kademlia/find.nim
```diff
@@ -112,35 +123,76 @@ proc updatePeers*(kad: KadDHT, peers: seq[(PeerId, seq[MultiAddress])]) {.raises
...
-    kad: KadDHT, target: Key
-): Future[seq[PeerId]] {.async: (raises: [CancelledError]).} =
-  ## Iteratively search for the k closest peers to a `target` key.
-
+proc iterativeLookup*(
+    kad: KadDHT,
+    target: Key,
+    dispatch: DispatchProc,
+    onReply: ReplyHandler,
+    stopCond: StopCond = nil,
```
gmelodie: Why would we want for `stopCond` to ever be nil? In theory this could be looped forever if no node in the DHT had a key, right? It would be interesting to have a test where this is `nil` and look for a key that is not in any node.

libp2p/protocols/kademlia/get.nim
```diff
@@ -91,19 +67,35 @@ proc getValue*(
...
-        discard
-    # filter out peers that have responded
-    candidates[] = candidates[].filterIt(not received.hasKey(it))
-    curTry.inc()
+  let onReply = proc(
+      peer: PeerId, msgOpt: Opt[Message], state: var LookupState
+  ): Future[void] {.async: (raises: []), gcsafe.} =
+    received[peer] = Opt.none(EntryRecord)
+
+    let reply = msgOpt.valueOr:
```
gmelodie, line 76:
```suggestion
    let reply = msgOpt.valueOr:
      debug "Empty reply", reply = reply
```

libp2p/protocols/kademlia/find.nim
richard-ramos: hm yeah, i agree, nil stop conditions make no sense. In any case this should be an assertion to ensure the dev using the iterative search function does not pass a nil function.

richard-ramos: Actually, i just remembered why i allow the stop condition to be nil. In findNode I dont use a stop condition, but instead rely on `selectCloserPeers` returning no peers

libp2p/protocols/kademlia/get.nim
richard-ramos: It returns an empty seq:
```
proc `[]`*[T](peerBook: PeerBook[T], peerId: PeerId): T {.public.} =
  ## Get all known metadata of a provided peer, or default(T) if missing
  peerBook.book.getOrDefault(peerId)
  ```
  In this case, it will not be able to dial the peerId, but should be fine as dispatchGetVal does expect the addresses to be there, or will just fail by emitting a `DialFailederror`

File: libp2p/protocols/kademlia/find.nim (hunk 3 above)
Comment (gmelodie) [reply libp2p/protocols/kademlia/find.nim]: In this case I think I'd be better to pass a function that does nothing in findNode, and never allow stopCond to be nil

Comment (richard-ramos) [reply libp2p/protocols/kademlia/find.nim]: Fixed in last commit

File: libp2p/protocols/kademlia/find.nim [hunk 5]
```diff
@@ -112,35 +123,80 @@ proc updatePeers*(kad: KadDHT, peers: seq[(PeerId, seq[MultiAddress])]) {.raises
...
+    false
+
+  let state = await kad.iterativeLookup(
+    target,
+    proc(
+        kad: KadDHT, peer: PeerId, target: Key
+    ): Future[Opt[Message]] {.
+        async: (raises: [CancelledError, DialFailedError, LPStreamError]), gcsafe
+    .} =
+      return await dispatchFindNode(kad, peer, target),
```
vladopajic: i would defined it above, in line with `let stop and ignoreReply` since this is same callback method.

libp2p/protocols/kademlia/find.nim
```diff
@@ -112,35 +123,80 @@ proc updatePeers*(kad: KadDHT, peers: seq[(PeerId, seq[MultiAddress])]) {.raises
...
+): Future[seq[PeerId]] {.base, async: (raises: [CancelledError]).} =
+  ## Iteratively search for the k closest peers to a `target` key.
+
+  let ignoreReply = proc(
+      _: PeerId, _: Opt[Message], _: var LookupState
+  ): Future[void] {.async: (raises: []), gcsafe.} =
+    discard
+
+  let stop = proc(state: LookupState): bool {.raises: [], gcsafe.} =
+    false
```
vladopajic, line 188: stop is always false, why ?

richard-ramos: This is because finding a node does not have a stop condition per se except for when there's no more nodes to query to traverse the DHT.  This happens inside the iterativeLookup proc:
```nim
    let toQuery = state.selectCloserPeers(kad.config.alpha)
    if toQuery.len() == 0:
      break
```

vladopajic: right! maybe would be nice to have a comment?

libp2p/protocols/kademlia/provider.nim
```diff
@@ -169,44 +169,31 @@ proc handleAddProvider*(
       ProviderRecord(
         provider: peer,
         expiresAt: chronos.Moment.now() + kad.config.providerExpirationInterval,
-        key: msg.key.toCid(),
+        key: msg.key,
```
gmelodie, line 176: This might be problematic as the spec says
> The target node verifies key is a valid CID

libp2p/protocols/kademlia/put.nim
```diff
@@ -25,7 +25,7 @@ proc isBestValue(kad: KadDHT, key: Key, record: EntryRecord): bool =
 proc dispatchPutVal*(
     switch: Switch, peer: PeerId, key: Key, value: seq[byte]
 ) {.async: (raises: [CancelledError, DialFailedError, LPStreamError]).} =
-  let conn = await switch.dial(peer, KadCodec)
+  let conn = await switch.dial(peer, switch.peerStore[AddressBook][peer], KadCodec)
```
vladopajic, line 28: `switch.peerStore[AddressBook][peer]` this this trow `ValueError`?

libp2p/protocols/kademlia/find.nim
vladopajic:
```
 let dispatchFind = proc(
        kad: KadDHT, peer: PeerId, target: Key
    ): Future[Opt[Message]] {.
        async: (raises: [CancelledError, DialFailedError, LPStreamError]), gcsafe
    .} =
      return await dispatchFindNode(kad, peer, target),
```

libp2p/protocols/kademlia/put.nim
richard-ramos: it does not as Peerstore overrides `[]` and it returns a @[]:

```nim
proc `[]`*[T](peerBook: PeerBook[T], peerId: PeerId): T {.public.} =
  ## Get all known metadata of a provided peer, or default(T) if missing
  peerBook.book.getOrDefault(peerId)
```

libp2p/protocols/kademlia/provider.nim
richard-ramos: Added a validation on the target node to check if the key is a valid multihash https://github.com/vacp2p/nim-libp2p/pull/2017/changes/63657ced2dde720f2ad76b01ff4a9d7229e60402 as the spec says that `It is also worth noting that the keys for provider records are multihashes.`

gmelodie (approved):

PR #2020, opened 2026-01-12
fix(kad-dht): stop condition on find_node and track whether node responded succesfully or not

richard-ramos:
Current code can potentially query the whole network. With this change it should stop when there are no unqueried peers left in the short list (as it  drop those whole  xor distance is greater than alpha), or, did queries and didnt discover anything closer than the best we already had.

h/t to Vlado for reporting this issue!

### DIFF
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -16,11 +16,15 @@ import ./[routingtable, protobuf, types]
 logScope:
   topics = "kad-dht find"
 
+type RespondedStatus* = enum
+  Failed
+  Success
+
 type LookupState* = object
   kad: KadDHT
   target*: Key
   shortlist*: Table[PeerId, XorDistance]
-  responded*: HashSet[PeerId]
+  responded*: Table[PeerId, RespondedStatus]
   attempts: Table[PeerId, int]
 
 type DispatchProc* = proc(kad: KadDHT, peer: PeerId, target: Key): Future[Opt[Message]] {.
@@ -88,8 +92,28 @@ proc selectCloserPeers*(
     # take at most alpha peers
     .take(amount)
 
+proc hasResponsesFromClosestAvailable(state: LookupState): bool {.raises: [], gcsafe.} =
+  ## True when all closest k AVAILABLE peers have responded.
+  let candidates = state.sortedShortlist(excludeResponded = false)
+  if candidates.len == 0:
+    return true
+
+  var closetsRespondedCnt = 0
+  for (c, _) in candidates:
+    if state.responded.hasKey(c):
+      try:
+        if state.responded[c] == RespondedStatus.Success:
+          closetsRespondedCnt.inc(1)
+      except KeyError:
+        raiseAssert "checked with hasKey"
+    else:
+      # It's a close peer but has not been queried yet
+      break
+
+  return closetsRespondedCnt >= state.kad.config.replication
+
 proc init*(T: type LookupState, kad: KadDHT, target: Key): T =
-  var res = LookupState(kad: kad, target: target, responded: initHashSet[PeerId]())
+  var res = LookupState(kad: kad, target: target)
   for pid in kad.rtable.findClosestPeerIds(target, kad.config.replication):
     res.shortlist[pid] = xorDistance(pid, target, kad.rtable.config.hasher)
 
@@ -132,7 +156,6 @@ proc iterativeLookup*(
     dispatch: DispatchProc,
     onReply: ReplyHandler,
     stopCond: StopCond,
-    countFailedAsResponded: bool = true,
 ): Future[LookupState] {.async: (raises: [CancelledError]).} =
   var state = LookupState.init(kad, target)
 
@@ -164,8 +187,8 @@ proc iterativeLookup*(
     for (fut, peerId) in zip(rpcBatch, toQuery):
       if not fut.completed():
         continue
-      if countFailedAsResponded or not fut.failed():
-        state.responded.incl(peerId)
+      state.responded[peerId] =
+        if fut.failed(): RespondedStatus.Failed else: RespondedStatus.Success
 
     for (peerId, msg) in completedRPCBatch:
       msg.withValue(reply):
@@ -186,9 +209,7 @@ method findNode*(
     discard
 
   let stop = proc(state: LookupState): bool {.raises: [], gcsafe.} =
-    # findNode does not have a stop condition per se except 
-    # for when there's no more nodes to query to traverse the DHT
-    false
+    state.hasResponsesFromClosestAvailable()
 
   let dispatchFind = proc(
       kad: KadDHT, peer: PeerId, target: Key
@@ -197,9 +218,7 @@ method findNode*(
   .} =
     return await dispatchFindNode(kad, peer, target)
 
-  let state = await kad.iterativeLookup(
-    target, dispatchFind, ignoreReply, stop, countFailedAsResponded = true
-  )
+  let state = await kad.iterativeLookup(target, dispatchFind, ignoreReply, stop)
 
   return state.selectCloserPeers(kad.config.replication, excludeResponded = false)
 
--- a/tests/libp2p/kademlia/test_find.nim
+++ b/tests/libp2p/kademlia/test_find.nim
@@ -9,7 +9,7 @@
 
 {.used.}
 
-import chronos, std/[enumerate, sets], sequtils, tables
+import chronos, std/[enumerate, sequtils, tables]
 import ../../../libp2p/[protocols/kademlia, switch, builders]
 import ../../tools/[unittest]
 import ./utils.nim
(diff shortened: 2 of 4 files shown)


libp2p/protocols/kademlia/find.nim
```diff
@@ -47,6 +48,56 @@ proc updateShortlist*(
...
+  var sorted = newSeqOfCap[(PeerId, XorDistance)](state.shortlist.len)
+  for pid, dist in state.shortlist.pairs():
+    sorted.add((pid, dist))
+
+  sorted.sort(
+    proc(a, b: (PeerId, XorDistance)): int =
+      cmp(a[1], b[1])
+  )
+
+  let keepCount = min(alpha, sorted.len)
```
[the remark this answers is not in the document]
richard-ramos: +1

libp2p/protocols/kademlia/find.nim
```diff
@@ -47,6 +48,56 @@ proc updateShortlist*(
...
+  for _, dist in state.shortlist.pairs():
+    if not hasValue or dist < minDist:
+      minDist = dist
+      hasValue = true
+  if hasValue:
+    return Opt.some(minDist)
+
+  Opt.none(XorDistance)
+
+proc trimShortlist(state: var LookupState, alpha: int) =
```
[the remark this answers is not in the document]
richard-ramos: +1

libp2p/protocols/kademlia/find.nim
```diff
@@ -137,6 +184,8 @@ proc iterativeLookup*(
     if stopCond(state):
       break
 
+    let beforeClosest = state.closestDistance()
+
     let toQuery = state.selectCloserPeers(kad.config.alpha)
     if toQuery.len() == 0:
```
vladopajic: `selectCloserPeers` can internally store distance from first node it returns to some property eg `bestNodeDistance`. then this propety could be used without `let beforeClosest = state.closestDistance()` completely

libp2p/protocols/kademlia/find.nim
```diff
@@ -170,6 +219,8 @@ proc iterativeLookup*(
         kad.updatePeers(newPeerInfos)
       await onReply(peerId, msg, state)
 
+    state.updateAfterRound(beforeClosest, kad.config.replication)
```
vladopajic: when we look at everything that this call does, some things are not needed here for other operations like get.
it would be better for this place to only have `state.trimShortlist(keep)`

and in `stop()` callback to have code that does this part:

```
  let afterClosest = state.closestDistance()
  if beforeClosest.isSome() and afterClosest.isSome():
    state.noNewNodesDiscovered = not (afterClosest.get() < beforeClosest.get())
  else:
    state.noNewNodesDiscovered = false
 ```

File: libp2p/protocols/kademlia/find.nim [hunk 5]
```diff
@@ -183,9 +234,8 @@ method findNode*(
     discard
 
   let stop = proc(state: LookupState): bool {.raises: [], gcsafe.} =
-    # findNode does not have a stop condition per se except 
-    # for when there's no more nodes to query to traverse the DHT
-    false
+    ## stop when no closer peers were discovered in the previous round
+    state.noNewNodesDiscovered
```
vladopajic: use `bestNodeDistance` propery from previous comment, then do

```
  let afterClosest = state.closestDistance()
  if bestNodeDistance.isSome() and afterClosest.isSome():
    return not (afterClosest.get() < bestNodeDistance.get())
  else:
    return false
```

libp2p/protocols/kademlia/find.nim
```diff
@@ -47,6 +48,52 @@ proc updateShortlist*(
...
+  var keepers = initHashSet[PeerId]()
+  for i in 0 ..< keepCount:
+    keepers.incl(sorted[i][0])
+
+  var toRemove: seq[PeerId]
+  for pid in state.shortlist.keys:
+    if not keepers.contains(pid):
+      toRemove.add(pid)
+  for pid in toRemove:
+    state.shortlist.del(pid)
```
vladopajic: could this be something like:
`state.shortlist = newShortlist`

and in the code above just creates `newShortlist` table

tests/libp2p/kademlia/test_find.nim
```diff
@@ -267,3 +267,44 @@ suite "KadDHT Find":
...
+    kad.populateRoutingTable(5)
+
+    var extraPeerIds: seq[PeerId]
+    for i in 0 ..< 10:
+      extraPeerIds.add(randomPeerId())
+
+    let replyMsg = Message(
+      msgType: MessageType.findNode,
+      closerPeers: extraPeerIds.mapIt(Peer(id: it.getBytes, addrs: @[])),
+    )
```
vladopajic: nit: better inside `dispatch` proc

libp2p/protocols/kademlia/find.nim
```diff
@@ -47,6 +48,52 @@ proc updateShortlist*(
...
+    return
+
+  var sorted = newSeqOfCap[(PeerId, XorDistance)](state.shortlist.len)
+  for pid, dist in state.shortlist.pairs():
+    sorted.add((pid, dist))
+
+  sorted.sort(
+    proc(a, b: (PeerId, XorDistance)): int =
+      cmp(a[1], b[1])
+  )
```
gmelodie: 1. create a proc called `sortedShortlist` to hold this code
2. change what is currently called `sortedShortlist` to something like `nextCandidates` and make it call the new `sortedShortlist`

libp2p/protocols/kademlia/find.nim
```diff
@@ -47,6 +48,52 @@ proc updateShortlist*(
...
+
+  let keepCount = keep
```
gmelodie: Just call the argument `keepCount` from the beginning

libp2p/protocols/kademlia/find.nim
```diff
@@ -47,6 +48,52 @@ proc updateShortlist*(
...
+  var keepers = initHashSet[PeerId]()
+  for i in 0 ..< keepCount:
+    keepers.incl(sorted[i][0])
```
gmelodie:
```suggestion
  let keepers = sorted.mapIt(it[0])[0..<keepCount].toHashSet()
```

libp2p/protocols/kademlia/find.nim
```diff
@@ -47,6 +48,52 @@ proc updateShortlist*(
...
+
+  var toRemove: seq[PeerId]
+  for pid in state.shortlist.keys:
+    if not keepers.contains(pid):
+      toRemove.add(pid)
```
gmelodie:
```suggestion
  let toRemove = state.shortlist.keys.toHashSet() - keepers
```

libp2p/protocols/kademlia/find.nim
```diff
@@ -47,6 +48,52 @@ proc updateShortlist*(
...
+  for pid in toRemove:
+    state.shortlist.del(pid)
+
+proc updateAfterRound*(
```
gmelodie: Could we not incorporate this into the stop condition instead of using a stateful variable (`noNewNodesDiscovered`)?

libp2p/protocols/kademlia/find.nim
```diff
@@ -85,8 +89,28 @@ proc selectCloserPeers*(
...
+  for (c, _) in candidates:
+    if state.responded.hasKey(c):
+      try:
+        if state.responded[c] == RespondedStatus.Success:
+          closetsRespondedCnt.inc(1)
+      except KeyError:
+        raiseAssert "checked with hasKey"
+    else:
+      # It's a close peer but has not been queried yet
+      break
```
vladopajic, line 111:
```suggestion
    try:
      if state.responded[c] == RespondedStatus.Success:
        closetsRespondedCnt.inc(1)
    except KeyError:
        # It's a close peer but has not been queried yet
       break
```

why `break`?  why not continue  to count? seems to like like code would be more robust if it not brakes here

for example what if somehow some candidate in early position in this proc call is not queried, because it was in later position when iterating - due to some unstable sorting or ... ? or this could not really happen?

vladopajic: i think code assumes that current peer is closer then next peer. but that doesn't have to be true - they could be at same distance

vladopajic: peer, distance to target
(1, 3), (2, 3), (3, 3), (4, 3)

candidates
(1, 2, 3, 4)

state.responded:
(1, true) (2, false), (3, true), (4, true)

replication 2

with these values code fails but there is response from 3 nodes

vladopajic: or example is missing something ?

richard-ramos: if node 2 is not present in `state.responded`, it means it has not been queried yet.
The spec says:
```
The lookup terminates when the initiator has queried and gotten responses from the k closest nodes it has seen.
```
At that moment node 2 has been seen (i.e. it is now part of the shortlist), but not queried nor gotten a response from yet.

richard-ramos: > why break? why not continue to count? seems to like like code would be more robust if it not brakes here

because we want to make sure that we only take into account that we have obtained a response from the closest nodes. If one of the closest nodes is still pending to be queried, we can't stop the iterative lookup yet

Issue #2013, opened 2026-01-09
Duplicate bootstrapping logs on findNode tests


gmelodie:
`Simple find node` test (and others) show a duplicate bootstrapping log, indicating that the function is being called twice

PR #2022, opened 2026-01-13
fix(kad): Bootstrap

rlve:
Addresses: https://github.com/vacp2p/nim-libp2p/issues/2006 (again) and https://github.com/vacp2p/nim-libp2p/issues/2013

Changes:
- refactor `bootstrap`: 
  - ensure it sends `findNode` with random key for all non-empty buckets on the first run - without `isStale` check (because all the non-empty buckets are fresh at the beginning)
  - on the consecutive runs it will refresh stale buckets only
- fix `hearbeat sleepFirst` - it didn't work, bootstrap was run twice on startup and that's why we observed duplicated logs
- add tests

### DIFF
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -13,6 +13,9 @@ import ../../[peerid, peerinfo, switch, multihash]
 import ../protocol
 import ./[routingtable, protobuf, types]
 
+logScope:
+  topics = "kad-dht find"
+
 type LookupState* = object
   kad: KadDHT
   target*: Key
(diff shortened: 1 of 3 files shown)


rlve:
> lgtm
> 
> not sure about `refactor` pr title tag, it suggest significant changes, but this pr doesn't really introduce any.
> 
> smaller things that are improved or fixed while adding tests can still be under `test` tag or `chore`. or it could be `fix` if pr fixes some issue that was discovered with tests and only scope of pr is to fix that bug and it makes sense to have it like that in history, eg "fix(kad): bootstrap failed on windows" then when making release users that use this library can see "aha they fixed this issue that i needed - nice!"

@vladopajic okay then, fix it is 😄

PR #2023, opened 2026-01-13
test(kad): bootstrap and refactor find tests

rlve:
Changes:
- add bootstrap tests:
  - `"bootstrap with mixed fresh and stale buckets refreshes only stale"`
  - `"KadDHT Bootstrap Component - bootstrap discovers new peers through network"`
- refactor find tests
  - do not rely on bootstrap since we have dedicated bootstrap tests, connect nodes directly instead
  - use common helpers everywhere

### DIFF
--- a/tests/libp2p/kademlia/utils.nim
+++ b/tests/libp2p/kademlia/utils.nim
@@ -187,7 +187,7 @@ proc nonEmptyBuckets*(kad: KadDHT): seq[int] =
 
 proc makeBucketStale*(bucket: var Bucket) =
   for peer in bucket.peers.mitems:
-    peer.lastSeen = Moment.now() - 40.minutes
+    peer.lastSeen = Moment.now() - (DefaultBucketStaleTime + 1.minutes)
 
 proc sortPeers*(
     peers: seq[PeerId], targetKey: Key, hasher: Opt[XorDHasher]
--- a/libp2p/protocols/kademlia/routingtable.nim
+++ b/libp2p/protocols/kademlia/routingtable.nim
@@ -129,7 +129,7 @@ proc isStale*(bucket: Bucket): bool =
   if bucket.peers.len == 0:
     return true
   for p in bucket.peers:
-    if Moment.now() - p.lastSeen > 30.minutes:
+    if Moment.now() - p.lastSeen > DefaultBucketStaleTime:
       return true
   return false
 
--- a/libp2p/protocols/kademlia/types.nim
+++ b/libp2p/protocols/kademlia/types.nim
@@ -20,6 +20,8 @@ const
   DefaultMaxBuckets* = 256
   DefaultTimeout* = 5.seconds
   DefaultBucketRefreshTime* = 10.minutes
+  DefaultBucketStaleTime* = 30.minutes
+    # peer not seen for this duration marks bucket stale
   DefaultRetries* = 5
   DefaultReplication* = 20 ## aka `k` in the spec
   DefaultAlpha* = 10 # concurrency parameter


tests/libp2p/kademlia/test_bootstrap.nim
```diff
@@ -51,7 +42,7 @@ suite "KadDHT Bootstrap":
 
     # Add peers - they will be fresh (just added)
     kad.populateRoutingTable(5)
-    check kad.countNonEmptyBuckets() >= 1
+    check kad.getNonEmptyBuckets().len >= 1
```
gmelodie: nit
```suggestion
    check kad.nonEmptyBuckets().len >= 1
```

tests/libp2p/kademlia/test_bootstrap.nim
```diff
@@ -65,35 +56,85 @@ suite "KadDHT Bootstrap":
...
+    check bucketIndices.len >= 2
+
+    for index in bucketIndices:
+      makeBucketStale(kad.rtable.buckets[index])
+
+    kad.findNodeCalls = @[]
+    await kad.bootstrap()
+
+    # Self lookup + one lookup per stale bucket
+    check kad.findNodeCalls.len == 1 + bucketIndices.len
```
gmelodie:
```suggestion
    check kad.findNodeCalls.len == bucketIndices.len + 1
```

tests/libp2p/kademlia/test_bootstrap.nim
```diff
@@ -65,35 +56,85 @@ suite "KadDHT Bootstrap":
...
-      bucketIndex(kad.rtable.selfId, peerId.toKey(), kad.rtable.config.hasher)
-    kad.rtable.buckets[bucketIdx].peers[0].lastSeen = Moment.now() - 40.minutes
+    # Make only the first bucket stale
+    let staleBucketIndex = bucketIndices[0]
+    makeBucketStale(kad.rtable.buckets[staleBucketIndex])
+    check kad.rtable.buckets[staleBucketIndex].isStale()
 
-    check kad.rtable.buckets[bucketIdx].isStale()
+    # Verify that the rest of non-empty buckets is fresh
+    for index in bucketIndices[1 ..^ 1]:
```
gmelodie: does this `[1 ..^ 1]` notation not mean the same as [1]? If not, can you make it a bit clearer?

rlve: > Backwards index operator can be used in a slice to address the array from the high bound. ^1 is equivalent to arr1.len-1

So it means from 2nd element till the last. If you read the whole scenario it's pretty clear what it does.

gmelodie:
```nim
for i in 1 ..< bucketIndices.len:
  let index = bucketIndices[i]
```

rlve: You did not convince me 😄  This notation is used across the codebase, so I'll keep it

gmelodie: It's an ugly one, it shouldn't be used anywhere

vladopajic: sometimes we can fall into the trap of using the same method for everything. there are always edge cases when other methods should be used, or general principle shouldn't be followed.

since this is a test i wouldn't personally care about either, but if this is library code then variant with `[1 .. 1]` introduces unnecessary complexity (copying takes time and memory is trashed right after). 

personally `for i in 1 .. buckets.len` would be the preferred way  and if this could not do it, then `[ x .. y ]` as a last resort.

vladopajic: when writing code one should not use excuses not to do better, if legacy code was built with a suboptimal approach, one shouldn't really conform to it. a better code variant should always be preferred in new prs. this means that there will be different standards, but transitions are always painful.

tests/libp2p/kademlia/utils.nim
```diff
@@ -145,6 +178,17 @@ proc getPeersfromRoutingTable*(kad: KadDHT): seq[PeerId] =
...
+proc nonEmptyBuckets*(kad: KadDHT): seq[int] =
+  var bucketIndices: seq[int]
+  for i, bucket in kad.rtable.buckets:
+    if bucket.peers.len > 0:
+      bucketIndices.add(i)
+  bucketIndices
+
+proc makeBucketStale*(bucket: var Bucket) =
+  for peer in bucket.peers.mitems:
+    peer.lastSeen = Moment.now() - 40.minutes
```
vladopajic: 40.munutes must come from wherever this value is stored

tests/libp2p/kademlia/test_bootstrap.nim
rlve: If notation `bucketIndices[1 ..^ 1]` is suboptimal due to performance reasons, then it's fine, I can change it. But this whole discussion started on it's being ugly, which is just an opinion

tests/libp2p/kademlia/utils.nim
rlve: https://github.com/vacp2p/nim-libp2p/blob/e468bf3bfb0d44afce3079f5191a9e19db93b7f4/libp2p/protocols/kademlia/routingtable.nim#L128-L134

Currently the threshold is not stored in any const.

vladopajic: aha, but where does `30 minutes` come from? we should use that value here

vladopajic: could we have constant for `30.minutes` and use it here?

rlve: sure

tests/libp2p/kademlia/test_bootstrap.nim
rlve: changed in https://github.com/vacp2p/nim-libp2p/pull/2023/changes/a65ae63a1bdc96251a53bf7a504d52003e339a11 🙄

tests/libp2p/kademlia/utils.nim
rlve: changed in https://github.com/vacp2p/nim-libp2p/pull/2023/changes/b23ecb88bbcffefc72399c62ff289989f0bb1209

PR #2024, opened 2026-01-13
chore(kad): add quorum ignored test

gmelodie:


tests/libp2p/kademlia/test_get.nim
```diff
@@ -265,3 +265,42 @@ suite "KadDHT Get":
...
+      chronos.seconds(1),
+    )
+
+    defer:
+      await allFutures(switch1.stop(), switch2.stop(), switch3.stop())
+
+    let
+      key = kad1.rtable.selfId
+      value1 = @[1.byte, 2, 3, 4, 5]
+      value2 = @[1.byte, 3, 2, 4, 5]
```
vladopajic: this should be true:
`value1 != value2`
otherwise asserts below do not correctly assert behavior

vladopajic: ah.... okay could we have some more obvious values like [1, 1, 1, 1]  and [2, 2, 2, 2]

vladopajic:
```suggestion
      value1 = @[1.byte, 1, 1, 1, 1]
      value2 = @[2.byte, 2, 2, 2, 2]
```

richard-ramos (approved):

PR #2025, opened 2026-01-14
test(interop): Kad test template

rlve:
Continuation of @Eshaan-byte PR: https://github.com/vacp2p/nim-libp2p/pull/1990

Addresses: https://github.com/vacp2p/nim-libp2p/issues/1987

Changes:
- move `kadInteropTest` to `tests/interop/kad`
- add tests for `kadInteropTest` proc

### DIFF
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -16,11 +16,15 @@ import ./[routingtable, protobuf, types]
 logScope:
   topics = "kad-dht find"
 
+type RespondedStatus* = enum
+  Failed
+  Success
+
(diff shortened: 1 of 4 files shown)


tests/interop/test_kad.nim
```diff
@@ -0,0 +1,68 @@
...
+    checkTrackers()
+
+  asyncTest "Happy path":
+    let switch = createSwitch()
+
+    await switch.start()
+    defer:
+      await switch.stop()
+
+    const ourAddress = "/ip4/127.0.0.1/tcp/0"
```
vladopajic: nit:  `ourAddress` as common const

PR #2026, opened 2026-01-14
feat(nix): build cbindings

gmelodie:
- Enable `nimble.lock` files in the root and the `cbind` directories (thus requiring `nimble setup` to be run in tests)
- Use `nimble.lock` to generate `nix/deps.nix` and `nix/cbind-deps.nix`
- Use `nix/deps.nix` and `nix/cbind-deps.nix` to build libp2p and cbindings in `flake.nix`

Future steps are to further use nix on tests, fully replacing current `nimble` vs `nim` shenanigans

Obs: this will be used to build logos-libp2p-module https://github.com/logos-co/logos-libp2p-module/pull/2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
           - platform:
               os: windows
             builder: windows-2022
-            shell: msys2 {0}
+            shell: bash
 
     defaults:
       run:
@@ -81,15 +81,25 @@ jobs:
           # The change happened on Nimble v0.14.0. Also forcing the deps to be reinstalled on each os and cpu.
           key: nimbledeps-${{ matrix.nim.ref }}-${{ matrix.builder }}-${{ matrix.platform.cpu }}-${{ hashFiles('.pinned') }} # hashFiles returns a different value on windows
 
-      - name: Install runtime deps (windows)
-        if: ${{ matrix.platform.os == 'windows'}}
+      - name: Restore llvm-mingw (Windows) from cache
+        if: runner.os == 'Windows'
+        id: windows-mingw-cache
+        uses: actions/cache@v4
+        with:
+          path: external/mingw-${{ matrix.platform.cpu }}
+          key: 'mingw-llvm-17-${{ matrix.platform.cpu }}'
+
+      - name: Install llvm-mingw dependency (Windows)
+        if: >
+          steps.windows-mingw-cache.outputs.cache-hit != 'true' &&
+          runner.os == 'Windows'
         run: |
-          pacman -S --noconfirm base-devel gcc nasm
-          
-      - name: Install deps
-        if: ${{ steps.deps-cache.outputs.cache-hit != 'true' }}
-        run: |
-          nimble install_pinned
+          mkdir -p external
+          MINGW_BASE="https://github.com/mstorsjo/llvm-mingw/releases/download/20230905"
+          MINGW_URL="$MINGW_BASE/llvm-mingw-20230905-ucrt-x86_64.zip"
+          curl -L "$MINGW_URL" -o "external/mingw-${{ matrix.platform.cpu }}.zip"
+          7z x -y "external/mingw-${{ matrix.platform.cpu }}.zip" -oexternal/mingw-${{ matrix.platform.cpu }}/
+          mv external/mingw-${{ matrix.platform.cpu }}/**/* ./external/mingw-${{ matrix.platform.cpu }}
 
       - name: Use gcc 14
         if: ${{ matrix.platform.os == 'linux-gcc-14'}}
@@ -100,6 +110,25 @@ jobs:
           # Set GCC-14 as the default
           sudo update-alternatives --set gcc /usr/bin/gcc-14
 
+      - name: Path to cached dependencies (Windows)
+        if: >
+          runner.os == 'Windows'
+        run: |
+          echo '${{ github.workspace }}'"/external/mingw-${{ matrix.platform.cpu }}/bin" >> $GITHUB_PATH
+          echo "." >> $GITHUB_PATH
+
+      - name: Install nasm (Windows)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          choco install nasm --no-progress -y
+          "C:\Program Files\NASM" | Out-File -FilePath $Env:GITHUB_PATH -Encoding utf8 -Append
+
+      - name: Install deps
+        if: ${{ steps.deps-cache.outputs.cache-hit != 'true' }}
+        run: |
+          nimble install_pinned
+
       - name: Run tests
         id: run-tests
         run: |
@@ -108,6 +137,10 @@ jobs:
           gcc --version
 
           export NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }}"
+
+          if [[ '${{ runner.os }}' == 'Windows' ]]; then
+            export NIMFLAGS="${NIMFLAGS} --cc:clang"
+          fi
           nimble test
 
       - name: Fix xml newlines
--- a/.github/actions/install_nim/action.yml
+++ b/.github/actions/install_nim/action.yml
@@ -72,7 +72,7 @@ runs:
         steps.windows-dlls-cache.outputs.cache-hit != 'true' &&
         inputs.os == 'Windows'
       run: |
-        mkdir external
+        mkdir -p external
         curl -L "https://nim-lang.org/download/windeps.zip" -o external/windeps.zip
         7z x external/windeps.zip -oexternal/dlls
 
--- a/.github/workflows/daily_tests_no_flags.yml
+++ b/.github/workflows/daily_tests_no_flags.yml
@@ -38,11 +38,6 @@ jobs:
           path: nimbledeps
           key: nimbledeps-${{ hashFiles('.pinned') }}
 
-      - name: Install runtime deps (windows)
-        if: ${{ matrix.platform.os == 'windows'}}
-        run: |
-          pacman -S --noconfirm base-devel gcc nasm
-
       - name: Install deps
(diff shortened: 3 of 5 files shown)


tools/gen-deps.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
```
vladopajic, line 9: maybe have this in `/tools` - that folder is hosting different tools `gen-deps` is just a tool that is implemented as bash script.

vladopajic (approved):

nimble.lock
```diff
@@ -0,0 +1,272 @@
+{
```
richard-ramos: cc: @arnetheduck as i am not sure, but shouldn't we have a nimble.lock file only in the cbind folder?, since nim-libp2p is meant to be a library, thus we cannot lock the dependency versions except for ranges defined in the .nimble file ... or does this not have effects downstream?

config.nims
```diff
@@ -1,6 +1,4 @@
 # to allow locking
-if dirExists("nimbledeps/pkgs"):
```
richard-ramos, line 2: This was added due to having the pinned file. If it's being removed, does it mean we no longer are gonna use `nimble install_pinned` ?

Makefile
```diff
@@ -0,0 +1,29 @@
+.PHONY: all build deps cbind-deps cbind clean
```
richard-ramos: Since this makefile is associated to the cbindings, why does it live in the root folder?

.github/workflows/ci.yml
```diff
@@ -107,7 +107,8 @@ jobs:
           nimble --version
           gcc --version
 
-          export NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }}" 
+          export NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }}"
+          nimble setup
```
richard-ramos: Why is `nimble setup` required in all the .yml files? I find it curious that this is needed considering that we have a `nimble install_pinned` in the same files, and for ex in the Dockerfile for interop/transport it was removed.

tools/gen-deps.sh
```diff
@@ -0,0 +1,81 @@
...
+set -euo pipefail
```
richard-ramos, line 11: Missing comments with license

flake.nix
```diff
@@ -12,23 +12,153 @@
...
+                --threads:on \
+                --opt:size \
+                --noMain \
+                --mm:refc \
+                --header \
+                --undef:metrics \
+                --nimMainPrefix:libp2p \
+                --nimcache:$TMPDIR/nimcache \
+                cbind/libp2p.nim
+            '';
```
richard-ramos: cbind's .nimble contains a similar command to build the library. Isn't there a way to avoid duplication, maybe by putting the build command in a bash script and have the .nimble file call the script with exec?

flake.nix
```diff
@@ -12,23 +12,153 @@
...
+            ];
+
+            buildPhase = ''
+              # make sure nim writes to a writable dir
+              export HOME=$TMPDIR
+              export XDG_CACHE_HOME=$TMPDIR/.cache
+              export NIMBLE_DIR=$TMPDIR/.nimble
+
+              mkdir -p build
+              mkdir -p $TMPDIR/nimcache
```
richard-ramos: 56 and 60 have the same value, so maybe we can have a variable containing $TMPDIR/.cache and use that variable both in the export and in the mkdir

config.nims
gmelodie: There are two of those statements in that file, one is `if dirExists("nimbledeps/pkgs"):` (nim 1, which we no longer maintain), the other is `if dirExists("nimbledeps/pkgs2"):`, which is being kept.

> does it mean we no longer are gonna use nimble install_pinned ?

I think eventually we should move away from `.pinned` and use `nimble.lock`, but I think getting rid of `.pinned` is outside of the scope of this PR (which already has some pretty important changes)

Makefile
gmelodie: It's not *only* for the cbindings. I'm hoping that eventually we'll be able to do things like `make` to compile libp2p (no cbindings) and `make test` to run tests using nix

.github/workflows/ci.yml
gmelodie: My idea is to move away from `.pinned` in favor of `nimble lock`. In theory CI could be changed just like  `interop/transport` was, but I didn't wanna change that much in one PR. That Dockerfile does seem to not complain though, hinting that we could in theory move away from .pinned in the near future.

Makefile
richard-ramos: If this is the case, then it probably makes sense to split up the make files with the cbind logic being moved in cbind/ folder,  similar to how the code for building the cbindings was extracted out from libp2p.nimble into its own .nimble file in https://github.com/vacp2p/nim-libp2p/pull/1939

flake.nix
gmelodie: The problem here is that we need to add the lib paths when using `nix`, I think the script would be even more of a hack.

richard-ramos: If there is no alternative that would avoid duplication, then maybe add a comment in both the .nimble and flake.nix so in the future any dev making changes in these compile commands know that they need to update also the one on the .nimble

flake.nix
```diff
@@ -12,23 +12,154 @@
...
+          cbindPathArgs =
+            builtins.concatStringsSep " "
+              (map (p: "--path:${p}")
+                   (builtins.attrValues cbindDeps));
+          libExt =
+            if pkgs.stdenv.hostPlatform.isWindows then "dll"
+            else if pkgs.stdenv.hostPlatform.isDarwin then "dylib"
+            else "so";
+        in {
+          default = pkgs.stdenv.mkDerivation {
```
jakubgs: Our standard in repos is to put the derivation under `nix/default.nix` and just include it in the flake.

Makes the flake more manageable/readable.

Examples:
https://github.com/status-im/nimbus-eth2/blob/stable/nix/default.nix
https://github.com/logos-messaging/logos-messaging-nim/blob/master/nix/default.nix
https://github.com/logos-storage/logos-storage-nim/blob/master/nix/default.nix
https://github.com/vacp2p/zerokit/blob/master/nix/default.nix

flake.nix
```diff
@@ -12,23 +12,154 @@
...
+                --opt:speed \
+                -d:libp2p_autotls_support \
+                -d:libp2p_mix_experimental_exit_is_dest \
+                -d:libp2p_gossipsub_1_4 \
+                libp2p.nim
+            '';
+          };
+
+          cbind = pkgs.stdenv.mkDerivation {
+            pname = "nim-libp2p-cbind";
```
jakubgs: Same, should be in `nix/cbind.nix` probably.

flake.nix
```diff
@@ -12,23 +12,154 @@
...
+        in {
+          default = pkgs.stdenv.mkDerivation {
+            pname = "nim-libp2p";
+            version = "dev";
+
+            src = ./.;
+
+            nativeBuildInputs = [
+              pkgs.nim-2_2
+              pkgs.git
```
jakubgs: Why do we need actual Git here? It's not like we can clone/fetch anything.
In many repos we provide a fake git in case pakages try to asses their own version:
https://github.com/status-im/nimbus-eth2/blob/stable/nix/default.nix#L35

tools/gen-deps.sh
```diff
@@ -0,0 +1,90 @@
...
+  echo "error: lockfile not found: $LOCKFILE"
+  exit 1
+fi
+
+echo "[*] Generating $OUTFILE from $LOCKFILE"
+
+mkdir -p "$(dirname "$OUTFILE")"
+
+cat > "$OUTFILE" <<'EOF'
+# AUTOGENERATED from nimble.lock
```
jakubgs, line 56: Maybe also reference `tools/gen-deps.sh` was used for clarity?

nimble.lock
arnetheduck: Indeed - lock files are only for binaries - they're not meaningful for libraries and a lot of developers get confused about the presence of a lock file in a library, thinking that it will somehow get used by consumers of the library. It won't.

This is usually a source of confusion for lots of reasons, but you have to look at it from the point of view of a consumer of a library: they want the library to work with as many versions of the dependencies as possible because that's needed to combine libp2p with other libraries (like json-rpc etc that shares deps with libp2p).

You can imagine that if libp2p were to select a particular version of chronos and json_rpc another, the combination of the two would not work.

Therefore, I tend to find it directly counterproductive to use lock files to test libraries and I've never been a fan of `.pinned` in libp2p .. even in the cbindings case, if you want to combine two libraries that both offer cbindings, you want some flexibility in the underlying dependencies, so I'm not convinced it should be there.

What would be better instead would be "extremity testing" - ie what libp2p would ideally be testing is the _lowest version_ of every declared dependency in PR:s as well as the _highest version_ of everything because this is where most problems happen. For example, if we have `requires "chronos >= 4.0.0" in the nimble file, 4.0.0 and (at the time of writing) 4.2.0 should be tested regularly (perhaps via a nightly job for the "max" version, since it keeps changing).

This way, you make sure that your `requires` accurately reflect the minimum version of for example chronos that libp2p can work with - the lower this version is, the better, but it must also be _correctly set_, something that you don't test when you use `.pinned` / a lock file / a nix file. This is more in line with how users of libp2p will "use" the information in the nimble file.

Sometimes, it's useful to test more versions than the minimal version - nim is an example where we test multiple nim versions - this is because nim is an "important dependency" - you can apply that thinking to other deps as well - ie ostensibly we could have a matrix for multiple chronos versions too - there's a balance to be had here though.

arnetheduck: here's a pretty good read on the subject btw: https://research.swtch.com/vgo-mvs

PR #2030, opened 2026-01-15
test(kad): put value 2 - new tests + small fix

rlve:
Changes:
- add `PUT_VALUE` tests:
  - `"PUT_VALUE rejects mismatched Message.key and Record.key"`
  - `"PUT_VALUE with no record / no value - malformed message handling"`
  - `"PUT_VALUE response echoes request"`
  - `"PUT_VALUE stores binary data with null and high bytes"`
- fix: `handlePutValue` crashed with `ResultDefect` when receiving a malformed `PUT_VALUE` message containing a record with no value. The code used `record.value.get()` wrapped in a `try/except KeyError`, but `Opt.get()` raises `ResultDefect`, not `KeyError`, so the exception was never caught.
- improve `containsData` helper, add `checkpoints` with meaningful messages

### DIFF
--- a/tests/libp2p/kademlia/test_put.nim
+++ b/tests/libp2p/kademlia/test_put.nim
@@ -145,37 +145,6 @@ suite "KadDHT Put":
       kads[0].containsNoData(msgKey)
       kads[0].containsNoData(recordKey)
 
-  asyncTest "PUT_VALUE response echoes request":
-    let kads = await setupKadSwitches(2)
-    defer:
-      await stopNodes(kads)
-
-    connectNodes(kads[0], kads[1])
-
-    let key = kads[0].rtable.selfId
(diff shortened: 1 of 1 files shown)


libp2p/protocols/kademlia/put.nim
```diff
@@ -82,12 +82,11 @@ proc handlePutValue*(
     error "Record key is different than Message key", msg = msg, conn = conn
     return
 
-  let entryRecord =
-    try:
-      EntryRecord(value: record.value.get(), time: $times.now().utc)
-    except KeyError:
-      error "No key, value or timeReceived in buffer", msg = msg, conn = conn
-      return
+  let value = record.value.valueOr:
```
richard-ramos, line 85: Good catch

richard-ramos (approved):

PR #2036, opened 2026-01-19
fix(kad): get value should reject mismatched key

rlve:
Changes:
- The `getValue` accepted responses without validating that `Record.key` matched the requested key. Add validation in the `onReply` callback to reject records where `record.key != key`.
- Add test
- Add log scope
- Improve `containsNoData` helper


tests/libp2p/kademlia/mock_kademlia.nim
```diff
@@ -23,3 +23,30 @@ method findNode*(
   kad.findNodeCalls.add(target)
   ## Return only locally known peers
   return kad.rtable.findClosestPeerIds(target, kad.config.replication)
+
+method handleGetValue*(
+    kad: MockKadDHT, conn: Connection, msg: Message
+) {.async: (raises: [CancelledError]).} =
+  let wrongKey = kad.mismatchedRecordKey.valueOr:
+    # No malicious behavior - call base implementation
+    await get.handleGetValue(kad, conn, msg)
```
vladopajic: `get.handleGetValue` what is `get` here, because i don't see it?

tests/libp2p/kademlia/test_get.nim
```diff
@@ -247,3 +247,36 @@ suite "KadDHT Get":
...
+
+  asyncTest "GET_VALUE rejects record where Record.key does not match requested key":
+    # Create a normal node (victim) and a malicious node
+    # The malicious node returns GET_VALUE responses with Record.key != requested key
+    let (victimSwitch, victim) =
+      await setupKadSwitch(PermissiveValidator(), CandSelector())
+
+    let
+      requestedKey = victim.rtable.selfId
+      wrongKey = @[0xFF.byte, 0xFF, 0xFF, 0xFF]
```
vladopajic: nit: 
`@[0xFF.byte, 0xFF, 0xFF, 0xFF]` -> 
`@[1.byte, 1, 1]` or `"key".toBytes()` i guess later is more readable.

vladopajic (approved):

tests/libp2p/kademlia/mock_kademlia.nim
rlve: I'm explicitly calling `handleGetValue` from `libp2p/protocols/kademlia/get.nim` (so it's coming from import)

vladopajic: aha... could we do it like this: `await handleGetValue(KadDHT(kad), conn, msg)`?

PR #2038, opened 2026-01-20
test(kad): get value 2 - new tests

rlve:
Changes:
- add get value tests:
  - `"GET_VALUE rejects response without record"`
  - `"GET_VALUE rejects record without value"`
  - `"GET_VALUE succeeds with some peers returning mismatched keys"`
  - `"GET_VALUE rejects records that fail validation"`
  - `"GET_VALUE succeeds when some peers are offline"`
  - `"GET_VALUE fails when too many peers are offline"`
  - `"GET_VALUE retrieves binary data with null and high bytes"`

### DIFF
--- a/tests/libp2p/kademlia/test_add_provider.nim
+++ b/tests/libp2p/kademlia/test_add_provider.nim
@@ -9,7 +9,7 @@ import
 import ../../tools/[unittest]
 import ./[mock_kademlia, utils]
 
-suite "KadDHT - ProviderManager":
+suite "KadDHT - Add Provider":
   teardown:
     checkTrackers()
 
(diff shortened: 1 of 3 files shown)


tests/libp2p/kademlia/mock_kademlia.nim
```diff
@@ -11,42 +11,66 @@ import chronos, chronicles, std/times
...
-type MockKadDHT* = ref object of KadDHT
-  findNodeCalls*: seq[Key]
-  mismatchedRecordKey*: Opt[Key]
-    ## If set, handleGetValue returns Record.key with this value
+type
+  GetValueResponseType* = enum
+    Normal ## Use base implementation
+    MismatchedKey ## Return record with wrong Record.key
+    EmptyRecord ## Return response without record field
+    NoValue ## Return record without value field
```
gmelodie: Just have the mock type take a list of responses it should reply with, like we do with [autotls mockapi](https://github.com/vacp2p/nim-libp2p/blob/e2bfec2c626598f5d9788c1c110a4e35f9b5322d/libp2p/autotls/acme/mockapi.nim#L17)

rlve: updated

tests/libp2p/kademlia/test_get.nim
```diff
@@ -243,34 +243,258 @@ suite "KadDHT Get":
         valueLocal
 
   asyncTest "GET_VALUE rejects record where Record.key does not match requested key":
-    # Create a normal node (victim) and a malicious node
-    # The malicious node returns GET_VALUE responses with Record.key != requested key
-    let (victimSwitch, victim) =
+    let (victimSwitch, victimKad) =
```
vladopajic: nit: i would call these without "victim" prefix/suffix. it doesn't feel like they are victim as nothing is really taken away from them. they are regular/expected behavior nodes. "malicious" for other seems okay, but it could also be "misbehaving" - not that i have strong option for this case, like i do for first.

vladopajic: or maybe just use "mock" instead of  "malicious"...

tests/libp2p/kademlia/test_get.nim
```diff
@@ -243,34 +243,258 @@ suite "KadDHT Get":
...
+      victimKad.containsNoData(wrongKey)
+
+  asyncTest "GET_VALUE rejects response without record":
+    let (victimSwitch, victimKad) =
+      await setupKadSwitch(PermissiveValidator(), CandSelector())
+
+    # Get value response with empty record
+    let
+      key = victimKad.rtable.selfId
+      mockMessage = Message(
```
vladopajic: `mockMessage = Message(`
`getValueEmtpyRecord = Message(` consider naming according to purpose

tests/libp2p/kademlia/test_get.nim
```diff
@@ -243,34 +243,258 @@ suite "KadDHT Get":
...
-      victim.containsNoData(requestedKey)
-      victim.containsNoData(wrongKey)
+      victimKad.containsNoData(key)
+      victimKad.containsNoData(wrongKey)
+
+  asyncTest "GET_VALUE rejects response without record":
+    let (victimSwitch, victimKad) =
+      await setupKadSwitch(PermissiveValidator(), CandSelector())
+
+    # Get value response with empty record
```
vladopajic: consider having this comment here:

`record: Opt.none(protobuf.Record), # get value with empty`

tests/libp2p/kademlia/test_get.nim
```diff
@@ -243,34 +243,258 @@ suite "KadDHT Get":
...
+      key = kads[0].rtable.selfId
+      value = @[1.byte, 2, 3, 4, 5]
+      wrongKey = @[1.byte, 1, 1, 1]
+      mockMessage = Message(
+        msgType: MessageType.getValue,
+        key: key,
+        record: Opt.some(
+          protobuf.Record(
+            key: wrongKey,
+            value: Opt.some(@[1.byte, 2, 3, 4]),
```
vladopajic: value: Opt.some(@[1.byte, 2, 3, 4]),
 ->

value: value,

?

tests/libp2p/kademlia/test_get.nim
```diff
@@ -243,34 +243,258 @@ suite "KadDHT Get":
...
+    )
+    defer:
+      await stopNodes(kads)
+      await maliciousSwitch.stop()
+
+    connectNodes(kads[0], kads[1])
+    connectNodes(kads[0], kads[2])
+    connectNodes(kads[0], maliciousKad)
+
+    # Good nodes have valid records
```
vladopajic: Good -> Compliant

PR #2040, opened 2026-01-20
feat(gossipsub): partial message extension

vladopajic:
`ft-ts:p2p:ift:2026q1-nimlibp2p-gossipsub-extensions:partial-messages`

this is the majority of partial message feature, some things are left for other pr.
- integration with switch builder
- implement unit tests test
- implement `gossipThePartsMetadata`
- eager pushing 

just like any other extensions, default behavior is unconfigured = unsupported, so it can be safely used with 100% code backwards compatibility.

### DIFF
--- a/docs/protocols_mix.md
+++ b/docs/protocols_mix.md
@@ -75,7 +75,7 @@ For detailed information on implementing custom spam protection mechanisms, see 
 ## Using experimental `exit == destination`
 
 1. Compile with: `-d:libp2p_mix_experimental_exit_is_dest`
-2. In `toConnection` you can now specify the behavior the exit node will have:`
+2. In `toConnection` you can now specify the behavior the exit node will have:
 
 ```nim
 # Exit != destination (the default)
@@ -103,7 +103,7 @@ A complete working example demonstrating the Mix protocol with Ping can be found
 
 ## RFC and Further Reading
 
-For a detailed technical specification and discussion, please refer to the [Mix Protocol RFC](https://rfc.vac.dev/vac/raw/mix/).
+For a detailed technical specification and discussion, please refer to the [Mix Protocol RFC](https://lip.logos.co/ift-ts/raw/mix.html).
 
 ## Acknowledgments
 
--- a/docs/protocols_mix_spam_protection.md
+++ b/docs/protocols_mix_spam_protection.md
@@ -2,7 +2,7 @@
 
 # Spam Protection Interface
 
-A pluggable interface for integrating spam protection mechanisms into the Mix protocol, as specified in [section 9.6 of the MIX specification](https://github.com/vacp2p/rfc-index/blob/main/vac/raw/mix.md#96-spam-protection-interface).
+A pluggable interface for integrating spam protection mechanisms into the Mix protocol, as specified in [section 9.6 of the MIX specification](https://lip.logos.co/ift-ts/raw/mix.html#96-spam-protection-interface).
 
 ## Architecture
 
@@ -36,8 +36,8 @@ See [test_spam_protection_interface.nim](../tests/libp2p/mix/test_spam_protectio
 
 For detailed specification and security considerations, see:
 
-- [MIX Protocol Specification](https://github.com/vacp2p/rfc-index/blob/main/vac/raw/mix.md)
-- [Spam Protection Interface (Section 9.6)](https://github.com/vacp2p/rfc-index/blob/main/vac/raw/mix.md#96-spam-protection-interface)
+- [MIX Protocol Specification](https://lip.logos.co/ift-ts/raw/mix.html)
+- [Spam Protection Interface (Section 9.6)](https://lip.logos.co/ift-ts/raw/mix.html#96-spam-protection-interface)
 
 ---
 


libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,45 @@
+# Nim-LibP2P
+# Copyright (c) 2023-2025 Status Research & Development GmbH
```
tersec: Pretty sure that the only relevant copyright year here is 2026

libp2p/protocols/pubsub/gossipsub/extension_test.nim
```diff
@@ -0,0 +1,39 @@
+# Nim-LibP2P
+# Copyright (c) 2023-2025 Status Research & Development GmbH
```
tersec: Only relevant copyright year is 2026

libp2p/protocols/pubsub/gossipsub/extensions_types.nim
```diff
@@ -0,0 +1,39 @@
+# Nim-LibP2P
+# Copyright (c) 2023-2025 Status Research & Development GmbH
```
tersec: Only relevant copyright year is 2026

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -568,7 +580,13 @@ proc unsubscribeAll*(p: PubSub, topic: string) {.public, gcsafe.} =
...
     p.updateTopicMetrics(topic)
 
-proc subscribe*(p: PubSub, topic: string, handler: TopicHandler) {.public.} =
+proc subscribe*(
+    p: PubSub,
+    topic: string,
+    handler: TopicHandler,
+    requestsPartial: bool = false,
+    supportsSendingPartial: bool = false,
+) {.public.} =
```
vladopajic, line 600: this shouldn't count as major release since `subscribe` will be backwards compatabile?

tests/libp2p/pubsub/extensions/test_extension_test.nim
```diff
@@ -0,0 +1,27 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
```
gmelodie, line 1: This file's naming is confusing: `test_extension_test.nim`, also what's the difference between this and `test_extensions.nim` in the same dir? Could they not be merged together?

Ah if you're testing `TestExtension` I still think that should all be in the same file (if there aren't a lot of tests) or call this `test_test_extention.nim` (which is still confusing but better imo)

tests/libp2p/pubsub/extensions/test_extension_test.nim
```diff
@@ -0,0 +1,27 @@
...
+import
+  ../../../../libp2p/protocols/pubsub/
+    [gossipsub/extension_test, gossipsub/extensions_types]
+import ../../../tools/[unittest, crypto]
+import ../utils
+
+suite "GossipSub Extensions :: Test Extension":
+  let peerId = PeerId.random(rng).get()
+
+  test "basic test":
```
gmelodie, line 17: improve naming e.g.:
```suggestion
  test "TestExtension handles PeerExtensions":
```

tests/libp2p/pubsub/extensions/test_extensions.nim
```diff
@@ -0,0 +1,73 @@
...
+
+{.used.}
+
+import chronos, results, options, sequtils
+import ../../../../libp2p/peerid
+import ../../../../libp2p/protocols/pubsub/[gossipsub/extensions, rpc/messages]
+import ../../../tools/[unittest, crypto]
+import ../utils
+
+proc makeRPC(extensions: ControlExtensions = ControlExtensions()): RPCMsg =
```
gmelodie, line 12: It doesn't seem like you're passing `extensions` anywhere. Unless you're using it on your subsequent PR, consider making this a `const` instead.

tests/libp2p/pubsub/extensions/test_extensions.nim
```diff
@@ -0,0 +1,73 @@
...
+  RPCMsg(control: some(ControlMessage(extensions: some(extensions))))
+
+suite "GossipSub Extensions :: State":
+  let peerId = PeerId.random(rng).get()
+
+  test "default unconfigured state":
+    # test does not assert anything explictly, but it should not crash
+    # when unconfigured state is used
```
gmelodie: What about we make the opposite? (either replace or add another test): a test that crashes and we `expect` the error

tests/libp2p/pubsub/extensions/test_extensions.nim
```diff
@@ -0,0 +1,73 @@
...
+
+    # peer sends ControlExtensions for the first time
+    state.handleRPC(peerId, makeRPC())
+
+    # when peer sends ControlExtensions after that, missbehavior should be reported
+    for i in 1 ..< 5:
+      state.handleRPC(peerId, makeRPC())
+      check reportedPeers[] == repeat[PeerId](peerId, i)
+
+  test "state reports missbehaving - many peers":
```
gmelodie, line 43: Test and variable names could be more descriptive than `misbehave(ing)` because that is not a term defined in the spec, which can cause confusion. What exactly is a misbehaving peer?

tests/libp2p/pubsub/test_message.nim
```diff
@@ -150,19 +150,32 @@ suite "Message":
...
+        ],
       messages: @[msg, msg], # 16 * 2 = 32 bytes
       ping: @[1'u8, 2], # 2 bytes
       pong: @[3'u8, 4], # 2 bytes
       control: some(control), # 12 + 3 + 3 + 17 + 3 = 38 bytes
+      partialMessageExtension: some(partialMessageExtensionRPC), # 4 bytes
     )
 
-    check byteSize(rpcMsg) == 28 + 32 + 2 + 2 + 38 # Total: 102 bytes
+    check byteSize(rpcMsg) == 30 + 32 + 2 + 2 + 38 + 4 # Total: 108 bytes
```
gmelodie: Hum... this is ugly. Is it safe to assume none of these fields will ever change? Even so... Can we not do this in `check` batches i.e.
```suggestion
    var partialMessageExtensionRPC = PartialMessageExtensionRPC(
      topicID: "a", # 1 byte
      groupID: @[1'u8], # 1 byte
      partialMessage: @[1'u8], # 1 byte
      partsMetadata: @[1'u8], # 1 byte
    )

var rpcMsg = RPCMsg(
subscriptions:
@[
          SubOpts(
            subscribe: true,
            topic: "a".repeat(12),
            requestsPartial: some(true),
            supportsSendingPartial: some(true),
          ), # 1 + 12 + 1 + 1 = 15 bytes
          SubOpts(subscribe: false, topic: "b".repeat(14)), # 1 + 14 = 15 bytes
        ],
messages: @[msg, msg], # 16 * 2 = 32 bytes
ping: @[1'u8, 2], # 2 bytes
pong: @[3'u8, 4], # 2 bytes
control: some(control), # 12 + 3 + 3 + 17 + 3 = 38 bytes
      partialMessageExtension: some(partialMessageExtensionRPC), # 4 bytes
)
    check:
      partialMessageExtensionRPC.len() == 4
      rpcMsg.subscriptions[0].len() == 15
      rpcMsg.subscriptions[1].len() == 15
      rpcMsg.messages.len() == 32
      rpcMsg.ping.len() == 2
      rpcMsg.pong.len() == 2
      rpcMsg.control.len() == 38
      rpcMsg.partialMessageExtension.len() == 4
      ...
    check byteSize(rpcMsg) == rpcMsg.subscriptions.len() + rpcMsg.messages.len()  + ... 
```

or use `byteSize` if for some reason `len` is not the right function, but you get the idea

tests/libp2p/pubsub/extensions/test_extension_partial_message.nim
```diff
@@ -0,0 +1,54 @@
...
+import chronos, results
+import ../../../../libp2p/peerid
+import
+  ../../../../libp2p/protocols/pubsub/
+    [gossipsub/extension_partial_message, gossipsub/extensions_types, rpc/messages]
+import ../../../tools/[unittest]
+import ../utils
+
+suite "GossipSub Extensions :: Partial Message Extension":
+  test "basic test":
```
gmelodie, line 15: same as other comment: better naming

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,294 @@
...
+): GroupState =
+  let key = TopicGroupKey.new(topic, groupId)
+
+  if key notin ext.groupState:
+    ext.groupState[key] = GroupState()
+
+  try:
+    return ext.groupState[key]
+  except KeyError:
+    raiseAssert "checked with if"
```
gmelodie:
```suggestion
  return ext.groupState.mgetOrPut(key, GroupState())
```
or
```suggestion
  return ext.groupState.mgetOrDefault(key)
```

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,294 @@
...
+
+proc getPeerState(gs: GroupState, peerId: PeerId): PeerGroupState =
+  if peerId notin gs.peerState:
+    gs.peerState[peerId] = PeerGroupState()
+
+  try:
+    return gs.peerState[peerId]
+  except KeyError:
+    raiseAssert "checked with if"
```
gmelodie: Same as above

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,294 @@
...
+        # when peer requested partial, then, by spec, they must support it
+    )
+  else:
+    ext.peerSubs.del(key)
+
+proc handlePartialRPC(
+    ext: PartialMessageExtension, peerId: PeerId, rpc: PartialMessageExtensionRPC
+) =
+  let result = ext.validateRPC(rpc)
+  if result.isErr():
```
gmelodie: iirc we should not use `result`

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,294 @@
...
+    # implements logic for checking if peer supports this "partial message" extension.
+    # default implementation is set by extensions state.
+
+  ValidateRPCProc =
+    proc(rpc: PartialMessageExtensionRPC): Result[void, string] {.gcsafe, raises: [].}
+    # implements logic for performing sanity checks on PartialMessageExtensionRPC.
+    # when error is returned extension will not proces PartialMessageExtensionRPC.
+    # needs to be implemented by application.
+
+  OnIncomingRPCProc =
```
gmelodie, line 45: Why did you choose to make these fields instead of base `method`s with sane defaults?

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,294 @@
...
+  var groupState = ext.getGroupState(topic, msgGroupId)
+
+  groupState.heartbeatsTillEviction = ext.heartbeatsTillEviction
+
+  proc publishPartialToPeer(peer: PeerId, peerRequestsPartial: bool) {.raises: [].} =
+    var rpc = PartialMessageExtensionRPC(topicID: topic, groupID: msgGroupId)
+    var peerState = groupState.getPeerState(peer)
+    var hasChanges: bool = false
+
+    # if partsMetada was changed, rpc sets new metadata
```
gmelodie:
```suggestion
    # if partsMetadata was changed, rpc sets new metadata 
```

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,294 @@
...
+proc publishPartial*(
+    ext: PartialMessageExtension, topic: string, pm: PartialMessage
+): int {.raises: [].} =
+  let msgGroupId = pm.groupId()
+  let msgPartsMetadata = pm.partsMetadata()
+  var groupState = ext.getGroupState(topic, msgGroupId)
+
+  groupState.heartbeatsTillEviction = ext.heartbeatsTillEviction
+
+  proc publishPartialToPeer(peer: PeerId, peerRequestsPartial: bool) {.raises: [].} =
```
gmelodie: This is not very readable inside `publishPartial`, please yank it out if possible

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,294 @@
...
+    if peerState.sentPartsMetadata != msgPartsMetadata:
+      hasChanges = true
+      rpc.partsMetadata = msgPartsMetadata.data
+      peerState.sentPartsMetadata = msgPartsMetadata
+
+    # if peer has requested partial messages, attempt to fulfill any 
+    # parts that peer is missing.
+    if peerRequestsPartial:
+      let result = pm.materializeParts(peerState.partsMetadata)
+      if result.isErr():
```
gmelodie: Don't use `result`

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,294 @@
...
+    let peerSubOpt = ext.peerSubs.getOrDefault(PeerTopicKey.new(p, topic))
+
+    # publish parital message if ...
+    if peerSubOpt.requestsPartial:
+      # 1) peer requested partial for topic (peer wants to receive partial message)
+      publishPartialToPeer(p, true)
+      publishedToCount.inc
+    elif peerSubOpt.supportsSendingPartial and ext.isRequestPartialByNode(topic):
+      # 2) peer supports sending partial for topic and
+      # this node wants to receive partial message for this topic.
```
gmelodie: I don't quite understand the wording here. How is this different than case `1)`?

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,294 @@
...
+
+    # if partsMetada was changed, rpc sets new metadata 
+    if peerState.sentPartsMetadata != msgPartsMetadata:
+      hasChanges = true
+      rpc.partsMetadata = msgPartsMetadata.data
+      peerState.sentPartsMetadata = msgPartsMetadata
+
+    # if peer has requested partial messages, attempt to fulfill any 
+    # parts that peer is missing.
+    if peerRequestsPartial:
```
gmelodie: We only try to fulfill on publish? What about when we receive i.e. in `handlePartialRPC`?

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -508,13 +518,15 @@ proc updateTopicMetrics(p: PubSub, topic: string) =
...
+    p.topics.withValue(topic, topicData):
+      libp2p_pubsub_topic_handlers.set(
+        topicData[].handlers.len.int64, labelValues = [topic]
+      )
     do:
       libp2p_pubsub_topic_handlers.set(0, labelValues = [topic])
   else:
     var others: int64 = 0
-    for key, val in p.topics:
+    for key, value in p.topics:
```
gmelodie:
```suggestion
    for key, _ in p.topics:
```
or
```suggestion
    for key in p.topics.keys():
```

tests/libp2p/pubsub/extensions/test_extensions.nim
vladopajic: code is designed not to crash so we cant expect that

tests/libp2p/pubsub/extensions/test_extension_test.nim
vladopajic: test file names follow convention "test_" + filename that they are testing.

"test_exntesion_test" is a test file for "extension_test". this extension is experimental extensions from spec called "test extension" which is very unfortunately named in spec in the first place. so this is root cause of confusion.

in `gossipsub` folder, i have chosen the following convention for file names: "extension" + "name of extension" 
extension_test 
extension_partial_message
extension_foo
extension_bar
...

and there is "extensions" which imports all extensions, configures and holds state. it is main file for extensions.

tests/libp2p/pubsub/extensions/test_extension_test.nim
vladopajic: let's nitpick tests when tests are implemented

tests/libp2p/pubsub/extensions/test_extensions.nim
vladopajic: let's nitpick tests when tests are implemented

tests/libp2p/pubsub/extensions/test_extensions.nim
vladopajic: from code:
> onMissbehave: PeerCallback # callback when peer does not follow extensions protocol.

`onMissbehave` was added as general callback when peer is not behaving according to protocol. there is only one case for now (implemented):
> An Extensions control message MUST NOT be sent more than once

(unimplemented)
> If a peer supports any extension, the Extensions control message MUST be included in the first message on the stream.

tests/libp2p/pubsub/extensions/test_extension_partial_message.nim
vladopajic: let's nitpick tests when tests are implemented

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: i saw that kademlia and mix uses `results`, so i went with it

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: it accesses values from closure like `groupState`, and some other.. i`ll see

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: these are callbacks there are configured to implement behavior and to create loose coupling with other components.
 they provide behavior from gossipsub, extensions and there are user specific behavior callbacks. it was much cleaner to have callback for each case then have 1, 2 or 3 types for these. functions. then i would have to have base type and type for any specific implementation. to many types unnecessary.

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: https://github.com/vacp2p/nim-libp2p/pull/2040#discussion_r2746764929

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: 1) peer (other node) requests partial messages
2) this node has requested partial messages and peer (other node) supports sending it

in this code, i have used the following convention:
- "node" is always this code, or current code or current peer 
- "peer" is any other node connected to this node

node is also a peer from other peers perspective

tests/libp2p/pubsub/extensions/test_extensions.nim
vladopajic: anway i'll think about test case name later

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: then it gets to application scope via `onIncomingRPC` where user implements logic. 
but i'll have to read spec again.

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
gmelodie: I couldn't find any occurrences of the `result` keyword in `libp2p/protocols/kademlia` or `libp2p/protocols/mix` except for importing the `results` module, which is something else

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
gmelodie: i see, please reword for clarity

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
gmelodie: Yeah, maybe not lay this part on the user

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: reword what exactly?

gmelodie (approved):

libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -238,21 +275,33 @@ proc byteSize*(imreceivings: seq[ControlIMReceiving]): int =
...
+static:
+  expectedFields(
+    PartialMessageExtensionRPC,
+    @["topicID", "groupID", "partialMessage", "partsMetadata"],
+  )
+proc byteSize(pme: PartialMessageExtensionRPC): int =
+  pme.topicID.len + #
+  pme.groupID.len + #
+  pme.partialMessage.len + #
+  pme.partsMetadata.len
```
richard-ramos: Missing comments?
```suggestion
  pme.topicID.len +
  pme.groupID.len +
  pme.partialMessage.len +
  pme.partsMetadata.len
```

libp2p/protocols/pubsub/rpc/protobuf.nim
```diff
@@ -286,16 +304,19 @@ proc decodeExtensions*(pb: ProtoBuffer): ProtoResult[ControlExtensions] {.inline
   trace "decodeExtensions: decoding message"
   var control = ControlExtensions()
 
+  if ?pb.getField(10, control.partialMessageExtension):
```
richard-ramos, line 307: in line 316 the fieldId is a constant. Should we do the same here?

libp2p/protocols/pubsub/rpc/protobuf.nim
```diff
@@ -286,16 +304,19 @@ proc decodeExtensions*(pb: ProtoBuffer): ProtoResult[ControlExtensions] {.inline
...
+    trace "decodeExtensions: partialMessageExtension is missing"
+
   # Experimental extensions must use field numbers larger than 0x200000 to be
   # encoded with 4 bytes
 
-  var testExtension: uint64
-  if ?pb.getField(filedId_RPCMsg_TestExtension, testExtension):
-    trace "decodeExtensions: read testExtension", testExtension = testExtension
-    control.testExtension = some(bool(testExtension))
+  if ?pb.getField(filedId_RPCMsg_TestExtension, control.testExtension):
```
richard-ramos:
```suggestion
  if ?pb.getField(fieldId_RPCMsg_TestExtension, control.testExtension):
```

libp2p/protocols/pubsub/rpc/protobuf.nim
```diff
@@ -453,23 +488,70 @@ proc encodeRpcMsg*(msg: RPCMsg, anonymize: bool): seq[byte] =
...
+  ok(some(pme))
+
 proc decodeRpcMsg*(msg: seq[byte]): ProtoResult[RPCMsg] {.inline.} =
   trace "decodeRpcMsg: decoding message", payload = msg.shortLog()
   var pb = initProtoBuffer(msg)
   var rpcMsg = RPCMsg()
   assign(rpcMsg.messages, ?pb.decodeMessages())
   assign(rpcMsg.subscriptions, ?pb.decodeSubscriptions())
   assign(rpcMsg.control, ?pb.decodeControl())
+  rpcMsg.partialMessageExtension = ?pb.decodePartialMessageExtensionRPC(10)
```
richard-ramos, line 548: fieldId_RPCMsg_PartialMessages ?

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -992,9 +1004,24 @@ method publish*(
...
+  let count = g.extensionsState.publishPartial(topic, pm)
+
+  if g.knownTopics.contains(topic):
+    libp2p_pubsub_messages_published_partial.inc(count.int64, labelValues = [topic])
+  else:
+    libp2p_pubsub_messages_published_partial.inc(count.int64, labelValues = ["generic"])
+
+  trace "Published partial message to peers", count = count
+
+  return count
```
richard-ramos: As discussed in https://github.com/vacp2p/nim-libp2p/pull/1582#discussion_r2243028640 , let's not return a count of peers as this number lends itself to confusion and it is not that useful

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -1025,24 +1052,57 @@ proc maintainDirectPeers(g: GossipSub) {.async: (raises: [CancelledError]).} =
...
+      cfg.isRequestPartialByNode = proc(topic: string): bool {.gcsafe, raises: [].} =
+        g.topics.withValue(topic, topicData):
+          return topicData[].requestsPartial
+        return false
+
+    g.parameters.partialMessageExtensionConfig = some(cfg)
+
+  return ExtensionsState.new(
+    onMissbehaveExtensions, #
+    g.parameters.testExtensionConfig, #
```
richard-ramos: missing comments?
```suggestion
    onMissbehaveExtensions,
    g.parameters.testExtensionConfig,
```

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,281 @@
...
+    groupState: Table[TopicGroupKey, GroupState]
+    peerSubs: Table[PeerTopicKey, SubState]
+
+proc new(
+    T: typedesc[PeerTopicKey], peerId: PeerId, topic: string
+): PeerTopicKey {.inline.} =
+  $peerId & keyDelimiter & topic
+
+proc hasPeer(key: PeerTopicKey, peerId: PeerId): bool =
+  return ($peerId & keyDelimiter) in cast[string](key)
```
richard-ramos: PeerTopicKey is already an alias of string. Is the cast[string] necessary? feels redundant. Same on line 98

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
richard-ramos: +1. Doing this probably translates into a warning due to overriding nim's `result`.

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,281 @@
...
+
+proc gossipThePartsMetadata(ext: PartialMessageExtension) =
+  # TODO: `partsMetadata` can be used during heartbeat gossip to inform non-mesh topic
+  # peers about parts this node has.
+  discard
+
+proc peerRequestsPartial*(
+    ext: PartialMessageExtension, peerId: PeerId, topic: string
+): bool =
+  let peerSubOpt = ext.peerSubs.getOrDefault(PeerTopicKey.new(peerId, topic))
```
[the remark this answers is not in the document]
richard-ramos: :-1:

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,281 @@
...
+      continue
+
+    let peerSubOpt = ext.peerSubs.getOrDefault(PeerTopicKey.new(p, topic))
+
+    # publish parital message if ...
+    if peerSubOpt.requestsPartial:
+      # 1) peer requested partial for topic (peer wants to receive partial message)
+      publishPartialToPeer(p, true)
+      publishedToCount.inc
+    elif peerSubOpt.supportsSendingPartial and ext.isRequestPartialByNode(topic):
```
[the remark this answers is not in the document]
richard-ramos: :-1:

libp2p/protocols/pubsub/gossipsub/partial_message.nim
```diff
@@ -0,0 +1,78 @@
...
+    ## - A partially available message
+    ## - No parts at all
+    ##
+    ## Implementations define how messages are partitioned into parts, how parts
+    ## are encoded, and how availability and requests for parts are represented.
+
+converter toPartsMetadata*(data: seq[byte]): PartsMetadata =
+  PartsMetadata(data: data)
+
+method merge*(a, b: PartsMetadata): PartsMetadata {.base, gcsafe, raises: [].} =
```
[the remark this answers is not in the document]
richard-ramos: Interesting!

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,281 @@
...
+  for key, group in ext.groupState:
+    group.peerState.del(peerId)
+
+  # remove peer subscription options from _peerSubs_
+  var toRemove: seq[PeerTopicKey] = @[]
+  for key, _ in ext.peerSubs:
+    if key.hasPeer(peerId):
+      toRemove.add(key)
+  for key in toRemove:
+    ext.groupState.del(key)
```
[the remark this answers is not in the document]
richard-ramos: This looks like a valid suggestion

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -1025,24 +1052,57 @@ proc maintainDirectPeers(g: GossipSub) {.async: (raises: [CancelledError]).} =
...
+      cfg.sendRPC = proc(
+          peerId: PeerId, rpc: PartialMessageExtensionRPC
+      ) {.gcsafe, raises: [].} =
+        g.peers.withValue(peerId, peer):
+          g.send(peer[], RPCMsg(partialMessageExtension: some(rpc)), false)
 
-  return ExtensionsState.new(onMissbehaveExtensions, some(testExtensionConfig))
+    if cfg.publishToPeers.isNil:
+      cfg.publishToPeers = proc(topic: string): seq[PeerId] {.gcsafe, raises: [].} =
+        let peers = g.makePeersForPublishDefault(topic)
```
richard-ramos, line 1088: > Fanout and Gossip messages
> Fanout and Gossip messages by definition come from non-mesh peers. Partial messages, without eager data, require an exchange of bitmaps before parts are transferred. In order for fanout and gossip messages to be useful, the Gossipsub implementation SHOULD include the peers that sent them in the list of peers to publish partial messages to. This allows the application to simply call PublishPartial to respond to both mesh and non mesh peers.

Reading the code for `makePeersForPublishDefault`, iiuc, it will return direct peers + mesh peers and will only use fanout if floodPublish is true 🤔 , thus not including all non-mesh peers.  Wouldnt we be diverging from the spec here?

libp2p/protocols/pubsub/rpc/protobuf.nim
richard-ramos, line 145: requestPartial and supportsSendingPartial are missing

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -0,0 +1,281 @@
...
+    # if peer has requested partial messages, attempt to fulfill any 
+    # parts that peer is missing.
+    if peerRequestsPartial:
+      let result = pm.materializeParts(peerState.partsMetadata)
+      if result.isErr():
+        # there might be error with last PartsMetadata so it is discarded,
+        # to avoid any error with future messages.
+        peerState.partsMetadata = newSeq[byte](0)
+      else:
+        peerState.partsMetadata = merge(peerState.partsMetadata, msgPartsMetadata)
```
richard-ramos: In the specs it says:
```
Upon receiving a partsMetadata a node SHOULD respond with only parts the peer wants.
A later partsMetadata replaces a prior one.
```
I could not find the implementation of `merge` but seems to me that the name of the proc might not be appropiate as we wouldnt merge but replace a partsMetadata if new one is received

libp2p/protocols/pubsub/gossipsub/behavior.nim
richard-ramos: > When Gossiping, a node that supports partial messages SHOULD NOT send an IHAVE to a peer that requested partial messages. The node SHOULD send a partial message instead.

The peers from getGossipPeersshould be filtered so we dont send an IHAVE to those that have requested partial messages

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -568,7 +580,13 @@ proc unsubscribeAll*(p: PubSub, topic: string) {.public, gcsafe.} =
 
     p.updateTopicMetrics(topic)
 
-proc subscribe*(p: PubSub, topic: string, handler: TopicHandler) {.public.} =
+proc subscribe*(
+    p: PubSub,
+    topic: string,
+    handler: TopicHandler,
+    requestsPartial: bool = false,
+    supportsSendingPartial: bool = false,
```
richard-ramos, line 599: > If a node requests partial messages, it MUST support sending partial messages.

Wouldnt passing a `requestPartial = true, supportsSendingPartial = false` be invalid?

libp2p/protocols/pubsub/rpc/messages.nim
vladopajic: no :) i did that to force formatting; don't like the way nph handles this, but i could remove it

libp2p/protocols/pubsub/pubsub.nim
vladopajic: no, it wouldn't be, because:

requestPartial = requestPartial
supportsSendingPartial = supportsSendingPartial or requestPartial

is the logic in the code

libp2p/protocols/pubsub/rpc/protobuf.nim
vladopajic: yeah i am not sure... i would prefer to avoid these consts here. seems annoying to be managing. every rule seems to have an exception, this is one exception against consts imo. 

- line gets  long length, from 2 chars `10` to  >20 chats `fieldId_RPCMsg_TestExtension`...
- the logic of these files is straightforward encoding/decoding, naming fields does not help in understanding of the code
- it's easier for me, to matching if the same filed id is used in encoding/decoding just by looking to number rather than reading text (totally personal) 


thing about that and let me know

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: oh, that i was thinking about "results" like anything from that import... yes i get it now

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -582,12 +600,16 @@ proc subscribe*(p: PubSub, topic: string, handler: TopicHandler) {.public.} =
     warn "Trying to subscribe to a topic not passing validation!", topic
     return
 
-  p.topics.withValue(topic, handlers):
+  p.topics.withValue(topic, topicData):
     # Already subscribed, just adding another handler
-    handlers[].add(handler)
+    topicData[].handlers.add(handler)
```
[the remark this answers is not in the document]
vladopajic: not sure....

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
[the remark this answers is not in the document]
vladopajic: yes

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: not necessary for 93, it was copy and paste. on line 98 it is necessary as GroupId is `seq[byte]`

libp2p/protocols/pubsub/rpc/protobuf.nim
vladopajic: https://github.com/vacp2p/nim-libp2p/pull/2040#discussion_r2747912067 
let's discuss there

libp2p/protocols/pubsub/rpc/messages.nim
vladopajic: removed...

libp2p/protocols/pubsub/gossipsub.nim
vladopajic: i'll read again spec. my intention was to do what is logical, and it was to call an existing method that that gets peers. it seemed alright since if it is not floodPublish then fanout peers are still used

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: a replacing wouldn't make sense at this place at all, because when this node publishes we can't replace because user might want parts 1,2,3 and we are publishing only part 3 here.

replace could make sense in `handlePartialRPC`. but even here merge is better:

- peer requested 1, 2, 3
- nodes receive request 1, 2, 3
- node sends 2, 3
- peer requested 1, 2, 3, 4  # note peer sends request (updated request adding 4 part) before node's response 2, 3 is received by peer. so new request by peer will completely discard what we sent. but since there is a merge it  is already doing this in highlighted code so we can reuse it here as well.

anyway i'll think about it still.

libp2p/protocols/pubsub/pubsub.nim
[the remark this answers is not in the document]
vladopajic: requestsPartial and supportsSendingPartial must be ignored because node has already sent subscription to peers.

libp2p/protocols/pubsub/rpc/protobuf.nim
richard-ramos: IMO it is okay to either using or not using constants as long as we are consistent (i.e. if we use numbers, then get rid of fieldId_RPCMsg_TestExtension). If it helps to take a decision, this protobuf encoding/decoding might change in the future if we use nim-protobuf-serialization, thus not having to worry about constants at all

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: i have replaced `merge` with `replace` only inside `handlePartialPRC`, better to go with per spec information.

---

> I could not find the implementation of merge

```
On the other side, when a node responds to a peer's request for parts, it should update the peer's 
partsMetadata with the **union** of its partsMetadata and the peers partsMetadata.
```

renaming `merge` -> `union`

libp2p/protocols/pubsub/pubsub.nim
richard-ramos: I see, in this case, shouldnt this be an enum?

PR #2048, opened 2026-01-23
test(kad): find & lookup 3

rlve:
Changes:
- add tests:
  - `"Find node merges results from parallel queries"`
  - `"Find node returns the actual k closest"`
  - `"Lookup stops when k closest nodes have responded successfully"`
  - `"Lookup doesn't stop when k successes but closer peer not responded"`
  - `"selectCloserPeers excludes peers that exhausted retries"`
- add log scope for `put`
- cleanup

### DIFF
--- a/tests/libp2p/kademlia/mock_kademlia.nim
+++ b/tests/libp2p/kademlia/mock_kademlia.nim
@@ -1,48 +1,30 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 # Copyright (c) Status Research & Development GmbH 
 
-import chronos, chronicles, std/times
+import chronos, chronicles
 import ../../../libp2p/protocols/kademlia/[types, routingtable, protobuf, get, provider]
 import ../../../libp2p/[peerid, stream/connection]
 
 type MockKadDHT* = ref object of KadDHT
   findNodeCalls*: seq[Key]
-  mismatchedRecordKey*: Opt[Key]
-    ## If set, handleGetValue returns Record.key with this value
+  getValueResponse*: Opt[Message]
   handleAddProviderMessage*: Opt[Message]
 
 method findNode*(
     kad: MockKadDHT, target: Key
 ): Future[seq[PeerId]] {.async: (raises: [CancelledError]).} =
-  # Track all findNode calls
   kad.findNodeCalls.add(target)
-  ## Return only locally known peers
   return kad.rtable.findClosestPeerIds(target, kad.config.replication)
 
 method handleGetValue*(
     kad: MockKadDHT, conn: Connection, msg: Message
 ) {.async: (raises: [CancelledError]).} =
-  let wrongKey = kad.mismatchedRecordKey.valueOr:
-    # No malicious behavior - call base implementation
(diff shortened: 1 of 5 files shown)


tests/libp2p/kademlia/test_find.nim
```diff
@@ -404,3 +477,88 @@ suite "KadDHT Find":
...
+    let targetKey = randomPeerId().toKey()
+    var state = LookupState.init(kads[0], targetKey)
+
+    # Add peers
+    var peers: seq[PeerId]
+    for i in 0 ..< 4:
+      peers.add(randomPeerId())
+      state.shortlist[peers[i]] =
+        xorDistance(peers[i], targetKey, kads[0].rtable.config.hasher)
+    let sorted = peers.sortPeers(targetKey, kads[0].rtable.config.hasher)
```
vladopajic: nit:  this would be better with utilities
let peers = generateRandomPeers(5)
addPeersToShortlist(state, peers)

tests/libp2p/kademlia/test_find.nim
```diff
@@ -84,6 +84,79 @@ suite "KadDHT Find":
...
+    defer:
+      await stopNodes(kads)
+
+    # Set replication factor to 3
+    let k = 3
+    kads[0].config.replication = k
+
+    # Node[0] is directly connected to all other nodes
+    for i in 1 ..< nodeCount:
+      connectNodes(kads[0], kads[i])
```
vladopajic: i think we called this `connectNodesHub`

tests/libp2p/kademlia/test_find.nim
```diff
@@ -404,3 +477,88 @@ suite "KadDHT Find":
...
     check allPeers ==
       @[peer1, peer2, peer3].sortPeers(targetKey, kads[0].rtable.config.hasher)
+
+  asyncTest "Lookup stops when k closest nodes have responded successfully":
+    let kads = await setupKadSwitches(1)
+    defer:
+      await stopNodes(kads)
+
+    let targetKey = randomPeerId().toKey()
+    var state = LookupState.init(kads[0], targetKey)
```
vladopajic, line 486: it seems like bad design when kad and switch needs to be started to test `LookupState` and kad is not really used.

does it make sense to have tests related to `LookupState` in separate file?

tests/libp2p/kademlia/test_find.nim
rlve: Changed in https://github.com/vacp2p/nim-libp2p/pull/2048/changes/8f78ff542110892a31e5678f3d05a3206267b695

tests/libp2p/kademlia/test_find.nim
rlve: Changed in https://github.com/vacp2p/nim-libp2p/pull/2048/changes/8f78ff542110892a31e5678f3d05a3206267b695

tests/libp2p/kademlia/test_find.nim
rlve: This has been addressed in https://github.com/vacp2p/nim-libp2p/pull/2051 due to the size of changes

PR #2049, opened 2026-01-25
fix(gossipsub): race condition issue causing nil pointer exception when sendConn is not set

chaitanyaprem:
**Bug** 

While testing with mix, i had noticed a crash happening on latest master in gossipsub randomly. 

Backtrace is the following, looks like sendCon is nil and we are trying to access something from it. This looks like a race condition which is why it was not happening consistently. 
I have tested this by disabling mix and just running few logos-messaging nodes trying to connect with each other and observed the crash happens.  

```
FAT 2026-01-25 09:35:52.263+05:30 Shutting down after receiving SIGSEGV topics="wakunode main" tid=386491520 file=wakunode2.nim:90 stacktrace=
Traceback (most recent call last)
~/Code/logos-messaging-nim/apps/wakunode2/wakunode2.nim(102) wakunode2
~/Code/logos-messaging-nim/vendor/nim-chronos/chronos/internal/asyncengine.nim(1210) runForever
~/Code/logos-messaging-nim/vendor/nim-chronos/chronos/internal/asyncengine.nim(150) poll
~/Code/logos-messaging-nim/vendor/nim-libp2p/libp2p/protocols/pubsub/gossipsub.nim(291) addPeerAfterConnected
~/Code/logos-messaging-nim/apps/wakunode2/wakunode2.nim(93) handleSigsegv
```

**Fix**
	•	The reordering ensures sendConn is initially valid when connectedFut completes
	•	But the connection can fail and closeSendConn() can run concurrently, setting sendConn back to nil
	•	The nil check handles this scenario gracefully - if the connection failed, we simply skip sending the extensions control message

Not sure if this affects any other flow though but this fix seem to work in my tests.

### DIFF
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -19,7 +19,7 @@ type LookupState* = object
   target*: Key
   shortlist*: Table[PeerId, XorDistance]
   responded*: Table[PeerId, RespondedStatus]
-  attempts: Table[PeerId, int]
+  attempts*: Table[PeerId, int]
 
 type DispatchProc* = proc(kad: KadDHT, peer: PeerId, target: Key): Future[Opt[Message]] {.
   async: (raises: [CancelledError, DialFailedError, LPStreamError]), gcsafe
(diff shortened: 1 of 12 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -282,7 +282,7 @@ method onNewPeer*(g: GossipSub, peer: PubSubPeer) =
       proc addPeerAfterConnected(): Future[void] {.async.} =
         await peer.connectedFut
 
-        if gossipExtensionsSupported(peer.sendConn.protocol):
+        if peer.sendConn != nil and gossipExtensionsSupported(peer.sendConn.protocol):
```
vladopajic, line 285: does it work without this?

chaitanyaprem: I too thought it would but it hit another scenario as mentioned above and crashed.
But now that i think about it, this nil check also may not gaurantee and prevent actual race condition right because what if the check passed and then sendConn is set to nil.

`• the connection can fail and `closeSendConn`() can run concurrently, setting `sendConn` back to nil
`

Issue #2044, opened 2026-01-21
docs: move all .md pages to docs folder


vladopajic:
https://github.com/vacp2p/nim-libp2p/pull/2037#discussion_r2712507553

- move all .md pages to docs folder
- make links from docs to new pages (so it is navigable just by clicks).

PR #2052, opened 2026-01-27
chore(docs): move documentation files to docs folder

chaitanyaprem:
Per #2044 moved docs for now. 

We can keep adding docs for other protocols and maintain them here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,12 +44,8 @@ You'll find the nim-libp2p documentation [here](https://vacp2p.github.io/nim-lib
 
 ## Development
 
-Read the [development guide](docs/development.md) to get started with the project and testing.
-
-Additional resources:
-
-- [Compile time flags](docs/compile_time_flags.md) - all available compile-time options
-- [Common hurdles](docs/common_hurdles.md) - frequently encountered issues and solutions
+See the [development guide](docs/development.md) to get started with the project and testing.
+For more details, refer to the [documentation](docs/README.md).
 
 ## Contributors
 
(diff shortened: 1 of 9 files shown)


docs/README.md
```diff
@@ -0,0 +1,19 @@
+# Documentation Index
+
+Welcome to the nim-libp2p documentation. Use the links below to navigate:
+
+## Guides
+
+- [Development Guide](development.md)
+- [Contributing](contributing.md)
+- [Compile Time Flags](compile_time_flags.md)
+- [Common Hurdles](common_hurdles.md)
```
vladopajic: would it make sense to have `[← Docs Index](../../README.md)` on all of these pages as well ?

README.md
vladopajic: this page kind of have some form of documentation index here... it would be too much to keep documentation index at two places (here and in docs/README.md). 

how about if we just add one item here to the end of the list:
`- [Documentation index](docs/README.md] - start reading the documentation from the beginning if you can’t find what you’re looking for`  (note feel free to suggest cta text here, it doesn't have to be this)

or what if we remove `[Compile time flags]` and `[Common hurdles]` with just `[Documentation index]`?

docs/README.md
vladopajic: - [Common Hurdles]  (i would put common hurdles as early on since many people can have problems even before they start developing, so let them have idea where they can find help)


---


- [Development Guide]
- [Common Hurdles] 

or 

- [Common Hurdles] 
- [Development Guide]

docs/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Protocols
+
+- [Mix Protocol](protocols/mix/README.md)
+  - [Spam Protection](protocols/mix/spam_protection.md)
+
+## Interop
+
+- [Hole Punching Interop](interop/hole-punching/README.md)
```
vladopajic: let's avoid any section (headers) here, we can have everthing under single bullet list

- [Common Hurdles]
- [Development Guide]
- ...
- Protocols
  - [Mix Protocol]
    - [Spam Protection]
- Interop (-erability)
  - [Hole Punching]

docs/protocols/mix/spam_protection.md
vladopajic: does link need a update?

docs/README.md
chaitanyaprem: Good idea to create a backward reference.

README.md
chaitanyaprem: Makes sense to have it inside and just refer it here.

Issue #2053, opened 2026-01-27
kad dht: RespondedStatus.Failed is never reached in iterative lookup


rlve:
In `libp2p/protocols/kademlia/find.nim`, the `RespondedStatus.Failed` enum variant is never assigned due to a logic error with chronos future state checks.

https://github.com/vacp2p/nim-libp2p/blob/27f6ec9d18186a6eb3ddd9ec400a341cd0d26dbc/libp2p/protocols/kademlia/find.nim#L183-L187

The check `fut.failed()` on line 187 can never return `true` because:

1. Chronos `FutureState` is an enum: `Pending, Completed, Cancelled, Failed`
2. `completed()` returns `true` only when `state == FutureState.Completed`
3. `failed()` returns `true` only when `state == FutureState.Failed`
4. These are mutually exclusive — a future cannot be both `Completed` and `Failed`
5. Line 184 skips all futures where `completed() == false`
6. Therefore, any future reaching line 187 has `completed() == true`, which means `failed() == false`

**Observed behavior:** All peer failures (dial errors, stream errors, timeouts) result in retries via the `attempts > retries` mechanism. The `responded` table only ever contains `Success` entries.

PR #2054, opened 2026-01-27
fix(kad): use `finished()` instead of `completed()` to track failed peers

rlve:
Resolves: #2053

Changes:
- use `finished()` instead of `completed()` which returns `true` for both completed and failed futures
- add test for retry path

### DIFF
--- a/tests/libp2p/kademlia/test_find.nim
+++ b/tests/libp2p/kademlia/test_find.nim
@@ -318,9 +318,9 @@ suite "KadDHT Find":
 
     check mockKad.handleFindNodeCalls == 0
 
-    let result = await kad.findNode(mockKad.rtable.selfId)
+    let peerIds = await kad.findNode(mockKad.rtable.selfId)
 
     # Lookup terminates gracefully after retry exhaustion
     check:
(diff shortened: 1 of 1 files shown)


libp2p/protocols/kademlia/find.nim
```diff
@@ -181,7 +181,7 @@ proc iterativeLookup*(
       completedRPCBatch = await rpcBatch.collectCompleted(kad.config.timeout)
 
     for (fut, peerId) in zip(rpcBatch, toQuery):
-      if not fut.completed():
+      if not fut.finished():
```
vladopajic, line 184: i remember being confused about this [here](https://github.com/vacp2p/nim-libp2p/pull/1980#pullrequestreview-3589097851).

could you check if there are other places in `kad` that are affected by similar issue? also could you check `collectCompleted` - not sure why i mentioned it there, but could you please double-check it as well?

rlve: I had a look and I didn't find other places with similar checks that could be affected by similar issue. `collectCompleted` also seems to be used correctly, to extract only successful results for processing.

PR #2055, opened 2026-01-27
feat(kad): service discovery

SionoiS:
Implementation of [logos-service-discovery](https://github.com/vacp2p/rfc-index/blob/master/docs/ift-ts/raw/logos-capability-discovery.md)

Note to reviewers:
- Libp2p folks should focus on architecture and code organisation.
- Research folks should focus on spec compliance and sound logic.
:pray:


libp2p/protocols/capability_discovery/iptree.nim
```diff
@@ -0,0 +1,87 @@
...
+  if ip.family != IpAddressFamily.IPv4:
+    return err(Ipv4Required)
+
+  var v = ipTree.root
+  let bytes = ip.address_v4
+
+  for i in 0 ..< 4:
+    let b = bytes[i]
+    for bit in countdown(7, 0):
+      v.counter -= 1
```
tinniaru3005: `removeIp` walks the tree with `v = v.left/right` without nil checks. If remove is called for an IP that was never inserted (or double-removed due to pruning/updates), `v` can become nil and crash. Also counters can go negative.

libp2p/protocols/capability_discovery/types.nim
```diff
@@ -0,0 +1,136 @@
...
+  Default_F_lookup* = 30 # stop when we have this many advertisers
+  Default_F_return* = 10 # max ads returned by a single registrar
+  Default_E* = 900.0 # advertisement expiry
+  Default_C* = 1_000.0 # advertisement cache capacity
+  Default_P_occ* = 10.0 # occupancy exponent
+  Default_G* = 1e-7 # safety parameter
+  Default_Delta* = chronos.seconds(1) # registration window
+  Default_M_buckets* = 16 # number of buckets in AdvT/DiscT/RegT
+
+  DefaultSelfSPRRereshTime* = chronos.minutes(10)
```
tinniaru3005: Is this a typo (`Refresh` instead of `Reresh`) or intentional?

libp2p/protocols/capability_discovery/advertiser.nim
```diff
@@ -0,0 +1,274 @@
...
+    return
+
+  # Update advTable with closer peers if service still exists
+  if serviceId in disco.advertiser.advTable:
+    for peerId in closerPeers:
+      try:
+        discard disco.advertiser.advTable[serviceId].insert(peerId.toKey())
+      except KeyError:
+        # Service was removed, ignore
+        break
```
tinniaru3005: Using `break` inside a for loop over `closerPeers` means if any peer insertion fails, we stop processing all remaining peers. This is likely incorrect. We can change `break` to `continue` to skip only the failed peer.

libp2p/protocols/capability_discovery/registrar.nim
```diff
@@ -0,0 +1,431 @@
...
+  ## Falls back to main DHT if table is empty or not found.
+
+  var closerPeerKeys: seq[Key] = @[]
+  block thisBlock:
+    if disco.serviceRoutingTables.hasService(serviceId):
+      let regTable = disco.serviceRoutingTables.getTable(serviceId).valueOr:
+        break thisBlock
+
+      for bucket in regTable.buckets:
+        let closerPeer = bucket.randomPeerInBucket(disco.rng)
```
richard-ramos: randomPeerInBucket expects at least one peer in the bucket, so probably makes sense to check the bucket.peers.len to be != 0

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -0,0 +1,431 @@
...
+  for existingAd in ads:
+    if existingAd.data.peerId == ad.data.peerId:
+      isDuplicate = true
+      # Update timestamp for the existing ad
+      let existingKey = existingAd.toAdvertisementKey()
+      disco.registrar.cacheTimestamps[existingKey] = now
+      break
+
+  if not isDuplicate:
+    ads.add(ad)
```
richard-ramos: Should this `ad` be stored back into the registrar cache?

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -0,0 +1,431 @@
...
+  let t_remaining = disco.processRetryTicket(regMsg, ad, t_wait, now)
+
+  if t_remaining <= 0:
+    cd_register_requests.inc(labelValues = [$kademlia_protobuf.Confirmed])
+    await disco.acceptAdvertisement(serviceId, ad, now, closerPeers, conn)
+    return
+
+  var init = now
+  if regMsg.ticket.isSome:
+    init = regMsg.ticket.get().tInit
```
richard-ramos, line 626: Do you think it would be necessary to validate the value of tInit? What would happen if the value of tInit here is really old?

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -0,0 +1,431 @@
...
+  ## Send Wait response with ticket or Rejected response
+
+  var ticket = ticket
+
+  ticket.tWaitFor = min(disco.discoConf.advertExpiry.uint32, t_remaining.uint32)
+  ticket.tMod = getTime().toUnix().uint64
+
+  ticket.sign(disco.switch.peerInfo.privateKey).isOkOr:
+    error "failed to sign ticket", error
+    await conn.sendRegisterReject(closerPeers)
```
richard-ramos, line 423:
```suggestion
    await conn.sendRegisterReject(closerPeers)
    return
```

Otherwise we'd send both a reject and a wait

libp2p/protocols/capability_discovery/advertiser.nim
```diff
@@ -0,0 +1,241 @@
...
+  disco.updateAdvertiserMetrics()
+
+proc runAdvertiseLoop*(disco: KademliaDiscovery) {.async: (raises: [CancelledError]).} =
+  ## Loop through all pre-scheduled actions and execute them at the correct time.
+
+  while true:
+    let queue = disco.advertiser.actionQueue
+
+    if queue.len == 0:
+      # The loop is restarted lazily by processAction()
```
richard-ramos: I think there's a problem with this and the condition on line 42:
```nim
  if disco.advertiseLoop.isNil and disco.advertiser.actionQueue.len > 0:
    disco.advertiseLoop = disco.runAdvertiseLoop()
```

If you return here,  `disco.advertiseLoop` will continue having a value, so the condition on line 42 of disco.advertiseLoop will never be `nil`,  so in practice `disco.runAdvertiseLoop` will only be executed once.

Imo the condition on line 42 for nil checking should be:

```nim
if ((disco.advertiseLoop.isNil or disco.advertiseLoop.finished()) and disco.advertiser.actionQueue.len > 0:
```

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -0,0 +1,241 @@
...
+    disco.scheduleAction(serviceId, registrar, bucketIdx, nextTime, Opt.some(newTicket))
+  of protobuf.RegistrationStatus.Rejected:
+    # Don't reschedule - this registrar rejected us
+    return
+
+  disco.processAction()
+
+proc addProvidedService*(disco: KademliaDiscovery, serviceId: ServiceId) =
+  ## Include this service in the set of services this node provides.
+
```
richard-ramos, line 173: Should the service be also added to `disco.services` so calls to `record` also return the service ?

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -0,0 +1,431 @@
...
+
+    let ipKey = $ip
+    let elapsedIp = float64(now - registrar.timestampIp.getOrDefault(ipKey, 0))
+    let boundIpVal = registrar.boundIp.getOrDefault(ipKey, 0.0)
+    if w > (boundIpVal - elapsedIp):
+      registrar.boundIp[ipKey] = w + float64(now)
+      registrar.timestampIp[ipKey] = now
+
+proc getRegistrarCloserPeers*(
+    disco: KademliaDiscovery, serviceId: ServiceId, count: int
```
richard-ramos, line 184: `count` parameter is not used at all?

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -21,6 +22,8 @@ type
     getProviders = 3
     findNode = 4
     ping = 5 # Deprecated
+    register = 6 # REGISTER for Capability Discovery
+    getAds = 7 # GET_ADS for Capability Discovery
```
richard-ramos: Are there plans to make capability discovery specs be part of libp2p? If not, perhaps we should make the numbers of these enum values to not be 6 or 7, but some larger uint32 value (i.e. 5892436 and 5892437) for the rare case in which the official libp2p specs for kad-dht update and add newer enum values, since it's highly likely they'd continue from last value (5)

libp2p/protocols/service_discovery/registrar.nim
SionoiS: it should be fine since we check the other timestamps.

@tinniaru3005 WDYT?

libp2p/protocols/service_discovery/registrar.nim
SionoiS: yes in `disco.rtable.findClosest(serviceId, count)`

libp2p/protocols/capability_discovery/advertiser.nim
```diff
@@ -0,0 +1,249 @@
...
+    registrar: PeerId,
+    bucketIdx: int,
+    scheduledTime: Moment,
+    ticket: Opt[Ticket] = Opt.none(Ticket),
+) =
+  ## Schedule an action at the given time
+
+  let action: PendingAction = (scheduledTime, serviceId, registrar, bucketIdx, ticket)
+  let idx = disco.advertiser.actionQueue.lowerBound(action, actionCmp)
+  disco.advertiser.actionQueue.insert(action, idx)
```
gmelodie: I really like the idea of the action queue, but I think it should contain all of the interface i.e. 
```suggestion
proc schedule*(
    queue: ActionScheduler,
    serviceId: ServiceId,
    registrar: PeerId,
    bucketIdx: int,
    scheduledTime: Moment,
    ticket: Opt[Ticket] = Opt.none(Ticket),
) =
  ## Schedule an action at the given time

  let action: PendingAction = (scheduledTime, serviceId, registrar, bucketIdx, ticket)
  let idx = queue.lowerBound(action, actionCmp)
  queue.insert(action, idx)
```


I think this could be generalized further:
-1. nit: change `ActionQueue` to `Scheduler` (scheduler is such a cool name, looks like a very shiny and durable german kitchen instrument)
0. Implement the ``<`` operator for the `Action` type and use nim's `[heapqueue](https://nim-lang.org/docs/heapqueue.html)` instead of the sequence. 
1. make `schedule` (former `scheduleAction`) take only `scheduler` and `action`, in this case the action needs to be constructed outside of the `schedule` function
2. from (1), further generalize an `action` to be just a closure (any closure) like: 
```nim
scheduler.schedule(1.seconds, proc () =
  sendMessage("peerA", "hello")
)

scheduler.schedule(2.seconds, proc () =
  reconnect("peerB", 3)
)

# ----------------- schedule implementation
proc schedule(delay: Duration, fn: proc() {.async.}) =
  asyncSpawn internalSchedule(delay, fn())  


proc internalSchedule(delay: Duration, fn: proc() {.async.}) {.async.} =
  await sleepAsync(delay)
  await fn()
```
3. move the scheduler out to `utils.nim` and add an `unschedule` function
4. ensure that no task is scheduled exactly at the same time (for most cases this won't be an issue, but it's a nice property to ensure)
5. replace heartbeats by the single scheduler (requires adding a repetition argument to the scheduler to allow a task to repeat indefinitely, or for a specific number of times too!)

Doing steps `-1`, `0`, `1`, `2` and `3` would be enough to remove a lot of the complexity from this file

libp2p/protocols/capability_discovery/advertiser.nim
```diff
@@ -0,0 +1,249 @@
...
+    disco: KademliaDiscovery,
+    serviceId: ServiceId,
+    registrar: PeerId,
+    bucketIdx: int,
+    scheduledTime: Moment,
+    ticket: Opt[Ticket] = Opt.none(Ticket),
+) =
+  disco.scheduleAction(serviceId, registrar, bucketIdx, scheduledTime, ticket)
+
+  disco.processAction()
```
gmelodie: This wouldn't be required with the scheduler approach
```suggestion
```

libp2p/protocols/capability_discovery/advertiser.nim
```diff
@@ -0,0 +1,249 @@
...
+    disco.serviceRoutingTables.insertPeer(serviceId, peerId.toKey())
+
+  case status
+  of protobuf.RegistrationStatus.Confirmed:
+    let nextTime = Moment.now() + chronos.seconds(int(disco.discoConf.advertExpiry))
+    disco.scheduleAction(serviceId, registrar, bucketIdx, nextTime, Opt.none(Ticket))
+  of protobuf.RegistrationStatus.Wait:
+    let newTicket = newTicketOpt.valueOr:
+      error "no ticket to retry with"
+      return
```
gmelodie: We only retry once?

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -0,0 +1,249 @@
...
+  disco.serviceRoutingTables.addService(
+    serviceId, disco.rtable, disco.config.replication, disco.discoConf.bucketsCount,
+    Provided,
+  )
+
+  cd_advertiser_services_added.inc()
+
+  let advTable = disco.serviceRoutingTables.getTable(serviceId).valueOr:
+    error "service not found", serviceId
+    return
```
gmelodie: addservice could return the table

libp2p/protocols/service_discovery/serviceroutingtables.nim
```diff
@@ -0,0 +1,144 @@
...
+    manager: ServiceRoutingTableManager, serviceId: ServiceId
+): Opt[RoutingTable] {.raises: [].} =
+  ## Get routing table for a service (immutable view)
+
+  let res = catch:
+    manager.tables[serviceId]
+  let table = res.valueOr:
+    return Opt.none(RoutingTable)
+
+  return Opt.some(table)
```
gmelodie: You can just do `manager.tables.get(serviceId)` directly to get an Opt I think

libp2p/protocols/service_discovery/serviceroutingtables.nim
```diff
@@ -0,0 +1,144 @@
...
+
+type
+  ServiceStatus* = enum
+    Interest = 0
+    Provided = 1
+    Both = 2
+
+  ServiceRoutingTableManager* = ref object
+    tables*: Table[ServiceId, RoutingTable]
+    serviceStatus*: Table[ServiceId, ServiceStatus]
```
gmelodie, line 23: I think this is overcomplicating things. Using one `table` with a pair of values would reduce the code a bit (i.e. `ServiceRoutingTableManager = TableRef[ServiceId, (RoutingTable, ServiceStatus)]`).

SionoiS:
The POC served us well. RIP

PR #2057, opened 2026-01-28
chore: set build env for windows to match nimbus

richard-ramos:
Requires: https://github.com/vacp2p/nim-lsquic/pull/59
This will avoid us having a diff in CI environment for Windows between what we use in nim-libp2p and what it is used in nimbus

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@
 </p>
 
 ## Background
+
 libp2p is a [Peer-to-Peer](https://en.wikipedia.org/wiki/Peer-to-peer) networking stack, with [implementations](https://github.com/libp2p/libp2p#implementations) in multiple languages derived from the same [specifications.](https://github.com/libp2p/specs)
 
 Building large scale peer-to-peer systems has been complex and difficult in the last 15 years and libp2p is a way to fix that. It strives to be a modular stack with secure defaults and useful protocols, while remaining open and extensible.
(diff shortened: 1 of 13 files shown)


vladopajic (approved):

.github/actions/install_nim/action.yml
```diff
@@ -72,7 +72,7 @@ runs:
         steps.windows-dlls-cache.outputs.cache-hit != 'true' &&
         inputs.os == 'Windows'
       run: |
-        mkdir external
+        mkdir -p external
```
gmelodie, line 75: Why use `-p` here? This is not a nested folder (is it?)

.github/workflows/ci.yml
```diff
@@ -81,15 +81,25 @@ jobs:
...
+          path: external/mingw-${{ matrix.platform.cpu }}
+          key: 'mingw-llvm-17-${{ matrix.platform.cpu }}'
+
+      - name: Install llvm-mingw dependency (Windows)
+        if: >
+          steps.windows-mingw-cache.outputs.cache-hit != 'true' &&
+          runner.os == 'Windows'
         run: |
-          nimble install_pinned
+          mkdir -p external
```
gmelodie, line 97: Same as above

.github/workflows/ci.yml
```diff
@@ -81,15 +81,25 @@ jobs:
...
+          MINGW_BASE="https://github.com/mstorsjo/llvm-mingw/releases/download/20230905"
```
gmelodie, line 98: Why use a version from 2023? can we not install it using `choco` as for `nasm` below?

.github/actions/install_nim/action.yml
richard-ramos: This makes the usage of mkdir safer, as it will be idempotent, i.e.:  it will not fail if the dir already exists, which is good, because if for some reason in the future we swtich the order of the steps in the CI, we wont have to care whether the external directory exists or not.

.github/workflows/ci.yml
richard-ramos: To use the same exact llvm-mingw they use in nimbus.

gmelodie: Then it should come from nimbus, in case they change it

PR #2058, opened 2026-01-28
chore(mix): make delay strategy pluggable

chaitanyaprem:
As per mix spec [delay strategy](https://lip.logos.co/ift-ts/raw/mix.html#62-delay-strategy) is supposed to be pluggable to make instantiations choose the type they want and configure it. 
- Made delay strategy pluggable with default being same as before
- provided a exponential strategy as well.

### DIFF
--- a/libp2p/protocols/mix/delay_strategy.nim
+++ b/libp2p/protocols/mix/delay_strategy.nim
@@ -10,31 +10,27 @@ import ../../crypto/crypto
 type DelayStrategy* = ref object of RootObj ## Abstract interface for delay strategies.
   rng: ref HmacDrbgContext
 
-import results
-
-method generateForEntry*(
-    self: DelayStrategy
-): Result[uint16, string] {.base, gcsafe, raises: [].} =
+method generateForEntry*(self: DelayStrategy): uint16 {.base, gcsafe, raises: [].} =
   ## Generate delay value to encode in packet (called by sender/entry node).
+  ## implementation should return some default value in case of errors
+  raiseAssert "generateForEntry must be implemented by concrete delay strategy types"
 
 method generateForIntermediate*(
     self: DelayStrategy, encodedDelayMs: uint16
 ): uint16 {.base, gcsafe, raises: [].} =
   ## Generate actual delay from encoded value (called by intermediate node).
   ## implementation should return some default value in case of errors
+  raiseAssert "generateForIntermediate must be implemented by concrete delay strategy types"
 
 type NoSamplingDelayStrategy* = ref object of DelayStrategy
   ## Default strategy: generates random delays [0-2]ms, uses them directly.
 
 proc new*(T: typedesc[NoSamplingDelayStrategy], rng: ref HmacDrbgContext): T =
+  doAssert(rng != nil, "random is not set")
   T(rng: rng)
 
-method generateForEntry*(
-    self: NoSamplingDelayStrategy
-): Result[uint16, string] {.gcsafe, raises: [].} =
-  if self.rng.isNil:
-    return err("RNG is nil")
-  ok((self.rng[].generate(uint64) mod 3).uint16)
+method generateForEntry*(self: NoSamplingDelayStrategy): uint16 {.gcsafe, raises: [].} =
+  self.rng[].generate(uint16) mod 3
 
 method generateForIntermediate*(
     self: NoSamplingDelayStrategy, encodedDelayMs: uint16
@@ -52,12 +48,13 @@ proc new*(
     meanDelayMs: uint16 = DefaultMeanDelayMs,
     rng: ref HmacDrbgContext,
 ): T =
+  doAssert(rng != nil, "random is not set")
   T(meanDelayMs: meanDelayMs, rng: rng)
 
 method generateForEntry*(
     self: ExponentialDelayStrategy
-): Result[uint16, string] {.gcsafe, raises: [].} =
-  ok(self.meanDelayMs)
+): uint16 {.gcsafe, raises: [].} =
+  self.meanDelayMs
 
 method generateForIntermediate*(
     self: ExponentialDelayStrategy, meanDelayMs: uint16
@@ -66,15 +63,7 @@ method generateForIntermediate*(
   ## Fall back to no delay in case of errors
   if meanDelayMs == 0:
     return 0u16
-  if self.rng.isNil:
-    return 0u16
   let randVal = self.rng[].generate(uint64)
   let u = (float64(randVal) + 1.0) / (float64(high(uint64)) + 1.0)
   let delay = -float64(meanDelayMs) * ln(u)
   min(delay, float64(high(uint16))).uint16
-
-proc exponentialDelayStrategy*(
-    meanDelayMs: uint16 = DefaultMeanDelayMs, rng: ref HmacDrbgContext
-): DelayStrategy =
-  ## Returns ExponentialDelayStrategy (recommended per spec).
-  ExponentialDelayStrategy.new(meanDelayMs, rng)
--- a/libp2p/protocols/mix/mix_protocol.nim
+++ b/libp2p/protocols/mix/mix_protocol.nim
@@ -368,7 +368,8 @@ proc handleMixMessages(
     trace "Intermediate node processing",
       peerId = mixProto.mixNodeInfo.peerId, multiAddr = mixProto.mixNodeInfo.multiAddr
     mix_messages_recvd.inc(labelValues = ["Intermediate"])
-    let actualDelayMs = mixProto.delayStrategy.generateForIntermediate(processedSP.delayMs.uint16)
+    let actualDelayMs =
+      mixProto.delayStrategy.generateForIntermediate(processedSP.delayMs.uint16)
     trace "Computed delay", encodedDelayMs = processedSP.delayMs, actualDelayMs
     await sleepAsync(milliseconds(actualDelayMs.int))
 
@@ -477,10 +478,12 @@ proc buildSurb(
         availableIndices.del(randomIndexPosition)
         debug "Selected mix node for surbs: ", indexInPath = i, peerId = randPeerId
         let mixPubInfo = mixProto.pubNodeInfo.getOrDefault(randPeerId)
-        let delayMillisec = mixProto.delayStrategy.generateForEntry().valueOr:
-          mix_messages_error.inc(labelValues = ["Entry/SURB", "NON_RECOVERABLE"])
-          return err("failed to generate delay: " & error)
-        (mixPubInfo.peerId, mixPubInfo.multiAddr, mixPubInfo.mixPubKey, delayMillisec)
+        (
+          mixPubInfo.peerId,
+          mixPubInfo.multiAddr,
+          mixPubInfo.mixPubKey,
+          mixProto.delayStrategy.generateForEntry(),
+        )
       else:
         (
           mixProto.mixNodeInfo.peerId, mixProto.mixNodeInfo.multiAddr,
@@ -742,9 +745,7 @@ proc anonymizeLocalProtocolSend*(
 
     let delayMillisec =
       if hop.len != PathLength - 1:
-        mixProto.delayStrategy.generateForEntry().valueOr:
-          mix_messages_error.inc(labelValues = ["Entry", "NON_RECOVERABLE"])
-          return err("failed to generate delay: " & error)
+        mixProto.delayStrategy.generateForEntry()
       else:
         0.uint16 # No delay for exit node
 
@@ -844,14 +845,12 @@ proc new*(
     spamProtection: Opt[SpamProtection] = default(Opt[SpamProtection]),
     delayStrategy: Opt[DelayStrategy] = Opt.none(DelayStrategy),
 ): T =
-  let actualDelayStrategy =
-    if delayStrategy.isSome:
-      delayStrategy.get
-    else:
-      NoSamplingDelayStrategy.new(rng)
+  let actualDelayStrategy = delayStrategy.valueOr:
+    NoSamplingDelayStrategy.new(rng)
   let mixProto = new(T)
   mixProto.init(
-    mixNodeInfo, pubNodeInfo, switch, tagManager, rng, spamProtection, actualDelayStrategy
+    mixNodeInfo, pubNodeInfo, switch, tagManager, rng, spamProtection,
+    actualDelayStrategy,
   )
   mixProto
 
--- a/tests/libp2p/mix/test_delay_strategy.nim
+++ b/tests/libp2p/mix/test_delay_strategy.nim
@@ -4,7 +4,6 @@
 {.used.}
 
 import std/[sets]
-import ../../../libp2p/crypto/crypto
 import ../../../libp2p/protocols/mix/delay_strategy
 import ../../tools/[unittest, crypto]
 
@@ -15,19 +14,13 @@ const
 
 suite "DelayStrategy":
   test "NoSamplingDelayStrategy generateForEntry returns values in [0, 2]":
-    let
-      rng = rng()
-      strategy = NoSamplingDelayStrategy.new(rng)
+    let strategy = NoSamplingDelayStrategy.new(rng())
 
     for _ in 0 ..< NumIterations:
-      let delay = strategy.generateForEntry().valueOr:
-        raiseAssert error
-      check delay <= 2
+      check strategy.generateForEntry() <= 2
 
   test "NoSamplingDelayStrategy generateForIntermediate returns encoded value":
-    let
-      rng = rng()
-      strategy = NoSamplingDelayStrategy.new(rng)
+    let strategy = NoSamplingDelayStrategy.new(rng())
 
     check:
       strategy.generateForIntermediate(100) == 100
@@ -37,20 +30,17 @@ suite "DelayStrategy":
     let rng = rng()
 
     check:
-      ExponentialDelayStrategy.new(50, rng).generateForEntry().get() == 50
-      ExponentialDelayStrategy.new(100, rng).generateForEntry().get() == 100
+      ExponentialDelayStrategy.new(50, rng).generateForEntry() == 50
+      ExponentialDelayStrategy.new(100, rng).generateForEntry() == 100
 
   test "ExponentialDelayStrategy generateForIntermediate returns 0 for mean 0":
-    let
-      strategy = ExponentialDelayStrategy.new(0, rng)
-      rng = rng()
+    let strategy = ExponentialDelayStrategy.new(0, rng())
 
     check strategy.generateForIntermediate(0) == 0
 
   test "ExponentialDelayStrategy generateForIntermediate samples from exponential distribution":
     let
-      rng = rng()
-      strategy = ExponentialDelayStrategy.new(100, rng)
+      strategy = ExponentialDelayStrategy.new(100, rng())
       meanDelayMs: uint16 = 100
       numSamples = 1000
     var sum: float64 = 0
@@ -67,8 +57,7 @@ suite "DelayStrategy":
 
   test "ExponentialDelayStrategy produces variable delays":
     let
-      rng = rng()
-      strategy = ExponentialDelayStrategy.new(100, rng)
+      strategy = ExponentialDelayStrategy.new(100, rng())
       meanDelayMs: uint16 = 100
 
     var delays = initHashSet[uint16]()


libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,81 @@
...
+    self: ExponentialDelayStrategy, rng: ref HmacDrbgContext, meanDelayMs: uint16
+): Result[uint16, string] {.gcsafe, raises: [].} =
+  ## Samples from exponential distribution: delay = -mean * ln(U)
+  if meanDelayMs == 0:
+    return ok(0u16)
+  if rng.isNil:
+    return err("RNG is nil")
+  let randVal = rng[].generate(uint64)
+  let u = (float64(randVal) + 1.0) / (float64(high(uint64)) + 1.0)
+  let delay = -float64(meanDelayMs) * ln(u)
```
gmelodie, line 68: Nit: I think this would be slightly less convoluted

```suggestion
  let delay = abs(float64(meanDelayMs) * ln(u))
```

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+
+    for _ in 0 ..< 100:
+      let delay = strategy.generateDelay(rng).valueOr:
+        raiseAssert error
+      check delay <= 2
+
+  test "NoSamplingDelayStrategy computeDelay returns encoded value":
+    let
+      strategy = NoSamplingDelayStrategy.new()
+      rng = newRng()
```
gmelodie: Use [`rng` from tests/tools/crypto.nim](https://github.com/vacp2p/nim-libp2p/blob/d16a307cba44c788215ea768abeb9f077b912429/tests/tools/crypto.nim#L13)

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+
+    let result100 = strategy.computeDelay(rng, 100).valueOr:
+      raiseAssert error
+    let result200 = strategy.computeDelay(rng, 200).valueOr:
+      raiseAssert error
+    check result100 == 100
+    check result200 == 200
```
gmelodie: You don't need to `raiseAssert` in tests
```suggestion
    strategy.computeDelay(rng, 100).get() == 100
    strategy.computeDelay(rng, 200).get() == 200
```

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+
+  test "ExponentialDelayStrategy generateDelay returns configured mean":
+    let rng = newRng()
+
+    let result50 = ExponentialDelayStrategy.new(50).generateDelay(rng).valueOr:
+        raiseAssert error
+    let result100 = ExponentialDelayStrategy.new(100).generateDelay(rng).valueOr:
+        raiseAssert error
+    check result50 == 50
+    check result100 == 100
```
gmelodie: Same as above

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+
+    let result100 = strategy.computeDelay(rng, 100).valueOr:
+      raiseAssert error
+    let result200 = strategy.computeDelay(rng, 200).valueOr:
+      raiseAssert error
+    check result100 == 100
+    check result200 == 200
+
+  test "ExponentialDelayStrategy generateDelay returns configured mean":
+    let rng = newRng()
```
gmelodie: Same as above

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+        raiseAssert error
+    let result100 = ExponentialDelayStrategy.new(100).generateDelay(rng).valueOr:
+        raiseAssert error
+    check result50 == 50
+    check result100 == 100
+
+  test "ExponentialDelayStrategy computeDelay returns 0 for mean 0":
+    let
+      strategy = ExponentialDelayStrategy.new()
+      rng = newRng()
```
gmelodie: Same as above

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+
+    let result0 = strategy.computeDelay(rng, 0).valueOr:
+      raiseAssert error
+    check result0 == 0
```
gmelodie: Same as above

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+
+  test "ExponentialDelayStrategy computeDelay samples from exponential distribution":
+    let
+      strategy = ExponentialDelayStrategy.new()
+      rng = newRng()
```
gmelodie: Same as above

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+      meanDelayMs: uint16 = 100
+      numSamples = 1000
+
+    var
+      sum: float64 = 0
+      samples: seq[uint16] = @[]
+
+    for _ in 0 ..< numSamples:
+      let delay = strategy.computeDelay(rng, meanDelayMs).valueOr:
+        raiseAssert error
```
gmelodie: Same as above

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+    let empiricalMean = sum / float64(numSamples)
+    # Allow 20% tolerance for statistical variation
+    check:
+      empiricalMean > float64(meanDelayMs) * 0.8
+      empiricalMean < float64(meanDelayMs) * 1.2
+
+  test "ExponentialDelayStrategy produces variable delays":
+    let
+      strategy = ExponentialDelayStrategy.new()
+      rng = newRng()
```
gmelodie: Same as above

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+      meanDelayMs: uint16 = 100
+
+    var delays: seq[uint16] = @[]
+    for _ in 0 ..< 10:
+      let delay = strategy.computeDelay(rng, meanDelayMs).valueOr:
+        raiseAssert error
```
gmelodie: Same as above

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+      delays.add(delay)
+
+    # Not all delays should be the same
+    var allSame = true
+    for i in 1 ..< delays.len:
+      if delays[i] != delays[0]:
+        allSame = false
+        break
+    check not allSame
```
gmelodie: Use `anyIt` from `sequtils`:

```suggestion
    check delays.anyIt(it != delays[0])
```

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+      empiricalMean < float64(meanDelayMs) * 1.2
+
+  test "ExponentialDelayStrategy produces variable delays":
+    let
+      strategy = ExponentialDelayStrategy.new()
+      rng = newRng()
+      meanDelayMs: uint16 = 100
+
+    var delays: seq[uint16] = @[]
+    for _ in 0 ..< 10:
```
gmelodie: Consider making `10` a `const`

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+      let delay = strategy.computeDelay(rng, meanDelayMs).valueOr:
+        raiseAssert error
+      samples.add(delay)
+      sum += float64(delay)
+
+    let empiricalMean = sum / float64(numSamples)
+    # Allow 20% tolerance for statistical variation
+    check:
+      empiricalMean > float64(meanDelayMs) * 0.8
+      empiricalMean < float64(meanDelayMs) * 1.2
```
gmelodie: Consider making `0.2` a const and getting `0.8` by doing `1 - yourConst`

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,93 @@
...
+import ../../../libp2p/protocols/mix/delay_strategy
+import ../../tools/[unittest]
+
+suite "delay_strategy_tests":
+  test "NoSamplingDelayStrategy generateDelay returns values in [0, 2]":
+    let
+      strategy = NoSamplingDelayStrategy.new()
+      rng = newRng()
+
+    for _ in 0 ..< 100:
```
gmelodie: Consider making `100` a `const`

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,81 @@
...
+  ok(min(delay, float64(high(uint16))).uint16)
```
gmelodie: Why do you have to do `min` here? Maybe if you do this before
```nim
  let randVal = rng[].generate(uint16)
```
Then you'd have the guarantee that it's not gonna exceed, correct? (I'm not sure)

chaitanyaprem: well, the min is to avoid truncation or overflow of converting float64 to uint16. 

We already generate randVal using uint64 itself. Did not get your suggestion.

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,81 @@
...
+  ## Default strategy: generates random delays [0-2]ms, uses them directly.
+
+proc new*(T: typedesc[NoSamplingDelayStrategy]): T =
+  T()
+
+method generateDelay*(
+    self: NoSamplingDelayStrategy, rng: ref HmacDrbgContext
+): Result[uint16, string] {.gcsafe, raises: [].} =
+  if rng.isNil:
+    return err("RNG is nil")
```
vladopajic: create `DelayStrategy` with the `rng`. so you don't have to pass it each time when generating delay.

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,81 @@
...
+    self: DelayStrategy, rng: ref HmacDrbgContext
+): Result[uint16, string] {.base, gcsafe, raises: [].} =
+  ## Generate delay value to encode in packet (called by sender).
+  raiseAssert "generateDelay must be implemented by concrete delay strategy types"
+
+method computeDelay*(
+    self: DelayStrategy, rng: ref HmacDrbgContext, encodedDelayMs: uint16
+): Result[uint16, string] {.base, gcsafe, raises: [].} =
+  ## Compute actual delay from encoded value (called by intermediate node).
+  raiseAssert "computeDelay must be implemented by concrete delay strategy types"
```
vladopajic: i would avoid `rng` from interface

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,81 @@
...
+
+type DelayStrategy* = ref object of RootObj ## Abstract interface for delay strategies.
+
+import results
+
+method generateDelay*(
+    self: DelayStrategy, rng: ref HmacDrbgContext
+): Result[uint16, string] {.base, gcsafe, raises: [].} =
+  ## Generate delay value to encode in packet (called by sender).
+  raiseAssert "generateDelay must be implemented by concrete delay strategy types"
```
vladopajic: since there are implementations of this interface consider not having body at for these methods.

libp2p/protocols/mix/delay_strategy.nim
vladopajic: since there are implementations of this interface consider not having body at for these methods.

vladopajic: no `raiseAssert`

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -365,9 +367,15 @@ proc handleMixMessages(
   of Intermediate:
     trace "Intermediate node processing",
       peerId = mixProto.mixNodeInfo.peerId, multiAddr = mixProto.mixNodeInfo.multiAddr
-    # Add delay
     mix_messages_recvd.inc(labelValues = ["Intermediate"])
-    await sleepAsync(milliseconds(processedSP.delayMs))
+    let actualDelayMs = mixProto.delayStrategy.computeDelay(
+      mixProto.rng, processedSP.delayMs.uint16
+    ).valueOr:
+      trace "Failed to compute delay", err = error
```
vladopajic: this should be no failure operation

chaitanyaprem: Well, this is mainly for very rare case where random number generation fails or for some reason delayStrategy is not able to compute due to whatever reason e.g mean is out of range or some other reason. 

It makes sense to leave it for such scenarios.

vladopajic: when you see it that way, it does make sense.

but also, if we create an interface that doesn't fail, then user defined delay will never fail, thus making things more robust here. and user can easily make it failure-less if they in case if random generation fails they can just return some hardcoded value - default to value.

vladopajic: we can do that instead of user. but it is better for users to do it, as there are many users choosing default value can result in some kind of randomness with fallback values.

interface can document this to explain users what to do when "true" random fails

chaitanyaprem: Yea, i am just forseeing some instantiation of mixnet where-in sender violates delay strategy defined by the network  or something similar. This can become a reason for intermediate node to not process further, but it is a weird way to look at it as it is not some kind of violation of protocol or something like that.  Atleast i am not sure at this point and wondering if leaving it open gives that flexibility.

libp2p/protocols/mix/delay_strategy.nim
arnetheduck: +1 for not making it part of any interface - we've been considering moving on from bearssl so we don't want to tie ourselves to any one particular rng - if the rng needs to be dynamic, it's better to use a callback like so:

```nim
type Rng = proc(v: var openArray[byte]) {.raises: [], gcsafe.}
```

this is the API that most CSRNGs operate on, ie the filling of a number of bytes with uniform random data.

libp2p/protocols/mix/mix_protocol.nim
vladopajic: i do think that case should be dealt with `computeDelay` - in my mind this operation needs to just to return delay. all failures to do this operation should result in return some fallback to value.  this should be maximum concern here.

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,81 @@
...
+  if rng.isNil:
+    return err("RNG is nil")
+  let randVal = rng[].generate(uint64)
+  let u = (float64(randVal) + 1.0) / (float64(high(uint64)) + 1.0)
+  let delay = abs(float64(meanDelayMs) * ln(u))
+  ok(min(delay, float64(high(uint16))).uint16)
+
+proc defaultDelayStrategy*(): DelayStrategy =
+  ## Returns NoSamplingDelayStrategy (backward compatible).
+  NoSamplingDelayStrategy.new()
```
vladopajic: this seems unnecessary

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -806,15 +812,16 @@ proc init*(
     tagManager: TagManager = TagManager.new(),
     rng: ref HmacDrbgContext = newRng(),
     spamProtection: Opt[SpamProtection] = default(Opt[SpamProtection]),
+    delayStrategy: DelayStrategy = defaultDelayStrategy(),
```
vladopajic: `delayStrategy: DelayStrategy = NoSamplingDelayStrategy.new(),`

libp2p/protocols/mix/mix_protocol.nim
chaitanyaprem: ok, to keep it simple for now i will leave error out. 
if needed in future we can always analyze and change, anyways it is not a clear scenario at this point of time.

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -840,9 +847,12 @@ proc new*(
...
```
vladopajic: why is there 

`new*` and `init*` and both have ` delayStrategy: DelayStrategy = defaultDelayStrategy(),`, make it only in once place

libp2p/protocols/mix/mix_protocol.nim
vladopajic: > ok, to keep it simple for now i will leave error out.
> if needed in future we can always analyze and change, anyways it is not a clear scenario at this point of time.

yes, would be better to leave it out. 
please, add to documentation that implementation should return some default to value in case of errors. this value can be generated at initialization.

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,81 @@
...
+    self: ExponentialDelayStrategy, rng: ref HmacDrbgContext, meanDelayMs: uint16
+): Result[uint16, string] {.gcsafe, raises: [].} =
+  ## Samples from exponential distribution: delay = -mean * ln(U)
+  if meanDelayMs == 0:
+    return ok(0u16)
+  if rng.isNil:
+    return err("RNG is nil")
+  let randVal = rng[].generate(uint64)
+  let u = (float64(randVal) + 1.0) / (float64(high(uint64)) + 1.0)
+  let delay = abs(float64(meanDelayMs) * ln(u))
```
richard-ramos: This is a nit, but the abs is not necessary, as you could write this instead as: `-float64(meanDelayMs) * ln(u)`,  because  `ln(u)` will always be <= 0, (due to `u` being > 0 and <= 1)

Image

also, that way you would match the formula in the comment on line 63

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,83 @@
...
+import ../../../libp2p/crypto/crypto
+import ../../../libp2p/protocols/mix/delay_strategy
+import ../../tools/[unittest, crypto]
+
+const
+  NumIterations = 100
+  NumSamples = 10
+  Tolerance = 0.2 # 20% tolerance for statistical tests
+
+suite "delay_strategy_tests":
```
vladopajic:
```suggestion
suite "DelayStrategy":
```

libp2p/protocols/mix/delay_strategy.nim
gmelodie: Sorry I meant `.generate(uint16)` in the suggestion above, corrected now

libp2p/protocols/mix/delay_strategy.nim
gmelodie: lul I was the one who suggested changing from `-float...` to `abs`, it was a nitpick though, fine with whatever

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,81 @@
...
+
+import std/math
+import bearssl/rand
+import ../../crypto/crypto
+
+type DelayStrategy* = ref object of RootObj ## Abstract interface for delay strategies.
+
+import results
+
+method generateDelay*(
```
vladopajic: `method generateDelay*(`
`method computeDelay*(`

these methods names are confusing and one wouldn't know where and how they are used because name is identical.

----

some hints:

- do not use `Delay` in these names, since types already have delay (so most likely variables are going to have it as well). example:

  `myDelay.generateDelay()` -> better if `myDelay.generate()`


- `computeDelay` can also be `generate`, since it has argument then signature is different.  (strategy.1 go with the flow, and make it identical)

- or it could be  `generateFrom` or `generateUsing`, or .... `generateXXXXX` (whatever suggestion for XXXXX you have) (strategy.2 make distinction)

- it could be `generateForEncoding` and `generateFromEncoded` ?


- whatever you choose, please have in mind that there is marginal difference between `generate` and `compute` here, so sticking with the same term would be better, imo.

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,83 @@
...
+      empiricalMean < float64(meanDelayMs) * (1 + Tolerance)
+
+  test "ExponentialDelayStrategy produces variable delays":
+    let
+      strategy = ExponentialDelayStrategy.new()
+      rng = rng()
+      meanDelayMs: uint16 = 100
+
+    var delays: seq[uint16] = @[]
+    for _ in 0 ..< NumSamples:
```
vladopajic:
```
var delays = initHashSet[uint16]()
....

check delays.len > NumSamples / 2
```

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,83 @@
...
+  test "ExponentialDelayStrategy computeDelay samples from exponential distribution":
+    let
+      strategy = ExponentialDelayStrategy.new()
+      rng = rng()
+      meanDelayMs: uint16 = 100
+      numSamples = 1000
+
+    var
+      sum: float64 = 0
+      samples: seq[uint16] = @[]
```
vladopajic: what is `samples` used for here?

libp2p/protocols/mix/delay_strategy.nim
chaitanyaprem: Yea, i could not think of better names and they are indeed confusing. 
Interface methods have comments as to which method has to be called by which node(sender/intermediate). Should be enough i guess. 

But i agree with your point about not having delay again in method names.

vladopajic: then, there you have it:

generateFromSender
generateFromIntermediate

vladopajic (changes requested):
since it's already approved, adding request for change to prevent accidental merge before giving final review

tests/libp2p/mix/test_delay_strategy.nim
chaitanyaprem: looks like remnant , removing it.

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,80 @@
...
+import ../../crypto/crypto
+
+type DelayStrategy* = ref object of RootObj ## Abstract interface for delay strategies.
+  rng: ref HmacDrbgContext
+
+import results
+
+method generateForEntry*(
+    self: DelayStrategy
+): Result[uint16, string] {.base, gcsafe, raises: [].} =
```
vladopajic: should also not use `Result`, just like `generateForIntermediate`
should have same comment about default value.

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,80 @@
...
+    self: DelayStrategy, encodedDelayMs: uint16
+): uint16 {.base, gcsafe, raises: [].} =
+  ## Generate actual delay from encoded value (called by intermediate node).
+  ## implementation should return some default value in case of errors
+
+type NoSamplingDelayStrategy* = ref object of DelayStrategy
+  ## Default strategy: generates random delays [0-2]ms, uses them directly.
+
+proc new*(T: typedesc[NoSamplingDelayStrategy], rng: ref HmacDrbgContext): T =
+  T(rng: rng)
```
vladopajic, line 30: `doAssert(rng != nil, "random is not set")`

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,80 @@
...
+
+method generateForEntry*(
+    self: NoSamplingDelayStrategy
+): Result[uint16, string] {.gcsafe, raises: [].} =
+  if self.rng.isNil:
+    return err("RNG is nil")
```
vladopajic: Unnecessary with `doAssert` in constructor

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,80 @@
...
+type ExponentialDelayStrategy* = ref object of DelayStrategy
+  ## Recommended strategy: encodes mean delay, samples from exponential distribution.
+  meanDelayMs: uint16
+
+proc new*(
+    T: typedesc[ExponentialDelayStrategy],
+    meanDelayMs: uint16 = DefaultMeanDelayMs,
+    rng: ref HmacDrbgContext,
+): T =
+  T(meanDelayMs: meanDelayMs, rng: rng)
```
vladopajic, line 52: doAssert(rng != nil, "random is not set")

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,80 @@
...
+
+method generateForIntermediate*(
+    self: ExponentialDelayStrategy, meanDelayMs: uint16
+): uint16 {.gcsafe, raises: [].} =
+  ## Samples from exponential distribution: delay = -mean * ln(U)
+  ## Fall back to no delay in case of errors
+  if meanDelayMs == 0:
+    return 0u16
+  if self.rng.isNil:
+    return 0u16
```
vladopajic: unnecessary with `doAssert` in constructor

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,80 @@
...
+  let randVal = self.rng[].generate(uint64)
+  let u = (float64(randVal) + 1.0) / (float64(high(uint64)) + 1.0)
+  let delay = -float64(meanDelayMs) * ln(u)
+  min(delay, float64(high(uint16))).uint16
+
+proc exponentialDelayStrategy*(
+    meanDelayMs: uint16 = DefaultMeanDelayMs, rng: ref HmacDrbgContext
```
vladopajic: `exponentialDelayStrategy` is unnecessary, `ExponentialDelayStrategy.new` should be used instead

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,79 @@
...
+const
+  NumIterations = 100
+  NumSamples = 10
+  Tolerance = 0.2 # 20% tolerance for statistical tests
+
+suite "DelayStrategy":
+  test "NoSamplingDelayStrategy generateForEntry returns values in [0, 2]":
+    let
+      rng = rng()
+      strategy = NoSamplingDelayStrategy.new(rng)
```
vladopajic:
```suggestion
    let strategy = NoSamplingDelayStrategy.new(rng())
```

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,79 @@
...
+
+    for _ in 0 ..< NumIterations:
+      let delay = strategy.generateForEntry().valueOr:
+        raiseAssert error
+      check delay <= 2
+
+  test "NoSamplingDelayStrategy generateForIntermediate returns encoded value":
+    let
+      rng = rng()
+      strategy = NoSamplingDelayStrategy.new(rng)
```
vladopajic:
```suggestion
    let strategy = NoSamplingDelayStrategy.new(rng())
```

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,79 @@
...
+
+    check:
+      strategy.generateForIntermediate(100) == 100
+      strategy.generateForIntermediate(200) == 200
+
+  test "ExponentialDelayStrategy generateForEntry returns configured mean":
+    let rng = rng()
```
vladopajic, line 30: `let rng` is also unnecessary here.

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,79 @@
...
+
+    check:
+      ExponentialDelayStrategy.new(50, rng).generateForEntry().get() == 50
+      ExponentialDelayStrategy.new(100, rng).generateForEntry().get() == 100
+
+  test "ExponentialDelayStrategy generateForIntermediate returns 0 for mean 0":
+    let
+      strategy = ExponentialDelayStrategy.new(0, rng)
+      rng = rng()
```
vladopajic: `let rng` is also unnecessary here. 
also 
 `strategy = ExponentialDelayStrategy.new(0, rng)` ->  `strategy = ExponentialDelayStrategy.new(0, rng()) `

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,79 @@
...
+
+    check strategy.generateForIntermediate(0) == 0
+
+  test "ExponentialDelayStrategy generateForIntermediate samples from exponential distribution":
+    let
+      rng = rng()
```
vladopajic: `let rng` is also unnecessary here.

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -0,0 +1,79 @@
...
+
+    let empiricalMean = sum / float64(numSamples)
+    # Allow 20% tolerance for statistical variation
+    check:
+      empiricalMean > float64(meanDelayMs) * (1 - Tolerance)
+      empiricalMean < float64(meanDelayMs) * (1 + Tolerance)
+
+  test "ExponentialDelayStrategy produces variable delays":
+    let
+      rng = rng()
```
vladopajic: `let rng` is also unnecessary here.

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -840,9 +842,17 @@ proc new*(
     tagManager: TagManager = TagManager.new(),
     rng: ref HmacDrbgContext = newRng(),
     spamProtection: Opt[SpamProtection] = default(Opt[SpamProtection]),
+    delayStrategy: Opt[DelayStrategy] = Opt.none(DelayStrategy),
 ): T =
+  let actualDelayStrategy =
+    if delayStrategy.isSome:
+      delayStrategy.get
+    else:
+      NoSamplingDelayStrategy.new(rng)
```
vladopajic: nit

```
let actualDelayStrategy = delayStrategy.valueOr():
  NoSamplingDelayStrategy.new(rng)
```

libp2p/protocols/mix/delay_strategy.nim
vladopajic: please revert to original code - with `raiseAssert`.

---

chatgpt made me believe that abstract methods are possible, so i advised that here. they are not, methods must have implementation.

additional gotcha is this:

```
# this does not compile
proc a(f: foo) {.base.} = 

# this does not compile
method a(f: foo) {.base.} = 

# this does not compile 
method a(f: foo) {.base.} = 
  # abstrat method

# this compiles !!!!
method a(f: foo) {.base.} = 
  ## abstrat method
```
note that `##` documentation comments which are formally part of the ast and therefore count as "something" so the method will compile. this is the case why this code compiled.

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -0,0 +1,80 @@
...
+
+method generateForEntry*(
+    self: DelayStrategy
+): Result[uint16, string] {.base, gcsafe, raises: [].} =
+  ## Generate delay value to encode in packet (called by sender/entry node).
+
+method generateForIntermediate*(
+    self: DelayStrategy, encodedDelayMs: uint16
+): uint16 {.base, gcsafe, raises: [].} =
+  ## Generate actual delay from encoded value (called by intermediate node).
```
vladopajic, line 21: https://github.com/vacp2p/nim-libp2p/pull/2058#discussion_r2743361076

libp2p/protocols/mix/delay_strategy.nim
chaitanyaprem: I dont like constructor's failing with exception, not sure what to do in such case though. 
should we set it by default to `newRng()`

libp2p/protocols/mix/delay_strategy.nim
chaitanyaprem: Thanks, this is what happens when you are sleepy 🤦

libp2p/protocols/mix/delay_strategy.nim
vladopajic: failing in constructors is usually safer than in methods.  since most constructors will be called at initialization any defect at that point are going to get noticed.

second thing is that this is programmer error. it is always fine to raise error, in this case.

libp2p/protocols/mix/delay_strategy.nim
chaitanyaprem: > chatgpt made me believe that abstract methods are possible, so i advised that here. they are not, methods must have implementation.

Oh..that is interesting. let me revert the changes then.

PR #2062, opened 2026-01-29
chore(kad): remove hard coded codec

SionoiS:
I removed the hard-coded codec so that in the future we can use a different one with no code change.


vladopajic (approved):
but still, i would consider removing this constant, as it is obviously causing antipattern behavior, in favor of code like this:

1) remove `const` at whole

```
proc new*(
    T: typedesc[KadDHT],
    switch: Switch,
    bootstrapNodes: seq[(PeerId, seq[MultiAddress])] = @[],
    config: KadDHTConfig = KadDHTConfig.new(),
    client: bool = false,
    codec: string = "/ipfs/kad/1.0.0",
): T {.raises: [].} =
```

2) or just have it unexported in file `kademlian.nim`, where no one can use it.


----

this is not the first time we have issue with these constants in types, that shouldn't be used but are actually used.

since defaults are added to "types" and types are included everywhere, it might be better to have "defaults" which are selectively imported only where they are used. but this will only make sense if number of imports like this will be one, two, three places.

PR #2064, opened 2026-01-29
feat(cbind): mix

richard-ramos:
Adds some functions required for using mix with the cbindings.

### DIFF
--- a/cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
+++ b/cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
@@ -114,11 +114,13 @@ proc parseBootstrapNodes(config: Libp2pConfig): seq[(PeerId, seq[MultiAddress])]
   return response
 
 proc toByteSeq(key: Libp2pPrivateKey): seq[byte] =
-  let data = newSeqUninit[byte](key.dataLen)
+  if key.dataLen == 0 or key.data.isNil():
+    return newSeq[byte](0)
+  let data = newSeqUninit[byte](key.dataLen.int)
   copyMem(addr data[0], key.data, key.dataLen)
   return data
 
-proc mountProtocols(libp2p: var LibP2P, config: Libp2pConfig) =
+proc mountGossipsub(libp2p: var LibP2P, config: Libp2pConfig) =
   var gossipSub = Opt.none(GossipSub)
   if config.mountGossipsub != 0:
     let gs = GossipSub.init(
@@ -126,16 +128,18 @@ proc mountProtocols(libp2p: var LibP2P, config: Libp2pConfig) =
     )
     libp2p.switch.mount(gs)
     gossipSub = Opt.some(gs)
+  libp2p.gossipSub = gossipSub
 
-  libp2p.switch.mount(Ping.new())
-
+proc mountKad(libp2p: var LibP2P, config: Libp2pConfig) =
   var kad = Opt.none(KadDHT)
   if config.mountKad != 0:
     let bootstrapNodes = parseBootstrapNodes(config)
     let k = KadDHT.new(libp2p.switch, bootstrapNodes = bootstrapNodes)
     libp2p.switch.mount(k)
     kad = Opt.some(k)
+  libp2p.kad = kad
 
+proc mountMix(libp2p: var LibP2P, config: Libp2pConfig) =
   var mix = Opt.none(MixProtocol)
   if config.mountMix != 0 and libp2p.mixNodeInfo.isSome:
     if config.mixNodesLen <= 0:
@@ -147,10 +151,17 @@ proc mountProtocols(libp2p: var LibP2P, config: Libp2pConfig) =
     )
     libp2p.switch.mount(mixProto)
     mix = Opt.some(mixProto)
-
-  libp2p.gossipSub = gossipSub
-  libp2p.kad = kad
   libp2p.mix = mix
+
+proc mountProtocols(libp2p: var LibP2P, config: Libp2pConfig) =
+  if config.mountGossipsub != 0:
+    libp2p.mountGossipsub(config)
+  if config.mountKad != 0:
+    libp2p.mountKad(config)
+
+  libp2p.switch.mount(Ping.new())
+
+  libp2p.mountMix(config)
 
 proc processGetPublicKey*(
     self: ptr LifecycleRequest, libp2p: ptr LibP2P
@@ -256,7 +267,7 @@ proc destroyShared(self: ptr LifecycleRequest) =
 
 proc process*(
     self: ptr LifecycleRequest, libp2p: ptr LibP2P
-): Future[Result[string, string]] {.async: (raises: [CancelledError]).} =
+): Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
   defer:
     destroyShared(self)
 
@@ -282,4 +293,4 @@ proc process*(
   else:
     raiseAssert "unsupported operation"
 
-  return ok("")
+  return ok()
--- a/cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
+++ b/cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
@@ -105,7 +105,7 @@ proc processDial*(
 
 proc processMixRegisterDestRead*(
     self: ptr StreamRequest, libp2p: ptr LibP2P
-): Future[Result[string, string]] {.async: (raises: [CancelledError]).} =
+): Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
   defer:
     destroyShared(self)
 
@@ -116,17 +116,15 @@ proc processMixRegisterDestRead*(
     return err("proto is empty")
 
   let sizeParam = self[].mixReadBehaviorParam.int
+  if sizeParam < 0:
+    return err("size must be >= 0")
   case MixReadBehaviorKind(self[].mixReadBehaviorKind)
   of MIX_READ_EXACTLY:
-    if sizeParam < 0:
-      return err("readExactly size must be >= 0")
     mixProto.registerDestReadBehavior($self[].proto, readExactly(sizeParam))
   of MIX_READ_LP:
-    if sizeParam < 0:
-      return err("readLp max size must be >= 0")
     mixProto.registerDestReadBehavior($self[].proto, readLp(sizeParam))
 
-  return ok("")
+  return ok()
 
 proc processMixDial*(
     self: ptr StreamRequest, libp2p: ptr LibP2P
@@ -161,7 +159,7 @@ proc processMixDial*(
 
 proc processMixSetNodeInfo*(
     self: ptr StreamRequest, libp2p: ptr LibP2P
-): Future[Result[string, string]] {.async: (raises: [CancelledError]).} =
+): Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
   defer:
     destroyShared(self)
 
@@ -218,13 +216,12 @@ proc processMixSetNodeInfo*(
       return err("could not mount mix: " & exc.msg)
 
     libp2p[].mix = Opt.some(mixProto)
-    echo "DSAFDSFS"
 
-  return ok("")
+  return ok()
 
 proc processMixNodePoolAdd*(
     self: ptr StreamRequest, libp2p: ptr LibP2P
-): Future[Result[string, string]] {.async: (raises: [CancelledError]).} =
+): Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
   defer:
     destroyShared(self)
 
@@ -248,11 +245,11 @@ proc processMixNodePoolAdd*(
     return err("libp2p public key invalid")
 
   mixProto.nodePool.add(MixPubInfo.init(peerId, maddr, mixPubKey, libp2pPubKey))
-  return ok("")
+  return ok()
 
 proc processClose*(
     self: ptr StreamRequest, libp2p: ptr LibP2P
-): Future[Result[string, string]] {.async: (raises: [CancelledError]).} =
+): Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
   defer:
     destroyShared(self)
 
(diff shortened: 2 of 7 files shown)


cbind/examples/mix.c
```diff
@@ -0,0 +1,334 @@
...
+  libp2p_curve25519_key_t mix_pub_keys[NUM_NODES] = {0};
+  // Needed for mix node pool entries (mix pool stores mix pubkey + libp2p pubkey).
+  libp2p_secp256k1_pubkey_t libp2p_pub_keys[NUM_NODES] = {0};
+  libp2p_private_key_t libp2p_priv_keys[NUM_NODES] = {0};
+
+  for (int i = 0; i < NUM_NODES; i++) {
+    libp2p_config_t cfg = {0};
+    cfg.flags = LIBP2P_CFG_MIX | LIBP2P_CFG_PRIVATE_KEY | LIBP2P_CFG_GOSSIPSUB;
+    cfg.mount_mix = 1;
+    cfg.mount_gossipsub = 0;
```
gmelodie, line 53: Having `LIBP2P_CFG_GOSSIPSUB` set but `cfg.mount_gossipsub = 0` doesn't make sense. We should even get rid of the double way of doing the same thing, with the `config` as it is (maybe with more fields like `mountKad`) and having a function populate the config from the flags:
```nim
proc populateFlags(var config: Libp2pConfig, flags: uint32) =
  config.mountGossipsub = flags and Libp2pCfgGossipsub
  config.mountGossipsubTriggerSelf = flags and Libp2pCfgGossipsubTriggerSelf
  config.mountKad = flags and Libp2pCfgKad
  ...

# then...
config.populateFlags(flags)
```

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
```diff
@@ -45,12 +54,19 @@ proc applyConfigDefaults(config: ptr Libp2pConfig): Libp2pConfig =
...
   if (flags and Libp2pCfgGossipsubTriggerSelf) != 0'u32:
     resolved.gossipsubTriggerSelf = config[].gossipsubTriggerSelf
   if (flags and Libp2pCfgKad) != 0'u32:
     resolved.mountKad = config[].mountKad
+  if (flags and Libp2pCfgMix) != 0'u32:
+    resolved.mountMix = config[].mountMix
+    resolved.mixIndex = config[].mixIndex
+    resolved.mixNodesLen = config[].mixNodesLen
+    if not config[].mixNodeInfoPath.isNil() and config[].mixNodeInfoPath[0] != '\0':
+      resolved.mixNodeInfoPath = config[].mixNodeInfoPath
```
gmelodie, line 69: Using the approach suggested above, this would become
```suggestion
  if (flags and Libp2pCfgMix) != 0'u32:
    resolved.mountMix = 
    resolved.mixIndex = 1
    ...
```

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
```diff
@@ -102,6 +118,56 @@ proc toByteSeq(key: Libp2pPrivateKey): seq[byte] =
...
   return data
 
+proc mountProtocols(libp2p: var LibP2P, config: Libp2pConfig) =
+  var gossipSub = Opt.none(GossipSub)
+  if config.mountGossipsub != 0:
+    let gs = GossipSub.init(
+      switch = libp2p.switch, triggerSelf = config.gossipsubTriggerSelf != 0
+    )
+    libp2p.switch.mount(gs)
+    gossipSub = Opt.some(gs)
```
gmelodie, line 130: nit: these should be each their own functions

```nim
proc mountProtocols(libp2p: var LibP2P, config: Libp2pConfig) =
  if config.mountGossipsub != 0:
    libp2p.mountGossipsub()
  if config.mountKad != 0:
    libp2p.mountKad()
```

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
```diff
@@ -97,6 +103,153 @@ proc processDial*(
...
+    self: ptr StreamRequest, libp2p: ptr LibP2P
+): Future[Result[string, string]] {.async: (raises: [CancelledError]).} =
+  defer:
+    destroyShared(self)
+
+  let mixProto = libp2p[].mix.valueOr:
+    return err("mix protocol is not mounted")
+
+  if self[].proto.isNil() or self[].proto[0] == '\0':
+    return err("proto is empty")
```
gmelodie, line 116: Do we want to destroy the request here? If so, maybe it's worth it to add more info on the error messages.

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
```diff
@@ -97,6 +103,153 @@ proc processDial*(
...
+  let sizeParam = self[].mixReadBehaviorParam.int
+  case MixReadBehaviorKind(self[].mixReadBehaviorKind)
+  of MIX_READ_EXACTLY:
+    if sizeParam < 0:
+      return err("readExactly size must be >= 0")
+    mixProto.registerDestReadBehavior($self[].proto, readExactly(sizeParam))
+  of MIX_READ_LP:
+    if sizeParam < 0:
+      return err("readLp max size must be >= 0")
+    mixProto.registerDestReadBehavior($self[].proto, readLp(sizeParam))
```
gmelodie, line 125:
```suggestion
  let sizeParam = self[].mixReadBehaviorParam.int
   if sizeParam < 0:
      return err("readExactly size must be >= 0")
  case MixReadBehaviorKind(self[].mixReadBehaviorKind)
  of MIX_READ_EXACTLY:
    mixProto.registerDestReadBehavior($self[].proto, readExactly(sizeParam))
  of MIX_READ_LP:
    mixProto.registerDestReadBehavior($self[].proto, readLp(sizeParam))
```

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
```diff
@@ -97,6 +103,153 @@ proc processDial*(
...
+
+  return ok(handle)
+
+proc processMixSetNodeInfo*(
+    self: ptr StreamRequest, libp2p: ptr LibP2P
+): Future[Result[string, string]] {.async: (raises: [CancelledError]).} =
+  defer:
+    destroyShared(self)
+
+  if self[].multiaddr.isNil() or self[].multiaddr[0] == '\0':
```
gmelodie, line 166: might as well
```suggestion
  if self[].multiaddr.isNil() or self[].multiaddr[0] == '\0' or self[].multiaddr.len() <= 1:
```

gmelodie: reasoning: if `multiaddr[0] != '0'` but `len == 1` means it doesn't have a `\0` terminator (maybe worth checking if it ends in `\0` even if `len > 1`?

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
```diff
@@ -97,6 +103,153 @@ proc processDial*(
...
+    )
+
+    try:
+      await mixProto.start()
+      libp2p[].switch.mount(mixProto)
+    except LPError as exc:
+      return err("could not mount mix: " & exc.msg)
+
+    libp2p[].mix = Opt.some(mixProto)
+    echo "DSAFDSFS"
```
gmelodie: Image

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
```diff
@@ -97,6 +103,153 @@ proc processDial*(
...
+      peerInfo.peerId, nodeAddr, mixPubKey, mixPrivKey, libp2pPubKey, libp2pPrivKey
+    )
+  )
+
+  if libp2p[].mix.isNone:
+    var mixProto = new(MixProtocol)
+    var delayStrategy = NoSamplingDelayStrategy.new(newRng())
+    mixProto.init(
+      libp2p[].mixNodeInfo.get(), libp2p[].switch, delayStrategy = delayStrategy
+    )
```
gmelodie, line 210: Why mount it in this function but fail on `processMixNodePoolAdd`?

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
```diff
@@ -97,6 +103,153 @@ proc processDial*(
...
+  return ok("")
+
+proc processMixNodePoolAdd*(
+    self: ptr StreamRequest, libp2p: ptr LibP2P
+): Future[Result[string, string]] {.async: (raises: [CancelledError]).} =
+  defer:
+    destroyShared(self)
+
+  let mixProto = libp2p[].mix.valueOr:
+    return err("mix protocol is not mounted")
```
gmelodie, line 229: Why not mount it?

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
```diff
@@ -97,6 +103,153 @@ proc processDial*(
...
+
+  if self[].peerId.isNil() or self[].peerId[0] == '\0':
```
gmelodie, line 231: Same as above (check `len <= 1`)

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
```diff
@@ -97,6 +103,153 @@ proc processDial*(
...
+    return err("peerId is empty")
+  if self[].multiaddr.isNil() or self[].multiaddr[0] == '\0':
```
gmelodie, line 233: Same as above

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
gmelodie: Also why is the ordering different? On `processMixSetNodeInfo` this comes last, why?

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
```diff
@@ -102,6 +118,56 @@ proc toByteSeq(key: Libp2pPrivateKey): seq[byte] =
...
+  if config.mountMix != 0 and libp2p.mixNodeInfo.isSome:
+    if config.mixNodesLen <= 0:
+      raise newException(LPError, "failed to create mix protocol: not enough mix nodes")
+    var mixProto = new(MixProtocol)
+    var delayStrategy = NoSamplingDelayStrategy.new(newRng())
+    mixProto.init(
+      libp2p.mixNodeInfo.get(), libp2p.switch, delayStrategy = delayStrategy
+    )
+    libp2p.switch.mount(mixProto)
+    mix = Opt.some(mixProto)
```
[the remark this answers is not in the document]
richard-ramos: Yeah that's fair. 
It seems to me that there's an architecture problem in mix, cause it needs the addresses to initialize the mix node info, yet the addresses are available AFTER the switch starts.... Will remove this path from the startup as suggested, since whoever calls libp2p_start, can also call immediatly set_mix_node_info anyway

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
```diff
@@ -97,6 +103,153 @@ proc processDial*(
...
+
+  let handle = cast[ptr Libp2pStream](createShared(Libp2pStream, 1))
+  handle[].conn = cast[pointer](conn)
+  libp2p[].connections[handle] = conn
+
+  return ok(handle)
+
+proc processMixSetNodeInfo*(
+    self: ptr StreamRequest, libp2p: ptr LibP2P
+): Future[Result[string, string]] {.async: (raises: [CancelledError]).} =
```
vladopajic: return values of this should be `Future[Result[void, string]] ` ?
success value: `string` -> `void`

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
richard-ramos: nim uses C `strlen` for `len` which assumes the cstring is `\0` terminated. If the string does not contain a `\0`, the return value behavior will be undefined. Tbh, probably the safest solution would be to not use cstrings at all but a (ptr, len) instead :thinking:  Maybe something to take into account for all logos core modules? cc:@iurimatias

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
richard-ramos: This is because processMixNodePoolAdd requires libp2p[].mix to be `Some`, which happens once mix_set_node_info is called.

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
richard-ramos: It's done in line 110 :thinking:

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
richard-ramos: good idea

cbind/examples/mix.c
richard-ramos: Will think this thoroughly and fix it in subsequent PR

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
richard-ramos: processMixSetNodeInfo is the function that mounts the mix protocol, since the mixnode info is required to start the protocol. processMixNodePoolAdd mounting mix is not valid as the mixnode info hasnt been set yet

PR #2065, opened 2026-01-30
chore(cbind): add private key

gmelodie:


cbind/libp2p.h
```diff
@@ -48,11 +48,24 @@ typedef struct libp2p_bootstrap_node {
   size_t multiaddrsLen;
 } libp2p_bootstrap_node_t;
 
+typedef enum {
```
richard-ramos: For ffi, use a pattern like this:
```
typedef uint32_t libp2p_pk_scheme;

enum {
  LIBP2P_PK_RSA = 0,
  LIBP2P_PK_ED25519 = 1,
  LIBP2P_PK_SECP256K1 = 2,
  LIBP2P_PK_ECDSA = 3,
};
```
to make it more portable, as enum size might change depending on the compiler / target.

In nim the function parameter would be then a `scheme: uint32` which you would translate back to a nim enum

cbind/libp2p.nim
```diff
@@ -196,6 +196,16 @@ proc libp2p_create_cid(
   callback(RET_OK.cint, addr cidStr[0], cast[csize_t](len(cidStr)), userData)
   RET_OK.cint
 
+proc libp2p_new_private_key*(
+    scheme: PKScheme
+): Libp2pPrivateKey {.dynlib, exportc, cdecl.} =
```
richard-ramos: Already shared comment via DM, but we should use callbacks here instead of returning a type.

cbind/config.nims
```diff
@@ -1,5 +1,4 @@
 # begin Nimble config (version 2)
---noNimblePath
```
richard-ramos, line 2: This change looks unrelated to the PR. Why is it needed?

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
```diff
@@ -33,6 +33,7 @@ const DefaultDnsResolver = "1.1.1.1:53"
 proc applyConfigDefaults(config: ptr Libp2pConfig): Libp2pConfig =
   var resolved = Libp2pConfig(
     flags: 0'u32,
+    passPrivKey: 0,
```
richard-ramos: needs a LIBP2P_CFG_PRIVATE_KEY which is evaluated in line 48-60. 

I added this `flags` as a way to indicate which flags were set, that way we can populate just the fields we care about in the config and use the default when the flag is not active, for example:
```
  libp2p_config_t cfg1 = {0};
  cfg1.flags =
      LIBP2P_CFG_GOSSIPSUB | LIBP2P_CFG_KAD;
  cfg1.mount_gossipsub = 1;
  cfg1.mount_kad = 1;
```
in here, i'm indicating that mount_gossipsub, and mount_kad are meant to be mounted. So, if we dont have a LIBP2P_CFG_PRIVATE_KEY flag, which we add in `cfg1.flags` it means that the nim code will assume we want to set as private key an `""` (empty) value.

Having said that, i understand it might be a bit unintuitive at first sight. Happy to change this mechanism to something else, but that should probably be done on a separate PR

cbind/config.nims
gmelodie: Not needed per se, but I found it annoying that I had to stash this change every time I had to run `nimble setup`, and it didn't seem to make any difference for me, but maybe I'm wrong

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
gmelodie: What I didn't understand is having the flags *and* `mount_kad`/`mount_gossipsub`, why can't we set those fields directly without having the `flags` (or the other way around: have `flags` and not `mount_kad`/`mount_gossipsub`

richard-ramos: This is because at the moment I am/was not sure which libp2p protocols we would mount by default. Using the flag would allow the bindings to use the default, while having just mount_* means that if you dont specify one of those variables, the module will not be mounted.

But, this can likely change assuming that no mount* feature is meant to be true by default. If that's the case, having just the variable and not the flag is more than enough. The flags can be kept for other values who have valid defaults.

cbind/libp2p.h
```diff
@@ -56,6 +69,8 @@ typedef struct {
   const char *dns_resolver;
   const libp2p_bootstrap_node_t *kad_bootstrap_nodes;
   size_t kad_bootstrap_nodes_len;
+  int pass_priv_key;
```
richard-ramos: This field is not really necessary, you can use the flags field, otherwise you'd need a `pass_*` field for each one of the configs.

You would use instead `LIBP2P_CFG_PRIVATE_KEY` by doing  
cfg.flags = .... | LIBP2P_CFG_PRIVATE_KEY to indicate you are passing a private key

cbind/libp2p.nim
```diff
@@ -129,11 +129,12 @@ if defined(android):
...
       echo logLevel, msg
 
-# Initializes the Nim runtime and foreign-thread GC
 proc initializeLibrary() {.exported.} =
+  ## Initializes the Nim runtime and foreign-thread GC
   if not initialized.exchange(true):
-    ## Every Nim library must call `<prefix>NimMain()` once
+    # Every Nim library must call `<prefix>NimMain()` once
     libp2pNimMain()
+    randomize() # initialize rng
```
richard-ramos: We do not use the pnrgn from the std lib but the one from bearssl

cbind/ffi_types.nim
```diff
@@ -9,6 +9,9 @@
 ################################################################################
 ### Exported types
 
+type Libp2pPrivateKey* = object
+  data*: pointer
+
```
richard-ramos, line 15: It is a good practice to have a dataLen, otherwise you would not know how much data to read from this pointer

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
```diff
@@ -98,7 +101,14 @@ proc parseBootstrapNodes(config: Libp2pConfig): seq[(PeerId, seq[MultiAddress])]
 proc createLibp2p(appCallbacks: AppCallbacks, config: Libp2pConfig): LibP2P =
   let dnsResolver =
     Opt.some(cast[NameResolver](DnsResolver.new(@[initTAddress($config.dnsResolver)])))
-  let switch = newStandardSwitch(nameResolver = dnsResolver)
+
+  var privKey = Opt.none(PrivateKey)
+  if config.passPrivKey != 0:
```
richard-ramos: passPrivKey is not necessary, as you have already a flag in config.flags that tells you if you are passing a private key or not.

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
```diff
@@ -98,7 +101,14 @@ proc parseBootstrapNodes(config: Libp2pConfig): seq[(PeerId, seq[MultiAddress])]
...
+    let src = cast[ptr PrivateKey](config.privKey.data)
```
richard-ramos: This cast is incorrect, as getRawBytes (used when generating the key), is the scheme specific serialization, so that data wont match the the struct layout.
What you should likely do is either use getBytes(), so it's serialized into a libp2p protobuf format, and then use PrivateKey.init, passing the serialized data

cbind/libp2p.nim
```diff
@@ -196,6 +196,38 @@ proc libp2p_create_cid(
...
+  ## Creates a new private key
+
+  initializeLibrary()
+
+  let key = PrivateKey.random(scheme, newRng()[]).valueOr:
+    echo "Could not generate private key"
+    return RET_ERR.cint
+
+  let keyData = key.getBytes().valueOr:
+    echo "Could not get raw bytes for private key"
```
richard-ramos:
```suggestion
    echo "Could not get bytes for private key"
```

cbind/examples/cbindings.c
```diff
@@ -58,11 +63,16 @@ int main(int argc, char **argv) {
 
   libp2p_config_t cfg1 = {0};
   cfg1.flags =
-      LIBP2P_CFG_GOSSIPSUB | LIBP2P_CFG_GOSSIPSUB_TRIGGER_SELF | LIBP2P_CFG_KAD;
+      LIBP2P_CFG_GOSSIPSUB | LIBP2P_CFG_GOSSIPSUB_TRIGGER_SELF | LIBP2P_CFG_KAD | LIBP2P_CFG_PRIVATE_KEY;
   cfg1.mount_gossipsub = 1;
   cfg1.gossipsub_trigger_self = 1;
   cfg1.mount_kad = 1;
 
+  libp2p_private_key_t priv_key = {0};
```
richard-ramos, line 71: the priv_key->data is not freed during cleanup

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
richard-ramos: In lines 106 to 198, it should be something like this:
```
let data = newSeqUninit[T](config.privKey.dataLen)
copyMem(addr data[0], addr config.privKey.data[0], config.privKey.dataLen)
let src = PrivateKey.init(data)
privKey = Opt.some(src)
```

but if you want to avoid this boilerplate here, you can make a proc for `Libp2pPrivateKey`  that returns you a sequence

richard-ramos (approved):

PR #2066, opened 2026-02-02
fix(mix): align tag management and add purging of old tags

chaitanyaprem:
- [x] As per [spec](https://lip.logos.co/ift-ts/raw/mix.html#861-shared-preprocessing) Spec (Section 8.6.1, Step 2) tags are suppsosed to be computed and stored as hash of shared secret. Currently raw secret is being stored, hence modifying it to match the spec.
- [x] There is also no purging of old tags which would just keep on increasing memory usage of the process. 
Added a purging mechanism with default purge interval of 1 hour.

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -2,11 +2,16 @@
 if dirExists("nimbledeps/pkgs2"):
   switch("NimblePath", "nimbledeps/pkgs2")
 
+switch("warningAsError", "BareExcept:on")
+switch("warningAsError", "CaseTransition:on")
+switch("warningAsError", "CStringConv:on")
+switch("warningAsError", "ImplicitDefaultValue:on")
+switch("warningAsError", "LockLevel:on")
+switch("warningAsError", "ObservableStores:on")
 switch("warningAsError", "UnusedImport:on")
 switch("warningAsError", "UseBase:on")
-switch("warning", "CaseTransition:off")
-switch("warning", "ObservableStores:off")
-switch("warning", "LockLevel:off")
+switch("hintAsError", "ConvFromXtoItselfNotNeeded:on")
+switch("hintAsError", "DuplicateModuleImport:on")
 
 --styleCheck:
   usages


tests/libp2p/mix/test_tag_manager.nim
```diff
@@ -1,65 +1,96 @@
...
+import chronos
+import ../../../libp2p/protocols/mix/tag_manager
+import ../../tools/unittest
+
+proc makeTag(seed: byte): Tag =
+  ## Helper to create a tag with a specific seed pattern
+  for i in 0 ..< result.len:
+    result[i] = byte((int(seed) * 32 + i) mod 256)
 
 suite "tag_manager_tests":
```
gmelodie:
```suggestion
suite "Tag Manager":
```

tests/libp2p/mix/test_tag_manager.nim
```diff
@@ -1,65 +1,96 @@
...
-      tag = generateRandomFieldElement().expect("should generate FE")
-      nonexistentTag = generateRandomFieldElement().expect("should generate FE")
+  test "add, check, and remove tags":
+    let tag1 = makeTag(1)
+    let tag2 = makeTag(2)
 
-    tm.addTag(tag)
+    # Initially empty
+    check tm.len == 0
+    check not tm.isTagSeen(tag1)
```
gmelodie, line 31:
```suggestion
    check not tm.isTagSeen(tag1)
    check not tm.isTagSeen(tag2)
```

libp2p/protocols/pubsub/timedcache.nim
```diff
@@ -86,6 +87,10 @@ func put*[K](t: var TimedCache[K], k: K, now = Moment.now()): bool =
   t.expire(now)
   t.ensureSizeBound()
 
+  # If not refreshing and already present, return early
+  if not t.refreshOnPut and k in t:
```
gmelodie: Nit: maybe use other names (reading `k in t` feels weird)

libp2p/protocols/mix/tag_manager.nim
```diff
@@ -1,31 +1,91 @@
...
+    tm.start()
+  tm
+
+proc addTag*(tm: TagManager, tag: Tag, now: Moment = Moment.now()) =
+  ## Add a tag to the manager. If already present, this is a no-op
+  ## (does not refresh expiry - first seen time is what matters for replay protection).
+  discard tm.cache.put(tag, now)
+
+proc isTagSeen*(tm: TagManager, tag: Tag): bool {.inline.} =
+  ## Check if a tag has been seen (and hasn't expired).
```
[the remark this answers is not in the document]
richard-ramos: This sounds like a good suggestion, wdyt, @chaitanyaprem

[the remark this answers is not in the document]
chaitanyaprem: This would make the check isTagSeen very heavy as we are running expiry which is traversing through linked list during replay check. This would introduce un-necessary delay during normal processing and better to avoid it.

Issue #1822, opened 2025-10-30
mix: Use peerStore to source mix nodes from the pool rather than separate pool


chaitanyaprem:
Rather than maintaining a separate pool of mix nodes, maybe it would make better sense to use peerStore by adding a new `MixPubKeyBook` to store node's mix public key. 
Similar implementation has been done in nwaku https://github.com/waku-org/nwaku/pull/3617, but would make better sense to make it part of mix protocol itself. 

This way whatever discovery mechanism is used to find mix nodes they just get added to the peerStore along with MixKey and we can remove API's like `setMixNodePool` from mix protocol. 

Thanks @jm-clius for this suggestion.

PR #2067, opened 2026-02-03
feat: implement a node pool and use mix node info from peerstore

chaitanyaprem:
implementing https://github.com/vacp2p/nim-libp2p/issues/1822

Also added a NodePool Mananger which in future can be enhanced to add local reputation/scoring probably.


libp2p/protocols/mix/pool.nim
```diff
@@ -0,0 +1,83 @@
...
+  ## MixPubKeyBook entry is always updated.
+  ## Address is added to AddressBook if not already present.
+  ## KeyBook is only set if not already present (to avoid overwriting
+  ## keys set by the Identify protocol).
+  pool.peerStore[MixPubKeyBook][info.peerId] = info.mixPubKey
+
+  # Add address if not already present
+  let existingAddrs = pool.peerStore[AddressBook][info.peerId]
+  if info.multiAddr notin existingAddrs:
+    pool.peerStore[AddressBook][info.peerId] = existingAddrs & @[info.multiAddr]
```
gmelodie, line 46: Come to think of it, should the addressBook be a `HashSet[MultiAddress]`? Or a `HeapQueue[MultiAddress]` if we need ordering (cc @richard-ramos)

libp2p/protocols/mix/pool.nim
```diff
@@ -0,0 +1,83 @@
...
+
+  # Only set key if peer has no key yet
+  let existingKey = pool.peerStore[KeyBook][info.peerId]
+  if existingKey.scheme == PKScheme.default:
+    pool.peerStore[KeyBook][info.peerId] =
+      PublicKey(scheme: Secp256k1, skkey: info.libp2pPubKey)
```
gmelodie, line 52: I don't understand how this part implements what the comment says, shouldn't it be something like the following instead?
```suggestion
  # Only set key if peer has no key yet
  if info.peerId not in pool.peerStore[KeyBook]:
    pool.peerStore[KeyBook][info.peerId] =
      PublicKey(scheme: Secp256k1, skkey: info.libp2pPubKey)
```
or
```suggestion
  # Only set key if peer has no key yet
  let existingKey = pool.peerStore[KeyBook][info.peerId]
  if existingKey.scheme != Secp256k1:
    pool.peerStore[KeyBook][info.peerId] =
      PublicKey(scheme: Secp256k1, skkey: info.libp2pPubKey)
```

libp2p/protocols/mix/pool.nim
```diff
@@ -0,0 +1,83 @@
...
+
+proc remove*(pool: MixNodePool, peerId: PeerId): bool =
+  ## Remove a mix node from the pool. Returns true if the node was present.
+  result = pool.peerStore[MixPubKeyBook].del(peerId)
```
gmelodie: We avoid the `result` keyword for IFT style guide reasons

gmelodie (approved):

PR #2069, opened 2026-02-03
chore(mix): run delay in background with proof generation

chaitanyaprem:
This is an optimization that works well when DoS/spam protection is enabled for intermediate hops.
This would ensure delay set at each hop and proof generation run in parallel. 

Note that by doing this the delay should be set to something that would take more time than proof generation, otherwise delay of packet would become constant (proof generation time) at each hop which defeats the purpose of random delays.

Thanks for the suggestion @jm-clius

### DIFF
--- a/libp2p/protocols/mix/mix_protocol.nim
+++ b/libp2p/protocols/mix/mix_protocol.nim
@@ -7,13 +7,16 @@ import
   ./[
     curve25519, fragmentation, mix_message, mix_node, sphinx, serialization,
     tag_manager, mix_metrics, exit_layer, multiaddr, exit_connection, spam_protection,
-    delay_strategy,
+    delay_strategy, pool,
   ]
 import stew/endians2
 import ../protocol
 import ../../utils/sequninit
 import ../../stream/[connection, lpstream]
 import ../../[switch, multicodec, peerinfo, varint]
+import ../../peerstore
+
+export pool
 
 when defined(enable_mix_benchmarks):
   import ./benchmark
@@ -41,8 +44,8 @@ type
 ## anonymize messages without modifying their core protocol behavior.
 type MixProtocol* = ref object of LPProtocol
   mixNodeInfo: MixNodeInfo
-  pubNodeInfo: Table[PeerId, MixPubInfo]
   switch: Switch
+  nodePool*: MixNodePool
   tagManager: TagManager
   exitLayer: ExitLayer
   rng: ref HmacDrbgContext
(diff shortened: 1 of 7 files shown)


libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -390,9 +389,23 @@ proc handleMixMessages(
...
 
     # Per-hop spam protection: Generate fresh proof for next hop and append
-    let outgoingPacket = mixProto.generateAndAppendProof(
-      processedSP.serializedSphinxPacket, "Intermediate"
-    ).valueOr:
+    # Run proof generation in parallel with delay to optimize latency.
+    # Note: The effective delay becomes max(configuredDelay, proofGenTime).
+    # If proof generation takes longer than the configured delay, the delay
+    # will be constant at proofGenTime. Implementers should ensure configured
+    # delays are >= expected proof generation time for variable delays to work.
```
vladopajic, line 387: would it make sense to add a log in case when proof generation takes longer then delay, so that person can see and tweak config?

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -390,9 +389,23 @@ proc handleMixMessages(
...
+    var outgoingPacketResult: Result[seq[byte], string]
+    let proofGenFut = (
+      proc(): Future[void] {.async.} =
+        outgoingPacketResult = mixProto.generateAndAppendProof(
```
vladopajic: can `proofGenFut` return this result instead of assigning it to a variable?

libp2p/protocols/mix/mix_protocol.nim
chaitanyaprem: Good idea, will add a warning log.

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -390,9 +389,30 @@ proc handleMixMessages(
...
+    # delays are >= expected proof generation time for variable delays to work.
+    let proofGenStartTime = Moment.now()
+    let proofGenFut = (
+      proc(): Future[Result[seq[byte], string]] {.async.} =
+        return mixProto.generateAndAppendProof(
+          processedSP.serializedSphinxPacket, "Intermediate"
+        )
+    )()
+    let delayFut = sleepAsync(milliseconds(actualDelayMs.int))
+
```
vladopajic, line 397: do first:
`let delayFut = sleepAsync(milliseconds(actualDelayMs.int))`

then this future yields to current thread;
then current thread runs this

```
 let proofGenFut = (
      proc(): Future[Result[seq[byte], string]] {.async.} =
        return mixProto.generateAndAppendProof(
          processedSP.serializedSphinxPacket, "Intermediate"
        )
    )()
```

then at some time `proofGenFut` will finish, and then `delayFut` will be ether finished or would need more time to wait

vladopajic: note, if `sleepAsync` is done after `proofGenFut`, it has high changes of being sequential, unless `proofGenFut` yield at some point.

vladopajic: `generateAndAppendProof` will never yield, so one must do `sleepAsync` first, then `proofGenFut`. because if sleep's timer is not started first then it will be started after proof is generated, as this future will never yields.

chaitanyaprem: I do like this suggestion, but as per [this](https://github.com/vacp2p/nim-libp2p/pull/2069#issuecomment-3842028217) looks like a threadpool is indeed required.

PR #2070, opened 2026-02-03
test(mix): refactor component tests

rlve:
Changes:

- refactor Mix component tests:
  - organize common utilities for setting up Mix nodes
  - remove duplicated code and simplify scenarios
  - remove shared variables

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/extension_test.nim
+++ b/libp2p/protocols/pubsub/gossipsub/extension_test.nim
@@ -7,13 +7,19 @@ import ./[extensions_types]
 
 type
   TestExtensionConfig* = object
-    onNegotiated*: PeerCallback
+    onNegotiated*: proc(peer: PeerId) {.gcsafe, raises: [].}
+      # called when this extensions has negotiated with the peer.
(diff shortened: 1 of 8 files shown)


gmelodie (approved):

tests/libp2p/mix/test_conn.nim
```diff
@@ -158,27 +98,20 @@ suite "Mix Protocol":
     check data == await nrProto.receivedMessages.get()
 
   when defined(libp2p_mix_experimental_exit_is_dest):
```
vladopajic, line 100: nit:

`when defined(libp2p_mix_experimental_exit_is_dest):`
if this line was inside the test, then all tests would have been on the same indentation level.

```
asyncTest ...
    when not defined(libp2p_mix_experimental_exit_is_dest):
       skip() 
       return
     
     test code here...
```

vladopajic: but, not sure if possible... :)

rlve: I think I'll leave it as it is, not sure if I like to have this check inside the test.

PR #2071, opened 2026-02-03
chore: run cbind examples and a test_all job under AddressSanitizer

gmelodie:
## Summary

Runs the C bindings examples and a dedicated `test_all` job under AddressSanitizer, so a memory error in the generated C bindings or in the vendored C fails CI instead of reaching a downstream user.

The cbind half is the reason this PR exists. The examples and the generated bindings pass buffers across the FFI boundary through C `malloc`, which is where a use-after-free or an overflow actually lives. `nimble examples` now builds the FFI library and every example with `-fsanitize=address`, including the vendored TinyCBOR objects that parse the binding payloads.

The test half needed one non-obvious decision. ASan cannot see Nim heap objects under the repository default of `--mm:refc`, because refc sub-allocates from its own arena and there are no per-object redzones. Measured against a deliberate `alloc` / `dealloc` use-after-free:

| memory management | `-d:useMalloc` | ASan report |
| --- | --- | --- |
| refc | no | none |
| refc | yes | none |
| orc | no | none |
| orc | yes | `heap-use-after-free` |

Only `--mm:orc -d:useMalloc` routes each object through `malloc`, so that is what the new `test-asan` job builds. It is a separate job rather than a flag on the existing matrix: six of the seven existing configurations are refc, where the flags would cost roughly 2x runtime and detect nothing on the Nim heap.

`ASAN_OPTIONS=detect_leaks=0` disables LeakSanitizer on that job. orc frees at collection time, so every object still live at exit is reported as leaked, which buries the memory-corruption reports this job exists to surface.

`make test ASAN=1` reproduces the job locally. It stays off by default because the daily `make test_integration` runs two Windows configurations and mingw gcc ships no ASan runtime. For the same reason `nimble examples` defaults to no sanitizer on Windows; `LIBP2P_SAN=asan|tsan|none` selects one explicitly, and the C bindings CI job sets it.

The cbindings job also runs the examples under ThreadSanitizer, with a suppression for the benign nimcrypto SHA-2 dispatch race (every thread writes the same pointer).

## Affected Areas

- [x] Build / Tooling
- [x] Core (fixes for what the sanitized builds reported)

Tooling: the new `test-asan` job in `.github/workflows/test_all.yml`, the tsan matrix in `.github/workflows/cbindings.yml`, the opt-in `ASAN=1` switch in `Makefile`, the sanitized `examples` task in `cbind/cbind.nimble`, and `--noNimblePath` in `config.nims` so the sanitized test build resolves deps the way `nimble setup` writes them.

Fixes the sanitizers reported once they could see the Nim heap:

- `peerstore.cleanup` passed a reference into `toClean[0]` to `del`, which shifts `toClean` and frees the slot under the reference. It now moves the id out before the delete.
- `now()` calls `localtime_r`, which TSan reports as a race under `--threads:on`. Relay, kademlia, autorelay and the TLS certificate check now use `getTime()`; each of them only needed UTC.
- `test_autotls` awaited `downloadCertificate` as the tail expression, so the deferred cleanup ran before the value was read. It now binds the result first.


richard-ramos (approved):

vladopajic:
@richard-ramos this pr needs ci required checks to adjust

remove:
```
Build C bindings Expected 
```
add:
```
C bindings / Build C bindings (asan)
C bindings / Build C bindings (tsan)
```

vladopajic (approved):

richard-ramos:
Done. Will add those as soon as this PR is merged

PR #2078, opened 2026-02-04
chore(mix): add ipv4 address filtering while returning from node pool

chaitanyaprem:


libp2p/protocols/mix/pool.nim
```diff
@@ -7,17 +7,32 @@
...
 import ../../crypto/curve25519
 import ./mix_node
 
 export mix_node.MixPubInfo
 
+const
+  # Mix protocol only supports IPv4 with TCP or QUIC-v1
+  TCP_IP4 = mapAnd(IP4, mapEq("tcp"))
+  UDP_IP4 = mapAnd(IP4, mapEq("udp"))
+  QUIC_V1_IP4 = mapAnd(UDP_IP4, mapEq("quic-v1"))
```
richard-ramos: wdyt about moving these to libp2p/multiaddress.nim so they can be reused if needed?

libp2p/protocols/mix/pool.nim
```diff
@@ -7,17 +7,32 @@
...
+
+func getSupportedMultiaddr(maddrs: seq[MultiAddress]): Option[MultiAddress] =
+  ## Returns the first multiaddress that is supported by the mix protocol.
```
richard-ramos, line 31: I wonder if this should also attempt to obtain the most external address? i.e., assuming the sequence of multiaddresses contains these values in this order: 127.0.0.1,  some private network addr, some external reachable addr.

libp2p/protocols/mix/pool.nim
gmelodie: +1

libp2p/protocols/mix/pool.nim
chaitanyaprem: Hmm, that would be good to do. I remember we did that in go-waku at some point. 
I am not sure if it would impact simulation environments where we want 127 address to be used. 

I would like to analyze the impact and then take up as a separate change.

Also that reminds me, maybe we should not pick any address but just pick address that is part of `LastSeenOutBook` to ensure it is reachable.

chaitanyaprem: ~Now we are using LastSeenOutBoundBook , i guess the most external address only gets picked as that would have been used to reach the peer last.~

PR #2082, opened 2026-02-05
test(tools): add lifecycle and topology

rlve:
This PR introduces reusable test utilities for managing node lifecycle and network topologies. It adds two new test tools:

`lifecycle.nim` - Provides generic functions for starting and stopping test nodes.

`topology.nim` - Offers utilities for connecting test nodes in common network patterns (star, hub, chain etc.).

The changes refactor test files across Kademlia DHT, Rendezvous, Kademlia Discovery, Mix protocol, and GossipSub component tests to use these new utilities. This consolidates previously duplicated helper code.

## Bug Fix - Kademlia Bootstrap Race Condition:

This PR also fixes a Nim assertion defect in the Kademlia bootstrap function. The issue was uncovered when applying the new test helpers and starting many Kad nodes with bootstrap concurrently. 

During bootstrap, `findNode()` can grow the routing table buckets dynamically, causing the bucket count to change mid-iteration. This triggered Nim's runtime assertion when iterating with `for i, bucket in kad.rtable.buckets`. The fix snapshots the bucket count before iteration to prevent the defect.

### DIFF
--- a/tests/tools/lifecycle.nim
+++ b/tests/tools/lifecycle.nim
@@ -5,6 +5,7 @@
 
 import chronos, sequtils
 import ../../libp2p/switch
+import ./futures
 
 proc startNodes*[T](nodes: seq[T]) {.async.} =
   await allFuturesRaising(nodes.mapIt(it.switch.start()))


tests/libp2p/kademlia/utils.nim
```diff
@@ -147,24 +123,13 @@ proc connectNodes*(kad1, kad2: KadDHT) =
     kad1.switch.peerInfo.addrs
 
 proc connectNodesStar*(nodes: seq[KadDHT]) =
```
vladopajic: i would get rid of these proc here.. they aren't doing anything

tests/tools/topology.nim
```diff
@@ -0,0 +1,117 @@
...
+import chronos
+import ../../libp2p/switch
+
+type
+  AsyncConnectProc*[T] = proc(a, b: T): Future[void] {.async.}
+  SyncConnectProc*[T] = proc(a, b: T)
+
+# Async Topology Builders
+
+proc connectNodesChain*[T](nodes: seq[T], connect: AsyncConnectProc[T]) {.async.} =
```
vladopajic: i think that that we can simplify... 

if there is concept:
```
type
  Connectable = concept a, b
    connect(a, b) is Future[void]
```

and then `connectXXX` utilities should be like this:
```
proc connectNodesRing*[T: Connectable ](nodes: seq[T]) {.async.} =
```
note: `connect: AsyncConnectProc[T]` is unnecessary, as `connect`  is part of concept.


now, every test group, like kad, mix, etc will have to have `connect(a, b)` utility that will full this concept for types (these are already implemented because these are passed down as args `connect: AsyncConnectProc[T]`)

---

to make things simpler, concept should only have async version of proc. and all `SyncConnectProc` should be converted to `async` with different helper `connect` halper:


```
connectSync(a, b) = 
  # connect sync, we have this

connect(a, b) Future[void] {.async.} =
   # connect enables Connectable concept, 
   # it is async and it wlil just call connectSync 
  connectSync(a,b)
```

tests/tools/topology.nim
```diff
@@ -0,0 +1,69 @@
...
+{.push raises: [].}
+
+import chronos
+import ../../libp2p/switch
+
+type Connectable =
+  concept a, b
+      connectNodes(a, b) is Future[void]
+
+proc connectNodesChain*[T: Connectable](nodes: seq[T]) {.async.} =
```
vladopajic: nit: would be bit better to no have `Nodes` in name of this interface, theoretical is should work on anything (not just nodes).

-> `connect(a, b) is Future[void]`
-> `connectChain`, etc..

but if this was done due to some naming collisions, then it's fine...

rlve: It just requires a bit more changes.. Done in https://github.com/vacp2p/nim-libp2p/pull/2082/commits/799b8c5ccc33333b96b31e5211eb4bf5191d21c6

PR #2083, opened 2026-02-05
chore(kad): add k closest to randomRecords proc

SionoiS:
If we don't add the K closest peers when initiating `randomRecords` we might not get any records in small networks.

This PR fixes this.

### DIFF
--- a/libp2p/protocols/kademlia_discovery/randomfind.nim
+++ b/libp2p/protocols/kademlia_discovery/randomfind.nim
@@ -22,13 +22,14 @@ proc randomRecords*(
 
   let randomKey = randomPeerId.toKey()
 
-  let queue = newAsyncQueue[(PeerId, Opt[Message])](disco.config.replication)
+  let queue = newAsyncQueue[(PeerId, Opt[Message])]()
 
   let peers = disco.rtable.findClosestPeerIds(randomKey, disco.config.replication)
   for peer in peers:
-    # ignore errors, queue capacity is the same as peer count
-    let _ = catch:
+    let addRes = catch:
       queue.addFirstNoWait((peer, Opt.none(Message)))
+    if addRes.isErr:
+      error "cannot enqueue peer", error = addRes.error.msg
 
   let findRes = catch:
     await disco.findNode(randomKey, queue)


libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -23,6 +23,15 @@ proc randomRecords*(
   let randomKey = randomPeerId.toKey()
 
   let queue = newAsyncQueue[(PeerId, Opt[Message])]()
+
+  let peers = disco.rtable.findClosestPeerIds(randomKey, disco.config.replication)
+  for peer in peers:
+    # queue cannot be full, we just created it...
+    # we add to the beginning of the queue to reverse the order
+    # since closer peers are lasts
```
vladopajic:
```suggestion
    # add to the beginning of the queue to reverse the order
    # since closer peers are lasts
```

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -23,6 +23,15 @@ proc randomRecords*(
   let randomKey = randomPeerId.toKey()
 
   let queue = newAsyncQueue[(PeerId, Opt[Message])]()
+
+  let peers = disco.rtable.findClosestPeerIds(randomKey, disco.config.replication)
+  for peer in peers:
+    # queue cannot be full, we just created it...
```
vladopajic:
```suggestion
    # ignoring raised errors since queue does not have capacity limitations
```

vladopajic: right ?

gmelodie: Yes it's highly unlikely, but we should still check (idk maybe one day the default queue size becomes super small and the default findNode `k` becomes super large. You never know. Can't hurt

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -23,6 +23,15 @@ proc randomRecords*(
   let randomKey = randomPeerId.toKey()
 
   let queue = newAsyncQueue[(PeerId, Opt[Message])]()
```
vladopajic: queue actually needs to be min heap here.

because each iteration of find node, and findClosestPeerIds will return some values that are ordered (but only for that result), and they are not order when we look into all results in queue

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -23,6 +23,15 @@ proc randomRecords*(
...
+
+  let peers = disco.rtable.findClosestPeerIds(randomKey, disco.config.replication)
```
vladopajic, line 27: why not in this order:
```
 await disco.findNode(randomKey, queue)
 if not enough elements in queue 
  queue add disco.rtable.findClosestPeerIds(randomKey, disco.config.replication)
```
?

libp2p/protocols/kademlia_discovery/randomfind.nim
vladopajic: seems to me like min heap should be used ether way. one should not think how elements are order. they should just insert and get without paying any special attention, like code in this pr is

libp2p/protocols/kademlia_discovery/randomfind.nim
SionoiS: :thinking: this would mean that if we get enough far nodes we don't get close nodes.

I would assume getting more nodes in general is better, no need to limit ourselves in this way.

libp2p/protocols/kademlia_discovery/randomfind.nim
vladopajic: @richard-ramos @gmelodie 

do we want to merge this pr without this, and then fix this later, or?

libp2p/protocols/kademlia_discovery/randomfind.nim
```diff
@@ -22,7 +22,14 @@ proc randomRecords*(
 
   let randomKey = randomPeerId.toKey()
 
-  let queue = newAsyncQueue[(PeerId, Opt[Message])]()
+  let queue = newAsyncQueue[(PeerId, Opt[Message])](disco.config.replication)
```
vladopajic: not sure if we need this capacity imitation here

libp2p/protocols/kademlia_discovery/randomfind.nim
richard-ramos: I'm okay with merging the PR without this as it escapes the scope of the PR. Issue for trackign this: https://github.com/vacp2p/nim-libp2p/issues/2084

libp2p/protocols/kademlia_discovery/randomfind.nim
SionoiS: Isn't it just safer to bound it?

vladopajic: i think it's not, since elements will not be globally sorted in entire queue

vladopajic: even if they are sorted, with min heap, i am not sure if  limitation should be `replication`

PR #2086, opened 2026-02-06
chore: some warning fixes

tersec:
- since `--hints:off` had been specified in the `nimble test` configuration, it hadn't been even processing hints. Removed that; it does show some other hints now, but generally, those `XDeclaredButNotUsed` hints should be addressed, and the build was already so noisy, it largely replaces (but in an actionable way) the warnings my previous PRs got rid of;

- `-d:debug` doesn't do anything in Nim. In particular, it does not cause debug-mode builds; and

- the `include` to `import` changes (not all of the `include`s used, just those two) were triggered by `DuplicateModuleImport`, i.e. when one module `include`s another, `DuplicateModuleImport` considers both `import`s to be in the same module

### DIFF
--- a/libp2p/protocols/mix/mix_protocol.nim
+++ b/libp2p/protocols/mix/mix_protocol.nim
@@ -395,12 +395,13 @@ proc handleMixMessages(
 
     await allFutures(proofGenFut, delayFut)
 
-    let proofGenTimeMs = (Moment.now() - proofGenStartTime).milliseconds
-    if proofGenTimeMs > actualDelayMs.int64:
-      warn "Proof generation time exceeds configured delay",
-        proofGenTimeMs,
-        delayMs = actualDelayMs,
-        hint = "Consider increasing delay to maintain variable timing"
+    if mixProto.spamProtection.isSome():
+      let proofGenTimeMs = (Moment.now() - proofGenStartTime).milliseconds
+      if proofGenTimeMs > actualDelayMs.int64:
+        warn "Proof generation time exceeds configured delay",
+          proofGenTimeMs,
+          delayMs = actualDelayMs,
+          hint = "Consider increasing delay to maintain variable timing"
 
     let outgoingPacket = proofGenFut.value().valueOr:
       error "Failed to generate spam protection proof for next hop", err = error
--- a/libp2p/protocols/mix/sphinx.nim
+++ b/libp2p/protocols/mix/sphinx.nim
@@ -402,7 +402,7 @@ proc processSphinxPacket*(
       ProcessedSphinxPacket(
         status: Intermediate,
         nextHop: address,
-        delayMs: uint16.fromBytes(delay).int,
+        delayMs: uint16.fromBytesBE(delay).int,
         serializedSphinxPacket: sphinxPkt.serialize(),
       )
     )


libp2p.nimble
```diff
@@ -23,12 +23,11 @@ let verbose = getEnv("V", "") notin ["", "0"]
...
 let cfg =
-  " --styleCheck:usages --styleCheck:error" &
-  (if verbose: "" else: " --verbosity:0 --hints:off") & " --skipUserCfg -f" &
-  " --threads:on --opt:speed"
+  " --styleCheck:usages --styleCheck:error" & (if verbose: "" else: " --verbosity:0") &
+  " --skipUserCfg -f" & " --threads:on --opt:speed"
 
 proc runTest(filename: string, moreoptions: string = "") =
-  var compileCmd = nimc & " " & lang & " -d:debug " & cfg & " " & flags
+  var compileCmd = nimc & " " & lang & " " & cfg & " " & flags
```
gmelodie, line 30: I'm not sure we want to remove `debug` for tests?

tersec: `-d:debug` doesn't do anything.

vladopajic: afaik:
- `raiseAssert` do not happen with `-d:release` 
- if developers use `-d:debug` in their source then we don't have those things when running tests. we did have some of those optimizations in `nim-lsquic` but not right now. maybe other libraries are using this flag
- builds are not the same (`-d:release` creates builds that are much faster than `-d:debug`, so there are some things added there with debug)

config.nims
```diff
@@ -15,6 +15,8 @@ switch("warningAsError", "UseBase:on")
 switch("hintAsError", "ConvFromXtoItselfNotNeeded:on")
 switch("hintAsError", "DuplicateModuleImport:on")
 
+switch("hint", "XCannotRaiseY:off")
```
vladopajic: why off? shouldn't we aim to fix those?

tests/libp2p/muxers/test_yamux.nim
```diff
@@ -3,12 +3,11 @@
 
 {.used.}
 
-import sugar
-import ../../../libp2p/[stream/bridgestream, muxers/yamux/yamux]
+import std/[sugar, tables]
+import chronos
+import ../../../libp2p/[stream/bridgestream, muxers/yamux/yamux, peerid]
```
vladopajic: would it be possible to include instead of import ?

config.nims
tersec: They can be trickier, but sure, can remove this. It's a tradeoff of how actionable each warning or hint is -- I've looked into fixing them in `nimbus-eth2` and `nimbus-eth1` and `XCannotRaiseY` are the least amenable, on average.

https://github.com/vacp2p/nim-libp2p/pull/2086/commits/32cfd0881174e0a1390f9519bb364d3e59ae4cf2

tests/libp2p/muxers/test_yamux.nim
tersec: Potentially, but it's not clear to me what the advantage is. But yes, can achieve a similar outcome with `include` if that's particularly important.

I'd generally argue that `import` is more realistic for a test, since generally speaking, non-test code isn't `include`ing either.

libp2p.nimble
tersec: Removing `-d:debug` does not enable `-d:release`.

tersec:
```
$ nim c e.nim 
......................................................................
CC: system/exceptions.nim
CC: std/private/digitsutils.nim
CC: system/dollars.nim
CC: system.nim
CC: e.nim
Hint:  [Link]
Hint: mm: orc; threads: on; opt: none (DEBUG BUILD, `-d:release` generates faster code)
28890 lines; 0.328s; 31.207MiB peakmem; proj: /tmp/e.nim; out: /tmp/e [SuccessX]
$ nim c -d:debug e.nim 
......................................................................
Hint:  [Link]
Hint: mm: orc; threads: on; opt: none (DEBUG BUILD, `-d:release` generates faster code)
28890 lines; 0.148s; 31.207MiB peakmem; proj: /tmp/e.nim; out: /tmp/e [SuccessX]
$ nim c -d:release e.nim 
......................................................................
CC: system/exceptions.nim
CC: std/private/digitsutils.nim
CC: system/dollars.nim
CC: system.nim
CC: e.nim
Hint:  [Link]
Hint: mm: orc; threads: on; opt: speed; options: -d:release
28890 lines; 0.959s; 31.242MiB peakmem; proj: /tmp/e.nim; out: /tmp/e [SuccessX]
```
It's identical with and without `-d:debug`.

gmelodie: @vladopajic I think @tersec is right, it's just a symbol and unless we're doing `when defined(debug): ...` somewhere (which we aren't) then it's not doing anything

gmelodie (approved):

libp2p.nimble
tersec: Also
> `raiseAssert` do not happen with `-d:release`

is not accurate.

The distinction is that `doAssert` [always triggers an assertion](https://nim-lang.org/docs/assertions.html#doAssert.t%2Cuntyped%2Cstring), but `assert` [doesn't under certain circumstances](https://nim-lang.org/docs/assertions.html#assert.t,untyped,string).

[raiseAssert](https://nim-lang.org/docs/assertions.html#raiseAssert%2Cstring) happens regardless.

vladopajic: got it :+1:

tests/libp2p/muxers/test_yamux.nim
vladopajic: i assume it was included because someone didn't want to make those fields accessible from outside

vladopajic: but since muxes are internal logic to libp2p, users will never touch it, i guess it's not big deal

tersec: Removed those changes: https://github.com/vacp2p/nim-libp2p/pull/2086/commits/b661385d1637d79ea386a93e501fe3ec5a65d91b

Issue #1946, opened 2025-12-09
test: change `discard` to `raiseAssert` if possible


vladopajic:
similarly to https://github.com/vacp2p/nim-libp2p/pull/1942/commits/df56e558dda48b77c875bab6c2e4b0d090b64044, search other places in tests and apply same change. 

if possible change `discard` in handler methods to `raiseAssert`. this gives more hardness to tests.

---- 
advanced:

explore ways in which `raiseAssert` could be used more in handlers. for example node that publishes messages in tests doesn't need same handler as node that receives messages. can publisher here have custom handler with `raiseAssert` ?

PR #2093, opened 2026-02-10
test(mix): add component tests

rlve:
Changes:
- add tests:
  - `"sender receives empty response when destination is unreachable"`
  - `"send fails when pool has not enough nodes"`
  - `"rate limit exceeded - message rejected at intermediate node"`
- simplify `MockMix` setup by changing constructor return type
- move `NoReplyProtocol` to `utils` to be reused between tests

### DIFF
--- a/libp2p.nimble
+++ b/libp2p.nimble
@@ -23,12 +23,11 @@ let verbose = getEnv("V", "") notin ["", "0"]
 
 # changes in run configs should be also reflected on flake.nix
 let cfg =
-  " --styleCheck:usages --styleCheck:error" &
-  (if verbose: "" else: " --verbosity:0 --hints:off") & " --skipUserCfg -f" &
-  " --threads:on --opt:speed"
+  " --styleCheck:usages --styleCheck:error" & (if verbose: "" else: " --verbosity:0") &
+  " --skipUserCfg -f" & " --threads:on --opt:speed"
 
 proc runTest(filename: string, moreoptions: string = "") =
-  var compileCmd = nimc & " " & lang & " -d:debug " & cfg & " " & flags
+  var compileCmd = nimc & " " & lang & " " & cfg & " " & flags
   if getEnv("CICOV").len > 0:
     compileCmd &= " --nimcache:nimcache/" & filename & "-" & $compileCmd.hash
   compileCmd &= " -d:libp2p_autotls_support"
--- a/libp2p/peeridauth/client.nim
+++ b/libp2p/peeridauth/client.nim
@@ -307,7 +307,7 @@ proc sendWithBearer(
 ): Future[(BearerToken, PeerIDAuthResponse)] {.
     async: (raises: [PeerIDAuthError, CancelledError])
 .} =
-  if bearer.expires.isSome() and DateTime(bearer.expires.get) <= now():
+  if bearer.expires.isSome() and bearer.expires.get <= now():
     raise newException(PeerIDAuthError, "Bearer expired")
   let authHeader = PeerIDAuthPrefix & " bearer=\"" & bearer.token & "\""
   let response =
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -18,9 +18,6 @@ import
   ../../utils/sequninit,
   ./bandwidth
 
-when defined(libp2p_gossipsub_1_4):
-  import ./bandwidth
-
 export peerid, connection, deques
 
 logScope:
--- a/tests/libp2p/mix/test_seq_no_generator.nim
+++ b/tests/libp2p/mix/test_seq_no_generator.nim
@@ -3,8 +3,8 @@
 
 {.used.}
 
-import chronicles, sets, std/[os]
-include ../../../libp2p/protocols/mix/seqno_generator
+import sets, std/[os]
+import ../../../libp2p/[peerid, protocols/mix/seqno_generator]
 import ../../tools/[unittest]
 
 const second = 1000
--- a/tests/libp2p/transports/test_ws.nim
+++ b/tests/libp2p/transports/test_ws.nim
@@ -150,7 +150,7 @@ suite "WebSocket transport":
     await transport1.stop()
 
 when defined(libp2p_autotls_support):
-  import ../../../libp2p/[autotls/service, autotls/mockservice]
+  import ../../../libp2p/autotls/mockservice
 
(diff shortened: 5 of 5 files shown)


libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -895,23 +894,24 @@ proc new*(
     rng: ref HmacDrbgContext = newRng(),
     spamProtection: Opt[SpamProtection] = default(Opt[SpamProtection]),
     delayStrategy: Opt[DelayStrategy] = Opt.none(DelayStrategy),
-): Result[T, string] =
+): T =
```
rlve, line 898: Using `Result[T, string]` as a return type in a constructor loses the subtype due to how `typedesc` parameter is resolved through it. It returns `Result[MixProtocol, string]` instead of `Result[MockMixProtocol, string]`.

Fix: return `T` directly instead of `Result[T, string]`.

vladopajic: this is also better, because constructors should be error-less.

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -895,23 +895,24 @@ proc new*(
...
   ## Constructs a new `MixProtocol` instance for the mix node at `index`,
   ## loading its private info from `nodeInfo` and populating the nodePool with
   ## the public info of all other nodes from `pubInfo`.
-  let mixNodeInfo = MixNodeInfo.readFromFile(index, nodeFolderInfoPath / fmt"nodeInfo").valueOr:
-    return err("Failed to load mix node info for index " & $index & " - err: " & error)
+  let mixNodeInfo = MixNodeInfo
+    .readFromFile(index, nodeFolderInfoPath / fmt"nodeInfo")
+    .expect("Failed to load mix node info for index " & $index)
```
vladopajic, line 904: we should refactor this, `mixNodeInfo` should be provided via constructor.

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -895,23 +895,24 @@ proc new*(
...
   # Load pub info into the nodePool
   for i in 0 ..< numNodes:
     if i == index:
       continue
-    let pubInfo = MixPubInfo.readFromFile(i, nodeFolderInfoPath / fmt"pubInfo").valueOr:
-      return err("Failed to load pub info from file: " & error)
+    let pubInfo = MixPubInfo.readFromFile(i, nodeFolderInfoPath / fmt"pubInfo").expect(
+        "Failed to load pub info from file"
+      )
     mixProto.nodePool.add(pubInfo)
```
vladopajic, line 916: similar as above, constructor should not read from file. values need to be provided to constructor.

tests/libp2p/mix/utils.nim
```diff
@@ -123,3 +115,29 @@ proc setupDestNode*[T: LPProtocol](
...
+    try:
+      buffer = await conn.readLp(1024)
+    except LPStreamError:
+      discard
+
+    await conn.close()
+    await nrProto.receivedMessages.put(buffer)
+
+  nrProto.handler = handler
+  nrProto.codec = NoReplyProtocolCodec
```
vladopajic, line 140:
```suggestion
  proc handler(conn: Connection, proto: string) {.async: (raises: [CancelledError]).} =
    try:
     let buffer = await conn.readLp(1024)
     await nrProto.receivedMessages.put(buffer)
    except LPStreamError:
      raiseAssert "shuld not happen"
    await conn.close()

  let nrProto = NoReplyProtocol()
  nrProto.receivedMessages = newAsyncQueue[seq[byte]]()
  nrProto.handler = handler
  nrProto.codec = NoReplyProtocolCodec
```

libp2p/protocols/mix/mix_protocol.nim
rlve: there is another constructor with MixNodeInfo as a parameter

https://github.com/vacp2p/nim-libp2p/blob/aae92b47fba77a2525cc1cccf3b5359980a66cb6/libp2p/protocols/mix/mix_protocol.nim#L869-L877

This one simplifies loading node info from the files.

libp2p/protocols/mix/mix_protocol.nim
vladopajic: https://github.com/vacp2p/nim-libp2p/issues/2095

libp2p/protocols/mix/mix_protocol.nim
vladopajic: then first one that reads from files should not exist. it should call proc:

```
let nodeInfo = readMixNodeInfoFromFile()
MixProtocol.new(nodeInfo .... )
```

etc..

tests/libp2p/mix/test_conn.nim
```diff
@@ -360,3 +336,52 @@ suite "Mix Protocol Component":
...
+        MixParameters(expectReply: Opt.some(true), numSurbs: Opt.some(byte(1))),
+      )
+      .expect("could not build connection")
+    defer:
+      await conn.close()
+
+    await conn.write(@[1.byte, 2, 3, 4, 5])
+
+    let response = await conn.readLp(1024).wait(10.seconds)
+    check response.len == 0
```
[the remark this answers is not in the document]
rlve: This `destReadBehavior` was used to simplify the scenario setup. Dest node is stopped anyway to test the behaviour when destination is unreachable.

libp2p/protocols/mix/mix_protocol.nim
vladopajic: https://github.com/vacp2p/nim-libp2p/issues/2095

tests/libp2p/mix/utils.nim
vladopajic: fyi: https://github.com/vacp2p/nim-libp2p/issues/1946

tests/libp2p/mix/utils.nim
```diff
@@ -123,3 +115,29 @@ proc setupDestNode*[T: LPProtocol](
...
   await switch.stop()
+
+###
+
+const NoReplyProtocolCodec* = "/test/1.0.0"
+
+type NoReplyProtocol* = ref object of LPProtocol
+  receivedMessages*: AsyncQueue[seq[byte]]
+
+proc newNoReplyProtocol*(): NoReplyProtocol =
```
vladopajic: nit: `newNoReplyProtocol` -> `NoReplyProtocol.new`

PR #2102, opened 2026-02-11
chore(mix): clean up mix_node

rlve:
Clean up after https://github.com/vacp2p/nim-libp2p/issues/1822.

Changes:
- clean up `mix_node`:
  - keep types `MixNodeInfo` and `MixPubInfo`
  - move test infrastructure to `tests/mix/utils` and simplify it
  - remove procs for writing/reading from the files with serialization/deserialization and other helpers
  - add `toMixPubInfo` helper to convert `MixNodeInfo`
- remove entirely 2nd `MixProtocol` constructor that was reading `MixNodeInfo` from the files
- add batch add to node pool
- delete `test_mix_node.nim` (no longer needed)
- update tests

### DIFF
--- a/libp2p/protocols/mix/mix_node.nim
+++ b/libp2p/protocols/mix/mix_node.nim
@@ -81,5 +81,5 @@ proc toMixPubInfo*(info: MixNodeInfo): MixPubInfo =
 
 proc includeAllExcept*(
     allNodes: seq[MixNodeInfo], exceptNode: MixNodeInfo
-): seq[MixPubInfo] =
+): seq[MixPubInfo] {.inline.} =
   allNodes.mapIt(it.toMixPubInfo()).filterIt(it.peerId != exceptNode.peerId)


examples/mix_ping.nim
```diff
@@ -35,36 +58,30 @@ proc createSwitch(
...
+
+    # Populate nodePool with all other nodes' public info
+    for otherIndex, otherNode in enumerate(mixNodes):
+      if otherIndex != index:
+        proto.nodePool.add(
+          MixPubInfo.init(
+            otherNode.peerId, otherNode.multiAddr, otherNode.mixPubKey,
+            otherNode.libp2pPubKey,
+          )
+        )
```
vladopajic:
```suggestion
let poolInfo = MixPubInfo.createFrom(mixNodes, mixNodes[index])
let proto = MixProtocol.new(mixNodes[index], switches[index], poolInfo)
```
note: finish the cleanup, pass data via constructor 

```
proc createFrom*(T: typedesc[MixPubInfo], mixNodes: seq[MixNode], exceptNode: mixNode): seq[MixPubInfo]
```
- utility creates `MixPubInfo` using all mixNodes, but it won't include `exceptNode`.
- should be publically accessible via `mix` becasue if it is needed in examples, it will probably be needed in other codebases, and tests, etc...

tests/libp2p/mix/utils.nim
```diff
@@ -68,41 +82,67 @@ proc setupMixNode[T: MixProtocol](
   switch.mount(proto)
   proto
 
+proc populateNodePool*(nodes: seq[MixProtocol], nodeInfos: seq[MixNodeInfo]) =
+  for i in 0 ..< nodes.len:
+    for j in 0 ..< nodeInfos.len:
+      if i == j:
+        continue
+      nodes[i].nodePool.add(nodeInfos[j].toMixPubInfo())
```
vladopajic: see comment above. this will not be needed here.

tests/libp2p/mix/utils.nim
```diff
@@ -68,41 +82,67 @@ proc setupMixNode[T: MixProtocol](
...
+      createSwitch(mixNodeInfo.multiAddr, Opt.some(mixNodeInfo.libp2pPrivKey))
+
     nodes.add(
       setupMixNode[MixProtocol](
-        index, numNodes, switches[index], destReadBehavior, spamProtectionRateLimit
+        mixNodeInfo, switch, destReadBehavior, spamProtectionRateLimit
       )
     )
+
+  populateNodePool(nodes, nodeInfos)
```
vladopajic: do not ` populateNodePool(nodes, nodeInfos)`, instead pass data via `setupMixNode`

tests/libp2p/mix/utils.nim
```diff
@@ -68,41 +82,67 @@ proc setupMixNode[T: MixProtocol](
...
+    let switch =
+      createSwitch(mixNodeInfo.multiAddr, Opt.some(mixNodeInfo.libp2pPrivKey))
     nodes.add(
-      setupMixNode[MixProtocol](
-        index, numNodes, switches[index], destReadBehavior, Opt.none(int)
-      )
+      setupMixNode[MixProtocol](mixNodeInfo, switch, destReadBehavior, Opt.none(int))
     )
 
+  populateNodePool(nodes, nodeInfos)
```
vladopajic: do not ` populateNodePool(nodes, nodeInfos)`, instead pass data via `setupMixNode`

examples/mix_ping.nim
rlve: I think node wouldn't usually know the rest of mix network at the time of creation to populate the node pool. That needs to be done by other discovery means.

I made it a little bit more convenient in https://github.com/vacp2p/nim-libp2p/pull/2102/changes/ab6ecdc55adade5eb5fc30bd9d49ad584aa68a6a

vladopajic: >  wouldn't usually know the rest of mix network at the time of creation

hmm, not sure, but iam guessing it must know, alos usage in this repository indicates that pool is known upfront (it's just matter of how code is being structure).  

anyhow, i am not insisting on this, since this pr as it has made things so much better.

tests/libp2p/mix/utils.nim
```diff
@@ -68,41 +78,67 @@ proc setupMixNode[T: MixProtocol](
   switch.mount(proto)
   proto
 
+proc populateNodePool*(nodes: seq[MixProtocol], nodeInfos: seq[MixNodeInfo]) =
+  for i in 0 ..< nodes.len:
+    let pubInfos = nodeInfos
+      .filterIt(it.peerId != nodes[i].switch.peerInfo.peerId)
+      .mapIt(it.toMixPubInfo())
+    nodes[i].nodePool.add(pubInfos)
```
vladopajic: unnecessary (see other comments for context)

examples/mix_ping.nim
```diff
@@ -35,36 +57,26 @@ proc createSwitch(
...
 
   # Set up mix protocols on each mix node
   for index, _ in enumerate(switches):
-    let proto = MixProtocol.new(index, switches.len, switches[index])
+    let proto = MixProtocol.new(mixNodes[index], switches[index])
+
+    # Populate nodePool with all other nodes' public info
+    let pubInfos = mixNodes.mapIt(it.toMixPubInfo()).filterIt(
+        it.peerId != proto.switch.peerInfo.peerId
+      )
```
vladopajic: consider having utility, not sure about name but say `MixNodeInfo.from(mixNodes, ignoredNode)`

tests/libp2p/mix/utils.nim
```diff
@@ -68,41 +78,67 @@ proc setupMixNode[T: MixProtocol](
...
   for index in 1 ..< numNodes:
+    let mixNodeInfo = generateMixNodeInfo(basePort + index)
+    nodeInfos.add(mixNodeInfo)
+    let switch =
+      createSwitch(mixNodeInfo.multiAddr, Opt.some(mixNodeInfo.libp2pPrivKey))
     nodes.add(
-      setupMixNode[MixProtocol](
-        index, numNodes, switches[index], destReadBehavior, Opt.none(int)
-      )
+      setupMixNode[MixProtocol](mixNodeInfo, switch, destReadBehavior, Opt.none(int))
```
vladopajic, line 99:
```
let switch = createSwitch...
let mixnode = setupMixNode(....)
mixnode.pool.add(     MixNodeInfo.from(mixNodes, ignoredNode = mixnode)`     )
nodes.add(mixnode)

```

note: use same code structure to create nodes everywhere. 
logic in utility `MixNodeInfo.from` seems to be reused across this pr, but existence of `populateNodePool` seems to introduced jet another usecase (unnecessary).
same utility is likely to be used in downstream dependencies.

rlve: I'm not sure about `MixNodeInfo.from` you're proposing. I think consumers wouldn't try to include self when populating node pool with peers, so an utility with `ignoredNode` parameter feels unnecessary. 🤔 

This situation feels purely related to tests/examples in this repo and how they're implemented. 

`populateNodePool` is just a helper for component tests because:
- we create all the switches with mix protocol first
- and then, once we have all the nodes ready, we populate their node pools

libp2p/protocols/mix/mix_node.nim
```diff
@@ -171,151 +76,10 @@ proc init*(
...
 
-proc initMixMultiAddrByIndex*(
-    self: var MixNodes, index: int, peerId: PeerId, multiAddr: MultiAddress
-): Result[void, string] =
-  if index < 0 or index >= self.len:
-    return err("Index must be between 0 and " & $(self.len))
-  self[index].multiAddr = multiAddr
-  self[index].peerId = peerId
-  ok()
+proc includeAllExcept*(
```
richard-ramos, line 82: This can probably be {.inline.}

libp2p/protocols/mix/mix_node.nim
```diff
@@ -35,118 +30,28 @@ proc initMixNodeInfo*(
...
-  let dMixNodeInfo = MixNodeInfo.deserialize(cast[seq[byte]](data)).valueOr:
-    return err("Mix node info deserialize error: " & error)
-  return ok(dMixNodeInfo)
-
-proc deleteNodeInfoFolder*(nodeInfoFolderPath: string = "./nodeInfo") =
-  ## Deletes the folder that stores serialized mix node info files
-  ## along with all its contents, if the folder exists.  
-  if dirExists(nodeInfoFolderPath):
-    removeDir(nodeInfoFolderPath)
+proc generateRandomMany*(
```
richard-ramos, line 48: I wonder if this function should live here and not in some utils file. It's useful for tests and examples but it's not going to be used IRL (i.e. you would likely only have a single switch in your machine, thus only a single MixNodeInfo known when node starts, with the others being discovered).

richard-ramos (approved):

libp2p/protocols/mix/mix_node.nim
vladopajic: utilities should be co-located with the types they operate on, regardless of where they're used. while it might seem counterintuitive to place a testing utility in the same module as production code, this is actually the right approach when the utility needs to be reused. 
these utilities are serving both the module itself and its downstream dependencies - waku can probably reuse this for their tests.

vladopajic: for example, `includeAllExcept` is utility that was needed in tests. but if it is part of same module as types it operates on then technically it can be reused is production code - it is very easy to imagine that there could be such use case.

libp2p/protocols/mix/mix_node.nim
rlve: added in https://github.com/vacp2p/nim-libp2p/pull/2102/changes/3fe40ed4c78543c58e3cdec9970c14cf2b2cf307

PR #2105, opened 2026-02-12
chore(partialmessage): push to selected peers

vladopajic:


libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -305,7 +305,11 @@ proc publishPartialToPeer(
   return hasChanges # aka has published
 
 proc publishPartial*(
-    ext: PartialMessageExtension, topic: string, pm: PartialMessage
+    ext: PartialMessageExtension,
+    topic: string,
+    pm: PartialMessage,
+    peers: seq[PeerId] = @[],
```
richard-ramos, line 311: Should this be a set?

vladopajic: no, by default empty peers mean we don't target any specific peer to push messages to - we push to all peers in topic. 
when peer is specified, we push to selected peers only - to fulfill their request

PR #2107, opened 2026-02-12
test(mix): anonymity + error cases + getMaxMessageSizeForCodec

rlve:
Changes:
- add tests:
  - `"path nodes are non-repeating"`
  - `"toConnection rejects expectReply without destReadBehavior"`
  - `"read from write-only connection raises error"`
  - `"write rejects oversized messages"`
  - `"no response sent back on failure"`
  - `"getMaxMessageSizeForCodec returns correct size"`
  - `"getMaxMessageSizeForCodec errors when overhead exceeds capacity"`

### DIFF
--- a/tests/libp2p/mix/test_conn.nim
+++ b/tests/libp2p/mix/test_conn.nim
@@ -19,7 +19,6 @@ import
     crypto/crypto,
     crypto/secp,
   ]
-from ../../../libp2p/protocols/mix/fragmentation import DataSize
 
 import ../../tools/[lifecycle, unittest]
 import ./utils
@@ -472,8 +471,13 @@ suite "Mix Protocol Component":
     defer:
       await conn.close()
 
+    # Write a message at exactly the maximum allowed size — should succeed
+    let maxMessageSize = getMaxMessageSizeForCodec(nrProto.codec, 0).get()
+    await conn.write(newSeq[byte](maxMessageSize))
+
+    # Write a message one byte over the limit — should be rejected
     expect LPStreamError:
-      await conn.write(newSeq[byte](DataSize + 1))
+      await conn.write(newSeq[byte](maxMessageSize + 1))
 
   asyncTest "no response sent back on failure":
     let nodes = await setupMixNodes(2)


tests/libp2p/mix/test_mix_message.nim
```diff
@@ -35,3 +35,38 @@ suite "mix_message_tests":
...
+    check:
+      sizeLong.isOk
+      sizeLong.get() < size0.get()
+
+  test "getMaxMessageSizeForCodec errors when overhead exceeds capacity":
+    let codec = "/test/1.0.0"
+
+    check:
+      getMaxMessageSizeForCodec(codec, 5).isOk # 5 SURB is max
+      getMaxMessageSizeForCodec(codec, 6).isErr
```
vladopajic: 5 should come from some `const` or wherever source of truth is for it

6 should be `that_const + 1`

tests/libp2p/mix/test_mix_message.nim
```diff
@@ -35,3 +35,38 @@ suite "mix_message_tests":
...
+    check:
+      size2.isOk
+      size2.get() == size0.get() - 2 * surbOverhead
+
+    # A longer codec should return a smaller max message size
+    let longCodec = "/test/with/a/much/longer/codec/name/1.0.0"
+    let sizeLong = getMaxMessageSizeForCodec(longCodec, 0)
+    check:
+      sizeLong.isOk
+      sizeLong.get() < size0.get()
```
vladopajic, line 63:
```
sizeLong.get() < size0.get()
sizeLong.get() + longCodec.len ==  size0.get() 
```
is last check true? if so would be nice to harden assertation

tests/libp2p/mix/utils.nim
```diff
@@ -119,15 +119,18 @@ const NoReplyProtocolCodec* = "/test/1.0.0"
 
 type NoReplyProtocol* = ref object of LPProtocol
   receivedMessages*: AsyncQueue[seq[byte]]
+  connPeerIds*: AsyncQueue[PeerId]
```
vladopajic:
```
type ProtoMessage* = object  # note sure about name....
   peerId: PeerId # or should be senderPeerId or # receiverPeerId - what is true? or should have both properties
   data: seq[byte]
receivedMessages: AsyncQueue[ProtoMessage]
``` 
use single structure for all data

tests/libp2p/mix/test_conn.nim
vladopajic:
```suggestion
      .toConnection(
        MixDestination.init(destNode.peerInfo.peerId, destNode.peerInfo.addrs[0]),
       nrProto.codec,
      )
```

vladopajic:
```
let conn = nodes[0].toConnDest(destNode) 
# or
let conn = nodes[0].toConnDest(destNode, nrProto)  
# if mix proto is not accessible via destNode
```

this should be one-liner with such utility

tests/libp2p/mix/test_conn.nim
```diff
@@ -71,6 +73,53 @@ suite "Mix Protocol Component":
 
     check data == await nrProto.receivedMessages.get()
 
+    # assert anonymity of the sender
+    let connPeerId = await nrProto.connPeerIds.get()
+    check:
+      connPeerId != nodes[0].switch.peerInfo.peerId # not the sender
+      connPeerId != destNode.peerInfo.peerId # not the destination itself
+      connPeerId in nodes.mapIt(it.switch.peerInfo.peerId)
```
vladopajic: would make sense to have more anonymity assertions, if applicable (iam not sure)

tests/libp2p/mix/test_conn.nim
```diff
@@ -71,6 +73,53 @@ suite "Mix Protocol Component":
...
+    let nodes = await setupMixNodes(10)
+    startAndDeferStop(nodes)
+
+    let (destNode, nrProto) = await setupDestNode(NoReplyProtocol.new())
+    defer:
+      await stopDestNode(destNode)
+
+    # Send multiple messages and track which mix node delivered each one
+    const numMessages = 20
+    var exitNodes: seq[PeerId]
```
vladopajic: `var exitNodes: seq[PeerId]` 
should alredy be
`var exitCounts: Table[PeerId, int]`

tests/libp2p/mix/test_conn.nim
```diff
@@ -71,6 +73,53 @@ suite "Mix Protocol Component":
...
+          MixDestination.init(destNode.peerInfo.peerId, destNode.peerInfo.addrs[0]),
+          NoReplyProtocolCodec,
+        )
+        .expect("could not build connection")
+
+      await conn.writeLp(@[byte(i)])
+      await conn.close()
+
+      discard await nrProto.receivedMessages.get().wait(5.seconds)
+      exitNodes.add(await nrProto.connPeerIds.get().wait(5.seconds))
```
vladopajic: `exitNodes` are not peer ids of exit node, they are peer ids of node that sent message to exit node?

tests/libp2p/mix/test_conn.nim
```diff
@@ -386,3 +435,83 @@ suite "Mix Protocol Component":
 
     expect LPStreamError:
       await conn.writeLp(@[1.byte, 2, 3])
+
+  asyncTest "toConnection rejects expectReply without destReadBehavior":
+    let nodes = await setupMixNodes(10) # no destReadBehavior registered
+    startAndDeferStop(nodes)
+
+    let (destNode, _) = await setupDestNode(Ping.new())
```
vladopajic: why `Ping` here?

tests/libp2p/mix/test_conn.nim
```diff
@@ -386,3 +435,83 @@ suite "Mix Protocol Component":
...
+    startAndDeferStop(nodes)
+
+    let (destNode, _) = await setupDestNode(NoReplyProtocol.new())
+    defer:
+      await stopDestNode(destNode)
+
+    let conn = nodes[0]
+      .toConnection(
+        MixDestination.init(destNode.peerInfo.peerId, destNode.peerInfo.addrs[0]),
+        NoReplyProtocolCodec, # no expectReply, connection is write-only
```
vladopajic: i dont get how this code is different then:
 https://github.com/vacp2p/nim-libp2p/pull/2107/changes#diff-45d84a3a3afe6ec2481aca98cd6a72a1554a72cbff8d919843b8993f471378d1R96
and this code does not raise LPStreamError?

tests/libp2p/mix/test_conn.nim
```diff
@@ -386,3 +435,83 @@ suite "Mix Protocol Component":
...
+    let targetAddr = nodes[1].switch.peerInfo.addrs[0]
+
+    # Dial the mix node directly
+    let conn = await nodes[0].switch.dial(targetPeerId, @[targetAddr], @[MixProtocolID])
+    defer:
+      await conn.close()
+
+    # Send a corrupted packet
+    let corruptedPacket = newSeq[byte](PacketSize)
+    await conn.writeLp(corruptedPacket)
```
vladopajic:
```suggestion
    # Send a "corrupted packet" - raw bytes that are not initiliazed according to the MixMessage structure
    await conn.writeLp(newSeq[byte](PacketSize))
```

tests/libp2p/mix/test_conn.nim
```diff
@@ -386,3 +435,83 @@ suite "Mix Protocol Component":
...
+
+    # Wait briefly to give the mix node time to process
+    # Then try to read — expect timeout because no bytes should come back
+    expect AsyncTimeoutError:
+      var buf = newSeq[byte](1)
+      await conn.readExactly(addr buf[0], 1).wait(1.seconds)
```
vladopajic: why should any bytes come back in first place? it's the destination node that should receive anything?

tests/libp2p/mix/test_mix_message.nim
rlve: It is. `sizeLong` means size of the message for a long codec. A longer codec should return a smaller max message size.

tests/libp2p/mix/utils.nim
rlve: I think `connPeerId` is the most clear as in the `PeerId` that delivered the message via the connection => which would be the `PeerId` of the exit node in this case. 

`senderPeerId` could get confused with the original sender of the message.

tests/libp2p/mix/test_conn.nim
rlve: I don't think we can assert anything else without implementing some way to inspect the mix path.

tests/libp2p/mix/test_conn.nim
rlve: `exitNode` != `destinationNode`. Exit node is the last node in the Mix path, that delivers the message to the destination.

rlve: In this test we basically assert that the `exitNode` is random and changes across 20 messages.

tests/libp2p/mix/test_conn.nim
rlve: The comment explains, we don't set up 3rd parameter (`expectReply`) of `toConnection`, which makes it write-only.

We get `LPStreamError` because we try to read from the connection.

In other cases, that are using `NoReplayProtocol` we don't try to read from the connection, that's the difference.

tests/libp2p/mix/test_conn.nim
rlve: It was supposed prove that `MixProtocol` doesn't send any information back in case of a failure - it's about adhering to the spec.

tests/libp2p/mix/utils.nim
```diff
@@ -113,21 +113,33 @@ proc setupDestNode*[T: LPProtocol](
 proc stopDestNode*(switch: Switch) {.async.} =
   await switch.stop()
 
+proc toMixDestination*(switch: Switch): MixDestination =
+  MixDestination.init(switch.peerInfo.peerId, switch.peerInfo.addrs[0])
+
+proc toMixDestination*(node: MixProtocol): MixDestination =
+  MixDestination.init(node.switch.peerInfo.peerId, node.switch.peerInfo.addrs[0])
```
vladopajic:
```suggestion
proc toMixDestination*(node: MixProtocol): MixDestination =
  node.switch.toMixDestination()
```

nit :)

gmelodie: It'd be good to have these helpers defined outside of tests and used in
- `docs/protocols_mix.md`
- `examples/mix_ping.nim`

tests/libp2p/mix/test_conn.nim
```diff
@@ -58,18 +60,54 @@ suite "Mix Protocol Component":
...
 
-    check data == await nrProto.receivedMessages.get()
+    let receivedMsg = await nrProto.receivedMessages.get().wait(2.seconds)
+    check data == receivedMsg.data
+
+    # assert anonymity of the sender
+    check:
+      receivedMsg.connPeerId != nodes[0].switch.peerInfo.peerId # not the sender
+      receivedMsg.connPeerId != destNode.peerInfo.peerId # not the destination itself
+      receivedMsg.connPeerId in nodes.mapIt(it.switch.peerInfo.peerId)
```
gmelodie, line 79: nit
```suggestion
    # assert anonymity of the sender
    let sender = nodes[0].switch.peerInfo.peerId
    let destination = destNode.peerInfo.peerId
    check:
      receivedMsg.connPeerId != sender
      receivedMsg.connPeerId != destination
      receivedMsg.connPeerId in nodes.mapIt(it.switch.peerInfo.peerId)
```

tests/libp2p/mix/test_conn.nim
vladopajic: aha i get it now. makes sense to have this check, but please include the same message from ^ comment.

> Assert that MixProtocol doesn't send any information back in case of a failure.

tests/libp2p/mix/test_conn.nim
```diff
@@ -58,18 +60,54 @@ suite "Mix Protocol Component":
...
+      exitNodes.mgetOrPut(receivedMsg.connPeerId, 0).inc()
+
+    # With 20 messages and 9 eligible nodes,
+    # random selection must produce at least 3 distinct exit nodes.
+    # Sender must never be exit and aestination must never be exit.
+    # No single node should monopolize the exit role.
+    check:
+      exitNodes.len >= 3
+      nodes[0].switch.peerInfo.peerId notin exitNodes
+      destNode.peerInfo.peerId notin exitNodes
```
gmelodie:
```suggestion
    # No single node should monopolize the exit role.
    let source = nodes[0].switch.peerInfo.peerId
    check:
      exitNodes.len >= 3
      source notin exitNodes
      destNode.peerInfo.peerId notin exitNodes
```

tests/libp2p/mix/test_mix_message.nim
rlve: sorry, misread your question

Assertion improved in https://github.com/vacp2p/nim-libp2p/pull/2107/changes/eb95b46fbb2df15601b955a8cfe4f8aa60192159

tests/libp2p/mix/test_conn.nim
```diff
@@ -371,18 +402,92 @@ suite "Mix Protocol Component":
...
+      await conn.writeLp(@[1.byte, 2, 3])
+
+  asyncTest "toConnection rejects expectReply without destReadBehavior":
+    let nodes = await setupMixNodes(10) # no destReadBehavior registered
+    startAndDeferStop(nodes)
+
+    let destNode = createSwitch()
+    await destNode.start()
+    defer:
+      await destNode.stop()
```
gmelodie, line 428: Here, you're not using `setupDestNode` because you want `destReadBehavior` to be missing? Might be worth it to add a comment.

tests/libp2p/mix/test_conn.nim
```diff
@@ -371,18 +402,92 @@ suite "Mix Protocol Component":
...
+      await stopDestNode(destNode)
+
+    let conn = nodes[0].toConnection(destNode.toMixDestination(), nrProto.codec).expect(
+        "could not build connection"
+      )
+    defer:
+      await conn.close()
+
+    expect LPStreamError:
+      await conn.write(newSeq[byte](DataSize + 1))
```
gmelodie: nit
```suggestion
      await conn.write(newSeq[byte](MaxDataSize + 1))
```

tests/libp2p/mix/test_mix_message.nim
```diff
@@ -35,3 +35,41 @@ suite "mix_message_tests":
...
       emptyMessage == string.fromBytes(dMixMsg.message)
       codec == dMixMsg.codec
+
+  test "getMaxMessageSizeForCodec returns correct size":
+    let codec = "/test/1.0.0"
+
+    let size0 = getMaxMessageSizeForCodec(codec, 0)
+    check:
+      size0.isOk
+      size0.get() > 0
```
gmelodie, line 46:
```suggestion
    check size0.get() > 0
```

gmelodie: Similarly for other occurrences

tests/libp2p/mix/utils.nim
rlve: These helpers are always using first address from the peer info, which works in our tests, but could be quite misleading for the consumers, as the address might not be dialable.

tests/libp2p/mix/test_conn.nim
rlve: It's a const coming from the implementation. Not going to rename.

gmelodie: Turns out it's not `max`, as message should be exactly that size. Please include a check for `DataSize - 1` as well

rlve: improved in https://github.com/vacp2p/nim-libp2p/pull/2107/changes/8c6e111044bdc2446e88ef3cafdaed1f4eeda13c

richard-ramos (approved):

PR #2109, opened 2026-02-12
refactor(conn): wait for peer-ready before handling streams

richard-ramos:
Fixes behaviour observed in nimbus in which the stream handler is executed before the connect even, thus the handshake fails

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,90 @@
+## Summary
+<!--
+Provide a clear and concise description of the purpose of this PR.
+
+Explain:
+- What this PR does
+- Why it is needed
+- The problem it solves or the improvement it introduces
+-->
+
+
+## Affected Areas
+<!--
+Indicate which parts of the codebase are impacted.
+Check all that apply and briefly describe the scope of the changes.
+-->
+
+- [ ] Gossipsub  
+  <!-- e.g. scoring changes, message validation, peer handling -->
+
+- [ ] Transports  
+  <!-- e.g. TCP, QUIC, WebSockets, Noise, etc. -->
+
+- [ ] Peer Management / Discovery
+
+- [ ] Protocol Logic
+
+- [ ] Build / Tooling
+
+- [ ] Other  
+  <!-- Describe below -->
+
+
+## Compatibility & Downstream Validation
+<!--
+For PRs affecting behavior on existing features, provide evidence that
+dependent projects build and function correctly with these changes.
+-->
+
+Reference PRs / branches / commits demonstrating successful integration:
+
+- **Nimbus:**  
+  <!-- Link PR or branch -->
+
+- **Waku:**  
+  <!-- Link PR or branch -->
+
+- **Codex:**  
+  <!-- Link PR or branch -->
+
+
+## Impact on Library Users
+<!--
+Describe how this affects downstream users of nim-libp2p.
+
+Examples:
+- API changes
+- Behavior changes
+- Performance implications
+- Migration requirements
+- No impact (internal refactor)
+-->
+
+
+## Risk Assessment
+<!--
+Identify potential risks introduced by this change.
+
+Consider:
+- Backward compatibility
+- Network behavior changes
+- Performance regressions
+- Security implications
+-->
+
+
+## References
+<!--
+Link to related:
+- Issues
+- Specs
+- Design discussions
+- Prior PRs
+-->
+
+
+## Additional Notes
+<!--
+Optional: any extra context reviewers should be aware of.
+-->
--- a/README.md
+++ b/README.md
@@ -62,8 +62,8 @@ Connect with other contributors in our [community channel](https://discord.com/c
 nim-libp2p is used by:
 
 - [Nimbus](https://github.com/status-im/nimbus-eth2), an Ethereum client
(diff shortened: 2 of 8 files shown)


libp2p/upgrademngrs/muxedupgrade.nim
```diff
@@ -86,12 +88,19 @@ proc new*(
...
 ): T =
-  let upgrader = T(muxers: muxers, secureManagers: @secureManagers, ms: ms)
+  let upgrader =
+    T(muxers: muxers, secureManagers: @secureManagers, ms: ms, connManager: connManager)
 
   upgrader.streamHandler = proc(conn: Connection) {.async: (raises: []).} =
     trace "Starting stream handler", conn
     try:
+      if not isNil(upgrader.connManager):
+        let ready = await upgrader.connManager.waitForPeerReady(conn.peerId)
```
vladopajic:
```
proc waitForPeerReady*( .... ): Future[bool] {.async: (raises: [CancelledError]).} =
```

```
upgrader.streamHandler = proc(conn: Connection) {.async: (raises: []).} =
```

raises mismatch here

libp2p/connmanager.nim
```diff
@@ -337,10 +391,16 @@ proc storeMuxer*(
...
 
-  await c.triggerConnEvent(
+  let connectedEvent = c.triggerConnEvent(
     peerId, ConnEvent(kind: ConnEventKind.Connected, incoming: dir == Direction.In)
   )
 
+  # this notifies that peer is ready once the Connected events have been started
+  # but before waiting for them to be completed, avoiding deadlocks where a 
+  # connected handler would need inbound streams to progress.
+  c.peerIsReady(peerId)
```
vladopajic: peerIsReady -> notifyPeerReady

IsReady  indicates that it returns `bool` if it is ready or not

libp2p/connmanager.nim
```diff
@@ -108,6 +111,56 @@ proc new*(
...
+  c.readyEvents.withValue(peerId, readyEvent):
+    if not readyEvent[].finished:
+      return readyEvent[]
+
+  let readyEvent = Future[void].Raising([CancelledError]).init()
+  c.readyEvents[peerId] = readyEvent
+  readyEvent
+
+proc peerIsReady(c: ConnManager, peerId: PeerId) =
+  c.readyPeers.incl(peerId)
```
vladopajic, line 126: `readyPeers` will keep growing. assume all peers wil get ready, then this set just keeps growing even after peer disconnects, right?

libp2p/connmanager.nim
```diff
@@ -108,6 +111,56 @@ proc new*(
...
+  ## Wait until `storeMuxer` has emitted the `Connected` conn event for `peerId`.
+  ## Existing ready peers bypass waiting.
+  if c.closed:
+    return false
+
+  if peerId in c.readyPeers:
+    return true
+
+  let readyEvent = c.getReadyEvent(peerId)
+  if c.closed:
```
vladopajic: if c.closed: already checked above

libp2p/upgrademngrs/muxedupgrade.nim
vladopajic: @arnetheduck why does this compile?

arnetheduck: not sure why it shouldn't? there's an `except CancelledError` that prevents said exception from propagating...

vladopajic: ah, right! i totally missed that

libp2p/transports/quictransport.nim
```diff
@@ -242,9 +244,20 @@ proc defaultCertGenerator(
 ): CertificateX509 {.gcsafe, raises: [TLSCertificateError].} =
   return generateX509(kp, encodingFormat = EncodingFormat.PEM)
 
-proc new*(_: type QuicTransport, u: Upgrade, privateKey: PrivateKey): QuicTransport =
+proc toOptConnManager(connManager: ConnManager): Opt[ConnManager] {.inline.} =
+  if isNil(connManager):
+    Opt.none(ConnManager)
+  else:
+    Opt.some(connManager)
```
gmelodie: I think this could be moved to `utility.nim` (maybe even to `nim-results`:
```nim
proc toOpt*[T](x: T): Opt[T] =
  if x.isNil: Opt.none(T) else: Opt.some(x)
```

libp2p/upgrademngrs/muxedupgrade.nim
```diff
@@ -86,12 +88,19 @@ proc new*(
     muxers: seq[MuxerProvider],
     secureManagers: openArray[Secure] = [],
     ms: MultistreamSelect,
+    connManager: ConnManager = nil,
```
gmelodie: Should this take an `Opt` instead? It feels confusing to mix up `nil` and `opt`

libp2p/connmanager.nim
richard-ramos: oops!

libp2p/connmanager.nim
richard-ramos: After the peer fully disconnects, the peer should be removed from readyPeers in `muxCleanup` when calling `peerIsNotReady`. 

But, reviewing this proc again, i see the name is not really good. I'll rename it to `clearPeerReadyState`

PR #2111, opened 2026-02-16
feat(kad):  DHT Metrics

SionoiS:
Hello!

I added metrics for the base Kad DHT impl.

Did I miss information that you think would be useful?

### DIFF
--- a/libp2p/protocols/rendezvous/rendezvous.nim
+++ b/libp2p/protocols/rendezvous/rendezvous.nim
@@ -244,17 +244,17 @@ proc discover*[E](
     return
   var nextOffset = cookie.offset
   let n = Moment.now()
-  var s = collect(newSeq()):
-    for index in namespaces:
-      var reg = rdv.registered[index]
-      if limit == 0:
-        break
-      if reg.expiration < n or index.uint64 < cookie.offset:
-        continue
-      limit.dec()
-      nextOffset = index.uint64 + 1
-      reg.data.ttl = Opt.some((reg.expiration - Moment.now()).seconds.uint64)
-      reg.data
+  var s: seq[Register] = @[]
+  for index in namespaces:
+    var reg = rdv.registered[index]
+    if limit == 0:
+      break
+    if reg.expiration < n or index.uint64 < cookie.offset:
+      continue
+    limit.dec()
+    nextOffset = index.uint64 + 1
+    reg.data.ttl = Opt.some((reg.expiration - Moment.now()).seconds.uint64)
+    s.add(reg.data)
   rdv.rng.shuffle(s)
   await conn.sendDiscoverResponse(s, Cookie(offset: nextOffset, ns: d.ns))
 


libp2p/protocols/kademlia/ping.nim
```diff
@@ -16,17 +16,29 @@ proc ping*(
...
+  kad_message_bytes_sent.inc(encoded.buffer.len.int64, labelValues = ["ping"])
+  let startTime = Moment.now()
+  await conn.writeLp(encoded.buffer)
 
-  let reply = Message.decode(await conn.readLp(MaxMsgSize)).tryGet()
+  let replyBuf = await conn.readLp(MaxMsgSize)
+  kad_message_bytes_received.inc(replyBuf.len.int64, labelValues = ["ping"])
+  kad_message_duration_ms.observe(
+    (Moment.now() - startTime).toFloatMs(), labelValues = ["ping"]
+  )
```
richard-ramos: For metrics that measure time, i think we can do something like this (iiuc the code in nim-metrics)
```nim
var replyBuf: seq[byte]
kad_message_duration_ms.time(["ping"]):
  await conn.writeLp(encoded.buffer)
  replyBuf = await conn.readLp(MaxMsgSize)
kad_message_bytes_received.inc(replyBuf.len.int64, labelValues = ["ping"])
```

libp2p/protocols/kademlia/kademlia_metrics.nim
```diff
@@ -0,0 +1,119 @@
...
+    "responses with closer peers", ["type"]
+
+  # Routing table metrics
+  declarePublicGauge kad_routing_table_peers, "total peers in routing table"
+  declarePublicGauge kad_routing_table_buckets, "number of buckets"
+  declarePublicGauge kad_routing_table_bucket_size, "peers per bucket", ["bucket"]
+  declarePublicCounter kad_routing_table_insertions, "peer insertions"
+  declarePublicCounter kad_routing_table_replacements, "peer replacements"
+
+  proc toFloatMs*(d: Duration): float64 =
```
richard-ramos: To avoid duplicating this proc, extract it from the if/else blocks

libp2p/protocols/kademlia/kademlia_metrics.nim
```diff
@@ -0,0 +1,119 @@
...
+    d.nanoseconds.float64 / 1_000_000.0
+
+else:
+  # No-op stubs when metrics disabled
+  import chronos
+
+  type
+    kad_messages_sent* = object
```
vladopajic: these should be `declareNoopCounter` and `declareNoopGauge`. 

```
declareNoopCounter kad_messages_received 
declareNoopGauge kad_message_bytes_sent
....
```

or 
```
var 
 kad_messages_received = noopCounter()
 kad_message_bytes_sent = noopGauge()
```

then code below would not need to define all those templates for each and every counter.  because 
`noopCounter()` and `noopGauge()` will return new instance of some types say `NoopMetrics` that has all those methods, so only this typed defines it.

type `NoopMetrics` can have all methods of gauge and counter, and then all metrics variables could be initialized to this value.

then there is futher improvment at that would be, to create new `declarePublicOrNoopCounter`  and `declarePublicOrNoopGauge`, templates that would based on some flag, `libp2p_kad_metrics`, create ether noop metrics or real metrics. then we dont have to repeat all code all over again

SionoiS: I've been thinking, do we even need to put metrics behind a flag? WDYT?

edit: there's already a flag to disable metric... I missed this.

vladopajic: even better :)

PR #2076, opened 2026-02-03
test(extensions): adding tests for partial message

vladopajic:
pr only adds tests for partial message extension. test for integration with gossipsub in later pr.

PR #2115, opened 2026-02-17
test(interop): add partial message nim interop test

vladopajic:
as part of https://github.com/vacp2p/nim-libp2p/issues/2098, this pr:
- adds `partialMessageInteropTest` proc that implements interop behavior of "nim peer"
- adds tests for `partialMessageInteropTest` logic 

future pr should add implementation of "other peer" using different implementation and integration with ci

### DIFF
--- a/tests/interop/partial_message.nim
+++ b/tests/interop/partial_message.nim
@@ -21,7 +21,7 @@ proc partialMessageInteropTest*(
     ourAddr: string,
     otherAddr: string,
     otherPeerId: PeerId,
-    timeout: Duration = 2.minutes,
+    timeout: Duration = 1.minutes,
 ): Future[bool] {.async.} =
   # this proc implements interop test for testing gossipsub partial message extension.
   # there are two entities:


tests/interop/partial_message.nim
```diff
@@ -0,0 +1,107 @@
...
+import chronos, stew/byteutils, tables, chronicles, sequtils
+import ../../libp2p/[builders, peerid, wire]
+import ../../libp2p/protocols/pubsub/[gossipsub, gossipsub/extensions, rpc/message]
+import ../tools/crypto
+import ../libp2p/pubsub/extensions/my_partial_message
+
+const partialTopic* = "logos-partial"
+
+proc makePartialMessage*(): MyPartialMessage =
+  MyPartialMessage(
```
vladopajic, line 15: please suggest naming for `MyPartialMessage`.
https://github.com/vacp2p/nim-libp2p/pull/2076#discussion_r2774391142

`TestPartialMessage`, `ExamplePartialMessage` have veto :)

tests/interop/partial_message.nim
```diff
@@ -0,0 +1,111 @@
...
+  ) {.gcsafe, raises: [].} =
+    # other peer must publish only one partial message.
+    # in this callback code asserts that other peer
+    # has published exactly that message and that nim peer has received.
+
+    if rpc.topicID != partialTopic:
+      error "partial message topic did not match",
+        got = $rpc.topicID, expected = $partialTopic
+      requestFulfilled.complete(false)
+      return
```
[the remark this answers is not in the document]
vladopajic: thats okay, it should riase

tests/interop/partial_message.nim
```diff
@@ -0,0 +1,111 @@
...
+  MyPartialMessage(
+    groupId: "interop-group".toBytes,
+    data: {1: "one".toBytes, 2: "two".toBytes, 3: "three".toBytes}.toTable,
+  )
+
+proc partialMessageInteropTest*(
+    ourAddr: string,
+    otherAddr: string,
+    otherPeerId: PeerId,
+    timeout: Duration = 5.minutes,
```
richard-ramos: why such a long timeout for tests?

richard-ramos (approved):
Looks good, but seems to me that this is not really an interop test because it is not testing with different implementations to see if anything breaks, tho. 
In theory nim to nim errors should have been captured with other tests

tests/interop/partial_message.nim
richard-ramos: ExpectedPartialMessage?   :stuck_out_tongue:

vladopajic:
> Looks good, but seems to me that this is not really an interop test because it is not testing with different implementations to see if anything breaks, tho. 
> In theory nim to nim errors should have been captured with other tests

this is part 1 of entire effort.
part 2, where "other peer" is implemented with go-libp2p is not implemented in this pr. and it will be in next pr.

tests/interop/partial_message.nim
vladopajic: copy paste from other iterop test logic

tests/interop/partial_message.nim
vladopajic: here i am looking for better type name for `MyPartialMessage` type that implements `PartialMessage`, and since we use it in many cases maybe we can have better name? 
why `ExpectedPartialMessage` ? were `Expected` comes from?
is `LibPartialMessage` or `B2PartialMessage` better than `MyPartialMessage`? (from comment)

tests/interop/test_partial_message.nim
```diff
@@ -0,0 +1,126 @@
...
+
+  (switch, gossipsub)
+
+suite "Gossipsub Partial Message Interop Tests with Nim nodes":
+  const ourAddress = "/ip4/127.0.0.1/tcp/0"
+
+  teardown:
+    checkTrackers()
+
+  asyncTest "Happy path":
```
gmelodie, line 59: Nit: don't like the name (maybe `partial message success` or sth)

PR #2116, opened 2026-02-17
fix: memory leak and stream removal

richard-ramos:
- Removes streams once connections are closed.
- Disables redundant connection timeout verification (since lsquic does it automatically)
- List of connections and streams changed from sequence to sets to not have to iterate each time we want to remove these from the list

### DIFF
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -6,11 +6,13 @@ import chronos, chronicles, metrics, results
 import lsquic
 import
   ../wire,
+  ../connmanager,
   ../multiaddress,
   ../multicodec,
   ../muxers/muxer,
   ../stream/connection,
-  ../upgrademngrs/upgrade
+  ../upgrademngrs/upgrade,
+  ../utility
 import ./transport
 import tls/certificate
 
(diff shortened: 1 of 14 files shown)


libp2p/transports/quictransport.nim
```diff
@@ -106,14 +110,20 @@ method closeImpl*(stream: QuicStream) {.async: (raises: []).} =
...
   procCall P2PConnection(session).isClosed or session.connection.isClosed
 
 method close*(session: QuicSession) {.async: (raises: []).} =
-  await noCancel allFutures(session.streams.mapIt(it.close()))
+  let streams = session.streams
+  session.streams.clear()
+  var futs = newSeqOfCap[Future[void]](streams.len)
+  for s in streams:
+    futs.add(s.close())
+  await noCancel allFutures(futs)
```
vladopajic:
```suggestion
  await noCancel allFutures(streams.mapIt(it.close()))
```

why not?

vladopajic (approved):
lgtm

what was memory leak ?

libp2p/transports/quictransport.nim
richard-ramos: I agree. Much more readable with mapIt

richard-ramos:
The leak happened here:
image

The stream was closed, yet a reference to the stream was kept until the connection was closed

PR #2119, opened 2026-02-18
test(interop): add partial message interop with go peer

vladopajic:
second part of https://github.com/vacp2p/nim-libp2p/issues/2098:

### DIFF
--- a/.pinned
+++ b/.pinned
@@ -8,7 +8,7 @@ json_serialization;https://github.com/status-im/nim-json-serialization@#2b1c5eb1
 metrics;https://github.com/status-im/nim-metrics@#6142e433fc8ea9b73379770a788017ac528d46ff
 ngtcp2;https://github.com/status-im/nim-ngtcp2@#791eb859145f9f268eb23eb9cbe777bdd7699c4d
 nimcrypto;https://github.com/cheatfate/nimcrypto@#19c41d6be4c00b4a2c8000583bd30cf8ceb5f4b1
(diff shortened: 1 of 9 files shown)


.github/workflows/interop.yml
```diff
@@ -120,3 +120,29 @@ jobs:
...
+
+      # build go and nim clients first to prevent one from waiting on the other to finish building.
+      # the go peer runs in the background because the interop test logic is implemented in the nim peer.
+      - name: Run Interop
+        run: |
+          nimble setup
+          cd interop/partial-message
+          nim c ./nim-peer/src/peer.nim
+          go build -C ./go-peer -o peer ./peer.go
+          ./go-peer/peer & ./nim-peer/src/peer
```
vladopajic, line 148: ideally this could be bash script in `interop/partial-message/run.sh`
this way running interop locally would be easy.
this should probably then be a pattern for other interops. 
maybe improvement for later? :)

tests/interop/partial_message.nim
```diff
@@ -32,12 +32,21 @@ proc partialMessageInteropTest*(
...
+    key = PrivateKey.init(keyBytes).expect("should have a private key")
+
   var switch = SwitchBuilder
     .new()
     .withRng(rng())
     .withAddresses(@[MultiAddress.init(ourAddr).tryGet()])
+    .withPrivateKey(key)
     .withTcpTransport()
-    .withMplex()
+    .withYamux()
```
gmelodie: Why yamux?

vladopajic: fixed

PR #2122, opened 2026-02-20
chore: resolve exception in logos -delivery

darshankabariya:
It’s a very small change to fix an exception in logos-delivery. 

https://github.com/logos-messaging/logos-delivery/issues/3731


Ivansete-status:
@darshankabariya 
This PR's title is a little bit misleading.
What it actually does is to hide an exception, instead of fixing it.
I'm not super happy about that one and I think we should understand why the exception happens so frequently instead of just ignoring it, as this PR suggests.
But maybe I'm missing something.
wdyt @richard-ramos

PR #2130, opened 2026-02-25
chore(cbind): remove duplicate flags

gmelodie:
## Summary
- Remove the `flags` field from cbinding configs in favor of explicit default configs
- Duplicated logic (`flags` + individual boolean options)

## Affected Areas
- [ x ] Cbindings  

## Impact on Library Users
- logos-libp2p-module (adjust to no longer use flags)

### DIFF
--- a/cbind/libp2p.nim
+++ b/cbind/libp2p.nim
@@ -262,6 +262,7 @@ proc libp2p_new_private_key(
   return RET_OK.cint
 
 proc libp2p_new_default_config(): Libp2pConfig {.dynlib, exportc, cdecl.} =
+  initializeLibrary()
   return Libp2pConfig.init()
 
 proc libp2p_new(


cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
```diff
@@ -193,6 +152,57 @@ proc createLibp2p(appCallbacks: AppCallbacks, config: Libp2pConfig): LibP2P =
 
   return ret
 
+proc copyConfig(config: ptr Libp2pConfig): Libp2pConfig =
+  var resolved = Libp2pConfig.default()
```
richard-ramos, line 170: When returning a default, the default dnsResolver is set to DefaultDnsResolver.cstring, which is not allocated via allocShared... 

destroyShared in line 225 will call deallocShared on self[].config.dnsResolver which is pointing to this cstring, and you will end up with undefined behavior which might cause crahses or corruption latter...

To fix this, you should probably create an allocShared copy of the default string inside the `default` proc.

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
```diff
@@ -193,6 +152,57 @@ proc createLibp2p(appCallbacks: AppCallbacks, config: Libp2pConfig): LibP2P =
...
+  if config[].manualPrivKey != 0 and not config[].privKey.data.isNil() and
+      config[].privKey.dataLen > 0:
+    let srcKey = config[].privKey
+    resolved.privKey.dataLen = srcKey.dataLen
+    resolved.privKey.data = allocShared(srcKey.dataLen.int)
+    copyMem(resolved.privKey.data, srcKey.data, srcKey.dataLen.int)
+
+  resolved.kadBootstrapNodesLen = config[].kadBootstrapNodesLen
+  if config[].kadBootstrapNodesLen > 0:
+    if not config[].kadBootstrapNodes.isNil() and config[].kadBootstrapNodesLen > 0:
```
richard-ramos: `config[].kadBootstrapNodesLen > 0` is redundant as you already check this in line 186

cbind/libp2p.nim
```diff
@@ -261,6 +263,22 @@ proc libp2p_new_private_key(
     )
   return RET_OK.cint
 
+proc default(T: typedesc[Libp2pConfig]): T =
```
richard-ramos: This should be exported, right? otherwise `copyConfig` will not use this but use the default zero inited struct when it calls Libp2pConfig.default()

cbind/ffi_types.nim
```diff
@@ -123,6 +112,7 @@ type Libp2pConfig* {.bycopy.} = object
   dnsResolver*: cstring
   kadBootstrapNodes*: ptr Libp2pBootstrapNode
   kadBootstrapNodesLen*: csize_t
+  manualPrivKey*: cint
```
richard-ramos: does this mean that any new field that can be non empty will need a corresponding `manual____` attribute?

richard-ramos (approved):

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
```diff
@@ -193,6 +154,64 @@ proc createLibp2p(appCallbacks: AppCallbacks, config: Libp2pConfig): LibP2P =
...
+  resolved.mountKad = config[].mountKad
+  resolved.mountMix = config[].mountMix
+  resolved.mountKadDiscovery = config[].mountKadDiscovery
+
+  if not config[].dnsResolver.isNil() and config[].dnsResolver[0] != '\0':
+    let src = config[].dnsResolver
+    var len = 0
+    while src[len] != '\0':
+      inc len
+    inc len # include null terminator
```
vladopajic, line 186: `let len = cstringLen(src)`

PR #2139, opened 2026-02-27
fix(kad): add configurable option to purge stale entries on bucketrefreshInterval

chaitanyaprem:
## Summary

Nodes such as logos-delivery/waku that don't have persistent node-keys would cause DHT bloat due to restarts. This happens especially apps that are running kademlia as each restart creates a new peerID. 

Added config flag for users to define this behaviour of purging stale entries on bucketRefreshInterval with default behaviour being not do do anything.

@SionoiS , @tinniaru3005  @jm-clius  wdyt? Does this solve the issue observed while testing with fleet or we need to do something else?

## Affected Areas

- **Waku:**  https://github.com/logos-messaging/logos-delivery/pull/3739

### DIFF
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -21,6 +21,8 @@ proc bootstrap*(
   ## Sends a findNode to find itself to keep nearby peers up to date
   ## Also sends a findNode to find a random key for each non-empty k-bucket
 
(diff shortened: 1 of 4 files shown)


SionoiS (approved):
LGTM

We still don't know what the problem is exactly but this can't hurt.

PR #2145, opened 2026-03-05
chore: harmonized proc name for no-op completed future

### LINKED ISSUE #1818: proc `completedFuture` should be `newFutureCompleted[T]`
https://github.com/vacp2p/nim-libp2p/blob/test-workaround/tests/utils/futures.nim#L63

- name should be changed from `completedFuture` to `newFutureCompleted` - to share same prefix as `newFuture`
- should have type parameter

rikettsie:
The proc returning an "already-completed" Future[void] with no pending work should be renamed to match the same prefix as `newFuture` and should accept a generic type parameter.

Closes #1818

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -112,6 +112,7 @@ proc init*(
     sendIDontWantOnPublish = false,
     testExtensionConfig = none(TestExtensionConfig),
     partialMessageExtensionConfig = none(PartialMessageExtensionConfig),
+    pingpongExtensionConfig = none(PingPongExtensionConfig),
 ): GossipSubParams =
   GossipSubParams(
     explicit: true,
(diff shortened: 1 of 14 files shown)


rikettsie:
About the failure in the transport interoperability tests, it seems that the `interop.yml` points to a broken commit of libp2p/tests-plans which were red in their [last days' commits](https://github.com/libp2p/test-plans/commits/master/) and they were suspended in the last [chore commit](https://github.com/libp2p/test-plans/commit/1e02a8765b2f9d58d6b7ebefb2f5d7242a64c753).

Do you think we should update our pin too? If so, I can add the following change at the 32 of the `interop.yml` :
```
uses: libp2p/test-plans/.github/actions/run-transport-interop-test@1e02a8765b2f9d58d6b7ebefb2f5d7242a64c753
```

tests/tools/futures.nim
```diff
@@ -2,11 +2,9 @@
 # Copyright (c) Status Research & Development GmbH 
 
 import chronos/futures, chronos, sequtils
+import ../../libp2p/utils/future
 
-proc completedFuture*(): Future[void] =
-  let f = newFuture[void]()
-  f.complete()
-  f
+export future.newFutureCompleted
```
vladopajic: instead of exporting, `../../libp2p/utils/future` should be used directly in those tests that use this utility.
importing then exporting should be avoided.

PR #2147, opened 2026-03-06
fix(kad): handle dial failed error early

gmelodie:
## Summary
Kademlia dispatch functions popped `DialFailedErrors` which could make operations to be cancelled when failing to dial one peer, log the error earlyand keep going.

## Affected Areas
- [ x ] Kademlia

/

## Impact on Library Users
Bug fix, no API or behavior changes

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,482 @@
+# nim-libp2p Coding Agent Instructions
+
+## Project Overview
+
+**nim-libp2p** is a native Nim implementation of the [libp2p](https://libp2p.io) peer-to-peer networking stack. It is used in production by [Nimbus (Ethereum client)](https://github.com/status-im/nimbus-eth2) and other projects.
+
+- **Language**: Nim ( see `libp2p.nimble` and `.github/workflows/ci.yml` for currently supported versions)
+- **License**: Apache 2.0 OR MIT
+- **Version**: 1.15.3
+- **Author**: Status Research & Development GmbH
+
+---
+
+## Repository Layout
+
+```
+nim-libp2p/
+├── libp2p/           # Main source code
+│   ├── switch.nim              # Core: manages connections, protocols, transports
+│   ├── standard_switch.nim     # Default switch implementation
+│   ├── builders.nim            # Builder pattern for switches
+│   ├── peerid.nim / peerinfo.nim / peerstore.nim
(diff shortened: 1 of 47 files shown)


libp2p/protocols/kademlia/find.nim
```diff
@@ -120,11 +120,14 @@ proc dispatchFindNode*(
...
-.} =
+): Future[Opt[Message]] {.async: (raises: [CancelledError, LPStreamError]), gcsafe.} =
   let addrs = addrs.valueOr(kad.switch.peerStore[AddressBook][peer])
-  let conn = await kad.switch.dial(peer, addrs, kad.codec)
+  let conn =
+    try:
+      await kad.switch.dial(peer, addrs, kad.codec)
+    except DialFailedError as e:
+      error "FindNode could not dial peer", description = e.msg
+      return Opt.none(Message)
```
richard-ramos, line 130: Returning an `Opt.none` here means that in iterativeLookup the `state.responded[peerId]` will be `RespondedStatus.Success`. Couldn't this cause the iterativeLookup to finish earlier than expected? 

Afaik the code there should handle cases when an exception is raised by checking if the future is failed:
```nim
      if not fut.finished():
        continue
      state.responded[peerId] =
        if fut.failed(): RespondedStatus.Failed else: RespondedStatus.Success
```

libp2p/protocols/kademlia/get.nim
richard-ramos, line 90: Code now returns a none for failures, thus the `onReply` will be executed.
`received[peer]` will be set even if there was a failure. and since the stop condition uses `received.len >= quorum`, a failure will contribute to the quorum, meaning that we can say now that there are not enough valid records earlier due to stopping the iteration earlier due to failures even tho we might have more peers to query available.


It makes me wonder in general if the try/except that we added in this PR should be handled within the iterativeLookup, and not in the dispatchs

libp2p/protocols/kademlia/find.nim
gmelodie: This is also true for code that was previously here, no? i.e.

```nim
  let reply = Message.decode(replyBuf).valueOr:
    debug "FindNode reply decode fail", error = error, conn = conn
    return Opt.none(Message)
```

gmelodie: https://github.com/vacp2p/nim-libp2p/issues/2163

PR #2161, opened 2026-03-13
chore: fix depracated warning for `Future.cancel()`

vladopajic:


tests/libp2p/protocols/test_perf.nim
```diff
@@ -93,8 +93,7 @@ proc runTestWithException(server: Switch, client: Switch) {.async.} =
     stats.uploadBytes == bytesToUpload
     stats.downloadBytes > 0
 
-  perfFut.cancel() # cancelling future will raise exception in perfClient
-  await sleepAsync(10.milliseconds)
```
vladopajic, line 97: sleep was needed to "wait" for cancel to be finished, now it is much better solution

PR #2164, opened 2026-03-14
chore(kad): handle dial failed exception

### LINKED ISSUE #2163: KadDHT: iteractiveLookup should handle exceptions
From https://github.com/vacp2p/nim-libp2p/pull/2147#discussion_r2908207478, `iteractiveLookup` should handle `DialFailedError`s and decode errors (which should be exceptions), instead of having `dispatchXXX` functions return `Opt.none` on those cases.

gmelodie:
## Summary

Previously, `DialFailedError` was silently swallowed in all Kademlia dispatch procs (`dispatchFindNode`, `dispatchGetVal`, `dispatchPutVal`, `dispatchGetProviders`, `dispatchAddProvider`, `ping`), returning `Opt.none` or early-returning on dial failure. This hid errors from the caller and prevented proper error attribution in `iterativeLookup`.

This PR propagates `DialFailedError` (and `ValueError` for decode failures) up the call stack so that `iterativeLookup` can distinguish between unreachable peers (dial failed) and peers that responded with malformed data, and log them accordingly.

## Affected Areas

- [ ] Gossipsub  
- [ ] Transports  
- [x] Protocol Logic  
  Kademlia DHT dispatch procs in:
  - `find.nim`
  - `get.nim`
  - `put.nim`
  - `provider.nim`
  - `ping.nim`  
  Error handling refactored to propagate rather than swallow exceptions.  
- [ ] Build / Tooling  
- [ ] Other  

- **Nimbus:** N/A  
- **Waku:** N/A  
- **Codex:** N/A  

## Impact on Library Users

- `DispatchProc` type signature now includes `DialFailedError` and `ValueError` in its `raises` set  
- Custom `DispatchProc` implementations must update their `raises` annotations accordingly  
- Behavioral change:
  - Dial failures and decode errors are no longer silently ignored  
  - Errors propagate to `iterativeLookup` and are logged at `error` level  

## Risk Assessment

- **Backward compatibility:**  
  - Wider `raises` set requires annotation updates  
  - No runtime impact for implementations that don’t raise these errors  

- **Network behavior:**  
  - Unchanged  
  - Failed peers still marked as `RespondedStatus.Failed` and excluded  

- **Performance:**  
  - No impact  

- **Security:**  
  - No impact  

## References

Fix #2163   

## Additional Notes

`iterativeLookup` now logs:
- `debug "Kad lookup: dial failed"` for unreachable peers  
- `debug "Kad lookup: RPC error"` for malformed responses


libp2p/protocols/kademlia/provider.nim
```diff
@@ -186,13 +181,12 @@ method handleAddProvider*(
 
 proc dispatchGetProviders*(
     kad: KadDHT, peer: PeerId, key: Key
-): Future[Opt[Message]] {.async: (raises: [CancelledError, LPStreamError]), gcsafe.} =
+): Future[Opt[Message]] {.
```
vladopajic: why is `Opt` needed as response type?

libp2p/protocols/kademlia/get.nim
```diff
@@ -12,13 +12,12 @@ logScope:
 
 proc dispatchGetVal*(
     kad: KadDHT, peer: PeerId, key: Key
-): Future[Opt[Message]] {.async: (raises: [CancelledError, LPStreamError]), gcsafe.} =
+): Future[Opt[Message]] {.
```
vladopajic: Opt[Message] why not just Message

vladopajic (commented):
looks good, with comments:

- in general, returning `Opt.none(T)` doesn't make much sense if proc are raising errors. because in general, all successfully executed procs should result with some value. if proc is called, but it can't retrieve value (there was error, or simply values was not set - like getting value from map) it should raise error.

libp2p/protocols/kademlia/provider.nim
gmelodie: Outside of the scope of this PR

libp2p/protocols/kademlia/get.nim
gmelodie: Outside of the scope of this PR

libp2p/protocols/kademlia/provider.nim
vladopajic: it is not. "outside of the scope" is veil of ignorance for not doing things. 

when change is introduced, it can have second order consequence. these consequences always have other implications on the code and if one keeps hiding under same veil, we could never explore those consequences. 

as well as, leaving partially finished changes with unnecessary things in interface, requires other changes to build on top of unnecessary things compounding on complexity.

pr is already introducing change in interface. that interface always needs to be rational.

gmelodie: We have survived with this return in this proc before this PR. The PR says `handle dial failed exception`. This is outside of the scope.

richard-ramos: it’s reasonable to question changes even within a scoped PR, but we also shouldn’t let PRs grow with side work.

let’s do the following:

* if this change is required for the fix, add a short justification in the PR
* if not, open a follow-up issue so it’s tracked and doesn’t get lost

the observation about Opt[Key] is valid, but the change itself can be handled as a follow-up

gmelodie: It's not required

richard-ramos (approved):

vladopajic (approved):

Issue #841, opened 2023-01-12
nim-protobuf-serialization integration


Menduist:
- https://github.com/status-im/nim-protobuf-serialization/issues/37 (pruned down to what is currently supported)
- https://github.com/status-im/nim-protobuf-serialization/issues/39
- fuzzing
- https://github.com/status-im/nim-libp2p/pull/838
- Adding missing features (oneOf, comments, imports, etc)
- Redo the conformance testing with every needed feature

vladopajic:
closing. nim-protobuf-serialization is now fully utilized for all messages

PR #2166, opened 2026-03-16
chore: Use protobuf_serialization in RendezVous protocol

nitely:
Ref: #841 #838

Changes:

- Replace minprotobuf by protobuf_serialization in `protocols/rendezvous`
- ~~Follow the proto definition https://github.com/libp2p/specs/blob/master/rendezvous/README.md#protobuf~~


libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -472,16 +497,18 @@ proc unsubscribe*[E](
   if ns.len < MinimumNamespaceLen or ns.len > MaximumNamespaceLen:
     raise newException(RendezVousError, "Invalid namespace")
 
-  let msg = encode(
-    Message(msgType: MessageType.Unregister, unregister: Opt.some(Unregister(ns: ns)))
+  let msg = Protobuf.encode(
+    Message(
+      msgType: pbSome(MsgTypeUnregister), unregister: pbSome(Unregister(ns: pbSome(ns)))
+    )
   )
```
vladopajic, line 502: i don't think this is the way to go.

ideally we should keep the good old `Message` type. since it is implemented using serialization agnostic types. new `Message` has dependency to `protobuf_serialization` and has to be included everywhere even tho we don't really need serialization there. just because fields of `Message` are depended on those types. 

this change should not change business logic and type like it currently does, and should only change underlay logic in `ecode` / `decode` for these types. or, look at it this way:
if we want to serialized `Message` with json. there should be just proc on `Message` type `ecnodeJson` or if we need to encode it to protobuf there should be just changes in proc `encodeProtobug` we should not change type nor businesses logic every time serialization is changed.

vladopajic (commented):
- ideally, change in serialization method should have no impact on business logic. other then 
`let bytes = msg.encodeMiniProtobuf()` - > 
`let bytes = msg.encodeProtobuf()` -> 
`let bytes = msg.encodeJson()`
... or not even this if just underlay logic of `encode` is changed.
- just because plain types now how `PBOption` as filed it will require unnecessary import just because of it
-  `PBOption` goes against our rule to use only `results` library, for optional values. this will indefinitely produce clash between these two

it would be better to enclose on protobuf changes to a single file, keep same old plain types, even at at cost of manual conversion from one type to another and having 1:1 mapping of plain old types <-> pb specific types.

libp2p/protocols/rendezvous/protobuf.nim
```diff
@@ -1,268 +1,63 @@
...
+  MsgTypeDiscover* = 3'u32
+  MsgTypeDiscoverResponse* = 4'u32
+  ResponseOk* = 0'u32
+  ResponseInvalidNamespace* = 100'u32
+  ResponseInvalidSignedPeerRecord* = 101'u32
+  ResponseInvalidTTL* = 102'u32
+  ResponseInvalidCookie* = 103'u32
+  ResponseNotAuthorized* = 200'u32
+  ResponseInternalError* = 300'u32
+  ResponseUnavailable* = 400'u32
```
gmelodie: Why not enums?

libp2p.nimble
```diff
@@ -13,7 +13,8 @@ requires "nim >= 2.0.0",
   "chronicles >= 0.11.0", "chronos >= 4.2.2", "metrics", "secp256k1", "stew >= 0.4.2",
   "unittest2", "results", "serialization", "lsquic >= 0.2.0",
   "https://github.com/status-im/nim-websock#42c37b4172519566db016810eccfce8a02cc1cdf",
-  "https://github.com/vacp2p/nim-jwt.git#057ec95eb5af0eea9c49bfe9025b3312c95dc5f2"
+  "https://github.com/vacp2p/nim-jwt.git#057ec95eb5af0eea9c49bfe9025b3312c95dc5f2",
+  "https://github.com/status-im/nim-protobuf-serialization#4619854df3c442694ee150d130c1a4ae473d941d"
```
gmelodie: It'd be good to have a release instead

gmelodie (changes requested):
I think this is very cool, but I agree with @vladopajic's main comment about `PBOption` because it has two implications (just paraphrasing what he said):
1. you need to import it everywhere
2. we'll constantly need to convert between `result` opts and `PBOptions`

Besides that, I think this is good to go

nitely:
protobuf2 optionals and results.Opt have different semantics. In proto2 [you can set a default value](https://protobuf.dev/programming-guides/proto2/#explicit-default) in the proto definition (ie: this is why `PBOption.get()` always returns a value even if `isSome` is false). It should also be rare to check `isSome`, as the absence of an optional field value usually implies the default value should be used; If the field value must be set, then the field is not really optional after all (and then it's hard to change the proto definition in a backward compatible way). In fact proto3 removed this distinction for a while by making every field optional and not ever writing default values in the wire (ie: the behavior you get with `PBOption.get()` if you never check isSome); they later added support for `optional` fields back in proto3 for the fringe cases.

That said, from what I can see there is not a single libp2p protocol that defines a default in the proto file, so in theory we could use Opt (and `Opt.get(default(T))` to mimic PBOption behavior) if I add support for custom types in protobuf-ser.

> you need to import it everywhere

@vladopajic @gmelodie eh, there is a `import protobuf_serialization; export protobuf_serialization` so not really. If you need to work with a type that uses `PBOption`, that module would export protobuf so I don't think you will need to import the proto library everywhere.

> we'll constantly need to convert between result opts and PBOptions

There is one single conversion in this PR, but ye, it's likely that conversion will be needed here and there for other protocols. I don't think there would be many but it's a good point.

> ideally, change in serialization method should have no impact on business logic.

~~Ideally yes, but this protocol (I've not checked the rest) is wrongly defined, in that it makes some fields "require". There is not a single required field in the spec proto file. By chance, I don't think this caused many business logic changes except for the cookie one in the discover API (which IMO is wrong to start with, by the notes in my top comment). So, you are saying you don't want to fix this, and want to keep the wrongly define required types? (meaning: `ResponseStatus`, `MessageType`; they are optional in the spec proto definition. The absence of these fields are likely to be treated as error in the application layer, so I get why they were set as required in the serialization layer, but then it does not follow the spec definitions)~~.

I misunderstood, nvm this.

> it would be better to enclose on protobuf changes to a single file, keep same old plain types, even at at cost of manual conversion from one type to another and having 1:1 mapping of plain old types <-> pb specific types.

Aside from `PBOptions`/`results.Opt` the original protocol.nim object types seem to specify the same types from what I can see. So if I add support for `results.Opt` it does not look like any conversion would be needed.

> Why not enums?

@gmelodie the other difference is enum types. Nim enum types are not compatible with proto enum types (proto enums can have duplcate values/aliases, and unknown values). There are plans to add a proto enum type that supports proto enum semantics. But I guess adding support for custom types would imply Nim enums can be supported with caveats. For what I've seen in the spec it's probably fine to use Nim enums in proto for libp2p, though.

libp2p/protocols/rendezvous/rendezvous.nim
nitely: > new Message has dependency to protobuf_serialization

That is the point of `protobuf_serialization`. You define the proto definition in the type itself through field pragmas and field types. That's how `protobuf_serialization` is designed. The alternative is what you already have, ie: `minprotobuf`.

That said, aside from `PBOptional` (and potentially proto enums when they are added), the rest of types are "agnostic", ie: they don't require to import `protobuf_serialization` to use the Message type.

As an aside, `protobuf_serialization` can be exported so it does not need to be imported "everywhere" as I said in other comment. I don't see this as a valid point. But I think adding results.Opt and Nim enum support would address the other points.

vladopajic:
> there is a `import ...`  `export ...` 

it's not just about that. everything will be depended on this `import` although you don't really need this serialization there. 

plain types should be used without any dependency on any specific serialization method.  serialization capabilities for specific type, could be imported by `import type_serialization_protobuf` or `import type_serialization_aeron` and then you gain serialization capacity on demand. this is goal that we should strive - clean sensible imports, instead of adding unnecessary dependencies. 

`import... export...` can be anti pattern. one should thing twice when going that way.

>  in that it makes some fields "require".

this is separate topic, which can be fixed with `Opt[T]` right?

>  So, you are saying you don't want to fix this, and want to keep the wrongly define required types? 

let's find more elegant solution

nitely:
> it's not just about that. everything will be depended on this import although you don't really need this serialization there.

well, the export could be changed to just export the `PBOption` type. But sure, that won't make it compatible with other serialization libs, and it requires the `Opt -> PBOption` conversion. So it'd be better to add support for `results.Opt`, ye.

> this is separate topic, which can be fixed with Opt[T] right?

Yes, true, it is, and yes it can be fixed with Opt[T]

> > So, you are saying you don't want to fix this, and want to keep the wrongly define required types?
> 
> let's find more elegant solution

As per the [proto spec](https://protobuf.dev/programming-guides/proto2/#field-labels) the solution is to move the validation to the application layer. Wherever the field is used you would do `let val = field.valueOr(raise SomeException)` for example.

vladopajic:
> As per the [proto spec](https://protobuf.dev/programming-guides/proto2/#field-labels) the solution is to move the validation to the application layer. Wherever the field is used you would do `let val = field.valueOr(raise SomeException)` for example.

protocol spec never defines how should your app internally represent the message and mapping of internal representation of message and message sent on wire does not have to, and _should not have to_, be the same.

message inside app domain should always be the best fit for usage inside the app. the most optimal, the most elegant.

explanation

consider `ProtocolAction` as plain old nim object, that is used to represent message of some protocol. and `ProtocolActionProtobuf` is 1:1 mapping of this type used just for serialization.
```
type ProtocolAction* = object
  id*: int
  action*: ActionEnum
  funds*: Opt[int]
  note*: string
  
 type ProtocolActionProtobuf = object
  id: PBOption[int]
  action: PBOption[string]
  funds: PBOption[int]
  note*: PBOption[string]
  ```
  
note that 
1) `ProtocolAction` does not have all `Opt` fields and that is because it will be cumbersome and naive to do `field.valueOr(raise SomeException)` really whenever field is used. cumbersome should be obvious _why_, and answer to nive _why_ is found in protocol specification of "my hypothetical action protocol".
2) semantically it can be the case that protocol message is not valid if some field is not defined, even on the wire. 
for example `ProtocolAction` message is by protocol definition action that node must preform, it has unique `id` that guaranties node not to execute more then 1 of these actions in 30days interval. what does that really mean? it means that this `id` must never be `Opt`, because how is node going to know what this `id` is? it can default to 0, but if 0 is valid value for `id` then it cant default to. therefore `id` _must not_  be opt. and why should app do `field.valueOr` in this case? it would be nonsense... 
similarly `action` filed must always be defined, because how is node going to know what action to run. but note that message on wire is string, for example "get", "store", "discard", etc... why would application bother with checking the string all the time, if there are 5 actions at most, it can easily be enum, and values could be represented with `uint8` type internally. and developer is doing this intentionally to harness power of type system as use of `case-of` constructs, for example...
then there is  `funds` filed which is optional. but that is because of how protocol is defied, and for example, protocol defies that funds can be omitted from message, in which case node should do action differently then case when funds are known. "before doing action, obtain current funds, then do the action" .
and finally `note`, which is as per protocol spec user defined note about action. and this is purely developers choice to represent message as just `string` because absence of value can be represented using `""` or `Opt.none(string)`, but for simplicity and elegance developer choose first. 


--- 

protobuf simply wants to delegate, raise up, parsing/validation of message from it's own responsibility to responsibility of app. and that in context of our story means this:

```nim

proc decode*(_: type[ProtocolAction], data: seq[byte]): Result[ProtocolAction, string] =
   # does whatever is necessary to parse data and validate according to correct use in 
   # language type system and as per protocol definition.
    
proc handler(data: seq[byte]) {.async: (raises: [CancelledError]).} =
   let action = ProtocolAction.decode(data).valueOr:
      # log error or whatever
      return 
    
   # after parse, application has handled anything to do with message 
   # validation and representation of message in application and language type system
   # for example, every use of `ProtocolAction.action` from here must assume that only correct
   # values will ever be here, because `decode` has already handled decoding of message and 
   # validation to type system representation. 
   
   
   # then `ProtocolAction` from here will get propagated to potentially different corners of app.
   # and imagine that every corner of that app has to `field.valueOr` or
   # be depended to `protobuf_serialization` unnecessary
   
```

nitely:
the scope of this PR is to replace minprotobuf usage with protobuf-ser, if you want to add a layer that does further validations/conversion that's fine. I can keep the required fields to do a 1-1 refactor and those can be turned into Opt to follow the libp2p spec in follow up PRs/tasks that implements something like what you suggest.

libp2p/protocols/rendezvous/protobuf.nim
```diff
@@ -1,268 +1,63 @@
...
-  if ?pb.getField(6, pbdr).optValue:
-    msg.discoverResponse = DiscoverResponse.decode(pbdr.buffer)
-    if msg.discoverResponse.isNone():
-      return Opt.none(Message)
-
-  Opt.some(msg)
+  MessageType* = uint32
+  ResponseStatus* = uint32
+
+  Cookie* {.proto2.} = object
```
moigagoo, line 35: I can't find Cookie definition in https://github.com/libp2p/specs/blob/master/rendezvous/README.md#protobuf

Also, this type isn't used anywhere else in this module.

Why is Cookie defined?

nitely: it seems the cookie is just bytes and what those bytes contain can be anything.  It's returned (as bytes) from some API for book keeping in following requests; ie: it's for internal use, kinda like an opaque type.

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -558,20 +585,22 @@ proc new*(
     try:
       let
         buf = await conn.readLp(4096)
-        msg = Message.decode(buf).tryGet()
-      case msg.msgType
-      of MessageType.Register:
+        msg = Protobuf.decode(buf, Message)
+      case msg.msgType.get()
```
richard-ramos: There's a change of behavior here.
According to the docs of protofub serialization:
> PBOptions are a generic type instantiated with the default value for that field. They serve as regular Options, except when they're none, they still return a value when get is called (the default value). 

This means that the `else` case in line 682 will never get executed, and instead it will enter line  599.

Probably it is a good idea to add a tryGet proc to the PBOption in nim-protobuf-serialization so you can do:
```
case msg.msgType.tryGet()
```
otherwise, an explicit check with `isSome` is required to be sure that we arent using the default value

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -269,10 +281,10 @@ proc advertisePeer[E](
       await conn.writeLp(msg)
       let
         buf = await conn.readLp(4096)
-        msgRecv = Message.decode(buf).tryGet()
-      if msgRecv.msgType != MessageType.RegisterResponse:
+        msgRecv = Protobuf.decode(buf, Message)
+      if msgRecv.msgType.get() != MsgTypeRegisterResponse:
         trace "Unexpected register response", peer, msgType = msgRecv.msgType
-      elif msgRecv.registerResponse.tryGet().status != ResponseStatus.Ok:
+      elif msgRecv.registerResponse.get().status.get() != ResponseOk:
```
richard-ramos: Similar to https://github.com/vacp2p/nim-libp2p/pull/2166/changes#r3149917690 , here code is using the default value when using `get`, so the status will be `ResponseOk`.

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -364,39 +381,47 @@ proc requestPeer[E](
...
+  if msgRcv.msgType.get() != MsgTypeDiscoverResponse:
     debug "Unexpected discover response", msgType = msgRcv.msgType
     return @[]
-  let resp = msgRcv.discoverResponse.valueOr:
+  if msgRcv.discoverResponse.isNone():
     debug "Discover response is empty"
     return @[]
-  if resp.status != ResponseStatus.Ok:
+  let resp = msgRcv.discoverResponse.get()
+  if resp.status.get() != ResponseOk:
```
richard-ramos: Same case as https://github.com/vacp2p/nim-libp2p/pull/2166/changes#r3149923962

richard-ramos (commented):
+1 to the observations of  @vladopajic and @gmelodie . `PBOption` leaking is not the way to go cause now call sites will need to know about protobuffer optional semantics a well as dealing with `get()`, since it is no longer a "field was present", but instead, a "get current value or default". 

> ^ Required fields are switched to optional to follow the spec. Also, note required fields are [discouraged](https://protobuf.dev/programming-guides/proto2/#field-labels). The user can send the default value in required fields, which is the same as an absent default value in optional fields. So is this default(T) value valid or not? if not, then the application must guard against this value; if yes, then why make it required?

I disagree with the "same as absent" part, cause for protocol logic, there are different questions to answer:
1. Is the field present?
 2. Is the value valid?
 3. If absent, does the protocol define a default behavior?

The `.get()` tries to solve question 1 and 3 with a single function, but this is somewhat sketchy. This is ok for when absence means use the default, but that is not always the case,

i.e. for protobuf, https://github.com/libp2p/specs/blob/master/rendezvous/README.md#protobuf  the default `MessageType` is `REGISTER`. so if you have a protobuffer that looks like this:
```
Message(msgType = absent, register = absent)
```
The decode would report you a non valid object when it reaches the 'protocol logic' layer. The default would indicate that this msg type is a `Register` (due to being default), but the payload is missing, thus we should reject this.

Or another case:
```
RegisterResponse(status = 0k)
```
`Ok` is a valid value (represented by a `0`), but if it was absent, it would not neccesarily be the same as an explicit `Ok`. If we were to treat absent status as `Ok`, then malformed responses could become succesful responses.

`required` fields are indeed discoraged, but not necessarily should make everything optional and use defaults instead. 

Perhaps the solution is to have two layers:
- lets keep the old public rendezvous types:
- and have protobuf specific objects.
You would then create encode/decode procs that would convert protocol layer objects from/to protobuf types

libp2p/protocols/rendezvous/protobuf.nim
```diff
@@ -23,246 +31,122 @@ type
...
+proc encode*(d: Discover): seq[byte] =
+  Protobuf.encode(d)
+
+proc encode*(dr: DiscoverResponse): seq[byte] =
+  Protobuf.encode(dr)
+
+proc encode*(msg: Message): seq[byte] =
+  Protobuf.encode(msg)
+
+proc decodeCookie(buf: seq[byte]): Opt[Cookie] =
```
nitely: all the `decodeXxx` exist to force the protobuf/serialization mixins resolution in this module. It should be properly fixed in nim-serialization I think (idk if it's possible). But they are needed for now to avoid the porotobuf imports at call-site (the outer most call-site that is not a generic).

libp2p/protocols/rendezvous/protobuf.nim
```diff
@@ -23,246 +31,122 @@ type
...
+  try:
+    Opt.some(Protobuf.decode(buf, Cookie))
+  except SerializationError:
+    Opt.none(Cookie)
```
vladopajic: can't the be generic?

```
proc decode[T](buf: seq[byte]): Opt[T] =
  try:
    Opt.some(Protobuf.decode(buf, T))
  except SerializationError:
    Opt.none(T)
```

then `decode` proc on any type would be: 

```
proc decode*(_: type Cookie, buf: seq[byte]): Opt[Cookie] =
  decode[Cookie](buf)
```
saves a lot of manual work

tests/libp2p/discovery/test_rendezvous_protobuf.nim
```diff
@@ -3,7 +3,8 @@
 
 {.used.}
 
-import ../../../libp2p/[protocols/rendezvous/protobuf, protobuf/minprotobuf]
+import protobuf_serialization
```
vladopajic: @vacp2p/p2p, i think we do not need `test_randezvouse_protobuf` anymore. purpose of these files is now obsolete, because these files tested encoding/decoding which was manual code before. now this code is done by library, so we don't have to test that, we should trust library that it is correct.

it only makes sense to have these tests when encoding/decoding is not directly delegating to `Protobuf.encode|decode`. case when there is intermediary type. but in this case seems unnecessary. 

at most we can have 1 roundtrip test per type:
```
let originalCookie = ........ create with all props set (is there generic way to generate type with random vals?)
check Cookie.decode(originalCookie.encode()) == originalCookie
```
asserting specific properties seems too much work for nothing

vladopajic (approved):
lgtm! looks very clean

tests/libp2p/discovery/test_rendezvous.nim
```diff
@@ -20,6 +20,7 @@ import
 import ../../tools/[lifecycle, topology, unittest, crypto]
 import ./utils
 
+# XXX protobuf_ser
```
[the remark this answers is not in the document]
vladopajic: +1, comments always need to make sense to person that has never looked into code before. if it is marker it need to explain it, or have link to issue if it is TODO.

gmelodie (approved):
nice!!

richard-ramos (approved):

libp2p/protocols/rendezvous/protobuf.nim
vladopajic: since `decode[T]` looks like it's going to be pattern then:
- this can be some general utility in nim-libp2p, or
- this could be utility in protobuf repo (`Protobuf.decodeAsOpt` or whatever) ?

nitely: > can't the be generic?

That seems to run into [the generic sandwich issue](https://github.com/nim-lang/Nim/issues/11225). The nim-serialization/protobuf mixins will get resolved at call-site, instead of the `rendezvous/protobuf.nim` module. Apparently, the `type[Cookie]` is a generic even though it looks like a concrete type. It's also mentioned in [the style guide](https://status-im.github.io/nim-style-guide/language.objconstr.html#cons) cons list. The workaround is using `decodeXxx`. Another workaround is using `bind`, but I tried that and it does not work (probably a Nim bug), plus if a new mixin is added to nim-serialization/protobuf then it will break here (ie: it'll need to be added to the `bind` list). Another workaround is exporting protobuf_serialization.

what we can do is:

```nim
proc decode[T](buf: seq[byte]): Opt[T] =
  try:
    Opt.some(Protobuf.decode(buf, T))
  except SerializationError:
    Opt.none(T)

proc decodeCookie(buf: seq[byte]): Opt[Cookie] =
  decode[Cookie](buf)

proc decode*(_: type Cookie, buf: seq[byte]): Opt[Cookie] =
  decodeCookie(buf)

proc decodeRegister(buf: seq[byte]): Opt[Register] =
  decode[Register](buf)

proc decode*(_: type Register, buf: seq[byte]): Opt[Register] =
  decodeRegister(buf)

# etc...
```

vladopajic: > what we can do is

still tiny bit better

vladopajic: can `proc decode[T]` be template?

vladopajic: hmmm i think i get it...

vladopajic: tbh whatever you decide would be fine, even that suggested example is okay

nitely: I've added `serializerFor` to generate encode/decode and avoid the code repetition. It could be added to nim-serialization maybe, but  I'll see if I can find a better solution.

vladopajic: this is perfect! i think `serializerFor` can stay here as utility for now.

tests/libp2p/discovery/test_rendezvous.nim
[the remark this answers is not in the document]
nitely: I updated this to use protobuf_serialization

tests/libp2p/discovery/test_rendezvous_protobuf.nim
nitely: ok, I left only 1 roundtrip test per type.

libp2p/peerid.nim
```diff
@@ -234,3 +235,32 @@ func getField*(
...
+) {.raises: [IOError].} =
+  writeField(stream, field, value.data, pbytes, skipDefault)
+
+proc readFieldInto*(
+    stream: InputStream,
+    value: var PeerId,
+    header: FieldHeader,
+    ProtoType: type ProtobufExt,
+): bool {.raises: [SerializationError, IOError].} =
+  readFieldInto(stream, value.data, header, pbytes)
```
nitely: this is missing the `peerId.init(buffer)` validation. I'll add it.

arnetheduck:
> The nim-serialization/protobuf mixins will get resolved at call-site

What we tend to recommend for nim-serialization code is to re-export all the nim-serialization/protobuf mixins will get resolved at call-sitel serialization code consistently, [like so](https://github.com/status-im/nimbus-eth2/blob/102be79c06b408c4901fc73d92cbc8397c39425f/beacon_chain/spec/eth2_ssz_serialization.nim#L24) and [so](https://github.com/status-im/nim-web3/blob/288f1ee57a510c6a722264083531bdd993bb5ba7/web3/conversions.nim#L29), indeed to ensure that the basic stuff is always available at the instantiation site.

nitely:
@arnetheduck I mentioned that in a few comments; but they don't like their modules to get polluted by serialization symbols "unnecessarily", or something along those lines.

arnetheduck:
>  polluted by serialization symbols "unnecessarily", or something along those lines.

ok, there's was a lot of back and forth I indeed did not read :) 

but the heart of the matter is that nim-serialization is _designed_ to be used with import/export - any library in nim that relies on generics being expanded requires the use of export to work reliably so that the compiler sees the same symbols at the instantiation point - this is probably something we should stress more in nim-ser itself - it's also something we developed the strict auto mode for so that you don't get accidental wrong serializations depending on what's visible / imported and where..

so long story short for the libp2p project: the way to use nim-serialization in general is indeed to reexport all _relevant_ serialization code - ie if a module requires serializing `Opt` then the `results` serialization should be re-exported.

Also, does nim-pb-ser do automatic object expansion? I did not have time to review, but if it does, we should probably limit that similar to json-ser so that the automatic object mode is opt-in.

nitely:
> Also, does nim-pb-ser do automatic object expansion? I did not have time to review, but if it does, we should probably limit that similar to json-ser so that the automatic object mode is opt-in.

It does not work as in json-ser. The object requires the `{.proto2.}`/`{.proto3.}` pragma. Plus, the extension system requires the field to contain the `ext` pragma. It doesn't have the issue of using the default serialization by mistake, ex: removing a given extension import or forgetting to export it will throw a compile time error, it won't fallback to the default/generic object serializer.

> so long story short for the libp2p project: the way to use nim-serialization in general is indeed to reexport all relevant serialization code - ie if a module requires serializing Opt then the results serialization should be re-exported.

So, in this specific case, the encode/decode defined in protobuf.nim is called from a few public generic functions in rendezvous.nim, and that module would need to export protobuf-ser then. Then if other module happens to call those generics from a public generic then it will need to also export protobuf-ser, and so on and so forth, even though there is no sign of protobuf-ser being use in that module. Not sure why would that be preferred over forcing the resolution at the place where everything needed to decode/encode is there.

PR #2190, opened 2026-03-25
test(interop): update transport script for unified-testing

rlve:
## Summary
Update Transport interop test script to be compatible with new [Unified testing](https://github.com/libp2p/unified-testing) repo (see https://github.com/libp2p/unified-testing/blob/master/docs/write-a-transport-test-app.md). 

Tested against a fork https://github.com/rlve/unified-testing/pull/1 for now. Currently `python-v0.x` (ws) is failing against `nim-v1.15`.

```
╲ Collecting results...
 ▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔
  → Results:
    → Total: 55
    ✓ Passed: 52
    ✗ Failed: 3
    - python-v0.x x nim-v1.15 (ws, noise, mplex)
    - nim-v1.15 x python-v0.x (ws, noise, yamux)
    - nim-v1.15 x python-v0.x (ws, noise, mplex)

  → Total time: 00:17:05

  ✗ 3 test(s) failed
```

> [!WARNING]
>  Currently the new repo doesn't support using their actions on the implementation side, that's why `run-transport-interop` job has to be disabled and the required check turned off. 

## Affected Areas
- [x] Tests

### DIFF
--- a/.github/workflows/interop.yml
+++ b/.github/workflows/interop.yml
@@ -13,29 +13,33 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  run-transport-interop:
-    name: Run transport interoperability tests
-    runs-on: ubuntu-22.04
-    steps:
-      - name: Free Disk Space
-        # For some reason we have space issues while running this action. Likely while building the image.
-        # This action will free up some space to avoid the issue.
-        uses: jlumbroso/free-disk-space@v1.3.1
-        with:
-          tool-cache: true
+  # run-transport-interop:
+  #   name: Run transport interoperability tests
+  #   runs-on: ubuntu-22.04
+  #   steps:
+  #     - name: Free Disk Space
+  #       # For some reason we have space issues while running this action. Likely while building the image.
+  #       # This action will free up some space to avoid the issue.
+  #       uses: jlumbroso/free-disk-space@v1.3.1
+  #       with:
+  #         tool-cache: true
 
-      - uses: actions/checkout@v4
-      - uses: docker/setup-buildx-action@v3
-      - name: Build image
-        run: docker buildx build --load -t nim-libp2p-transport-head -f interop/transport/Dockerfile .
-      - name: Run tests
-        uses: libp2p/test-plans/.github/actions/run-transport-interop-test@master
-        with:
-          test-filter: nim-libp2p-transport-head
-          # without suffix action fails because "hole-punching-interop" artifacts have 
-          # the same name as "transport-interop" artifacts
-          test-results-suffix: transport-interop
-          extra-versions: ${{ github.workspace }}/interop/transport/version.json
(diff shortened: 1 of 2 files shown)


.github/workflows/interop.yml
```diff
@@ -13,33 +13,33 @@ concurrency:
...
-    name: Run transport interoperability tests
-    runs-on: ubuntu-22.04
-    steps:
-      - name: Free Disk Space
-        # For some reason we have space issues while running this action. Likely while building the image.
-        # This action will free up some space to avoid the issue.
-        uses: jlumbroso/free-disk-space@v1.3.1
-        with:
-          tool-cache: true
+  # run-transport-interop:
```
vladopajic, line 16: if we are not going to need this why keep it cmmented?

vladopajic (approved):

.github/workflows/interop.yml
rlve: I didn't remove it as I'm not sure if it's not supported by design or not supported just yet. From what I saw it should be fairly easy to modify to run it here as well.

interop/transport/main.nim
```diff
@@ -93,12 +94,10 @@ proc main() {.async.} =
...
-    echo Json.encode(
-      ResultJson(
-        handshakePlusOneRTTMillis: float(totalDelay.milliseconds),
-        pingRTTMilllis: float(pingDelay.milliseconds),
-      )
-    )
+    echo "latency:"
+    echo "  handshake_plus_one_rtt: " & $float(totalDelay.milliseconds)
+    echo "  ping_rtt: " & $float(pingDelay.milliseconds)
+    echo "  unit: ms"
```
vladopajic, line 100: what type of output it is now? 
if it is json we should use json, if it yaml. we should marshal some object with yaml

rlve: it requires yaml output to stdout

.github/workflows/interop.yml
vladopajic: Please consider always removing as much code as possible (unnecessary code). what you can do here to be more specific is:

```
# run-transport-interop:
#   currently this job is not implemented because of this and that. and it will be added later on after this and that
```

vladopajic: that is remove all comments of old implementation right away. it is dead code

interop/transport/main.nim
vladopajic: because looking in the code it is not clear what is going on. would be bettert ot have some object that is marshaled to stdout

`echo Yaml.marshal(result)`

richard-ramos (approved):

PR #2191, opened 2026-03-25
chore: add `with*` constructors to messages

vladopajic:
most often use case of creating rpc types is setting only one property:
- when `RPCMsg` is created we only want to set "control", "messages" or "subscriptions" properties
- similarly with `ControlMessage` we only want to set one property here.

 
 based on this usecase, this pr adds `with*` constructs that are handy for creating rpc types with only one field set. these helpers seem to be especially usefully when `RPCMsg` is created with `control` property, because this is the case when code becomes much more readable and compact.

### DIFF
--- a/libp2p/protocols/pubsub/rpc/messages.nim
+++ b/libp2p/protocols/pubsub/rpc/messages.nim
@@ -352,30 +352,6 @@ proc byteSize*(rpc: RPCMsg): int =
   rpc.preambleExtension.withValue(v):
     result += v.byteSize
 
-proc withPreamble*(_: typedesc[RPCMsg], preamble: seq[Preamble]): RPCMsg =
-  RPCMsg(preambleExtension: Opt.some(PreambleExtensionRPC(preamble: preamble)))
-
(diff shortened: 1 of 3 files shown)


libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -352,6 +352,61 @@ proc byteSize*(rpc: RPCMsg): int =
   rpc.preambleExtension.withValue(v):
     result += v.byteSize
 
+proc withIWant*(_: typedesc[ControlMessage], msgIds: seq[MessageId]): ControlMessage =
+  ControlMessage(iwant: @[ControlIWant(messageIDs: msgIds)])
```
richard-ramos, line 356: non actionable nitpick: you can do this to only mention the type once 
```suggestion
proc withIWant*(T: typedesc[ControlMessage], msgIds: seq[MessageId]): T =
  T(iwant: @[ControlIWant(messageIDs: msgIds)])
```

PR #2192, opened 2026-03-25
Test PR to see if claude can review - DO NOT MERGE -

richard-ramos:


libp2p/transports/quictransport.nim
```diff
@@ -192,7 +185,7 @@ method handle*(m: QuicMuxer): Future[void] {.async: (raises: []).} =
   proc handleStream(stream: QuicStream) {.async: (raises: []).} =
     ## call the muxer stream handler for this channel
     ##
-    await m.streamHandler(stream)
+    m.streamHandler(stream)
```
richard-ramos, line 188: High: removing await here changes the muxer behavior. streamHandler returns Future[void], so this call no longer waits for the handler to finish before asserting that the stream is closed. In the normal case the assertion now runs immediately and can fire before the handler closes the stream.

libp2p/transports/quictransport.nim
```diff
@@ -83,13 +83,6 @@ method readOnce*(
     except StreamError as e:
       raise (ref LPStreamError)(msg: "error in readOnce: " & e.msg, parent: e)
```
richard-ramos, line 84: High: removing the block below leaves this method without any successful return path. After the awaited read completes, the proc now resolves to the default int value instead of readLen, and it also stops marking EOF on a zero-length read. That breaks the stream read contract for every QUIC read.

PR #2195, opened 2026-03-26
test(multiaddress): getField protobuf tests

rlve:
> [!NOTE]
> Keeping the description for the context. It was decided to not proceed with the PR. Just keeping the added unit tests. 

## Summary
When tested `transport interop` tests from `unified-testing` against `nim-v1.15`. 
```
╲ Collecting results...
 ▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔
  → Results:
    → Total: 55
    ✓ Passed: 51
    ✗ Failed: 4
    - python-v0.x x nim-v1.15 (ws, noise, mplex)
    - python-v0.x x nim-v1.15 (ws, noise, yamux)
    - nim-v1.15 x python-v0.x (ws, noise, yamux)
    - nim-v1.15 x python-v0.x (ws, noise, mplex)
  → Total time: 00:16:34
  ✗ 4 test(s) failed
```

All 4 `python x nim` WS tests fail (both directions, both muxers). TCP and QUIC work fine.

#### Root cause

Python's identify response includes an empty `observed_addr` protobuf field (field 4 is present but 0 bytes) for WS connections. Which is probably not correct on their side.

`MultiAddress.init` rejects empty bytes, which crashes the whole identify with "Incorrect message received!".

#### Possible fix

When a multiaddress protobuf field is present but contains an empty bytes, treat it as "field not present" (`ok(false)`) instead of passing the empty bytes to `MultiAddress.init`

Passed run: https://github.com/rlve/unified-testing/actions/runs/23592534753/job/68701055207?pr=1

```
╲ Collecting results...
 ▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔
  → Results:
    → Total: 55
    ✓ Passed: 55
    ✗ Failed: 0

  → Total time: 00:17:35

  ✓ All tests passed!
```

### DIFF
--- a/.github/workflows/daily_ci_report.yml
+++ b/.github/workflows/daily_ci_report.yml
@@ -1,0 +1,188 @@
+name: Daily CI Failure Report
+
+on:
+  schedule:
+    - cron: "0 12 * * *"
+  workflow_dispatch:
+
+permissions:
+  issues: write
+
+jobs:
+  check-and-report:
+    name: Check daily CI and report failures
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check daily CI status and open/update issue on failure
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
+        run: |
+          python3 << 'PYEOF'
+          import os
+          import sys
+          import requests
+          from datetime import datetime, timezone, timedelta
+
+          TOKEN = os.environ["GITHUB_TOKEN"]
+          REPO = os.environ["REPO"]
+          DEFAULT_BRANCH = os.environ.get("DEFAULT_BRANCH") or "master"
+
+          HEADERS = {
+              "Authorization": f"Bearer {TOKEN}",
+              "Accept": "application/vnd.github+json",
+              "X-GitHub-Api-Version": "2022-11-28",
+          }
+          BASE = f"https://api.github.com/repos/{REPO}"
+
+          # Daily workflows to monitor
+          WORKFLOWS = [
+              ("Daily Nimbus", "daily_nimbus.yml"),
+              ("Daily amd64", "daily_amd64.yml"),
+              ("Daily i386", "daily_i386.yml"),
+          ]
+
+          LABEL_NAME = "daily-ci-failure"
+          LABEL_COLOR = "d93f0b"
+          LABEL_DESC = "Tracks daily CI job failures"
+
+          now = datetime.now(timezone.utc)
+          # Allow up to 26 hours so we always catch the 6:30 UTC scheduled run
+          cutoff = now - timedelta(hours=26)
+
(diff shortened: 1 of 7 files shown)


libp2p/multiaddress.nim
```diff
@@ -1102,7 +1102,7 @@ proc getField*(
 ): ProtoResult[bool] {.inline.} =
   var buffer: seq[byte]
   let res = ?pb.getField(field, buffer)
-  if not (res):
+  if not (res) or buffer.len == 0:
```
vladopajic: if we decide to go with this; it would be cool to have unit test for this. so we don't relay on interop

tests/libp2p/multiformat/test_multiaddress.nim
```diff
@@ -381,6 +381,48 @@ suite "MultiAddress test suite":
...
+
+  test "getField returns ok(false) for empty multiaddress":
+    var pb = initProtoBuffer()
+    pb.write(1, newSeq[byte]())
+    pb.finish()
+
+    var decoded: MultiAddress
+    check:
+      pb.getField(1, decoded).get() == false
+      decoded == MultiAddress()
```
rlve: It's this case. Previously it would return `err(ProtoError.IncorrectBlob)`.

libp2p/multiaddress.nim
rlve: Added in https://github.com/vacp2p/nim-libp2p/pull/2195/changes/3b237c928621725a787488a78fe92e80657e5702

tests/libp2p/multiformat/test_multiaddress.nim
vladopajic: tbh, i have no idea what this case should be.
logically it makes sense for both options to return `ok(false)` and `err(ProtoError.IncorrectBlob)`, tho later one is bit more logical (subjectively).

anyhow we should keep these unit tests that are added here to assert expected behavior in the future.

---

to get answer to what should we decide here, we should:
- see if other libp2p implementation are compatible here with python - because maybe it's just python making this decision up.
  - if other libp2p implementation (say go-libp2p) is passing tests with python and passing tests with nim, but nim is not passing test with python then how so?? 
- what is convention for decoding other values when bytes are @[] vs not set?

vladopajic: ok, i just read description of pr. i think we should do following:
- keep multi-address code as is
- keep unit tests we added here (matching existing logic)
- ignore `python ws` from `unified-testing`  with comment explaining why

rlve: All right, will do

PR #2198, opened 2026-03-27
fix(mix): delay correlation risks

chaitanyaprem:
## Summary

Implements the mix-delay timing-correlation mitigations discussed [here](https://forum.research.logos.co/t/timing-correlation-risks-in-mixnet-delay-strategies/677).
The change removes fixed-value artifacts at the practical upper bound and at the low end of spam-protected delays, while keeping sampling runtime bounded.

## Affected Areas

- [x] Protocol Logic - exponential mix delays sample directly from the bounded practical window instead of truncating to a fixed ceiling or relying on unbounded rejection retries.
- [x] API / Configuration - added `SpamProtectionDelayStrategy`, the recommended delay strategy when per-hop spam protection is enabled.
- [x] API / Configuration - `MixProtocol` no longer mutates caller-provided delay strategies to inject a floor.
- [x] Test Helpers - spam-protected mix test nodes now default to `SpamProtectionDelayStrategy` when no explicit delay strategy is provided.
- [x] Observability - warning text reports when proof generation exceeds the sampled delay, which matches the actual comparison in the code.

NA

## Impact on Library Users
- Users enabling per-hop spam protection should prefer `SpamProtectionDelayStrategy` to avoid timing correlation between proof generation and short exponential delays.
- `ExponentialDelayStrategy` remains generic and is no longer implicitly modified by `MixProtocol`.
- No API breakage for existing strategy types, but the spam-protection timing mitigation is now an explicit strategy choice instead of hidden protocol mutation.

## Risk Assessment
Low. The change is limited to the mix delay sampling path and the strategy selection surface, and is covered by unit tests plus spam-protection integration tests.

## Verification
- `tests/libp2p/mix/test_delay_strategy.nim`: `12/12` passed
- `tests/libp2p/mix/component/test_spam_protection.nim`: `4/4` passed
- `tests/test_all.nim -d:path=mix`: `134/134` passed

## References
https://forum.research.logos.co/t/timing-correlation-risks-in-mixnet-delay-strategies/677

### DIFF
--- a/tests/libp2p/kademlia/test_get_providers.nim
+++ b/tests/libp2p/kademlia/test_get_providers.nim
@@ -266,7 +266,9 @@ suite "KadDHT - Get Providers":
 
   asyncTest "Get providers aggregates providers from multiple peers":
     # Topology: kads[0] <-> kads[1] <-> kads[2] <-> kads[3] <-> kads[4]
-    let kads = setupKadSwitches(5)
+    # Use a long republishProvidedKeysInterval so the background republish heartbeat
+    # doesn't fire during the test and race with getProviders dials.
+    let kads = setupKadSwitches(5, republishProvidedKeysInterval = chronos.hours(1))
     startAndDeferStop(kads)
 
     await connectChain(kads)


chaitanyaprem:
Not sure why the CI keeps failing to install deps.

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -39,27 +39,69 @@ method generateForIntermediate*(
...
+    minimumDelayMs: minimumDelayMs,
+  )
+
+proc minimumDelayMs*(self: ExponentialDelayStrategy): uint16 {.inline, raises: [].} =
+  self.minimumDelayMs
+
+proc setMinimumDelayMs*(
+    self: ExponentialDelayStrategy, minimumDelayMs: uint16
+) {.inline, raises: [].} =
+  self.minimumDelayMs = minimumDelayMs
```
gmelodie: Can we not make the field public?

libp2p/protocols/mix/delay_strategy.nim
```diff
@@ -39,27 +39,69 @@ method generateForIntermediate*(
...
 type ExponentialDelayStrategy* = ref object of DelayStrategy
   ## Recommended strategy: encodes mean delay, samples from exponential distribution.
-  ## The distribution is truncated at -mean*ln(negligibleProb), discarding the
-  ## impractically long tail while preserving the mixing properties.
+  ## Samples are drawn directly from the exponential distribution conditioned on
+  ## the configured [minimumDelayMs, practicalMaxDelayMs] window. This preserves
+  ## a smooth bounded distribution without fixed spikes at either bound.
   meanDelayMs: uint16
   negligibleProb: float64
+  minimumDelayMs: uint16
```
vladopajic, line 60: instead of using units in variable names, it would be better to use `Duration` type. 

i opened issue not to block on this in this pr: https://github.com/vacp2p/nim-libp2p/issues/2206

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -87,3 +93,34 @@ suite "DelayStrategy":
...
+      )
+    var minimumDelayHits = 0
+
+    for _ in 0 ..< 10000:
+      let delay = strategy.generateForIntermediate(meanDelayMs)
+      check delay >= minimumDelayMs
+      if delay == minimumDelayMs:
+        inc minimumDelayHits
+
+    check minimumDelayHits < 500
```
vladopajic: - why `10000` here?
- why `500` here? - why not `100`? or `200`?

tests/libp2p/mix/component/test_spam_protection.nim
```diff
@@ -89,3 +97,15 @@ suite "Mix Protocol - Spam Protection":
...
+    let delayStrategy = ExponentialDelayStrategy.new(100, rng(), minimumDelayMs = 0)
+
+    let nodes = await setupMixNodes(
+      4,
+      spamProtectionRateLimit = Opt.some(10),
+      delayStrategy = Opt.some(DelayStrategy(delayStrategy)),
+    )
+    startAndDeferStop(nodes)
+
+    check delayStrategy.minimumDelayMs() == 100'u16
```
vladopajic: what's going on here? 

why is `check delayStrategy.minimumDelayMs() == 100'u16` this expected? 

could you add another test case where `spamProtectionRateLimit = Opt.some(10),` this is not used so we can see different value in this case?

libp2p/protocols/mix/mix_protocol.nim
```diff
@@ -417,6 +419,12 @@ proc proofSize(sp: Opt[SpamProtection]): int =
     return 0
   return sp.get().proofSize
 
+proc configureDelayFloor(mixProto: MixProtocol) {.inline.} =
+  if mixProto.delayStrategy of ExponentialDelayStrategy:
+    let expDelayStrategy = ExponentialDelayStrategy(mixProto.delayStrategy)
+    if mixProto.spamProtection.isSome() and expDelayStrategy.minimumDelayMs() == 0:
+      expDelayStrategy.setMinimumDelayMs(DefaultSpamProtectionDelayFloorMs)
+
```
vladopajic: mix node should not be doing this. the one who created node should do this.

for example consider code from tests:

```
    let delayStrategy = ExponentialDelayStrategy.new(100, rng(), minimumDelayMs = 0)

    let nodes = await setupMixNodes(
      4,
      spamProtectionRateLimit = Opt.some(10),
      delayStrategy = Opt.some(DelayStrategy(delayStrategy)),
    )
```

this code already knows that is using spam protection, so this code should already create `DelayStrategy` accordingly.

1) 
that is code should ether use some other delay strategy, say `ExponentialDelayStrategyWithFloor`

```
    let delayStrategy = ExponentialDelayStrategyWithFloor.new(100, rng(), minimumDelayMs = 0)

    let nodes = await setupMixNodes(
      4,
      spamProtectionRateLimit = Opt.some(10),
      delayStrategy = Opt.some(DelayStrategy(delayStrategy)),
    )
```

or 

2) is this programmer error because if one knows what `ExponentialDelayStrategy` then one should configure it accordingly that is one should not use `minimumDelayMs = 0` because one knows that is using spamProtection.

```
    let delayStrategy = ExponentialDelayStrategy.new(100, rng(), minimumDelayMs = 0)

    let nodes = await setupMixNodes(
      4,
      spamProtectionRateLimit = Opt.some(10),
      delayStrategy = Opt.some(DelayStrategy(delayStrategy)),
    )
```

vladopajic: in other words, MixProtocol should not inspect for specific type of spam protection it must work for whatever is configured. 

in the process of creation of DelayStrategy, the type in this case `ExponentialDelayStrategy` should take care of `"configureDelayFloor"`.

libp2p/protocols/mix/delay_strategy.nim
chaitanyaprem: Thanks, that was remnant from a previous attempt to expose and set this from outside and realized it is too complex. Will modify it.

libp2p/protocols/mix/mix_protocol.nim
chaitanyaprem: The idea was not to expose this complexity to the user of mix protocol, because this is a challenge/issue in the way mix protocol applies delay strategy and spam protection. 
It would be better to handle it internally, but i do get the concern you are raising. Let me think of a better way to address this.

chaitanyaprem: I have created a new DelayStrategy that is recommeded to be used for SPamProtection whic handles this logic and keeps mix protocol agnostic to it. Thanks for the comment, this makes code look cleaner.

tests/libp2p/mix/test_delay_strategy.nim
chaitanyaprem: used vars and added comments explaining why.

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -87,3 +99,54 @@ suite "DelayStrategy":
...
+  test "SpamProtectionDelayStrategy allows overriding the default delay floor":
+    let
+      meanDelayMs: uint16 = 100
+      minimumDelayMs: uint16 = 250
+      strategy = SpamProtectionDelayStrategy.new(
+        meanDelayMs, rng(), minimumDelayMs = minimumDelayMs
+      )
+
+    for _ in 0 ..< BoundarySamples:
+      check strategy.generateForIntermediate(meanDelayMs) >= minimumDelayMs
```
vladopajic: nit: yes ` >= ` will ensure this check, but test case might be satisfied with `==` if there is some bug.
and that's not what delay strategy should do, that is, tests needs to assert that some will be `>` and others `==` to minimumDelayMs. 
this is just extra check very specific to DelayStrategy testing.

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -87,3 +99,54 @@ suite "DelayStrategy":
...
+
+    check strategy.generateForIntermediate(meanDelayMs) == minimumDelayMs
+
+  test "SpamProtectionDelayStrategy applies the default delay floor":
+    let
+      meanDelayMs: uint16 = 100
+      strategy = SpamProtectionDelayStrategy.new(meanDelayMs, rng())
+
+    for _ in 0 ..< BoundarySamples:
+      check strategy.generateForIntermediate(meanDelayMs) >=
```
vladopajic: https://github.com/vacp2p/nim-libp2p/pull/2198/changes#r3014381774
similar

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -87,3 +99,54 @@ suite "DelayStrategy":
 
     for _ in 0 ..< 10000:
       check strategy.generateForIntermediate(meanDelayMs) <= maxDelayMs
+
+  test "ExponentialDelayStrategy bounds values below the configured minimum":
```
vladopajic: `below` or `above`?

tests/libp2p/mix/test_delay_strategy.nim
```diff
@@ -67,16 +73,22 @@ suite "DelayStrategy":
 
     check delays.len > NumSamples div 2
 
-  test "ExponentialDelayStrategy truncates at negligible probability threshold":
+  test "ExponentialDelayStrategy bounds values above the practical maximum":
```
vladopajic: `above` or `below`?

tests/libp2p/mix/component/test_spam_protection.nim
```diff
@@ -89,3 +89,35 @@ suite "Mix Protocol - Spam Protection":
...
+    )
+    startAndDeferStop(nodes)
+
+    var sawDelayBelowFloor = false
+    for _ in 0 ..< 100:
+      if delayStrategy.generateForIntermediate(100) < DefaultSpamProtectionDelayFloorMs:
+        sawDelayBelowFloor = true
+        break
+
+    check sawDelayBelowFloor
```
vladopajic: this test case is not needed. we should guard this with code review.
there is no guarantee that even if one mutates  `DelayStrategy` in mix that they will mutate original reference (passed down to `setupMixNodes`) because one can create copy, then mutate copy.

tests/libp2p/mix/component/test_spam_protection.nim
```diff
@@ -89,3 +89,35 @@ suite "Mix Protocol - Spam Protection":
...
+    let nodes = await setupMixNodes(
+      4,
+      spamProtectionRateLimit = Opt.some(10),
+      delayStrategy = Opt.some(DelayStrategy(delayStrategy)),
+    )
+    startAndDeferStop(nodes)
+
+    for _ in 0 ..< 100:
+      check delayStrategy.generateForIntermediate(100) >=
+        DefaultSpamProtectionDelayFloorMs
```
vladopajic: why shouldn't it work ?

libp2p/protocols/mix/mix_protocol.nim
vladopajic: thanks prem, this is much better

vladopajic (approved):

PR #2200, opened 2026-03-29
feat(kad): add client mode for capability discovery without Kad-DHT participation

tinniaru3005:
## Summary

This PR introduces client mode for capability discovery, allowing nodes to discover peers for specific services without participating in the Kad-DHT.

## Affected Areas

- [ ] Gossipsub  
  

- [ ] Transports  
  

- [ ] Protocol Logic

- [ ] Build / Tooling

- [ ] Other  
  

- **Nimbus:**  
  

- **Waku:**  
  

- **Codex:**


SionoiS (approved):
If it works, it works but it's a bit of a weird way to do client mode.

The best would be a new type which cannot call functionality that don't exist in client mode.

Approved because libp2p guys may have even better ideas. 🤷

vladopajic:
> If it works, it works but it's a bit of a weird way to do client mode.
> 
> The best would be a new type which cannot call functionality that don't exist in client mode.
> 
> Approved because libp2p guys may have even better ideas. 🤷

yes, it appears that separate type should be cleaner approach. but then code reuse between these two types might require more effort.
 
 ---

other note is that functions (methods, procs) of kad in client mode, in current logic, better:
`doAssert clientMode, "not supported in client mode"`

handling of unsupported/undefined behavior of client mode must raise error because if developer is calling such methods it means that there is issue with the code. returning error `err "msg"` can be seen as normal flow. semantically these two have different meaning.

then we see difference that one can call `start()` but one can't call `startAdvertising()`, `start` will early return but `startAdvertising` will return error. there is inconsistency here. 

i don't have strong opinion of going with separate type, although it could be cleaner if implemented correctly, but client mode procs and methods better raise errors do to better semantics.

libp2p/protocols/kad_disco.nim
```diff
@@ -84,6 +84,7 @@ proc new*(
     services: toHashSet(services),
     discoConf: discoConf,
     xprPublishing: xprPublishing,
+    clientMode: client,
```
vladopajic: why not

`clientMode: clientMode,`

?

gmelodie (approved):

libp2p/protocols/kad_disco.nim
tinniaru3005: Good point — `clientMode: clientMode` would be more explicit. I used client since that’s the constructor parameter, but I can rename it for better clarity and consistency.

libp2p/protocols/kademlia.nim
```diff
@@ -75,7 +75,7 @@ proc new*(
   kad.updatePeers(bootstrapNodes)
 
   kad.codec = codec
-  if client:
+  if clientMode:
     return kad
```
richard-ramos: What would be the behavior of KadDHT client mode, if i call start() and stop() on it?

libp2p/protocols/kad_disco.nim
jm-clius, line 1: I think we can probably uncommit the changes to these files where only comments are added.

libp2p/protocols/kademlia.nim
jm-clius, line 1: Same here. I don't think we need to commit changes that are only a few comments. I'd discard changes to this file.

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -14,6 +14,12 @@ import ./[types, serviceroutingtables, service_discovery_metrics]
 logScope:
   topics = "service-disco advertiser"
 
+proc isClientMode(disco: KademliaDiscovery): bool {.inline, raises: [].} =
+  disco.handler.isNil
```
jm-clius: Wait, isn't there already a `disco.client` boolean? It will be cleaner than inferring this from just the handler being unset.

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -35,6 +41,8 @@ proc sendRegister*(
 .} =
   ## Send REGISTER request to a peer
 
+  requireServerMode(disco, "sendRegister")
```
jm-clius: I don't think we need either the `requireServerMode()` or `isClientMode()` indirections. Why not simply check if `disco.client` is true and `return err` when it is?

jm-clius (changes requested):
Thanks! :) I'm glad that there's now no unnecessary constructs related to service discovery clients introduced in the kademlia and kad_disco files. However, I'm not quite sure I see the need for all the indirection: unless I'm missing something, checking if `disco.client` is true should be a simple way to return an error on functions that are not supposed to be called by clients.

libp2p/protocols/kademlia.nim
tinniaru3005: yup removed :)

libp2p/protocols/kad_disco.nim
tinniaru3005: yup sone. Thanks :)

libp2p/protocols/service_discovery/advertiser.nim
tinniaru3005: right now `disco.client` does not seem to exist, which is why trying to access it from `advertiser.nim` gives undeclared field error. `disco` is of type [`KademliaDiscovery`](https://github.com/vacp2p/nim-libp2p/blob/1e7e90351cd204586747dd9a7db40241fde53548/libp2p/protocols/kademlia_discovery/types.nim#L17)  which is defined as `ref object of KadDHT`. But it doesn't have any `client` field which is being set not is it inheriting that from any parent object. `client` appears as a parameter in the [constructor](https://github.com/vacp2p/nim-libp2p/blob/1e7e90351cd204586747dd9a7db40241fde53548/libp2p/protocols/kademlia.nim#L52). `client` is beng used to return early [here](https://github.com/vacp2p/nim-libp2p/blob/1e7e90351cd204586747dd9a7db40241fde53548/libp2p/protocols/kademlia.nim#L78) before setting `kad.handler`and it is not persisted as object state. 

To use `disco.client` we have to most probably add a field `clientMode` kademlia [`type.nim`](https://github.com/vacp2p/nim-libp2p/blob/1e7e90351cd204586747dd9a7db40241fde53548/libp2p/protocols/kademlia_discovery/types.nim#L17)  file and set it in `kad_disco.nim` and `kademlia.nim` files. But always open to any other ways that might be cleaner.

jm-clius: Ah, apologies! You're right. I hallucinated a `client` property for `disco`. :) Yes, I think the right approach would be to just persist `client` as a new field (I think that would be preferable as a field name to `clientMode`) for the `KadDHT` object. Since `KademliaDiscover` is a `ref object of KadDHT` it will inherit all the fields already existing in `KadDHT`, so you should be able to just read the `client` field then.

tinniaru3005: yup done :)

libp2p/protocols/service_discovery/advertiser.nim
tinniaru3005: ya done:)
Thanks a lot for the suggestion

jm-clius (approved):
Thanks!

richard-ramos (approved):

PR #2211, opened 2026-04-01
chore(test): utilize rng() template

vladopajic:
instead of using default argument value for random (aka `rng`) that is creating new random, this pr is passing already created random from singleton variable (accessed using template `rng`).  

this fix fixes issue with test, but https://github.com/vacp2p/nim-libp2p/issues/2212 was opened to make changes to avoid any issues in the future (potential issues).


tests/libp2p/pubsub/component/test_gossipsub_extensions.nim
```diff
@@ -183,12 +183,10 @@ suite "GossipSub Component - Extensions":
           sendIDontWantOnPublish = true,
         )
         .toGossipSub()
-      n0 = nodes[0]
-      n1 = nodes[1]
```
vladopajic, line 187: removing unsued n0 value.

tests/libp2p/pubsub/utils.nim
```diff
@@ -212,13 +214,15 @@ proc generateNodes*(
 ): seq[PubSub] =
   for i in 0 ..< num:
     let switch = newStandardSwitch(
+      rng = rng(),
```
vladopajic, line 217: because rng argument is not explicitly set, `newRng()` will be called as default value for argument. calling `newRng()` can exhaust seed entropy. 

in general we should avoid having default value for `rng` because, default values are optional, do we can forget to initialize them, we can cause issues down the line.

PR #2216, opened 2026-04-01
fix: nil pointer dereference issues caused when sha256 is used in ref object

vladopajic:
## Summary

fixes issue when nil pointer dereference issues caused when sha256 is used in ref object

## Affected Areas

- [x] Gossipsub 

n/a

## Impact on Library Users

n/a


libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -237,7 +237,7 @@ method publish*(
 method initPubSub*(f: FloodSub) {.raises: [InitializationError].} =
   procCall PubSub(f).initPubSub()
   f.seen = TimedCache[SaltedId].init(2.minutes)
-  f.seenSalt.init()
```
arnetheduck, line 240: fyi, depending on the nimcrypto version used, this code is impacted by https://github.com/cheatfate/nimcrypto/issues/98 - ie 0.7.* series of nimcrypto is broken when used inside a ref object, like FloodSub

vladopajic: thanks, is workaround possible? 
- could nimcrypto avoid this pragma? 
  - could nimcrupto avoid this pragma on specific platforms?
- or would it be okay for nim libp2p to use older version of nimcrypto?
- or ?

arnetheduck: we're discussing the options but none that are good exist for now (the pragma is needed to use hardware extensions but it only works for stack vars in refc)

> or would it be okay for nim libp2p to use older version of nimcrypto?

this is what nimbus does, for now - ie technically it's the downstream application that chooses versions, not libp2p (within the bounds that libp2p establishes in its nimble file) .. in this particular case, only refc is broken, so it's tricky and part of the reason the options are bad.

arnetheduck: btw, one option that exists is to allocate the memory for the context in a local variable on the stack and not keep it in pubsub, if that's possible .. otherwise, it's also possible to allocate aligned memory manually, ie with malloc (and release on close)

vladopajic: > local variable on the stack

global variable :) 

please review

vladopajic: only issues i see is that all gossipsubs in same process will have same salt (with global var). but this might not be issue?

vladopajic: opt1: global var
https://github.com/vacp2p/nim-libp2p/pull/2216/changes/1714b7b6e2f4f56cafc574ad6dca586b6014c579

opt2: local var
https://github.com/vacp2p/nim-libp2p/pull/2216/commits/c98e2cd0c1953d3cf509b74909628adb9500b494
( overhead is one more `.update()` )

vladopajic: ^ @arnetheduck

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -32,14 +32,14 @@ type FloodSub* {.public.} = ref object of PubSub
     # We use a salted id because the messages in this cache have not yet
     # been validated meaning that an attacker has greater control over the
     # hash key and therefore could poison the table
-  seenSalt*: sha256
-    # The salt in this case is a partially updated SHA256 context pre-seeded
-    # with some random data
+  seenSalt: array[32, byte] # random data used as salt
```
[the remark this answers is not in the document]
vladopajic: no one should be accessing that; `seenSalt` was previously unintentionally exposed.

gmelodie (approved):

libp2p/protocols/pubsub/floodsub.nim
arnetheduck: >  gossipsubs in same process will have same salt 

this should be fine, ie the salt is there to prevent remote attacks against the hashing method used in Table (collision attacks).

PR #2240, opened 2026-04-07
test(interop): GossipSub - core instructions (part 1)

rlve:
## Summary

Adds a `GossipSub` interop test peer, a binary that implements the [gossipsub-interop test contract](https://github.com/libp2p/test-plans/blob/master/gossipsub-interop/test-specs/implementation.md) so `nim-libp2p` can be integrated.

It reads a JSON script (same format used by `go-libp2p` and `rust-libp2p` peers), executes instructions sequentially against a `GossipSub` node, and outputs structured JSON events to stdout for the analysis pipeline.

This PR covers the core (non-partial-message) instructions with tests.

## Affected Areas
- [x] Interop
  New `interop/gossipsub/` directory.

No library code is modified. No downstream impact.

## Impact on Library Users

None.


tests/interop/test_gossipsub.nim
```diff
@@ -0,0 +1,267 @@
...
+  proc getAddr(node: GossipSub): MultiAddress =
+    node.switch.peerInfo.addrs[0]
+
+  asyncTest "script runs connect + subscribe + publish":
+    await setupNodes()
+    defer:
+      await teardownNodes()
+
+    const topic = "foobar"
+    var receivedMsgId: string
```
vladopajic: `receivedMsgId: string` ->  r`eceivedMsgIdFut: Future[string]`
would be bit nicer becasue we can the just wait on future.
in general, we should avoid `checkUntilTimeout` as much as possible.

interop/gossipsub/src/lib.nim
```diff
@@ -0,0 +1,121 @@
...
+    listenAddr: MultiAddress,
+    gossipSubParams: GossipSubParams = GossipSubParams.init(),
+): GossipSub =
+  let privKey = nodePrivKey(nodeId)
+
+  let switch = SwitchBuilder
+    .new()
+    .withRng(rng())
+    .withAddresses(@[listenAddr])
+    .withPrivateKey(privKey)
```
vladopajic:
```suggestion
    .withPrivateKey(nodePrivKey(nodeId))
```

interop/gossipsub/src/runner.nim
```diff
@@ -0,0 +1,113 @@
...
+# Forward declaration
+proc executeInstruction*(runner: ScriptRunner, instruction: ScriptInstruction) {.async.}
+
+proc executeConnect(runner: ScriptRunner, connectTo: seq[int]) {.async.} =
+  for targetId in connectTo:
+    let targetPid = nodePeerId(targetId)
+    try:
+      let targetAddr = runner.resolveAddr(targetId)
+      await runner.node.switch.connect(targetPid, @[targetAddr])
+    except Exception as e:
```
richard-ramos: CancelledError shouldnt be captured

interop/gossipsub/src/runner.nim
```diff
@@ -0,0 +1,113 @@
...
+  if target > now:
+    await sleepAsync(target - now)
+
+proc executeSubscribeToTopic(runner: ScriptRunner, topicId: string) {.async.} =
+  let logStream = runner.logStream
+
+  proc handler(topic: string, data: seq[byte]) {.async.} =
+    if data.len >= 8:
+      let msgId = extractMsgId(data)
+      logReceivedMessage(logStream, $msgId, topic)
```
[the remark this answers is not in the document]
richard-ramos: This is correct

interop/gossipsub/src/runner.nim
```diff
@@ -0,0 +1,113 @@
...
+
+proc executeSetTopicValidationDelay(
+    runner: ScriptRunner, validationTopicID: string, delaySeconds: float64
+) {.async.} =
+  let delay = milliseconds(int64(delaySeconds * 1000.0))
+  runner.node.addValidator(
+    @[validationTopicID],
+    proc(topic: string, message: messages.Message): Future[ValidationResult] {.async.} =
+      await sleepAsync(delay)
+      return ValidationResult.Accept,
```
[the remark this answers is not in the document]
richard-ramos: This is correct too

interop/gossipsub/src/instructions.nim
```diff
@@ -0,0 +1,133 @@
...
+  )
+
+proc parseWaitUntil(j: JsonNode): ScriptInstruction =
+  ScriptInstruction(kind: WaitUntil, elapsedSeconds: j["elapsedSeconds"].getInt())
+
+proc parseSubscribeToTopic(j: JsonNode): ScriptInstruction =
+  ScriptInstruction(
+    kind: SubscribeToTopic,
+    topicID: j["topicID"].getStr(),
+    partial:
```
richard-ramos, line 116: This does not seem to be used?

interop/gossipsub/src/runner.nim
```diff
@@ -0,0 +1,113 @@
...
+    publishTopicID: string,
+    messageSizeBytes: int,
+    publishMessageID: int,
+) {.async.} =
+  var data = newSeq[byte](messageSizeBytes)
+  let msgIdU64 = uint64(publishMessageID)
+  data[0 ..< 8] = toBytesBE(msgIdU64)
+  try:
+    discard await runner.node.publish(publishTopicID, data)
+  except CatchableError as e:
```
richard-ramos, line 75: This is capturing CancelledError too

interop/gossipsub/src/instructions.nim
```diff
@@ -0,0 +1,133 @@
...
+      if j.hasKey("partial"):
+        j["partial"].getBool()
+      else:
+        false,
+  )
+
+proc parsePublish(j: JsonNode): ScriptInstruction =
```
vladopajic, line 123: in general, it would be much nicer if all types represented by JsonNode are parsed directly to it's type representation. i am not really familiar with nim, but something similar to this:

```
var parsed: ScriptInstruction
jsonDeserialize(j, parsed)
return parsed
```

interop/gossipsub/src/instructions.nim
```diff
@@ -0,0 +1,133 @@
...
+    IfNodeIDEquals = "ifNodeIDEquals"
+    WaitUntil = "waitUntil"
+    SubscribeToTopic = "subscribeToTopic"
+    Publish = "publish"
+    SetTopicValidationDelay = "setTopicValidationDelay"
+
+  ScriptInstruction* = object
+    case kind*: InstructionKind
+    of InitGossipSub:
+      gossipSubParams*: JsonNode
```
vladopajic: consider not using `JsonNode` in type here.  `JsonNode` is generic type and can represent anything, an elephant for example, but we only want `gossipSubParams` type to be stored here. that is, better keep type safety all across.

interop/gossipsub/src/instructions.nim
vladopajic: in general, `JsonNode` we should probably never used directly.

interop/gossipsub/src/runner.nim
```diff
@@ -0,0 +1,113 @@
...
+  let target = runner.startTime + seconds(elapsedSeconds)
+  let now = Moment.now()
+  if target > now:
+    await sleepAsync(target - now)
+
+proc executeSubscribeToTopic(runner: ScriptRunner, topicId: string) {.async.} =
+  let logStream = runner.logStream
+
+  proc handler(topic: string, data: seq[byte]) {.async.} =
+    if data.len >= 8:
```
vladopajic, line 55: why `if data.len >= 8:`  through the code?

interop/gossipsub/src/runner.nim
```diff
@@ -0,0 +1,149 @@
...
+proc executePublish(
+    runner: ScriptRunner,
+    publishTopicID: string,
+    messageSizeBytes: int,
+    publishMessageID: int,
+) {.async.} =
+  var data = newSeq[byte](messageSizeBytes)
+  let msgIdU64 = uint64(publishMessageID)
+  data[0 ..< 8] = toBytesBE(msgIdU64)
+  try:
```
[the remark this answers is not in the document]
vladopajic: probably `doAssert` ...

interop/gossipsub/src/runner.nim
```diff
@@ -0,0 +1,113 @@
...
+    discard await runner.node.publish(publishTopicID, data)
+  except CatchableError as e:
+    warn "Publish failed", messageID = publishMessageID, error = e.msg
+
+proc executeSetTopicValidationDelay(
+    runner: ScriptRunner, validationTopicID: string, delaySeconds: float64
+) {.async.} =
+  let delay = milliseconds(int64(delaySeconds * 1000.0))
```
vladopajic: not sure how chronos does it but:

delaySeconds.seconds.toMilliseconds

step 1 create duration 
step 2 from duration return in whatever unit you need it

interop/gossipsub/config.nims
```diff
@@ -0,0 +1,8 @@
+# begin Nimble config (version 2)
+when withDir(thisDir(), system.fileExists("nimble.paths")):
+  include "nimble.paths"
+# end Nimble config
+
+# Send chronicles logs to stderr so stdout is reserved for structured JSON output
+switch("define", "chronicles_sinks=textlines[stderr]")
+switch("define", "chronicles_log_level=INFO")
```
vladopajic, line 10: can we use all fancy `config.nims` from root? like UnusedImport, UnreachableCode etc...

interop/gossipsub/src/instructions.nim
rlve: Not yet, will be in the next PR.

interop/gossipsub/src/runner.nim
```diff
@@ -0,0 +1,127 @@
...
+      warn "Connect failed", target = targetId, error = e.msg
+
+proc executeIfNodeIDEquals(
+    runner: ScriptRunner, nodeID: int, inner: ScriptInstruction
+) {.async.} =
+  if nodeID == runner.nodeId:
+    await runner.executeInstruction(inner)
+
+proc executeWaitUntil(runner: ScriptRunner, elapsedSeconds: int) {.async.} =
+  await sleepAsync(elapsedSeconds.seconds())
```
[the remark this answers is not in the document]
rlve: correct

interop/gossipsub/src/instructions.nim
```diff
@@ -0,0 +1,169 @@
...
+    case kind*: InstructionKind
+    of InitGossipSub:
+      gossipSubParams*: GossipSubParams
+    of Connect:
+      connectTo*: seq[int]
+    of IfNodeIDEquals:
+      nodeID*: int
+      inner*: ref ScriptInstruction
+    of WaitUntil:
+      elapsedSeconds*: int
```
vladopajic: i competently understand why this field was named like this. 

however i would prefer filed like this:
`elapsed*: Duration` because all duration values should better use Duration type and units should not be in variable name.

we can easily parse  ` j["elapsedSeconds"].getInt())` to Duration as set it. then avoiding any conversion later on.

if you decide to go with this please consider similar change with `delaySeconds`.

tests/interop/test_gossipsub.nim
```diff
@@ -0,0 +1,269 @@
...
+    multiaddress,
+    peerid,
+    protocols/pubsub/gossipsub,
+    protocols/pubsub/rpc/message,
+    switch,
+  ]
+import ../tools/[unittest]
+import ../../interop/gossipsub/src/[node, instructions, runner]
+
+suite "GossipSub Interop":
```
vladopajic, line 18: maybe split in files?
- `test_gossipsub_[json/parsing/serializing...]`
- `test_gossipsub_runner`

tests/interop/test_gossipsub.nim
```diff
@@ -0,0 +1,269 @@
...
+      params.heartbeatInterval == 2.seconds
+      params.fanoutTTL == 1.minutes
+      params.pruneBackoff == 30.seconds
+      params.unsubscribeBackoff == 10.seconds
+      params.historyLength == 10
+      params.historyGossip == 5
+      params.dLazy == 8
+      params.gossipFactor == 0.5
+
+suite "GossipSub Interop - Script runner":
```
vladopajic, line 169: would it make sense to add some other test like, connect to node that does not exist... 

or in general some `ScriptInstruction` that:
- are invalid (can't connect to node that does not exist)
- that produce no effect (like publishing on topic that does not exists etc....)

tests/interop/test_gossipsub.nim
```diff
@@ -0,0 +1,269 @@
...
+  proc getAddr(node: GossipSub): MultiAddress =
+    node.switch.peerInfo.addrs[0]
+
+  asyncTest "script runs connect + subscribe + publish":
+    await setupNodes()
+    defer:
+      await teardownNodes()
+
+    const topic = "foobar"
+    let receivedMsgIdFut = newFuture[string]()
```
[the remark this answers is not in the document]
vladopajic: i think we can ignore this in tests because they are short lived futures

interop/gossipsub/src/runner.nim
```diff
@@ -0,0 +1,130 @@
...
+  data[0 ..< 8] = toBytesBE(msgIdU64)
+  try:
+    discard await runner.node.publish(publishTopicID, data)
+  except CancelledError as e:
+    raise e
+  except CatchableError as e:
+    warn "Publish failed", messageID = publishMessageID, error = e.msg
+
+proc executeSetTopicValidationDelay(
+    runner: ScriptRunner, validationTopicID: string, delaySeconds: float64
```
vladopajic: `delaySeconds: float64` ->
`delay: Duration`

vladopajic (approved):

tests/interop/test_gossipsub.nim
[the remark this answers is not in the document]
vladopajic: the reason why we add name to future is, just in case, if we want to see runtime information of ongoing futures to be able to see what futures are alive. we don't really need this for unit test.

interop/gossipsub/src/instructions.nim
rlve: This is following the `GossipSub` interop tests contract:

https://github.com/libp2p/test-plans/blob/fbeebf2120cbc0d07dc922a6b810d81db6e0cb81/gossipsub-interop/script_instruction.py#L21

interop/gossipsub/peer.nim
```diff
@@ -0,0 +1,99 @@
...
+        if p.val.len > 0:
+          paramsPath = p.val
+        else:
+          expectParams = true
+      elif p.key == "node-id":
+        nodeIdOpt = parseInt(p.val)
+    of cmdShortOption:
+      discard
+
+  doAssert paramsPath.len > 0, "Must provide --params <path>"
```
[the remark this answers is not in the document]
vladopajic: i would prefer more quit then doAssert

interop/gossipsub/src/runner.nim
rlve: This is following the `GossipSub` interop tests contract:

https://github.com/libp2p/test-plans/blob/fbeebf2120cbc0d07dc922a6b810d81db6e0cb81/gossipsub-interop/script_instruction.py#L59

tests/interop/test_gossipsub.nim
rlve: Will split in the next PR, when there is more tests.

tests/interop/test_gossipsub.nim
rlve: I'm not sure it makes sense to spend time on it now, as the important thing is to make it working with `Gossipsub` Interop tests and there's still work to do.

interop/gossipsub/src/instructions.nim
vladopajic: i understand, but we can have json `elapsedSeconds` - watherver contract is, but in our nim code we can use type that fits the best for us.

for example we use `GossipSubParams` type - json contract is different then our type. but we use our type.

interop/gossipsub/src/runner.nim
rlve: changed in https://github.com/vacp2p/nim-libp2p/pull/2240/changes/57d1fc424edee91cb2534fdd332a5088a43d8602

interop/gossipsub/src/instructions.nim
rlve: changed in https://github.com/vacp2p/nim-libp2p/pull/2240/changes/57d1fc424edee91cb2534fdd332a5088a43d8602

interop/gossipsub/peer.nim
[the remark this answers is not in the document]
rlve: changed in https://github.com/vacp2p/nim-libp2p/pull/2240/changes/57d1fc424edee91cb2534fdd332a5088a43d8602

interop/gossipsub/src/instructions.nim
```diff
@@ -102,7 +102,7 @@ proc parseIfNodeIDEquals(
   )
 
 proc parseWaitUntil(j: JsonNode): ScriptInstruction =
-  ScriptInstruction(kind: WaitUntil, elapsedSeconds: j["elapsedSeconds"].getInt())
+  ScriptInstruction(kind: WaitUntil, elapsed: j["elapsedSeconds"].getInt().seconds())
```
vladopajic: what if `elapsedSeconds` is something like `1.35` ?

i think we should better have one proc that parses json duration to Duration type. then use it here `j["elapsedSeconds"].parseDuration()` and in code below:
` j["delaySeconds"].parseDuration()`

interop/gossipsub/src/instructions.nim
```diff
@@ -44,7 +44,7 @@ proc nsToDuration(ns: float64): Duration =
 proc getDuration(node: JsonNode, default: Duration): Duration =
   if node == nil or node.kind == JNull:
     return default
-  nsToDuration(node.getFloat())
+  node.getFloat().nsToDuration()
```
vladopajic: not sure if this is used?

interop/gossipsub/src/instructions.nim
rlve: `elapsedSeconds` is an `int` according to the contract, so I only expect full seconds

image

interop/gossipsub/src/instructions.nim
rlve: it's used when parsing `GossipSub` duration params

image

rlve: they come as nanoseconds according to the contract

interop/gossipsub/src/instructions.nim
vladopajic: got it, i think we can still have the same method for parsing `Duration` type:
- because if we get float from JsonNode we will be able to get `int` and `float` value with that, i assume.
- if we cant use getFloat then `parseDuration` logic could be:
  - try to get value using getFloat() -> if error try to get value using getInt() -> if error invalid input

interop/gossipsub/src/instructions.nim
vladopajic: aha, could we have then method:

```
parseDuration(node  JsonNode, unit: Duration): Duration= 
   return node.getFloat() * unit
```

then use it like this:
- in `GossipSubParams`

    ```
    ... =  j.getOrDefault("FanoutTTL").parseDuration(1.nanosecond)
    ```
- for `elapsedSeconds` and `delaySeconds`

  ```
  ... = j["delaySeconds"].parseDuration(1.seconds)
  ```

note:
- there might be better type for `unit: Duration` -> `unit: Unit` ( i haven't checked that in chronos)
- otherwise we could have `parseDurationSec` and `parseDurationNano` to avoid units arugment
- i avoided default values for simplifying example

vladopajic: `parseDurationSec` should work with int and float value in JsonNode (because there is not reason it shouldn't work with both - both type of  values could be converted to Duration type), similarly `parseDurationNano`. then we have really clean and robust parsing.

interop/gossipsub/src/node.nim
```diff
@@ -0,0 +1,81 @@
...
+  ]
+import ../../../tests/tools/crypto
+
+# Peer Id
+
+proc getNodeId*(): int =
+  let hostname = getHostname()
+  doAssert hostname.startsWith("node"),
+    "Expected hostname like 'node42', got: " & hostname
+  parseInt(hostname[4 ..^ 1])
```
gmelodie, line 24: This could panic, no?
```suggestion
let suffix = hostname[4 .. ^1]

if suffix.len == 0:
  raise newException(ValueError,
    "Missing node id in hostname: " & hostname)

try:
  return parseInt(suffix)
except ValueError:
  raise newException(ValueError,
    "Invalid node id in hostname: " & hostname)
```

gmelodie (approved):

interop/gossipsub/src/instructions.nim
rlve: handled in https://github.com/vacp2p/nim-libp2p/pull/2240/changes/37869a17c15a103c0607c7619138ea35f8c2b478

interop/gossipsub/src/instructions.nim
rlve: handled in https://github.com/vacp2p/nim-libp2p/pull/2240/changes/37869a17c15a103c0607c7619138ea35f8c2b478

interop/gossipsub/src/node.nim
rlve: Claudio says no

> suffix.len == 0 check is unnecessary — if hostname is "node" with nothing after it, parseInt("") already raises ValueError, so this guard adds nothing.
> 
> The try/except just re-raises the same exception type — catching ValueError from parseInt only to raise a new ValueError with a slightly different message is noise. The original parseInt error message already says what went wrong.

interop/gossipsub/src/instructions.nim
```diff
@@ -0,0 +1,174 @@
...
+
+proc parseDurationNano*(node: JsonNode): Duration =
+  ## Parse a JSON number (int or float) as nanoseconds into a Duration.
+  int64(node.getFloat()).nanoseconds()
+
+proc parseDurationSec*(node: JsonNode): Duration =
+  ## Parse a JSON number (int or float) as seconds into a Duration.
+  let secs = node.getFloat()
+  int64(secs * 1_000_000_000.0).nanoseconds()
+
```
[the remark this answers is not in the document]
rlve: off-by-1ns fine by me

PR #2241, opened 2026-04-07
feat(gossipsub): extend message priorities to H/M/L

richard-ramos:
## Summary
This PR replaces the current H/L priority outbound msg queue with a 3 level: High, Medium, and Low. 
Protocol-critical control traffic stays High, locally published messages move to Medium, and relayed / IWANT-reply traffic moves to Low.

This is needed to avoid low priority messages from grow without bound.. With this PR, if max high priority messages are exceeded, gossip will still disconnect slow peers, while medium/low traffic will just drop the messages.

## Affected Areas
- [x] Gossipsub  
- [ ] Transports  
- [] Protocol Logic
- [ ] Build / Tooling
- [ ] Other  

## Risk Assessment
The main risk is the change of behavior for low/medium priority messages.

## References
- https://roadmap.vac.dev/p2p/ift/2026q2-nimlibp2p-gossipsub-queues

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -101,11 +101,33 @@ proc init*(
     maxHighPriorityQueueLen = DefaultMaxHighPriorityQueueLen,
     maxMediumPriorityQueueLen = DefaultMaxMediumPriorityQueueLen,
     maxLowPriorityQueueLen = DefaultMaxLowPriorityQueueLen,
+    maxNumElementsInNonPriorityQueue: int = -1,
+      # Deprecated: use maxMediumPriorityQueueLen and maxLowPriorityQueueLen instead.
     sendIDontWantOnPublish = false,
     testExtensionConfig = Opt.none(TestExtensionConfig),
     partialMessageExtensionConfig = Opt.none(PartialMessageExtensionConfig),
     pingpongExtensionConfig = Opt.none(PingPongExtensionConfig),
 ): GossipSubParams =
+  if maxNumElementsInNonPriorityQueue >= 0:
+    warn "maxNumElementsInNonPriorityQueue is deprecated. Use maxMediumPriorityQueueLen and maxLowPriorityQueueLen"
+
+  let resolvedMaxMediumPriorityQueueLen =
+    if maxNumElementsInNonPriorityQueue >= 0 and
+        maxMediumPriorityQueueLen == DefaultMaxMediumPriorityQueueLen:
+      maxNumElementsInNonPriorityQueue
+    else:
+      maxMediumPriorityQueueLen
+
+  let resolvedMaxLowPriorityQueueLen =
+    if maxNumElementsInNonPriorityQueue >= 0 and
+        maxLowPriorityQueueLen == DefaultMaxLowPriorityQueueLen:
+      maxNumElementsInNonPriorityQueue
+    else:
+      maxLowPriorityQueueLen
+
+  let deprecatedMaxNumElementsInNonPriorityQueue =
+    if maxNumElementsInNonPriorityQueue >= 0: maxNumElementsInNonPriorityQueue else: 0
+
   GossipSubParams(
     explicit: true,
     pruneBackoff: pruneBackoff,
@@ -142,8 +164,9 @@ proc init*(
     overheadRateLimit: overheadRateLimit,
     disconnectPeerAboveRateLimit: disconnectPeerAboveRateLimit,
     maxHighPriorityQueueLen: maxHighPriorityQueueLen,
-    maxMediumPriorityQueueLen: maxMediumPriorityQueueLen,
-    maxLowPriorityQueueLen: maxLowPriorityQueueLen,
+    maxMediumPriorityQueueLen: resolvedMaxMediumPriorityQueueLen,
+    maxLowPriorityQueueLen: resolvedMaxLowPriorityQueueLen,
+    maxNumElementsInNonPriorityQueue: deprecatedMaxNumElementsInNonPriorityQueue,
     sendIDontWantOnPublish: sendIDontWantOnPublish,
     testExtensionConfig: testExtensionConfig,
     partialMessageExtensionConfig: partialMessageExtensionConfig,
@@ -1076,11 +1099,18 @@ method initPubSub*(g: GossipSub) {.raises: [InitializationError].} =
     g.parameters = GossipSubParams.init()
 
   # If deprecated maxNumElementsInNonPriorityQueue was set
-  # and the new field is still at its default, copy the value over.
-  if g.parameters.maxNumElementsInNonPriorityQueue > 0 and
-      g.parameters.maxLowPriorityQueueLen == DefaultMaxLowPriorityQueueLen:
-    warn "maxNumElementsInNonPriorityQueue is deprecated. Use maxLowPriorityQueueLen"
-    g.parameters.maxLowPriorityQueueLen = g.parameters.maxNumElementsInNonPriorityQueue
+  # and the new fields are still at their defaults, copy the value over.
+  if g.parameters.maxNumElementsInNonPriorityQueue > 0 and (
+    g.parameters.maxMediumPriorityQueueLen == DefaultMaxMediumPriorityQueueLen or
+    g.parameters.maxLowPriorityQueueLen == DefaultMaxLowPriorityQueueLen
+  ):
+    warn "maxNumElementsInNonPriorityQueue is deprecated. Use maxMediumPriorityQueueLen and maxLowPriorityQueueLen"
+    if g.parameters.maxMediumPriorityQueueLen == DefaultMaxMediumPriorityQueueLen:
+      g.parameters.maxMediumPriorityQueueLen =
+        g.parameters.maxNumElementsInNonPriorityQueue
+    if g.parameters.maxLowPriorityQueueLen == DefaultMaxLowPriorityQueueLen:
+      g.parameters.maxLowPriorityQueueLen =
+        g.parameters.maxNumElementsInNonPriorityQueue
 
   let validationRes = g.parameters.validateParameters()
   if validationRes.isErr:
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -1083,7 +1083,7 @@ method initPubSub*(g: GossipSub) {.raises: [InitializationError].} =
   # and the new field is still at its default, copy the value over.
   if g.parameters.maxNumElementsInNonPriorityQueue > 0 and
       g.parameters.maxLowPriorityQueueLen == DefaultMaxLowPriorityQueueLen:
-    warn "maxNumElementsInNonPriorityQueue is deprecated. Use maxNumElementsInNonPriorityQueue"
+    warn "maxNumElementsInNonPriorityQueue is deprecated. Use maxLowPriorityQueueLen"
     g.parameters.maxLowPriorityQueueLen = g.parameters.maxNumElementsInNonPriorityQueue
 
   let validationRes = g.parameters.validateParameters()
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -409,7 +409,7 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
         libp2p_pubsub_broadcast_prune.inc(labelValues = [g.topicLabel(prune.topicID)])
 
     trace "sending control message", payload = shortLog(respControl), peer
-    g.send(peer, RPCMsg.withControl(respControl), isHighPriority = true)
+    g.send(peer, RPCMsg.withControl(respControl), priority = MessagePriority.High)
 
   if messages.len > 0:
     for i, smsg in messages:
@@ -423,7 +423,7 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
 
     # iwant replies have lower priority
     trace "sending iwant reply messages", peer
-    g.send(peer, RPCMsg.withMessages(messages), isHighPriority = false)
+    g.send(peer, RPCMsg.withMessages(messages), priority = MessagePriority.Low)
 
 proc sendIDontWant(
     g: GossipSub,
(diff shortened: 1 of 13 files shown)


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -960,7 +968,11 @@ proc createExtensionsState(g: GossipSub): ExtensionsState =
     if cfg.onNegotiated.isNil:
       cfg.onNegotiated = proc(peerId: PeerId) {.gcsafe, raises: [].} =
         g.peers.withValue(peerId, peer):
-          g.send(peer[], RPCMsg(testExtension: Opt.some(TestExtensionRPC())), false)
+          g.send(
+            peer[],
+            RPCMsg(testExtension: Opt.some(TestExtensionRPC())),
+            priority = MessagePriority.Low,
```
richard-ramos: Should this be high priority instead? in the OG code it was low priority, but it seems to me that the priority should be higher, otherwise we could risk dropping the test extension message

tests/libp2p/pubsub/component/test_gossipsub_message_handling.nim
```diff
@@ -90,7 +90,7 @@ suite "GossipSub Component - Message Handling":
     gossip1.broadcast(
       gossip1.mesh[topic],
       RPCMsg.withControl(ControlMessage.withIHave(topic, iwantMessageIds)),
-      isHighPriority = false,
+      priority = MessagePriority.Low,
     )
```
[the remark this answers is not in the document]
richard-ramos: This uses the same priority it had from before. It's probably a good idea to update the tests to match how the priority those messages should really have

tests/libp2p/pubsub/component/test_gossipsub_message_cache.nim
```diff
@@ -264,7 +264,7 @@ suite "GossipSub Component - Message Cache":
     nodes[2].broadcast(
       @[nodes[2].getPeerByPeerId(topic, nodes[0].peerInfo.peerId)],
       RPCMsg.withControl(ControlMessage.withIWant(@[messageId1, messageId2])),
-      isHighPriority = false,
+      priority = MessagePriority.Low,
     )
```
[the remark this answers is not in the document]
richard-ramos: This uses the same priority it had from before. It's probably a good idea to update the tests to match how the priority those messages should really have

tests/libp2p/pubsub/component/test_gossipsub_control_messages.nim
```diff
@@ -298,7 +314,9 @@ suite "GossipSub Component - Control Messages":
 
     # When an IHAVE message is sent from node0
     let p1 = n0.getOrCreatePeer(n1.peerInfo.peerId, @[GossipSubCodec_12])
-    n0.broadcast(@[p1], RPCMsg.withControl(ihaveMessage), isHighPriority = false)
+    n0.broadcast(
+      @[p1], RPCMsg.withControl(ihaveMessage), priority = MessagePriority.Low
+    )
```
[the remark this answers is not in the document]
richard-ramos: This uses the same priority it had from before. It's probably a good idea to update the tests to match how the priority those messages should really have

tests/libp2p/pubsub/component/test_gossipsub_control_messages.nim
```diff
@@ -61,8 +61,12 @@ suite "GossipSub Component - Control Messages":
...
     let p0 = n1.getOrCreatePeer(n0.peerInfo.peerId, @[GossipSubCodec_12])
     let p1 = n0.getOrCreatePeer(n1.peerInfo.peerId, @[GossipSubCodec_12])
-    n0.broadcast(@[p1], RPCMsg.withControl(graftMessage), isHighPriority = false)
-    n1.broadcast(@[p0], RPCMsg.withControl(graftMessage), isHighPriority = false)
+    n0.broadcast(
+      @[p1], RPCMsg.withControl(graftMessage), priority = MessagePriority.Low
+    )
+    n1.broadcast(
+      @[p0], RPCMsg.withControl(graftMessage), priority = MessagePriority.Low
+    )
```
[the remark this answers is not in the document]
richard-ramos: This uses the same priority it had from before. It's probably a good idea to update the tests to match how the priority those messages should really have

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -548,10 +612,14 @@ proc canAskIWant*(p: PubSubPeer, msgId: MessageId): bool =
...
   return false
 
-proc sendNonPriorityTask(p: PubSubPeer) {.async: (raises: [CancelledError]).} =
+proc sendNonHighPriorityTask(p: PubSubPeer) {.async: (raises: [CancelledError]).} =
   while true:
     # we send non-priority messages only if there are no pending priority messages
-    let msg = await p.rpcmessagequeue.nonPriorityQueue.popFirst()
+    while p.rpcmessagequeue.mediumPriorityQueue.len == 0 and
+        p.rpcmessagequeue.lowPriorityQueue.len == 0:
+      p.rpcmessagequeue.dataAvailableEvent.clear()
```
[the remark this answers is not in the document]
richard-ramos: I dont think this can happen. Checking the length and the clear are sync operations that will be executed one after the other, and since we run on a single thread it shouldnt be possible for another signal to arrive between the len check and the clear call

libp2p/protocols/pubsub/gossipsub.nim
vladopajic: this should be priority of whatever that is not risk of dropping the message. 
but note that this will be only used in testing extensions interop and here there wont be many messages transmitted. so priority = whatever will also work

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -409,7 +417,7 @@ proc handleControl(g: GossipSub, peer: PubSubPeer, control: ControlMessage) =
         libp2p_pubsub_broadcast_prune.inc(labelValues = [g.topicLabel(prune.topicID)])
 
     trace "sending control message", payload = shortLog(respControl), peer
-    g.send(peer, RPCMsg.withControl(respControl), isHighPriority = true)
+    g.send(peer, RPCMsg.withControl(respControl), priority = MessagePriority.High)
```
vladopajic: in general, i think code does not need to be `... priority = value)` because there is only one property that can can match `MessagePriority`. doing `priority = MessagePriority.High` seems verbose/unnecessary.

before, i think, `isHighPriority = value` was used because there were two boolean arguments - so naming was necessary.

richard-ramos: Good idea

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -374,7 +411,7 @@ proc sendMsgSlow(p: PubSubPeer, msg: seq[byte]) {.async: (raises: [CancelledErro
 
 proc sendMsg(
```
vladopajic, line 408: change in this proc now adds a backpressure when message is sent. and idea was to do not have any - not ideally documented.

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -399,69 +436,100 @@ proc sendMsg(
...
-        libp2p_pubsub_disconnects_over_non_priority_queue_limit.inc()
-        p.closeSendConn(PubSubPeerEventKind.DisconnectionRequested)
+        libp2p_gossipsub_high_priority_queue_size.inc(labelValues = [$p.peerId])
+      # Check high-priority queue bound — if exceeded, the peer is too slow
+      # to keep up with protocol-critical messages and must be disconnected.
+      if p.rpcmessagequeue.sendPriorityQueue.len > p.maxHighPriorityQueueLen:
+        if not p.disconnected:
+          p.disconnected = true
+          libp2p_pubsub_disconnects_over_high_priority_queue_limit.inc()
+          p.closeSendConn(PubSubPeerEventKind.DisconnectionRequested)
```
vladopajic: i wounder if this case can be `p.disconnect()` we don't really need to know implementation details how peer is disconnecting.

```

proc disconnect(p: PubSubPeer) = 
 if not p.disconneted
   ...
```

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -399,69 +436,100 @@ proc sendMsg(
...
+          newFutureCompleted[void]()
       else:
-        Future[void].completed()
+        f
     else:
-      let f = p.rpcmessagequeue.nonPriorityQueue.addLast(
+      f
+  elif priority == MessagePriority.Medium:
+    if p.rpcmessagequeue.mediumPriorityQueue.len >= p.maxMediumPriorityQueueLen:
+      libp2p_pubsub_medium_priority_queue_drops.inc()
```
vladopajic: ` when defined(pubsubpeer_queue_metrics):`

should this be also true for `libp2p_pubsub_medium_priority_queue_drops` these metrics?

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -399,69 +436,100 @@ proc sendMsg(
...
+      trace "low priority queue full, dropping message", p
+      newFutureCompleted[void]()
+    else:
+      p.rpcmessagequeue.lowPriorityQueue.addLast(
+        QueuedMessage(data: msg, useCustomConn: useCustomConn)
+      )
+      when defined(pubsubpeer_queue_metrics):
+        libp2p_gossipsub_low_priority_queue_size.inc(labelValues = [$p.peerId])
+      p.rpcmessagequeue.dataAvailableEvent.fire()
+      newFutureCompleted[void]()
```
vladopajic: consider reducing complexity of  `sendEncoded` using proc for each case:

```
if highPriority:
  p.sendHighPriorityMessage(QueuedMessage)
elif mediumPriority:
  p.sendMediumPriorityMessage(QueuedMessage)
else:
 p.sendLowPriorityMessage(QueuedMessage)
```

sendXxxxMsg or queueXxxxMsg

libp2p/protocols/pubsub/pubsubpeer.nim
vladopajic: https://github.com/vacp2p/nim-libp2p/pull/1666

libp2p/protocols/pubsub/gossipsub.nim
vladopajic: i would go with high or medium, since there will be only 1 message ever sent of this kind - in testing environents

gmelodie (approved):

libp2p/protocols/pubsub/pubsubpeer.nim
gmelodie: Adding to this, it might be worth it to have a PriorityQueue type with an associated `send` method and a `maxLen` field inside (avoid having ugly fields like `maxMediumPriorityQueueLen`).

libp2p/protocols/pubsub/pubsubpeer.nim
richard-ramos: Good call. I fixed this to not add backpressure

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -410,58 +443,102 @@ proc sendMsg(
...
+  if not f.finished:
+    p.rpcmessagequeue.sendPriorityQueue.addLast(f)
+    when defined(pubsubpeer_queue_metrics):
+      libp2p_gossipsub_high_priority_queue_size.inc(labelValues = [$p.peerId])
+  return f
+
+proc sendMediumPriorityMessage(
+    p: PubSubPeer, msg: seq[byte], useCustomConn: bool
+): Future[void] =
+  let queuedMsg = QueuedMessage(data: msg, useCustomConn: useCustomConn)
```
vladopajic, line 470: consider reducing scope of `queuedMsg` to else branch

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -410,58 +443,102 @@ proc sendMsg(
...
+    when defined(pubsubpeer_queue_metrics):
+      libp2p_gossipsub_medium_priority_queue_size.inc(labelValues = [$p.peerId])
+    p.rpcmessagequeue.dataAvailableEvent.fire()
+
+  return newFutureCompleted[void]()
+
+proc sendLowPriorityMessage(
+    p: PubSubPeer, msg: seq[byte], useCustomConn: bool
+): Future[void] =
+  let queuedMsg = QueuedMessage(data: msg, useCustomConn: useCustomConn)
```
vladopajic, line 486: consider reducing scope of `queuedMsg` to else branch

tests/libp2p/pubsub/test_priority_queues.nim
```diff
@@ -0,0 +1,278 @@
...
+
+    await conn.close()
+    let drain1 = callbacksDrained(conn.pendingWrites[0])
+    let drain2 = callbacksDrained(conn.pendingWrites[1])
+    let drain3 = callbacksDrained(conn.pendingWrites[2])
+
+    checkUntilTimeout:
+      drain1.finished
+      drain2.finished
+      drain3.finished
```
vladopajic:
```suggestion
   await allFuturesRaising(@[drain1, drain2, drain3])
```

why not wait on futures themselves?
if accepted, please review other instances of same pattern. 

---

note, usage of `checkUntilTimeout` is a smell and it's existence is because of technical depth. we should try to avoid it's usage as much as possible, because it introduces unnecessary time waste (it always finishes after the condition), it can be source of flakiness (mostly in theory, but still..), it's a bit harder to debug then other synchronization construct like future, it is less expressive and in can never be as elegant as waiting on _right_ condition. that said, it's usage is completely fine, at least in `pubusub/component` tests.

vladopajic (approved):

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -220,7 +220,7 @@ proc send*(
     p: PubSub,
     peer: PubSubPeer,
     msg: RPCMsg,
```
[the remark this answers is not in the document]
vladopajic: this is interesting suggestion...
i general we could try approach to keep api backwards comparable, for 1-2 releases, before dropping it. we could do this only if it is effortless for us, because otherwise we may introduce risk.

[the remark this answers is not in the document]
vladopajic: but not sure, change should also be trivial to do...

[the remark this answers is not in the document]
richard-ramos: Yeah, this is a good suggestion and maintenance of it is trivial. I'll include it

PR #2238, opened 2026-04-06
feat(service-disco): add kademlia extensions

gmelodie:
## Summary

Extends the Kademlia DHT stack with the foundational types, routing infrastructure, and protocol plumbing needed for the Logos Capability / Service Discovery protocol. The work introduces a per-service routing table manager, a registrar cache with lower-bound enforcement, an advertiser, and the supporting metrics layer, while also refactoring `KademliaDiscovery` construction, hardening the routing table API (removing spurious `var` requirements), and adding bias-free random-selection utilities to the crypto module.

## Changes

- **`libp2p/protocols/service_discovery/types.nim` (new):**
  - Defines all service-discovery domain types: `ServiceId`, `Advertisement`, `Registrar`, `Advertiser`, `KademliaDiscoveryConfig`, `IpTree`, `PendingAction`, `AdvertiseTask`
  - Protocol constants (`Default_K_register`, `Default_K_lookup`, `Default_F_lookup`, etc.)
  - Helpers: `hashServiceId`, `advertisesService`, `toAdvertisementKey`
  - `Ticket` sign/verify via `PrivateKey` / `PublicKey`

- **`libp2p/protocols/service_discovery/service_routing_tables.nim` (new):**
  - `ServiceRoutingTableManager`: per-service `RoutingTable` instances with Interest / Provided / Both tracking
  - API: `addService`, `removeService`, `insertPeer`, `getTable`, `refreshAllTables`, `count`, `serviceIds`, `clear`
  - Prometheus metrics updates on every mutation

- **`libp2p/protocols/service_discovery/service_discovery_metrics.nim` (new):**
  - Declares public Prometheus counters, gauges, and histograms for:
    - Messages, registrations, lookups
    - Registrar cache
    - Service routing tables
    - IP tree
    - Advertiser

- **`libp2p/protocols/kademlia_discovery/types.nim`:**
  - `KademliaDiscovery` gains:
    - `registrar`, `advertiser`, `serviceRoutingTables`, `discoConf`, `xprPublishing`
    - Three async loop handles
  - Extracts `record()` from `refreshSelfSignedPeerRecord` to produce `SignedExtendedPeerRecord`
  - `ExtEntryValidator` / `ExtEntrySelector` updated to decode `SignedExtendedPeerRecord`
  - Removes `toKey(ServiceInfo)` and inline `ExtendedPeerRecord.init`

- **`libp2p/protocols/kad_disco.nim`:**
  - Moves `KademliaDiscovery.new` here (from types)
[...]

PR #2245, opened 2026-04-09
chore(crypto): add pick utilities

gmelodie:
## Summary

Adds `pick` and `pickOne` generic utilities to `libp2p/crypto/crypto.nim`, backed by `HmacDrbgContext` (CSPRNG) with rejection sampling to eliminate modulo bias. These are used to introduce `randomPeer` on the Kademlia routing table, enabling unbiased random peer selection from a bucket.

Also refactors `randomKeyInBucket` to take `ref HmacDrbgContext` instead of `var`, simplifies its bit-randomization logic (boundary byte + bulk slice instead of per-bit iteration), renames `routingtable.nim` → `routing_table.nim` and `randomfind.nim` → `random_find.nim` for snake_case consistency, and removes spurious `var` from `insert` and `peerIndexInBucket` procs (ref semantics made `var` redundant).

## Affected Areas

- [ ] Gossipsub  
- [ ] Transports  
  `routing_table.nim`: new `randomPeer` proc; `randomKeyInBucket` signature change and logic simplification.  
- [x] Protocol Logic  
  `kademlia.nim`, `find.nim`, `random_find.nim`: updated imports after file renames.  
- [ ] Build / Tooling  
- [x] Other  
  `crypto/crypto.nim`: new `pick`, `pickOne`, and `randBelow` procs.

Internal-only changes. No public API additions visible to downstream consumers beyond the new `pick` / `pickOne` / `randomPeer` procs. File renames affect only internal imports — all call-sites updated in this PR.

- **Nimbus:** N/A — no behavioral changes  
- **Waku:** N/A  
- **Codex:** N/A  

## Impact on Library Users

- New API additions: `pick`, `pickOne`, `randomPeer`
- No removals or breaking changes
- Internal module renames (`routingtable` → `routing_table`, `randomfind` → `random_find`) may affect direct imports, but these are not part of the public API surface

## Risk Assessment

Low.

- `randBelow` uses standard rejection sampling (no modulo bias)
- `randomKeyInBucket` preserves semantics (random key within XOR-distance bucket)
- Covered by existing and new unit tests

## References

Derived from #2238 

## Additional Notes

N/A

### DIFF
--- a/libp2p/crypto/crypto.nim
+++ b/libp2p/crypto/crypto.nim
@@ -180,6 +180,45 @@ proc shuffle*[T](rng: ref HmacDrbgContext, x: var openArray[T]) =
       y = rand mod i
     swap(x[i], x[y])
 
+proc randBelow(rng: ref HmacDrbgContext, max: uint32): int =
+  ## Returns a uniformly random integer in the range [0, max).
+  ## Uses 32-bit rejection sampling to eliminate modulo bias and to
+  ## support max values larger than 65536.
+  # threshold = 2^32 mod max. In uint32 arithmetic: (0 - umax) mod umax.
+  # Values in [0, threshold) are rejected to eliminate modulo bias.
+  let threshold = (0'u32 - max) mod max
+  while true:
+    var bytes: array[4, byte]
+    hmacDrbgGenerate(rng[], bytes)
+    let r =
+      bytes[0].uint32 or (bytes[1].uint32 shl 8) or (bytes[2].uint32 shl 16) or
+      (bytes[3].uint32 shl 24)
+    if r >= threshold:
+      return (r mod max).int
+
+proc pick*[T](rng: ref HmacDrbgContext, x: openArray[T], n: int): Opt[seq[T]] =
+  doAssert n >= 0, "n must be non-negative"
+  if x.len == 0:
+    return Opt.none(seq[T])
+  if n == 0:
+    return Opt.some(newSeq[T]())
+  var indices = newSeq[int](x.len)
+  for i in 0 ..< x.len:
+    indices[i] = i
+  let count = min(n, x.len)
+  var output = newSeq[T](count)
+  for i in 0 ..< count:
+    let j = i + rng.randBelow((x.len - i).uint32)
+    swap(indices[i], indices[j])
+    output[i] = x[indices[i]]
+  Opt.some(output)
+
+proc pickOne*[T](rng: ref HmacDrbgContext, x: openArray[T]): Opt[T] =
+  if x.len == 0:
+    return Opt.none(T)
+
+  Opt.some(x[rng.randBelow(x.len.uint32)])
+
 proc random*(
     T: typedesc[PrivateKey],
     scheme: PKScheme,
--- a/libp2p/crypto/crypto.nim
+++ b/libp2p/crypto/crypto.nim
@@ -202,9 +202,11 @@ proc pick*[T](rng: ref HmacDrbgContext, x: openArray[T], n: int): Opt[seq[T]] =
     return Opt.none(seq[T])
   if n == 0:
     return Opt.some(newSeq[T]())
+
   var indices = newSeq[int](x.len)
   for i in 0 ..< x.len:
     indices[i] = i
+
   let count = min(n, x.len)
   var output = newSeq[T](count)
   for i in 0 ..< count:
--- a/libp2p/protocols/kademlia/routing_table.nim
+++ b/libp2p/protocols/kademlia/routing_table.nim
@@ -6,6 +6,7 @@ import bearssl/rand, chronos, chronicles, results
 import ./types
 import ./kademlia_metrics
 import ../../peerid
+import ../../crypto/crypto
 import ../../utils/sequninit
 
 logScope:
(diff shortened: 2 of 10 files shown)


libp2p/crypto/crypto.nim
```diff
@@ -180,6 +180,46 @@ proc shuffle*[T](rng: ref HmacDrbgContext, x: var openArray[T]) =
...
+      return (r mod umax).int
+
+proc pick*[T](rng: ref HmacDrbgContext, x: openArray[T], n: int): Opt[seq[T]] =
+  if x.len == 0:
+    return Opt.none(seq[T])
+  if n == 0:
+    return Opt.some(newSeq[T]())
+  var indices = newSeq[int](x.len)
+  for i in 0 ..< x.len:
+    indices[i] = i
```
vladopajic, line 208:
```suggestion

  var indices = newSeq[int](x.len)
  for i in 0 ..< x.len:
    indices[i] = i

```

libp2p/crypto/crypto.nim
```diff
@@ -180,6 +180,46 @@ proc shuffle*[T](rng: ref HmacDrbgContext, x: var openArray[T]) =
       y = rand mod i
     swap(x[i], x[y])
 
+proc randBelow(rng: ref HmacDrbgContext, max: int): int =
```
vladopajic: use unit32 for max
- doAssert is not needed
- conversion below is not needed

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -156,22 +157,21 @@ proc randomKeyInBucket*(selfId: Key, bucketIndex: int, rng: var HmacDrbgContext)
...
-      raw[byteIdx] = raw[byteIdx] or (1'u8 shl bitInByte)
-    else:
-      raw[byteIdx] = raw[byteIdx] and not (1'u8 shl bitInByte)
+  # randomize remaining bytes
+  if tgtByte + 1 < raw.len:
+    hmacDrbgGenerate(rng[], raw.toOpenArray(tgtByte + 1, raw.len - 1))
 
   return raw
+
+proc randomPeer*(bucket: Bucket, rng: ref HmacDrbgContext): Opt[Key] =
```
vladopajic, line 173: it doesn't make sense for `randomPeer()` to return `Opt[Key]` - naming-wise something feels off. 

we can close eyes and ignore this, but if this is systemic issue in this package, it should be approached form mindset of resolving it. rather then add and adding more same _bad_ decisions.

from previous prs consider:

```
bucket.randomPeer(rng).toKey()
```

vladopajic (approved):

libp2p/crypto/crypto.nim
gmelodie: what?

libp2p/protocols/kademlia/routing_table.nim
gmelodie: what happens if you call random peer on an empty bucket?

libp2p/crypto/crypto.nim
richard-ramos: @vladopajic are you making the code breathe? :eyes:

richard-ramos (approved):

libp2p/crypto/crypto.nim
vladopajic: yes :)

libp2p/protocols/kademlia/routing_table.nim
vladopajic: what happens now?

`randomPeer(b: Bucket): Opt[Peer]`
`toKey(peer: Opt[Peer]): Opt[Key]`

vladopajic: above utilities will stack

libp2p/crypto/crypto.nim
gmelodie: jesus christ

libp2p/protocols/kademlia/routing_table.nim
gmelodie: Is this what you want?  `randomPeer(b: Bucket): Peer` and have a `doAssert bucket.len > 0` inside it?

vladopajic: i just advise on reimplementing this with 2 mentioned utilities. 

```
randomPeer(b: Bucket): Opt[Peer]
toKey(peer: Opt[Peer]): Opt[Key]
```
these utilities return Opt, because if there are no peers random will return Opt.none.  
and toKey is handy utility that just return key, from optional value.

because when used in conjunction they can stack and they are logical from naming and usage perspective.

question: what will happen when code will need you to really get random peer from bucket - and not `Key`?
what is new method gonna be named `randomPeerReallyPeerThisTime(b: Bucket): Opt[Peer]`?

vladopajic: okay method overloading is a thing in nim, but still there will be 2 implementations of `randomPeer` that will do same thing.

vladopajic:
```
proc randomPeer*(bucket: Bucket, rng: ref HmacDrbgContext): Opt[Peer] =
  rng.pickOne(bucket.peers)

proc toKey(peer: Opt[Peer]): Opt[Key] =
  peer.withValue(p):
    Opt.some(p.nodeId)
  Opt.none(Key)
```

vladopajic: then further changes would be ->

move `randomPeer` utility to location where `Bucket` is defined, since all utilities should be in package were type they operate on are.
same with `toKey`

now utilities are in different place, but they will be always there were you need them. and same should always be done for all utilities.

vladopajic: that is recipe

PR #2246, opened 2026-04-09
chore(service-disco): add service discovery

gmelodie:
## Summary

Renames the `KademliaDiscovery` protocol to `ServiceDiscovery` and lays the groundwork for per-service routing tables. The rename reflects that the protocol’s purpose is general service discovery, not Kademlia-specific logic.

This PR also fixes two bugs that prevented `randomRecords` from returning results:
- `ExtEntryValidator` was decoding peer records as `SignedPeerRecord` instead of `SignedExtendedPeerRecord`, causing `putValue` to always reject records.
- `ServiceDiscovery.start()` was not launching the self-signed peer record maintenance loop, so nodes never stored their own records.

## Affected Areas

- [ ] Gossipsub  
- [ ] Transports  

  `ServiceDiscovery` (formerly `KademliaDiscovery`) and its routing table infrastructure.  
  `randomRecords` / `lookup` now function correctly end-to-end.

- [x] Protocol Logic  
  `routing_table.nim` (renamed from `routingtable.nim`):
  - `insert` and `peerIndexInBucket` no longer require `var` receiver  
  - `randomKeyInBucket` now takes `ref HmacDrbgContext`  
  - Adds `randomPeer`  
  - `service_discovery.nim` start/stop lifecycle fixed  

- [ ] Build / Tooling  

- [x] Other  
  `libp2p/crypto/crypto.nim`:
  - Adds `pick` and `pickOne` (bias-free random selection via `HmacDrbgContext`)  
  - Used by `randomPeer`  

- **Nimbus:** N/A  
- **Waku:** N/A  
- **Codex:** N/A  

## Impact on Library Users

- **Renames:**
  - `KademliaDiscovery` → `ServiceDiscovery`
  - `ExtendedKademliaDiscoveryCodec` → `ExtendedServiceDiscoveryCodec`
  - `kad_disco` → `service_discovery`
  - `kademlia_discovery/` → `service_discovery/`
  - `routingtable` → `routing_table`
- `cbind` FFI field `mountKadDiscovery` → `mountServiceDiscovery`

- **API changes:**
  - `randomKeyInBucket`: `var HmacDrbgContext` → `ref HmacDrbgContext`
  - `RoutingTable.insert` no longer requires `var`

## Risk Assessment

- **Network incompatibility:**  
  Codec changed from `/logos/kad/1.0.0` to `/logos/service-discovery/1.0.0`

- **Behavioral fixes:**  
  - `ExtEntryValidator` now correctly accepts extended peer records  
  - Self-signed peer record loop now runs  
  - `randomRecords` now returns results as expected  

- **Performance:**  
  No regressions expected; `pick` / `pickOne` use O(1) amortized rejection sampling

## References

#2238 and #2245 

## Additional Notes

`ServiceRoutingTableManager` (per-service routing tables keyed by `ServiceId`) and `ServiceDiscoveryConfig` are scaffolded but not yet wired into live code paths. They serve as placeholders for upcoming service routing work.


libp2p/protocols/service_discovery.nim
richard-ramos, line 1: With this rename, are we getting rid of KademliaDiscovery?

I might have understood wrong but shouldnt it be like this?:
- KadDHT: Base protocol
- KademliaDiscovery: Extension to Kademlia (fully compatible with KadDHT)
- Service Discovery: Different protocol

cc: @SionoiS for clarification

cbind/ffi_types.nim
```diff
@@ -126,7 +126,7 @@ type Libp2pConfig* {.bycopy.} = object
   gossipsubTriggerSelf*: cint
   mountKad*: cint
   mountMix*: cint
-  mountKadDiscovery*: cint
+  mountServiceDiscovery*: cint
```
richard-ramos, line 129: This should be changed as well: https://github.com/vacp2p/nim-libp2p/blob/master/cbind/libp2p.h#L148 (and the cbindings example)

libp2p/protocols/service_discovery.nim
gmelodie: I decided to get rid of it as it was causing me some confusion (it required types from service discovery and vice versa) and it wasn't really being used by any of our clients, so I decided we drop it altogether until there's an actual need for it, and when that's the case we can properly abstract service discovery. Right now I think it'd be just over abstracting too soon and adding unnecessary complexity

SionoiS: > * KadDHT: Base protocol
> * KademliaDiscovery: Extension to Kademlia (fully compatible with KadDHT)
> * Service Discovery: Different protocol

Service discovery is technically just an extension of KadDHT (2 new msg types)

What I called at first `KademliaDiscovery` (the XPR publishing thing) is just a specific way of using KadDHT.

> I decided to get rid of it as it was causing me some confusion (it required types from service discovery and vice versa) and it wasn't really being used by any of our clients, so I decided we drop it altogether until there's an actual need for it, and when that's the case we can properly abstract service discovery. Right now I think it'd be just over abstracting too soon and adding unnecessary complexity

Make sense to me since everything is 100% backward compatible with plain KadDHT (with some caveats).

richard-ramos (approved):

libp2p/protocols/service_discovery/types.nim
```diff
@@ -13,11 +13,29 @@ import ../kademlia/types
...
+  ServiceStatus* = enum
+    Interest = 0
+    Provided = 1
+    Both = 2
+
+  ServiceRoutingTableManager* = ref object
+    tables*: Table[ServiceId, RoutingTable]
+    serviceStatus*: Table[ServiceId, ServiceStatus]
+
+  ServiceDiscoveryConfig* = object ## placeholder for now
```
vladopajic: seems unrelated

vladopajic (approved):

PR #2247, opened 2026-04-10
chore(connmanager): add muxer store

vladopajic:
## Summary

as a part of `ift-ts:p2p:ift:2026q1-nimlibp2p-connection-management` commitment this pr has preparation work.

- adding `MuxerStore` to connection manager that encapsulates logic with muxers.
  - this has enabled to avoid `muxCleanup`
- dropping `slotsAvailable`

## Affected Areas

- [X] Connection Manager

this pr only has under the hood changes.

## Impact on Library Users

this pr only has under the hood changes.

## Risk Assessment

should be fully verifiable with current unit tests - no contract change.

## References


libp2p/muxer_store.nim
```diff
@@ -0,0 +1,74 @@
+import std/[tables, sequtils]
+from peerinfo import PeerId
+from muxers/muxer import Muxer
+from stream/connection import Connection, Direction
```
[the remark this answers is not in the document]
vladopajic: muxer.connection.peerId

libp2p/connmanager.nim
```diff
@@ -483,7 +452,7 @@ proc dropPeer*(c: ConnManager, peerId: PeerId) {.async: (raises: [CancelledError
   ##
   trace "Dropping peer", peerId
 
-  let muxers = c.muxed.getOrDefault(peerId)
+  let muxers = c.muxerStore.remove(peerId)
```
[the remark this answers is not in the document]
vladopajic: @copilot please check if newest code has this issue. add short summary to this comment.

do not fix.
do not make any fixes.
do not make any changes.
do not push commits to pr.

libp2p/connmanager.nim
```diff
@@ -289,32 +272,13 @@ proc closeMuxer(muxer: Muxer) {.async: (raises: [CancelledError]).} =
       trace "Exception in close muxer handler", description = exc.msg
   trace "Cleaned up muxer", m = muxer
 
-proc muxCleanup(c: ConnManager, mux: Muxer) {.async: (raises: []).} =
```
vladopajic, line 291: existence of `muxCleanup` is hardly justified, because there is `onClose` with the same signature. 
there should be only 1 method that does the cleanup - `onClose`.

libp2p/connmanager.nim
```diff
@@ -430,7 +392,7 @@ proc getOutgoingSlot*(
 func semaphore(c: ConnManager, dir: Direction): AsyncSemaphore {.inline.} =
   return if dir == In: c.inSema else: c.outSema
 
-proc slotsAvailable*(c: ConnManager, dir: Direction): int =
```
vladopajic, line 432: dropping `slotsAvailable` in favor of `availableSlots` which use is already more widespread

libp2p/muxer_store.nim
```diff
@@ -0,0 +1,77 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Copyright (c) Status Research & Development GmbH
+
+import std/[tables, sequtils]
+from peerinfo import PeerId
+from muxers/muxer import Muxer
+from stream/connection import Connection, Direction
```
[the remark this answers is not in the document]
vladopajic: vai `muxer.connection.peerid`

libp2p/connmanager.nim
```diff
@@ -496,22 +459,18 @@ proc close*(c: ConnManager) {.async: (raises: [CancelledError]).} =
...
   trace "Closing ConnManager"
   c.closed = true
-  let muxed = c.muxed
-  c.muxed.clear()
 
   let expected = c.expectedConnectionsOverLimit
   c.expectedConnectionsOverLimit.clear()
 
   for _, fut in expected:
     await fut.cancelAndWait()
```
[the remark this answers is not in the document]
vladopajic: @copilot  please check if newest code has this issue. add short summary to this comment.

do not fix.
do not make any fixes.
do not make any changes.
do not push commits to pr.

libp2p/connmanager.nim
```diff
@@ -289,32 +272,13 @@ proc closeMuxer(muxer: Muxer) {.async: (raises: [CancelledError]).} =
...
-  except CatchableError as exc:
-    # This is top-level procedure which will work as separate task, so it
-    # do not need to propagate CancelledError and should handle other errors
-    warn "Unexpected exception peer cleanup handler", mux, description = exc.msg
+proc onPeerDisconnected(c: ConnManager, peerId: PeerId) {.async: (raises: []).} =
+  c.clearPeerReadyState(peerId)
+  libp2p_peers.set(c.muxerStore.countPeers.int64)
+  await noCancel c.triggerPeerEvents(peerId, PeerEvent(kind: PeerEventKind.Left))
+  await noCancel c.triggerConnEvent(peerId, ConnEvent(kind: ConnEventKind.Disconnected))
+  if not c.peerStore.isNil:
```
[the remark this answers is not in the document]
vladopajic: @copilot  please check if newest code has this issue. add short summary to this comment.

do not fix.
do not make any fixes.
do not make any changes.
do not push commits to pr.

libp2p/connmanager.nim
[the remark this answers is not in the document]
vladopajic: @copilot  please check if newest code has this issue. add short summary to this comment.

do not fix.
do not make any fixes.
do not make any changes.
do not push commits to pr.

libp2p/connmanager.nim
vladopajic: @copilot please provide assessment what approach is better - code that was before or after the change?

do not fix.
do not make any fixes.
do not make any changes.
do not push commits to pr.

libp2p/connmanager.nim
```diff
@@ -370,25 +335,23 @@ proc storeMuxer*(
...
-    doAssert muxer notin muxers[]
-    muxers[].add(muxer)
-  do:
-    c.muxed[peerId] = @[muxer]
-    newPeer = true
-  libp2p_peers.set(c.muxed.len.int64)
+  let isNewPeer = c.muxerStore.count(peerId) == 0
+
+  if not c.muxerStore.add(muxer):
+    raise newException(LPError, "muxer already stored")
```
richard-ramos, line 356: This changes how the previous existance of a muxer is handled, with previously being critical enough to stop the execution of nim-libp2p. 

What happens now when this exception is being raised?

libp2p/connmanager.nim
```diff
@@ -328,16 +291,18 @@ proc onClose(c: ConnManager, mux: Muxer) {.async: (raises: []).} =
...
+    let peerId = mux.connection.peerId
+    let removed = c.muxerStore.remove(mux)
+    if removed and c.muxerStore.count(peerId) == 0:
+      await c.onPeerDisconnected(peerId)
+    await noCancel c.triggerConnEvent(
+      peerId, ConnEvent(kind: ConnEventKind.Disconnected)
+    )
 
-proc selectMuxer*(c: ConnManager, peerId: PeerId, dir: Direction): Muxer =
+proc selectMuxer*(c: ConnManager, peerId: PeerId, dir: Direction): Muxer {.inline.} =
```
richard-ramos:
```suggestion
proc selectMuxer*(c: ConnManager, peerId: PeerId, dir: Direction): Muxer =
```
https://status-im.github.io/nim-style-guide/language.inline.html

copilot should have warned about it tho :( https://github.com/vacp2p/nim-libp2p/blob/master/.github/copilot-instructions.md#inline-functions

libp2p/connmanager.nim
```diff
@@ -483,9 +447,11 @@ proc dropPeer*(c: ConnManager, peerId: PeerId) {.async: (raises: [CancelledError
   ##
   trace "Dropping peer", peerId
 
-  let muxers = c.muxed.getOrDefault(peerId)
+  let muxers = c.muxerStore.remove(peerId)
```
richard-ramos, line 454: This might cause issues if there is a reconnection when closing:

It goes like this:
- In line 450 you remove the peer from mux store immedialtly
- `dropPeer` will then await the `close`s in line 452. During this wait a dropped peer could reconnect
- when this happen, a new muxer will be stored (Perhaps this is a case for locks to be used?)
- Since you cleaned the store,  the reconnect will be counted as a new peer
- And then when dropPeer finishes waiting, it will still call c.onPeerDisconnected for a peer that has reconnected, emiting the Left and removing peer from peerstore

libp2p/connmanager.nim
richard-ramos, line 449: Since this returns a future that can be cancelled, something that could happen is that you end up with the peer being removed from the mux store in line 450, but not really dropped if the cancellation happens while line 452 is being awaited

richard-ramos (commented):
Hm... reading this PR made me realize that both the code introduced in this PR, and mine should be tested https://github.com/vacp2p/nim-libp2p/pull/2241 for some time both by DST and also in one of the testnets in nimbus before we do the next release

libp2p/connmanager.nim
vladopajic: same muxer should not be tracked, aka stored, more then once. if happens it means that libp2p developer has made an error. 

since method already raises, it is convenient to raise error indicating wrong behavior.

libp2p/connmanager.nim
vladopajic: good catch i see now.

vladopajic: in new change, `dropPeer` has be made atomic. after all it should be.

libp2p/connmanager.nim
vladopajic: this should be the same issue as https://github.com/vacp2p/nim-libp2p/pull/2247#discussion_r3065475197

libp2p/connmanager.nim
vladopajic: first, case of cancellation was not ideal even in code previously:

- `dropPeer` _could_ drop connection without emitting events
- `dropPeer` will _never_ do as expected – remove all resources associated with peer (it will always partially remove muxers). when user wants to drop peer, manager should drop all knowledge related to it.
- because some peer connections will stay in memory, _always_ will in case of cancellation, then it means that gc will not be triggered for those muxers

with new change:

- it is guaranteed that `dropPeer` will remove all resources for peer – even when cancelled
- cancellation will only interrupt process of closing muxers
  - this can be easily changed to make this process uninterruptible, but i dont think we should or need to do it
  - why? when cancellation happens, manager will clear all resources (guaranteed), then only thing left is to close muxers. if this part is cancelled (closing muxers), it is not managers responsibility if graceful close was interrupted, but responsibility of caller – it is actually callers _fulfilled want_.
- removing all references to muxers encourages nim gc to do cleanup even when interrupted

libp2p/connmanager.nim
richard-ramos: I think the problem still happens in last commit with this sequence:
1. In line 450 the peer is removed from the mux store
2. In line 453 you wait for the muxers to close
3. In line 455 you always call onPeerDisconnected in the `finally`

In step 2 it is still possible during that wait a disconnected peer reconnects, and then in storeMuxer the peer is added back as a new connection. `onPeerDisconnected` would not know this, thus still execute for that peer.

Maybe adding the `noCancel` and a `if` is enough? i.e.:
```nim
let muxers = c.muxerStore.remove(peerId)
if muxers.len > 0:
  try:
    await noCancel allFutures(muxers.mapIt(closeMuxer(it)))  <--------------
  finally:
    if c.muxerStore.count(peerId) == 0:  <---------------------------
      await noCancel c.onPeerDisconnected(peerId)
```

I vibecoded this demo of how the bug would happen: https://gist.github.com/richard-ramos/ff00008baac198d99d3ab53fd8b5a279#file-drop_peer_race_demo-nim-L80

vladopajic: https://github.com/vacp2p/nim-libp2p/pull/2247/commits/3caade6080ea22ada10ba039919b443b410c9766 will solve it

thanks for example, i wouldn't really get to your point without it.
described issue is very unlikely to be reproduced in real nim-libp2p usages, perhaps impossible?, considering current context. reasoning for this is that example assumes:
 1) `dropPeer` is not waited on right away
     -  nim-libp2p: never does this. equivalent would be that it's future is discarded, what we should never do.
2) `storeMuxer` is called is called within same caller's stack frame as `dropPeer`
     - nim-libp2: never the case. `storeMuxer` and `dropPeer` never share the same caller's stack frame
3) uses mix of async and sync calls
    -  nim-libp2p: `storeMuxer` is async, example assumes that it is not async
4) example in general has _wrong_ logic

however, it makes sense to pay attention and write code to be resilient even for scenarios like these.  :+1:

libp2p/muxer_store.nim
```diff
@@ -0,0 +1,77 @@
...
+    if idx >= 0:
+      muxers[].del(idx)
+      if muxers[].len == 0:
+        s.muxed.del(peerId)
+      return true
+  return false
+
+proc add*(s: MuxerStore, muxer: Muxer): bool =
+  let peerId = muxer.connection.peerId
+  s.muxed.withValue(peerId, muxers):
```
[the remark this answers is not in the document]
vladopajic: muxer stork works with non nil values by design, all check are done in lvl above, conn manager

libp2p/connmanager.nim
```diff
@@ -265,18 +254,12 @@ proc expectConnection*(
     c.expectedConnectionsOverLimit.del(key)
 
 proc contains*(c: ConnManager, peerId: PeerId): bool =
-  peerId in c.muxed
+  return c.muxerStore.contains(peerId)
```
gmelodie, line 256: Maybe `{.inline.}` since it's one liner? Also idk if this is doable but it'd be nice to be able to do `peerId in c.muxerStore` if possible

libp2p/connmanager.nim
```diff
@@ -289,32 +272,17 @@ proc closeMuxer(muxer: Muxer) {.async: (raises: [CancelledError]).} =
...
+    return
 
-    await c.triggerConnEvent(peerId, ConnEvent(kind: ConnEventKind.Disconnected))
-  except CatchableError as exc:
-    # This is top-level procedure which will work as separate task, so it
-    # do not need to propagate CancelledError and should handle other errors
-    warn "Unexpected exception peer cleanup handler", mux, description = exc.msg
+  c.clearPeerReadyState(peerId)
+  if not c.peerStore.isNil:
+    c.peerStore.cleanup(peerId)
```
gmelodie, line 282: Can we make `cleanup` check if peerstore is nil?

libp2p/connmanager.nim
```diff
@@ -328,16 +296,18 @@ proc onClose(c: ConnManager, mux: Muxer) {.async: (raises: []).} =
     debug "Unexpected exception in connection manager's cleanup",
       description = exc.msg, mux
   finally:
-    await c.muxCleanup(mux)
+    let peerId = mux.connection.peerId
+    let removed = c.muxerStore.remove(mux)
+    if removed and c.muxerStore.count(peerId) == 0:
+      await c.onPeerDisconnected(peerId)
```
gmelodie, line 301: Do you think it makes sense to have these checks inside `onPeerDisconnected` so we could do just
```suggestion
    await c.onPeerDisconnected(peerId)
```

libp2p/connmanager.nim
```diff
@@ -370,25 +340,23 @@ proc storeMuxer*(
     peerId = muxer.connection.peerId
     dir = muxer.connection.dir
 
-  if c.connCount(peerId) > c.maxConnsPerPeer:
+  if c.muxerStore.count(peerId) > c.maxConnsPerPeer:
     let key = (peerId, dir)
     let expectedConn = c.expectedConnectionsOverLimit.getOrDefault(key)
     if expectedConn != nil and not expectedConn.finished:
       expectedConn.complete(muxer)
```
gmelodie, line 346: obs: i would expect this to be
```nim
muxer.complete(expectedConn)
```
no?

libp2p/connmanager.nim
```diff
@@ -370,25 +340,23 @@ proc storeMuxer*(
...
     else:
-      debug "Too many connections for peer", conns = c.connCount(peerId), peerId, dir
+      debug "Too many connections for peer",
+        conns = c.muxerStore.count(peerId), peerId, dir
 
       raise newTooManyConnectionsError()
```
gmelodie, line 351: I don't quite understand this part. If `if c.muxerStore.count(peerId) <= c.maxConnsPerPeer:` (which is the equivalent of the else clause), then we error with too many connections? Why isn't it the opposite?

libp2p/connmanager.nim
```diff
@@ -483,11 +452,14 @@ proc dropPeer*(c: ConnManager, peerId: PeerId) {.async: (raises: [CancelledError
...
 
-  let muxers = c.muxed.getOrDefault(peerId)
-  for muxer in muxers:
-    await closeMuxer(muxer)
+  let muxers = c.muxerStore.remove(peerId)
+  if muxers.len > 0:
+    try:
+      await allFutures(muxers.mapIt(closeMuxer(it)))
+    finally:
+      await noCancel c.onPeerDisconnected(peerId)
```
gmelodie, line 459: Is there a case where `closeMuxer` won't call `onPeerDisconnected`? If not, can we `closeMuxer` call `onPeerDisconnected`?

libp2p/connmanager.nim
```diff
@@ -510,8 +480,11 @@ proc close*(c: ConnManager) {.async: (raises: [CancelledError]).} =
...
       readyEvent.cancelSoon()
 
-  for _, muxers in muxed:
-    for mux in muxers:
-      await closeMuxer(mux)
+  let muxed = c.muxerStore.getAll()
+  c.muxerStore.clear()
+  for peerId, muxers in muxed:
+    if muxers.len > 0:
+      await allFutures(muxers.mapIt(closeMuxer(it)))
```
gmelodie, line 486: If any of `closeMuxer` gets cancelled, would this leave the store in a weird state? Would you still want to call `onPeerDisconnected` then?

gmelodie (approved):
cool!

libp2p/connmanager.nim
vladopajic: described this case here 
https://github.com/vacp2p/nim-libp2p/pull/2247#discussion_r3068069478

libp2p/connmanager.nim
vladopajic: https://github.com/vacp2p/nim-libp2p/pull/2242#discussion_r3051033386
:dagger:  :)

gmelodie: fuck 2 space indents

libp2p/connmanager.nim
vladopajic: https://github.com/vacp2p/nim-libp2p/pull/2247#discussion_r3065309455

tbh i have no problem of `{.inline.}` for all one liners.
but we need to be straight on that

libp2p/connmanager.nim
vladopajic: not really, ` c.muxerStore.remove(mux)` needs to be outside `onPeerDisconnected`

richard-ramos (approved):

PR #2262, opened 2026-04-13
fix(autotls): letsencrypt challenge type change

gmelodie:
## Summary

Let's Encrypt staging recently introduced a new challenge type (`dns-persist-01`) that is not present in the `ACMEChallengeType` enum.

The previous implementation deserialized the entire `challenges` array in one shot via `json.to(ACMEAuthorizationsResponse)`, so any unrecognized enum value caused `parseEnum` to raise a `ValueError`, crashing the daily integration jobs.

This PR fixes that by parsing each challenge individually and silently skipping entries with unrecognized fields, which is the correct behaviour per RFC 8555.

It also fixes a missing `libp2p_autotls_support` guard on the integration test file that prevented the test from compiling correctly.

### Changes

- **libp2p/autotls/acme/api.nim**
  - Changed `requestAuthorizations` to iterate over the `challenges` array element-by-element
  - Catch `ValueError` / `JsonKindError` per entry
  - Skip unrecognized challenge types with a debug log instead of propagating an error

- **tests/integration/test_autotls_integration.nim**
  - Added `and defined(libp2p_autotls_support)` to the `when` guard
  - Ensures the file only compiles when autotls support is enabled (consistent with other autotls tests)
  - Fixes visibility issue where `ACMEApi.new` was not available to the compiler

---

## Affected Areas

- Gossipsub
- Transports
- Peer Management / Discovery
- Protocol Logic
  - ACME challenge parsing in the AutoTLS subsystem is now resilient to unknown challenge types
- Build / Tooling
- Other

---

The change is backward compatible.

Public types remain unchanged:
- `ACMEChallengeType`
- `ACMEChallenge`
- `ACMEAuthorizationsResponse`

Behavioral change:
- `requestAuthorizations` now returns a filtered list of challenges (only recognized ones)
- Previously, it raised an error when encountering unknown types

Impact:
- Code filtering by specific challenge types (e.g. `DNS01`) is unaffected
- Nimbus, Waku, and Codex do not consume this internal ACME API directly
- No downstream validation required

---

## Impact on Library Users

- **`requestAuthorizations` — behavior change**
  - Previously: raised `ACMEError("Failed to decode JSON")` if any challenge had unknown type/status
  - Now: silently skips invalid challenges and returns the rest
  - Existing callers already filter by type, so effective behavior remains the same

- No new types, procs, or fields introduced

---

## Risk Assessment

- **Silent challenge skipping**
  - If Let's Encrypt replaces known challenge types with unknown ones, result may be empty
  - This leads to `ACMEError("No challenges received")`
  - Considered correct failure mode
  - **Risk: Low**

- **Per-element try/except masking real decode errors**
  - Malformed JSON in valid challenge types may be silently dropped
  - Mitigated via debug logging
  - **Risk: Low**

---

## References

* #2261


libp2p/autotls/acme/api.nim
```diff
@@ -348,7 +348,13 @@ when defined(libp2p_autotls_support):
     handleError("requestAuthorizations"):
       doAssert authorizations.len > 0
       let acmeResponse = await self.get(parseUri(authorizations[0]))
-      acmeResponse.body.to(ACMEAuthorizationsResponse)
+      var challenges: seq[ACMEChallenge]
+      for challenge in acmeResponse.body.getOrDefault("challenges").getElems():
+        try:
+          challenges.add(challenge.to(ACMEChallenge))
+        except ValueError, JsonKindError:
+          debug "Skipping challenge with unrecognized fields", challenge = $challenge
```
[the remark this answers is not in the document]
gmelodie: Same error message for both, like it is rn

richard-ramos, line 358: What should happen here id the number of challenges is 0 due to some of them failing in line 356. Should a ACMEError be raised indicating that it was not possible to request an autorization?

gmelodie: That's exactly what happens on line 374:
```nim
    let authorizationsResponse =
      await self.requestAuthorizations(orderResponse.authorizations, key, kid)
    if authorizationsResponse.challenges.len == 0:
      raise newException(ACMEError, "No challenges received")
```

vladopajic: debatable but:

by RFC 8555 is it ever possible to receive zero `authorizations` ?
if answer is `no` then:
  - code should raise if it fail to parse at least 1 challenge in `ACMEAuthorizationsResponse.parse()`
  - we shouldn't do `if authorizationsResponse.challenges.len ==0` check as per RFC  and since raise in parsing guarantee that authorizations response will always have challenges 
      - note it's our decision where to check `authorizationsResponse.challenges.len` once when parsing all every time where we use this struct

vladopajic (approved):

richard-ramos (approved):

PR #2265, opened 2026-04-13
feat(switchbuilder): add withHolePunching

SionoiS:
I added an option to the switch builder so that hole punching is easier to setup.

It uses default values when possible but doesn't allow much configuration. Is that fine? Please Advise.

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -22,6 +22,7 @@ import
   protocols/[identify, secure/secure, secure/noise, rendezvous, kademlia],
   protocols/connectivity/[
     autonat/server,
+    autonat/client,
     autonat/service,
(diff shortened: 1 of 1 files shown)


libp2p/builders.nim
```diff
@@ -317,6 +320,18 @@ proc withAutonatV2*(
   )
   b
 
+proc withHolePunching*(
+    b: SwitchBuilder, maxNumRelays: int = 5, onReservationHandler: proc = nil
```
vladopajic: if we are not really sure that these should be default values then better not have default values.

vladopajic (approved):

gmelodie (approved):

richard-ramos (approved):

PR #2267, opened 2026-04-13
feat(service-disco): add ipTree, signatures

gmelodie:
## Summary

This PR lays the groundwork for the Logos capability-discovery protocol by introducing two self-contained building blocks:

* The `IpTree` data structure for Sybil-resistance rate-limiting
* Cryptographic `sign`/`verify` on the `Ticket` type for registration flow control

Both components include full unit test coverage.

### Changes

* `libp2p/protocols/service_discovery/types.nim`

  * Added `IpTreeNode` and `IpTree` ref types

* `libp2p/protocols/service_discovery/iptree.nim` (new)

  * `IpTree.new()` — constructs an empty 32-level binary prefix tree
  * `insertIp` / `removeIp` — increment/decrement counters along an IPv4 address's 32-bit path

    * IPv6 is rejected with an error
    * Counters never go below zero
  * `ipScore` — returns a `[0.0, 1.0]` similarity score measuring how many prefix nodes exceed the expected-count threshold (`root.counter / 2^depth`)

    * Used by the registrar to detect subnet clustering

* `libp2p/protocols/kademlia/protobuf.nim`

  * Added `stew/endians2` and `crypto/crypto` imports
  * `sign(ticket, privateKey)`

    * Serializes `advertisement || tInit || tMod || tWaitFor || expiresAt || nonce` (big-endian)
    * Signs with the given key and stores in `ticket.signature`
  * `verify(ticket, publicKey)`

    * Reconstructs the same byte string and verifies the stored signature

* `tests/libp2p/service_discovery/test_iptree.nim` (new)

  * 21 tests covering:

    * Construction
    * Insert/remove correctness
    * Counter floor behavior
    * IPv6 rejection
    * Full `ipScore` gradient from `/32` to cross-octet differences

* `tests/libp2p/service_discovery/test_signatures.nim` (new)

  * 12 tests covering:

    * Happy-path sign/verify
    * Wrong-key rejection
    * Empty/corrupted signatures
    * Field-level tamper detection for all signed fields
    * Boundary values
    * Re-sign overwrite semantics

* `tests/libp2p/service_discovery/utils.nim`

  * Added `makeTicket` and `signedTicket` helpers
  * Re-exports protobuf so tests can access `Ticket` without extra imports

---

## Affected Areas

* Gossipsub
* Transports
* Peer Management / Discovery

  * Introduces `IpTree` (subnet similarity scoring) and `Ticket` sign/verify
  * Core primitives for the capability-discovery registration protocol
* Protocol Logic

  * `kademlia/protobuf.nim` gains:

    * New dependencies (`crypto`, `endians2`)
    * New procs on the existing `Ticket` type
* Build / Tooling
* Other

---

All existing encode/decode paths in `kademlia/protobuf.nim` are unaffected — changes are purely additive.

* No existing `Ticket` fields are modified
* `IpTreeNode` / `IpTree` are appended and do not alter existing APIs

Downstream impact:

* **Nimbus**: no impact (does not use service discovery types)
* **Waku**: no impact (protobuf encode/decode unchanged)
* **Codex**: no impact (does not depend on these modules)

---

## Impact on Library Users

* New types:

  * `IpTreeNode`, `IpTree` (exported)

* New procs:

  * `IpTree.new`, `insertIp`, `removeIp`, `ipScore`
  * `Ticket.sign(privateKey)`
  * `Ticket.verify(publicKey)`

* Dependency change:

  * `kademlia/protobuf` now imports `crypto/crypto`
  * Users vendoring this file directly must include the transitive dependency

---

## Risk Assessment

* `protobuf.nim` now imports `crypto/crypto`

  * Introduces a compile-time dependency on the crypto subsystem
  * Runtime behavior unchanged
  * Possible minor build-time increase
    **Risk: low**

* `ipScore` divides by `2^depth` across 32 levels

  * No division-by-zero (depth 0 divisor = 1)
  * `nil` guard ensures safe handling of sparse trees
    **Risk: low**

* `removeIp` returns `ok()` even if the IP/path does not exist

  * Ensures idempotent behavior
  * Cannot distinguish "not present" vs "removed"
    **Risk: low (intentional design)**

---

## References

* Feature branch: `feat--logos-capability-discovery`


libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -284,3 +285,35 @@ proc decode*(T: type Message, pb: ProtoBuffer): ProtoResult[T] =
...
+  ## Sign the ticket with the given private key.
+  ## Signature covers: advertisement || tInit || tMod || tWaitFor || expiresAt || nonce
+  var sigInput =
+    newSeqOfCap[byte](ticket.advertisement.len + 8 + 8 + 4 + 8 + ticket.nonce.len)
+  sigInput.add(ticket.advertisement)
+  sigInput.add(@(toBytesBE(ticket.tInit)))
+  sigInput.add(@(toBytesBE(ticket.tMod)))
+  sigInput.add(@(toBytesBE(ticket.tWaitFor)))
+  sigInput.add(@(toBytesBE(ticket.expiresAt)))
+  sigInput.add(ticket.nonce)
```
richard-ramos: This seems to be missing encoding the actual length of the advertisement and nonces i.e.

```suggestion
  sigInput.add(@(toBytesBE(uint32(ticket.advertisement.len))))
  sigInput.add(ticket.advertisement)
  sigInput.add(@(toBytesBE(ticket.tInit)))
  sigInput.add(@(toBytesBE(ticket.tMod)))
  sigInput.add(@(toBytesBE(ticket.tWaitFor)))
  sigInput.add(@(toBytesBE(ticket.expiresAt)))
sigInput.add(@(toBytesBE(uint32(ticket.nonce.len))))
  sigInput.add(ticket.nonce)
```

libp2p/protocols/service_discovery/iptree.nim
```diff
@@ -0,0 +1,103 @@
...
+      v.counter += 1
+      if (b and (1'u8 shl bit)) == 0:
+        if v.left.isNil:
+          v.left = IpTreeNode(counter: 0)
+        v = v.left
+      else:
+        if v.right.isNil:
+          v.right = IpTreeNode(counter: 0)
+        v = v.right
+  ok()
```
richard-ramos: @copilot , is this incrementing the current node before descending? wouldnt this mean that the last leaf gets created but never counted? 

Could you implement a test case that tests this is not the case as a comment here? 

DO NOT COMMIT NOR MAKE ANY CHANGES TO THIS PR!!

libp2p/protocols/kademlia/protobuf.nim
gmelodie: I think this is correct, or, at least according to @SionoiS's implementation [here](https://github.com/vacp2p/nim-libp2p/pull/2055/changes#diff-9cf8a98a04175697bc5f864f0131d8f9d94ee8dafb8a94f19af34df3e15b357fR155-R169)

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -284,3 +285,35 @@ proc decode*(T: type Message, pb: ProtoBuffer): ProtoResult[T] =
...
+): Result[void, CryptoError] {.raises: [], gcsafe.} =
+  ## Sign the ticket with the given private key.
+  ## Signature covers: advertisement || tInit || tMod || tWaitFor || expiresAt || nonce
+  var sigInput =
+    newSeqOfCap[byte](ticket.advertisement.len + 8 + 8 + 4 + 8 + ticket.nonce.len)
+  sigInput.add(ticket.advertisement)
+  sigInput.add(@(toBytesBE(ticket.tInit)))
+  sigInput.add(@(toBytesBE(ticket.tMod)))
+  sigInput.add(@(toBytesBE(ticket.tWaitFor)))
+  sigInput.add(@(toBytesBE(ticket.expiresAt)))
```
vladopajic: `var ticketBytes = ticket.toBytes()`

add `toBytes` utility

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -284,3 +285,35 @@ proc decode*(T: type Message, pb: ProtoBuffer): ProtoResult[T] =
...
+proc verify*(ticket: Ticket, publicKey: PublicKey): bool {.raises: [], gcsafe.} =
+  ## Verify the ticket signature against the given public key.
+  var sigInput =
+    newSeqOfCap[byte](ticket.advertisement.len + 8 + 8 + 4 + 8 + ticket.nonce.len)
+  sigInput.add(ticket.advertisement)
+  sigInput.add(@(toBytesBE(ticket.tInit)))
+  sigInput.add(@(toBytesBE(ticket.tMod)))
+  sigInput.add(@(toBytesBE(ticket.tWaitFor)))
+  sigInput.add(@(toBytesBE(ticket.expiresAt)))
+  sigInput.add(ticket.nonce)
```
vladopajic: `var ticketBytes = ticket.toBytes()`

ruse `toBytes` utility here

libp2p/protocols/service_discovery/types.nim
```diff
@@ -16,6 +16,13 @@ const
   ExtendedServiceDiscoveryCodec* = "/logos/service-discovery/1.0.0"
 
 type
+  IpTreeNode* = ref object
+    counter*: int
+    left*, right*: IpTreeNode
+
+  IpTree* = ref object
+    root*: IpTreeNode
+
```
vladopajic: what was reasoning for not having type with implementation?

tests/libp2p/utils/test_iptree.nim
```diff
@@ -0,0 +1,240 @@
...
+    # Inserting the same IP again increments counters again
+    discard tree.insertIp(ip1)
+    check tree.root.counter == 4
+    check tree.root.left.counter == 2
+
+  test "removeIp decrements counters correctly":
+    let tree = IpTree.new()
+    let ip1 = parseIpAddress("64.0.0.1")
+    let ip2 = parseIpAddress("192.168.1.1")
+    let ip3 = parseIpAddress("128.0.0.1")
```
vladopajic: this is way to go...
move scope of ip variables up, so that all tests can use them. do not do `parseIpAddress` in any of tests anymore

tests/libp2p/service_discovery/utils.nim
```diff
@@ -3,10 +3,30 @@
...
+    expiresAt: 0,
+    nonce: @[],
+    signature: @[],
+  )
+
+proc signedTicket*(privateKey: PrivateKey): Ticket =
+  var t = makeTicket()
+  let res = t.sign(privateKey)
+  doAssert res.isOk(), "sign failed in test helper"
+  t
```
vladopajic: this should be inside `test_signatures.nim` file for now, for now....

libp2p/protocols/service_discovery/iptree.nim
```diff
@@ -0,0 +1,106 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
```
vladopajic, line 1: type `IpTree` could have easily be defined outside of `discovery` package. consider moving this out to `nim-libp2p/libp2p/utils` for example
- it existence is more closer to existence to `std/net` package then `discovery`
- when looking from perspective of `discovery` package, the `IpTree` type should be considered as type provided other library as it is not tied to `discovery`'s package business logic
- it will reduce total code of `discovery` (total files, types and everything that we have to manage here)

libp2p/protocols/service_discovery/iptree.nim
```diff
@@ -0,0 +1,106 @@
...
+# Copyright (c) Status Research & Development GmbH
+
+import std/net
+import results
+import ./types
+
+proc new*(T: typedesc[IpTree]): T =
+  T(root: IpTreeNode(counter: 0))
+
+proc insertIp*(ipTree: IpTree, ip: IpAddress): Result[void, string] {.raises: [].} =
```
vladopajic:
```
proc insertIp*(ipTree: IpTree, ip: IpAddress): Result[void, string] {.raises: [].}
proc removeIp*(ipTree: IpTree, ip: IpAddress): Result[void, string] {.raises: [].}
proc ipScore*(ipTree: IpTree, ip: IpAddress): Result[float64, string] {.raises: [].}
...
```

all `proc`s of `IpTree` return `Result` which seems like bad decision. it makes usage cumbersome.
since all what result method does is process only one case where ipv6 address is used, `IpTree` should only assume correct behavior in it's implementation - that is only `ipv4` addresses will ever be provided. if anyone does otherwise - shame on you.

`IpTree` seems like type that sits at very bottom on information flow, `ipv4` constraint should be handled in levels above. or if necessary better do it before calling procs of this type.
the same analogy as we had yesterday: `AcmeResponse.parse()` should ensure that no response with zero challenges are passed down to lower layers. similarly here, `IpTree` type should assume the same, no ipv4 address will ever be passed down, but who will be responsible for that will will see later.

libp2p/protocols/service_discovery/types.nim
SionoiS: Historical reason, which is not good reason...

libp2p/protocols/kademlia/protobuf.nim
SionoiS: The specification says sign the concat of `ad || t_init || t_mod || t_wait`

https://github.com/vacp2p/rfc-index/blob/master/docs/ift-ts/raw/logos-service-discovery.md#base-message-structure

I never noticed this... Where do nonces and expire_at come from 🤔

edit: I checked the code both are barely used and can be removed.

libp2p/protocols/service_discovery/iptree.nim
```diff
@@ -0,0 +1,5 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Copyright (c) Status Research & Development GmbH
+
+import ../../utils/iptree
+export iptree
```
vladopajic: why? why not just import it as `import ../../utils/iptree` were needed?

libp2p/protocols/service_discovery/types.nim
```diff
@@ -8,8 +8,11 @@ import
     peerid, switch, multihash, cid, multicodec, routing_record, extended_peer_record
   ]
 import ../../protobuf/minprotobuf
+import ../../utils/iptree
 import ../kademlia/types
 
+export iptree
+
```
vladopajic: why? we don't really see usage of `iptree` in service_disco to justify this

vladopajic (approved):
lgtm, but unnecessary imports must go, and should be added when needed if needed how needed

richard-ramos (approved):

PR #2269, opened 2026-04-13
feat(gossip): reduce score of slow peers

richard-ramos:
## Summary

This PR adds a separate slow-peer scoring signal for gossipsub peers that repeatedly overflow their medium- or low-priority outbound queues.

It is needed because high-priority queue overflow already disconnects a peer immediately, but repeated medium/low queue drops previously had no direct effect on peer score or pruning. The change introduces a mild secondary penalty that can help deprioritize persistently slow peers without changing existing gossip parameters

Concretely, this PR:
- adds `slowPeerPenaltyWeight`, `slowPeerPenaltyThreshold`, and `slowPeerPenaltyDecay` to `GossipSubParams`
- tracks `slowPeerPenalty` in both live peer state and retained peer stats
- increments that penalty only when medium/low queue overflow causes message drops
- applies and decays the penalty during score updates
- keeps high-priority overflow behavior unchanged: it still disconnects and does not affect score
- refactors queue-action selection into a dedicated helper to make the behavior explicit and directly testable

## Affected Areas

- [x] Gossipsub  
  - Adds a new slow-peer scoring component, parameter validation, retained-state restore logic, and queue-overflow integration.  Changes how medium/low queue overflow contributes to pruning pressure, while preserving high-priority disconnect behavior.
  - Adds test coverage for params, score calculation/decay, queue-action decisions, retained score state, and heartbeat pruning.  
- [ ] Transports
- [ ] Protocol Logic  
- [ ] Build / Tooling
- [ ] Other  

- **Nimbus:**  
  Not yet validated in a downstream integration branch.

- **Waku:**  
  Not yet validated in a downstream integration branch.

- **Codex:**  
  Not yet validated in a downstream integration branch.

## Impact on Library Users

Downstream users get three new gossipsub tuning parameters:
- `slowPeerPenaltyWeight`
- `slowPeerPenaltyThreshold`
- `slowPeerPenaltyDecay`

Behaviorally, peers that repeatedly overflow medium/low outbound queues can now accumulate a mild score penalty and eventually be pruned. The default settings on this branch are conservative but enabled by default: `weight = -0.05`, `threshold = 2.0`, `decay = 0.2`.

There is also a new public queue-action helper in `pubsubpeer` used to make queue behavior deterministic and testable, but it does not require migration for normal library consumers.

## Risk Assessment

Main risk is network-behavior drift: peers that are locally slow to drain medium/low queues may now see lower scores and be pruned more often than before.

Additional risks:
- the signal is based on local backpressure, so its impact is workload- and deployment-dependent
- defaults are conservative but active by default
- queue-action logic was refactored, so regressions in send/queue routing are possible even though focused tests pass

The change does not expand the disconnect surface for high-priority overflow, and it does not modify mesh delivery penalties or protocol-misbehavior scoring.

## References

- Prior related work: `vacp2p/nim-libp2p#1083`
- Upstream comparison considered during design: `go-libp2p-pubsub` and `rust-libp2p` slow-peer behavior

### DIFF
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -95,16 +95,10 @@ type
       ## Relayed messages and IWANT replies
       ## They are dropped if exceeded max. Peer is not disconnected.
 
-  QueueActionType* {.pure.} = enum
-    SendHigh
-    SendMedium
-    DropMedium
-    SendLow
-    DropLow
-    Disconnect
-
   QueueAction* = object
-    action*: QueueActionType
+    priority*: MessagePriority
+    send*: bool
+      # if true msg should be sent respecting priority, if false it means drop/disconnect
     slowPeerPenaltyDelta*: float64
 
   PubSubPeerEventKind* {.pure.} = enum
@@ -244,21 +238,29 @@ func determineQueueAction*(
     sendPriorityQueueLen == 0 and mediumPriorityQueueLen == 0 and
     lowPriorityQueueLen == 0
 
+  var send = true
+  var slowPeerPenaltyDelta = 0.0
+  var priority = priority
+
   if priority == MessagePriority.High or emptyQueues:
+    # If queues are empty, we can send medium/low messages immediatly
+    # thus setting the priority of the message to high
+    priority = MessagePriority.High
+
     if sendPriorityQueueLen >= maxHighPriorityQueueLen:
-      QueueAction(action: QueueActionType.Disconnect, slowPeerPenaltyDelta: 0.0)
-    else:
-      QueueAction(action: QueueActionType.SendHigh, slowPeerPenaltyDelta: 0.0)
+      send = false
   elif priority == MessagePriority.Medium:
     if mediumPriorityQueueLen >= maxMediumPriorityQueueLen:
-      QueueAction(action: QueueActionType.DropMedium, slowPeerPenaltyDelta: 1.0)
-    else:
-      QueueAction(action: QueueActionType.SendMedium, slowPeerPenaltyDelta: 0.0)
+      send = false
+      slowPeerPenaltyDelta = 1.0
   else:
     if lowPriorityQueueLen >= maxLowPriorityQueueLen:
-      QueueAction(action: QueueActionType.DropLow, slowPeerPenaltyDelta: 1.0)
-    else:
-      QueueAction(action: QueueActionType.SendLow, slowPeerPenaltyDelta: 0.0)
+      slowPeerPenaltyDelta = 1.0
+      send = false
+
+  QueueAction(
+    priority: priority, send: send, slowPeerPenaltyDelta: slowPeerPenaltyDelta
+  )
 
 func determineQueueAction*(
     priority: MessagePriority,
@@ -525,7 +527,7 @@ proc enqueueNonHighPriorityMessage(
     when defined(pubsubpeer_queue_metrics):
       libp2p_gossipsub_low_priority_queue_size.inc(labelValues = [$p.peerId])
   of MessagePriority.High:
-    doAssert false, "high-priority messages are not enqueued"
+    raiseAssert "high-priority messages are not enqueued"
   p.rpcmessagequeue.dataAvailableEvent.fire()
 
   return newFutureCompleted[void]()
@@ -544,7 +546,7 @@ proc dropNonHighPriorityMessage(
       libp2p_pubsub_low_priority_queue_drops.inc()
     trace "low priority queue full, dropping message", p
   of MessagePriority.High:
-    doAssert false, "high-priority messages are not dropped via queue overflow scoring"
+    raiseAssert "high-priority messages are not dropped via queue overflow scoring"
   return newFutureCompleted[void]()
 
 proc sendEncoded*(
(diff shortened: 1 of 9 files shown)


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -222,6 +235,42 @@ func outbound*(p: PubSubPeer): bool =
   # This behaviour is presrcibed to counter sybil attacks and ensures that a coordinated inbound attack can never fully take over the mesh
   if not p.sendConn.isNil and p.sendConn.transportDir == Direction.Out: true else: false
 
+func determineQueueAction*(
+    priority: MessagePriority,
+    sendPriorityQueueLen, mediumPriorityQueueLen, lowPriorityQueueLen: int,
+    maxHighPriorityQueueLen, maxMediumPriorityQueueLen, maxLowPriorityQueueLen: int,
+): QueueAction =
```
gmelodie, line 236: Maybe add a comment explaining that we want to be connected to high, medium and low peers to ensure good network health

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -519,26 +570,36 @@ proc sendEncoded*(
...
+        queueAction.slowPeerPenaltyDelta, MessagePriority.Medium
+      )
+    of QueueActionType.SendLow:
+      p.enqueueNonHighPriorityMessage(msg, useCustomConn, MessagePriority.Low)
+    of QueueActionType.DropLow:
+      p.dropNonHighPriorityMessage(
+        queueAction.slowPeerPenaltyDelta, MessagePriority.Low
+      )
+    of QueueActionType.Disconnect:
+      p.disconnectPeer()
```
gmelodie: Can we not have more general functions (e.g. `sendMessage(priority, msg)` or `dropMessage(priority, msg)`)?

gmelodie (approved):

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -443,57 +492,59 @@ proc sendMsg(
...
     p.rpcmessagequeue.mediumPriorityQueue.addLast(queuedMsg)
     when defined(pubsubpeer_queue_metrics):
       libp2p_gossipsub_medium_priority_queue_size.inc(labelValues = [$p.peerId])
-    p.rpcmessagequeue.dataAvailableEvent.fire()
+  of MessagePriority.Low:
+    p.rpcmessagequeue.lowPriorityQueue.addLast(queuedMsg)
+    when defined(pubsubpeer_queue_metrics):
+      libp2p_gossipsub_low_priority_queue_size.inc(labelValues = [$p.peerId])
+  of MessagePriority.High:
+    doAssert false, "high-priority messages are not enqueued"
```
vladopajic:
```suggestion
    raiseAssert "high-priority messages are not enqueued"
```

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -443,57 +492,59 @@ proc sendMsg(
...
       libp2p_pubsub_low_priority_queue_drops.inc()
     trace "low priority queue full, dropping message", p
-  else:
-    p.rpcmessagequeue.lowPriorityQueue.addLast(queuedMsg)
-    when defined(pubsubpeer_queue_metrics):
-      libp2p_gossipsub_low_priority_queue_size.inc(labelValues = [$p.peerId])
-    p.rpcmessagequeue.dataAvailableEvent.fire()
-
+  of MessagePriority.High:
+    doAssert false, "high-priority messages are not dropped via queue overflow scoring"
```
vladopajic:
```suggestion
    raiseAssert "high-priority messages are not dropped via queue overflow scoring"
```

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -95,6 +95,18 @@ type
...
+    SendHigh
+    SendMedium
+    DropMedium
+    SendLow
+    DropLow
+    Disconnect
+
+  QueueAction* = object
+    action*: QueueActionType
+    slowPeerPenaltyDelta*: float64
```
vladopajic, line 102:
```
  QueueAction* = object
    priority: MessagePriority
    send*: bool  # if true msg should be sent respecting priority, if false it means drop/disconnect
    slowPeerPenaltyDelta*: float64
```

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -222,6 +235,42 @@ func outbound*(p: PubSubPeer): bool =
...
+  elif priority == MessagePriority.Medium:
+    if mediumPriorityQueueLen >= maxMediumPriorityQueueLen:
+      QueueAction(action: QueueActionType.DropMedium, slowPeerPenaltyDelta: 1.0)
+    else:
+      QueueAction(action: QueueActionType.SendMedium, slowPeerPenaltyDelta: 0.0)
+  else:
+    if lowPriorityQueueLen >= maxLowPriorityQueueLen:
+      QueueAction(action: QueueActionType.DropLow, slowPeerPenaltyDelta: 1.0)
+    else:
+      QueueAction(action: QueueActionType.SendLow, slowPeerPenaltyDelta: 0.0)
```
vladopajic:
```nim
  var send: bool = true
  var slowPeerPenaltyDelta: 0.0
  
  if priority == MessagePriority.High or emptyQueues:
    send = sendPriorityQueueLen >= maxHighPriorityQueueLen:
  elif priority == MessagePriority.Medium:
    if mediumPriorityQueueLen >= maxMediumPriorityQueueLen
         slowPeerPenaltyDelta = 1.0
         send = false
  else:
    if lowPriorityQueueLen >= maxLowPriorityQueueLen:
         slowPeerPenaltyDelta = 1.0
         send = false
  
  QueueAction(priority, send, slowPeerPenaltyDelta)
```

vladopajic (commented):
`QueueActionType` encodes semantics of `MessagePriority`. 
it's better not to have `QueueActionType` as it duplicates `MessagePriority` type in very small context of code.

libp2p/protocols/pubsub/pubsubpeer.nim
vladopajic: note that for `QueueAction`, `priority` is already there as `MessagePriority`.
converting `MessagePriority` to different type is just same thing with extra steps.

vladopajic (approved):

arnetheduck:
> The change introduces a mild secondary penalty that can help deprioritize persistently slow peers

this is somewhat risky - ie consider a slow peer that nonetheless is able to keep up _on average_ - we don't want to descore those since this will prevent the participation of these peers in the network generally as they will get descored. 

Consider a bursty protocol that posts 1000 1kb messages in a short burst and then is silent - if the "network requirement" is 1mb/s, this protocol works as long as there is one such burst per second but the 10mb/s peers will clear their queue faster even though 2mb/s peers are _also_ fast enough - in such a network, we don't want to discriminate against 2mb/s peers.

richard-ramos:
@arnetheduck , I'll change this to be opt-in in next PR just to be safe

richard-ramos:
https://github.com/vacp2p/nim-libp2p/pull/2429

PR #2282, opened 2026-04-15
refactor(connmanager): constructors

vladopajic:
## Summary

this pr adds specialized constructors for connection manager that avoid explosion of augments for (new) following connection manager operation modes. 

```
ConnManager.newMaxTotal(...)
ConnManager.newMaxInOut(...)
ConnManager.newWatermark(...)
... and others
```

## Affected Areas

- [x] ConnectionManager

- **Nimbus:**  
 - does not use  `ConnManager.new` 

- **Waku:**  
 - does not use `ConnManager.new` 

- **Codex:**  
  

## Impact on Library Users

there are under the hood changes. connection manager is not directly used.

## Risk Assessment

low risk.

## References

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -267,7 +267,6 @@ proc withMaxIn*(
     b: SwitchBuilder, maxIn: int
 ): SwitchBuilder {.public, deprecated: "Use withMaxInOut() instead".} =
   ## Maximum concurrent incoming connections. Should be used with `withMaxOut<#withMaxOut,SwitchBuilder,int>`_
-  doAssert maxIn > 0, "`maxIn` must be greater than 0"
   b.maxIn = maxIn
   b
 
@@ -275,7 +274,6 @@ proc withMaxOut*(
     b: SwitchBuilder, maxOut: int
 ): SwitchBuilder {.public, deprecated: "Use withMaxInOut() instead".} =
   ## Maximum concurrent outgoing connections. Should be used with `withMaxIn<#withMaxIn,SwitchBuilder,int>`_
-  doAssert maxOut > 0, "`maxOut` must be greater than 0"
   b.maxOut = maxOut
   b
 
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -394,11 +394,16 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
     else:
       Identify.new(peerInfo, b.sendSignedPeerRecord)
 
-  let
-    connManager =
-      ConnManager.new(b.maxConnsPerPeer, b.maxConnections, b.maxIn, b.maxOut)
-    ms = MultistreamSelect.new()
-    muxedUpgrade = MuxedUpgrade.new(b.muxers, secureManagerInstances, ms, connManager)
+  var connManager: ConnManager
+  if b.maxIn > 0 or b.maxOut > 0:
+    connManager = ConnManager.newMaxInOut(b.maxIn, b.maxOut, b.maxConnsPerPeer)
+  elif b.maxConnections > 0:
+    connManager = ConnManager.newMaxTotal(b.maxConnections, b.maxConnsPerPeer)
+  else:
+    connManager = ConnManager.newMaxTotal()
+
+  let ms = MultistreamSelect.new()
+  let muxedUpgrade = MuxedUpgrade.new(b.muxers, secureManagerInstances, ms, connManager)
 
   b.autotls.withValue(autotlsService):
     b.services.add(autotlsService)
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -259,16 +259,33 @@ proc withMaxConnections*(
 ): SwitchBuilder {.public.} =
   ## Maximum concurrent connections of the switch. You should either use this, or
   ## `withMaxIn <#withMaxIn,SwitchBuilder,int>`_ & `withMaxOut<#withMaxOut,SwitchBuilder,int>`_
+  doAssert maxConnections > 0, "`maxConnections` must be greater than 0"
   b.maxConnections = maxConnections
   b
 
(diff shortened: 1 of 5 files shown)


libp2p/builders.nim
```diff
@@ -394,11 +394,16 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
...
       Identify.new(peerInfo, b.sendSignedPeerRecord)
 
-  let
-    connManager =
-      ConnManager.new(b.maxConnsPerPeer, b.maxConnections, b.maxIn, b.maxOut)
-    ms = MultistreamSelect.new()
-    muxedUpgrade = MuxedUpgrade.new(b.muxers, secureManagerInstances, ms, connManager)
+  var connManager: ConnManager
+  if b.maxIn > 0 or b.maxOut > 0:
+    connManager = ConnManager.newMaxInOut(b.maxIn, b.maxOut, b.maxConnsPerPeer)
```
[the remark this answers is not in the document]
vladopajic: not necesserly

gmelodie (approved):
Okay, but I think this behavior should be a `ConnManagerConfig` with all three configs `maxIn`, `maxOut`, `maxConnections`, `maxConnectionsPerPeer` and an enum for the max interpretation/behavior. The config would then be passed as argument of `new`

```nim
type
  MaxConnectionsBehavior* = enum
    MaxInOut,
    MaxTotal
```

vladopajic:
> Okay, but I think this behavior should be a `ConnManagerConfig` 

that would be same thing as arguments, just in new type + extra enum.

```
let cfg = ConnManagerConfig(mode: MaxConnectionsBehavior.MaxInOut, maxIn: 10, maxOut: 20)
let cm = ConnManager.new(cfg)
```
vs
```
let cm = ConnManager.newMaxInOut(10, 20)
```

libp2p/builders.nim
```diff
@@ -394,11 +394,19 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
     else:
       Identify.new(peerInfo, b.sendSignedPeerRecord)
 
-  let
-    connManager =
-      ConnManager.new(b.maxConnsPerPeer, b.maxConnections, b.maxIn, b.maxOut)
-    ms = MultistreamSelect.new()
-    muxedUpgrade = MuxedUpgrade.new(b.muxers, secureManagerInstances, ms, connManager)
+  var connManager: ConnManager
```
richard-ramos, line 412: Should a doAssert be added `maxConnections > 0 && (maxIn > 0 || maxOut > 0)` to not use them together ?

libp2p/connmanager.nim
```diff
@@ -78,35 +78,46 @@ type
...
   result = newException(TooManyConnectionsError, "Too many connections")
 
-proc new*(
+proc newMaxTotal*(
     C: type ConnManager,
-    maxConnsPerPeer = MaxConnectionsPerPeer,
     maxConnections = MaxConnections,
-    maxIn = -1,
-    maxOut = -1,
+    maxConnsPerPeer = MaxConnectionsPerPeer,
```
richard-ramos, line 84: nit: let's keep the original order of these two parameters:
```
    maxConnsPerPeer = MaxConnectionsPerPeer,
    maxConnections = MaxConnections,
```

libp2p/builders.nim
```diff
@@ -394,11 +394,19 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
...
+  if b.maxIn > 0 or b.maxOut > 0:
+    if b.maxIn > 0 and b.maxOut > 0:
+      connManager = ConnManager.newMaxInOut(b.maxIn, b.maxOut, b.maxConnsPerPeer)
+    else:
+      raiseAssert "withMaxIn() should be paired with withMaxOut()"
```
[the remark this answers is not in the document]
richard-ramos: This makes me think that since both parameters are to be configured together, maybe we should deprecate `withMaxIn` and `withMaxOut`, and create a `withMaxInOut(in, out)` :)

libp2p/builders.nim
richard-ramos, line 263:
```suggestion
  doAssert maxConnections > 0, "Invalid connection counts!"
  b.maxConnections = maxConnections
```

since we got rid of this condition in `new`, we should probably add it here to avoid letting the dev introduce a bad default

libp2p/builders.nim
vladopajic: it will be braking change, because current code allows that

libp2p/builders.nim
richard-ramos: Hm... thinking it more throughly, this should still continue in new. Perhaps something like this, to not change the behavior of the connection manager constructor :thinking: :

```
  var connManager: ConnManager 
  if b.maxIn > 0 or b.maxOut > 0:
    if b.maxIn > 0 and b.maxOut > 0:
      connManager = ConnManager.newMaxInOut(b.maxIn, b.maxOut, b.maxConnsPerPeer)
    else:
      raiseAssert "withMaxIn() should be paired with withMaxOut()" 
  elif b.maxConnections > 0:
    connManager = ConnManager.newMaxTotal(b.maxConnections, b.maxConnsPerPeer)
  else:
    raiseAssert, "Invalid connection counts!"
    ```

vladopajic: better in with proc sto stop it right away

libp2p/builders.nim
vladopajic: we should allow anything configured. but use the option with more 'prority'

libp2p/connmanager.nim
vladopajic: it was moved by intention to be last place to allow constructor without naming arguments `newMaxTotal(1)` means "max total connections" and not "max connections per peer"

libp2p/builders.nim
vladopajic: if nothing is specified we can always create `ConnManager` with default values. there is not reason to crash there

richard-ramos (approved):

PR #2284, opened 2026-04-15
feat(connmanager): watermark

vladopajic:
## Summary

as per `2026q1-nimlibp2p-connection-management` commitment target, this pr adds:
- hi/low watermark mode
- protect/unprotect 

## Affected Areas

- [x] ConnectionManager

pr has only backwards compatible changes

## Impact on Library Users

no impact; internal changes

## Risk Assessment

low. all changes covered with unit tests.

## References

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -98,6 +98,7 @@ type
     observedAddrManager: ObservedAddrManager
     enableWildcardResolver: bool
     addressPolicy: PeerAddressPolicy
+    watermarkCfg: Opt[WatermarkConfig]
 
 proc new*(T: type[SwitchBuilder]): T {.public.} =
   ## Creates a SwitchBuilder
@@ -121,6 +122,7 @@ proc new*(T: type[SwitchBuilder]): T {.public.} =
     kad: Opt.none(KadInfo),
     enableWildcardResolver: true,
     addressPolicy: defaultAddressPolicy,
+    watermarkCfg: Opt.none(WatermarkConfig),
   )
 
 proc withPrivateKey*(
@@ -296,6 +298,27 @@ proc withMaxConnsPerPeer*(
   b.maxConnsPerPeer = maxConnsPerPeer
   b
 
+proc withWatermark*(
+    b: SwitchBuilder,
+    lowWater: int,
+    highWater: int,
+    gracePeriod: Duration,
+    silencePeriod: Duration,
+): SwitchBuilder {.public.} =
+  ## Enable hi/lo watermark connection management.
+  ## When connected peers exceed `highWater`, the connection manager trims
+  ## down to `lowWater`, skipping peers within `gracePeriod` and protected peers.
+  ## Takes priority over `withMaxConnections`/`withMaxIn`/`withMaxOut`.
+  b.watermarkCfg = Opt.some(
+    WatermarkConfig(
+      lowWater: lowWater,
+      highWater: highWater,
+      gracePeriod: gracePeriod,
+      silencePeriod: silencePeriod,
+    )
+  )
+  b
+
 proc withPeerStore*(b: SwitchBuilder, capacity: int): SwitchBuilder {.public.} =
   b.peerStoreCapacity = Opt.some(capacity)
   b
@@ -433,7 +456,9 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError], public.} =
       Identify.new(peerInfo, b.sendSignedPeerRecord)
 
   var connManager: ConnManager
-  if b.maxIn > 0 or b.maxOut > 0:
+  if b.watermarkCfg.isSome():
+    connManager = ConnManager.newWatermark(b.watermarkCfg.get(), b.maxConnsPerPeer)
+  elif b.maxIn > 0 or b.maxOut > 0:
     if b.maxIn > 0 and b.maxOut > 0:
       connManager = ConnManager.newMaxInOut(b.maxIn, b.maxOut, b.maxConnsPerPeer)
     else:
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -302,8 +302,8 @@ proc withWatermark*(
     b: SwitchBuilder,
     lowWater: int,
     highWater: int,
-    gracePeriod: Duration,
-    silencePeriod: Duration,
+    gracePeriod: Duration = 1.minutes,
+    silencePeriod: Duration = 10.seconds,
 ): SwitchBuilder {.public.} =
   ## Enable hi/lo watermark connection management.
   ## When connected peers exceed `highWater`, the connection manager trims
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -3,7 +3,7 @@
 
 {.push raises: [].}
 
-import std/[tables, sets, sequtils]
+import std/[algorithm, tables, sets, sequtils]
 import pkg/[chronos, chronicles, metrics]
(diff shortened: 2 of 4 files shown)


libp2p/connmanager.nim
```diff
@@ -471,6 +524,69 @@ proc dropPeer*(c: ConnManager, peerId: PeerId) {.async: (raises: [CancelledError
...
+      continue
+    candidates.add((connTime, peerId))
+
+  candidates.sort(
+    proc(a, b: (Moment, PeerId)): int =
+      a[0].cmp(b[0])
+  )
+
+  for (_, peerId) in candidates:
+    if c.muxerStore.countPeers() <= wm.lowWater:
```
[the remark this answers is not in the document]
vladopajic: well it depends how we want it.

protect is now implemented as hard guarantees - protect these peers at all cost. then, if libp2p is miss-configured/misused - say high water is 5 and 10 peers are protected, off course that rule will still be respected. 

if high water has more weight then protect, then this is different stance, and then this comment is valid.

libp2p/connmanager.nim
```diff
@@ -471,6 +527,78 @@ proc dropPeer*(c: ConnManager, peerId: PeerId) {.async: (raises: [CancelledError
 
   trace "Peer dropped", peerId, connCount = muxers.len
 
+proc protect*(c: ConnManager, peerId: PeerId, tag: string) =
+  ## Mark `peerId` as protected from watermark trimming under `tag`.
+  ## A peer remains protected as long as at least one tag is active.
+  c.protectedPeers.mgetOrPut(peerId, initHashSet[string]()).incl(tag)
+
+proc unprotect*(c: ConnManager, peerId: PeerId, tag: string): bool =
```
richard-ramos, line 535: Should this be called also inside `onPeerDisconnect`?

libp2p/builders.nim
```diff
@@ -278,6 +280,27 @@ proc withMaxConnsPerPeer*(
   b.maxConnsPerPeer = maxConnsPerPeer
   b
 
+proc withWatermark*(
+    b: SwitchBuilder,
+    lowWater: int,
+    highWater: int,
+    gracePeriod: Duration,
+    silencePeriod: Duration,
```
richard-ramos: Any recommended values for these two?

richard-ramos (approved):

libp2p/connmanager.nim
vladopajic: i don't think so... you when you protect peer you want to protect it once. peer can connect - disconnect - reconnect at any time. same protection should apply.

libp2p/builders.nim
vladopajic: added https://github.com/vacp2p/nim-libp2p/pull/2284/commits/b5bd55867e0dedd73063d218af4f0fa4b4ac45c0

libp2p/connmanager.nim
vladopajic: when i think of it again, deferentially we don't want to remove protection. it is value that user has set, so user should unset it.

libp2p/connmanager.nim
```diff
@@ -368,6 +412,12 @@ proc storeMuxer*(
 
   libp2p_peers.set(c.muxerStore.countPeers().int64)
 
+  if c.watermark.isSome:
+    if isNewPeer:
+      c.connectedAt[peerId] = Moment.now()
+    if c.muxerStore.countPeers() > c.watermark.get().highWater:
+      c.triggerTrim()
```
gmelodie, line 419: `withValue`

libp2p/connmanager.nim
```diff
@@ -391,28 +441,34 @@ proc storeMuxer*(
...
+      trace "Too many outgoing connections"
+      raise newTooManyConnectionsError()
   return ConnectionSlot(connManager: c, direction: Out)
 
 func semaphore(c: ConnManager, dir: Direction): AsyncSemaphore {.inline.} =
   return if dir == In: c.inSema else: c.outSema
 
 proc availableSlots*(c: ConnManager, dir: Direction): int =
+  if c.watermark.isSome:
+    return connectionsUnlimited # unlimited slots in watermark mode
```
gmelodie, line 466: It's not actually `unlimited` though right? Would it make sense to return the number of slots before trimming?

gmelodie (commented):
What happens when protected + grace period >= watermark high? Might be useful to have a test for that

libp2p/connmanager.nim
vladopajic: well it certainly is limited, by some physical factors of machine etc.., but we say "unlimited" because `availableSlots` proc does not really have meaning in watermark mode. it should always return the highest number possible as any other code that uses this proc should never fail thinking user can't connect.

conceptually watermark mode does not have "slots".

returning any big number is okay, even `3151` is okay, or `131` is okay here as well. but `1` is not okay as autonat (`hasEnoughIncomingSlots`) would fail. for same reasons 2 is better, 3 is still better then 2, `number of slots before trimming` is better all smaller values, but `int.high` is the best.

therefore, returning any other number makes less sense.

vladopajic:
> What happens when protected + grace period >= watermark high? Might be useful to have a test for that

`protected` peers stay protected. 

not sure what exactly you meant with this, but is this what you mean https://github.com/vacp2p/nim-libp2p/pull/2284#discussion_r3088862079?

libp2p/connmanager.nim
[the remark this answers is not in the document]
vladopajic: this is not good as it code can return 0

gmelodie:
> not sure what exactly you meant with this, but is this what you mean https://github.com/vacp2p/nim-libp2p/pull/2284#discussion_r3088862079?

That comment is about the trigger not happening, my concern is about no peers being trimmed even if trigger happens. If `highWatermark = 10; protectedPeers = 5; peersInGracefulWait = 5`, does this mean the trim will skip all peers and trim will keep getting triggered and not trimming anything?

vladopajic:
> trim will keep getting triggered and not trimming anything? 

yes

vladopajic:
cant see what i pasted via gh app, might be wrong link, but this one should be correct one:

https://github.com/vacp2p/nim-libp2p/pull/2284#discussion_r3088862079

libp2p/connmanager.nim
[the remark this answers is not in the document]
gmelodie: If protected peers are taking up all the space, I think it's a configuration problem (user should choose better values) aka they should not be trimmed

The issue I see particularly is with unprotected peers on the grace period. I think they should be trimmed 

to summarize
- if protected peers > highwater: don't even trigger trim
- else...
  - trim unprotected peers that are not in grace period
  - if still not below low water, trim peers in grace period until you reach low water
  - if trimmed all grace period peers and still not in low water, trim is complete

gmelodie (approved):
I'll leave the trimming issue to your judgement

libp2p/connmanager.nim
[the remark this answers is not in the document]
vladopajic: what you are referring here is different meaning of grace period - different protection guarantee. you would like grace period to be softer guarantee, where connection can be trimmed.

grace period usually means safe period. in this context grace period == protected. 

and the difference here is - how we want it to be. 

and the reasoning behind how it currently is, is because watermark mode is different type of guarantee then it allows burst of connections that are safe within grace period. 

protection guarantee schema that you indicate would not allow burst of connections. it would be similar to simple limiting mode because all connections will get trimmed down to limit (low watermark), with one exception of having protection.

PR #2286, opened 2026-04-15
feat(builder): `withPrivateAddressFilter`

### LINKED ISSUE #2248: feat: filter out private IPs
KademliaDHT (and probably Rendezvous as well) return addresses that belong to the private address space as defined in https://datatracker.ietf.org/doc/html/rfc1918#section-3 . Having a public facing DHT (or any kind of public discovery mechanism, really) return nodes with unreachable addresses pollutes the routing tables with nodes that are not really useful.

The specs for routing records (used on identify) indicates the following:
https://github.com/libp2p/specs/blob/master/RFC/0003-routing-records.md

> A peer SHOULD only include addresses that it believes are routable via the public internet, ideally having confirmed that this is the case via some external mechanism such as a successful AutoNAT dial-back.
> In some cases we may want to include localhost or LAN-local address; for example, when testing the DHT using many processes on a single machine. To support this, implementations may use a global runtime configuration flag or environment variable to control whether local addresses will be included.

Based on this, questions and things to verify are the following, before deciding on a solution
1 - Does Autonat fixes this? should Autonat remove the list of unreachable addresses from the announced addresses?
2 - If we decide to not announcing rfc1918 addresses be something that affects libp2p in general (i.e. identify), or be limited just to discovery mechanisms.
3 - We should make sure whatever we implement, to make it Opt-In either via a new switch option, or by adding parameters to the specific modules being affected

cc: @SionoiS

richard-ramos:
## Summary

Adds the ability to filter private/non-routable IP addresses from a node's announced addresses and from peer address records received from the network. This is done by adding two new `SwitchBuilder` methods:
- `withAddressPolicy(policy)`: accepts any custom policy
- `withPrivateAddressFilter()`: convenience wrapper that applies publicRoutableAddressPolicy()

Enabling one of those options propagates the policy to `PeerInfo`, `PeerStore` and `Kademlia`

Only addresses in `peerStore[AddressBook]` are affected by this change. SignedPeerRecords will still contain local addresses, because changing the record invalidates it. We could add a getter function that returns the filtered addresses should the need arise, though

Fixes: https://github.com/vacp2p/nim-libp2p/issues/2248

cc: @SionoiS 

## Affected Areas
- [ ] Gossipsub  
- [ ] Transports
- [ ] Protocol Logic
- [ ] Build / Tooling
- [ ] Other  : Switch Builder options

## Impact on Library Users
New options were added that are Opt-in.

## Risk Assessment
- Nodes behind NAT will lose all addresses. Those nodes should not enable this option
- Mix protocol might bypass this policy if adding entries manually to the MixNodePool

### DIFF
--- a/libp2p/peeraddrpolicy.nim
+++ b/libp2p/peeraddrpolicy.nim
@@ -7,42 +7,38 @@ import std/sequtils
 import results
 import multiaddress, routing_record, wire
 
-type PeerAddressPolicy* = ref object
+type PeerAddressPolicy* = object
   accept*: proc(ma: MultiAddress): bool {.gcsafe, raises: [].}
 
+proc defaultAddressPolicy*(): PeerAddressPolicy =
+  ## The default policy that accepts all addresses. Used so that consumers can
+  ## always rely on a policy with a non-nil `accept` proc.
+  PeerAddressPolicy(
+    accept: proc(ma: MultiAddress): bool {.gcsafe, raises: [].} =
+      true
+  )
+
 proc accepts*(policy: PeerAddressPolicy, ma: MultiAddress): bool =
-  policy.isNil or policy.accept.isNil or policy.accept(ma)
+  doAssert not policy.accept.isNil, "invalid peerAddr policy"
+  policy.accept(ma)
 
 proc filterAddrs*(
     policy: PeerAddressPolicy, addrs: openArray[MultiAddress]
 ): seq[MultiAddress] =
-  if policy.isNil or policy.accept.isNil:
-    return @addrs
+  doAssert not policy.accept.isNil, "invalid peerAddr policy"
   addrs.filterIt(policy.accept(it))
 
 proc preservesAddrs*(policy: PeerAddressPolicy, addrs: openArray[MultiAddress]): bool =
   policy.filterAddrs(addrs).len == addrs.len
 
 proc filterPeerRecord*(policy: PeerAddressPolicy, record: PeerRecord): Opt[PeerRecord] =
-  if policy.isNil or policy.accept.isNil:
-    return Opt.some(record)
+  doAssert not policy.accept.isNil, "invalid peerAddr policy"
 
   var filtered = record
   filtered.addresses.keepItIf(policy.accepts(it.address))
   if filtered.addresses.len == 0:
     return Opt.none(PeerRecord)
   Opt.some(filtered)
-
-proc allowsSignedPeerRecord*(
-    policy: PeerAddressPolicy, signedPeerRecord: seq[byte]
-): bool =
-  if policy.isNil or policy.accept.isNil:
-    return true
-
-  let spr = SignedPeerRecord.decode(signedPeerRecord).valueOr:
-    return false
-
-  policy.preservesAddrs(spr.data.addresses.mapIt(it.address))
 
 proc publicRoutableAddressPolicy*(): PeerAddressPolicy =
   PeerAddressPolicy(
--- a/libp2p/peeraddrpolicy.nim
+++ b/libp2p/peeraddrpolicy.nim
@@ -1,0 +1,55 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Copyright (c) Status Research & Development GmbH
+
+{.push raises: [].}
+
+import std/sequtils
+import results
+import multiaddress, routing_record, wire
+
+type PeerAddressPolicy* = ref object
+  accept*: proc(ma: MultiAddress): bool {.gcsafe, raises: [].}
+
+proc accepts*(policy: PeerAddressPolicy, ma: MultiAddress): bool =
+  policy.isNil or policy.accept.isNil or policy.accept(ma)
+
+proc filterAddrs*(
+    policy: PeerAddressPolicy, addrs: openArray[MultiAddress]
+): seq[MultiAddress] =
+  if policy.isNil or policy.accept.isNil:
+    return @addrs
+  addrs.filterIt(policy.accept(it))
+
+proc preservesAddrs*(
+    policy: PeerAddressPolicy, addrs: openArray[MultiAddress]
+): bool =
+  policy.filterAddrs(addrs).len == addrs.len
+
+proc filterPeerRecord*(
+    policy: PeerAddressPolicy, record: PeerRecord
+): Opt[PeerRecord] =
+  if policy.isNil or policy.accept.isNil:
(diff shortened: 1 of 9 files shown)


libp2p/protocols/kademlia/find.nim
```diff
@@ -158,14 +158,20 @@ proc dispatchFindNode*(
...
+    addressPolicy: PeerAddressPolicy,
+    rtable: RoutingTable,
+    peerInfos: seq[PeerInfo],
 ) {.raises: [].} =
   for p in peerInfos:
-    if rtable.insert(p.peerId):
+    let addrs = addressPolicy.filterAddrs(p.addrs)
+    if addrs.len == 0:
+      continue
+    if rtable.insert(p.peerId) and addrs.len > 0:
```
gmelodie: These two checks are redundant

libp2p/peeraddrpolicy.nim
```diff
@@ -0,0 +1,55 @@
...
+type PeerAddressPolicy* = ref object
+  accept*: proc(ma: MultiAddress): bool {.gcsafe, raises: [].}
+
+proc accepts*(policy: PeerAddressPolicy, ma: MultiAddress): bool =
+  policy.isNil or policy.accept.isNil or policy.accept(ma)
+
+proc filterAddrs*(
+    policy: PeerAddressPolicy, addrs: openArray[MultiAddress]
+): seq[MultiAddress] =
+  if policy.isNil or policy.accept.isNil:
```
gmelodie: These checks are annoying, why is policy even allowed to be nil? There should be a default and one should always be needed

libp2p/peeraddrpolicy.nim
```diff
@@ -0,0 +1,55 @@
...
+    return @addrs
+  addrs.filterIt(policy.accept(it))
+
+proc preservesAddrs*(
+    policy: PeerAddressPolicy, addrs: openArray[MultiAddress]
+): bool =
+  policy.filterAddrs(addrs).len == addrs.len
```
gmelodie: One liner -> `{.inline.}`

richard-ramos: Inline is to be avoided according to https://status-im.github.io/nim-style-guide/language.inline.html

libp2p/protocols/kademlia/find.nim
richard-ramos: i agree

libp2p/protocols/kademlia/find.nim
```diff
@@ -158,14 +158,20 @@ proc dispatchFindNode*(
...
-    switch: Switch, rtable: RoutingTable, peerInfos: seq[PeerInfo]
+    switch: Switch,
+    addressPolicy: PeerAddressPolicy,
+    rtable: RoutingTable,
+    peerInfos: seq[PeerInfo],
 ) {.raises: [].} =
   for p in peerInfos:
+    let addrs = addressPolicy.filterAddrs(p.addrs)
+    if addrs.len == 0:
+      continue
```
vladopajic, line 169: if this is first line inside of `updatePeers` why don't do it outside of `updatePeers` ?

libp2p/peeraddrpolicy.nim
richard-ramos: I got rid of one of the nil checks, and converted the check for policy.accept.isNil into an assertion, because there's nothing stopping us from passing a policy.accept that is nil (because callbacks/closures are nillable)

tests/libp2p/services/test_private_addr_filter.nim
```diff
@@ -0,0 +1,267 @@
...
+    peerstore,
+    protocols/identify,
+    protocols/kademlia,
+  ]
+import ../../tools/[unittest, crypto]
+
+proc ma(s: string): MultiAddress =
+  MultiAddress.init(s).tryGet()
+
+suite "isFilterablePrivateMA":
```
vladopajic: `isFilterablePrivateMA` suite should be in `tests/libp2p/test_wire` ?

vladopajic: in general these suites should be split following existing convention

vladopajic (approved):

libp2p/peeraddrpolicy.nim
```diff
@@ -0,0 +1,47 @@
...
+# Copyright (c) Status Research & Development GmbH
+
+{.push raises: [].}
+
+import std/sequtils
+import results
+import multiaddress, routing_record, wire
+
+type PeerAddressPolicy* = object
+  accept*: proc(ma: MultiAddress): bool {.gcsafe, raises: [].}
```
vladopajic: consider simplifying this to:
```
type PeerAddressPolicy proc(ma: MultiAddress): bool {.gcsafe, raises: [].}
```

then  `defaultAddressPolicy` would be 

```
proc defaultAddressPolicy*(): PeerAddressPolicy =
 proc(ma: MultiAddress): bool {.gcsafe, raises: [].} =
      true
```
or  put it in cost to avoid `()` 

```
const defaultAddressPolicy* =  proc(ma: MultiAddress): bool {.gcsafe, raises: [].} =
      true
```

libp2p/peeraddrpolicy.nim
```diff
@@ -0,0 +1,47 @@
...
+proc defaultAddressPolicy*(): PeerAddressPolicy =
+  ## The default policy that accepts all addresses. Used so that consumers can
+  ## always rely on a policy with a non-nil `accept` proc.
+  PeerAddressPolicy(
+    accept: proc(ma: MultiAddress): bool {.gcsafe, raises: [].} =
+      true
+  )
+
+proc accepts*(policy: PeerAddressPolicy, ma: MultiAddress): bool =
+  doAssert not policy.accept.isNil, "invalid peerAddr policy"
```
vladopajic: `doAssert not policy.accept.isNil` i am not sure about these kind of checks, why not leave default nil pointer de-reference error? are we going to expect doing this kind of checks everywhere?

libp2p/peeraddrpolicy.nim
vladopajic: the follow up it would be better if default value of type is "default address policy" - return's `true`.

for example current code 
```
type PeerAddressPolicy* = object
  accept*: proc(ma: MultiAddress): bool {.gcsafe, raises: [].} = defaultAddressPolicy()
```
avoids any checks `.isNill`...

gmelodie (approved):

PR #2288, opened 2026-04-16
test(gossipsub): extensions race condition + fix

rlve:
## Summary

Fixes a race condition where GossipSub extensions control messages could be sent *after* topic subscriptions on a newly opened stream, violating the protocol requirement that extensions control must be the first message exchanged.

The bug only manifests when a node already has topic subscriptions at the time it connects to a new peer, which is the normal startup order in production and interop tests (subscribe, then discover peers), but not how the existing unit tests were structured (connect, then subscribe).

### Root cause

Extensions control was sent from `onNewPeer`, which fires when the `PubSubPeer` object is created - before the stream exists. At that point `peer.codec` is empty, so the send was deferred through an async future (`await peer.codecInitializedFut`). When the stream later opened, the base class `PubSub.onPubSubPeerEvent` handler sent subscriptions synchronously and immediately. The async extensions future resolved in a later event loop iteration, arriving second.

### Fix

1. **Moved `sendExtensionsControl` from `onNewPeer` to the `StreamOpened` event handler.** At `StreamOpened` time the codec is already negotiated (set from `newConn.protocol` before the event fires), so the send executes synchronously - before `procCall FloodSub(p).onPubSubPeerEvent` sends subscriptions in the same call chain.

2. **Removed the async fallback path.** Since the call site now guarantees the codec is set, the `asyncSpawn` + `await codecInitializedFut` path was dead code. Replaced it with a `doAssert peer.codec != ""` so future callers from the wrong lifecycle point get a loud failure instead of a silent ordering violation.

3. **Made `addPeer` idempotent with a separate query.** Added `hasControlBeenSent` to check whether we've already sent extensions to a peer (e.g. after stream reconnect without full peer disconnect), keeping `addPeer` as a clean void mutation. Prevents duplicate extensions control messages and duplicate `onNegotiated` callbacks.

## Affected Areas

- [x] Gossipsub  
  Extensions control message ordering, peer event handling

- **Interop:** Fixes the failing 2-Nim-node GossipSub interop test in `test-plans/gossipsub-interop`

## Impact on Library Users

No API changes. Internal fix to message ordering on stream open.

## Risk Assessment

- **Low risk.** The change narrows the execution path (removes async, adds assert) rather than adding complexity.
- The ordering guarantee now relies on synchronous execution order within `onPubSubPeerEvent`, rather than on `MessagePriority.High` FIFO ordering across async boundaries.
- Stream reconnect (without full peer disconnect) is safe: `hasControlBeenSent` returns true, no duplicate control is sent, remote peer retains extensions state from the first handshake.

## Additional Notes

The existing "Test Extension" unit test didn't catch this because it connects nodes *before* subscribing to topics. With zero topics at connect time, the base class `StreamOpened` handler (`if p.topics.len > 0: p.sendSubs(...)`) is a no-op - there's nothing to race against. The new test subscribes before connecting to reproduce the real-world ordering.


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -284,30 +284,23 @@ proc usesExtensions(g: GossipSub): bool =
...
-    g.extensionsState.addPeer(peer.peerId)
-    g.send(
-      peer,
-      RPCMsg.withControl(
-        ControlMessage.withExtensions(g.extensionsState.makeControlExtensions())
-      ),
-      MessagePriority.High, # must be the first message on the stream
-    )
+  # Called from StreamOpened where codec is always set from the negotiated connection.
+  doAssert peer.codec != "", "sendExtensionsControl requires codec to be negotiated"
```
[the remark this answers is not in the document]
rlve: This is incorrect right? Would be good to add this to copilot-instructions because it always complains about it.

https://nim-lang.org/docs/assertions.html#doAssert

libp2p/protocols/pubsub/gossipsub/extensions.nim
```diff
@@ -137,15 +137,19 @@ proc heartbeat*(state: ExtensionsState) =
...
+  peerId in state.sentExtensions
 
-  state.sentExtensions.incl(peerId)
+proc addPeer*(state: ExtensionsState, peerId: PeerId) =
+  # called after peer has connected to node, when gossipsub is about to send
+  # extensions control message and mark the peer as sent from our side.
 
   # when node has received control extensions from peer then extensions have negotiated
-  if peerId in state.peerExtensions:
+  if peerId notin state.sentExtensions and peerId in state.peerExtensions:
```
vladopajic:
```suggestion
  if not state.hasControlBeenSent(peerId) and peerId in state.peerExtensions:
```

consider always reusing same logic

libp2p/protocols/pubsub/gossipsub/extensions.nim
```diff
@@ -137,15 +137,19 @@ proc heartbeat*(state: ExtensionsState) =
 
   state.onHeartbeat()
 
-proc addPeer*(state: ExtensionsState, peerId: PeerId) =
-  # called after peer has connected to node and extensions control message is sent by gossipsub.
+proc hasControlBeenSent*(state: ExtensionsState, peerId: PeerId): bool =
```
vladopajic: nit:

`isControlSent` / `hasControlSent` or `controlSent`

tests/libp2p/pubsub/component/test_gossipsub_extensions.nim
```diff
@@ -51,6 +51,235 @@ suite "GossipSub Component - Extensions":
...
         negotiatedPeersSorted == nodesPeerIdSorted
 
+  asyncTest "Extensions control is sent before subscriptions on stream open":
+    # If node has pre-existing subscriptions on a dial,
+    # then it sends the subscriptions control message before extensions control message on a newly opened stream,
+    # which violates the protocol contract.
+    const topic = "logos-extensions-ordering"
+    var
+      dialerFirstOutgoing = Opt.none(RPCMsg)
+      receiverFirstOutgoing = Opt.none(RPCMsg)
```
vladopajic: 1) `let dialerFirstMsgFut: Future[RPCMsg]`
2) then complete it from observer with first message
3) await for message `let msg = await dialerFirstMsgFut` 

`checkUntilTimeout` we should avoid/minimize using

tests/libp2p/pubsub/component/test_gossipsub_extensions.nim
```diff
@@ -51,6 +51,235 @@ suite "GossipSub Component - Extensions":
...
+    )
+
+    startAndDeferStop(nodes)
+    nodes.subscribeAllNodes(topic, voidTopicHandler)
+    await connect(nodes[1], nodes[0])
+
+    waitSubscribeStar(nodes, topic)
+
+    # Wait a few heartbeats to ensure no late duplicates arrive.
+    await nodes[1].waitForHeartbeatByEvent(5)
```
vladopajic: this will be potentially flaky and slow. two options:

1) if you need to wait on count things use WaitGroup:
```nim
let blocker = newWaitGroup(1) # 1 can be any number of things waited on

# if condition ....
   blocker.done() # "1 thing is done" aka reduce count by -1

await blocker.wait() # wait for all things to get done
```

2) use future (is limited to only one count) 
```nim
let fut: Future

# if condition ...
   fut.complete() 

await fut
```

tests/libp2p/pubsub/component/test_gossipsub_extensions.nim
```diff
@@ -51,6 +51,235 @@ suite "GossipSub Component - Extensions":
...
+    nodes.subscribeAllNodes(topic, voidTopicHandler)
+    await connect(nodes[1], nodes[0])
+
+    # Wait for subscriptions to propagate.
+    waitSubscribeStar(nodes, topic)
+
+    # None of the sent messages should contain extensions control.
+    check:
+      outgoingMsgs.len >= 1
+
```
vladopajic, line 236: what if messages do not happen right now. could this be flaky?

tests/libp2p/pubsub/component/test_gossipsub_extensions.nim
```diff
@@ -51,6 +51,235 @@ suite "GossipSub Component - Extensions":
...
+  asyncTest "Extensions re-negotiated after disconnect/reconnect with correct ordering":
+    const topic = "foobar"
+
+    var negotiatedPeers: seq[PeerId]
+    proc onNegotiated(peer: PeerId) {.gcsafe, raises: [].} =
+      negotiatedPeers.add(peer)
+
+    # Track all outgoing messages per (sender, receiver) pair in send order.
+    var outgoingMsgs: Table[string, seq[RPCMsg]]
```
vladopajic: `var outgoingMsgs: Table[string, Future[RPCMsg]]`
or
`var outgoingMsgs: Table[string, AsyncQueue[RPCMsg]]`

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -284,30 +284,23 @@ proc usesExtensions(g: GossipSub): bool =
...
-  # only then node should send extensions control message.
+  if not gossipExtensionsSupported(peer.codec):
+    return
 
-  # it's very likely that codecs will not be set at this point as connection was 
-  # not established and details of negotiated protocol (codecs) are not known.
-  # in this case node needs to wait for connection to be established or 
-  # more precisely for codecs to be initialized.
-  proc sendAfterCodecInitialized(): Future[void] {.async: (raises: [CancelledError]).} =
-    await peer.codecInitializedFut
```
vladopajic, line 306: `peer.codecInitializedFut` this can also be removed now

vladopajic (approved):
lgtm, send control extension logic is better-cleaner now. tests should be improved before merge.

consider avoiding `checkUntilTimeout`;  it is a smell and it's existence is because of technical depth. we should try to avoid it's usage as much as possible.
it is wasteful pattern, it never finishes right at condition - it always finishes X amount time after condition. it unnecessary interrupts execution by waking up checking condition and blocking again. it is fix for flakiness but in theory it can be source of flakiness as well (that mf is controversial). it is also not as elegant as waiting for right condition (future done, WaitGroup done etc). using future and WaitGroup has better semantic `waitOnFirstMessageFur`, `allMesagesSentWg`, etc...


`checkUntilTimeout` is last resort tool where we must keep checking for condition.

tests/libp2p/pubsub/component/test_gossipsub_extensions.nim
rlve: This is negative assertion confirming that only one control message arrives.

There is no explicit event I can wait for.

It can be either sleepAsync or a few heartbeats like that.

vladopajic: aha, got it. since this is not really depended on heartbeats better just have empirically chosen sleep https://github.com/vacp2p/nim-libp2p/blob/copilot/fix-inconsistent-start-behaviour/tests/libp2p/pubsub/component/test_gossipsub_message_handling.nim#L117

tests/libp2p/pubsub/component/test_gossipsub_extensions.nim
rlve:
```nim
    # Wait for subscriptions to propagate.
    waitSubscribeStar(nodes, topic)
```

This above ensures that control messages with peer subscriptions were received.

gmelodie (approved):

PR #2300, opened 2026-04-17
fix: raise exception rather than defect on malformed DNS response

jm-clius:
This is a suggestion, rather than necessarily a fix.

I notice that the third party DNS library will raise an `Exception` on malformed DNS responses. I can confirm that this happens from time to time (root cause unknown, but observed a few times in WSL2 environment). To me the `AssertDefect` and accompanying fatal panic seems excessive, as in most cases we can recover and simply ignore the unresolved DNS lookup.

There may be some impact though that I'm not thinking of, so feel free to LMK what you think. :)


libp2p/nameresolving/dnsresolver.nim
```diff
@@ -88,8 +88,8 @@ proc getDnsResponse(
     except Exception as exc:
       # Nim 1.6: parseResponse can has a raises: [Exception, ..] because of
       # https://github.com/nim-lang/Nim/commit/035134de429b5d99c5607c5fae912762bebb6008
-      # it can't actually raise though
-      raiseAssert "Exception parsing DN response: " & exc.msg
+      # Let's raise as an IOError
+      raise newException(IOError, "Exception parsing DNS response: " & exc.msg)
```
gmelodie, line 92: Since we're no longer supporting nim 1.6, can we remove this entirely?

jm-clius: Eh, to be honest I wasn't even paying attention to the Nim 1.6 in the comments. I compiled with Nim 2.2.4 as set by the build system and can confirm that a malformed DNS response leads to this raiseAssert.

gmelodie: `dnsclient.nim` is unmaintained, we should get rid of it altogether https://github.com/vacp2p/nim-libp2p/issues/2301. This is a problem for another day though, aproving.

gmelodie (approved):

richard-ramos (approved):

PR #2285, opened 2026-04-15
feat(service-disco): add advertiser

gmelodie:
## Summary

This PR adds the advertiser subsystem to `ServiceDiscovery`, building on the registrar and per-service routing tables introduced in the parent branch.

It wires up the `Advertiser` type (already declared in `types.nim`) into the `ServiceDiscovery` object and implements the full outbound advertisement lifecycle:
- Encoding the local peer record
- Sending `REGISTER` messages
- Handling `Confirmed` / `Wait` / `Rejected` responses with appropriate back-off
- Feeding closer peers back into the service routing table

`addProvidedService` and `removeProvidedService` are added as the primary API for managing which services a node actively advertises.

---

### `libp2p/protocols/service_discovery/advertiser.nim` (new)

- **sendRegister**
  - Dials a registrar peer, sends a `REGISTER` message, and returns:
    `(RegistrationStatus, Opt[Ticket], seq[PeerId])`

- **startAdvertising**
  - Long-running loop per registrar peer
  - Builds the signed advertisement from `peerInfo` (if not externally supplied)
  - Handles response states:
    - `Confirmed` → sleep `advertExpiry`
    - `Wait` → sleep `min(advertExpiry, tWaitFor)`
    - `Rejected` → stop
  - Feeds closer peers into the service routing table on each reply

- **addProvidedService**
  - Registers the service as `Provided` in `rtManager` (seeded from main routing table)
  - Shuffles peers in each bucket
  - Spawns up to `kRegister` `startAdvertising` tasks per non-empty bucket

- **removeProvidedService**
  - Cancels and removes all pending tasks for the service
  - Calls `rtManager.removeService`
  - Updates metrics

- **clear**
  - Bulk-cancels all running advertise tasks

---

### `libp2p/protocols/service_discovery/types.nim`

- Adds:
  - `advertiser*: Advertiser` field to `ServiceDiscovery`

---

### `libp2p/protocols/service_discovery.nim`

- Initializes `Advertiser` via `Advertiser.new()` in the `ServiceDiscovery` constructor

---

### `tests/libp2p/service_discovery/test_advertiser.nim` (new)

- 12 tests across three suites:
  - `addProvidedService` (5 async)
  - `removeProvidedService` (3 async)
  - `startAdvertising / stopAdvertising` (4 sync)

---

### `tests/libp2p/service_discovery/utils.nim`

- Adds test helpers:
  - `populateRoutingTable(disco, n)`
  - `populateAdvTable(disco, serviceId)`
[...]

PR #2306, opened 2026-04-20
feat(service-disco): add discoverer

gmelodie:
## Summary

This PR introduces the **Discoverer** component and completes the service discovery protocol by implementing the previously stubbed `REGISTER` and `GET_ADS` handlers.  

- `lookup` enables a node to query the network for providers of a given service ID by iterating its per-service routing table and issuing `GET_ADS` requests.  
- Server-side handlers (`handleRegister`, `handleGetAds`) complete the round-trip:
  - Registrars can now accept registrations with rate-limiting and signed wait tickets
  - Registrars can respond to advertisement queries from their cache

---

### libp2p/protocols/service_discovery/discoverer.nim (new)

- `sendGetAds`
  - Dials a peer
  - Sends a `GET_ADS` message
  - Decodes returned advertisements (filters mismatched service IDs)
  - Collects closer peers

- `lookup`
  - Iterates the per-service routing table bucket by bucket
  - Queries up to `kLookup` peers per bucket via `sendGetAds`
  - Seeds the routing table with returned closer peers
  - Accumulates up to `fLookup` valid advertisements

---

### libp2p/protocols/service_discovery/registrar.nim

- `handleRegister`
  - Validates the advertisement
  - Computes rate-limiting wait time
  - Adjusts wait using a valid retry ticket (if provided)
  - Responds:
    - `Confirmed` → calls `acceptAdvertisement`
    - `Wait` → returns a signed `Ticket`
    - `Rejected` → fallback on signing failure

- `handleGetAds`
  - Looks up registrar cache for the requested service ID
  - Returns up to `fReturn` advertisements
  - Attaches closer peers from the routing table

---

### libp2p/protocols/service_discovery.nim

- Added `registrar` and `discoverer` imports
- Exported `discoverer` (exposes `lookup`)
- Replaced `"Unimplemented"` stubs:
  - `MessageType.register` → `await disco.handleRegister(conn, msg)`
  - `MessageType.getAds` → `await disco.handleGetAds(conn, msg)`

---

### tests/libp2p/service_discovery/test_discoverer.nim (new)

Covers:

- Routing table creation on first call  
- Empty table returns `ok([])`  
- Idempotent repeated calls  
- Independent tables per service ID  
- Result count capped (see risk note)

---

### tests/libp2p/service_discovery/utils.nim

- Exported `routing_table_manager`
- Added helpers:
  - `populateRoutingTable`
  - `populateAdvTable`
  - `populateSearchTable`

---

## Affected Areas

- **Peer Management / Discovery**  
  Implements client-side provider lookup over per-service Kademlia routing tables.

- **Protocol Logic**  
  `REGISTER` and `GET_ADS` are now fully handled server-side (previously dropped).

- **Gossipsub**  
- **Transports**  
- **Build / Tooling**  
- **Other**  

(No direct changes in these areas.)

---

- No changes to existing KadDHT encode/decode or APIs  
- Message types `register` (16) and `getAds` (17) already existed; only dispatch was missing  
- Mixed-version behavior:
  - Older nodes silently drop messages
  - Advertisers retry on timeout → graceful degradation

No validation required for standard KadDHT consumers (Nimbus, Waku, Codex).

ServiceDiscovery users should verify that `stop()` correctly drains in-flight advertising tasks.

---

## Impact on Library Users

- **New API**
  - `lookup*(disco: ServiceDiscovery, serviceId: ServiceId): Future[Result[seq[Advertisement], string]]`

- **Behavior changes**
  - `REGISTER` is now processed (returns `Confirmed`, `Wait`, or `Rejected`)
  - `GET_ADS` is now processed (returns cached advertisements)

---

## Risk Assessment

- **Retry ticket `tInit` not bounded to wall time**  
  A manipulated past `tInit` could produce a large negative adjustment.  
  Mitigated by retry-window validation in `processRetryTicket`.  
  **Risk:** Low

- **Deterministic advertisement ordering**  
  `handleGetAds` returns ads in insertion order (no randomization).  
  May bias results under load.  
  **Risk:** Low (can be improved later)

- **Ticket signing fallback to `Rejected`**  
  If signing fails, advertiser backs off for full expiry instead of retrying sooner.  
  Safe but may introduce delays.  
  **Risk:** Low

- **Test uses `kLookup` instead of `fLookup`**  
  Result cap assertion may be weaker than intended if values differ.  
  **Risk:** Low

---

## References

- Base branch: `feat/service-disco/advertiser`
- Related: `feat/service-disco/registrar`
- #2285 
- #2055


libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -0,0 +1,127 @@
...
+      break
+
+    var bucket = searchTable.buckets[bucketIdx]
+    if bucket.peers.len == 0:
+      continue
+
+    disco.rng.shuffle(bucket.peers)
+
+    for peerId in bucket.peersToQuery(disco.discoConfig.kLookup):
+      let (ads, closerPeers) = (await sendGetAds(disco, peerId, serviceId)).valueOr:
```
richard-ramos: Should this have a timeout like with kademlia in which kad.config.timeout is used? because otherwise this might take a long time depending on the speed of the peers. 

Also, this is done sequentially. Should it be done concurrently for all peers?

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -0,0 +1,127 @@
...
+    return err("connection writing failed: " & writeRes.error.msg)
+  let replyBuf = readRes.valueOr:
+    return err("connection reading failed: " & readRes.error.msg)
+
+  cd_messages_received.inc(labelValues = [$MessageType.getAds])
+  cd_message_bytes_received.inc(
+    replyBuf.len.float64, labelValues = [$MessageType.getAds]
+  )
+
+  let reply = Message.decode(replyBuf).valueOr:
```
richard-ramos, line 79: Peerstore should be updated with the peers obtained here, right? i.e.
```nim
# for each peer
switch.peerStore[AddressBook].extend(p.peerId, addrs)
```

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -391,3 +391,96 @@ proc acceptAdvertisement*(
...
+
+  let ad = validateRegisterMessage(regMsg).valueOr:
+    await sendRegisterResponse(
+      conn, kademlia_protobuf.RegistrationStatus.Rejected, closerPeers
+    )
+    cd_register_requests.inc(
+      labelValues = [$kademlia_protobuf.RegistrationStatus.Rejected]
+    )
+    return
+
```
richard-ramos, line 424: Is it necessary to check if we advertise the service?
```
if not ad.advertisesService(serviceId):
   await sendRegisterResponse(
      conn, kademlia_protobuf.RegistrationStatus.Rejected, closerPeers
    )
    cd_register_requests.inc(
      labelValues = [$kademlia_protobuf.RegistrationStatus.Rejected]
    )
    return
```

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -391,3 +391,96 @@ proc acceptAdvertisement*(
...
+    await sendRegisterResponse(
+      conn, kademlia_protobuf.RegistrationStatus.Wait, closerPeers, Opt.some(ticket)
+    )
+    cd_register_requests.inc(labelValues = [$kademlia_protobuf.RegistrationStatus.Wait])
+
+proc handleGetAds*(
+    disco: ServiceDiscovery, conn: Connection, msg: Message
+) {.async: (raises: [CancelledError]).} =
+  let serviceId = msg.key
+  let ads = disco.registrar.cache.getOrDefault(serviceId, @[])
```
richard-ramos, line 468: I was trying to find calls to `pruneExpiredAds` unsuccesfully. Without that, the cache entries will never expire

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -0,0 +1,127 @@
...
+    return err("failed to decode message response: " & $error)
+
+  let getAdsMsg = reply.getAds.valueOr:
+    return err("get ads message response not found")
+
+  return
+    ok((getAdsMsg.advertisements.validAds(serviceId), reply.closerPeers.toPeerIds()))
+
+proc peersToQuery(bucket: Bucket, kLookup: int): seq[PeerId] =
+  var peerIds: seq[PeerId]
```
vladopajic: consider pre-allocation `peerIds`

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -0,0 +1,127 @@
...
+  let reply = Message.decode(replyBuf).valueOr:
+    return err("failed to decode message response: " & $error)
+
+  let getAdsMsg = reply.getAds.valueOr:
+    return err("get ads message response not found")
+
+  return
+    ok((getAdsMsg.advertisements.validAds(serviceId), reply.closerPeers.toPeerIds()))
+
+proc peersToQuery(bucket: Bucket, kLookup: int): seq[PeerId] =
```
vladopajic:
```
pluckFirstPeers(bucked: Bucket, count: int)  seq[PeerId]
# gets first up to `count` peers from bucket.
```

`pluckFirstPeers` -> `pluckFirstNPeers` ...

vladopajic (approved):

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -0,0 +1,127 @@
...
+
+  for bucketIdx in 0 ..< searchTable.buckets.len:
+    if found.len >= disco.discoConfig.fLookup: # done
+      break
+
+    var bucket = searchTable.buckets[bucketIdx]
+    if bucket.peers.len == 0:
+      continue
+
+    disco.rng.shuffle(bucket.peers)
```
vladopajic: it's probably not ideal to shuffle original values
- pick up to K elements at random (do we have utility for this already)
- or clone then, then shuffle

vladopajic: that's why bucket has to be var and `bucketIdx` is needed?

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -391,3 +391,96 @@ proc acceptAdvertisement*(
...
+      labelValues = [$kademlia_protobuf.RegistrationStatus.Confirmed]
+    )
+  else:
+    disco.registrar.updateLowerBounds(serviceId, ad, tWait, now)
+
+    let tInit =
+      if regMsg.ticket.isSome:
+        regMsg.ticket.unsafeGet().tInit
+      else:
+        now
```
vladopajic: it would be good if there is construct here

`let tInit  =regMsg.ticket.getOrDefault(now)`

just like tables have.
`
especially because this one liner could be used right in the same lime as `tInit` field of `Ticket` constructor bellow; making variable unnecessery

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -391,3 +391,96 @@ proc acceptAdvertisement*(
...
+    let encoded = ad.encode().valueOr:
+      error "failed to encode advertisement", error
+      continue
+    adBytes.add(encoded)
+
+  let closerPeers = disco.findClosestPeers(serviceId)
+  let response = Message(
+    msgType: MessageType.getAds,
+    getAds: Opt.some(GetAdsMessage(advertisements: adBytes)),
+    closerPeers: closerPeers,
```
vladopajic: nit `closerPeers: disco.findClosestPeers(serviceId)`

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -391,3 +391,96 @@ proc acceptAdvertisement*(
...
+  let ads = disco.registrar.cache.getOrDefault(serviceId, @[])
+
+  var adBytes: seq[seq[byte]]
+  for ad in ads:
+    if adBytes.len >= disco.discoConfig.fReturn:
+      break
+    let encoded = ad.encode().valueOr:
+      error "failed to encode advertisement", error
+      continue
+    adBytes.add(encoded)
```
vladopajic: could this be utility

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -0,0 +1,127 @@
...
+      let (ads, closerPeers) = (await sendGetAds(disco, peerId, serviceId)).valueOr:
+        error "failed to get ads", error
+        continue
+
+      for nodeId in closerPeers:
+        disco.rtManager.insertPeer(serviceId, nodeId.toKey())
+
+      let remaining = disco.discoConfig.fLookup - found.len
+      if remaining > 0:
+        found.add(ads[0 ..< min(remaining, ads.len)])
```
vladopajic: is there issue if we return more then `disco.discoConfig.fLookup`. in theory wasting information is always bad.

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -0,0 +1,127 @@
...
+      else:
+        # done
+        break
```
vladopajic: first if in next cycle should handle it, unnecessary redudant

libp2p/protocols/service_discovery/discoverer.nim
gmelodie: calling updatePeers now

libp2p/protocols/service_discovery/registrar.nim
gmelodie: The problem here is that we want `ticket.get().tInit` or `now`, not `ticket.get()` or `now`, otherwise we could just do `ticket.get(now)`. Claude came up with a working but kinda ugly solution. Open to other alternatives

libp2p/protocols/service_discovery/discoverer.nim
gmelodie: So [spec](https://github.com/logos-co/logos-lips/blob/aaf158aa59edb2ce0841a345725d16355218c196/vac/raw/logos-capability-discovery.md) says: 
> Stop early if enough advertiser peers (F_lookup) have been found — no need to continue searching.

but i guess since we already did all the hard work of querying the peers, it couldn't hurt to use the information. I wonder if this is a case of "fuck the spec" or do we implement it like this and question the spec and reimplement later if decided this is a better approach 

cc @richard-ramos

libp2p/protocols/service_discovery/registrar.nim
gmelodie: Fixed, added pruning of service tables cache wich was missing too

libp2p/protocols/service_discovery/discoverer.nim
gmelodie: used `pick` instead (god damn fish memory, i added that last week)

libp2p/protocols/service_discovery/registrar.nim
vladopajic: aha i see... then it needs to be specialized utility `regMsg.ticket.tInitOrDefault(now)`

libp2p/protocols/service_discovery/discoverer.nim
vladopajic: > no need to continue searching

algorithm is not continuing to search, it is just returning all the information.

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -0,0 +1,140 @@
...
+
+  discard disco.rtManager.addService(
+    serviceId, disco.rtable, disco.config.replication, disco.discoConfig.bucketsCount,
+    Interest,
+  )
+
+  let searchTable = disco.rtManager.getTable(serviceId).valueOr:
+    return err("service table not found for service id: " & $serviceId)
+
+  var found = newSeqOfCap[Advertisement](disco.discoConfig.fLookup)
```
vladopajic, line 183: `disco.discoConfig.fLookup`
why is this not 
`disco.config.fLookup`

libp2p/protocols/service_discovery/discoverer.nim
vladopajic: also on the same topic, request should include value "dont send me more then X ads", since values are gonna get dropped anyway. 

and also processing responses here with `collectCompleted` is inefficient, if first peer gets you everything then code is waiting on slowest peer that is never contributing to final result. which will always happen for any configuration. say `fLookup` is 5 and currently there are 4 found, next batch queries 5 more peers. and first one gets you result in 500ms while last one takes 2seconds. there is 1.5secon of waste here + all the traffic.

libp2p/protocols/service_discovery/discoverer.nim
gmelodie: `ServiceDiscovery` is of type `KadDHT` which already has a `config` field

libp2p/protocols/service_discovery/discoverer.nim
gmelodie: makes sense! please recheck

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -0,0 +1,192 @@
...
+
+    let completedFut = wrapper.value()
+    pending.del(pending.find(completedFut))
+
+    if completedFut.completed():
+      completedFut.value().withValue(response):
+        disco.processResponse(serviceId, response, found, limit)
+
+    if found.len >= limit:
+      disco.drainCompletedPeers(serviceId, pending)
```
vladopajic, line 164: i would do this in defer, since this is not really part of main logic here, but rather cleanup

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -0,0 +1,192 @@
...
+    disco: ServiceDiscovery, serviceId: ServiceId, peers: seq[PeerId], limit: int
+): Future[seq[Advertisement]] {.async: (raises: [CancelledError]).} =
+  var found = newSeqOfCap[Advertisement](limit)
+  var pending: seq[Future[Opt[GetAdsResult]]] =
+    peers.mapIt(Future[Opt[GetAdsResult]](dispatchGetAds(disco, it, serviceId)))
+  defer:
+    for fut in pending:
+      if not fut.finished():
+        fut.cancelSoon()
+
```
vladopajic, line 145: `pending.cancelSoon()`

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -0,0 +1,192 @@
...
+  for fut in pending.filterIt(it.completed()):
+    fut.value().withValue(response):
+      disco.insertCloserPeers(serviceId, response.closerPeers)
+
+proc collectBucketAds(
+    disco: ServiceDiscovery, serviceId: ServiceId, peers: seq[PeerId], limit: int
+): Future[seq[Advertisement]] {.async: (raises: [CancelledError]).} =
+  var found = newSeqOfCap[Advertisement](limit)
+  var pending: seq[Future[Opt[GetAdsResult]]] =
+    peers.mapIt(Future[Opt[GetAdsResult]](dispatchGetAds(disco, it, serviceId)))
```
vladopajic: `dispatchGetAds(disco, it, serviceId, limit)`

dispatch limit in request so that one serving request does not add more then necessary?

libp2p/protocols/service_discovery/discoverer.nim
vladopajic: still, should be doable. `KadDHT` will have type param `C`  that represents config. then `ServiceDiscovery`  extends `KadDHT` with config type `DiscoveryConfig` that extends `KadConfig`.

but talking out from top of my head 1) not sure how this will work in nim 2) not sure how these would stack if there is really senergy or not

tests/libp2p/service_discovery/test_discoverer.nim
```diff
@@ -0,0 +1,74 @@
...
+
+    check disco.rtManager.hasService(sid1)
+    check disco.rtManager.hasService(sid2)
+    check disco.rtManager.count() == 2
+
+  asyncTest "kRegister cap: result length never exceeds kRegister":
+    let kRegister = 5
+    let disco = makeMockDiscovery(
+      discoConfig = ServiceDiscoveryConfig.new(kRegister = kRegister, bucketsCount = 16)
+    )
```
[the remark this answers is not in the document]
vladopajic: should also have ` check res.get().len > 0`

libp2p/protocols/service_discovery/types.nim
```diff
@@ -116,6 +118,16 @@ proc hash*(t: AdvertiseTask): Hash =
 proc toAdvertisementKey*(ad: Advertisement): AdvertisementKey {.raises: [].} =
   (peerId: ad.data.peerId, seqNo: ad.data.seqNo)
 
+proc encode*(ads: seq[Advertisement], fReturn: int): seq[seq[byte]] {.raises: [].} =
```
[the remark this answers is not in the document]
vladopajic: legit

how does tests pass with this kind of error?

richard-ramos (approved):

PR #2310, opened 2026-04-20
test(gossipsub): partial messages with fanout bug + fix

rlve:
## Summary

The partial message extension had a bug in `publishPartial` that prevented fanout publishers from working correctly. A fanout publisher is a node that publishes partial messages to a topic it is not subscribed to. Two things were broken:

1. When picking publish targets (no explicit peer list), the function only used the default `publishToPeers` result. It missed peers that already exchanged metadata for the group but never sent a subscription RPC. This broke the reply path when a remote peer is itself a fanout publisher.

2. When deciding whether to send to a given peer, the function required `nodeSubOpt.supportsSendingPartial OR nodeSubOpt.requestsPartial`. That condition always failed for an unsubscribed node (fanout publisher), so no messages went out.

**Fixes:**
- Broadcast publish now extends the default peer list with any peer that has per-group state in `groupState.peerState`.
- When a peer requests partials, the node sends to it regardless of its own subscription opts.
- When this node requests partials, the check now accepts `peerHasState` (peer already sent group metadata) as a valid signal alongside `peerSubOpt.supportsSendingPartial`.

## Affected Areas

- [x] Gossipsub
  - `extension_partial_message.nim`: publish routing in `publishPartial`

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

Behavior change in the partial message extension. Nodes calling `publishPartial` on a topic they are not subscribed to now work correctly. There is no API change.

## Risk Assessment

Low. The change relaxes publish conditions so it is strictly more permissive. The fanout path was previously broken, so there is no regression risk against working behavior. Three levels of tests cover the new scenarios: unit (extension logic), component (full gossipsub node), and interop (two-node script runner).

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
+++ b/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 # Copyright (c) Status Research & Development GmbH
 
-import tables, chronicles, results
+import tables, sets, chronicles, results
 import ../../../utils/tablekey
 import ../../../[peerid]
 import ../rpc/messages
@@ -331,10 +331,13 @@ proc publishPartial*(
       # with peers that already exchanged metadata for this group.
       # This preserves replies to a remote peer acting as an unsubscribed
       # fanout publisher.
+      var seen = publishTargets.toHashSet()
       for p in groupState.peerState.keys:
-        if p notin publishTargets:
+        if not seen.containsOrIncl(p):
           publishTargets.add(p)
       publishTargets
+
+  let nodeRequestsPartial = ext.config.nodeTopicOpts(topic).requestsPartial
 
   var publishedToCount: int = 0
   for p in publishToPeers:
@@ -354,7 +357,6 @@ proc publishPartial*(
     # advertised partial support for the topic or already have per-group
     # state. peerHasState covers peers that already sent metadata for this
     # group without ever sending a subscription RPC.
-    let nodeRequestsPartial = ext.config.nodeTopicOpts(topic).requestsPartial
     let peerHasState = groupState.peerState.hasKey(p)
     if nodeRequestsPartial and (peerSubOpt.supportsSendingPartial or peerHasState):
       if ext.publishPartialToPeer(topic, pm, groupState, p, false):
--- a/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
+++ b/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
@@ -322,7 +322,7 @@ proc publishPartial*(
   groupState.heartbeatsTillEviction = ext.config.heartbeatsTillEviction
   groupState.lastPublishedMetadata = pm.partsMetadata()
 
-  var peersToTry =
+  var publishToPeers =
     if peers.len > 0:
       peers
     else:
@@ -337,7 +337,7 @@ proc publishPartial*(
       publishTargets
 
   var publishedToCount: int = 0
-  for p in peersToTry:
+  for p in publishToPeers:
     if not ext.config.isSupported(p):
       continue
 
--- a/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
+++ b/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 # Copyright (c) Status Research & Development GmbH
 
-import tables, sets, chronicles, results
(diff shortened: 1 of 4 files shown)


libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -322,29 +322,43 @@ proc publishPartial*(
...
+      var publishTargets = ext.config.publishToPeers(topic)
+      # Extend this node's current publish targets
+      # with peers that already exchanged metadata for this group.
+      # This preserves replies to a remote peer acting as an unsubscribed
+      # fanout publisher.
+      var seen = publishTargets.toHashSet()
+      for p in groupState.peerState.keys:
+        if not seen.containsOrIncl(p):
+          publishTargets.add(p)
+      publishTargets
```
vladopajic:
```suggestion
      var peers = ext.config.publishToPeers(topic).toHashSet()
      # Extend this node's current publish targets
      # with peers that already exchanged metadata for this group.
      # This preserves replies to a remote peer acting as an unsubscribed
      # fanout publisher.
      for p in groupState.peerState.keys:
        peers.incl(p)
      peers.toSeq()
```

just hashset would be sufficient here?

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -322,29 +322,43 @@ proc publishPartial*(
...
+      continue
+
+    # If this node requests partials, send metadata to peers that either
+    # advertised partial support for the topic or already have per-group
+    # state. peerHasState covers peers that already sent metadata for this
+    # group without ever sending a subscription RPC.
+    let peerHasState = groupState.peerState.hasKey(p)
+    if nodeRequestsPartial and (peerSubOpt.supportsSendingPartial or peerHasState):
       if ext.publishPartialToPeer(topic, pm, groupState, p, false):
         publishedToCount.inc
```
vladopajic, line 364: i see... but should condition be
```
if nodeRequestsPartial and (peerSubOpt.supportsSendingPartial or peerSubOpt.requestsPartial or peerHasState):
```
in other words i understand additional `or peerHasState`; but why was ` or peerSubOpt.requestsPartial` dropped from condition?

----
nit:

`if nodeRequestsPartial and (peerSubOpt.supportsSendingPartial or groupState.hasPeer(p))`

avoids map search unless really necessary for condition

tests/interop/gossipsub/test_runner.nim
```diff
@@ -194,3 +194,70 @@ suite "GossipSub Interop - Script runner - Component":
...
+      ScriptInstruction(kind: WaitUntil, elapsed: 6.seconds),
+    ]
+
+    await allFutures(@[runner0.runScript(node0Script), runner1.runScript(node1Script)])
+
+    checkUntilTimeout:
+      key in runner0.messages
+      key in runner1.messages
+      runner0.messages[key].isComplete()
+      runner1.messages[key].isComplete()
```
vladopajic, line 261: should not part of this pr. but later we could improve this is `AsyncQueue`:

```nim
let interopMsg = await runner0.messageQueue

check:
  interopMsg.key == key
  interopMsg.msg == {{actual message}}

```

tests/libp2p/pubsub/component/test_gossipsub_extensions.nim
```diff
@@ -362,6 +362,72 @@ suite "GossipSub Component - Extensions":
...
+    const topic = "logos-partial"
+    const groupId = "group-id-1".toBytes
+
+    proc validateRPC(
+        rpc: PartialMessageExtensionRPC
+    ): Result[void, string] {.gcsafe, raises: [].} =
+      checkLen(rpc.partsMetadata)
+      return ok()
+
+    var incomingRPC: Table[PeerId, seq[PartialMessageExtensionRPC]]
```
vladopajic: probably old code has the same smell, but we should brake the old patterns.

```
 let incomingRPC: Table[PeerId, AsyncQueue[PartialMessageExtensionRPC]]
```

vladopajic (approved):
lgtm

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: > dropped from condition

thb i forgot what the code was, if `supportsSendingPartial` was set to `supportsSendingPartial or requestsPartial` then it's okay.

rlve: The new branch right above (`if peerSubOpt.requestsPartial:`) catches any peer that requests partials and does `continue`, so by the time we hit this condition `peerSubOpt.requestsPartial` is always false.

richard-ramos (approved):

PR #2314, opened 2026-04-21
feat(cbind): register custom protocol handlers

richard-ramos:
## Summary
Add C binding support for mounting custom libp2p protocols and handling incoming protocol streams from C.

This PR introduces `libp2p_mount_protocol`, which lets C callers register a protocol ID and a callback handler. The handler receives the incoming stream handle and can interact with it through the existing `libp2p_stream_*` APIs.

It also adds a C echo protocol example that demonstrates the intended callback flow:

`mount protocol -> dial protocol -> read LP frame -> write LP frame -> closeWithEOF -> release`

This is needed because the C bindings previously supported dialing and consuming protocols as clients, but did not expose a way for C users to define server-side protocol handlers.

## Affected Areas
- [ ] Gossipsub  
 - [ ] Transports  
- [ ] Protocol Logic
- [x] Build / Tooling
- [x] Other
  Updates the public C binding API and FFI request dispatch path.
  Adds C binding support for mounting custom `LPProtocol` handlers and managing incoming stream lifetime from C.
  Updates the cbind examples task to build and run the new echo example.

- **Nimbus:**  
  Not tested.

- **Waku:**  
  Not tested.

- **Codex:**  
  Not tested.

Local validation:
- Built the cbind static library successfully.
- Built and ran the new `cbind/examples/echo.c` example successfully.
- Built and ran the existing `cbindings` and `mix` C examples against the updated static library.

## Impact on Library Users
This adds a new C binding API. No impact on logos-core module. Also, this likely

```c
typedef void (*Libp2pProtocolHandler)(
    libp2p_ctx_t *ctx,
    libp2p_stream_t *stream,
    const char *proto,
    size_t protoLen,
    void *userData);

int libp2p_mount_protocol(libp2p_ctx_t *ctx,
                          const char *proto,
                          Libp2pProtocolHandler handler,
                          Libp2pCallback callback,
                          void *userData);
```

C users can now mount custom protocols and handle inbound streams.

Handlers are callback-based and must not block the libp2p worker thread. A handler should schedule stream operations using `libp2p_stream_read*` / `libp2p_stream_write*`, then eventually call `libp2p_stream_release` when it is done with the incoming stream.

Existing C binding APIs remain compatible.

## Risk Assessment
Limited to the cbind layer.

## Potential risks:

Incorrect stream release behavior in C handlers can keep incoming protocol handler futures alive longer than intended.
C protocol handlers run on the libp2p worker thread, so blocking in user callbacks can still stall that node.
The direct-dispatch path for re-entrant worker-thread requests changes request scheduling behavior for callbacks invoked from the worker thread, but avoids deadlocks when C handlers call `libp2p_stream_*`.
No wire protocol changes are introduced.


cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
```diff
@@ -302,6 +302,14 @@ proc processRelease*(
   if not libp2p[].connections.hasKey(handle):
     return err("unknown connection handle")
 
+  # For incoming custom-protocol streams, release completes the Nim protocol
+  # handler that has been waiting for C to finish its callback chain.
+  let releaseWaiter = libp2p[].streamReleaseWaiters.getOrDefault(handle, nil)
+  if not releaseWaiter.isNil():
```
gmelodie:
```suggestion
  libp2p[].streamReleaseWaiters.handle.withValue(releaseWaiter):
```

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
```diff
@@ -302,6 +302,14 @@ proc processRelease*(
...
+    libp2p[].streamReleaseWaiters.del(handle)
+    if not releaseWaiter.finished:
+      releaseWaiter.complete()
```
gmelodie, line 315: i think you can just
```suggestion
    releaseWaiter.complete()
```

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_protocol_requests.nim
```diff
@@ -0,0 +1,104 @@
...
+
+  libp2p[].customProtocols[proto] = mountedProtocol
+  ok()
+
+proc process*(
+    self: ptr ProtocolRequest, libp2p: ptr LibP2P
+): Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
+  case self[].operation
+  of ProtocolMsgType.MOUNT:
+    await self.processMount(libp2p)
```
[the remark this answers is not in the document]
richard-ramos: It will return the result of that `await`.

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_protocol_requests.nim
```diff
@@ -0,0 +1,104 @@
...
+    return err(exc.msg)
+
+  libp2p[].customProtocols[proto] = mountedProtocol
+  ok()
+
+proc process*(
+    self: ptr ProtocolRequest, libp2p: ptr LibP2P
+): Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
+  case self[].operation
+  of ProtocolMsgType.MOUNT:
```
gmelodie: `case` for only 1 case?

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_protocol_requests.nim
```diff
@@ -0,0 +1,104 @@
...
+        if selectedProto.len > 0:
+          cast[ptr cchar](unsafeAddr selectedProto[0])
+        else:
+          cast[ptr cchar](nil)
+      handler(ctx, stream, protoPtr, cast[csize_t](selectedProto.len), protocolUserData)
+
+    await releaseWaiter
+
+  let mountedProtocol = LPProtocol.new(codecs = @[proto], handler = handle)
+  await mountedProtocol.start()
```
gmelodie, line 98: I thought `switch.start` was supposed to `start` every mounted handler. why do we have to start it here instead?

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_protocol_requests.nim
```diff
@@ -0,0 +1,104 @@
...
+  if self[].handler.isNil():
+    return err("handler is nil")
+  if libp2p.isNil() or libp2p[].switch.isNil():
+    return err("libp2p switch is not initialized")
+
+  let proto = $self[].proto
+  let peerInfo = libp2p[].switch.peerInfo
+  let mountedInSwitch = not peerInfo.isNil() and proto in peerInfo.protocols
+  if libp2p[].customProtocols.hasKey(proto) or mountedInSwitch:
+    return err("protocol already mounted: " & proto)
```
gmelodie: I wonder in which case would `peerInfo` be `nil`. If that actually did happen I'd expect the error message to be `peerInfo is nil`
```suggestion
  if peerInfo.isNil():
    return err("unexpected: peerInfo in switch is nil")
  if libp2p[].customProtocols.hasKey(proto) or proto in peerInfo.protocols:
    return err("protocol already mounted: " & proto)
```

cbind/libp2p_thread/libp2p_thread.nim
```diff
@@ -114,6 +121,20 @@ proc destroyLibP2PThread*(ctx: ptr LibP2PContext): Result[void, string] =
 proc sendRequestInternal(
     ctx: ptr LibP2PContext, req: ptr LibP2PThreadRequest
 ): Result[void, string] =
+  # Normal C callers enqueue requests from another thread and wait until the
+  # worker receives them. Custom protocol handlers are different: they are
+  # invoked by libp2p on the worker thread, and are expected to call
+  # libp2p_stream_* from their callbacks. If we used the queue path here, the
```
gmelodie, line 127: Makes sense, since the handlers for the custom protocols come from the C thread, not the nim/chronos thread.

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_protocol_requests.nim
richard-ramos: was gonna add UNMOUNT in subsequent PR. Will remove it here to only introduce stuff as we need on PRs

gmelodie: It's not incorrect though, so I think you can leave it

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_stream_requests.nim
richard-ramos: Sadly if you call `complete()` on an already completed future, you will end up with a `An attempt was made to complete a Future more than once.` error

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_protocol_requests.nim
richard-ramos: it will start it, but this is for these cases in which the protocol is created and mounted after the switch has started:
```
libp2p_start(...);
libp2p_mount_protocol(...);
```

gmelodie (approved):

vladopajic (approved):

PR #2315, opened 2026-04-21
chore: add local testing addr policy

SionoiS:
I added a new policy that I needed for testing discovery and refactored a bit how policies are handled.


libp2p/wire.nim
```diff
@@ -170,3 +170,25 @@ proc isFilterablePrivateMA*(ma: MultiAddress): bool =
...
+  let hostIP = initTAddress(ma).valueOr:
+    return true
+
+  if hostIP.isGlobal():
+    return false
+
+  if hostIP.isLoopback():
+    return false
+
+  return true
```
gmelodie: I think inverting the logic will make it more straightforward. Also removing the loopback part from here to make the utility more reusable, checking for loopback addresses can be done outside this proc

```suggestion
proc isGloballyRoutable*(ma: MultiAddress): bool =
  ## Returns true if this address globally routable.
  ## Circuit relay addresses are always considered globally routable
  ## even if the relay itself has a private IP
  ## since the relay address may still provide connectivity.
  
  if ma.contains(multiCodec("p2p-circuit")).get(false):
    return true

  if DNS.matchPartial(ma):
    return true

  let hostIP = initTAddress(ma).valueOr:
    return false

  if hostIP.isGlobal():
    return true

  return false
```

SionoiS: The goal was to create a duplicate of https://github.com/vacp2p/nim-libp2p/blob/e1bbda4f666ae74e5e3909767e83508157c2b97b/libp2p/wire.nim#L165 except that this one allow loopback addrs.

> checking for loopback addresses can be done outside this proc

How and where?

gmelodie: In that case maybe we can just skip this function entirely and just do

```nim
const localNoPrivateAddressPolicy* = proc(
    ma: MultiAddress
): bool {.gcsafe, raises: [].} =
  isFilterablePrivateMA(ma) and not isLoopback(ma)
```
(and only introduce `isLoopback`)

Or some other variation of this. So the overall comment here is
1. avoid double negatives (`not isNot...`, `if not... return false`)
2. make utilities as composable/reusable/tiny as possible

SionoiS: > In that case maybe we can just skip this function entirely and just do
> 
> ```nim
> const localNoPrivateAddressPolicy* = proc(
>     ma: MultiAddress
> ): bool {.gcsafe, raises: [].} =
>   isFilterablePrivateMA(ma) and not isLoopback(ma)
> ```

That would call `initTAddress(ma)` twice is that ok?


> 1. avoid double negatives (`not isNot...`, `if not... return false`)
> 2. make utilities as composable/reusable/tiny as possible

Yes, I agree.

SionoiS:
I simplified the composition. Almost perfect but we still call `initTAddress()` multiple times. :shrug:

gmelodie (approved):

gmelodie (approved):
oh yeah!

libp2p/wire.nim
```diff
@@ -160,13 +160,15 @@ proc isPublicMA*(ma: MultiAddress): bool =
...
   return hostIP.isGlobal()
 
-proc isFilterablePrivateMA*(ma: MultiAddress): bool =
-  ## Returns true if this address should be filtered out because it is private
-  ## or not globally routable. Circuit relay addresses are never filtered even
-  ## if the relay itself has a private IP, since the relay address may still
-  ## provide connectivity.
+proc isCircuitRelayMA*(ma: MultiAddress): bool =
   if ma.contains(multiCodec("p2p-circuit")).get(false):
+    return true
```
vladopajic, line 168:
```suggestion
  return ma.contains(multiCodec("p2p-circuit")).get(false):
```

drop `if`...

vladopajic (approved):

PR #2320, opened 2026-04-22
chore(service-disco): various fixes

SionoiS:
Various fixes and missing features from the POC.

@gmelodie

PR #2321, opened 2026-04-22
feat(service-disco): add missing API functions

gmelodie:
## Summary

This PR completes the service discovery API by adding:

- `startDiscovering`
- `stopDiscovering`
- `startAdvertising`
- `stopAdvertising`
- A `ServiceInfo`-based `lookup` overload
---

### Peer Management / Discovery

- `discoverer.nim`
  - Added:
    - `startDiscovering`
    - `stopDiscovering`
    - `lookup(ServiceInfo)`
  - These are symmetric to the advertiser API:
    - `addProvidedService`
    - `removeProvidedService`

---

Reference integrations:

- **Nimbus:** N/A — additive API only, no breaking changes
- **Waku:** N/A
- **Codex:** N/A

---

### Service Discovery (New API)

- `startDiscovering(ServiceInfo) → bool`
- `stopDiscovering(ServiceInfo) → bool`
- `lookup(ServiceInfo)`

All changes are additive. No existing APIs were modified.

- No impact on users not using service discovery.

---

## Risk Assessment

- Service discovery:
  - Purely additive
  - No wire protocol changes

---

## References
#2320 

---

## Additional Notes
N/A


libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -207,3 +207,24 @@ proc removeServiceInterest*(
...
+  let status = disco.rtManager.serviceStatus.getOrDefault(serviceId)
+  if status in {Interest, Both}:
+    disco.rtManager.removeService(serviceId, Interest)
+    return false
+  return true
+
+proc lookup*(
+    disco: ServiceDiscovery, service: ServiceInfo
+): Future[Result[seq[Advertisement], string]] {.async: (raises: [CancelledError]).} =
+  return await disco.lookup(service.id.hashServiceId())
```
SionoiS: Would it not be better to place those close to the Logos module logic (IDK where that is...)?

Also, stop and `startDiscovering` are the same as `addServiceInterest`

libp2p/protocols/service_discovery/advertiser.nim
richard-ramos, line 126: With changing `continue` to `return` the advertising loop will stop advertising this task if there's a failure. If this is intentional, should there be any cleanup associated to this advertisement?

libp2p/protocols/service_discovery/random_find.nim
richard-ramos, line 14:
```suggestion
proc randomRecords*(
```
test_find_random.nim will not find the proc otherwise

libp2p/protocols/service_discovery.nim
richard-ramos: `selfSignedPeerRecordLoop` is created conditionally on line 141, so no guarantees it will contain a future in this point
```suggestion
  if not disco.selfSignedPeerRecordLoop.isNil:
    disco.selfSignedPeerRecordLoop.cancelSoon()
    disco.selfSignedPeerRecordLoop = nil
```

libp2p/protocols/service_discovery/discoverer.nim
richard-ramos, line 82: Why delete this? addresses wont get stored in peer store without this

interop/gossipsub/src/logger.nim
richard-ramos, line 38: This seems unrelated to this PR

interop/gossipsub/src/node.nim
richard-ramos, line 73: Seems unrelated to this PR

richard-ramos (commented):
Hm... i just realized there might be a rebase issue here? I see changes that seem unrelated to the PR

libp2p/protocols/service_discovery/advertiser.nim
gmelodie: Introduced in #2320

libp2p/protocols/service_discovery/discoverer.nim
gmelodie: > Would it not be better to place those close to the Logos module logic 

My idea was to have them here and then expose them as `cbindings` and use the bindings in the logos libp2p core module. This is the pipeline we've been using.

> Also, stop and startDiscovering are the same as addServiceInterest

What do you mean? The point here is to have an API exactly like described in the RFC for simplicity of usage

SionoiS: > What do you mean? The point here is to have an API exactly like described in the RFC for simplicity of usage

Technically the API is for the Logos discovery module not libp2p but it doesn't really matter. If you prefer this way, we good.

vladopajic (approved):

richard-ramos (approved):

PR #2326, opened 2026-04-23
fix(service-disco): registrar wait formula ceiling

SionoiS:
For service discovery, I added a `ceil()` to the registrar waiting time. The effect is that now the value is quantized.

### DIFF
--- a/libp2p/protocols/service_discovery/registrar.nim
+++ b/libp2p/protocols/service_discovery/registrar.nim
@@ -146,7 +146,7 @@ proc waitingTime*(
     advertCacheCap: uint64,
     serviceId: ServiceId,
     now: uint64,
-): float64 =
+): chronos.Duration =
   doAssert advertCacheCap > 0, "advertCacheCap must be > 0"
   let c = registrar.cacheTimestamps.len.uint64
   let c_s = registrar.cache.getOrDefault(serviceId, @[]).len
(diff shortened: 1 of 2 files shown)


libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -187,7 +187,12 @@ proc waitingTime*(
       if ipLowerBound > w:
         w = ipLowerBound
 
-  return secsAsDuration(max(0.0, w))
+  # Bound & Quantize W
+  w = max(0.0, w)
+  w = min(w, float64(uint32.high))
+  w = ceil(w)
+
+  return w
```
gmelodie: Why is quantization desired here exactly? Also, you can quantize it and still use `chronos.Duration`s, which is what we prefer when dealing with types that represent time.
```suggestion
  # Bound & Quantize W
  w = max(0.0, w)
  w = min(w, float64(uint32.high))
  w = ceil(w)

   return secsAsDuration(w)
```

gmelodie (changes requested):

libp2p/protocols/service_discovery/registrar.nim
SionoiS: > Why is quantization desired here exactly?

To simplify things a bit. We only use integer when serializing anyway. In testing we noticed truncation of floats resulting in 0.0 wait time.

> Also, you can quantize it and still use `chronos.Duration`s, which is what we prefer when dealing with types that represent time.

True, I may have been overeager... I'll change it back.

gmelodie (approved):

tinniaru3005 (approved):
Looks good to me

vladopajic (approved):

jm-clius (approved):
Thanks!

PR #2329, opened 2026-04-24
chore(connmanager): consts tidy

vladopajic:
- `MaxConnections` should be called `DefaultMaxConnections` because this const tells what should be value of `maxConnections` when value is not specified.
  - similar for `MaxConnectionsPerPeer`
- avoided exporting these consts. these consts should be internal properties of `ConnManager`. any use of `ConnManager` should specify value if that value should be respected, or should leave these values unspecified in which case `ConnManager` uses defaults. therefore use of these consts outside of `ConnManager.new` is bad practice.


richard-ramos (approved):

gmelodie (approved):
Is there a benefit of using `-1` instead of `Opt`?

vladopajic:
> Is there a benefit of using `-1` instead of `Opt`?

in this case, other parts of code was already using `-1`, here:
https://github.com/vacp2p/nim-libp2p/pull/2329/changes#diff-f40cd5fb8dc2ec79f888364dc76713a731f959d7e9309628162dcb452a9801ceR599-R600
and in other places...

so i don't really wanted to touch that now because it is not really necessary for const tidy goal.  

therefore `-1`. but `Opt` might be slightly more nicer in general.

vladopajic:
> > Is there a benefit of using `-1` instead of `Opt`?
> 
> in this case, other parts of code was already using `-1`, here: https://github.com/vacp2p/nim-libp2p/pull/2329/changes#diff-f40cd5fb8dc2ec79f888364dc76713a731f959d7e9309628162dcb452a9801ceR599-R600 and in other places...
> 
> so i don't really wanted to touch that now because it is not really necessary for const tidy goal.
> 
> therefore `-1`. but `Opt` might be slightly more nicer in general.

`-1` was already choosen as unspecified value, this pr didn't change that. just avoided using default value const in favor of unspecified value because this was real intention of that code

PR #2334, opened 2026-04-24
chore(service-disco): refactor registrar time types

SionoiS:
## Summary

This PR refactors the internal time representation used by the service-discovery registrar. All timestamps and durations that were previously tracked as raw `uint64` Unix epoch seconds or `float64` seconds are replaced with `chronos.Moment` and `chronos.Duration` throughout.

Lower bounds are now stored as absolute future `Moment` deadlines (`now + waitDuration`) rather than relative `float64` offsets. This eliminates a class of drift bugs and removes the need for the `safeSub` saturation helper.

As a side-effect, the three numeric parameters of `ServiceDiscoveryConfig` (`advertExpiry`, `occupancyExp`, `registrationWindow`) are changed from `chronos.Duration` to `float64` seconds to decouple the algorithm's scalar arithmetic from Chronos types. A single `toChronos(float64): Duration` conversion helper is introduced in `types.nim`.

---

### `libp2p/protocols/service_discovery/types.nim`
- Changed `ServiceDiscoveryConfig` fields `advertExpiry`, `occupancyExp`, and `registrationWindow` from `chronos.Duration` to `float64` (seconds); updated defaults accordingly.
- Changed registrar bound/timestamp table value types from `float64` / `uint64` to `Moment`.
- Added `toChronos(secs: float64): Duration` conversion helper (nanosecond precision).

### `libp2p/protocols/service_discovery/registrar.nim`
- Replaced all `uint64` Unix timestamps with `Moment` and all `float64` wait-seconds with `Duration`.
- Stored lower bounds as absolute `Moment` deadlines instead of relative residual floats; updated `waitingTime` and `updateLowerBounds` accordingly.
- Moved the "bound & quantize W" block before IP/service lower-bound comparisons to keep the logic consistent.
- Removed `safeSub`, `secsAsDuration`, and the caller-supplied `now: uint64` parameter from `processRetryTicket` (ticket window is now checked against wall time internally).
- Fixed `findOldestKey` to handle an empty table without relying on `high(uint64)` sentinel.

### `libp2p/protocols/service_discovery/advertiser.nim`
- Updated `sleepAsync` calls to convert `float64` expiry via `toChronos`.
- Simplified `waitSecs` min-expression by removing the spurious `.seconds.uint32` intermediate.

### `libp2p/protocols/service_discovery.nim`
- Updated heartbeat interval and `pruneExpiredAds` argument to use `toChronos`.

#### `tests/libp2p/service_discovery/test_registrar.nim`
- Replaced all `uint64` / `float64` literal timestamps with `initMoment(secs)` and `seconds(n)`.
- Updated assertions to match the new absolute-bound semantics.

#### `tests/libp2p/service_discovery/test_components.nim`
- Updated `processRetryTicket` call sites to drop the `now` argument.
- Updated `tWait` to use `Duration`.
- Updated ticket fixtures to use real wall-time offsets so window checks pass correctly.

#### `tests/libp2p/service_discovery/utils.nim`
- Changed `fillCache` signature from `uint64` to `Moment`.
- Removed full `std/times` import in favour of selective `from std/times import`.

---

## Affected Areas

- Gossipsub  
- Transports  
- Peer Management / Discovery  
  - Core service-discovery registrar and advertiser time-tracking logic  
- Protocol Logic  
  - Registration wait-time computation and retry-ticket window validation  
- Build / Tooling  
- Other  

---

- The on-wire protobuf format is unchanged:
  - Ticket fields (`tInit`, `tMod`, `tWaitFor`) remain `uint64` Unix seconds.
  - Conversion to/from `Moment` / `Duration` happens at the boundary in `handleRegister` and `processRetryTicket`.
  - Existing encode/decode paths are unaffected.

- **Breaking change**:
  - `ServiceDiscoveryConfig` constructor API:
    - `advertExpiry`, `occupancyExp`, `registrationWindow` now require `float64` seconds instead of `chronos.Duration`.

- Downstream consumers (Nimbus, Waku, Codex) using non-default values must update call sites accordingly.

---

### Breaking Changes
- `ServiceDiscoveryConfig.new()`
  - Parameters `advertExpiry`, `occupancyExp`, `registrationWindow` changed from `chronos.Duration` to `float64`.
  - Calls like `chronos.seconds(N)` must become `N.float64`.

- `Registrar` struct fields:
  - `cacheTimestamps`, `boundService`, `timestampService`, `boundIp`, `timestampIp`
  - Types changed (`uint64` → `Moment`, `float64` → `Moment`).

- `pruneExpiredAds(registrar, expiry)`
  - `expiry` changed from `uint64` to `chronos.Duration`.

- `processRetryTicket(disco, regMsg, ad, tWait)`
  - Removed `now: uint64` parameter.
  - `tWait`: `float64` → `Duration`.
  - Return type: `float64` → `Duration`.

- `waitingTime(...)` and `updateLowerBounds(...)`
  - `now`: `uint64` → `Moment`.
  - Wait/bound parameters: `float64` → `Duration`.

### Additions
- New helper:
  - `toChronos(secs: float64): chronos.Duration` (exported from `types.nim`)

---

## Risk Assessment

- **Lower-bound semantics change (low risk)**  
  Bounds are now stored as `Moment` deadlines (`now + w`) instead of residual floats.  
  Remaining wait becomes `bound - now`, mathematically equivalent but avoids floating-point drift and removes the need for `safeSub`. Verified by tests.

- **`processRetryTicket` uses wall time internally (low risk)**  
  Previously relied on caller-supplied `now`. Now uses `getTime().toUnix()` internally.  
  Tests updated to use real wall-time offsets.

- **`findOldestKey` sentinel change (low risk)**  
  Replaced `high(uint64)` sentinel with a boolean flag since `Moment` lacks a max sentinel.  
  Behavior unchanged for non-empty tables; empty case remains safe.

- **`advertExpiry.uint32` vs `.seconds.uint32` (low risk)**  
  Since `advertExpiry` is now `float64` seconds, `.seconds` is unnecessary.  
  Semantics remain identical.

---

## References

- N/A


SionoiS:
closing, chronos lib are the preferred time types to use.

SionoiS:
I'll use chronos types instead of std/times

SionoiS:
@tinniaru3005 I made changes to algo and tests. Could you double check my work since your more knowledgeable.

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -157,68 +146,74 @@ proc waitingTime*(
...
 
   for addressInfo in ad.data.addresses:
     let ip = addressInfo.address.getIp().valueOr:
       continue
 
     let ipKey = $ip
-    let elapsedIp = float64(safeSub(now, registrar.timestampIp.getOrDefault(ipKey, 0)))
-    let boundIpVal = registrar.boundIp.getOrDefault(ipKey, 0.0)
+    let prevTimestamp = registrar.timestampService.getOrDefault(serviceId, Moment())
+    let prevBoundTimestamp = registrar.boundService.getOrDefault(serviceId, Moment())
```
tinniaru3005: Here IP-level lower bounds are updated based on the service-level state. I think we should use `timestampIp` and `boundIp`

SionoiS: ah good catch my mistake

libp2p/protocols/service_discovery/types.nim
```diff
@@ -236,3 +236,6 @@ proc toPeerInfos*(peers: seq[Peer]): seq[PeerInfo] =
     peerInfos.add(peerInfo)
 
   return peerInfos
+
+func toChronos*(secs: float64): chronos.Duration =
+  chronos.nanoseconds(int64(secs * 1_000_000_000))
```
vladopajic: `toChronos` -> `secondsAsDuration` or `secondsToDuration`

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -444,11 +440,12 @@ proc handleRegister*(
...
 
+    let nowUnix = getTime().toUnix()
     var ticket = Ticket(
       advertisement: regMsg.advertisement,
-      tInit: regMsg.ticket.tInitOrDefault(now),
-      tMod: now,
-      tWaitFor: uint32(tWait),
+      tInit: regMsg.ticket.tInitOrDefault(nowUnix.uint64),
+      tMod: nowUnix.uint64,
+      tWaitFor: tWait.seconds.uint32,
```
vladopajic: why these fields are not `Duration` or `Moment`?

libp2p/protocols/service_discovery/types.nim
```diff
@@ -66,11 +66,11 @@ type
     kLookup*: int
     fLookup*: int
     fReturn*: int
-    advertExpiry*: chronos.Duration
+    advertExpiry*: float64
```
vladopajic: why?

vladopajic: > I'll use chronos types instead of std/times

???

libp2p/protocols/service_discovery/registrar.nim
gmelodie: I think it's because it'll be sent over the wire (i.e. [the holy spec](https://github.com/logos-co/logos-lips/blob/aaf158aa59edb2ce0841a345725d16355218c196/vac/raw/logos-capability-discovery.md#ticket))

libp2p/protocols/service_discovery/types.nim
gmelodie: I think it's all a balance of converting between float/int and Duration/Moment. In this particular case, I agree that it should be `chronos.Duration`, but in `Ticket` above, it should probably be float/int since tickets are supposed to be encoded/signed/sent over the wire.

gmelodie (approved):
Thanks for doing this! One thing I'm not 100% sure about is in files like `test_components.nim` when you do `from std/times import getTime, toUnix` and then use `Moment` and `Duration`, if that'll be `times.Duration` (we don't want that) or `chronos.Duration`.

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -157,68 +146,74 @@ proc waitingTime*(
...
-    let boundServiceVal = registrar.boundService.getOrDefault(serviceId, 0.0)
-    let serviceLowerBound = boundServiceVal - elapsedService.float64
-    if serviceLowerBound > w:
-      w = serviceLowerBound
+    let elapsedDuration =
+      now - registrar.timestampService.getOrDefault(serviceId, Moment())
+    let prevBoundTimestamp = registrar.boundService.getOrDefault(serviceId, Moment())
+    let lowerBound = prevBoundTimestamp - elapsedDuration
+    if lowerBound > now and waitDuration < lowerBound - now:
+      waitDuration = lowerBound - now
```
tinniaru3005: I think waitingTime subtracts elapsed twice. `boundService` / `boundIp` now store absolute timestamps as `now + waitDuration`, but waitingTime does:
```let lowerBound = prevBoundTimestamp - elapsedDuration```
Then compares lowerBound - now. If the previous bound is already an absolute time, it should probably just use:
```let remaining = prevBoundTimestamp - now```

i think we are considering:
timestampService = t₀ (when bound was set)
boundService = t₀ + w₀ (absolute timestamp)

prevBoundTimestamp = t₀ + w₀
elapsedDuration = now - t₀

then:
lowerBound = prevBoundTimestamp - elapsedDuration
           = (t₀ + w₀) - (now - t₀)
           = t₀ + w₀ - now + t₀
           = 2t₀ + w₀ - now

but it should be:
remaining = (t₀ + w₀) - now

i think changing this part to:
```let prevBoundTimestamp = registrar.boundService.getOrDefault(serviceId, Moment())

if prevBoundTimestamp > now and waitDuration < prevBoundTimestamp - now:
  waitDuration = prevBoundTimestamp - now
```

Do you think this reasoning makes sense, or am I missing something here?

libp2p/protocols/service_discovery/registrar.nim
SionoiS: we use integers on the wire

libp2p/protocols/service_discovery/types.nim
SionoiS: For a config file this is simpler to handle. We convert internally and don't expose chronos types to the outside.

I can change it back if ya want.

tests/libp2p/service_discovery/test_registrar.nim
```diff
@@ -157,56 +158,56 @@ suite "Service Discovery Registrar - Lower Bound Enforcement":
...
     # bound = 1500, timestamp = 1000 → effective = 1500 - 0 = 1500
-    registrar.boundService[serviceId] = 1500.0
-    registrar.timestampService[serviceId] = 1000
+    registrar.boundService[serviceId] = initMoment(1500)
+    registrar.timestampService[serviceId] = initMoment(1000)
 
     let w = registrar.waitingTime(discoConfig, ad, 1000, serviceId, now)
 
-    check abs(w.inFloatSecs - 1500.0) < 0.001
+    check w >= seconds(500)
```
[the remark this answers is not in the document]
SionoiS: @tinniaru3005 WDYT?

tests/libp2p/service_discovery/test_registrar.nim
```diff
@@ -594,42 +593,41 @@ suite "Service Discovery Registrar - Configuration Variations":
...
 
-    let discoConfig1 = ServiceDiscoveryConfig.new(occupancyExp = chronos.seconds(1))
+    let discoConfig1 = ServiceDiscoveryConfig.new(occupancyExp = 1.0)
     let w1 = registrar.waitingTime(discoConfig1, ad, 1000, serviceId, now)
 
-    let discoConfig2 = ServiceDiscoveryConfig.new(occupancyExp = chronos.seconds(20))
+    let discoConfig2 = ServiceDiscoveryConfig.new(occupancyExp = 20.0)
     let w2 = registrar.waitingTime(discoConfig2, ad, 1000, serviceId, now)
 
     check w2 >= w1
```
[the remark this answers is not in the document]
SionoiS: @tinniaru3005 WDYT?

libp2p/protocols/service_discovery/types.nim
gmelodie: Yeah leave it as `chronos.Duration`, if we ever wanna support a config file we'll implement a parser that converts between the strings to the proper types

libp2p/protocols/service_discovery/registrar.nim
vladopajic: > we use integers on the wire

this is not the wire tho...

---

this is currently ticket encode method:

```nim
proc encode*(ticket: Ticket): ProtoBuffer {.raises: [], gcsafe.} =
  var pb = initProtoBuffer()
   ....
  pb.write(4, ticket.tWaitFor)
  ....
  return pb
```

but it should be


```nim
proc encode*(ticket: Ticket): ProtoBuffer {.raises: [], gcsafe.} =
  var pb = initProtoBuffer()
   ....
  pb.write(4, ticket.tWaitFor.seconds.uint32) <-------------------------------------- for example convert to on the wire representation when you actually send on wire / receive from wire
  ....
  return pb
```

vladopajic (changes requested):
- all fields should use chronos duration, there is no reason why this can't happen
- convert to appropriate on-the-wire value when doing encode /decode

libp2p/protocols/service_discovery/types.nim
vladopajic: >  float/int since tickets are supposed to be encoded/signed/sent over the wire

all duration values must be of `Duration` type. when doing encode/decode, duration value should be converted to appropriate on the wire representation.

SionoiS:
Ok I changed the logic so that chronos types are used everywhere and do the conversion when de/encoding.

libp2p/protocols/service_discovery/registrar.nim
SionoiS: `w_2 ≥ w_1 - (t_2 - t_1)` that's the invariant from the spec, could we not write the logic so that it match?

```
if waitDuration < (bound - timestamp) - (now - timestamp):
    waitDuration = (bound - timestamp) - (now - timestamp)
```

I'm not sure :thinking: this is complicated... 

edit: it works tests passes :tada:

gmelodie (approved):

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -46,9 +47,9 @@ type
   # Nested within Register message
   Ticket* {.public.} = object
     advertisement*: seq[byte] # field 1 - Copy of the original advertisement
-    tInit*: uint64 # field 2 - Ticket creation timestamp (Unix time in seconds)
-    tMod*: uint64 # field 3 - Last modification timestamp (Unix time in seconds)
-    tWaitFor*: uint32 # field 4 - Remaining wait time in seconds
+    tInit*: Moment # field 2 - Ticket creation timestamp (Unix time in seconds)
+    tMod*: Moment # field 3 - Last modification timestamp (Unix time in seconds)
+    tWaitFor*: Duration # field 4 - Remaining wait time in seconds
```
vladopajic, line 52: descriptions explaining actual type should not be needed anymore?

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -192,15 +193,22 @@ proc decode*(T: type Peer, pb: ProtoBuffer): ProtoResult[T] =
...
+    ticket = Ticket()
+    tInit: uint64 = 0
+    tMod: uint64 = 0
+    tWaitFor: uint32 = 0
 
   discard ?pb.getField(1, ticket.advertisement)
-  discard ?pb.getField(2, ticket.tInit)
-  discard ?pb.getField(3, ticket.tMod)
-  discard ?pb.getField(4, ticket.tWaitFor)
+  discard ?pb.getField(2, tInit)
```
vladopajic, line 203:
```
discard ?pb.getField(2, tInit)
```
this would be nicer with helper

```
discard ?pb.getField(2, ticket.tInit, Second)
```
```
nim-libp2p/libp2p/protobuf/minprotobuf.nim

proc getField*(
    pb: ProtoBuffer, field: int, output: var Moment, unit: 
): ProtoResult[bool] {.inline.} =
```

but there will be soon initiative to replace `minprotobuf` with `protobuf_serialization`. 
but still feel free if you want to

libp2p/protocols/service_discovery/registrar.nim
tinniaru3005: looks good to me :)

tinniaru3005 (approved):
Looks good to me

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -167,58 +155,66 @@ proc waitingTime*(
     discoConfig.advertExpiry.seconds.float64 * occupancy *
     (serviceSim + ipSim + discoConfig.safetyParam)
 
+  # Bound & Quantize W
+  w = max(0.0, w)
+  w = min(w, float64(uint32.high))
+  w = ceil(w)
+
+  var waitDuration = chronos.seconds(w.int64)
```
vladopajic: usually code here does `.seconds` or `.unit` at the end of statment 
```
  var waitDuration = (w.int64).seconds
```

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -167,58 +155,66 @@ proc waitingTime*(
...
-    let elapsedService =
-      safeSub(now, registrar.timestampService.getOrDefault(serviceId, 0))
-    let boundServiceVal = registrar.boundService.getOrDefault(serviceId, 0.0)
-    let serviceLowerBound = boundServiceVal - elapsedService.float64
-    if serviceLowerBound > w:
-      w = serviceLowerBound
+    let
+      prevTimestamp = registrar.timestampService.getOrDefault(serviceId, Moment())
+      prevBound = registrar.boundService.getOrDefault(serviceId, Moment())
+      elapsedDuration = now - prevTimestamp
```
vladopajic:
```
prevTimestamp = registrar.timestampService.getOrDefault(serviceId, Moment())
elapsedDuration = now - prevTimestamp
```

this can be negative if `registrar.timestampService` does not have `serviceId`.

does it make sense to do this:
```
prevTimestamp = registrar.timestampService.getOrDefault(serviceId, now)
elapsedDuration = now - prevTimestamp
```

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -300,12 +295,14 @@ proc findAdIdx*(ads: seq[Advertisement], peerId: PeerId): int =
 
 proc findOldestKey(disco: ServiceDiscovery): AdvertisementKey =
   var oldestKey: AdvertisementKey
-  var oldestTime = high(uint64)
+  var oldestTime: Moment
+  var first = true
```
vladopajic: `var oldestTime: Momet.high`

vladopajic: on the side note, this proc assumes there will always be result. not sure if there are really guarantees for that.

libp2p/protocols/service_discovery/types.nim
```diff
@@ -21,11 +21,11 @@ const
...
   Default_F_lookup* = 30
   Default_F_return* = 10
-  Default_E* = 900.0
+  Default_E*: int64 = 900
   Default_C*: uint64 = 1_000
-  Default_P_occ* = chronos.seconds(10)
+  Default_P_occ* = 10.0
   Default_G* = 1e-7
-  Default_Delta* = chronos.seconds(1)
+  Default_Delta*: int32 = 1
```
vladopajic: was this by intention?

libp2p/protocols/service_discovery/types.nim
```diff
@@ -66,11 +66,11 @@ type
     kLookup*: int
     fLookup*: int
     fReturn*: int
-    advertExpiry*: chronos.Duration
+    advertExpiry*: timer.Duration
```
vladopajic: why not `chronos` ?

tests/libp2p/service_discovery/test_components.nim
```diff
@@ -46,138 +46,175 @@ suite "processRetryTicket":
       status: Opt.none(kad_protobuf.RegistrationStatus),
       ticket: Opt.none(Ticket),
     )
-    let tResult = disco.processRetryTicket(regMsg, ad, 900.0, 1000)
-    check abs(tResult - 900.0) < 0.001
+    let tWait = seconds(900)
```
vladopajic: `let tWait = 900.seconds` 
here and elsewhere

libp2p/protocols/service_discovery/types.nim
```diff
@@ -66,11 +66,11 @@ type
...
     advertCacheCap*: uint64
-    occupancyExp*: chronos.Duration
+    occupancyExp*: float64
     safetyParam*: float64
-    registrationWindow*: chronos.Duration
+    registrationWindow*: timer.Duration
```
vladopajic: ?

libp2p/protocols/service_discovery/types.nim
```diff
@@ -66,11 +66,11 @@ type
     kLookup*: int
     fLookup*: int
     fReturn*: int
-    advertExpiry*: chronos.Duration
+    advertExpiry*: timer.Duration
     advertCacheCap*: uint64
-    occupancyExp*: chronos.Duration
+    occupancyExp*: float64
```
vladopajic, line 71: ?

vladopajic (approved):
there are few places where `timer.Duration` is used; whole codebase needs to converge to singe `Duration` type. please stick to `chronos`, and if `chronos.Duration` is not sufficient enough for whatever reason we will need to address that, ether by improving type in library or changing direction and utilizing   `timer.Duration` in nim-libp2p library instead.

those comments must be addressed before merge

vladopajic:
disabled auto-merge since there are comments ^ that needs to be addressed

libp2p/protocols/service_discovery/registrar.nim
SionoiS: The `-` proc doesn't allow negative `Duration` but in this case maybe we should return an error if there's no value to get. :thinking: It's a situation that should never happen normally.

libp2p/protocols/service_discovery/registrar.nim
SionoiS: > on the side note, this proc assumes there will always be result. not sure if there are really guarantees for that.

I'll check the code this this specifically, we might want to log errors in cases where this guaranty is broken.

libp2p/protocols/service_discovery/types.nim
SionoiS: When adding back chronos types, I used `timer.X` instead my bad.

libp2p/protocols/service_discovery/types.nim
SionoiS: This one is indeed on purpose, this value is not time related.

libp2p/protocols/service_discovery/registrar.nim
vladopajic: if it is theoretically impossible, and you are sure, to not get value, then you should not do anything.
otherwise it always better to be on safe side

libp2p/protocols/service_discovery/types.nim
SionoiS: First one yes, that value is not time related. Second one no, I'll change that.

libp2p/protocols/kademlia/protobuf.nim
SionoiS: Still very useful info even with the correct types.

libp2p/protocols/kademlia/protobuf.nim
SionoiS: The protobuf code would have to know about chronos which is a bit weird.

I'll let you guys handle this.

richard-ramos:
@copilot resolve the merge conflicts in this pull request

PR #2344, opened 2026-04-27
fix(connmanager): max conns per peer limit

### LINKED ISSUE #2328: connmanager: `# FIXME This should be 1 instead of 0, it will get fixed soon`
address this comment

vladopajic:
## Summary

`connmanager`'s property `maxConnsPerPeer` was not correctly implemented. it had off by one error and when value was set to `maxConnsPerPeer = 1` it actually meant  "peer should have up to 2 connections".

in this pr corrects `maxConnsPerPeer` to actually mean what it is supposed to mean.

because `defaultMaxConnsPerPeer`  had value of  `1` and since this actually meant `2`, `defaultMaxConnsPerPeer` is now increased to `2`. this is done because:
-  otherwise all code will to be updated to similarly to [this](https://github.com/vacp2p/nim-libp2p/pull/2344/changes#diff-653923ae1438326da4aadabaf283ea0645f7b5317c473cb068da8d26d2027655R29) change
- this will keep backwards comparability in cases when `.withMaxConnsPerPeer()` builder is not used

however, this will be "braking change" for those that used `withMaxConnsPerPeer` with custom value, because if they used it with value `3` this will now really mean `3` and not `4`.

## Affected Areas

- [x] Connection Manager

comparability is not changed with this.

## Impact on Library Users

"braking change" for those that used `withMaxConnsPerPeer`  with custom value.

## Risk Assessment

low.

## References

https://github.com/vacp2p/nim-libp2p/issues/2328


tests/libp2p/test_conn_manager.nim
```diff
@@ -115,7 +115,6 @@ suite "Connection Manager":
     await connMngr.close()
 
   asyncTest "get conn with direction":
-    # This would work with 1 as well cause of a bug in connmanager that will get fixed soon
```
vladopajic, line 124: this is not true, because 2 muxers are create for same peer, therefore `maxConnsPerPeer` has to be 2.

tests/libp2p/test_conn_manager.nim
```diff
@@ -176,30 +175,28 @@ suite "Connection Manager":
...
     await connMngr.close()
     await allFuturesRaising(muxs.mapIt(it.close()))
 
   asyncTest "expect connection from peer":
-    # FIXME This should be 1 instead of 0, it will get fixed soon
-    let connMngr = newMaxTotal(maxConnsPerPeer = 0)
-    let peerId = PeerId.random(rng).tryGet()
-
-    await connMngr.storeMuxer(makeMuxer(peerId))
+    let connMngr = newMaxTotal(maxConnsPerPeer = 1)
```
vladopajic, line 185: fixed now

libp2p/connmanager.nim
```diff
@@ -183,8 +180,7 @@ proc new*(
   T(
     muxerStore: MuxerStore.new(),
     maxConnsPerPeer:
-      if maxConnsPerPeer >= 0: maxConnsPerPeer else: DefaultMaxConnectionsPerPeer,
-      # issue#2328 must never be 0
+      if maxConnsPerPeer > 0: maxConnsPerPeer else: DefaultMaxConnectionsPerPeer,
     maxConnectionsIn: maxInArg,
```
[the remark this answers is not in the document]
vladopajic: should be common sense... `maxConnsPerPeer = 0` should mean node is not connected to anything. there is no use of such node. 

except for some testing use case where node will terminate connection, but this case should be done with specialized mechanism anyway.

libp2p/connmanager.nim
```diff
@@ -417,19 +413,17 @@ proc storeMuxer*(
   let
     peerId = muxer.connection.peerId
     dir = muxer.connection.dir
+    peerConnsCount = c.muxerStore.count(peerId)
+    isNewPeer = peerConnsCount == 0
```
gmelodie, line 435: Is this ever used?

gmelodie (approved):

libp2p/connmanager.nim
gmelodie: (i mean `isNewPeer`)

vladopajic: yes, here https://github.com/vacp2p/nim-libp2p/pull/2344/changes#diff-3ddd628fc83a4e98d3f7fce526358838819a6c0aa75d0f7d24cd61a43b578392R434

vladopajic: https://github.com/vacp2p/nim-libp2p/pull/2344/changes#diff-3ddd628fc83a4e98d3f7fce526358838819a6c0aa75d0f7d24cd61a43b578392L440

vladopajic: L440 or L434

libp2p/connmanager.nim
```diff
@@ -151,9 +151,6 @@ proc maxInOut*(T: type LimitsConfig, maxIn: int, maxOut: int): LimitsConfig =
   doAssert maxOut > 0, "maxOut must be > 0"
   LimitsConfig(maxIn: maxIn, maxOut: maxOut)
 
-proc newTooManyConnectionsError(): ref TooManyConnectionsError {.inline.} =
-  result = newException(TooManyConnectionsError, "Too many connections")
-
 proc new*(
     T: type ConnManager,
     maxConnsPerPeer: int = -1,
```
richard-ramos: Maybe this should be 0? since now in line 201 and in  libp2p/builders.nim:571 checks for > 0?
Same for the comment in line 163

richard-ramos (approved):

libp2p/connmanager.nim
vladopajic: any value `<= 0` will cause `maxConnsPerPeer` to use default value, because these values are not really valid - all are considered as "undefined".

i have chosen `-1` as default value here as it subjectively  appears more, as undefined value in this context. `0` can appear as valid value.

but i don't have strong opinion here. ether should be okay.

vladopajic: https://github.com/vacp2p/nim-libp2p/pull/2344/commits/be82e09b06e946aef7499975c31824e4bc765767

PR #2345, opened 2026-04-27
fix(kad): race in randomFind

### LINKED ISSUE #2281: possible race condition in random find
https://github.com/vacp2p/nim-libp2p/blob/620b6a44c701d0c20259b298bfe71dbbb5719a69/libp2p/protocols/service_discovery/random_find.nim#L37-L38

This code can block forever if the queue becomes empty while findNodeFut is not finished yet,
The condition to loop only depends on `findNodeFut` not being finished and the queue is not empty
This is not a problem the first time the loop is executed because the queue will be seeded with closest peers (on line 30)
but then, in future iterations, the line with `await.queue.popFirst()` can be problematic, because that queue can be later filled by async lookups done on `findNode`, so the following condition might happen:
1. async lookup are still in progress (`findNodeFut` is not finished yet) and the queue is empty (because it already processed everything in past iterations)
2. the code proceeds to line with `await queue.popFirst()`.
3. The lookup finishes, with no further items added to the queue
4. code gets stuck

gmelodie:
## Summary

Fixes a race condition in `randomRecords` where the processing loop could block indefinitely on an empty `AsyncQueue`. The loop condition (`not findNodeFut.finished or not queue.empty()`) permits entry when `findNodeFut` is still running even if the queue is empty, and the unconditional `await queue.popFirst()` would then suspend forever if `findNodeFut` completed without enqueuing further peers.

The fix detects the empty-queue case and awaits `findNodeFut` directly, ensuring the loop either resumes with newly enqueued peers or exits cleanly once no more work can arrive.

- `libp2p/protocols/service_discovery/random_find.nim`:
  - Added an `if queue.empty()` guard at the top of the processing loop.
  - When the queue is empty (but `findNodeFut` is still running), the loop now awaits `findNodeFut` before continuing, preventing an indefinite block on `popFirst`.
  - The post-loop `await findNodeFut` is unaffected; re-awaiting a completed Chronos future is a no-op that returns the stored result immediately.

## Affected Areas

- Gossipsub  
- Transports  
- Peer Management / Discovery  
  - Fixes a hang in the Kademlia random-walk peer discovery loop.
- Protocol Logic  
  - Corrects the async control flow in `randomRecords` / `lookupRandom`.
- Build / Tooling  
- Other  

The public API (`lookupRandom`) is unchanged. The fix only alters internal async scheduling within `randomRecords`: callers receive the same `seq[ExtendedPeerRecord]` result and no existing encode/decode paths are touched.

Downstream consumers (Nimbus, Waku, Codex) that call `lookupRandom` do not need revalidation beyond confirming the previously hanging scenario now terminates correctly.

## Impact on Library Users

- No API or ABI changes — all types, procs, and signatures are identical to before.
- Behavior change only: `lookupRandom` now always terminates instead of hanging indefinitely when the Kademlia lookup finishes with an empty queue.

## Risk Assessment

- Low — loop termination change: previously the loop could hang forever (high-severity bug). Now it exits as soon as `findNodeFut` completes and the queue is empty. There is no scenario where the old non-hanging path produces a different result from the new path, because awaiting `findNodeFut` followed by `continue` still drains any items enqueued before completion.
- Low — double-await of `findNodeFut`: Chronos allows re-awaiting a completed future; it returns the cached result synchronously. The post-loop error-logging block is therefore unaffected.

## References

Fix #2281

### DIFF
--- a/libp2p/protocols/service_discovery/random_find.nim
+++ b/libp2p/protocols/service_discovery/random_find.nim
@@ -38,17 +38,25 @@ proc randomRecords(
     let peerId =
       if queue.empty():
         # The queue is temporarily empty while findNodeFut may still enqueue
-        # more peers. Wait for whichever happens first: a new queued peer or
-        # completion of findNodeFut.
+        # more peers. Use an event to wake up when either a peer is enqueued
+        # OR findNodeFut finishes, so we react to whichever comes first.
         let popFirstFut = queue.popFirst()
-        await one(popFirstFut, findNodeFut)
-
-        if popFirstFut.finished:
-          let (peerId, _) = popFirstFut.read()
+        let wakeEvent = newAsyncEvent()
+        popFirstFut.addCallback(
+          proc(_: pointer) {.gcsafe, raises: [].} =
+            wakeEvent.fire()
+        )
+        findNodeFut.addCallback(
+          proc(_: pointer) {.gcsafe, raises: [].} =
+            wakeEvent.fire()
+        )
+        await wakeEvent.wait()
+        if popFirstFut.completed:
+          let (peerId, _) = await popFirstFut
           peerId
         else:
           await popFirstFut.cancelAndWait()
-          continue
+          break
       else:
         let (peerId, _) = await queue.popFirst()
         peerId
--- a/libp2p/protocols/service_discovery/random_find.nim
+++ b/libp2p/protocols/service_discovery/random_find.nim
@@ -50,7 +50,11 @@ proc randomRecords(
           proc(_: pointer) {.gcsafe, raises: [].} =
             wakeEvent.fire()
         )
-        await wakeEvent.wait()
+        try:
+          await wakeEvent.wait()
+        except CancelledError as e:
+          await popFirstFut.cancelAndWait()
+          raise e
         if popFirstFut.completed:
           let (peerId, _) = await popFirstFut
           peerId
--- a/libp2p/protocols/service_discovery/random_find.nim
+++ b/libp2p/protocols/service_discovery/random_find.nim
@@ -35,15 +35,23 @@ proc randomRecords(
 
   var buffers: seq[seq[byte]]
   while not findNodeFut.finished or not queue.empty():
-    if queue.empty():
-      # findNodeFut is still running but the queue is temporarily empty.
-      # Wait for it to finish — it may enqueue more peers, or once done we know none will come.
-      let waitRes = catch:
-        await findNodeFut
-      discard waitRes
-      continue
-    let (peerId, _) = await queue.popFirst()
+    let peerId =
+      if queue.empty():
+        # The queue is temporarily empty while findNodeFut may still enqueue
+        # more peers. Wait for whichever happens first: a new queued peer or
+        # completion of findNodeFut.
+        let popFirstFut = queue.popFirst()
+        await one(popFirstFut, findNodeFut)
 
+        if popFirstFut.finished:
+          let (peerId, _) = popFirstFut.read()
+          peerId
+        else:
+          await popFirstFut.cancelAndWait()
+          continue
+      else:
+        let (peerId, _) = await queue.popFirst()
+        peerId
     let res = catch:
       await disco.dispatchGetVal(peerId, peerId.toKey())
     let msgOpt = res.valueOr:
--- a/libp2p/protocols/service_discovery/random_find.nim
+++ b/libp2p/protocols/service_discovery/random_find.nim
@@ -11,6 +11,38 @@ import ./[types]
 logScope:
   topics = "ext-kad-dht random records"
 
+proc makeFireCallback(event: AsyncEvent): CallbackFunc =
+  proc(_: pointer) {.gcsafe, raises: [].} =
+    event.fire()
+
+proc nextPeer(
+    queue: AsyncQueue[(PeerId, Opt[Message])], findNodeFut: Future[seq[PeerId]]
+): Future[Opt[PeerId]] {.async: (raises: [CancelledError]).} =
+  ## Pop the next peer from the queue, blocking until one arrives or
+  ## findNodeFut finishes. Returns none when no more peers will come.
+  if not queue.empty():
+    let (peerId, _) = await queue.popFirst()
+    return Opt.some(peerId)
+
+  # Queue is temporarily empty while findNodeFut may still enqueue more peers.
+  # Wait for whichever comes first.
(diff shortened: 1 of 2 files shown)


libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -35,6 +35,13 @@ proc randomRecords(
...
   var buffers: seq[seq[byte]]
   while not findNodeFut.finished or not queue.empty():
+    if queue.empty():
+      # findNodeFut is still running but the queue is temporarily empty.
+      # Wait for it to finish — it may enqueue more peers, or once done we know none will come.
+      let waitRes = catch:
+        await findNodeFut
+      discard waitRes
+      continue
     let (peerId, _) = await queue.popFirst()
```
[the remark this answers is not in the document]
vladopajic: this seems like a more robust suggestion
@gmelodie wdyt?

libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -35,8 +35,23 @@ proc randomRecords(
...
+        # completion of findNodeFut.
+        let popFirstFut = queue.popFirst()
+        await one(popFirstFut, findNodeFut)
+
+        if popFirstFut.finished:
+          let (peerId, _) = popFirstFut.read()
+          peerId
+        else:
+          await popFirstFut.cancelAndWait()
+          continue
```
vladopajic: hmmm this `continue` ...

what happens if `queue` was empty and `findNodeFut` got cancelled or failed ?

vladopajic: this code would be much more easier with csp semantics, use queue as channel, push to queue asynchronously from `findNode`, when push is done close queue. 
in while loop read until queue is closed. end of story.

gmelodie: well it'll hit continue and loop will not run again, no?

vladopajic: aha i see now `finished` in while condition.

vladopajic (approved):

libp2p/protocols/service_discovery/random_find.nim
vladopajic: could it be `brake` (instead of `continue`)? it should convey intentions better?

libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -35,8 +35,31 @@ proc randomRecords(
...
+        let wakeEvent = newAsyncEvent()
+        popFirstFut.addCallback(
+          proc(_: pointer) {.gcsafe, raises: [].} =
+            wakeEvent.fire()
+        )
+        findNodeFut.addCallback(
+          proc(_: pointer) {.gcsafe, raises: [].} =
+            wakeEvent.fire()
+        )
+        await wakeEvent.wait()
```
richard-ramos: If a call to `randomRecords` is cancelled while awaiting here, `popFirstFut` will leak, maybe do something like:

```suggestion
try:
  await wakeEvent.wait()
except CancelledError as exc:
  await popFirstFut.cancelAndWait()
  raise exc
```


I asked claude for a small repro test unit to reproduce the behavior so it's easier to see what's going on:

```nim
import chronos
import unittest2

proc waitLikeRandomFind[T](
    queue: AsyncQueue[T], producer: Future[void]
): Future[void] {.async: (raises: [CancelledError]).} =
  if queue.empty():
    let popFirstFut = queue.popFirst()
    let wakeEvent = newAsyncEvent()

    popFirstFut.addCallback(proc(_: pointer) {.gcsafe, raises: [].} =
      wakeEvent.fire()
    )
    producer.addCallback(proc(_: pointer) {.gcsafe, raises: [].} =
      wakeEvent.fire()
    )

    await wakeEvent.wait()

    if popFirstFut.completed:
      discard await popFirstFut
    else:
      await popFirstFut.cancelAndWait()

suite "random find cancellation repro":
  test "cancelling event wait leaves popFirst future attached to queue":
    waitFor((proc() {.async.} =
      let queue = newAsyncQueue[int]()
      let producer = newFuture[void]("producer")

      let waiter = waitLikeRandomFind(queue, producer)
      await sleepAsync(10.millis)
      await waiter.cancelAndWait()

      queue.addLastNoWait(42)
      await sleepAsync(10.millis)

      check await queue.popFirst().withTimeout(50.millis)
      # The expected behavior is that 42 remains available after cancelling waiter; the current PR-shaped logic lets the orphaned popFirstFut consume it.

      await producer.cancelAndWait()
    )())
```

gmelodie: Nice catch! Accepted and also added a regression check test

libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -35,8 +35,35 @@ proc randomRecords(
...
+            wakeEvent.fire()
+        )
+        findNodeFut.addCallback(
+          proc(_: pointer) {.gcsafe, raises: [].} =
+            wakeEvent.fire()
+        )
+        try:
+          await wakeEvent.wait()
+        except CancelledError as e:
+          await popFirstFut.cancelAndWait()
```
vladopajic: `findNodeFut` cancel this one as well?

libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -35,8 +35,35 @@ proc randomRecords(
...
+          raise e
+        if popFirstFut.completed:
```
vladopajic:
```suggestion

        if popFirstFut.completed:
```
gives a brake to reader

libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -35,8 +35,35 @@ proc randomRecords(
...
+      if queue.empty():
+        # The queue is temporarily empty while findNodeFut may still enqueue
+        # more peers. Use an event to wake up when either a peer is enqueued
+        # OR findNodeFut finishes, so we react to whichever comes first.
+        let popFirstFut = queue.popFirst()
+        let wakeEvent = newAsyncEvent()
+        popFirstFut.addCallback(
+          proc(_: pointer) {.gcsafe, raises: [].} =
+            wakeEvent.fire()
+        )
```
vladopajic:
```
popFirstFut.addCallback(makeFireCallback(wakeEvent))
findNodeFut.addCallback(makeFireCallback(wakeEvent))
```

richard-ramos (approved):

PR #2348, opened 2026-04-28
chore: update NIMBLE_COMMIT to version v0.99.1

jmgomez:
Latest nimble test


gmelodie (approved):

richard-ramos (commented):
Looks like most CI jobs do fail at the nimble install step with this change

jmgomez:
> Looks like most CI jobs do fail at the nimble install step with this change

Yeah, its a test to see what breaks. Working on a fix

jmgomez:
FYI @richard-ramos There is a contradiction between what .pinned says and what the nimble file requires. The pinned bearssl commit (b16f70bd) has version 0.2.6 in its nimble file, but chronos (also pinned) requires bearssl >= 0.2.7. I updated the bearssl pin to v0.2.8 (22c6a76c) which fixes the dependency resolution.

richard-ramos:
@jmgomez good catch!  thank you!

richard-ramos (approved):

richard-ramos:
Hm. This is strange. It's failing now in 
``` 
   Fetching https://github.com/status-im/nim-chronos (>= 4.0.4) using git
/home/runner/work/_temp/c19884a1-ac25-4028-bb16-9951562d6be6.sh: line 14: 14575 Segmentation fault      (core dumped) nimble test
```
when executing `nimble test`...

anyway, this is in 2.0.x, and we are thinking to drop this 
¯\_(ツ)_/¯

richard-ramos:
Alright, I just removed the "Required" flag for 2.0.x  jobs, and once it's green, it will get automatically merged. 
Thanks @jmgomez!

PR #2233, opened 2026-04-06
test(mix): test delta integrity beyond leading zeros

mghazwi:
## Summary
The PR adds a test to show the integrity problem as described in vacp2p/rfc-index#306
The test mainly follows the test right on top of it, but tampering is done after the leading zeros. 
The test fails now since the integrity issue is not resolved yet, but should pass once it's fixed.

## Affected Areas
- [x] Mix

## References
vacp2p/rfc-index#306

## Additional Notes
This PR doesn't need to be merged now, but can be merged later when the issue is resolved.

PR #2349, opened 2026-04-28
feat(mix): use LIONESS for Sphinx payload encryption

chaitanyaprem:
## Summary

Replaces AES-CTR delta (payload) encryption in the Sphinx mix protocol with LIONESS, a wide-block PRP. Header encryption stays AES-CTR — only the payload is migrating.

This closes the payload integrity gap as per https://github.com/logos-co/logos-lips/issues/306: under AES-CTR, tampering past the leading-zeros window was undetectable because XOR encryption is malleable. With LIONESS, any single-bit flip in the ciphertext diffuses across the whole plaintext on decrypt, scrambling the leading-zeros tag at the exit hop. The integrity test from #2233 is included verbatim and now passes.

`processReply` additionally verifies the leading-zeros tag after unwrapping (was unchecked under AES-CTR).

For construction details see the [design notes](https://hackmd.io/@codex-storage/SkRHiWKo-l). A formal spec update is in progress (vacp2p/rfc-index#306).

## Affected Areas

- [x] Protocol Logic — Mix protocol payload encryption (Sphinx delta)

Wire-incompatible with prior AES-CTR-only nodes on the payload path. Header format unchanged.

- **Nimbus:** N/A
- **Waku:** integrated and verified end-to-end via 5-mix-node simulation in [logos-messaging/logos-delivery#3837](https://github.com/logos-messaging/logos-delivery/pull/3837) — cover traffic exercises encrypt → multi-hop forward → unwrap with zero LIONESS errors
- **Codex:** Not yet integrated with Mix

## Impact on Library Users

- No public API surface change in `mix_protocol.nim`.
- New module `libp2p/protocols/mix/lioness.nim` exports the standalone cipher.
- Operators must roll all mix nodes together due to the wire-incompatible payload format.

## Risk Assessment

Payload encryption is security-critical. **Requesting review from cryptography researchers before merge** — round structure, SHAKE128 KDF, and key-derivation chain are the focus areas.

## References

- Spec issue: vacp2p/rfc-index#306
- Design notes: https://hackmd.io/@codex-storage/SkRHiWKo-l
- Reference implementation (Rust): https://github.com/mghazwi/lioness_blockcipher
- Original integrity demonstration: #2233

## Additional Notes

- No new top-level dependencies.
- 166/166 mix tests pass (13 new LIONESS tests, 1 new Sphinx integrity test).
- Draft pending spec finalisation.

### DIFF
--- a/libp2p/protocols/mix/lioness.nim
+++ b/libp2p/protocols/mix/lioness.nim
@@ -1,0 +1,210 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Copyright (c) Status Research & Development GmbH
+
+## LIONESS wide-block cipher (Anderson & Biham, 1996), instantiated with
+## ChaCha20 (stream cipher), keyed Blake2b-256 (hash), and SHAKE128 (KDF).
+##
+## The block ``B = L || R`` is split into a 32-byte left half and a right half
+## of size ``len(B) - 32``, then four Feistel rounds are applied:
+##
+##   round 1:  R := R XOR ChaCha20(key = L XOR K1)
+##   round 2:  L := L XOR Blake2b_K2(R)
+##   round 3:  R := R XOR ChaCha20(key = L XOR K3)
+##   round 4:  L := L XOR Blake2b_K4(R)
+##
+## ``K1..K4`` are derived from a 32-byte master key by feeding it into SHAKE128
+## and reading 128 bytes (= 4 * 32) of output. The 12-byte ChaCha20 IV is
+## supplied by the caller; ``sphinx.nim`` derives it per-hop from the shared
+## secret with the same labeled-SHA-256 pattern it uses for the header
+## AES-CTR keys. See ``tests/libp2p/mix/test_lioness.nim`` for vectors.
+##
+## LIONESS itself does not provide integrity. The Sphinx construction prepends
+## ``k`` zero bytes to the plaintext before encryption and verifies them at the
+## destination after decryption: tampering anywhere in the ciphertext scrambles
+## the entire plaintext through the wide-block PRP, so the leading zeros are
+## destroyed with overwhelming probability. See the migration design note for
+## details.
+
+{.push raises: [].}
+
+import chronicles
+import results
+import nimcrypto/[blake2, keccak, utils]
+import bearssl/abi/bearssl_block
+
+logScope:
+  topics = "libp2p mix lioness"
+
+const
+  LionessLeftLen* = 32
+    ## Size of the left half ``L``. Must equal both the stream cipher key size
+    ## and the hash output size, since L is XORed with each.
+  LionessHashKeyLen* = 32
+    ## Blake2b MAC key size used in the hash rounds. Sized to the master key
+    ## (``LionessMasterKeyLen``); a larger MAC key cannot create entropy that
+    ## the master key does not already provide.
+  LionessMasterKeyLen* = 32
+    ## Size of the per-hop shared-secret master key from which round keys are
+    ## derived.
+  LionessMinBlockLen* = LionessLeftLen * 2
+    ## Minimum supported block size. The construction works for any
+    ## ``|m| > LionessLeftLen``, but real Sphinx payloads are several KB; the
+    ## stricter bound rejects degenerate inputs that callers never need.
+
+  LionessIvLen* = 12
+    ## ChaCha20 nonce size. The IV is supplied by the caller; in this
+    ## codebase ``sphinx.nim`` derives it per-hop from the shared secret
+    ## using the same labeled-SHA-256 pattern it uses for the header
+    ## AES-CTR ``aes_key``/``iv`` (see ``deriveLionessIv``).
+
+type
+  LionessError* {.pure.} = enum
+    BlockTooSmall
+    InvalidMasterKey
+    InvalidIv
+
+  RoundKeys = object
+    k1: array[LionessLeftLen, byte]
+    k2: array[LionessHashKeyLen, byte]
+    k3: array[LionessLeftLen, byte]
+    k4: array[LionessHashKeyLen, byte]
+
+  Lioness* = object
+    ## Stateless LIONESS instance. Construct via
+    ## ``Lioness.init(masterKey, iv)`` and call ``clear`` when no longer
+    ## needed to wipe round keys from memory.
+    keys: RoundKeys
+    iv: array[LionessIvLen, byte]
+
+func clear(self: var RoundKeys) =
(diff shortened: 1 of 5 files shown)


libp2p/protocols/mix/lioness.nim
```diff
@@ -0,0 +1,185 @@
...
+    ## hash key is absorbed into Blake2b's state, never XORed with ``L``.
+  LionessMasterKeyLen* = 32
+    ## Size of the per-hop shared-secret master key from which round keys are
+    ## derived.
+  LionessMinBlockLen* = LionessLeftLen + 1
+    ## Minimum supported block size: at least one byte of right-half data.
+
+  KeyMaterialLen = 2 * LionessLeftLen + 2 * LionessHashKeyLen # 192
+
+  ChaCha20Iv*: array[12, byte] =
```
bkomuves: while in theory, this looks "ok", i don't like it

we are deriving a lot of keys anyway, why don't just derive a fresh key (domain-separated!) for the IV too?

libp2p/protocols/mix/lioness.nim
```diff
@@ -0,0 +1,185 @@
...
+
+import results
+import nimcrypto/[blake2, keccak, utils]
+import bearssl/abi/bearssl_block
+
+const
+  LionessLeftLen* = 32
+    ## Size of the left half ``L``. Must equal both the stream cipher key size
+    ## and the hash output size, since L is XORed with each.
+  LionessHashKeyLen* = 64
```
bkomuves: So how keyed Blake2b works (as far as I understand), is that:

- the native key size is 64 bytes
- if the key is shorter than 64 bytes, it's padded with zeros
- if the key is longer than 64 bytes, it's hashed into 64 bytes (with what? prolly Blake2b itself :)

In any case, we are deriving all keys from a 32 byte "master key".

To me, it doesn't make any sense to have a 64 byte key here, which was derived from a 32 byte key to start with. But maybe (unlikely though) I'm missing some subtle detail here. I remember tho (cannot find right now) a statement saying that 16 bytes (128 bits) is to minimum key size, which indeed makes sense if we are targeting 128 bits of security, which is the standard assumption.

libp2p/protocols/mix/lioness.nim
```diff
@@ -0,0 +1,185 @@
...
+    ## Blake2b MAC key size used in the hash rounds. 64 is Blake2b's maximum
+    ## key length without internal truncation (RFC 7693), giving the highest
+    ## security margin available without spec deviation. Unlike the stream
+    ## cipher key, this size is not constrained by ``LionessLeftLen``: the
+    ## hash key is absorbed into Blake2b's state, never XORed with ``L``.
+  LionessMasterKeyLen* = 32
+    ## Size of the per-hop shared-secret master key from which round keys are
+    ## derived.
+  LionessMinBlockLen* = LionessLeftLen + 1
```
bkomuves: yeah i think this should be at minimum `LionessLeftLen + 16`, but more likely `LionessLeftLen + 32`.

not familiar enough with the underlying theory, but `+1` sounds _rather_ suspicious, while `+32` sounds safe; furthermore we will never user it for anything small (that's the whole point!), the minimum size will be something like 3000-4000 bytes, so yeah, just make it `+32`

libp2p/protocols/mix/lioness.nim
```diff
@@ -0,0 +1,185 @@
...
+    ctx: shake128
+    material: array[KeyMaterialLen, byte]
+  ctx.init()
+  ctx.update(masterKey)
+  ctx.xof()
+  discard ctx.output(addr material[0], uint(KeyMaterialLen))
+  ctx.clear()
+
+  var keys: RoundKeys
+  copyMem(addr keys.k1[0], addr material[0], LionessLeftLen)
```
bkomuves: huh, nim is pretty shitty lol

libp2p/protocols/mix/lioness.nim
```diff
@@ -0,0 +1,185 @@
...
+
+# ---------------------------------------------------------------------------
+# Round helpers — operate on the whole block in place; the split point is
+# always ``LionessLeftLen``.
+# ---------------------------------------------------------------------------
+
+proc streamRound(blk: var openArray[byte], subkey: openArray[byte]) =
+  ## ``R ^= ChaCha20(key = L XOR subkey, iv = ChaCha20Iv, counter = 0)``
+  doAssert subkey.len == LionessLeftLen
+  doAssert blk.len > LionessLeftLen
```
bkomuves: should be `> LionessMinBlockLen` as far i as i understand, but this is pretty horrible code to be honest

libp2p/protocols/mix/lioness.nim
```diff
@@ -0,0 +1,185 @@
...
+    addr blk[LionessLeftLen],
+    csize_t(rightLen),
+  )
+
+  burnMem(roundKey)
+
+proc hashRound(blk: var openArray[byte], subkey: openArray[byte]) =
+  ## ``L ^= Blake2b_{subkey}(R)``  (32-byte digest)
+  doAssert subkey.len == LionessHashKeyLen
+  doAssert blk.len > LionessLeftLen
```
bkomuves: same as above

libp2p/protocols/mix/lioness.nim
chaitanyaprem: For now just followed the rust reference implementation by Mo. But should have made it align with other primitives as per mix spec.
Will change it to derive fresh key for IV .

libp2p/protocols/mix/lioness.nim
chaitanyaprem: Thanks!

You must be referring to the [mix spec](https://lip.logos.co/ift-ts/raw/mix.html#82-cryptographic-primitives) which specifies that all cryptographic ops target  a min of 128bits.

Got it, and changing it to 32 bytes.
You are not missing any detail , i just used the rust implementation as a reference to write this, but should have looked at mix security params to review.

libp2p/protocols/mix/lioness.nim
chaitanyaprem: You mean the assertions? 
I don't like having assertions either(but since we are already doing this in other places, just followed suite) , let me return and log error which is better.

libp2p/protocols/mix/lioness.nim
chaitanyaprem: well, dont blame nim here...it is my knowledge of it. 
Found out a way to not require copying and will change it.

chaitanyaprem:
Closing this PR as this will be now tracked at https://github.com/logos-co/nim-libp2p-mix/pull/1

Issue #2336, opened 2026-04-24
The OS Backlog Trap: `switch.nim` blocks `accept()`, causing remote nodes to hang and timeout


radiken:
## Description

Currently, the `accept` loop blocks on a semaphore (`getIncomingSlot()`) *before* calling `transport.accept()`. When the node reaches its maximum connection limit, the accept loop freezes. However, the underlying Operating System kernel does not freeze. The kernel continues to complete 3-way TCP handshakes and queues the established connections into the listen backlog.

Because the TCP handshake succeeds at the OS level, remote connecting nodes believe they have successfully connected. They send their first cryptographic handshake packet (e.g., Noise) and wait for a response. Because `nim-libp2p` never calls `accept()`, it never reads the packet. The remote node hangs in the dark until it hits a 10-second timeout, failing the connection.

## Impact

This impacts the network during scenarios like DHT warmups, bootstrap node surges.
- **Latency Spikes:** Connecting nodes waste 10 seconds waiting for a response that will never come, massively slowing down their dialing phases.
- **Resource Exhaustion on Remote Nodes:** Remote nodes keep their file descriptors and memory allocated for 10 seconds for a "dead" connection.
- **Confusing Logs:** Operators see successful `connect()` calls followed by bizarre cryptographic handshake timeouts.

## Root Cause Code

**File:** `libp2p/switch.nim` (inside the `accept` loop)
```nim
# 1. Blocks here if MaxConnections limit is reached
let slot = await s.connManager.getIncomingSlot() 

# 2. Never executes until a slot frees! The OS handles TCP handshakes in the meantime.
conn = await transport.accept()                  
```

## Steps to Reproduce
1. Checkout to this commit: https://github.com/vacp2p/nim-libp2p/commit/0b744d7eaa921cda6d2c205846732c62591108bb
2. run `~/.nimble/bin/nim c -r reproduce_backlog_trap.nim`

## Proposed Solution

Do not block the `accept()` loop waiting for a slot. The application should always pull the connection from the OS as fast as possible to prevent the backlog from filling up with "zombie" connections.

The logic should be inverted:
1. Call `await transport.accept()` unconditionally.
2. Once the connection is accepted, check if there is an available slot (non-blockingly).
3. If the `MaxConnections` limit is reached, **immediately call `conn.close()`**.

[...]

PR #2360, opened 2026-04-28
fix(muxer): accept loop stuck

gmelodie:
## Summary

This PR fixes a bottleneck in the switch's incoming connection accept loop that caused bootstrap and warmup phases to stall when deploying 80+ nodes.

Previously, `upgrades.acquire()` (bounded by `ConcurrentUpgrades = 4`) was called **before** `transport.accept()`. This meant the loop could only accept a new TCP connection once an upgrade slot was free. Under load, this caused the OS TCP accept queue to fill up, and dialing nodes would hit their upgrade timeout before their connection was ever accepted.

The fix moves `upgrades.acquire()` into `upgradeMonitor`, so it runs **after acceptance**:
- TCP connections are accepted immediately (bounded only by `MaxConnections`)
- The upgrade concurrency limit is enforced asynchronously in the spawned upgrade task

### Changes

- **libp2p/switch.nim**
  - `upgradeMonitor`:
    - Added `await upgrades.acquire()` at the top (before calling `switch.upgrader`)
    - Guarded by a `semAcquired` boolean so the `finally` block only releases the semaphore if it was actually acquired
  - **Accept loop**:
    - Removed the `await upgrades.acquire()` / `except CancelledError: return` block at the top of each iteration
    - Removed the three corresponding `upgrades.release()` calls from:
      - nil-connection path
      - `CancelledError` path
      - `CatchableError` path
    - The loop no longer owns the semaphore

---

## Affected Areas

- Gossipsub  
- Transports  
  - Accept loop in `Switch` no longer serializes on the upgrade semaphore before calling `transport.accept()`  
- Peer Management / Discovery  
- Protocol Logic  
- Build / Tooling  
- Other  

---

- Public API of `Switch`, `Transport`, and `ConnManager` is unchanged  
- Upgrade concurrency limit (`ConcurrentUpgrades = 4`) is preserved  
- 30-second upgrade timeout is preserved  
- `MaxConnections` invariant is still enforced via `getIncomingSlot()`  
- Existing encode/decode and dial paths are unaffected  

Downstream consumers (Nimbus, Waku, Codex) interact via:
- `switch.connect()`
- muxer streams
- protocol handlers  

None of these are impacted. No additional validation is required beyond running existing integration test suites.

---

## Impact on Library Users

- No new types, procs, or fields added or removed  
- `ConcurrentUpgrades` remains exported and unchanged  
- **Behavior change (non-breaking):**
  - Under high inbound load, TCP connections are accepted earlier
  - Connections waiting for upgrade slots are held in-memory (as async tasks) instead of the OS TCP accept queue
  - This is observable via connection-state metrics but does not affect the API surface

---

## Risk Assessment

- **Semaphore never-release regression (low)**  
  Previously, `upgrades.release()` was called unconditionally in `finally`. Now it is guarded by `semAcquired`.  
  If `upgrades.acquire()` raises `CancelledError` before setting the flag, the semaphore is correctly not released.  
  A bug here would drain the semaphore permanently, but the logic mirrors existing patterns (`upgradeSuccessful`).

- **Accepted-but-never-upgraded connections (low)**  
  If cancellation occurs between `transport.accept()` and slot acquisition:
  - Connection is closed in `upgradeMonitor`'s `finally` (`conn.close()`)
  - Slot is released by the connection tracking monitor  
  → No resource leak

- **Upgrade queue depth (low)**  
  Up to `MaxConnections` `upgradeMonitor` tasks may queue waiting for one of 4 upgrade slots.  
  Each task holds a TCP connection reference. With typical values (50–200 connections, ~1s upgrades), the queue drains quickly and memory overhead is negligible.

---

## References

- Upstream issue: https://github.com/vacp2p/nim-libp2p/issues/2336  
- Related experiment: KAD-DHT Planning (internal Notion)


richard-ramos (commented):
there's a problem with this approach,

The server will now first accept the connection, and count it as an active incoming connection,and only after that it will wait for a upgrade slot. The problem is that the 30 second upgrade timeout will start ONLY when the upgrade process starts, and not while the connection is waiting for the slot.

So, this means a connection can use one connection slot while doing nothing and waiting... thus making libp2p more prone to DoS. if many malicious or slow peers try this, they will quickly fill all incoming connection slot, then real peers (in case of malicious peers) cannot connect due to not having available slots... and this can last much longer than 30 seconds due to the flow introduced in this PR:

1. Connection is accepted
2. Consumes an incoming connection slot
3. It waits for one of the 4 upgrade slot (this step can take a long time, waiting for an upgrade slot).
4. After you get the upgrade slot, code will run: `await switch.upgrader(trans, conn).wait(30.seconds)`. 

Before this PR, you could still have a wait longer than 30s, but it was less annoying because the wait would be at the tcp transport layer, waiting until the app called accept, and not affecting the incoming slot count. 
Before this change, we waited for an upgrade slot to be available before the next connection, so these waiting connections did not use a connection slot...

I think to fix this, the 30s timeout should cover both waiting for an upgrade slot and running the upgrade itself,  something +- like this.
```
let timeout = Moment.now() + 30.seconds
await upgrades.acquire().wait(timeout)
await switch.upgrader(transport, conn).wait(timeout)
```

tests/libp2p/test_switch.nim
```diff
@@ -35,6 +35,40 @@ const TestCodec = "/test/proto/1.0.0"
...
+    for _ in 0 ..< 10:
+      try:
+        await switch.connect(peerId, addrs)
+        return
+      except DialFailedError as e:
+        lastErr = e
+        await sleepAsync(200.milliseconds)
+    raise lastErr
+  else:
+    await switch.connect(peerId, addrs)
```
vladopajic: i am now realizing that this is not clean solution, because doing this all over the place would be dirty.
how about:

- change transport in switch to use something else, transport _should_ be irrelevant here for purpose of these tests. because these are testing switch and not the transport itself?
  - could we use quic, or even memory transport (we have built it for this actually)
- is there anything better?

libp2p/transports/quictransport.nim
```diff
@@ -131,10 +128,7 @@ method write*(
           bytes.len.int64, labelValues = [stream.shortAgent]
         )
   except StreamError as e:
-    if e of ref StreamResetError:
-      raise newLPStreamResetError()
-
-    raise newLPStreamEOFError()
+    raise newLPStreamResetError()
```
richard-ramos, line 131: This change seems unrelated to this PR and also changes how reset works for Quic 🤔

richard-ramos (approved):

libp2p/switch.nim
```diff
@@ -233,31 +238,36 @@ proc upgradeMonitor(
     trace "Connection upgrade failed", description = e.msg, conn
     libp2p_failed_upgrades_incoming.inc()
   finally:
+    await deadlineFut.cancelAndWait()
```
vladopajic, line 241:
```suggestion
    deadlineFut.cancel()
```

consider not doing `await` here, it is unnecessary to pause current execution to await on future which reference is expiring after few statements as well.

vladopajic (approved):

libp2p/switch.nim
vladopajic: disabled auto-merge in case ^

PR #2361, opened 2026-04-28
feat(kad): hide connection info

gmelodie:
## Summary

Make connection hiding configurable.

This PR adds a `hideConnectionInfo` configuration flag to the Kademlia DHT that, when enabled (default: `true`), causes all outbound protobuf-encoded messages to advertise `notConnected` for every peer's connection type instead of the actual local connection state.

This prevents the node from leaking internal topology information (i.e., which peers it is actively connected to) to the wider DHT network, acting as a privacy and security hardening measure.

---

## Changes

- **libp2p/protocols/kademlia/types.nim**
  - Added `hideConnectionInfo*: bool` field to `KadDHTConfig`
  - Added `hideConnectionInfo: bool = true` parameter to `KadDHTConfig.new()` and wired it into the constructor

- **libp2p/protocols/kademlia/protobuf.nim**
  - Added `hideConnection: bool = true` parameter to `encode*(peer: Peer, ...)`
    - When `true`, the serialized connection-type field is always `notConnected`
  - Added `hideConnection: bool = true` parameter to `encode*(msg: Message, ...)`
    - Propagated to both `closerPeers` and `providerPeers` encodings

- **libp2p/protocols/kademlia/find.nim**
  - Passed `kad.config.hideConnectionInfo` to `msg.encode()` in:
    - `dispatchFindNode` (outbound request)
    - `handleFindNode` (outbound response)

- **libp2p/protocols/kademlia/get.nim**
  - Passed `kad.config.hideConnectionInfo` to:
    - `msg.encode()` / `response.encode()` in `dispatchGetVal`
    - Both response paths in `handleGetValue`

- **libp2p/protocols/kademlia/ping.nim**
  - Passed `kad.config.hideConnectionInfo` to `encode()` in:
    - `ping` (request)
    - `handlePing` (response)

- **libp2p/protocols/kademlia/provider.nim**
  - Added `hideConnection: bool` parameter to `dispatchAddProvider`
  - Threaded `kad.config.hideConnectionInfo` through `addProvider`
  - Passed `kad.config.hideConnectionInfo` to `encode()` in:
    - `dispatchGetProviders`
    - `handleGetProviders`

- **libp2p/protocols/kademlia/put.nim**
  - Passed `kad.config.hideConnectionInfo` to `msg.encode()` in `handlePutValue`

- **tests/libp2p/kademlia/test_encoding.nim**
  - Updated existing round-trip tests to call `encode(hideConnection = false)`
  - Added new tests covering:
    - `hideConnection = true` always emits `notConnected`
    - `hideConnection = false` preserves actual type
    - Decoding all four `ConnectionType` values
    - Message-level hiding for both peer lists
    - Default `KadDHTConfig.hideConnectionInfo == true`

---

## Affected Areas

- Gossipsub  
- Transports  
- Peer Management / Discovery  
- Protocol Logic  
  - Kademlia DHT message encoding now conditionally suppresses connection-type metadata  
- Build / Tooling  
- Other  

---

- New `hideConnection` parameter on:
  - `encode(peer)`
  - `encode(msg)`
  - Defaults to `true`, changing default encoding behavior

- Previously:
  - Actual `ConnectionType` was always emitted

- Now:
  - `notConnected` is emitted unless explicitly overridden

- Decode path is unchanged:
  - All four `ConnectionType` values are still accepted inbound

- Downstream consumers (Nimbus, Waku, Codex):
  - Will observe `notConnected` in received peer records
  - This is spec-compliant (field is advisory)
  - Should verify routing-table logic does not depend on remote connection-type hints

---

## Impact on Library Users

- `KadDHTConfig`:
  - New field: `hideConnectionInfo: bool` (default: `true`)
  - Constructor updated with optional parameter (backward-compatible)

- Encoding API:
  - `encode*(peer: Peer, hideConnection: bool = true)`
  - `encode*(msg: Message, hideConnection: bool = true)`
  - Behavior change: default no longer preserves real connection type

- Internal API:
  - `dispatchAddProvider` gains positional `hideConnection: bool`
  - May break forks calling it directly

---

## Risk Assessment

- **Default behavior change in encode (medium)**  
  Old behavior exposed real `ConnectionType`; new default hides it.  
  Low risk for interoperability (field is advisory), but medium risk for internal consumers relying on it.  
  Mitigation: explicitly pass `hideConnection = false`.

- **Test suite validity (low)**  
  Existing tests updated to opt out of hiding. Intent preserved.

- **`dispatchAddProvider` signature change (low)**  
  Positional parameter addition may break external forks.  
  All in-tree usages updated; Nim compile-time checks mitigate risk.

- **Performance (negligible)**  
  Adds a single conditional branch per encoded peer.

---

## References

- N/A


libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -125,7 +126,9 @@ proc encode*(getAdsMsg: GetAdsMessage): ProtoBuffer {.raises: [], gcsafe.} =
   pb.finish()
   return pb
 
-proc encode*(msg: Message): ProtoBuffer {.raises: [], gcsafe.} =
+proc encode*(
+    msg: Message, hideConnection: bool = true
+): ProtoBuffer {.raises: [], gcsafe.} =
```
vladopajic, line 140: `encode()` with default `hideConnection` value is never used, better not have default value. 
also seems like important to declare each time.

vladopajic (commented):
proper name should be:
`ConnectionType` -> `ConnectionStatus` because it tells status connected/not connected, 
`ConnectionInfo` would be less optimal name, as `info` could represent anything.
`ConnectionType` is the least optimal name here

therefore: 
`hideConnection` -> `hideConnectionStatus`
`hideConnectionInfo` -> `hideConnectionStatus`

----

variable that represents the same thing should be used with same naming across code. that is `hideConnectionInfo` in case of this pr and not sometimes shorter name sometimes longer

vladopajic:
i understand that `ConnectionType` was added earlier, but suggested rename should be easy to do (it wont add more the 20 lines of code).

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -84,12 +84,13 @@ proc encode*(record: Record): ProtoBuffer {.raises: [].} =
...
 
-proc encode*(peer: Peer): ProtoBuffer {.raises: [].} =
+proc encode*(peer: Peer, hideConnection: bool = true): ProtoBuffer {.raises: [].} =
   var pb = initProtoBuffer()
   pb.write(1, peer.id)
   for address in peer.addrs:
     pb.write(2, address.data.buffer)
-  pb.write(3, uint32(ord(peer.connection)))
+  let connType = if hideConnection: ConnectionType.notConnected else: peer.connection
+  pb.write(3, uint32(ord(connType)))
```
vladopajic: `connType.hide(hideConnection)`

would be bit nicer with utility that encapsulates logic for "hiding"

`pb.write(3, connType.hide(hideConnection).ord.uint32)`

richard-ramos (approved):

vladopajic (approved):

PR #2363, opened 2026-04-29
refactor(connmanager): add limits config

vladopajic:
## Summary

- `ConnManager.new` previously had two mutually exclusive limit modes that were configured via arguments. this is error prone design and is avoided via `LimitsConfig` that can be created in only two ways `maxTotal` or  `maxInOut` constructors. 
this avoids:
  - `-1` values for connections limit arguments escaping outside of connection manager (they are still present but with limited scope - other components are not aware of that)
  - validation of arguments in different places (now it is only in  `maxTotal` and  `maxInOut` constructors)
  - if/else to decide what argument to pass to `ConnManager`

## Affected Areas

- [X] Connection Manager  
- [X] Switch Builder

## Impact on Library Users

`SwitchBuilder`'s  `withMaxIn(...)` and `withMaxOut(...)` were previously deprecated, but they actually need to go...
configuring values independently could be achieved but deciding to remove these since it should be trivial for library users to replace usage with `withMaxInOut(...)`

## Risk Assessment

low risk


libp2p/connmanager.nim
```diff
@@ -47,6 +47,14 @@ type
     gracePeriod*: Duration ## newly connected peers are exempt from trimming
     silencePeriod*: Duration ## minimum interval between trim cycles
 
+  LimitsConfig* = object
```
vladopajic, line 50: `LimitsConfig` or `ConnLimitsConfig`?

libp2p/builders.nim
```diff
@@ -261,37 +258,26 @@ proc withRng*(b: SwitchBuilder, rng: ref HmacDrbgContext): SwitchBuilder {.publi
...
-): SwitchBuilder {.public, deprecated: "Use withMaxInOut() instead".} =
-  ## Maximum concurrent incoming connections. Should be used with `withMaxOut<#withMaxOut,SwitchBuilder,int>`_
-  b.maxIn = maxIn
-  b
-
-proc withMaxOut*(
-    b: SwitchBuilder, maxOut: int
-): SwitchBuilder {.public, deprecated: "Use withMaxInOut() instead".} =
-  ## Maximum concurrent outgoing connections. Should be used with `withMaxIn<#withMaxIn,SwitchBuilder,int>`_
-  b.maxOut = maxOut
```
vladopajic, line 275: these must go.. it should be trivial for users  to replace with `withMaxInOut`

libp2p/builders.nim
```diff
@@ -595,9 +570,7 @@ proc newStandardSwitchBuilder*(
     secureManagers: openArray[SecureProtocol] = [SecureProtocol.Noise],
     inTimeout: Duration = 5.minutes,
     outTimeout: Duration = 5.minutes,
-    maxConnections = -1,
-    maxIn = -1,
-    maxOut = -1,
+    limits: Opt[LimitsConfig] = Opt.none(LimitsConfig),
```
richard-ramos, line 554: This proc is {.public.}, thus this change is a breaking change. We need a version with the prev parameters and a {.deprecated.} pragma.

Same situation happens in newStandardSwitch in 638

richard-ramos (changes requested):

richard-ramos (approved):

libp2p/builders.nim
richard-ramos: As discussed in channel, we'll move to new major version on next release, so this observation no longer applies

libp2p/builders.nim
```diff
@@ -261,37 +258,26 @@ proc withRng*(b: SwitchBuilder, rng: ref HmacDrbgContext): SwitchBuilder {.publi
...
-  b
-
-proc withMaxOut*(
-    b: SwitchBuilder, maxOut: int
-): SwitchBuilder {.public, deprecated: "Use withMaxInOut() instead".} =
-  ## Maximum concurrent outgoing connections. Should be used with `withMaxIn<#withMaxIn,SwitchBuilder,int>`_
-  b.maxOut = maxOut
+  ## Maximum concurrent connections of the switch. You should either use this,
+  ## or `withMaxInOut <#withMaxInOut,SwitchBuilder,int,int>`_.
+  b.limitsConfig = Opt.some(LimitsConfig.maxTotal(maxConnections))
```
gmelodie, line 264: Here we need to check for negative values on `maxConnections`, correct?

libp2p/builders.nim
```diff
@@ -261,37 +258,26 @@ proc withRng*(b: SwitchBuilder, rng: ref HmacDrbgContext): SwitchBuilder {.publi
...
 
 proc withMaxInOut*(
     b: SwitchBuilder, maxIn: int, maxOut: int
 ): SwitchBuilder {.public.} =
   ## Maximum concurrent incoming and outgoing connections.
-  doAssert maxIn > 0, "`maxIn` must be greater than 0"
-  doAssert maxOut > 0, "`maxOut` must be greater than 0"
-  b.maxIn = maxIn
-  b.maxOut = maxOut
+  b.limitsConfig = Opt.some(LimitsConfig.maxInOut(maxIn, maxOut))
```
gmelodie, line 269: Same as above

libp2p/connmanager.nim
gmelodie: To me `LimitsConfig` makes sense if it has information about other limits (not only connection limits). We can leave as is and include more limits as we need in the future?

tests/libp2p/test_conn_manager.nim
```diff
@@ -20,10 +20,16 @@ method newStream*(
   Connection.new(m.peerId, Direction.Out)
 
 proc newMaxTotal(maxConnections = 10, maxConnsPerPeer = 1): ConnManager =
```
gmelodie, line 22: Same as above: check for `limits > 0`

tests/libp2p/test_conn_manager.nim
```diff
@@ -20,10 +20,16 @@ method newStream*(
...
-  ConnManager.new(maxConnections = maxConnections, maxConnsPerPeer = maxConnsPerPeer)
+  ConnManager.new(
+    maxConnsPerPeer = maxConnsPerPeer,
+    limits = Opt.some(LimitsConfig.maxTotal(maxConnections)),
+  )
 
 proc newMaxInOut(maxIn: int, maxOut: int, maxConnsPerPeer = 1): ConnManager =
```
gmelodie, line 28: Same as above: check for `limits > 0`

gmelodie (approved):

libp2p/builders.nim
vladopajic: that's done inside `maxTotal`

libp2p/builders.nim
vladopajic: check is done inside `maxInOut`

tests/libp2p/test_conn_manager.nim
vladopajic: will be checked inside `maxTotal`

tests/libp2p/test_conn_manager.nim
vladopajic: check is done inside `maxInOut`

PR #2370, opened 2026-04-29
fix(pubsub): track pending sends without adding  backpressure

richard-ramos:
## Summary

This updates pubsub peer sending so pending transport writes are tracked by the peer queue.

Before this change, `sendMsg` started `conn.writeLp(msg)` but did not keep the returned future. This meant pubsub could think a peer had no pending send even when the transport was still writing. As a result, more messages could keep being sent to the same peer and build up inside the transport buffers.

This is especially problematic for QUIC, where bandwidth can look fine while message latency grows because messages spend more time waiting in hidden queues.

With this change:

- `sendMsg` returns a future that represents the actual write lifecycle.
- high-priority sends store that future in the peer queue when the write is still pending.
- `sendEncoded`, `broadcast`, and `publish` still get fire-and-forget behavior and do not wait for the transport write to finish.
- medium/low priority peer queue processing can now see pending high-priority sends and wait before sending more messages to that peer.

With this, pubsub will be aware of slow peer sends without adding backpressure to publishers or receive handlers.

## Affected Areas

- [x] Gossipsub  
  Pubsub peer send queue behavior is updated so pending sends are tracked correctly.
- [ ] Transports
- [x] Protocol Logic  
  Internal message scheduling between high, medium, and low priority pubsub messages is affected.
- [ ] Build / Tooling
- [ ] Other

## Impact on Library Users

No API changes.

`publish`, `broadcast`, and normal pubsub send paths remain fire-and-forget. This PR should not make application publishing wait for network writes.

The behavior change is internal: pubsub now tracks pending per-peer sends and avoids sending lower-priority messages to a peer while a high-priority send is still pending.

Expected impact:

- reduced hidden transport buffering for slow peers
- better queue visibility through existing pubsub peer queue metrics
- no migration required for downstream users

## Risk Assessment

Risk is moderate because this changes internal pubsub send scheduling.

Main behavior risk:

- medium and low priority messages may be delayed more consistently behind pending high-priority sends for the same peer.

This is intended, but it can affect timing-sensitive tests or workloads.

### KEEP THIS INTO ACCOUNT WHILE REVIEWING CODE:
Backpressure risk:
- this PR should not add caller backpressure.
- `sendEncoded`, `broadcast`, and `publish` still return without waiting for transport writes.
- receive handlers should not block on pubsub writes.

Network behavior risk:
- slow peers may see more delayed or dropped medium/low priority messages when their queues fill.
- this is preferable to silently accumulating latency inside transport buffers.

### DIFF
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -427,7 +427,7 @@ proc clearSendPriorityQueue(p: PubSubPeer) =
     )
 
 proc sendMsgContinue(conn: Connection, msgFut: Future[void]) {.async: (raises: []).} =
-  # Continuation for a pending `sendMsg` future from below
+  # Continuation for a pending transport write from below
   try:
     await msgFut
     trace "sent pubsub message to remote", conn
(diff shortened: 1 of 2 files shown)


gmelodie (approved):

vladopajic (approved):
nice!

PR #2387, opened 2026-05-01
feat(kad): add flag to disable bootstrapping

SionoiS:


richard-ramos (approved):

libp2p/protocols/kademlia/types.nim
```diff
@@ -354,3 +354,4 @@ type KadDHT* = ref object of LPProtocol
   dataTable*: LocalTable
   providerManager*: ProviderManager
   config*: KadDHTConfig
+  bootstrapping*: bool
```
gmelodie: `bootstrapping` should be `disableBootstrapping` (with `false` as default) and should be inside of `KadDHTConfig`

gmelodie (changes requested):
Curious as to why this is needed

SionoiS:
> Curious as to why this is needed

Because otherwise we can't isolate service discovery for testing.

gmelodie (approved):

PR #2394, opened 2026-05-01
feat(kad): add provider rejection and spillover

gmelodie:
## Summary

Adds overload protection for Kademlia provider records behind the `-d:kadProviderRejection` compilation flag.

- **Without the flag**: All existing behavior is preserved byte-for-byte.
- **With the flag enabled**:
  - Nodes can reject `ADD_PROVIDER` requests when a configurable per-key provider limit (`maxProvidersPerKey`) is reached.
  - Advertising nodes automatically backtrack to progressively farther peers ("spillover") when the closest candidates are full.
  - Ensures records can still be stored even on heavily loaded keys.

---

### `libp2p/protocols/kademlia/protobuf.nim`
- Added `AddProviderStatus` enum:
  - `accepted = 0`
  - `rejected = 1`
- Encoded in protobuf field **10** of the `Message` type.
- Added optional field:
  - `providerStatus: Opt[AddProviderStatus]`
- Encoding/decoding is always active for clean round-tripping, but runtime population only occurs under the flag.

### `libp2p/protocols/kademlia/types.nim`
- Added constant:
  - `DefaultMaxProvidersPerKey = 0` (0 = unlimited)
- Added config field:
  - `maxProvidersPerKey: int` in `KadDHTConfig`
- Includes documentation and constructor default.

### `libp2p/protocols/kademlia/find.nim`
- Added:
  - `allSortedPeers*(state: LookupState): seq[PeerId]`
- Exposes full iterative-lookup shortlist sorted by XOR distance.
- Used for spillover beyond top-k candidates.

### `libp2p/protocols/kademlia/provider.nim`
- Added module-level documentation for both modes (standard and overload-protection).
- Changed:
  - `dispatchAddProvider` return type:
    - From: `Result[void, string]`
    - To: `Result[AddProviderStatus, string]`
  - Under the flag:
    - Reads receiver response.
    - Falls back to `accepted` for backward compatibility.
- Modified `addProvider`:
  - Performs full `iterativeLookup`.
  - Builds larger candidate pool.
  - Sends `ADD_PROVIDER` in XOR-distance order.
  - Performs spillover when a batch is unanimously rejected.
- Modified `handleAddProvider`:
  - Enforces `maxProvidersPerKey`.
  - Exempts re-advertisements by known providers.
  - Sends acceptance/rejection responses.
  - Updates metrics.
- Added helper:
  - `sendAddProviderResponse` to centralize response writing.

### `libp2p/protocols/kademlia/kademlia_metrics.nim`
- Added metrics:
  - `kad_provider_rejections_sent`
  - `kad_provider_spillover_rounds`

### `tests/libp2p/kademlia/test_add_provider.nim`
- Added:
  - `{.define: kadProviderRejection.}` to enable flag during tests.
- Added five async tests:
  - Per-key limit rejection
  - Same-provider refresh exemption
  - Spillover to farther node
  - Normal replication when all nodes accept
  - Protobuf round-trip for `AddProviderStatus`

---

## Affected Areas

- Gossipsub
- Transports
- Peer Management / Discovery
- Protocol Logic:
  - Kademlia `ADD_PROVIDER` handling now supports optional rejection responses and spillover logic
- Build / Tooling
- Other

---

- Fully backward compatible.
- Protobuf field **10** is optional:
  - Not written without the flag.
  - Unflagged peers decode messages unchanged.
- `dispatchAddProvider` change:
  - Internal (not exported)
  - No external impact
- Flagged → unflagged interaction:
  - Missing response treated as `accepted`
  - Preserves interoperability

### Downstream Consumers

- **Unaffected unless opting in**:
  - Nimbus
  - Waku
  - Codex
- Opt-in users should:
  - Set `maxProvidersPerKey`
  - Validate behavior under rejection load

---

## Impact on Library Users

- `AddProviderStatus`:
  - New public enum (additive)
- `Message.providerStatus`:
  - New optional field
  - `none()` unless flag is active and message is a reply
- `KadDHTConfig.maxProvidersPerKey`:
  - Default: `0` (unlimited)
  - No effect without flag
- `KadDHTConfig.new`:
  - New optional parameter
- `LookupState.allSortedPeers()`:
  - New public method
- Metrics:
  - `kad_provider_rejections_sent`
  - `kad_provider_spillover_rounds`
- `dispatchAddProvider`:
  - Internal change only

---

### Iterative Lookup Overhead
- `addProvider` uses `iterativeLookup` under the flag.
- Adds extra round-trip and network traffic.
- Low risk:
  - Opt-in only
  - Same traversal already used by `getProviders`

### Response Handling
- Sender waits for response after `ADD_PROVIDER`.
- If receiver fails or is unflagged:
  - Read error treated as `accepted`
- Low risk:
  - Correct fallback
  - Tested under partial failure

### Test Flag Scope
- `{.define: kadProviderRejection.}` applies to entire compilation unit.
- All ~115 Kademlia tests run under the flag.
- Low risk:
  - Existing tests verified under flag

### Re-advertisement Exemption
- Providers can always refresh their own record.
- Limit applies to **distinct providers**, not messages.
- Low risk:
  - Behavior documented and tested

---

## References
https://github.com/vacp2p/roadmap/pull/445/changes#diff-40a4fb71e95fc9a8144b7c1f1e8ac53f49cb50ddf996d67c960b23d362b68bb1


tests/config.nims
```diff
@@ -17,6 +17,8 @@ import strutils, os
   libp2p_mplex_metrics
 --d:
   unittestPrintTime
+--d:
+  libp2p_kademlia_provider_rejection
```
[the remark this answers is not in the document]
gmelodie: This makes absolutely no sense whatsoever

libp2p/protocols/kademlia/provider.nim
```diff
@@ -168,20 +239,45 @@ proc manageExpiredProviders*(kad: KadDHT) {.async: (raises: [CancelledError]).}
...
-    if not PeerId.init(peer.id).isOk():
-      continue
-
+  let validPeers =
+    msg.providerPeers.filterIt(it.id == peerBytes and PeerId.init(it.id).isOk())
+
+  when defined(libp2p_kademlia_provider_rejection):
+    kad.config.maxProvidersPerKey.withValue(limit):
+      let existingProviders =
+        kad.providerManager.knownKeys.getOrDefault(msg.key, initHashSet[Provider]())
```
richard-ramos, line 267: This made me realize that we need to create a hasher for `Provider`, otherwise we risk having duplicates if for example we create two providers with same id but different addresses!
```
proc hash*(provider: Provider): Hash =
   hash(provider.id)

proc `==`*(a, b: Provider): bool =
  a.id == b.id
```

libp2p/protocols/kademlia/provider.nim
```diff
@@ -121,23 +131,84 @@ proc dispatchAddProvider(
...
+      return ok(AddProviderStatus.accepted)
+
+    return ok(reply.providerStatus.get(AddProviderStatus.accepted))
+  else:
+    return ok(AddProviderStatus.accepted)
+
+proc sendBatch(kad: KadDHT, peers: seq[PeerId], key: Key): auto =
+  peers.mapIt(
+    kad.switch.dispatchAddProvider(
+      it, key, kad.codec, kad.config.hideConnectionStatus, kad.config.timeout
```
richard-ramos: Should this timeout here be smaller than the timeout on line 197? or timeout in line 197 larger? because otherwise it will cause unnecessary spillovers for cases in which you have a mixed network of peers that use spillovers and those that do not.

If you have a peer that does not use `libp2p_kademlia_provider_rejection`, they are not going to send a ADD_PROVIDER response, so the code will wait for the peer timeout and then treat the request as accepted, but the batch timeout can fire before the peer timeout finishes, so `countResults()` will see that future as unfinished and skip it,  so the `stored` might be to low and continue to do spillover even tho that peer might have succesfully stored the record.

Maybe since each request already has a timeout, the spillover logic should wait for requests futures to finish before calling `countResults()`?

This might be and edge case that is rare in practice tho

richard-ramos (commented):
Is there a spec PR proposed for libp2p/spec to discuss this extension?

libp2p/utils/future.nim
```diff
@@ -38,3 +38,11 @@ template newFutureCompleted*[T](): auto =
...
   for fut in futs:
     fut.cancelSoon()
+
+proc awaitBatch*[Fut](
+    rpcBatch: seq[Fut], timeout: Duration
+) {.async: (raises: [CancelledError]).} =
+  try:
+    await rpcBatch.allFutures().wait(timeout)
+  except AsyncTimeoutError:
+    discard
```
vladopajic: since this is defined as general utility, then `Batch` part here is misleading, because not all futures will be batch of something

since this utility just calls `allFutures` then name of should keep that as base. and since it is swallowing error then it better reflect that as well.

 ```
allFuturesWaitOrTimeout  
allFuturesWaitIgnoreTimeout 
allFuturesWithTimeoutNoThrow
```

or something under those lines....

File: libp2p/protocols/kademlia/put.nim [hunk 5]
```diff
@@ -81,13 +82,9 @@ proc putValue*(
...
-    let rpcBatch = chunk.mapIt(kad.switch.dispatchPutVal(it, key, value, kad.codec))
-    try:
-      await rpcBatch.allFutures().wait(kad.config.timeout)
-    except AsyncTimeoutError:
-      debug "One or more PutValue messages timed out"
-      # Dispatch will timeout if any of the calls don't receive a response (which is normal)
-      discard
+    await chunk.mapIt(kad.switch.dispatchPutVal(it, key, value, kad.codec)).awaitBatch(
+      kad.config.timeout
+    )
```
vladopajic: this would be much nicer in two lines. 
it would be easier on reader to understand it.
it would be still less or equal lines then current code (3 lines, with ` kad.config.timeout` taking whole line :( ).

libp2p/protocols/kademlia/get.nim
```diff
@@ -139,11 +140,7 @@ proc getValue*(
       # value is invalid or not best
       rpcBatch.add(kad.switch.dispatchPutVal(p, key, best.value, kad.codec))
 
-  try:
-    await rpcBatch.allFutures().wait(chronos.seconds(5))
-  except AsyncTimeoutError:
-    # Dispatch will timeout if any of the calls don't receive a response (which is normal)
-    discard
+  await rpcBatch.awaitBatch(chronos.seconds(5))
```
vladopajic: nit: `5.seconds` we use this pattern in general through codebase

but real question is where does 5 come from?
most of code is using some value for config. if this should not be the case then that should be documented with reasoning why timeout here differs. would be even nicer to have const and document that const.

gmelodie: Nice find, should be `kad.config.timeout`

libp2p/utils/future.nim
```diff
@@ -38,3 +38,11 @@ template newFutureCompleted*[T](): auto =
 template cancelSoon*[T](futs: seq[T]) =
   for fut in futs:
     fut.cancelSoon()
+
+proc allFuturesWaitOrTimeout*[Fut](
+    rpcBatch: seq[Fut], timeout: Duration
```
vladopajic: `rpcBatch` -> `futs`

libp2p/protocols/kademlia/provider.nim
```diff
@@ -121,23 +131,93 @@ proc dispatchAddProvider(
...
+          rejected.inc()
+    (accepted, rejected)
+
+  proc addProviderSpillover(
+      kad: KadDHT, key: Key
+  ) {.async: (raises: [CancelledError]).} =
+    let noReply = proc(
+        peerId: PeerId, msgOpt: Opt[Message], state: var LookupState
+    ): Future[void] {.async: (raises: []), gcsafe.} =
       discard
```
vladopajic: this can be defined as general callback next to `iterativeLookup` so it can be reused.

libp2p/protocols/kademlia/provider.nim
```diff
@@ -121,23 +131,93 @@ proc dispatchAddProvider(
...
+    let stop = proc(state: LookupState): bool {.raises: [], gcsafe.} =
+      state.hasResponsesFromClosestAvailable()
```
vladopajic: in relation to previews comment, this as well.

libp2p/protocols/kademlia/provider.nim
```diff
@@ -121,23 +131,93 @@ proc dispatchAddProvider(
...
+    let dispatchFind = proc(
+        kad: KadDHT, peer: PeerId, target: Key
+    ): Future[Result[Message, string]] {.
+        async: (raises: [CancelledError]), gcsafe, closure
+    .} =
+      return await dispatchFindNode(kad, peer, target)
```
vladopajic: same here

vladopajic (approved):
in continuation to the [comment](https://github.com/vacp2p/nim-libp2p/pull/2391#pullrequestreview-4214715787), the pr has the same _smell_ - why having compile time flag if same could be achieved with `KadDHTConfig.providerRejection`? 

- this pr introduces about 400 loc; compiling difference is nothing
- configuration field gives advantage: it is possible to create unit test (aka component tests) that can create two kad configs, one with other without support, and assert behavior. because you can't really create same binary with two different configurations via compile time flags that easily. 
- it's not just interaction with two components; but also test separately with future and without; sure it is possible to do that with compile time flags, but at which cost?
- it seems that practice of using compile time flags is habit that should be broken; by going with non compile time configs by default. there are cases when compile time flags could be more beneficial, and we should come up with framework how to make that decisions; the "feature gating policy":
   - is feature experimental: could it be regular config, still?
   - does feature introduce new dependencies: experimental feature that introduce new dependency like quic was at the time, can be under compile flag to guard downstream dependencies from unnecessary nimble headache
   - is it expected form downstream users to use this? it is okay to be compile time flag if we don't expect
   - etc...

please note that if compile time flags usage explode it will be harder to test, maintain and use this repository.

libp2p/protocols/kademlia/provider.nim
```diff
@@ -121,23 +133,79 @@ proc dispatchAddProvider(
...
-proc addProvider*(kad: KadDHT, key: Key) {.async: (raises: [CancelledError]), gcsafe.} =
-  ## Find the closest nodes to the key via FIND_NODE and send ADD_PROVIDER with self's peerInfo to each of them
 
-  let peers = await kad.findNode(key)
-  for chunk in peers.toChunks(kad.config.alpha):
-    let rpcBatch = chunk.mapIt(
-      kad.switch.dispatchAddProvider(
-        it, key, kad.codec, kad.config.hideConnectionStatus
-      )
+  if providerRejection:
```
vladopajic: code can now avoid indentation here:
```
if not providerRejection:
   return ok(AddProviderStatus.accepted)

...
```

richard-ramos:
> why having compile time flag if same could be achieved with `KadDHTConfig.providerRejection`?

I propose this rule to be added to the instructions https://github.com/vacp2p/nim-libp2p/pull/2416

richard-ramos (approved):

PR #2406, opened 2026-05-04
feat(service-disco): Local registrar access

SionoiS:
Advertiser and discoverer now access the local registrar for either ad placement or search.

### DIFF
--- a/libp2p/protocols/service_discovery/connection.nim
+++ b/libp2p/protocols/service_discovery/connection.nim
@@ -10,7 +10,7 @@ import ./[types, service_discovery_metrics, registrar]
 logScope:
   topics = "service-disco connection"
 
-proc sendRemoteMessage*(
+proc send*(
     disco: ServiceDiscovery, peerId: PeerId, msg: Message
 ): Future[Result[Message, string]] {.async: (raises: [CancelledError]), gcsafe.} =
   let addrs = disco.switch.peerStore[AddressBook][peerId]
--- a/libp2p/protocols/service_discovery/connection.nim
+++ b/libp2p/protocols/service_discovery/connection.nim
@@ -50,7 +50,7 @@ proc send*(
 
   return ok(reply)
 
-proc handleServiceDiscoMessage*(
+proc handleMessage*(
     disco: ServiceDiscovery, conn: Connection, msg: Message
 ) {.async: (raises: [CancelledError]).} =
   cd_messages_received.inc(labelValues = [$msg.msgType])
--- a/libp2p/protocols/service_discovery/connection.nim
+++ b/libp2p/protocols/service_discovery/connection.nim
@@ -1,0 +1,72 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Copyright (c) Status Research & Development GmbH
+
+import chronos, chronicles, results
+import ../../[peerid, switch, multiaddress, extended_peer_record]
+import ../kademlia
+import ../kademlia/types
+import ./[types, service_discovery_metrics, registrar]
+
+logScope:
+  topics = "service-disco connection"
+
+proc sendRemoteMessage*(
+    disco: ServiceDiscovery, peerId: PeerId, msg: Message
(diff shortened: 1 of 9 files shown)


SionoiS:
WIP working on tests

SionoiS:
I separated the connection from the protocol logic. This pattern could be also used elsewhere too, but it was out of scope.

libp2p/protocols/service_discovery/connection.nim
```diff
@@ -0,0 +1,72 @@
...
+import chronos, chronicles, results
+import ../../[peerid, switch, multiaddress, extended_peer_record]
+import ../kademlia
+import ../kademlia/types
+import ./[types, service_discovery_metrics, registrar]
+
+logScope:
+  topics = "service-disco connection"
+
+proc sendRemoteMessage*(
```
gmelodie: nit: maybe 
```suggestion
proc send*(
```

libp2p/protocols/service_discovery/connection.nim
```diff
@@ -0,0 +1,72 @@
...
+
+  cd_messages_received.inc(labelValues = [$msg.msgType])
+  cd_message_bytes_received.inc(replyBuf.len.float64, labelValues = [$msg.msgType])
+
+  let reply = Message.decode(replyBuf).valueOr:
+    return err("failed to decode message response: " & $error)
+
+  return ok(reply)
+
+proc handleServiceDiscoMessage*(
```
gmelodie: nit
```suggestion
proc handleMessage*(
```

gmelodie (approved):

libp2p/protocols/service_discovery/connection.nim
SionoiS: :thinking: but one can also send a message locally now...

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -412,22 +410,29 @@ proc tInitOrDefault(ticket: Opt[Ticket], default: Moment): Moment =
...
+
+  let regMsg = inMsg.register.valueOr:
+    error "no register message"
+    return msg
 
   let ad = validateRegisterMessage(regMsg, serviceId).valueOr:
-    await sendRegisterResponse(
-      conn, kademlia_protobuf.RegistrationStatus.Rejected, closerPeers
-    )
-    cd_register_requests.inc(
```
richard-ramos, line 427: This metric is no longer needed in registration*?

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -208,10 +177,14 @@ proc lookup*(
     if bucket.peers.len == 0:
       continue
 
+    var peers = disco.peersToQuery(bucket)
+
+    if once:
+      peers.add(disco.switch.peerInfo.peerId)
```
richard-ramos, line 183: This will cause the switch peerid to be added twice, right?

`acceptAdvertisement` will put the switch peer ID in the service routing table in line 390 when called in `libp2p/protocols/service_discovery/advertiser.nim` line 177, . So `peersToQuery(bucket)` may already include `disco.switch.peerInfo.peerId`. This line always adds it again.

That means lookup can ask the local registrar two times and return the same local ad duplicated.  

Could we only add the local peer if it is not already in `peers`, or remove duplicate ads before returning them?

SionoiS: A node should never end up in any of it's own tables. I think we have checks for that. Let me verify!

edit: indeed we always prevent that case
https://github.com/vacp2p/nim-libp2p/blob/0ac30a9413923878a9c592bfc5681ac8708636e0/libp2p/protocols/kademlia/routing_table.nim#L79

libp2p/protocols/service_discovery/registrar.nim
SionoiS: nice catch! I'll add this metric back

richard-ramos (approved):

PR #2378, opened 2026-04-30
chore: extract mix to logos-co/nim-libp2p-mix

chaitanyaprem:
## Summary

Mix has been extracted out of nim-libp2p into its own repository at
[**logos-co/nim-libp2p-mix**](https://github.com/logos-co/nim-libp2p-mix)
so it can evolve independently of the core libp2p library. This PR
removes the mix code from nim-libp2p.

## What's removed

- **Source**: \`libp2p/protocols/mix/\` (tree) and \`libp2p/protocols/mix.nim\` (facade)
- **Tests**: \`tests/libp2p/mix/\`
- **Example**: \`examples/mix_ping.nim\`
- **Peer store**: \`MixPubKeyBook\` from \`libp2p/peerstore.nim\` (only used by mix)
- **C bindings**: 7 \`libp2p_mix_*\` exported procs, related types/fields, and request processors in \`cbind/\`
- **Compile flag**: \`-d:libp2p_mix_experimental_exit_is_dest\` from the default test cfg (the new repo sets it in its own nimble file)
- **Docs**: \`docs/protocols_mix.md\`, \`docs/protocols_mix_spam_protection.md\`, README & copilot-instructions entries

Net change: **49 files deleted, 16 modified (−538 / +6).**

## Consumer migrations (already up)

These should land **before** this PR merges, otherwise consumers break:

- [\`logos-co/mix-rln-spam-protection-plugin#6\`](https://github.com/logos-co/mix-rln-spam-protection-plugin/pull/6) — switch import to \`libp2p_mix/spam_protection\`
- [\`logos-messaging/logos-delivery#3842\`](https://github.com/logos-messaging/logos-delivery/pull/3842) — switch all waku imports, add \`libp2p_mix\` nimble dep, add new submodule

## Verification

- \`nim check\` on \`libp2p.nim\` and \`cbind/libp2p.nim\` after deletion: clean ✓
- 5-node mixnet simulation in logos-delivery (chat2mix + lightpush) runs end-to-end with the new \`libp2p_mix\` package ✓

## History preservation

Full mix git history is preserved in \`logos-co/nim-libp2p-mix\` with original
authors, dates, and PR references rewritten as cross-repo links to
\`vacp2p/nim-libp2p#NNNN\`.

## Test plan

- [x] CI green on this branch (full test suite + interop)
- [x] Close LIONESS PR #2349 with a pointer to logos-co/nim-libp2p-mix#mix-payload-lioness

PR #2407, opened 2026-05-05
fix(daily): cover traffic leaking channels

gmelodie:
## Summary
Cover traffic integration tests were leaking channels. The `slot exhaustion` test now stops mix nodes before destNode. Previously (LIFO defer order), destNode stopped first, leaving in-flight exit→destNode channels in a partially-open state when the mix nodes' switches finally cleaned up.

## References
- #2404


vladopajic (approved):
FYI #2378 mix is not here anymore

richard-ramos (approved):

Issue #2408, opened 2026-05-05
SwitchBuilder.new() does not initialize b.rng


richard-ramos:
https://github.com/vacp2p/nim-libp2p/pull/2398#discussion_r3189372074
> SwitchBuilder.new() does not initialize b.rng, but withAutotls now immediately passes b.rng into AutotlsService.new. This makes calling SwitchBuilder.new().withAutotls(...) (or any builder method that uses b.rng before build()) prone to nil dereference / Defect. Consider initializing b.rng in SwitchBuilder.new() (e.g., to newRng()) or asserting b.rng is non-nil in withAutotls/withRendezVous/etc. with a clear error message.

Ideally it shouldnt matter the order in which you put the build options, i,e,:
```nim
.withRng()
.withAutoTLS()

# vs 

.withAutoTLS()
.withRng()
```

The PR in which **copilot** raised that comment, it was fixed by adding an assertion that forces the dev to use withRng() before the autotls call, since AutoTLS receives the rng as a constructor option. To avoid this, it should either have an initialization function that sets up the rng in the `build()` proc, or, we initialize autotls inside the `build()` proc, instead of doing it in the option.

PR #2410, opened 2026-05-05
fix(autotls): withAutotls independent ordering

gmelodie:
## Summary
`.withAutotls()` no longer depends on `withRng` being declared first, rng is set on `build`.

## References
#2408


richard-ramos (approved):
Could we do the same with rendezvous?

https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/builders.nim#L340

vladopajic (approved):

PR #2412, opened 2026-05-05
feat: identify pusher

vladopajic:
## Summary

this pr is main effort for adding auto identify push on peer changes. 
currently peer broadcasts new peer info when new protocol is mounted via `mount` or when `swtich.updateAddrs()` is called. 

following prs could address improvements for identify push to happen really automatically, avoiding need for n `swtich.updateAddrs()` 

## Affected Areas

- [x] Switch  
  

## Impact on Library Users

identify push is currently by default added to switch. all peers will have this.

## Risk Assessment

low.

## References

`ift-ts:p2p:ift:2026q2-nimlibp2p-maintenance:identify-push`

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -474,6 +474,7 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError].} =
   )
 
   switch.mount(identify)
+  switch.setupIdentifyPush()
 
   if not isNil(b.autonatV2Client):
     b.autonatV2Client.setup(switch)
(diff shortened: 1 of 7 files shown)


libp2p/builders.nim
```diff
@@ -474,6 +474,7 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError].} =
   )
 
   switch.mount(identify)
+  switch.setupIdentifyPush()
```
gmelodie, line 478: Would it make sense to have `setupIdentifyPush` also mount `identify`?

gmelodie (approved):

libp2p/builders.nim
vladopajic: not really, `Identify` and `IdentifyPush` are different protocols

vladopajic (commented):
test

richard-ramos (approved):

PR #2421, opened 2026-05-06
feat(PeerInfo): add observer

vladopajic:
## Summary

this pr adds an observer-like pattern for PeerInfo changes. since there are many places where peer info is updated, and the identify pusher needs the latest peer info each time it pushes to peers, the most flexible approach is to listen for peer info changes - wherever they originate now or in the future.

## Affected Areas

- [X] PeerInfo
- [x] IdentifyPusher  
 

internal changes only.

## Impact on Library Users

not impact, internal changes only.

## Risk Assessment

low.

## References

continuation of effort for: https://github.com/vacp2p/nim-libp2p/pull/2412

### DIFF
--- a/libp2p/identify_pusher.nim
+++ b/libp2p/identify_pusher.nim
@@ -39,6 +39,7 @@ type
     peerInfo: PeerInfo
     onIdentifiedHandler: PeerEventHandler
     onLeftHandler: PeerEventHandler
+    onPeerInfoUpdated: PeerInfoObserver
 
 proc new*(
     T: type IdentifyPusher,
@@ -94,6 +95,13 @@ proc start*(p: IdentifyPusher) =
   ## Construct the IdentifyPush protocol and register the connection manager
   ## hooks that maintain `pushPeers`. The caller is responsible for mounting
   ## `p.identifyPush` on a Switch.
+
+  proc onPeerInfoUpdated(peerInfo: PeerInfo) {.gcsafe, raises: [].} =
+    p.broadcast()
+
+  p.onPeerInfoUpdated = onPeerInfoUpdated
+  p.peerInfo.addObserver(p.onPeerInfoUpdated)
+
   proc onIncomingPush(peerId: PeerId, info: IdentifyInfo) {.async.} =
     p.peerStore.updatePeerInfo(info)
     if IdentifyPushCodec in info.protos:
@@ -121,14 +129,16 @@ proc start*(p: IdentifyPusher) =
 
 proc stop*(p: IdentifyPusher) {.async: (raises: []).} =
   ## Cancel any in-flight pushes and unregister connection manager hooks.
-  if p.onIdentifiedHandler != nil:
+  if not p.onIdentifiedHandler.isNil:
     p.connManager.removePeerEventHandler(
       p.onIdentifiedHandler, PeerEventKind.Identified
     )
     p.onIdentifiedHandler = nil
-  if p.onLeftHandler != nil:
+  if not p.onLeftHandler.isNil:
     p.connManager.removePeerEventHandler(p.onLeftHandler, PeerEventKind.Left)
     p.onLeftHandler = nil
+  if not p.onPeerInfoUpdated.isNil:
+    p.peerInfo.removeObserver(p.onPeerInfoUpdated)
 
   let pending = move(p.ongoingSend)
   pending.cancelSoon()
--- a/libp2p/identify_pusher.nim
+++ b/libp2p/identify_pusher.nim
@@ -32,8 +32,9 @@ type
 
   IdentifyPusher* = ref object
     identifyPush*: IdentifyPush
-    pushPeers*: HashSet[PeerId]
-    ongoingSend*: seq[PushSendFut]
+    pushPeers: HashSet[PeerId]
+    started: bool
+    ongoingSend: seq[PushSendFut]
     connManager: ConnManager
     peerStore: PeerStore
     peerInfo: PeerInfo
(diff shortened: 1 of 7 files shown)


gmelodie (approved):
I don't like the `observer` name though. To me it seems like a struct/object rather than a function, so it took me a bit to understand:

```nim
proc update*(p: PeerInfo) {.async: (raises: [CancelledError]).} =
  defer:
    for observer in p.observers:
      observer(p)
```

Maybe another name would be preferrable, but I couldn't think of something better to suggest

vladopajic:
> I don't like the `observer` name though. 

that's the name of pattern https://en.wikipedia.org/wiki/Observer_pattern

libp2p/peerinfo.nim
```diff
@@ -75,6 +83,14 @@ proc update*(p: PeerInfo) {.async: (raises: [CancelledError]).} =
     info "Can't update the signed peer record"
     return
```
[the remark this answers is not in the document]
vladopajic: this is not early exit, data was already initialized.

libp2p/identify_pusher.nim
```diff
@@ -121,14 +129,16 @@ proc start*(p: IdentifyPusher) =
...
     p.connManager.removePeerEventHandler(
       p.onIdentifiedHandler, PeerEventKind.Identified
     )
     p.onIdentifiedHandler = nil
-  if p.onLeftHandler != nil:
+  if not p.onLeftHandler.isNil:
     p.connManager.removePeerEventHandler(p.onLeftHandler, PeerEventKind.Left)
     p.onLeftHandler = nil
+  if not p.onPeerInfoUpdated.isNil:
+    p.peerInfo.removeObserver(p.onPeerInfoUpdated)
```
richard-ramos, line 154:
```suggestion
  if not p.onPeerInfoUpdated.isNil:
    p.peerInfo.removeObserver(p.onPeerInfoUpdated)
    p.onPeerInfoUpdated = nil
```

(to match the style of the other handlers being niled after they're removed)

libp2p/peerinfo.nim
```diff
@@ -64,6 +68,10 @@ proc expandAddrs*(
   p.addressPolicy.filterAddrs(addrs)
 
 proc update*(p: PeerInfo) {.async: (raises: [CancelledError]).} =
+  defer:
+    for observer in p.observers:
```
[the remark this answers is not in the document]
vladopajic: i dont see this ever being the issue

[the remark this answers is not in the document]
vladopajic: @copilot add test case that ensures that observers are not changed via add / remove while from observe callback

libp2p/peerinfo.nim
```diff
@@ -64,6 +68,10 @@ proc expandAddrs*(
   p.addressPolicy.filterAddrs(addrs)
 
 proc update*(p: PeerInfo) {.async: (raises: [CancelledError]).} =
```
richard-ramos, line 71: With this change you'd be sending addrs updates to all peers regardless of the address really changing or not, like if you returned  in line 84 OR  In the case of autonat v1 and v2 this is annoying because of code like this https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/connectivity/autonatv2/service.nim#L144 which updates the peerInfo periodically, asking multiple peers in short sucession

So, it probably makes sense to check if `p.addr` really changed after applying the mappers in line 75 instead of this happening on a defer,
i.e:

```nim
  let oldAddrs = p.addrs

  p.addrs = p.listenAddrs
  for mapper in p.addressMappers:
    p.addrs = await mapper(p.addrs)

  p.addrs = p.addressPolicy.filterAddrs(p.addrs)

  p.signedPeerRecord = SignedPeerRecord.init(
    p.privateKey, PeerRecord.init(p.peerId, p.addrs)
  ).valueOr:
    info "Can't update the signed peer record"
    return
  
  if p.addrs != oldAddrs:
       for observer in p.observers:
          observer(p)
```

(might require converting oldAddrs to hashset and p.addr in the comparison for checking equality)

libp2p/peerinfo.nim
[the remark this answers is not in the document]
vladopajic: this doesn't make sense actually....

libp2p/identify_pusher.nim
vladopajic: already done

libp2p/peerinfo.nim
richard-ramos: Actually... looking more into this, i'm not even sure that the current code in master is correct.  It seems to me that we should update `p` only if the signed record was updated  correctly :thinking: 

i.e. perhasp shoulb be something like this:
```nim

  let oldAddrs = p.addrs

  var newAddrs = p.listenAddrs
  for mapper in p.addressMappers:
    newAddrs = await mapper(newAddrs)

  newAddrs= p.addressPolicy.filterAddrs(newAddrs)

  p.signedPeerRecord = SignedPeerRecord.init(
    p.privateKey, PeerRecord.init(p.peerId, newAddrs)
  ).valueOr:
    info "Can't update the signed peer record"
    return

 p.addrs = newAddrs
  
  if newAddrs != oldAddrs:
       for observer in p.observers:
          observer(p)
```

richard-ramos (approved):

libp2p/peerinfo.nim
vladopajic: i'll look into that separately. currently i don't see how could this ever fail, since this is only internal data. maybe one of those 1) if private key is not set, or peer id is nil, or some similar condition that should never happen at this stage.

vladopajic:
dropping this pr in favor of #2436

PR #2424, opened 2026-05-06
fix(service-disco): startDiscovering not sending messages

### LINKED ISSUE #2417: service-disco: what is `startDiscovering` meant to do?
I'm a bit confused about the discovery API.

`startDiscovering` only registers an `Interest` entry in `rtManager` and seeds a per-service routing table from the main one. No messages go out. The refresh loop runs `findNode` for registered services but never `GET_ADS`, so no ads get fetched. Only lookup does that, via `collectBucketAds`.

Two things I'm unsure about.

`lookup` itself calls `addService(..., Interest)` and the entry persists. There's no `stopLookup`, so a lookup-only caller leaves an entry the refresh loop keeps maintaining, removable only via `stopDiscovering`. Intended?

The name reads to me like it should kick off background discovery I can read later, similar to `startProviding` in plain Kad DHT immediately calling `addProvider`. But it sends nothing and there's no accessor for accumulated ads. The spec also only describes a one-shot `LOOKUP` with no persistent subscription.

How it's supposed to work?

gmelodie:
## Summary

`startDiscovering` was silently broken: it registered an Interest routing table but never launched a background task to actually fetch advertisements from registrars, so callers who relied on it to populate results would get nothing back on lookup.

This PR wires up the missing background discovery loop, adds a `Discoverer` object (with a running-task set and an ad cache), makes lookup cache-aware when a discovery session is active, and cleans up the one-shot lookup path so it no longer leaves orphaned routing tables behind.

### Changes

- **libp2p/protocols/service_discovery/types.nim**
  - Added `DiscoverTask` ref object (future + `serviceId`) and `Discoverer` ref object (running task set + ad cache).
  - Added `Discoverer.new()` constructor and `hash(DiscoverTask)`.
  - Renamed `hashServiceId(string)` → `toServiceId(ServiceInfo)` and added `==` overloads between `ServiceInfo` and `ServiceId` for ergonomic comparisons.
  - Added `discoverer` field to `ServiceDiscovery`.
  - Removed unused `stew/byteutils` import and unused `toKey` proc.

- **libp2p/protocols/service_discovery/discoverer.nim**
  - Added `mergeAds` helper that deduplicates by `PeerId`, keeping the highest `seqNo`.
  - Extracted `fetchRemoteAds` from the inline lookup body so it can be shared between the background loop and one-shot lookups.
  - Added `discoverFromRegistrar` background coroutine (using heartbeat) that periodically refreshes the cache.
  - Rewrote `startDiscovering` (now `addDiscoveredService` + thin `startDiscovering` shim): spawns `discoverFromRegistrar` and tracks it in `discoverer.running`.
  - Rewrote `stopDiscovering` to async: cancels and awaits the running task, clears the cache entry, removes the routing table.
  - Rewrote `lookup`:
    - Returns from cache immediately when a discovery session is active and cache is full.
    - Otherwise performs a one-shot remote fetch and defers routing-table removal on exit.

- **libp2p/protocols/service_discovery/advertiser.nim**
  - Updated `hashServiceId` call sites to `toServiceId`.

- **libp2p/protocols/service_discovery.nim**
  - Initializes `Discoverer.new()` in `ServiceDiscovery.new()`.
  - `stop` now cancels all running discoverer tasks.
  - `lookup(ServiceInfo)` updated to use `toServiceId` instead of `hashServiceId`.

- **tests/libp2p/service_discovery/test_discoverer.nim**
  - Converted sync tests that called `startDiscovering`/`stopDiscovering` to `asyncTest`.
  - Asserted that one-shot lookup does not leave routing tables behind (inverted previous assertions).
  - Added new test: *"lookup returns cache when startDiscovering is active"*.

- **tests/libp2p/service_discovery/test_advertiser.nim**
  - Updated all `hashServiceId` call sites to `toServiceId`.

- **tests/libp2p/service_discovery/component/test_advertise_discover.nim**
  - Updated `startDiscovering`/`stopDiscovering` call signatures.
  - Added `checkUntilTimeout` assertion ensuring the discoverer cache is populated before calling `lookup`.

- **tests/libp2p/service_discovery/component/test_lookup_get_ads.nim** & `test_register.nim`
  - Updated `hashServiceId` call sites to `toServiceId(ServiceInfo(...))`.

- **tests/libp2p/service_discovery/test_registrar.nim**
  - Updated remaining `hashServiceId` call site to `toServiceId`.

---

## Affected Areas

- Gossipsub  
- Transports  
- Peer Management / Discovery  

Two-stage change:
- Fixes the broken background discovery loop.
- Makes `lookup` cache-aware.

- **Protocol Logic**
  - `lookup` control flow changed; one-shot path now cleans up its routing table on exit.

- **Build / Tooling**
- **Other**

---

`toServiceId(ServiceInfo)` produces the same SHA-256 digest as the old `hashServiceId(string)` when called with the same `service.id` string, so wire format and stored keys are identical.

Breaking changes:
- `startDiscovering` / `stopDiscovering` now take `ServiceInfo` instead of `string`.
- `stopDiscovering` is now async.

Downstream integrations (Nimbus, Waku, Codex) should be audited for:
- Passing raw strings instead of `ServiceInfo`.
- Missing `await` on `stopDiscovering`.

---

## Impact on Library Users

- `startDiscovering(service: ServiceInfo): bool`  
  - Breaking: renamed from `startDiscovering(serviceId: string)`

- `stopDiscovering(service: ServiceInfo): Future[void]`  
  - Breaking: was sync, now async

- `Discoverer`, `DiscoverTask`  
  - New exported types

- `ServiceDiscovery.discoverer`  
  - New exported field

- `toServiceId(ServiceInfo): ServiceId`  
  - New replacement for removed `hashServiceId(string)`

- `hashServiceId`  
  - Removed

- `toKey(ServiceInfo): Key`  
  - Removed (unused)

- `==` overloads for (`ServiceInfo`, `ServiceId`)  
  - Added

---

## Risk Assessment

- **One-shot lookup routing table cleanup**
  - Previously retained indefinitely; now removed on exit via `defer`.
  - Risk: low (table rebuilt on next call; avoids stale/wasteful state).

- **`mergeAds` deduplication (PeerId / seqNo)**
  - Lower `seqNo` ads dropped if out of order.
  - Matches registrar logic.
  - Risk: low.

- **`stopDiscovering` now async**
  - Callers not awaiting may leave dangling tasks.
  - Risk: medium.

## References
Fixes #2417


libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -135,56 +136,132 @@ proc collectBucketAds(
...
+proc mergeAds(
+    existing, incoming: seq[Advertisement]
+): seq[Advertisement] {.raises: [].} =
+  var best: Table[PeerId, Advertisement]
+  for ad in existing & incoming:
+    let pid = ad.data.peerId
+    best.withValue(pid, curr):
+      if ad.data.seqNo > curr[].data.seqNo:
+        curr[] = ad
+    do:
```
vladopajic: i can swallow `withValue`, but  `withValue` + `do` is so ugly construct

libp2p/protocols/service_discovery/types.nim
```diff
@@ -135,13 +147,22 @@ proc encode*(ads: seq[Advertisement], fReturn: int): seq[seq[byte]] {.raises: []
     adBytes.add(encoded)
   adBytes
 
-proc hashServiceId*(serviceStr: string): ServiceId =
-  let digest = sha256.digest(serviceStr)
+proc toServiceId*(service: ServiceInfo): ServiceId =
```
vladopajic: `toServiceId` ->

`deriveServiceId` / `computeServiceId` / `makeServiceId`

vladopajic (approved):

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -135,56 +136,129 @@ proc collectBucketAds(
...
+
+  return found
+
+proc discoverFromRegistrar(
+    disco: ServiceDiscovery, serviceId: ServiceId
+) {.async: (raises: [CancelledError]).} =
+  heartbeat "discover service ads", disco.discoConfig.advertExpiry:
+    let remoteAds = await disco.fetchRemoteAds(serviceId, disco.discoConfig.fLookup)
+    disco.discoverer.cache[serviceId] =
+      mergeAds(disco.discoverer.cache.getOrDefault(serviceId, @[]), remoteAds)
```
richard-ramos, line 188: Should the ads stored in this cache have a expiry time? because otherwise the logic (iiuc) will keep adding to the cache and only remove them when stopDiscovering is called

SionoiS:
> never launched a background task to actually fetch advertisements from registrars

That's is not the goal. Yes true we are missing a feature here but it's only populating the table AKA calling `refreshTable` on it once.

> would get nothing back on lookup.

Wut? A lookup will populate the table as it search for adverts.

> makes lookup cache-aware when a discovery session is active

We may consider this in the future but for now that's a no. Open a PR to modify the spec so that we can discuss this.

PR #2425, opened 2026-05-07
feat(peerstore): address TTL

gmelodie:
## Summary

Adds TTL-based expiry to `AddressBook` so that stale peer addresses are
automatically pruned when they have not been refreshed within a configurable
window.

Without this, addresses accumulate indefinitely. After an identify exchange
or address update, the timestamp is refreshed; if a peer's addresses are
never refreshed within the TTL window, the next read silently removes them.

A `ZeroDuration` TTL opts out of pruning entirely. The default TTL is 1 hour
(`defaultAddressTtl`).

## Affected Areas

- [ ] Gossipsub
- [ ] Transports

  - `AddressBook` gains `ttl` / `lastUpdated` fields and lazy-expiry
    overrides for `[]`, `[]=`, `del`, `contains`, and `extend`.
  - `PeerStore` gains `addressTtl: Duration` field and
    `defaultAddressTtl = 1.hours` constant.
  - `PeerStore.new()` gains an `addressTtl` keyword parameter.
  - `SwitchBuilder` gains `withAddressTtl()` and a backing
    `addressTtl: Opt[Duration]` field.

- [ ] Protocol Logic
- [ ] Build / Tooling
- [ ] Other

The TTL is a new opt-in configuration knob; no existing call sites that omit
`addressTtl` change their signature. However, **behavior changes by default**:
previously, addresses were kept indefinitely; after this PR they expire after
1 hour of no refresh.

Projects that rely on long-lived address caches (e.g. across long idle
periods) may need to call `.withAddressTtl(ZeroDuration)` or tune the value.

- **Nimbus:** N/A (pending downstream validation)
- **Waku:** N/A (pending downstream validation)
- **Codex:** N/A (pending downstream validation)

## Impact on Library Users

- **New API**: `SwitchBuilder.withAddressTtl(ttl: Duration)` lets callers
  configure the address TTL. Pass `ZeroDuration` to restore the old
  "keep forever" behavior.

- **New API**:
  `PeerStore.new(identify, capacity, addressTtl)` — `addressTtl` is a new
  optional keyword argument defaulting to `1.hours`.

- **Behavior change**: The default TTL is now **1 hour**. Addresses not
  refreshed (via identify or direct `AddressBook` writes) within that window
  will be lazily removed on the next read.

- Expiry is lazy — no background timer — so stale entries are only collected
  at access time.

- Change handlers registered on `AddressBook` are notified when an entry is
  lazily evicted.

## Risk Assessment

- **Backward compatibility**: Existing call sites compile unchanged, but the
  default behavior silently shifts from "keep addresses forever" to
  "expire after 1 hour." This could cause unexpected dial failures for peers
  that are idle for more than an hour if no re-identify occurs.

- **Network behavior**: Peers that are regularly connected will have their
  addresses refreshed on every identify exchange
  (`updatePeerInfo` calls `peerStore[AddressBook][peerId] = addrs`), so the
  risk is mainly for peers that are known but not actively connected.

- **Performance**: Lazy eviction adds a `Moment.now()` comparison on each
  address lookup. There is no background sweeper, so the overhead is minimal
  and proportional to access frequency.

- **Security**: No security implications.

## References

N/A

## Additional Notes

Expiry precision is limited by read frequency — an entry won't be removed
until it is next read after the TTL has elapsed.

If proactive cleanup is needed in future (e.g. for memory-constrained nodes),
a periodic sweeper can be added on top of this foundation.

---


libp2p/peerstore.nim
```diff
@@ -5,9 +5,6 @@
 runnableExamples:
   # Will keep info of all connected peers +
   # last 50 disconnected peers
-  # Passing `nil` for `identify` is only safe for simple peer-book usage like
-  # this example. APIs that rely on identify metadata require a real
-  # `Identify` instance when constructing the `PeerStore`.
```
vladopajic, line 10: ?

libp2p/builders.nim
```diff
@@ -77,6 +77,7 @@ type
     agentVersion: string
     nameResolver: NameResolver
     peerStoreCapacity: Opt[int]
+    addressTtls: Opt[AddressConfidenceTtls]
```
vladopajic:
```suggestion
    addressTtls: AddressConfidenceTtls = AddressConfidenceTtls()
```
or 

```suggestion
    addressTtls: AddressConfidenceTtls = defaultAddressConfidenceTtls
```

`Opt` is unnecessary

libp2p/builders.nim
```diff
@@ -284,6 +285,11 @@ proc withPeerStore*(b: SwitchBuilder, capacity: int): SwitchBuilder =
   b.peerStoreCapacity = Opt.some(capacity)
   b
 
+proc withAddressTtls*(b: SwitchBuilder, ttls: AddressConfidenceTtls): SwitchBuilder =
```
vladopajic: `withAddressConfidenceTtls`

libp2p/peerstore.nim
```diff
@@ -59,10 +57,37 @@ type
...
+
+  AddressEntry* = object
+    address*: MultiAddress
+    confidence*: AddressConfidence
+    lastUpdated*: Moment
+
+  AddressConfidenceTtls* = object
+    low*: Duration ## TTL for Low-confidence addresses (default 15 minutes)
+    medium*: Duration ## TTL for Medium-confidence addresses (default 1 hour)
+    high*: Duration ## TTL for High-confidence addresses (default 24 hours)
```
vladopajic: default values are not really lined to these properties; changing defaults will be inconsistent with comment
- avoid mentioning defaults
- or use defaults 

```suggestion
    low*: Duration = 15.minutes ## TTL for Low-confidence addresses
    medium*: Duration = 1.hours ## TTL for Medium-confidence addresses
    high*: Duration = 24.hour s## TTL for High-confidence addresses
```

libp2p/peerstore.nim
```diff
@@ -80,10 +105,25 @@ type
...
+    pruneHandle: Future[void]
+
+const defaultAddressConfidenceTtls* =
+  AddressConfidenceTtls(low: 15.minutes, medium: 1.hours, high: 24.hours)
+
+proc new*(
+    Self: type PeerStore,
+    identify: Identify,
+    capacity = 1000,
+    addressTtls = defaultAddressConfidenceTtls,
```
vladopajic:
```suggestion
    addressTtls = AddressConfidenceTtls()
```

based on usage it looks like the best option would be like this, avoiding `defaultAddressConfidenceTtls`

libp2p/peerstore.nim
```diff
@@ -124,6 +164,153 @@ proc addHandler*[T](peerBook: PeerBook[T], handler: PeerBookChangeHandler) =
...
+  let elapsed = Moment.now() - entry.lastUpdated
+  case entry.confidence
+  of AddressConfidence.Low:
+    elapsed > ttls.low
+  of AddressConfidence.Medium:
+    elapsed > ttls.medium
+  of AddressConfidence.High:
+    elapsed > ttls.high
+  of AddressConfidence.Infinite:
+    false
```
vladopajic: `return elapsed > entry.confidence.getTtls(ttls)`
would be nicer like that, with
```proc getTtls*(confidence: AddressConfidence, ttls: AddressConfidenceTtls): Duration =```



but then:
- `getTtls` should be `ttls` ?
- then `ttls` arg should be something else .. ?

`return elapsed > entry.confidence.ttls(ttlsConfig)` ???

File: libp2p/peerstore.nim [hunk 7]
```diff
@@ -124,6 +164,153 @@ proc addHandler*[T](peerBook: PeerBook[T], handler: PeerBookChangeHandler) =
...
+
+proc mergeConfidence(
```
vladopajic: `mergeConfidence` -> `higherConfidence` ?

libp2p/peerstore.nim
```diff
@@ -124,6 +164,153 @@ proc addHandler*[T](peerBook: PeerBook[T], handler: PeerBookChangeHandler) =
...
+  ## Called after a successful outbound connection to `ma`.
+  ## Upgrades the address to High confidence and refreshes its lastUpdated.
+  ## If the address is not yet in the book it is added.
+  let now = Moment.now()
+  var entries = addressBook.book.getOrDefault(peerId)
+  var idx = -1
+  for i, e in entries:
+    if e.address == ma:
+      idx = i
+      break
```
vladopajic:
```suggestion
  let idx = entries.findWithAddress(ma)
```

libp2p/peerstore.nim
```diff
@@ -124,6 +164,153 @@ proc addHandler*[T](peerBook: PeerBook[T], handler: PeerBookChangeHandler) =
...
+  if idx >= 0:
+    entries[idx].confidence = max(entries[idx].confidence, AddressConfidence.High)
+    entries[idx].lastUpdated = now
+  else:
+    entries.add(
+      AddressEntry(address: ma, confidence: AddressConfidence.High, lastUpdated: now)
+    )
+  addressBook.book[peerId] = entries
+  for handler in addressBook.changeHandlers:
```
vladopajic: give be a brake `\n` here

libp2p/peerstore.nim
```diff
@@ -124,6 +164,153 @@ proc addHandler*[T](peerBook: PeerBook[T], handler: PeerBookChangeHandler) =
...
+        break
+    if idx >= 0:
+      entries[idx].confidence = max(entries[idx].confidence, confidence)
+      entries[idx].lastUpdated = now
+    else:
+      entries.add(AddressEntry(address: ma, confidence: confidence, lastUpdated: now))
+  if entries.len > 0:
+    addressBook.book[key] = entries
+  for handler in addressBook.changeHandlers:
+    handler(key)
```
vladopajic: same comments as above

how is this code note reused with `markConnected`?

libp2p/peerstore.nim
```diff
@@ -124,6 +164,153 @@ proc addHandler*[T](peerBook: PeerBook[T], handler: PeerBookChangeHandler) =
...
+    key: PeerId,
+    addrs: seq[MultiAddress],
+    confidence = AddressConfidence.Medium,
+) =
+  ## Add addresses for `key` without removing existing ones.
+  ## For addresses already present the confidence is upgraded to
+  ## `max(existing, confidence)` and lastUpdated is refreshed.
+  ## New addresses are added with `confidence`.
+  let now = Moment.now()
+  var entries = addressBook.book.getOrDefault(key)
```
vladopajic, line 308: i dont know if this is going to work; afik values of `getOrDefault` are usually not mutable? it it returns default value is it going to be stored in book as well?

libp2p/peerstore.nim
```diff
@@ -132,24 +319,60 @@ macro getTypeName(t: type): untyped =
   let typ = getTypeImpl(t)[1]
   newLit(repr(typ.owner()) & "." & repr(typ))
 
+proc `[]`*(p: PeerStore, _: type[AddressBook]): AddressBook =
+  ## Get the AddressBook, initialising it with the store's configured TTLs.
+  let name = getTypeName(AddressBook)
+  var book = AddressBook(p.books.getOrDefault(name))
+  if book.isNil:
```
vladopajic: this is strange.... `AddressBook` is create then there is `.isNil`.

code should be build in such way that if ` var book = AddressBook()` it must mean that book is created in next line because this should be obvious usecase.

libp2p/peerstore.nim
vladopajic: this might work of `ref` objects, but ether way would be nicer if we stick to the same pattern through codebase:
- use `map.getOrDefault(key)` when value retrieved is not modified. 
- use `map.withValue(key, value):` if we intend to modify it

above will be universally true for any object type (assuming); not sure from top of my head if `map.withValue` for ref values will work, why not?

libp2p/dialer.nim
```diff
@@ -90,6 +90,8 @@ method dialAndUpgrade*(
 
       doAssert not isNil(mux), "connection died after upgrade " & $dialed.dir
       debug "Dial successful", peerId = mux.connection.peerId
+      if mux.connection.peerId != default(PeerId):
+        self.peerStore[AddressBook].markConnected(mux.connection.peerId, addrs)
```
richard-ramos: Shouldn't be the case, since this is AFTER you connect to a peer, but for the sake of being consistent, the address policy needs to be applied, otherwise we risk storing addresses we are not really interested into in the peerstore

```suggestion
        let filteredAddrs = self.peerStore.addressPolicy.filterAddrs(addrs)
        self.peerStore[AddressBook].markConnected(mux.connection.peerId, filteredAddrs)
```

tests/libp2p/test_peer_store.nim
```diff
@@ -133,3 +133,126 @@ suite "PeerStore":
...
+      address: addr1,
+      confidence: AddressConfidence.Infinite,
+      lastUpdated: Moment.now() - 365.days,
+    )
+    check entry.isExpired(ttls) == false
+
+  test "pruneExpired removes only expired entries, keeps live ones":
+    let book = makeBook(1.seconds, 1.hours, 24.hours)
+    book.set(peerId1, @[addr1], AddressConfidence.Low)
+    book.set(peerId1, @[addr2], AddressConfidence.Medium)
```
richard-ramos: This `.set` is replacing the previous address, so addr1 will in practice not exist in the address book. This should probably be `.extend`

libp2p/peerstore.nim
richard-ramos:
```suggestion
proc isExpired*(entry: AddressEntry, ttls: AddressConfidenceTtls): bool =
  let ttl =
    case entry.confidence
    of AddressConfidence.Low:
      ttls.low
    of AddressConfidence.Medium:
      ttls.medium
    of AddressConfidence.High:
      ttls.high
    of AddressConfidence.Infinite:
      return false

  if ttl <= ZeroDuration:
    return false

  Moment.now() - entry.lastUpdated > ttl
```
otherwise, `ZeroDuration` entries will be expired automatically

libp2p/peerstore.nim
```diff
@@ -132,24 +319,60 @@ macro getTypeName(t: type): untyped =
...
+  heartbeat "AddressBook TTL pruning", interval, sleepFirst = true:
+    peerStore[AddressBook].pruneExpired()
+
+proc startAddressPruning*(peerStore: PeerStore) =
+  ## Start the periodic per-address TTL pruning loop. No-op if already running.
+  ## The loop fires at the Low-confidence TTL interval (shortest possible expiry).
+  if not peerStore.pruneHandle.isNil:
+    return
+
+  let interval = peerStore.addressTtls.low
```
richard-ramos: This only checks for `low` being `ZeroDuration`, but would not apply if medium or high is set to `ZeroDuration`

libp2p/peerstore.nim
gmelodie: code be like
image

libp2p/peerstore.nim
gmelodie: It's var so we can add values to `entries`, not because it's `ref` (might be, but doesn't matter). `entries` is written back at the end in line 295: 
```nim
 if entries.len > 0:
    addressBook.book[key] = entries
```

libp2p/peerstore.nim
```diff
@@ -159,7 +390,7 @@ proc updatePeerInfo*(
   if len(info.addrs) > 0:
     let addrs = peerStore.addressPolicy.filterAddrs(info.addrs)
     if addrs.len > 0:
-      peerStore[AddressBook][info.peerId] = addrs
+      peerStore[AddressBook].set(info.peerId, addrs, AddressConfidence.Medium)
```
richard-ramos, line 407: Something interesting happening here is that you could overwrite  the dialing address what was set to `High` in dialer.nim, line 97, (that happens before Identify is run).

libp2p/dialer.nim
```diff
@@ -90,6 +91,10 @@ method dialAndUpgrade*(
 
       doAssert not isNil(mux), "connection died after upgrade " & $dialed.dir
       debug "Dial successful", peerId = mux.connection.peerId
+      if mux.connection.peerId != default(PeerId):
```
vladopajic: how come can `peerId` be uninitialized?

libp2p/peerstore.nim
```diff
@@ -124,6 +164,156 @@ proc addHandler*[T](peerBook: PeerBook[T], handler: PeerBookChangeHandler) =
...
+      )
+    )
+
+  if newEntries.len > 0:
+    addressBook.book[peerId] = newEntries
+  elif peerId in addressBook.book:
+    addressBook.book.del(peerId)
+
+  for handler in addressBook.changeHandlers:
+    handler(peerId)
```
vladopajic:
```suggestion
  ## Replace the address list for `peerId`. For addresses that already exist
  ## the confidence is upgraded to `max(existing, confidence)` so that a
  ## high-confidence entry is never downgraded by a lower-confidence update.
  defer: 
    for handler in addressBook.changeHandlers:
      handler(peerId)
    
  if addrs.len == 0:
    addressBook.book.del(peerId)
    return
  
  let now = Moment.now()
  var prevConf: Table[MultiAddress, AddressConfidence]
  for entry in addressBook.book.getOrDefault(peerId):
    prevConf[entry.address] = entry.confidence

  var newEntries = newSeqOfCap[AddressEntry](addrs.len)
  for ma in addrs:
    newEntries.add(
      AddressEntry(
        address: ma,
        confidence: higherConfidence(prevConf, ma, confidence),
        lastUpdated: now,
      )
    )

  addressBook.book[peerId] = newEntries
```

- since `changeHandlers` should always be executed have them in defer at the begging
- case when addresses are cleared, does not need to go through same logic unnecessary. it is also better to clearly separate that case from case where addresses are updated.

libp2p/peerstore.nim
```diff
@@ -124,6 +164,156 @@ proc addHandler*[T](peerBook: PeerBook[T], handler: PeerBookChangeHandler) =
...
+proc markConnected*(addressBook: AddressBook, peerId: PeerId, ma: MultiAddress) =
+  ## Called after a successful outbound connection to `ma`.
+  ## Upgrades the address to High confidence and refreshes its lastUpdated.
+  ## If the address is not yet in the book it is added.
+  let now = Moment.now()
+  var entries = addressBook.book.getOrDefault(peerId)
+  entries.upsertAddress(ma, AddressConfidence.High, now)
+  addressBook.book[peerId] = entries
+  for handler in addressBook.changeHandlers:
+    handler(peerId)
```
vladopajic: `addressBook.notifyChangeHandlers(peerId)`

here and in other places

libp2p/peerstore.nim
```diff
@@ -132,24 +322,65 @@ macro getTypeName(t: type): untyped =
   let typ = getTypeImpl(t)[1]
   newLit(repr(typ.owner()) & "." & repr(typ))
 
+proc `[]`*(p: PeerStore, _: type[AddressBook]): AddressBook =
+  ## Get the AddressBook, initialising it with the store's configured TTLs.
+  let name = getTypeName(AddressBook)
+  p.books.withValue(name, bookPtr):
+    return AddressBook(bookPtr[])
+  let book = AddressBook.new()
```
vladopajic, line 345: break after early return; it's new block of code

libp2p/peerstore.nim
```diff
@@ -132,24 +322,65 @@ macro getTypeName(t: type): untyped =
...
+    return
+
+  let ttls = peerStore.addressTtls
+  var interval = ZeroDuration
+  for t in [ttls.low, ttls.medium, ttls.high]:
+    if t > ZeroDuration and (interval == ZeroDuration or t < interval):
+      interval = t
+
+  if interval <= ZeroDuration:
+    return
```
vladopajic:
```
let ttlsSeq = [ttls.low, ttls.medium, ttls.high]

let interval = ttlsSeq.filterIt(it > ZeroDuration).foldl(min(a,b))

or

let interval = ttlsSeq.filterIt(it > ZeroDuration).min()
```

this can be more cleaner and intuitive with one liners

tests/libp2p/test_peer_store.nim
```diff
@@ -133,3 +134,126 @@ suite "PeerStore":
       peerStore.cleanup(randomPeerId)
 
     check peerStore[AgentBook].len == 30
+
+suite "AddressBook TTL / confidence":
+  let
+    keyPair1 = KeyPair.random(ECDSA, rng()).get()
+    peerId1 = PeerId.init(keyPair1.seckey).get()
```
vladopajic: `let peerId1 = PeerId.random()` 

and code that later needs pub key should be: `peerId1.getPubKey()`

tests/libp2p/test_peer_store.nim
```diff
@@ -133,3 +134,126 @@ suite "PeerStore":
...
+    b
+
+  test "isExpired - Low expires after low TTL":
+    let ttls = AddressConfidenceTtls(low: 1.seconds, medium: 1.hours, high: 24.hours)
+    let entry = AddressEntry(
+      address: addr1,
+      confidence: AddressConfidence.Low,
+      lastUpdated: Moment.now() - 2.seconds,
+    )
+    check entry.isExpired(ttls) == true
```
vladopajic: `check entry.isExpired(ttls)`
and 
`check not entry.isExpired(ttls)` 

perhaps this notation is more closer to natural language

vladopajic (approved):
lgtm; cosmetic suggestion

richard-ramos (approved):

PR #2427, opened 2026-05-07
test(interop): GossipSub - fix message id

rlve:
## Summary

Uncovered while updating PR and fixing merge conflicts in https://github.com/libp2p/test-plans/pull/850.

Fix the interop gossipsub message ID provider to emit the raw 8 bytes of the message payload prefix instead of the base-10 ASCII string of the decoded uint64. This matches the wire format used by the go and rust interop binaries, so message IDs line up across implementations during interop runs.

## Affected Areas

- [x] Interop tests
  Interop test harness only (`interop/gossipsub/src/node.nim`). No library code changed.

## Impact on Library Users

None.

## Risk Assessment

Low.


interop/gossipsub/src/node.nim
```diff
@@ -43,12 +43,11 @@ proc extractMsgId*(data: openArray[byte]): uint64 =
 
 proc interopMsgIdProvider*(m: Message): Result[MessageId, ValidationResult] =
   ## Message ID provider for interop tests.
-  ## Reads first 8 bytes of message data as big-endian uint64,
-  ## returns base-10 string representation as bytes.
+  ## The message ID is the raw first 8 bytes of the message data (big-endian u64),
+  ## matching the wire format used by the go and rust interop binaries.
   if m.data.len < 8:
```
vladopajic: would be nice if 8 is in const, since it is everywhere

vladopajic (approved):

gmelodie (approved):

PR #2433, opened 2026-05-07
fix(ci): add skip to autotls integration tests

gmelodie:
## Summary

Fixes a reliability issue with AutoTLS integration tests. Since it depends on Let's Encrypt's staging servers, and sometimes they've been unavailable causing tests to fail. With this we retry three times to make sure.


vladopajic:
i am not convinced in this.. github is now failing due to high platform usage, and it is not obvious if this is going to solve general network connectivity issue. 
the golden bullet fix would be to run http server locally, that will mock staging server functionality. 

but then second question is how come internal logic is not robust enough to handle temporal connectivity drops? if they are really temporal better solve it in the code.

tests/tools/unittest.nim
```diff
@@ -44,6 +44,36 @@ template asyncTest*(name: string, body: untyped): untyped =
       )()
     )
 
+template asyncTestRetry*(
```
vladopajic: this as general solution will not always work. if test throws exception in will trow it at any point. and depending on test it can leave open connections (or other resources) thus failing again (due to checkers) even if tests succeeds on next attempt.

gmelodie:
> github is now failing due to high platform usage, and it is not obvious if this is going to solve general network connectivity issue.

This not the issue tackled here, what I'm aiming to do is reduce failures due to Let's Encrypt servers.

 > but then second question is how come internal logic is not robust enough to handle temporal connectivity drops? if they are really temporal better solve it in the code.
 
 Aham agreed. Will implement

vladopajic:
> This not the issue tackled here, what I'm aiming to do is reduce failures due to Let's Encrypt servers.

how do we know that let's encrypt servers are making those issues, and it's not on other side - the github runner?

occasionally we can see, mostly lately, that steps fail and random steps. afaiks, it appears to be some temporal connectivity drop (fetching deps, installing nim, etc..) as next run always succeeds.

vladopajic (approved):

richard-ramos (approved):

PR #2434, opened 2026-05-07
fix(kad): table insertion check

SionoiS:
In the context of kademlia, a table prevent the local node from being included by checking if it equals the table Id, the problem is that in the case of service discovery, tables Ids don't equal the node Id anymore which allow it to be wrongly included.

### DIFF
--- a/libp2p/protocols/service_discovery/routing_table_manager.nim
+++ b/libp2p/protocols/service_discovery/routing_table_manager.nim
@@ -31,7 +31,6 @@ proc addService*(
     replication: int,
     bucketsCount: int,
     status: ServiceStatus,
-    localNodeId: Key,
 ): bool =
   # Fast path: service already exists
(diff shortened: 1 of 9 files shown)


gmelodie (commented):
In all `addService` calls the new argument is `disco.rtable.selfId`, if this is always the case, then just change `addService` to get it from `rtable` which is the second argument it takes.

libp2p/protocols/service_discovery/routing_table_manager.nim
```diff
@@ -31,6 +31,7 @@ proc addService*(
     replication: int,
     bucketsCount: int,
     status: ServiceStatus,
+    localNodeId: Key,
```
richard-ramos: It should be possible to obtain the localNodeId from `mainRoutingTable.localNodeId`

richard-ramos (approved):

gmelodie (approved):

PR #2440, opened 2026-05-08
refactor(IdentifyPusher): as Service

vladopajic:
## Summary

- initial `IdentifyPusher` was unfortunately added to Switch as ad-hoc component. however, there is more elegant approach which is to use switch `Service`. 
- adding `withIdentifyPusher` to `SwitchBuilder`
- utilizing `PeerInfoObserver`

## Affected Areas

- [x] IdentifyPusher
- [x] Switch

## References

follow up to: https://github.com/vacp2p/nim-libp2p/pull/2412 completing feature


libp2p/services/identify_pusher.nim
```diff
@@ -0,0 +1,204 @@
...
+    p.peerStore.updatePeerInfo(info)
+    if IdentifyPushCodec in info.protos:
+      p.pushPeers.incl(peerId)
+    else:
+      p.pushPeers.excl(peerId)
+
+  p.identifyPush = IdentifyPush.new(onIncomingPush)
+
+  try:
+    switch.mount(p.identifyPush)
```
richard-ramos, line 139: In practice it will likely not happen since no one keeps starting/stopping a switch during the execution of a program, but, having this mount here without umount somewhere else, means that each call of switch.start() will keep this handler around and add a new one for the same protocol

richard-ramos (approved):

libp2p/services/identify_pusher.nim
vladopajic: true but i don't think issue is with this code, as much as there is issue with current service lifecycle logic:
- `setup` should happen once to bootstrap service to swithc
- `start` / `stop` should be able to happen any amount of time, in any order after `setup`

libp2p/services/identify_pusher.nim
```diff
@@ -0,0 +1,204 @@
...
+  if not p.started:
+    return
+
+  for peerId in p.pushPeers.toSeq():
+    let fut = p.sendOne(peerId)
+    p.ongoingSend.add(fut)
+    fut.addCallback proc(udata: pointer) =
+      let idx = p.ongoingSend.find(fut)
+      if idx >= 0:
+        p.ongoingSend.del(idx)
```
gmelodie, line 113: this `addCallback` is very cool

libp2p/services/identify_pusher.nim
```diff
@@ -0,0 +1,204 @@
...
+  p.connManager.addPeerEventHandler(onIdentified, PeerEventKind.Identified)
+  p.connManager.addPeerEventHandler(onLeft, PeerEventKind.Left)
+
+method stop*(
+    p: IdentifyPusher, switch: Switch
+): Future[bool] {.async: (raises: [CancelledError]).} =
+  let hasBeenStopped = await procCall Service(p).stop(switch)
+  if hasBeenStopped:
+    if not p.started:
+      return hasBeenStopped
```
gmelodie, line 183:
```suggestion
  if hasBeenStopped and not p.started:
    return true
```

libp2p/services/identify_pusher.nim
```diff
@@ -0,0 +1,204 @@
...
+      p.connManager.removePeerEventHandler(
+        p.onIdentifiedHandler, PeerEventKind.Identified
+      )
+      p.onIdentifiedHandler = nil
+    if not p.onLeftHandler.isNil:
+      p.connManager.removePeerEventHandler(p.onLeftHandler, PeerEventKind.Left)
+      p.onLeftHandler = nil
+    if not p.onPeerInfoUpdated.isNil:
+      p.peerInfo.removeObserver(p.onPeerInfoUpdated)
+      p.onPeerInfoUpdated = nil
```
gmelodie, line 197: Maybe add some wrappers here so this part would be:
```
p.removeEventHandlers()
```

libp2p/services/identify_pusher.nim
```diff
@@ -0,0 +1,204 @@
...
+
+    let pending = move(p.ongoingSend)
+    await pending.cancelAndWait()
```
gmelodie, line 200: What's the reason to have this `move` here instead of

```suggestion
    await p.ongoingSend.cancelAndWait()
```

libp2p/utils/future.nim
```diff
@@ -35,6 +35,12 @@ template newFutureCompleted*[T](): auto =
   fut.complete()
   fut
 
+template cancelAndWait*[T](futs: seq[T]): auto =
+  var cancelFuts = newSeqOfCap[Future[void].Raising([])](futs.len)
+  for fut in futs:
+    cancelFuts.add(fut.cancelAndWait())
+  allFutures(cancelFuts)
+
```
gmelodie, line 43: Are there other places where we could use this? Add this to this PR or create issue maybe?

libp2p/builders.nim
```diff
@@ -357,6 +359,14 @@ proc withServices*(b: SwitchBuilder, services: seq[Service]): SwitchBuilder =
...
   b
 
+proc withIdentifyPusher*(b: SwitchBuilder, enabled: bool = true): SwitchBuilder =
+  ## When enabled, the IdentifyPush protocol is mounted on the
+  ## switch and an `IdentifyPusher` service tracks which connected peers
+  ## advertise IdentifyPush, broadcasting our updated `PeerInfo` to all
+  ## tracked peers whenever it changes.
+  b.identifyPusherEnabled = enabled
+  b
+
```
gmelodie, line 369:
```suggestion
proc withIdentifyPusher*(b: SwitchBuilder): SwitchBuilder =
  ## When enabled, the IdentifyPush protocol is mounted on the
  ## switch and an `IdentifyPusher` service tracks which connected peers
  ## advertise IdentifyPush, broadcasting our updated `PeerInfo` to all
  ## tracked peers whenever it changes.
  b.identifyPusherEnabled = true
  b

```

tests/libp2p/test_switch.nim
```diff
@@ -1267,3 +1267,54 @@ suite "Switch":
...
+
+    let dummy = LPProtocol.new(@[codecs], dummyHandler)
+    await dummy.start()
+    switch2.mount(dummy)
+
+    # switch1 should get information about new protocol
+    checkUntilTimeout:
+      codecs in switch1.peerStore[ProtoBook][switch2.peerInfo.peerId]
+
+  asyncTest "broadcasts on updateAddrs":
```
gmelodie, line 1308:
```suggestion
  asyncTest "broadcasts on peerInfo update":
```

gmelodie (approved):
Cool!

libp2p/services/identify_pusher.nim
vladopajic: `move` will effectively do `p.ongoingSend.clear()` so i don't have to

libp2p/services/identify_pusher.nim
vladopajic: will be addressed in future pr; i kinda did this intentionally since `hasBeenStopped` will be removed, so i wanted to leave other if.

libp2p/builders.nim
vladopajic: no, this is intentionally like this because you can use it:

```
.withIdentifyPusher()
or
.withIdentifyPusher(someconfig.IsEnabled)
```

libp2p/utils/future.nim
vladopajic: @copilot please open issue to track this

PR #2441, opened 2026-05-08
chore(service-disco): rename `startDiscovering` to `registerInterest` + improve component test

### LINKED ISSUE #2417: service-disco: what is `startDiscovering` meant to do?

rlve:
## Summary

Renames `startDiscovering` / `stopDiscovering` to `registerInterest` / `unregisterInterest` in the service discovery API. The old names suggested the call kicks off background discovery and accumulates ads, but it does neither. It only registers an `Interest` entry in `rtManager` and seeds a per-service routing table from the main one so that subsequent `lookup` calls have somewhere to start. See vacp2p/nim-libp2p#2417 for the discussion that led to the rename.

The cbind layer (Nim enum, exported C symbols, header) is renamed to match.

Also refactors the component test that was previously called `"startDiscovering sends messages and finds registered peer"`. The old test was misleading (the call sends no messages) and largely duplicated the end-to-end flow already covered by the test above it.

## Affected Areas

- [x] Service Discovery
- [x] cbind

## Impact on Library Users

Breaking API change.

Nim API:
- `startDiscovering` -> `registerInterest`
- `stopDiscovering` -> `unregisterInterest`

C bindings:
- `libp2p_service_disco_start_discovering` -> `libp2p_service_disco_register_interest`
- `libp2p_service_disco_stop_discovering` -> `libp2p_service_disco_unregister_interest`
- enum `SD_START_DISCOVERING` / `SD_STOP_DISCOVERING` -> `SD_REGISTER_INTEREST` / `SD_UNREGISTER_INTEREST`

Behavior is unchanged. Migration is a straight rename.

## Risk Assessment

Mechanical rename. No behavioral change. Anyone using the C bindings needs to update the renamed function and enum names.

## References

- vacp2p/nim-libp2p#2417


SionoiS (approved):
LGTM! Thanks.

gmelodie (approved):

richard-ramos (approved):
Since bindings were modified, https://github.com/logos-co/logos-libp2p-module will need to be updated
cc: @gmelodie

PR #2443, opened 2026-05-08
chore(service-disco): dedupe identical ads

### LINKED ISSUE #2431: service-disco: lookup returns duplicates when multiple queried peers hold the same ad
`lookup(serviceId)` does not deduplicate results across the peers it queries. If the same advertisement is held by more than one of the queried peers, with the same `(peerId, seqNo)` returned by both, each copy ends up in the result.

After one lookup, the discoverer learns about the advertiser via closerPeers from the registrar's reply and inserts it into the service-specific routing table. On the next lookup, the discoverer queries both the registrar and the advertiser itself. The advertiser also caches its own ad locally (the local-advertising path in `addProvidedService`), so both peers return the same record. The caller receives two identical advertisements.

Reproduction in `test_advertise_discover.nim:128-136`: one advertiser registers two services against one registrar, then the discoverer looks up each service id in turn. The second lookup returns len == 2 for a single advertiser.

rlve:
Fixes [vacp2p/nim-libp2p#2431. ](https://github.com/vacp2p/nim-libp2p/issues/2431) 

`lookup(serviceId)` was returning duplicates when multiple queried peers held the same advertisement. Specifically, an advertiser caches its own ad locally, so once the discoverer learned about the advertiser via `closerPeers`, subsequent lookups would query both the registrar and the advertiser and get back two byte-identical copies.

The fix switches the two ad collectors in the discoverer (`collectBucketAds` and `lookup`) from `seq[Advertisement]` to `HashSet[Advertisement]`, with `hash(Advertisement)` keyed on `envelope.signature.data`. Dedup is byte-identity only. Adverts that share `(peerId, seqNo)` but differ in content are kept distinct.

## Affected Areas

- [x] Service Discovery

## Impact on Library Users

`lookup()` no longer returns byte-duplicate adverts. The return type is unchanged (`seq[Advertisement]`), but result ordering is no longer guaranteed since the dedup happens through a `HashSet`. Callers that relied on insertion order would need to sort explicitly, but that wasn't a documented guarantee before either.

## Risk Assessment

Low. The dedup is byte-identity only, so no semantic behavior changes. The bucket-level `limit` short-circuit in `collectBucketAds` now bounds *unique* ads instead of raw ads, which is the correct interpretation of the cap.

## References

- Closes vacp2p/nim-libp2p#2431
- Test coverage added in `tests/libp2p/service_discovery/component/test_advertise_discover.nim`:
  - Updated existing "one advertiser provides two services" test
  - New "lookup dedups byte-identical adverts but keeps same (peerId, seqNo) variants" test exercising both contracts in one shot


SionoiS (approved):
LGTM Thank you!

gmelodie (approved):
\* amazing \*

richard-ramos (approved):

Issue #1867, opened 2025-11-11
Makefile for tests


gmelodie:
Use Makefile instead of nimble for running most of the tests

vladopajic:
makefile could be used for all tasks, it can be entry point for running commands. those commands can invoke nimble task when necessary (for example `nimble website` could be `make build_website` where make just calls `nimble website`) .

richard-ramos:
Implemented in #2454

PR #2454, opened 2026-05-11
chore: makefile instead of nimble

gmelodie:
## Summary

This PR replaces direct `nimble test` invocations in CI workflows with `make test` / `make testintegration`, and adds the corresponding `Makefile` targets that were previously missing. Makefile targets run nim directly, reduce the overall time to run each target. I'm not sure why that is, maybe because nimble needs to be compiled before it runs.

### Changes

- `.github/workflows/ci.yml`
  - Replaced `nimble test` with `make test` in the main test step.

- `.github/workflows/coverage.yml`
  - Replaced `nimble test` with `make test` in the coverage test step.

- `.github/workflows/daily_common.yml`
  - Replaced:
    ```sh
    nimble test -- -d:chronicles_log_level=DEBUG
    ```
    with:
    ```sh
    export NIMFLAGS=...
    make test
    ```
  - Replaced the ad-hoc `nim r` integration test invocation with:
    ```sh
    export NIMFLAGS=...
    make testintegration
    ```
    removing the redundant `nim --version` / `nimble --version` lines.

- `Makefile`
  - Added:
    - `NIMC`
    - `NIMFLAGS`
    - `V`
    - `VERBOSITY_FLAG`
    - `NIM_FLAGS`
    - `RUNNER_FLAGS`
    - `TEST_PATH`
  - Added targets:
    - `test`
    - `_test_all`
    - `testmultiformatexts`
    - `testintegration`
    - `testpath`
  - These targets invoke `nim r` directly with a consistent set of flags.
  - Retained thin wrappers that delegate to `nimble` for tasks that still require it:
    - `install_pinned`
    - `pin`
    - `unpin`
    - `gen_multicodec`
    - `format`
  - Removed an inline comment on the `cbind` target.

## Affected Areas

- Gossipsub
- Transports
- Peer Management / Discovery
- Protocol Logic
- Build / Tooling
  - CI workflows and `Makefile` updated to run tests via `make` instead of `nimble`.
- Other

All public library APIs and encode/decode paths are completely unaffected — this PR only changes how tests are invoked in CI and locally.

Downstream consumers (Nimbus, Waku, Codex) that depend on the library itself will see no change. No downstream validation is needed.

## Impact on Library Users

- No public API, ABI, or behavioral changes.
- This is a build/tooling-only change.

## Risk Assessment

- **Compiler flag parity**
  - The `Makefile`’s `NIM_FLAGS` must stay in sync with any flags previously passed via `nimble test`.
  - If a flag was silently added by `nimble` and is now absent, tests could behave differently.
  - Risk: low — the flag set is now explicit and auditable in one place.

- **`NIMFLAGS` env var propagation**
  - The daily workflow now sets `NIMFLAGS` as an environment variable before calling `make`.
  - The `Makefile` appends `$(NIMFLAGS)` at the end of `NIM_FLAGS`, so extra flags compose correctly.
  - Risk: low — this is the standard `make` override pattern.

- **`-f` (force recompile) flag**
  - `NIM_FLAGS` includes `-f`, which forces full recompilation on every `make test` run.
  - This was likely the existing `nimble` behavior; if it was not, CI build times may increase slightly.
  - Risk: low.

## References
- #1867

### DIFF
--- a/.github/workflows/daily_common.yml
+++ b/.github/workflows/daily_common.yml
@@ -163,11 +163,11 @@ jobs:
           if [[ '${{ matrix.config.cc }}' == 'clang' ]]; then
             export NIMFLAGS="${NIMFLAGS} --cc:clang"
           fi
-          nimble test -- -d:chronicles_log_level=DEBUG
+          export NIMFLAGS="${NIMFLAGS} -d:chronicles_log_level=DEBUG"
+          make test
 
       - name: Run integration tests
         if: ${{ matrix.config.os == 'linux' && matrix.config.cpu == 'amd64' }}
         run: |
-          nim --version
-          nimble --version
-          nim r -d:libp2p_autotls_support -d:chronicles_log_level=DEBUG tests/integration/test_all.nim
+          export NIMFLAGS="${NIMFLAGS} -d:chronicles_log_level=DEBUG"
+          make testintegration
--- a/.github/workflows/daily_common.yml
+++ b/.github/workflows/daily_common.yml
@@ -170,4 +170,4 @@ jobs:
         if: ${{ matrix.config.os == 'linux' && matrix.config.cpu == 'amd64' }}
         run: |
           export NIMFLAGS="${NIMFLAGS} -d:chronicles_log_level=DEBUG"
-          make testintegration
+          make test_integration
--- a/.github/workflows/daily_common.yml
+++ b/.github/workflows/daily_common.yml
@@ -23,8 +23,20 @@ jobs:
     name: Delete github action cache
     runs-on: ubuntu-latest
     continue-on-error: true
+    permissions:
+      actions: write
     steps:
-      - uses: toshimaru/delete-action-cache@main
+      - name: Delete caches for this ref
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REF: ${{ github.ref }}
+          REPO: ${{ github.repository }}
+        run: |
+          set -u
+          gh cache list -R "$REPO" --ref "$REF" --limit 100 --json id --jq '.[].id' \
+            | while read -r id; do
+                gh cache delete -R "$REPO" "$id" || echo "skipped $id (already gone)"
+              done
 
   resolve_deps:
     name: Resolve newest dependencies
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,29 @@
-.PHONY: all build deps cbind clean
+.PHONY: all build deps cbind clean test _test_all testmultiformatexts testintegration testpath \
+        install_pinned pin unpin gen_multicodec format
+
+NIMC     ?= nim
+NIMFLAGS ?=
+V        ?= 0
+
+ifeq ($(V),0)
+VERBOSITY_FLAG = --verbosity:0
+else
+VERBOSITY_FLAG =
+endif
+
+NIM_FLAGS = \
+  --styleCheck:usages --styleCheck:error \
+  $(VERBOSITY_FLAG) \
+  --skipUserCfg \
+  -f \
+  --threads:on \
+  --opt:speed \
+  -d:libp2p_autotls_support \
+  $(NIMFLAGS)
+
+RUNNER_FLAGS = --output-level=VERBOSE --console
+
+TEST_PATH ?=
 
 all: build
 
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -13,11 +38,77 @@ deps: nix/deps.nix
...
+	nimble pin
+
+unpin:
+	nimble unpin
+
+gen_multicodec:
+	nimble gen_multicodec
+
+format:
+	nimble format
```
vladopajic: this should be `nph ./. *.nim` or smt

vladopajic: because this will be run usually

Makefile
```diff
@@ -13,11 +38,77 @@ deps: nix/deps.nix
...
+	  $(if $(CICOV),--nimcache:nimcache/test_all_multiformat,) \
+	  -d:libp2p_multicodec_exts=../tests/libp2p/multiformat_exts/multicodec_exts.nim \
+	  -d:libp2p_multiaddress_exts=../tests/libp2p/multiformat_exts/multiaddress_exts.nim \
+	  -d:libp2p_multihash_exts=../tests/libp2p/multiformat_exts/multihash_exts.nim \
+	  -d:libp2p_multibase_exts=../tests/libp2p/multiformat_exts/multibase_exts.nim \
+	  -d:libp2p_contentids_exts=../tests/libp2p/multiformat_exts/contentids_exts.nim \
+	  -d:path=multiformat_exts \
+	  tests/test_all.nim -- $(RUNNER_FLAGS) --xml:tests/results_test_all_multiformat.xml
+
+testintegration: nimble.paths
```
vladopajic: `test_integration`

Makefile
```diff
@@ -13,11 +38,77 @@ deps: nix/deps.nix
...
+	done
+
+test: nimble.paths _test_all testmultiformatexts
+
+_test_all: nimble.paths
+	$(NIMC) r $(NIM_FLAGS) \
+	  $(if $(CICOV),--nimcache:nimcache/test_all,) \
+	  tests/test_all.nim -- $(RUNNER_FLAGS) --xml:tests/results_test_all.xml
+
+testmultiformatexts: nimble.paths
```
vladopajic: `test_multiformat_exts`

Makefile
```diff
@@ -13,11 +38,77 @@ deps: nix/deps.nix
...
+	    if [ -n "$$src" ] && [ "$$src" != "." ]; then \
+	      path="$$pkg/$$src"; \
+	    else \
+	      path="$$pkg"; \
+	    fi; \
+	    printf 'switch("path", "%s")\n' "$$path" >> $@; \
+	  done; \
+	done
+
+test: nimble.paths _test_all testmultiformatexts
```
richard-ramos: I think this should be like this to avoid compiling in parallel if you ever decide to use the `-j` flag of `make`
```suggestion
test: nimble.paths
	$(MAKE) _test_all
	$(MAKE) testmultiformatexts
```

Makefile
```diff
@@ -13,11 +38,77 @@ deps: nix/deps.nix
...
+
```
richard-ramos, line 93: What should be the source of truth for `test` target? 'cause we still have `nimble test` target which does exactly the same, and i dont think maintaining both this target and the nimble tasks is the way to go

richard-ramos (commented):
> The goal is to centralize test execution logic in the Makefile so that developers can run the same commands locally as CI does, reducing the “works on my machine” surface area and making it easier to tweak compiler flags in one place.

can you expand a bit on what kind of situations do happen that make it not possible to use `nimble test` succesfully today?

Makefile
```diff
@@ -1,4 +1,29 @@
-.PHONY: all build deps cbind clean
```
vladopajic, line 1: would it be possible to include target 
`make test {{path}}` that replaces `nim c -r -d:path={{path}} tests/test_all.nim` 
that one would be really useful

Makefile
vladopajic: yes!

this effort should be to make `make` targets default source of truth. because it doesn't really make sense to compile nimble just for ` exec "nph ./. *.nim"` commands. not even for tests... 

we can have nimble just as package manager `nimble install` `nimble setup` and no more from that. all other day-to-day targets should be `make`

Makefile
```diff
@@ -1,4 +1,29 @@
...
+.PHONY: all build deps cbind clean test _test_all testmultiformatexts testintegration testpath \
```
vladopajic: other target would be `make clean-nim`:

```
rm -rf ~/.nimble
installs nim using choose nim
sets nim to 2.2.10
nimble install nimble
nimble install nph@0.7.0
```

i need this usually

.github/workflows/daily_common.yml
```diff
@@ -163,11 +163,11 @@ jobs:
...
+          make test
 
       - name: Run integration tests
         if: ${{ matrix.config.os == 'linux' && matrix.config.cpu == 'amd64' }}
         run: |
-          nim --version
-          nimble --version
-          nim r -d:libp2p_autotls_support -d:chronicles_log_level=DEBUG tests/integration/test_all.nim
+          export NIMFLAGS="${NIMFLAGS} -d:chronicles_log_level=DEBUG"
+          make testintegration
```
vladopajic: `test_integration`

libp2p.nimble
```diff
@@ -122,6 +58,3 @@ task unpin, "Restore global package use":
 task gen_multicodec,
   "Download the multicodec CSV and regenerate libp2p/multicodec_table.nim":
   exec nimc & " c -r tools/gen_multicodec.nim"
```
vladopajic, line 60: this can easily be `make` target; it's one liner

Makefile
```diff
@@ -13,11 +52,87 @@ deps: nix/deps.nix
...
+	nimble unpin
+
+gen_multicodec:
+	nimble gen_multicodec
+
+format:
+	find . -name '*.nim' -not -path './nimbledeps/*' | xargs nph
+
+clean-nim:
+	rm -rf ~/.nimble/pkgs2 ~/.nimble/pkgcache
```
vladopajic: this does not work for me really it has to be clean install. everything has to be removed under `.nimble` directory;

vladopajic: not that if everything is removed, choosenim has to be installed again `curl https://nim-lang.org/choosenim/init.sh -sSf | sh`

vladopajic (approved):

Makefile
```diff
@@ -13,11 +52,87 @@ deps: nix/deps.nix
...
+unpin:
+	nimble unpin
+
+gen_multicodec:
+	nimble gen_multicodec
+
+format:
+	find . -name '*.nim' -not -path './nimbledeps/*' | xargs nph
+
+clean-nim:
```
richard-ramos, line 133: We shouldn't have this target, as it has a scope larger than nim-libp2p and you are basically nuking up your whole nim installation.

In any case, perhaps what it should do instead is to remove:
1 - ./nimbledeps directory
2 - nimble.locks
3 - nimble.paths

If one needs to clean the whole nimble directory to fix a build issue, this should be instead reported to nimble.

also, you could have instead a custom sh script that does this in your ~/bin directory .

gmelodie:
> can you expand a bit on what kind of situations do happen that make it not possible to use nimble test succesfully today?

The part you highlighted was not on point indeed. As to why the makefile is useful: it reduces nimble overhead on many testing tasks.

richard-ramos (approved):

PR #2461, opened 2026-05-12
test(service-disco): lookup component

rlve:
## Summary

Three new tests for `lookup()`:

- Discoverer drops ads returned by a registrar when the ad's XPR services don't match the requested key.
- Once one registrar fills the result up to F_lookup, no further buckets are queried.
- Buckets are visited in ascending index order (farthest first).

The order-dependent tests use a `setupRegistrarsInDistinctBuckets` helper that spawns registrars until two land in different buckets and returns them in querying order.

Also adds a `randomKey()` helper in `utils.nim`.

## Affected Areas

- [x] Tests

## Impact on Library Users

None.

## Risk Assessment

None.


SionoiS (approved):
LGTM Thank you!

richard-ramos (approved):

gmelodie (approved):

PR #2462, opened 2026-05-13
refactor(Switch): Service lifecycle

vladopajic:
changes:
- return value `bool` from all `Service` method is remove as it was unused (discarded) thus making it unnecessary 
- `Switch` will call `setup()` on all `Service`s right after it is created, this is the time for `Service` to get anything they need form `Switch`
- `setup()` now raises `ServiceSetupError` when service could not be setup
  -  previously errors when setting up service did nothing
- `run()` was renamed to `start()`
- `start()` is called right when `Switch` is started
   - previously it was called from `setup`, this was not okay as switch was not starting at that point
- `stop()` is unchanged

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -123,7 +123,6 @@ switch("warningAsError", "UnusedImport:on")
 switch("warningAsError", "UseBase:on")
 switch("hintAsError", "ConvFromXtoItselfNotNeeded:on")
 switch("hintAsError", "DuplicateModuleImport:on")
-switch("hintAsError", "XCannotRaiseY:on")
 --styleCheck: usages
 --styleCheck: error
 --mm: refc          # Reference counting (not ORC yet)
--- a/libp2p/autotls/service.nim
+++ b/libp2p/autotls/service.nim
@@ -153,11 +153,11 @@ when defined(libp2p_autotls_support):
         self.config.ipAddress = Opt.some(getPublicIPAddress())
       except ValueError as e:
         raise newException(
-          ServiceSetupError, "Failed to get public IP address. Reason" & $e.msg
+          ServiceSetupError, "Failed to get public IP address. Reason: " & $e.msg
         )
       except OSError as e:
         raise newException(
-          ServiceSetupError, "Failed to get public IP address. Reason" & $e.msg
+          ServiceSetupError, "Failed to get public IP address. Reason: " & $e.msg
         )
 
   method issueCertificate(
--- a/libp2p/autotls/service.nim
+++ b/libp2p/autotls/service.nim
@@ -266,7 +266,7 @@ when defined(libp2p_autotls_support):
       await sleepAsync(self.config.issueRetryTime)
     error "Failed to issue certificate"
 
-  method run*(
+  method start*(
       self: AutotlsService, switch: Switch
   ) {.async: (raises: [CancelledError]).} =
     trace "Starting Autotls management"
--- a/libp2p/autotls/service.nim
+++ b/libp2p/autotls/service.nim
@@ -153,9 +153,10 @@ when defined(libp2p_autotls_support):
         self.config.ipAddress = Opt.some(getPublicIPAddress())
       except ValueError, OSError:
         raise newException(
-          ServiceSetupError, "Failed to get public IP address. Reason: " & getCurrentExceptionMsg()
+          ServiceSetupError,
+          "Failed to get public IP address. Reason: " & getCurrentExceptionMsg(),
         )
-        
+
   method issueCertificate(
       self: AutotlsService
   ): Future[bool] {.
--- a/libp2p/autotls/service.nim
+++ b/libp2p/autotls/service.nim
@@ -151,15 +151,11 @@ when defined(libp2p_autotls_support):
     if self.config.ipAddress.isNone():
       try:
         self.config.ipAddress = Opt.some(getPublicIPAddress())
-      except ValueError as e:
+      except ValueError, OSError:
         raise newException(
-          ServiceSetupError, "Failed to get public IP address. Reason: " & $e.msg
+          ServiceSetupError, "Failed to get public IP address. Reason: " & getCurrentExceptionMsg()
         )
-      except OSError as e:
-        raise newException(
-          ServiceSetupError, "Failed to get public IP address. Reason: " & $e.msg
-        )
-
+        
   method issueCertificate(
       self: AutotlsService
   ): Future[bool] {.
--- a/libp2p/autotls/service.nim
+++ b/libp2p/autotls/service.nim
@@ -146,23 +146,19 @@ when defined(libp2p_autotls_support):
       rng: rng,
     )
 
-  method setup*(
-      self: AutotlsService, switch: Switch
-  ): Future[bool] {.async: (raises: [CancelledError]).} =
+  method setup*(self: AutotlsService, switch: Switch) {.raises: [ServiceSetupError].} =
     trace "Setting up AutotlsService"
-    let hasBeenSetup = await procCall Service(self).setup(switch)
-    if hasBeenSetup:
-      if self.config.ipAddress.isNone():
-        try:
-          self.config.ipAddress = Opt.some(getPublicIPAddress())
-        except ValueError as exc:
-          error "Failed to get public IP address", err = exc.msg
-          return false
-        except OSError as exc:
-          error "Failed to get public IP address", err = exc.msg
-          return false
-      self.managerFut = self.run(switch)
-    return hasBeenSetup
+    if self.config.ipAddress.isNone():
+      try:
+        self.config.ipAddress = Opt.some(getPublicIPAddress())
+      except ValueError as e:
+        raise newException(
+          ServiceSetupError, "Failed to get public IP address. Reason" & $e.msg
+        )
+      except OSError as e:
+        raise newException(
+          ServiceSetupError, "Failed to get public IP address. Reason" & $e.msg
+        )
 
   method issueCertificate(
       self: AutotlsService
@@ -283,28 +279,31 @@ when defined(libp2p_autotls_support):
       error "Could not find a running TcpTransport in switch"
       return
 
-    heartbeat "Certificate Management", self.config.renewCheckTime:
-      if self.cert.isNone():
-        await self.tryIssueCertificate()
+    proc manageCert() {.async: (raises: []).} =
+      try:
+        heartbeat "Certificate Management", self.config.renewCheckTime:
+          if self.cert.isNone():
+            await self.tryIssueCertificate()
 
-      # AutotlsService will renew the cert 1h before it expires
-      let cert = self.cert.valueOr:
-        error "Could not issue certificate"
-        return
-      let waitTime = cert.expiry - Moment.now - self.config.renewBufferTime
-      if waitTime <= self.config.renewBufferTime:
-        await self.tryIssueCertificate()
+          # AutotlsService will renew the cert 1h before it expires
+          let cert = self.cert.valueOr:
+            error "Could not issue certificate"
+            return
+          let waitTime = cert.expiry - Moment.now - self.config.renewBufferTime
+          if waitTime <= self.config.renewBufferTime:
+            await self.tryIssueCertificate()
+      except CancelledError:
+        trace "Autotls management cancelled"
+
+    self.managerFut = manageCert()
 
   method stop*(
       self: AutotlsService, switch: Switch
-  ): Future[bool] {.async: (raises: [CancelledError]).} =
-    let hasBeenStopped = await procCall Service(self).stop(switch)
-    if hasBeenStopped:
-      if not self.acmeClient.isNil():
-        await self.acmeClient.close()
-      if not self.brokerClient.isNil():
-        await self.brokerClient.close()
-      if not self.managerFut.isNil():
-        await self.managerFut.cancelAndWait()
-        self.managerFut = nil
-    return hasBeenStopped
+  ) {.async: (raises: [CancelledError]).} =
+    if not self.acmeClient.isNil():
+      await self.acmeClient.close()
+    if not self.brokerClient.isNil():
+      await self.brokerClient.close()
+    if not self.managerFut.isNil():
+      await self.managerFut.cancelAndWait()
+      self.managerFut = nil
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -463,6 +463,7 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError].} =
       PeerStore.new(identify, capacity)
     else:
       PeerStore.new(identify)
+  peerStore.addressPolicy = b.addressPolicy
 
   if b.enableWildcardResolver:
     b.services.add(WildcardAddressResolverService.new())
@@ -479,8 +480,6 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError].} =
   if b.identifyPusherEnabled:
     b.services.add(IdentifyPusher.new())
 
-  peerStore.addressPolicy = b.addressPolicy
-
   let switch = newSwitch(
     peerInfo = peerInfo,
     transports = transports,
@@ -492,6 +491,7 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError].} =
     services = b.services,
     rng = b.rng,
   )
+  switch.setupServices()
 
   switch.mount(identify)
 
--- a/libp2p/protocols/connectivity/autonat/service.nim
+++ b/libp2p/protocols/connectivity/autonat/service.nim
@@ -221,6 +221,7 @@ method start*(
 
   if self.enableAddressMapper:
     switch.peerInfo.addressMappers.add(self.addressMapper)
+    await switch.peerInfo.update()
 
   self.scheduleInterval.withValue(interval):
     if self.scheduleHandle.isNil:
--- a/libp2p/protocols/connectivity/autonat/service.nim
+++ b/libp2p/protocols/connectivity/autonat/service.nim
@@ -210,7 +210,6 @@ method setup*(self: AutonatService, switch: Switch) {.raises: [ServiceSetupError
     ): Future[void] {.async: (raises: [CancelledError]).} =
       discard askPeer(self, switch, peerId)
 
-
 method start*(
     self: AutonatService, switch: Switch
 ) {.async: (raises: [CancelledError]).} =
@@ -219,7 +218,7 @@ method start*(
   switch.connManager.addPeerEventHandler(
     self.newConnectedPeerHandler, PeerEventKind.Joined
   )
-  
+
   if self.enableAddressMapper:
     switch.peerInfo.addressMappers.add(self.addressMapper)
 
--- a/libp2p/protocols/connectivity/autonat/service.nim
+++ b/libp2p/protocols/connectivity/autonat/service.nim
@@ -210,15 +210,16 @@ method setup*(self: AutonatService, switch: Switch) {.raises: [ServiceSetupError
     ): Future[void] {.async: (raises: [CancelledError]).} =
       discard askPeer(self, switch, peerId)
 
-    switch.connManager.addPeerEventHandler(
-      self.newConnectedPeerHandler, PeerEventKind.Joined
-    )
 
 method start*(
     self: AutonatService, switch: Switch
 ) {.async: (raises: [CancelledError]).} =
   trace "Running AutonatService"
 
+  switch.connManager.addPeerEventHandler(
+    self.newConnectedPeerHandler, PeerEventKind.Joined
+  )
+  
   if self.enableAddressMapper:
     switch.peerInfo.addressMappers.add(self.addressMapper)
 
--- a/libp2p/protocols/connectivity/autonat/service.nim
+++ b/libp2p/protocols/connectivity/autonat/service.nim
@@ -175,7 +175,7 @@ proc schedule(
     service: AutonatService, switch: Switch, interval: Duration
 ) {.async: (raises: [CancelledError]).} =
   heartbeat "Scheduling AutonatService run", interval:
-    await service.run(switch)
+    await service.askConnectedPeers(switch)
 
 proc addressMapper(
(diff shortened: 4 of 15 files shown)


.github/copilot-instructions.md
```diff
@@ -123,7 +123,6 @@ switch("warningAsError", "UnusedImport:on")
 switch("warningAsError", "UseBase:on")
 switch("hintAsError", "ConvFromXtoItselfNotNeeded:on")
 switch("hintAsError", "DuplicateModuleImport:on")
-switch("hintAsError", "XCannotRaiseY:on")
```
vladopajic, line 126: dropping `XCannotRaiseY` because some methods of `Service` raise `ServiceSetupError` some don't. and because of this hint compiling is getting errors. 

what we could do is disable hints in specific services, like it was done here:
https://github.com/vacp2p/nim-libp2p/pull/2462/changes#diff-a99c4291e04a9bde36fc1088e4d03053d57e0fcb54d659fcbc4b313362cf7278L626

but it would be to much effort for nothing...

libp2p/switch.nim
```diff
@@ -186,6 +181,10 @@ proc dial*(
 
   dial(s, peerId, addrs, @[proto])
 
+proc setupServices*(s: Switch) {.raises: [ServiceSetupError].} =
+  for service in s.services:
+    service.setup(s)
+
```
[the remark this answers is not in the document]
vladopajic: there is no need for this; there is and probably will always be, only one location where `setupServices` is called

libp2p/switch.nim
```diff
@@ -349,26 +348,13 @@ proc start*(s: Switch) {.async: (raises: [CancelledError, LPError]).} =
     return
 
   debug "starting switch for peer", peerInfo = s.peerInfo
+
   var startFuts: seq[Future[void]]
   for t in s.transports:
     let addrs = s.peerInfo.listenAddrs.filterIt(t.handles(it))
-
     s.peerInfo.listenAddrs.keepItIf(it notin addrs)
```
vladopajic, line 347: what is idea of this two lines?

vladopajic: `s.peerInfo.listenAddrs` should always be empty at this point?

libp2p/switch.nim
```diff
@@ -349,26 +348,13 @@ proc start*(s: Switch) {.async: (raises: [CancelledError, LPError]).} =
...
-        let ready = newAsyncEvent()
-        s.acceptFuts.add(s.accept(t, some(ready)))
-        s.peerInfo.listenAddrs &= t.addrs
-        await ready.wait()
-
-  # some transports require some services to be running
-  # in order to finish their startup process
-  for service in s.services:
-    discard await service.setup(s)
+    startFuts.add(t.start(addrs))
```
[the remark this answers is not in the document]
vladopajic: no, `Switch` should not do these checks. it should just delegate start / stop without knowing details.

tests/libp2p/protocols/test_autonat_service.nim
```diff
@@ -273,20 +273,6 @@ suite "Autonat Service":
...
-
-    check (await autonatService.setup(switch)) == true
-    check (await autonatService.setup(switch)) == false
-
-    check (await autonatService.stop(switch)) == true
-    check (await autonatService.stop(switch)) == false
-
-    await allFuturesRaising(switch.stop())
-
   asyncTest "Must bypass maxConnectionsPerPeer limit":
```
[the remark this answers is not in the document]
vladopajic: no one should ever call `setup` twice.

libp2p/autotls/service.nim
```diff
@@ -146,23 +146,19 @@ when defined(libp2p_autotls_support):
...
+      try:
+        self.config.ipAddress = Opt.some(getPublicIPAddress())
+      except ValueError as e:
+        raise newException(
+          ServiceSetupError, "Failed to get public IP address. Reason: " & $e.msg
+        )
+      except OSError as e:
+        raise newException(
+          ServiceSetupError, "Failed to get public IP address. Reason: " & $e.msg
+        )
```
gmelodie: I don't think we can do the following, can we?

```suggestion
      except ValueError, OSError as e:
        raise newException(
          ServiceSetupError, "Failed to get public IP address. Reason: " & $e.msg
        )
```
The fact that we can't is... unbelievable

libp2p/autotls/service.nim
```diff
@@ -283,28 +279,31 @@ when defined(libp2p_autotls_support):
...
+          let cert = self.cert.valueOr:
+            error "Could not issue certificate"
+            return
+          let waitTime = cert.expiry - Moment.now - self.config.renewBufferTime
+          if waitTime <= self.config.renewBufferTime:
+            await self.tryIssueCertificate()
+      except CancelledError:
+        trace "Autotls management cancelled"
+
+    self.managerFut = manageCert()
```
gmelodie, line 295: very nice, but do you not have to `raise` the `CancelledError`?

```suggestion
    proc manageCert() {.async: (raises: []).} =
      try:
        heartbeat "Certificate Management", self.config.renewCheckTime:
          if self.cert.isNone():
            await self.tryIssueCertificate()

          # AutotlsService will renew the cert 1h before it expires
          let cert = self.cert.valueOr:
            error "Could not issue certificate"
            return
          let waitTime = cert.expiry - Moment.now - self.config.renewBufferTime
          if waitTime <= self.config.renewBufferTime:
            await self.tryIssueCertificate()
      except CancelledError as e:
        trace "Autotls management cancelled"
        raise e

    self.managerFut = manageCert()
```

libp2p/protocols/connectivity/autonat/service.nim
```diff
@@ -196,58 +196,49 @@ proc addressMapper(
...
+    switch.peerInfo.addressMappers.add(self.addressMapper)
 
-method run*(
+method start*(
     self: AutonatService, switch: Switch
 ) {.async: (raises: [CancelledError]).} =
   trace "Running AutonatService"
-  await askConnectedPeers(self, switch)
+  self.scheduleInterval.withValue(interval):
+    if self.scheduleHandle.isNil:
```
gmelodie, line 227: Probably not the scope of this PR but... `scheduleHandle` and `scheduleInterval` should be inside the same `Opt`, no? Since you don't really use one without the other.

gmelodie (approved):

libp2p/autotls/service.nim
vladopajic: to whome does it raise to?

libp2p/autotls/service.nim
vladopajic: i think this is possible if getErrorMessage() is used

libp2p/protocols/connectivity/autonat/service.nim
vladopajic: not sure... `scheduleInterval` is passed down from constructor, `scheduleHandle` is created inside.

libp2p/switch.nim
```diff
@@ -379,16 +365,16 @@ proc start*(s: Switch) {.async: (raises: [CancelledError, LPError]).} =
...
       )
 
-  for t in s.transports: # for each transport
-    if t.addrs.len > 0 or t.running:
-      if t of TcpTransport:
-        continue # already added previously
-      s.acceptFuts.add(s.accept(t))
-      s.peerInfo.listenAddrs &= t.addrs
+  for t in s.transports:
+    s.acceptFuts.add(s.accept(t))
```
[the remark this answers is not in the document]
vladopajic: can never happen

vladopajic:
@copilot stop reviewing this pr

libp2p/autotls/service.nim
vladopajic: @copilot+gpt-5.3-codex explain what will happen with `CancelledError` in this case? to whom is this error raised to? who raises cancelled error and when?

libp2p/switch.nim
```diff
@@ -349,26 +348,13 @@ proc start*(s: Switch) {.async: (raises: [CancelledError, LPError]).} =
...
 
-    if addrs.len > 0 or t.running:
-      let fut = t.start(addrs)
-      startFuts.add(fut)
-      if t of TcpTransport:
-        await fut
-        let ready = newAsyncEvent()
-        s.acceptFuts.add(s.accept(t, some(ready)))
-        s.peerInfo.listenAddrs &= t.addrs
-        await ready.wait()
```
richard-ramos, line 366: In switch in line 265, it says:
```
      # Signal readiness after acquiring the slot but before awaiting
      # transport.accept(). In chronos, fire() schedules the waiter
      # callback without suspending here, so the loop continues
      # synchronously into transport.accept() — on Windows this posts
      # AcceptEx before switch.start() returns to the caller.
```

removing these lines of code means that we are starting the accept loop without awaiting for the ready event.

The documentation still comments about the need for this readiness event for windows+tcp, so getting rid of this introduces a possible race condition again, right? how is this solved?

cc: @gmelodie

libp2p/switch.nim
```diff
@@ -349,26 +348,13 @@ proc start*(s: Switch) {.async: (raises: [CancelledError, LPError]).} =
...
-
-  # some transports require some services to be running
-  # in order to finish their startup process
-  for service in s.services:
-    discard await service.setup(s)
```
richard-ramos, line 371: This one is also an interesting comment, 'cause for example WsTransport requires AutoTLS to start first when starting a secure ws listener, without passing a manual cert, notice how it waits for autls.running before continuing in wstransport.nim line 329. 

So the flow with this change is +- like this IIUC: 
1. WsTransport.start begins.
2. It waits for autotls.running.
3. AutotlsService.start has not been called yet.
4. `Switch.start` is waiting for `WsTransport.start` to finish before calling `AutotlsService.start`.
5. WS times out after `DefaultAutotlsWaitTimeout`, stops itself, and returns.
Then AutoTLS starts, but it will be too late for that WS listener.

How is this solved in this case?

libp2p/builders.nim
```diff
@@ -492,6 +491,7 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError].} =
     services = b.services,
     rng = b.rng,
   )
+  switch.setupServices()
```
richard-ramos, line 494: Based on this comment on the description:
> Switch will call setup() on all Services right after it is created, this is the time for Service to get anything they need form Switch

we should probably move this line to 528, because there are still items being setup in the switch (from lines 496 to to 527 that services might need in the future

libp2p/services/hpservice.nim
```diff
@@ -87,52 +87,48 @@ proc newConnectedPeerHandler(
...
-        discard await self.autoRelayService.setup(switch)
-      elif networkReachability == NetworkReachability.Reachable and
-          self.autoRelayService.isRunning():
-        discard await self.autoRelayService.stop(switch)
+  self.onNewStatusHandler = proc(
+      networkReachability: NetworkReachability, confidence: Opt[float]
+  ) {.async: (raises: [CancelledError]).} =
+    if networkReachability == NetworkReachability.NotReachable and
+        not self.autoRelayService.isRunning():
+      await self.autoRelayService.start(switch)
```
richard-ramos, line 116: autorelay can lose its address mapper after it is stopped once. notice that `setup()` adds the mapper, `stop()` will remove it, but `start()` does not add it back. So if reachability changes, for example: `NotReachable` to `Reachable` to  `NotReachable`, autorelay starts again but its relay addresses are no longer added to peerInfo.addrs. 

So we might to take the mappers that are currently in `setup` in line 79 of autorelayservice.nim and move them `start`

libp2p/services/autorelayservice.nim
```diff
@@ -143,14 +137,11 @@ method run*(
...
-  let hasBeenStopped = await procCall Service(self).stop(switch)
-  if hasBeenStopped:
-    self.running = false
-    self.runner.cancelSoon()
-    switch.peerInfo.addressMappers.keepItIf(it != self.addressMapper)
-    await switch.peerInfo.update()
-  return hasBeenStopped
+) {.async: (raises: [CancelledError]).} =
+  self.running = false
+  self.runner.cancelSoon()
```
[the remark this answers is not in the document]
richard-ramos: Why is this not the case?  the code in switch indicates that a transport might fail to start, and will call the switch `stop` when that happens, stopping all services,

```nim
  for fut in startFuts:
    if fut.failed:
      await s.stop()
      raise newException(
        LPError, "starting transports failed: " & $fut.error.msg, fut.error
      )
```

But then, the service startup happens later in line 75, so it should be indeed possible for this error to happen, since autorelay only sets the runner in start()

libp2p/services/wildcardresolverservice.nim
```diff
@@ -171,19 +171,17 @@ method setup*(
...
       listenAddrs: seq[MultiAddress]
   ): Future[seq[MultiAddress]] {.async: (raises: [CancelledError]).} =
     return expandWildcardAddresses(self.networkInterfaceProvider, listenAddrs)
 
-  debug "Setting up WildcardAddressResolverService"
-  let hasBeenSetup = await procCall Service(self).setup(switch)
-  if hasBeenSetup:
-    switch.peerInfo.addressMappers.add(self.addressMapper)
-  return hasBeenSetup
+  switch.peerInfo.addressMappers.add(self.addressMapper)
```
richard-ramos: to keep the lifecycle consistent, this should be moved to start(), otherwise you are not able to start/stop/start this service, otherwise you lose these mappers since they are removed on stop

(which is strange, i admit, but it can happen (like it does with autorelay which starts/stops from time to time)

libp2p/protocols/connectivity/autonat/service.nim
```diff
@@ -196,58 +196,49 @@ proc addressMapper(
...
 
-    if self.enableAddressMapper:
-      switch.peerInfo.addressMappers.add(self.addressMapper)
+    switch.connManager.addPeerEventHandler(
+      self.newConnectedPeerHandler, PeerEventKind.Joined
+    )
 
-  return hasBeenSetup
+  if self.enableAddressMapper:
+    switch.peerInfo.addressMappers.add(self.addressMapper)
```
richard-ramos: to keep the lifecycle consistent, the peer event handler and the address mapper should be moved to start(), since they are removed on stop

libp2p/protocols/connectivity/autonatv2/service.nim
```diff
@@ -241,27 +236,22 @@ method setup*(
       self.newConnectedPeerHandler, PeerEventKind.Joined
     )
 
-  self.config.scheduleInterval.withValue(interval):
-    self.scheduleHandle = schedule(self, switch, interval)
-
   if self.config.enableAddressMapper:
     switch.peerInfo.addressMappers.add(self.addressMapper)
```
richard-ramos: to keep the lifecycle consistent, the peer event handler and the address mapper should be moved to start(), since they are removed on stop

libp2p/services/autorelayservice.nim
```diff
@@ -54,35 +54,29 @@ proc reserveAndUpdate(
...
+  proc handlePeerLeft(
+      peerId: PeerId, event: PeerEvent
+  ) {.async: (raises: [CancelledError]).} =
+    trace "Peer Left", peerId
+    self.relayPeers.withValue(peerId, future):
+      future[].cancelSoon()
+
+  switch.addPeerEventHandler(handlePeerIdentified, Identified)
+  switch.addPeerEventHandler(handlePeerLeft, Left)
+  switch.peerInfo.addressMappers.add(self.addressMapper)
```
richard-ramos: to keep the lifecycle consistent, the address mapper should be moved to start(), since it is removed on stop

richard-ramos (changes requested):

libp2p/services/autorelayservice.nim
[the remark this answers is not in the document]
vladopajic: aha, now i see!

libp2p/switch.nim
vladopajic: i don't understand this code or it doesn't really make sense. 
for now i have removed it.

libp2p/builders.nim
vladopajic: my first tough wast to put `setupServices` inside `newSwitch`, because it is the place where `Switch` and `Service`s meet. then, decided to put it in method, my reasoning was this:
- lets have `newSwitch` to be just constructor - there is no right/wrong here doing `setupServices` inside is also 100% clean
- builder is building switch so i wanted that place to be responsible of building everything; therefore exposing `setupServices` here was for `build()` - it is building and not switch

 
for now i would not like to move it at later stages of build, because if in future someone needs change like that than doing it would be less effort then writing this comment; and in the same time it will give signal in code review that there are new dependencies introduced. `Service` should really depend on components that make up `Switch` in construction|creation time (`newSwitch`), and they should not really depend on things that happen after that. 

tldr; what comment "right after it is created" meant there was that setup is called right after `newSwitch` is created (all components that make up switch are created), and not that switch is fully built by `build()` proc.

libp2p/switch.nim
vladopajic: fixed it now by starting services and transports asynchronously and not awaiting on them separately. this will always give opportunity to one that is depended on to complete start.

libp2p/switch.nim
richard-ramos: fair enough. The reason I asked was that this code in particular was added last week in https://github.com/vacp2p/nim-libp2p/pull/2360/ , so i find it curious it is now being reverted.

libp2p/builders.nim
richard-ramos: I understand the point, but the lifecycle is easier to understand if `setupServices()` runs after the builder has finished configuring the switch.

In this PR `build()` still changes the switch after `setup()` is called. For a service author, I would expect `setup()` to receive the final switch, just before it can be started, or have documentation indicating what is not expected to be available in the switch the moment a service is setup.

Also needed to take into consideration: services not only come from nim-libp2p, but might come from downstream, as we have the `.withServices` option, so users of nim-libp2p are able to define their own service, and this wont be caught in a review

vladopajic: > and this wont be caught in a review

it would... if downstream dependency has some code - the new service and suddenly location of `setupServices` is important, then they would need to report issue or open pr. and if this case happens, it is better for us to know when this is happening.

another thing is that most of that code (after `setupServices`)  should not be there, it should be moved before switch is created; those are also services and they should be created using the same approach - there is no difference between `RendezVous` and `IdentifyPusher`. 
this should be done in future effort to make refactoring transition smoother.

richard-ramos: Alright, approving since this will be solved in a future PR.

That said, for downstream users, I do not think “they can report an issue later” is a good approach. Since `.withServices()` is public, users may reasonably expect `setup()` to receive the switch after the builder has finished configuring it, but before `start()`.

If we already know this is prone to issues, we should avoid putting the burden on users to identify and report them later

richard-ramos (approved):

libp2p/builders.nim
vladopajic: ... whole discussion about downstream dependencies is irrelevant here. 

the only place where `withServices` metter is within `build` method. any scope outside of it is irrelevant.


downstream deps, or tests, or any other use will do this:
```
let switch = builder.build()
# add service or protocol, or whatever
# at this place it is irrelevant where `withServices` is called, because 
# it is integral part of build()
```

---

and inside `build()`, it should never matter where it is called. because there should not be dependencies between services other then dependency on  `Switch`, and that should happen right as soon as those two meet.

richard-ramos: i understand the design goal that services should only depend on things that exist when the switch is created, but that's not what i am discussing about. My concern is about the current API behavior for anyone that one day decide to create custom services.

currently `setup()` is called before `build()` finishes configuring the switch, but this is not obvious from the API. for someone writing a custom service with `.withServices`, `setup(switch)` looks like the switch is already fully configured.

i created a small downstream example to show the issue more clearly:

https://github.com/richard-ramos/downstream-nim-libp2p-app

if `setupServices()` is not moved after the switch configuration is finished, then i think we should clearly document that `setup()` is an early initialization step, and that services should not rely on data from switch attributes available during setup, since some parts of the switch may still change after it.

otherwise those custom services may erroneously depend on build-order details that are currently implicit and undocumented.

vladopajic: 1) custom services don't need `withServices()` from `SwitchBuilder`:

```
let switch = builder.build()

# do whatever you want, in whatever order you need
let myService = MyService.new(switch)
switch.mount(myService.myProtocol)
```



2) service can mount protocol


```
method setup*(
    self: CapabilityAdvertiserService, switch: Switch
) {.raises: [ServiceSetupError].} =
  self.myCapabilityRecord = CapabilityRecord(
    peerId: $switch.peerInfo.peerId,
    protocols: sortedCopy(switch.peerInfo.protocols),
  )
  
  
  # it is 100% fine for custom service to mount protocol at this point
  ```


but in general `withServices` shouldn't really be used for anything complex.  the app should you switch builder to create switch, then mounting custom protocols and custom services is independent on building switch. `build` should not go that far to solve hypothetical problems that are unnecessary for it, since mounting and setting up extras is at users discretion - switch is already there.

vladopajic: the more i think about this, it seems that `withServices` should not be exposed by `SwitchBuilder`. as it doesn't do anything. it just calls `setup(switch)` after switch is built, but that completely unnecessary, as users will have switch few lines after.

`withServices`, or rather `Service` type is only needed by `Switch`, it should be internal logic that only helps `nim-libp2p` to implement components of switch.

all user defined protocols or services are-should be added after `Switch` is created. and we don't really need to know how those `Services` look like as they do not shape `Switch`.

____


- it completely unnecessary for user to create Service early on, just to pass it to `withServices`. it doesn't have any impact on `Switch`
- user services can be of whatever type, and whatever lifecycle, it is unnecessary for user to comply to `Switch-Service`
   -  users service can just be `future` it doesn't have to be `Service`
- Service should only remain as `nim-libp2p` internal type

vladopajic: i'll think more about this and create pr on monday if it still makes sense

vladopajic: i'll think more about this and create pr on monday if it still makes sense

PR #2464, opened 2026-05-13
test(service-disco): error handling component

rlve:
## Summary

Adds component error-handling test suite.

The new suite covers malformed and invalid protocol inputs:
- unknown `MessageType` closes without a reply
- `REGISTER` without a register body returns `Rejected`
- `REGISTER` with an empty advertisement returns `Rejected`
- `REGISTER` with an advertisement for another service returns `Rejected`

The existing malformed advertisement test was moved out of the register suite.

## Affected Areas

- [x] Tests

## Impact on Library Users

None.

## Risk Assessment

None.

### DIFF
--- a/tests/libp2p/service_discovery/component/test_error.nim
+++ b/tests/libp2p/service_discovery/component/test_error.nim
@@ -4,14 +4,13 @@
 
 import chronos, results
 import
-  ../../../../libp2p/
-    [
-      protobuf/minprotobuf,
-      protocols/service_discovery/advertiser,
-      protocols/service_discovery/types,
-      stream/connection,
-      switch,
-    ]
+  ../../../../libp2p/[
+    protobuf/minprotobuf,
+    protocols/service_discovery/advertiser,
+    protocols/service_discovery/types,
+    stream/connection,
+    switch,
+  ]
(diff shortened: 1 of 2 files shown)


vladopajic (approved):

SionoiS (approved):
Thank you!

tests/libp2p/service_discovery/component/test_error.nim
```diff
@@ -0,0 +1,150 @@
...
+    var pb = initProtoBuffer()
+    pb.write(1, 99'u32)
+    pb.finish()
+    let invalidMsg = pb.buffer
+
+    expect LPStreamError:
+      discard
+        await clientSwitch.sendRawMessage(registrarNode, invalidMsg).wait(2.seconds)
+
+  asyncTest "REGISTER without register body returns Rejected":
```
gmelodie, line 64: Could these last three tests be one? You could define the invalid messages and just assert that they all return rejected.

gmelodie (approved):

tests/libp2p/service_discovery/component/test_error.nim
rlve: They could, but I don't think it's worth to consolidate them. It wouldn't save much execution time and this way the intent is more clear.

PR #2466, opened 2026-05-14
chore(SwitchBuilder): `build()` proc cleanup

vladopajic:
## Summary

changes related to cleanup of `SwitchBuilder.build()` proc. previously `build()` was one big long proc. in this pr it was splitted in three sub proc that cleanly isolate each stage of building `Switch`:
- `buildSwitch` is first sub-stage that creates `Switch` with core components - it is MVP for `Switch`
- `setupServices` and `mountProtocols` are additional sub-stages that will add service and mount protocols

## Affected Areas

- [x] SwitchBuilder

## Impact on Library Users

only internal changes.

## Risk Assessment

low. no functional changes, only reorganization.


libp2p/switch.nim
```diff
@@ -181,10 +183,6 @@ proc dial*(
 
   dial(s, peerId, addrs, @[proto])
 
-proc setupServices*(s: Switch) {.raises: [ServiceSetupError].} =
```
vladopajic, line 184: it was wrong to expose this method. this is only need for building `Switch`, therefore it should not be exposed. interface of `Switch` should be to what is needed by apps.

libp2p/switch.nim
```diff
@@ -51,6 +52,7 @@ type
     peerInfo*: PeerInfo
     connManager*: ConnManager
     transports*: seq[Transport]
+    muxedUpgrade*: MuxedUpgrade
```
vladopajic, line 52: creation of `circuitRelay` depends on this component of `Switch`. it was not problem for nim-libp2p to have access to it it was created in `build()`. but, what if downstream dependency wants to have something similar to `circuitRelay`? 

in order to guarantee this, `setupServices` and `mountProtocols` are given `Switch` created in the first stage, so anything that service or protocol needs should be accessible from `Switch`. this will ensure that downstream dependence can access to same.

libp2p/switch.nim
```diff
@@ -370,33 +368,3 @@ proc start*(s: Switch) {.async: (raises: [CancelledError, LPError]).} =
   s.peerStore.startAddressPruning()
 
   debug "Started libp2p node", peer = s.peerInfo
```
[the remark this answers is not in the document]
vladopajic: this proc is 100% unnecessary, objects can be created manually.
and no one should use this in downstream deps.

gmelodie (approved):

tests/libp2p/protocols/test_noise.nim
```diff
@@ -72,8 +72,17 @@ proc makeSwitch(ma: MultiAddress, outgoing: bool, plaintext: bool = false): Swit
     ms = MultistreamSelect.new()
     muxedUpgrade = MuxedUpgrade.new(muxers, secureManagers, ms)
     transports = @[Transport(TcpTransport.new(upgrade = muxedUpgrade))]
-
-  newSwitch(peerInfo, transports, secureManagers, connManager, ms, peerStore)
+    dialer = Dialer.new(peerInfo.peerId, connManager, peerStore, transports, nil)
```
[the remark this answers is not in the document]
vladopajic: not really needed for these tests

libp2p/builders.nim
```diff
@@ -430,50 +430,63 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError].} =
     else:
       Identify.new(peerInfo, b.sendSignedPeerRecord)
 
-  let connManager = ConnManager.new(
+  var peerStore = block:
+    b.peerStoreCapacity.withValue(capacity):
+      PeerStore.new(identify, capacity, b.addressTtls)
+    else:
+      PeerStore.new(identify, addressTtls = b.addressTtls)
+  peerStore.addressPolicy = b.addressPolicy
```
vladopajic, line 438: in future, this better be constructor argument

libp2p/builders.nim
```diff
@@ -430,50 +430,63 @@ proc build*(b: SwitchBuilder): Switch {.raises: [LPError].} =
...
+
+  var connManager = ConnManager.new(
     maxConnsPerPeer = b.maxConnsPerPeer,
     limits = b.limits,
     watermark = b.watermark,
     scoring = b.scoring,
   )
+  connManager.peerStore = peerStore
```
vladopajic, line 446: this better be constructor argument

richard-ramos (approved):

PR #2469, opened 2026-05-14
feat: use BoringSSL for noise crypto primitives

richard-ramos:
## Summary
Replaces the BearSSL-backed crypto helpers used by Noise with BoringSSL implementations for ChaCha20-Poly1305, Curve25519/X25519, and HKDF.

The change also propagates BoringSSL failure reporting through the Noise path: AEAD decrypt and Curve25519 multiplication now return failure status, and Noise raises a handshake error when DH fails.

## Affected Areas
- [x] Transports: Noise secure channel crypto and handshake error handling.
- [x] Other: crypto primitives in `libp2p/crypto`.

## Impact on Library Users
Direct callers of `ChaChaPoly.decrypt` and `Curve25519.mul` now need to handle the returned `bool`.

## Risk Assessment
The main risk is behavior drift from switching crypto backends, especially around AEAD authentication failures, empty inputs, HKDF output layout, and X25519 low-order point handling.

### DIFF
--- a/libp2p/crypto/hkdf.nim
+++ b/libp2p/crypto/hkdf.nim
@@ -6,7 +6,7 @@
 {.push raises: [].}
 
 import nimcrypto
-import bearssl/[kdf, hash]
+import boringssl
 
 type HkdfResult*[len: static int] = array[len, byte]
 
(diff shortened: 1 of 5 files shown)


gmelodie (approved):

arnetheduck:
I'm not convinced this is a great idea - the noise backend was audited together with bearssl for security issues and I don't quite see the practical benefit of switching here. Noise is on its way out (ie it's being replaced by QUIC where the new cryptography is being used) so I'd probably lean towards holding the current implementat constant instead of introducing this churn.

richard-ramos:
@arnetheduck,

i understand the concern about changing crypto code, and that the BearSSL code was audited, so we should not change it without a good reason, but I think there is still a good reason here:

indeed QUIC is the preferred transport when UDP works as it is supposed to befaster and already includes security and multiplexing. But QUIC does not fully replace TCP in libp2p. The QUIC spec says QUIC should be offered, but TCP is [recommended to be offered as fallback because UDP can be blocked](
https://github.com/libp2p/specs/tree/master/quic)

So TCP fallback according to the specs is still part of the expected libp2p deployment model. Specs also say libp2p still needs a secure channel and a stream muxer. The connection spec says the baseline security protocol is Noise, and the [baseline stream muxer is Yamux](https://github.com/libp2p/specs/tree/master/connections). So in practice, TCP + Noise + Yamux is still important as a fallback mechanism.

Because of that, I do not think we can treat Noise nor Yamux as dead/deprecated code. Even if QUIC is the main direction, Noise is still part of the fallback, and fallback security code still needs to be maintained.

The old audit is useful, but it is also old, because code and dependencies changes over time, and besides, the project now also has QUIC/BoringSSL in the stack. IMO we should do another audit, and I think it is better to audit a more modern crypto backend we would want to keep using in the future, instead of keeping BearSSL only because it was audited years ago.

There is also the [email exchange you had with Thomas Pornin](https://discord.com/channels/613988663034118151/785478304181387274/1412316063986876416). My understanding of it is that BearSSL TLS 1.3 and QUIC support are in limbo, and that he cannot promise the kind of active maintenance needed. That makes BearSSL a very risky dependency to have.

libp2p/crypto/hkdf.nim
```diff
@@ -15,33 +15,31 @@ proc hkdf*[T: sha256, len: static int](
...
-  var ctx: HkdfContext
-  hkdfInit(
-    ctx,
-    addr sha256Vtable,
+  if outputs.len == 0:
+    return
+
+  let totalLen = outputs.len * len
+  if totalLen == 0:
+    return
```
vladopajic, line 23: first `if` is not needed because of second `if`.

vladopajic (approved):

richard-ramos:
As discussed with @kaiserd, and based on comments above from @arnetheduck, closing this PR for now. 
https://github.com/vacp2p/nim-libp2p/issues/2474 was opened to track this initiative in case we want to revise it again in the future

PR #2472, opened 2026-05-15
test(Noise): cosmetics

vladopajic:
just a small cosmetic touch

### DIFF
--- a/tests/libp2p/protocols/test_noise.nim
+++ b/tests/libp2p/protocols/test_noise.nim
@@ -255,7 +255,8 @@ suite "Noise":
 
   asyncTest "e2e: test wrong secure negotiation":
     var switch1 = makeSwitch(ma, false)
-    var switch2 = makeSwitch(ma, true, true) # PlainText enabled; mismatched with Noise, so we want this to fail
+    var switch2 = makeSwitch(ma, true, true)
+      # PlainText enabled; mismatched with Noise, so we want this to fail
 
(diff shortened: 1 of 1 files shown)


tests/libp2p/protocols/test_noise.nim
```diff
@@ -50,15 +50,13 @@ method init(p: TestProto) {.gcsafe.} =
...
-    ma: MultiAddress, outgoing: bool, plaintext: bool = false
-): (Switch, PeerInfo) =
+proc makeSwitch(ma: MultiAddress, outgoing: bool, plaintext: bool = false): Switch =
   var
     privateKey = PrivateKey.random(ECDSA, rng()).get()
     peerInfo = PeerInfo.new(privateKey, @[ma])
 
   proc createMplex(conn: Connection): Muxer =
-    result = Mplex.new(conn)
+    Mplex.new(conn)
```
richard-ramos, line 59: It seems to me that at this point we could just call  Mplex.new(conn) directly. This proc only saves us from writing 4 extra characters 😆

richard-ramos (approved):

gmelodie (approved):

tests/libp2p/protocols/test_noise.nim
vladopajic: i tried, it doesn't work. i think it was because `createMplex` returns `Muxer` and `Mplex.new` returns `Mplex`.

PR #2475, opened 2026-05-18
chore(SwitchBuilder): drop `withServices`

vladopajic:
user defined services do not really need to be service, they can be future or anything else. `Service` should be only internal abstraction of `Switch`. but anything that user wants to build on top of `Switch` should not be dependency to creating and building `Switch`; also user should be able to choose how they want to organize code - forcing Service abstraction is unnecessary for all.

idea originated here: https://github.com/vacp2p/nim-libp2p/pull/2462#discussion_r3252656136 

----

to complete cleanup these issues should be addressed:
https://github.com/vacp2p/nim-libp2p/issues/2485
https://github.com/vacp2p/nim-libp2p/issues/2484


libp2p/builders.nim
```diff
@@ -333,13 +326,6 @@ proc withAutonatV2*(
   )
   b
 
-proc withAutonatV2Service*(
```
vladopajic, line 352: - this must not be exposed to users since they should not use this

libp2p/builders.nim
```diff
@@ -307,14 +306,8 @@ proc withNameResolver*(b: SwitchBuilder, nameResolver: NameResolver): SwitchBuil
   b.nameResolver = nameResolver
   b
 
-proc withAutonat*(b: SwitchBuilder): SwitchBuilder =
-  b.autonat = true
-  b
-
-proc withAutonatService*(
```
vladopajic, line 330: - this must not be exposed to users since they should not use this

libp2p/switch.nim
```diff
@@ -166,20 +168,27 @@ method dial*(
...
 proc dial*(
     s: Switch, peerId: PeerId, addrs: seq[MultiAddress], proto: string
-): Future[Stream] {.async: (raises: [DialFailedError, CancelledError], raw: true).} =
+): Future[Connection] {.async: (raises: [DialFailedError, CancelledError], raw: true).} =
   ## Connected to a peer and open a stream
   ## with the specified `proto`
 
   dial(s, peerId, addrs, @[proto])
 
+proc add*(s: Switch, service: Service) {.raises: [ServiceSetupError].} =
```
vladopajic: this shouldn't really exist (as public proc); any externally created service does not need to part of switch because only benefit is that that service will be started and stopped together with switch. but since switch is started and stopped externally then all these services could be started and stopped in the same way.

the reason why `add` is still here is because:
- test have this pattern and i didn't want to go into fixing those tests,
- if some users really want to add externally created services to switch, they can can have it

vladopajic: https://github.com/vacp2p/nim-libp2p/issues/2484

tests/libp2p/services/test_wildcard_resolver.nim
```diff
@@ -28,16 +28,19 @@ proc getAddressesMock(
...
     .new()
     .withRng(rng())
     .withAddresses(addrs, false)
     .withTcpTransport()
     .withMplex()
     .withNoise()
-    .withServices(@[svc])
     .build()
 
+  switch.add(svc)
```
vladopajic, line 40: adding externally created `svc` here is unnecessary

it is not needed for switch at all; it only means that this service will be started and stopped together with switch. but service could be started and stopped manually where switch is started and stopped - which is way to go.

same pattern is in all tests here.

richard-ramos (approved):

libp2p/switch.nim
```diff
@@ -180,6 +182,18 @@ proc dial*(
 
   dial(s, peerId, addrs, @[proto])
 
+proc add*(
```
gmelodie, line 185: `add`??? wtf
```suggestion
proc addService*(
```

libp2p/switch.nim
```diff
@@ -180,6 +182,18 @@ proc dial*(
...
+    s: Switch, service: Service
+) {.
+    raises: [ServiceSetupError],
+    deprecated: "externally created services should not be added to Switch"
+.} =
+  if service.isNil:
+    return
```
gmelodie, line 192: `debug`/`trace` log maybe?

gmelodie (approved):

libp2p/switch.nim
vladopajic: it is irrelevant; proc should be removed in future efforts

libp2p/switch.nim
vladopajic: it is irrelevant; proc should be removed in future efforts
adding nil service is okay since tests are doing that in some branches.

PR #2477, opened 2026-05-18
chore: dcutr protobuf serialization using nim-protobuf-serialization

moigagoo:
## Summary

Rewrite dcutr protobuf serialization using [nim-protobuf-serialization](https://github.com/status-im/nim-protobuf-serialization).

## Affected Areas
- MultiAddress type: make protobuf-serializable
- dcutr core: serialize nim-protobuf-serialization
- dcutr server: update the `decode` calls
- dcutr client: update the `decode` calls


libp2p/protocols/connectivity/dcutr/core.nim
```diff
@@ -7,48 +7,35 @@ import std/sequtils
...
+  MsgType* {.pure.} = enum
     Connect = 100
     Sync = 300
 
-  DcutrMsg* = object
-    msgType*: MsgType
-    addrs*: seq[MultiAddress]
+  DcutrMsg* {.proto2.} = object
+    msgType* {.fieldNumber: 1, required, ext.}: MsgType
+    addrs* {.fieldNumber: 2.}: seq[MultiAddress]
```
moigagoo: @nitely Here, I'm not using an extension to serialize MultiAddress but instead rely on the default serialization. To make it possible, VBuffer needs to be serializable as well, and therefore you see it being [annotated as well](https://github.com/vacp2p/nim-libp2p/pull/2477/changes#diff-b2e967a925d8fefc7f83349849b02e5b695cab9d336fb1be42cb819e31efc526R13).

If I put an `ext` here, I get an error: `Error: fatal error: Serializing FlatType as field type is not supported: DcutrMsg.addrs`

I can't understand what specifically is wrong by this error message.

libp2p/vbuffer.nim
```diff
@@ -5,12 +5,14 @@
...
+
 import varint, strutils
 import ./utils/sequninit
 
-type VBuffer* = object
-  buffer*: seq[byte]
-  offset*: int
+type VBuffer* {.proto2.} = object
+  buffer* {.fieldNumber: 1, required.}: seq[byte]
+  offset* {.fieldNumber: 2, required, sint.}: int64
```
moigagoo: `int` isn't serializable by default in nim-protobuf-serialization but int64 is.

This looks like a dirty hack because `int` could be `int32` on another platform.

libp2p/protocols/connectivity/dcutr/core.nim
nitely: > If I put an ext here, I get an error: Error: fatal error: Serializing FlatType as field type is not supported: DcutrMsg.addrs

That error is being raised from a function, probably one of: `readFieldInto`, `writeFieldInto`, `computeFieldSize`, that needs to be implemented for `MultiAddress`.

moigagoo: @nitely 

> That error is being raised from a function, probably one of: readFieldInto, writeFieldInto, computeFieldSize, that needs to be implemented for MultiAddress.

Do you mean `writeField`? `writeFieldInto` doesn't seem to exist.

Still, even with those procs overriden for `MultiAddress`, I'm getting the error `Error: fatal error: MultiAddress: missing {.proto2.} or {.proto3.}`. Annotating `MultiAddress` kind of goes against the goal of using extensions but even if I do annotate that, than the same error appears for `VBuffer`. And if I annotate VBuffer as well then it's basically my current PR, no extensions necessary.

I can't understand why `MultiAddress` requires `proto2` whereas `PeerId` doesn't.

nitely: the `missing {.proto2.}` error is fixed in https://github.com/status-im/nim-protobuf-serialization/pull/78

libp2p/multiaddress.nim
```diff
@@ -1263,3 +1263,41 @@ func shortLog*(addrs: seq[MultiAddress], maxAddrs: int): string =
...
+proc readFieldInto*(
+    stream: InputStream,
+    value: var MultiAddress,
+    header: FieldHeader,
+    ProtoType: type ProtobufExt,
+): bool {.raises: [SerializationError, IOError].} =
+  var buffer = default(seq[byte])
+
+  if readFieldInto(stream, buffer, header, pbytes):
+    value.data.buffer = buffer
```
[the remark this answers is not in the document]
gmelodie: Sound reasonable

libp2p/multiaddress.nim
```diff
@@ -1263,3 +1263,41 @@ func shortLog*(addrs: seq[MultiAddress], maxAddrs: int): string =
...
+    true
+  else:
+    false
```
gmelodie, line 1385:
```suggestion
    return true
  return false
```

.pinned
```diff
@@ -19,4 +19,4 @@ websock;https://github.com/status-im/nim-websock@#387a8eb7e961e8fdd3b1a717d36bc5
 zlib;https://github.com/status-im/nim-zlib@#e680f269fb01af2c34a2ba879ff281795a5258fe
 bearssl_pkey_decoder;https://github.com/vacp2p/bearssl_pkey_decoder@#d34aa46bf9d0a3ffff810fbd3c4d2fa024eb9368
 jwt;https://github.com/vacp2p/nim-jwt@#057ec95eb5af0eea9c49bfe9025b3312c95dc5f2
-protobuf_serialization;https://github.com/status-im/nim-protobuf-serialization@#46753f2b90365035bc0f75c6894e160c35880be1
+protobuf_serialization;https://github.com/status-im/nim-protobuf-serialization@#38d24eb3bd93e605fb88199da71d36b1ec0ad60d
```
gmelodie, line 22: We should make sure this doesn't clash with https://github.com/vacp2p/nim-libp2p/pull/2508. Maybe wait until that one is merged to merge this one

libp2p/multiaddress.nim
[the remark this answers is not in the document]
richard-ramos: yeah, it's a valid concern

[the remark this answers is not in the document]
moigagoo: I actually doubted up until the last moment which variant to keep... The variant with `init` kind of confused me because it creates a new instance and then assigns that to `value` whereas `PeerId` has a version of `init` that modifies an object in place.

libp2p/multiaddress.nim
moigagoo: @gmelodie I thought explicit `rerurn`s are discouraged except for early exits: https://status-im.github.io/nim-style-guide/language.result.html#practical-notes

richard-ramos (approved):

libp2p/multiaddress.nim
gmelodie: Then 
```nim
    return true
  false
```

gmelodie (approved):

libp2p/multiaddress.nim
nitely: > I thought explicit rerurns are discouraged except for early exits: https://status-im.github.io/nim-style-guide/language.result.html#practical-notes

ye, return should not be used here. See https://discord.com/channels/613988663034118151/1197844948477821019/1490959718036934787

nitely: see also https://github.com/status-im/nim-style-guide/pull/31/changes ; nvm actually using an early return here, like @gmelodie says is probably ok. In some cases the early return can save some nesting/indentation and make code more readable, here it does nothing, so probably it's just preference.

gmelodie: > so probably it's just preference

yup

nitely: but why do you prefer it? if you make it an early return and you remove the `return true`, the code will compile and return false instead of true. If you keep it the way it's now, and remove the `true`, the compiler will complain about the branch not returning true/false. Failing at compile time is better than some misbehavior at runtime. Why would you ever remove the `return true`?, by mistake in some refactor most likely.

libp2p/multiaddress.nim
```diff
@@ -1327,3 +1327,42 @@ func shortLog*(addrs: seq[MultiAddress], maxAddrs: int): string =
...
+    stream: InputStream,
+    value: var MultiAddress,
+    header: FieldHeader,
+    ProtoType: type ProtobufExt,
+): bool {.raises: [SerializationError, IOError].} =
+  var buffer = default(seq[byte])
+
+  if readFieldInto(stream, buffer, header, pbytes):
+    value = MultiAddress.init(buffer).valueOr:
+      raise newException(ProtobufValueError, "Invalid protobuf MultiAddress")
```
nitely, line 1365: On a closer look, the [serializer read for seq[MultiAddress]](https://github.com/vacp2p/nim-libp2p/blob/d353f79a1eeae81621d7eb43482e3d3f55695c11/libp2p/multiaddress.nim#L1179) has a different behaviour than just reading `MultiAddress` and putting into the seq. It ignores `MultiAddress` init errors... so this requires changing the protobuf extension system to support custom `seq[T]` :/

nitely: and [these tests](https://github.com/vacp2p/nim-libp2p/blob/069a05bde492d717fce26e7df8b5b0c4221664f0/tests/libp2p/multiformat/test_multiaddress.nim#L374-L449) need to be replicated for the protobuf extension using `Protobuf.encode/decode`

nitely: >so this requires changing the protobuf extension system to support custom seq[T] :/

implemented in https://github.com/status-im/nim-protobuf-serialization/pull/80

.pinned
moigagoo: @gmelodie since #2508 is merged, this is no longer an issue, is it?

libp2p/multiaddress.nim
moigagoo: If it's a matter of preference, I strongly prefer the variant without return:
1. It's symmetric and reads very naturally: ff this, then true, if that, then false. With two returns, it's still symmetric but overly explicit and with one return, it's asymmetric which makes you think about why there's a return in one place and no return in the other.
2. It's consistent with PeerId's code. If we have essentially the same logic in two places implemented just slightly differently, that's confusing and requires explanation.

libp2p/multiaddress.nim
moigagoo: @nitely I lost the plot, sorry.

If we need custom logic when reading to a `seq[MultiAddress]` that mimics the logic in `getRepeatedField`, shouldn't we just [add a `readFieldInto` for `seq[MultiAddress]`](https://github.com/vacp2p/nim-libp2p/pull/2477/changes/a04a23753a729778d3850249b7e80ef034b28ebf)?

So, the [changes in pb-ser](https://github.com/status-im/nim-protobuf-serialization/pull/80) aren't required for this?

> and [these tests](https://github.com/vacp2p/nim-libp2p/blob/069a05bde492d717fce26e7df8b5b0c4221664f0/tests/libp2p/multiformat/test_multiaddress.nim#L374-L449) need to be replicated for the protobuf extension using Protobuf.encode/decode

Here I'm seeing a problem. If I call `Protobuf.encode` or `decode` directly, they won't work because `MultiAddress` is not annotated with `proto2` as it's only serialized as part of DcutrMsg:

```nim
suite "MultiAddress test suite (protobuf_serialization)":
  test "getField returns ok(true) for valid multiaddress (protobuf_serialization)":
    let
      ma = MultiAddress.init("/ip4/1.2.3.4/tcp/80").get()
      maEncoded = Protobuf.encode(ma)
      maDecoded = Protobuf.decode(maEncoded, MultiAddress)

    check:
      maDecoded == ma
```
```shell
Error: fatal error: MultiAddress: missing {.proto2.} or {.proto3.}
```

If MultiAddress is supposed to be serializable outside of the Dcutr context, I guess, it has to be annotated, is this correct? But that would mean `VBuffer` needs to have extensions defined for it as well, which kind of brings us back to the initial PR state.

moigagoo: @nitely

I tried adding `readFieldInto` for `seq[MultiAddress]` replicating the original logic from `getRepeatedField` and added a test for pb-ser of `MultiAddress`. But to make this test pass, I had to extend the PR significantly, I'm not sure this is the right move...

I've returned the PR to draft stage for the time being.

moigagoo: @nitely moreover, since VBuffer has to use extensions anyway, MultiAddress may not use them at all. I've pushed the variant with only VBuffer extensions.

Please check out the PR.

nitely: > Here I'm seeing a problem. If I call Protobuf.encode or decode directly, they won't work because MultiAddress is not annotated with proto2 as it's only serialized as part of DcutrMsg:

you can create a type in the tests with a field of MultiAddress and add all the required proto pragmas to this test type.

> If MultiAddress is supposed to be serializable outside of the Dcutr context, I guess, it has to be annotated, is this correct?

No. MultiAddress is not a protobuf message, is a type extension.

> If we need custom logic when reading to a seq[MultiAddress] that mimics the logic in getRepeatedField, shouldn't we just [add a readFieldInto for seq[MultiAddress]](https://github.com/vacp2p/nim-libp2p/pull/2477/changes/a04a23753a729778d3850249b7e80ef034b28ebf)?

> So, the https://github.com/status-im/nim-protobuf-serialization/pull/80 aren't required for this?

it probably works without that PR because it overloads the default generic readFieldInto. But it could be if the readFieldInto is not imported then it fallsback to the default which would be wrong. The PR makes it so failing to import the custom readFileInto results in a compile time error.

There are some issues with https://github.com/vacp2p/nim-libp2p/pull/2477/changes/a04a23753a729778d3850249b7e80ef034b28ebf implementation: readFieldInto does not work like `let res = ?pb.getRepeatedField(field, items)`; it's called each time the field is found as you can see [here](https://github.com/status-im/nim-protobuf-serialization/pull/80/changes#diff-2cb4b9cfbd4418b239fab96f249d1f9b3be4f446a47aa517eb190b95441afcbdR47) notice it calls readFieldInto for a seq item and adds it to the seq, it never gets all values at once (there is no way to do that efficiently, minprotobuf is probably not efficient about it, I have not looked into it though). Which also means you cannot check if the sequence is empty inside `readFieldInto`, that check probably needs to be moved to the `decode` function; ex: after decoding, before returning the object it would check the seq is not empty.

tests/libp2p/protocols/test_dcutr.nim
```diff
@@ -26,7 +26,8 @@ suite "Dcutr":
     let connectMsg = DcutrMsg(msgType: MsgType.Connect, addrs: addrs)
 
     let pb = connectMsg.encode()
-    let connectMsgDecoded = DcutrMsg.decode(pb.buffer)
+    let connectMsgDecoded = DcutrMsg.decode(pb).valueOr:
+      raise newException(DcutrError, "Failed to decode a Connect message.")
 
     check connectMsg == connectMsgDecoded
```
nitely, line 35: this needs an extra check for `check pb == "hardcoded_encoded_value"` where the hardcoded value is taken from what master branch produces (before these PR changes). So we can make sure the encoded value is still the same. Otherwise you could change a field number for example and this test will still pass. That's I think the reason why the tests are currently passing, even though making MultiAddress a proto message is wrong and so they should be failing.

nitely: for the encoded value you can use `hexToSeqByte` from `stew/byteutils`: `check pb == hexToSeqByte("hardcoded_encoded_hex_value")` and you can use `echo toHex(pb)` to produce the hex string using the master branch.

moigagoo: @nitely thanks for the hint! I've added the check to connect and sync encode/decode tests.

libp2p/multiaddress.nim
moigagoo: @nitely thanks for clarifying this.

1. I've reverted the latest changes so that MultiAddress is a type extension again.
2. I've copied the logic of readFieldInto for seq from your PR so that the seq items are now read one by one.

> that check probably needs to be moved to the decode function; ex: after decoding, before returning the object it would check the seq is not empty.

That is something that should be done in pr-ser, right?

nitely: > I've copied the logic of readFieldInto for seq from your PR so that the seq items are now read one by one.

that's what the default readFieldInto for seq does. The point of the overload is so the MultiAddress that fail to init are skipped instead of raising the protobuf exception. Probably something like:

```
proc readFieldInto*(
    stream: InputStream,
    value: var seq[MultiAddress],
    header: FieldHeader,
    ProtoType: type ProtobufExt,
): bool {.raises: [SerializationError, IOError].} =
  var data = default(seq[byte])

  if readFieldInto(stream, data, header, pbytes):
    let ma = MultiAddress.init(data).valueOr:
      debug "Unsupported MultiAddress in blob", ma = data
      return false
    value.add(ma)
    true
  else:
    false
```

> That is something that should be done in pr-ser, right?

I don't think this is something that can be added to protobuf-ser. I think the place to add it is probably `dcutr/core.nim`, instead of `Protobuf.serializerFor([DcutrMsg])` you would define the encode/decode (exactly the same code serializerFor produces) and add the check in the decode proc.

nitely: > readFieldInto does not work like let res = ?pb.getRepeatedField(field, items); it's called each time the field is found as you can see [here](https://github.com/status-im/nim-protobuf-serialization/pull/80/changes#diff-2cb4b9cfbd4418b239fab96f249d1f9b3be4f446a47aa517eb190b95441afcbdR47) notice it calls readFieldInto for a seq item and adds it to the seq, it never gets all values at once (there is no way to do that efficiently, minprotobuf is probably not efficient about it, I have not looked into it though).

FWIW, minprotobuf seems to traverse the entire encoded data for each field, which has `fields_count*data_len` runtime.

moigagoo: @nitely 

> Probably something like:

It works but for some reason debug won't work: `rror: undeclared identifier: 'activeChroniclesStream'`.

Have you encountered this error?

> think the place to add it is probably dcutr/core.nim, instead of Protobuf.serializerFor([DcutrMsg])

I feel that this deviates from how things work now.

Currently, MultiAddress module is responsible for determining whether serialization is successful or not: getRepeatedField populates a seq of MultiAddress and if the resulting seq is empty returns an error. So DcutrMsg or any other MutliAddress consumer never gets an empty seq of MultiAddress.

If we move the logic to core.nim, it's the Dcutr module whoc decides if an empty seq is an error or not. Which is reasonable: readFieldInto never sees the entire sequence, it just determines if this particular item is added to the seq or not.

If there are any other MultiAddress consumers that rely on the existing MultiAddress check, the must implement decode as well.

I'm not saying this is a bad thing. But I think it is worth mentioning that non-empty addresses policy is now being enforced by the dcutr protocol and not the mutliaddress type. The tests will have to reflect that too.

nitely: > Have you encountered this error?

did it work before? with master?

> I feel that this deviates from how things work now.

true, it does. ^ @richard-ramos [minprotobuf seq[MultiAddress] decoding](https://github.com/vacp2p/nim-libp2p/blob/d353f79a1eeae81621d7eb43482e3d3f55695c11/libp2p/multiaddress.nim#L1179) errors out if all received values are invalid, but it does not error out if the message contains no MultiAddress. In both cases the resulting seq is empty. There is no way to make this distinction in protobuf-ser. You can only check if the seq is empty or not after decoding. Any idea what could be done here?

I don't see a way to support this in protobuf-ser, see https://github.com/vacp2p/nim-libp2p/pull/2477#discussion_r3288295801

moigagoo: @nitely 

> did it work before? with master?

Well, this proc doesn't exist in master 😅 But I'm seeing `debug` in `getRepeatedField` and it compiles just fine. It must be because of the context where `readFieldInto` gets executed.

nitely: I meant it worked for `getRepeatedField ` but if it's a compile error then surely it worked. Try `export chronicles` in libp2p/multiaddress.nim, and if that works then likely the problem is readFieldInto is a generic. If that does not work, then idk.

moigagoo: Yeah I tried both `export`ing chronicles from multiaddress.nim and and `import`int it in test_multiaddress.nim.

The result is surprising:
```shell
$ nim r tests/libp2p/multiformat/test_multiaddress.nim "Protobuf*::"
Hint: used config file '/home/moigagoo/.nimble/pkgs2/nim-2.2.10-17ec440fdb89f8903db29a17898af590087d2b64/config/nim.cfg' [Conf]
Hint: used config file '/home/moigagoo/.nimble/pkgs2/nim-2.2.10-17ec440fdb89f8903db29a17898af590087d2b64/config/config.nims' [Conf]
Hint: used config file '/home/moigagoo/projects/nim-libp2p/config.nims' [Conf]
Hint: used config file '/home/moigagoo/projects/nim-libp2p/tests/config.nims' [Conf]
................................................................................................................................................................................................................................................................................................................................
/home/moigagoo/projects/nim-libp2p/libp2p/peerstore.nim(337, 19) Warning: owner is deprecated [Deprecated]
....
CC: system.nim
CC: sets.nim
CC: protobuf_serialization/codec.nim
CC: protobuf_serialization/reader.nim
CC: results.nim
CC: libp2p/multiaddress.nim
CC: test_multiaddress.nim
/home/moigagoo/.cache/nim/test_multiaddress_d/@mtest_multiaddress.nim.c: In function ‘runTestX60gensym1177___test95multiaddress_u15613’:
/home/moigagoo/.cache/nim/test_multiaddress_d/@mtest_multiaddress.nim.c:22214:126: error: incompatible type for argument 3 of ‘getField__OOZOOZOOZlibp2pZmultiaddress_u5610’
22214 |                         nimln_(440);                    T4_ = getField__OOZOOZOOZlibp2pZmultiaddress_u5610((&pb_1), ((NI)1), decoded_1);
      |                                                                                                                              ^~~~~~~~~
      |                                                                                                                              |
      |                                                                                                                              tyObject_MultiAddress__b8Mi1sCvnYnJUOL9c9ci6LNQ
/home/moigagoo/.cache/nim/test_multiaddress_d/@mtest_multiaddress.nim.c:20434:219: note: expected ‘tyObject_MultiAddress__b8Mi1sCvnYnJUOL9c9ci6LNQ *’ but argument is of type ‘tyObject_MultiAddress__b8Mi1sCvnYnJUOL9c9ci6LNQ’
20434 | multiaddress_u5610)(tyObject_ProtoBuffer__HMI60XWymeNggabotsliEw* pb_p0, NI field_p1, tyObject_MultiAddress__b8Mi1sCvnYnJUOL9c9ci6LNQ* value_p2) {
      |                                                                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~

/home/moigagoo/.cache/nim/test_multiaddress_d/@mtest_multiaddress.nim.c: In function ‘runTestX60gensym1225___test95multiaddress_u16035’:
/home/moigagoo/.cache/nim/test_multiaddress_d/@mtest_multiaddress.nim.c:23047:126: error: incompatible type for argument 3 of ‘getField__OOZOOZOOZlibp2pZmultiaddress_u5610’
23047 |                         nimln_(451);                    T3_ = getField__OOZOOZOOZlibp2pZmultiaddress_u5610((&pb_1), ((NI)1), decoded_1);
      |                                                                                                                              ^~~~~~~~~
      |                                                                                                                              |
      |                                                                                                                              tyObject_MultiAddress__b8Mi1sCvnYnJUOL9c9ci6LNQ
/home/moigagoo/.cache/nim/test_multiaddress_d/@mtest_multiaddress.nim.c:20434:219: note: expected ‘tyObject_MultiAddress__b8Mi1sCvnYnJUOL9c9ci6LNQ *’ but argument is of type ‘tyObject_MultiAddress__b8Mi1sCvnYnJUOL9c9ci6LNQ’
20434 | multiaddress_u5610)(tyObject_ProtoBuffer__HMI60XWymeNggabotsliEw* pb_p0, NI field_p1, tyObject_MultiAddress__b8Mi1sCvnYnJUOL9c9ci6LNQ* value_p2) {
      |                                                                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~

Error: execution of an external compiler program 'gcc -c  -w -fmax-errors=3 -fno-strict-aliasing -pthread -I/home/moigagoo/projects/nim-libp2p/nimbledeps/pkgs2/bearssl-0.2.8-da4dd7ae96d536bdaf42dca9c85d7aed024b6a86/bearssl/abi -I/home/moigagoo/projects/nim-libp2p/nimbledeps/pkgs2/bearssl-0.2.8-da4dd7ae96d536bdaf42dca9c85d7aed024b6a86/bearssl/abi/../csources/src/ -I/home/moigagoo/projects/nim-libp2p/nimbledeps/pkgs2/bearssl-0.2.8-da4dd7ae96d536bdaf42dca9c85d7aed024b6a86/bearssl/abi/../csources/inc/ -I/home/moigagoo/projects/nim-libp2p/nimbledeps/pkgs2/bearssl-0.2.8-da4dd7ae96d536bdaf42dca9c85d7aed024b6a86/bearssl/abi/../csources/tools/ -DBR_USE_UNIX_TIME=1 -DBR_USE_URANDOM=1 -DBR_LE_UNALIGNED=1 -DBR_64=1  -DBR_amd64=1 -DBR_INT128=1 -fno-omit-frame-pointer   -I/home/moigagoo/.nimble/pkgs2/nim-2.2.10-17ec440fdb89f8903db29a17898af590087d2b64/lib -I/home/moigagoo/projects/nim-libp2p/tests/libp2p/multiformat -o /home/moigagoo/.cache/nim/test_multiaddress_d/@mtest_multiaddress.nim.c.o /home/moigagoo/.cache/nim/test_multiaddress_d/@mtest_multiaddress.nim.c' failed with exit code: 1
```

nitely: > I don't see a way to support this in protobuf-ser, it seems only possible in minprotobuf because it does not follow the proto spec, see https://github.com/vacp2p/nim-libp2p/pull/2477#discussion_r3288295801

A possible workaround is to add `default(MultiAddress)` (empty which is invalid) on init error to the seq, then do a post validation in `decode` + filter-out the empty values... but then that needs to be done in every protocol containing seq[MultiAddress] fields... :/

richard-ramos: > @nitely: minprotobuf seq[MultiAddress] decoding errors out if all received values are invalid, but it does not error out if the message contains no MultiAddress. In both cases the resulting seq is empty. There is no way to make this distinction in protobuf-ser. You can only check if the seq is empty or not after decoding. Any idea what could be done here?

I think we do not need to keep this exact old minprotobuf behavior. For DCUtR, the important thing is is whether we have usable addresses or not. So I think the protobuf MultiAddress extension can skip invalid addresses.

so maybe the custom reader for seq[MultiAddress] could do roughly this?
```
let m = MultiAddress.init(data)
if m.isErr():
  continue  # we just ignore it...
else:
   add it to the seq
```

Then DCUtR code can check the result:
- for Connect message, if there are no addresses, it will abort and not do DCUtR.
- in Sync message, empty addresses are OK, because Sync is sent with `@[]` anyway. ( `DcutrMsg.decode `should not fail only because addrs is empty, that would make Sync break)

moigagoo: @richard-ramos thanks, that sounds a lot more reasonable than the existing len check: https://github.com/vacp2p/nim-libp2p/pull/2477/commits/ce7e446783ce276e25a165489e501bba293b845e

moigagoo:
I've removed `debug` from readFieldInto because adding it breaks compilation.

libp2p/protocols/connectivity/dcutr/core.nim
```diff
@@ -7,51 +7,54 @@ import std/sequtils
...
+    case decoded.msgType
+    of MsgType.Connect:
+      if len(decoded.addrs) > 0:
+        Opt.some(decoded)
+      else:
+        Opt.none(DcutrMsg)
+    of MsgType.Sync:
+      Opt.some(decoded)
+  except SerializationError:
+    Opt.none(DcutrMsg)
```
nitely: [this is already checked in the app logic](https://github.com/vacp2p/nim-libp2p/blob/069a05bde492d717fce26e7df8b5b0c4221664f0/libp2p/protocols/connectivity/dcutr/server.nim#L57-L61), so it's not needed. We can go back to using `serializerFor`. The minprotobuf decode allowed the seq[MultiAddress] to be empty so I figured the check was already been done in app logic.

nitely (approved):

PR #2480, opened 2026-05-18
test(service-disco): validateRegisterMessage + message size

rlve:
## Summary

Adds new tests:

- Registrar rejects advertisements whose service set doesn't include the queried service, or has no services at all, and accepts multi-service advertisements where one matches.
- Advertiser builds records without enforcing a size limit on per-service data or on the encoded XPR.
- Component-level test confirming that an oversized message on the wire is dropped with no reply.

## Affected Areas

- [x] Tests

## Impact on Library Users

None.

## Risk Assessment

None.


tests/libp2p/service_discovery/test_advertiser.nim
```diff
@@ -146,3 +152,40 @@ suite "Advertiser - removeProvidedService":
...
+
+    let record = disco.record()
+    check record.isOk()
+
+    let encoded = record.get().encode()
+
+    check:
+      encoded.isOk()
+      encoded.get().len > MaxMsgSize
+      record.get().data.addresses.len == addrs.len
```
rlve, line 191: I added these two tests to show that there is no validation on the record size and it's possible to create messages bigger than `MaxMsgSize`. The [spec](https://github.com/logos-co/logos-lips/blob/f7bb80a4b05ebf8039c3795d9cb9fe67b9a6228b/docs/anoncomms/raw/logos-service-discovery.md?plain=1#L506) recommends the following constraints:

```
**Size constraints:**

- Each `ServiceInfo.data` field SHOULD be ≤ 33 bytes
- Total encoded XPR SHOULD be ≤ 1024 bytes
```

Up to you whether this should be addressed and an issue created.

gmelodie (approved):

vladopajic (approved):

tests/libp2p/service_discovery/test_advertiser.nim
SionoiS: Ah good catch! Yes it should be enforced.

SionoiS (approved):

PR #2482, opened 2026-05-18
chore: better names for streams and connections

richard-ramos:
## Summary

This PR is a semantic naming cleanup. It does not change runtime behavior.

It introduces clearer aliases for existing connection endpoints:

- `Stream` for protocol-facing stream endpoints
- `RawConn` for transport and upgrade boundaries before security/muxing
- `MuxedStream` for streams opened or handled by muxers

These names are aliases over the existing implementation types, not a semantic type split. The goal is to make APIs read closer to what each value represents today, while making a future stronger refactor easier to stage.

The PR also updates call sites, bindings, examples, tutorials, and tests to use stream-oriented names consistently. Public pubsub send/custom stream names now use `Stream` terminology, and the upgrade path names the secure stage with `SecureConn` before muxing.

## Affected Areas

- [x] Gossipsub
  Pubsub peer send/custom stream APIs now use `sendStream`, `GetStream`, `CustomStreamCallbacks`, and `useCustomStream` naming.
- [x] Transports
  Transport accept/dial and upgrade boundaries use `RawConn`; relay keeps `RelayConnection` as the transport-facing wrapper over a relay protocol stream.
  Connection manager and muxer-store APIs keep established peer sessions as `Muxer`, while Kademlia and discovery protocol stream call sites use stream naming.
- [x] Protocol Logic
  Protocol handlers, multistream negotiation, switch dialing, secure upgrade, and muxer stream handlers use `Stream`, `SecureConn`, or `MuxedStream` according to stack position.
- [x] Other
  cbind APIs, C examples, Nim examples, and tutorials now use stream-oriented names such as `StreamCallback` and `libp2p_stream_t *stream`.

## Impact on Library Users

There is no runtime behavior change.

For Nim users, `Connection` remains the underlying implementation type, so this does not introduce new runtime representations. Some public API names were changed to reflect stream semantics, such as pubsub send/custom stream callbacks and stream handler parameters.

For cbind users, stream callbacks and stream handles now use `StreamCallback` / `stream` naming instead of the older connection-oriented wording.

## Risk Assessment

Risk is low because the core change is semantic aliasing and call-site renaming. The main risk is stale naming in API docs, examples, bindings, or tests. This PR updates those alongside the library code.

No network behavior, protocol behavior, security behavior, or performance changes are intended.

### DIFF
--- a/cbind/examples/cbindings.c
+++ b/cbind/examples/cbindings.c
@@ -54,8 +54,8 @@ static void peerstore_entry_handler(int callerRet,
                                     const char *msg, size_t len,
                                     void *userData);
 
-static void connection_handler(int callerRet, libp2p_stream_t *conn,
-                               const char *msg, size_t len, void *userData);
+static void stream_handler(int callerRet, libp2p_stream_t *stream,
+                           const char *msg, size_t len, void *userData);
 static void create_cid_handler(int callerRet, const char *msg, size_t len,
                                void *userData);
 static void read_handler(int callerRet, const uint8_t *data, size_t dataLen,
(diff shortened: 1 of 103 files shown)


gmelodie (approved):

vladopajic (approved):

vladopajic:
:100:

PR #2483, opened 2026-05-18
feat(nat): announcedAddrs

gmelodie:
### Summary

Adds a public **"announced addresses"** API so a node can bind locally while advertising a different, externally-reachable address — the typical scenario being a host behind a NAT with a known static port-forward, or any setup where the operator wants to override what the mapper chain would otherwise produce.

The change has two halves:

1. **New surface area** — `SwitchBuilder.withAnnouncedAddresses` / `withAnnouncedAddress` and a matching `announcedAddrs` field on `PeerInfo`. When non-empty, `PeerInfo.expandAddrs` returns this set verbatim (still subject to `addressPolicy` filtering) and skips the listen-addrs + `addressMappers` pipeline entirely.
2. **Builder DSL cleanup motivated by the new API** — the previously-public `autonatService*`, `autonatV2Service*`, `hpService*` fields on `SwitchBuilder` are made private, and dedicated setters (`withAutonatService`, `withAutonatV2Service`) are added so all configuration goes through the fluent builder. `withAutotls` / `withCircuitRelay` / `withRendezVous` now treat a `nil` argument as "clear", which lets call sites drop their `if x != nil:` guards.

Two consumers of "what addresses do we announce?" — DCUtR (`dcutr/server.nim`) and hole-punching (`services/hpservice.nim`) — now prefer `switch.peerInfo.addrs` (the expanded announce set, which respects the new API and existing UPnP/autonat mappers) over the older per-listen-addr `guessDialableAddr` fallback whenever an expanded set is available.

### Affected Areas

- `libp2p/builders.nim` — new `announcedAddrs` field, `withAnnouncedAddresses` / `withAnnouncedAddress` / `withAutonatService` / `withAutonatV2Service` setters, visibility tightening on three service fields, nil-tolerant `withAutotls` / `withCircuitRelay` / `withRendezVous`, `announcedAddrs` threaded into `PeerInfo.new` in `buildSwitch`.
- `libp2p/peerinfo.nim` — new `announcedAddrs` field, `expandAddrs` short-circuit when it's non-empty, new `announcedAddrs` kwarg on `PeerInfo.new`.
- `libp2p/protocols/connectivity/dcutr/server.nim` — DCUtR uses `peerInfo.addrs` over `listenAddrs.mapIt(guessDialableAddr)` when available.
- `libp2p/services/hpservice.nim` — same fallback change for hole-punching's connected-peer handler.
- `interop/hole-punching/hole_punching.nim` — refactored to thread `relayClient` through `withCircuitRelay` now that it tolerates nil.
- `libp2p/autotls/service.nim`, `libp2p/protocols/rendezvous/rendezvous.nim` — drive-by `# todo: should not be ref object` comments on `AutotlsConfig` / `RendezVousConfig` only.
- Tests:
  - `tests/libp2p/test_peer_info.nim` — new cases: "Announced addresses override mapper chain", "addressPolicy still filters announced addresses".
  - `tests/libp2p/test_switch.nim` — new `asyncTest` "withAnnouncedAddresses overrides what peers see" verifying the wildcard resolver is bypassed.
  - `test_autonat.nim`, `test_autonat_service.nim`, `test_autonat_v2_service.nim`, `test_relay_v2.nim`, `test_hp.nim` — refactored to use the new builder setters / nil-tolerant overloads instead of poking private fields.

### Impact on Library Users

- **New API**: `SwitchBuilder.withAnnouncedAddresses(seq[MultiAddress])`, `withAnnouncedAddress(MultiAddress)`, `withAutonatService(Opt[AutonatService])`, `withAutonatV2Service(Opt[AutonatV2Service])`. New optional kwarg `announcedAddrs` on `PeerInfo.new`. New public field `PeerInfo.announcedAddrs`.
- **Breaking for downstream callers who poked builder fields directly**: `b.autonatService = ...`, `b.autonatV2Service = ...`, `b.hpService = ...` no longer compile. Migration is the corresponding `with*` setter (`withAutonatService`, `withAutonatV2Service`, `withHolePunching`); the in-repo test refactors are worked examples.
- **Behavior change when `announcedAddrs` is used**: the entire `addressMappers` chain (including the wildcard resolver, UPnP, autonat) is skipped — callers opting in must supply already-resolved addresses. `addressPolicy` filtering still applies.
- **Behavior change in DCUtR / hole-punching fallback**: when zero addrs have been observed yet, both protocols now use `peerInfo.addrs` (if non-empty) instead of `listenAddrs.mapIt(guessDialableAddr(...))`. For callers without `withAnnouncedAddresses` and no other address mappers, `peerInfo.addrs` will still be empty here, so behavior is unchanged; for callers with mappers or announce set, they get the better address.
- **Nil-tolerant overloads**: `withAutotls(nil)`, `withCircuitRelay(nil)`, `withRendezVous(nil)` now clear the option instead of storing `Opt.some(nil)`.

### Risk Assessment

- **Backward compatibility**: the visibility change on `autonatService` / `autonatV2Service` / `hpService` is a source-level break for anyone setting them directly. No on-the-wire change.
- **Network behavior**: a user calling `withAnnouncedAddresses` who hasn't actually arranged for those addresses to be reachable will advertise unreachable addresses to peers — same hazard as today's manual mapper hacks, but more discoverable. The mapper-chain bypass is documented inline in `builders.nim` and `peerinfo.nim`.
- **DCUtR / hole-punching**: the fallback shift is gated on `peerInfo.addrs.len > 0`, so callers with no address mappers and no announced addrs see no change. Callers with a working mapper chain were previously falling back to a per-listen-addr guess that was usually wrong; this is a strict improvement for them.
- **Performance**: none expected — `expandAddrs` skips work in the announce-override path; no new allocations elsewhere.
- **Security**: none — `addressPolicy` still filters announced addresses (e.g. `withPrivateAddressFilter` still drops RFC1918 entries from the announce set), and a covering test pins this behavior.

### References

https://github.com/status-im/nimbus-eth2/blob/6fb05f36804d53c2e8e014cfeeea8ad7996a5efe/beacon_chain/networking/eth2_network.nim#L2484-L2486


richard-ramos (approved):

tests/libp2p/test_switch.nim
```diff
@@ -1193,6 +1193,36 @@ suite "Switch":
...
+        .build()
+
+    await switch.start()
+    defer:
+      await switch.stop()
+
+    # Transport actually bound a local socket — listenAddrs is populated, and
+    # the *original* zero-port address has been resolved to something else.
+    check switch.peerInfo.listenAddrs.len == 1
+    check switch.peerInfo.listenAddrs[0] != listenAddr
```
vladopajic, line 1221: i don't thing this should be asserted here, it is not related to test and it drives attention from test

vladopajic (approved):

PR #2488, opened 2026-05-19
test(service-disco): non-32-byte keys + misc component

rlve:
## Summary

Adds new tests:

- advertiser REGISTER refresh after advertExpiry
- K_lookup limiting when multiple registrars are in the same bucket
- REGISTER with non-32-byte keys returning Rejected
- GET_ADS with non-32-byte keys returning an empty response

## Affected Areas

- [x] Tests

## Impact on Library Users

None.

## Risk Assessment

None.

### DIFF
--- a/tests/libp2p/service_discovery/component/test_error.nim
+++ b/tests/libp2p/service_discovery/component/test_error.nim
@@ -164,3 +164,57 @@ suite "Service Discovery Component - Error Handling":
     expect LPStreamError:
       discard
         await clientSwitch.sendRawMessage(registrarNode, oversizedMsg).wait(2.seconds)
+
+  asyncTest "REGISTER with non-32-byte key returns Rejected":
+    # Spec calls for rejection on bad key length.
+    # Impl has no length check, the key reaches the service-membership check.
+    let registrarNode = setupServiceDiscoveryNode()
+    let clientNode = setupServiceDiscoveryNode()
+    startAndDeferStop(@[registrarNode, clientNode])
+    await connect(registrarNode, clientNode)
+
+    let adBytes =
+      makeAdvertisement("service", clientNode.switch.peerInfo.privateKey).encode().get()
+
+    for keyLen in [0, 31, 33, 64]:
+      var key = newSeq[byte](keyLen)
+      for i in 0 ..< keyLen:
+        key[i] = i.byte
+
+      let msg = kad_protobuf.Message(
+        msgType: kad_protobuf.MessageType.register,
+        key: key,
+        register: Opt.some(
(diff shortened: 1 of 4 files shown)


gmelodie (approved):

tests/libp2p/service_discovery/component/test_lookup_get_ads.nim
```diff
@@ -19,22 +19,50 @@ import ../../../../libp2p/protocols/kademlia/protobuf as kad_protobuf
...
   var a = setupServiceDiscoveryNode(discoConfig = conf)
   var b = setupServiceDiscoveryNode(discoConfig = conf)
-  while bucketOf(a) == bucketOf(b):
+  for _ in 0 ..< MaxRegistrarSetupAttempts:
+    if bucketOf(a, serviceId) != bucketOf(b, serviceId):
+      if bucketOf(a, serviceId) < bucketOf(b, serviceId):
+        return (a, b)
+      else:
+        return (b, a)
+
```
SionoiS, line 40: We could pre-compute a set of service id that fall into different buckets and use a lookup table instead.

Would that be useful?

SionoiS (approved):
LGTM Thank you!

tests/libp2p/service_discovery/component/test_error.nim
```diff
@@ -164,3 +164,57 @@ suite "Service Discovery Component - Error Handling":
...
+    # Impl has no length check, cache lookup misses on the arbitrary key.
+    let registrarNode = setupServiceDiscoveryNode()
+    let clientNode = setupServiceDiscoveryNode()
+    startAndDeferStop(@[registrarNode, clientNode])
+    await connect(registrarNode, clientNode)
+
+    for keyLen in [0, 31, 33, 64]:
+      var key = newSeq[byte](keyLen)
+      for i in 0 ..< keyLen:
+        key[i] = i.byte
```
vladopajic: this could some utility; what we care here is just to create initialized seq[byte] of some size. which is what we do when creating messages as well.

vladopajic (approved):

tests/libp2p/service_discovery/component/test_lookup_get_ads.nim
rlve: Will reconsider if it turns out to be flaky.

tests/libp2p/service_discovery/component/test_error.nim
rlve: This wasn't really needed, simplified with seq of 0s.

vladopajic: that makes sense as well

PR #2493, opened 2026-05-19
feat(nat): NATService skeleton

gmelodie:
## Summary

Adds piece 2 of the NAT traversal refinement: a first-class `NATService` skeleton in nim-libp2p. Consumers can now configure a static external IP through a single opt-in builder call instead of implementing NAT setup themselves.

Changes:
- Added `NATConfig` and `NATService` (`libp2p/services/natservice.nim`)
- Added `.withNAT(config)` builder (`libp2p/builders.nim`)
- `natModeExplicitIp` registers an `AddressMapper` that rewrites bound TCP/QUIC listen addresses with the configured external IP and populates `peerInfo.announcedAddrs`
- `natModeAuto` is currently a documented no-op until future AutoNAT / hole-punching integration

Builds on piece 1 (`withAnnouncedAddresses`). Future work:
- Piece 3: UPnP / NAT-PMP integration
- Piece 3a: AutoNAT + hole punching integration
- Pieces 4–5: Waku and Nimbus adoption

## Affected Areas

- [x] Other — new opt-in service (`NATService`) and builder (`withNAT`)
- [ ] Transports
- [ ] Peer management / discovery
- [ ] Protocol logic
- [ ] Build / tooling

## Compatibility

Opt-in only. Existing behavior is unchanged unless `.withNAT(...)` is used.

- New public APIs: `NATMode`, `NATConfig`, `NATService`, `withNAT`
- No removals or behavioral changes
- `standard_switch` does not enable NAT support by default
- `natModeExplicitIp` automatically populates `peerInfo.announcedAddrs`
- `natModeAuto` remains inactive until piece 3a

## Risk Assessment

- **Compatibility:** No breaking changes in this PR
- **Network behavior:** Unchanged unless explicitly enabled
- **Performance:** Explicit-IP mapping runs once after transport binding
- **Security:** No probing or external I/O; IP is caller-supplied
- **Known limitation:** Announced addresses are frozen after initial mapping and do not track runtime port changes

## Tests

Added `tests/libp2p/services/test_natservice.nim` covering:

1. Explicit IP populates announced addresses
2. `natModeAuto` is a no-op
3. `stop()` clears announced addresses
4. `.withNAT(...)` end-to-end builder integration

All tests pass locally.

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -40,7 +40,7 @@ import services/wildcardresolverservice
 export
   switch, peerid, peerinfo, peeraddrpolicy, connection, multiaddress, crypto, errors,
   TLSPrivateKey, TLSCertificate, TLSFlags, ServerFlags, connmanager.ConnectionLimits,
-  connmanager.maxTotal, connmanager.maxInOut
+  connmanager.maxTotal, connmanager.maxInOut, natservice
 
 const MemoryAutoAddress* = memorytransport.MemoryAutoAddress
 
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -40,7 +40,7 @@ import services/wildcardresolverservice
 export
   switch, peerid, peerinfo, peeraddrpolicy, connection, multiaddress, crypto, errors,
   TLSPrivateKey, TLSCertificate, TLSFlags, ServerFlags, connmanager.ConnectionLimits,
-  connmanager.maxTotal, connmanager.maxInOut, natservice
+  connmanager.maxTotal, connmanager.maxInOut, natservice.NATConfig
 
 const MemoryAutoAddress* = memorytransport.MemoryAutoAddress
 
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -27,7 +27,7 @@ import
     relay/client,
     relay/rtransport,
   ],
-  services/[autorelayservice, hpservice, identify_pusher],
+  services/[autorelayservice, hpservice, identify_pusher, natservice],
   connmanager,
   upgrademngrs/muxedupgrade,
   observedaddrmanager,
@@ -84,6 +84,7 @@ type
     autonatV2Client: AutonatV2Client
     autonatV2Service: Opt[AutonatV2Service]
     hpService: Opt[HPService]
+    natConfig: Opt[NATConfig]
     autotlsConfig: Opt[AutotlsConfig]
     circuitRelay: Opt[Relay]
     rdvConfig: Opt[RendezVousConfig]
@@ -341,6 +342,12 @@ proc withAutonatV2*(
   )
   b
 
+proc withNAT*(b: SwitchBuilder, config: NATConfig): SwitchBuilder =
+  ## Enable a NAT traversal service.
+  ## TODO: wire in autonat / hole-punching.
+  b.natConfig = Opt.some(config)
+  b
+
 proc withHolePunching*(
     b: SwitchBuilder, maxNumRelays: int, onReservationHandler: proc
 ): SwitchBuilder =
@@ -519,6 +526,9 @@ proc setupServices(b: SwitchBuilder, switch: Switch) {.raises: [LPError].} =
 
   b.hpService.withValue(hpservice):
     switch.services.add(hpservice)
+
+  b.natConfig.withValue(natCfg):
+    switch.services.add(NATService.new(natCfg))
 
   if b.identifyPusherEnabled:
     switch.services.add(IdentifyPusher.new())
--- a/libp2p/multiaddress.nim
+++ b/libp2p/multiaddress.nim
@@ -875,6 +875,18 @@ proc getPart*(ma: MultiAddress, codec: MultiCodec): MaResult[MultiAddress] =
       return ok(part)
   err("no such codec in multiaddress")
 
+proc getProtocolArgument*(ma: MultiAddress, codec: MultiCodec): MaResult[seq[byte]] =
+  ## Returns the raw byte argument of the first component in ``ma`` matching
+  ## ``codec`` (e.g. the 2-byte port for ``/tcp/<port>``, the 4-byte address
+  ## bytes for ``/ip4/<addr>``).
+  for item in ma:
+    let
+      ritem = ?item
+      code = ?ritem.protoCode()
+    if code == codec:
+      return ok(?ritem.protoAddress())
+  err("Multiaddress codec has not been found")
+
 proc getProtocol(name: string): MAProtocol {.inline.} =
   let mc = MultiCodec.codec(name)
   if mc != InvalidMultiCodec:
--- a/libp2p/multiaddress.nim
+++ b/libp2p/multiaddress.nim
@@ -1271,16 +1271,12 @@ proc replaceIp*(ma: MultiAddress, ip: IpAddress): MaResult[MultiAddress] =
   let
     ip4 = multiCodec("ip4")
     ip6 = multiCodec("ip6")
-    targetCodec =
-      case ip.family
-      of IpAddressFamily.IPv4: ip4
-      of IpAddressFamily.IPv6: ip6
     newIp =
       case ip.family
       of IpAddressFamily.IPv4:
-        ?MultiAddress.init(targetCodec, ip.address_v4)
+        ?MultiAddress.init(ip4, ip.address_v4)
       of IpAddressFamily.IPv6:
-        ?MultiAddress.init(targetCodec, ip.address_v6)
+        ?MultiAddress.init(ip6, ip.address_v6)
 
   var
     found = false
--- a/libp2p/multiaddress.nim
+++ b/libp2p/multiaddress.nim
@@ -1263,10 +1263,11 @@ proc getIp*(ma: MultiAddress): Opt[IpAddress] =
 
 proc replaceIp*(ma: MultiAddress, ip: IpAddress): MaResult[MultiAddress] =
   ## Returns a copy of ``ma`` with its leading IP4/IP6 component replaced by
-  ## ``ip``. The remainder of the multiaddress (transport, port, and any
-  ## suffix such as ``/quic-v1``, ``/ws``, ``/wss``, ``/tls/ws``) is
-  ## preserved. Returns an error if ``ma`` has no IP component or its IP
-  ## family does not match ``ip``.
+  ## ``ip``. If ``ip``'s family differs from the original, the IP codec is
+  ## swapped accordingly (``/ip6/...`` becomes ``/ip4/...`` or vice versa).
+  ## The remainder of the multiaddress (transport, port, and any suffix such
+  ## as ``/quic-v1``, ``/ws``, ``/wss``, ``/tls/ws``) is preserved. Returns
+  ## an error if ``ma`` has no IP component.
   let
     ip4 = multiCodec("ip4")
     ip6 = multiCodec("ip6")
@@ -1288,8 +1289,6 @@ proc replaceIp*(ma: MultiAddress, ip: IpAddress): MaResult[MultiAddress] =
     let part = ?item
     let code = ?part.protoCode
     if not found and (code == ip4 or code == ip6):
-      if code != targetCodec:
-        return err("multiaddress: IP family mismatch")
       ?res.append(newIp)
       found = true
     else:
--- a/libp2p/multiaddress.nim
+++ b/libp2p/multiaddress.nim
@@ -1249,6 +1249,43 @@ proc getIp*(ma: MultiAddress): Opt[IpAddress] =
       cursor.offset = cursor.offset + skipLen
     # Marker - nothing to skip
 
+proc replaceIp*(ma: MultiAddress, ip: IpAddress): MaResult[MultiAddress] =
+  ## Returns a copy of ``ma`` with its leading IP4/IP6 component replaced by
+  ## ``ip``. The remainder of the multiaddress (transport, port, and any
+  ## suffix such as ``/quic-v1``, ``/ws``, ``/wss``, ``/tls/ws``) is
+  ## preserved. Returns an error if ``ma`` has no IP component or its IP
+  ## family does not match ``ip``.
+  let
+    ip4 = multiCodec("ip4")
+    ip6 = multiCodec("ip6")
+    targetCodec =
+      case ip.family
+      of IpAddressFamily.IPv4: ip4
+      of IpAddressFamily.IPv6: ip6
+    newIp =
+      case ip.family
+      of IpAddressFamily.IPv4:
+        ?MultiAddress.init(targetCodec, ip.address_v4)
+      of IpAddressFamily.IPv6:
+        ?MultiAddress.init(targetCodec, ip.address_v6)
+
+  var
+    found = false
+    res = MultiAddress.init()
+  for item in ma.items:
+    let part = ?item
+    let code = ?part.protoCode
+    if not found and (code == ip4 or code == ip6):
+      if code != targetCodec:
+        return err("multiaddress: IP family mismatch")
+      ?res.append(newIp)
+      found = true
+    else:
+      ?res.append(part)
+  if not found:
+    return err("multiaddress: no IP component to replace")
+  ok(res)
+
 const AvgMultiAddressStringLength = 32
 
 func shortLog*(addrs: seq[MultiAddress], maxAddrs: int): string =
--- a/libp2p/services/natservice.nim
+++ b/libp2p/services/natservice.nim
@@ -116,6 +116,9 @@ method stop*(self: NATService, switch: Switch) {.async: (raises: [CancelledError
   trace "Stopping NATService"
   if self.addressMapper != nil:
     switch.peerInfo.addressMappers.keepItIf(it != self.addressMapper)
-  # Clear the explicit announce so the switch reverts to mapper-chain output.
+  # Clear the explicit announce owned by this service, but do not trigger
+  # peerInfo.update() during shutdown because it may notify observers and
+  # cause them to publish address changes while the switch is tearing down.
+  # Address recalculation is deferred until the next Switch.start or an
+  # explicit refresh performed at a safe lifecycle point.
   switch.peerInfo.announcedAddrs = @[]
-  await switch.peerInfo.update()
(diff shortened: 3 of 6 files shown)


libp2p/builders.nim
```diff
@@ -341,6 +342,12 @@ proc withAutonatV2*(
   )
   b
 
+proc withNAT*(b: SwitchBuilder, config: NATConfig): SwitchBuilder =
```
vladopajic, line 345: `withXxx` procs should only collect configuration. creating services and protocols should happen based on those configurations in `build()` proc.

libp2p/services/natservice.nim
```diff
@@ -0,0 +1,124 @@
...
+    listenAddr.getProtocolArgument(netCodec).withValue(portArg):
+      let port = Port(uint16.fromBytesBE(portArg))
+      let base = MultiAddress.init(explicitIp, transportProto, port)
+      if isQuic:
+        let quicV1 = MultiAddress.init("/quic-v1").valueOr:
+          continue
+        base.concat(quicV1).withValue(maddress):
+          addrs.add(maddress)
+      else:
+        addrs.add(base)
```
vladopajic: this could have been simplified 

```nim
 for listenAddr in listenAddrs:
  
  let tcpAddr = makeExplicitTcpAddress(...)
  tcpAddr.withValue(a):
    addrs.add(a)
    continue

  let quicAddr = makeExplicitQuicAddress(...)
  quicAddr.withValue(a):
    addrs.add(a)
    continue
```

would be easier to follow the code as well

libp2p/services/natservice.nim
```diff
@@ -0,0 +1,124 @@
...
+
+  case self.config.mode
+  of natModeAuto:
+    # TODO: wire autonat / hole-punching in here.
+    discard
+  of natModeExplicitIp:
+    let explicitIp = self.config.explicitIp.valueOr:
+      raise newException(
+        ServiceSetupError, "NATService natModeExplicitIp requires explicitIp to be set"
+      )
```
vladopajic: it should not be possible to create invalid config, or config to go all the way to here and to get validated here

tests/libp2p/services/test_natservice.nim
```diff
@@ -0,0 +1,97 @@
...
+): Switch {.raises: [LPError].} =
+  let switch = SwitchBuilder
+    .new()
+    .withRng(rng())
+    .withAddresses(listenAddrs, false)
+    .withTcpTransport()
+    .withMplex()
+    .withNoise()
+    .build()
+  switch.add(natService)
```
vladopajic: `switch.add` is deprecated and should not be used here

tests/libp2p/services/test_natservice.nim
```diff
@@ -0,0 +1,97 @@
...
+      explicitIp = parseIpAddress("203.0.113.7")
+      cfg = NATConfig.init(natModeExplicitIp, Opt.some(explicitIp))
+      switch = SwitchBuilder
+        .new()
+        .withRng(rng())
+        .withAddresses(@[MultiAddress.init("/ip4/127.0.0.1/tcp/0").tryGet()], false)
+        .withTcpTransport()
+        .withMplex()
+        .withNoise()
+        .withNAT(cfg)
```
vladopajic: `makeStandardSwitchBuilder` ?

tests/libp2p/services/test_natservice.nim
```diff
@@ -0,0 +1,97 @@
...
+      switch =
+        createSwitch(natService, @[MultiAddress.init("/ip4/127.0.0.1/tcp/0").tryGet()])
+
+    await switch.start()
+    defer:
+      await switch.stop()
+
+    check switch.peerInfo.listenAddrs.len == 1
+    let boundPort = switch.peerInfo.listenAddrs[0][multiCodec("tcp")].get
+    let expected = MultiAddress.init("/ip4/203.0.113.7" & $boundPort).tryGet()
```
vladopajic: would be better if `explicitIp` is used here to construct `expected`

tests/libp2p/services/test_natservice.nim
```diff
@@ -0,0 +1,97 @@
...
+# Copyright (c) Status Research & Development GmbH
+
+{.used.}
+
+import std/net
+import chronos
+import ../../../libp2p/[builders, switch, services/natservice, multiaddress, multicodec]
+import ../../tools/[unittest, crypto]
+
+proc createSwitch(
```
vladopajic: `makeSwitch`

tests/libp2p/services/test_natservice.nim
vladopajic: or should be `makeSwitch` from above?

libp2p/services/natservice.nim
```diff
@@ -0,0 +1,124 @@
...
+    let (netCodec, transportProto, isQuic) =
+      if tcpMatch:
+        (multiCodec("tcp"), tcpProtocol, false)
+      else:
+        (multiCodec("udp"), udpProtocol, true)
+
+    listenAddr.getProtocolArgument(netCodec).withValue(portArg):
+      let port = Port(uint16.fromBytesBE(portArg))
+      let base = MultiAddress.init(explicitIp, transportProto, port)
+      if isQuic:
```
richard-ramos: you need another condition for `ws`, `wss` and `tls/ws`, otherwise websocket addresses will lose the suffix, i.e.:
`/ip4/1.2.3.4/tcp/123/wss` will become `/ip4/1.2.3.4/tcp/123/`

libp2p/services/natservice.nim
```diff
@@ -0,0 +1,124 @@
...
+method stop*(self: NATService, switch: Switch) {.async: (raises: [CancelledError]).} =
+  trace "Stopping NATService"
+  if self.addressMapper != nil:
+    switch.peerInfo.addressMappers.keepItIf(it != self.addressMapper)
+  # Clear the explicit announce owned by this service, but do not trigger
+  # peerInfo.update() during shutdown because it may notify observers and
+  # cause them to publish address changes while the switch is tearing down.
+  # Address recalculation is deferred until the next Switch.start or an
+  # explicit refresh performed at a safe lifecycle point.
+  switch.peerInfo.announcedAddrs = @[]
```
richard-ramos: Wont this clear also those addresses you would set with `withAnnouncedAddresses`?

libp2p/services/natservice.nim
```diff
@@ -0,0 +1,124 @@
...
+      ## is performed.
+      ## TODO: In this skeleton it is a no-op;
+    natModeExplicitIp
+      ## Static external IP, no probing. The configured `explicitIp` is combined
+      ## with the bound listen ports to produce `peerInfo.announcedAddrs`.
+
+  NATConfig* = object
+    mode*: NATMode
+    explicitIp*: Opt[IpAddress]
+
```
richard-ramos, line 30: To avoid creating incorrect configs, (like for example:
```nim
NATConfig(mode: natModeExplicitIp, explicitIp: Opt.none(...))
```

you can use a variant object like this:

```nim
 NATConfig* = object
    case mode*: NATMode
    of natModeExplicitIp:
      explicitIp*: IpAddress
    else:
      discard
```

libp2p/services/natservice.nim
```diff
@@ -0,0 +1,124 @@
...
+  NATMode* = enum
+    natModeAuto
+      ## Use libp2p-native mechanisms (autonat / hole-punching). No port-mapping
+      ## is performed.
+      ## TODO: In this skeleton it is a no-op;
+    natModeExplicitIp
+      ## Static external IP, no probing. The configured `explicitIp` is combined
+      ## with the bound listen ports to produce `peerInfo.announcedAddrs`.
+
+  NATConfig* = object
```
richard-ramos, line 24: These items need to be re-exported in builders.nim, (by adding `natservice` to `libp2p/builders.nim` line 40) otherwise someone using `withNAT` will also need to `import libp2p/services/natservice`

libp2p/services/natservice.nim
```diff
@@ -0,0 +1,124 @@
...
+  T(config: config)
+
+proc init*(
+    T: typedesc[NATConfig],
+    mode: NATMode,
+    explicitIp: Opt[IpAddress] = Opt.none(IpAddress),
+): T =
+  T(mode: mode, explicitIp: explicitIp)
+
+proc explicitIpMultiAddrs(
```
richard-ramos: The content of this function is a bit complex. Probably makes sense to split it into separate helpers

libp2p/services/natservice.nim
```diff
@@ -0,0 +1,124 @@
...
+# Copyright (c) Status Research & Development GmbH
+
+{.push raises: [].}
+
+import std/[net, sequtils]
+import stew/endians2
+import chronos, chronicles, results
+import ../[multiaddress, multicodec]
+import ../switch
+import ./wildcardresolverservice
```
richard-ramos: Was wondering why `wildcardresolverservice` was used, but looks like it is only for the `getProtocolArgument` proc.

This is not introduced in this PR, but doesn't it seems odd that this proc lives there and not inside multiaddress.nim?

libp2p/services/natservice.nim
vladopajic: based on code i though it should only work for quic and tcp, but it looks like it should work on anything on top of ipv4/ipv6? in that case tor address will also fail, or anything introduced later.

right way would be to have some method that will do `replaceIp(explictIp)` in `MultiAddress` type. for example currently there is method `getIp` in `MultiAddress`, so something like that that will replace value.

richard-ramos: worth mentioning:  this issue also happens in wildcardresolverservice.nim, so probably the solution that is used here, could also be used there, which is a big plus. Here's a test that demonstrates the problem in 
tests/libp2p/services/test_wildcard_resolver.nim: 

```nim
  asyncTest "WildcardAddressResolverService must preserve websocket suffix when resolving wildcard addresses":
    let svc: Service =
      WildcardAddressResolverService.new(networkInterfaceProvider = getAddressesMock)
    let switch = createSwitchWithoutTransports(
      svc, @[MultiAddress.init("/ip4/0.0.0.0/tcp/4/ws").tryGet()]
    )

    await switch.start()

    echo "resolved websocket wildcard addresses:"
    for address in switch.peerInfo.addrs:
      echo "  " & $address

    check switch.peerInfo.addrs ==
      @[
        MultiAddress.init("/ip4/127.0.0.1/tcp/4/ws").get,
        MultiAddress.init("/ip4/192.168.1.22/tcp/4/ws").get,
      ]

    await switch.stop()
```

Expected `switch.peerInfo.addrs`:
```
/ip4/127.0.0.1/tcp/4/ws
/ip4/192.168.1.22/tcp/4/ws
```
Output (note the suffix is gone):
```
resolved websocket wildcard addresses:
  /ip4/127.0.0.1/tcp/4
  /ip4/192.168.1.22/tcp/4
```

libp2p/services/natservice.nim
vladopajic: this is not good suggestion after realizing this https://github.com/vacp2p/nim-libp2p/pull/2493/changes#r3274071334.

libp2p/builders.nim
```diff
@@ -40,7 +40,7 @@ import services/wildcardresolverservice
 export
   switch, peerid, peerinfo, peeraddrpolicy, connection, multiaddress, crypto, errors,
   TLSPrivateKey, TLSCertificate, TLSFlags, ServerFlags, connmanager.ConnectionLimits,
-  connmanager.maxTotal, connmanager.maxInOut
+  connmanager.maxTotal, connmanager.maxInOut, natservice
```
vladopajic: note how we should only export things that are actually need to be exported -> `natservice.NATConfig`

tests/tools/test_multiaddress.nim
```diff
@@ -41,3 +41,28 @@ suite "MultiAddress testing tools":
...
       MultiAddress.init("/ip4/5.6.7.8/tcp/80/ws").get().getIp() ==
         Opt.some(IpAddress(family: IpAddressFamily.IPv4, address_v4: [5'u8, 6, 7, 8]))
+
+  test "replaceIp":
+    let
+      ip4 = parseIpAddress("203.0.113.7")
+      ip6 = parseIpAddress("2001:db8::1")
+    check:
+      MultiAddress.init("/ip4/1.2.3.4/tcp/1234").get.replaceIp(ip4).get ==
+        MultiAddress.init("/ip4/203.0.113.7/tcp/1234").get
```
vladopajic: some local helper would really come handy here

`check replace("/ip4/1.2.3.4/tcp/1234", "203.0.113.7", "/ip4/203.0.113.7/tcp/1234" )`

libp2p/multiaddress.nim
```diff
@@ -1249,6 +1261,42 @@ proc getIp*(ma: MultiAddress): Opt[IpAddress] =
...
+    targetCodec =
+      case ip.family
+      of IpAddressFamily.IPv4: ip4
+      of IpAddressFamily.IPv6: ip6
+    newIp =
+      case ip.family
+      of IpAddressFamily.IPv4:
+        ?MultiAddress.init(targetCodec, ip.address_v4)
+      of IpAddressFamily.IPv6:
+        ?MultiAddress.init(targetCodec, ip.address_v6)
```
vladopajic:
```suggestion
    newIp =
      case ip.family
      of IpAddressFamily.IPv4:
        ?MultiAddress.init(ip4, ip.address_v4)
      of IpAddressFamily.IPv6:
        ?MultiAddress.init(ip6, ip.address_v6)
```

`targetCodec` appears unnecessary

libp2p/services/wildcardresolverservice.nim
```diff
@@ -63,97 +62,47 @@ proc new*(
...
-                )
-              )
-            else:
-              addresses.add(listenAddr)
-        else:
-          addresses.add(listenAddr)
-    else:
+    if not (TCP_IP.matchPartial(listenAddr) or QUIC_V1_IP.matchPartial(listenAddr)):
+      addresses.add(listenAddr)
+      continue
```
vladopajic: not sure if this is needed?

check below should be enough: if listen address has `ip` (and is wildcard ip) replace it.

libp2p/services/natservice.nim
```diff
@@ -0,0 +1,84 @@
...
+  case self.config.mode
+  of natModeAuto:
+    # TODO: wire autonat / hole-punching in here.
+    discard
+  of natModeExplicitIp:
+    let explicitIp = self.config.explicitIp
+    self.addressMapper = proc(
+        listenAddrs: seq[MultiAddress]
+    ): Future[seq[MultiAddress]] {.async: (raises: [CancelledError]).} =
+      return explicitIpMapped(listenAddrs, explicitIp)
```
vladopajic: nit: avoid closure if not necessary; better use `self.config.explicitIp`

in general use closure only when needed because accessing closure has more cost then local variable or object filed.

vladopajic: it's not that it matters here; but for sake future code

libp2p/services/natservice.nim
```diff
@@ -0,0 +1,84 @@
...
+      discard
+
+  NATService* = ref object of Service
+    config: NATConfig
+    addressMapper: AddressMapper
+
+proc new*(T: typedesc[NATService], config: NATConfig): T =
+  T(config: config)
+
+proc explicitIpMapped(
```
vladopajic: this can be easily unit tested

libp2p/services/natservice.nim
```diff
@@ -0,0 +1,84 @@
...
+  ## ``/tls/ws``) are preserved. Addresses without an IP component, or with a
+  ## mismatching family, are dropped.
+  var addrs: seq[MultiAddress]
+  for listenAddr in listenAddrs:
+    let ip = listenAddr.getIp().valueOr:
+      continue
+    if ip.family != explicitIp.family:
+      continue
+    listenAddr.replaceIp(explicitIp).withValue(remapped):
+      addrs.add(remapped)
```
richard-ramos: when you have enabled wildcard resolver in the builder, you might end up with duplicated addresses here, i.e, if you set up this:
- listen address: `/ip4/0.0.0.0/tcp/50000`
- explicit ip  `203.0.113.7`
And if your machine has multiple interfaces, then, wildcard resolver could resolve the wildcard addresses to something like this:
```
/ip4/192.168.1.10/tcp/50000
/ip4/10.0.0.5/tcp/50000
```
Then, the code here will replace only the IP and preserve the port/suffix, thus you could end up with
```
/ip4/203.0.113.7/tcp/50000
/ip4/203.0.113.7/tcp/50000
```
thus announcing the same address twice. 
An easy fix is to dedupe before returning.

richard-ramos: Here's a quick example where u can see what i mean. Assuming you got multiple network interfaces, you should get an output like this:

```
listenAddrs:
  /ip4/0.0.0.0/tcp/34555
announced addrs after wildcard resolver + NAT:
  /ip4/203.0.113.7/tcp/34555
  /ip4/203.0.113.7/tcp/34555
duplicates:
  2x /ip4/203.0.113.7/tcp/34555
```


```nim
import std/[net, tables]
import chronos
import libp2p/[builders, multiaddress]

proc main() {.async.} =
  let
    explicitIp = parseIpAddress("203.0.113.7")
    cfg = NATConfig(mode: natModeExplicitIp, explicitIp: explicitIp)
    switch = SwitchBuilder
      .new()
      .withAddresses(@[MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet()])
      .withTcpTransport()
      .withMplex()
      .withNoise()
      .withNAT(cfg)
      .build()

  await switch.start()
  defer:
    await switch.stop()

  echo "listenAddrs:"
  for address in switch.peerInfo.listenAddrs:
    echo "  ", address

  echo "announced addrs after wildcard resolver + NAT:"
  var counts = initCountTable[string]()
  for address in switch.peerInfo.addrs:
    let key = $address
    counts.inc(key)
    echo "  ", key

  echo "duplicates:"
  var found = false
  for address, count in counts:
    if count > 1:
      found = true
      echo "  ", count, "x ", address
  if not found:
    echo "  none"

waitFor main()

```

richard-ramos (approved):

vladopajic (approved):

Issue #2499, opened 2026-05-20
service-disco: valid service-table peers are dropped when Kad bucketIndex is 16 or higher


rlve:
## Problem

ServiceDiscovery builds service-specific tables with 16 buckets, but inserts peers through the generic Kad routing table logic.

The spec says service-specific tables use `m = 16`, but they do not use raw Kad bucket indexes directly. They must map the 256-bit XOR distance into 16 service buckets:

`i = min(floor((lz * m) / 256), m - 1)`

So a peer with `lz = 16` is valid. It should go into service bucket `1`, but current implementation drops it.

## Spec

The spec defines `m = 16` for service-specific tables ([link](https://github.com/logos-co/logos-lips/blob/1383661f9d566dde0d1b0de0e3ed01ed602ed34a/docs/anoncomms/raw/logos-service-discovery.md?plain=1#L282)).

It says service tables are centered on `service_id_hash` ([link](https://github.com/logos-co/logos-lips/blob/master/docs/anoncomms/raw/logos-service-discovery.md?plain=1#L295-L302)):

```
`KadDHT(peerID)` table is centered on `peerID`.
`AdvT(service_id_hash)`, `DiscT(service_id_hash)`
and `RegT(service_id_hash)` are service-specific tables
and MUST be centered on `service_id_hash`.
Bootstrapping of service-specific tables
MAY be done from the `KadDHT(peerID)` table.
```

It defines the required bucket formula ([link](https://github.com/logos-co/logos-lips/blob/master/docs/anoncomms/raw/logos-service-discovery.md?plain=1#L304-L317)):

```
When inserting a peer into the service-specific tables,
the bucket index into which the peer will be inserted MUST be determined by:

- x = reference ID which is the `service_id_hash`
- y = target peer ID `peerID`
- L = 256 = bit length of IDs
- `m` = 16 = number of buckets in the advertise/search table
- `d = x ⊕ y = service_id_hash ⊕ peerID` = bitwise XOR distance (interpreted as an unsigned integer)

The bucket index `i` where `y` is placed in `x`'s table is calculated as:

- Let `lz = CLZ(d)` = number of leading zeros in the 256-bit representation of `d`
- `i = min( ⌊ (lz * m) / 256 ⌋ , m − 1 )`
- Special case: For `d = 0` (same ID), `i = m - 1`
```

## Current implementation

[...]

PR #2500, opened 2026-05-20
test(service-disco): add repro test for #2499

rlve:
## Summary

This PR adds a ServiceDiscovery component test for #2499.

The test pins one registrar identity so its Kad `bucketIndex` is exactly `16` for the advertised service. That registrar is present in the advertiser's main Kad routing table, but it is dropped when `addProvidedService` seeds the service-specific table.

A second registrar stays inside the current bucket range and proves the happy path still works.

The test also updates the ServiceDiscovery test helper so tests can create a node with a fixed private key.

## Affected Areas

- [x] Tests

## Impact on Library Users

None.

## Risk Assessment

None.

## References

- https://github.com/vacp2p/nim-libp2p/issues/2499
- https://github.com/logos-co/logos-lips/blob/master/docs/anoncomms/raw/logos-service-discovery.md?plain=1#L304-L317

### DIFF
--- a/tests/libp2p/service_discovery/utils.nim
+++ b/tests/libp2p/service_discovery/utils.nim
@@ -69,15 +69,19 @@ proc makeAdvertisement*(
   )
   SignedExtendedPeerRecord.init(privateKey, extRecord).get()
 
-proc createSwitch*(): Switch =
-  SwitchBuilder
+proc createSwitch*(privateKey: Opt[PrivateKey] = Opt.none(PrivateKey)): Switch =
+  var builder = SwitchBuilder
     .new()
     .withRng(rng())
     .withAddresses(@[MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet()])
     .withTcpTransport()
     .withMplex()
     .withNoise()
-    .build()
+
+  privateKey.withValue(key):
+    discard builder.withPrivateKey(key)
+
+  builder.build()
 
 proc testKadDHTConfig(): KadDHTConfig =
   KadDHTConfig.new(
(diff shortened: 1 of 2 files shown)


SionoiS (approved):
Thanks!

tests/libp2p/service_discovery/utils.nim
```diff
@@ -69,15 +69,19 @@ proc makeAdvertisement*(
...
     .new()
     .withRng(rng())
     .withAddresses(@[MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet()])
     .withTcpTransport()
     .withMplex()
     .withNoise()
-    .build()
+
+  privateKey.withValue(key):
+    discard builder.withPrivateKey(key)
```
vladopajic, line 82: it would be nicer to add additional `with` method to `SwitchBuilder` in order to avoid braking building.


```
withPrivateKey(b: SwitchBuilder, privateKey: Opt[PrivateKey]): SwitchBuilder
  privateKey.withValue(key):
    b.withPrivateKey(key)
  b
```
note: 
  - `privateKey` in this `with` should not have default value (it should always be specified)
  -  when `privateKey` is set, delegate it to non-optional `withPrivateKey` proc


then you can just add one line to `createSwitch`... `.withPrivateKey(privateKey)`

vladopajic (approved):

tests/libp2p/service_discovery/utils.nim
vladopajic: bonus: 
```
return makeStandardSwitchBuilder(MultiAddress.init("/ip4/0.0.0.0/tcp/0").get())
.withPrivateKey(privateKey)
.build()
```

gmelodie (approved):

PR #2501, opened 2026-05-20
feat(streams): use sink for write buffers

richard-ramos:
## Summary

This PR changes the stream write boundary to accept owned `seq[byte]` buffers via `sink`, allowing write buffers to be moved instead of copied where ownership is transferred.

The change updates the base `LPStream.write` method and matching write implementations across the transport, muxer, secure channel, relay, and bridge stream paths. It also moves locally constructed buffers at clear ownership handoff points, such as length-prefixed writes, muxer send queues, encrypted Noise frames, and relay forwarding.

## Affected Areas

- [x] Transports
  QUIC, WebSocket, and Chronos stream write implementations now match the sink-aware stream API.
- [x] Protocol Logic
  Noise and relay stream writes move owned buffers when forwarding to the next stream layer.
- [x] Other
  Mplex, Yamux, BridgeStream, and stream test mocks were updated to match the new write signature.

## Impact on Library Users

Callers can continue passing `seq[byte]` to `write`; Nim preserves normal value semantics and can move buffers when they are temporaries or last uses.

Implementations that override `LPStream.write` or `Connection.write` must update their method signature to `msg: sink seq[byte]`.

## Risk Assessment

The behavioral surface is intended to stay the same: the change affects ownership at stream write boundaries, not wire format or protocol logic.

The main compatibility risk is source-level for downstream custom stream implementations that override `write` with the old `seq[byte]` parameter.

### DIFF
--- a/libp2p/muxers/mplex/lpchannel.nim
+++ b/libp2p/muxers/mplex/lpchannel.nim
@@ -198,7 +198,7 @@ method readOnce*(
     raise newLPStreamConnDownError(exc)
 
 proc prepareWrite(
-    s: LPChannel, msg: seq[byte]
+    s: LPChannel, msg: sink seq[byte]
 ): Future[void] {.async: (raises: [CancelledError, LPStreamError]).} =
   # prepareWrite is the slow path of writing a message - see conditions in
   # write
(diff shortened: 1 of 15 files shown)


libp2p/muxers/mplex/lpchannel.nim
```diff
@@ -267,12 +267,14 @@ proc completeWrite(
...
 method write*(
-    s: LPChannel, msg: seq[byte]
+    s: LPChannel, msg: sink seq[byte]
 ): Future[void] {.async: (raises: [CancelledError, LPStreamError], raw: true).} =
   ## Write to mplex channel - there may be up to MaxWrite concurrent writes
   ## pending after which the peer is disconnected
 
-  let closed = s.closedLocal or s.conn.closed
+  let
+    msgLen = msg.len
```
vladopajic, line 276: `msgLen` seems unused

vladopajic (approved):

gmelodie (approved):

libp2p/muxers/mplex/lpchannel.nim
richard-ramos: It's used in line 288

PR #2502, opened 2026-05-20
fix(service-disco): build advert once to keep adverts byte-identical

### LINKED ISSUE #2490: flaky test: one advertiser provides two services - both discoverable
https://github.com/vacp2p/nim-libp2p/actions/runs/26106282250/job/76770894936

```
========================================================
  /home/runner/work/nim-libp2p/nim-libp2p/tests/test_all 'Service Discovery Component - Advertise Discover::one advertiser provides two services - both discoverable'
--------------------------------------------------------
    [TIMEOUT] Timeout was reached and the conditions were not true. Check if the code is working as expected or consider increasing the timeout param.
    /home/runner/work/nim-libp2p/nim-libp2p/tests/libp2p/service_discovery/component/test_advertise_discover.nim(133, 44): Check failed: registrarNode.countAdsInCache(svcAId) == 1
    /home/runner/work/nim-libp2p/nim-libp2p/tests/libp2p/service_discovery/component/test_advertise_discover.nim(135, 6): Check failed: block:
  let foundA = await discovererNode.lookup(svcAId)
  let foundB = await discovererNode.lookup(svcBId)
  foundA.get().anyIt(it.data.peerId == advertiserNode.switch.peerInfo.peerId) and
      foundA.get().len == 1 and
      foundB.get().anyIt(it.data.peerId == advertiserNode.switch.peerInfo.peerId) and
      foundB.get().len == 1

```

rlve:
## Summary

Fixes a flaky failure in the service discovery test "one advertiser provides two services - both discoverable". In CI, `lookup` would sometimes return the same provider twice.

**Why it failed.** When a node calls `addProvidedService`, it spawns several `advertiseToRegistrar` tasks: one per remote registrar, plus one for local advertising where the node registers the ad with itself. Each task built its own advertisement by calling `disco.record()`. That record stamps `seqNo` from `Moment.now().epochSeconds`, which has 1-second resolution. When two tasks ran on opposite sides of a second boundary, they produced records with different `seqNo`, and so different signatures and bytes.

The discoverer's `lookup` collects advertisements from every registrar and dedups them by exact byte identity. Two records of the same provider that differ only in `seqNo` are not byte-identical, so both survive and `lookup` returns the provider twice. CI's slower scheduling made straddling the second boundary more likely, so the test failed intermittently.

**The fix.** Build the advertisement once in `addProvidedService` and hand the same bytes to every `advertiseToRegistrar` task. Every registrar now caches a byte-identical record, the dedup collapses them, and `lookup` returns one entry per provider.

## Affected Areas

- [x] Service Discovery

## Impact on Library Users

`advertiseToRegistrar` is exported and its signature changed. `advert: Opt[seq[byte]]` is now `advert: seq[byte]`. Callers passing `Opt.some(bytes)` pass `bytes` directly. Callers that relied on `Opt.none` to get an auto-built record now build it themselves. `addProvidedService` and `startAdvertising` are unchanged.

## Risk Assessment

Low. The change makes advert generation deterministic instead of timing-dependent. One behavior change worth noting: the `services` list in an advert is now snapshotted when `addProvidedService` is called rather than when the spawned task first runs. The old timing was non-deterministic, so this tightens behavior rather than regressing it.

### DIFF
--- a/libp2p/protocols/service_discovery/advertiser.nim
+++ b/libp2p/protocols/service_discovery/advertiser.nim
@@ -76,21 +76,13 @@ proc advertiseToRegistrar*(
     serviceId: ServiceId,
     registrar: PeerId,
     ticket: Opt[Ticket],
-    advert: Opt[seq[byte]],
+    advert: seq[byte],
 ) {.async: (raises: [CancelledError]).} =
   doAssert not disco.clientMode, "not supported in client mode"
 
   if not disco.rtManager.hasService(serviceId):
     error "no service routing table found", serviceId
     return
-
-  let advertBuff = advert.valueOr:
-    let extRecord = disco.record().valueOr:
-      error "failed create extended peer record", error
-      return
-    extRecord.encode().valueOr:
-      error "failed to encode advertisement", error
-      return
 
   cd_advertiser_actions_executed.inc()
 
@@ -100,7 +92,7 @@ proc advertiseToRegistrar*(
 
   while true:
     let response = (
-      await disco.sendRegister(registrar, serviceId, advertBuff, currentTicket)
+      await disco.sendRegister(registrar, serviceId, advert, currentTicket)
     ).valueOr:
       error "failed to register ad", error
       return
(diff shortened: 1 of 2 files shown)


vladopajic (approved):
make sense even if it is not a fix

SionoiS (approved):
Much better like this. Thanks!

gmelodie (approved):

PR #2503, opened 2026-05-20
feat(pubsub): use sink for RPC buffers

richard-ramos:
## Summary

Use `sink seq[byte]` for PubSub RPC buffers that are encoded, queued, sent, or decoded.

This makes ownership explicit on the PubSub send/decode path and moves encoded buffers into queued messages instead of assigning them by value.

## Affected Areas

- [x] Gossipsub / PubSub RPC
  - Encoded send path
  - RPC queue storage
  - RPC constructor helpers
  - RPC decode entry point

## Impact on Library Users

No wire protocol behavior changes are intended. PubSub APIs that accept owned RPC buffers now express that ownership with `sink`.

## Risk Assessment

Low behavioral risk: this is an ownership/refcounting cleanup around existing PubSub message buffers. The main review risk is ensuring moved buffers are not reused after handoff.

### DIFF
--- a/libp2p/protocols/pubsub/rpc/messages.nim
+++ b/libp2p/protocols/pubsub/rpc/messages.nim
@@ -352,45 +352,51 @@ proc byteSize*(rpc: RPCMsg): int =
   rpc.preambleExtension.withValue(v):
     result += v.byteSize
 
-proc withIWant*(_: typedesc[ControlMessage], msgIds: seq[MessageId]): ControlMessage =
-  ControlMessage(iwant: @[ControlIWant(messageIDs: msgIds)])
+proc withIWant*(
+    _: typedesc[ControlMessage], msgIds: sink seq[MessageId]
+): ControlMessage =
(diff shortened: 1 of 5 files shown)


libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -352,63 +352,69 @@ proc byteSize*(rpc: RPCMsg): int =
   rpc.preambleExtension.withValue(v):
     result += v.byteSize
```
gmelodie, line 353: While we're at it, might wanna remove these `result`s

gmelodie (approved):

libp2p/protocols/pubsub/rpc/messages.nim
richard-ramos: I'll ask copilot on a separate PR to replace all instances of `result` across the codebase

vladopajic (approved):

PR #2508, opened 2026-05-21
chore: use `protobuf_serialization` 0.4.0

richard-ramos:
Also bumps the `deps.nix` file and fixes a minor issue in `gen-deps.sh`


vladopajic (approved):

gmelodie (approved):
Is there not a release version for protobuf_serialization? We should really stop using the comit hashes in nimble file

richard-ramos:
> Is there not a release version for protobuf_serialization? We should really stop using the comit hashes in nimble file

but this is what the PR does! :smile:

gmelodie:
> but this is what the PR does! 😄

ah sry i was looking at the pinned file :/

PR #2514, opened 2026-05-21
fix(service-disco): skip double-hashing service ID in bucket index

rlve:
## Summary                                                                                                                                                                             
Service IDs are already SHA-256 hashes of the service name string. When the routing table computed bucket indices, it hashed the self ID unconditionally via `hashFor`, so service tables were computing `XOR(H(serviceId), H(peer))` instead of `XOR(serviceId, H(peer))`. That extra hash means peers end up in wrong buckets, because the XOR distance is measured from a different point.
           
Service discovery routing tables now skip hashing the service ID when computing bucket indices. The PR adds a  `selfIdPreHashed` flag to `RoutingTableConfig` and sets it in the service table constructor. Also adds a direct `xorDistance(a, b: Key)` overload that works on raw keys without hashing.                                                                                                                                                                               
                                                                                                                                                                                         
Tests verify the flag is set and that peers land in buckets computed from `XOR(serviceId, H(peer))` rather than `XOR(H(serviceId), H(peer))`.

  ## Affected Areas

  - [x] Service Discovery
    Service discovery routing table bucketing logic
  - [x] Kademlia
    `bucketIndex` and `xorDistance` in `types.nim` / `routing_table.nim`
  - [x] Tests
  
## Impact on Library Users

`bucketIndex` now takes an optional `selfIdPreHashed` parameter (defaults to `false`). Existing callers are unaffected. `hashFor` is now exported. A new two-argument `xorDistance(a, b: Key)` overload is available for raw key distance without hashing.

## Risk Assessment

Low. The `selfIdPreHashed` flag defaults to `false`, so only code that opts in (service discovery) changes behavior. The previous double-hashing meant service table buckets didn't match the spec. Existing Kademlia routing tables are unaffected.


SionoiS (approved):
Very nice bug catching.

Thanks a lot!

gmelodie (approved):

richard-ramos (approved):

PR #2525, opened 2026-05-25
chore(MultistreamSelect): handle cosmetics

vladopajic:
<s>
this pr fixes issues when all matched protocol handlers are called. this is not correct as first matched protocol handler should win.

`go-libp2p` reference:
https://github.com/multiformats/go-multistream/blob/master/multistream.go#L178C1-L191C1
</s>

just cosmetics changes in the end, like reuse `lookupProtocol` and reduce `try-except` to parts that are only affected

### DIFF
--- a/libp2p/multistream.nim
+++ b/libp2p/multistream.nim
@@ -180,42 +180,44 @@ proc handle*(
     m: MultistreamSelect, stream: Stream, active: bool = false
 ) {.async: (raises: [CancelledError]).} =
   trace "Starting multistream handler", stream, handshaked = active
-  var
-    protos: seq[string]
-    matchers: seq[Matcher]
-  for h in m.handlers:
-    if h.match != nil:
-      matchers.add(h.match)
-    for proto in h.protos:
-      protos.add(proto)
-
-  try:
-    let ms = await MultistreamSelect.handle(stream, protos, matchers, active)
-    for h in m.handlers:
-      if (h.match != nil and h.match(ms)) or h.protos.contains(ms):
-        trace "found handler", stream, protocol = ms
-
-        let protocol = h.protocol
-
-        if not protocol.reserveIncoming(stream.peerId):
-          debug "Inbound stream budget exceeded, rejecting incoming stream",
-            stream, protocol = ms, peerId = stream.peerId
-          return
-
-        try:
-          await protocol.handler(stream, ms)
-        finally:
-          protocol.releaseIncoming(stream.peerId)
-        return
-    debug "no handlers", stream, ms
-  except CancelledError as exc:
-    raise exc
-  except CatchableError as exc:
-    trace "Exception in multistream", stream, description = exc.msg
-  finally:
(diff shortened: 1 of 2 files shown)


libp2p/multistream.nim
```diff
@@ -180,42 +180,47 @@ proc handle*(
...
+          matchers.add(h.match)
+        for proto in h.protos:
+          protos.add(proto)
+      await MultistreamSelect.handle(stream, protos, matchers, active)
+    except LPStreamError as e:
+      trace "Exception in multistream", stream, description = e.msg
+      return
+    except MultiStreamError as e:
+      trace "Exception in multistream", stream, description = e.msg
+      return
```
vladopajic, line 209: getting compiler error when using this code:

```
except LPStreamError, MultiStreamError:
  trace "Exception in multistream", stream, description = getCurrentErrorMsg()
  return
```


```
.assertions.nim(38)       raiseAssert
Error: unhandled exception: closureiters.nim(335, 17) `c[i].kind == nkType`  [AssertionDefect]
```

gmelodie (approved):

libp2p/multistream.nim
```diff
@@ -180,42 +180,44 @@ proc handle*(
...
+          protos.add(proto)
+      await MultistreamSelect.handle(stream, protos, matchers, active)
+    except LPStreamError as e:
+      trace "Exception in MultistreamSelect.handle", stream, description = e.msg
+      return
+    except MultiStreamError as e:
+      trace "Exception in MultistreamSelect.handle", stream, description = e.msg
+      return
+
+  m.lookupProtocol(ms).withValue(p):
```
vladopajic, line 211: now realizing that `withValue` should have been called `ifValue`, perhaps?

`withValue` is more suited when user is giving value to proc, and not when it is getting (here code is getting - templates is actually calling `.get`)..

vladopajic: them last `else` would appear more suited to this template (because of `if-else` naming)

libp2p/multistream.nim
```diff
@@ -180,42 +180,44 @@ proc handle*(
...
-        if not protocol.reserveIncoming(stream.peerId):
-          debug "Inbound stream budget exceeded, rejecting incoming stream",
-            stream, protocol = ms, peerId = stream.peerId
-          return
-
-        try:
-          await protocol.handler(stream, ms)
-        finally:
-          protocol.releaseIncoming(stream.peerId)
-        return
```
richard-ramos, line 209: Not sure I understand what changed with this PR besides some code cleanup. This `return` here in old code meant that once there was a match, the proc would return immediatly, and not continue with other protocols 🤔

richard-ramos (approved):

libp2p/multistream.nim
vladopajic: ah, totally missed that. i'll rewrite the pr title and still keep pr

Issue #2526, opened 2026-05-25
service-disco: registrar closerPeers do not use RegT


rlve:
## Description

The registrar's service-specific routing table (RegT) is populated but never read. Both `REGISTER` and `GET_ADS` responses populate `closerPeers` from the main Kad routing table via [`findClosestPeers`](https://github.com/vacp2p/nim-libp2p/blob/d6e8dc475e5c343d04eb7aa8fb4fa49c66b74fc4/libp2p/protocols/kademlia/find.nim#L314-L319), which sorts by XOR distance and returns the K closest peers. The service table created in [`rtManager.addService`](https://github.com/vacp2p/nim-libp2p/blob/d6e8dc475e5c343d04eb7aa8fb4fa49c66b74fc4/libp2p/protocols/service_discovery/routing_table_manager.nim#L27) and written to in [`acceptAdvertisement`](https://github.com/vacp2p/nim-libp2p/blob/d6e8dc475e5c343d04eb7aa8fb4fa49c66b74fc4/libp2p/protocols/service_discovery/registrar.nim#L383-L390) is never consulted by the registrar.

This has four consequences:

1. **closerPeers are service-agnostic.** Two unrelated services get identical closerPeers because both come from the same main Kad table. Advertisers and discoverers cannot refine their service-specific routing tables toward peers that are relevant to a particular service.

2. **closerPeers lack bucket diversity.** `findClosestPeers` returns K peers sorted by XOR distance, which cluster in a few buckets. The spec's GETPEERS returns one random peer per bucket across all m=16 buckets, spreading suggestions across the keyspace.

3. **GET_ADS does not add discoverer to RegT.** [`getAdvertisements`](https://github.com/vacp2p/nim-libp2p/blob/d6e8dc475e5c343d04eb7aa8fb4fa49c66b74fc4/libp2p/protocols/service_discovery/registrar.nim#L489-L501) returns a response without writing the discoverer's peer to any service table.

[...]

PR #2528, opened 2026-05-25
test(service-disco): closer peers

rlve:
## Summary
These tests document spec violations tracked in #2526: both REGISTER and GET_ADS responses source closer peers from the main Kad routing table instead of the per-service RegT. 

  New tests:
  - REGISTER closerPeers come from main Kad table, ignoring peers that exist only in RegT
  - GET_ADS closerPeers come from main Kad table, same problem
  - REGISTER with Wait status does not create or write to RegT
  - REGISTER with Confirmed is the only path that writes to RegT
  - GET_ADS does not add the discoverer to RegT

  ## Affected Areas

  - [x] Tests

  ## Impact on Library Users

  None.

  ## Risk Assessment

  None.

  ## References

  - #2526


tests/libp2p/service_discovery/component/test_registrar_closer_peers.nim
```diff
@@ -0,0 +1,200 @@
...
+    let response = await advertiserNode.sendRegister(
+      registrarNode.switch.peerInfo.peerId, serviceId, adBytes
+    )
+    check response.isOk()
+    check response.get().status == kad_protobuf.RegistrationStatus.Wait
+
+    # Spec: the registrar adds the advertiser to RegT on every REGISTER.
+    # Implementation: RegT is not created until Confirmed.
+    let serviceTable = registrarNode.rtManager.getTable(serviceId)
+    check serviceTable.isNone()
```
gmelodie: Would it be better to have the serviceTable have a couple entries and check that this register that returns Wait doesn't add anything to it?

gmelodie: Then we could collapse this and the next test into only one

gmelodie (approved):
Comments and PR title reference `RegT` but code doesn't, would be better that the code had a variable named `regT` or that the PR title/comments made the link between the two names

tests/libp2p/service_discovery/component/test_registrar_closer_peers.nim
rlve: Updated and combined these 2 tests.

RegT refers to a Service table and the role of Registrar as per the spec.

SionoiS:
Wut? I don't understand, we are testing that the code is indeed bugged?

rlve:
> Wut? I don't understand, we are testing that the code is indeed bugged?

These tests also work as a documentation of a current behaviour and a proof.

It's also good to see them fail, once the code is fixed, plus less work for the implementer to switch the assertion instead of developing a new test from scratch.

richard-ramos (approved):

PR #2531, opened 2026-05-26
chore: more standard builders

vladopajic:
all that boilerplate shouldn't really repeat 

bonus changes:
- removed default address in switch builder
- fixing autorelay flaky test (started appearing with no obvious change in this pr)


libp2p/builders.nim
```diff
@@ -97,12 +98,9 @@ type
 proc new*(T: type[SwitchBuilder]): T =
   ## Creates a SwitchBuilder
 
-  let address =
-    MultiAddress.init("/ip4/127.0.0.1/tcp/0").expect("Should initialize to default")
-
```
vladopajic, line 102: this can lead to bugs in downstream deps:

if user has
```
SwitchBuilder.new()
  .withTcpTransport()
  ...
  . build()
```

without `withAddress` (because it will work); and if the they want to switch to other transport

```
SwitchBuilder.new()
  .withQuicTransport()
  ...
  . build()
```

starting switch will fail as there is not quic address. this shouldn't really exists in code.

libp2p/builders.nim
```diff
@@ -433,6 +442,9 @@ proc buildSwitch(b: SwitchBuilder): Switch {.raises: [LPError].} =
   if b.rng == nil: # newRng could fail
     raise newException(Defect, "Cannot initialize RNG")
 
+  if b.addresses.len == 0:
+    raise newException(Defect, "use withAddress() to specify listen address")
```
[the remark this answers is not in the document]
vladopajic: makes sense; reverted this; 

it is also not really needed as https://github.com/vacp2p/nim-libp2p/blob/more-standard-switch/tests/libp2p/transports/basic_tests.nim#L92 this tests will ensure that all transports raise error if address is not provided.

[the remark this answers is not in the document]
vladopajic: but this makes valid point; if user only wants do dial, currently, they would still have to `start()` switch. otherwise services will not be working.

gmelodie (approved):

libp2p/builders.nim
```diff
@@ -97,12 +98,9 @@ type
 proc new*(T: type[SwitchBuilder]): T =
   ## Creates a SwitchBuilder
 
-  let address =
-    MultiAddress.init("/ip4/127.0.0.1/tcp/0").expect("Should initialize to default")
-
   SwitchBuilder(
     privKey: Opt.none(PrivateKey),
-    addresses: @[address],
+    addresses: @[],
```
richard-ramos, line 103: This is a nice find. There's nothing in the spec that says that we should have at least one address. In other implementations you can also create non listening clients like in go-libp2p:

```go
package main

import (
	"context"
	"fmt"
	"time"

	libp2p "github.com/libp2p/go-libp2p"
	"github.com/libp2p/go-libp2p/core/peer"
	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
)

func main() {
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()

	// Node B listens.
	nodeB, err := libp2p.New(
		libp2p.ListenAddrStrings("/ip4/127.0.0.1/tcp/0"),
	)
	if err != nil {
		panic(err)
	}
	defer nodeB.Close()

	// Node A does not listen on any multiaddress.
	nodeA, err := libp2p.New(
		libp2p.NoListenAddrs,
	)
	if err != nil {
		panic(err)
	}
	defer nodeA.Close()

	fmt.Println("node A listen addrs:", nodeA.Addrs()) // should be []
	fmt.Println("node B listen addrs:", nodeB.Addrs())

	bInfo := peer.AddrInfo{
		ID:    nodeB.ID(),
		Addrs: nodeB.Addrs(),
	}

	if err := nodeA.Connect(ctx, bInfo); err != nil {
		panic(err)
	}

	res := <-ping.Ping(ctx, nodeA, nodeB.ID())
	if res.Error != nil {
		panic(res.Error)
	}

	fmt.Println("ping RTT:", res.RTT)
}
```

richard-ramos: Hm... looking more into the code, it seems that when starting the transport we are making the code require at least one address in quictransport, tcptransport and wstransport  🤔 .

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
```diff
@@ -251,17 +251,18 @@ proc createLibp2p(appCallbacks: AppCallbacks, config: Libp2pConfig): LibP2P =
...
 
   case transport
   of TransportType.QUIC:
     switchBuilder = switchBuilder.withQuicTransport()
+    if addrs.len == 0:
+      addrs.add(MultiAddress.init("/ip4/127.0.0.1/udp/0/quic-v1").get())
   of TransportType.TCP:
     switchBuilder = switchBuilder.withTcpTransport()
+    if addrs.len == 0:
+      addrs.add(MultiAddress.init("/ip4/127.0.0.1/tcp/0").get())
```
richard-ramos: It shouldn't be necessary to have a default address. Not necessarily you want to start the 'server' side of a transport, and might instead just do the 'client' side only. 

(altho currently it seems to not be possible the way transports check for the existence of at least one address before raising a `no addr was provided.` error :( )

vladopajic: the only reason why code worked before is because there was default address in builder. currently how code is set up address needs to be specified.

libp2p/builders.nim
vladopajic: yes but you dont have to start transport, if you want to dial you can just `.dial(...)`

vladopajic: start aka start listening 
dial is dial
stop akak stop listening

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
vladopajic: ah no

vladopajic: maybe example needa to have addressed instead

richard-ramos (approved):

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_lifecycle_requests.nim
richard-ramos: yeah, that way it would look closer to how switch builder behaves by default with this PR

PR #2534, opened 2026-05-26
chore(autotls): encapsulate broker interaction behind AutotlsBroker

### LINKED ISSUE #2520: autotls: followups
- enforce correct naming `brokerClient` should not be variable name for `PeerIDAuthClient` not here is no broker in later 

- if broker is abstraction on top of `PeerIDAuthClient` then introduce that type

- enforce api encapsulation of auto tls broker api
  this type is supposed to encapsulate all interaction with broker. currently it doesn't fully encapsulate it's logic, for example:

   ```nim
     https://github.com/vacp2p/nim-libp2p/blob/4de121c602743045573525ffa76e8d140ce038e1/libp2p/autotls/service.nim
      let strMultiaddresses: seq[string] = addrs.mapIt($it)
      let payload = %*{"value": keyAuth, "addresses": strMultiaddresses}
      let registrationURL =
        parseUri("https://" & self.config.brokerURL & "/v1/_acme-challenge")
  
      trace "Sending challenge to AutoTLS broker"
      let (bearer, response) =
        await self.brokerClient.send(registrationURL, self.peerInfo, payload, self.bearer)
   ```
  here details of payload are constructed in service, then passed to `brokerClient.send`. instead `brokerClient.send` should internally create request and give back response. that his 

   `let resp = await brokerClient.sendChalleng(peerInfo, addrs, ...)` 
   note details of how addresses are constructed is also encapsulated.

- ideally `AutotlsService` should not deal with anything https any creation of request, payloads or checking results such as `response.status != HttpOk:` should be encapsulated in `api` / `client` types.  `AutotlsService` should not check status code it should make a call (to `api` or `client` however we call it) and should get response or error if call has failed.

- rename `ACMEChallengeResponseWrapper` to `ACMEChallengeDns01Response` as this is specifc type than only exposes `dns01` challenge (it also has property of same name). 

- ensure API error handling 
  make sure that all http/api errors are propagated
  make sure that all http/api errors are correctly handled by service. for example let's encrypt can returns temporary error like "server buys" - could code attempt to create certificate with some timeout?

gmelodie:
## Summary

Follow-ups to the AutoTLS work, addressing #2520.

Previously `AutotlsService` reached directly into HTTP/PeerID-Auth details: it built the JSON payload and registration URL, called `PeerIDAuthClient.send`, juggled the bearer token, and inspected `response.status`. The field was even named `brokerClient: PeerIDAuthClient`, a misnomer since `PeerIDAuthClient` knows nothing about the broker.

This PR introduces an `AutotlsBroker` type that owns all broker interaction, so the service simply makes a call and gets back success or an error.

- **New `AutotlsBroker`** (`libp2p/autotls/broker.nim`): wraps `PeerIDAuthClient`, owns the broker URL and bearer token, and exposes `sendChallenge(peerInfo, addrs, keyAuth)`. Request construction (multiaddress list, payload, URL) and response validation are internal.
- **`AutotlsService`** no longer touches HTTP: the ~25-line broker block collapses to one `await self.broker.sendChallenge(...)`. Removed the encapsulation-leaking `bearer` field and the exported `HttpOk`/`HttpCreated` constants; the broker URL default moved to `DefaultBrokerURL` (renamed from `AutoTLSBroker` to avoid a Nim identifier clash with the new type).
- **Renamed** `ACMEChallengeResponseWrapper` → `ACMEChallengeDns01Response`, since the type only exposes the `dns01` challenge.
- **Error propagation**: the broker now raises `AutoTLSError` (including the response body) on a non-OK broker status instead of the service silently returning `false`. Combined with the existing `tryIssueCertificate` retry loop (`issueRetries`/`issueRetryTime`), transient broker/ACME failures such as Let's Encrypt "server busy" are propagated and retried rather than swallowed.

## Affected Areas

- [x] Protocol Logic
  
- [x] Build / Tooling
  

Internal refactor of the AutoTLS module (gated behind `-d:libp2p_autotls_support`). No downstream integration changes required.

## Impact on Library Users

Minimal. AutoTLS is opt-in via `-d:libp2p_autotls_support`. Public surface changes within that module:
- `AutotlsService.brokerClient`/`.bearer` fields removed, replaced by `AutotlsService.broker: AutotlsBroker`.
- `AutoTLSBroker` constant renamed to `DefaultBrokerURL`.
- Exported `HttpOk`/`HttpCreated` constants removed from `autotls/service`.
- `ACMEChallengeResponseWrapper` renamed to `ACMEChallengeDns01Response`.

## Risk Assessment

Low. Pure refactor with no behavior change to the issuance flow; the only semantic change is that a failed broker registration now raises (and is retried) instead of returning `false` (which also triggered a retry). Existing unit suites pass: AutoTLS ACME (12), config (4), peer-id-auth (4).

## References

- Closes #2520


richard-ramos (approved):

libp2p/autotls/service.nim
vladopajic: these logs are quite strange since they are not logging real values

vladopajic: better move these to `checkDNSRecords`, they are more related to that code there

vladopajic: it seems unnecessary for this proc to return `bool`.
- all cases that lead to unsuccessful outcome should raise error. code is already like this; better follow in through in reaming 1-2 cases.
- if you prefer to return something, return certificate.

vladopajic: `let acmeClient = self.acmeClient` unnecessery

vladopajic: `let domain = api.Domain("*." & baseDomain)`

better:
- move it as close to as where it is needed
- or avoid this variable

vladopajic: better raise error

vladopajic: nit `: string`, unnecessery

vladopajic: theoretically this could fail; but unlikely

vladopajic, line 196: maybe better if moved closer to `let certKeyPair`

vladopajic: better: just inline this in `TLSPrivateKey.init(derPrivKey.pemEncode("PRIVATE KEY")),`

vladopajic: always log error message

vladopajic: this `trace` can be higher log level. it doesn't happen often and it seems quite nice info to see in logs.
maybe improve message "AutoTLS successfully renewed certificate"?

libp2p/autotls/service.nim
```diff
@@ -8,28 +8,25 @@ import chronos/apps/http/httpclient
 
 import
   ./acme/client,
+  ./broker,
```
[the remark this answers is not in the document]
vladopajic: seems legit

vladopajic (approved):

vladopajic:
disabling auto merge, please have one more cosmetic swipe

libp2p/autotls/service.nim
vladopajic: what i wanted to say/suggest is to do safe get

gmelodie:
Removed `run-daily-ci`, was getting rate limited

PR #2536, opened 2026-05-26
chore: replace unmaintained dnsclient with internal DNS codec

### LINKED ISSUE #2301: Get rid of unmaintained dnsclient.nim
Currently we're using https://github.com/ba0f3/dnsclient.nim, we should fork and maintain it or switch to another maintained nim dns client. Maybe https://github.com/fox0430/chronos-dnsc?

gmelodie:
## Summary

Closes #2301.

`nim-libp2p` depended on the unmaintained [`ba0f3/dnsclient.nim`](https://github.com/ba0f3/dnsclient.nim), but used only a thin slice of it. The UDP transport is already done in-house via chronos `DatagramTransport` (`getDnsResponse`); `dnsclient` was used purely as a **wire codec** — building a query and decoding `A`/`AAAA`/`TXT` answers.

This PR drops the dependency entirely and vendors a minimal codec instead of forking or swapping in another (also lightly maintained) library. Benefits:

- No external DNS dependency to track or maintain — removed from `libp2p.nimble`, `.pinned`, and all `nix/` build files.
- `libp2p/nameresolving/dnsmessage.nim` (new) is pure bytes-in/bytes-out under `{.push raises: [].}`, raising only `ValueError` on malformed input (never a `Defect`). It uses `stew/endians2`, already a dependency.
- `dnsresolver.nim` shrinks substantially: `questionToBuf`, the `StringStream` round-trip, and every `except Exception` / `raiseAssert` / "Nim 1.6" workaround are gone — those existed only because the old library leaked `Exception`. The async `raises` set narrows to `[CancelledError, IOError, TransportError, ValueError]`.

Resolver behavior is unchanged: name-server failover/rotation, AAAA-first ordering for IPv6 servers, and the 5s timeout are preserved.

> [!NOTE]
> The codec parses answer rdata (`A`/`AAAA`/`TXT`) but only *skips* answer names (it still follows/terminates compression pointers correctly) since the resolver never reads them. If a future record type needs the decoded name, that logic would need to be added back.

## Affected Areas

- [x] Peer Management / Discovery — DNS name resolution (`libp2p/nameresolving`)
- [x] Build / Tooling — `libp2p.nimble`, `.pinned`, `nix/{deps,cbind,default}.nix`

No public API change — `DnsResolver` keeps the same constructor and `resolveIp`/`resolveTxt` signatures. No downstream integration PRs required.

## Impact on Library Users

Internal refactor with no API change. Downstream users get one fewer (unmaintained) transitive dependency. The DNS resolution surface is unchanged.

## Risk Assessment

- **Backward compatibility:** none affected — same types and method signatures.
- **Correctness:** the new codec is validated against the real captured `status.im` `A`/`AAAA` responses (which use `0xc00c` compression pointers) that already back the resolver tests, so parity with the old library is demonstrated. Malformed/truncated inputs now raise `ValueError` rather than risking an `IndexDefect`.
- **Security:** all reads are bounds-checked; oversized labels/names and over-512-byte packets are rejected per RFC 1035.

## References

- Issue: #2301

## Additional Notes

Verification:
- New `tests/libp2p/test_dnsmessage.nim` (auto-discovered): query encoding, illegal-name rejection, `A`/`AAAA` decoding against compressed fixtures, `TXT` concatenation, and malformed-input handling — **6/6 pass**.
- Existing `tests/libp2p/test_name_resolve.nim` (drives the full `DnsResolver` against a mock UDP server) — **10/10 pass**.
- `nim check` clean on the codec, resolver, and library root; `nph`-formatted.
- `rg -i dnsclient` returns no hits.


libp2p/nameresolving/dnsresolver.nim
```diff
@@ -3,13 +3,7 @@
...
 {.push raises: [].}
 
-import
-  std/[streams, sets, sequtils],
-  chronos,
-  chronicles,
-  stew/byteutils,
-  dnsclientpkg/[protocol, types],
-  ../utility
+import std/[sets, sequtils, random], chronos, chronicles, ./dnsmessage
```
richard-ramos: Could we use libp2p's Rng instead of the one from std?

libp2p/nameresolving/dnsmessage.nim
```diff
@@ -0,0 +1,166 @@
...
+  ## `A`/`AAAA`/`TXT` answers are returned; other record types are skipped.
+  if data.len < HeaderSize:
+    raise newException(ValueError, "DNS response shorter than header")
+  if data.len > MaxPacketSize:
+    raise newException(ValueError, "DNS response exceeds 512 bytes")
+
+  var r = DnsReader(data: @data, pos: 0)
+
+  discard r.readShort() # id
+  discard r.readShort() # flags
```
richard-ramos: Shouldn't these two be validated?

libp2p/nameresolving/dnsmessage.nim
```diff
@@ -0,0 +1,166 @@
...
+  let qdcount = r.readShort()
```
richard-ramos, line 152: Should we validate there's at least one question?

libp2p/nameresolving/dnsmessage.nim
```diff
@@ -0,0 +1,166 @@
...
+
+proc parseTxt(rdata: openArray[byte]): string =
+  ## TXT rdata is a sequence of length-prefixed character-strings.
+  var text: string
+  var i = 0
+  while i < rdata.len:
+    let length = rdata[i].int
+    inc i
+    if i + length > rdata.len:
+      break
```
richard-ramos:
```suggestion
    if i + length > rdata.len:
      raise newException(ValueError, "invalid TXT record")
```
or maybe log it? for the strange case in which a malformed packet is received. This operation only happens on startup, so it should happen rarely enough that having an `error` log will not be noisy at all and might be useful to catch these scenarios.

libp2p/nameresolving/dnsmessage.nim
```diff
@@ -0,0 +1,166 @@
...
+    discard r.readShort() # class
+    discard r.readShort() # ttl high
+    discard r.readShort() # ttl low
+    let rdlength = r.readShort().int
+    let rdata = r.readBytes(rdlength)
+
+    case rrType
+    of A.uint16:
+      if rdata.len == 4:
+        answers.add(DnsAnswer(kind: A, value: ipv4ToString(rdata)))
```
richard-ramos:
```suggestion
      if rdata.len == 4:
        answers.add(DnsAnswer(kind: A, value: ipv4ToString(rdata)))
      else:
        raise newException(ValueError, "invalid A record")
```
or maybe log it?

libp2p/nameresolving/dnsmessage.nim
```diff
@@ -0,0 +1,166 @@
...
+    of AAAA.uint16:
+      if rdata.len == 16:
+        answers.add(DnsAnswer(kind: AAAA, value: ipv6ToString(rdata)))
```
richard-ramos:
```suggestion
      if rdata.len == 16:
        answers.add(DnsAnswer(kind: AAAA, value: ipv6ToString(rdata)))
      else:
        raise newException(ValueError, "invalid AAA record")
```
or maybe log it?

libp2p/nameresolving/dnsmessage.nim
```diff
@@ -0,0 +1,166 @@
...
+  var i = 0
+  while i < rdata.len:
+    let length = rdata[i].int
+    inc i
+    if i + length > rdata.len:
+      break
+    for j in 0 ..< length:
+      text &= char(rdata[i + j])
+    inc i, length
+  return text
```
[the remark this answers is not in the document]
pcmoen: ✅ **Fikset** — Bruker nå `entries()` og entry-key (trimmet) som `userObjectId` for konsistens med assessments-keys.

libp2p/utility.nim
```diff
@@ -40,6 +40,27 @@ func shortLog*(item: string): string =
     s &= item[(item.len - split) .. item.high]
     s
 
+func ipv4ToString*(bytes: openArray[byte]): string {.raises: [ValueError].} =
```
vladopajic: have in mind #2538

libp2p/nameresolving/dnsmessage.nim
```diff
@@ -0,0 +1,166 @@
...
+proc readShort(r: var DnsReader): uint16 {.raises: [ValueError].} =
+  if r.pos + 2 > r.data.len:
+    raise newException(ValueError, "Truncated DNS message")
+  let value = fromBytesBE(uint16, r.data.toOpenArray(r.pos, r.pos + 1))
+  inc r.pos, 2
+  return value
+
+proc readBytes(r: var DnsReader, n: int): seq[byte] {.raises: [ValueError].} =
+  if n < 0 or r.pos + n > r.data.len:
+    raise newException(ValueError, "Truncated DNS message")
```
vladopajic: `template raiseErr(msg: string)`

libp2p/nameresolving/dnsmessage.nim
```diff
@@ -0,0 +1,166 @@
...
+    of A.uint16:
+      if rdata.len == 4:
+        answers.add(DnsAnswer(kind: A, value: ipv4ToString(rdata)))
+    of AAAA.uint16:
+      if rdata.len == 16:
+        answers.add(DnsAnswer(kind: AAAA, value: ipv6ToString(rdata)))
+    of TXT.uint16:
+      answers.add(DnsAnswer(kind: TXT, value: parseTxt(rdata)))
+    else:
+      discard # unsupported record type, already skipped via rdlength
```
vladopajic: okay, but still raise error

vladopajic (approved):

libp2p/utility.nim
gmelodie: Done — after the #2538 merge `utility.nim` is gone; `ipv4ToString`/`ipv6ToString` now live in `libp2p/utils/conversion.nim`, with dedicated tests in `tests/libp2p/utils/test_conversion.nim`.

richard-ramos (approved):

PR #2543, opened 2026-05-27
chore(future): add completeOnce template

vladopajic:
this pr adds `completeOnce` template as replacement for code patterns where code checks if future is completed before completing it.

 ----

there are patterns, not many, that have following pattern:

```
if not fut.finished:
  fut.complete()
  # ... do some stuff here when future is being completed
```

which could be:

```
fut.completeOnceAnd:
  # ... do some stuff here when future is being completed
```

or something similar... but that can be implemented later.


gmelodie (approved):

richard-ramos (approved):
lgtm but the name is a bit weird :)

vladopajic:
> lgtm but the name is a bit weird :) 

would be better if original `complete` is already like that. 

when i was thinking about naming `sync.Once` from go was first thing that came to my mind.

PR #2547, opened 2026-05-27
fix(SwitchBuilder): initialization of services dependent of Rng

richard-ramos:
## Summary

This fixes `SwitchBuilder` paths that created RNG-dependent services before the builder had initialized its default RNG.

The PR:
- Defers AutoNAT v2 and hole-punching service construction until `buildSwitch`, after RNG initialization
- Avoids generating a throwaway default private key when the caller already supplied one
- Makes duplicate muxer registration checks use `doAssert`
- Gives the hole-punching reservation callback an explicit type

## Affected Areas

  - AutoNAT v2 and hole-punching builder initialization
- [x] Other
  - `SwitchBuilder` setup and validation

## Impact on Library Users

Users can configure AutoNAT v2 or hole punching before calling `withRng`, or rely on the builder's default RNG, without those services retaining a nil RNG.

The hole-punching reservation callback now has an explicit type, so incompatible callback signatures fail at compile time.

## Risk Assessment

Low. The change is limited to builder construction order and validation. It does not change the runtime behavior of the services after they are created.

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -89,9 +89,7 @@ type
     autonatV2ServerConfig: Opt[AutonatV2Config]
     autonatV2Config: Opt[AutonatV2ServiceConfig]
     autonatV2Client: AutonatV2Client
-    autonatV2Service: Opt[AutonatV2Service]
     holePunchingConfig: Opt[HolePunchingConfig]
-    hpService: Opt[HPService]
     natConfig: Opt[NATConfig]
     autotlsConfig: Opt[AutotlsConfig]
     circuitRelay: Opt[Relay]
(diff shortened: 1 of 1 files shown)


libp2p/builders.nim
```diff
@@ -434,8 +443,24 @@ proc buildSwitch(b: SwitchBuilder): Switch {.raises: [LPError].} =
...
+
+  b.holePunchingConfig.withValue(config):
+    let
+      autonatService = AutonatService.new(AutonatClient(), b.rng)
+      autoRelayService = AutoRelayService.new(
+        config.maxNumRelays, RelayClient.new(), config.onReservationHandler, b.rng
+      )
+      hpService = HPService.new(autonatService, autoRelayService)
+
+    b.hpService = Opt.some(hpService)
```
vladopajic: services should be created in `setupServices` proc.

note that in this case `b.hpService` and `b.autonatV2Service` fields should be removed as well.

vladopajic (approved):

gmelodie (approved):

PR #2548, opened 2026-05-27
chore: harden ZeroQueue edge cases

richard-ramos:
## Summary

Harden `ZeroQueue` boundary behavior and make queue length checks constant-time.

This PR:
- tracks the queued byte length as state instead of recalculating it by walking every chunk
- keeps the cached length updated when chunks are pushed, partially popped, fully popped, or cleared
- makes non-positive `popChunkSeq` requests return an empty sequence without consuming queued data
- lets zero-length `consumeTo` calls return immediately, including when the destination pointer is `nil`

## Affected Areas

- [x] Other

`libp2p/utils/zeroqueue.nim` and its focused tests.

## Impact on Library Users

No API migration is required. Existing `ZeroQueue.len` callers keep the same interface, with cheaper repeated length checks. Boundary calls such as `popChunkSeq(0)` and `popChunkSeq(-1)` now return an empty sequence instead of raising.

## Risk Assessment

Risk is limited to `ZeroQueue` consumers. The main behavior change is the cached byte count, so correctness depends on keeping it in sync with push, pop, and clear operations.

### DIFF
--- a/libp2p/utils/zeroqueue.nim
+++ b/libp2p/utils/zeroqueue.nim
@@ -22,22 +22,21 @@ type ZeroQueue* = object
   # byte sequences `seq[byte]` (called chunks). This type is useful for streaming or buffering 
   # scenarios where chunks of binary data are accumulated and consumed incrementally.
   chunks: Deque[Chunk]
-  length: int64
+  queuedBytes: int64
 
 proc clear*(q: var ZeroQueue) =
   q.chunks.clear()
-  q.length = 0
+  q.queuedBytes = 0
 
 proc isEmpty*(q: ZeroQueue): bool =
-  q.length == 0
+  q.chunks.len() == 0
 
 proc len*(q: ZeroQueue): int64 =
-  q.length
+  q.queuedBytes
 
 proc push*(q: var ZeroQueue, b: sink seq[byte]) =
-  let length = b.len
-  if length > 0:
-    q.length += length.int64
+  if b.len > 0:
+    q.queuedBytes += b.len.int64
     q.chunks.addLast(newChunk(b))
 
 proc popChunk(q: var ZeroQueue, count: int): Chunk {.inline.} =
@@ -46,9 +45,8 @@ proc popChunk(q: var ZeroQueue, count: int): Chunk {.inline.} =
 
   # first chunk has up to requested count elements,
   # queue will return this chunk (chunk might have less then requested)
-  let firstLen = first.len()
-  if firstLen <= count:
-    q.length -= firstLen.int64
+  if first.len() <= count:
+    q.queuedBytes -= first.len().int64
     return first
 
   # first chunk has more elements then requested count, 
@@ -57,12 +55,10 @@ proc popChunk(q: var ZeroQueue, count: int): Chunk {.inline.} =
   ret.size = ret.start + count
   first.start += count
   q.chunks.addFirst(first)
-  q.length -= count.int64
+  q.queuedBytes -= count.int64
   return ret
 
 proc consumeTo*(q: var ZeroQueue, pbytes: pointer, nbytes: int): int =
-  if nbytes == 0:
-    return 0
   var consumed = 0
   while consumed < nbytes and not q.isEmpty():
     let chunk = q.popChunk(nbytes - consumed)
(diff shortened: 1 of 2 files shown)


libp2p/utils/zeroqueue.nim
```diff
@@ -22,29 +22,33 @@ type ZeroQueue* = object
...
 proc len*(q: ZeroQueue): int64 =
-  var l: int64
-  for b in q.chunks.items():
-    l += b.len()
-  return l
+  q.length
 
 proc push*(q: var ZeroQueue, b: sink seq[byte]) =
-  if b.len > 0:
+  let length = b.len
```
vladopajic: `let length = b.len`  is unnecessary

`.len` should always be used as source of truth for all container types. assigning it to variable should only be done when there is strong reason for doing so.
this pattern actually complicates code.

libp2p/utils/zeroqueue.nim
```diff
@@ -22,29 +22,33 @@ type ZeroQueue* = object
...
     q.chunks.addLast(newChunk(b))
 
 proc popChunk(q: var ZeroQueue, count: int): Chunk {.inline.} =
+  doAssert(count > 0, "count must be positive integer")
   var first = q.chunks.popFirst()
 
   # first chunk has up to requested count elements,
   # queue will return this chunk (chunk might have less then requested)
-  if first.len() <= count:
+  let firstLen = first.len()
```
vladopajic: `let firstLen = first.len()` is unnecessary

libp2p/utils/zeroqueue.nim
```diff
@@ -53,9 +57,12 @@ proc popChunk(q: var ZeroQueue, count: int): Chunk {.inline.} =
   ret.size = ret.start + count
   first.start += count
   q.chunks.addFirst(first)
+  q.length -= count.int64
   return ret
 
 proc consumeTo*(q: var ZeroQueue, pbytes: pointer, nbytes: int): int =
+  if nbytes == 0:
+    return 0
```
vladopajic: this early return is unnecessary as `while consumed < nbytes` already has this purpose to guard.
assembler code difference for these two is neglectable.

libp2p/utils/zeroqueue.nim
```diff
@@ -67,12 +74,16 @@ proc consumeTo*(q: var ZeroQueue, pbytes: pointer, nbytes: int): int =
...
 proc popChunkSeq*(q: var ZeroQueue, count: int): seq[byte] =
-  if q.isEmpty:
+  if count <= 0 or q.isEmpty:
     return @[]
 
   let chunk = q.popChunk(count)
-  var dest = newSeqUninit[byte](chunk.len())
+  let chunkLen = chunk.len()
+  if chunkLen == 0:
+    return @[]
```
vladopajic: this is never going to be true. zero push should be ignored in the first place.

libp2p/utils/zeroqueue.nim
```diff
@@ -67,12 +74,16 @@ proc consumeTo*(q: var ZeroQueue, pbytes: pointer, nbytes: int): int =
   return consumed
 
 proc popChunkSeq*(q: var ZeroQueue, count: int): seq[byte] =
-  if q.isEmpty:
+  if count <= 0 or q.isEmpty:
     return @[]
 
   let chunk = q.popChunk(count)
-  var dest = newSeqUninit[byte](chunk.len())
+  let chunkLen = chunk.len()
```
vladopajic: `let chunkLen = chunk.len()` as above

libp2p/utils/zeroqueue.nim
```diff
@@ -22,29 +22,33 @@ type ZeroQueue* = object
...
 
 proc push*(q: var ZeroQueue, b: sink seq[byte]) =
-  if b.len > 0:
+  let length = b.len
+  if length > 0:
+    q.length += length.int64
     q.chunks.addLast(newChunk(b))
 
 proc popChunk(q: var ZeroQueue, count: int): Chunk {.inline.} =
+  doAssert(count > 0, "count must be positive integer")
```
vladopajic, line 43: this assert seems unnecessary.

`popChunk` is internal method that will never be called for `count<=0` case. if such case do happen better to guard it with unit tests.

tests/libp2p/utils/test_zeroqueue.nim
```diff
@@ -23,6 +23,9 @@ suite "ZeroQueue":
     check q.len() == 5
     check not q.isEmpty()
 
+    check q.popChunkSeq(0) == newSeq[byte]()
+    check q.popChunkSeq(-1) == newSeq[byte]()
```
vladopajic: this can be ether expected or counter expected behavior. in some cases it might be better to get errors if calling with `-1` then not to get any. later can theoretically cause errors to never get noticed - unless there are unit tests that assert higher lvl logic. 

i don't really have strong opinion in this case.

vladopajic (approved):
length related change is okay. the reason why it was not implemented in the first place is because `len` was never used on the hot path, while `isEmpty` was used - but it was O(1) already. `len`was there just for unit test purposes. initially i preferred simplicity.

libp2p/utils/zeroqueue.nim
vladopajic: https://github.com/nim-lang/Nim/blob/version-2-2/lib/pure/collections/deques.nim#L97

vladopajic: no one should really optimize that

richard-ramos:
Yeah... i actually dont feel strongly confident in the need for this PR, honestly :thinking:

gmelodie (approved):

PR #2551, opened 2026-05-27
chore: remove explicit inline pragmas

richard-ramos:
## Summary

Removes explicit `{.inline.}` pragmas from existing Nim declarations across the codebase.

This follows the Status Nim style guide, which says to avoid explicit inline functions because they add clutter and leave inlining decisions to the compiler/LTO where possible.

## Affected Areas

- [x] Other

Code style cleanup across multiformats, crypto helpers, protobuf helpers, pubsub, Kademlia, service discovery, transports, stream utilities, and cbind helper code.

## Impact on Library Users

No API or network behavior change. This only removes explicit compiler hints from existing declarations.

## Risk Assessment

Low. The change is mechanical; remaining pragmas such as `raises` and `async` are preserved. Any risk is limited to compiler codegen choices after removing manual inline hints.

## References

- [Status Nim style guide: inline functions](https://status-im.github.io/nim-style-guide/language.inline.html)

### DIFF
--- a/cbind/libp2p_thread/inter_thread_communication/requests/libp2p_kademlia_requests.nim
+++ b/cbind/libp2p_thread/inter_thread_communication/requests/libp2p_kademlia_requests.nim
@@ -44,9 +44,7 @@ type RandomRecordsResult* = object
   records*: ptr Libp2pExtendedPeerRecord
   recordsLen*: csize_t
 
-proc allocServiceInfoArrayFromSeq(
-    services: seq[ServiceInfo]
-): ptr Libp2pServiceInfo {.inline.} =
+proc allocServiceInfoArrayFromSeq(services: seq[ServiceInfo]): ptr Libp2pServiceInfo =
   if services.len == 0:
     return nil
   let ret =
--- a/libp2p/cid.nim
+++ b/libp2p/cid.nim
@@ -270,11 +270,11 @@ proc repr*(cid: Cid): string =
   s.add($(cid.mhash()))
   s
 
-proc write*(vb: var VBuffer, cid: Cid) {.inline.} =
+proc write*(vb: var VBuffer, cid: Cid) =
   ## Write CID value ``cid`` to buffer ``vb``.
   vb.writeArray(cid.data.buffer)
 
-proc hash*(cid: Cid): Hash {.inline.} =
+proc hash*(cid: Cid): Hash =
   hash(cid.data.buffer)
 
 proc `$`*(cid: Cid): string =
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -208,7 +208,7 @@ proc new*(
     scoring: scoring,
   )
 
-proc connCount*(c: ConnManager, peerId: PeerId): int {.inline.} =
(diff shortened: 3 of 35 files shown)


cbind/libp2p_thread/inter_thread_communication/requests/libp2p_kademlia_requests.nim
```diff
@@ -44,9 +44,7 @@ type RandomRecordsResult* = object
   records*: ptr Libp2pExtendedPeerRecord
```
vladopajic, line 44: i wonder if `avoid` means to be cautious or much stricter - do not use it!

i also wonder why nim language use it in their source: https://github.com/nim-lang/Nim/blob/version-2-2/lib/pure/collections/deques.nim#L97

vladopajic: it certainly is the case that even with notes form style guide some cases do make sense to inline - if you want to. and i could see cases, from this pr, were inline was pointless and this is why style guide made that statement. 

i feel that style guide needs to be better with explaining because this is not binary case.

vladopajic (approved):

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_kademlia_requests.nim
vladopajic: for example, inlining `allocServiceInfoArrayFromSeq` doesn't make sense  while ```proc `-`(x: uint32): uint32 {.inline.}``` makes.

richard-ramos: cc: @tersec and @arnetheduck for their thoughts . Some clarity on this is greatly appreciated

arnetheduck: the effect of `inline` is that it places a copy of the function in each C source file and marks it `static` on the C side - this allows the C compiler to be more aggressive about inlining it since it knows it won't be used outside of that translation unit. This has several consequences:

* it doesn't make any difference to make private nim functions `inline`
* for public nim functions, compile times increase because there are many copies of the same code (one in each file that uses the function)
* it makes inlining _possible_ without LTO
* it's still not a guarantee that the function will be inlined, ie the C compiler can still decide to not inline it (this depends on how and where the function is called)

However, when you use LTO, _all_ functions are effectively `static` and can be inlined but because the same function is called from multiple places, the C compiler might be less aggressive about inlining in some edge cases where the function is large. 

Thus, it's a tradeoff between increasing compile times for everyone at the expense of _sometimes_ missing an inlining -> use `inline` in cases where you've measured an actual difference _with LTO enabled_, ie in hotspots, instead of spreading it randomly throughout the codebase.

The std lib also does more inlining than needed because nim code is not compiled with LTO _by default_ - we do lto everywhere that matters but if you don't, you need to use `inline` as a workaround.

gmelodie (approved):

PR #2553, opened 2026-05-27
fix(ping): keep inbound ping streams reusable

richard-ramos:
## Summary

The libp2p ping spec says the dialing peer may send another 32-byte payload on the same stream and that the listening peer should loop and echo the next payload. This PR updates the Nim ping listener to follow that behavior by keeping inbound ping streams open for repeated payloads until the peer closes the stream.

It also:
- treats EOF as normal ping stream completion instead of logging it as a generic handler exception
- asserts that `Ping.new` receives a non-nil RNG
- removes unused imports from the ping protocol module
- gets rid of a `CatchableError`

## Affected Areas

- [x] Protocol Logic
  - Updates `/ipfs/ping/1.0.0` listener behavior and ping callback typing.

## Impact on Library Users

`PingHandler` is now typed as an async non-raising callback:

```nim
proc(peer: PeerId): Future[void] {.async: (raises: []), gcsafe.}
```

Callers passing a ping callback may need to make the callback match that stricter async signature.

## Risk Assessment

Low risk. The listener now keeps the negotiated ping stream alive for repeated payloads, matching the spec. The callback type change is a small API break for users that pass custom ping handlers.

## References

- [libp2p ping spec](https://github.com/libp2p/specs/blob/master/ping/ping.md)

### DIFF
--- a/libp2p/protocols/ping.nim
+++ b/libp2p/protocols/ping.nim
@@ -7,11 +7,7 @@
 
 import chronos, chronicles
 import
-  ../stream/connection,
-  ../peerid,
-  ../crypto/crypto,
-  ../protocols/protocol,
-  ../errors
+  ../stream/connection, ../peerid, ../crypto/crypto, ../protocols/protocol, ../errors
 
 export chronicles, rng, connection
 
@@ -49,9 +45,6 @@ method init*(p: Ping) =
         await stream.write(@buf)
         if not isNil(p.pingHandler):
           await p.pingHandler(stream.peerId)
-    except CancelledError as exc:
-      trace "cancelled ping handler"
-      raise exc
     except LPStreamEOFError as exc:
       trace "ping stream closed", description = exc.msg, stream
     except LPStreamError as exc:
--- a/libp2p/protocols/ping.nim
+++ b/libp2p/protocols/ping.nim
@@ -7,11 +7,9 @@
 
 import chronos, chronicles
 import
-  ../peerinfo,
   ../stream/connection,
   ../peerid,
   ../crypto/crypto,
-  ../multiaddress,
   ../protocols/protocol,
   ../errors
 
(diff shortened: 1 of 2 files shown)


gmelodie (approved):

libp2p/protocols/ping.nim
```diff
@@ -44,16 +39,19 @@ method init*(p: Ping) =
...
-        await p.pingHandler(stream.peerId)
-    except CancelledError as exc:
+      while true:
+        await stream.readExactly(addr buf[0], PingSize)
+        trace "echoing ping", stream, pingData = @buf
+        await stream.write(@buf)
+        if not isNil(p.pingHandler):
+          await p.pingHandler(stream.peerId)
+    except CancelledError as e:
       trace "cancelled ping handler"
```
vladopajic: or just avoid catching it

libp2p/protocols/ping.nim
```diff
@@ -44,16 +39,19 @@ method init*(p: Ping) =
     try:
       trace "handling ping", stream
       var buf: array[PingSize, byte]
-      await stream.readExactly(addr buf[0], PingSize)
-      trace "echoing ping", stream, pingData = @buf
-      await stream.write(@buf)
-      if not isNil(p.pingHandler):
-        await p.pingHandler(stream.peerId)
-    except CancelledError as exc:
+      while true:
```
vladopajic, line 42: so protocol changed from single ping to many pings?

vladopajic (approved):

libp2p/protocols/ping.nim
vladopajic: aha i see description ok!

richard-ramos: yeah, we were actually not following the specs!

Issue #2084, opened 2026-02-05
kad: use min heap in randomRecords


richard-ramos:
> queue actually needs to be min heap here.
> 
> because each iteration of find node, and findClosestPeerIds will return some values that are ordered (but only for that result), and they are not order when we look into all results in queue 

 _Originally posted by @vladopajic in [#2083](https://github.com/vacp2p/nim-libp2p/pull/2083/changes/BASE..e4e2bbc3f424255e9ef721f6dbe21f68837f6b76#r2769210339)_

PR #2559, opened 2026-05-28
refactor(kad): rewrite randomRecords around new lookOnce primitive

gmelodie:
## Summary

  `randomRecords` previously called `findNode` with a shared
  `AsyncQueue[(PeerId, Opt[Message])]` so it could consume discovered peers
  while the lookup was still running. That hook leaked lookup internals onto
  `findNode`'s public signature, and the simplifications during review left
  `randomRecords` waiting for `findNode` to finish entirely before issuing a
  single `getValue` — undoing the concurrency the queue was added for.

  This PR drops the queue and rebuilds `randomRecords` around a new lookup
  primitive.

  ## Changes

  ### `libp2p/protocols/kademlia/find.nim`

  * Extract `lookOnce` from `iterativeLookup`. Runs one round of the lookup
    against a caller-owned `LookupState`; returns `false` when the shortlist
    is exhausted.
  * `iterativeLookup` is now a thin loop around `lookOnce`.
  * Drop the `queue`/`PeerDistance`/`newPeerDistanceHeap` API from
  * `LookupState` becomes a `ref object` so it can be passed across an
    `await` boundary as a parameter.

  ### `libp2p/protocols/service_discovery/random_find.nim`

  * Drive the lookup directly via `lookOnce` instead of `findNode`.
  * Fire `dispatchGetVal` on each newly-responded peer as the lookup
    progresses, so getValue overlaps with the next lookup round.
  * Stop on `closestAvailableStop`, matching `findNode`'s prior termination.

  ### `tests/libp2p/kademlia/mock_kademlia.nim`,
  `tests/libp2p/kademlia/utils.nim`

  Updated to match the new `findNode` and `LookupState` signatures.

  ## Impact on Library Users

  `KadDHT.findNode`'s optional `queue` parameter is gone. The only in-tree
  caller passing one (`service_discovery/random_find`) no longer needs it.
  External callers using a custom queue must migrate to driving `lookOnce`
  themselves.

  `LookupState` is now `ref object`. `ReplyHandler` and `updateShortlist`
  drop their `var` qualifier — mutations propagate through the reference.

  ## References

  * Follow-up to #2083
  * Related: #2084


tests/libp2p/kademlia/test_find.nim
```diff
@@ -309,3 +311,30 @@ suite "KadDHT Find":
     check:
       responsiveKad.switch.peerInfo.peerId in peerIds
       mockKad.handleFindNodeCalls == retries + 1 # (initial call + retries)
+
+suite "KadDHT PeerDistance heap":
```
vladopajic: what does this test assert?

heap queue should have already been tested in it's unit test?

libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -16,31 +17,31 @@ proc makeFireCallback(event: AsyncEvent): CallbackFunc =
...
+    queue: AsyncHeapQueue[PeerDistance], findNodeFut: Future[seq[PeerId]]
 ): Future[Opt[PeerId]] {.async: (raises: [CancelledError]).} =
-  ## Pop the next peer from the queue, blocking until one arrives or
+  ## Pop the closest peer from the heap, blocking until one arrives or
   ## findNodeFut finishes. Returns none when no more peers will come.
   if not queue.empty():
-    let (peerId, _) = await queue.popFirst()
-    return Opt.some(peerId)
+    let entry = await queue.pop()
+    return Opt.some(entry.peerId)
```
vladopajic: is it necessary to `await` when code already knows that queue is not empty?

libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -54,14 +55,12 @@ proc randomRecords(
...
   let peers = disco.rtable.findClosestPeerIds(randomKey, disco.config.replication)
   for peer in peers:
-    let addRes = catch:
-      queue.addFirstNoWait((peer, Opt.none(Message)))
-    if addRes.isErr:
-      error "cannot enqueue peer", error = addRes.error.msg
+    let distance = xorDistance(peer, randomKey, disco.rtable.config.hasher)
+    queue.push(PeerDistance(peerId: peer, distance: distance))
 
   let findNodeFut = disco.findNode(randomKey, queue)
```
vladopajic: why code does not await here? it would simplify code by a lot all that complex logic in `nextPeer` will be unnecessary + `newAsyncHeapQueue` would be unnecessary - just use regular heap

vladopajic (approved):
slices complexity away!

libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -54,47 +22,33 @@ proc randomRecords(
...
-    let addRes = catch:
-      queue.addFirstNoWait((peer, Opt.none(Message)))
-    if addRes.isErr:
-      error "cannot enqueue peer", error = addRes.error.msg
+    let distance = xorDistance(peer, randomKey, disco.rtable.config.hasher)
+    queue[].push(PeerDistance(peerId: peer, distance: distance))
 
-  let findNodeFut = disco.findNode(randomKey, queue)
+  # CancelledError propagates; findNode reports any other failure via logs.
+  discard await disco.findNode(randomKey, queue)
```
richard-ramos: This `await` will slow down things, since now you need to wait for  findNode to be done before doing a getValue Call, compared to previously in which peers were consumed while findNodFut was still running

vladopajic: ^ before we answer that

- is this heap necessary at all? the initially hypothesis to use heap could also be wrong.
- while loop does not really care if goes to `dispatchGetVal` to farthest nodes first - which is case before this pr?
- in previous code, why code doesn't `dispatchGetVal` as batch requests?
- `findNode` has queue as input parameter and returns peer as result - why?
- does find random loop has some stop criteria like number of peers? 


the more i look into this code the more i get confused

vladopajic: > is this heap necessary at all?

now i realize that when i made original comment i was thinking that queue from arguments is source of peers where we need to select next peers for search. and since we need to select closer peers then heap is needed. now i realize that `findNode` select peers from `state` and not form queue.

vladopajic: the `queue` needs to answer _why_ it is there in the first place. 
in `findNode` why is `ignoreReply` called like that - why it adds elements to this queue?

vladopajic: for regular find operation `queue` is 100% unnecessary and therefore it shouldn't belong there

vladopajic: for `randomRecords` it would be better not to use `findNode`, but to use internal functions that build `findNode` inside while loop:
 
 pseudo code:
 
 ```
state = LookupState.new()
while {there better be some stop condition explicitly defined or see stop below} 
  state.lookOnce(target) # target is random key
  if no new results stop 
  
  get peers state then query get values
 ```

libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -1,47 +1,25 @@
...
-    Opt.none(PeerId)
+proc cancelPending(
+    futs: seq[Future[Result[Message, string]].Raising([CancelledError])]
+) {.async: (raises: []).} =
+  var pending: seq[FutureBase]
+  for fut in futs:
+    if not fut.finished:
+      pending.add(fut)
+  if pending.len > 0:
+    await noCancel allFutures(pending.mapIt(it.cancelAndWait()))
```
vladopajic: it is always safe to cancel all futures in no mater the state:

```
await noCancel allFutures(futs.mapIt(it.cancelAndWait()))
```

libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -54,47 +32,47 @@ proc randomRecords(
...
-      let reply = msgOpt.valueOr:
-        continue
+      # Issue getValue requests as soon as peers respond, so they run in
+      # parallel with the next lookup round.
+      for peerId, status in state.responded:
+        if status == RespondedStatus.Success and peerId notin queried:
+          queried.incl(peerId)
+          getValFuts.add(disco.dispatchGetVal(peerId, peerId.toKey()))
+  except CancelledError as e:
+    await cancelPending(getValFuts)
```
vladopajic, line 53: `await cancelPending(getValFuts)` this is called in defer at the begging you can ignore catching this error and handling it all places here in order to cancel those futures. it should be safe to always cancel them in defer no matter the state.

libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -54,47 +32,47 @@ proc randomRecords(
...
-      let record = reply.record.valueOr:
-        continue
+  var buffers: seq[seq[byte]]
+  for fut in getValFuts:
+    let res =
+      try:
+        await fut
+      except CancelledError as e:
+        await cancelPending(getValFuts)
+        raise e
```
vladopajic, line 63:
```
await allFututes(getValFuts)
for fut in getValFuts:
  # futue is completed just get value
```
perhaps?

vladopajic (approved):
makes more sense

richard-ramos (approved):
With latest changes this does look way better than before, but I still do have mixed feelings about this PR because:

- In current master: node knew some peers locally, so it could start asking records with GET_VALUE right away
- In this PR: we first ask these peers with FIND_NODE, then ask the nodes that respond successfully for GET_VALUE

I guess this is probably ok, I checked the specs and did not find anything that says we cannot do this, but it still means the first GET_VALUE calls will be dispatched a bit later than before. Was this intentional?

PR #2560, opened 2026-05-28
chore(kad): limits

### LINKED ISSUE #1851: Kad DHT limits
`ProviderManager` already has capacity limits (`DefaultProviderRecordCapacity`, `DefaultProvidedKeyCapacity`) with eviction policies. The following components are missing equivalent limits:

- [ ] `provider.nim` — Providers per Key
- `knownKeys: Table[Key, HashSet[Provider]]` is unbounded.
- A single popular key can accumulate unlimited providers.
- **Recommendation:** Add a `maxProvidersPerKey` config (e.g., default `20`, matching replication factor).

- [ ] `find.nim` — Lookup Shortlist
- The shortlist (`Table[PeerId, XorDistance]`) and visited set grow unbounded during a lookup.
- **Recommendation:** Add a cap (e.g., `maxShortlistSize`, default `replication * 2 = 40`) and drop farther peers when exceeded.

- [ ] `get.nim` — `ReceivedTable`
- `ReceivedTable: TableRef[PeerId, Opt[EntryRecord]]` per query has no size cap.
- **Recommendation:** Add a max entries limit tied to replication.

- [ ] `put.nim` / `get.nim` — Value Size
- Individual record values (`seq[byte]`) have no size limit.
- **Recommendation:** Add a `maxValueSize` config (e.g., `8KB` default) enforced on store and on incoming messages.

- [ ] `kademlia.nim` — Concurrent RPCs
- No global limit on concurrent in-flight RPC calls.
- **Recommendation:** Add `maxConcurrentRpcs` or reuse a semaphore shared across `find`, `get`, and `put` operations.

## Existing Configs (Reference)
Defined in `types.nim`:

- `DefaultProviderRecordCapacity = 500`
- `DefaultProvidedKeyCapacity = 500`
- `MaxMsgSize = 4096`
- `DefaultAlpha = 10` (per-operation concurrency)
- `DefaultReplication = 20`

---

## Implementation Note
All new limits should be added to `KadDHTConfig` with sensible defaults and exposed consistently with existing configuration parameters.

gmelodie:
## Summary

Adds configurable upper bounds to several Kademlia DHT data structures that
previously could grow without limit (#1851).

### Bounded Structures

* `provider.nim`

  * Per-key provider sets (`knownKeys[k]`) are now bounded by
    `maxProvidersPerKey` regardless of `providerRejection`.
  * With `providerRejection = true`, the receiver still replies with a status.
  * With `providerRejection = false`, over-cap advertisements are silently
    dropped.

* `find.nim`

  * `LookupState.shortlist` is now bounded by `maxShortlistSize`.
  * When at capacity:

    * the farthest unresponded peer is evicted only if the incoming peer is
      closer
    * responded peers are preserved so their results remain usable

* `get.nim`

  * The per-query `ReceivedTable` stops accepting new peers after reaching
    `maxReceivedSize`.

* `put.nim` / `get.nim`

  * Both the `putValue` API and inbound PUT/GET handlers now reject records
    whose value exceeds `maxValueSize`.

* `kademlia.nim`

  * All outbound RPC dispatchers now share a global
    `AsyncSemaphore` bounded by `maxConcurrentRpcs`.
  * Enforcement is implemented through a new `withRpcSlot` template.

### Dispatcher Changes

The following procedures were changed from `Switch` receivers to `KadDHT`
receivers so they can access the shared semaphore:

* `dispatchPutVal`
* `dispatchAddProvider`

## Affected Areas

* [x] Protocol Logic — Kademlia DHT

  * provider records
  * iterative lookup
  * GET / PUT / ADD_PROVIDER RPCs

This is a pure addition of bounded defaults:

* no wire-format changes
* no new required configuration

All existing tests in `tests/libp2p/kademlia/` pass (125 total).

One test was updated to reflect the new always-enforced provider cap:

```text
providerRejection=false: receiver never sends rejection replies
```

### API Changes

`KadDHTConfig` gains five new optional configuration fields:

* `maxShortlistSize`
* `maxReceivedSize`
* `maxValueSize`
* `maxConcurrentRpcs`
* `maxProvidersPerKey`

  * existing field
  * default changed from `Opt.none` to `Opt.some(20)`

All new fields have defaults, so callers using:

```nim
KadDHTConfig.new()
```

require no migration.

### Behavioral Changes

With default configuration, nodes now enforce:

* maximum of 20 providers per key
* maximum shortlist size of 40 peers
* maximum `ReceivedTable` size of 20 entries
* maximum record value size of 8 KiB
* maximum of 100 concurrent in-flight RPCs

Additionally:

* `dispatchPutVal`
* `dispatchAddProvider`

now have source-incompatible signatures for any external caller invoking them
directly outside the `kademlia` module (none exist in-tree).

## Risk Assessment

* The new defaults are intentionally generous (at or above replication factor),
  so well-behaved nodes should not hit these limits during normal operation.

* Over-cap inbound RPCs are:

  * silently dropped, or
  * rejected with a status reply when `providerRejection` is enabled

  No on-wire error-code changes are introduced.

* The shared RPC semaphore introduces minor self-throttling during bursty
  lookup activity.

  With:

```text
maxConcurrentRpcs = 100
alpha = 10
```

roughly 10 concurrent lookup operations can proceed without contention.

## References

* Closes #1851

### DIFF
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -50,9 +50,7 @@ proc getFarthest(
   else:
     Opt.none((PeerId, XorDistance))
 
-proc tryEvictFarthest(
-    state: var LookupState, newDist: XorDistance
-): bool {.raises: [].} =
+proc tryEvictFarthest(state: LookupState, newDist: XorDistance): bool {.raises: [].} =
   ## Drop the worst (farthest) peer from the shortlist if it is farther than
   ## ``newDist``. Considers all peers — including ones that already responded —
   ## because the iterative lookup needs the closer candidate to make progress.
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -34,6 +34,22 @@ type ReplyHandler* = proc(
 
 type StopCond* = proc(state: LookupState): bool {.raises: [], gcsafe.}
 
+proc getFarthest(
+    t: Table[PeerId, XorDistance]
+): Opt[(PeerId, XorDistance)] {.raises: [].} =
+  var worstPid: PeerId
+  var worstDist: XorDistance
+  var found = false
+  for pid, d in t.pairs():
+    if not found or worstDist < d:
+      worstPid = pid
+      worstDist = d
+      found = true
+  if found:
+    Opt.some((worstPid, worstDist))
+  else:
+    Opt.none((PeerId, XorDistance))
+
 proc tryEvictFarthest(
     state: var LookupState, newDist: XorDistance
 ): bool {.raises: [].} =
@@ -42,19 +58,13 @@ proc tryEvictFarthest(
   ## because the iterative lookup needs the closer candidate to make progress.
   ## A responded peer's contribution is already merged into the shortlist, so
   ## evicting it costs nothing beyond bookkeeping.
-  var worstPid: PeerId
-  var worstDist: XorDistance
-  var found = false
-  for pid, d in state.shortlist.pairs():
-    if not found or worstDist < d:
-      worstPid = pid
-      worstDist = d
-      found = true
-  if not found or newDist >= worstDist:
+  let (pid, dist) = state.shortlist.getFarthest().valueOr:
     return false
-  state.shortlist.del(worstPid)
-  state.attempts.del(worstPid)
-  state.responded.del(worstPid)
+  if newDist >= dist:
+    return false
+  state.shortlist.del(pid)
+  state.attempts.del(pid)
+  state.responded.del(pid)
   return true
 
 proc updateShortlist*(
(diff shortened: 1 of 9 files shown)


libp2p/protocols/kademlia/types.nim
```diff
@@ -363,10 +399,14 @@ proc new*(
     hideConnectionStatus: bool = true,
     disableBootstrapping: bool = false,
     providerRejection: bool = false,
-    maxProvidersPerKey: Opt[int] = Opt.none(int),
+    limits: KadDHTLimits = KadDHTLimits.new(),
```
richard-ramos: I dont think we should have `limits` use `KadDHTLimits.new` as default.

Those defaults (like `DefaultMaxShortlistSize` and `DefaultReceivedSize`), cause these are defined based on the `DefaultReplicationSize`, i.e.:
```
maxShortlistSize: DefaultMaxShortlistSize              # DefaultReplication * 2 = 40
maxReceivedSize: DefaultReceivedSize                   # DefaultReplication = 20
```
meaning that if you only set some of the values, the others wont be updated accordingly and you could end up with a mismatched configuration,

for example, if you were to set a replication of 50 and a quorum of 30 here, and you dont specify the limits, you will end up in astrange situation, just to give an example:

with a `replication` of 50, it means we should try to work with 50 closest peers, the `maxShortlistSize` has a default of 40, thus, the shortlist will not have enough candidates to meet the requested replication, and the `maxProvidersPerKey beign 20 by default, means that you could store providers way below the replication target.

And for example if you were to setup a `quorum` of 30 and `maxReceivedSize` of 20 (the default), `getValue` will never be able to get 30 entries, so it will fail due to not meeting quorum.

So, this makes me think that  perhaps the limits should perhaps look like this:
```
    limits: Opt[KadDHTLimits] = Opt.none(KadDHTLimits),
```

and then inside `new`, if `limits.isNone`, you would derivate the values based on the replication, so the defaults at least make sense

libp2p/protocols/kademlia/types.nim
```diff
@@ -32,6 +32,17 @@ const
 
   MaxMsgSize* = 4096
 
+  DefaultMaxProvidersPerKey* = 20 ## upper bound on providers stored per key
+  DefaultMaxShortlistSize* = DefaultReplication * 2
+    ## upper bound on iterative-lookup shortlist
+  DefaultMaxReceivedSize* = DefaultReplication
+    ## upper bound on per-query ``ReceivedTable``
+  DefaultMaxValueSize* = MaxMsgSize
```
richard-ramos: Hm.. not sure about this either, cause remember that the MaxMsgSize is used to limit the protobuffer request and that includes some overhead due to the other fields, and also due to the serialization. This MaxValueSize should be smaller than the MaxMsgSize, but not sure what a proper default could be.... maybe 3kb instead of 4kb (the MaxMsgSize default)? that would leave you 1kb left for the data from other fields and serialization overhead

libp2p/protocols/kademlia/types.nim
```diff
@@ -363,10 +399,14 @@ proc new*(
...
-    maxProvidersPerKey: Opt[int] = Opt.none(int),
+    limits: KadDHTLimits = KadDHTLimits.new(),
 ): K {.raises: [].} =
-  doAssert maxProvidersPerKey.isNone or maxProvidersPerKey.get() > 0,
+  doAssert limits.maxProvidersPerKey.isNone or limits.maxProvidersPerKey.get() > 0,
     "maxProvidersPerKey must be > 0; use Opt.none(int) for unlimited"
+  doAssert limits.maxShortlistSize > 0, "maxShortlistSize must be > 0"
+  doAssert limits.maxReceivedSize > 0, "maxReceivedSize must be > 0"
+  doAssert limits.maxValueSize > 0, "maxValueSize must be > 0"
+  doAssert limits.maxConcurrentRpcs > 0, "maxConcurrentRpcs must be > 0"
```
richard-ramos: I wonder if we should also enfoce these:
- `maxShortlistSize` >= `replication`
- `maxReceivedSize` >= `quorum`
thoughts?

libp2p/protocols/kademlia/find.nim
```diff
@@ -34,20 +34,44 @@ type ReplyHandler* = proc(
...
 
+proc tryEvictFarthestUnresponded(
+    state: var LookupState, newDist: XorDistance
+): bool {.raises: [].} =
+  var worstPid: PeerId
+  var worstDist: XorDistance
+  var found = false
+  for pid, d in state.shortlist.pairs():
+    if state.responded.contains(pid):
+      continue
```
richard-ramos: I am also not sure about this either.

Wouldnt you end up dropping better peers when the shortlist.len >= maxShortListSize and peers inside it all responded?

Because then if you find a new peer, and that peer has a closer xordistance, you are gonna ignore it even if that peer is better than what we have.

It makes me think that we should instead drop the worse peers from the list (and remove them from responded and attempt)

richard-ramos (approved):

libp2p/protocols/kademlia/find.nim
```diff
@@ -34,20 +34,48 @@ type ReplyHandler* = proc(
...
+  var worstPid: PeerId
+  var worstDist: XorDistance
+  var found = false
+  for pid, d in state.shortlist.pairs():
+    if not found or worstDist < d:
+      worstPid = pid
+      worstDist = d
+      found = true
+  if not found or newDist >= worstDist:
+    return false
```
vladopajic, line 62: let (pid, found) = state.shortlist.getFarthest()

vladopajic (approved):

libp2p/protocols/kademlia/provider.nim
```diff
@@ -103,60 +105,50 @@ proc addProviderRecord(pm: ProviderManager, record: ProviderRecord) =
...
-  let encoded = msg.encode(hideConnectionStatus)
-  kad_messages_sent.inc(labelValues = [$MessageType.addProvider])
-  kad_message_bytes_sent.inc(
-    encoded.buffer.len.int64, labelValues = [$MessageType.addProvider]
-  )
-  let writeRes = catch:
-    await stream.writeLp(encoded.buffer)
-  if writeRes.isErr:
-    return err(writeRes.error.msg)
+  withRpcSlot(kad):
```
vladopajic: there is no need to indent code, template can just lock and defer release

vladopajic (changes requested):

libp2p/protocols/kademlia/types.nim
```diff
@@ -401,3 +454,17 @@ type KadDHT* = ref object of LPProtocol
...
+  ## Acquire one ``rpcSem`` slot for the duration of ``body``. The slot is
+  ## released whether ``body`` completes normally, raises, or is cancelled.
+  await kad.rpcSem.acquire()
+  try:
+    body
+  finally:
+    try:
+      kad.rpcSem.release()
+    except AsyncSemaphoreError:
+      raiseAssert "rpcSem released without acquire"
```
vladopajic: must not use `body`, release in defer

vladopajic (approved):

PR #2567, opened 2026-05-28
fix(quic): harden certificate parsing and peer verification

richard-ramos:
## Summary

This PR tightens libp2p TLS certificate validation and QUIC peer identity checks.

- validates certificate inputs and libp2p extension data before OpenSSL/ASN.1 parsing paths use them
- preserves 24-hour UTC values when formatting and parsing certificate validity times
- verifies QUIC peer certificates against the expected `PeerId` when one is provided
- forwards the optional `PeerId` through the base `Transport.dial(address, peerId)` wrapper so transport-level checks receive it

## Affected Areas

- [x] Transports
  QUIC certificate verification and base transport dial peer-ID forwarding.
- [x] Other
  TLS certificate parsing and generation helpers.

## Impact on Library Users

`P2pCertificate.verify` now requires an expected `PeerId`. Callers that only need structural certificate validation should use `verifiedIdentityKey()`.

QUIC dials that pass an expected peer ID now fail when the remote certificate belongs to a different peer. Unknown-peer dial paths still validate the certificate structure, but cannot bind it to a peer ID until one is known.

The `Transport.dial(address, peerId)` convenience wrapper now preserves the provided peer ID instead of dropping it.

## Risk Assessment

The main behavior change is stricter QUIC certificate acceptance when an expected peer ID is available.

The certificate parser now rejects empty or malformed certificate/extension data earlier.

### DIFF
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -282,23 +282,36 @@ type QuicTransport* = ref object of Transport
   rng: Rng
   certGenerator: CertGenerator
 
-proc makeCertificateVerifier(): CertificateVerifier =
-  proc certificateVerifier(serverName: string, certificatesDer: seq[seq[byte]]): bool =
-    if certificatesDer.len != 1:
-      trace "CertificateVerifier: expected one certificate in the chain",
-        cert_count = certificatesDer.len
+proc verifyCertificates(
+    certificatesDer: seq[seq[byte]], expectedPeerId: Opt[PeerId] = Opt.none(PeerId)
+): bool =
+  if certificatesDer.len != 1:
+    trace "CertificateVerifier: expected one certificate in the chain",
+      cert_count = certificatesDer.len
+    return false
+
+  let cert =
+    try:
+      parse(certificatesDer[0])
+    except CertificateParsingError as e:
+      trace "CertificateVerifier: failed to parse certificate", msg = e.msg
       return false
 
-    let cert =
-      try:
-        parse(certificatesDer[0])
-      except CertificateParsingError as e:
-        trace "CertificateVerifier: failed to parse certificate", msg = e.msg
-        return false
+  if expectedPeerId.isSome:
+    let peerId = expectedPeerId.get()
+    if not cert.verify(peerId):
+      trace "CertificateVerifier: certificate did not match expected peer id",
+        expectedPeerId = peerId
+      return false
+    return true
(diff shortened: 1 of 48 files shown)


vladopajic (approved):

libp2p/transports/quictransport.nim
```diff
@@ -282,23 +282,36 @@ type QuicTransport* = ref object of Transport
...
-    if certificatesDer.len != 1:
-      trace "CertificateVerifier: expected one certificate in the chain",
-        cert_count = certificatesDer.len
+proc verifyCertificates(
+    certificatesDer: seq[seq[byte]], expectedPeerId: Opt[PeerId] = Opt.none(PeerId)
+): bool =
+  if certificatesDer.len != 1:
+    trace "CertificateVerifier: expected one certificate in the chain",
+      cert_count = certificatesDer.len
+    return false
```
gmelodie: Why not accept more than one cert in chain?

libp2p/transports/quictransport.nim
```diff
@@ -282,23 +282,36 @@ type QuicTransport* = ref object of Transport
...
+      trace "CertificateVerifier: certificate did not match expected peer id",
+        expectedPeerId = peerId
+      return false
+    return true
 
-    return cert.verify()
+  if cert.verifiedIdentityKey().isNone:
+    trace "CertificateVerifier: certificate verification failed"
+    return false
+  return true
```
gmelodie: We should aim for all last return statements to not use `return` 
```suggestion
  true
```

libp2p/transports/quictransport.nim
```diff
@@ -506,6 +522,12 @@ method dial*(
 
     let client = self.client.get()
     let quicConnection = await client.dial(taAddress)
+    if peerId.isSome and not verifyCertificates(quicConnection.certificates(), peerId):
```
gmelodie: If you're passing an `Opt` of a peerid to `verifymcertificates`, either have that function verify that `isSome` or extract value here and have that take a peerid directly

gmelodie (approved):

libp2p/transports/quictransport.nim
richard-ramos: https://github.com/libp2p/specs/blob/master/tls/tls.md
> Endpoints MUST NOT send a certificate chain that contains more than one certificate.

PR #2568, opened 2026-05-28
fix(service-disco): multi service ads caching

SionoiS:
I noticed that when a registrar accept an advert it was not cached under ALL the services it contained, this is now the case.

### DIFF
--- a/libp2p/protocols/service_discovery/registrar.nim
+++ b/libp2p/protocols/service_discovery/registrar.nim
@@ -55,6 +55,19 @@ proc removeAd*(ipTree: IpTree, ad: Advertisement) {.raises: [].} =
 proc isExpired(now, ts: Moment, expiry: Duration): bool {.inline.} =
   now - ts > expiry
 
+proc hasAdReference(
+    registrar: Registrar, key: AdvertisementKey, excludeSid: ServiceId = default(ServiceId)
+): bool {.raises: [].} =
+  ## Returns true if the ad key is still present in any service's cache list
+  ## (optionally ignoring one service, e.g. the one currently being pruned).
+  for sid, sads in registrar.cache:
+    if sid == excludeSid:
+      continue
+    for a in sads:
+      if a.toAdvertisementKey() == key:
+        return true
+  false
+
 proc pruneAdsForService(
     registrar: Registrar,
     serviceId: ServiceId,
@@ -67,14 +80,22 @@ proc pruneAdsForService(
   while i < ads.len:
     let ad = ads[i]
     let key = ad.toAdvertisementKey()
+    var doDelete = false
     registrar.cacheTimestamps.withValue(key, ts):
       if isExpired(now, ts[], advertExpiry):
+        doDelete = true
+    do:
+      # key not present -> orphan, delete from list
+      doDelete = true
+    if doDelete:
+      ads.delete(i)
+      if not hasAdReference(registrar, key, excludeSid = serviceId):
         registrar.ipTree.removeAd(ad)
-        registrar.cacheTimestamps.del(key)
-        ads.delete(i)
+        if key in registrar.cacheTimestamps:
+          registrar.cacheTimestamps.del(key)
         inc(expiredCount)
-      else:
-        inc(i)
+    else:
+      inc(i)
 
 proc pruneEmptyServices(registrar: Registrar) =
   var toRemove: seq[ServiceId] = @[]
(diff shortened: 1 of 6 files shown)


libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -67,14 +82,22 @@ proc pruneAdsForService(
...
         registrar.ipTree.removeAd(ad)
-        registrar.cacheTimestamps.del(key)
-        ads.delete(i)
+        if key in registrar.cacheTimestamps:
+          registrar.cacheTimestamps.del(key)
         inc(expiredCount)
-      else:
-        inc(i)
+    else:
+      inc(i)
```
gmelodie: Asked claude to come up with a more readable version:
```nim
var writeIdx = 0
for readIdx in 0 ..< ads.len:
  let
    ad = ads[readIdx]
    key = ad.toAdvertisementKey()

  # Absent key counts as an orphan, so default to expired.
  var expired = true
  registrar.cacheTimestamps.withValue(key, ts):
    expired = isExpired(now, ts[], advertExpiry)

  if expired:
    if not hasAdReference(registrar, key, excludeSid = serviceId):
      registrar.ipTree.removeAd(ad)
      registrar.cacheTimestamps.del(key)
      inc expiredCount
  else:
    ads[writeIdx] = ad
    inc writeIdx

ads.setLen(writeIdx)
```

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -312,30 +335,50 @@ proc findOldestKey(disco: ServiceDiscovery): Opt[AdvertisementKey] =
...
+      if a.toAdvertisementKey() == oldestKey:
+        oldestAd = a
+        found = true
+        break
+    if found:
+      break
+  if not found:
+    # Stale ts entry with no list refs; just drop it
+    disco.registrar.cacheTimestamps.del(oldestKey)
+    return
```
gmelodie: extract into fn

```nim
func findAd(registrar: Registrar, key: AdvertisementKey): Opt[Advertisement] =
  for sads in registrar.cache.values:
    for a in sads:
      if a.toAdvertisementKey() == key:
        return Opt.some(a)
  Opt.none(Advertisement)
```

call like
```nim
let oldestAd = disco.registrar.findAd(oldestKey).valueOr:
  # Stale ts entry with no list refs; just drop it
  disco.registrar.cacheTimestamps.del(oldestKey)
  return
```

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -312,30 +335,50 @@ proc findOldestKey(disco: ServiceDiscovery): Opt[AdvertisementKey] =
...
+      else:
+        inc(i)
+    if removedHere and sads.len == 0:
+      emptiedSids.add(sid)
+    if sid == serviceId and removedHere:
+      ads = sads
 
-          break search
+  for sid in emptiedSids:
+    disco.registrar.cache.del(sid)
```
gmelodie: extract as fn
```nim
proc removeAdFromLists(
    registrar: Registrar, key: AdvertisementKey
): seq[ServiceId] =
  var emptiedSids: seq[ServiceId]
  for sid, sads in registrar.cache.mpairs:
    var
      writeIdx = 0
      removedHere = false
    for readIdx in 0 ..< sads.len:
      if sads[readIdx].toAdvertisementKey() == key:
        removedHere = true
      else:
        sads[writeIdx] = sads[readIdx]
        inc writeIdx
    if removedHere:
      sads.setLen(writeIdx)
      if sads.len == 0:
        emptiedSids.add(sid)
  emptiedSids
```
calling with
```nim
let emptiedSids = disco.registrar.removeAdFromLists(oldestKey)
```

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -380,26 +423,76 @@ proc insertNewAd*(
...
+        let ex = sads[i]
+        if ex.data.peerId == peerId and ex.data.seqNo < ad.data.seqNo:
+          let exKey = ex.toAdvertisementKey()
+          sads.delete(i)
+          if exKey in disco.registrar.cacheTimestamps:
+            disco.registrar.ipTree.removeAd(ex)
+            disco.registrar.cacheTimestamps.del(exKey)
+          shouldUpdateMetrics = true
+        else:
+          inc(i)
```
gmelodie: Same as above, this should be a separate function

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -380,26 +423,76 @@ proc insertNewAd*(
...
     else:
-      disco.insertNewAd(serviceId, ads, ad, now)
-
-  disco.registrar.cache[serviceId] = ads
+      if adKey in disco.registrar.cacheTimestamps:
+        # Already live under other service(s) - just reference it here too
+        ads.add(ad)
+        disco.registrar.cache[sid] = ads
+        disco.registrar.cacheTimestamps[adKey] = now
+        shouldUpdateMetrics = true
```
gmelodie: separate function

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -380,26 +423,76 @@ proc insertNewAd*(
...
+      else:
+        # Brand new advertisement
+        if disco.registrar.cacheTimestamps.len.uint64 >= disco.discoConfig.advertCacheCap:
+          evictOldestAd(disco, sid, ads)
+        ads.add(ad)
+        disco.registrar.cacheTimestamps[adKey] = now
+        disco.registrar.ipTree.insertAd(ad)
+        disco.registrar.cache[sid] = ads
+        shouldUpdateMetrics = true
```
gmelodie: separate function

gmelodie (commented):
As a general rule of thumb, if there's 
```
# comment1 explaining a block of code
if/for/while:
  ...

# comment2 explaining another block
if/for/while:
  ...
```
I like to always exctract for readability (saves you the comments)
```
fn saysWhatComment1Does():
  ...
fn saysWhatComment2Does():
  ...

saysWhatComment1Does()
saysWhatComment2Does()
```

tests/libp2p/service_discovery/component/test_advertise_discover.nim
```diff
@@ -112,42 +112,6 @@ suite "Service Discovery Component - Advertise Discover":
...
-      block:
-        let foundA = await discovererNode.lookup(svcAId)
-        foundA.get().anyIt(it.data.peerId == advertiserNode.switch.peerInfo.peerId) and
-          foundA.get().len == 1
-      block:
-        let foundB = await discovererNode.lookup(svcBId)
-        foundB.get().anyIt(it.data.peerId == advertiserNode.switch.peerInfo.peerId) and
-          foundB.get().len == 1
-          # Regression for vacp2p/nim-libp2p#2431: this lookup previously returned a duplicate.
-
```
rlve, line 150: Why did you remove this test?

SionoiS:
> As a general rule of thumb, if there's
> 
> ```
> # comment1 explaining a block of code
> if/for/while:
>   ...
> 
> # comment2 explaining another block
> if/for/while:
>   ...
> ```
> 
> I like to always exctract for readability (saves you the comments)
> 
> ```
> fn saysWhatComment1Does():
>   ...
> fn saysWhatComment2Does():
>   ...
> 
> saysWhatComment1Does()
> saysWhatComment2Does()
> ```

yeah kinda slop this new assistant. I'll make the changes :+1:

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -467,7 +560,7 @@ proc registration*(disco: ServiceDiscovery, peerId: PeerId, inMsg: Message): Mes
   tWait = disco.processRetryTicket(regMsg, ad, tWait)
```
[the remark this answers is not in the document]
SionoiS: We will not short-circuit or change the wait mechanism in any way since it would lead to DOS vector in the protocol.

SionoiS:
I'm thinking of refactoring the cache into a proper min-DB in another PR so that it can be abstracted away. This code is getting ridiculously hard to follow...

tests/libp2p/service_discovery/component/test_advertise_discover.nim
SionoiS: Here's my thinking on this.

- The test case of advertising multiple services for one advert is already covered by `auto-advetise on start`
- The case of re-registering an advert cannot happen per the spec even if we modify the API into multi-services one.
- If we change the config to make it possible (safetyParam = 0.0), what are we even testing?

WDYT?

rlve: Ah right, I forgot about `auto-advertise on start`. I guess it's okay to remove it then.

If/when the multi-service API is added then new tests should be added anyway. 

> If we change the config to make it possible (safetyParam = 0.0), what are we even testing?

I guess we're simulating a situation where a node advertises serviceA, waits 15min, and then advertises serviceB 😃 not sure if it's valuable though

SionoiS:
I just had a good discussion with @jm-clius about registering all services an advert contains. The gist of it is that we specifically don't want to do that because it changes the protocols intent. 

The protocol is designed for advertisers to register an advert (whatever it is) at a registrar after some wait period.

This PR assumes what the advert is and changes the wait calculation to "wait for all (instead of one) your services to be discoverable at this location".

I will close this and this feature will not be implemented.

PR #2573, opened 2026-05-28
feat(nat): upnp

gmelodie:
## Summary
  - Adds `libp2p/services/nat/upnp_mapper.nim`: worker-thread-based async
  wrapper around the blocking miniupnpc C API
  - Holds a `Miniupnp` instance for the worker's lifetime; main-thread async
  dispatches one request at a time via two `ThreadSignalPtr`s
  - All cross-thread data is fixed-size POD — no Nim GC heap crosses the thread
   boundary

  Part 3 of 6. Not yet wired into NATService — that comes in part 5.

### DIFF
--- a/libp2p/services/nat/upnp_mapper.nim
+++ b/libp2p/services/nat/upnp_mapper.nim
@@ -61,6 +61,15 @@ type
     # and orders close() after any in-flight dispatch.
     lock: AsyncLock
 
+proc free(ctx: ptr UpnpWorkerCtx) =
+  # Tolerates partial initialization — a signal field that was never assigned
+  # (or that was already closed in-place) is left as nil and skipped here.
+  if not ctx.reqSignal.isNil:
+    discard ctx.reqSignal.close()
+  if not ctx.respSignal.isNil:
+    discard ctx.respSignal.close()
+  freeShared(ctx)
+
 proc setError(resp: var UpnpResponse, msg: string) =
   resp.success = false
   resp.ip.reset()
@@ -136,6 +145,8 @@ proc handleUnmap(upnp: Miniupnp, req: UpnpRequest, resp: var UpnpResponse) =
 
 proc upnpWorker(ctx: ptr UpnpWorkerCtx) {.thread.} =
   let upnp = newMiniupnp()
+  defer:
+    upnp.close()
 
   while true:
     let w = ctx.reqSignal.waitSync()
@@ -145,6 +156,7 @@ proc upnpWorker(ctx: ptr UpnpWorkerCtx) {.thread.} =
       # further requests — exit. Closing respSignal unblocks any awaiter.
       error "UpnpMapper worker reqSignal wait failed", err = w.error
       discard ctx.respSignal.close()
+      ctx.respSignal = nil # don't double-close in free()
       break
     if ctx.shutdown.load():
       error "UpnpMapper worker shutdown"
@@ -167,9 +179,8 @@ proc upnpWorker(ctx: ptr UpnpWorkerCtx) {.thread.} =
       error "UpnpMapper worker respSignal fire failed",
         err = (if fr.isErr: fr.error else: "timeout")
       discard ctx.respSignal.close()
+      ctx.respSignal = nil # don't double-close in free()
       break
-
-  upnp.close()
 
 proc dispatch(
     self: UpnpMapper, req: sink UpnpRequest
@@ -182,7 +193,7 @@ proc dispatch(
     try:
       self.lock.release()
     except AsyncLockError as e:
-      warn "UpnpMapper lock release failed", err = e.msg
+      raiseAssert "UpnpMapper lock release failed: " & e.msg
 
   if self.closed.load():
     return err("UpnpMapper closed")
@@ -216,11 +227,10 @@ proc dispatch(
 proc newUpnpMapper*(): UpnpMapper {.raises: [ResourceExhaustedError].} =
   let ctx = createShared(UpnpWorkerCtx, 1)
   ctx.reqSignal = ThreadSignalPtr.new().valueOr:
-    freeShared(ctx)
+    free(ctx)
     raise newException(ResourceExhaustedError, "UpnpMapper reqSignal: " & error)
   ctx.respSignal = ThreadSignalPtr.new().valueOr:
-    discard ctx.reqSignal.close()
-    freeShared(ctx)
+    free(ctx)
     raise newException(ResourceExhaustedError, "UpnpMapper respSignal: " & error)
 
   let mapper = UpnpMapper(ctx: ctx, lock: newAsyncLock())
@@ -228,15 +238,11 @@ proc newUpnpMapper*(): UpnpMapper {.raises: [ResourceExhaustedError].} =
   try:
     createThread(mapper.thread, upnpWorker, ctx)
   except ValueError as e:
-    discard ctx.reqSignal.close()
-    discard ctx.respSignal.close()
-    freeShared(ctx)
+    free(ctx)
     raise
       newException(ResourceExhaustedError, "UpnpMapper thread create failed: " & e.msg)
   except ResourceExhaustedError as e:
-    discard ctx.reqSignal.close()
-    discard ctx.respSignal.close()
-    freeShared(ctx)
+    free(ctx)
     raise
       newException(ResourceExhaustedError, "UpnpMapper thread create failed: " & e.msg)
 
@@ -287,7 +293,11 @@ method unmap*(
     return err(resp.getError())
   ok()
 
-method close*(self: UpnpMapper) {.async: (raises: [CancelledError]), gcsafe.} =
+method close*(self: UpnpMapper) {.async: (raises: []), gcsafe.} =
+  # No cancellable awaits: the only await is `noCancel(lock.acquire())`, and
+  # the lock release in the defer below is guarded against AsyncLockError —
+  # which is a developer error (release without acquire), not a runtime
+  # failure mode.
   if self.closed.exchange(true):
     return
 
@@ -298,7 +308,7 @@ method close*(self: UpnpMapper) {.async: (raises: [CancelledError]), gcsafe.} =
     try:
       self.lock.release()
     except AsyncLockError as e:
-      warn "UpnpMapper lock release failed", err = e.msg
+      raiseAssert "UpnpMapper lock release failed: " & e.msg
 
   # Tell the worker to exit. A separate shutdown flag (rather than a
   # urShutdown request) avoids racing with worker reads of ctx.request.
@@ -310,9 +320,8 @@ method close*(self: UpnpMapper) {.async: (raises: [CancelledError]), gcsafe.} =
     # Force the worker's waitSync to return so it can exit and joinThread
     # below does not deadlock.
     discard self.ctx.reqSignal.close()
+    self.ctx.reqSignal = nil # don't double-close in free()
 
   joinThread(self.thread)
 
-  discard self.ctx.reqSignal.close()
-  discard self.ctx.respSignal.close()
-  freeShared(self.ctx)
+  free(self.ctx)
--- a/libp2p/services/nat/upnp_mapper.nim
+++ b/libp2p/services/nat/upnp_mapper.nim
@@ -57,6 +57,11 @@ type
     ctx: ptr UpnpWorkerCtx
     thread: Thread[ptr UpnpWorkerCtx]
     closed: Atomic[bool]
+    # Active dispatch + close participants. Whoever decrements this to zero
+    # after `closed` is set performs the final teardown, so ctx and the signal
+    # handles outlive any in-flight `dispatch` coroutine.
+    refs: Atomic[int]
+    finalized: Atomic[bool]
     lock: AsyncLock
 
 proc setError(resp: var UpnpResponse, msg: string) =
@@ -103,13 +108,16 @@ proc handleDiscover(upnp: Miniupnp, req: UpnpRequest, resp: var UpnpResponse) =
     resp.setError("upnp parseIpAddress: " & e.msg)
 
 proc handleMap(upnp: Miniupnp, req: UpnpRequest, resp: var UpnpResponse) =
+  # Clamp via uint64 so the int() conversion is safe on 32-bit targets where
+  # high(int) < high(uint32).
+  let lease = int(min(req.mapLease.uint64, uint64(high(int))))
   let r = upnp.addPortMapping(
     externalPort = $req.mapExternal,
     protocol = toUpnpProto(req.mapProto),
     internalHost = upnp.lanAddr,
     internalPort = $req.mapInternal,
     desc = "nim-libp2p",
-    leaseDuration = int(req.mapLease),
+    leaseDuration = lease,
   )
   if r.isErr:
     resp.setError("upnp addPortMapping: " & $r.error)
@@ -150,36 +158,54 @@ proc upnpWorker(ctx: ptr UpnpWorkerCtx) {.thread.} =
 
   upnp.close()
 
+proc finalize(self: UpnpMapper) =
+  if self.finalized.exchange(true):
+    return
+  discard self.ctx.reqSignal.close()
+  discard self.ctx.respSignal.close()
+  freeShared(self.ctx)
+
+proc release(self: UpnpMapper) =
+  if self.refs.fetchSub(1) == 1 and self.closed.load:
+    self.finalize()
+
 proc dispatch(
     self: UpnpMapper, req: sink UpnpRequest
 ): Future[Result[UpnpResponse, string]] {.async: (raises: [CancelledError]), gcsafe.} =
   if self.closed.load:
     return err("UpnpMapper closed")
 
-  await self.lock.acquire()
+  discard self.refs.fetchAdd(1)
   try:
     if self.closed.load:
       return err("UpnpMapper closed")
 
-    self.ctx.request = req
-    let fr = self.ctx.reqSignal.fireSync()
-    if fr.isErr or not fr.get():
-      return err("UpnpMapper signal fire failed")
-
+    await self.lock.acquire()
     try:
-      await self.ctx.respSignal.wait()
-    except AsyncError as e:
-      return err("UpnpMapper wait: " & e.msg)
+      if self.closed.load:
+        return err("UpnpMapper closed")
 
-    if self.closed.load:
-      return err("UpnpMapper closed")
+      self.ctx.request = req
+      let fr = self.ctx.reqSignal.fireSync()
+      if fr.isErr or not fr.get():
+        return err("UpnpMapper signal fire failed")
 
-    ok(self.ctx.response)
+      try:
+        await self.ctx.respSignal.wait()
+      except AsyncError as e:
+        return err("UpnpMapper wait: " & e.msg)
+
+      if self.closed.load:
+        return err("UpnpMapper closed")
+
+      return ok(self.ctx.response)
+    finally:
+      try:
+        self.lock.release()
+      except AsyncLockError as e:
+        warn "UpnpMapper lock release failed", err = e.msg
   finally:
-    try:
-      self.lock.release()
-    except AsyncLockError as e:
-      warn "UpnpMapper lock release failed", err = e.msg
+    self.release()
 
 proc newUpnpMapper*(): UpnpMapper {.raises: [ResourceExhaustedError].} =
   let ctx = createShared(UpnpWorkerCtx, 1)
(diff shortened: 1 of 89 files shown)


libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,310 @@
...
+  self.ctx.request = UpnpRequest(kind: urShutdown)
+  let fr = self.ctx.reqSignal.fireSync()
+  if fr.isErr or not fr.get():
+    warn "UpnpMapper shutdown signal failed",
+      err = (if fr.isErr: fr.error else: "timeout")
+
+  try:
+    joinThread(self.thread)
+  except Exception as exc:
+    warn "UpnpMapper joinThread failed", err = exc.msg
```
vladopajic: should never catch `Exception`, but catching `CatchableError` is okay

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,310 @@
...
+method unmap*(
+    self: UpnpMapper, externalPort: Port, proto: MapProto
+): Future[Result[void, string]] {.async: (raises: [CancelledError]), gcsafe.} =
+  await self.dispatch(
+    UpnpRequest(kind: urUnmap, unmapExternal: uint16(externalPort), unmapProto: proto)
+  )
+
+  if self.ctx.response.success:
+    return Result[void, string].ok()
+  Result[void, string].err(self.ctx.response.getError())
```
vladopajic: usually errors are in early return branches

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,310 @@
...
+  resp.success = false
+  resp.ipFamily = 0
+  resp.externalPort = 0
+  let n = min(msg.len, ErrBufLen)
+  for i in 0 ..< n:
+    resp.errorBuf[i] = msg[i]
+  resp.errorLen = n
+
+proc getError(resp: UpnpResponse): string =
+  result = newString(resp.errorLen)
```
vladopajic: `result`

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,310 @@
...
+  let ctx = createShared(UpnpWorkerCtx, 1)
+  ctx.reqSignal = ThreadSignalPtr.new().valueOr:
+    freeShared(ctx)
+    raise newException(ResourceExhaustedError, "UpnpMapper reqSignal: " & error)
+  ctx.respSignal = ThreadSignalPtr.new().valueOr:
+    discard ctx.reqSignal.close()
+    freeShared(ctx)
+    raise newException(ResourceExhaustedError, "UpnpMapper respSignal: " & error)
+
+  result = UpnpMapper(ctx: ctx)
```
vladopajic: `result`

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,310 @@
...
+      unmapExternal: uint16
+      unmapProto: MapProto
+    of urShutdown:
+      discard
+
+  UpnpResponse = object
+    success: bool
+    errorLen: int
+    errorBuf: array[ErrBufLen, char]
+    ipFamily: int # 0 = unset, 4 = IPv4, 6 = IPv6
```
vladopajic: `Opt[IpAddressFamily]` avoids need for any comment here

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,310 @@
...
+# ----------------------------------------------------------------------------
+
+proc toUpnpProto(p: MapProto): UPNPProtocol =
+  case p
+  of mpTcp: UPNPProtocol.TCP
+  of mpUdp: UPNPProtocol.UDP
+
+proc handleDiscover(upnp: var Miniupnp, req: UpnpRequest, resp: var UpnpResponse) =
+  if upnp == nil:
+    upnp = newMiniupnp()
```
vladopajic: seem strange to initialized variable managed outside from here

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,310 @@
...
+
+  let ipr = upnp.externalIPAddress()
+  if ipr.isErr:
+    resp.setError("upnp externalIPAddress: " & $ipr.error)
+    return
+
+  try:
+    let ip = parseIpAddress(ipr.get())
+    resp.success = true
+    resp.setIp(ip)
```
vladopajic:
```suggestion
    resp.success = true
    resp.setIp(parseIpAddress(ipr.get()))
```

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,310 @@
...
+  except ValueError as exc:
```
vladopajic: e?

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,310 @@
...
+proc dispatch(
+    self: UpnpMapper, req: sink UpnpRequest
+): Future[void] {.async: (raises: [CancelledError]).} =
+  if self.closed.load:
+    raise newException(CancelledError, "UpnpMapper closed")
+
+  self.ctx.request = req
+  let fr = self.ctx.reqSignal.fireSync()
+  if fr.isErr or not fr.get():
+    raise newException(CancelledError, "UpnpMapper signal fire failed")
```
vladopajic: `CancelledError` ?

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,310 @@
...
+      unmapProto: MapProto
+    of urShutdown:
+      discard
+
+  UpnpResponse = object
+    success: bool
+    errorLen: int
+    errorBuf: array[ErrBufLen, char]
+    ipFamily: int # 0 = unset, 4 = IPv4, 6 = IPv6
+    ipBuf: array[IpBufLen, byte]
```
vladopajic: why is it not possible to use `IpAddress` type? why does `getIp` and `setIp` exist?

vladopajic (commented):
would it be possible to have some unit test for this file?

libp2p/services/nat/upnp_mapper.nim
richard-ramos: e

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,310 @@
...
+    respSignal: ThreadSignalPtr
+    request: UpnpRequest
+    response: UpnpResponse
+
+  UpnpMapper* = ref object of PortMapper
+    ctx: ptr UpnpWorkerCtx
+    thread: Thread[ptr UpnpWorkerCtx]
+    closed: Atomic[bool]
+
+# ----------------------------------------------------------------------------
```
richard-ramos: These --------------------------------------------------------- are kinda noisy

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,310 @@
...
+      mapInternal: uint16(internalPort),
+      mapExternal: uint16(externalPort),
+      mapProto: proto,
+      mapLease: lease,
+    )
+  )
+
+  if self.ctx.response.success:
+    return Result[Port, string].ok(Port(self.ctx.response.externalPort))
+  Result[Port, string].err(self.ctx.response.getError())
```
richard-ramos: I think you can avoid the prefix to make this more readable.
```suggestion
    return ok(Port(self.ctx.response.externalPort))
  err(self.ctx.response.getError())
```
there are other parts of the code like this

richard-ramos (commented):
Same comments i wrote in https://github.com/vacp2p/nim-libp2p/pull/2574 apply here, specially the one related to the asynclock, as there's this comment of "The async API on top dispatches one request at a time " in the header, which is not enforced

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,272 @@
...
+    if self.closed.load:
+      return err("UpnpMapper closed")
+
+    self.ctx.request = req
+    let fr = self.ctx.reqSignal.fireSync()
+    if fr.isErr or not fr.get():
+      return err("UpnpMapper signal fire failed")
+
+    try:
+      await self.ctx.respSignal.wait()
```
richard-ramos: Something that will likely never happen, but it is worth mentioning (just for code cleanliness) is that evenwith the lock, if this is cancelled while `awaiting` here, you'll get the lock released, but upnpWorker would still keep running and trigger the fireSignal... meaning that a new dispatch could read the response that was meant for the original cancelled worker.

Dunno if worth fixing tho 🤔

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,272 @@
...
+    return err(resp.getError())
+  ok()
+
+method close*(self: UpnpMapper) {.gcsafe, raises: [].} =
+  if self.closed.exchange(true):
+    return
+
+  # Wake any in-flight dispatch waiting on respSignal so it observes `closed`
+  # and returns promptly instead of hanging.
+  discard self.ctx.respSignal.fireSync()
```
richard-ramos: Since this is dealing with a lock protected variable, it should also acquire/release the lock

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,290 @@
...
+
+  resp.success = true
+
+proc upnpWorker(ctx: ptr UpnpWorkerCtx) {.thread.} =
+  let upnp = newMiniupnp()
+
+  while true:
+    let w = ctx.reqSignal.waitSync()
+    if w.isErr or ctx.shutdown.load:
+      break
```
vladopajic, line 163: perhaps log error?

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,290 @@
...
+
+  upnp.close()
+
+proc dispatch(
+    self: UpnpMapper, req: sink UpnpRequest
+): Future[Result[UpnpResponse, string]] {.async: (raises: [CancelledError]), gcsafe.} =
+  if self.closed.load:
+    return err("UpnpMapper closed")
+
+  await self.lock.acquire()
```
vladopajic, line 191: let's prefer pattern with defer release, reasons:
- in general `try` body should be as small as possible because otherwise code gets harder to follow
- in general we should never-ever have nested `try` statements - it has been source of many bugs before 
- it avoid unnecessary indentation, which is especially annoying when code is changed, when indentation level changes then whole block is seen as diff; 
  - this is especially true if code later on wants to reduce lock level, or other changes
- seeing release right after acquire you immediately understand lock level
- `try-finally` - has no advantage?

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,290 @@
...
+      handleUnmap(upnp, ctx.request, ctx.response)
+
+    discard ctx.respSignal.fireSync()
+
+  upnp.close()
+
+proc dispatch(
+    self: UpnpMapper, req: sink UpnpRequest
+): Future[Result[UpnpResponse, string]] {.async: (raises: [CancelledError]), gcsafe.} =
+  if self.closed.load:
```
vladopajic: by default we should use `()`, if function adjective we can avoid `()`. 

"load" is clearly action and we better use `()` there.

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,290 @@
...
+  let ctx = createShared(UpnpWorkerCtx, 1)
+  ctx.reqSignal = ThreadSignalPtr.new().valueOr:
+    freeShared(ctx)
+    raise newException(ResourceExhaustedError, "UpnpMapper reqSignal: " & error)
+  ctx.respSignal = ThreadSignalPtr.new().valueOr:
+    discard ctx.reqSignal.close()
+    freeShared(ctx)
+    raise newException(ResourceExhaustedError, "UpnpMapper respSignal: " & error)
+
+  result = UpnpMapper(ctx: ctx, lock: newAsyncLock())
```
vladopajic: `result`

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,290 @@
...
+    self.ctx.request = req
+    let fr = self.ctx.reqSignal.fireSync()
+    if fr.isErr or not fr.get():
+      return err("UpnpMapper signal fire failed")
+
+    try:
+      await self.ctx.respSignal.wait()
+    except AsyncError as e:
+      return err("UpnpMapper wait: " & e.msg)
+    except CancelledError as exc:
```
vladopajic: `e`

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,290 @@
...
+
+  result = UpnpMapper(ctx: ctx, lock: newAsyncLock())
+
+  try:
+    createThread(result.thread, upnpWorker, ctx)
+  except ValueError, ResourceExhaustedError:
+    discard ctx.reqSignal.close()
+    discard ctx.respSignal.close()
+    freeShared(ctx)
+    raise newException(ResourceExhaustedError, "UpnpMapper thread create failed")
```
vladopajic: as a rule of thumb we should always keep original error message

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,290 @@
...
+    return err(resp.getError())
+  ok()
+
+method close*(self: UpnpMapper) {.async: (raises: [CancelledError]), gcsafe.} =
+  if self.closed.exchange(true):
+    return
+
+  # Wait uncancellably for any in-flight dispatch to release the lock, so the
+  # signal handles and ctx are not torn down while still in use.
+  await noCancel(self.lock.acquire())
```
vladopajic, line 306: release in `defer`

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,317 @@
...
+): Future[Result[void, string]] {.async: (raises: [CancelledError]), gcsafe.} =
+  let resp = ?await self.dispatch(
+    UpnpRequest(kind: urUnmap, unmapExternal: uint16(externalPort), unmapProto: proto)
+  )
+
+  if not resp.success:
+    return err(resp.getError())
+  ok()
+
+method close*(self: UpnpMapper) {.async: (raises: [CancelledError]), gcsafe.} =
```
vladopajic: based on code this proc should never raise `CancelledError`?

only async operation is wait for lock, but there is `noCancel` there?

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,317 @@
...
+    return
+
+  # Wait uncancellably for any in-flight dispatch to release the lock, so the
+  # signal handles and ctx are not torn down while still in use.
+  await noCancel(self.lock.acquire())
+  defer:
+    try:
+      self.lock.release()
+    except AsyncLockError as e:
+      warn "UpnpMapper lock release failed", err = e.msg
```
vladopajic: class of this `except` construct is same as constructs in nim that handle for example `KeyError`.

our code is always designed in such way that that `KeyError` never happens, so that block is never going to be executed. if error do happen then it is likely developer error so adding logs doesn't make much sense as that log will never get noticed. 
in these cases code usually has `raiseAssert "this should never happen but it did"` which makes it much more easier to notice errors if they do happen. 

tld; swap `warn` with `raiseAssert`

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,317 @@
...
+    raise newException(ResourceExhaustedError, "UpnpMapper respSignal: " & error)
+
+  let mapper = UpnpMapper(ctx: ctx, lock: newAsyncLock())
+
+  try:
+    createThread(mapper.thread, upnpWorker, ctx)
+  except ValueError as e:
+    discard ctx.reqSignal.close()
+    discard ctx.respSignal.close()
+    freeShared(ctx)
```
vladopajic: `free()`

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,317 @@
...
+    raise
+      newException(ResourceExhaustedError, "UpnpMapper thread create failed: " & e.msg)
+  except ResourceExhaustedError as e:
+    discard ctx.reqSignal.close()
+    discard ctx.respSignal.close()
+    freeShared(ctx)
```
vladopajic: `free()`

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,317 @@
...
+      err = (if fr.isErr: fr.error else: "timeout")
+    # Force the worker's waitSync to return so it can exit and joinThread
+    # below does not deadlock.
+    discard self.ctx.reqSignal.close()
+
+  joinThread(self.thread)
+
+  discard self.ctx.reqSignal.close()
+  discard self.ctx.respSignal.close()
+  freeShared(self.ctx)
```
vladopajic: `free()`

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,317 @@
...
+
+  if self.closed.load():
+    return err("UpnpMapper closed")
+
+  ok(self.ctx.response)
+
+proc newUpnpMapper*(): UpnpMapper {.raises: [ResourceExhaustedError].} =
+  let ctx = createShared(UpnpWorkerCtx, 1)
+  ctx.reqSignal = ThreadSignalPtr.new().valueOr:
+    freeShared(ctx)
```
vladopajic: `free()`

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,317 @@
...
+    raise newException(ResourceExhaustedError, "UpnpMapper reqSignal: " & error)
+  ctx.respSignal = ThreadSignalPtr.new().valueOr:
+    discard ctx.reqSignal.close()
+    freeShared(ctx)
```
vladopajic: `free()`

libp2p/services/nat/upnp_mapper.nim
vladopajic: headsup, it is still free, just becasue `reqSignal` is nill then `.close()` is not happening.

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,317 @@
...
+    let fr = ctx.respSignal.fireSync()
+    if fr.isErr or not fr.get():
+      # The awaiting dispatch() can't see this response. Close respSignal so
+      # the wait errors out instead of hanging, then exit the worker.
+      error "UpnpMapper worker respSignal fire failed",
+        err = (if fr.isErr: fr.error else: "timeout")
+      discard ctx.respSignal.close()
+      break
+
+  upnp.close()
```
vladopajic: would be nicer to have it in `defer`

libp2p/services/nat/upnp_mapper.nim
```diff
@@ -0,0 +1,317 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
```
vladopajic, line 1: all new code should have some, at least basic unit test

richard-ramos (approved):

tests/libp2p/services/test_upnp_mapper.nim
```diff
@@ -0,0 +1,82 @@
...
+
+    var order: seq[int]
+    proc tag(i: int): Future[void] {.async: (raises: [CancelledError]).} =
+      discard await m.discover(50.milliseconds)
+      order.add(i)
+
+    let f1 = tag(1)
+    let f2 = tag(2)
+    await allFutures(f1, f2)
+    check order == @[1, 2]
```
vladopajic, line 82: this check will always be true?

not sure if it is possible assert this

tests/libp2p/services/test_upnp_mapper.nim
```diff
@@ -0,0 +1,82 @@
...
+    # in-flight dispatch to release the lock before tearing the worker down.
+    let m = newUpnpMapper()
+    let fut = m.discover(200.milliseconds)
+    await sleepAsync(10.milliseconds)
+    fut.cancelSoon()
+
+    try:
+      discard await fut
+    except CancelledError:
+      discard
```
vladopajic, line 62: this is a race, fut can finish before sleep. 


usually we would want to to do:
```
expect CancelledError:
 discard await fut
```
but then we need to make sure that fut is not finished before `expect`. prehaps

```
if not fut.finished:
   expect ....
else:
   try again ? dunno
```

vladopajic (approved):

PR #2574, opened 2026-05-28
feat(nat): nat-pmp

gmelodie:
## Summary
  - Adds `libp2p/services/nat/natpmp_mapper.nim`: mirrors the worker-thread
  structure of `upnp_mapper.nim` around the `nat_traversal/natpmp` sync
  bindings

  Part 4 of 6.

### DIFF
--- a/libp2p/services/nat/natpmp_mapper.nim
+++ b/libp2p/services/nat/natpmp_mapper.nim
@@ -177,10 +177,7 @@ proc dispatch(
 ): Future[Result[MapperResponse, string]] {.async: (raises: [CancelledError]), gcsafe.} =
   await self.lock.acquire()
   defer:
-    try:
-      self.lock.release()
-    except AsyncLockError as e:
-      raiseAssert "NatPmpMapper lock release failed: " & e.msg
+    self.lock.safeRelease("NatPmpMapper")
 
   await self.drainPrevious()
   # Close can only have run at the awaits above; synchronous code below is
@@ -219,19 +216,21 @@ proc dispatch(
 
 proc newNatPmpMapper*(): NatPmpMapper {.raises: [ResourceExhaustedError].} =
   let ctx = createShared(NatPmpWorkerCtx, 1)
-  ctx.reqSignal = ThreadSignalPtr.new().valueOr:
-    free(ctx)
-    raise newException(ResourceExhaustedError, "NatPmpMapper reqSignal: " & error)
-  ctx.respSignal = ThreadSignalPtr.new().valueOr:
-    free(ctx)
-    raise newException(ResourceExhaustedError, "NatPmpMapper respSignal: " & error)
+  initSignals(ctx, "NatPmpMapper")
 
   let mapper = NatPmpMapper(ctx: ctx, lock: newAsyncLock())
   try:
     createThread(mapper.thread, natpmpWorker, ctx)
-  except ValueError, ResourceExhaustedError:
+  except ValueError as e:
     free(ctx)
-    raise newException(ResourceExhaustedError, "NatPmpMapper thread create failed")
+    raise newException(
+      ResourceExhaustedError, "NatPmpMapper thread create failed: " & e.msg
+    )
+  except ResourceExhaustedError as e:
+    free(ctx)
+    raise newException(
+      ResourceExhaustedError, "NatPmpMapper thread create failed: " & e.msg
+    )
   mapper
 
 method discover*(
@@ -327,10 +326,7 @@ method close*(self: NatPmpMapper) {.async: (raises: []), gcsafe.} =
   # signal handles and ctx are not mutated/torn down while still in use.
   await noCancel(self.lock.acquire())
   defer:
-    try:
-      self.lock.release()
-    except AsyncLockError as e:
-      raiseAssert "NatPmpMapper lock release failed: " & e.msg
+    self.lock.safeRelease("NatPmpMapper")
 
   self.ctx.request = NatPmpRequest(kind: nrShutdown)
   let fireR = self.ctx.reqSignal.fireSyncOrErr()
--- a/libp2p/services/nat/natpmp_mapper.nim
+++ b/libp2p/services/nat/natpmp_mapper.nim
@@ -341,11 +341,11 @@ method close*(self: NatPmpMapper) {.async: (raises: [CancelledError]), gcsafe.} 
   if self.closed.exchange(true):
     return
 
-  # Unblock any dispatch currently parked on respSignal so the caller sees
-  # CancelledError instead of waiting for a response that may never come if
+  # Unblock any dispatch currently parked on respSignal so it can release the
+  # lock — without this, noCancel(lock.acquire) below would block forever if
   # the worker is stuck inside libnatpmp. Also cancel any drain left over
-  # from a previous timeout, since joinThread below would otherwise wait
-  # for the worker to fire respSignal first.
+  # from a previous timeout, since joinThread further down would otherwise
+  # wait for the worker to fire respSignal first.
   let inflight = self.inflight
   if inflight != nil:
     inflight.cancelSoon()
@@ -353,6 +353,15 @@ method close*(self: NatPmpMapper) {.async: (raises: [CancelledError]), gcsafe.} 
   let drain = self.drainPending
   if drain != nil:
     drain.cancelSoon()
+
+  # Wait uncancellably for any in-flight dispatch to release the lock so the
+  # signal handles and ctx are not mutated/torn down while still in use.
+  await noCancel(self.lock.acquire())
+  defer:
+    try:
+      self.lock.release()
+    except AsyncLockError as e:
+      warn "NatPmpMapper lock release failed", err = e.msg
 
   self.ctx.request = NatPmpRequest(kind: nrShutdown)
   let fr = self.ctx.reqSignal.fireSync()
--- a/libp2p/services/nat/natpmp_mapper.nim
+++ b/libp2p/services/nat/natpmp_mapper.nim
@@ -14,9 +14,7 @@ import ./portmapper
 logScope:
   topics = "libp2p natservice natpmp"
 
-const
-  ErrBufLen = 256
-  IpBufLen = 16
+const ErrBufLen = 256
 
 type
   NatPmpReqKind = enum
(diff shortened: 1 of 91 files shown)


libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,290 @@
...
+  if ipOpt.isNone:
+    return Result[IpAddress, string].err("natpmp discover: missing IP in response")
+  Result[IpAddress, string].ok(ipOpt.get())
+
+method map*(
+    self: NatPmpMapper,
+    internalPort: Port,
+    externalPort: Port,
+    proto: MapProto,
+    lease: uint32,
```
richard-ramos, line 252: We should validate that lease is > 0, otherwise in libnatpmp, it will delete the mapping. https://github.com/miniupnp/libnatpmp/blob/master/natpmp.h#L174

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,290 @@
...
+    discard ctx.respSignal.fireSync()
+
+proc dispatch(
+    self: NatPmpMapper, req: sink NatPmpRequest
+): Future[void] {.async: (raises: [CancelledError]).} =
+  if self.closed.load:
+    raise newException(CancelledError, "NatPmpMapper closed")
+
+  self.ctx.request = req
+  let fr = self.ctx.reqSignal.fireSync()
```
richard-ramos: 🤔 should we protect the ctx with an asyncLock? I see there's potential for this function to be called concurrently from diff procs

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,290 @@
...
+): Future[Result[void, string]] {.async: (raises: [CancelledError]), gcsafe.} =
+  await self.dispatch(
+    NatPmpRequest(kind: nrUnmap, unmapExternal: uint16(externalPort), unmapProto: proto)
+  )
+
+  if self.ctx.response.success:
+    return Result[void, string].ok()
+  Result[void, string].err(self.ctx.response.getError())
+
+method close*(self: NatPmpMapper) {.gcsafe, raises: [].} =
```
richard-ramos: calling `close*` should also cancel any `dispatch` that is currently waiting on line 205

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,361 @@
...
+
+    discard ctx.respSignal.fireSync()
+
+proc dispatch(
+    self: NatPmpMapper, req: sink NatPmpRequest, timeout: Duration = InfiniteDuration
+): Future[void] {.async: (raises: [CancelledError]).} =
+  if self.closed.load:
+    raise newException(CancelledError, "NatPmpMapper closed")
+
+  await self.lock.acquire()
```
richard-ramos, line 178: This is the wrong place for acquiring / releasing the lock.

If you go for this approach, the dispatch function should instead return a copy of the response instead of modifying self.ctx.response

But if not, then the lock acquire/release should be moved to the caller because otherwise, the context can be modified before the callers read it.

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,361 @@
...
+          await waitFut
+        else:
+          if not await waitFut.withTimeout(timeout):
+            # Timed out; the worker is still running its blocking call (libnatpmp
+            # has its own internal retry timeout). Drain respSignal before
+            # releasing the lock so the next dispatch starts with clean state.
+            let drainFut = self.ctx.respSignal.wait()
+            self.inflight = drainFut
+            try:
+              await drainFut
```
richard-ramos: Do note that even tho you're using the timeout as suggested by copilot, this await can still block, if respSignal never arrives, exceeding the timeout duration.

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,389 @@
...
+        raise newException(CancelledError, "NatPmpMapper wait: " & exc.msg)
+    finally:
+      self.inflight = nil
+
+    if timedOut:
+      # waitFut is still pending; the worker will fire respSignal once
+      # libnatpmp's internal retry loop gives up. Stash the wait so the
+      # NEXT dispatch drains it, instead of blocking this caller past its
+      # requested timeout.
+      self.drainPending = waitFut
```
richard-ramos: I'm not sure if this will work, because if there's a timeout, `waitFut` will end up being cancelled. The next `dispatch` call, will wait for a `drainPending` that is already cancelled, raise a `CancelledError`, and catch it and `discard` it in line 208, and not consume `respSignal`

So when eventually `respSignal` is fired by the original worker, the next request will consume a response that does not belong to the request that initiated it , but to the older request instead

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,368 @@
...
+  # for the worker to fire respSignal first.
+  let inflight = self.inflight
+  if inflight != nil:
+    inflight.cancelSoon()
+
+  let drain = self.drainPending
+  if drain != nil:
+    drain.cancelSoon()
+
+  self.ctx.request = NatPmpRequest(kind: nrShutdown)
```
richard-ramos, line 331: You gotta acquire the lock here before using `self.ctx`, similar to how you did in https://github.com/vacp2p/nim-libp2p/blob/07d44d1c2271c576d7fcc9c1744030fc48487d8a/libp2p/services/nat/upnp_mapper.nim#L293-L300

richard-ramos (approved):
LGTM
Only an observation to be fixed in the `close` method.

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,368 @@
...
+  let fr = self.ctx.reqSignal.fireSync()
+  if fr.isErr() or not fr.get():
+    warn "NatPmpMapper shutdown signal failed",
+      err = (if fr.isErr(): fr.error() else: "timeout")
+
+  try:
+    joinThread(self.thread)
+  except Exception as e:
```
vladopajic: `CatchabeError`

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,368 @@
...
+  await self.drainPrevious()
+  # Close can only have run at the awaits above; synchronous code below is
+  # atomic, so one check here gates every subsequent self.ctx access.
+  self.checkOpen()
+
+  self.ctx.request = req
+  let fr = self.ctx.reqSignal.fireSync()
+  if fr.isErr() or not fr.get():
+    raise newException(CancelledError, "NatPmpMapper signal fire failed")
+
```
vladopajic, line 192: not sure in raising `CancelledError`. i think we should never raise is manually, we should only respond to it, or re-raise.

this code might work, but depending on how it is being used there is chance to interrupt whole program execution.  

better, have one error type, then raise it.

```
type MyOperationAbortedError = object of CatchableError

raise newException(MyOperationAbortedError, "User requested abort")
```

it just adds 1 line to total code but makse it much more safer.

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,368 @@
...
+  freeShared(ctx)
+
+proc newNatPmpMapper*(): NatPmpMapper {.raises: [ResourceExhaustedError].} =
+  let ctx = createShared(NatPmpWorkerCtx, 1)
+  ctx.reqSignal = ThreadSignalPtr.new().valueOr:
+    freeShared(ctx)
+    raise newException(ResourceExhaustedError, "NatPmpMapper reqSignal: " & error)
+  ctx.respSignal = ThreadSignalPtr.new().valueOr:
+    discard ctx.reqSignal.close()
+    freeShared(ctx)
```
vladopajic: `destroyCtx`

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,368 @@
...
+
+method unmap*(
+    self: NatPmpMapper, externalPort: Port, proto: MapProto
+): Future[Result[void, string]] {.async: (raises: [CancelledError]), gcsafe.} =
+  # NAT-PMP identifies a mapping by (internal port, protocol). The PortMapper
+  # API only carries the external port through unmap(), so look up the
+  # internal port we recorded when map() created the rule.
+  let key: MappingKey = (uint16(externalPort), proto)
+  if not self.mappings.hasKey(key):
+    return Result[void, string].err(
```
vladopajic: `Result[void, string]` can we avoid this, here and elsewhere 
`Result[Port, string]`, etc..

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,368 @@
...
+      mapProto: proto,
+      mapLease: lease,
+    )
+  )
+
+  if resp.success:
+    let mapped = Port(resp.externalPort)
+    self.mappings[(uint16(mapped), proto)] = uint16(internalPort)
+    return Result[Port, string].ok(mapped)
+  Result[Port, string].err(resp.getError())
```
vladopajic:
```suggestion
  if not resp.success:
     err(resp.getError())

  self.mappings[(resp.externalPort, proto)] = internalPort.uint16
  ok(Port(resp.externalPort))
```

nits:
- all code in this file generally does early return with error, and successfully outcomes are when all errors are cleared out, would be nice to adhere to that here. not only here but this is generally in programming this is how code flow is, adopting this practice in general will always result in better code.
- line brake helps code breath, especial for nim
- unnecessary type conversions 
- using `.uint16` after value to avoid parentheses (not sure if this works for port but would be nice here as well)

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,368 @@
...
+
+  let drain = self.drainPending
+  if drain != nil:
+    drain.cancelSoon()
+
+  self.ctx.request = NatPmpRequest(kind: nrShutdown)
+  let fr = self.ctx.reqSignal.fireSync()
+  if fr.isErr() or not fr.get():
+    warn "NatPmpMapper shutdown signal failed",
+      err = (if fr.isErr(): fr.error() else: "timeout")
```
vladopajic: this line `err =` took me quite some time to understand what it is. 

`else: "timeout")` is the strangest part, just set it to error whatever error is, and "timeout" should be set where error originates not here

libp2p/services/nat/natpmp_mapper.nim
```diff
@@ -0,0 +1,368 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
```
vladopajic, line 1: code looks plausible correct, but without tests its hard to see if it will really work

vladopajic (approved):

PR #2575, opened 2026-05-28
feat(nat): NATService

gmelodie:
## Summary
  - Extends `natservice.nim` with `natModeUpnp`/`natModeNatPmp`,
  `upnpConfig`/`natPmpConfig` helpers, `setupMappings`, and a refresh loop that
   re-issues mappings before the lease expires
  - Updates `builders.nim`: `withNAT` accepts an optional `PortMapperFactory`
  (production picks the UPnP/NAT-PMP backend matching `config.mode`; tests
  inject mocks)
  - Mappings hook into `peerInfo.addressMappers`, so post-bind listen addrs get
   rewritten to public `(extIp, extPort)` multiaddrs

  Part 5 of 6.

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -41,8 +41,7 @@ export
   switch, peerid, peerinfo, peeraddrpolicy, connection, multiaddress, crypto, errors,
   TLSPrivateKey, TLSCertificate, TLSFlags, ServerFlags, connmanager.ConnectionLimits,
   connmanager.maxTotal, connmanager.maxInOut, natservice.NATConfig, natservice.NATMode,
-  natservice.PortMapperFactory, natservice.upnpConfig, natservice.natPmpConfig,
-  natservice.NATService
+  natservice.PortMapperFactory, natservice.NATService
 
 const MemoryAutoAddress* = memorytransport.MemoryAutoAddress
 
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -41,7 +41,8 @@ export
   switch, peerid, peerinfo, peeraddrpolicy, connection, multiaddress, crypto, errors,
   TLSPrivateKey, TLSCertificate, TLSFlags, ServerFlags, connmanager.ConnectionLimits,
   connmanager.maxTotal, connmanager.maxInOut, natservice.NATConfig, natservice.NATMode,
-  natservice.PortMapperFactory, natservice.NATService
+  natservice.PortMapperFactory, natservice.NATService, natservice.upnpConfig,
+  natservice.natPmpConfig
 
 const MemoryAutoAddress* = memorytransport.MemoryAutoAddress
 
@@ -548,12 +549,7 @@ proc setupServices(b: SwitchBuilder, switch: Switch) {.raises: [LPError].} =
     switch.services.add(HPService.new(autonatService, autoRelayService))
 
   b.natConfig.withValue(natCfg):
-    let svc =
-      if b.natPortMapperFactory.isNil:
-        NATService.new(natCfg)
-      else:
-        NATService.new(natCfg, b.natPortMapperFactory)
-    switch.services.add(svc)
+    switch.services.add(NATService.new(natCfg, b.natPortMapperFactory))
 
   if b.identifyPusherEnabled:
     switch.services.add(IdentifyPusher.new())
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -373,12 +373,11 @@ proc withHolePunching*(
   )
   b
 
-when defined(libp2p_autotls_support):
-  proc withAutotls*(
-      b: SwitchBuilder, config: AutotlsConfig = AutotlsConfig.new()
-  ): SwitchBuilder =
-    b.autotlsConfig = Opt.some(config)
-    b
+proc withAutotls*(
+    b: SwitchBuilder, config: AutotlsConfig = AutotlsConfig.new()
+): SwitchBuilder =
+  b.autotlsConfig = Opt.some(config)
+  b
 
 proc withCircuitRelay*(b: SwitchBuilder, r: Relay = Relay.new()): SwitchBuilder =
   if r.isNil:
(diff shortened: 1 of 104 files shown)


libp2p/builders.nim
```diff
@@ -541,7 +548,12 @@ proc setupServices(b: SwitchBuilder, switch: Switch) {.raises: [LPError].} =
     switch.services.add(HPService.new(autonatService, autoRelayService))
 
   b.natConfig.withValue(natCfg):
-    switch.services.add(NATService.new(natCfg))
+    let svc =
+      if b.natPortMapperFactory.isNil:
+        NATService.new(natCfg)
+      else:
+        NATService.new(natCfg, b.natPortMapperFactory)
+    switch.services.add(svc)
```
vladopajic: this is not ideal. if you have to manually work around arguments for code that is 100% added by you then there is design issue. design code so that it suits you.

here it seems strange that second argument can't be nil but it can be left out. default of left out should be nil.


--

this could should have been something like: 
```
let svc = NATService.new(natCfg, b.natPortMapperFactory)
switch.services.add(svc)
```

or even one line

libp2p/services/natservice.nim
```diff
@@ -5,35 +5,77 @@
...
-    of natModeExplicitIp:
+    of ExplicitIp:
       explicitIp*: IpAddress
-    of natModeAuto:
+    of Auto:
       discard
+    of Upnp, NatPmp:
+      refreshInterval*: Duration
+      discoveryTimeout*: Duration
+      leaseDuration*: uint32
```
vladopajic: uint32 -> Duration

libp2p/services/natservice.nim
```diff
@@ -5,35 +5,77 @@
...
+    portMapperFactory: PortMapperFactory
+    mapper: PortMapper
+    refreshLoopFut: Future[void]
+    mappedPorts: seq[(Port, MapProto)]
+    externalIp*: Opt[IpAddress]
+
+const
+  DefaultRefreshInterval* = 30.minutes
+  DefaultDiscoveryTimeout* = 10.seconds
+  DefaultLeaseDuration*: uint32 = 3600
```
vladopajic: where is this used?

libp2p/services/natservice.nim
```diff
@@ -5,35 +5,77 @@
...
+
+proc defaultPortMapperFactory(mode: NATMode): PortMapper {.gcsafe, raises: [].} =
+  try:
+    case mode
+    of Upnp:
+      newUpnpMapper()
+    of NatPmp:
+      newNatPmpMapper()
```
vladopajic: we should stick to some convention:
- newMyType() or 
- MyType.new()

libp2p/services/natservice.nim
```diff
@@ -56,31 +98,212 @@ proc explicitIpMapped*(
...
+    discard
+  of ExplicitIp:
+    if self.addressMapper != nil:
+      switch.peerInfo.addressMappers.keepItIf(it != self.addressMapper)
+    # Do not touch peerInfo.announcedAddrs here: those may have been set by the
+    # user via withAnnouncedAddresses, and triggering peerInfo.update during
+    # shutdown can cause observers (e.g. IdentifyPusher) to broadcast while the
+    # switch is tearing down.
+  of Upnp, NatPmp:
+    if self.refreshLoopFut != nil and not self.refreshLoopFut.finished:
```
vladopajic: `self.refreshLoopFut.finished:` unnecessary

libp2p/services/natservice.nim
```diff
@@ -56,31 +98,212 @@ proc explicitIpMapped*(
...
+  while true:
+    try:
+      await sleepAsync(self.config.refreshInterval)
+    except CancelledError:
+      return
+
+    try:
+      await switch.peerInfo.update()
+    except CancelledError:
+      return
```
vladopajic: is it necessary to catch CancelledError at all?

libp2p/services/natservice.nim
```diff
@@ -56,31 +98,212 @@ proc explicitIpMapped*(
...
+    self.addressMapper = proc(
+        listenAddrs: seq[MultiAddress]
+    ): Future[seq[MultiAddress]] {.async: (raises: [CancelledError]).} =
+      let announced = await self.setupMappings(listenAddrs)
+      # If we got nothing back (e.g. discovery failed, no private listenAddrs)
+      # fall back to passing the listenAddrs through unchanged so the next
+      # mapper in the chain has something to work with.
+      if announced.len == 0:
+        return listenAddrs
+      return announced
```
richard-ramos, line 345: The mappers should be setup in `start` proc, otherwise when you `stop()` and then `start()` again, the mappers will be gone

libp2p/services/natservice.nim
```diff
@@ -56,31 +98,212 @@ proc explicitIpMapped*(
...
+    return Opt.none(MultiAddress)
+  let srcTa = initTAddress(listenAddr).valueOr:
+    return Opt.some(withIp)
+  if srcTa.port == externalPort:
+    return Opt.some(withIp)
+  replaceTransportPort(withIp, externalPort)
+
+proc findMappableListenPorts(listenAddrs: seq[MultiAddress]): seq[ListenPort] =
+  listenAddrs
+    .filterIt(it.isPrivateMA)
```
richard-ramos, line 187: we gotta filter out ipv6 addresses as that is not supported in natpmp, and for upnp, the readme says it's still [pending](https://github.com/status-im/nim-nat-traversal#todo).

libp2p/services/natservice.nim
```diff
@@ -5,35 +5,77 @@
...
+
+proc defaultPortMapperFactory(mode: NATMode): PortMapper {.gcsafe, raises: [].} =
+  try:
+    case mode
+    of Upnp:
+      newUpnpMapper()
+    of NatPmp:
+      newNatPmpMapper()
+    else:
+      nil
```
richard-ramos: nit: perhaps this proc should return an Opt[PortMapper] so we dont have to deal with nils

libp2p/services/natservice.nim
richard-ramos: The self.addressMapper should also be moved to start

richard-ramos (approved):

vladopajic (approved):

PR #2577, opened 2026-05-28
feat(LPProtocol): add metrics

vladopajic:
adds metrics to LPProtocol for:
- protocol stream budget rejections 
- protocol stream instances currently open

due to issues with metrics library test are in separate pr. this code has been tested with fixed version from local.

### DIFF
--- a/libp2p/protocols/protocol.nim
+++ b/libp2p/protocols/protocol.nim
@@ -4,11 +4,29 @@
 {.push raises: [].}
 
 import std/[tables]
-import chronos, results
+import chronos, results, metrics
 import ../stream/connection
 import ../utils/opt
 
 export results
+
+declareCounter(
+  libp2p_protocol_stream_cap_rejections_total,
+  "protocol stream budget rejections",
+  labels = ["protocol", "direction", "scope"],
+)
+
+declareGauge(
+  libp2p_protocol_streams_open,
+  "protocol stream instances currently open",
+  labels = ["protocol", "direction"],
+)
+
+export libp2p_protocol_stream_cap_rejections_total, libp2p_protocol_streams_open
+
+const
+  scopeTotal = "total"
+  scopePeerPeer = "per_peer"
 
 type
   LPProtoHandler* = proc(stream: Stream, proto: string): Future[void] {.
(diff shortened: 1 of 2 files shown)


gmelodie (approved):

libp2p/protocols/protocol.nim
```diff
@@ -81,76 +99,104 @@ proc new*(
...
+  let pb = budget.perPeerIncoming[peerId]
+  if pb == 0:
+    return # there was no reserve for this peer
+  elif pb == 1:
     budget.perPeerIncoming.del(peerId)
+  else:
+    budget.perPeerIncoming.inc(peerId, -1)
+
+  budget.totalIncoming.dec
+  libp2p_protocol_streams_open.dec(labelValues = [p.codec, "in"])
```
richard-ramos, line 166: Maybe this should be in line 150 like this:

```
proc releaseIncoming*(p: LPProtocol, peerId: PeerId) =
  defer:
        libp2p_protocol_streams_open.dec(labelValues = [p.codec, "in"])
```

because otherwise, if you dont set a budget, this metric wont work

libp2p/protocols/protocol.nim
```diff
@@ -81,76 +99,104 @@ proc new*(
...
+  if not canAccept:
+    libp2p_protocol_stream_cap_rejections_total.inc(
+      labelValues = [p.codec, "out", scope]
+    )
+    return false
+
   budget.totalOutgoing.inc
   budget.perPeerOutgoing.inc(peerId)
+  libp2p_protocol_streams_open.inc(labelValues = [p.codec, "out"])
   return true
```
richard-ramos, line 190: Similar to prev comment so the metric works regardless of budget existing or not

libp2p/protocols/protocol.nim
vladopajic: i see, it shouldn't dec in all cases like when `pb == 0`. 

ill add manually when there is no budget

vladopajic: or something else...

vladopajic: `if budget.isNil:` was unnecessary

vladopajic: no wait

libp2p/protocols/protocol.nim
```diff
@@ -81,76 +99,87 @@ proc new*(
...
+  let pb = p.streamBudget.perPeerOutgoing[peerId]
+  if pb == 0:
+    return # there was no reserve for this peer
+  elif pb == 1:
+    p.streamBudget.perPeerOutgoing.del(peerId)
+  else:
+    p.streamBudget.perPeerOutgoing.inc(peerId, -1)
+
+  p.streamBudget.totalOutgoing.dec
+  libp2p_protocol_streams_open.dec(labelValues = [p.codec, "out"])
```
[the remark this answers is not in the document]
vladopajic: riveted this change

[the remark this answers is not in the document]
vladopajic: riveted this change

vladopajic:
^ apparently there is issue with `MultistreamSelect` where `addHandler` could add protocol without codes. this should not be allowed. this should be handled separately.

richard-ramos (approved):

PR #2578, opened 2026-05-29
fix(service-disco): fix accepted time value for invalid ticket

### LINKED ISSUE #2542: service-disco: invalid retry tickets are not rejected and preserve forged tInit in registrar-signed tickets
### Description

When a `REGISTER` request carries a ticket that fails validation (bad signature, mismatched ad, or outside the retry window), `processRetryTicket` silently returns the original `t_wait` instead of signaling a rejection. The `registration` flow then falls through to the `Wait` path and issues a fresh registrar-signed ticket.

The problem is that the new ticket's `tInit` is copied from the incoming unvalidated ticket via `tInitOrDefault`:

https://github.com/vacp2p/nim-libp2p/blob/0ab8ca2dd6bdfdca2b2ad4a42ed963f2fc6879da/libp2p/protocols/service_discovery/registrar.nim#L407-L411

https://github.com/vacp2p/nim-libp2p/blob/0ab8ca2dd6bdfdca2b2ad4a42ed963f2fc6879da/libp2p/protocols/service_discovery/registrar.nim#L466-L470

An attacker can forge a ticket with `tInit` far in the past and an invalid signature. The registrar ignores the bad signature but copies the forged `tInit` into a new registrar-signed ticket. On retry, `totalWaitSoFar = now - forgedTInit` is massive, which drives the remaining wait time negative and results in immediate `Confirmed`.

**This is an edge case.** It only matters when `t_wait` increases significantly between the first attempt and the retry, which requires the cache to fill up during that window. In normal operation with low cache occupancy, `t_wait` is near zero and the exploit has no effect. The scenario becomes relevant under adversarial conditions where an attacker is actively trying to flood the registrar cache. That is exactly when the waiting time back-pressure is supposed to protect the registrar, and the forged `tInit` neutralizes it: a legitimate peer may need multiple retry cycles as `t_wait` keeps growing, while the attacker gets guaranteed one-retry admission.

### Spec Recommendation

The spec says to respond with `Rejected` when any ticket validation check fails. The three checks are: valid registrar signature, matching advertisement, and retry within the registration window.

[spec L972-L981](https://github.com/logos-co/logos-lips/blob/f1017b9f30cfdf8cb4f53d5f3286e4bdf0bdfcd1/docs/anoncomms/raw/logos-service-discovery.md?plain=1#L972-L981)

### Implementation

`processRetryTicket` returns `t_wait` unchanged on every failure path instead of signaling rejection:

https://github.com/vacp2p/nim-libp2p/blob/0ab8ca2dd6bdfdca2b2ad4a42ed963f2fc6879da/libp2p/protocols/service_discovery/registrar.nim#L239-L270

`registration` has no way to distinguish "no ticket" from "invalid ticket" and falls through to the `Wait` path in both cases:

https://github.com/vacp2p/nim-libp2p/blob/0ab8ca2dd6bdfdca2b2ad4a42ed963f2fc6879da/libp2p/protocols/service_discovery/registrar.nim#L447-L487

Two things need to change:
1. `processRetryTicket` (or `registration`) should return `Rejected` when a provided ticket fails any validation check.
2. `tInit` should only be carried forward from tickets that passed validation. On first attempt or rejected ticket, `tInit = now`.

SionoiS:
Fixes #2542

## Problem

In the service discovery registrar, a `REGISTER` request carrying an invalid retry ticket could still receive a fresh registrar-signed ticket.

When `processRetryTicket` encountered a ticket that failed validation (invalid signature, mismatched advertisement, or outside the retry window), it silently returned the original `t_wait`. The registration flow then proceeded to issue a new ticket, copying `tInit` from the unvalidated ticket via `tInitOrDefault`.

This allowed an attacker to forge a ticket with a `tInit` timestamp far in the past and an invalid signature. Although the registrar ignored the forged signature, it still propagated the forged `tInit` into a newly signed ticket. On retry, `totalWaitSoFar = now - forgedTInit` would be arbitrarily large, causing the peer to be confirmed immediately and bypassing the back-pressure mechanism intended to protect the registrar under cache pressure.

## Fix

* Added an `isValidTicket` helper and validate incoming tickets at the start of `registration`.
* Reject invalid tickets immediately with a `Rejected` response, as required by the specification.
* Only preserve `tInit` when the ticket passes the retry-window validation; otherwise, initialize `tInit` to `now`.
* Removed `tInitOrDefault`.
* Threaded `now` through `processRetryTicket` to ensure consistent time validation.

## Tests

Added component tests in:

`tests/libp2p/service_discovery/component/test_error.nim`

The new tests verify rejection of:

* Tickets with mismatched advertisements.
* Tickets with invalid signatures.
* Tickets outside the retry window.


libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -269,6 +265,24 @@ proc processRetryTicket*(
 
   return t_wait
 
+proc isTicketAuthentic(
+    disco: ServiceDiscovery, regMsg: RegisterMessage
+): bool {.raises: [].} =
+  let ticket = regMsg.ticket.valueOr:
+    return true
```
gmelodie: Should this not be false? Or error somehow?

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -479,12 +496,15 @@ proc registration*(disco: ServiceDiscovery, peerId: PeerId, inMsg: Message): Mes
...
-    advertisement: regMsg.advertisement,
-    tInit: regMsg.ticket.tInitOrDefault(now),
-    tMod: now,
-    tWaitFor: tWait,
-  )
+  var ticket =
+    Ticket(advertisement: regMsg.advertisement, tInit: now, tMod: now, tWaitFor: tWait)
+
+  if regMsg.ticket.isSome():
+    let t = regMsg.ticket.get()
```
gmelodie:
```suggestion
  if regMsg.ticket.withValue(t):
```

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -479,12 +496,15 @@ proc registration*(disco: ServiceDiscovery, peerId: PeerId, inMsg: Message): Mes
...
+    let windowStart = t.tMod + t.tWaitFor
+    let windowEnd = windowStart + disco.discoConfig.registrationWindow
+    if now >= windowStart and now <= windowEnd:
```
gmelodie: completely up to you, but this is equivalent (might make the <= and >= less explicit though)
```suggestion
    if now in windowStart..windowEnd:
```

gmelodie (approved):

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -269,6 +265,24 @@ proc processRetryTicket*(
...
+    return false
+
+  let registrarPubKey = disco.switch.peerInfo.privateKey.getPublicKey().valueOr:
+    error "Failed to get registrar public key", error
+    return false
+
+  if not ticket.verify(registrarPubKey):
+    return false
+
+  return true
```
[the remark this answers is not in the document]
richard-ramos: Seems like a valid concern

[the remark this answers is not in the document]
SionoiS: yes totally

tests/libp2p/service_discovery/component/test_register.nim
```diff
@@ -228,46 +220,20 @@ suite "Service Discovery Component - Register":
...
+    check registrarNode.countAdsInCache(serviceId) == 0
 
-    # The response ticket is registrar-signed but carries the unvalidated tInit.
-    check maliciousTicket.tInit == oldTInit
+    let maliciousResp2 =
+      await maliciousNode.sendRegister(registrarPeerId, serviceId, maliciousAdBytes)
+    check maliciousResp2.isOk()
+    let m2Status = maliciousResp2.get().status
+    check m2Status in
+      {kad_protobuf.RegistrationStatus.Wait, kad_protobuf.RegistrationStatus.Confirmed}
```
rlve: Test should be deterministic and assertions definitive.

richard-ramos (approved):

PR #2581, opened 2026-05-30
chore: add autonat v1 protobuf serialization powered by nim-protobuf-serialization

moigagoo:
## Summary

This PR replaces minprotobuf-based serialization for autonat v1 protocol with nim-protobuf-serialization-based one.

## Affected Areas
- autonat v1 type definitions
- autonat v1 client
- autonat v1 server
- autonat v1 tests

### DIFF
--- a/libp2p/protocols/connectivity/autonat/server.nim
+++ b/libp2p/protocols/connectivity/autonat/server.nim
@@ -31,15 +31,17 @@ proc sendResponseError(
 ) {.async: (raises: [CancelledError]).} =
   let pb = AutonatMsg(
     msgType: MsgType.DialResponse,
-    response: Opt.some(AutonatDialResponse(
-      status: status,
-      text:
-        if text == "":
-          Opt.none(string)
-        else:
-          Opt.some(text),
-      ma: Opt.none(MultiAddress),
-    ))
+    response: Opt.some(
+      AutonatDialResponse(
+        status: status,
+        text:
+          if text == "":
+            Opt.none(string)
+          else:
+            Opt.some(text),
+        ma: Opt.none(MultiAddress),
+      )
+    ),
   ).encode()
   try:
     await stream.writeLp(pb)
@@ -51,9 +53,11 @@ proc sendResponseOk(
 ) {.async: (raises: [CancelledError]).} =
   let pb = AutonatMsg(
     msgType: MsgType.DialResponse,
-    response: Opt.some(AutonatDialResponse(
-      status: ResponseStatus.Ok, text: Opt.some("Ok"), ma: Opt.some(ma)
-    ))
+    response: Opt.some(
+      AutonatDialResponse(
+        status: ResponseStatus.Ok, text: Opt.some("Ok"), ma: Opt.some(ma)
+      )
+    ),
   ).encode()
   try:
     await stream.writeLp(pb)
--- a/libp2p/protocols/connectivity/autonat/server.nim
+++ b/libp2p/protocols/connectivity/autonat/server.nim
@@ -29,28 +29,34 @@ type Autonat* = ref object of LPProtocol
 proc sendResponseError(
     stream: Stream, status: ResponseStatus, text: string = ""
 ) {.async: (raises: [CancelledError]).} =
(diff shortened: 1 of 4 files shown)


moigagoo:
@richard-ramos I really wanted to test the hypothesis that I proposed earlier that if nimlangserver works for nim-libp2p (specifically, the MCP server), the remaining tasks to integrate nim-protobuf-serialization can be delegated to AI.

[I fixed nimlangserver](https://github.com/nim-lang/langserver/pull/416) and created this PR with Gemini CLI. I'm not seeing any issues with it, looks solid. The only confusing decision is to use `AutonatMsg` instead of `AutonatDial` and `AutonatDialResponse` but that seems justified as the latter were used before as workarounds.

Could you please take a look and tell if this is viable? If it is, I think we can do the other protocols this way.

libp2p/protocols/connectivity/autonat/server.nim
```diff
@@ -29,28 +29,38 @@ type Autonat* = ref object of LPProtocol
...
+  let pb = AutonatMsg(
+    msgType: MsgType.DialResponse,
+    response: Opt.some(
+      AutonatDialResponse(
+        status: status,
+        text:
+          if text == "":
+            Opt.none(string)
+          else:
+            Opt.some(text),
```
vladopajic, line 41: wrapping string as `Opt` seems unnecessary if empty string `""` already means absence of value.
similarly wrapping `ref` types to `Opt` would also be pointless since we can do `.isNil` on ref types (not issue here just making a point).

libp2p/protocols/connectivity/autonat/types.nim
```diff
@@ -30,22 +35,22 @@ type
...
+    status* {.fieldNumber: 1, required, ext.}: ResponseStatus
+    text* {.fieldNumber: 2.}: Opt[string]
+    ma* {.fieldNumber: 3, ext.}: Opt[MultiAddress]
 
-  AutonatMsg* = object
-    msgType*: MsgType
-    dial*: Opt[AutonatDial]
-    response*: Opt[AutonatDialResponse]
+  AutonatMsg* {.proto2.} = object
+    msgType* {.fieldNumber: 1, required, ext.}: MsgType
```
richard-ramos, line 51: For lines 46 and 51, 

The [AutoNAT v1 spec](https://github.com/libp2p/specs/blob/master/autonat/autonat-v1.md#rpc-messages) defines the  `type` as `optional`, but this marks it as required. 

Same issue applies to `status`. 

In practice these fields should have been marked as required in the specs tho :(

arnetheduck: > required

https://protobuf.dev/programming-guides/style/#required

gmelodie: @arnetheduck do you mean we should never use `required` cuz it doesn't do anything, or that it's fine to leave it since it doesn't do anything?

arnetheduck: the code should not use required, but neither should protocols _in general_

libp2p/protocols/connectivity/autonat/server.nim
moigagoo: The return type is Opt though so it is expected that whatever you're getting it's an Opt.

Also, this is pre-existing code so changing this would probably break some code that consumes this.

vladopajic: > The return type is Opt though so it is expected that whatever you're getting it's an Opt.

using `Opt` wrapper for types that naturally have absence of value implemented in type system is error prone, cumbersome and doesn't really bring value.

 > changing this would probably break some code

perhaps, and fixing it could be nonetheless trivial relative to making the change

vladopajic (approved):

libp2p/protocols/connectivity/autonat/server.nim
gmelodie: I'm fine with whatever as long as integration tests work

gmelodie (approved):
A mazing!

PR #2590, opened 2026-06-02
test(conn-manager): watermark/scoring component 1

rlve:
## Summary

Adds a component test suite for the connection manager's watermark and peer-scoring behavior. These tests exercise the trim path end-to-end over QUIC switches, covering the watermark policy, grace/silence periods, and the scoring inputs that decide which peers survive a trim.

Test Cases:
- Watermark trims the node down to `lowWater`, pruning the oldest peer first
- Grace period exempts newly connected peers, both fully (all peers young) and partially (older peer trimmed, recent ones kept)
- Silence period throttles back-to-back trims
- Protected peers survive a trim
- Higher-scored peers (static tag) survive a trim
- Outbound peers survive over inbound ones via `outboundBonus`

## Affected Areas

  - [x] Tests

  ## Impact on Library Users

  None.

  ## Risk Assessment

  None.


tests/libp2p/test_conn_manager_component.nim
```diff
@@ -0,0 +1,225 @@
...
+proc newWatermarkSwitch(
+    lowWater: int,
+    highWater: int,
+    gracePeriod: Duration = 0.seconds,
+    silencePeriod: Duration = 0.seconds,
+    outboundBonus: int = 0,
+    decayResolution: Duration = 1.minutes,
+    maxConnections: int = 0,
+): Switch {.raises: [LPError].} =
+  var builder = makeStandardSwitchBuilder("", TransportType.QUIC)
```
vladopajic:
```suggestion
  var builder = makeStandardSwitchBuilder(QuicAutoAddress)
```

tests/libp2p/test_conn_manager_component.nim
```diff
@@ -0,0 +1,225 @@
...
+    .withWatermarkPolicy(lowWater, highWater, gracePeriod, silencePeriod)
+    .withPeerScoring(
+      PeerScoring(outboundBonus: outboundBonus, decayResolution: decayResolution)
+    )
+  if maxConnections > 0:
+    builder = builder.withMaxConnections(maxConnections)
+  builder.build()
+
+proc newSwitch(): Switch {.raises: [LPError].} =
+  makeStandardSwitch("", TransportType.QUIC)
```
vladopajic:
```suggestion
  makeStandardSwitch(QuicAutoAddress)
```

tests/libp2p/test_conn_manager_component.nim
vladopajic: or just `()` it will create quic by default

tests/libp2p/test_conn_manager_component.nim
vladopajic: or `()`

vladopajic (approved):

libp2p/connmanager.nim
```diff
@@ -718,8 +718,9 @@ proc triggerTrim(c: ConnManager) {.gcsafe, raises: [].} =
     return
 
   c.watermark.withValue(wm):
-    if Moment.now() - c.lastTrim < wm.silencePeriod:
-      return
+    c.lastTrim.withValue(lastTrim):
+      if Moment.now() - lastTrim < wm.silencePeriod:
+        return
```
gmelodie, line 723: Is this a fix you found with the tests? If not, can we keep it not an opt? If so, nice find!

tests/libp2p/test_conn_manager_component.nim
```diff
@@ -0,0 +1,225 @@
...
+    .withWatermarkPolicy(lowWater, highWater, gracePeriod, silencePeriod)
+    .withPeerScoring(
+      PeerScoring(outboundBonus: outboundBonus, decayResolution: decayResolution)
+    )
+  if maxConnections > 0:
+    builder = builder.withMaxConnections(maxConnections)
+  builder.build()
+
+proc newSwitch(): Switch {.raises: [LPError].} =
+  makeStandardSwitch()
```
gmelodie:
```suggestion
```

tests/libp2p/test_conn_manager_component.nim
```diff
@@ -0,0 +1,225 @@
...
+
+proc newSwitches(count: int): seq[Switch] {.raises: [LPError].} =
+  (0 ..< count).mapIt(newSwitch())
+
+template startAndDeferStop(switches: seq[Switch]): untyped =
+  await allFuturesRaising(switches.mapIt(it.start()))
+  defer:
+    await allFuturesRaising(switches.mapIt(it.stop()))
```
gmelodie: Is this not defined somewhere else? Try not to redefine it here

tests/libp2p/test_conn_manager_component.nim
```diff
@@ -0,0 +1,225 @@
...
+
+proc connect(dialer, listener: Switch) {.async.} =
+  await dialer.connect(listener.peerInfo.peerId, listener.peerInfo.addrs)
```
gmelodie, line 32: Same as above, this seems like it could/should be reused by other testing modules

tests/libp2p/test_conn_manager_component.nim
```diff
@@ -0,0 +1,225 @@
...
+  # short wait between connects
+  await sleepAsync(50.millis)
+
+proc peerCount(s: Switch): int =
+  s.connManager.getConnections().len
```
gmelodie, line 37: Let's not hide the logic behind functions unnecessarily. It's somewhat of a personal preference I think, but I'd prefer not having to chase down `peerCount` to see what it actually gets (i.e. gets connection count from connManager)
```suggestion
```

gmelodie (approved):

richard-ramos (approved):

libp2p/connmanager.nim
rlve: Yes, this is the fix specifically.

tests/libp2p/test_conn_manager_component.nim
rlve: It wasn't defined for bare switches before. Improved in: https://github.com/vacp2p/nim-libp2p/pull/2590/changes/7a07b7199cca6461aeb351ea4a91f558319e228b

tests/libp2p/test_conn_manager_component.nim
rlve: We do that in tests all the time. Repeating construction like `s.connManager.getConnections().len` multiple times gets very tedious.

libp2p/connmanager.nim
richard-ramos: should we get this in release 2.0.0?

rlve: @richard-ramos It's a fix for an edge case.

Here is an explanation from Claude:

```
I've got the full diagnosis confirmed. Here's why.

  Root cause: lastTrim starts at zero, and chronos Moment.now() is time-since-boot

  The silence-period gate in connmanager.nim:714:

  proc triggerTrim(c: ConnManager) =
    if not c.trimFut.isNil and not c.trimFut.finished: return
    c.watermark.withValue(wm):
      if Moment.now() - c.lastTrim < wm.silencePeriod:   # <-- line 721
        return
      c.trimFut = c.trimConnections()

  lastTrim: Moment (connmanager.nim:116) is never initialized, so it defaults to Moment(value: 0). The gate is supposed to throttle only the second trim, but on the very first call it computes Moment.now() - 0, i.e. the
  absolute clock value itself.

  Now the clock. Chronos defaults to -d:asyncTimer=mono (confirmed in chronos/timer.nim:57). On Linux that's clock_gettime(CLOCK_MONOTONIC) — nanoseconds since system boot, not since epoch. So:

  - CI runs in a freshly-booted runner/container. Uptime is a few minutes. Moment.now() ≈ a few minutes in nanoseconds. With silencePeriod = 1.hours, the condition Moment.now() - 0 < 1.hours is true → the first trim is 
  suppressed. The node never trims to lowWater, so checkUntilTimeout: node.peerCount == lowWater (line 140) hits [TIMEOUT]. Exactly the failure you pasted.
  - Locally your box has up 11 days. Moment.now() ≈ 11 days in ns, which is > 1.hours, so the first trim runs normally and the test passes.

  That's also why only this test fails: it's the only one using a large silencePeriod. The grace-period tests use connectedAt timestamps (relative deltas), so they're immune.

  This is an implementation bug, not a test bug
```

PR #2591, opened 2026-06-02
test(conn-manager): watermark/scoring component 2

rlve:
## Summary

Extends the connection-manager component suite with cases for protection lifecycle, decaying scores, and the interaction between the watermark trim and the hard connection cap.

New cases:
- `unprotect` re-exposes a peer so a later trim can prune it
- Multi-tag protection holds until every tag is removed (`isProtected` flips only on the last `unprotect`)
- A decaying tag (`tagPeerDecaying` + `decayFixed`) keeps a peer until its score decays to zero, then the next trim drops it
- Trim stops short of `lowWater` when more than `lowWater` peers are protected, settling at the protected count
- Hard cap and watermark running together: trimmed connections release their semaphore slots, and available inbound slots settle at `cap - lowWater - 1` (accept loop keeps one slot pre-acquired)
- Hard cap rejects new connections when the trim has no prunable candidates: inbound `getIncomingSlot` blocks, outbound `getOutgoingSlot` raises `TooManyConnectionsError`
- Dial fails with `DialFailedError` when reaching `lowWater` forces the trim to drop the dialing peer's own freshly stored connection

## Affected Areas

- [x] Tests

## Impact on Library Users

None.

## Risk Assessment

None.

### DIFF
--- a/tests/libp2p/test_conn_manager_component.nim
+++ b/tests/libp2p/test_conn_manager_component.nim
@@ -215,3 +215,237 @@ suite "Connection Manager Watermark/Scoring Component":
       node.peerCount == lowWater
 
     check node.isConnected(peers[0].peerInfo.peerId)
+
+  asyncTest "unprotect re-exposes a peer to trimming":
+    # peers[0] is the oldest peer, the first to be trimmed.
+    # protection keeps it through the first trim.
+    # once unprotected the second trim drops it.
+    const
+      lowWater = 2
+      highWater = 3
+    let node = newWatermarkSwitch(lowWater, highWater)
+    # twice highWater peers so a second batch can trigger a second trim
+    let peers = newSwitches(2 * highWater)
+    let all = @[node] & peers
+
+    startAndDeferStop(all)
+
+    for peer in peers[0 ..< highWater]:
+      await connect(peer, node)
+
+    # protect peers[0] before the trim-triggering dial
+    node.connManager.protect(peers[0].peerInfo.peerId, "keep")
+
+    await connect(peers[highWater], node)
+
+    # the first trim settles at lowWater with protected peers[0] still connected
+    checkUntilTimeout:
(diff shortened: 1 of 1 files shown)


tests/libp2p/test_conn_manager_component.nim
```diff
@@ -223,3 +223,237 @@ suite "Connection Manager Watermark/Scoring Component":
...
+      node.connManager.availableSlots(Direction.In) == 0
+
+    # with the cap full the semaphore rejects new connections
+    check not (await node.connManager.getIncomingSlot().withTimeout(100.millis))
+    expect TooManyConnectionsError:
+      discard node.connManager.getOutgoingSlot()
+
+  asyncTest "dial fails when the trim prunes the new connection":
+    # reaching lowWater can force the trim to drop the dialing peer's own freshly stored connection
+    # the dial then fails with DialFailedError instead of connecting and being trimmed cleanly later
```
rlve, line 433: @vladopajic have a look on this case, is this acceptable or should I create an issue?

vladopajic: ~~it works as expected; but questions is should it be improved~~

is this it: 
when node drops connection on its side, other node must also drop same connection (connection has to have to points). 
if this is is then it's fine

vladopajic (approved):

tests/libp2p/test_conn_manager_component.nim
rlve: I was rather thinking if this should be handled gracefully, connection allowed to happen and then trimmed? Anyway this is an edge case and happens because of `gracePeriod == 0`.

vladopajic: that is fine.because all connections are scored when trimming.

gmelodie (approved):

richard-ramos (approved):

PR #2592, opened 2026-06-02
fix(service-disco): service discovery key bucketing 

### LINKED ISSUE #2499: service-disco: valid service-table peers are dropped when Kad bucketIndex is 16 or higher

SionoiS:
Fixes #2499

## Problem

`ServiceDiscovery` builds service-specific routing tables (`AdvT`, `DiscT`, and `RegT`) with `m = 16` buckets. However, the generic Kademlia routing table indexes peers directly by `leadingZeros` and rejects any peer whose bucket index is `>= maxBuckets`.

As a result:

* Peers with a Kademlia bucket index `>= 16` are silently dropped.
* Peers that are retained are placed into incorrect service buckets.

According to the service discovery specification, routing tables centered on `service_id_hash` must derive bucket indices using:

```
i = min(floor((lz * m) / 256), m - 1)
```

with `i = m - 1` when `d = 0`.

Under this mapping, a peer with `lz = 16` belongs in service bucket `1`, not outside the table.

## Impact

* Advertisers may miss valid registrar candidates and therefore send fewer registrations.
* Discoverers query incomplete routing tables.
* Registrars may return incorrect `closerPeers` results.
* Service buckets `1..15` contain incorrectly assigned peers while excluding peers that should belong to them.

## Fix

* `bucketIndex` now maps `leadingZeros` into the configured bucket range using the specification-defined scaling formula.
* The `idx >= maxBuckets` rejection in `insert` has been removed.
* `randomKeyInBucket` now selects target `lz` values consistent with the new bucket mapping, ensuring `refreshTable` queries the correct portion of the keyspace.
* The component test previously titled *"drops a known registrar at Kad bucketIndex 16"* has been updated to *"maps registrars via scaled bucketing"* and now verifies that a high-`lz` registrar is correctly included in the routing table and receives advertisements.


libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -208,18 +213,37 @@ proc isStale*(bucket: Bucket): bool =
       return true
   return false
 
-proc randomKeyInBucket*(selfId: Key, bucketIndex: int, rng: Rng): Key =
+proc randomKeyInBucket*(
+    selfId: Key, bucketIndex: int, rng: Rng, maxBuckets: int = DefaultMaxBuckets
+): Key =
+  let targetLz =
```
gmelodie: What does `targetLz` stand for? Can we have a more descriptive name?

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -208,18 +213,37 @@ proc isStale*(bucket: Bucket): bool =
...
+      # lz in [lo, hi] satisfy (lz * M) div 256 == bucketIndex
+      let lo = (bucketIndex * 256 + M - 1) div M
+      let hi = min(256, ((bucketIndex + 1) * 256 + M - 1) div M - 1)
+      if lo > hi:
+        bucketIndex * (256 div max(1, M))
+      else:
+        let span = hi - lo + 1
+        var r: array[1, byte]
+        rng.generate(r)
+        lo + (int(r[0]) mod span)
```
gmelodie: This is very cryptic
1. Encapsulate logic in functions with good descriptive names.
2. Why create a new variable `M`? I was a bit confused when reading and took me a bit to understand where it came from. Also it should not be capitalized. 

```nim
let targetLz = getTargetLz()
...

proc satisfiesThisCond(maxBuckets, bucketIndex) =
  maxBuckets == 256 or bucketIndex >= maxBuckets

proc doAnotherThing() =
      let span = hi - lo + 1
        var r: array[1, byte]
        rng.generate(r)
        lo + (int(r[0]) mod span)

proc doSthElse() = 
     let M = maxBuckets
      # lz in [lo, hi] satisfy (lz * M) div 256 == bucketIndex
      let lo = (bucketIndex * 256 + M - 1) div M
      let hi = min(256, ((bucketIndex + 1) * 256 + M - 1) div M - 1)
      if lo > hi:
        bucketIndex * (256 div max(1, M))
      else:
        doAnotherThing()

proc getTargetLz() = 
  if maxBuckets <= 1:
      0
  elif satisfiesThisCond(maxBuckets, bucketIndex): # <--- why max
    bucketIndex
  else:
    doSthElse()
```

tests/libp2p/service_discovery/component/test_lookup_get_ads.nim
```diff
@@ -34,40 +38,48 @@ proc setupRegistrarsInDistinctBuckets(
...
 
-  raiseAssert "could not find registrars in distinct buckets"
+  let svcName = "service"
+  let pk0 = getStaticPrivateKeyForBucket(svcName, 0)
+  # Get a key for a higher bucket (we have entries for 1).
+  var pk1 = pk0
+  for b in 1 ..< conf.bucketsCount:
+    if b in precomputedStaticBucketKeys[svcName] and
+        precomputedStaticBucketKeys[svcName][b].len > 0:
+      pk1 = getStaticPrivateKeyForBucket(svcName, b)
```
gmelodie: Nit: these variables and function names are too big, can you maybe reduce them?

tests/libp2p/service_discovery/component/test_lookup_get_ads.nim
```diff
@@ -34,40 +38,48 @@ proc setupRegistrarsInDistinctBuckets(
...
+      break
+  let n0 = setupServiceDiscoveryNode(discoConfig = conf, privateKey = Opt.some(pk0))
+  let n1 = setupServiceDiscoveryNode(discoConfig = conf, privateKey = Opt.some(pk1))
+  let bb0 = bucketOf(n0, serviceId)
```
gmelodie: Why `bb0` and not `b0`? Or better `bucket0`?

tests/libp2p/service_discovery/component/test_lookup_get_ads.nim
```diff
@@ -34,40 +38,48 @@ proc setupRegistrarsInDistinctBuckets(
...
-    registrarsByBucket[bucket].add(registrar)
-
-    if registrarsByBucket[bucket].len == count:
-      return registrarsByBucket[bucket]
-
-  raiseAssert "could not find enough registrars in one bucket"
+  doAssert count > 0, "count must be > 0"
+  let svcName = "service"
+  # Find a bucket that has enough precomputed keys (will be 0 for "service").
+  var chosenB = -1
```
gmelodie: nit
```suggestion
  var chosenBucket = -1
```

tests/libp2p/service_discovery/utils.nim
```diff
@@ -154,3 +154,107 @@ proc containsPeer*(
...
+    "08011240CD04A0C51250F7A2545C61FC084E160BFD01E760D7D88016111F8CE175890D605D25CDE8DFA9C3EAA70D9C8D8AD4F7F7792C63CD3C83AA8F8FEF031103A91261",
+    "08011240DD6480FA1CC5CF3C86E496CAC93093F8F9CC2C85C212E4C84DDB59DA6E407AE54A0ACF94451A00CED3C59FDBCFCBBE28F780637A4B984153D4EB72A566C88DD1",
+  ]
+  inner0[1] = @[
+    "0801124021B12E50EF13E4E40597DFAC04309E811969A9E76DB0130E7837AF3AA2D1E95F4726D31731A6956F8FE7CA59DFFC918A323758D6ACCE01ECDBC4F00C974BC197",
+    "080112401A9447EB01AF2E716212E19EF7A217A3629226AF7A25977CCE2E11EB55CA83B58BDB4B193C2A0E0B055C07C943A610AA6DD8755CDB7BEE9061300FB1980696A9",
+  ]
+  var m = initTable[string, Table[int, seq[string]]]()
+  m["service"] = inner0
+  m
```
gmelodie: This is really ugly. I don't think it should be an issue to generate this every time, should it? Unless this takes an absurd amount of time, just have it generate on the fly for every test.

gmelodie (changes requested):
I think we should have a function that maps raw kad 256 space to 16 service table space, and another that does the reverse, then just use this throughout. Here we're embedding the logic inside `randomKeyInBucket`.  

Asking for changes mostly because of the commented out testing stuff, otherwise this looks ok!

tests/libp2p/service_discovery/utils.nim
SionoiS: Unfortunately, it does take a long time. On my powerful desktop that one test file takes 90s to complete. In CI it would take multiple minutes.

libp2p/protocols/kademlia/routing_table.nim
SionoiS: lz = leading zeros but yeah not the best name.

libp2p/protocols/kademlia/routing_table.nim
SionoiS: AI sucks at this... I'll have to go full artisan mode.

tests/libp2p/service_discovery/utils.nim
richard-ramos: It's probably an overkill, but it would probably be nicer if instead of having the generator code commented there, you had it on a separate nim script, that once executed it would generate in a new file containing the precomputed keys, and then import those keys.

SionoiS: > nim script, that once executed it would generate in a new file containing the precomputed keys, and then import those keys.

Is there already some code I could look at that does something similar?

gmelodie: @SionoiS yes! just put it under [`tools`](https://github.com/vacp2p/nim-libp2p/tree/master/tools) with a name of `generatorXYZ.nim` or something, have the script print to the screen. Also write the usage instructions somewhere e.g. docs

richard-ramos: @SionoiS , maybe something like this code generated by codex? 
https://github.com/vacp2p/nim-libp2p/compare/fix-kad-bucketing...codex/pr2592-static-key-generator?expand=1

I did not read it in detail to know if code style matches this repo, etc, but should work and serve to illustrate the idea

tests/libp2p/service_discovery/utils.nim
```diff
@@ -154,3 +154,107 @@ proc containsPeer*(
...
+  ]
+  var m = initTable[string, Table[int, seq[string]]]()
+  m["service"] = bucketKeysTable
+  m
+
+proc getKeyForBucket*(serviceName: string, bucket: int): PrivateKey =
+  ## Returns a PrivateKey (cycling through the precomputed list for the
+  ## (serviceName, bucket) pair) that is known to land in that scaled bucket
+  ## under the current bucketIndex rules used by service tables.
+  var counters {.threadvar.}: Table[(string, int), int]
```
richard-ramos: Why is a threadvar used here? 🤔 .
Nothing wrong with it, but just curious

richard-ramos (approved):
LGTM
IMO getting the commented code out of the test and having a generator create it should be done in this PR, but i dont consider it a blocker.

tests/libp2p/service_discovery/utils.nim
SionoiS: I reviewed where the keys are used and maybe we could just delete the generator. I don't think we will need more keys in the future. Much simpler this way.

tests/libp2p/service_discovery/utils.nim
SionoiS: overcomplicated way of not reusing the same keys. I changed it.

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -98,12 +106,9 @@ proc insert*(rtable: RoutingTable, nodeId: Key): bool =
     return false # No self insertion
 
   let idx = bucketIndex(
-    rtable.selfId, nodeId, rtable.config.hasher, rtable.config.selfIdPreHashed
+    rtable.selfId, nodeId, rtable.config.hasher, rtable.config.maxBuckets,
+    rtable.config.selfIdPreHashed,
```
gmelodie: We could even do `let idx = rtable.bucketIndex()`

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -208,32 +213,34 @@ proc isStale*(bucket: Bucket): bool =
...
-  let lsbMask = (1'u8 shl tgtBitInByte) - 1
-  if lsbMask != 0:
-    var rb: array[1, byte]
-    rng.generate(rb)
-    raw[tgtByte] = (raw[tgtByte] and not lsbMask) or (rb[0] and lsbMask)
+proc randomKeyInBucket*(
+    selfId: Key, bucketIndex: int, rng: Rng, maxBuckets: int = DefaultMaxBuckets
+): Key =
+  let
+    index = clamp(bucketIndex, 0, (maxBuckets - 1))
```
gmelodie, line 212: TIL about `clamp`!

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -208,32 +213,34 @@ proc isStale*(bucket: Bucket): bool =
...
 
-  # randomize remaining bytes
-  if tgtByte + 1 < raw.len:
-    rng.generate(raw.toOpenArray(tgtByte + 1, raw.len - 1))
+  # The bit at the boundary should be the opposite of selfId. So that the XOR is not 0.
+  if selfId[byteIdx].testBit(boundBitIdx) == key[byteIdx].testBit(boundBitIdx):
+    key[byteIdx].flipBit(boundBitIdx)
 
-  return raw
+  return key
```
gmelodie: nit
```suggestion
  key
```

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -208,32 +213,34 @@ proc isStale*(bucket: Bucket): bool =
...
+    leadingZeros = index * 256 div buckets
+    (byteIdx, rem) = divmod(leadingZeros, 8)
+    boundBitIdx = 7 - rem
+
+  var key = selfId
+
+  for i in byteIdx ..< IdLength:
+    rng.generate(key[i])
+
+  # From 0 to boundBitIdx the bits should be random.
```
gmelodie: Place this comment before the last `for`, it made me really confused seeing it here

tests/libp2p/service_discovery/test_routing_table_manager.nim
```diff
@@ -369,16 +369,56 @@ suite "ServiceRoutingTableManager - service id hashing":
...
+        serviceId,
+        peer,
+        Opt.none(XorDHasher),
+        DefaultMaxBuckets,
+        selfIdPreHashed = false,
+      )
+
+    check:
+      preHashBucket != doubleHashBucket
+      table.buckets.len > preHashBucket
```
gmelodie: What is this asserting?

tests/libp2p/service_discovery/test_routing_table_manager.nim
```diff
@@ -369,16 +369,56 @@ suite "ServiceRoutingTableManager - service id hashing":
 
     manager.insertPeer(serviceId, peer)
 
-    # The peer lands in the bucket determined by XOR(serviceId, H(peer)),
-    # not XOR(H(serviceId), H(peer))
+    # Service tables use selfIdPreHashed=true so that bucketIndex uses the
+    # already-hashed serviceId directly (as self), hashing only the peer key
+    # once. This test verifies that (for a chosen pair where single- vs
+    # double-hashing the serviceId would produce different buckets) the peer
+    # is placed according to the pre-hashed serviceId path.
```
gmelodie: One thing i hate about AI are these convoluted unnecessarily long comments:


```suggestion
```

gmelodie (commented):
>  IMO getting the commented code out of the test and having a generator create it should be done in this PR, but i dont consider it a blocker.

I agree with @richard-ramos here. Can you please include it in a script? Or at least open an issue and paste the code in a code block so we can implement it later?

gmelodie (approved):

tests/libp2p/service_discovery/test_routing_table_manager.nim
SionoiS: I refactored the test. it should be clearer now

PR #2594, opened 2026-06-02
fix(multistream): flaky peer limit test

gmelodie:
## Summary

Fix a race in `e2e - inbound per-peer limit` (`tests/libp2p/test_multistream.nim`) that was failing CI on the NATService PR (#2575) and others.

The test starts two concurrent dialers, then a third, expecting the third to be rejected because `maxIncomingStreamsPerPeer = 2`. All three streams race on `reserveIncoming` in `multistream.handle`, so any one of the three can lose the race. When the third dialer happens to win a slot, it blocks on the handler's `blocker.wait()` instead of getting `LPStreamEOFError`, tripping the 1s connector timeout — exactly the `Timeout while waiting for connector` failure seen on `macos-15-arm64 (Nim v2.2.4 / clang)`.

The fix gates the third dial on a `WaitGroup(2)` that the handler decrements on entry, so the first two reservations are guaranteed to land before the third dial starts. `makeBlockedHandler` takes a new optional `entered: WaitGroup` parameter so the other tests using it stay unchanged.

## Affected Areas

- [x] Other — test only (`tests/libp2p/test_multistream.nim`, suite `Multistream :: stream limits`)

No production code changed.

N/A — test-only change.

## Impact on Library Users

No impact — test-only change.

## Risk Assessment

Minimal. The change adds a synchronization point in a single `asyncTest` and threads an optional `WaitGroup` argument through a test-local helper (`makeBlockedHandler`). Stress-tested locally with 20 consecutive runs of `test_multistream`, all green.

## References

- Failing job: https://github.com/vacp2p/nim-libp2p/actions/runs/26835945447/job/79129733417
- Blocks: #2575 (feat(nat): NATService)

### DIFF
--- a/tests/libp2p/test_multistream.nim
+++ b/tests/libp2p/test_multistream.nim
@@ -652,10 +652,7 @@ suite "Multistream :: stream limits":
     await reserved.wait(5.seconds)
 
     expect LPStreamEOFError:
-      try:
-        await connector().wait(1.seconds)
-      except AsyncTimeoutError:
-        raiseAssert "Timeout while waiting for connector"
+      await connector()
 
     await dialers.cancelAndWait()
     await transport2.stop()
--- a/tests/libp2p/test_multistream.nim
+++ b/tests/libp2p/test_multistream.nim
@@ -552,7 +552,7 @@ suite "Multistream :: stream limits":
       let stream = await transport.accept()
       asyncSpawn acceptedOne(stream)
 
-  proc makeBlockedHandler(): LPProtoHandler =
+  proc makeBlockedHandler(entered: WaitGroup = nil): LPProtoHandler =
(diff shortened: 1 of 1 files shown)


tests/libp2p/test_multistream.nim
```diff
@@ -641,6 +649,8 @@ suite "Multistream :: stream limits":
     for _ in 0 ..< 2:
       dialers.add(connector())
 
+    await reserved.wait(5.seconds)
+
     expect LPStreamEOFError:
       try:
         await connector().wait(1.seconds)
```
vladopajic: or increase or remove timeout? that shouldn't really be there

vladopajic (approved):

richard-ramos (approved):

PR #2596, opened 2026-06-02
feat(nat): NATService

gmelodie:
## Summary

Wires the UPnP and NAT-PMP port-mapping backends from parts 1-4 into the
end-to-end NAT traversal flow. This is part 5 of 6.

- Extends `libp2p/services/natservice.nim` with two new `NATMode`s (`Upnp`,
  `NatPmp`), `upnpConfig` / `natPmpConfig` constructors, a `setupMappings`
  routine that discovers the IGD and requests a mapping for every private
  IPv4 listen address, and a `refreshLoop` that re-issues mappings before
  the lease expires (also invalidating the cached external IP so DHCP
  renewals / failovers are picked up).
- The address mapper rewrites each listen multiaddr in-place to the public
  `(extIp, extPort)` so it flows through `peerInfo.addressMappers` and is
  surfaced via identify like any other announced address. Falls back to
  passing listen addrs through unchanged when discovery yields nothing,
  so it composes cleanly with other mappers in the chain.
- `start` / `stop` are symmetric: backends are constructed in `start` (so
  a stop/start cycle re-creates them) and `stop` tears down mappings,
  closes the mapper, and clears the cached external IP.
- `builders.nim`: `withNAT` gains an optional `PortMapperFactory`
  parameter. Production picks the UPnP / NAT-PMP backend matching
  `config.mode` via `defaultPortMapperFactory`; tests inject mocks.
- Reopened from #2575 which was accidentally merged early.

Renames the existing `NATMode` enum values from `natModeAuto` /
`natModeExplicitIp` to `Auto` / `ExplicitIp` for consistency with the new
`Upnp` / `NatPmp` variants.

## Affected Areas

- [x] Peer Management / Discovery — new external-address path via NAT
  port mapping, surfaced through `peerInfo.addressMappers`.
- [x] Build / Tooling — `NATService.new` now takes an optional
  `PortMapperFactory`; `withNAT` exposes the same hook.
- [x] Other — `NATMode` enum value rename (`natModeAuto` →
  `Auto`, `natModeExplicitIp` → `ExplicitIp`).

Downstream impact is limited to the renamed enum values and the new
optional `withNAT` parameter (additive). No integration runs yet — will
be done as part of the part 6 wrap-up if needed.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

- **API addition:** `NATConfig` now supports `Upnp` and `NatPmp` modes
  with `upnpConfig(...)` / `natPmpConfig(...)` helpers exposing
  `refreshInterval`, `discoveryTimeout`, and `leaseDuration` knobs.
- **Behavior addition:** when one of those modes is configured, the
  switch performs IGD discovery on start, requests port mappings for
  each private IPv4 listen address, and announces public `(extIp,
  extPort)` multiaddrs in place of the private ones.
- **Breaking rename (small):** `NATMode.natModeAuto` /
  `natModeExplicitIp` → `NATMode.Auto` / `ExplicitIp`. Existing
  downstreams using these names need a trivial search/replace.
- `withNAT` gains an optional `portMapperFactory` argument; existing
  callers compile unchanged.

## Risk Assessment

- **Backward compatibility:** enum rename is source-breaking but
  trivial; no on-the-wire change.
- **Network behavior:** when an IGD is unreachable or mapping fails,
  the service warns and falls back to passing listen addrs through
  unchanged — no addresses are dropped silently. Non-IPv4 listen
  addresses are filtered out at the entry point because both backends
  only support IPv4 today.
- **Lifecycle:** `stop` unmaps everything before closing the backend,
  so the IGD shouldn't accumulate stale leases across restarts; the
  refresh loop also unmaps stale entries each cycle (e.g. when an IGD
  returns a different external port on the next `map()`).
- **Performance:** discovery and mapping run off the hot path inside
  the address-mapper chain and on a periodic heartbeat; no extra cost
  for `Auto` / `ExplicitIp` users.

## References

- Reopens #2575 (accidentally merged early)
- Builds on #2571 (portmapper interface) and parts 3 / 4 of the series

### DIFF
--- a/libp2p/services/natservice.nim
+++ b/libp2p/services/natservice.nim
@@ -3,7 +3,7 @@
 
 {.push raises: [].}
 
-import std/[net, sequtils, strutils]
+import std/[net, sequtils]
 import chronos, chronicles, results
 import ../[multiaddress, multicodec, wire]
 import ../switch
@@ -16,27 +16,27 @@ logScope:
 
 type
   NATMode* = enum
-    natModeAuto
+    Auto
       ## Use libp2p-native mechanisms (autonat / hole-punching). No port-mapping
       ## is performed here.
       ## TODO: In this skeleton it is a no-op; folded in by piece 3a.
-    natModeExplicitIp
+    ExplicitIp
       ## Static external IP, no probing. The configured ``explicitIp`` is combined
       ## with the bound listen addresses to produce the announced address set.
-    natModeUpnp
+    Upnp
       ## Discover an IGD via UPnP, request port mappings for each private
       ## listen address, and surface the public ``(extIp, extPort)`` multiaddrs
       ## through the peerInfo addressMappers chain. Mappings are refreshed
       ## periodically.
-    natModeNatPmp ## Same as ``natModeUpnp`` but using the NAT-PMP protocol.
+    NatPmp ## Same as ``Upnp`` but using the NAT-PMP protocol.
 
   NATConfig* = object
     case mode*: NATMode
-    of natModeExplicitIp:
+    of ExplicitIp:
       explicitIp*: IpAddress
-    of natModeAuto:
+    of Auto:
       discard
-    of natModeUpnp, natModeNatPmp:
+    of Upnp, NatPmp:
       refreshInterval*: Duration
       discoveryTimeout*: Duration
       leaseDuration*: uint32
@@ -53,40 +53,16 @@ type
     externalIp*: Opt[IpAddress]
 
 const
-  DefaultUpnpRefreshInterval* = 30.minutes
-  DefaultUpnpDiscoveryTimeout* = 10.seconds
-  DefaultUpnpLeaseDuration*: uint32 = 3600
-
-proc upnpConfig*(
-    refreshInterval = DefaultUpnpRefreshInterval,
-    discoveryTimeout = DefaultUpnpDiscoveryTimeout,
-    leaseDuration = DefaultUpnpLeaseDuration,
-): NATConfig =
-  NATConfig(
-    mode: natModeUpnp,
-    refreshInterval: refreshInterval,
-    discoveryTimeout: discoveryTimeout,
-    leaseDuration: leaseDuration,
-  )
-
-proc natPmpConfig*(
-    refreshInterval = DefaultUpnpRefreshInterval,
-    discoveryTimeout = DefaultUpnpDiscoveryTimeout,
-    leaseDuration = DefaultUpnpLeaseDuration,
-): NATConfig =
-  NATConfig(
-    mode: natModeNatPmp,
-    refreshInterval: refreshInterval,
-    discoveryTimeout: discoveryTimeout,
-    leaseDuration: leaseDuration,
-  )
+  DefaultRefreshInterval* = 30.minutes
+  DefaultDiscoveryTimeout* = 10.seconds
+  DefaultLeaseDuration*: uint32 = 3600
 
 proc defaultPortMapperFactory(mode: NATMode): PortMapper {.gcsafe, raises: [].} =
   try:
     case mode
-    of natModeUpnp:
+    of Upnp:
       newUpnpMapper()
-    of natModeNatPmp:
+    of NatPmp:
       newNatPmpMapper()
     else:
       nil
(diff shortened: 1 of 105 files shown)


vladopajic (approved):

libp2p/services/natservice.nim
```diff
@@ -56,31 +123,258 @@ proc explicitIpMapped*(
...
+    self.addressMapper = proc(
+        listenAddrs: seq[MultiAddress]
+    ): Future[seq[MultiAddress]] {.async: (raises: [CancelledError]).} =
+      let announced = await self.setupMappings(listenAddrs)
+      # If we got nothing back (e.g. discovery failed, no private listenAddrs)
+      # fall back to passing the listenAddrs through unchanged so the next
+      # mapper in the chain has something to work with.
+      if announced.len == 0:
+        return listenAddrs
+      return announced
```
richard-ramos, line 339: setupMappings will filter the private addresses and attempt to do the natpmp/upnp mapping. This is fine if you only have private addresses, since announced will return the new mappings.

However, if you already had a publicly reachable address before calling setupMappings, the return on line 345 will not contain those, so you end up missing addresses.

The return on line 345 must keep non private addresses too

libp2p/services/natservice.nim
```diff
@@ -56,31 +123,258 @@ proc explicitIpMapped*(
         addrs.add(remapped)
   addrs
 
+# --------------------------------------------------------------------------
+# UPnP / NAT-PMP helpers
+# --------------------------------------------------------------------------
```
richard-ramos: ✂️

libp2p/services/natservice.nim
```diff
@@ -5,35 +5,102 @@
...
 type
   NATMode* = enum
-    natModeAuto
+    Auto
       ## Use libp2p-native mechanisms (autonat / hole-punching). No port-mapping
-      ## is performed.
-      ## TODO: In this skeleton it is a no-op;
-    natModeExplicitIp
+      ## is performed here.
+      ## TODO: In this skeleton it is a no-op; folded in by piece 3a.
```
richard-ramos: wut

libp2p/services/natservice.nim
```diff
@@ -56,31 +123,258 @@ proc explicitIpMapped*(
...
+  if self.mapper.isNil:
+    debug "No port mapper available; skipping NAT port mapping"
+    return @[]
+  let listenPorts = findMappableListenPorts(listenAddrs)
+  if listenPorts.len == 0:
+    debug "No private listen addresses to map; releasing any prior mappings"
+    await self.unmapAll()
+    return @[]
+  if self.externalIp.isNone and not await self.discoverExternalIp():
+    return @[]
```
richard-ramos:
```suggestion
  if self.externalIp.isNone and not await self.discoverExternalIp():
    await self.unmapAll()
    return @[]
```
to close the older active mappings right?

libp2p/services/natservice.nim
```diff
@@ -56,31 +123,258 @@ proc explicitIpMapped*(
...
+proc mapOnePort(
+    self: NATService, lp: ListenPort, externalIp: IpAddress
+): Future[Opt[MappedEntry]] {.async: (raises: [CancelledError]).} =
+  let lease = uint32(self.config.leaseDuration.seconds)
+  let mapRes = await self.mapper.map(lp.port, lp.port, lp.proto, lease)
+  if mapRes.isErr:
+    warn "NAT port mapping failed", port = lp.port, proto = lp.proto, err = mapRes.error
+    return Opt.none(MappedEntry)
+
+  let extPort = mapRes.get()
```
richard-ramos:
```suggestion
  let extPort = await self.mapper.map(lp.port, lp.port, lp.proto, lease).valueOr:
    warn "NAT port mapping failed", port = lp.port, proto = lp.proto, err = error
    return Opt.none(MappedEntry)
```

libp2p/services/natservice.nim
```diff
@@ -56,31 +123,258 @@ proc explicitIpMapped*(
...
+  ## and reissues mappings before the lease expires. Invalidates the cached
+  ## external IP first so DHCP renewals / failovers are picked up.
+  heartbeat "NATService refresh", self.config.refreshInterval, sleepFirst = true:
+    self.externalIp = Opt.none(IpAddress)
+    await switch.peerInfo.update()
+
+# --------------------------------------------------------------------------
+# Service lifecycle
+# --------------------------------------------------------------------------
+
```
richard-ramos: ✂️

libp2p/services/natservice.nim
```diff
@@ -56,31 +123,258 @@ proc explicitIpMapped*(
...
+      # fall back to passing the listenAddrs through unchanged so the next
+      # mapper in the chain has something to work with.
+      if announced.len == 0:
+        return listenAddrs
+      return announced
+    self.refreshLoopFut = self.refreshLoop(switch)
 
-method start*(self: NATService, switch: Switch) {.async: (raises: [CancelledError]).} =
-  trace "Starting NATService", mode = self.config.mode
   if self.addressMapper != nil:
```
richard-ramos: nit: Noticed that we do use `!= nil` and `== nil` in the code. Giving a quick search, it seems we use `.isNil()` / `not foo.isNil()` more consistently. Perhaps we should do that in nat related code as well?

libp2p/services/natservice.nim
```diff
@@ -56,31 +123,258 @@ proc explicitIpMapped*(
...
+proc discoverExternalIp(
+    self: NATService
+): Future[bool] {.async: (raises: [CancelledError]).} =
+  let discoverRes = await self.mapper.discover(self.config.discoveryTimeout)
+  if discoverRes.isErr:
+    warn "NAT discovery failed; not announcing mapped addresses",
+      err = discoverRes.error
+    self.externalIp = Opt.none(IpAddress)
+    return false
+  self.externalIp = Opt.some(discoverRes.get())
```
richard-ramos:
```suggestion
  let discover = await self.mapper.discover(self.config.discoveryTimeout).valueOr:
    warn "NAT discovery failed; not announcing mapped addresses", err = error
    self.externalIp = Opt.none(IpAddress)
    return false
  self.externalIp = Opt.some(discover)
```

tests/libp2p/test_multistream.nim
```diff
@@ -478,14 +478,16 @@ suite "Multistream :: stream limits":
       let stream = await transport.accept()
       asyncSpawn acceptedOne(stream)
 
-  proc makeBlockedHandler(): LPProtoHandler =
+  proc makeBlockedHandler(entered: WaitGroup = nil): LPProtoHandler =
```
richard-ramos, line 481: Are the changes in this file required for this PR?

richard-ramos (approved):

PR #2597, opened 2026-06-03
fix(gossipsub): Partial Messages - fanout regression

### LINKED ISSUE #2586: gossipsub: unsubscribed partial-message fanout publisher can no longer fulfill parts
After #2569 an unsubscribed fanout publisher can announce its parts but can never deliver the actual part data, so the fanout flow no longer works.

The PR gated `onIncomingRPC` in `handlePartialRPC`:

https://github.com/vacp2p/nim-libp2p/blob/072961b767457d6a1b2680a45bc9123ee5a336f8/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim#L252-L259

Delivery is a round-trip: the publisher announces metadata, the consumer replies with what it is missing, then the publisher fulfills. The fulfill step is only triggered reactively through `onIncomingRPC`. For an unsubscribed publisher the gate is always false (it neither requests partials nor can advertise `supportsSendingPartial` without subscribing), so the publisher holds the complete message and knows what the consumer wants but never sends it. 

The changed `test_runner` "fanout scenario" was rewritten to assert the consumer never completes, showing the regression.

rlve:
## Summary

Fixes #2586. After PR #2569, an unsubscribed fanout publisher could no longer fulfill partial message requests. The gating in `handlePartialRPC` only let a node respond if it requested partials or advertised support for sending them. A fanout publisher does neither: it can't declare `supportsSendingPartial` without subscribing to the topic, yet it still holds the full message data consumers are missing. So consumers asking for parts never got them.

The fix adds a third path: if the peer requested partials and we have published to this `(topic, group)`, handle the RPC. Published state is tracked via the new `hasPublished()` helper (`lastPublishedMetadata.len > 0`).

The gating logic moved out of `handlePartialRPC` into a dedicated `shouldHandlePartialRPC` proc, restructured as guard clauses so it reads top to bottom and only does the group lookup on the path that needs it.

## Affected Areas
- [x] Gossipsub

## Impact on Library Users

Behavior change for the partial message extension: unsubscribed fanout publishers now correctly respond to part requests from consumers, restoring the pre-#2569 fulfillment flow. No API changes.

## Risk Assessment

Low. The change only widens when `onIncomingRPC` fires, and only for a peer that explicitly requested partials and a group we have actually published to. The new `getOrDefault` lookup does not allocate group state for unrelated RPCs.

## Reference
- Tested https://github.com/vacp2p/nim-libp2p/pull/2597/commits/6dd17d86f37c0460cc537a93da73358fad77517d against GossipSub Interop tests in https://github.com/rlve/test-plans/pull/10/changes/b3988ba687e950a356f2680d5fd5b7eaeaf1b1af - https://github.com/rlve/test-plans/actions/runs/26888908126/job/79308942071

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
+++ b/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
@@ -233,6 +233,27 @@ proc handleSubscribeRPC(ext: PartialMessageExtension, peerId: PeerId, rpc: SubOp
   else:
     ext.peerTopicOpts.del(key)
 
+proc shouldHandlePartialRPC(
+    ext: PartialMessageExtension, peerId: PeerId, rpc: PartialMessageExtensionRPC
+): bool =
+  let nodeTopicOpts = ext.config.nodeTopicOpts(rpc.topicID)
+  if nodeTopicOpts.requestsPartial:
+    return true
+
+  # We only send parts to a peer that asked for them.
+  if not ext.peerRequestsPartial(peerId, rpc.topicID):
+    return false
+
+  if nodeTopicOpts.supportsSendingPartial:
+    return true
+
+  # An unsubscribed fanout publisher has no topic opts set, but it has published
+  # to this group, so it can still fulfill parts the peer is missing.
+  # Look up without creating state.
+  let groupState =
+    ext.groupState.getOrDefault(TopicGroupKey.new(rpc.topicID, rpc.groupID))
+  groupState != nil and groupState.hasPublished()
+
 proc handlePartialRPC(
     ext: PartialMessageExtension, peerId: PeerId, rpc: PartialMessageExtensionRPC
 ) =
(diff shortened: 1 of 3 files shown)


libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -250,11 +253,17 @@ proc handlePartialRPC(
...
+    nodeTopicOpts.supportsSendingPartial or (group != nil and group.hasPublished())
+
   let shouldHandlePartialRPC =
-    nodeTopicOpts.requestsPartial or (
-      nodeTopicOpts.supportsSendingPartial and
-      ext.peerRequestsPartial(peerId, rpc.topicID)
-    )
+    nodeTopicOpts.requestsPartial or
+    (nodeCanSendPartial and ext.peerRequestsPartial(peerId, rpc.topicID))
   if shouldHandlePartialRPC:
```
vladopajic: `shouldHandlePartialRPC` might be converted to `proc` since code has grown. if this is done then `proc` can be implemented in such way to avoid unnecessary table lookups:

```
let nodeTopicOpts = ext.config.nodeTopicOpts(rpc.topicID) <-- lookup #1
if nodeTopicOpts.supportsSendingPartial:
  return true

let group = ext.groupState.getOrDefault(TopicGroupKey.new(rpc.topicID, rpc.groupID)) <--- lookup #2 
if  nodeTopicOpts.supportsSendingPartial and group != nil and group.hasPublished():
  return true

if ext.peerRequestsPartial(peerId, rpc.topicID): <-- lookup #3 
 return true

return false
```

note that lookup#2 and lookup#3 are not executed if condition after lookup#1 is satisfied.

vladopajic (approved):

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
rlve: It couldn't be exactly as suggested, but I refactored it in https://github.com/vacp2p/nim-libp2p/pull/2597/commits/1749f9d1241b022a8ffb65becfcfce3728782bbc

gmelodie (approved):

PR #2600, opened 2026-06-04
fix(autonatv2): include observed addresses in dial request candidates

2-towns:
## Summary

The AutonatV2 service sent only `peerInfo.addrs` as dial-request candidates. Until the node is confirmed reachable, the address mapper leaves only listen addresses there, so a node behind NAT never submits a dialable candidate: its reachability can never be confirmed (chicken-and-egg).

This PR lets the service add the addresses observed by other peers via identify as additional candidates, behind a new opt-in config flag `enableDialableCandidates` (default `false`, so behavior is unchanged unless explicitly enabled).

When enabled, `askPeer` builds the candidate list as:

- `guessDialableAddr(listenAddr)` for each listen address, kept only when an observed-IP quorum actually changed it (so we never submit a raw / private / wildcard address)
- plus `getMostObservedProtosAndPorts()` when a full IP+port observation reached quorum
- merged with `peerInfo.addrs`, deduplicated, then filtered through `peerInfo.addressPolicy`

It also guards against an empty candidate list: `askPeer` returns `Unknown` without dialing instead of sending an empty dial request.

## Affected Areas

- [ ] Gossipsub
- [ ] Transports
- [x] Protocol Logic
  
- [ ] Build / Tooling
- [ ] Other

Additive, opt-in config field. Off by default, so no behavior change for existing consumers.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

Off by default: existing consumers see no change. Consumers that opt in via `AutonatV2ServiceConfig.new(enableDialableCandidates = true)` get richer dial-request candidates derived from observed addresses (useful for nodes that change networks, e.g. mobile). The new field is additive: no breaking change, no migration needed. The `peerInfo.addressPolicy` and explicit announced addresses are still respected.

## Risk Assessment

Low. Disabled by default. When enabled, only a few extra candidates are added for dial-back testing.

### DIFF
--- a/libp2p/protocols/connectivity/autonatv2/service.nim
+++ b/libp2p/protocols/connectivity/autonatv2/service.nim
@@ -161,18 +161,19 @@ proc askPeer(
 
   var reqAddrs = switch.peerInfo.addrs
   if self.config.enableDialableCandidates:
-    # Until the node is confirmed reachable, the address mapper leaves only
-    # listen addresses in peerInfo.addrs, so a node behind NAT would never
-    # submit a dialable candidate (chicken-and-egg).
-    # Add first the guessDialableAddr that uses the listen port
-    # and the observed address as a fallback.
-    var observedCandidates =
-      switch.peerInfo.listenAddrs.mapIt(switch.peerStore.guessDialableAddr(it))
+    # A node behind NAT only has listen addresses until it is confirmed
+    # reachable, so derive extra candidates from the observed addresses to break
+    # the chicken-and-egg.
+    var observedCandidates: seq[MultiAddress]
+    for listenAddr in switch.peerInfo.listenAddrs:
+      let guessed = switch.peerStore.guessDialableAddr(listenAddr)
+      if guessed != listenAddr:
+        observedCandidates.add(guessed)
     observedCandidates &= switch.peerStore.getMostObservedProtosAndPorts()
 
-    # Combine observed candidates with the node's current addresses to form the
-    # candidate list
-    reqAddrs = deduplicate(reqAddrs & observedCandidates)
+    reqAddrs = deduplicate(reqAddrs & observedCandidates).filterIt(
+        switch.peerInfo.addressPolicy(it)
+      )
     if reqAddrs.len == 0:
       debug "No candidate addresses to test, not asking peer"
       return Unknown
--- a/libp2p/protocols/connectivity/autonatv2/service.nim
+++ b/libp2p/protocols/connectivity/autonatv2/service.nim
@@ -158,13 +158,12 @@ proc askPeer(
 
   # Until the node is confirmed reachable, the address mapper leaves only
   # listen addresses in peerInfo.addrs, so a node behind NAT would never
-  # submit a dialable candidate (chicken-and-egg). Add the addresses
-  # observed by other peers via identify as additional candidates.
-  var observedCandidates = switch.peerStore.getMostObservedProtosAndPorts()
-  if observedCandidates.len == 0:
-    # Fallback to guessDialableAddr if no observed candidates are available
-    observedCandidates =
-      switch.peerInfo.listenAddrs.mapIt(switch.peerStore.guessDialableAddr(it))
+  # submit a dialable candidate (chicken-and-egg).
+  # Add first the guessDialableAddr that uses the listen port
+  # and the observed address as a fallback.
+  var observedCandidates =
+    switch.peerInfo.listenAddrs.mapIt(switch.peerStore.guessDialableAddr(it))
+  observedCandidates &= switch.peerStore.getMostObservedProtosAndPorts()
 
   # Combine observed candidates with the node's current addresses to form the
   # candidate list
--- a/libp2p/protocols/connectivity/autonatv2/service.nim
+++ b/libp2p/protocols/connectivity/autonatv2/service.nim
@@ -38,6 +38,7 @@ type
     maxQueueSize: int
     minConfidence: float
     enableAddressMapper: bool
+    enableDialableCandidates: bool
 
   AutonatV2Service* = ref object of Service
     config*: AutonatV2ServiceConfig
@@ -65,6 +66,7 @@ proc new*(
     maxQueueSize: int = 10,
     minConfidence: float = 0.3,
     enableAddressMapper = true,
+    enableDialableCandidates = false,
 ): T =
   return T(
     scheduleInterval: scheduleInterval,
@@ -73,6 +75,7 @@ proc new*(
     maxQueueSize: maxQueueSize,
     minConfidence: minConfidence,
     enableAddressMapper: enableAddressMapper,
+    enableDialableCandidates: enableDialableCandidates,
   )
 
 proc new*(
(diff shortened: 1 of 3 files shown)


tests/libp2p/protocols/test_autonat_v2_service.nim
```diff
@@ -587,3 +588,63 @@ suite "AutonatV2 Service":
...
+    let
+      (service, client) = newService(NetworkReachability.Reachable, expectedDials = 1)
+      switch = createSwitch(Opt.some(service))
+      switches = @[createSwitch()]
+
+    # Simulate identify reports from other peers: the same public address
+    # must be observed at least minCount (3) times to become a candidate.
+    let observedAddr = MultiAddress.init("/ip4/8.8.8.8/tcp/4040").tryGet()
+    for _ in 0 ..< 3:
+      discard switch.peerStore.identify.observedAddrManager.addObservation(observedAddr)
```
gmelodie: Why add it three times?

gmelodie (commented):
> The AutonatV2 service sent only peerInfo.addrs as dial-request candidates. Until the node is confirmed reachable, the address mapper leaves only listen addresses there, so a node behind NAT never submits a dialable candidate: its reachability can never be confirmed (chicken-and-egg).

It's only chicken-and-egg if you assume AutoNAT is the only way to assert reachability, which it is not. However, I do agree that *if it were* the only way, it should still work. But it seems to me that the fix belongs to the address mapper, not autonat, since maybe other protocols could suffer from the same issue, no?

2-towns:
> It's only chicken-and-egg if you assume AutoNAT is the only way to assert reachability, which it is not. However, I do agree that _if it were_ the only way, it should still work. But it seems to me that the fix belongs to the address mapper, not autonat, since maybe other protocols could suffer from the same issue, no?

Yes but I think is this the common case when you are behind a NAT: you don't know your IP so it seems natural to rely on the Autonat + Idendity to find that IP. 

Regarding the address mapper I am not sure to understand... I thought each protocol was defining its own Address Mapper. And I am not sure of the side effect to fix that issue into it.

tests/libp2p/protocols/test_autonat_v2_service.nim
2-towns: Because you want to reach the quorum and assert the observed address

gmelodie: Ah! So better would be to have sth like `for _ in 0 ..< quorum:`

gmelodie (approved):
@2-towns nvm! Changing the address mapper would affect other protocols too.

vladopajic (approved):

richard-ramos (approved):

2-towns:
Previously the observed IP + observed source port was tried first, with `guessDialableAddr` only as a fallback. But that source port is an ephemeral outbound port, usually different from the listen port. 

So a node actually reachable on its listen port (e.g. a manual port-forward) got probed on the wrong port and becomes NotReachable.

That's why I changed the order in my last commit: the  `guessDialableAddr` is added first and the fallback is `getMostObservedProtosAndPorts`. 

Can you please review it again ?

vladopajic (approved):

gmelodie (approved):

libp2p/protocols/connectivity/autonatv2/service.nim
```diff
@@ -156,10 +156,27 @@ proc askPeer(
...
+  # observed by other peers via identify as additional candidates.
+  var observedCandidates = switch.peerStore.getMostObservedProtosAndPorts()
+  if observedCandidates.len == 0:
+    # Fallback to guessDialableAddr if no observed candidates are available
+    observedCandidates =
+      switch.peerInfo.listenAddrs.mapIt(switch.peerStore.guessDialableAddr(it))
+
+  # Combine observed candidates with the node's current addresses to form the
+  # candidate list
+  let reqAddrs = deduplicate(switch.peerInfo.addrs & observedCandidates)
```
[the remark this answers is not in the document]
richard-ramos: This looks like a valid concern

libp2p/protocols/connectivity/autonatv2/service.nim
```diff
@@ -156,10 +156,26 @@ proc askPeer(
...
+  # submit a dialable candidate (chicken-and-egg).
+  # Add first the guessDialableAddr that uses the listen port
+  # and the observed address as a fallback.
+  var observedCandidates =
+    switch.peerInfo.listenAddrs.mapIt(switch.peerStore.guessDialableAddr(it))
+  observedCandidates &= switch.peerStore.getMostObservedProtosAndPorts()
+
+  # Combine observed candidates with the node's current addresses to form the
+  # candidate list
+  let reqAddrs = deduplicate(switch.peerInfo.addrs & observedCandidates)
```
richard-ramos: This is also adding the original listen address when no quorum exists (when there is no quorum, guessDialableAddr will retun the original addresses (including raw/private/wildcard); and also bypasses both the `peerIfno.addressPolicy`, and the addresses defined `withAnnouncedAddresses` which are part of peerInfo.addrs


Probably code should be +- like this:

```suggestion
  var observedCandidates: seq[MultiAddress]

  if switch.peerInfo.announcedAddrs.len == 0:
    for listenAddr in switch.peerInfo.listenAddrs:
      let guessed = switch.peerStore.guessDialableAddr(listenAddr)
      if guessed != listenAddr:
        observedCandidates.add(guessed)

    observedCandidates &= switch.peerStore.getMostObservedProtosAndPorts()

  let reqAddrs = deduplicate(switch.peerInfo.addrs & observedCandidates).filterIt(switch.peerInfo.addressPolicy(it))
```

libp2p/protocols/connectivity/autonatv2/service.nim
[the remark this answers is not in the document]
2-towns: it depends what do you want to prioritise ? Is it the addresses in the `peerInfo` or the `observedCandidates` ?

tests/libp2p/protocols/test_autonat_v2_service.nim
2-towns: Okay I applied the change

libp2p/protocols/connectivity/autonatv2/service.nim
2-towns: Nice !  Instead of 

```nim
if switch.peerInfo.announcedAddrs.len == 0:
```

I introduce a a config boolean `enableDialableCandidates` to make the choice explicit

2-towns:
@richard-ramos I applied your suggestion but instead of checking the length of `announceAddrs`, I introduced a bool `enableDialableCandidates` (default false) to make it explicit and avoid breaking changes. Sorry to ask again but can you review this change please ?

libp2p/protocols/connectivity/autonatv2/service.nim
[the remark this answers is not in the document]
2-towns: Actually I think it makes sense with explicit `enableDialableCandidates` option.

libp2p/protocols/connectivity/autonatv2/service.nim
```diff
@@ -156,10 +159,30 @@ proc askPeer(
...
+    var observedCandidates: seq[MultiAddress]
+    for listenAddr in switch.peerInfo.listenAddrs:
+      let guessed = switch.peerStore.guessDialableAddr(listenAddr)
+      if guessed != listenAddr:
+        observedCandidates.add(guessed)
+    observedCandidates &= switch.peerStore.getMostObservedProtosAndPorts()
+
+    reqAddrs = deduplicate(observedCandidates & reqAddrs).filterIt(
+        switch.peerInfo.addressPolicy(it)
+      )
```
richard-ramos, line 176: Looking at this change, i do see something we gotta improve in follow up PRs:  since autonat does not dial all addresses but loops thru the addresses in order, and selects the first  adress that can be dialed, and will return immedialy (regardless of dialling being succesful or not).

With this PR, when `enableDialableCandidates == true`,  we'll add the observed addresses before the `peerInfo.addrs`, so.. if the observed address is not really dialable might be tested first, fail, and then other addresses from the the node wont be tested

Not a blocker for the PR since this it opt-in, but we likely should think on improving this in follow ups PRs, like for example, by creating some heuristics to determine which is the most likely dialable address, so we put it first in the list, adding other addresses as fallback, and sending requests...   we would keep track of the main address that was tested, and would later send new reqeusts rotating the address at index 0 so we try a different one first.

richard-ramos (approved):
LGTM

libp2p/protocols/connectivity/autonatv2/service.nim
2-towns: Yes I agree

PR #2601, opened 2026-06-04
chore(Identify): utilize protobuf_serialization

vladopajic:
should only be change in protobuf serialization method

### DIFF
--- a/libp2p/protocols/identify.nim
+++ b/libp2p/protocols/identify.nim
@@ -90,7 +90,7 @@ chronicles.expandIt(IdentifyMsg):
 proc makeIdentifyMsg(
     pi: PeerInfo, observedAddr: Opt[MultiAddress], sign: bool
 ): IdentifyMsg =
-  var spr: Opt[seq[byte]]
+  var spr = Opt.none(seq[byte])
   ## Optionally populate signedPeerRecord field.
   ## See https://github.com/libp2p/go-libp2p/blob/ddf96ce1cfa9e19564feb9bd3e8269958bbc0aba/p2p/protocol/identify/pb/identify.proto for reference.
   if sign:
--- a/libp2p/protocols/identify.nim
+++ b/libp2p/protocols/identify.nim
@@ -110,7 +110,9 @@ proc makeIdentifyMsg(
 
 proc makeIdentifyInfo(peer: PeerId, msg: IdentifyMsg): IdentifyInfo =
   var spr = Envelope.decode(msg.signedPeerRecord, PeerRecord.payloadDomain).toOpt()
-  if spr.isSome and msg.publicKey.isSome and (spr.get().publicKey != msg.publicKey.get()):
+  if spr.isSome and msg.publicKey.isSome and (
+    spr.get().publicKey != msg.publicKey.get()
+  ):
     spr = Opt.none(Envelope)
 
   IdentifyInfo(
--- a/libp2p/protocols/identify.nim
+++ b/libp2p/protocols/identify.nim
@@ -99,6 +99,8 @@ proc makeIdentifyMsg(
     agentVersion: Opt.some(if pi.agentVersion == "": AgentVersion else: pi.agentVersion),
     signedPeerRecord:
       if sign:
+        ## Optionally populate signedPeerRecord field.
+        ## See https://github.com/libp2p/go-libp2p/blob/ddf96ce1cfa9e19564feb9bd3e8269958bbc0aba/p2p/protocol/identify/pb/identify.proto for reference.
         pi.signedPeerRecord.envelope.encode().valueOr:
           info "failed to encode signed peer record", msg = $error
           @[]
@@ -107,6 +109,10 @@ proc makeIdentifyMsg(
   )
 
 proc makeIdentifyInfo(peer: PeerId, msg: IdentifyMsg): IdentifyInfo =
+  var spr = Envelope.decode(msg.signedPeerRecord, PeerRecord.payloadDomain).toOpt()
+  if spr.isSome and msg.publicKey.isSome and (spr.get().publicKey != msg.publicKey.get()):
+    spr = Opt.none(Envelope)
+
   IdentifyInfo(
     peerId: peer,
     pubkey: msg.publicKey,
@@ -115,8 +121,7 @@ proc makeIdentifyInfo(peer: PeerId, msg: IdentifyMsg): IdentifyInfo =
     observedAddr: msg.observedAddr,
     protoVersion: msg.protoVersion,
     agentVersion: msg.agentVersion,
-    signedPeerRecord:
-      Envelope.decode(msg.signedPeerRecord, PeerRecord.payloadDomain).toOpt(),
+    signedPeerRecord: spr,
   )
 
 proc new*(
@@ -138,10 +143,8 @@ method init*(p: Identify) =
     trace "handling identify request", stream
 
     let msg = makeIdentifyMsg(p.peerInfo, stream.observedAddr, p.sendSignedPeerRecord)
-    var bytes = msg.encode()
-
     try:
-      await stream.writeLp(move bytes)
+      await stream.writeLp(msg.encode())
       debug "identify: info sent", stream, info = p.peerInfo
     except LPError as e:
       trace "identify handler failed to write message", description = e.msg, stream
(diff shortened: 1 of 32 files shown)


gmelodie (approved):

libp2p/protocols/identify.nim
```diff
@@ -61,71 +76,48 @@ type
...
-    pb.write(3, proto)
-  observedAddr.withValue(observed):
-    pb.write(4, observed.data.buffer)
-  pb.write(5, ProtoVersion)
-  let agentVersion =
-    if peerInfo.agentVersion.len <= 0: AgentVersion else: peerInfo.agentVersion
-  pb.write(6, agentVersion)
-
-  ## Optionally populate signedPeerRecord field.
-  ## See https://github.com/libp2p/go-libp2p/blob/ddf96ce1cfa9e19564feb9bd3e8269958bbc0aba/p2p/protocol/identify/pb/identify.proto for reference.
```
vladopajic, line 96: keep this

libp2p/protocols/identify.nim
```diff
@@ -61,71 +76,48 @@ type
...
+    peerId: peer,
+    pubkey: msg.publicKey,
+    addrs: msg.listenAddrs,
+    protos: msg.protocols,
+    observedAddr: msg.observedAddr,
+    protoVersion: msg.protoVersion,
+    agentVersion: msg.agentVersion,
+    signedPeerRecord:
+      Envelope.decode(msg.signedPeerRecord, PeerRecord.payloadDomain).toOpt(),
+  )
```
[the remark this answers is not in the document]
vladopajic: hm

libp2p/protocols/identify.nim
```diff
@@ -61,71 +76,55 @@ type
...
+  IdentifyInfo(
+    peerId: peer,
+    pubkey: msg.publicKey,
+    addrs: msg.listenAddrs,
+    protos: msg.protocols,
+    observedAddr: msg.observedAddr,
+    protoVersion: msg.protoVersion,
+    agentVersion: msg.agentVersion,
+    signedPeerRecord: spr,
+  )
```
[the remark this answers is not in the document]
vladopajic: code currently implemeted matches old code

libp2p/protocols/identify.nim
```diff
@@ -40,9 +41,23 @@ type
...
+
+  IdentifyMsg {.proto2.} = object
+    publicKey* {.fieldNumber: 1, ext.}: Opt[PublicKey]
+    listenAddrs* {.fieldNumber: 2, ext.}: seq[MultiAddress]
+    protocols* {.fieldNumber: 3.}: seq[string]
+    observedAddr* {.fieldNumber: 4, ext.}: Opt[MultiAddress]
+    protoVersion* {.fieldNumber: 5.}: Opt[string]
+    agentVersion* {.fieldNumber: 6.}: Opt[string]
+    delta* {.fieldNumber: 7.}: Opt[Delta]
+    signedPeerRecord* {.fieldNumber: 8.}: seq[byte]
```
richard-ramos: This is changing the wire encoding when `sendSignedPeerRecord` is false. Previously this field was omitted completely, but now it is a `seq[byte]` that will be encoding using the default `@[]`, thus being present.

This introduces a breaking behavior against older nim-libp2p, because the prev code does this:
```
if ?pb.getField(8, signedPeerRecord).toOpt() and
    pubkey == signedPeerRecord.envelope.publicKey:
  iinfo.signedPeerRecord = Opt.some(signedPeerRecord.envelope)
```

`getField(8. signedPeerRecord)` will fail because empty bytes are not a valid signed envelope, thus failing cause `decodeMsg` to fail.

The fix is easy tho,  `signedPeerRecord` must be an `Opt[seq[byte]]`

richard-ramos: Vibecoded proof:

```
Old nim-libp2p encoder, no SPR:
  field 8: omitted
  old decoder accepts: true

PR #2601 encoder, no SPR:
  field 8: present, len = 0
  old decoder accepts: false   <-------------- should have been true
```

code:

```nim
{.push raises: [].}

import
  libp2p/crypto/crypto,
  libp2p/protobuf/minprotobuf,
  libp2p/routing_record,
  libp2p/utils/opt

import tests/tools/crypto

proc decodeLikeOldNimLibp2p(buf: seq[byte]): Opt[bool] =
  ## This is the important old nim-libp2p decode behavior:
  ## if field 8 is present, it must be a valid signed peer record.
  var
    pubkey: PublicKey
    signedPeerRecord: SignedPeerRecord

  var pb = initProtoBuffer(buf)

  if ?pb.getField(1, pubkey).toOpt():
    if ?pb.getField(8, signedPeerRecord).toOpt() and
        pubkey == signedPeerRecord.envelope.publicKey:
      discard signedPeerRecord

  Opt.some(true)

proc encodeLikeOldNimLibp2pNoSpr(pubkey: PublicKey): seq[byte] =
  ## Old nim-libp2p behavior when sendSignedPeerRecord = false:
  ## field 8 is not written at all.
  var pb = initProtoBuffer()
  pb.write(1, pubkey.getBytes().expect("valid key"))
  pb.finish()
  pb.buffer

proc encodeLikePr2601NoSpr(pubkey: PublicKey): seq[byte] =
  ## PR #2601 behavior when signedPeerRecord = @[]:
  ## field 8 is written as present bytes with length 0.
  var pb = initProtoBuffer()
  pb.write(1, pubkey.getBytes().expect("valid key"))
  pb.write(8, default(seq[byte]))
  pb.finish()
  pb.buffer

proc field8State(buf: seq[byte]): string =
  var
    pb = initProtoBuffer(buf)
    sprBytes: seq[byte]

  let present = pb.getField(8, sprBytes).expect("valid protobuf")
  if present:
    "present, len = " & $sprBytes.len
  else:
    "omitted"

let
  privKey = PrivateKey.random(rng()).expect("valid private key")
  pubkey = privKey.getPublicKey().expect("valid public key")

let oldWire = encodeLikeOldNimLibp2pNoSpr(pubkey)
let pr2601Wire = encodeLikePr2601NoSpr(pubkey)

echo "Old nim-libp2p encoder, no SPR:"
echo "  field 8: ", field8State(oldWire)
echo "  old decoder accepts: ", decodeLikeOldNimLibp2p(oldWire).isSome()

echo ""
echo "PR #2601 encoder, no SPR:"
echo "  field 8: ", field8State(pr2601Wire)
echo "  old decoder accepts: ", decodeLikeOldNimLibp2p(pr2601Wire).isSome()
```

libp2p/protocols/identify.nim
```diff
@@ -61,71 +76,55 @@ type
...
+      else:
+        @[],
+  )
+
+proc makeIdentifyInfo(peer: PeerId, msg: IdentifyMsg): IdentifyInfo =
+  var spr = Envelope.decode(msg.signedPeerRecord, PeerRecord.payloadDomain).toOpt()
+  if spr.isSome and msg.publicKey.isSome and (
+    spr.get().publicKey != msg.publicKey.get()
+  ):
+    spr = Opt.none(Envelope)
```
richard-ramos: this should use `SignedPeerRecord.decode` instead of `Envelope.decode` because the later only validates the envelope itself but does not decode and validate the actual `PeerRecord`.

Also, since `IdentifyPush` can omit the `publicKey`, this condition on line 113 could potentially allow storing a peer record that does not belong to a public key (when `spr.isSome` and `msg.publicKey.isNone`), so the check needs to be done.

this can be fixed like this, assuming that `signedPeerRecord` has been modified so it is a `Opt[seq[byte]]`:
```suggestion
  var spr = Opt.none(Envelope)
  msg.signedPeerRecord.withValue(sprBytes):
    SignedPeerRecord.decode(sprBytes).toOpt().withValue(signedPeerRecord):
      if signedPeerRecord.data.peerId == peer:
        spr = Opt.some(signedPeerRecord.envelope)

```

richard-ramos (changes requested):

libp2p/protocols/identify.nim
nitely: fwiw, `signedPeerRecord* {.fieldNumber: 8.}: seq[byte]` should not compile, but the validation is lacking (I'll fix it). `Bytes`/`seq[byte]` type needs to be either required or optional or repeated in proto2.

richard-ramos (approved):

PR #2604, opened 2026-06-05
feat(nat): absorb autonat / hole-punching into NATService

gmelodie:
## Summary

Move AutoNAT (v1/v2) and hole-punching wiring out of the builder and into
`NATService`, so `withNAT(NATConfig)` is the single entry point for all NAT
traversal concerns.

`NATConfig` is now three independent, optional concerns:

- `portMapping` — UPnP / NAT-PMP / explicit-IP, built via `upnpConfig`,
  `natPmpConfig`, `explicitIpConfig`.
- `reachability` — AutoNAT v1/v2 probing, built via `autonatConfig(version, …)`.
- `holePunching` — DCUtR + AutoRelay + AutoNAT v1, built via `holePunchingConfig(…)`.

`withNAT` may be called once per distinct concern and merges them; configuring
the same concern twice is a programmer error. `reachability` and `holePunching`
are mutually exclusive (hole-punching already drives its own AutoNAT v1) and
pairing them is rejected at setup with `ServiceSetupError`.

`withAutonatV2` and `withHolePunching` are **deprecated** — kept as thin wrappers
over `withNAT`, not removed.

## Affected Areas

- [x] Protocol Logic — NAT, AutoNAT, hole-punching
- [x] Build / Tooling — builder API surface

## Impact on Library Users

Breaking API changes:

- `NATMode` is split into `PortMappingMode` (`ExplicitIp`/`Upnp`/`NatPmp`) and
  `AutonatVersion` (`AutonatV1`/`AutonatV2`). The old `Auto` mode is gone — an
  empty `NATConfig()` is the no-op.
- Construct `NATConfig` via the helper procs (`upnpConfig` / `natPmpConfig` /
  `explicitIpConfig` / `autonatConfig` / `holePunchingConfig`) instead of
  `NATConfig(mode: …)`.
- `withAutonatV2(…)` / `withHolePunching(…)` now emit deprecation warnings;
  migrate to `withNAT(autonatConfig(AutonatV2, …))` /
  `withNAT(holePunchingConfig(…))`.
- `NATService.new` now takes `rng` as a required second argument.
- `PortMapperFactory` now receives `PortMappingMode` instead of `NATMode`.
- The live AutoNAT v2 service is exposed via the
  `autonatV2Service: Opt[AutonatV2Service]` accessor on `NATService` (derived
  from the active reachability service).

## References

Follows on from the `feat/nat/upnp-pmp` series; replaces the standalone
`chore/nat/include-autonat-hole-punching` branch.

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -36,7 +36,11 @@ import
 export
   switch, peerid, peerinfo, peeraddrpolicy, connection, multiaddress, crypto, errors,
   TLSPrivateKey, TLSCertificate, TLSFlags, ServerFlags, connmanager.ConnectionLimits,
-  connmanager.maxTotal, connmanager.maxInOut, natservice
+  connmanager.maxTotal, connmanager.maxInOut, natservice.NATConfig,
+  natservice.PortMappingMode, natservice.AutonatVersion, natservice.PortMapperFactory,
+  natservice.NATService, natservice.upnpConfig, natservice.natPmpConfig,
+  natservice.explicitIpConfig, natservice.autonatConfig, natservice.holePunchingConfig,
+  natservice.AutonatV2ServiceConfig, natservice.AutonatV2Service, natservice.natService
 
 const MemoryAutoAddress* = memorytransport.MemoryAutoAddress
 
@@ -338,26 +342,15 @@ proc withAutonatV2Server*(
   b.autonatV2ServerConfig = Opt.some(config)
   b
 
-proc mergeInto(dst: var NATConfig, src: NATConfig) =
-  template mergeField(field: untyped) =
-    src.field.withValue(v):
-      doAssert dst.field.isNone,
-        "withNAT: " & astToStr(field) & " configured more than once"
-      dst.field = Opt.some(v)
-
-  mergeField(portMapping)
-  mergeField(reachability)
-  mergeField(holePunching)
-
 proc withNAT*(
     b: SwitchBuilder, config: NATConfig, portMapperFactory: PortMapperFactory = nil
 ): SwitchBuilder =
-  ## Build ``config`` with `upnpConfig`/`natPmpConfig`/`explicitIpConfig`/`autonatConfig`/`holePunchingConfig`.
-  ## May be called repeatedly for distinct concerns; setting one concern twice is a programmer error.
+  ## Build ``config`` with the `*Config` helpers; call once per distinct concern,
+  ## setting the same concern twice is a programmer error.
   var merged = b.natConfig.get(NATConfig())
   merged.mergeInto(config)
   b.natConfig = Opt.some(merged)
-  if not portMapperFactory.isNil:
+  if not portMapperFactory.isNil():
     b.natPortMapperFactory = portMapperFactory
   b
 
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -16,19 +16,9 @@ import
   transports/[transport, tcptransport, wstransport, quictransport, memorytransport],
   muxers/[muxer, mplex/mplex, yamux/yamux],
   protocols/[identify, secure/secure, secure/noise, rendezvous, kademlia],
-  protocols/connectivity/[
-    autonat/server,
-    autonat/client,
-    autonat/service,
-    autonatv2/server,
-    autonatv2/service,
-    autonatv2/client,
-    relay/relay,
-    relay/client,
-    relay/rtransport,
-  ],
-  services/
-    [autorelayservice, hpservice, identify_pusher, natservice, wildcardresolverservice],
+  protocols/connectivity/
+    [autonat/server, autonatv2/server, relay/relay, relay/client, relay/rtransport],
+  services/[identify_pusher, natservice, wildcardresolverservice],
   connmanager,
   upgrademngrs/muxedupgrade,
   observedaddrmanager,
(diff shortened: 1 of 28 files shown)


vladopajic (approved):

libp2p/services/natservice.nim
```diff
@@ -37,12 +54,19 @@ type
...
     mapper: PortMapper
     refreshLoopFut: Future[void]
     mappedPorts: seq[(Port, MapProto)]
     externalIp*: Opt[IpAddress]
+    # AutoNAT / hole-punching sub-services. Populated by setup() based on config.
+    autonatService*: AutonatService
+    autonatV2Service*: AutonatV2Service
+    autonatV2Client*: AutonatV2Client
+    autoRelayService*: AutoRelayService
+    hpService*: HPService
```
richard-ramos: are all of these attributes required to be public?

libp2p/builders.nim
```diff
@@ -353,34 +330,15 @@ proc withAutonatV2Server*(
...
-  b
-
 proc withNAT*(
     b: SwitchBuilder, config: NATConfig, portMapperFactory: PortMapperFactory = nil
 ): SwitchBuilder =
-  ## Enable a NAT traversal service.
-  ## TODO: wire in autonat / hole-punching.
-  ## ``portMapperFactory`` is intended for tests; when ``nil``, NATService picks
-  ## the production UPnP / NAT-PMP backend matching ``config.mode``.
+  ## Enable NAT traversal. Build ``config`` with `upnpConfig`, `natPmpConfig`, `autonatConfig`, or `holePunchingConfig`.
```
richard-ramos: I think this API is confusing now. with this PR, it looks like this is the only place to enable NAT features, and `withNAT` being a single setter, butas it is right now, it only stores a single `NATConfig`. Each one of the helper functions (`natPmpConfig, `autonatConfig` and `holePunchConfig` each return a full `NATConfig` and the intention behind these functions is not clear. I could see an user doing something like this:
```nim
builder
  .withNAT(upnpConfig())
  .withNAT(autonatConfig(AutonatV1))
```
expecting UPnP and AutoNAT, but the second call what it will do is replace the first config, so they lose UPnP... so in theory to get both you should do:
```
var cfg = upnpConfig()
cfg.autonat = Opt.some(AutonatV1)
builder.withNAT(cfg)
```
which is not obvious at all.  Perhaps `withNAT` should reject repeated calls instead of just overwriting. Ideally we should not make it easier for the dev to commit mistakes

libp2p/services/natservice.nim
richard-ramos: Since this PR absorbs autonat and holepunching,  this `return` is problematic, because if for some reason the port mapper cannot be created, then `start` will return early before starting AutoNAT/Holepunching services. This means that a failure in upnp/natpmp could cause autonatv1/v2 and holepunching to not start.  (these are not dependent on each other, as they attack a different problem)

libp2p/services/natservice.nim
```diff
@@ -7,22 +7,39 @@ import std/[net, sequtils]
...
 
   NATConfig* = object
+    ## Hole-punching uses AutoNAT v1; pairing it with ``autonat = some(AutonatV2)`` is rejected at setup.
+    autonat*: Opt[AutonatVersion]
+    enableHolePunching*: bool
+    maxNumRelays*: int
+    onReservation*: OnReservationHandler
+    autonatV1ScheduleInterval*: Opt[Duration]
+    autonatV2ServiceConfig*: Opt[AutonatV2ServiceConfig]
     case mode*: NATMode
```
richard-ramos: This config got a bit confusing now, because mode previously only controlled which address/nat strategy to use, but now with all other configs that were added, it's not clear that mode refers just to port mapping... This `mode` should be renamed to something else...

Some strange configurations can be done now for example
- `NatConfig(mode: Auto)`:  this basically does nothing
- `NatConfig(mode: Auto, enableHolePunching: true, autonat: Opt.none())` - no autonat but seems to start autonat v1 because holepunching requires it (TIL??)
- `NatConfig(mode: Upnp, autonatV2ServiceConfig: some(...), autonat: Opt.none())`: accepted but ignores autonat v2 is ignored because v1 is defined.

This makes me think that the configuration should be splitted within NATConfig, so it's easier to understand what's going on, +- like this:
```
NATConfig = object
  portMapping: Opt[PortConfig]   # config related to upnp/natpmp/explicitIp
  reachability: Opt[AutoNATConfig]      # config for v1/v2
  holePunching: Opt[HolePunchingConfig]
```

libp2p/services/natservice.nim
```diff
@@ -274,35 +329,89 @@ proc refreshLoop(
...
+
+proc setupHolePunching(
+    self: NATService, switch: Switch
+) {.raises: [ServiceSetupError].} =
+  if self.config.autonat == Opt.some(AutonatV2):
+    raise newException(
+      ServiceSetupError,
+      "NATService: enableHolePunching currently requires AutoNAT v1; " &
+        "set NATConfig.autonat to none or some(AutonatV1).",
+    )
```
richard-ramos: wth. Holepunching should be independent of the autonat version. This seems to be something happening prior to this PR. Seems like we missed it when we were working on Autonat. Not a blocker for this PR but please create a follow up issue.

richard-ramos (commented):
I like the idea of moving NAT traversal setup out of `SwitchBuilder`. It makes sense that `NATService` owns this logic, but in general we should be more careful with breaking changes.  I think this should be done in a backwards compatible way. 

Previously (when discussing the removal of `{.public.}` pragma, it was said the following:
> If we introduce breaking changes or things suddenly stop being backward compatible,  lets bump major versions. Minor versions are for backward compatible additions only or just new stuff. And the patch version just for bugs (tbf, it's rarelly used). *If we can avoid the breaking change via `{.deprecated.}` , no need to increment new version*

And this is for sure something that we can avoid in this PR. (Let's leave major version updates when they can cause more impact).

I do think as well that the API was made harder to understand with this PR, due to reasons described below

tests/libp2p/services/test_natservice.nim
```diff
@@ -378,6 +383,188 @@ suite "NATService":
...
+    let nat = findNatService(switch)
+    check:
+      # HPService mounts DCUtR and drives AutoNAT v1, not v2.
+      dcutrMounted(switch)
+      nat.autonatV2Service.isNone()
+
+  test "hole-punching paired with AutonatV2 reachability is rejected at setup":
+    # The realistic path: two withNAT calls for the conflicting concerns.
+    expect ServiceSetupError:
+      discard SwitchBuilder
```
vladopajic: standard switch builder perhaps?

vladopajic (approved):

richard-ramos (approved):
Image

Issue #1163, opened 2024-08-05
Removing `discard Future` and `asyncspawn`


lchenut:
Discarding Future is generally a bad idea because the Future created could be "garbage collected" and cause segfault with a wrong condition.

Even if it's a better way of doing things, asyncspawn is also not recommended because it's uncontrollable. It creates a Future artifact to which nobody has access.

For both this problems, it's best to store the resulting Future in some object so that we can, at the very least, cancel it.

vladopajic:
there has been significant effort to improve this. 
also team has adopted this knowledge and new code never does this. 
also there copilot review instructions to warn if patterns like this happen in pull requests.

PR #2605, opened 2026-06-08
fix: track and cancel asyncSpawn-ed futures on teardown

gmelodie:
## Summary

Replaces every `asyncSpawn` in `libp2p/` source with a future stored on its owning object so it can be tracked and cancelled on teardown.

```
$ grep -rn "asyncSpawn" libp2p/
(no matches)
```

The tracking pattern is shared via a single `trackFut` helper in `utils/future`: it prunes already-finished futures, then takes ownership of the new one. Owners hold a `Future[void]` field (or a `seq[Future[void]]` / `Table[PeerId, Future[void]]`) and cancel it in their `close`/`stop`. Where the future was previously protected by `noCancel`, we now allow cancellation and catch `CancelledError` so close paths can short-circuit it.

Sites converted, grouped by owner:
- `RelayConnection.durationCheckFut`
- `SecureConn.cleanupFut`
- `LPChannel.{resetMessageFut, cleanupFut, handlerFut}` / `YamuxChannel.{cleanupFut, handlerFut, sendLoopFut}`
- `TcpTransport.closeFuts` / `QuicTransport.closeFuts` / `QuicMuxer.handlerFuts`
- `Switch.upgradeFuts`
- `ConnManager.{onCloseFuts, peerEventFuts, slotMonitorFuts}`
- `AutoRelayService.backingOff` (`seq[PeerId]` → `Table[PeerId, Future[void]]`)
- `GossipSub.pendingTasks` (disconnect + validateAndRelay fan-in)
- `PubSubPeer.{connectFut, sendFuts}` + `trackSend` helper

## Affected Areas

- [x] Gossipsub
- [x] Transports
- [x] Protocol Logic
- [x] Other — mplex / yamux

## Impact on Library Users

Additive only — no existing public symbol's signature was changed or removed. The change does, however, add new exported symbols (in Nim, `*` is public API):

- `utils/future.trackFut*` — shared future-tracking helper
- `LPChannel.cleanupFut*` / `handlerFut*`
- `GossipSub.pendingTasks*`
- `PubSubPeer.trackSend*`

These are exported only because another module assigns or reads them.

## Behavior Changes

This is not a pure refactor — teardown semantics change:

- `noCancel` is dropped from `RelayConnection.checkDurationConnection`; the new `except CancelledError` keeps `raises: []` and skips the timeout-driven close because close is already underway.
- `validateAndRelay` now drops its `validationSeen` bookkeeping via `defer` (and reads with `getOrDefault` instead of `pop`), so the entry is cleared even if the task is cancelled mid-flight.
- Owners now actively cancel/drain these futures on `close`/`stop` instead of leaking them.

## Risk Assessment

- `cancelSoon` no-ops on finished futures, so double-close paths are safe.
- `YamuxChannel.sendLoopFut`: callers now gate on `not channel.isSending` to avoid overwriting a live in-flight future with a finished one from the internal re-entry guard.
- `PubSubPeer.connect` adds a `not connectFut.finished` guard to avoid orphaning the connector loop on repeated `subscribePeer` calls.

## References

- Issue #1163


libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -1083,6 +1086,8 @@ method stop*(g: GossipSub): Future[void] {.async: (raises: [], raw: true).} =
   g.directPeersLoop.cancelSoon()
   g.scoringHeartbeatFut.cancelSoon()
   g.heartbeatFut.cancelSoon()
+  g.pendingTasks.cancelSoon()
```
richard-ramos, line 1087: What will happen with entries in `validationSeen` that are usually removed in `validateAndRelay` after `await g.validate` is succesful ? (line 506)

libp2p/transports/tcptransport.nim
```diff
@@ -177,6 +179,9 @@ method stop*(self: TcpTransport): Future[void] {.async: (raises: []).} =
         await acceptFut.value().closeWait()
     self.acceptFuts = @[]
 
+    await noCancel self.closeFuts.cancelAndWait()
+    self.closeFuts = @[]
```
richard-ramos: let's use this so the cleanup to close all connections happens
```suggestion
    discard await noCancel allFinished(self.closeFuts)
    self.closeFuts = @[]
```

libp2p/transports/quictransport.nim
```diff
@@ -423,6 +428,9 @@ method stop*(transport: QuicTransport) {.async: (raises: []).} =
   let futs = transport.connections.mapIt(it.close())
   await noCancel allFutures(futs)
 
+  await noCancel transport.closeFuts.cancelAndWait()
+  transport.closeFuts = @[]
```
richard-ramos: same here
```suggestion
  discard await noCancel allFinished(transport.closeFuts)
  transport.closeFuts = @[]
```

libp2p/connmanager.nim
```diff
@@ -734,6 +742,13 @@ proc close*(c: ConnManager) {.async: (raises: [CancelledError]).} =
   if not c.trimFut.isNil:
     await c.trimFut.cancelAndWait()
 
+  await c.peerEventFuts.cancelAndWait()
+  c.peerEventFuts = @[]
+  await c.slotMonitorFuts.cancelAndWait()
+  c.slotMonitorFuts = @[]
+  await c.onCloseFuts.cancelAndWait()
+  c.onCloseFuts = @[]
```
richard-ramos: I think these two lines should go in line 769, so the muxers are closed first and not canceled
```suggestion
```

libp2p/connmanager.nim
richard-ramos, line 774: Also, for the cleanup code to run, it should be like this:

```suggestion

  discard await allFinished(c.onCloseFuts)
  c.onCloseFuts = @[]
```

libp2p/muxers/mplex/mplex.nim
```diff
@@ -158,7 +158,7 @@ method handle*(m: Mplex) {.async: (raises: []).} =
         if m.streamHandler != nil:
           # Launch handler task
           # All the errors are handled inside `handleStream()` procedure.
-          asyncSpawn m.handleStream(channel)
+          channel.handlerFut = m.handleStream(channel)
```
richard-ramos, line 173: Besides removing the asyncSpawns from code, what would be the purpose of these new attributes? because as they're right now, they're unused after setting them

libp2p/muxers/yamux/yamux.nim
```diff
@@ -154,6 +154,9 @@ type
     closedRemotely: AsyncEvent
     closedLocally: bool
     receivedData: AsyncEvent
+    cleanupFut: Future[void] # tracks the muxer-side cleanup task
+    handlerFut: Future[void] # tracks the muxer's streamHandler invocation
+    sendLoopFut: Future[void] # tracks the singleton sendLoop
```
richard-ramos: Same question here about the purpose of the attributes. Should these futures be canceled on close?

libp2p/protocols/connectivity/relay/rconn.nim
```diff
@@ -53,9 +58,11 @@ proc new*(
   if limitDuration > 0:
     proc checkDurationConnection() {.async: (raises: []).} =
       try:
-        await noCancel stream.join().wait(limitDuration.seconds())
+        await stream.join().wait(limitDuration.seconds())
       except AsyncTimeoutError:
         await stream.close()
+      except CancelledError:
+        discard
```
richard-ramos: why remove `noCancel` here and catch `CancelledError`? maybe add a small comment in the `except` for the future to understand the intention behind this change.

richard-ramos:
This removes `asyncSpawn`, but i dont think it should close #1163. The issue also mentions discarded futures in general. Looking at the code, there are `discard someAsyncProc()` cases that still exist. Also, some cases in which the futures are dropped like using
```
g.pendingTasks.cancelSoon()
g.pendingTasks = @[]
```
i.e. drops the reference to the future... so maybe the GC could then remove those futures causing the problem described in the issue?  (i genuinely dont know).

libp2p/protocols/pubsub/gossipsub.nim
gmelodie: Good point!

gmelodie:
> This removes asyncSpawn, but i dont think it should close https://github.com/vacp2p/nim-libp2p/issues/1163

Not at all, this is only a part of the issue.

richard-ramos:
> > @richard-ramos: This removes asyncSpawn, but i dont think it should close #1163
> 
> @gmelodie:  Not at all, this is only a part of the issue.

I removed this part of the description then:  `Closes #1163`

vladopajic (approved):

richard-ramos (approved):

libp2p/muxers/mplex/mplex.nim
```diff
@@ -50,6 +50,16 @@ proc newTooManyChannels(): ref TooManyChannels =
 proc newInvalidChannelIdError(): ref InvalidChannelIdError =
   newException(InvalidChannelIdError, "max allowed channel count exceeded")
 
+proc drainChannelTasks(channs: seq[LPChannel]) {.async: (raises: []).} =
```
richard-ramos, line 53: We should likely include resetMessageFut in here to drain it

libp2p/transports/tcptransport.nim
```diff
@@ -177,6 +178,9 @@ method stop*(self: TcpTransport): Future[void] {.async: (raises: []).} =
         await acceptFut.value().closeWait()
     self.acceptFuts = @[]
 
+    discard await noCancel allFinished(self.closeFuts)
+    self.closeFuts = @[]
```
richard-ramos, line 182: should this be done too in the else of line 193 for the cases when stop() is called on a non listening tcp transport? (i.e. when you only do outgoing connections?)

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -724,8 +732,16 @@ proc startSendNonHighPriorityTask(p: PubSubPeer) =
...
+proc stopTasks*(p: PubSubPeer) =
+  ## Peer-wide teardown: cancels the connector loop, in-flight sends, and the
+  ## non-high-priority sender task; clears its priority queues.
+  if not p.connectFut.isNil():
+    p.connectFut.cancelSoon()
+    p.connectFut = nil
+  for fut in p.sendFuts:
+    fut.cancelSoon()
+  p.sendFuts = @[]
+  if not p.rpcmessagequeue.sendNonHighPriorityTask.isNil():
```
richard-ramos, line 744: Gotta also cancel the futures in sendPriorityQueue

vladopajic (approved):

libp2p/muxers/mplex/mplex.nim
```diff
@@ -50,6 +50,18 @@ proc newTooManyChannels(): ref TooManyChannels =
...
+  ## Awaits the per-channel tasks so they don't outlive the muxer; callers must
+  ## have torn channels and connection down first.
+  var futs: seq[Future[void]]
+  for chann in channs:
+    futs.add(chann.cleanupFut)
+    if not chann.handlerFut.isNil():
+      futs.add(chann.handlerFut)
+    if not chann.resetMessageFut.isNil():
+      futs.add(chann.resetMessageFut)
+  discard await noCancel allFinished(futs)
```
arnetheduck, line 63:
```suggestion
  await noCancel allFutures(futs)
```

this avoids the discard (that costs an extra copy of `futs`)

PR #2610, opened 2026-06-08
chore(Noise): utilize protobuf_serialization

vladopajic:


gmelodie (approved):

libp2p/protocols/secure/noise.nim
```diff
@@ -514,45 +518,37 @@ method handshake*(
...
-            "Failed to deserialize remote public key bytes. (initiator: " & $initiator &
-            ")"
-        )
-      if not remoteProof.getField(2, remoteSigBytes).valueOr(false):
-        raise (ref NoiseHandshakeError)(
-          msg:
-            "Failed to deserialize remote signature bytes. (initiator: " & $initiator &
-            ")"
+      remoteMsg = NoiseHandshakePayloadMsg.decode(handshakeRes.remoteP2psecret).valueOr:
+        raise newException(NoiseHandshakeError, error)
```
richard-ramos, line 540: decode returns an `Opt.none(NoiseHandshakePayloadMsg)` in https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protobuf/utils.nim#L20-L24 so, `error` wont contain anything. A hardcoded message must be used instead like it was done previously

vladopajic: this pr is based of https://github.com/vacp2p/nim-libp2p/tree/protobuf-improvements branch that has changed that. repeating same text all over (or precisely it always different message) and hiding decoding errors is not ideal

richard-ramos (approved):

PR #2612, opened 2026-06-09
chore(autonatv2):  utilize protobuf_serialization

vladopajic:


gmelodie (approved):
-345 >> +185!!!

libp2p/protocols/connectivity/autonatv2/client.nim
```diff
@@ -195,4 +192,4 @@ method sendDialRequest*(
   finally:
     # rollback any changes
     self.expectedNonces.del(nonce)
-  return dialResp.asAutonatV2Response(testAddrs)
+  return dialResp.get().asAutonatV2Response(testAddrs)
```
richard-ramos: If the exception in line 190 happens, `dialResp` could potentially be none, so this .get() will fail.

libp2p/protocols/connectivity/autonatv2/mock.nim
richard-ramos: How is it possible this line compiles if minprotobuffer was removed from the imports 🤔 (same question for line 38, and line 31 for the `MsgType`)

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -91,13 +91,11 @@ proc dialBack(
...
-    await stream.writeLp(DialBack(nonce: nonce).encode().buffer)
+    await stream.writeLp(DialBack(nonce: Opt.some(nonce)).encode())
 
     # receive DialBackResponse
-    discard DialBackResponse.decode(
-      initProtoBuffer(await stream.readLp(AutonatV2MsgLpSize))
-    ).valueOr:
-      trace "DialBack failed, could not decode DialBackResponse"
+    discard DialBackResponse.decode(await stream.readLp(AutonatV2MsgLpSize)).valueOr:
+      trace "DialBack failed, could not decode DialBackResponse", error = error
```
richard-ramos, line 100: Should we check after decoding for the status to be defined and  `OK`?

As it is right now, the code is okay because currently the protobuffer looks like this:
```
message DialBackResponse {
    enum DialBackStatus {
        OK = 0;
    }
    DialBackStatus status = 1;
}
```
meaning that since it's proto3, the default value for the status field would be OK (if they omit it), but then in the future if a new status is added, then not checking for the status here would become a bug

libp2p/protocols/connectivity/autonatv2/types.nim
richard-ramos, line 1: The [spec](https://raw.githubusercontent.com/libp2p/specs/master/autonat/autonat-v2.md) uses `proto3` when defining its protobuf. [go-libp2p](https://github.com/libp2p/go-libp2p/blob/8f1ddf32f7f4c9756bce70f77275fe7597110a6a/p2p/protocol/autonatv2/pb/autonatv2.proto) and [rust-libp2p](https://github.com/umgefahren/rust-libp2p/blob/ec56cbc645e2caaa083162a00c8843c0e859ecdd/protocols/autonat/src/v2/generated/structs.proto) also use `proto3`. Let's do the same.

libp2p/protocols/connectivity/autonatv2/types.nim
```diff
@@ -60,198 +56,41 @@ type
...
-# DialDataRequest
-proc encode*(dialDataReq: DialDataRequest): ProtoBuffer =
-  var encoded = initProtoBuffer()
-  encoded.write(1, dialDataReq.addrIdx)
-  encoded.write(2, dialDataReq.numBytes)
-  encoded.finish()
-  encoded
+  DialRequest* {.proto2.} = object
+    addrs* {.fieldNumber: 1, ext.}: seq[MultiAddress]
+    nonce* {.fieldNumber: 2, pint.}: Opt[Nonce]
```
richard-ramos: Why use `Opt[Nonce]` instead of `Nonce` directly? for a dial request it would be strange to not send a nonce. We could assume the field is always there (even if omitted, and the default 0 is returned) . If using proto3 this makes even more sense

Same question applies to the nonce on `DialBack`

And the same could be said of the `status` field in `DialBackResponse` and the `addrIdx` and `numBytes` in `DialDataRequest`

-----

These objects should look like this, because in most cases having them optional does not serve any purpose in the logic since we can work fine with their defaults without changing the purpose/semantics of the fields in the code. 
We should still keep the fields of the AutonatV2Msg object optional tho

```nim
type
  DialRequest* {.proto3.} = object
    addrs* {.fieldNumber: 1, ext.}: seq[MultiAddress]
    nonce* {.fieldNumber: 2, fixed.}: Nonce

  DialResponse* {.proto3.} = object
    status* {.fieldNumber: 1, ext.}: ResponseStatus
    addrIdx* {.fieldNumber: 2, pint.}: AddrIdx
    dialStatus* {.fieldNumber: 3, ext.}: DialStatus

  DialBack* {.proto3.} = object
    nonce* {.fieldNumber: 1, fixed.}: Nonce

  DialBackResponse* {.proto3.} = object
    status* {.fieldNumber: 1, ext.}: DialBackStatus

  DialDataRequest* {.proto3.} = object
    addrIdx* {.fieldNumber: 1, pint.}: AddrIdx
    numBytes* {.fieldNumber: 2, pint.}: NumBytes

  DialDataResponse* {.proto3.} = object
    data* {.fieldNumber: 1.}: seq[byte]

  AutonatV2Msg* {.proto2.} = object
    dialReq* {.fieldNumber: 1.}: Opt[DialRequest]
    dialResp* {.fieldNumber: 2.}: Opt[DialResponse]
    dialDataReq* {.fieldNumber: 3.}: Opt[DialDataRequest]
    dialDataResp* {.fieldNumber: 4.}: Opt[DialDataResponse]
```

nitely: if you are going to use proto3, keep in mind using [Nim enums is not conformant with the proto3 spec](https://github.com/status-im/nim-protobuf-serialization/blob/d9aa950b9d9e8bfc8a201740042b5e8ea5880875/protobuf_serialization/std/enums.nim#L18-L22), and it'll set the enum value to 0 for unknown values (this is expected in proto2), ex: if they update the spec and add a new enum value, that will be set to 0. Using `Opt[MyEnum]` won't help in the rare case the encoded msg contains many enum values for the same field (only the last *known* enum value will be set, instead of the last one which may be unknown). You can use (pint) int32/uint32 constants for now instead of enum. [There is oneof support](https://github.com/status-im/nim-protobuf-serialization/blob/d9aa950b9d9e8bfc8a201740042b5e8ea5880875/tests/test_oneof.nim#L14-L30) if you want to use that instead of optional. If you decide to use optional, you can make `AutonatV2Msg` a `{.proto3.}`, Opt works the same.

vladopajic: > Why use `Opt[Nonce]` instead of `Nonce` directly?

i have raised similar question in other prs, but we still decide to go with opt. i dont really understand how to make a decision here so i just made all optional as everything is (a assuming going to be) on other protos.

libp2p/protocols/connectivity/autonatv2/mock.nim
vladopajic: file is unused

libp2p/protocols/connectivity/autonatv2/server.nim
vladopajic: > Should we check after decoding

probably, but in this effort i am just doing surgical precision on refactoring protobuf serialization

libp2p/protocols/connectivity/autonatv2/types.nim
```diff
@@ -60,198 +57,41 @@ type
...
-# DialDataRequest
-proc encode*(dialDataReq: DialDataRequest): ProtoBuffer =
-  var encoded = initProtoBuffer()
-  encoded.write(1, dialDataReq.addrIdx)
-  encoded.write(2, dialDataReq.numBytes)
-  encoded.finish()
-  encoded
+  DialRequest* {.proto3.} = object
+    addrs* {.fieldNumber: 1, ext.}: seq[MultiAddress]
+    nonce* {.fieldNumber: 2, pint.}: Nonce
```
richard-ramos:
```suggestion
    nonce* {.fieldNumber: 2, fixed.}: Nonce
```

libp2p/protocols/connectivity/autonatv2/types.nim
```diff
@@ -60,198 +57,41 @@ type
...
+    dialStatus* {.fieldNumber: 3, ext.}: Opt[DialStatus]
 
-# DialDataResponse
-proc encode*(dialDataResp: DialDataResponse): ProtoBuffer =
-  var encoded = initProtoBuffer()
-  encoded.write(1, dialDataResp.data)
-  encoded.finish()
-  encoded
+  DialBack* {.proto3.} = object
+    nonce* {.fieldNumber: 1, pint.}: Nonce
```
richard-ramos:
```suggestion
    nonce* {.fieldNumber: 1, fixed.}: Nonce
```

libp2p/protocols/connectivity/autonatv2/types.nim
```diff
@@ -60,198 +57,41 @@ type
...
+  DialDataResponse* {.proto3.} = object
+    data* {.fieldNumber: 1.}: seq[byte]
 
-proc decode*(T: typedesc[AutonatV2Msg], pb: ProtoBuffer): Opt[T] =
-  var msg: ProtoBuffer
+  AutonatV2Msg* {.proto3.} = object
+    dialReq* {.fieldNumber: 1.}: Opt[DialRequest]
+    dialResp* {.fieldNumber: 2.}: Opt[DialResponse]
+    dialDataReq* {.fieldNumber: 3.}: Opt[DialDataRequest]
+    dialDataResp* {.fieldNumber: 4.}: Opt[DialDataResponse]
```
richard-ramos: @nitely , in a previous comment you mentioned nim-protobuf-serialization supports `oneof`. 

How can we use it here?

libp2p/protocols/connectivity/autonatv2/types.nim
richard-ramos: yeah, that's fair. maybe this is something to do in a follow-up for other protocols?
just thought about this a bit more, it does make sense for `Opt` to be used when a none actually means something... otherwise, the extra handling doesn't seem to benefit us much.

libp2p/protocols/connectivity/autonatv2/types.nim
nitely:
```nim
# nim c -r -d:LogGeneratedTypes autonat2.nim

import protobuf_serialization, protobuf_serialization/proto_parser

importProto3("autonat2.proto")
```

generates `autonat2.nim.generated.nim` which contains:

```nim

## Generated at line 3
type
  DialStatus* {.pure, proto3.} = enum
    UNUSED = 0, E_DIAL_ERROR = 100, E_DIAL_BACK_ERROR = 101, OK = 200
  DialBackStatus* {.pure, proto3.} = enum
    OK = 0
  DialBackResponse* {.proto3.} = object
    status* {.fieldNumber: 1, ext.}: DialBackStatus
  DialBack* {.proto3.} = object
    nonce* {.fieldNumber: 1, fixed.}: uint64
  DialDataResponse* {.proto3.} = object
    data* {.fieldNumber: 1.}: seq[byte]
  ResponseStatus* {.pure, proto3.} = enum
    E_INTERNAL_ERROR = 0, E_REQUEST_REJECTED = 100, E_DIAL_REFUSED = 101,
    OK = 200
  DialResponse* {.proto3.} = object
    status* {.fieldNumber: 1, ext.}: ResponseStatus
    addrIdx* {.fieldNumber: 2, pint.}: uint32
    dialStatus* {.fieldNumber: 3, ext.}: DialStatus
  DialDataRequest* {.proto3.} = object
    addrIdx* {.fieldNumber: 1, pint.}: uint32
    numBytes* {.fieldNumber: 2, pint.}: uint64
  DialRequest* {.proto3.} = object
    addrs* {.fieldNumber: 1.}: seq[seq[byte]]
    nonce* {.fieldNumber: 2, fixed.}: uint64
  MsgKind* {.pure, proto3.} = enum
    notSet = 0, dialRequest = 1, dialResponse = 2, dialDataRequest = 3,
    dialDataResponse = 4
  Msg* {.proto3, oneof.} = object
    case kind*: MsgKind
    of MsgKind.notSet:
      nil
    of MsgKind.dialRequest:
      dialRequest* {.fieldNumber: 1.}: DialRequest
    of MsgKind.dialResponse:
      dialResponse* {.fieldNumber: 2.}: DialResponse
    of MsgKind.dialDataRequest:
      dialDataRequest* {.fieldNumber: 3.}: DialDataRequest
    of MsgKind.dialDataResponse:
      dialDataResponse* {.fieldNumber: 4.}: DialDataResponse
  Message* {.proto3.} = object
    msg* {.oneof.}: Msg
```

then after decoding you can check the kind to see if the expected field was set. For encoding you do `Message(msg: Msg(kind: MsgKind.dialRequest, dialRequest: theDialRequest))`. That said the only case the Opt version fails is if multiple *different* fields were set (which I think seems odd here). With oneof the last that was set is set.

nitely: > That said the only case the Opt version fails is if multiple different fields were set (which I think seems odd here). With oneof the last that was set is set.

mmh, also if the oneof field is a message and it's split and interleaved with other oneof field in the encoded data, oneof won't merge the message with the previous one if the kind got changed in between (it will overwrite), but Opt[T] will always merge... that may give a different decoded result.

see [this oneof test case](https://github.com/status-im/nim-protobuf-serialization/blob/d9aa950b9d9e8bfc8a201740042b5e8ea5880875/tests/test_oneof.nim#L343-L356) and [this Opt test case](https://github.com/status-im/nim-protobuf-serialization/blob/d9aa950b9d9e8bfc8a201740042b5e8ea5880875/tests/test_pkg_results.nim#L97-L109).

libp2p/protocols/connectivity/autonatv2/types.nim
```diff
@@ -60,198 +57,58 @@ type
...
-    Opt.none(AutonatV2Msg)
-
-# Custom `==` is needed to compare since AutonatV2Msg is a case object
+  DialRequest* {.proto3.} = object
+    addrs* {.fieldNumber: 1, ext.}: seq[MultiAddress]
+    nonce* {.fieldNumber: 2, fixed.}: Nonce
+
+  DialResponse* {.proto3.} = object
+    status* {.fieldNumber: 1, ext.}: ResponseStatus
+    addrIdx* {.fieldNumber: 2, fixed.}: Opt[AddrIdx]
```
richard-ramos:
```suggestion
    addrIdx* {.fieldNumber: 2, pint.}: Opt[AddrIdx]
```

libp2p/protocols/connectivity/autonatv2/types.nim
```diff
@@ -60,198 +57,58 @@ type
...
+
+  DialBack* {.proto3.} = object
+    nonce* {.fieldNumber: 1, fixed.}: Nonce
+
+  DialBackResponse* {.proto3.} = object
+    status* {.fieldNumber: 1, ext.}: DialBackStatus
+
+  DialDataRequest* {.proto3.} = object
+    addrIdx* {.fieldNumber: 1, fixed.}: AddrIdx
+    numBytes* {.fieldNumber: 2, fixed.}: NumBytes
```
richard-ramos:
```suggestion
    addrIdx* {.fieldNumber: 1, pint.}: AddrIdx
    numBytes* {.fieldNumber: 2, pint.}: NumBytes
```

richard-ramos (commented):
The `fixed` change was too broad. It only applies to the `nonce` fields (see [protobuf scheme](https://github.com/libp2p/specs/blob/master/autonat/autonat-v2.md#rpc-messages). That's the only obs left for this PR.

libp2p/protocols/connectivity/autonatv2/types.nim
```diff
@@ -60,198 +57,58 @@ type
...
+    notSet = 0
+    DialRequest = 1
+    DialResponse = 2
+    DialDataRequest = 3
+    DialDataResponse = 4
+
+  AutonatV2MsgOneof* {.proto3, oneof.} = object
+    case kind*: MsgKind
+    of MsgKind.notSet:
+      nil
```
nitely:
```suggestion
      discard
```

richard-ramos (approved):

libp2p/protocols/connectivity/autonatv2/types.nim
nitely: mmh, it's probably a good idea to check the encoded data is still the same that used to be generated with minprotobuf. At least one test should have been failed because of `fixed` but they didn't. Similar to what Constantine did here https://github.com/vacp2p/nim-libp2p/blob/059018413d191e8ec78f2a4f66f574deee3ecab5/tests/libp2p/protocols/test_dcutr.nim#L36

> for the encoded value you can use `hexToSeqByte` from `stew/byteutils`: `check pb == hexToSeqByte("hardcoded_encoded_hex_value")` and you can use `echo toHex(pb)` to produce the hex string using a commit previous to this PR that used minprotobuf.

PR #2614, opened 2026-06-10
chore(routing-record): utilize protobuf_serialization

vladopajic:


tests/libp2p/test_routing_record.nim
```diff
@@ -82,7 +82,7 @@ suite "Signed Routing Record":
 
     check SignedPeerRecord.decode(buffer).error == EnvelopeInvalidSignature
 
-  test "Decode doesn't fail if some addresses are invalid":
```
vladopajic, line 85: tests have been inverted. because MultiAddress decoding already raises error when value is invalid. changing parsing of MultiAddress will change behavior in other places as well.

richard-ramos (approved):

gmelodie (approved):

PR #2616, opened 2026-06-10
chore(relay): utilize protobuf_serialization

vladopajic:


gmelodie (approved):

libp2p/protocols/connectivity/relay/messages.nim
```diff
@@ -156,21 +82,21 @@ proc checkValid*(spr: SignedVoucher): Result[void, EnvelopeError] =
   else:
     ok()
 
-# Circuit Relay V2 Hop Message
+# Circuit Relay V2 Messages
```
richard-ramos, line 54: The spec for circuitv2 uses protov3. Let's use that https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md  (with no required fields)

libp2p/protocols/connectivity/relay/messages.nim
```diff
@@ -156,21 +82,21 @@ proc checkValid*(spr: SignedVoucher): Result[void, EnvelopeError] =
...
+    expire* {.fieldNumber: 1, required, pint.}: uint64
+    addrs* {.fieldNumber: 2, ext.}: seq[MultiAddress]
+    svoucher* {.fieldNumber: 3.}: Opt[seq[byte]]
 
-  Limit* = object
-    duration*: uint32 # seconds
-    data*: uint64 # bytes
+  Limit* {.proto2.} = object
+    duration* {.fieldNumber: 1, required, pint.}: uint32
+    data* {.fieldNumber: 2, required, pint.}: uint64
```
richard-ramos: The spec says that these fields are optional

libp2p/protocols/connectivity/relay/messages.nim
```diff
@@ -156,21 +82,21 @@ proc checkValid*(spr: SignedVoucher): Result[void, EnvelopeError] =
...
 
-# Circuit Relay V2 Hop Message
+# Circuit Relay V2 Messages
 
 type
-  Peer* = object
-    peerId*: PeerId
-    addrs*: seq[MultiAddress]
+  Peer* {.proto2.} = object
+    peerId* {.fieldNumber: 1, required, ext.}: PeerId
```
richard-ramos: Marked as optional in the spec

libp2p/protocols/connectivity/relay/messages.nim
```diff
@@ -156,21 +82,21 @@ proc checkValid*(spr: SignedVoucher): Result[void, EnvelopeError] =
...
+    addrs* {.fieldNumber: 2, ext.}: seq[MultiAddress]
 
-  Reservation* = object
-    expire*: uint64 # required, Unix expiration time (UTC)
-    addrs*: seq[MultiAddress] # relay address for reserving peer
-    svoucher*: Opt[seq[byte]] # optional, reservation voucher
+  Reservation* {.proto2.} = object
+    expire* {.fieldNumber: 1, required, pint.}: uint64
```
richard-ramos: Marked as optional in the spec

libp2p/protocols/connectivity/relay/messages.nim
```diff
@@ -36,111 +39,34 @@ type
...
 # Voucher
 
-type Voucher* = object
-  relayPeerId*: PeerId # peer ID of the relay
-  reservingPeerId*: PeerId # peer ID of the reserving peer
-  expiration*: uint64 # UNIX UTC expiration time for the reservation
+type Voucher* {.proto2.} = object
+  relayPeerId* {.fieldNumber: 1, required, ext.}: PeerId
+  reservingPeerId* {.fieldNumber: 2, required, ext.}: PeerId
+  expiration* {.fieldNumber: 3, required, pint.}: uint64
```
richard-ramos: Optional in the spec

libp2p/protocols/connectivity/relay/messages.nim
```diff
@@ -187,152 +113,23 @@ type
...
-      lpb.write(2, msg.limit.data)
-    lpb.finish()
-    pb.write(4, lpb.buffer)
-  msg.status.withValue(status):
-    pb.write(5, status.ord.uint)
-
-  pb.finish()
-  pb
+  HopMessage* {.proto2.} = object
+    msgType* {.fieldNumber: 1, required, ext.}: HopMessageType
```
richard-ramos: Optional in the spec

libp2p/protocols/connectivity/relay/messages.nim
```diff
@@ -187,152 +113,23 @@ type
...
-  msg.msgType = StopMessageType(msgTypeOrd)
-
-  var pbPeer: ProtoBuffer
-  if ?pb.getField(2, pbPeer).toOpt():
-    var peer: Peer
-    ?pbPeer.getRequiredField(1, peer.peerId).toOpt()
-    discard ?pbPeer.getRepeatedField(2, peer.addrs).toOpt()
-    msg.peer = Opt.some(peer)
+  StopMessage* {.proto2.} = object
+    msgType* {.fieldNumber: 1, required, ext.}: StopMessageType
```
richard-ramos: Optional in the specs

libp2p/protocols/connectivity/relay/messages.nim
```diff
@@ -36,111 +41,30 @@ type
...
-  RelayMessage* = object
-    msgType*: Opt[RelayType]
-    srcPeer*: Opt[RelayPeer]
-    dstPeer*: Opt[RelayPeer]
-    status*: Opt[StatusV1]
+  RelayMessage* {.proto3.} = object
+    msgType* {.fieldNumber: 1, ext.}: Opt[RelayType]
+    srcPeer* {.fieldNumber: 2.}: Opt[RelayPeer]
+    dstPeer* {.fieldNumber: 3.}: Opt[RelayPeer]
+    status* {.fieldNumber: 4, ext.}: Opt[StatusV1]
```
richard-ramos, line 50: The comment about using proto3 applied to circuitv2. CircuitV1 objects should remain [proto2](https://github.com/libp2p/specs/blob/master/relay/circuit-v1.md#wire-format)

libp2p/protocols/connectivity/relay/messages.nim
```diff
@@ -187,152 +112,21 @@ type
...
-    if msg.limit.duration > 0:
-      lpb.write(1, msg.limit.duration)
-    if msg.limit.data > 0:
-      lpb.write(2, msg.limit.data)
-    lpb.finish()
-    pb.write(4, lpb.buffer)
-  msg.status.withValue(status):
-    pb.write(5, status.ord.uint)
+  HopMessage* {.proto3.} = object
+    msgType* {.fieldNumber: 1, ext.}: HopMessageType
```
richard-ramos: In the spec it says the following for both the `type` on on `HopMessage` and `StopMessage`:

```
  // This field is marked optional for backwards compatibility with proto2.
  // Users should make sure to always set this.
  optional Type type = 1;
```

So these cannot be a plain enum, but an Opt[enum]. However,In a diff comment @nitely mentioned:

> if you are going to use proto3, keep in mind using [Nim enums is not conformant with the spec](https://github.com/status-im/nim-protobuf-serialization/blob/d9aa950b9d9e8bfc8a201740042b5e8ea5880875/protobuf_serialization/std/enums.nim#L18-L22), and it'll set the enum value to 0 for unknown values, ex: if they update the spec and add a new enum value, that will be set to 0. You can use (pint) int32/uint32 constants for now instead.

That means that `HopMessageType` and `StopMessageType`should likely be changed from enum to `pint` as suggested, but also, that would mean that we need to check for the presence of that field with isSome() wherever it is used, and reject the message if not present. 

Not sure why it is the case, but looking at how it is used, think it is so it allows you can distinguish that you are specifically sending a RESERVE or a CONNECT msg type instead of assuming the default value. (0 in both cases)

vladopajic: suggested approach of using `pint` instead of `Opt[enum]` defeats the purpose of refactoring effort - why do those checks manually if old minprotobuf would be able to handle encoding/decoding perfectly so nothing really escapes to business logic?

case when user sends value 5 for enum where there are only [0, 1] values. could be seen as okay, because although value was not strictly [0, 1] it can still be valid (if other data is valid). 

however decision to fallback to 0 value was probably not ideal, or why not raise serialization error? or why not default to some other value: 1) could you default to value of first enum item (so if first time has value -1, then it's that value) or 2) default to value of special enum name (`Enum.ProtoUndefinedValue`)?

richard-ramos: >  You can use (pint) int32/uint32 constants for now instead.

The `for now` makes me think that this is meant to be a workaround and  will be fixed in the future?

nitely: > old minprotobuf would be able to handle encoding/decoding perfectly

Consider a enum of values 0 and 1 (that later gets updated to 0, 1, 2 but there are nim-libp2p instances out there with the old version). If the encoded data contains the field repeated multiple times: [0, 1, 2] in proto2 you get the enum set to 1; value 2 is discarded as it is unknown, this is the expected proto2 closed enums behaviour. In proto3 you get the enum set to 2, because it supports values that are not part of the enum (Nim enums do not support this).

That said it's an edge case (probably very rare), so it depends on whether you want to follow the proto spec strictly. In case the enum field appears only once in the encoded data (the most common case) Opt[NimEnum] works ok because the optional is simply not set if the value is unknown.

> The for now makes me think that this is meant to be a workaround and will be fixed in the future?

There is an open issue https://github.com/status-im/nim-protobuf-serialization/issues/39 . Proto3 enums will likely be implemented as `penum` pragma + `distinct int32/uint32` as it's common when interfacing with C for the same reason. Then the user defines the enum values as constants of `type MyEnum = distinct int32`.

nitely: btw, if minprotobuf sets the enum to 0 on unknown values, that does not follow the proto2 spec (neither the proto3 spec). I have not checked what it does.

vladopajic: > btw, if minprotobuf sets the enum to 0 on unknown values, that does not follow the proto2 spec (neither the proto3 spec). I have not checked what it does.

it depends case by case. filed reads value as int then you can have watever logic to convert to enum

vladopajic: so what needs to be done here?

richard-ramos: If this is an edge case as @nitely mentions, and the most probable outcome is that `Opt[NimEnum] works ok because the optional is simply not set if the value is unknown`, what we could do is create a follow up issue to revisit this code when https://github.com/status-im/nim-protobuf-serialization/issues/39 is closed.

richard-ramos (approved):
@vladopajic, i just went over the [specs](https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md) and noticed some other fields that are optional:

```proto3
message Peer {
  // This field is marked optional for backwards compatibility with proto2.
  // Users should make sure to always set this.
  optional bytes id = 1;   <----------------------------------
  repeated bytes addrs = 2;
}

message Reservation {
  // This field is marked optional for backwards compatibility with proto2.
  // Users should make sure to always set this.
  optional uint64 expire = 1; // Unix expiration time (UTC)  <-----------
  repeated bytes addrs = 2;   // relay addrs for reserving peer
  optional bytes voucher = 3; // reservation voucher
}
```  

would you prefer to do the logic for those on a follow up PR? i can create an issue for tracking

libp2p/protocols/connectivity/relay/messages.nim
richard-ramos: follow up issue - https://github.com/vacp2p/nim-libp2p/issues/2624

libp2p/protocols/connectivity/relay/messages.nim
```diff
@@ -36,143 +39,41 @@ type
...
-  RelayMessage* = object
-    msgType*: Opt[RelayType]
-    srcPeer*: Opt[RelayPeer]
-    dstPeer*: Opt[RelayPeer]
-    status*: Opt[StatusV1]
+  RelayMessage* {.proto2.} = object
+    msgType* {.fieldNumber: 1, ext.}: Opt[RelayType]
+    srcPeer* {.fieldNumber: 2.}: Opt[RelayPeer]
+    dstPeer* {.fieldNumber: 3.}: Opt[RelayPeer]
+    status* {.fieldNumber: 4, ext.}: Opt[StatusV1]
```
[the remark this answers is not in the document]
richard-ramos: wtf copilot, this does not make sense

vladopajic:
> would you prefer to do the logic for those on a follow up PR? i can create an issue for tracking

it will be done in this pr

vladopajic:
> message Peer {
>   // This field is marked optional for backwards compatibility with proto2.
>   // Users should make sure to always set this.
>   optional bytes id = 1;   <----------------------------------

this is v1:
```
  message Peer {
    required bytes id = 1;    // peer id
    repeated bytes addrs = 2; // peer's known addresses
  }
  ```
  
  it was marked as required there (?)

PR #2617, opened 2026-06-10
chore(kad): protobuf_serialization

vladopajic:
utilizes `protobuf_serialization` in kademlia and service discovery.


libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -121,37 +132,62 @@ proc encode*(
...
+    tMod: Moment.init(cast[int64](m.tMod), Second),
+    tWaitFor: m.tWaitFor.secs,
+    signature: m.signature.get(@[]),
+  )
 
-  pb.finish()
-  return pb
+Protobuf.serializerFor([TicketMsg])
+
+Protobuf.extensionDefaults(Ticket, defaultWriteSeq = false)
```
nitely:
```suggestion
Protobuf.extensionDefaults(Ticket, defaultSeq = false, packed = false)
```

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -121,37 +132,62 @@ proc encode*(
...
+    stream: InputStream,
+    value: var Ticket,
+    header: FieldHeader,
+    ProtoType: type ProtobufExt,
+): bool {.raises: [SerializationError, IOError].} =
+  var data = default(seq[byte])
+  if readFieldInto(stream, data, header, pbytes):
+    let ticketMsg = TicketMsg.decode(data).valueOr:
+      raise (ref ProtobufValueError)(msg: "Invalid Ticket")
+    value = ticketMsg.toTicket()
```
nitely: this won't support message merging.

From https://protobuf.dev/programming-guides/encoding/#last-one-wins

> The effect of these rules is that parsing the concatenation of two encoded messages produces exactly the same result as if you had parsed the two messages separately and merged the resulting objects.

but this code seems to overwrite the previous instance instead of merging.

vladopajic: this is temporal, just to make test pass. the idea here is not merging, but i was to use TicketMsg for protobuf for representing Ticket the actual type. i was having some compile errors and when i realized what might have been wrong (Duration needs extension?) i already had code and just decided to make it to see if tests will compile/pass before calling it a day. thanks for heads up

nitely: ah ok, ye, Duration and Moment likely need an extension.

tests/libp2p/kademlia/test_encoding.nim
```diff
@@ -117,41 +73,20 @@ suite "KadDHT Protobuffers":
...
     let maddrs = @[MultiAddress.init("/ip4/127.0.0.1/tcp/9000").get()]
     for ct in [connected, canConnect, cannotConnect, notConnected]:
       let peer = Peer(id: @[1'u8, 2, 3], addrs: maddrs, connection: ct)
-      let decoded = Peer
-        .decode(initProtoBuffer(peer.encode(hideConnectionStatus = false).buffer))
-        .get()
+      let decoded = Peer.decode(peer.encode(hideConnectionStatus = false)).get()
       check decoded.connection == ct
 
-  test "decode all four ConnectionStatus values":
```
vladopajic, line 145: test unnecessary, decoding already tested above

tests/libp2p/kademlia/test_encoding.nim
```diff
@@ -44,49 +38,11 @@ suite "KadDHT Protobuffers":
     )
     check msg == Message.decode(msg.encode(hideConnectionStatus = false)).get()
 
-  test "decode record with missing fields":
```
vladopajic, line 47: tests with missing/invalid data are unnecessary, we trust that protobuf_serialization is going to correctly serialize/deserialize

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -3,19 +3,21 @@
...
+import protobuf_serialization/pkg/results
+import protobuf_serialization/std/enums
+import ../../utils/[protobuf, protobuf_chronos]
 
 type
-  Record* = object
-    key*: seq[byte]
-    value*: Opt[seq[byte]]
-    timeReceived*: Opt[string]
+  Record* {.proto2.} = object
```
nitely, line 17: the spec seems to be using proto3 https://github.com/libp2p/specs/tree/master/kad-dht#rpc-messages  it seems the `syntax = "proto2";` was added later, but the singular fields have implicit presence which is only supported in proto3.

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -33,10 +35,10 @@ type
     canConnect = 2 # Unused
     cannotConnect = 3 # Unused
 
-  Peer* = object
-    id*: seq[byte]
-    addrs*: seq[MultiAddress]
-    connection*: ConnectionStatus
+  Peer* {.proto2.} = object
+    id* {.fieldNumber: 1.}: Opt[seq[byte]]
+    addrs* {.fieldNumber: 2, ext.}: Opt[seq[MultiAddress]]
```
nitely, line 40: This should not compile but there is a missing validation in protobuf-ser. A field cannot be optional and repeated.

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -54,290 +56,117 @@ type
...
-    key*: seq[byte]
-    record*: Opt[Record]
-    closerPeers*: seq[Peer]
-    providerPeers*: seq[Peer]
-    providerStatus*: Opt[AddProviderStatus]
-      # field 11 - ADD_PROVIDER response status (nim extension)
-    register*: Opt[RegisterMessage] # field 21 -  REGISTER message
-    getAds*: Opt[GetAdsMessage] # field 22 -  GET_ADS message
+  GetAdsMessage* {.proto2.} = object
+    advertisements* {.fieldNumber: 1.}: Opt[seq[seq[byte]]]
```
nitely: same as above, cannot be optional and repeated.

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -54,290 +56,116 @@ type
...
-    tWaitFor*: Duration # field 4 - Remaining wait time in seconds
-    signature*: seq[byte] # field 5 - Ed25519 signature
+  Ticket* {.proto2.} = object
+    advertisement* {.fieldNumber: 1, required.}: seq[byte]
+      # field 1 - Copy of the original advertisement
+    tInit* {.fieldNumber: 2, ext.}: Opt[Moment]
+      # field 2 - Ticket creation timestamp (Unix time in seconds)
+    tMod* {.fieldNumber: 3, ext.}: Opt[Moment]
+      # field 3 - Last modification timestamp (Unix time in seconds)
+    tWaitFor* {.fieldNumber: 4, ext.}: Opt[Duration]
```
richard-ramos, line 66: These fields are expected to be in seconds yet Duration and Moment proto extension use nanoseconds.

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -54,290 +56,116 @@ type
...
-    providerStatus*: Opt[AddProviderStatus]
-      # field 11 - ADD_PROVIDER response status (nim extension)
-    register*: Opt[RegisterMessage] # field 21 -  REGISTER message
-    getAds*: Opt[GetAdsMessage] # field 22 -  GET_ADS message
+  GetAdsMessage* {.proto2.} = object
+    advertisements* {.fieldNumber: 1.}: seq[seq[byte]]
+      # field 1 - List of encoded advertisements
+
+  Message* {.proto2.} = object
+    msgType* {.fieldNumber: 1, required, ext.}: MessageType
```
richard-ramos: The spec does not mark this field as required. Also, https://protobuf.dev/programming-guides/proto2/ says this for proto2:
> required: Do not use. Required fields are so problematic they were removed from proto3 and editions. Semantics for required field should be implemented at the application layer. When it is used, a well-formed message must have exactly one of this field.

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -54,290 +56,116 @@ type
...
+    key* {.fieldNumber: 2, required.}: seq[byte]
```
richard-ramos: Let's remove required, due to language guide saying to not use this, andd also because valid messages do not need to return the key at all times, like in FIND_NODE response in which only closerPeers is expected. 

Also seems to be the case for other implementations
- go-libp2p omits the key: https://github.com/libp2p/go-libp2p-kad-dht/blob/master/handlers.go#L252-L282
- rust-libp2p: https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/protocol.rs#L304-L338 and the same for responses of getProviders and getValue
 
Code should still check for this field to be present where it makes sense

richard-ramos (commented):
Interestingly enough, go-libp2p and rust-libp2p use proto3 even tho the spec says to use proto2 :( 

- https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/generated/dht.proto
- https://github.com/libp2p/go-libp2p-kad-dht/blob/master/pb/dht.proto

So I wonder if we should use the spec or the code as reference? Or if perhaps we should send an update to kademlia specs so they use proto3? 

I'm okay with using proto2, tho :shrug:

libp2p/protocols/kademlia.nim
```diff
@@ -102,7 +102,11 @@ proc new*(
       kad_messages_received.inc(labelValues = [$msg.msgType])
       kad_message_bytes_received.inc(bufLen.int64, labelValues = [$msg.msgType])
 
-      case msg.msgType
+      if msg.msgType.isNone:
+        debug "MessageType not set"
+        return
+
+      case msg.msgType.get()
```
richard-ramos: This would likely break interop, as they set PUT_VALUE as `0`, and in proto3 that will get omited, thus when it arrives here, it will be a `Opt.none` since the field is absent.

- go-libp2p: 
  - https://github.com/libp2p/go-libp2p-kad-dht/blob/master/pb/protocol_messenger.go#L64-L80
  - https://github.com/libp2p/go-libp2p-kad-dht/blob/master/handlers.go#L147-L214

- rust-libp2p:
  - https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/protocol.rs#L396-L400
  - https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/protocol.rs#L438-L447


Perhaps the way to go here is to use `msg.msgType.get(MessageType.putValue)` wherever `msg.msgType` is accesed or use
```
msgType* {.fieldNumber: 1.}: PBOption[MessageType.putValue]
```
because in the docs for protobuf serialization it says: 
> PBOption. PBOptions are a generic type instantiated with the default value for that field. They serve as regular Options, except when they're none, they still return a value when `get` is called (the default value).

vladopajic: `PBOption` is not really way to go, because "They serve as regular Options", and then using it more will defeat the purpose of using `results`. would be much better if default value is like this:

```
msgType* {.fieldNumber: 1.}: Opt[MessageType] = Opt.some(MessageType.putValue)
```

just like regular default value for property.

libp2p/protocols/kademlia.nim
```diff
@@ -102,7 +102,7 @@ proc new*(
       kad_messages_received.inc(labelValues = [$msg.msgType])
       kad_message_bytes_received.inc(bufLen.int64, labelValues = [$msg.msgType])
```
```suggestion
kad_messages_received.inc(labelValues = [$msg.msgType.get(MessageType.putValue)])
kad_message_bytes_received.inc(bufLen.int64, labelValues = [$msg.msgType.get(MessageType.putValue)])
```

richard-ramos (approved):

libp2p/protocols/kademlia/protobuf.nim
gmelodie: @nitely wdym "and repeated"? Which field is repeated here? Do you mean repeated as in `seq`? If so that makes sense

gmelodie (approved):

libp2p/protocols/kademlia/protobuf.nim
nitely: @gmelodie yes, that's what I mean. `foo: Opt[seq[T]]` (where `T: not byte`) would be translated to `optional repeated T = ...` in the proto schema, which is not legal/allowed.

PR #2619, opened 2026-06-12
chore(protobuf): add chronos/time extensions

vladopajic:
this pr adds protobuf extensions for chronos/time types. so it can be reused allover the repo.

### DIFF
--- a/libp2p/utils/protobuf_chronos.nim
+++ b/libp2p/utils/protobuf_chronos.nim
@@ -5,7 +5,7 @@
 
 import chronos, results, protobuf_serialization, protobuf_serialization/extension
 
-Protobuf.extensionDefaults(Duration, defaultSeq = true, packed = true)
+Protobuf.extensionDefaults(Duration, defaultSeq = true)
 
 func computeFieldSize*(
     field: int, value: Duration, ProtoType: type ProtobufExt, skipDefault: static bool
@@ -34,7 +34,7 @@ proc readFieldInto*(
   else:
     false
 
-Protobuf.extensionDefaults(Moment, defaultSeq = true, packed = true)
+Protobuf.extensionDefaults(Moment, defaultSeq = true)
 
 func computeFieldSize*(
     field: int, value: Moment, ProtoType: type ProtobufExt, skipDefault: static bool
--- a/libp2p/utils/protobuf_chronos.nim
+++ b/libp2p/utils/protobuf_chronos.nim
@@ -1,0 +1,63 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Copyright (c) Status Research & Development GmbH
+
+{.push raises: [].}
+
+import chronos, results, protobuf_serialization, protobuf_serialization/extension
+
+Protobuf.extensionDefaults(Duration, defaultSeq = true, packed = true)
+
+func computeFieldSize*(
+    field: int, value: Duration, ProtoType: type ProtobufExt, skipDefault: static bool
+): int =
+  computeFieldSize(field, value.nanoseconds, pint64, skipDefault)
+
+proc writeField*(
+    stream: OutputStream,
+    field: int,
+    value: Duration,
+    ProtoType: type ProtobufExt,
+    skipDefault: static bool = false,
+) {.raises: [IOError].} =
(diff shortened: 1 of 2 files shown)


libp2p/utils/protobuf_chronos.nim
```diff
@@ -0,0 +1,63 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Copyright (c) Status Research & Development GmbH
+
+{.push raises: [].}
+
+import chronos, results, protobuf_serialization, protobuf_serialization/extension
+
+Protobuf.extensionDefaults(Duration, defaultSeq = true, packed = true)
```
richard-ramos: since `packed` is `true` in here and in line 37, isnt it necessary then to define `computeFieldSizePacked`, `writeFieldPacked`, and `readFieldPackedInto`? because seems to me that this would fail the moment you try to encode a seq[Moment] or seq[Duration]

(because of this https://github.com/status-im/nim-protobuf-serialization/blob/f45476a3c1f4e7bff73845e6450d686be040ddeb/protobuf_serialization/reader.nim#L26-L32)

richard-ramos (commented):
Not sure if it makes sense to have an extension for `Moment`, as it is a monotonic clock, and shouldnt be used across peers.  I imagine that this was done because of `Ticket` in kademlia/protobuf since  `tInit` and `tMod` are `Moment`, but reading that code, it seems to me that was the wrong type to use and it should have been a `Duration`, `DateTime`, `Time`, or some `Timestamp = int64` type instead. (i.e. you would want a wall clock value instead, right?)

libp2p/utils/protobuf_chronos.nim
```diff
@@ -0,0 +1,63 @@
...
+{.push raises: [].}
+
+import chronos, protobuf_serialization, protobuf_serialization/extension
+
+Protobuf.extensionDefaults(Duration, defaultSeq = true, packed = true)
+
+func computeFieldSize*(
+    field: int, value: Duration, ProtoType: type ProtobufExt, skipDefault: static bool
+): int =
+  computeFieldSize(field, value.nanoseconds, pint64, skipDefault)
```
gmelodie, line 13: Really cool!

gmelodie (approved):
I didn't know nim-protobuf was this easy to extend. Really cool!

Question: should we consider having this code inside nim-chronos? Not sure where the scope of protobuffers is (the consumer of a lib or upstream)

vladopajic:
> Not sure if it makes sense to have an extension for `Moment`, as it is a monotonic clock, and shouldnt be used across peers.  I imagine that this was done because of `Ticket` in kademlia/protobuf since  `tInit` and `tMod` are `Moment`, but reading that code, it seems to me that was the wrong type to use and it should have been a `Duration`, `DateTime`, `Time`, or some `Timestamp = int64` type instead. (i.e. you would want a wall clock value instead, right?)

yes pr is done because protobufs in kademlia already use moment. refactoring from moment to duration can be don in following effort after protobuf serialization is utilized.

richard-ramos:
Created https://github.com/vacp2p/nim-libp2p/issues/2622 for follow up

libp2p/utils/protobuf_chronos.nim
vladopajic: packed was copy past error. fixed it

richard-ramos (approved):

libp2p/utils/protobuf_chronos.nim
nitely: you do need to implement packed for pint type extensions, if you want to support seq[T].

https://github.com/status-im/nim-protobuf-serialization/blob/cec5f1da897c0b3e6d3a1f2da6a36b4bbdc3a1a8/protobuf_serialization/codec.nim#L75

enum example: https://github.com/status-im/nim-protobuf-serialization/blob/master/protobuf_serialization/std/enums.nim

nitely:
> should we consider having this code inside nim-chronos? Not sure where the scope of protobuffers is (the consumer of a lib or upstream)

you can add it to protobuf-ser, here if you want: https://github.com/status-im/nim-protobuf-serialization/tree/master/protobuf_serialization/pkg

PR #2620, opened 2026-06-12
test(conn-manager): event ordering issue

rlve:
## Summary

Adds tests documenting an event ordering bug in the connection manager watermark trim. #2621

When a new connection pushes the peer count over `highWater`, `storeMuxer` triggers the trim before emitting `Connected` and `Joined`. With `gracePeriod` at zero and enough protected peers, the trim can prune the just-stored connection itself. The pruned peer's lifecycle events then come out in the wrong order and its ready state survives the disconnect.

The tests assert the current wrong behaviour:
- unit: events for the pruned peer are emitted as `Left, Disconnected, Connected, Joined` instead of `Connected, Joined, Left, Disconnected`
- unit: `waitForPeerReady` returns true for the pruned peer, because `notifyPeerReady` runs after the trim already cleared the ready state
- component: gossipsub unsubscribes on `Left` as a no-op, then the late `Joined` subscribes the pruned peer. The node ends up with a gossipsub subscription for a disconnected peer

In the component test the inversion only shows up when a `Connected` handler takes longer to complete than the connection close. `storeMuxer` emits `Joined` after all `Connected` handlers complete, so a handler that awaits I/O lets the prune finish first. For example that is current downstream usage: nimbus-eth2's `onConnEvent` awaits `performProtocolHandshakes` on every connection. The component test models it with a delaying handler.

  ## Affected Areas
  - [x] Tests

  ## Impact on Library Users

  None.

  ## Risk Assessment

  None

### DIFF
--- a/tests/libp2p/test_conn_manager.nim
+++ b/tests/libp2p/test_conn_manager.nim
@@ -555,6 +555,7 @@ suite "Connection Manager Watermark":
     await connMngr.close()
 
   asyncTest "lifecycle events are emitted not in order when the trim prunes the new connection":
+    # TODO nim-libp2p#2621 conn-manager: trim of the just-stored connection emits Left before Joined
     # storeMuxer triggers the trim before emitting Connected and Joined for the stored peer.
     # the trim can prune the just-stored connection, emitting Left and Disconnected first.
     # consumers assume the order: Connected, Joined, Left, Disconnected.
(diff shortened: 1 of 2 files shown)


gmelodie (approved):

vladopajic (approved):
good find

richard-ramos (approved):

PR #2625, opened 2026-06-12
feat(cbind): expose metrics via libp2p_collect_metrics

gmelodie:
## Summary

Exposes nim-libp2p's metrics through the C bindings so downstream consumers can scrape them.

- Adds `libp2p_collect_metrics(ctx, callback, userData)` to the cbind API. It walks the global metrics registry, serializes the resulting `{name, type, help, labels, value, timestamp}` records to a JSON array, and hands the bytes back through the callback's `(msg, len)`. The symbol uses the same `{.dynlib, exportc.}` convention as every other exported proc.
- Builds the cbind with `-d:metrics` instead of `--undef:metrics`. Previously every collector compiled to a no-op, so no values were ever recorded.

The registry is process-global and internally locked, so the export reads it directly on the caller's thread without dispatching to the libp2p thread. The bytes handed to the callback point into a stack-local string and are only valid for the duration of the synchronous call; consumers must copy before returning.

## Affected Areas

- [x] Build / Tooling
  

- [x] Other
  

## Impact on Library Users

- New, additive FFI symbol \`libp2p_collect_metrics\`; no existing bindings change.
- cbind shared/static libraries now record metrics at runtime (\`-d:metrics\`).

## Risk Assessment

- API surface is additive: existing FFI symbols are unchanged.
- The flag flip from \`--undef:metrics\` to \`-d:metrics\` is a runtime behavior change — collectors that previously compiled to no-ops now actually record, adding the usual collector bookkeeping cost. This matches how nim-libp2p is built across CI and downstream.

## Additional Notes

Consumed by logos-co/logos-libp2p-module, which parses the returned JSON array.


cbind/libp2p.nim
```diff
@@ -1423,5 +1427,104 @@ proc libp2p_peerstore_delete_peer(
...
+  var labelPairs = newSeq[LabelPair](labels.len)
+  for i in 0 ..< labels.len:
+    labelPairs[i] = LabelPair(name: labels[i], value: labelValues[i])
+  entries[].add MetricEntry(
+    name: name,
+    `type`: metricType,
+    help: help,
+    labels: labelPairs,
+    value: value,
+    timestamp: timestamp.toUnixFloat(),
```
richard-ramos: why toUnixFloat? shouldnt it be unix timestamps in milliseconds? 🤔

cbind/libp2p.nim
```diff
@@ -1423,5 +1427,104 @@ proc libp2p_peerstore_delete_peer(
...
+    entries: ptr seq[MetricEntry],
+    metricType: string,
+    help: string,
+    name: string,
+    value: float64,
+    labels, labelValues: openArray[string],
+    timestamp: Time,
+) {.gcsafe, raises: [].} =
+  if name.endsWith("_created"):
+    return
```
richard-ramos, line 1518: Curious to know why _created metrics are dropped?

gmelodie: So apparently nim-metrics adds these creation timestamps to every counter and summary [here](https://github.com/status-im/nim-metrics/blob/b4b70a88fe1755d281366cbc3f22d7515240d192/metrics.nim#L553-L569), and in these cases we just want `_total`

richard-ramos (approved):

vladopajic (approved):

PR #2631, opened 2026-06-15
chore(autonatv2): normalize response address selection

richard-ramos:
keep the reachability result and return no selected address for addresses not part of the original request


libp2p/protocols/connectivity/autonatv2/utils.nim
```diff
@@ -37,8 +37,16 @@ proc asAutonatV2Response*(
       dialResp: self,
       addrs: Opt.none(MultiAddress),
     )
+
+  if uint64(addrIdx) >= uint64(testAddrs.len):
```
vladopajic: notation without `()` -> `addrIdx.uint64` is more eye pleasing also 1 character shorter

vladopajic (approved):

PR #2637, opened 2026-06-16
feat(protobuf): add universal protobuf metrics

vladopajic:
this pr adds universal protobuf metrics for total read/write bytes when `libp2p_protobuf_metrics` compile time flag is used.

`libp2p_protobuf_metrics` flag was created earlier but it only had metrics for protobufs encode/decode on pubsub scope (`libp2p_pubsub_rpc_bytes_write` and `libp2p_pubsub_rpc_bytes_read`).

in order to enable similar metrics in new era with `protobuf_ser` this pr introduces universal protobuf metrics that can be used to measure for any type. metrics for types are enabled by setting `withMetrics=true` argument to `serializerFor` macro. developers have to choose which types should have metrics manually, by choosing main uttermost container types, because metrics for types such as `PeerId` are unnecessary as they appear as properties on main types so their metrics are not really individually interesting. 

this pr introduced universal metrics `libp2p_protobuf_bytes_read` and `libp2p_protobuf_bytes_write`. going forward idea is for metrics `libp2p_pubsub_rpc_bytes_write` and `libp2p_pubsub_rpc_bytes_read` to become obsolete, or replaced. causing braking change in metrics.


libp2p/utils/protobuf.nim
```diff
@@ -5,16 +5,28 @@
...
 
-macro serializerFor*(_: type Protobuf, Types: untyped): untyped =
+when defined(libp2p_protobuf_metrics):
+  import ./protobuf_metrics
+
+macro serializerFor*(
+    _: type Protobuf, Types: untyped, withMetrics: bool = false
+): untyped =
   ## This generates encode/decode protobuf procs for `Types`
+  let doMetrics = newLit(withMetrics.eqIdent("true"))
```
richard-ramos, line 15:
```suggestion
    _: type Protobuf, Types: untyped, withMetrics: static[bool] = false
): untyped =
## This generates encode/decode protobuf procs for `Types`
  let doMetrics = newLit(withMetrics)
```

Maybe this so it works both with the `true` literal and also with constants like this?

```nim
const someConstant = true
Protobuf.serializerFor([Msg], someConstant)

# or

Protobuf.serializerFor([Msg], defined(someConstant))
```

richard-ramos (approved):

libp2p/utils/protobuf.nim
vladopajic: tbh, const usage looks like unnecessary const

gmelodie (approved):
Code is ok, but can you clarify why you're using both the compile time flag and the `doMetrics` argument? Can we not get rid of the compile time flag and only use `dometrics` instead?

vladopajic:
> Code is ok, but can you clarify why you're using both the compile time flag and the `doMetrics` argument? Can we not get rid of the compile time flag and only use `dometrics` instead?

`withMetrics` is already explained in pr description.

tldr, metrics are not interesting for some types, for example types like `PeerId` or any other type like `MultiAddress`. those are types of inner properties of types that we really care like `IdentifyMsg`. that's why developer should manually opt in for metrics for types that meter.

PR #2638, opened 2026-06-16
chore(pubsub)!: protobuf_serialization

vladopajic:
utilize protobuf_serialization in pubsub; should have minimal functional changes (only to support `T → Opt[T]`).

**braking change**:
- some values in pubsub interface have changed type to `Opt[T]` form `T`
   - `Message`, `PartialMessageExtensionRPC` now have `Opt[T]` on all properties
- old pubsub metrics have be abandoned in favor for general pubsub metrics (as explained in #2637). 
   - still metrics will be utilized in later pr for all protocols/messages altogether


gmelodie (approved):

libp2p/protocols/pubsub/rpc/message.nim
```diff
@@ -110,9 +108,8 @@ proc init*(
     topic: string,
     seqno: Opt[uint64],
 ): Message {.gcsafe, raises: [].} =
-  var msg = Message(data: move(data), topic: topic)
-  msg.fromPeer = peerId
+  var msg = Message(fromPeer: Opt.some(peerId), data: move(data), topic: topic)
```
richard-ramos: How come this compiles? the fields in `Message` are marked as `Opt`
🤔 

```
  Message* {.proto2.} = object
    fromPeer* {.fieldNumber: 1, ext.}: Opt[PeerId]
    data* {.fieldNumber: 2.}: Opt[seq[byte]]
    seqno* {.fieldNumber: 3.}: Opt[seq[byte]]
    topic* {.fieldNumber: 4.}: Opt[string]
    signature* {.fieldNumber: 5.}: Opt[seq[byte]]
    key* {.fieldNumber: 6.}: Opt[seq[byte]]
```

libp2p/protocols/pubsub/rpc/message.nim
```diff
@@ -25,26 +25,27 @@ declareCounter(
...
     Opt.some(pubkey)
-  elif m.key.len > 0 and pubkey.init(m.key):
+  elif m.key.isSome and pubkey.init(m.key.get()):
     # check if peerId extracted from m.key is the same as m.fromPeer
     let derivedPeerId = PeerId.init(pubkey).valueOr:
       warn "could not derive peerId from key field"
       return Opt.none(PublicKey)
 
-    if derivedPeerId != m.fromPeer:
+    if derivedPeerId != m.fromPeer.get():
```
richard-ramos: the `from` in messages is optional, depending on the signature policy, it might not be used at all (`StrictNoSign`, `LaxNoSign`)

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -232,22 +235,24 @@ proc getPeers(prune: ControlPrune, peer: PubSubPeer): seq[(PeerId, Opt[PeerRecor
   var routingRecords: seq[(PeerId, Opt[PeerRecord])]
   for record in prune.peers:
     var peerRecord = Opt.none(PeerRecord)
-    if record.signedPeerRecord.len > 0:
-      SignedPeerRecord.decode(record.signedPeerRecord).toOpt().withValue(spr):
-        if record.peerId != spr.data.peerId:
+    if record.signedPeerRecord.isSome:
+      SignedPeerRecord.decode(record.signedPeerRecord.get()).toOpt().withValue(spr):
+        if record.peerId.get() != spr.data.peerId:
```
richard-ramos: in https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#protobuf `PeerInfo.peerId` is optional, so this .get() can crash the app, so maybe a default `PeerId()` in the `.get()` should be used here.

Looking at other impl:
- go-libp2p: accepts the empty peerId but will later fail the dial
- rust-libp2p: filters missing peerIDs

libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -451,3 +474,34 @@ proc withIMReceiving*(_: typedesc[RPCMsg], preamble: sink Preamble): RPCMsg =
...
+    msg
+
+func validate*(msg: RPCMsg): Result[void, string] =
+  # validates RPCMsg after it is received and decoded.
+  for m in msg.subscriptions:
+    if m.topic.isNone:
+      return err("Subsciption topic must be set")
+  for m in msg.messages:
+    if m.topic.isNone or m.topic.get().len == 0:
+      return err("Message missing required topic field")
```
richard-ramos: The spec says that the topic is required
```
message Message {
  optional bytes from = 1;
  optional bytes data = 2;
  optional bytes seqno = 3;
  required string topic = 4;
  optional bytes signature = 5;
  optional bytes key = 6;
}
```
However, protobuf recommendations for proto2 say that we shouldnt use `required`. But, we still want this validation to happen, and currently it is being done only in here, but not when decoding a `Message`.

Also, this validation is being more strict due to not allowing `""` than what the spec suggest and other implementations do, thus breaking interop.

to fix that, this validation should be extracted to its own function, and only check for presence, not for content length:
 
```nim
func validate*(msg: Message): Result[void, string] =
  if msg.topic.isNone:
    return err("Message missing required topic field")
  ok()
```

Then: 
```suggestion
  for m in msg.messages:
    ?message.validate()
```

And also inside the `decode` proc for `Message`: 
```nim
proc decode*(_: type Message, buf: seq[byte]): Result[Message, string] =
  try:
    let msg = decodeMessage(buf)
    ?msg.validate()
    ok(msg)
  except SerializationError as e:
    err("failed to decode Message from protobuf bytes. " & e.msg)
```

This validate function should also be called inside the encode function for `Message` and `RpcMessage` to ensure we are not sending an Opt.none (which does not happen now but might be done by mistake later)

libp2p/protocols/pubsub/gossipsub/preamblestore.nim
```diff
@@ -35,9 +35,9 @@ proc init*(
     expiresAt: Moment,
 ): T =
   PreambleInfo(
-    messageId: preamble.messageID,
-    messageLength: preamble.messageLength,
-    topicId: preamble.topicID,
+    messageId: preamble.messageID.get(),
+    messageLength: preamble.messageLength.get(),
```
richard-ramos: messageLength is optional in the protobuffer, so this can potentially fail.
```suggestion
    messageLength: preamble.messageLength.get(0),
```

richard-ramos (commented):
I am not convinced we should expose all protobuf field optionality as a breaking change on our public API... we'd be intentionally forcing the release of a new major version (if we respect semver) and changes downstream for something whose impact is limited to maintainability.

Looking downstream projects like nimbus and logos delivery use fields like the `message.data`, `message.topic` and `message.fromPeer` in validators, message providers, etc, and they would now need to add `.get()` or `valueOr` handling for data that we in nim-libp2p would have rejected or where providing defaults would mean no changes in the API while achieving the same result.

Can we keep the protobuf library migration internal where possible and only use Opt when the absence of a field is meaningful and not just because the protobuf field is marked as optional? and in cases where the public API changes are needed, to try keep them minimal? i.e. we are using optional topics because of `required` not being recommended to be used... but having it optional in the public API does not provide much value, since an `Opt.none` topic would still be considered invalid anyway. The same for the data being a `Opt[seq[byte]]`, when we could assume that an empty sequence means no data.

-----

> old pubsub metrics have be abandoned in favor for general pubsub metrics (as explained in https://github.com/vacp2p/nim-libp2p/pull/2637).
>    still metrics will be utilized in later pr for all protocols/messages altogether

Are these abandoned pubsub metrics used in dashboards on logos-delivery / nimbus-eth2 ? I'd like to understand the impact this change might have.
cc: @tersec , @Ivansete-status

libp2p/protocols/pubsub/rpc/message.nim
vladopajic: interesting; i assume that this proc is not really used so it actually doesn't compile

vladopajic: it doesn't compile squared

Ivansete-status:
Hello @richard-ramos , I think is safe to delete those metrics from `logos-delivery` PoV.

EDITED: These are the metrics that I'm considering as being removed:
```
libp2p_pubsub_rpc_bytes_read
libp2p_pubsub_rpc_bytes_write
```

These are the libp2p metrics used in the major `logos-delivery` Grafana board:
## PubSub / GossipSub

| Metric | Type | Labels used | Panels |
|---|---|---|---|
| `libp2p_pubsub_peers` | gauge | — | "Libp2p PubSub Peers" (52), "Connected Peers" (54), "LibP2P PubSub Peers" (8, summed by instance) |
| `libp2p_pubsub_topics` | gauge | — | "Pubsub Topics" (60) |
| `libp2p_pubsub_subscriptions_total` | counter | — | "Pubsub Topics" (60) |
| `libp2p_pubsub_unsubscriptions_total` | counter | — | "Pubsub Topics" (60) |
| `libp2p_pubsub_validation_success_total` | counter | — | "LibP2P Validations" (9) |
| `libp2p_pubsub_validation_failure_total` | counter | — | "LibP2P Validations" (9) |
| `libp2p_pubsub_validation_ignore_total` | counter | — | "LibP2P Validations" (9) |

## Connections / peers / streams

| Metric | Type | Labels used | Panels |
|---|---|---|---|
| `libp2p_peers` | gauge | `fleet` (also drives the Fleet template var via `label_values(libp2p_peers, fleet)`) | "LibP2P Peers" (2) |
| `libp2p_open_streams` | gauge | `type` | "LibP2P Open Streams" (3) |
| `libp2p_network_bytes_total` | counter | `direction` (in/out) | "Inbound Traffic" (78), "Outbound Traffic" (79) — rated |

## Dialing

| Metric | Type | Panels |
|---|---|---|
| `libp2p_total_dial_attempts_total` | counter | "LibP2P Dials" (7) |
| `libp2p_failed_dials_total` | counter | "LibP2P Dials" (7) |
| `libp2p_successful_dials_total` | counter | "LibP2P Dials" (7) |

## AutoNAT

| Metric | Type | Labels used | Panels |
|---|---|---|---|
| `libp2p_autonat_reachability_confidence` | gauge | `reachability` (Reachable / NotReachable) | "Node is Reachable (Experimental)" (81), "Node is NotReachable (Experimental)" (82) |

-----

Adding @AlbertoSoutullo to the loop to confirm whether they are using any of the removed metrics in their DST validations.
Cheers.

libp2p/protocols/pubsub/rpc/messages.nim
vladopajic: >  should also be called inside the encode function

decode is also check for encode. having validation in encode will not increase guarantees, because of how testing works. information source are types (type system), and encode-decode always happen in pairs in exactly this order - when we talk about testing. so if any error do happen in test, it will be stopped by `decode`.

types coming from users should also not be checked at encode because user actions should not propagate to that step (encode is last step before data hits wire) if they are invalid. this data should be validated as soon as user gives in the data. this is not only more efficient, but also gives users more informative errors early on.

vladopajic: > validation is being more strict due to not allowing ""

it was done because, ai explains it like this:

> Required fields must be explicitly set to a non-empty value for the message to be considered valid
> An empty string is treated as "not set" because the default value for string is ""

therefore, not allowing "" is exactly by protocol spec

vladopajic:
> Can we keep the protobuf library migration internal where possible and only use Opt when the absence of a field is meaningful and not just because the protobuf field is marked as optional? 

this pr has made effort following requirements from previous refactoring effort. please suggest approach that pubsub should take here. 

> having it optional in the public API does not provide much value

this was true for internal API as well, Opt just added bloat.

tersec:
> I am not convinced we should expose all protobuf field optionality as a breaking change on our public API... we'd be intentionally forcing the release of a new major version (if we respect semver) and changes downstream for something whose impact is limited to maintainability.
> 
> Looking downstream projects like nimbus and logos delivery use fields like the `message.data`, `message.topic` and `message.fromPeer` in validators, message providers, etc, and they would now need to add `.get()` or `valueOr` handling for data that we in nim-libp2p would have rejected or where providing defaults would mean no changes in the API while achieving the same result.
> 
> Can we keep the protobuf library migration internal where possible and only use Opt when the absence of a field is meaningful and not just because the protobuf field is marked as optional? and in cases where the public API changes are needed, to try keep them minimal? i.e. we are using optional topics because of `required` not being recommended to be used... but having it optional in the public API does not provide much value, since an `Opt.none` topic would still be considered invalid anyway. The same for the data being a `Opt[seq[byte]]`, when we could assume that an empty sequence means no data.
> 
> > old pubsub metrics have be abandoned in favor for general pubsub metrics (as explained in #2637).
> > still metrics will be utilized in later pr for all protocols/messages altogether
> 
> Are these abandoned pubsub metrics used in dashboards on logos-delivery / nimbus-eth2 ? I'd like to understand the impact this change might have. cc: @tersec , @Ivansete-status

Currently https://metrics.status.im/d/pgeNfj2Wz23/nimbus-fleet-testnets?orgId=1&var-instance=geth-02.ih-eu-mda1.nimbus.hoodi&var-container=beacon-node-hoodi-testing&from=now-24h&to=now&refresh=15m&timezone=utc shows one dashboard we use.

Some of the libp2p metrics there are
- `libp2p_open_streams`
- `libp2p_peers_traffic_read_total` (might not be working now)
- `libp2p_peers_traffic_write_total` (might not be working now)
- `libp2p_peers_identity` (might not be working now)

libp2p/protocols/pubsub/rpc/message.nim
richard-ramos: uuuh... :confused: . I'd have thought the compiler would have caught that. I guess file is not used?

libp2p/protocols/pubsub/rpc/messages.nim
richard-ramos: This AI explanation is incorrect or it is mixing proto3 behavior. For proto2 `required` means field presence, not non empty string. A field is present if it appears on the wire regardless of its value being empty.  


i.e.: In https://www.protobufpal.com/ you can see this behavior for the following protobuf: 
```proto
syntax = "proto2";

message Message {
  optional string topic = 1;
  optional string test = 2;
}
```
using the following JSONs, how the encoded output changes:

**Both fields have data**
```json
{
  "topic":"aaaa",
  "test": "bbbb"
}
```
Output (hex): `0a0461616161120462626262`. 

The encoding works like this (only for topic field):
- `0a` = field 1, wire type 2 (length-delimited)
- `04` = length 4
- `61616161` = `"aaaa"`


**Empty topic**
```json
{
  "topic":"",
  "test": "bbbb"
}
```
Output (hex): `0a00120462626262`
- `0a` = field 1
- `00` = length 0

**Topic is `none`**
```json
{
  "test": "bbbb"
}
```
Output (hex): `120462626262`
- Field 1 is omited

richard-ramos:
> > Can we keep the protobuf library migration internal where possible and only use Opt when the absence of a field is meaningful and not just because the protobuf field is marked as optional?
> 
> this pr has made effort following requirements from previous refactoring effort. please suggest approach that pubsub should take here.
> 
> > having it optional in the public API does not provide much value
> 
> this was true for internal API as well, Opt just added bloat.


Yeah... this is indeed an annoyance i guess due to proto2 fields optional fields having to be `Opt` or `PBOption`. 

IMO there are two possible paths:

1. update the spec/protobuf schema so presence is only modeled where it actually matters, using instead proto3 + explicit optional wherever a field not being present is significant. That would need coordination with libp2p/specs reviewers ofc.

or

2. It's not really necessary to make the raw protobuffer types part of the public API, because at the end of the day, these protobuffer types are meant for plumbing. You could have separate 'domain' types, for example:
```
# Internal type used just for protobuf serialization
type ProtoMessage {.proto2.} = object
  fromPeer: Opt[PeerId]
  data: Opt[seq[byte]]
  ...
  ...

# The public API type, basically the same type we have before this PR
type Message = object
  fromPeer*: PeerId
  data*: seq[byte]
  ...
  ...
```
with functions that convert from/to field types. That way we avoid polluting the API with `Opt`, but instead the protobuff types are used only where encoding/decoding happens, and this have the nice side effect of avoiding breaking changes

cc-ing @nitely and @arnetheduck for ideas too

libp2p/protocols/pubsub/rpc/messages.nim
nitely: > However, protobuf recommendations for proto2 say that we shouldnt use required. But, we still want this validation to happen, and currently it is being done only in here, but not when decoding a Message.

If the proto schema says the field is required then make it require. The problem is making a field require when the proto schema says it's optional. Also, making a field Opt when the schema says it's required may cause issues too, for example if the Opt field is not set when encoding. Also, proto3 *implicit* fields (optional is fine if set) seem incompatible with proto2 required, since default values won't appear in the encoded message, I mention this just in case you thought of implementing this as proto3.

AFAIK, the issue with required fields is they cannot be removed and they cannot be added, since they will fail to parse any old message lacking the field. This means the spec will never be able to remove (or change) those fields, but that's not a nim-libp2p problem.

> For proto2 required means field presence, not non empty string. A field is present if it appears on the wire regardless of its value being empty.

True.

richard-ramos: I see, makes sense to use required when spec already says to use required then. Thank you for the clarification.

nitely:
> update the spec/protobuf schema so presence is only modeled where it actually matters, using instead proto3 + explicit optional wherever a field not being present is significant. That would need coordination with libp2p/specs reviewers ofc.

afaik, it's not possible to switch a proto2 schema that contains required fields to proto3 in a backward compatible way. The current protobuf docs advice is to make [singular fields optional](https://protobuf.dev/programming-guides/proto3/#field-labels) to maximize proto2/proto3 compatibility. Which seems to be what libp2p is doing for new protocols.

But this is for defining proto file schemas. If you are implementing one of these schemas, and make a field optional as the schema says, but all you do is `the_field = opt_field.valueOr(raise newInvalidMessage)` then what's the difference with making the field implementation required? the only difference is the latter will fail at decode time instead of later in your application. cc @arnetheduck no?

The problem may be when encoding and sending messages; because the required field will always be present even if it should not because the value was not explicitly set. Then some application may treat this differently than if the field was optional and not set (ie: absent). But then I think you will have this issue anyway if you go with option 2, and convert public object to protobuf object (how to know if the Opt needs to be set or not if the public object field value is a default)... but I guess at least this gives you the chance to decide if the value should be set or not based on the field, which you don't have with required fields. Also, the separation of public object and protobuf object may be the cleaner approach.

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -232,22 +235,24 @@ proc getPeers(prune: ControlPrune, peer: PubSubPeer): seq[(PeerId, Opt[PeerRecor
...
+      SignedPeerRecord.decode(record.signedPeerRecord.get()).toOpt().withValue(spr):
+        if record.peerId.isSome and record.peerId.get() != spr.data.peerId:
           trace "peer sent envelope with wrong public key", peer
         else:
           peerRecord = Opt.some(spr.data)
       else:
         trace "peer sent invalid SPR", peer
 
-    routingRecords.add((record.peerId, peerRecord))
+    routingRecords.add((record.peerId.get(), peerRecord))
```
richard-ramos, line 250: Could this .get() trigger a defect if receiving a malformed PRUNE message with no peerId? (i.e. Opt.none(PeerId))

richard-ramos (approved):
lgtm except for one comment, i left below, but approving in the meantime.

I also went ahead and looked downstream how would nimbus and logos-delivery be affected by this PR:

Nimbus is affected in `eth2_network.nim:2497`:
- `m.data` is passed to `snappy.decode`; PR changes it to `Opt[seq[byte]]`.
- Validators use `message.data.len`, `snappy.decode(message.data, ...)`, and pass `message.fromPeer` as `PeerId` at `eth2_network.nim:2653 and `eth2_network.nim:2660`.
- Nimbus should probably reject messages missing `data` or `fromPeer` anyway (we gotta check that).

Logos Delivery is affected in `waku_relay/protocol.nim:271`  and `waku_relay/message_id.nim:26`:
- `msg.data` / `message.data` are decoded, hashed, and measured with `.len`; PR changes them to `Opt[seq[byte]]`.
- `graft.topicID` and `prune.topicID` are inserted into `HashSet[string]` at `waku_relay/protocol.nim:307`. PR changes those to `Opt[string]`. I am not sure cause i havent checked the code, but the impact downstream must be little because missing graft/prune topic IDs should likely be ignored at libp2p level

i think we can accept some breakage here because the changes make the code more compliant with the protobuf specification than it was before... The affected call sites are few and should be straightforward to fix, and we can send PRs to update those projects in the next release

libp2p/protocols/pubsub/gossipsub/behavior.nim
vladopajic: added check just it case

arnetheduck:
> I am not convinced we should expose all protobuf field optionality as a breaking change on our public API..

Indeed, I think this PR should be updated to not make this change to the public types - `Opt` adds (significant) overhead for `seq` for a type that is in the hot path - also, breaking the public api should not be done lightly and in this particular case, there's no reason to do so _semantically_, ie from a usage point of view, we did not need to semantically differentiate "empty" from "not present" which is what Opt does in this context. When presenting "high-level" API types, this is the key thing to look out for, ie is there semantic meaning attached to the full expressivity of the underlying protobuf encoding or can for example "empty" and "not present" be collapsed to a single representation.

The way to deal with this is to use a custom (de)serializer for the Message type specifically - ie the type annotations for protobuf-ser are meant to cover the 90% use case but then there's the remaining 10% which requires special care and attention (just like you don't optimize all code, just the hot path), you write manual overriders for writeValue/readValue .. 

Finally, this kind of "collapsing" of empty and not present might be common enough to warrant a "feature" in protobuf-ser, ie maybe as a field annotation. The reason we have PBOption and the like _by default_ is that these types are required to fully express the underlying protobuf encoding but sometimes that full expressivity is not needed and it would make sense for protobuf-ser to have an "opt-in" way of reducing the cases.

Finally, breaking the public API is very disruptive, for many reasons - also because it becomes harder to roll back to a previous version when things go wrong. When making breaking changes, these should ideally be available as a separate opt-in for a transition period using a separate type, with the old approach deprecated but not removed.

richard-ramos:
I'll implement this in a follow up PR today.

richard-ramos:
@arnetheduck , @nitely , @vladopajic 
let me know your thoughts in
- https://github.com/vacp2p/nim-libp2p/pull/2893
- https://github.com/status-im/nim-protobuf-serialization/pull/108

PR #2643, opened 2026-06-16
chore(kad): refine routing table peer admission

richard-ramos:
## Summary
Improves how Kademlia adds peers to the routing table.
Uses known public IP addresses to keep routing entries more balanced.

cc: @SionoiS

### DIFF
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -1,9 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 # Copyright (c) Status Research & Development GmbH
 
-import std/[tables, sequtils, algorithm]
+import std/[tables, sequtils, algorithm, net]
 import chronos, chronicles, results
-import ../../[peerid, peerinfo, switch, multihash, peeraddrpolicy]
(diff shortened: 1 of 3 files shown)


libp2p/protocols/kademlia/find.nim
```diff
@@ -201,21 +201,149 @@ proc dispatchFindNode*(
...
+  for ip in addrs:
+    if ip.subnet24() == subnet:
+      return true
+  false
+
+proc sharesSubnet64(addrs: seq[Ipv6Address], subnet: Ipv6Subnet64): bool =
+  for ip in addrs:
+    if ip.subnet64() == subnet:
+      return true
+  false
```
gmelodie, line 255: What if we made these more flexible as two `shareSubnet` with a `subnet` being a type that wraps an `int` between 1 -> 31 (ipv4) or 1 -> 127 (ipv6)

gmelodie (approved):
cool!

vladopajic (approved):

SionoiS (approved):
Good stuff! 🫡

libp2p/protocols/kademlia/find.nim
richard-ramos: pls expand on this

PR #2649, opened 2026-06-18
chore(examples): utilize protobuf_serialization

vladopajic:


gmelodie (approved):

examples/tutorial_3_protobuf.nim
```diff
@@ -174,7 +166,7 @@ proc main() {.async.} =
   await stream.close()
 
   for metric in metrics.metrics:
-    echo metric.name, " = ", metric.value
+    echo metric.name.get(), " = ", metric.value.get()
```
richard-ramos, line 169: For the example it is okay, but the fields being optional mean that the `.get()` could potentially fail IRL

richard-ramos (approved):
no proto3? :(

examples/tutorial_3_protobuf.nim
vladopajic: > could potentially fail IRL

this example is impossible to fail, `.get()` is guarded by `validate(m: Metric)`

> no proto3? :(

maybe there could be another example; proto2 was added because libpp2 uses proto2 mainly

richard-ramos: > proto2 was added because libpp2 uses proto2 mainly

:disappointed:

PR #2650, opened 2026-06-18
chore(signed-envelope): utilize protobuf-serialization

vladopajic:
utilizing `protobuf-serialization` for `Envelope` type.

- filed `domain` from `Envelope` type has been removed - it is not there in the spec
  - `domain` field was not really needed/used; but some places did check value of this field. since filed is removed new method `verify(domain)` was added that will check signature with specified domain. when `true` domain is correct, when `false` signature is is not correct (domain is not as expected. payload or key). in places where it was used it doesn't really matter which is not correct 
- `Envelope` proc for protobuf encoding now returns `seq[byte]` instead of ` Result[seq[byte], CryptoError]`

### DIFF
--- a/libp2p/protocols/rendezvous/rendezvous.nim
+++ b/libp2p/protocols/rendezvous/rendezvous.nim
@@ -327,8 +327,7 @@ proc advertise*[E](
     info "Can't create the signed peer record", error = error
     return
 
-  let pBuff = signedPeerRecord.encode().valueOr:
-    raise newException(AdvertiseError, "Wrong Custom Peer Record")
+  let pBuff = signedPeerRecord.encode()
   await rdv.advertise(ns, ttl, peers, Opt.some(pBuff))
 
 proc advertise*[E](
--- a/libp2p/protocols/rendezvous/rendezvous.nim
+++ b/libp2p/protocols/rendezvous/rendezvous.nim
@@ -364,8 +364,7 @@ method advertise*(
   if rdv.switch.isNil:
     # I don't like this, but adding this as i don't understand why we have a constructor without switch as arg
     raise newException(AdvertiseError, "Rendezvous not setup with a switch")
-  let sprBuff = rdv.switch.peerInfo.signedPeerRecord.encode().valueOr:
-    raise newException(AdvertiseError, "Wrong Signed Peer Record")
+  let sprBuff = rdv.switch.peerInfo.signedPeerRecord.encode()
   await rdv.advertise(ns, lttl, rdv.peers, sprBuff)
 
 proc requestLocally*[E](rdv: GenericRendezVous[E], ns: string): seq[E] =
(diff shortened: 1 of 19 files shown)


gmelodie (approved):

libp2p/protocols/rendezvous/rendezvous.nim
richard-ramos: 🤔 another case of code compiling when it should not? because `signedPeerRecord` is a `SignedPayload[E]`, and `encode` returns `seq[byte]` now. I dont understand how this compiles

```
proc encode*[T](msg: SignedPayload[T]): seq[byte] =
```

libp2p/signed_envelope.nim
```diff
@@ -67,49 +43,37 @@ proc init*(
...
 
-  envelope.signature = ?privateKey.sign(envelope.getSignatureBuffer())
+  envelope.signature = ?privateKey.sign(envelope.getSignatureBuffer(domain))
 
   ok(envelope)
 
-proc encode*(env: Envelope): Result[seq[byte], CryptoError] =
-  var pb = initProtoBuffer()
+proc verify*(envelope: Envelope, domain: string): bool =
+  envelope.signature.verify(envelope.getSignatureBuffer(domain), envelope.publicKey)
```
richard-ramos, line 55: the `Envelope` being `proto3` means that fields `publicKey` can be omited. When that happens we will receive a default value, but that's not good, for the type `PublcKey` the default value is invalid, because it will be a:
```
PublicKey(scheme: RSA, rsakey: nil)
```
so when `envelope.signature.verify(buffer, envelope.publicKey)` is called, it will trigger this assertion if the protobuffer includes a signature, but not a public key:
```
  doAssert((not isNil(sig)) and (not isNil(pubkey)))
```
https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/crypto/rsa.nim#L794

And having a valid public key and no signature will cause the verifier to return a false anyway, thus making the envelope invalid anyway...

so here we should likely validate that if we are including both a valid public key (by using `not envelope.publicKey.getBytes().isErr`), and a non empty signature

vladopajic: should this be checked in `decode` instead?

eg

```
  if envelope.publicKey.getBytes().isErr:
    return err(EnvelopeFieldMissing)
  if envelope.signature.data.len == 0:
    return err(EnvelopeFieldMissing)
```

libp2p/protocols/rendezvous/rendezvous.nim
vladopajic: just checked, tests compile without this proc (commented out) - it is unused

libp2p/signed_envelope.nim
richard-ramos: that could work!

richard-ramos (approved):

PR #2652, opened 2026-06-18
chore(gossipsub): cap partial-message group state

gmelodie:
Adds upper bounds on the number of partial-message groups GossipSub keeps in memory, so a flood of distinct (topic, group) ids cannot grow the per-extension state without limit.

Two independent caps:
- `maxPartialGroups` (default 1000) — global cap. Enforced lazily on each heartbeat: when the group count exceeds the cap, the least-recently-used groups are trimmed. Recency is the existing `heartbeatsTillEviction` counter, refreshed whenever a group sees new metadata, so actively used groups are trimmed last.
- `maxPartialGroupsPerPeer` (default 100) — per-peer cap. A group created from a peer's received metadata is attributed to that peer; once the peer is at its budget, further new groups are rejected and never allocated.

Each group records its creating peer. The per-peer budget is released when a group is evicted (heartbeat expiry or LRU trim) and when the peer is removed; on peer removal, groups the peer created that are now empty are evicted immediately instead of lingering until heartbeat expiry. Locally-published groups are uncounted (no creator) and bypass the per-peer cap, but remain subject to the global LRU trim.

Two metrics:
- `libp2p_gossipsub_partial_message_groups` — gauge of groups currently tracked, updated at each allocation/eviction.
- `libp2p_gossipsub_partial_message_groups_dropped` — counter of groups that did not survive a limit, covering both per-peer rejections and global LRU trims.

Tests cover the global LRU trim (including least-recently-used ordering), the per-peer cap, and budget reclaim on both peer removal and heartbeat eviction.

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
+++ b/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 # Copyright (c) Status Research & Development GmbH
 
-import tables, sets, sequtils, chronicles, results, metrics
+import tables, sets, sequtils, algorithm, chronicles, results, metrics
 import ../../../utils/tablekey
 import ../../../[peerid]
 import ../rpc/messages
@@ -63,9 +63,8 @@ type
       # needs to be implemented by application.
     heartbeatsTillEviction*: int
       # number of heartbeats for which metadata will be retained before eviction
-    maxPartialGroups*: int = 1000
-      # upper bound on (topic, group) entries; new groups are rejected at the limit
-    maxPartialGroupsPerPeer*: int = 100 # upper bound on groups a single peer can create
+    maxPartialGroups*: int = 1000 # global cap; excess trimmed LRU each heartbeat
+    maxPartialGroupsPerPeer*: int = 100
 
   PeerGroupState = ref object
     receivedPartsMetadata: Opt[PartsMetadata] # parts metadata peer sent to this node
@@ -146,6 +145,7 @@ proc evictGroup(ext: PartialMessageExtension, key: TopicGroupKey) =
   group.creator.withValue(creator):
     ext.releasePeerGroup(creator)
   ext.groupState.del(key)
+  ext.updateGroupCountMetric()
 
 proc reduceHeartbeatsTillEviction(ext: PartialMessageExtension) =
   var toRemove: seq[TopicGroupKey] = @[]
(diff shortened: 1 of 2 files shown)


libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -250,6 +314,18 @@ proc shouldHandlePartialRPC(
...
   groupState != nil and groupState.hasPublished()
 
+proc recordReceivedMetadata(
+    ext: PartialMessageExtension, peerId: PeerId, rpc: PartialMessageExtensionRPC
+) =
+  if rpc.partsMetadata.len == 0:
+    return
+  let groupState = ext.acquireGroupState(peerId, rpc.topicID, rpc.groupID)
+  if groupState.isNil():
+    return
```
vladopajic, line 350: deepening on how limits are configured, and how nimbus expects to use it this can eater be okay strategy or sub-optimal. 
- if limits are comfortable to fit honest peers usage - all good
- if limits are not comfortable for hones peers or they are but there are few peers that will max out their limit in order to get everyone to `maxPartialGroups` all new metadata, from everyone, will be discarded (not stored).

it seems that instead of forbidding storing metadata better strategy in all cases would be to always allow storing but evict (probably oldest) group state before storing new. 

if above is true then, does it make sense to:
- keep all code as it is (for simplicity) and just trim excess groups in heartbeat phase (oldest first)? have `maxPartialGroups` config, in each heartbeat trim anything old down to this limit. this strategy drastically reduce possible memory overflow with groups. 
- ^ but still one peer could be spamming and taking everyone space. (we should check if this is possible after first)
   - if they really are spamming (sending a lot of messages) then better disconnect and add penalty 
   - it also makes sense to have peer peer limit, but it better be (???) implemented like circular buffer (always allow new data)
- state groups that are created by node (data is not received by rpc) should never be in these limits. only heartbeatsTillEviction should apply to them

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -114,20 +130,64 @@ proc peerRequestsPartial*(
...
+    return nil
+
+  let group = GroupState(creator: Opt.some(peerId))
+  ext.groupState[key] = group
+  ext.peerGroupCount[peerId] = ext.peerGroupCount.getOrDefault(peerId) + 1
+  ext.updateGroupCountMetric()
+  group
+
+proc trackedGroups*(ext: PartialMessageExtension): int =
+  ext.groupState.len
```
vladopajic: `libp2p_testing` perhaps

we should start using `libp2p_testing` flag bit more, just to get used to it. it looks like we cant go wrong with it we only use this flag to export things or to expose some properties (like here) aka getters. i don't think we should use it for setter or any conditional code compile (that is not export or getter).

the only risk is, exported things will be visible in test builds, so we can build feature that only works in test but doesn't work in non tests. but we should be good with this because 1) it's unlikely for developer to make this mistake (+ there is code review, etc) 2) there are other non testing build is ci that build different binaries - those can fail in this case

vladopajic (approved):

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -76,6 +89,7 @@ type
     config: PartialMessageExtensionConfig
     groupState: Table[TopicGroupKey, GroupState]
     peerTopicOpts: Table[PeerTopicKey, TopicOpts]
+    peerGroupCount: Table[PeerId, int] # groups allocated per peer, keyed by creator
```
vladopajic: CountTable[PeerId]

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: also, maybe split effort LRU + maxgroups and per peer limit after (if we see above as reasonable)

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -54,6 +63,9 @@ type
       # needs to be implemented by application.
     heartbeatsTillEviction*: int
       # number of heartbeats for which metadata will be retained before eviction
+    maxPartialGroups*: int = 1000
```
richard-ramos: Should `publishPartial` also check for `maxPartialGroups`?

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -203,6 +263,10 @@ method onRemovePeer*(
   # remove peer data from _groupState_
   for key, group in ext.groupState:
     group.peerState.del(peerId)
+    group.creator.withValue(creator):
+      if creator == peerId:
+        group.creator = Opt.none(PeerId)
+  ext.peerGroupCount.del(peerId)
```
richard-ramos: Should evictGroup be called for the peer groups too?

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
vladopajic: not really, node should always be able to publish, and limits should not apply to node. node will never do memory attack on themself.

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
```diff
@@ -114,20 +130,64 @@ proc peerRequestsPartial*(
...
-  # reduce heartbeatsTillEviction and remove groups that hit 0
   var toRemove: seq[TopicGroupKey] = @[]
   for key, group in ext.groupState.mpairs:
     group.heartbeatsTillEviction.dec
     if group.heartbeatsTillEviction <= 0:
       toRemove.add(key)
   for key in toRemove:
-    ext.groupState.del(key)
+    ext.evictGroup(key)
+  ext.updateGroupCountMetric()
```
vladopajic: `updateGroupCountMetric` not sure if this really needs to be that precise (update on each change) - or it could be simplified to calling it just once on each heartbeat. would simplify other code as well (`getGroupState` could remain as template, etc... )

vladopajic: 700ms is heartbeat for nimbus

libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim
gmelodie: I like the LRU idea, lemme see if that's too much for a single PR, otherwise I'll split in two

richard-ramos (approved):

vladopajic (approved):

PR #2653, opened 2026-06-18
chore(ci): parallelize test execution

gmelodie:
## Summary

`make test` built and ran the `test_all` slices and `test_multiformat_exts` strictly serially, which dominated CI wall-clock time. This adds a `TEST_JOBS` knob so the independent test binaries are **run** concurrently via `make -j`, while compilation stays serial.

- `TEST_JOBS ?= 1` keeps local `make test` serial by default; CI sets it to 3.
- `test` now compiles all slices serially (`compile_tests`) and then runs the resulting binaries under `make -j$(TEST_JOBS)` (`run_tests`).
- Compilation is intentionally left serial: a single `nim c` already parallelizes its C backend across all cores, so compiling slices concurrently mostly oversubscribes the C compiler while multiplying peak memory.
- Per-slice `--nimcache` is now unconditional (previously gated behind `CICOV`) so each binary keeps its own cache and coverage `.gcda` files stay isolated.

## Affected Areas

- [x] Build / Tooling
  Makefile test targets and the CI/coverage/daily workflows.

N/A — CI/build-tooling only; no library code changes, so downstream builds are unaffected.

## Impact on Library Users

No impact. Internal build/CI change only; no API or behavior changes to the library.

## Risk Assessment

- Compilation stays serial, so the 32-bit ~3 GB compiler address-space limit is unaffected by this change.
- Main risk is runtime contention: under `make -j` multiple test binaries run concurrently, so tests that bind fixed ports or share fixtures could flake under load. Local `make test` stays serial (`TEST_JOBS=1`) by default.

## References

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,8 +177,13 @@ jobs:
             export NIMFLAGS="${NIMFLAGS} --cc:clang"
           fi
 
-          # Compile test binaries serially, run them concurrently.
-          make TEST_JOBS=3 test
+          # Build/run test binaries concurrently; fewer lanes on 32-bit to stay
+          # under the ~3 GB address cap.
+          TEST_JOBS=3
+          if [[ '${{ matrix.platform.cpu }}' == 'i386' ]]; then
+            TEST_JOBS=2
+          fi
+          make TEST_JOBS=$TEST_JOBS test
 
       - name: Fix xml newlines
         if: ${{ !cancelled() && steps.run-tests.outcome == 'failure' }}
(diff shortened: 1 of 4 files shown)


vladopajic (approved):
would be nice to know the difference - run 3 times same ci and compare it to master.
these machines that run ci are probably already fully utilized when compiling single binary. compiling two binaries simultaneously (`$(MAKE) -j$(TEST_JOBS) _run_all_tests`)  might not have any benefit in which case it would be better to keep things simple. maybe `--nimcache` alone is enough. 

maybe we can try `--incremental:on`?

richard-ramos (approved):

PR #2663, opened 2026-06-23
test(peerstore): address TTL 2

rlve:
## Summary

Adds one component test to the PeerStore Address TTL suite that documents what happens to a peer's stored address on a long-lived but idle connection.

An address's `lastUpdated` only moves when something writes it again: a re-dial, a fresh identify, or an identify-push from the peer. On a connection that stays open without any of those, nothing refreshes the timestamp. Once the TTL for the entry's confidence band elapses, the background pruning loop removes it, even though the switch is still connected to that peer.

**Question**: is it an issue or an acceptable situation?


tests/libp2p/test_peer_store_component.nim
```diff
@@ -171,3 +171,37 @@ suite "PeerStore Address TTL - Component":
...
+      dialer.peerStore[AddressBook].entries(peerId).len > 0
+
+    # Nothing refreshes the address.
+    # Back-date the entries beyond the TTL.
+    for i in 0 ..< dialer.peerStore[AddressBook].book[peerId].len:
+      dialer.peerStore[AddressBook].book[peerId][i].lastUpdated = Moment.now() - 2.hours
+
+    # The prune loop drops the old address while the connection is still open.
+    checkUntilTimeout:
+      dialer.peerStore[AddressBook].entries(peerId).len == 0
```
vladopajic, line 219: interesting. i guess it would be better to save as long as peer is connected. or TTL should only count after peer disconnects.

tests/libp2p/test_peer_store_component.nim
```diff
@@ -171,3 +171,37 @@ suite "PeerStore Address TTL - Component":
...
       switch.peerStore[AddressBook].entries(zeroTtlPeer).len == 1
       switch.peerStore[AddressBook].entries(infinitePeer).len == 1
+
+  asyncTest "connected-but-idle address expires while still connected":
+    # An address only gets a fresh timestamp when it is written again:
+    # on a re-dial, an identify, or an identify push from the peer.
+    # If a connection does none of these, the address gets stale and the
+    # prune loop drops it while we are still connected to the peer.
+    let
+      listener = makeStandardSwitch(TcpAutoAddress)
```
vladopajic: consider not specifying address (`TcpAutoAddress`) unless it is necessary. this test should be transport agnostic afaik?

vladopajic (approved):

tests/libp2p/test_peer_store_component.nim
rlve: Yeah, I'm flagging it to decide if it requires an action or not.

gmelodie (approved):

tests/libp2p/test_peer_store_component.nim
vladopajic: @gmelodie @richard-ramos

rlve: Since the PR is merged, I opened an issue to track this: https://github.com/vacp2p/nim-libp2p/issues/2684

PR #2667, opened 2026-06-23
fix(pubsub): cap number of topics a peer can subscribe to

gmelodie:
## Description

Adds a per-peer limit on the number of distinct topics a peer can be tracked as subscribed to, and uses it to bound the subscription bookkeeping we keep for each peer.

Previously `topicsHigh` defaulted to `int.high`, so the per-peer subscription state was effectively unbounded. This PR sets a sensible default (`DefaultTopicsHigh = 1024`) and enforces it when handling incoming `SUBSCRIBE` messages: once a peer reaches the limit, further subscriptions to new topics are ignored.

## Behavior change

The default `topicsHigh` changes from `int.high` (unlimited) to `1024`. Deployments that legitimately track more than 1024 distinct topics per peer must now set `topicsHigh` explicitly, otherwise subscriptions to new topics beyond the limit are dropped.

## Changes

- Add `DefaultTopicsHigh = 1024` and use it as the default `topicsHigh` instead of `int.high`.
- Track the number of distinct subscribed topics per peer via a new `subscribedTopics` counter on `PubSubPeer`, kept in sync on subscribe/unsubscribe.
- In both FloodSub and GossipSub, skip adding a subscription for a new topic once the peer is at `topicsHigh` (existing subscriptions are unaffected).
- In GossipSub, apply the existing `behaviourPenalty` and a new `libp2p_gossipsub_over_topics_high_subscription` metric when a subscription is skipped for exceeding the limit.
- Factor the repeated `0.1` penalty value into a named `SubscriptionFloodPenalty` constant.
- Use the `addPeer`/`removePeer`/`hasPeer` helpers consistently in FloodSub's subscribe handling.
- On peer disconnect, FloodSub's `unsubscribePeer` now removes the peer via `removePeer`, which deletes emptied topic entries instead of leaving empty sets behind. This prevents the topic table from growing unbounded across reconnect cycles.

## Tests

- FloodSub: subscriptions beyond `topicsHigh` are not added; a freed slot can be reused after unsubscribing.
- FloodSub: `unsubscribePeer` clears emptied topic entries from the table.
- GossipSub: subscriptions beyond `topicsHigh` are rejected and the peer's behaviour penalty / metric are updated.


vladopajic (approved):

libp2p/protocols/pubsub/floodsub.nim
richard-ramos: This should also delete empty topics, because otherwise imagine you subscribe to 1024 topics, disconnect, and reconnect, and subscribe to new topics. This would still grow . I wonder if we should use f.floodsub.removePeer like gossipsub does? i.e. something like
```
for t in toSeq(f.floodsub.keys):
  f.floodsub.removePeer(t, pubSubPeer)

```

richard-ramos (approved):

PR #2668, opened 2026-06-23
feat(cbind): user-provided XPRs

gmelodie:
## Summary

Adds a C API utility, `libp2p_create_xpr`, that lets an encapsulating application build and sign its **own node's** Extensible Peer Record (XPR) — choosing the advertised service set, addresses, and `seqNo` — and get the signed, protobuf-encoded bytes back.

Previously the only XPR path was the record auto-generated internally by service discovery during advertising; there was no cbind entry point to construct a `SignedExtendedPeerRecord` directly.

```c
int libp2p_create_xpr(libp2p_ctx_t *ctx, const char **addrs, size_t addrsLen,
                      const Libp2pServiceInfo *services, size_t servicesLen,
                      uint64_t seqNo, Libp2pBufferCallback callback,
                      void *userData);
```

Behavior:

- The record is signed on the worker thread with the node's own `peerInfo.privateKey`, so no private key crosses the FFI boundary. As a consequence, `peerId` is always the node's own — this call builds *the node's* record, not an arbitrary peer's.
- Unsupplied addresses (`addrs == NULL` / `addrsLen == 0`) fall back to the node's listen addresses.
- `seqNo == 0` defaults to the current unix time (`0` is used as the "unset" sentinel since the C ABI takes a plain `uint64_t`).
- Inputs are validated at the boundary: a non-NULL length with a NULL `addrs`/`services` pointer, or a service with `dataLen > 0` and NULL `data`, fails via the callback. The encoded record is checked against `MaxServiceDataSize` / `MaxXPRSize` before being returned.

The validate + sign + size-check logic shared with the existing `ServiceDiscovery.record()` is factored into a single `SignedExtendedPeerRecord.build` helper in `extended_peer_record.nim`, and `record()` is refactored to call it so the two paths can't drift.

## Affected Areas

- [x] Build / Tooling
  

`libp2p_create_xpr` is purely additive. The `record()` refactor preserves behavior — same validation, signing, and size check in the same order — with the only observable change being unified error-message strings. No downstream integration required.

## Impact on Library Users

New C API; no behavior change to existing APIs.

- `libp2p_create_xpr` is added to `cbind/libp2p.h`.
- Signing uses the running node's key, so the node context (`ctx`) is required. This call does **not** advertise the record — it only returns the signed bytes.
- New internal `SignedExtendedPeerRecord.build` helper in `extended_peer_record.nim` (validate + sign + size-check in one place), now shared with `ServiceDiscovery.record()`.

## Risk Assessment

Low. The request reuses the existing worker-thread request / `READ` buffer-callback machinery and the shared `build` helper. It does not require service discovery to be mounted (it only reads the switch's `peerInfo`). The `record()` refactor changes no behavior.

## Scope / Follow-up

This PR is the create-and-sign half only: it returns signed XPR bytes but nothing in cbind consumes them yet. Advertising a user-provided XPR would require `libp2p_service_disco_start_advertising` to accept optional pre-built XPR bytes and forward them to the existing `startAdvertising(service, advert)` path in the Nim layer. That wiring is intentionally left out and tracked as follow-up.

## References

Branch: `feat/cbind/user-provided-xprs`

### DIFF
--- a/cbind/libp2p.nim
+++ b/cbind/libp2p.nim
@@ -1075,6 +1075,17 @@ proc libp2p_create_xpr(
   initializeLibrary()
   checkLibParams(ctx, callback, userData)
 
+  if addrsLen > 0 and addrs.isNil():
+    failWithBufferMsg(callback, userData, "addrs are not set")
+
+  if servicesLen > 0 and services.isNil():
+    failWithBufferMsg(callback, userData, "services are not set")
+
+  let serviceArray = cast[ptr UncheckedArray[Libp2pServiceInfo]](services)
+  for i in 0 ..< servicesLen.int:
+    if serviceArray[i].dataLen > 0 and serviceArray[i].data.isNil():
+      failWithBufferMsg(callback, userData, "service data is not set")
+
   libp2p_thread.sendRequestToLibP2PThread(
     ctx,
     RequestType.SERVICE_DISCOVERY,
--- a/cbind/libp2p.nim
+++ b/cbind/libp2p.nim
@@ -1062,6 +1062,39 @@ proc libp2p_service_disco_start_advertising(
 
   RET_OK.cint
 
+proc libp2p_create_xpr(
+    ctx: ptr LibP2PContext,
+    addrs: ptr cstring,
+    addrsLen: csize_t,
+    services: ptr Libp2pServiceInfo,
+    servicesLen: csize_t,
+    seqNo: uint64,
+    callback: Libp2pBufferCallback,
+    userData: pointer,
+): cint {.dynlib, exportc, cdecl.} =
+  initializeLibrary()
+  checkLibParams(ctx, callback, userData)
+
+  libp2p_thread.sendRequestToLibP2PThread(
+    ctx,
+    RequestType.SERVICE_DISCOVERY,
+    ServiceDiscoveryRequest.createSharedXpr(
+      addrs = addrs,
+      addrsLen = addrsLen,
+      services = services,
+      servicesLen = servicesLen,
+      seqNo = seqNo,
+    ),
+    callback,
+    CallbackKind.READ,
+    userData,
+  ).isOkOr:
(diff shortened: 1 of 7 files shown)


vladopajic (approved):

cbind/libp2p.nim
```diff
@@ -1062,6 +1062,39 @@ proc libp2p_service_disco_start_advertising(
...
+    addrsLen: csize_t,
+    services: ptr Libp2pServiceInfo,
+    servicesLen: csize_t,
+    seqNo: uint64,
+    callback: Libp2pBufferCallback,
+    userData: pointer,
+): cint {.dynlib, exportc, cdecl.} =
+  initializeLibrary()
+  checkLibParams(ctx, callback, userData)
+
```
richard-ramos, line 1081:
```
  if servicesLen > 0 and services.isNil():
    failWithMsg(callback, userData, "services are not set")
```

cbind/libp2p_thread/inter_thread_communication/requests/libp2p_service_discovery_requests.nim
```diff
@@ -37,9 +46,44 @@ proc createShared*(
...
+  ret[].seqNo = seqNo
+  ret[].servicesLen = servicesLen.int
+
+  if servicesLen > 0 and not services.isNil():
+    ret[].services = cast[ptr UncheckedArray[SharedService]](allocShared0(
+      sizeof(SharedService) * servicesLen.int
+    ))
+    let src = cast[ptr UncheckedArray[Libp2pServiceInfo]](services)
+    for i in 0 ..< servicesLen.int:
+      ret[].services[i].id = src[i].id.alloc()
```
richard-ramos, line 69:
```suggestion
      if src[i].dataLen > 0 and src[i].data.isNil():
        failWithBufferMsg(callback, userData, "service data is nil")
      ret[].services[i].id = src[i].id.alloc()
```

PR #2669, opened 2026-06-23
chore: fully remove minprotobuf

vladopajic:
final effort that fully removes `minprotobuf`


libp2p/protobuf/minprotobuf.nim
richard-ramos, line 1: minprotobuf is actively being used in logos-delivery. We have to deprecate this and open an issue there to encourage their migration to protobuf-serialiation

richard-ramos: It's also used in logos-storage

vladopajic: issues opened. i'll close this pr and open new one that does final cleanup without removing minprotobuf code

PR #2670, opened 2026-06-24
fix(connmanager): emit Joined before trimming new peers

### LINKED ISSUE #2621: conn-manager: trim of the just-stored connection emits Left before Joined
### Description

When a new connection pushes the peer count over `highWater`, `storeMuxer` triggers the watermark trim inline, before it emits `Connected` and `Joined` for the stored peer. The trim can prune the just-stored connection itself, and the drop emits `Left` and `Disconnected`. If the `Connected` handlers finish after the connection close completes, the events come out as `Left, Disconnected, Connected, Joined`.

Preconditions for the wrong order:
- `gracePeriod` is zero, the field default, so the just-stored connection is a trim candidate
- protected or graced peers force the trim down to the just-stored connection
- a `Connected` handler that awaits, delaying `Joined` past the close

The potential issue: consumers that create peer state on `Joined` and destroy it on `Left` are left with stale state, since the destroy runs before the create. For example pubsub subscribes on `Joined` and unsubscribes on `Left`, so a node can end up with a gossipsub subscription for the pruned peer.

A second effect of the same root cause: `storeMuxer` calls `notifyPeerReady` after the trim already dropped the peer and cleared its ready state, so `waitForPeerReady` keeps reporting the pruned peer ready. This needs the whole drop to complete without suspending, so it reproduces with in-memory connections but not over TCP or QUIC.

A `gracePeriod` above zero exempts the new connection from its own trim, so it is an effective mitigation. Deferring the trim until after the `Connected` and `Joined` emission, or excluding the just-stored peer from trim candidates, would fix the order.

richard-ramos:
## Summary

This fixes connection-manager lifecycle ordering when watermark trimming prunes a newly stored connection. The peer now emits `Connected` and `Joined` before trim can emit `Left` and `Disconnected`, preserving the order expected by event consumers such as pubsub.

The regression coverage also checks that a peer pruned during trim does not remain in the ready state.

## Affected Areas

- [x] Peer Management / Discovery - connection-manager watermark trimming and peer lifecycle events.

## Impact on Library Users

No API changes. Event consumers should observe a consistent lifecycle for peers pruned immediately after connection: `Connected`, `Joined`, `Left`, `Disconnected`.

## Risk Assessment

Low. The change is scoped to connection-manager bookkeeping around newly stored peers in watermark mode. Trim is intentionally delayed until after the joined peer event is processed.

## References

- Fixes #2621


richard-ramos:
Running also CI in nimbus https://github.com/status-im/nimbus-eth2/pull/8650

libp2p/connmanager.nim
```diff
@@ -471,12 +475,21 @@ proc storeMuxer*(
...
+    c.peerEventFuts.trackFut(joinedEvent)
+
+  if c.watermark.isSome:
+    if isNewPeer:
+      c.connectedAt[peerId] = Moment.now()
+    if c.muxerStore.countPeers() > c.watermark.get().highWater:
+      if joinedEvent.isNil:
+        c.triggerTrim()
+      else:
+        c.peerEventFuts.trackFut(c.triggerTrimAfter(joinedEvent))
```
vladopajic: would be nicer if all code related to staring trim is in one place, now trim is started here and in `triggerTrimAfter`. 

for example, code here can be only this:
```
if c.muxerStore.countPeers() > c.watermark.get().highWater:
   triggerTrimAfter(joinedEvent)
```

and code inside `triggerTrimAfter` can encapsulate all logic for doing `triggerTrim` right away or waiting if future is not nil..

vladopajic (approved):

gmelodie (approved):

Issue #2671, opened 2026-06-24
remove `minprotobuf` related code


vladopajic:
after downstream deps transition off `minprotobuf` remove all `minprotobuf` related code.

vladopajic:
closing in favor of #2615

PR #2672, opened 2026-06-24
chore: removing usage of minprotobuf

vladopajic:
final effort that removes remaining `minprotobuf` usage. 
usage in PeerId, MultiAddress remains as it still might be useful to downstream deps. 

after downstream deps transition off `minprotobuf` it will be trivial to just remove `minprotobuf` related code (it's delete only action). https://github.com/vacp2p/nim-libp2p/issues/2671


gmelodie (approved):
Image

richard-ramos (approved):
lgtm 🚀

PR #2674, opened 2026-06-24
chore(ci): run each test subsystem as its own CI job

gmelodie:
## Summary

CI built and ran the whole test suite concurrently inside one step via `make -j$(TEST_JOBS) test`, so the slices' output interleaved into a single unreadable log. The first attempt at fixing this split the suite into `make -j`-free index slices (`test_slice_0`, `test_slice_1`), but a slice number is meaningless to anyone reading CI and its membership silently shifts whenever a test file is added.

This instead partitions the suite into named subsystem groups (`pubsub`, `discovery`, `transports`, `protocols`, `crypto`, `utils`, `misc`), each run as its own CI job. A red job now names the affected subsystem instead of an opaque slice, and every job's log is self-contained.

`tests/test_groups.json` is the single source of truth: the CI matrix derives its job list from it (a `compute-matrix` job emits `jq -c 'keys + ["multiformat"]'`, consumed via `fromJSON`), and the test harness (`tests/imports.nim`) reads the same file to decide which files each group compiles. They cannot drift.

The harness asserts at compile time that the partition stays total and disjoint: every test file maps to exactly one group, no named group matches zero files (so a renamed directory fails the build instead of silently dropping its tests), and exactly one catch-all (`misc`) exists. Groups are also each a small enough translation unit to keep the 32-bit Nim compiler under its ~3 GB address cap, which the unsharded suite overran on orc.

## Affected Areas

- [x] Build / Tooling
  CI workflows (`ci.yml`, `daily_common.yml`, `coverage.yml`), the test `Makefile`, and the test import harness (`tests/imports.nim`, `tests/test_all.nim`, `tests/test_groups.json`).

N/A — CI/tooling-only change. No library code is touched, and the set of test files that runs is unchanged (143 files), only how they are bucketed into jobs.

## Impact on Library Users

No impact. Internal CI/build change only; no API, behavior, or performance change to the library.

## Risk Assessment

- Increases CI job count per platform/Nim combination (one job per subsystem group + multiformat), in exchange for clean logs and a failure that names the affected area.
- Local `make test` now compiles every group in one translation unit; on a 32-bit/low-RAM machine this can exceed the compiler's address cap (CI is unaffected — it builds one group per job). `make test_group GROUP=<name>` builds a single group.
- The `misc` catch-all is a broad bucket (core libp2p + interop + tools), so a failure there is less specific than the named groups.

## References

None.

## Additional Notes

Group names and the directories they cover live in `tests/test_groups.json`; add a directory to a group's pattern list (or rely on the `misc` catch-all) and the compile-time check guarantees it still runs somewhere.


.github/workflows/ci.yml
```diff
@@ -47,6 +47,13 @@ jobs:
             memory_management: refc
           - ref: v2.2.10
             memory_management: orc
+        # Each test slice and the multiformat suite runs as its own job so
+        # their logs stay separate instead of interleaving under one `make -j`.
+        # Slice count must match TEST_ALL_SLICES in the Makefile.
+        target:
+          - test_slice_0
```
vladopajic: what is `test_slice_0` ?

vladopajic: ah so all tests are split into 2 groups?

vladopajic: i am not sure if we should do that, like that. ci pays the cost 2 times for compiling tests (ci will compile same things all over in all workflows). so it will only speed execution. but for execution there is better approach with `unittest3` which i should complete at one point.

however it should be fine to do these two separately:
- test_multiformat_exts
- test_all 

i would prefer simpler ci code over splitting

gmelodie: @vladopajic  Oh this pr is not ready for review yet. 

> but for execution there is better approach with unittest3 which i should complete at one point.

this is the actual proper way to do it indeed

gmelodie:
This PR bloats CI usage from 3h currently to 13h! No good.

PR #2677, opened 2026-06-25
chore(protobuf): main type encode/decode metrics

vladopajic:
- adding encode/decode metrics for all main types
- added `domain` argument to `serializerFor` to avoid any clash when types have same name (kademlia: Message, rendezvous: Message, etc....)
- added `trackEncodeBytes` and `trackDecodeBytes` utility templates for custom (manual) encode/decode implementations

### DIFF
--- a/libp2p/protocols/pubsub/rpc/protobuf.nim
+++ b/libp2p/protocols/pubsub/rpc/protobuf.nim
@@ -30,12 +30,12 @@ proc decode*(_: type Message, buf: seq[byte]): Result[Message, string] =
 
 proc encode*(msg: RPCMsg, anonymize: bool): seq[byte] =
   let encoded = encode(Protobuf, msg.anonymize(anonymize))
-  trackEncodeBytes(encoded.len, $RPCMsg, "gossipsub")
+  trackEncodeBytes(encoded.len, RPCMsg, "gossipsub")
   encoded
 
 proc decodeRPCMessage(buf: seq[byte]): RPCMsg {.raises: [SerializationError].} =
   let msg = decode(Protobuf, buf, RPCMsg)
-  trackDecodeBytes(buf.len, $RPCMsg, "gossipsub")
+  trackDecodeBytes(buf.len, RPCMsg, "gossipsub")
   msg
 
 proc decode*(_: type RPCMsg, buf: seq[byte]): Result[RPCMsg, string] =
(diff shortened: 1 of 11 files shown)


vladopajic:
kademlia has these metrics:
```
#received
kad_messages_received.inc(labelValues = [$msgType])
kad_message_bytes_received.inc(bufLen.int64, labelValues = [$msgType])

#sent
kad_messages_sent.inc(labelValues = [$MessageType.findNode])
kad_message_bytes_sent.inc(encoded.len.int64, labelValues = [$MessageType.findNode])
```

seems like protobuf can have `protobuf_messages_received` and `protobuf_messages_sent` metrics as well.
we could also extend protobuf label adding one more value that can be type of message. 
then protobuf metrics could replace these metrics in kademlia. it would be more elegant because kad metrics are all over the place and if they were in encode/decode it would be single location.

gmelodie (approved):

libp2p/utils/protobuf.nim
```diff
@@ -8,16 +8,29 @@ import std/macros, results, protobuf_serialization
...
+
+macro decodeFor*(
+    _: type Protobuf, Types: untyped, withMetrics: bool = false, domain: string = ""
+): untyped =
   ## This generates decode protobuf procs for `Types`
   let doMetrics = newLit(withMetrics.eqIdent("true"))
   var stmts = newStmtList()
   for T in Types:
     let decodeName = ident("decode" & $T)
+    let metricLabel = newLit(domain.strVal & "." & $T)
```
richard-ramos: can the "." be appended depending of domain being `""` or not? because otherwise you can end up with a ".MetricName"  without domain if you forget to set it up

libp2p/protocols/pubsub/rpc/protobuf.nim
```diff
@@ -29,10 +29,14 @@ proc decode*(_: type Message, buf: seq[byte]): Result[Message, string] =
...
 proc encode*(msg: RPCMsg, anonymize: bool): seq[byte] =
-  encode(Protobuf, msg.anonymize(anonymize))
+  let encoded = encode(Protobuf, msg.anonymize(anonymize))
+  trackEncodeBytes(encoded.len, RPCMsg, "gossipsub")
+  encoded
 
 proc decodeRPCMessage(buf: seq[byte]): RPCMsg {.raises: [SerializationError].} =
-  decode(Protobuf, buf, RPCMsg)
+  let msg = decode(Protobuf, buf, RPCMsg)
+  trackDecodeBytes(buf.len, RPCMsg, "gossipsub")
```
richard-ramos:
```suggestion
  trackDecodeBytes(buf.len, RPCMsg, "gossipsub")
  let msg = decode(Protobuf, buf, RPCMsg)
```
so it matches behavior of the `macro decodeFor` and metrics are updated regardless of protobuf decoding being succesful or not?

richard-ramos (approved):

libp2p/utils/protobuf.nim
vladopajic: done

richard-ramos (approved):

PR #2679, opened 2026-06-25
chore(cbind): [superseded by nim-ffi] pubsub TopicHandler backpressure

gmelodie:
## Status: superseded by the nim-ffi migration (#2681) — safe to close

This PR added a synchronous backpressure channel by changing the pubsub
`TopicHandler` C callback from `void` to returning an `int` status (0 = accepted,
nonzero = consumer full → drop + `trace` log).

That mechanism does not exist in the new cbindings. After #2681 (`migrate to
nim-ffi`), pubsub delivery is no longer a per-subscription C callback with a
return value — it is the `on_pubsub_message` **FFI event**:

- The C listener signature is `void (*)(const PubsubMessageEvent*, void*)` — it
  returns nothing, and nim-ffi's dispatch (`notifyListeners`) discards any return.
- Events are delivered asynchronously: the payload is enqueued on a bounded SPSC
  queue and drained on a separate event thread, so there is no synchronous point
  at which libp2p could observe "consumer full" and drop a specific message.

The intent of this PR — backpressure when a consumer can't keep up — is already
handled one layer down, and better than an advisory per-message drop:

- nim-ffi's event queue is bounded (`EventQueueCapacity = 1024`).
- On overflow it sets a sticky "stuck" flag and fires the `not_responding`
  event, giving the consumer a real congestion signal.

This is effectively the "libp2p-owned bounded queue + poll/drop policy" that this
PR's own *Additional Notes* flagged as the out-of-scope follow-up. It now lives in
nim-ffi, so the `int`-return change is obsolete.

Closing. If we want to *surface* the existing backpressure to C consumers (expose
the `not_responding` / queue-stuck state), that's a separate, design-first change.


gmelodie:
superseeded by #2681

PR #2681, opened 2026-06-25
feat(cbind)!: migrate to nim-ffi

gmelodie:
**Praise:** Thank you @Ivansete-status for all the good work on nim-ffi that allowed this refactor

## Summary

Migrates `cbind/` from its hand-rolled C-FFI runtime to the [`nim-ffi`](https://github.com/logos-messaging/nim-ffi) framework (package `ffi`), pinned by commit.

`cbind/` previously hand-rolled an entire FFI runtime: a worker thread, a request enum + dispatch, per-domain `createShared`/`process*` objects, C callback typedefs, shared-memory allocation, foreign-thread GC, and a hand-written `libp2p.h`. `nim-ffi` already provides all of that (`FFIContext[T]` threads, request channel, handle/event registries, CBOR codec) behind a DSL, and *generates* the C + CDDL bindings. So this PR is mostly **deletion**: the plumbing is replaced by annotations on a single, shorter `cbind/libp2p.nim`. The generated bindings are consumed by logos-co/logos-libp2p-module#77.

### What stays vs. what changes

- The libp2p-specific logic (switch builder, `mountGossipsub`/`mountKad`/`mountProtocols`, every request body) is **ported** into `{.ffi.}` proc bodies — only the plumbing is removed.
- The surface is uniformly `{.ffi.}` / `{.ffiCtor.}` / `{.ffiDtor.}` / `{.ffiEvent.}`; `genBindings()` emits the C header + CDDL schema. C export names are `camelToSnakeCase(procName)` (e.g. `libp2pGossipsubPublish` → export `libp2p_gossipsub_publish`), which consumers reach through the generated per-context wrappers `libp2p_ctx_<name>` (e.g. `libp2p_ctx_gossipsub_publish`).
- Requests/responses are CBOR `{.ffi.}` objects; `cstring`/`ptr+len` become `string`/`seq`. Stream handles become plain `uint64` ids tracked in `LibP2P` state (nim-ffi mints framework handles only on a `{.ffi.}` return path, and an incoming-stream protocol handler has no `ctx`).
- Protocol/pubsub handlers become events: `on_incoming_stream` and `on_pubsub_message`. The release-waiter handshake is preserved so multistream doesn't close incoming streams early. The old single `libp2p_set_event_callback` is superseded by the generated `libp2p_add_event_listener`.

### Deleted

`cbind/libp2p.h`, `ffi_types.nim`, `types.nim`, `alloc.nim`, and all of `cbind/libp2p_thread/**`. The root `cbind/.clang-format` is retained as `cbind/examples/.clang-format`, since the examples are still hand-written C.

### Examples

`cbind/examples/` ships two C programs, built and run against the generated `c_bindings/libp2p.h` and the `buildffi` shared library: `echo.c` (custom-protocol stream echo) and `gossipsub.c` (pub/sub over a shared topic). The `Build Examples` CI job builds and runs them.

### Hardening and cleanups

- Stream reads are bounded: `libp2p_stream_read_exactly` / `libp2p_stream_read_lp` validate the requested length against a 64 MiB `MaxReadBytes` cap and reject values that would truncate on 32-bit targets (e.g. the `linux-i386` job), closing an unbounded-preallocation DoS.
- Host-supplied config is validated instead of silently swallowed: an invalid `privKey` or `dnsResolver` now returns a precise `err(...)` (a bad private key previously fell through to a random identity), and `kad_get_value` rejects a nonsensical `quorum == 0`.
- The three loose stream fields on `LibP2P` (`streams`, `nextStreamId`, `streamReleaseWaiters`) are folded into a `StreamRegistry` with `register`/`get`/`release`, collapsing six copy-pasted lookup-and-`err` prologues into one call each.
- `gossipsub_publish` now returns a `PublishResponse { peerCount }` instead of discarding the publish fan-out count.
- `libp2pPeerinfo` → `libp2pPeerInfo` (C symbol `libp2p_peer_info`), aligning the export with the neighbouring `PeerInfoResponse` type.

### Consumer integration

Two follow-up fixes land the downstream build (logos-co/logos-libp2p-module#77):

- **Flat header install** (`nix/cbind.nix`): the generated headers (`libp2p.h` + its `nim_ffi_*.h` companions) install flat into `$out/include`, so a consumer that stages them with `cp $out/include/*.h` finds `libp2p.h` instead of a nested `c_bindings/` directory it would miss.
- **`libp2pDecodeXpr`**: adds the XPR decode proc (and `DecodeXprRequest`) mirroring `libp2pCreateXpr` — decodes a signed extended peer record, verifies the signature, and returns an `ExtendedPeerRecordEntry`. The module's `decodeXpr` binds to it.

### Dependency pinning

nim-ffi and the cbind-only deps (`taskpools`, `cbor_serialization`, `ffi`) are pinned **by commit** in `cbind/.pinned` and installed by a cbind-scoped `nimble install_pinned` task, deliberately kept out of the repo-root `.pinned` so the main libp2p CI (which doesn't use them, including its Nim 2.2.4 job) stays untouched. A branch name with a `/` is rejected by `nimble install`, so exact commits are required for reproducibility.

## Affected Areas

- [x] Build / Tooling
- [x] Other — C FFI bindings layer (`cbind/`) rewritten on top of nim-ffi; CBOR + event-based foreign ABI replaces the C-struct/callback ABI.


arnetheduck:
wait, why is this switching to C++?

.pinned
```diff
@@ -19,3 +19,6 @@ zlib;https://github.com/status-im/nim-zlib@#e680f269fb01af2c34a2ba879ff281795a52
 protobuf_serialization;https://github.com/status-im/nim-protobuf-serialization@#8406e7287196661614ce6a8e8be20f755376af7f
 nat_traversal;https://github.com/status-im/nim-nat-traversal@#860e18c37667b5dd005b94c63264560c35d88004
 npeg;https://github.com/zevv/npeg@#409f6796d0e880b3f0222c964d1da7de6e450811
+taskpools;https://github.com/status-im/nim-taskpools@#9e8ccc754631ac55ac2fd495e167e74e86293edb
+cbor_serialization;https://github.com/vacp2p/nim-cbor-serialization@#1664160e04d153573373afddc552b9cbf6fbe4dc
+ffi;https://github.com/logos-messaging/nim-ffi@#362ce635a0e79b70d50b1483429807b0417fb707
```
vladopajic: this better be in `/cbind/.pinned`

gmelodie:
> wait, why is this switching to C++?

~~@arnetheduck  nim-ffi currently does not generate C, only C++, and the only consumer of the cbindings is logos-libp2p-module. If there's a requirement of it being C from other clients, we can add C codegen in nim-ffi. This is just a nim-ffi feasibility test though, don't take this draft too seriously~~

migrating to C instead

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
-  var msgPtr: ptr cchar = nil
-  if msgLen > 0:
-    msgPtr = cast[ptr cchar](addr localMsg[0])
-  callback(RET_ERR.cint, nil, msgPtr, msgLen, userData)
-  return RET_ERR.cint
-
-template failIfStreamNil(
-    stream: ptr Libp2pStream, callback: Libp2pCallback, userData: pointer, msg: string
-) =
+## C FFI bindings for nim-libp2p, built on top of `nim-ffi`.
```
Ivansete-status, line 4: A bit aside from PR's purpose: I prefer `library` term instead of `cbind` for the folder name.

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
+##
+## The whole FFI runtime (worker thread, request channel, CBOR codec, event
```
Ivansete-status: I would not add to much ffi implementation details as they can vary eventually.

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
+## queue and the generated C/CDDL bindings) is provided by `nim-ffi`. This
+## file only declares the library state, the request/response shapes and the
+## libp2p-specific bodies; `genBindings()` at the bottom emits the foreign
+## bindings consumed by `logos-co/logos-libp2p-module`.
+
+import ffi
```
Ivansete-status, line 11: It is possible to split this module into several smaller meaningful modules and then, perform `include` for all the modules that annotate `.ffi.` procs.

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
+  ok(true)
+
+proc libp2pStreamRelease*(
+    lib: LibP2P, streamId: uint64
+): Future[Result[bool, string]] {.ffi.} =
+  discard ?lib.streams.get(streamId)
+  lib.streams.release(streamId)
+  ok(true)
+
+proc libp2pMountProtocol*(
```
Ivansete-status, line 628: Likely aside from PR's purpose but it seems there is too much logic done in ffi layer.
I think it is better to have a pure Nim API facade that contains all that logic and then make ffi a very thin layer on top of it. Maybe something to consider in upcoming PRs.

Ivansete-status (approved):

.github/workflows/cbindings.yml
```diff
@@ -1,4 +1,4 @@
-name: C bindings examples
```
vladopajic, line 1: isn't point of this workflow running `nimble examples` command?

cbind/examples/README.md
```diff
@@ -0,0 +1,41 @@
+# cbind examples
```
vladopajic, line 1: consider having `cbind/README.md` instead

this file can in general explain `cbind` and commands. since 
> From cbind, produce the runtime library and the generated header first

code here is already related parent folder.

then file can have section explain examples.

vladopajic (approved):

cbind/libp2p.nim
gmelodie: @richard-ramos ? i'm fine with whatever

cbind/.clang-format
richard-ramos, line 1: I'd keep this for the examples tho, since they're still C code

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
+      DefaultDnsServers
+    else:
+      @[initTAddress(config.dnsResolver)]
+
+  let rng = newRng()
+
+  var privKey = Opt.none(PrivateKey)
+  if config.privKey.len > 0:
+    PrivateKey.init(config.privKey).withValue(copyKey):
+      privKey = Opt.some(copyKey)
```
richard-ramos: Looks like we lost the code that failed if the private key passed to init was invalid

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
+
+type MuxerType {.pure.} = enum
+  MPLEX
+  YAMUX
+
+type BootstrapNode {.ffi.} = object
+  peerId: string
+  multiaddrs: seq[string]
+
+type Libp2pConfig {.ffi.} = object
```
richard-ramos, line 90: Previously we had a libp2p_new_default_config() that automatically set up the config with some useful defaults (like setting tcp and mplex as default transports). Was it intentionally removed?

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
-    help: string
-    labels: seq[LabelPair]
-    value: float64
-    timestamp: int64
+  let xpr = SignedExtendedPeerRecord.build(peerInfo.privateKey, peerRecord).valueOr:
+    return err(error)
+
+  ok(xpr.encode())
+
+proc libp2pDecodeXpr*(
```
richard-ramos, line 916: This shouldnt require the libp2p node to be executed.

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
+
+proc libp2pPeerstoreDeletePeer*(
+    lib: LibP2P, peerId: string
+): Future[Result[bool, string]] {.ffi.} =
+  let pid = PeerId.init(peerId).valueOr:
+    return err($error)
+  lib.switch.peerStore.del(pid)
+  ok(true)
+
+proc libp2pCreateCid*(
```
richard-ramos, line 1024: This shouldnt require a libp2p node to be executed.

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
+
+  let mh = MultiHash.digest(req.hash, req.data).valueOr:
+    return err("multihash error: " & $error)
+
+  let cid = Cid.init(cidVer, mc, mh).valueOr:
+    return err("cid init error: " & $error)
+
+  ok($cid)
+
+proc libp2pNewPrivateKey*(
```
richard-ramos, line 1046: This shouldnt require the libp2p node to be executed. (Especially since this could be used to obtain a private key that can be passed to the node config to setup the node)

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
+    return err(e.msg)
+  lib.gossipSub = Opt.some(gs)
+  ok()
+
+proc mountKad(lib: LibP2P, config: Libp2pConfig): Result[void, string] =
+  if not (config.mountKad or config.mountServiceDiscovery):
+    return ok()
+  let bootstrapNodes = parseBootstrapNodes(config).valueOr:
+    return err(error)
+  # Validator/selector are host callbacks that can't cross the FFI boundary, so use defaults.
```
richard-ramos, line 312: Why not? with previous ffi we could configure callbacks both for validators and selectors. Without this, the usage of kademlia is very limited since you're practically accepting any kind of value in the DHT


(and we will require similar behavior for gossipsub in the future to be able to set observers and validators)

cbind/libp2p.h
richard-ramos, line 1: This file previously contained doc comments on the functions that were exposed. How does nim-ffi solve this? does it add comments to the functions based on the nim comments?

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
+    return err("switch peerInfo is nil")
+
+  var addresses = parseMultiaddrs(req.addrs).valueOr:
+    return err(error)
+  if addresses.len == 0:
+    addresses = peerInfo.addrs
+
+  let seqNo =
+    if req.seqNo == 0:
+      Moment.now().epochSeconds.uint64
```
richard-ramos:
```suggestion
      getTime().toUnix().uint64
```
Or is there an speicfic reason why to use a monotonic clock here (i think i missed this on prev reviews)

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
+    return err("kad-dht not initialized")
+  let res = await kad.putValue(req.key, req.value)
+  if res.isErr():
+    return err(res.error)
+  ok(true)
+
+proc libp2pKadGetValue*(
+    lib: LibP2P, req: KadGetValueRequest
+): Future[Result[ReadResponse, string]] {.ffi.} =
+  let kad = lib.kad.valueOr:
```
richard-ramos, line 735: In prev bindings we had a `quorumOverride == 0` validation that went missing here. (quorum of 0 makes no sense).  This observation applies to putValue as well

cbind/libp2p.nim
```diff
@@ -1,1528 +1,1069 @@
...
+  try:
+    return createLibp2pNode(config)
+  except CatchableError as e:
+    return err("could not create libp2p node: " & e.msg)
 
-  let cidRes = Cid.init(cidVer, mc, mh)
-  let cid = cidRes.valueOr:
-    failWithMsg(callback, userData, "cid init error: " & $cidRes.error)
+proc libp2pDestroy*(lib: LibP2P) {.ffiDtor.} =
+  discard
```
richard-ramos, line 440: The previous destroy 'waited' for the thread to return before returning. Does this do the same or just gets rid of `lib`? because if so, maybe we should defensively run a libp2pStop here?

richard-ramos (commented):
This PR is extremely large thus I'm sure i missed stuff to review.
Would have been better to do the migration in separate PRs moving diff sections of the code at a time, even if it meant having to live with both existing bindings and new bindings for a while.

.github/workflows/cbindings.yml
```diff
@@ -54,13 +54,29 @@ jobs:
         run: |
           nimble install_pinned
 
-      - name: Build and run c bindings examples
+      - name: Restore build cache
```
vladopajic: building examples takes 5s, cache seems unnecessary. also i would avoid caching new stuff since main cache is close to limits. 

in general consider doing ci optimizations later. 

some notes: 
- `Build FFI library and generate C/CDDL bindings` step takes the most time consider improving this step
- maybe split into different sub steps 
- or cache final output (if there is some) form this step

vladopajic: cache `cbind` dependencies?

gmelodie:
closing in favor of 

1. #2772 — deps + pinning
2. #2773 — scaffold + CI
3. #2774 — connectivity
4. #2775 — streams (+ echo)
6. #2777 — kad + CID + crypto
7. #2778 — service discovery
9. #2780 — flip + delete legacy ← this PR

PR #2682, opened 2026-06-25
chore(protobuf): cosmetics

vladopajic:
- `serializerFor` is only needed for types that we explicitly ecnode/decode - for inner types this is not needed. therefore removing these 
  - if for some reason it is needed in future, it will be trivial to add case by case when needed 
- removed tests with `checkEncodeDecode` - since we are not doing it manually anymore, we should trust that protobuf-serialization is working correctly
   - tests that have custom logic (like hiding connection status) are kept


gmelodie (approved):

richard-ramos (approved):

nitely:
> removed tests with checkEncodeDecode - since we are not doing it manually anymore, we should trust that protobuf-serialization is working correctly 

beware, protobuf-ser cannot check you used the right `fieldNumber` (as per the libp2p spec) or the right field type, it also cannot check the type extensions are implemented correctly. Maybe some interop tests do catch these, but fwiw at least in #2477 there is proof the tests for dcutr were not enough as they passed in some commit where it shouldn't (discussion here: https://github.com/vacp2p/nim-libp2p/pull/2477#discussion_r3287834215). But also, the `checkEncodeDecode` tests won't catch these issues, because they are lacking the encoded data check; ex: they only check the value can be encoded/decoded but not the encoded value is correct.

PR #2696, opened 2026-06-26
fix(kad): cancel timed-out lookup RPCs so they don't leak rpc slots

radiken:
In a big bootstrap, a large fraction of nodes hang forever in `KadDHT.bootstrap()` and never finish. The iterative lookup (`lookOnce`, `kademlia/find.nim`) fires a batch of FIND_NODE RPCs and waits for them with `collectCompleted(timeout)` (`utils/future.nim`). On timeout, `collectCompleted` discards the futures that didn't answer but never cancels them. Each one is still parked in `readLp`, holding the `rpcSem` slot it acquired via `withRpcSlot`. `rpcSem` caps concurrent outbound RPCs at `maxConcurrentRpcs`, so every slow or unresponsive peer leaks one slot per lookup round. They pile up, and once they hold the whole pool the next FIND_NODE blocks on the semaphore and the lookup deadlocks.

**The fix:** cancel the still-pending futures in `collectCompleted` after the wait, so they release their stream and rpc slot instead of running on after their result has been abandoned.

### DIFF
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -21,6 +21,7 @@ type LookupState* = ref object
   shortlist*: Table[PeerId, XorDistance]
   responded*: Table[PeerId, RespondedStatus]
   attempts*: Table[PeerId, int]
+  inflight*: Table[PeerId, seq[FutureBase]]
 
 type DispatchProc* = proc(
   kad: KadDHT, peer: PeerId, target: Key
@@ -393,9 +394,10 @@ proc lookOnce*(
     let msg = await dispatch(kad, peerId, target)
     return (peerId, msg)
 
-  let
-    rpcBatch = toQuery.mapIt(dispatchWithPeer(it))
-    completedRPCBatch = await rpcBatch.collectCompleted(kad.config.timeout)
+  let rpcBatch = toQuery.mapIt(dispatchWithPeer(it))
+  for (fut, peerId) in zip(rpcBatch, toQuery):
+    state.inflight.mgetOrPut(peerId, @[]).add(FutureBase(fut))
+  let completedRPCBatch = await rpcBatch.collectCompleted(kad.config.timeout)
 
   for (fut, peerId) in zip(rpcBatch, toQuery):
     if not fut.finished():
@@ -420,6 +422,15 @@ proc lookOnce*(
     )
     await onReply(peerId, Opt.some(reply), state)
 
+  var doneFuts: seq[FutureBase]
+  for peerId in toQuery:
+    if state.responded.hasKey(peerId) or
+        state.attempts.getOrDefault(peerId, 0) > kad.config.retries:
+      doneFuts.add(state.inflight.getOrDefault(peerId).filterIt(not it.finished()))
+      state.inflight.del(peerId)
+  if doneFuts.len > 0:
+    await doneFuts.cancelAndWait()
+
   return true
 
 proc iterativeLookup*(
@@ -435,6 +446,12 @@ proc iterativeLookup*(
   while not stopCond(state):
     if not await kad.lookOnce(state, rtable, dispatch, onReply):
       break
+
+  var leftover: seq[FutureBase]
+  for futs in state.inflight.values:
+    leftover.add(futs.filterIt(not it.finished()))
+  if leftover.len > 0:
+    await leftover.cancelAndWait()
 
   state
 
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -400,7 +400,7 @@ proc lookOnce*(
   let completedRPCBatch = await rpcBatch.collectCompleted(kad.config.timeout)
 
   for (fut, peerId) in zip(rpcBatch, toQuery):
-    if not fut.finished():
+    if not fut.finished() or fut.cancelled():
       continue
     if fut.failed():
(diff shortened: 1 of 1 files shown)


libp2p/utils/future.nim
```diff
@@ -90,13 +90,19 @@ proc collectCompleted*[T, E](
...
+  ## slots instead of running on after their result has been abandoned.
   try:
     await futs.allFutures().wait(timeout)
   except AsyncTimeoutError:
     # Some futures didn’t finish in time, ignore
     discard
 
+  for fut in futs:
+    if not fut.finished():
+      await fut.cancelAndWait()
```
vladopajic:
```suggestion
  await futs.cancelAndWait()
```

vladopajic (approved):

richard-ramos:
Codex does suggest to also do the following:
```diff
diff --git a/libp2p/protocols/kademlia/find.nim b/libp2p/protocols/kademlia/find.nim
index <old>..<new> 100644
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -399,6 +399,8 @@ proc lookOnce*(
   for (fut, peerId) in zip(rpcBatch, toQuery):
     if not fut.finished():
       continue
+    if fut.cancelled():
+      continue
     if fut.failed():
       state.responded[peerId] = RespondedStatus.Failed
     else:
```

because in a later step the code does the following since a cancelled future is not `failed()`, thus the call to  `value()` in else branch can fail
```nim
    if fut.failed():
      state.responded[peerId] = RespondedStatus.Failed
    else:
      let (_, res) = fut.value()  <--------------------------
```

gmelodie (approved):

libp2p/utils/future.nim
radiken: thanks! updated

radiken:
I reworked this since it didn't pass the CI tests. The old version cancelled timed-out FIND_NODE futures every round, which also cancelled in-flight retries and broke the lookup retry path under load.

New approach: track in-flight futures per peer and cancel them only once a peer has responded or run out of retries, plus a sweep at lookup end. Slots still get recycled and the bootstrap deadlock is still fixed, but an active retry is never cancelled.

libp2p/protocols/kademlia/find.nim
```diff
@@ -420,6 +422,15 @@ proc lookOnce*(
...
     await onReply(peerId, Opt.some(reply), state)
 
+  var doneFuts: seq[FutureBase]
+  for peerId in toQuery:
+    if state.responded.hasKey(peerId) or
+        state.attempts.getOrDefault(peerId, 0) > kad.config.retries:
+      doneFuts.add(state.inflight.getOrDefault(peerId).filterIt(not it.finished()))
+      state.inflight.del(peerId)
+  if doneFuts.len > 0:
+    await doneFuts.cancelAndWait()
```
vladopajic: `if doneFuts.len > 0:` is not really necessary. `cancelAndWait` must work when `doneFuts` is empty (it should be noop). if it doesn't work like this then better fix `cancelAndWait` template.

only draw-down is that when `doneFuts.len == 0` then there is unnecessary await. this is smaller nitpick which we do not really pay attention to. and in cases like this we can assume that there will always be something to cancel.

libp2p/protocols/kademlia/find.nim
```diff
@@ -420,6 +422,15 @@ proc lookOnce*(
     )
     await onReply(peerId, Opt.some(reply), state)
 
+  var doneFuts: seq[FutureBase]
+  for peerId in toQuery:
+    if state.responded.hasKey(peerId) or
+        state.attempts.getOrDefault(peerId, 0) > kad.config.retries:
+      doneFuts.add(state.inflight.getOrDefault(peerId).filterIt(not it.finished()))
```
vladopajic: why `not it.finished()`? should this be `it.finished()` since `doneFuts`?

libp2p/protocols/kademlia/find.nim
```diff
@@ -21,6 +21,7 @@ type LookupState* = ref object
   shortlist*: Table[PeerId, XorDistance]
   responded*: Table[PeerId, RespondedStatus]
   attempts*: Table[PeerId, int]
+  inflight*: Table[PeerId, seq[FutureBase]]
```
vladopajic, line 24: `inflight` is shared variable for any `lookOnce` calls. theoretically there could be many `lookOnce` calls that are concurrently executing. in that case `state.inflight` will not be correct for both of those calls.

can `inflight` be variable local to `lookOnce` call?

gmelodie (approved):

libp2p/protocols/kademlia/find.nim
richard-ramos: I renamed `doneFuts` to `toCancel` to make the intention behind the code clearer, since what we want to cancel is the pending futures

libp2p/protocols/kademlia/find.nim
richard-ramos: hmmm... but `LookupState` is meant to be used sequentially. Concurrent `lookOnce` calls with the same state would cause races not only on `inflight`, but also on other variables `shortlist`, `attempts`, and `responded`. :thinking:

vladopajic: oh right, `LookupState` is created once for each request. this first comment is not on the point.

vladopajic (approved):

PR #2697, opened 2026-06-26
fix(ci): increase i386 ASAN timeout

gmelodie:
## Summary

The `linux-i386 (Nim v2.2.10 / orc)` lane of Continuous Integration fails on `master` and on every open PR, getting cancelled at the 55-minute mark (`The operation was canceled.`).

The lane was switched to AddressSanitizer in #2695. ASAN instrumentation combined with the Makefile's `--opt:speed` makes GCC's compilation of the two `test_all` slices roughly 3x slower: in the failing runs the test step spends ~50 minutes purely compiling before the first test binary even starts, leaving only a few minutes before the shared `timeout-minutes: 55` ceiling kills the job. The lane has never actually completed — it timed out even on #2695's own CI.

This change gives only that lane a larger budget (120 min) via a matrix-conditional `timeout-minutes`, while keeping the tight 55-min runaway ceiling on every other lane. The gate mirrors the ASAN gate in the run-tests step exactly (`os == 'linux' && cpu == 'i386' && memory_management == 'orc'`) so the two never drift.

## Affected Areas

- [x] Build / Tooling
  

N/A — CI configuration only; no library code changes.

## Impact on Library Users

No impact. CI-only change with no effect on the published library.

## Risk Assessment

Low for correctness: no build flags, source, or test selection change, so the test behavior that already passes is untouched. The expression only raises the kill ceiling for the ASAN lane; all other lanes keep the 55-min ceiling and already finish in 20-30 min.

Caveat worth verifying: `120` is an estimate (~50 min compile + a projected ASAN test run), since the lane has never run to completion. A single `workflow_dispatch` of this branch would confirm it finishes under 120 before this is marked ready.

Known trade-off: this only buys time, it does not address why ASAN compilation takes ~50 min. The cheaper structural fix (drop `--opt:speed` to ASAN's recommended `-O1`, or raise `TEST_ALL_SLICES` for this lane) changes test runtime/Makefile semantics and is left out of scope here intentionally.

## References

- #2695 — enabled AddressSanitizer on the i386 + orc lane (root cause)

### DIFF
--- a/tests/libp2p/pubsub/component/test_gossipsub_heartbeat.nim
+++ b/tests/libp2p/pubsub/component/test_gossipsub_heartbeat.nim
@@ -156,12 +156,12 @@ suite "GossipSub Component - Heartbeat":
   asyncTest "Fanout maintenance during heartbeat - expired peers are dropped":
     const
       numberOfNodes = 10
-      heartbeatInterval = 200.milliseconds
+      heartbeatInterval = 1000.milliseconds
     let
       nodes = generateNodes(
           numberOfNodes,
           gossip = true,
-          fanoutTTL = 60.milliseconds,
+          fanoutTTL = 200.milliseconds,
           heartbeatInterval = heartbeatInterval,
         )
         .toGossipSub()
@@ -195,7 +195,7 @@ suite "GossipSub Component - Heartbeat":
   asyncTest "Fanout maintenance during heartbeat - fanout peers are replenished":
     const
       numberOfNodes = 10
-      heartbeatInterval = 200.milliseconds
+      heartbeatInterval = 1000.milliseconds
     let
       nodes = generateNodes(
           numberOfNodes, gossip = true, heartbeatInterval = heartbeatInterval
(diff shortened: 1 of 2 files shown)


vladopajic (approved):

richard-ramos (approved):

tests/libp2p/pubsub/component/test_gossipsub_heartbeat.nim
```diff
@@ -156,12 +156,12 @@ suite "GossipSub Component - Heartbeat":
   asyncTest "Fanout maintenance during heartbeat - expired peers are dropped":
     const
       numberOfNodes = 10
-      heartbeatInterval = 200.milliseconds
+      heartbeatInterval = 1000.milliseconds
```
vladopajic, line 159: not sure about this part, this will change for all.

vladopajic (changes requested):

gmelodie:
this is not a proper workaround, job is taking 120mins+

PR #2701, opened 2026-06-26
ci: i386 temp workflow

vladopajic:
temporal workflow that catches nil pointer exceptions on i386 and test_gossipsub_message_handling tests

### DIFF
--- a/.github/workflows/i386_temp.yml
+++ b/.github/workflows/i386_temp.yml
@@ -72,29 +72,30 @@ jobs:
         run: |
           nim c \
             --styleCheck:usages --styleCheck:error \
-            --verbosity:0 \
+            --verbosity:2 \
             --skipUserCfg \
             -f \
             --threads:on \
             --opt:speed \
             --mm:orc \
             -d:useMalloc \
+            -d:chronicles_log_level=TRACE \
             --passC:-fsanitize=address \
             --passL:-fsanitize=address \
             -o:tests/libp2p/pubsub/component/test_gossipsub_message_handling \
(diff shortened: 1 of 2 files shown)


richard-ramos (approved):

gmelodie (approved):

.github/workflows/i386_temp.yml
```diff
@@ -0,0 +1,89 @@
...
+        run: make nimble.paths nat_libs
+
+      - name: Compile test_gossipsub_message_handling
+        id: compile-test
+        run: |
+          nim --version
+          nimble --version
+          gcc --version
+
+          nim c --styleCheck:usages --styleCheck:error --verbosity:0 --skipUserCfg --opt:speed --mm:orc \
```
richard-ramos:
```suggestion
          nim c --styleCheck:usages --styleCheck:error --verbosity:0 --skipUserCfg --opt:speed --mm:orc --threads:on  \
```
so it matches the same config as `make test`

.github/workflows/i386_temp.yml
```diff
@@ -0,0 +1,89 @@
...
+      matrix:
+        nim:
+          - ref: v2.2.10
+            memory_management: orc
+
+    defaults:
+      run:
+        shell: bash
+
+    name: "linux-i386 orc+asan test_gossipsub_message_handling"
```
richard-ramos: The flags for address sanitizer are missing (unless we're dropping tht from the CI, if so, then this job needs to be renamed)

vladopajic: i think we do not need it for catching nil pointer exception in test ?

PR #2707, opened 2026-06-30
ci: add line trace

vladopajic:
add line trace via `libbacktrace` to CI `test_all` workflow to get more detail information on strack trace 

**NOTE**
build times, even when cache, has gotten worse with this change. test execution time is unchanged. it behave is if there is not cache - judging by compile times.


arnetheduck:
alternative, if you want to retain baseline performance: https://github.com/status-im/nim-libbacktrace#system-stack-trace-override

vladopajic:
> alternative, if you want to retain baseline performance: https://github.com/status-im/nim-libbacktrace#system-stack-trace-override

thanks for suggestion 

it will really be useful to have some basic repo with CI implemented and configured with best options, that will server as an example for compiling nim tests and then running them. and maybe have this example repository in `nim-style-guide`

richard-ramos (approved):

gmelodie (approved):

PR #2709, opened 2026-06-30
ci: rework test workflow

vladopajic:
- test workflow has been split into to `test_all.yml` and `test_multiformat_exts.yml`
  - `test_all.yml` is test binary that runs `test_all.nim` tests for all platforms
  - `test_multiformat_exts.yml` is test binary that runs `test_multiformat_exts` tests for all platforms (but randomly choosing platform on each run). 
  - having separate workflows reduces total execution time of `test_all` by up to 1m (what it takes for  test_multiformat_exts to build and execute). 
  - cons is that `test_multiformat_exts` is not executed on all platforms always, but there is no need as these tests do not appear to be platform specific
- all test workflows have separate steps that build and run tests
- added build cache 

improves total ci execution time, now it takes no more then 13m per job (except i386 orc). which is overall improvement by ~50%


.github/workflows/test_all.yml
```diff
@@ -1,11 +1,10 @@
-name: Continuous Integration
+name: Test All
 
 on:
   push:
     branches:
       - master
   pull_request:
-  merge_group:
```
richard-ramos, line 8: It doesnt hurt to keep this `merge_group` for the moment we try enabling merge queues again (and to add it in test_multiformat_exts.yml too)

.github/workflows/test_multiformat_exts.yml
```diff
@@ -0,0 +1,170 @@
...
+    steps:
+      - name: Pick random platform
+        id: pick
+        run: |
+          declare -a OSS=("linux" "linux" "linux-gcc-14" "macos-15" "windows" "windows")
+          declare -a CPUS=("amd64" "i386" "amd64" "arm64" "amd64" "amd64")
+          declare -a CCS=("gcc" "gcc" "gcc" "clang" "clang" "gcc")
+          declare -a BUILDERS=("ubuntu-22.04" "ubuntu-22.04" "ubuntu-24.04" "macos-15" "windows-2022" "windows-2022")
+          declare -a SHELLS=("bash" "bash" "bash" "bash" "bash" "msys2 {0}")
+          PLATFORM_INDEX=$((RANDOM % 6))
```
richard-ramos, line 34: Why pick a random platform instead of running test on all?

Could it be possible to override this decision via a parameter to be set when running the workflow via the `Actions` tab of github, in case you want to run this on a specific platform?

.github/workflows/test_all.yml
```diff
@@ -160,30 +156,45 @@ jobs:
...
-          fi
+      - name: Build prerequisites
+        run: make nimble.paths nat_libs
 
-          export NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }}"
+      - name: Restore build cache
+        uses: actions/cache@v5
+        with:
+          path: nimcache
+          key: ${{ github.ref_name }}-nimcache-test_all-${{ matrix.platform.os }}-${{ matrix.platform.cpu }}-${{ matrix.platform.cc }}-${{ matrix.nim.ref }}-${{ matrix.nim.memory_management }}
```
richard-ramos, line 166: Codex said some interesting comment both for here and test_multiformat_exts.tml:

----

the workflows restore nimcache with branch-scoped keys, then call nim c directly without the Makefile’s -f rebuild flag. 

A cache hit on the same PR/master ref *can reuse stale generated/object files across commits*. 

On i386/orc, both slice builds also share --nimcache:nimcache/test_all, so sliceIdx=1 can inherit artifacts from sliceIdx=0. 

Use the Makefile targets, or add -f plus separate nimcache dirs per binary/slice and preferably make the primary cache key commit-scoped with broad restore keys.


If Nim decides an object in nimcache is still usable, CI can accidentally test stale compiled code instead of the current checkout. The i386/orc slice path makes this worse: both slice builds use the same --nimcache:nimcache/test_all:
```
nim c $NIMFLAGS -d:sliceIdx=0 -o:tests/test_all_0 ./tests/test_all
nim c $NIMFLAGS -d:sliceIdx=1 -o:tests/test_all_1 ./tests/test_all
```
Because NIMFLAGS already contains --nimcache:nimcache/test_all, slice 1 can reuse artifacts generated while compiling slice 0, even though sliceIdx changes which tests are imported. That risks producing an incorrect second binary.

.github/workflows/test_all.yml
gmelodie: ? 

If we want to enable queues in the future, then we readd the merge groups

.github/workflows/test_all.yml
```diff
@@ -96,7 +95,7 @@ jobs:
           shell: ${{ matrix.shell }}
           nim_ref: ${{ matrix.nim.ref }}
 
-      - name: Restore deps from cache
+      - name: Restore dependencies from cache
```
gmelodie, line 98: Image

gmelodie (approved):

.github/workflows/test_all.yml
richard-ramos: Just wondering why go with the effort of removing it in the first place :shrug:

gmelodie: I think it's fair to remove while we're at it since we're not using it

.github/workflows/test_multiformat_exts.yml
vladopajic: > having separate workflows reduces total execution time of test_all by up to 1m (what it takes for test_multiformat_exts to build and execute).
cons is that test_multiformat_exts is not executed on all platforms always, but there is no need as these tests do not appear to be platform specific

vladopajic: about override we can dalay implementing that to moment when it is needed. i dont see this code being change nor these test needing this.

.github/workflows/test_all.yml
vladopajic: this seems like halucination brcause test binaries will change in commits. if build cache has this issue here it will have even when we push new commit?

richard-ramos (approved):

.github/workflows/test_all.yml
vladopajic: this was new worflowfrođm the start https://github.com/vacp2p/nim-libp2p/commit/c43885c7d3ee7808d62b428c9d8913d3cf688ebc

but github sees it as edit to old one. 

i didn't really had in mind to change this it just happend so i left it like this

Issue #2713, opened 2026-06-30
transports: inconsistency when stopping pending accept


rlve:
### Description

When a transport is stopped while an `accept()` is blocked, TCP, WS raise a transport-stopped error while QUIC returns `nil`.

### Implementation

TCP raises when stopped while waiting:

https://github.com/vacp2p/nim-libp2p/blob/172f55c1380f93857a36991ef4addd3cb67e4578/libp2p/transports/tcptransport.nim#L263-L265

In QUIC, the post-wait guard is skipped and the await finished raises a QuicError that gets swallowed, so the method returns nil.

https://github.com/vacp2p/nim-libp2p/blob/172f55c1380f93857a36991ef4addd3cb67e4578/libp2p/transports/quictransport.nim#L519-L536

The check at the top of QUIC accept() only fires when accept() is called after stop().

https://github.com/vacp2p/nim-libp2p/blob/172f55c1380f93857a36991ef4addd3cb67e4578/libp2p/transports/quictransport.nim#L499-L501

### Related

- #2194
- #2714

PR #2714, opened 2026-06-30
test(transports): multiple addresses

### LINKED ISSUE #2194: test(transport): tcp,quic should allow multiple local addresses
we should extend `basic_test` with additional test case that asserts that transport can be started with multiple addresses. 
this should work for `tcp` and `quic` and for all other transports if it works.

do this after https://github.com/vacp2p/nim-libp2p/pull/2193 is merged

---
test cases:
- default test case: where multiple addresses are provided; dial from clients and ensure that dialed connection works
example: `test_quic.nim::should allow multiple local addresse`
- cancel dialing: same as above, but here we dial (with timeout so it gets cancelled) and do not drive accept so connection is not created (dial is cancelled). in this case we need to ensure that all already established connections will still work and that connections that we attempted to dial will not work (future will be canceled). 
- cancel accept: similar case to above, here we want to ensure that allready established connections will still work and that no new connections will be made because accept fut was cancelled. example: `basic_test.nim::handle accept cancellation`
- stopping transport while accepting: - should stop transport, close all connections; future that was blocked on `accept` should throw - transport stopped
- please consider and suggest other edge cases that can happen when stopping/canceling transport/futures

rlve:
## Summary

Addresses #2194

Adds transport tests covering cancellation and shutdown edge cases around `accept()` and `dial()`. 

Three things:
- A dual-stack test proving that cancelling an in-flight `dial()` on one address family leaves an already-established connection fully working.
- A dual-stack test proving that cancelling a parked `accept()` doesn't disturb an established connection and doesn't spawn a stray one.
- A basic-transport test proving that stopping a transport unblocks a pending `accept()`. QUIC resolves the parked accept to nil, the others raise `TransportClosedError`. Inconsistency raised in https://github.com/vacp2p/nim-libp2p/issues/2713

Supporting this, a small `runStreamHandler` helper in `utils.nim` opens a stream on an existing muxer and runs a handler on it, so tests can reuse an established connection to prove it still exchanges data after a cancellation.

## Affected Areas

- [x] Tests

### DIFF
--- a/tests/libp2p/transports/stream_tests.nim
+++ b/tests/libp2p/transports/stream_tests.nim
@@ -101,6 +101,107 @@ template streamTransportTest*(
       clientStreamHandler,
     )
 
+  asyncTest "dial cancellation leaves established connection working":
+    if addressIP6.isNone:
+      skip() # scoped to dual-stack transports (tcp, quic)
+      return
+
+    proc serverStreamHandler(stream: MuxedStream) {.async: (raises: []).} =
+      noExceptionWithStreamClose(stream):
+        var buffer: array[clientMessage.len, byte]
+        await stream.readExactly(addr buffer, clientMessage.len)
+        check string.fromBytes(buffer) == clientMessage
+        await stream.write(serverMessage)
+
+    proc clientStreamHandler(stream: MuxedStream) {.async: (raises: []).} =
+      noExceptionWithStreamClose(stream):
+        await stream.write(clientMessage)
+        var buffer: array[serverMessage.len, byte]
+        await stream.readExactly(addr buffer, serverMessage.len)
+        check string.fromBytes(buffer) == serverMessage
+
+    let server = transportProvider()
+    await server.start(@[addressIP4, addressIP6.get()])
+    defer:
+      await server.stop()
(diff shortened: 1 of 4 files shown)


gmelodie (approved):

tests/libp2p/transports/stream_tests.nim
```diff
@@ -101,6 +101,107 @@ template streamTransportTest*(
...
+        await stream.readExactly(addr buffer, clientMessage.len)
+        check string.fromBytes(buffer) == clientMessage
+        await stream.write(serverMessage)
+
+    proc clientStreamHandler(stream: MuxedStream) {.async: (raises: []).} =
+      noExceptionWithStreamClose(stream):
+        await stream.write(clientMessage)
+        var buffer: array[serverMessage.len, byte]
+        await stream.readExactly(addr buffer, serverMessage.len)
+        check string.fromBytes(buffer) == serverMessage
```
vladopajic: `check (await stream.readExactlyAsStr(serverMessage.len)) ==  serverMessage`

vladopajic (approved):

PR #2719, opened 2026-06-30
fix(autotls): decouple ACME RSA keys from libp2p identity schemes

richard-ramos:
## Summary

AutoTLS ACME now uses `RsaPrivateKey` directly for ACME account keys, certificate keys, CSR generation, and JWS signing instead of routing those RSA-only operations through the generic libp2p `KeyPair`.

Nimbus needs this because it builds nim-libp2p with `libp2p_pki_schemes=secp256k1`, which removes RSA from the generic libp2p identity key union. The previous AutoTLS ACME code still dereferenced RSA fields through `KeyPair`, causing Nimbus builds that import AutoTLS-related modules to fail at compile time.

## Affected Areas

- [x] Other  
  AutoTLS ACME key handling and related tests.

## Impact on Library Users

This changes lower-level AutoTLS ACME APIs from `KeyPair` parameters to `RsaPrivateKey` parameters. Callers using `ACMEApi` or `ACMEClient` directly should generate/pass RSA private keys directly.

Higher-level AutoTLS service usage continues to generate the required certificate key internally.

## Risk Assessment

Behavioral risk is low: ACME still uses RSA/RS256 as before, but the type boundary is now explicit. The main compatibility risk is source-level migration for callers of the lower-level ACME APIs.

## References

- Nimbus CI failure: https://github.com/status-im/nimbus-eth2/actions/runs/28470173703/job/84380389813

### DIFF
--- a/libp2p/autotls/acme/api.nim
+++ b/libp2p/autotls/acme/api.nim
@@ -7,8 +7,8 @@ import chronos/apps/http/httpclient, results, chronicles
 
 import ./jws
 import ./utils
-import ../../crypto/crypto
 import ../../crypto/rsa
+import ../../utils/opt
 
 export ACMEError, ACMENetworkError
 
@@ -108,7 +108,7 @@ type ACMECertificateResponse* = object
 type ACMECertificate* = object
   rawCertificate*: string
   certificateExpiry*: DateTime
-  certKeyPair*: KeyPair
+  certKeyPair*: RsaPrivateKey
 
 const
   Alg = "RS256"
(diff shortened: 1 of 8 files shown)


libp2p/autotls/acme/api.nim
```diff
@@ -108,7 +108,7 @@ type ACMECertificateResponse* = object
 type ACMECertificate* = object
   rawCertificate*: string
   certificateExpiry*: DateTime
-  certKeyPair*: KeyPair
+  certKeyPair*: RsaPrivateKey
```
vladopajic, line 111: cant this be just `PrivateKey`?

vladopajic (approved):

libp2p/autotls/acme/api.nim
richard-ramos: No, because Nimbus builds with `libp2p_pki_schemes=secp256k1`, so RSA attribute `rsakey`  is not available in `PrivateKey`. If we use it here, then the compile error from https://github.com/status-im/nimbus-eth2/actions/runs/28470173703/job/84380389813 happens.

Since ACME always needs an RSA key and this is only used internal API of autotls, using `RsaPrivateKey` directly makes the change very low impact, luckily

PR #2723, opened 2026-07-02
fix: avoid duplicate sending of control packet

etan-status:
## Summary

Send IHAVE only once per peer instead of spamming it with duplicates for each topic for which they did not already requested partial.

## Affected Areas

- [x] Gossipsub  
  

- [ ] Transports  
  

- [ ] Protocol Logic

- [ ] Build / Tooling

- [ ] Other

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -759,14 +759,16 @@ proc onHeartbeat(g: GossipSub) =
   let peers = g.getGossipPeers()
   for peer, control in peers:
     # only ihave from here
+    var didSend = false
     for ihave in control.ihave:
       libp2p_pubsub_broadcast_ihave.inc(labelValues = [g.topicLabel(ihave.topicID)])
 
-      if ihave.topicID.isSome and
+      if not didSend and ihave.topicID.isSome and
           not g.extensionsState.peerRequestsPartial(peer.peerId, ihave.topicID.get()):
         # send IHAVE only if peer has not requested partial for topic.
         # these peers will receive gossip of partial metadata via extension.
         g.send(peer, RPCMsg.withControl(control), MessagePriority.High)
+        didSend = true
 
   g.mcache.shift() # shift the cache
 


etan-status:
Also, from https://nim-lang.org/docs/destructors.html#sink-parameters

> A location that is passed to a `sink` parameter should not be used afterward.

`RPCMsg.withControl(control)` marks `control` with `sink`.

vladopajic (approved):

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -759,14 +759,16 @@ proc onHeartbeat(g: GossipSub) =
   let peers = g.getGossipPeers()
   for peer, control in peers:
     # only ihave from here
+    var didSend = false
```
vladopajic, line 762: this will result in node seeding only one `IHAVE` for peer per heartbeat?

instead what if node had `senIHaves: Table[PeerId, seq[TopicID]]` property so node does not send message if it already has sent this message to peer ?

vladopajic: ah wait

etan-status: the old logic sent duplicate RPCMsg.withControl(control) for each topic that doesn't pass peerRequestsPartial. with the full IHAVE list in each of the copies (or some undefined thing, depending on how RPCMsg.withControl(control) deals with the "sink" argument still being used afterward)

vladopajic:
```
control.ihave.filterIt(not g.extensionsState.peerRequestsPartial(peer.peerId, it.topicID.get()))
    
for ihave in control.ihave:
  libp2p_pubsub_broadcast_ihave.inc(labelValues = [g.topicLabel(ihave.topicID)])
   
g.send(peer, RPCMsg.withControl(control), MessagePriority.High)
```

vladopajic: ^ this code should be like this actually, `control.ihave` needs to filter out when peer requested partial for this topic

vladopajic: maybe change `getGossipPeers` so that this proc does not return topics if peer has requested partial and code here would be simple:

```
for peer, control in peers:
  g.send(peer, RPCMsg.withControl(control), MessagePriority.High)
  for ihave in control.ihave:
      libp2p_pubsub_broadcast_ihave.inc(labelValues = [g.topicLabel(ihave.topicID)])
```

note g.send is under first `for` and not second; second `for` is just for metrics. 

edit do `g.send` first and then update metrics to avoid issue if `g.send` is under second for.

etan-status: feel free to push to the PR :-) I'm not too familiar with the partial request feature.

vladopajic: okay, but the issue is that `g.send` is under second `for` that was only there to update metrics

vladopajic: https://github.com/vacp2p/nim-libp2p/pull/2724 @etan-status please review

etan-status: Looks good, regarding the "not sending duplicates" angle. Closing this PR here.

etan-status:
Superseded by #2724

PR #2724, opened 2026-07-02
fix(gossipsub): sending duplicate IHAVE

vladopajic:
the problem existed because:
`g.send(peer, RPCMsg.withControl(control), MessagePriority.High)` 
would be sent for every topic id in the control, this was wrong and issue was likely do to invalid indentation. `g.send` was under for that was updating metrics it should not be.

based on #2723

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -644,7 +644,7 @@ proc replenishFanout*(g: GossipSub, topic: string) =
 
   trace "fanout replenished with peers", peers = g.fanout.peers(topic)
 
-proc getGossipPeers*(g: GossipSub): Table[PubSubPeer, ControlMessage] =
+proc makeGossipControlMessages*(g: GossipSub): Table[PubSubPeer, ControlMessage] =
   ## gossip iHave messages to peers
   ##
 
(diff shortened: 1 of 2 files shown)


libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -759,15 +764,11 @@ proc onHeartbeat(g: GossipSub) =
...
+    g.send(peer, RPCMsg.withControl(control), MessagePriority.High)
+
     for ihave in control.ihave:
       libp2p_pubsub_broadcast_ihave.inc(labelValues = [g.topicLabel(ihave.topicID)])
 
-      if ihave.topicID.isSome and
-          not g.extensionsState.peerRequestsPartial(peer.peerId, ihave.topicID.get()):
-        # send IHAVE only if peer has not requested partial for topic.
-        # these peers will receive gossip of partial metadata via extension.
-        g.send(peer, RPCMsg.withControl(control), MessagePriority.High)
```
vladopajic, line 769: this was under `for ihave in control.ihave:` which is completely wrong

etan-status (commented):
Looks good to me, regarding the "not sending duplicates" angle.

richard-ramos (approved):

gmelodie (approved):

Issue #2731, opened 2026-07-02
ipaddr: ipAddrMatches only compares the ip4 component, so autonat v2 never skips amplification attack prevention for IPv6 clients


rlve:
### Description

The autonat v2 server should skip amplification attack prevention when the requested address matches the client's observed IP address. For IPv6 clients this never happens: `ipAddrMatches` only compares the ip4 component, so an IPv6 observed address never matches.

The behavior is pinned by the test "amplification attack prevention not skipped when observed IPv6 addr matches a requested addr" in `tests/libp2p/protocols/test_autonat_v2.nim`, alongside an IPv4 twin proving the exemption works when only the
address family differs.

### Spec Recommendation

The exemption is plain IP equality with no family qualifier: "If the selected address has the same IP address as the client's observed IP address, server proceeds to the next step skipping Amplification Attack Prevention steps"
([autonat v2 spec](https://github.com/libp2p/specs/blob/master/autonat/autonat-v2.md)).

### Implementation

`ipAddrMatches` selects the family with a bool that no caller ever sets:

https://github.com/vacp2p/nim-libp2p/blob/b3a0dc2818bac0ca767164a7cb1d9a3e901e6bf7/libp2p/utils/ipaddr.nim#L54-L72

The only call site, in the dial request handler:

https://github.com/vacp2p/nim-libp2p/blob/b3a0dc2818bac0ca767164a7cb1d9a3e901e6bf7/libp2p/protocols/connectivity/autonatv2/server.nim#L241-L253

### Related

- #2710

Issue #2710, opened 2026-06-30
ipaddr: isPrivate classifies every IPv6 address as public, so autonat canDial dials private IPv6 targets


rlve:
### Description

`isPrivate` in `utils/ipaddr.nim` matches IPv4 dotted-decimal string prefixes only, so `isPrivate` returns `false` for every IPv6 address, including ULA (`fc00::/7`), link-local (`fe80::/10`) and loopback (`::1`).

https://github.com/vacp2p/nim-libp2p/blob/de00b8292902ad1dc62aae496426bd83b41ecd7a/libp2p/utils/ipaddr.nim#L14-L30

The concrete consequence is in autonat v2. `canDial` filters private targets through this `isPrivate`:

https://github.com/vacp2p/nim-libp2p/blob/de00b8292902ad1dc62aae496426bd83b41ecd7a/libp2p/protocols/connectivity/autonatv2/server.nim#L155-L171

With `allowPrivateAddresses = false`, a private IPv6 target like a ULA `fd00::1` is not rejected, because `isPrivate("fd00::1")` is `false`. So the server will attempt to dial a private IPv6 address that it would refuse if it were IPv4.

### Note: there are two unrelated isPrivate paths

`wire.nim` has a separate `isPrivateMA` that converts to a chronos `TransportAddress` and uses the chronos `isPrivate` / `isShared` / `isLinkLocal` predicates, which are family-aware:

https://github.com/vacp2p/nim-libp2p/blob/de00b8292902ad1dc62aae496426bd83b41ecd7a/libp2p/wire.nim#L178-L185

`natservice` uses `isPrivateMA`, not the `ipaddr` `isPrivate`, so NAT port mapping and announce filtering are not affected by this bug:

https://github.com/vacp2p/nim-libp2p/blob/de00b8292902ad1dc62aae496426bd83b41ecd7a/libp2p/services/natservice.nim#L284-L288

https://github.com/vacp2p/nim-libp2p/blob/de00b8292902ad1dc62aae496426bd83b41ecd7a/libp2p/services/natservice.nim#L362-L364

PR #2733, opened 2026-07-02
test(autonat): dual stack 4

rlve:
## Summary

Adds AutoNAT v2 dual-stack test coverage.

New tests in `test_autonat_v2.nim`:

- an IPv4-only server refuses a DialRequest for an IPv6 addr with `EDialRefused`, and an IPv6-only server refuses an IPv4 addr the same way.
- a dual-stack server dials back a loopback IPv6 addr even with `allowPrivateAddresses = false`, because `isPrivate` classifies every IPv6 address as non-private. Pins the protocol-level consequence of #2710.
- amplification attack prevention is skipped when the observed IPv4 addr matches a requested addr, but never skipped for IPv6. Pins #2731.

Supporting changes: `setupAutonat` takes an optional `dstAddrs` so tests control the server listen families, and the `TCP_IP6` pattern is now exported from `multiaddress.nim` next to `TCP_IP4`. The identical local definitions in `test_identify.nim` and `tests/tools/test_multiaddress.nim` are replaced with the exported patterns.

## Affected Areas

- [x] Tests
  Test-only coverage for AutoNAT v2, plus one new exported multiaddress pattern (`TCP_IP6`).

## Impact on Library Users

`TCP_IP6` is a new exported const in `multiaddress.nim`.

### DIFF
--- a/tests/libp2p/protocols/test_autonat_v2.nim
+++ b/tests/libp2p/protocols/test_autonat_v2.nim
@@ -20,21 +20,13 @@ import
 import ../../tools/[unittest, crypto, switch_builder, multiaddress]
 
 proc setupAutonat(
-    srcAddrs: seq[MultiAddress] = newSeq[MultiAddress](),
-    dstAddrs: seq[MultiAddress] = newSeq[MultiAddress](),
+    srcAddrs: seq[MultiAddress] = @[TcpAutoAddress],
+    dstAddrs: seq[MultiAddress] = @[TcpAutoAddress],
     config: AutonatV2Config = AutonatV2Config.new(),
 ): Future[(Switch, Switch, AutonatV2Client)] {.async.} =
-  var srcBuilder = makeStandardSwitchBuilder(TcpAutoAddress)
-  if srcAddrs.len > 0:
-    srcBuilder = srcBuilder.withAddresses(srcAddrs)
-
-  var dstBuilder = makeStandardSwitchBuilder(TcpAutoAddress).withAutonatV2Server(config)
-  if dstAddrs.len > 0:
-    dstBuilder = dstBuilder.withAddresses(dstAddrs)
-
   let
-    src = srcBuilder.build()
-    dst = dstBuilder.build()
+    src = makeStandardSwitchBuilder(srcAddrs).build()
+    dst = makeStandardSwitchBuilder(dstAddrs).withAutonatV2Server(config).build()
     client = AutonatV2Client.new(rng())
 
   client.setup(src)
@@ -76,7 +68,7 @@ suite "AutonatV2":
     ) == Reachable
 
   asyncTest "asAutonatV2Response":
-    let addrs = @[MultiAddress.init("/ip4/127.0.0.1/tcp/4000").get()]
+    let addrs = @[ma("/ip4/127.0.0.1/tcp/4000")]
     let errorDialResp = DialResponse(
       status: ResponseStatus.Ok,
       addrIdx: Opt.none(AddrIdx),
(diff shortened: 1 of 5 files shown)


richard-ramos (approved):

tests/libp2p/protocols/test_autonat_v2.nim
```diff
@@ -202,6 +208,115 @@ suite "AutonatV2":
         addrs: Opt.some(reqAddrs[0]),
       )
 
+  asyncTest "DialRequest with IPv6 addr refused by IPv4-only server":
+    # server listens on IPv4 only, so it refuses to dial any IPv6 addr
+    let (src, dst, client) = await setupAutonat()
+    defer:
+      await allFutures(src.stop(), dst.stop())
+
+    let reqAddrs = @[MultiAddress.init("/ip6/2001:db8::1/tcp/4040").get()]
```
vladopajic:
```suggestion
    let reqAddrs = @[ma("/ip6/2001:db8::1/tcp/4040")]
```

`ma` utility template from test/tools/multiaddress

tests/libp2p/protocols/test_autonat_v2.nim
```diff
@@ -20,15 +21,20 @@ import ../../tools/[unittest, crypto, switch_builder, multiaddress]
...
+    dstAddrs: seq[MultiAddress] = newSeq[MultiAddress](),
     config: AutonatV2Config = AutonatV2Config.new(),
 ): Future[(Switch, Switch, AutonatV2Client)] {.async.} =
   var srcBuilder = makeStandardSwitchBuilder(TcpAutoAddress)
   if srcAddrs.len > 0:
     srcBuilder = srcBuilder.withAddresses(srcAddrs)
 
+  var dstBuilder = makeStandardSwitchBuilder(TcpAutoAddress).withAutonatV2Server(config)
+  if dstAddrs.len > 0:
+    dstBuilder = dstBuilder.withAddresses(dstAddrs)
```
vladopajic: ideally it would be better to change `makeStandardSwitchBuilder`, it currently accepts single `MultiAddress` but it can accept sequence as well. with that change then


`dstAddrs: seq[MultiAddress] = @[TcpAutoAddress]`  default to singe tpc address when not specified

 `var dstBuilder = makeStandardSwitchBuilder(dstAddrs).withAutonatV2Server(config)` just pass `dstAddrs` t

actually you will not need `dstBuilder` in this case

vladopajic (approved):

gmelodie (approved):

tests/libp2p/protocols/test_autonat_v2.nim
rlve: Addressed in https://github.com/vacp2p/nim-libp2p/pull/2733/changes/608eb56b2b47b09af2d399fb52c121c18cabe28e

vladopajic: :fire:

tests/tools/switch_builder.nim
```diff
@@ -8,28 +8,42 @@ import ./[crypto, multiaddress]
...
-  if QUIC_V1.match(address):
+  ## Transports are added automatically to match the listen `addresses`.
+
+  var b = SwitchBuilder.new().withRng(rng()).withNoise().withAddresses(addresses)
+
+  # the listen addresses decide which transports to use, deduped across addresses
+  var hasQuic, hasTcp, hasWs, hasMemory = false
+  for address in addresses:
+    if QUIC_V1.match(address):
+      hasQuic = true
```
vladopajic, line 21: nit: this code can add transport as well
```
  var addedQuic ....
  for ....
    if QUIC_V1.match(address) and not addedQuic:
      addedQuic = true
      b = b.withQuicTransport()
   ....
```

rlve: Not sure if it adds clarity though

PR #2734, opened 2026-07-02
fix(peerstore): keep connected peer addresses from expiring

### LINKED ISSUE #2684: peerstore: should an address expire while we're still connected to the peer?
The `AddressBook` expires entries on a TTL measured from each entry's `lastUpdated`, and the background prune loop deletes expired entries without ever checking connection state. So an address can be pruned while we hold an active connection to that peer.

`lastUpdated` is only refreshed when the entry is written: a dial (`markConnected`, High, 24h), identify on connect (`set`, Medium, 1h), or an identify-push from the remote. There's no periodic re-identify, and identify-push only fires when the remote's own advertised addresses change. So on a stable long-lived connection between two nodes with stable addresses, nothing refreshes the timestamp, and the entry eventually crosses its TTL and gets dropped.

The TTL check ignores connection state entirely:

https://github.com/vacp2p/nim-libp2p/blob/cd35cc04f7ddda58f39ec49ee84d30784501b6df/libp2p/peerstore.nim#L178-L194

https://github.com/vacp2p/nim-libp2p/blob/cd35cc04f7ddda58f39ec49ee84d30784501b6df/libp2p/peerstore.nim#L331-L346

A test documents the behavior (`connected-but-idle address expires while still connected`): the entry is pruned while `isConnected` is still true.

https://github.com/vacp2p/nim-libp2p/blob/cd35cc04f7ddda58f39ec49ee84d30784501b6df/tests/libp2p/test_peer_store_component.nim#L189-L221

### Possible consequences

The question is whether this matters in practice, and how much. An active connection is arguably the strongest evidence an address is valid, so pruning it feels off, but I'm not confident about the real-world impact. A few angles worth checking:

1. **Redial after disconnect.** While connected, losing the address is invisible because new streams reuse the existing muxer (`tryReusingConnection` runs before `addrs` is even read). The address only matters again once the connection drops and something wants to redial. nim-libp2p doesn't redial dropped connections on its own, so does any consumer actually rely on the peerstore still holding the address at that point, or do they keep their own list?

2. **Kademlia FIND_NODE responses.** Kademlia builds responses from `peerStore[AddressBook][peer]`. A connected-but-expired peer would yield an empty address list, so we'd serve records with no addresses for a peer we're connected to right now. Is that a real problem or negligible?

https://github.com/vacp2p/nim-libp2p/blob/cd35cc04f7ddda58f39ec49ee84d30784501b6df/libp2p/protocols/kademlia/find.nim#L508-L514

3. **`contains` check.** `AddressBook.contains` returns false once entries expire, so a connected peer reads as "address unknown". Does anything gate on that?

Frequency, for reference: Medium (1h) inbound addresses would hit this on any connection lasting over an hour, High (24h) dialed addresses on connections lasting over a day (validators, bootstrap, relays).

richard-ramos:
## Summary

Keep known peer addresses available while the peer is still connected.

Previously, AddressBook TTL pruning could expire a peer's addresses even while an active connection existed. This made connected peers appear addressless to peerstore readers and could leave no address available immediately after disconnect. This PR tracks connected peers in the peerstore, keeps their existing addresses live while connected, and refreshes stale entries before removing the connected state on disconnect.

## Affected Areas

## Impact on Library Users
Connected peers no longer lose their known addresses due to TTL expiry while the connection is active. After the last disconnect, those addresses return to normal TTL-based expiry behavior.

This adds peerstore lifecycle hooks used by the connection manager; no migration is required for existing callers.

## References

- Fixes #2684

### DIFF
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -380,6 +380,7 @@ proc onPeerDisconnected(c: ConnManager, peerId: PeerId) {.async: (raises: []).} 
   c.clearPeerReadyState(peerId)
   c.connectedAt.del(peerId)
   if not c.peerStore.isNil:
+    c.peerStore.markPeerDisconnected(peerId)
     c.peerStore.cleanup(peerId)
   libp2p_peers.set(c.muxerStore.countPeers.int64)
   await noCancel c.triggerPeerEvents(peerId, PeerEvent(kind: PeerEventKind.Left))
@@ -463,6 +464,9 @@ proc storeMuxer*(
     raise newException(LPError, "muxer already stored")
 
   libp2p_peers.set(c.muxerStore.countPeers().int64)
+
+  if isNewPeer and not c.peerStore.isNil:
+    c.peerStore.markPeerConnected(peerId)
 
   c.onCloseFuts.trackFut(c.onClose(muxer))
 
--- a/libp2p/peerstore.nim
+++ b/libp2p/peerstore.nim
@@ -90,6 +90,7 @@ type
 
   AddressBook* = ref object of PeerBook[seq[AddressEntry]]
     ttls*: AddressConfidenceTtls
+    connectedPeers: HashSet[PeerId]
 
   KeyBook* = ref object of PeerBook[PublicKey]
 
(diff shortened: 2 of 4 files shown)


vladopajic (approved):

rlve (commented):
👌

gmelodie (approved):

PR #2737, opened 2026-07-02
fix(connmanager): avoid late Joined after peer drop

richard-ramos:


vladopajic (approved):

tests/libp2p/test_conn_manager.nim
```diff
@@ -260,6 +260,62 @@ suite "Connection Manager":
...
+          events.add(kind)
+
+    connMngr.addConnEventHandler(connectedHandler, ConnEventKind.Connected)
+    connMngr.addConnEventHandler(disconnectedHandler, ConnEventKind.Disconnected)
+    connMngr.addPeerEventHandler(peerHandler("Joined"), PeerEventKind.Joined)
+    connMngr.addPeerEventHandler(peerHandler("Left"), PeerEventKind.Left)
+
+    let storeFut = connMngr.storeMuxer(muxer)
+
+    check await connectedStarted.wait().withTimeout(1.seconds)
```
vladopajic:
```suggestion
    check await connectedStarted.wait()
```

timeouts introduce flakiness, let's not have them unless really needed.

tests/libp2p/test_conn_manager.nim
```diff
@@ -260,6 +260,62 @@ suite "Connection Manager":
...
+    unblockConnected.fire()
+    await storeFut
+
+    checkUntilTimeout:
+      "Disconnected" in events
+
+    check:
+      "Connected" in events
+      "Left" in events
+      "Joined" notin events
```
vladopajic:
```suggestion
    checkUntilTimeout:
      events == @["Connected", "Left", "Disconnected"]
```
assert ordering as well ?

vladopajic (approved):

tests/libp2p/test_conn_manager.nim
vladopajic: if we ever switch to `unittest3`, test like this will be really flaky. because `unittest3` starts X tests and then waits on them. it can happen that test like this is started but other tests take cpu time first. and when this test gets to it's cpu time, timeout time might have elapsed. and test would fail but this would be completely valid case if cpu race happened in different order.

gmelodie (approved):

PR #2739, opened 2026-07-03
ci(daily): remove concurrent test job

vladopajic:
removing concurrent test job and split running test in two parts (compile & execute).

### DIFF
--- a/.github/workflows/daily_common.yml
+++ b/.github/workflows/daily_common.yml
@@ -165,29 +165,34 @@ jobs:
           echo "nimble install -y $DEPS"
           nimble install -y $DEPS
 
+      - name: Build test_all
+        run: |
+          NIMFLAGS="--nimcache:nimcache/test_all --mm:${{ matrix.nim.memory_management }} --opt:speed --styleCheck:usages --styleCheck:error --parallelBuild:0 --threads:on --skipUserCfg"
+          NIMFLAGS="$NIMFLAGS -d:chronicles_log_level=DEBUG"
+          NIMFLAGS="$NIMFLAGS --debugger:native --stacktrace:off --import:libbacktrace --define:nimStackTraceOverride"
+          if [[ '${{ matrix.platform.cc }}' == 'clang' ]]; then
+            NIMFLAGS="$NIMFLAGS --cc:clang"
+          fi
+          if [[ '${{ matrix.platform.cpu }}' == 'i386' && '${{ matrix.nim.memory_management }}' == 'orc' ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/daily_common.yml
```diff
@@ -165,29 +165,34 @@ jobs:
...
+          NIMFLAGS="$NIMFLAGS -d:chronicles_log_level=DEBUG"
+          NIMFLAGS="$NIMFLAGS --debugger:native --stacktrace:off --import:libbacktrace --define:nimStackTraceOverride"
+          if [[ '${{ matrix.platform.cc }}' == 'clang' ]]; then
+            NIMFLAGS="$NIMFLAGS --cc:clang"
           fi
-
-          export NIMFLAGS="${NIMFLAGS} --mm:${{ matrix.nim.memory_management }}"
-          if [[ '${{ matrix.config.cc }}' == 'clang' ]]; then
-            export NIMFLAGS="${NIMFLAGS} --cc:clang"
+          if [[ '${{ matrix.platform.cpu }}' == 'i386' && '${{ matrix.nim.memory_management }}' == 'orc' ]]; then
```
richard-ramos: Instead of `matrix.platform`, you should use `matrix.config` (defined in line 81)

.github/workflows/daily_common.yml
richard-ramos, line 145: Should we add 
```
      - name: Build prerequisites
        run: make nimble.paths nat_libs
```

like in test_all.yml?

vladopajic: yes

.github/workflows/daily_common.yml
vladopajic: yes, this was copy / paste error

richard-ramos (approved):

gmelodie (commented):
> note: devel errors fixed in https://github.com/vacp2p/nim-libp2p/pull/2740

Maybe point this pr to that one?

Having two if statements in different steps (this one, and the one below which runs slices 0 and 1) is error-prone, since one could forget to update one while updating the other. Question: why is not using slices necessary?

vladopajic:
> Maybe point this pr to that one?

#2740 is independent pr that fixes issue. #2740 was pointed here as hint that these tests are in fact passing with other fix.

> why is not using slices necessary?

because it is impossible to find issues in output like this https://github.com/vacp2p/nim-libp2p/actions/runs/28645967510/job/84952529946 for example


> Having two if statements in different steps (this one, and the one below which runs slices 0 and 1) is error-prone, since one could forget to update one while updating the other.
 
 wdym?

gmelodie (approved):

PR #2742, opened 2026-07-03
test: utilize new switch builder with multiple addresses

vladopajic:
utilizing new switch builder that accepts multiple addresses.

### DIFF
--- a/tests/tools/switch_builder.nim
+++ b/tests/tools/switch_builder.nim
@@ -31,11 +31,15 @@ proc makeStandardSwitchBuilder*(addresses: seq[MultiAddress]): SwitchBuilder =
   if hasQuic:
     b = b.withQuicTransport()
   if hasTcp:
-    b = b.withTcpTransport().withMplex()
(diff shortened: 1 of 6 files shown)


tests/tools/switch_builder.nim
```diff
@@ -31,11 +31,15 @@ proc makeStandardSwitchBuilder*(addresses: seq[MultiAddress]): SwitchBuilder =
...
   if hasWs:
-    b = b.withWsTransport().withMplex()
+    b = b.withWsTransport()
   if hasMemory:
-    b = b.withMemoryTransport().withMplex()
+    b = b.withMemoryTransport()
+
+  # mplex is default muxer
+  if hasTcp or hasWs or hasMemory:
+    b = b.withMplex()
```
vladopajic, line 43: has to be added once in case when there is multiple address of same type

richard-ramos (approved):

gmelodie (approved):

PR #2746, opened 2026-07-03
test(misc): dual stack 5

rlve:
## Summary

Adds test coverage for IPv4/IPv6 dual-stack behavior across three areas: service-discovery IP-similarity scoring, the wildcard address resolver, and the NAT service.

Two of the pinned behaviors got their own issues:

  - #2756
  - #2757

The rest is new tests: NAT service drops a v6 external from the IGD (maps the port, announces nothing), and the wildcard resolver's per-family expansion and passthrough of non-wildcard / non-IP addresses.

The only non-test change is making `expandWildcardAddresses` public in tests so it can be tested directly.

 ## Affected Areas

- [x] Tests

### DIFF
--- a/libp2p/services/wildcardresolverservice.nim
+++ b/libp2p/services/wildcardresolverservice.nim
@@ -69,7 +69,7 @@ proc isWildcardIp(ip: IpAddress): bool =
   of IpAddressFamily.IPv6:
     ip.address_v6 == AnyAddress6.address_v6
 
-proc expandWildcardAddresses*(
+proc expandWildcardAddresses(
     networkInterfaceProvider: NetworkInterfaceProvider, listenAddrs: seq[MultiAddress]
 ): seq[MultiAddress] =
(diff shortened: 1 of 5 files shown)


gmelodie (approved):

libp2p/services/wildcardresolverservice.nim
```diff
@@ -69,7 +69,7 @@ proc isWildcardIp(ip: IpAddress): bool =
   of IpAddressFamily.IPv6:
     ip.address_v6 == AnyAddress6.address_v6
 
-proc expandWildcardAddresses(
+proc expandWildcardAddresses*(
```
vladopajic: `when defined(libp2p_testing)`

fyi: https://github.com/vacp2p/nim-libp2p/pull/2652#discussion_r3441830699

vladopajic (approved):

PR #2752, opened 2026-07-06
ci: dropping daily pinned

vladopajic:
dropping daily pinned CI workflow as these tests are already running on each commit in master and user branches.

### DIFF
--- a/.github/workflows/daily_common.yml
+++ b/.github/workflows/daily_common.yml
@@ -4,11 +4,6 @@ name: Daily Common
 on:
   workflow_call:
     inputs:
-      pinned_deps:
-        description: "Should dependencies be installed from pinned file or use latest versions"
-        required: false
-        type: boolean
-        default: false
       nim:
         description: "Nim Configuration"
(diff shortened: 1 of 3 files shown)


.github/workflows/daily_common.yml
```diff
@@ -65,16 +58,11 @@ jobs:
...
-        if: ${{ inputs.pinned_deps }}
-        run: |
-          echo "Pinned dependencies requested; skipping newest dependency resolution."
-
   test:
     needs: [delete_cache, resolve_deps]
     env:
       NIMBLE_COMMIT: 9207e8b2bbdf66b5a4d1020214cff44d2d30df92 # v0.20.1
-    timeout-minutes: 55
+    timeout-minutes: 70
```
vladopajic, line 65: all latest deps use bit more time since everything is downloaded and compiled for scratch with no cache.

richard-ramos (approved):

gmelodie (approved):

Issue #666, opened 2021-12-01
Failed `accept` are not `re-spawned` or handled in any way.


dryajov:
Currently, the switch will spawn a transport's `accept` which will run unattended. If the accept fails for any reason there is no way of notifying the calling application, nor does the switch attempt to re-spawn it. It would be nice if we could have more granular control over this flow as there might be cases where a switch ends up without any active accepts and is effectively unreachable.

Menduist:
https://github.com/status-im/nim-libp2p/pull/658 for instance caused the WS accept loop to crash and thus ws was unreachable

There are multiple ways to look at it, but if you think of it like an `asyncSpawn` the loop should never fail, and if it does, it should be a defect basically

Now, is there some cases where the accept may legitimately fail and not be restarted, but the program continues? Probably, and we may need to find how to handle it. It kinda goes back to the runtime mounting/unmouting of transports you talked about

dryajov:
> There are multiple ways to look at it, but if you think of it like an asyncSpawn the loop should never fail, and if it does, it should be a defect basically

Yeah, I agree with this, what I'm saying is that we don't address this in any way. One way might be that we always try to re-spawn the accept after it exited, this sounds like an easy quick fix to mend things for now, but this does introduce it's own issues, for example if accept keeps failing what then, do we keep re-spawning indefinitely?

All in all, this sort of things are a bit hard to reason about in a "framework/library", how much do you expose to the user and how much do you try to handle internally? This are hard questions to the tune of the discussion in #662.

vladopajic:
@rlve, we can have test for this?

for example we can create some new transport, that will only live in tests, and this flawed implementation should always fail `accept`. then we can test different things when using this transport. 

this will serve us to understand what's currently happening in code, and what is currently _expected_ behavior. 

if code is currently doing retry until `accept` succeeds, we can have test that will use flawed implementation transport impl that on second attempt makes `accept` successful`.

anyhow, if this makes sense could we address this in future? in this case let's create new issue for tests part so we can track that effort separately, just copy paste text or whatever.

PR #2759, opened 2026-07-06
test(transports): accept fails

### LINKED ISSUE #2755: transports: no tests for what happens when a transport's `accept` fails
### Description

When a transport's `accept` raises, the switch's accept loop logs an error and exits, and it never restarts. The transport stays marked as `running` and its address stays advertised, so the switch keeps looking reachable while nothing is actually accepting connections.

We have no tests covering any of this, so it's easy to break again without noticing.

### What to do

Add a transport that only lives in tests and lets us control what `accept` does. With it we can pin down the current behavior and understand what's actually expected:

- `accept` always fails, so we can see what the switch does today when a transport can't accept.
- `accept` fails once then succeeds, to check whether the loop retries or gives up.

Right now the two failure modes behave differently, and the tests should make that explicit:

- `accept` returns `nil`: the loop treats it as non-fatal and retries.

https://github.com/vacp2p/nim-libp2p/blob/458a507d1c6b38e8ff6f2aa9b2a4c902c5d9d65a/libp2p/switch.nim#L289-L295

- `accept` raises: the loop logs and returns, and is never respawned.

https://github.com/vacp2p/nim-libp2p/blob/458a507d1c6b38e8ff6f2aa9b2a4c902c5d9d65a/libp2p/switch.nim#L306-L312

A test-only transport is easy to plug in via `withTransport`, same way `withMemoryTransport` does it, and subclassing `MemoryTransport` avoids touching a real socket.

Note: these tests describe how the code behaves now, not how it should behave. What *should* happen when accept fails (respawn, backoff, notify the app, treat as a defect) is still open in #666. The tests are a first step so we can reason about that with the current behavior nailed down.

### Related

Splitting the testing effort out of #666 to track it separately.
#666
#658

rlve:
## Summary

Addresses #2755

Adds test coverage for what happens when a transport's `accept` fails.

The switch runs one accept loop per transport. Today the two failure modes behave differently and neither was pinned by a test:
  - `accept` returns `nil`: treated as a transient miss, the loop retries.
  - `accept` raises: the loop logs "Exception in accept loop, exiting" and returns for good. The transport stays `running` with its address still advertised, so the node looks reachable while nothing is accepting.

This PR documents current behavior only. Any refactor is tracked in #666.

New `tests/libp2p/test_switch_accept.nim` covers:
  - accept raising exits the loop while the transport still reports running and advertised
  - accept returning nil is non-fatal and the loop keeps retrying
  - inbound dials are dropped once a transport's accept loop has died
  - a transport recovers and serves a real connection after transient nil failures
  - one transport's accept failure does not stop other transports from accepting

## Affected Areas

- [x] Tests

## Impact on Library Users

Two minor surface changes, both for tests only:
- `MemoryTransport.rng` is now an exported field.
- `Switch.acceptFuts` accessor added under `when defined(libp2p_testing)`

### DIFF
--- a/tests/libp2p/test_switch_accept.nim
+++ b/tests/libp2p/test_switch_accept.nim
@@ -12,7 +12,7 @@ import
 import ../stubs/transportstub
 import ../tools/[unittest, crypto, lifecycle, multiaddress, switch_builder]
 
-proc newStubServer(
+proc newStubAcceptSwitch(
     behavior: StubAcceptBehavior, nilCount = 0, withTcp = false, maxIn = 0
 ): (Switch, MemoryTransportStub) =
   var addrs = @[MemoryAutoAddress()]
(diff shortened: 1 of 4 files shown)


richard-ramos (approved):

tests/libp2p/test_switch_accept.nim
```diff
@@ -0,0 +1,114 @@
...
+  let switch = b.build()
+  (switch, MemoryTransportStub(switch.transports[0]))
+
+suite "Switch accept-loop failure handling":
+  teardown:
+    checkTrackers()
+
+  asyncTest "accept raising exits the loop while the transport still looks reachable":
+    # a single inbound slot lets us check the loop hands it back when accept fails
+    let (server, stub) = newStubAcceptSwitch(RaiseAlways, maxIn = 1)
```
vladopajic: `stub` -> `transport`

tests/libp2p/test_switch_accept.nim
```diff
@@ -0,0 +1,114 @@
...
+
+    check server.connManager.availableSlots(Direction.In) == 1
+
+  asyncTest "accept returning nil is non-fatal and the loop keeps retrying":
+    let (server, stub) = newStubAcceptSwitch(NilAlways)
+    startAndDeferStop(@[server])
+
+    # nil is treated as a transient miss, so the loop keeps calling accept
+    checkUntilTimeout:
+      stub.acceptCalls >= 5
```
vladopajic: and the loop keeps retrying for 5 attempts or indefinitely? if first case then this should be const, and that const should be used here instead of 5. if indefinitely, then just improve comment a bit
 
```
 # nil is treated as a transient miss, so the loop keeps calling accept indefinitely. 
```

vladopajic (approved):
lgtm minor comments

gmelodie (approved):

PR #2772, opened 2026-07-07
chore(cbind): nim-ffi migration [1/9] — deps + pinning scaffold

gmelodie:
**nim-ffi cbind migration — PR train**

1. #2772 — deps + pinning ← this PR
2. #2773 — scaffold + CI
3. #2774 — connectivity
4. #2775 — streams (+ echo)
5. #2776 — pubsub (+ gossipsub)
6. #2777 — kad + CID + crypto
7. #2778 — service discovery
8. #2779 — relay / peerstore / metrics
9. #2780 — flip + delete legacy

---
Part 1 of a 9-PR train that replaces the hand-rolled `cbind/` C bindings with the generated **nim-ffi** codegen, landing it domain by domain so each PR stays reviewable and CI-green (build the FFI lib + generate bindings + build/run examples).

Targets `master` (stacked).

**This PR adds**
- Adds `cbind/.pinned` pinning `nim-ffi` (SHA-pinned at `7e3fd96`), its transitive `cbor_serialization`, and `taskpools`.
- Adds the matching `cbor_serialization` and `ffi` entries to `nix/cbind-deps.nix` (`taskpools` was already present).
- Declares ffi in `cbind/cbind.nimble` via `requires \"ffi\"` — by name, like `taskpools`. The rev lives in `cbind/.pinned` and `nix/cbind-deps.nix`, not hardcoded in the `.nimble`, so there's one fewer place to keep in sync.
- Adds an `install_pinned` task to `cbind/cbind.nimble` and wires it into `cbindings.yml`.
- Keeps `nim-ffi` **out** of the repo-root `.pinned` on purpose: it requires Nim >= 2.2.6, and the root `install_pinned` runs on the Nim 2.2.4 CI matrix, which would break. cbind gets these deps via its own scoped lock.
- No behaviour change — the legacy `cbind/` still builds and ships unchanged.

PR #2773, opened 2026-07-07
chore(cbind): nim-ffi migration [2/9] — new-library scaffold + CI

gmelodie:

1. #2772 — deps + pinning
2. #2773 — scaffold + CI ← this PR
3. #2774 — connectivity
4. #2775 — streams (+ echo)
6. #2777 — kad + CID + crypto
7. #2778 — service discovery

---
Part 2 of a 9-PR train that replaces the hand-rolled `cbind/` C bindings with the generated **nim-ffi** codegen, landing it domain by domain so each PR stays reviewable. Targets `chore/cbind/ffi/01-deps-pinning` (stacked).

**This PR adds**
- `cbind/libp2p_ffi.nim`, the new nim-ffi library, developed **in parallel** to the legacy `cbind/libp2p.nim` so the old bindings keep shipping until the final flip. It carries the library state (`LibP2P`, `StreamRegistry`), `declareLibrary`, the `MaxReadBytes` read-ceiling reserved for the stream reads landing in a later PR, the node builder, the lifecycle entry points, and `genBindings()`.
- The `{.ffi.}` config types (`Libp2pConfig`, `BootstrapNode`, the transport/muxer enums) and their parsing live in `cbind/libp2p_ffi/config.nim`. `config.parse()` validates the raw config once and returns a `ParsedConfig` of ready-to-use Nim values, so the node builder never touches a raw string or ordinal. The file is `include`d rather than imported: nim-ffi reads `{.ffi.}` object fields from the AST, and exporting the fields would leak their `*` into the generated field names.
- Protocol mounting orchestrated by `mountProtocols`, with Kademlia and service discovery mounting through separate helpers instead of one function branching on config flags.
- `buildffi`/`genbindings_c`/`genbindings_cddl` nimble tasks, a temporary `.#cbind-ffi` nix target (which reuses the repo Makefile's `nat_libs` target), and a temporary `c-bindings-ffi` CI job — all pointed at the new file. The old `.#cbind` target and job are untouched.

**Lifecycle & shutdown**
- `libp2pNew` / `libp2pStart` / `libp2pStop` / `libp2pDestroy`. `libp2pDestroy` **owns graceful shutdown**: it stops the switch on the FFI worker loop before the threads are joined and the context is freed, so destroy alone is sufficient. `libp2pStop` is a thin, idempotent wrapper over the same `shutdownSwitch` path, kept for an explicit, error-reporting stop. A `stopped` flag makes stop-then-destroy a no-op.
- This relies on nim-ffi's async teardown hook for `{.ffiDtor.}`; the `ffi` dep is re-pinned to logos-messaging/nim-ffi#115 (`d1575f0`). The FFI build compiles with `-d:ffiThreadExitTimeoutMs=5000` to give a graceful `switch.stop()` room beyond the 1500 ms default before the context is leaked.
- `libp2p_new_default_config()` is preserved for hosts that still call it, as a hand-maintained C-ABI mirror struct (the `{.ffi.}` config crosses as CBOR, not a C struct, so it isn't in the generated bindings).

**CI status**
- The `c-bindings-ffi` job runs on macOS (`macos-15` / arm64 / clang) but is **currently disabled** (`if: false`): `libp2p_ffi.nim` is built up domain by domain across PRs 2–9 and only compiles end-to-end at the flip, so a green run isn't meaningful until then. It's re-enabled (drop `if: false`) in the final PR. The legacy `c-bindings` job is unchanged and stays green throughout.

**Review artifact:** the generated `libp2p.h` appears for the first time here, with just the lifecycle entry points.

### DIFF
--- a/.github/workflows/cbindings.yml
+++ b/.github/workflows/cbindings.yml
@@ -69,7 +69,13 @@ jobs:
   # Temporary parallel job: builds the new nim-ffi library (cbind/libp2p_ffi.nim)
   # and generates its C/CDDL bindings. Folded into the main flow at the flip PR,
   # when libp2p_ffi.nim replaces the legacy libp2p.nim.
+  #
+  # Disabled until the final PR of the nim-ffi migration series: libp2p_ffi.nim is
+  # built up incrementally across PRs 2..9 and only compiles end-to-end at the
+  # flip. Re-enable (drop `if: false`) once the library is complete and expected
+  # to pass.
   c-bindings-ffi:
+    if: false
     timeout-minutes: 30
     strategy:
       fail-fast: false
@@ -81,7 +87,7 @@ jobs:
         shell: bash
 
     name: "Build FFI library (nim-ffi)"
-    runs-on: ubuntu-22.04
+    runs-on: macos-15 # fastest to compile and run; see daily_tests_no_flags.yml
     steps:
       - name: Checkout
         uses: actions/checkout@v6
@@ -92,8 +98,8 @@ jobs:
         uses: "./.github/actions/install_nim"
         with:
           shell: bash
-          os: linux
-          cpu: amd64
+          os: macos-15
+          cpu: arm64
           nim_ref: v2.2.10
 
       - name: Restore deps from cache
@@ -101,7 +107,7 @@ jobs:
         uses: actions/cache@v5
         with:
           path: nimbledeps
-          key: nimbledeps-linux-amd64-gcc-${{ hashFiles('.pinned', 'cbind/.pinned') }}
+          key: nimbledeps-macos-15-arm64-clang-${{ hashFiles('.pinned', 'cbind/.pinned') }}
 
       - name: Install deps
         if: ${{ steps.deps-cache.outputs.cache-hit != 'true' }}
@@ -112,12 +118,13 @@ jobs:
         run: |
           nim --version
           nimble --version
-          gcc --version
+          clang --version
 
           nimble setup
           cd cbind
           nimble install_pinned
           nimble setup
+          make -C .. nat_libs
           nimble buildffi
           nimble genbindings_c
           nimble genbindings_cddl
--- a/.github/workflows/cbindings.yml
+++ b/.github/workflows/cbindings.yml
@@ -101,7 +101,7 @@ jobs:
         uses: actions/cache@v5
         with:
           path: nimbledeps
-          key: nimbledeps-linux-amd64-gcc-${{ hashFiles('.pinned') }}
+          key: nimbledeps-linux-amd64-gcc-${{ hashFiles('.pinned', 'cbind/.pinned') }}
 
       - name: Install deps
         if: ${{ steps.deps-cache.outputs.cache-hit != 'true' }}
--- a/.github/workflows/cbindings.yml
+++ b/.github/workflows/cbindings.yml
@@ -65,3 +65,59 @@ jobs:
           nimble install_pinned
           nimble setup
           nimble examples
+
+  # Temporary parallel job: builds the new nim-ffi library (cbind/libp2p_ffi.nim)
+  # and generates its C/CDDL bindings. Folded into the main flow at the flip PR,
+  # when libp2p_ffi.nim replaces the legacy libp2p.nim.
+  c-bindings-ffi:
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+    env:
+      NIMBLE_COMMIT: 9207e8b2bbdf66b5a4d1020214cff44d2d30df92 # v0.20.1
+
+    defaults:
+      run:
+        shell: bash
+
+    name: "Build FFI library (nim-ffi)"
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          submodules: true
+
+      - name: Setup Nim
+        uses: "./.github/actions/install_nim"
+        with:
+          shell: bash
+          os: linux
+          cpu: amd64
+          nim_ref: v2.2.10
+
+      - name: Restore deps from cache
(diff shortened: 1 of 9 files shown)


vladopajic (approved):

.github/workflows/cbindings.yml
```diff
@@ -65,3 +65,59 @@ jobs:
...
+      fail-fast: false
+    env:
+      NIMBLE_COMMIT: 9207e8b2bbdf66b5a4d1020214cff44d2d30df92 # v0.20.1
+
+    defaults:
+      run:
+        shell: bash
+
+    name: "Build FFI library (nim-ffi)"
+    runs-on: ubuntu-22.04
```
vladopajic: `macos-15-arm64-clang` is the fastest ci usally finishes compile first, and test are usually faster to execute. maybe use it here, if it will work out of the box?  (note other changes that will have to be done, cache entry...)

cbind/libp2p_ffi.nim
```diff
@@ -0,0 +1,218 @@
...
+  ok(lib)
+
+proc libp2pNew*(config: Libp2pConfig): Future[Result[LibP2P, string]] {.ffiCtor.} =
+  try:
+    return createLibp2pNode(config)
+  except CatchableError as e:
+    return err("could not create libp2p node: " & e.msg)
+
+proc libp2pDestroy*(lib: LibP2P) {.ffiDtor.} =
+  discard
```
richard-ramos: The previous destroy 'waited' for the thread to return before returning. Does this do the same or just gets rid of lib? because if so, maybe we should defensively run a libp2pStop here?

.github/workflows/cbindings.yml
```diff
@@ -65,3 +65,59 @@ jobs:
...
+        run: |
+          nim --version
+          nimble --version
+          gcc --version
+
+          nimble setup
+          cd cbind
+          nimble install_pinned
+          nimble setup
+          nimble buildffi
```
richard-ramos, line 128:
```suggestion
          make -C .. nat_libs
          nimble buildffi
```

shouldnt we build nat_libs first?

cbind/libp2p_ffi.nim
```diff
@@ -0,0 +1,218 @@
...
+  if cfg.autonatV2:
+    switchBuilder = switchBuilder.withNAT(autonatConfig(AutonatV2))
+
+  if cfg.autonatV2Server:
+    switchBuilder = switchBuilder.withAutonatV2Server()
+
+  let switch =
+    try:
+      switchBuilder.build()
+    except CatchableError as e:
```
richard-ramos: Lets use a more specific exception

cbind/libp2p_ffi.nim
```diff
@@ -0,0 +1,218 @@
...
+  let lib = LibP2P(switch: switch, rng: rng, relayClient: relayClientOpt)
+
+  ?mountProtocols(lib, cfg)
+
+  ok(lib)
+
+proc libp2pNew*(config: Libp2pConfig): Future[Result[LibP2P, string]] {.ffiCtor.} =
+  try:
+    return createLibp2pNode(config)
+  except CatchableError as e:
```
richard-ramos: Lets use a more specific exception

cbind/libp2p_ffi.nim
```diff
@@ -0,0 +1,218 @@
...
+
+proc libp2pDestroy*(lib: LibP2P) {.ffiDtor.} =
+  discard
+
+proc libp2pStart*(lib: LibP2P): Future[Result[bool, string]] {.ffi.} =
+  try:
+    await lib.switch.start()
+  except LPError as e:
+    return err(e.msg)
+  ok(true)
```
richard-ramos: Since this function will only return ok(true) and never ok(false), the return could be changed to `Future[Result[void, string]]`, right?  (same on `libp2pStop`)

cbind/libp2p_ffi/config.nim
```diff
@@ -0,0 +1,181 @@
...
+
+type MuxerType {.pure.} = enum
+  MPLEX
+  YAMUX
+
+type BootstrapNode {.ffi.} = object
+  peerId: string
+  multiaddrs: seq[string]
+
+type Libp2pConfig {.ffi.} = object
```
richard-ramos, line 31: Previously we had libp2p_new_default_config(). Should we keep it? or not worth the effort?

gmelodie: Back!

cbind/libp2p_ffi.nim
gmelodie: Good catch! Implemented a dtor hook that we can use here to make destroy call stop (needs this because destroy and stop are on different threads). Now they both call shutdown.

cbind/libp2p_ffi.nim
```diff
@@ -0,0 +1,262 @@
...
+proc shutdownSwitch(lib: LibP2P) {.async.} =
+  ## Single source of truth for graceful shutdown. Idempotent: safe to call from
+  ## both an explicit `libp2pStop` and `libp2pDestroy`'s teardown.
+  if lib.stopped:
+    return
+  lib.stopped = true
+  await lib.switch.stop()
+
+proc libp2pStart*(lib: LibP2P): Future[Result[void, string]] {.ffi.} =
+  try:
```
richard-ramos, line 216:
```suggestion
  if not lib.stopped:
    return
  try:
```

cbind/libp2p_ffi.nim
```diff
@@ -0,0 +1,262 @@
...
+    await lib.switch.start()
+  except LPError as e:
+    return err(e.msg)
+  ok()
```
richard-ramos, line 221:
```suggestion
  lib.stopped = false
  ok()
```

richard-ramos (approved):

PR #2774, opened 2026-07-07
chore(cbind): nim-ffi migration [3/9] — connectivity & identity

gmelodie:

1. #2772 — deps + pinning
2. #2773 — scaffold + CI
3. #2774 — connectivity ← this PR
4. #2775 — streams (+ echo)
6. #2777 — kad + CID + crypto
7. #2778 — service discovery

---
Part 3 of a 9-PR train that replaces the hand-rolled `cbind/` C bindings with the generated **nim-ffi** codegen, landing it domain by domain so each PR stays reviewable and CI-green (build the FFI lib + generate bindings + build/run examples).

Targets `chore/cbind/ffi/02-scaffold` (stacked).

**This PR adds**
- `libp2pPublicKey` — the node's own public key, serialized in its scheme-native form so any key type round-trips (the default builder uses Ed25519), not just secp256k1.
- `libp2pConnect` / `libp2pDisconnect` — connect to / drop a peer, with a caller-supplied dial timeout (`timeoutMs`; `<= 0` means no bound).
- `libp2pPeerInfo` / `libp2pConnectedPeers` — own peer id + addrs; connected peers by direction.
- `libp2pDial` / `libp2pDialCircuitRelay` — open a stream to a peer (optionally via a relay), returning a `streamId` tracked in the `StreamRegistry`. Both take the same per-call `timeoutMs`.
- Their `{.ffi.}` request/response types and the shared `parseMultiaddrs` helper.

**Timeouts.** Every dial entry point exposes a per-call `timeoutMs` in its request rather than hardcoding one, so the caller owns the bound. The `{.ffi: "timeout = 30000".}` on those procs is a separate concern: it's the FFI runtime's handler backstop (default 5s) raised to 30s to cover libp2p's UpgradeTimeout, not the dial timeout.

**Also folded in.** Fresh nodes are now constructed with `stopped = true`, so `libp2pStart`'s idempotency guard (`if not lib.stopped: return`) no longer early-returns before ever starting the switch.

**Review artifact:** Header diff shows only these connectivity entry points. The `streamId`s handed out here become useful once stream ops land in [4/9].


vladopajic (approved):

cbind/libp2p_ffi.nim
```diff
@@ -259,4 +284,106 @@ proc libp2pNewDefaultConfig(): CLibp2pConfig {.
...
+    of PKScheme.Secp256k1:
+      peerInfo.publicKey.skkey
+    else:
+      return err("peerInfo public key must be secp256k1")
+
+  ok(@(pubKey.getBytes()))
+
+proc libp2pConnect*(
+    lib: LibP2P, req: ConnectRequest
+): Future[Result[void, string]] {.ffi: "timeout = 30000".} =
```
richard-ramos, line 313: What is this `timeout = 30000` for? is it for the max duration libp2pConnect can last?  In `ConnectRequest` there's a `timeoutMs` that sets the timeout for dialing (with timeoutMs <= 0 allowing `InfiniteDuration`. The advantage of it is that we let the user of the library determine the timeout duration instead of having it enforced for them.

Also, in libp2pDial, libp2pDialCircuitRelay there's a hardcoded timeout as well. Shouldnt those be part of the request instead?

cbind/libp2p_ffi.nim
```diff
@@ -259,4 +284,106 @@ proc libp2pNewDefaultConfig(): CLibp2pConfig {.
...
+  let peerInfo = lib.switch.peerInfo
+  if peerInfo.isNil():
+    return err("switch peerInfo is nil")
+
+  let pubKey =
+    case peerInfo.publicKey.scheme
+    of PKScheme.Secp256k1:
+      peerInfo.publicKey.skkey
+    else:
+      return err("peerInfo public key must be secp256k1")
```
richard-ramos: I know this wasnt introduced on this PR, but wondering if we should support other types of keys? since the default we use in the switch builder is Ed25519 anyway 🤔

cbind/libp2p_ffi.nim
gmelodie: This timeout is a nim-ffi timeout for the request (i.e. the function we're calling). The only real restriction on this is that it has to be greater than the time we need to do whatever it is that the function does (hence the 30s, which should be more than enough for whatever, i *think*?) and be lower than `infinite`.

gmelodie: > Also, in libp2pDial, libp2pDialCircuitRelay there's a hardcoded timeout as well. Shouldnt those be part of the request instead?

Yes! Added, but both `req.timeoutMs` and the ffi `timeout` should exist

richard-ramos: > it has to be greater than the time we need to do whatever it is that the function does (hence the 30s, which should be more than enough for whatever, i think?) and be lower than infinite.

if this is the case, should there be a validation to make sure that  `0 < timeoutMs <= timeout`,  because the logic as it is right now allows you to set an infinite duration when `timeoutMs <= 0`, also there's nothing stopping you from putting more than 30s as `timeoutMs`.

Perhaps the 30000 could be a constant too instead of having the timeout = 30000 everywhere? (dunno if possible)

but having said this, one thing i'm less sure about is if we should be setting up this timeout instead of having the users do it. Also, to take into account: the 30s would apply the whole operation... we should likely document this, as you can use multiple addresses here, and nim-libp2p will dial them sequentially, so these 30s are not used only for a single connection attempt but potentially for multiple addresses.

....
...

this makes me think about another point: `libp2p.h` contains doc comments on the API (i admit that we got sloppy with latest additions, so they're missing docs), but with nim-ffi, these comments are not present at all. Does nim-ffi support adding comments to the exported API?

gmelodie: > Perhaps the 30000 could be a constant too instead of having the timeout = 30000 everywhere? (dunno if possible)

Yes I thought about that. This is doable, but because of macro shenanigans we need to tweak something in nim-ffi to make this work (i can, if you want). This could even be a compile flag.

> but having said this, one thing i'm less sure about is if we should be setting up this timeout instead of having the users do it. Also, to take into account: the 30s would apply the whole operation... we should likely document this, as you can use multiple addresses here, and nim-libp2p will dial them sequentially, so these 30s are not used only for a single connection attempt but potentially for multiple addresses.

Hum.... This is a good point indeed. This may need more input/discussions though. If there's an issue with the timeouts, this would only touch ffi users (which, for now, is not a lot of people). Do you think we should wait/discuss, patch something like the compile time flag, or leave as-is opening an issue for discussion and patch in the future with a more proper fix? 

> Does nim-ffi support adding comments to the exported API?

No! But totally could. In many aspects nim-libp2p is being a testbed for nim-ffi's devx, so if you think this is desirable, we can make it happen (that's actually the point of this exercise).

richard-ramos: > Hum.... This is a good point indeed. This may need more input/discussions though. If there's an issue with the timeouts, this would only touch ffi users (which, for now, is not a lot of people). Do you think we should wait/discuss, patch something like the compile time flag, or leave as-is opening an issue for discussion and patch in the future with a more proper fix?

Can be done as a follow up, after discussions on what to do.

> No! But totally could. In many aspects nim-libp2p is being a testbed for nim-ffi's devx, so if you think this is desirable, we can make it happen (that's actually the point of this exercise).

It would be a nice to have!

richard-ramos (approved):

Issue #120, opened 2020-03-27
Less confusing init patterns


sinkingsugar:
Just an idea to make the api more intuitive about resource management.

### Init patterns

While trying to figure out why many tests were failing in the `test-stream` branch, I noticed that the resource creation and destruction API is counter-intuitive.
Most of the debugged issues were just lingering resources in the tests.

We are networking library so it's very common that *releasing* resources is spelled with `close()` and actually this is the case.
Yet most of those resources are created with a `newXX` pattern while the most natural would be `openXX`, so when we write code we can mentally automatically think and write stuff like, I know this pattern often will not be of use unless its a test but it still matters
```nim
let mplexDial = openMplex(conn)
# e.g. in the case of a test
defer: await mplexDial.close()
```
Without the need of thinking anything and without the need of any linter (very broken in nim) because we know already that a `open` has a `close`.

Also `new` is often associated with *forget me* `ref` objects.

------

### new pattern research
```
131: newSeq
184: newException
2: newSkContext
43: newConnection
17: newDaemonApi
1: newStringTable
3: newStringOfCap
47: newFuture
1: newPool
7: newAsyncEvent
7: newString
2: newMultistreamHandshakeException
22: newMultistream
2: newInvalidVarintException
1: newInvalidVarintSizeException
2: newChannel
16: newLPStreamEOFError
3: newStreamInternal
4: newLPStreamLimitError
21: newMplex
13: newStream
4: newMuxer
5: newMuxerProvider
9: newSeqOfCap
9: newIdentify
4: newMessage
5: newTimedCache
6: newMCache
3: newAsyncLock
15: newPubSub
63: newBufferStream
12: newPubSubPeer
9: newNoise
2: newPlainText
2: newSecioConn
4: newSecio
7: newStandardSwitch
48: newTransport
5: newSwitch
2: newAlreadyPipedError
4: newNotWritableError
8: newLPStreamIncompleteError
2: newChronosStream
2: newAsyncStreamReader
2: newAsyncStreamWriter
6: newLPStreamReadError
9: newLPStreamIncorrectError
4: newLPStreamWriteError
6: newNoPubSubException
3: newTestSelectStream
2: newTestLsStream
2: newTestNaStream
```

arnetheduck:
[...]

PR #116, opened 2020-03-25
Fix/misc

dryajov:
Fixes #110, #109.

PR #2775, opened 2026-07-07
chore(cbind): nim-ffi migration [4/9] — streams & custom protocols (+ echo example)

gmelodie:

1. #2772 — deps + pinning
2. #2773 — scaffold + CI
3. #2774 — connectivity
4. #2775 — streams (+ echo) ← this PR
6. #2777 — kad + CID + crypto
7. #2778 — service discovery

---

Part 4 of a 9-PR train that replaces the hand-rolled `cbind/` C bindings with the generated **nim-ffi** codegen, landing it domain by domain so each PR stays reviewable and CI-green (build the FFI lib + generate bindings + build/run examples).

Targets `chore/cbind/ffi/03-connectivity` (stacked).

**This PR adds**
- Stream ops over `streamId` handles: `readExactly`, `readLp`, `write`, `writeLp`, `close`, `closeWithEof`, `release`.
- `libp2pMountProtocol` + the `onIncomingStream` `{.ffiEvent.}`, plus the `StreamRegistry` release-waiter logic that keeps custom-protocol handlers alive across the FFI boundary until the host releases the stream.
- `validateReadLength` hardening — rejects negative lengths and anything past the `MaxReadBytes` (64 MiB) DoS ceiling before it sizes an allocation.
- The **echo** C example plus its build harness: `common.h`, `CMakeLists.txt`, `README.md`, updated `Makefile`, and the `examples` nimble task.

**Lifecycle procs now return `bool`**

`libp2pStart`, `libp2pStop`, `libp2pConnect`, and `libp2pDisconnect` return `Result[bool, string]` instead of `Result[void, string]`. This is the first PR that compiles `libp2p_ffi.nim` in the always-on `c-bindings` job, and nim-ffi's async dispatch can't marshal a `void` success — its generated handler lambda unwraps the result value, which doesn't exist for `void`. A trivial `bool` gives every lifecycle call a uniform `on_bool` reply, which the echo example already consumes.

**Also**
- Bumps pinned nim-ffi to `7ef58f4` (latest master): picks up compile-time `--nimMainPrefix` validation (#116), the loud scalar-fast-path binding drop (#120), and the shorter genbindings incantation (#119). All are satisfied by the current build flags (`--nimMainPrefix:liblibp2p`) and the default CBOR ABI, so no binding source change was needed.
- Points `buildffi`/`genbindings`/`examples` at the installed `ffi` and `cbor_serialization` packages via `--path`. The `c-bindings` job runs under nimble's project-local deps mode, and those two packages aren't cbind `requires` (not in the registry), so `nimble setup` leaves them off `nimble.paths`. Their transitive deps are libp2p deps already on the inherited root paths.

**Review artifact:** the generated header diff shows the stream/protocol entry points and the incoming-stream event.


vladopajic (approved):

cbind/examples/echo.c
```diff
@@ -1,375 +1,204 @@
...
-  }
-
+  AWAIT_BOOL(bw, libp2p_ctx_stop(client, on_bool, &bw), "stop client");
+  AWAIT_BOOL(bw, libp2p_ctx_stop(server, on_bool, &bw), "stop server");
+  libp2p_ctx_destroy(client);
+  libp2p_ctx_destroy(server);
   return status;
 }
-
-static void event_handler(int callerRet, const char *msg, size_t len,
```
gmelodie, line 173: These and lines below got moved to `common.h`

cbind/libp2p_ffi.nim
```diff
@@ -397,4 +438,124 @@ proc libp2pDialCircuitRelay*(
...
+  let streams = lib.streams
+  proc handle(
+      stream: Stream, selectedProto: string
+  ) {.async: (raises: [CancelledError]).} =
+    let streamId = streams.register(stream)
+    defer:
+      streams.release(streamId)
+    let releaseWaiter =
+      Future[void].Raising([CancelledError]).init("cbind custom protocol release")
+    streams.releaseWaiters[streamId] = releaseWaiter
```
richard-ramos, line 545: Should we do a stream_release on all releaseWaiters in libp2pStop / destroy?

richard-ramos (approved):

PR #2778, opened 2026-07-07
chore(cbind): nim-ffi migration [7/9] — service discovery + extended peer records

gmelodie:

1. #2772 — deps + pinning
2. #2773 — scaffold + CI
3. #2774 — connectivity
4. #2775 — streams (+ echo)
6. #2777 — kad + CID + crypto
7. #2778 — service discovery ← this PR

---
Part 7 of a 9-PR train that replaces the hand-rolled `cbind/` C bindings with the generated **nim-ffi** codegen, landing it domain by domain so each PR stays reviewable and CI-green (build the FFI lib + generate bindings + build/run examples).

Targets `chore/cbind/ffi/06-kad-crypto` (stacked).

**This PR adds**
- Service discovery: `start`, `stop`, `startAdvertising`, `stopAdvertising`, `lookup`, `randomLookup`, `registerInterest`, `unregisterInterest`.
- `libp2pCreateXpr` and `libp2pDecodeXpr` (the extended-peer-record encode/decode, including the previously-missing `decode_xpr` path).
- Their `{.ffi.}` request/response types.

**Review artifact:** Header diff shows the service-discovery + xpr entry points; verify `libp2p_ctx_decode_xpr` is present.

### DIFF
--- a/cbind/libp2p_ffi.nim
+++ b/cbind/libp2p_ffi.nim
@@ -14,7 +14,8 @@
 
 import ffi
 
-import std/[tables, sequtils, times]
+import std/[tables, sequtils]
+from std/times import getTime, toUnix
 
 import ../libp2p
 import ../libp2p/[multiaddress, peerid]
--- a/cbind/libp2p_ffi.nim
+++ b/cbind/libp2p_ffi.nim
@@ -864,9 +864,7 @@ proc libp2pCreateXpr*(
   for s in req.services:
     services.add(ServiceInfo(id: s.id, data: Opt.some(s.data)))
 
-  let peerRecord = ExtendedPeerRecord.init(
-    peerId = peerInfo.peerId, addresses = addresses, seqNo = seqNo, services = services
-  )
+  let peerRecord = ExtendedPeerRecord.init(peerInfo.peerId, addresses, seqNo, services)
 
   let xpr = SignedExtendedPeerRecord.build(peerInfo.privateKey, peerRecord).valueOr:
     return err(error)
--- a/cbind/libp2p_ffi.nim
+++ b/cbind/libp2p_ffi.nim
@@ -14,7 +14,7 @@
 
 import ffi
 
-import std/[tables, sequtils]
+import std/[tables, sequtils, times]
 
 import ../libp2p
 import ../libp2p/[multiaddress, peerid]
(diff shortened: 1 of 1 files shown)


cbind/libp2p_ffi.nim
```diff
@@ -764,6 +776,116 @@ proc libp2pKadRandomRecords*(
...
+      ExtendedPeerRecord.init(
+        peerId = peerInfo.peerId, addresses = addresses, services = services
+      )
+    else:
+      ExtendedPeerRecord.init(
+        peerId = peerInfo.peerId,
+        addresses = addresses,
+        seqNo = req.seqNo,
+        services = services,
+      )
```
vladopajic: this pattern indicates that `ExtendedPeerRecord.init` is not ideal for it's use case. nim-libp2p being a library should pay attention to use case and improve it. in example here having to work with 2 `ExtendedPeerRecord.init`  instead of one:


call site should do this:
```
   seqNo = if req.seqNo != 0: Opt.some(req.seqNo) else: Opt.none(uint64)
```
^ or whatever else
```
   seqNo = if req.seqNo != 0: req.seqNo else: ExtendedPeerRecord.defaultSeqNo()
```


interface first case can be this:
```
proc init*(
    ...
    seqNo: uint64 | Opt[uint64] = getTime().toUnix().uint64,
): T =
  T(
    seqNo: .... use argument or use default (if Opt.none was used)
  )
  
```
and for second case is unchanged.

vladopajic (approved):

cbind/libp2p_ffi.nim
vladopajic: or passing 0 value for `seqNo` to `init` should also mean to use default seq no?

```
proc init*(
    ...
    seqNo: uint64 = 0
): T =
  T(
    seqNo: if seqNo !=0: seqNo else: getTime().toUnix().uint64,
  )
```

vladopajic: ^ this one seems the most elegant? but not sure if `0` should be valid possible value

richard-ramos: For other impls, at least for the PeerRecord (not the extended one), the 0 is a valid value. It does not make much difference to start at 0 or 1 tho, since the important thing is that the sequence increments everytime you sent an update

cbind/libp2p_ffi.nim
```diff
@@ -760,6 +777,113 @@ proc libp2pKadRandomRecords*(
...
+      getTime().toUnix().uint64
+    else:
+      req.seqNo
+
+  var services: seq[ServiceInfo]
+  for s in req.services:
+    services.add(ServiceInfo(id: s.id, data: Opt.some(s.data)))
+
+  let peerRecord = ExtendedPeerRecord.init(
+    peerId = peerInfo.peerId, addresses = addresses, seqNo = seqNo, services = services
```
vladopajic: nit avoid named arguments

richard-ramos (approved):

PR #2793, opened 2026-07-09
fix(identify): handle missing optional pub key

### LINKED ISSUE #2786: identify: absent publicKey is rejected and drops the connection
### Description

When nim-libp2p identifies a peer whose Identify message omits the `publicKey` field, nim rejects it and the connection is dropped. This breaks interop. In the libp2p unified-testing transport suite, nim against eth-p2p-z over QUIC fails in both dial directions.

Dialer error:

```
Failed to finish outgoing upgrade in internalConnect: No pubkey in identify
```

Unlike #2781, the message decodes fine. The failure is a post-decode check: `identify()` treats `publicKey` as mandatory and raises `IdentityInvalidMsgError("No pubkey in identify")` when it is absent. nim runs identify on both the incoming and outgoing upgrade, so both directions fail. When nim is the listener it identifies the dialer too and tears the connection down on the same raise, which the dialer observes as `EndOfStream`. A single fix to the check below makes both directions pass.

eth-p2p-z's transport-interop peer builds its Identify from default options, which leave `publicKey` unset, so the field is genuinely off the wire (see [eth-p2p-z main.zig L269](https://github.com/zen-eth/eth-p2p-z/blob/bb368b1932aca9c32bd9da74f592a5c5f1ba3bc0/interop/transport/main.zig#L269), `IdentifyHandler.init`). This is within spec: on an authenticated transport (QUIC-TLS, Noise) the remote peer id is already established by the security handshake and passed into `identify()` as `remotePeerId`, so the `publicKey` field only restates that identity.

### Spec Recommendation

The identify spec makes `publicKey` optional and says nothing about a receiver rejecting a message that omits it.

- The field is [`optional bytes publicKey = 1;`](https://github.com/libp2p/specs/blob/master/identify/README.md?plain=1#L80).
- The [entire normative text on the field](https://github.com/libp2p/specs/blob/master/identify/README.md?plain=1#L104-L107) is "This is the public key of the peer, marshalled in binary form as specified in peer-ids." No MUST/SHOULD to send it, no receiver-side requirement to verify or reject.
- "[Note that missing fields should be ignored](https://github.com/libp2p/specs/blob/master/identify/README.md?plain=1#L68-L71), as peers may choose to send partial updates."

A peer omitting `publicKey` is within spec.

### Implementation

`publicKey` is a single `Opt[PublicKey]`:

https://github.com/vacp2p/nim-libp2p/blob/bd689933bddb3e3ef3a3b76cbf3916a328becf7f/libp2p/protocols/identify.nim#L48

`identify()` raises when the field is absent, then enforces the derived peer id matches `remotePeerId`:

https://github.com/vacp2p/nim-libp2p/blob/bd689933bddb3e3ef3a3b76cbf3916a328becf7f/libp2p/protocols/identify.nim#L177-L185

The single `PublicKey` field decoder in crypto.nim raises `ProtobufValueError` on a malformed key, the same anti-pattern #2781 fixed for `MultiAddress`. eth-p2p-z omits the field so it never reaches this path, but it is a latent whole-message abort worth fixing in the same pass:

https://github.com/vacp2p/nim-libp2p/blob/bd689933bddb3e3ef3a3b76cbf3916a328becf7f/libp2p/crypto/crypto.nim#L1010-L1025

### Reference behavior (go-libp2p)

go-libp2p treats a missing `publicKey` as a non-event and uses the connection's authenticated peer id. The call site says so directly. See [go-libp2p id.go L834-L835](https://github.com/libp2p/go-libp2p/blob/95be6665b014e3e29bdb639ec10f8944d141969d/p2p/protocol/identify/id.go#L834-L835):

```go
// get the key from the other side. we may not have it (no-auth transport)
ids.consumeReceivedPubKey(c, mes.PublicKey)
```

The consumer logs at debug and returns when the key is absent, so identify completes normally. See [go-libp2p id.go L877-L886](https://github.com/libp2p/go-libp2p/blob/95be6665b014e3e29bdb639ec10f8944d141969d/p2p/protocol/identify/id.go#L877-L886):

```go
func (ids *idService) consumeReceivedPubKey(c network.Conn, kb []byte) {
	lp := c.LocalPeer()
	rp := c.RemotePeer()

	if kb == nil {
		log.Debug("did not receive public key for remote peer",
			"local_peer", lp,
			"remote_peer", rp)
		return
	}
```

The identity go-libp2p trusts is `c.RemotePeer()`, the peer id from the security handshake, which is the `remotePeerId` fallback proposed above. go-libp2p only requires a pubkey in the separate `signedPeerRecord` field, which eth-p2p-z is not using here.

### Related

#2781 - the same identify path, empty `observedAddr` (a decoder raise rather than a mandatory-field check).

rlve:
## Summary

Addresses https://github.com/vacp2p/nim-libp2p/issues/2786

1. **identify: accept a missing `publicKey`.** identify treated the optional `publicKey` field as mandatory and dropped the connection when it was absent, breaking interop with minimal QUIC-only peers. It now falls back to the peer id from the security handshake (`remotePeerId`) when the field is missing, and still enforces the match when it is present.

2. **test-only: compare advertised addresses as sets.** The identify-push tests asserted exact equality between the peer store's address list and `peerInfo.addrs`. The `AddressBook` keeps a unique address per peer, but wildcard listen addrs (`/ip6/::`) expand to every interface address, which can include duplicates when one link-local is shared across interfaces (macOS `awdl0`/`llw0`). The tests now compare address sets, matching the assertion's intent. 
This only fixes local failures on macOS. CI was not affected.

## Affected Areas

    identify `publicKey` handling and its tests.

Backward compatible. Only relaxes a receiver-side check, no wire format or API signature change, so no downstream build or integration work is required.

## Impact on Library Users

A peer that sends identify without `publicKey` is now accepted instead of dropped. On authenticated transports (Noise, TLS, QUIC) the remote identity is unchanged because it comes from the handshake. For such peers `IdentifyInfo.pubkey` is `none`, so `KeyBook` is not populated for them. Existing consumers already treat that field as optional.

## Risk Assessment

Low. The change removes a rejection path rather than adding one. A present-but-mismatched `publicKey` still raises `IdentityNoMatchError`.

### DIFF
--- a/tests/libp2p/protocols/test_identify.nim
+++ b/tests/libp2p/protocols/test_identify.nim
@@ -3,8 +3,7 @@
 
 {.used.}
 
-import std/sets
-import chronos, chronicles
+import chronos, chronicles, protobuf_serialization, sets
 import
   ../../../libp2p/[
     protocols/identify,
@@ -21,7 +20,6 @@ import
     crypto/crypto,
     upgrademngrs/upgrade,
   ]
-import protobuf_serialization
 import ../../tools/[unittest, crypto, multiaddress, switch_builder]
 
 type IdentifyMsgNoPubKey {.proto2.} = object
--- a/tests/libp2p/protocols/test_identify.nim
+++ b/tests/libp2p/protocols/test_identify.nim
@@ -21,7 +21,11 @@ import
     crypto/crypto,
     upgrademngrs/upgrade,
   ]
+import protobuf_serialization
 import ../../tools/[unittest, crypto, multiaddress, switch_builder]
+
+type IdentifyMsgNoPubKey {.proto2.} = object
+  protocols {.fieldNumber: 3.}: seq[string]
 
 suite "Identify":
   teardown:
@@ -110,6 +114,38 @@ suite "Identify":
       check id.agentVersion.get() == customAgentVersion
       check id.protos == @["/test/proto1/1.0.0", "/test/proto2/1.0.0"]
       check id.signedPeerRecord.isNone()
+
+    asyncTest "identify without publicKey falls back to remotePeerId":
+      # A minimal peer (e.g. eth-p2p-z over QUIC) may omit the optional publicKey field.
+      # The security handshake already authenticated the remote, so identify
+      # must accept the message and fall back to remotePeerId instead of rejecting.
+      const testProtos = @["/test/proto1/1.0.0"]
+      proc noPubKeyHandler(
+          stream: Stream, proto: string
+      ) {.async: (raises: [CancelledError]).} =
+        try:
+          await stream.writeLp(
+            Protobuf.encode(IdentifyMsgNoPubKey(protocols: testProtos))
+          )
+        except LPError as e:
+          trace "failed to write pubkey-less identify", description = e.msg
+        finally:
+          await stream.closeWithEOF()
+
+      msListen.addHandler(LPProtocol.new(@[IdentifyCodec], noPubKeyHandler))
+      proc acceptHandler(): Future[void] {.async.} =
+        let c = await transport1.accept()
+        await msListen.handle(c)
+
+      acceptFut = acceptHandler()
+      conn = await transport2.dial(transport1.addrs[0])
+
+      discard await msDial.select(conn, IdentifyCodec)
+      let id = await identifyProto2.identify(conn, remotePeerInfo.peerId)
+
+      check id.pubkey.isNone()
+      check id.peerId == remotePeerInfo.peerId
+      check id.protos == testProtos
 
     asyncTest "handle failed identify":
       msListen.addHandler(identifyProto1)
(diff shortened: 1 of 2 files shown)


tests/libp2p/protocols/test_identify.nim
```diff
@@ -110,6 +115,38 @@ suite "Identify":
...
+      const testProtos = @["/test/proto1/1.0.0"]
+      proc noPubKeyHandler(
+          stream: Stream, proto: string
+      ) {.async: (raises: [CancelledError]).} =
+        try:
+          await stream.writeLp(
+            Protobuf.encode(IdentifyMsgNoPubKey(protocols: testProtos))
+          )
+        except LPError as e:
+          trace "failed to write pubkey-less identify", description = e.msg
```
vladopajic: `raiseAssert` ?

logs in tests are usually unnecessary, unless when debugging

vladopajic (approved):

richard-ramos (approved):

gmelodie (approved):

PR #2794, opened 2026-07-09
feat(tests): tests as subproject of repo

### LINKED ISSUE #2768: make `test` directory as subroject just like `cbind` or `examples`
making test as sub project will enable library source code dependencies to be split for tests. this will be adventurous to nim-libp2p downstream deps as they will not have to download deps that are used for tests only.

vladopajic:
this PR switches `tests` dir to be subproject of repo, much like `cbind` and `examples`.

benefit is that downstream dependencies will not have to resolve dependencies that are needed only for testing purpose of this repo. they can have completely different approach (different libraries required for testing).

closes:  #2768


gmelodie (approved):

Makefile
richard-ramos, line 188: Should we also do:
```
install_pinned:
   nimble install_pinned
   cd tests && nimble install_pinned
```

so unittest2 and libbacktrace get downloaded too in the test folder? because as it is right now, those two would get skipped when running `make install_pinned`

richard-ramos:
```
    find . -name '*.nim' -not -path './nimbledeps/*' -not -path './tests/nimbledeps/*' | xargs nph
```

richard-ramos (approved):

PR #2797, opened 2026-07-09
ci: dropping i386 orc from tests

vladopajic:
dropping i386 orc from tests as it was causing OOM issue. also removing hacks related to supporting running tests for this platform.

### DIFF
--- a/.github/workflows/test_all.yml
+++ b/.github/workflows/test_all.yml
@@ -47,6 +47,14 @@ jobs:
             memory_management: refc
           - ref: v2.2.10
             memory_management: orc
+        exclude:
+          - platform:
+              os: linux
+              cpu: i386
+              cc: gcc
+            nim:
+              ref: v2.2.10
+              memory_management: orc
         include:
           - platform:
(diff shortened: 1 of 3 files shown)


.github/workflows/test_all.yml
```diff
@@ -47,6 +47,14 @@ jobs:
...
           - ref: v2.2.10
             memory_management: orc
+        exclude:
+          - platform:
+              os: linux
+              cpu: i386
+              cc: gcc
+            nim:
+              ref: v2.2.10
+              memory_management: orc
```
gmelodie, line 57: To be clear, then we still running i386 + nim 2.2.4, and this is what we want, correct? (why not drop 2.2.4 as well)

gmelodie (approved):

.github/workflows/test_all.yml
vladopajic: i386 + orc + nim 2.2.4 is not in the matrix

richard-ramos (approved):

Issue #2800, opened 2026-07-09
identify: outgoing identify hangs against peers that don't serve identify


rlve:
### Description

When nim dials a peer that does not serve identify, the outgoing identify hangs forever. There is no error and no timeout. Confirmed in the libp2p unified-testing transport suite with `nim-v2.2.0 x lua-v0.1.0` over tcp/noise/yamux,
where lua serves ping only. nim as the dialer hung for 180s until the harness killed it. nim as the listener was fine.

Root cause. `PeerStore.identify` opens a stream and multistream-selects `/ipfs/id/1.0.0`. The peer answers `na` because it does not serve identify. `select` returns `""` and nim correctly skips identify. But the `finally` then calls `closeWithEOF()` on every non-cancelled path, including this one. `closeWithEOF` sends `Fin` and blocks in `readOnce` waiting for the remote to half-close. A peer that answered `na` is under no obligation to half-close, so the read never returns.

`closeWithEOF`'s own contract says it must only be used after an in-protocol notification that no more data will arrive. A multistream `na` is a negotiation result, not such a notification. On the `na` path the stream should be reset.

This is not specific to identify. The general dial path `Dialer.negotiateStream` does the same `closeWithEOF`-after-`na` on its failure branch, so any protocol dialed against a peer that does not support it can hang the same way. Identify is the only path that speculatively probes every outbound connection regardless of peer support, which is why it surfaces here first. `IdentifyPusher` closes the same way in its `finally`.

### Implementation

The identify probe. `select` returns `""` on `na`, and the `finally` still runs `closeWithEOF`:

https://github.com/vacp2p/nim-libp2p/blob/766d4da40dcf1c24de58a659e950d9bb40a095fa/libp2p/peerstore.nim#L551-L571

`closeWithEOF` sends Fin then blocks reading for the remote EOF. Note the contract in the doc comment:

https://github.com/vacp2p/nim-libp2p/blob/766d4da40dcf1c24de58a659e950d9bb40a095fa/libp2p/stream/lpstream.nim#L304-L338

`na` maps to an empty selected protocol:

https://github.com/vacp2p/nim-libp2p/blob/766d4da40dcf1c24de58a659e950d9bb40a095fa/libp2p/multistream.nim#L86-L88

Same pattern in the general dial path. The unselected branch and the outgoing-budget branch both `closeWithEOF`:

[...]

PR #2802, opened 2026-07-09
test(identify): repro test for #2800

rlve:
## Summary

Adds a regression test for #2800

The test reproduces this with an in-memory mplex pair. The listener serves ping only and leaves the stream open after answering `na`. The dialer runs the real `PeerStore.identify`, which blocks in `closeWithEOF`.


tests/libp2p/protocols/test_identify.nim
```diff
@@ -337,3 +341,51 @@ suite "Identify":
...
+      except CatchableError:
+        discard
+      await noCancel stream.close()
+
+    let
+      remoteInfo = PeerInfo.new(PrivateKey.random(PKScheme.Ed25519, rng()).get())
+      peerStore = PeerStore.new(Identify.new(remoteInfo))
+    let identifyFut = peerStore.identify(muxDialer, Direction.Out)
+
+    await sleepAsync(200.milliseconds)
```
vladopajic, line 385: why is sleep needed? in general we should always document it - if we cant have non sleep solution like waiting on event or synchronizing with futs, waitgroups or eventqueue

vladopajic (approved):

tests/libp2p/protocols/test_identify.nim
vladopajic: could you do `await handlerCalledFut` here and in handler on begging do `handlerCalledFut.complete()`

gmelodie (approved):

tests/libp2p/protocols/test_identify.nim
rlve: I have added a comment explaining why the sleep is needed. The other suggestion wouldn't work in this case.

PR #2805, opened 2026-07-09
fix(service-disco): extend IP-similarity scoring to IPv6

### LINKED ISSUE #2756: service-disco: extend IP-similarity scoring to IPv6
### Description

The IP tree that backs IP-similarity scoring is IPv4-only. Before any address reaches the tree, `filterIPv4` drops every IPv6 address, and the tree operations (`insertIp` / `removeIp` / `ipScore`) hard-assert IPv4. So `adScore`, `insertAd` and `removeAd` only ever see IPv4 addresses.

The consequence is a Sybil-throttle gap. An advertisement whose addresses are all IPv6 scores `adScore == 0.0`, so it contributes nothing to the `ipSim` term of the waiting time. A flood of pure-IPv6 ads from a single /48 or /64 evades the IP-similarity component of the waiting-time throttle that the same flood over an IPv4 /24 would hit. Dual-stack ads are unaffected, they are still scored on their IPv4 address.

This is behaviorally pinned by two component tests in https://github.com/vacp2p/nim-libp2p/pull/2746: an IPv6-only node gets consecutive advertisements Confirmed with no waiting time, while a dual-stack node is told to Wait on its second advertisement.

```nim
  asyncTest "an IPv6 node registers three services one after another":
    # A node listening only on IPv6 advertises only IPv6 addresses, which never
    # enter the IPv4-only IP tree, so its advertisements draw no IP-similarity
    # waiting time: each is Confirmed and cached, even under default settings.
    let registrarNode = setupServiceDiscoveryNode()
    let advertiserNode = setupServiceDiscoveryNode(addresses = @[TcpAutoAddressIP6()])
    startAndDeferStop(@[registrarNode, advertiserNode])
    await connect(registrarNode, advertiserNode)

    for i in 1 .. 3:
      let service = makeServiceInfo("ipv6-service-" & $i)
      let serviceId = service.id.hashServiceId()
      advertiserNode.addProvidedService(service)
      checkUntilTimeout:
        registrarNode.countAdsInCache(serviceId) == 1
```

### Spec Recommendation

The spec RECOMMENDS the IP tree as a 32-level binary tree over the 32 bits of an IPv4 address, and makes IPv6 an explicit optional extension:

> Implementations can extend the IP tree algorithms to IPv6 by using a 128-level binary tree, corresponding to the 128-bit length of IPv6 addresses.

So the IPv4-only tree is spec-compliant today, and this is a request to implement the optional IPv6 extension so IPv6 and IPv6-heavy advertisements are also subject to IP-similarity scoring. See the [IP Similarity Score section](https://lip.logos.co/anoncomms/raw/logos-service-discovery.html#ip-similarity-score).

tinniaru3005:
## Summary

Fixes #2756. The IP-similarity scoring used to throttle Sybil floods
in service discovery only worked for IPv4:

- `filterIPv4` dropped every IPv6 address before it reached the IP tree.
- `insertIp` / `removeIp` / `ipScore` in the IP tree hard-asserted IPv4.

An advertisement whose addresses were all IPv6 therefore always scored
`adScore == 0.0`, contributing nothing to the `ipSim` term of the
waiting-time formula. A flood of pure-IPv6 ads from a single /48 or
/64 could bypass the IP-similarity throttle that the same flood over
an IPv4 /24 would hit. Dual-stack ads were unaffected since they still
had an IPv4 address to score.

### Fix: a separate tree per address family

Rather than merging IPv6 addresses into the existing IPv4 tree, `IpTree`
now holds two independent roots:

```nim
IpTree* = ref object
  root*: IpTreeNode   ## IPv4 tree root: a 32-level binary tree.
  root6*: IpTreeNode  ## IPv6 tree root: a 128-level binary tree.
```

## Affected Areas

- [ ] Gossipsub  
  

- [ ] Transports  
  

- [ ] Protocol Logic

- [ ] Build / Tooling

- [ ] Other  
  

- **Nimbus:**  
  

- **Waku:**  
  

- **Codex:**


richard-ramos (approved):

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -21,34 +21,35 @@ logScope:
 proc updateRegistrarMetrics(registrar: Registrar) {.raises: [].} =
   cd_registrar_cache_ads.set(registrar.cacheTimestamps.len.float64)
   cd_registrar_cache_services.set(registrar.cache.len.float64)
-  cd_iptree_unique_ips.set(registrar.ipTree.root.counter.float64)
+  cd_iptree_unique_ips.set(
+    registrar.ipTree.root.counter.float64 + registrar.ipTree.root6.counter.float64
+  )
```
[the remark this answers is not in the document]
gmelodie: Not really an issue introduced by this PR, but would be nice to fix it here.

libp2p/utils/iptree.nim
```diff
@@ -1,29 +1,43 @@
...
+  of IpAddressFamily.IPv6:
+    @(ip.address_v6)
+
+proc treeRoot(ipTree: IpTree, ip: IpAddress): IpTreeNode =
+  case ip.family
+  of IpAddressFamily.IPv4: ipTree.root
+  of IpAddressFamily.IPv6: ipTree.root6
 
-  var v = ipTree.root
+proc insertIp*(ipTree: IpTree, ip: IpAddress) {.raises: [].} =
```
gmelodie, line 43: To fix the issue raised by copilot above, I think the path is making `insertIp` first check if ip is already inserted, which would require two passes on the tree, which should be ok/negligible overhead

libp2p/utils/iptree.nim
gmelodie: nit: maybe we could make this logic consistent with insert (and wherever else we traverse the tree) by extracting it into a function that receives an optional 

at the very least use the `nxt` trick to avoid `if...else` on update

but ideally we could rework this (shouldn't be much code):

```nim
Path = object
    nodes: array[MaxBits + 1, Node] # nodes[0] is the root
    len: int

func last(p: Path): Node =
  p.nodes[p.len - 1]

func bitSet(key: openArray[byte], bit: int): bool =
  ## MSB-first, so tree order matches prefix order.
  (key[bit shr 3] and (0x80'u8 shr uint8(bit and 7))) != 0

proc descend(t: var Tree, key: openArray[byte], create: bool): Opt[Path] =
  ## Walks `key` from the root, recording every node visited. With `create`,
  ## missing links are allocated and the result is always `some`; otherwise the
  ## first gap yields `none`.
  if t.root.isNil:
    if not create:
      return Opt.none(Path)
    t.root = Node()

  var
    path = Path(len: 1)
    v = t.root
  path.nodes[0] = v

  for bit in 0 ..< key.len * 8:
    let right = key.bitSet(bit)
    var nxt =
      if right: v.right
      else: v.left

    if nxt.isNil:
      if not create:
        return Opt.none(Path)
      nxt = Node()
      if right:
        v.right = nxt
      else:
        v.left = nxt

    v = nxt
    path.nodes[path.len] = v
    inc path.len

  Opt.some(path)
```

libp2p/utils/iptree.nim
```diff
@@ -73,36 +85,37 @@ proc removeIp*(ipTree: IpTree, ip: IpAddress) {.raises: [].} =
       dec n.counter
```
gmelodie: Also adjacent, but shouldn't this delete the created nodes instead of just decreasing the counter? (the suggestion above removes the nodes)

libp2p/utils/iptree.nim
gmelodie: Then the operations would be
```nim
proc contains*(t: var Tree, key: openArray[byte]): bool =
  let path = t.descend(key, create = false).valueOr:
    return false
  path.last.present

proc insert*(t: var Tree, key: openArray[byte]): bool =
  ## True if newly added.
  let v = t.descend(key, create = true).expect("create always succeeds").last
  let isNew = not v.present
  v.present = true
  isNew

proc remove*(t: var Tree, key: openArray[byte]): bool =
  ## True if the key was present. Prunes nodes left empty.
  let path = t.descend(key, create = false).valueOr:
    return false

  let v = path.last
  if not v.present:
    return false
  v.present = false

  for i in countdown(path.len - 1, 1):
    let n = path.nodes[i]
    if n.present or not n.left.isNil or not n.right.isNil:
      break

    let p = path.nodes[i - 1]
    if p.left == n:
      p.left = nil
    else:
      p.right = nil

  if not t.root.isNil and not t.root.present and t.root.left.isNil and
      t.root.right.isNil:
    t.root = nil

  true
```

libp2p/utils/iptree.nim
```diff
@@ -73,36 +85,37 @@ proc removeIp*(ipTree: IpTree, ip: IpAddress) {.raises: [].} =
...
+  let nBits = bytes.len * 8
 
-  for i in 0 ..< 4:
-    let b = bytes[i]
+  var depth = 0
+  for b in bytes:
     for bit in countdown(7, 0):
-      let depth = i * 8 + (7 - bit) # 0 .. 31; child node sits at tree level depth+1
-      let threshold = total / float64(1'u64 shl (depth + 1))
+      let threshold = total / pow(2.0, float64(depth + 1))
```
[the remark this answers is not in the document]
gmelodie: Indeed! If `total` is loop-invariant, `threshold` simply halves each step and you need no inverse at all

```nim
var threshold = total * 0.5
for bit in countdown(7, 0):
  threshold *= 0.5
  # ...
```

libp2p/protocols/service_discovery/registrar.nim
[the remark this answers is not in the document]
tinniaru3005: Yup thats a good point.

libp2p/utils/iptree.nim
tinniaru3005: I'd rather not dedupe `insertIp` — two reasons:

1. **Spec conflict.** `ADD_IP_TO_TREE()` in the RFC increments unconditionally, no presence check. The root counter is documented as *"total IPs currently in `ad_cache`"* — total, not unique. Current behavior matches spec.

2. **Weakens the throttle.** If a repeat address stops incrementing counters after the first insert, an attacker spamming many ads from one IP always looks like "1" to the tree — the score never rises no matter how many ads pile up from that address. That defeats what `ipSim` is for.

Went with renaming the metric instead (`cd_iptree_unique_ips` → `cd_iptree_total_ips`)

libp2p/utils/iptree.nim
tinniaru3005: Extracted a shared `step(v, goRight, create)` helper — `insertIp`, `removeIp`, and `ipScore` all call it now instead of each doing their own inline child-selection. `insertIp` used to write the nil-check-and-create logic out twice (once per branch); now it's one call, same as `removeIp` already did.

Didn't take the fuller `Path`/`descend`/`Opt`/`contains`/pruning-API sketch as-is — that design assumes a presence-based membership trie (`present: bool` per node). Our tree is counter-based: every ancestor's count feeds the subnet-similarity threshold in `ipScore`, not just leaf presence. Adopting that shape as-is would silently turn "how many ads from this subnet" into "how many distinct addresses".

libp2p/utils/iptree.nim
tinniaru3005: Good catch — and it led to something bigger than pruning: `removeIp` had a pre-existing off-by-one where it recorded each node into the removal path before stepping to its child, so it walked root through depth-(n-1) but never touched the actual leaf. The leaf's counter was never decremented on removal.

Fixed both: `removeIp` now decrements root through the actual leaf, and prunes the trailing run of now-empty, childless nodes back toward the root.

libp2p/utils/iptree.nim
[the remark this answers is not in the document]
tinniaru3005: Good call — swapped pow(2.0, depth+1) for incremental halving.

libp2p/utils/iptree.nim
gmelodie: alright!

gmelodie (approved):

PR #2807, opened 2026-07-10
fix(routing-record): use proto3 peer record encoding

richard-ramos:
## Summary

Switch peer routing records from proto2-style serialization to proto3, matching the libp2p routing records specification.

The spec defines `PeerRecord` as `syntax = "proto3"` with:
- `bytes peer_id = 1`
- `uint64 seq = 2`
- `repeated AddressInfo addresses = 3`

This PR updates `PeerRecord`, `AddressInfo`, and `ExtendedPeerRecord` to use proto3 serialization directly. Because proto3 does not enforce `required` fields, decoding now explicitly rejects records with a missing peer id or invalid address entries.

This is needed for interoperability with spec-compliant routing records, including records where proto3 default-valued scalar fields may be omitted from the wire encoding.

## Affected Areas

  - Signed peer routing records and extended peer records.
- [x] Protocol Logic
  - Peer record encode/decode validation and protobuf compatibility.

## Impact on Library Users

Peer routing record wire encoding now follows the libp2p proto3 format. Valid proto3 records that omit default-valued fields can be decoded, while malformed records are still rejected through explicit validation.

## Risk Assessment

The main risk is around decode compatibility at the peer-record boundary. This PR replaces proto2 `required` enforcement with explicit validation for fields that nim-libp2p still needs to treat as mandatory, such as peer id and valid multiaddresses.

## References

- https://github.com/libp2p/specs/blob/6b6203ee6f62938ce67efdb33498173f475851c0/RFC/0003-routing-records.md
```

### CODE REVIEW

File: libp2p/utils/protobuf.nim [hunk 1]
```diff
@@ -79,9 +80,20 @@ macro serializerFor*(
...
-          ok(`decodeName`(buf))
-        except SerializationError as e:
-          err("failed to decode " & $(`T`) & " from protobuf bytes. " & e.msg)
+        let decoded =
+          try:
+            `decodeName`(buf)
+          except SerializationError as e:
+            return err("failed to decode " & $(`T`) & " from protobuf bytes. " & e.msg)
+
+        when compiles(`T`.validateDecoded(decoded)):
```
gmelodie, line 89: cool!

gmelodie (approved):

vladopajic (approved):

PR #2808, opened 2026-07-10
fix(ci): stop dcutr test leaking a channel; run full daily CI on label

gmelodie:
## Summary

Three related changes:

1. **Fix the flaky `LPChannel` leak** in the daily-CI `Dcutr` suite that spills into every subsequent test.
2. **Make the `run-daily-ci` label trigger the full daily set**, so this fix can actually be validated on the runner that catches it.
3. **Fix the `Daily Nimbus` job** so it actually compiles against the PR's commit instead of nimbus's pinned `nim-libp2p`.

### 1. The Dcutr channel leak

`Dcutr::All client connect attempts fail` fails `checkTrackers()` with one leaked mplex `LPChannel` (`Opened 3242 / Closed 3241`). Because the chronos trackers are process-global and the orphaned channel never closes, every test that runs after it also fails its `teardown: checkTrackers()` — the observed `Dcutr .....FFFF` cascade off a single leak.

The leak is a teardown race in the inbound-upgrade path (`switch.nim` `upgrader`):

```
await connManager.storeMuxer(muxed)  # connCount becomes 2 HERE
await peerStore.identify(muxed, dir) # opens a NEW LPChannel (muxer.newStream)
triggerPeerEvents(..., Identified)   # fires only after identify closes its stream
```

The client tests waited with `checkUntilTimeout: connCount == 2`, which returns the instant `storeMuxer` bumps the count — before `peerStore.identify` has opened/closed its stream. The tests then stop both switches immediately, so on a slow runner (the macOS daily job) `switch.stop()` races the in-flight identify `LPChannel` and orphans it. On fast Linux the identify finishes first, so it does not reproduce there.

Fix: wait for the direct connection to be **identified** (`PeerEventKind.Identified`, fired only after the identify stream is opened and closed), not merely **connected**, before teardown — the same event-based approach already used for the QUIC test in #2762. Applied to the two client paths ("DCUtR establishes a new connection" and the `ductrClientTest` template covering "Client connect timeout" and "All client connect attempts fail").

The server tests (`connCount == 1`) and the QUIC test are left unchanged: the server tests' `dir = In` dial never completes a mux handshake against a plain accept (so it opens no `LPChannel`), and QUIC doesn't use mplex `LPChannel` — neither is a leak source, they only inherited the spill.

### 2. `run-daily-ci` label triggers the full daily set

The leak only surfaces on the macOS `Daily without feature flags` job, but the `run-daily-ci` label previously only triggered `Daily test all (latest dependencies)` (linux amd64/i386, windows) and `Daily runnable examples`. The two remaining daily workflows monitored by `daily_ci_report.yml` — `Daily without feature flags` (macos-15) and `Daily Nimbus` — ran on schedule only, so labeling a PR could not reproduce the failing job.

Added the existing `pull_request: [labeled, synchronize]` + `run-daily-ci` guard + concurrency block to those two workflows (`daily_test_all_no_flags.yml`, `daily_nimbus.yml`), so the label now runs exactly the set that constitutes daily CI. Scheduled/dispatch behavior is unchanged. `Daily i386` needs no separate workflow — the i386 matrix entry already lives in `Daily test all (latest dependencies)`, which is label-triggered.

### 3. `Daily Nimbus` compiles the PR's commit

The `Daily Nimbus` job wasn't testing `github.sha`: `git submodule set-branch --branch <sha>` had no effect because the subsequent `make` re-checked out the `nim-libp2p` pinned by nimbus, and it built the full nimbus target twice. It now clones nimbus-eth2, checks out `github.sha` in `vendor/nim-libp2p` via `git fetch`/`git checkout FETCH_HEAD`, and builds only `nimbus_beacon_node` (the target that uses libp2p). The redundant `actions/checkout` step is dropped, since the source comes from the nimbus submodule.

## Affected Areas

- [x] Build / Tooling
  

## Impact on Library Users

No impact — test- and CI-only changes. No library/API/behavior changes.


vladopajic (approved):

.github/workflows/daily_i386.yml
```diff
@@ -0,0 +1,27 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test_i386_latest:
+    name: Daily i386 (latest dependencies)
+    if: github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run-daily-ci')
+    uses: ./.github/workflows/daily_common.yml
```
richard-ramos: This was removed in `master`.

.github/workflows/daily_i386.yml
richard-ramos, line 1: Isn't this workflow redundant with daily_test_all_latest_deps.yml?  it already includes the same architectures and run-daily-ci code

.github/workflows/daily_nimbus.yml
richard-ramos: I just realized we dont really need these two lines, since we are cloning nimbus and using the libp2p source code from the submodule!
```suggestion
```

richard-ramos, line 33: So, i just realized that this daily job wasnt really testing the `github.sha` because the `make` in line 30 will reset the submodules and use those that are pinned by nimbus, and building the `nimbus_beacon_node`  twice, since it already does it with `make`, lol 🤦 

we should do something like this instead (and also just build the nimbus_beacon_node which is what uses libp2p)

```make
      - name: Compile nimbus using nim-libp2p
        env:
          GITHUB_SHA: ${{ github.sha }}
        run: |
          git clone --branch unstable --single-branch https://github.com/status-im/nimbus-eth2.git
          cd nimbus-eth2
          git submodule update --init --recursive
          cd vendor/nim-libp2p
          git fetch https://github.com/vacp2p/nim-libp2p.git "$GITHUB_SHA"
          git checkout FETCH_HEAD
          cd ../..

          make -j"$(nproc)" nimbus_beacon_node
```

richard-ramos (commented):
~~The .yml changes seem unrelated to this PR.~~
nvm. I just read the description

.github/workflows/daily_nimbus.yml
gmelodie: Unrelated to this pr tho, correct? or do you want me to add these changes here?

gmelodie: nvm claude figured out a fix, seems simple enough to add here

richard-ramos (approved):

Issue #2330, opened 2026-04-24
flaky test(floodsub): FloodSub message size validation 2


vladopajic:
https://github.com/vacp2p/nim-libp2p/actions/runs/24878623084/job/72841251826?pr=2327

```
[9](https://github.com/vacp2p/nim-libp2p/actions/runs/24878623084/job/72841251826?pr=2327#step:11:2740)
==================================
  D:\a\nim-libp2p\nim-libp2p\tests\test_all.exe "FloodSub Component::FloodSub message size validation 2"
----------------------------------
    [TIMEOUT] Timeout was reached and the conditions were not true. Check if the code is working as expected or consider increasing the timeout param.
    D:\a\nim-libp2p\nim-libp2p\tests\tools\unittest.nim(145, 11): Check failed: messageReceived == 1
```

PR #2810, opened 2026-07-11
fix(pubsub): resolve flaky FloodSub multiple-peer tests

### LINKED ISSUE #2798: bug: FloodSub multiple peers, no self trigger is flaky
https://github.com/vacp2p/nim-libp2p/actions/runs/29021282108/job/86128659350

```
========================================
  /home/runner/work/nim-libp2p/nim-libp2p/tests/test_all 'FloodSub Component::FloodSub multiple peers, no self trigger'
----------------------------------------
    [TIMEOUT] Timeout was reached and the conditions were not true. Check if the code is working as expected or consider increasing the timeout param.
    /home/runner/work/nim-libp2p/nim-libp2p/tests/tools/unittest.nim(93, 2): Check failed: nodes.allIt(it.floodsub.getOrDefault(topic).len == nodes.len - 1)

  [FAILED ] (  52.1s) FloodSub multiple peers, no self trigger

==========================================
  /home/runner/work/nim-libp2p/nim-libp2p/tests/test_all 'FloodSub Component::FloodSub multiple peers, with self trigger'
------------------------------------------
    [TIMEOUT] Timeout was reached and the conditions were not true. Check if the code is working as expected or consider increasing the timeout param.
    /home/runner/work/nim-libp2p/nim-libp2p/tests/tools/unittest.nim(93, 2): Check failed: nodes.allIt(it.floodsub.getOrDefault(topic).len == nodes.len - 1)
    [TIMEOUT] Timeout was reached and the conditions were not true. Check if the code is working as expected or consider increasing the timeout param.
    /home/runner/work/nim-libp2p/nim-libp2p/tests/libp2p/pubsub/component/test_floodsub.nim(218, 30): Check failed: n.floodsub[topic].len == numberOfNodes - 1
    [TIMEOUT] Timeout was reached and the conditions were not true. Check if the code is working as expected or consider increasing the timeout param.
    /home/runner/work/nim-libp2p/nim-libp2p/tests/libp2p/pubsub/component/test_floodsub.nim(218, 30): Check failed: n.floodsub[topic].len == numberOfNodes - 1
    [TIMEOUT] Timeout was reached and the conditions were not true. Check if the code is working as expected or consider increasing the timeout param.
    /home/runner/work/nim-libp2p/nim-libp2p/tests/libp2p/pubsub/component/test_floodsub.nim(218, 30): Check failed: n.floodsub[topic].len == numberOfNodes - 1
/home/runner/work/nim-libp2p/nim-libp2p/tests/tools/unittest.nim(18) test_floodsub
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/unittest2-0.2.5-02bb3751ba9ddc3c17bfd89f2e41cb6bfb8fc0c9/unittest2.nim(1175) testSuite
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/unittest2-0.2.5-02bb3751ba9ddc3c17bfd89f2e41cb6bfb8fc0c9/unittest2.nim(1094) runDirect
/home/runner/work/nim-libp2p/nim-libp2p/tests/tools/unittest.nim(40) runTest`gensym862
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/chronos-4.2.2-3a4c9477df8cef20a04e4f1b54a2d74fdfc2a3d0/chronos/internal/asyncfutures.nim(665) waitFor
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/chronos-4.2.2-3a4c9477df8cef20a04e4f1b54a2d74fdfc2a3d0/chronos/internal/asyncfutures.nim(640) pollFor
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/chronos-4.2.2-3a4c9477df8cef20a04e4f1b54a2d74fdfc2a3d0/chronos/internal/asyncengine.nim(150) poll
/home/runner/work/nim-libp2p/nim-libp2p/tests/tools/unittest.nim(110) anonymous
/home/runner/work/nim-libp2p/nim-libp2p/tests/libp2p/pubsub/component/test_floodsub.nim(218) checkExpiringInternal`gensym945
/home/runner/work/nim-libp2p/nim-libp2p/nim/lib/pure/collections/tables.nim(244) []
/home/runner/work/nim-libp2p/nim-libp2p/nim/lib/pure/collections/tables.nim(232) raiseKeyError
[[reraised from:
/home/runner/work/nim-libp2p/nim-libp2p/tests/tools/unittest.nim(18) test_floodsub
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/unittest2-0.2.5-02bb3751ba9ddc3c17bfd89f2e41cb6bfb8fc0c9/unittest2.nim(1175) testSuite
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/unittest2-0.2.5-02bb3751ba9ddc3c17bfd89f2e41cb6bfb8fc0c9/unittest2.nim(1094) runDirect
/home/runner/work/nim-libp2p/nim-libp2p/tests/tools/unittest.nim(40) runTest`gensym862
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/chronos-4.2.2-3a4c9477df8cef20a04e4f1b54a2d74fdfc2a3d0/chronos/internal/asyncfutures.nim(665) waitFor
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/chronos-4.2.2-3a4c9477df8cef20a04e4f1b54a2d74fdfc2a3d0/chronos/internal/asyncfutures.nim(640) pollFor
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/chronos-4.2.2-3a4c9477df8cef20a04e4f1b54a2d74fdfc2a3d0/chronos/internal/asyncengine.nim(150) poll
/home/runner/work/nim-libp2p/nim-libp2p/tests/tools/unittest.nim(110) anonymous
]]
[[reraised from:
/home/runner/work/nim-libp2p/nim-libp2p/tests/tools/unittest.nim(18) test_floodsub
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/unittest2-0.2.5-02bb3751ba9ddc3c17bfd89f2e41cb6bfb8fc0c9/unittest2.nim(1175) testSuite
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/unittest2-0.2.5-02bb3751ba9ddc3c17bfd89f2e41cb6bfb8fc0c9/unittest2.nim(1094) runDirect
/home/runner/work/nim-libp2p/nim-libp2p/tests/tools/unittest.nim(40) runTest`gensym862
/home/runner/work/nim-libp2p/nim-libp2p/nimbledeps/pkgs2/chronos-4.2.2-3a4c9477df8cef20a04e4f1b54a2d74fdfc2a3d0/chronos/internal/asyncfutures.nim(665) waitFor
]]

    Unhandled error: key not found: foobar [KeyError]

  [FAILED ] ( 134.2s) FloodSub multiple peers, with self trigger

```

paschal533:
Fixes #2798.

## What was going wrong

I dug into both failures separately since they turned out to have different root causes.

**"FloodSub multiple peers, with self trigger" (the KeyError crash)**

After the publish/receive part of the test finishes, there's a loop that calls `unsubscribeAll` on every node and then asserts that `floodsub[topic]` still has `numberOfNodes - 1` entries (the comment says "we keep the peers in table," meaning a node's own unsubscribe shouldn't affect what it knows about its peers' subscriptions).

The problem is that `unsubscribeAll` does more than update local state - it actually broadcasts a real "unsubscribe" RPC to every connected peer. And the assertion was wrapped in `checkUntilTimeout`, which polls on a `sleepAsync`, so it yields back to the event loop between checks. Since the loop unsubscribes all 10 nodes back to back, by the time later nodes in the loop get checked, earlier nodes' unsubscribe broadcasts have already arrived and been processed, so their peer count has already started shrinking. `PeerTable.removePeer` deletes a topic's table entry outright once its peer set hits zero, so eventually a node gets checked after every one of its peers has already unsubscribed, and the test's direct `floodsub[topic]` index (not `getOrDefault`) throws an unhandled KeyError, which takes the whole suite down with it - that's why the CI log shows a handful of `[TIMEOUT]` checkpoints before the crash (those are normal failed checks from nodes with a partially-drained but non-empty set) followed by the fatal exception.

Fix: run that assertion synchronously, immediately after each node's own `unsubscribeAll` call, with no `await` in between. Since Nim/chronos is single-threaded and cooperatively scheduled, nothing else can process incoming network data unless we yield, so this makes the check deterministic instead of racing the network. I also switched to `getOrDefault` there as a defensive measure so a future regression fails as a normal check instead of crashing the process.

**"FloodSub multiple peers, no self trigger" (the timeout)**

This one fails purely at mesh formation (`waitSubscribeStar`), before any publish happens. Setting up a full star of 10 nodes means 90 pubsub streams need to be dialed and negotiated, and the default 30s window is apparently not always enough on a loaded CI runner. This is the same failure signature as #2330, which got fixed in #2331 by bumping the timeout for a different floodsub test in the same file. I did the same thing here: `waitSubscribeStar` now takes an optional `timeout` param (default unchanged at 30s for every other caller) and both large floodsub star tests pass 90s.

## Verification

No Nim toolchain was available in my environment, so I set up a from-scratch Nim 2.2.10 + mingw-w64 build to actually reproduce this rather than fix it blind. With the fix applied, a clean run of the full FloodSub suite passes 12/12, including both previously flaky tests:

```
[Test   ]           FloodSub multiple peers, no self trigger
[OK     ] (  18.4s) FloodSub multiple peers, no self trigger
[Test   ]           FloodSub multiple peers, with self trigger
[OK     ] (  7.58s) FloodSub multiple peers, with self trigger
...
[Summary] 12 tests run (40.3s): 12 OK, 0 FAILED, 0 SKIPPED
```

I also saw one run in my sandbox get stuck entirely at the QUIC connection stage, and one run where both tests failed right after I'd force-killed a hung process on the same box - both look like local environment/resource artifacts (no UDP sockets ever bound for that process, consistent with something blocking QUIC in my sandbox) rather than anything related to this change, since the same binary ran cleanly end to end when the environment wasn't under that kind of stress. CI runs on real infrastructure without that constraint, so I'd treat this PR's own CI run as the actual confirmation.

## Changes

- `tests/libp2p/pubsub/component/test_floodsub.nim`: synchronous check instead of a polling one for the unsubscribeAll assertion, `getOrDefault` instead of direct indexing, 90s timeout on both large star tests.
- `tests/libp2p/pubsub/utils.nim`: `waitSubscribeStar` gains an optional `timeout` parameter (defaults to the previous 30s).


vladopajic (approved):

vladopajic:
@paschal533 there is just small issue with code formatting, that needs to be fixed to satisfy CI.

```
[1](https://github.com/vacp2p/nim-libp2p/actions/runs/29144712842/job/86820841761?pr=2810#step:3:292)
diff --git a/tests/libp2p/pubsub/component/test_floodsub.nim b/tests/libp2p/pubsub/component/test_floodsub.nim
index 1b72635..f4a1c1b 100644
--- a/tests/libp2p/pubsub/component/test_floodsub.nim
+++ b/tests/libp2p/pubsub/component/test_floodsub.nim
@@ -229,7 +229,8 @@ suite "FloodSub Component":
     for node in nodes:
       node.unsubscribeAll(topic)
       check:
-        node.floodsub.getOrDefault(topic).len == numberOfNodes - 1 # we keep the peers in table
+        node.floodsub.getOrDefault(topic).len == numberOfNodes - 1
+          # we keep the peers in table
         node.topics.len == 0 # remove the topic tho
```

usually, this check adds comments to PR with formatting fixes, but in this case revewdog failed

tests/libp2p/pubsub/component/test_floodsub.nim
```diff
@@ -211,12 +215,23 @@ suite "FloodSub Component":
...
+    # This assertion must run synchronously, immediately after unsubscribeAll
+    # and without any intervening `await`. `unsubscribeAll` broadcasts a real
+    # "unsubscribe" RPC to every connected peer, and `PeerTable.removePeer`
+    # deletes a topic's table entry entirely once its peer set becomes empty.
+    # If we polled/awaited here (e.g. via checkUntilTimeout), earlier nodes'
+    # broadcasts could reach later nodes in this loop before they're checked,
+    # draining - and eventually removing - their `floodsub[topic]` entry and
+    # crashing with an unhandled KeyError instead of a normal check failure.
+    # `getOrDefault` is used defensively so any regression here surfaces as a
+    # failed check rather than an unhandled exception that aborts the suite.
```
gmelodie:
```suggestion
    #
    # This assertion must run synchronously, immediately after unsubscribeAll.
    # `unsubscribeAll` broadcasts "unsubscribe" RPC messages to every connected 
    # peer. If we polled/awaited here (e.g. via checkUntilTimeout), earlier nodes'
    # broadcasts could reach later nodes before they're checked, eventually removing
    #  their `floodsub[topic]` entry and crashing with an unhandled KeyError.
```

tests/libp2p/pubsub/utils.nim
```diff
@@ -349,14 +349,21 @@ template waitSubscribeHub*[T: PubSub](hub: T, nodes: seq[T], topic: string): unt
...
+    nodes: seq[T], topic: string, timeout: Duration = 30.seconds
+): untyped =
   ## Star: 1-2; 1-3; 2-1; 2-3, 3-1, 3-2
-  ## 
+  ##
+  ## `timeout` defaults to the same 30s used by `checkUntilTimeout`. Larger
+  ## meshes (e.g. a full star of 10+ nodes) need every peer to dial and
+  ## negotiate a pubsub stream with every other peer before this converges,
+  ## which can occasionally exceed 30s on a loaded CI runner - callers with
+  ## many nodes should pass a larger `timeout`.
```
gmelodie:
```suggestion
```

tests/libp2p/pubsub/component/test_floodsub.nim
```diff
@@ -200,7 +203,8 @@ suite "FloodSub Component":
     await connectStar(nodes)
 
     subscribeAllNodes(nodes, topic, futs.mapIt(it[1]))
-    waitSubscribeStar(nodes, topic)
+    # See the equivalent comment in "FloodSub multiple peers, no self trigger".
```
gmelodie:
```suggestion
```

tests/libp2p/pubsub/component/test_floodsub.nim
```diff
@@ -164,7 +164,10 @@ suite "FloodSub Component":
     await connectStar(nodes)
 
     subscribeAllNodes(nodes, topic, futs.mapIt(it[1]))
-    waitSubscribeStar(nodes, topic)
+    # A full star of numberOfNodes needs numberOfNodes * (numberOfNodes - 1)
+    # pubsub streams to be dialed and negotiated; give this more headroom
+    # than the 30s default on loaded CI runners (see #2798).
```
gmelodie, line 169: 👍

gmelodie (approved):

richard-ramos (approved):
LGTM.
The only thing i'd change is that the new comments are a bit too extensive
but it's a nit

PR #2812, opened 2026-07-11
test: Expand stream tests to yamux

etan-status:
## Summary

streamTransportTest is only run on mplex not yamux. Add missing test.

- ipv4 / ipv6 tests separated into their own suite, otherwise they clash with their test names.

- yamux closes the stream if too much data is queued up, so split the tests into two: one for large transfer, and one for burst of small. when the stream is closed early, the test gets stuck as the future never completes.

- needs #2809 to pass

## Affected Areas

- [ ] Gossipsub  
  

- [X] Transports  
  

- [ ] Protocol Logic

- [ ] Build / Tooling

- [ ] Other

### DIFF
--- a/tests/libp2p/transports/stream_tests.nim
+++ b/tests/libp2p/transports/stream_tests.nim
@@ -220,7 +220,11 @@ template streamTransportTest*(
         await serverResetDone
 
         var buffer: array[1, byte]
-        check (await stream.readOnce(addr buffer[0], 1)) == 0
+        # EOF until this side reads the RST frame, reset once it lands
+        try:
+          check (await stream.readOnce(addr buffer[0], 1)) == 0
+        except LPStreamResetError:
+          discard
 
         expect LPStreamResetError:
           await stream.writeLp(fromHex("1234"))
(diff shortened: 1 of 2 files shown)


vladopajic (approved):

vladopajic (commented):
lgtm, it would be nice to have `yamux` coverage here, but this will create burden on us to maintain in - resolve flakiness that ci has now started to show. and there was initiative to remove / deprecate Yamux: https://github.com/vacp2p/nim-libp2p/issues/1635.

gmelodie (approved):

richard-ramos:
@etan-status ,  is nimbus looking into using yamux?
I'm curious, because of this https://github.com/vacp2p/nim-libp2p/issues/1635

etan-status:
No, it isn't. This was coincidental, I tried to run a test to verify that EOF is correctly reported on all transports as part of https://github.com/status-im/nimbus-eth2/pull/8744 and noticed that the test appears generic but doesn't actually work, so fixed it.

PR #2780, opened 2026-07-07
chore(cbind): nim-ffi migration [9/9] — flip to libp2p.nim, delete legacy cbind

gmelodie:
**nim-ffi cbind migration — PR train**

1. #2772 — deps + pinning
2. #2773 — scaffold + CI
3. #2774 — connectivity
4. #2775 — streams (+ echo)
5. #2776 — pubsub (+ gossipsub)
6. #2777 — kad + CID + crypto
7. #2778 — service discovery
8. #2779 — relay / peerstore / metrics
9. #2780 — flip + delete legacy ← this PR

---

Part 9 of a 9-PR train that replaces the hand-rolled `cbind/` C bindings with the generated **nim-ffi** codegen, landing it domain by domain so each PR stays reviewable. Rebased onto `master` now that parts 1–8 have merged.

> **Reviewing note — the diff looks bigger than the change.** GitHub renders `cbind/libp2p.nim` as ~2.5k churned lines, but the new file is **byte-for-byte the old `cbind/libp2p_ffi.nim`** (built up and reviewed across parts 2–8) except for a **single line** — the config include path (`include "libp2p_ffi/config"` → `include "libp2p/config"`). Git can't show this as a rename because `libp2p.nim` already existed with legacy content on both sides, so it pairs the ffi content against the deleted legacy file instead. Verify the real (one-line) change with:
>
> ```
> git diff <merge-base>:cbind/libp2p_ffi.nim HEAD:cbind/libp2p.nim
> ```

**This PR does**

- Makes `cbind/libp2p.nim` the nim-ffi module: the contents built up in `cbind/libp2p_ffi.nim` across parts 2–8 replace the legacy `libp2p.nim`, and `libp2p_ffi.nim` is deleted. See the reviewing note above — this is a verbatim move, not a rewrite.
- Deletes the legacy implementation: `alloc.nim`, `ffi_types.nim`, `types.nim`, `libp2p.h`, the whole `libp2p_thread/` tree, and `examples/cbindings.c`. All pure deletions.
- Points every build at the flipped module: `nix/cbind.nix`, `cbind/cbind.nimble`, and the mobile `nix/cbind-ffi-{android,ios}.nix` derivations now compile `cbind/libp2p.nim`.
- Drops the temporary `.#cbind-ffi` nix target and re-enables the `c-bindings-ffi` CI job (removes `if: false`) as macOS coverage, alongside the linux `cbindings` job that builds the FFI lib, generates the C/CDDL bindings, and runs the examples.
[...]

PR #2813, opened 2026-07-12
build(android): add libp2p_ffi mobile targets

richard-ramos:
## Summary

Adds Android build support for the `cbind/libp2p_ffi.nim` C ABI target.

This PR adds Nix packages for:

- `cbind-ffi-android-arm64-v8a`
- `cbind-ffi-android-x86_64`
- `cbind-ffi-android`

Each Android package builds shared/static `liblibp2p` artifacts, generated C headers/CDDL, required runtime/vendor libraries, and a small Android check harness that creates, starts, stops, and destroys a libp2p node through the generated C API.

It also adds Android CI coverage, Android build documentation, updates `nim-lsquic` to `0.5.5`, and fixes the `libp2p_ffi` start/stop path needed by the generated FFI bindings.

## Affected Areas

- [x] Build / Tooling  
  Adds Android Nix build targets, Android CI, Android artifact checks, and Android build documentation.

- [x] Other  
  Extends the nim-ffi C ABI packaging path used by downstream native consumers.

## Impact on Library Users

Downstream Android applications can build and package `liblibp2p.so` or `liblibp2p.a` for `arm64-v8a` and `x86_64`.

The Android outputs include the generated headers, CDDL schema, TinyCBOR headers, `libc++_shared.so`, and the vendored NAT traversal static libraries needed for downstream linking.

## Risk Assessment

Most risk is isolated to Android/Nix packaging and CI build time.

The `libp2p_ffi` start/stop changes are required for the generated FFI API to compile and for a newly created node to start correctly. Existing legacy cbind support is not extended or replaced by this PR.

## References

Depends on / should wait for:

- https://github.com/vacp2p/nim-libp2p/pull/2780

That PR flips the nim-ffi migration to make `libp2p_ffi` the active C binding path. This Android work should merge after that lands.

### DIFF
--- a/.github/workflows/mobile_android.yml
+++ b/.github/workflows/mobile_android.yml
@@ -54,8 +54,6 @@ jobs:
           test -f "$out/lib/liblibp2p.so"
           test -f "$out/lib/liblibp2p.a"
           test -f "$out/lib/libc++_shared.so"
-          test -f "$out/lib/libminiupnpc.a"
-          test -f "$out/lib/libnatpmp.a"
           test -f "$out/include/libp2p.h"
           test -f "$out/include/nim_ffi_cbor.h"
           test -f "$out/include/nim_ffi_prelude.h"
--- a/.github/workflows/mobile_android.yml
+++ b/.github/workflows/mobile_android.yml
@@ -1,0 +1,73 @@
+name: Mobile Android
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  merge_group:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  cbind-ffi-android:
+    timeout-minutes: 60
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - abi: arm64-v8a
+            attr: cbind-ffi-android-arm64-v8a
+            machine: AArch64
+          - abi: x86_64
+            attr: cbind-ffi-android-x86_64
+            machine: Advanced Micro Devices X86-64
+
+    defaults:
+      run:
+        shell: bash
+
+    name: "Android ${{ matrix.abi }} FFI check build"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          submodules: true
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+
+      - name: Build Android C ABI package
+        run: |
+          nix --extra-experimental-features "nix-command flakes" \
+            build ".#${{ matrix.attr }}" \
+            -o "result-${{ matrix.abi }}"
+
+      - name: Verify Android artifacts
+        run: |
+          out="result-${{ matrix.abi }}"
+
+          test -f "$out/lib/liblibp2p.so"
+          test -f "$out/lib/liblibp2p.a"
+          test -f "$out/lib/libc++_shared.so"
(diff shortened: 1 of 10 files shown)


vladopajic (approved):

vladopajic (approved):

.github/workflows/mobile_android.yml
```diff
@@ -0,0 +1,73 @@
+name: Mobile Android
```
gmelodie, line 1: I'd be good to have a similar thing for all nix-generated artifacts, not only mobile, maybe we can make this a bit more generic?

cbind/libp2p_ffi.nim
```diff
@@ -216,7 +190,9 @@ proc createLibp2pNode(config: Libp2pConfig): Result[LibP2P, string] =
     except LPError as e:
       return err("could not create libp2p node: " & e.msg)
 
-  let lib = LibP2P(switch: switch, rng: rng, relayClient: relayClientOpt, stopped: true)
+  let lib = LibP2P(
+    switch: switch, rng: rng, relayClient: relayClientOpt, stopped: true
```
gmelodie: Hum... this should be the default I think

gmelodie: Also I think since the original intent was not to allow stat -> stop -> start (which has been superseeded) it'd be better to call this field `running`.

cbind/libp2p_ffi.nim
```diff
@@ -210,19 +212,19 @@ proc shutdownSwitch(lib: LibP2P) {.async.} =
   lib.stopped = true
   await lib.switch.stop()
```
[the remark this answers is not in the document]
gmelodie: hummm indeed good catch

gmelodie (commented):
I would like to see the node actually running on android CI instead of just cross compiling the artifacts e.g. append sth like this to `mobile_android.yml` (claude-generated):
```yaml
        - name: Enable KVM
          if: matrix.abi == 'x86_64'
          run: |
            echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666",
  OPTIONS+="static_node=kvm"' \
              | sudo tee /etc/udev/rules.d/99-kvm4all.rules
            sudo udevadm control --reload-rules
            sudo udevadm trigger --name-match=kvm

        - name: Run FFI check on emulator
          if: matrix.abi == 'x86_64'
          uses: reactivecircus/android-emulator-runner@v2
          with:
            api-level: 23
            arch: x86_64
            target: default
            force-avd-creation: false
            emulator-options: -no-window -no-audio -no-boot-anim
  -gpu swiftshader_indirect
            disable-animations: true
            script: |
              out="result-x86_64"
              adb push "$out/lib/liblibp2p.so"
  /data/local/tmp/
              adb push "$out/lib/libc++_shared.so"
  /data/local/tmp/
              adb push "$out/bin/libp2p_ffi_android_check"
  /data/local/tmp/
              adb shell chmod 755
  /data/local/tmp/libp2p_ffi_android_check
              adb shell "LD_LIBRARY_PATH=/data/local/tmp
  /data/local/tmp/libp2p_ffi_android_check; echo RC=\$?" | tee
  run.log
              grep -q '^RC=0' run.log
```

And then maybe have the check start/stop/start/stop/destroy a node.

vladopajic:
>  node actually running on android CI

not sure how long it takes to bootstrap android emulator, but if it takes time it better be daily test.

.github/workflows/mobile_android.yml
richard-ramos: Will try to do this as part of a separate PR, since it seems to have a larger scope than just mobile

gmelodie (approved):

PR #2821, opened 2026-07-13
chore(nat): replace nim-nat-traversal with nim-libplum

gmelodie:
## Summary

Replaces the NAT port-mapping backend from [nim-nat-traversal](https://github.com/status-im/nim-nat-traversal) (separate miniupnpc + libnatpmp C wrappers) with [nim-libplum](https://github.com/logos-storage/nim-libplum), a Nim binding for [libplum](https://github.com/paullouisageneau/libplum).

libplum is a single library that owns its own internal thread, discovers the gateway, auto-selects a working protocol (PCP → NAT-PMP → UPnP-IGD), and keeps mappings refreshed on its own. This collapses the two hand-rolled worker-thread mappers into one, removes our per-service refresh loop and lease bookkeeping, and lets the external IP arrive together with each mapping instead of via a separate discovery round-trip.

It also simplifies the build: libplum's vendored C compiles into libp2p via nim's `{.compile.}` pragmas (same pattern nim-lsquic already uses), so the `nat_libs` Makefile machinery, the per-toolchain `.a` rebuilds, and the writable-copy dance in the nix cbind derivations all go away.

## Affected Areas

  NAT port mapping: `natservice.nim`, `nat/portmapper.nim`, new `nat/plum_mapper.nim`, removed `nat/upnp_mapper.nim` + `nat/natpmp_mapper.nim`.

- [x] Build / Tooling
  Dependency swap in `libp2p.nimble` / `.pinned` / `nix/deps.nix`; removed `nat_libs` from the `Makefile`, the nix cbind/cbind-ffi derivations, `cbind.nimble`, `flake.nix`, and the CI workflows.

This changes the public NAT config surface (see below), so downstream callers that construct `upnpConfig`/`natPmpConfig` with `refreshInterval`/`leaseDuration` need updating. Downstream integration branches are not yet linked.

- **Nimbus:** N/A — pending
- **Waku:** N/A — pending
- **Codex:** N/A — pending

## Impact on Library Users

- New `natConfig()` (libplum tries every protocol, recommended default) added alongside the existing `upnpConfig()` / `natPmpConfig()`; `PortMappingMode` gains an `Auto` case.
- `upnpConfig()` / `natPmpConfig()` signatures changed: `refreshInterval` and `leaseDuration` are gone (libplum handles refresh internally); they now take `discoveryTimeout` and `mappingTimeout`.
- Exported consts `DefaultRefreshInterval` and `DefaultLeaseDuration` removed; `DefaultMappingTimeout` added.
- The `PortMapper` base interface changed for custom-factory users: `discover` is gone, and `map` no longer takes a `lease` and now returns the external IP and port together as `MappedPort`.
- No change to `explicitIpConfig`, `autonatConfig`, or `holePunchingConfig`.

## Risk Assessment

- `plum_init`/`plum_cleanup` is a process-global singleton (a second init fails while a client is live). `PlumMapper` ref-counts init across live mappers; if two mappers request different protocol filters, the first one wins and a warning is logged. The ref-count and active-filter globals are unsynchronized, so every `PlumMapper` must be created and closed from the same chronos event-loop thread.
- Behavior change: no per-service refresh loop and no re-announce on a timer — libplum's internal recheck keeps mappings alive, and the external IP is taken from the first successful mapping rather than a standalone discovery step.
- IPv4-only mapping is preserved (non-IPv4 listen addrs are still filtered before mapping).
- Build: libplum's vendored C compiles into libp2p via nim's `{.compile.}`, and the nix path uses `fetchSubmodules = true`. `libp2p.nimble` pins libplum by exact commit rather than `>= 0.6.0` on purpose: the `v0.6.0` tag predates the `installFiles` entry for `libplum_units.c`, so version-range resolution (e.g. `nimble setup`) fetches a package that fails to compile with `cannot find: libplum_units.c`. Pinning the requires forces every resolution path onto the fixed commit.

## References

- nim-libplum: https://github.com/logos-storage/nim-libplum (pinned `acefbe424`, branch `fix/missing-install-file` — not yet a tagged release)
- libplum: https://github.com/paullouisageneau/libplum

## Additional Notes

The new `test_plum_mapper.nim` covers lifecycle and error paths (create/close, double-close, map-after-close, unmap-without-map, ref-counted shared instance) without requiring a real gateway, mirroring the old mapper tests. `test_natservice.nim` was updated to the new interface (a single recording mock returns `MappedPort`; the refresh-loop and lease-validation tests are dropped).


vladopajic (approved):

libp2p/services/natservice.nim
```diff
@@ -288,34 +278,23 @@ proc findMappableListenPorts(listenAddrs: seq[MultiAddress]): seq[ListenPort] =
...
+  tuple[entry: (Port, MapProto), externalIp: IpAddress, announced: Opt[MultiAddress]]
 
 proc mapOnePort(
-    self: NATService, lp: ListenPort, externalIp: IpAddress
+    self: NATService, lp: ListenPort
 ): Future[Opt[MappedEntry]] {.async: (raises: [CancelledError]).} =
-  let lease = uint32(self.portMapping().leaseDuration.seconds)
-  let extPort = (await self.mapper.map(lp.port, lp.port, lp.proto, lease)).valueOr:
+  # libplum returns the external address with the mapping; no separate discovery.
+  let mapped = (await self.mapper.map(lp.port, lp.port, lp.proto)).valueOr:
```
richard-ramos, line 288: Some interesting comment from codex. Not sure if true:

[libp2p/services/natservice.nim:288] setupMappings still calls self.mapper.map(...) every time the address mapper runs. With libplum, each call creates a persistent mapping id, but [libp2p/services/nat/plum_mapper.nim:109] stores ids by (externalPort, proto), so a repeated update that returns the same external port overwrites the previous id. unmapStale then keeps the entry because the port/proto did not change, leaving the old mapping active and no longer individually destroyable. Reuse existing mappings for unchanged lis

libp2p/services/nat/plum_mapper.nim
```diff
@@ -0,0 +1,133 @@
...
+  DefaultMappingTimeout = 10.seconds
+
+type
+  MappingKey = tuple[externalPort: uint16, proto: MapProto]
+
+  PlumMapper* = ref object of PortMapper
+    filter: ProtocolFilter
+    closed: bool
+    mappings: Table[MappingKey, cint] ## -> libplum mapping id, to destroy later
+
```
richard-ramos, line 41: similar to upnp_mapper, you'll likely need a lock, but to protect `mappings` in `map`, `unmap` and `close`.

richard-ramos (commented):
- In `cbind/cbind.cnimble` there are still a call to do a `make -C .. nat_libs`

libp2p/services/nat/plum_mapper.nim
```diff
@@ -0,0 +1,133 @@
...
+  ?acquirePlum(filter, discoverTimeout, mappingTimeout)
+  ok(PlumMapper(filter: filter))
+
+method map*(
+    self: PlumMapper, internalPort: Port, externalPort: Port, proto: MapProto
+): Future[Result[MappedPort, string]] {.async: (raises: [CancelledError]), gcsafe.} =
+  if self.closed:
+    return err("PlumMapper closed")
+
+  let res = ?await createMapping(
```
richard-ramos, line 131: Should the `mappingTimeout` be passed here too?

libp2p/services/natservice.nim
```diff
@@ -517,12 +481,11 @@ proc startPortMapping(self: NATService, switch: Switch) =
     case pm.mode
     of ExplicitIp:
       self.addressMapper = explicitIpMapper(pm.explicitIp)
-    of Upnp, NatPmp:
+    of Upnp, NatPmp, Auto:
       self.mapper = self.buildPortMapper(pm.mode).valueOr:
         warn "Could not build port mapper; port mapping inactive", mode = pm.mode
         return
       self.addressMapper = self.portMappingMapper()
-      self.refreshLoopFut = self.refreshLoop(switch)
```
richard-ramos, line 525: Curious about why was this refresh loop removed?

libp2p/services/natservice.nim
gmelodie: makes sense!

libp2p/services/natservice.nim
gmelodie: Ah! because now it's libplum that handles the renewal of leases

richard-ramos (approved):

PR #2825, opened 2026-07-14
fix(service-disco): ticket time window

SionoiS:
The advertiser was sleeping a relative duration after Wait, capped by advertExpiry:

let waitSecs = min(disco.discoConfig.advertExpiry, newTicket.tWaitFor.get())

Under load, waitingTime often yields tWaitFor > advertExpiry.


richard-ramos (approved):

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -281,11 +293,12 @@ proc advertiseToRegistrar*(
 
       currentTicket = Opt.some(newTicket)
 
-      let waitSecs = min(disco.discoConfig.advertExpiry, newTicket.tWaitFor.get())
+      let waitSecs = ticketRetryDelay(newTicket, Moment.now())
 
       debug "waiting for registrar", serviceId, registrar, wait = $waitSecs
```
vladopajic: seems like this debug should be in `if` statement bellow, as there is no waiting otherwise?

vladopajic (approved):

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -226,6 +226,18 @@ proc sendRegister*(
     )
   )
 
+proc ticketRetryDelay*(ticket: Ticket, now: Moment): Duration {.raises: [].} =
+  ## Time to sleep before presenting ``ticket`` so the registrar sees it inside
+  ## the eligibility window ``[tMod + tWaitFor, tMod + tWaitFor + Δ]``.
```
gmelodie:
```suggestion
  ## the eligibility window ``[tMod + tWaitFor, tMod + tWaitFor + delta]``.
```

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -226,6 +226,18 @@ proc sendRegister*(
...
+  ##
+  ## Uses absolute ticket timestamps (not a relative wait from response receipt,
+  ## and not capped by advert expiry) so retries under load are not rejected as
```
gmelodie:
```suggestion
  ## Uses absolute ticket timestamps so retries under load are not rejected as
```

gmelodie (approved):

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -488,10 +488,13 @@ proc registration*(disco: ServiceDiscovery, peerId: PeerId, inMsg: Message): Mes
...
+  # Floor to whole seconds so in-memory ticket times match wire encode/decode
+  # and the signature payload (epochSeconds / Duration.seconds).
+  let nowSec = Moment.init(now.epochSeconds, Second)
   var ticket = Ticket(
     advertisement: regMsg.advertisement,
-    tInit: Opt.some(now),
-    tMod: Opt.some(now),
+    tInit: Opt.some(nowSec),
+    tMod: Opt.some(nowSec),
     tWaitFor: Opt.some(tWait),
```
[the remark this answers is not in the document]
SionoiS: This value is already rounded in `waitingTime()`

libp2p/protocols/service_discovery/advertiser.nim
SionoiS: I just removed that part it's way too verbose anyway.

libp2p/protocols/service_discovery/registrar.nim
jm-clius, line 1: As far as I can tell, the issue lies in this file - [the spec states](https://lip.logos.co/anoncomms/raw/logos-service-discovery.html#handling-register-requests) that the waiting time issued by the Registrar should be capped at `advertExpiry` and afaics this is not implemented.

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -226,6 +226,22 @@ proc sendRegister*(
...
+
+  let tWaitFor = ticket.tWaitFor.valueOr:
+    return ZeroDuration
+
+  let eligibility = tMod + tWaitFor
+
+  if now >= eligibility:
+    return ZeroDuration
+
+  return eligibility - now
```
jm-clius, line 243: I don't think this can work - `Moment` has no absolute meaning outside of a single process clock. It cannot be used to compare time across different clock domains. The tests here will pass, because they run on the same clock. Furthermore, [the spec](https://lip.logos.co/anoncomms/raw/logos-service-discovery.html#registrar-algorithms) specifically states that "clock synchronization" should not be necessary because we only deal with relative wait times. If I understand the issue correctly, we see waiting times issued that are larger than the advert expiry time under load? If so, the fix  should probably be registrar-side - capping issued waiting times there? This is what the spec mandates too, but afaics has not been implemented.

jm-clius (changes requested):
Afaics, the issue is that [the spec-mandated capping of waiting time](https://lip.logos.co/anoncomms/raw/logos-service-discovery.html#handling-register-requests) to the advert expiry time on the _Registrar_ side has not been implemented. We do not want to introduce clock synchronisation requirements.

SionoiS:
> Afaics, the issue is that [the spec-mandated capping of waiting time](https://lip.logos.co/anoncomms/raw/logos-service-discovery.html#handling-register-requests) to the advert expiry time on the _Registrar_ side has not been implemented. We do not want to introduce clock synchronisation requirements.

Yes that make sense in this context, I understand better now.

re-reading the spec I think waiting time can be zero. I'll look into it and fix that too.

PR #2827, opened 2026-07-14
feat(android): simulator

richard-ramos:
## Summary

Adds runtime coverage to the Android mobile CI by running the x86_64 `libp2p_ffi_android_check` binary inside an Android emulator.

The workflow now:
- Enables KVM for the x86_64 Android job
- Boots an API 23 x86_64 emulator
- Pushes `liblibp2p.so`, `libc++_shared.so`, and the check binary to the device
- Runs the check with `LD_LIBRARY_PATH` set and fails CI unless it exits with `RC=0`

## Affected Areas

- [x] Build / Tooling  
  Android GitHub Actions workflow for the mobile FFI build.

## Impact on Library Users

No API or runtime behavior changes for library users. This only expands CI validation for Android artifacts.

## Risk Assessment

Low risk. The new emulator step is limited to the x86_64 Android CI matrix entry; arm64 continues to build and verify artifacts without emulator execution.

### DIFF
--- a/.github/workflows/mobile_android.yml
+++ b/.github/workflows/mobile_android.yml
@@ -54,8 +54,6 @@ jobs:
           test -f "$out/lib/liblibp2p.so"
           test -f "$out/lib/liblibp2p.a"
           test -f "$out/lib/libc++_shared.so"
-          test -f "$out/lib/libminiupnpc.a"
-          test -f "$out/lib/libnatpmp.a"
           test -f "$out/include/libp2p.h"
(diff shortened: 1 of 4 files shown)


gmelodie (approved):
a mazing

vladopajic (approved):

gmelodie (approved):

arnetheduck:
hm, this looks interesting - could you backport something like this to nim-chronos so that its test suite run on android as well?

PR #2828, opened 2026-07-14
fix(dialer): reset streams after failed protocol negotiation

### LINKED ISSUE #2800: identify: outgoing identify hangs against peers that don't serve identify

richard-ramos:
## Summary

Fixes hangs after failed multistream protocol negotiation when the remote peer leaves the stream open after responding `na`.

Failed identify probes, identify push attempts, and general dial negotiation failures now reset the stream instead of closing with EOF. Successful identify and identify-push paths still close with EOF after completing the protocol.

Adds regression coverage for outbound identify against a peer that rejects identify, plus general failed protocol negotiation paths.

## Impact on Library Users

Prevents outbound identify and protocol dials from hanging against peers that do not support the requested protocol. No API changes.

## Risk Assessment

Low. The change is limited to failure-path stream cleanup; successful negotiation paths preserve the existing EOF close behavior.

## References

Closes #2800

### DIFF
--- a/libp2p/dialer.nim
+++ b/libp2p/dialer.nim
@@ -315,7 +315,7 @@ proc negotiateStream*(
   trace "Negotiating stream", stream, protos
   let selected = await MultistreamSelect.select(stream, protos)
   if not protos.contains(selected):
-    await stream.closeWithEOF()
+    await stream.reset()
     raise newException(
       DialFailedError,
       "Unable to select sub-protocol. Selected: " & $selected & ". Available: " & $protos,
@@ -323,7 +323,7 @@ proc negotiateStream*(
 
   self.ms.lookupProtocol(selected).withValue(protocol):
     if not protocol.reserveOutgoing(stream.peerId):
-      await stream.closeWithEOF()
+      await stream.reset()
       raise newException(
         DialFailedError, "Outbound stream budget exceeded for protocol: " & selected
       )
--- a/libp2p/peerstore.nim
+++ b/libp2p/peerstore.nim
@@ -546,7 +546,7 @@ proc identify*(
   if stream == nil:
     return
 
-  var cancelled = false
+  var identifyCompleted = false
   try:
     if (await MultistreamSelect.select(stream, peerStore.identify.codec())):
       let info = await peerStore.identify.identify(stream, stream.peerId)
@@ -561,14 +561,14 @@ proc identify*(
         muxer.setShortAgent(knownAgent)
 
       peerStore.updatePeerInfo(info, stream.observedAddr, Opt.some(dir))
+      identifyCompleted = true
   except CancelledError as exc:
-    cancelled = true
     raise exc
   finally:
-    if cancelled:
-      await noCancel stream.reset()
-    else:
+    if identifyCompleted:
       await noCancel stream.closeWithEOF()
+    else:
+      await noCancel stream.reset()
 
 proc getMostObservedProtosAndPorts*(self: PeerStore): seq[MultiAddress] =
   return self.identify.observedAddrManager.getMostObservedProtosAndPorts()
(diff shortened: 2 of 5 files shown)


gmelodie (approved):
nice fix!

vladopajic (approved):

Issue #2826, opened 2026-07-14
kad-dht: closerPeers advertise a peer's observed connection address


rlve:
### Description

The unified-testing kad-dht suite runs a bootstrap, a provider, and a querier across implementations. Every py and dotnet combination passes. `nim_x_nim_x_py` (nim bootstrap, nim provider, py querier) is flaky. The py querier reaches the nim bootstrap and runs its provider lookup, but every address it is then handed to dial points at an ephemeral port nothing listens on: the provider's inbound source address, and (via `closerPeers`) an address that resolves back to the querier's own host. The dial fails, and the lookup finishes with no reachable provider.

The difference is where each implementation gets the addresses it puts in `closerPeers`.

go, py, and dotnet carry each peer's listen addresses in the DHT messages. py attaches a signed peer record to every request and response and consumes it into the peerstore on receipt, then reads closer-peer addresses from the peerstore; go reads closer-peer addresses from the peerstore as well. So the moment a provider sends `ADD_PROVIDER` or `FIND_NODE`, the receiver already has the provider's listen address, and every `closerPeers` reply it later sends carries diallable listen addresses. There is no dependency on Identify, and no connection address is ever advertised.

nim does neither. Its kad messages carry no signed peer record. `closerPeers` addresses come from the `AddressBook`, and three call sites populate it with the inbound stream's `observedAddr`. `observedAddr` is the remote end of the connection, the address the peer dialed from. A peer that dialed in used an ephemeral source port, not the port it listens on, so no one else can reach it there. The listen address reaches the AddressBook only later, through Identify. Even the listen address that arrives inside the `ADD_PROVIDER` message is filed only under provider records, not the AddressBook that feeds `closerPeers`. So a peer that walks `closerPeers` to route toward the key is handed the provider's ephemeral source address rather than a diallable listen address; before Identify supplies the listen address, that ephemeral address is the only one there, and nothing accepts connections on it.

`observedAddr` is stored at `AddressConfidence.Low`; the listen address Identify later supplies is stored at `AddressConfidence.Medium`. Once the listen address is present, a caller that tries every advertised address can still connect, which is why a nim querier and the pure-nim tests do not notice. But once recorded, the observed address stays in `closerPeers`, and before Identify it is the only address there.

[...]

PR #2832, opened 2026-07-15
test(interop): kad-dht for unified-testing

rlve:
## Summary

Adds a kad-dht interop peer for the libp2p unified-testing harness. One binary, `interop/kad-dht-v2/peer.nim`, implements the three roles the harness drives through the `ROLE` env var: bootstrap, provider, and querier. Peers coordinate through Redis over a fixed tcp/noise/yamux stack.

The peer exercises the `KadDHT` API end to end: the provider runs `addProvider` and `putValue`, the querier runs `getProviders` and `getValue` and prints `status: pass` once it finds both. This lets nim-libp2p take part in the cross-implementation kad-dht test matrix alongside the `py` and `dotnet` peers.

Tested with https://github.com/libp2p/unified-testing/pull/119

Current results: 
```
Running test: nim_x_nim_x_py
  [FAILED]

Tests complete. Collecting results...
Ran: 27, Skipped: 0, Passed: 26, Failed: 1
```

One outstanding issue tracked in #2826


interop/kad-dht-v2/peer.nim
```diff
@@ -0,0 +1,201 @@
...
+  switch.mount(kad)
+  await switch.start()
+  defer:
+    await switch.stop()
+
+  let myAddr = switch.peerInfo.fullAddrs.tryGet()[0]
+  redisClient.publishValue(config.testKey, BootstrapAddrSuffix, $myAddr)
+  info "Bootstrap published multiaddr", address = $myAddr
+
+  await sleepAsync(1.hours) # harness aborts the stack once the querier exits
```
vladopajic: might be better to:
- `switch.start()` /  `switch.stop()` in main, since it creates it it owns it
-  `await sleepAsync(1.hours)` do this in main as well
   - alternatively do something like `await waitSignal(SIGINT)` instead of sleep

vladopajic (approved):

interop/kad-dht-v2/peer.nim
```diff
@@ -0,0 +1,201 @@
...
+) {.async.} =
+  ## Fetch the bootstrap multiaddr from Redis, dial it, and seed the routing table.
+  let raw = await redisClient.fetchValue(testKey, BootstrapAddrSuffix, RedisWaitTimeout)
+  let bootstrapMA = MultiAddress.init(raw).tryGet()
+  let bootstrapPeerId = await switch.connect(bootstrapMA)
+  kad.updatePeers(@[(bootstrapPeerId, @[bootstrapMA])])
+  info "Connected to bootstrap", peerId = bootstrapPeerId
+
+proc runBootstrap(config: KadConfig, switch: Switch, redisClient: Redis) {.async.} =
+  let kad = newKad(switch)
```
vladopajic: all cases are creating `kad` maybe have that in main as well. the role cases should only run specific logic for already configured switch+kad

interop/kad-dht-v2/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# syntax=docker/dockerfile:1.5-labs
```
vladopajic, line 1: is there a way to run unified-testing in `nim-libp2p` ci?

currently there is growing source code related to unified-testing here. and there is no assertions of that code, not even if they compile. nim-libp2p may introduce braking change and since CI is not even checking if `kad-dht-v2/peer.nim` compiles. it could easily stop to compile.

therefore:
- it would be niece to run unified testing in ci (at least daily)
- if above is not possible then we could at least only build these binaries that are used for unified-testing?

interop/kad-dht-v2/peer.nim
```diff
@@ -0,0 +1,201 @@
...
+    await switch.stop()
+
+  await connectBootstrap(switch, kad, redisClient, config.testKey)
+
+  await kad.addProvider(provideKey(config.testKey))
+  info "Provider announced key"
+
+  let putRes = await kad.putValue(valueKey(config.testKey), ProviderValue.toBytes())
+  if putRes.isErr:
+    raise newException(CatchableError, "putValue failed: " & putRes.error)
```
gmelodie: nit: maybe...
```suggestion
    raise newException(ValueError, "putValue failed: " & putRes.error)
```

interop/kad-dht-v2/peer.nim
```diff
@@ -0,0 +1,201 @@
...
+  await switch.start()
+  defer:
+    await switch.stop()
+
+  # Wait until the provider is done before connecting.
+  discard
+    await redisClient.fetchValue(config.testKey, ProviderDoneSuffix, RedisWaitTimeout)
+  await connectBootstrap(switch, kad, redisClient, config.testKey)
+
+  # Test 2: find the provider record.
```
gmelodie: What is `Test 2`, `Test 4`? Is this a unified-interop thing? If so, will both tests always run or it's either one or the other? Could we encapsulate them in functions for readability?

gmelodie (approved):
Only nits. Thank you for doing this! Amazing work :)

interop/kad-dht-v2/Dockerfile
rlve: It's not possible currently to run unified testing suites in `nim-libp2p`'s CI.

But a workflow building these dockerfiles is a good idea, will create a PR!

PR #2834, opened 2026-07-15
fix(service-disco): wait time cap

SionoiS:
A much better version of #2825 

Thanks @jm-clius


gmelodie (approved):

richard-ramos (approved):

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -458,7 +459,8 @@ proc registration*(disco: ServiceDiscovery, peerId: PeerId, inMsg: Message): Mes
 
     return msg
 
-  let now = Moment.now()
+  #Always use seconds granularity
```
tinniaru3005, line 461: `pruneExpiredAds` still uses raw `Moment.now()`. This comment reads like a project-wide rule, but it's actually only true within registration(). Suggest tightening the comment so it doesn't get treated as a broader invariant.

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -185,6 +184,8 @@ proc waitingTime*(
       if waitDuration < prevWaitDuration - elapsedDuration:
         waitDuration = prevWaitDuration - elapsedDuration
 
+  waitDuration = min(discoConfig.advertExpiry, waitDuration)
+
```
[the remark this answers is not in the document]
tinniaru3005: Good catch. This is a genuine gap.

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -185,6 +184,8 @@ proc waitingTime*(
       if waitDuration < prevWaitDuration - elapsedDuration:
         waitDuration = prevWaitDuration - elapsedDuration
 
+  waitDuration = min(discoConfig.advertExpiry, waitDuration)
```
tinniaru3005: The `advertExpiry` cap runs here, before `updateWaitAfterRetry()` subtracts elapsed retry time in `registration()` (line 478). That inverts the intended order — the cap should apply to `t_remaining` (after subtracting elapsed time), i.e. `t_wait_for = min(E, t_wait - elapsed)`, not to the raw formula output before subtraction.

Suggest moving this clamp to `registration()`, applied to `tWait` after the `updateWaitAfterRetry()` call, not inside `waitingTime()`.

libp2p/protocols/service_discovery/registrar.nim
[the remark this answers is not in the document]
SionoiS: The type itself prevent negative value. It cannot happen AFAIK.

libp2p/protocols/service_discovery/registrar.nim
SionoiS: good catch I miss that one.

PR #2835, opened 2026-07-15
fix(daily): nimbus

richard-ramos:


richard-ramos:
FYI @gmelodie  @vladopajic 
Fixing this will require nimbus to update to 2.3.0 (once available), because there are some breaking changes (some in gossip that require using Opt instead of direct value due to protobuffers being changed, and also libplum)

PR #2837, opened 2026-07-15
fix(quic): propagate accept errors instead of nil

### LINKED ISSUE #2713: transports: inconsistency when stopping pending accept

richard-ramos:
## Summary

Prevent QUIC accept failures from producing nil connections that can later cause a segmentation fault.

QUIC now:

- marks the transport stopped before closing endpoints, matching TCP shutdown ordering
- raises typed transport errors instead of swallowing accept failures
- rejects nil connections when constructing a QUIC muxer
- reports `TransportClosedError` when shutdown interrupts a pending accept

## Affected Areas

- [x] Transports  
  QUIC accept and shutdown behavior.

## Impact on Library Users

Pending QUIC accepts now raise `TransportClosedError` during shutdown instead of resolving to nil. Other QUIC accept failures are also propagated to callers with their original error as the parent.

## Risk Assessment

Callers that explicitly relied on a nil result from a QUIC accept interrupted by shutdown must handle `TransportClosedError` instead. This aligns QUIC with the existing TCP behavior.

## References

- [Failing daily CI run](https://github.com/vacp2p/nim-libp2p/actions/runs/29333063999)

### DIFF
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -532,13 +532,13 @@ method accept*(
     let conn = await finished
     return self.wrapConnection(conn, Direction.In)
   except QuicError as exc:
-    warn "QUIC accept failed", description = exc.msg
+    debug "Quic Error", description = exc.msg
     raise (ref QuicTransportError)(msg: "QUIC accept failed: " & exc.msg, parent: exc)
   except common.TransportError as exc:
-    warn "QUIC transport accept failed", description = exc.msg
+    debug "Transport Error", description = exc.msg
     raise newTransportClosedError(exc)
   except TransportOsError as exc:
-    warn "QUIC OS accept failed", description = exc.msg
+    debug "OS Error", description = exc.msg
     raise (ref QuicTransportError)(msg: "QUIC OS accept failed: " & exc.msg, parent: exc)
 
 proc listenerEndpointFor(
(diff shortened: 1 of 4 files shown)


libp2p/transports/quictransport.nim
```diff
@@ -211,6 +211,9 @@ type QuicMuxer* = ref object of Muxer
 proc new*(
     _: type QuicMuxer, conn: P2PConnection, peerId: Opt[PeerId] = Opt.none(PeerId)
 ): QuicMuxer {.raises: [CertificateParsingError, LPError].} =
+  if conn.isNil:
+    raise (ref QuicTransportError)(msg: "cannot create QUIC muxer from nil connection")
```
vladopajic, line 215:
```
raise newException(QuicTransportError, "cannot create QUIC muxer from nil connection")
```

vladopajic (approved):

rlve (approved):

gmelodie (approved):

PR #2838, opened 2026-07-15
fix(dcutr): preserve QUIC server role during hole punching

### LINKED ISSUE #2806: dcutr: quic hole punching dials from both peers
### Description

QUIC hole punching fails against a spec-compliant peer because nim dials the direct connection from both DCUtR roles. For QUIC that means both peers emit a client Initial, and neither acts as the server, so the handshake collides.

The DCUtR spec makes QUIC hole punching asymmetric. Only the `Sync` receiver (`A`) dials. The `Sync` sender (`B`) must not dial. It sends random-byte UDP packets and becomes the QUIC server. nim treats every hole-punchable address the same and dials from both roles, which is right for TCP simultaneous open but wrong for QUIC.

`B` is the DCUtR initiator in `client.nim`. It sends `Sync`, waits half the RTT, then dials every dialable address, QUIC included:

https://github.com/vacp2p/nim-libp2p/blob/c69bec5407f07d20cf33bd5037ad598267bc859b/libp2p/protocols/connectivity/dcutr/client.nim#L69-L85

Note `B` already labels itself the server here: it dials with `dir = In` and registers an expected incoming connection. For TCP that relabeling is harmless. For QUIC the client and server roles are real on the wire, so `B` emitting a client Initial defeats the very role it just claimed.

`A` is the responder in `server.nim`. It dials on receiving `Sync`, which is the correct role for both TCP and QUIC:

https://github.com/vacp2p/nim-libp2p/blob/c69bec5407f07d20cf33bd5037ad598267bc859b/libp2p/protocols/connectivity/dcutr/server.nim#L81-L97

Both paths take addresses from `getHolePunchableAddrs`, which returns TCP and QUIC_V1 addresses mixed together with no transport or role distinction:

https://github.com/vacp2p/nim-libp2p/blob/c69bec5407f07d20cf33bd5037ad598267bc859b/libp2p/protocols/connectivity/dcutr/core.nim#L80-L92

This is not a NAT-mapping reuse problem. The QUIC transport already reuses the listener endpoint for dialing, so a punched connection leaves from the listener UDP port:

https://github.com/vacp2p/nim-libp2p/blob/c69bec5407f07d20cf33bd5037ad598267bc859b/libp2p/transports/quictransport.nim#L565-L572

### Spec Recommendation

[DCUtR Simultaneous Connect, step 5](https://github.com/libp2p/specs/blob/cfcf0230b2f5f11ed6dd060f97305faa973abed2/relay/DCUtR.md?plain=1#L87-L101):

> For a QUIC address:
> - Upon receiving the `Sync`, `A` immediately dials the address to `B`.
> - Upon expiry of the timer, `B` starts to send UDP packets filled with random
>   bytes to `A`'s address. Packets should be sent repeatedly in random intervals
>   between 10 and 200 ms.
> - This will result in a QUIC connection where `A` is the client and `B` is the
>   server.

`A` receives `Sync` and maps to `server.nim`. `B` sends `Sync` and maps to `client.nim`. For a QUIC address, `B` diverges from the spec: it dials instead of sending random-byte UDP packets.

### Reproduction

unified-testing hole-punch suite, `rust-v0.56 x nim-v2.2 (quic-v1)`, rust relay, linux router - https://github.com/libp2p/unified-testing/pull/78

- rust dials the direct connection as `role_override: Dialer, port_use: Reuse`. It is spec `A`.
- nim is the listener, so it initiates DCUtR and is spec `B`. It dials too.
- rust closes the direct connection: `Connection closed with <nim peerid>:
  Some(IO(Custom { kind: Other, error: Error(Right(Decode(Io(Custom { kind: Other, error: Failed })))) }))`.

TCP hole punching passes in both directions. QUIC over the relay works. Only the direct hole-punched QUIC connection fails.

richard-ramos:
## Summary

Fix QUIC DCUtR hole punching so the Sync sender retains the listener/server role required by the specification.

Listener-role QUIC dial attempts now send random UDP packets from the matching listener socket and wait for the inbound QUIC connection, rather than initiating a QUIC client connection. This mirrors go-libp2p’s transport-level handling and prevents both peers from acting as QUIC clients during hole punching.

## Affected Areas

- [x] Transports — propagate connection direction to transports and add QUIC listener-role hole punching
- [x] Protocol Logic — preserve the asymmetric QUIC roles required by DCUtR

## Risk Assessment

The behavioral change is limited to QUIC dial attempts marked `Direction.In`; normal outbound QUIC dialing remains unchanged. Selecting a listener requires exactly one endpoint matching the target address family.

## References

- Resolves [[#2806](https://github.com/vacp2p/nim-libp2p/issues/2806)](https://github.com/vacp2p/nim-libp2p/issues/2806)
- [DCUtR simultaneous connect specification](https://github.com/libp2p/specs/blob/master/relay/DCUtR.md#simultaneous-connect)

### DIFF
--- a/libp2p/crypto/rng.nim
+++ b/libp2p/crypto/rng.nim
@@ -113,6 +113,19 @@ proc randBelow(rng: Rng, max: uint32): int =
     if r >= threshold:
       return (r mod max).int
 
+proc rand*(rng: Rng, low, high: int): int =
+  ## Return a uniformly random integer in the inclusive range [`low`, `high`].
+  doAssert low <= high, "random range must not be empty"
+
+  let width =
+    if low >= 0 or high < 0:
+      uint64(high - low) + 1
+    else:
+      uint64(-(low + 1)) + uint64(high) + 2
+  doAssert width <= uint64(uint32.high), "random range is too large"
+
+  low + rng.randBelow(width.uint32)
+
 proc pick*[T](rng: Rng, x: openArray[T], n: int): Opt[seq[T]] =
   doAssert n >= 0, "n must be non-negative"
   if x.len == 0:
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -26,6 +26,8 @@ export transport
 logScope:
   topics = "libp2p quictransport"
 
+const QuicHolePunchPacketSize* = 64
+
 type
   P2PConnection = connection.Connection
   QuicConnection = lsquic.Connection
@@ -608,7 +610,7 @@ method dial*(
       # UDP packets from the listener socket and let the expected inbound
       # connection complete the DCUtR attempt.
       while true:
-        let payload = self.rng.generateBytes(64)
+        let payload = self.rng.generateBytes(QuicHolePunchPacketSize)
         await endpoint.get().datagramTransport().sendTo(taAddress, payload)
         let delay = self.rng.rand(10, 200)
         await sleepAsync(delay.milliseconds)
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -5,6 +5,7 @@ import std/[hashes, sets, sequtils]
 import chronos, chronicles, metrics, results
 import lsquic
 import
+  ../crypto/rng,
   ../wire,
   ../connmanager,
   ../multiaddress,
@@ -350,11 +351,15 @@ proc new*(
     _: type QuicTransport,
     u: Upgrade,
     privateKey: PrivateKey,
+    rng: Rng,
     connManager: ConnManager = nil,
 ): QuicTransport =
+  doAssert not rng.isNil, "Rng is nil"
+
   let self = QuicTransport(
     upgrader: QuicUpgrade(ms: u.ms, connManager: connManager.toOpt()),
     privateKey: privateKey,
+    rng: rng,
     certGenerator: defaultCertGenerator,
   )
   procCall Transport(self).initialize()
@@ -364,12 +369,16 @@ proc new*(
     _: type QuicTransport,
     u: Upgrade,
     privateKey: PrivateKey,
+    rng: Rng,
     certGenerator: CertGenerator,
     connManager: ConnManager = nil,
 ): QuicTransport =
+  doAssert not rng.isNil, "Rng is nil"
+
   let self = QuicTransport(
     upgrader: QuicUpgrade(ms: u.ms, connManager: connManager.toOpt()),
     privateKey: privateKey,
+    rng: rng,
     certGenerator: certGenerator,
   )
   procCall Transport(self).initialize()
@@ -576,6 +585,7 @@ method dial*(
     hostname: string,
     address: MultiAddress,
     peerId: Opt[PeerId] = Opt.none(PeerId),
+    dir: Direction = Direction.Out,
 ): Future[RawConn] {.async: (raises: [transport.TransportError, CancelledError]).} =
   let taAddress =
     try:
@@ -586,6 +596,23 @@ method dial*(
       )
 
   try:
+    if dir == Direction.In:
+      let endpoint = self.listenerEndpointFor(taAddress)
+      if endpoint.isNone():
+        raise newException(
+          QuicTransportDialError,
+          "error in QUIC hole punch: no unique listener for address family",
+        )
+
+      # The Sync sender is the QUIC server. Open its NAT mapping with random
+      # UDP packets from the listener socket and let the expected inbound
+      # connection complete the DCUtR attempt.
+      while true:
+        let payload = self.rng.generateBytes(64)
+        await endpoint.get().datagramTransport().sendTo(taAddress, payload)
+        let delay = self.rng.rand(10, 200)
+        await sleepAsync(delay.milliseconds)
+
     let endpoint = self.dialEndpointFor(taAddress)
     let quicConnection = await endpoint.dial(taAddress)
     peerId.withValue(expectedPeerId):
@@ -606,6 +633,8 @@ method dial*(
     )
   except TransportOsError as e:
     raise newException(QuicTransportDialError, "error in quic dial:" & e.msg, e)
+  except chronos.TransportError as e:
+    raise newException(QuicTransportDialError, "error in quic dial: " & e.msg, e)
   except DialError as e:
     raise newException(QuicTransportDialError, "error in quic dial:" & e.msg, e)
   except QuicError as e:
--- a/tests/libp2p/transports/test_quic.nim
+++ b/tests/libp2p/transports/test_quic.nim
@@ -9,6 +9,7 @@ import
     transports/transport,
     transports/quictransport,
     upgrademngrs/upgrade,
+    utils/future,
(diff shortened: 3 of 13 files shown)


tests/libp2p/transports/test_quic.nim
```diff
@@ -55,6 +56,36 @@ suite "Quic transport":
...
+      await puncher.stop()
+
+    let remoteAddr = MultiAddress
+      .init("/ip4/127.0.0.1/udp/" & $receiver.localAddress().port & "/quic-v1")
+      .tryGet()
+    let punchFut = puncher.dial("", remoteAddr, Opt.none(PeerId), Direction.In)
+    defer:
+      await punchFut.cancelAndWait()
+
+    check (await packetReceived.wait(1.seconds)) == 64
```
vladopajic: it might be better not to do `.wait(1.seconds)`, if we ever switch to `unittest3` patterns like this can be source of flakiness - test may run and at this point other test may take over executing and when this test resumes it may be over 1.sec (say 1.2sec).

`unttest3` will have timeout for tests (default to 60s) so if test hangs due to some contract being broken then we will see - but with bigger timeout. timeout will be configurable per test, like this `asyncTest "listener-role dial sends UDP hole-punch packets", timeout = 10.seconds:`

vladopajic: `64` better come from const:

```
const dcutrMsgSize = 64
when defined(libp2p_testing):
  export dcutrMsgSize

 ...

self.rng.generateBytes(dcutrMsgSize)

... 

check (await packetReceived) == dcutrMsgSize
```

libp2p/crypto/rng.nim
```diff
@@ -113,6 +113,19 @@ proc randBelow(rng: Rng, max: uint32): int =
     if r >= threshold:
       return (r mod max).int
 
+proc rand*(rng: Rng, low, high: int): int =
```
vladopajic, line 116: nit: `proc int*` since it returns `int`. 
`rand.int()` >> `rand.rand()`

tests/libp2p/transports/test_quic.nim
vladopajic: i understand purpose of using timeout while developing, but not sure if we should check in that code because of flakiness

libp2p/transports/quictransport.nim
```diff
@@ -586,6 +596,23 @@ method dial*(
...
+        )
+
+      # The Sync sender is the QUIC server. Open its NAT mapping with random
+      # UDP packets from the listener socket and let the expected inbound
+      # connection complete the DCUtR attempt.
+      while true:
+        let payload = self.rng.generateBytes(64)
+        await endpoint.get().datagramTransport().sendTo(taAddress, payload)
+        let delay = self.rng.rand(10, 200)
+        await sleepAsync(delay.milliseconds)
```
vladopajic, line 616: understood, but it feels strange. for transport to implement this logic.

it also needs better explanation: `while true:` needs to be explained - this `dial` future is expected to be cancelled (must be), it will never return on it's own. this is also divination from behavior of other transports, when dialing using quic server it never returns, while dialing from other transport returns?

vladopajic, line 616: does it make sense to do this:

- attempt to dial endpint, with timeout, if successful return connection
- if it gets timeout out, sent random bytes
- repeat with 1

the difference is that eventually this dial will return connection

vladopajic (approved):

rlve (approved):
I have run hole punch unified-testing suite locally with your branch and all the cases are green now ✅ 

```
## Test Results

| Test | Dialer | Listener | Transport | Status | Duration |
|------|--------|----------|-----------|--------|----------|
| nim-v2.2 x rust-v0.56 (tcp, noise, yamux) [dr - linux, rly - rust-v0.56, lr - linux] | nim-v2.2 | rust-v0.56 | tcp | ✅ | 3s |
| nim-v2.2 x rust-v0.56 (tcp, noise, mplex) [dr - linux, rly - rust-v0.56, lr - linux] | nim-v2.2 | rust-v0.56 | tcp | ✅ | 5s |
| nim-v2.2 x rust-v0.56 (quic-v1) [dr - linux, rly - rust-v0.56, lr - linux] | nim-v2.2 | rust-v0.56 | quic-v1 | ✅ | 5s |
| rust-v0.56 x nim-v2.2 (tcp, noise, yamux) [dr - linux, rly - rust-v0.56, lr - linux] | rust-v0.56 | nim-v2.2 | tcp | ✅ | 7s |
| rust-v0.56 x nim-v2.2 (tcp, noise, mplex) [dr - linux, rly - rust-v0.56, lr - linux] | rust-v0.56 | nim-v2.2 | tcp | ✅ | 7s |
| nim-v2.2 x nim-v2.2 (tcp, noise, yamux) [dr - linux, rly - rust-v0.56, lr - linux] | nim-v2.2 | nim-v2.2 | tcp | ✅ | 7s |
| nim-v2.2 x nim-v2.2 (tcp, noise, mplex) [dr - linux, rly - rust-v0.56, lr - linux] | nim-v2.2 | nim-v2.2 | tcp | ✅ | 7s |
| nim-v2.2 x nim-v2.2 (quic-v1) [dr - linux, rly - rust-v0.56, lr - linux] | nim-v2.2 | nim-v2.2 | quic-v1 | ✅ | 7s |
| rust-v0.56 x nim-v2.2 (quic-v1) [dr - linux, rly - rust-v0.56, lr - linux] | rust-v0.56 | nim-v2.2 | quic-v1 | ✅ | 7s |

```

libp2p/transports/quictransport.nim
vladopajic: fyi @rlve, ^ this is blind spot in unit tests, there should be test that assert what happens when server is dialing. code needs to be the same for every transport.

richard-ramos: I don’t think this works for QUIC because `endpoint.dial()` would make both peers act as clients which is the original problem i try to solve here. 

In the spec it explicitly requires that the sender of `Sync` should remain as server and only send random UDP packets

libp2p/crypto/rng.nim
richard-ramos: While it sounds cool, it does cause annoying compile errors:
```
Error: ambiguous identifier: 'int' -- use one of the following:
  system.int: int
  rng.int: proc (rng: Rng, low: int, high: int): int
```

libp2p/transports/quictransport.nim
gmelodie, line 616: `while true...sleep`: can we replace this with a `heartbeat`?

gmelodie (approved):

libp2p/transports/quictransport.nim
vladopajic: aha, got it. in that case then this is okay,

richard-ramos: Nah :( heartbeat is meant to be use a fixed interval, but for DCUtR we should use a randomized sleep interval. https://github.com/libp2p/specs/blob/cfcf0230b2f5f11ed6dd060f97305faa973abed2/relay/DCUtR.md?plain=1#L98-L99

PR #2840, opened 2026-07-16
test(interop): build unified-testing dockerfiles workflow

rlve:
## Summary

Two interop-testing changes that go together.

First, a new CI job builds the Dockerfiles for the unified-testing interop peers (`hole-punching-v2`, `kad-dht-v2`, `transport-v2`, `perf`). Because these peers are only ever compiled by the external suite, nothing in the CI catches it when they stop building against the current tree. This job just runs `docker build` on each so a broken peer fails here instead of silently rotting.

Second, the unified-testing interop peer now defaults to `yamux` instead of `mplex`. These peers are driven by the external unified-testing suite, and the relayed-circuit hole-punch path failed against rust-libp2p because rust has dropped mplex and is yamux-only. When the nim peer defaulted to mplex, the DCUtR channel over the relayed circuit had no common muxer and negotiation failed with `no muxer available`. Defaulting to yamux fixes the mismatch. `MUXER` env var still overrides it.

### DIFF
--- a/.github/workflows/interop.yml
+++ b/.github/workflows/interop.yml
@@ -129,8 +129,8 @@ jobs:
           go build -C ./go-peer -o peer ./peer.go
           ./go-peer/peer & ./nim-peer/src/peer
 
-  # The -v2 interop peers are driven by the external unified-testing suite, so
-  # they are never built anywhere else in CI. Build their Dockerfiles here
+  # These interop peers are driven by the external unified-testing suite,
+  # so they are never built anywhere else in CI. Build their Dockerfiles here
   # to guarantee the peers keep compiling against the current nim-libp2p tree.
   build-unified-testing-interop-images:
     name: Build unified-testing interop image (${{ matrix.name }})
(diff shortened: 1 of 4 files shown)


.github/workflows/interop.yml
```diff
@@ -128,3 +128,24 @@ jobs:
           nim c ./nim-peer/src/peer.nim
           go build -C ./go-peer -o peer ./peer.go
           ./go-peer/peer & ./nim-peer/src/peer
+
+  build-unified-testing-interop-images:
```
vladopajic: it would be slightly better (just personal opinion) if this is done daily. reasoning:

it is less likely to for us to f-up this code on every commit, if we ever f-up catching issues with a day lag should be fine. most of these errors are only build related so it should be easy to fix. we also have stronger practice for taking care of daily tests `daily_ci_report.yml` workflow creates issue anytime any daily CI fails and devs take turn in maintain it with priority. 
- i feel like this will doing these builds only every commit is just too much, and i would prefer if we in general start with having different priority for CI

do this only if @richard-ramos @gmelodie approve


headsup, if we go with adding new daily workflow, include it to report daily_ci_report.yml here

vladopajic (approved):

.github/workflows/interop.yml
rlve: I agree, it feels excessive to run on every commit. I'll change to daily.

gmelodie (approved):

PR #2842, opened 2026-07-16
fix(kad): target the requested bucket when refreshing

gmelodie:
## Summary

Per-bucket refresh never queried the bucket it was asked to refresh. It degenerated into an unbiased random lookup, landing in bucket 0 roughly half the time regardless of the target — so the maintenance Kademlia relies on to keep sparse buckets alive was effectively not happening.

`randomKeyInBucket` built a key by manipulating the raw bits of `selfId` so that the *unhashed* XOR against `selfId` had exactly `i` leading zeros. But `bucketIndex` hashes the key before measuring distance (`key.hashFor(rtable.config.hasher)`), and both routing tables are constructed without a hasher (`kademlia.nim`, `routing_table_manager.nim`), so `hashFor` falls through to `defaultHasher` = sha256. Hashing destroys the prefix relationship the generator so carefully built, making the generated key's actual bucket uniformly random.

Measured on the production config (default sha256 hasher), asking for a bucket over 2000 draws:

| bucket requested | times actually hit | landed in bucket 0 |
|---|---|---|
| 1 | 477 / 2000 | 990 |
| 3 | 125 / 2000 | 1010 |
| 6 | 20 / 2000 | 1001 |
| 10 | 0 / 2000 | 982 |

Every existing test passed `hasher = Opt.some(noOpHasher)`, the identity hasher — the one configuration where the old code was correct, and why this stayed green.

The target is sent over the wire as a raw key that each remote peer hashes itself (`dispatchFindNode` -> `handleFindNode` -> `findClosestPeers`), so a target cannot be generated in hash space, and sha256 cannot be inverted. The target therefore has to be found by preimage search. This is the same conclusion go-libp2p reached; it pays the cost at build time via a precomputed prefix table (`GenRandPeerID` / `keyPrefixMap`) rather than at runtime.

`randomKeyInBucket` now takes the `RoutingTable` — it needs the hasher and `selfIdPreHashed`, which the old free-function signature could not see — and draws candidate keys until one *hashes* into the wanted bucket, returning `Opt[Key]`. `bucketIndex` is split so the search loop can hoist the self-hash out instead of re-hashing `selfId` on every attempt.

Because the search costs ~2^lz hashes, buckets nearer than `MaxRefreshLeadingZeros` (12) return `none` and are skipped by `refreshTable`. That cap is not a coverage gap: those near buckets are exactly the ones the `findNode(selfId)` at the top of `refreshTable` already refreshes, and a bucket past 12 only populates in a network of ~8k+ peers. 12 was chosen by measurement — a search at the cap costs ~14ms, where go-libp2p's equivalent of 15 would block the event loop ~110ms.

After the fix, targeting accuracy is 200/200 for every reachable bucket (0, 1, 3, 6, 10, 12), near buckets are declined rather than silently mistargeted, and the service-discovery table shape (`maxBuckets=16`, prehashed self) resolves correctly.

## Affected Areas

- [x] Protocol Logic

Scope of changes:

- `libp2p/protocols/kademlia/routing_table.nim` — `randomKeyInBucket` rewritten to search in hash space; `bucketIndex` split into `bucketIndexFor` (accepts a precomputed self-hash) plus a private `selfHash` helper; new private `minLeadingZeros`.
- `libp2p/protocols/kademlia.nim` — `refreshTable` handles `Opt[Key]` and skips buckets with no reachable target.
- `libp2p/protocols/kademlia/types.nim` — new `MaxRefreshLeadingZeros` constant.
- `tests/libp2p/kademlia/test_routing_table.nim` — call sites updated; new coverage on the default hasher.

Both the main DHT routing table and the service-discovery per-service tables (which refresh through the same `refreshTable` via `refreshAllTables`) were affected by the bug.

No wire-format or protocol change — the FIND_NODE key on the wire is still an ordinary raw key, and peers handle it exactly as before. The change is confined to which target the local node picks for its own refresh lookups, so no downstream integration branches are needed to validate it.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

**Breaking API change** to an exported proc, though one with no known external callers:

```nim
# before
proc randomKeyInBucket*(selfId: Key, bucketIndex: int, rng: Rng, maxBuckets: int = DefaultMaxBuckets): Key

# after
proc randomKeyInBucket*(rtable: RoutingTable, bucketIndex: int, rng: Rng): Opt[Key]
```

It takes the routing table instead of a bare `selfId` + `maxBuckets` (it needs the table's hasher to be correct at all) and returns `Opt[Key]`, since a target is not always obtainable. Callers pass the `RoutingTable` they already hold and handle `none` by skipping.

`bucketIndex` is now a `func` rather than a `proc`. This is a strengthening — existing call sites are unaffected.

Behavior change: bucket refresh now actually explores the bucket it targets, which is the intended Kademlia maintenance. Expect refresh lookups to reach sparse/distant buckets that previously went unrefreshed.

Performance: each refresh target now costs a bounded preimage search — negligible for low buckets (~0.01ms at bucket 0-1, ~0.2ms at bucket 6) and ~14ms at the bucket-12 cap. This is paid at most once per stale bucket per `bucketRefreshTime` (default 10 minutes), and only for buckets that are both non-empty and stale.

## Risk Assessment

- **Backward compatibility:** no wire or on-disk format change. Source-breaking only for direct callers of `randomKeyInBucket` (signature above).
- **Network behavior:** refresh traffic now spreads across buckets as designed instead of concentrating on bucket 0. Volume is unchanged — the same number of `findNode` calls, aimed at the right places. Buckets past the cap are covered by the existing self-lookup.
- **Performance:** the search is synchronous and blocks the event loop for its duration. Bucket occupancy is influenceable by a sybil that farms peer IDs to fill buckets 0-12, but the total is self-limiting: each bucket costs half the next, so the whole ladder sums to ~2^13 draws (~15ms) per 10-minute refresh cycle. Worst case for a single bucket is ~14ms at the cap.
- **Security:** targets are drawn from the same `Rng` as before and remain uniformly distributed within the target bucket, so refresh targets stay unpredictable. The wire key is now uniform random rather than carrying `selfId`'s prefix bits.

## References

- Kademlia bucket refresh: <https://docs.libp2p.io/concepts/discovery-routing/kaddht/>
- Prior art for the cap and the preimage-search approach: go-libp2p-kbucket `GenRandPeerID` / `maxCplForRefresh`, which precomputes a 16-bit prefix preimage table.

## Additional Notes

Verification was done against a standalone harness driving the real `RoutingTable` under the default sha256 hasher, which produced the before/after numbers above.

The three new tests exercise the **default sha256 hasher** — the configuration the bug actually lived in — rather than only `noOpHasher`: exact bucket targeting across several buckets, `none` past the cap, and the narrow 16-bucket service-discovery split. `Run Kademlia DHT interoperability tests` passes on this branch.

One note on the attempt budget: it is deliberately generous rather than tight. A budget of ~4x the expected draws would fail ~1.8% of the time per call, which across the 25-call loops in the existing tests would have made them flaky roughly a third of the time. Overshooting is free, since the loop exits on the first success.

### DIFF
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -36,11 +36,10 @@ proc refreshTable*(
     if not (forceRefresh or bucket.isStale()):
       continue
 
-    # Buckets too near to target are covered by the findNode on selfId above.
-    let randomKey = randomKeyInBucket(rtable, i, kad.rng).valueOr:
+    let target = rtable.refreshTarget(i, kad.rng).valueOr:
       trace "No refresh target for bucket", bucket = i
       continue
-    discard await kad.findNode(randomKey, rtable)
+    discard await kad.findNode(target, rtable)
 
 proc bootstrap*(
     kad: KadDHT, forceRefresh = false
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -36,8 +36,10 @@ proc refreshTable*(
     if not (forceRefresh or bucket.isStale()):
       continue
 
-    let randomKey =
-      randomKeyInBucket(rtable.selfId, i, kad.rng, rtable.config.maxBuckets)
+    # Buckets too near to target are covered by the findNode on selfId above.
+    let randomKey = randomKeyInBucket(rtable, i, kad.rng).valueOr:
+      trace "No refresh target for bucket", bucket = i
+      continue
     discard await kad.findNode(randomKey, rtable)
 
 proc bootstrap*(
--- a/libp2p/protocols/kademlia/routing_table.nim
+++ b/libp2p/protocols/kademlia/routing_table.nim
@@ -237,3 +237,15 @@ proc randomKey*(bucket: Bucket, rng: Rng): Opt[Key] =
     proc(e: NodeEntry): Key =
       e.nodeId
   )
+
+proc refreshTarget*(rtable: RoutingTable, bucketIndex: int, rng: Rng): Opt[Key] =
+  ## Key to run a findNode against in order to refresh `bucketIndex`.
+  let random = randomKeyInBucket(rtable, bucketIndex, rng)
+  if random.isSome():
+    return random
+
+  # Buckets near self need a shared prefix too long to draw at random. A peer
+  # the bucket already holds has that prefix by construction.
+  if bucketIndex notin 0 ..< rtable.buckets.len:
+    return Opt.none(Key)
+  rtable.buckets[bucketIndex].randomKey(rng)
--- a/libp2p/protocols/kademlia/routing_table.nim
+++ b/libp2p/protocols/kademlia/routing_table.nim
@@ -210,11 +210,6 @@ proc isStale*(bucket: Bucket): bool =
   return false
 
 proc randomKeyInBucket*(rtable: RoutingTable, bucketIndex: int, rng: Rng): Opt[Key] =
-  ## Returns a random key that `rtable` places in `bucketIndex`, for use as a
-  ## bucket-refresh lookup target. A key's bucket is decided by its hash, which
-  ## cannot be inverted, so candidates are drawn until one hashes into the
-  ## bucket, costing ~2^lz draws. Buckets nearer than `MaxRefreshLeadingZeros`,
-  ## and searches that exhaust their budget, return `none`.
   let lz = clamp(bucketIndex, 0, bucketCount(rtable.config.maxBuckets) - 1)
   if lz > MaxRefreshLeadingZeros:
     return Opt.none(Key)
@@ -222,11 +217,10 @@ proc randomKeyInBucket*(rtable: RoutingTable, bucketIndex: int, rng: Rng): Opt[K
   # A draw hits the bucket with probability ~2^-(lz+1); overshooting is free.
   let maxAttempts = 32 shl lz
   let selfHash = rtable.selfHash()
-  var candidate: array[IdLength, byte]
+  var key = newSeqUninit[byte](IdLength)
 
   for _ in 0 ..< maxAttempts:
-    rng.generate(candidate)
-    let key = @candidate
+    rng.generate(key)
     if rtable.bucketIndexFor(selfHash, key) == lz:
       return Opt.some(key)
 
--- a/libp2p/protocols/kademlia/routing_table.nim
+++ b/libp2p/protocols/kademlia/routing_table.nim
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 # Copyright (c) Status Research & Development GmbH
 
-import algorithm, sequtils, math, bitops
+import algorithm, sequtils
 import chronos, chronicles, results
 import ./types
 import ./kademlia_metrics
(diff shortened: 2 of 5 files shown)


SionoiS (approved):
LGTM

Thank you

libp2p/protocols/kademlia.nim
```diff
@@ -36,8 +36,10 @@ proc refreshTable*(
     if not (forceRefresh or bucket.isStale()):
       continue
 
-    let randomKey =
-      randomKeyInBucket(rtable.selfId, i, kad.rng, rtable.config.maxBuckets)
+    # Buckets too near to target are covered by the findNode on selfId above.
+    let randomKey = randomKeyInBucket(rtable, i, kad.rng).valueOr:
+      trace "No refresh target for bucket", bucket = i
+      continue
```
richard-ramos: Some interesting comment from claude:

---
Medium — Near buckets can become permanently unrefreshed. [kademlia.nim:39-42](https://github.com/vacp2p/nim-libp2p/blob/2b50c652b52e1648ebc89fff01b0acb5453d51d0/libp2p/protocols/kademlia.nim#L39-L42) assumes findNode(selfId) covers every bucket rejected by the 12-bit cutoff. It only traverses the globally closest peers, so intermediate stale buckets can remain untouched. More seriously, service-discovery tables have 16 buckets: buckets 1–15 require at least 16 leading zeros and are therefore all skipped, even with forceRefresh. Because their selfId is already hashed, the initial wire lookup hashes it again and does not target the service table’s center. Consider falling back to a random existing entry from the non-empty bucket or using precomputed/nonblocking target generation.

richard-ramos (approved):

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -40,20 +40,29 @@ proc new*(
...
+
+func bucketIndex*(rtable: RoutingTable, key: Key): int =
+  rtable.bucketIndexFor(rtable.selfHash(), key)
+
+func minLeadingZeros(rtable: RoutingTable, bucketIndex: int): int =
+  ## Fewest leading zeros in the XOR distance that lands a key in `bucketIndex`.
+  if rtable.config.maxBuckets <= 1:
+    return 0
+
+  bucketIndex * 256 div rtable.config.maxBuckets
```
tinniaru3005: I think this should be just `bucketIndex` to match the latest formula

libp2p/protocols/kademlia.nim
gmelodie: Opened an issue to fix `findNode`: https://github.com/vacp2p/nim-libp2p/issues/2846

libp2p/protocols/kademlia/routing_table.nim
gmelodie: hum... but for example if `maxBuckets == 16` (in service disco) and `bucketIndex == 1` the result would have to be `16`, no?

Digging in did turn up a real bug though: `bucketIndexFor` floors, so its inverse has to round up, and the `div` here floors as well. It's off by one for any `maxBuckets` that doesn't divide `256` (96/100 buckets wrong at `maxBuckets == 100`). Both current configs are divisors so nothing's broken today, but I've switched it to `ceilDiv` (same result for `256` and `16`, correct for the rest).

tinniaru3005 (approved):
LGTM

vladopajic (approved):

PR #2845, opened 2026-07-16
fix(service-disco): ads pruning

SionoiS:
Ads are stored per service (`cache[ServiceId]`), but timestamps are global by `(peerId, seqNo)`. When the same ad appeared under multiple services, the first prune deleted the shared timestamp; the next service hit `withValue` with no match and never advanced `i`, so `pruneAdsForService` looped forever and left orphans uncleared.

### DIFF
--- a/libp2p/protocols/service_discovery/registrar.nim
+++ b/libp2p/protocols/service_discovery/registrar.nim
@@ -54,6 +54,20 @@ proc removeAd*(ipTree: IpTree, ad: Advertisement) {.raises: [].} =
 proc isExpired(now, ts: Moment, expiry: Duration): bool =
   now - ts > expiry
 
+proc removeAdvertisementEverywhere(registrar: Registrar, ad: Advertisement) =
+  let key = ad.toAdvertisementKey()
+
+  registrar.cacheTimestamps.del(key)
+
+  for _, sads in registrar.cache.mpairs:
+    var j = 0
+    while j < sads.len:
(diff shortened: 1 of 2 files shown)


tinniaru3005 (approved):
LGTM :)

AlbertoSoutullo:
Hmmm I don't know if I am fully convinced about this approach (personally). I mean yes, this will fix the infinite loop because the prune loop now guarantees progress when a cache entry has no matching timestamp, so the `withValue` miss can no longer leave `i` unchanged forever.

But there still is what we talked in the thread. This leaves the deletion a bit implicit. A multi-service advertisement can exist in multiple `registrar.cache[serviceId]` lists while sharing one `cacheTimestamps[(peerId, seqNo)]` entry. The new logic removes orphaned copies as pruning reaches them, which fixes the CPU issue, but it does not centralize the "deleting an advertisement key should clean all cache copies for that key". 
I don't know how you want to tackle this, but if you want to leave it like this I think it would be good if we either documented that lazy cleanup is intentional, or introduced a helper for deleting an `AdvertisementKey` from all registrar indexes. Personally I would suggest going with the second option, but not my call.

If option 2 ends up happening, I’d also add these tests for example:
- orphan ad present in `cache` but missing from `cacheTimestamps`
- same expired ad present under multiple service IDs
- a second prune after cleanup being a no-op

SionoiS:
closing in favor of https://github.com/vacp2p/nim-libp2p/pull/2849

PR #2849, opened 2026-07-17
chore(service-disco): refactor registrar cache

SionoiS:
A complete rewrite of the advertisement cache.


libp2p/protocols/service_discovery/types.nim
```diff
@@ -50,10 +50,24 @@ type
...
-    cacheTimestamps*: Table[AdvertisementKey, Moment]
+  AdPutResult* = enum
+    AdIgnored
+    AdRefreshed
+    AdInserted
+    AdReplaced
+
+  CachedAd* = ref object
+    ad*: Advertisement
+    timestamp*: Moment
```
jm-clius, line 56: Yes! Thanks. Makes much more sense than keeping timestamps cached separately. On the whole, though, I'm questioning the need for a single ad cache (or at least what it's purpose should be). See my comments elsewhere.

libp2p/protocols/service_discovery/types.nim
```diff
@@ -138,6 +152,13 @@ proc toAdvertisementKey*(ad: Advertisement): AdvertisementKey {.raises: [].} =
 proc hash*(ad: Advertisement): Hash {.raises: [].} =
   hash(ad.envelope.signature.data)
 
+proc new*(T: typedesc[AdvertisementCache]): T {.raises: [].} =
+  T(
+    entries: initTable[AdvertisementKey, CachedAd](),
```
jm-clius: This obscures the fact that an `ad` belongs solely to the single service ID against which it was advertised on the wire. Since the `AdvertisementKey` here is only the `peerId` and `seqNo`, we would not be able to distinguish between `ad` s from the same peerID but that was registered against different service IDs. Even if these ads are byte-identical, they are "different" ads based on the fact that they advertise a different service.

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,231 @@
...
+
+    if hasPeer:
+      if ad.data.seqNo == oldEntry.ad.data.seqNo:
+        oldEntry.timestamp = now
+        return AdRefreshed
+      elif ad.data.seqNo < oldEntry.ad.data.seqNo:
+        return AdIgnored
+      else:
+        c.replaceEverywhere(existingKey, ad, now)
+        return AdReplaced
```
jm-clius: As mentioned elsewhere, I'm not convinced that the Registrar should take heed of `seqNo`. We are simply following the specced Service Discovery Protocol (i.e. ads are placed if they pass admission control), and this would translate to a non-specified addition to the protocol. Of course, we should deduplicate byte-equivalent ads and update the associated timestamp, but this should (perhaps?) only happen within a specific service ID routing table.

jm-clius (commented):
Thanks!

Adding my comments here that I added on Discord as well. This already ignores the "services" encoded inside the `ad`, which is great, but I think the consequences can probably lead to even further simplification:

---
I think we'll have a much cleaner design if we imagine (or enforce!) that the service information _inside_ the `Advertisement` is mostly opaque on service discovery level (i.e. none of registrar, discoverer, advertiser "knows" anything about what's encoded inside the ad). The encoded `services` field is _only_ of interest to the consuming application of the ad. The advertiser may choose to encode information about the advertised service (e.g. mix key) or even other services that it supports - but only the application that receives the ad should know about that and decide if it's interesting. This means that only the `serviceId` against which the `ad` was registered on the wire protocol is ever considered within the discovery protocol. Importantly, vice versa the `ad` has no meaning without the externally provided `serviceId`. For that reason, something like `updateAd(ad: Advertisement)` wouldn't make sense - because the ad has no meaning without the `serviceId` it was advertised against. The `AdvertisementKey` that we extract would then also not make sense - an ad cannot only be indexed against included `peerId` and `seqNo`, because the exact same `ad` against a different `serviceId` _is_ a different ad. As it stands, we wouldn't allow the same peer to advertise against multiple services, because [this table](https://github.com/vacp2p/nim-libp2p/pull/2849/changes#diff-9cf8a98a04175697bc5f864f0131d8f9d94ee8dafb8a94f19af34df3e15b357fR65) only allows one cached ad per peerId,seqNo tuple.

I would therefore also view the `seqNo` (probably) as something related to service information and that's therefore only interesting to the application. It helps the app to distinguish between different versions of service information - if I discover `mix` peers and I get multiple ads for the same peer, I (as an application) can use the `seqNo` to decide which to keep. From registrar's perspective, any new `ad` register request is treated in the same way - apply IP-based and service-based admission control - regardless of `seqNo`. Afaict the DISC-NG paper and our LIP also doesn't consider `seqNo` as part of the discovery protocol. Of course, byte-equivalent ads for the same service and peerId should be deduplicated within the service table. This would greatly simplify the structure if we don't need multiple (ambiguous) indexes for each ad - an ad is something that belongs only to a single service-centered routing table and has no meaning outside of that service-specific context. I might of course be missing many nuances here, but I would suggest this as starting point for the conversation. Either way, it seems structural rethink (already started) is the way to go here rather than trying to work around the symptoms we've noticed in the tests.

SionoiS:
I simplified the cache as much as possible and now the questions is what additional features do we really need if any?

@jm-clius @tinniaru3005

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,131 @@
...
+
+proc put*(c: AdvertisementCache, serviceId: ServiceId, ad: Advertisement, now: Moment) =
+  ## Append a new ad slot for `serviceId`. Evicts the oldest slot when full.
+  if c.len.uint64 >= c.capacity:
+    c.evictOldest()
+
+  if serviceId notin c.byService:
+    c.byService[serviceId] = @[]
+  c.byService.withValue(serviceId, slots):
+    slots[].add(CachedAd(ad: ad, timestamp: now))
```
tinniaru3005: I think we need a check before this for whether an identical ad is already cached for this `serviceId`, so a resubmission refreshes the timestamp instead of appending a duplicate slot.

SionoiS:
The spec will be updated to specify what is an identical ad. I will then add this feature and this PR will be ready for review.

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,138 @@
...
+import ./types
+
+export types
+
+proc getIPs(addrsInfos: seq[AddressInfo]): seq[IpAddress] =
+  var ips: seq[IpAddress]
+  for addrInfo in addrsInfos:
+    addrInfo.address.getIp().withValue(ip):
+      ips.add(ip)
+  ips
```
vladopajic: this proc is utility of `AddressInfo`, therefore it should be defined with this type. it should not be moved left-and-right with each refactor.

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,138 @@
...
+
+proc insertAd(ipTree: IpTree, ad: Advertisement) =
+  for ip in ad.data.addresses.getIPs():
+    ipTree.insertIp(ip)
+
+proc removeAd(ipTree: IpTree, ad: Advertisement) =
+  for ip in ad.data.addresses.getIPs():
+    ipTree.removeIp(ip)
+
+proc adScore(ipTree: IpTree, ad: Advertisement): float64 =
```
vladopajic: `adScore` or `maxScore`? why is exposed naming different the underlying logic?

```nim
var ad: Advertisement


ad.adScore()
# vs
ad.maxScore()

# later usage is bit nicer
```

libp2p/protocols/service_discovery/advertisement_cache.nim
vladopajic: also `var ips: seq[IpAddress]` can be preallocated as max size is known

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,138 @@
...
+
+  for serviceId, slots in c.byService:
+    for i, slot in slots:
+      if slot.timestamp < oldestTime:
+        oldestTime = slot.timestamp
+        oldestService = serviceId
+        oldestIndex = i
+
+  if oldestIndex >= 0:
+    c.removeSlot(oldestService, oldestIndex)
```
vladopajic:
```
 let oldestIndex = c.findOldest()
 if oldestIndex >= 0:
    c.removeSlot(oldestService, oldestIndex)
```

vladopajic (commented):
pardon me for premature nitpicks

SionoiS:
> pardon me for premature nitpicks

Nitpicks always welcome :D

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,137 @@
...
+  return oldestEntree
+
+proc evictOldest(c: AdvertisementCache) =
+  let optIndex = c.findOldestIndex()
+  if optIndex.isSome():
+    let (oldestService, oldestIndex) = optIndex.get()
+    c.removeSlot(oldestService, oldestIndex)
+
+proc put*(c: AdvertisementCache, serviceId: ServiceId, ad: Advertisement, now: Moment) =
+  ## Append a new ad slot for `serviceId`. Evicts the oldest slot when full.
```
richard-ramos: Should this function check if the seqNo of the advertisement being added is lower than the one it exists in case there's already an advertisement for the same `ad` in `slots`, to avoid inserting a stale record?

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -463,6 +290,15 @@ proc registration*(disco: ServiceDiscovery, peerId: PeerId, inMsg: Message): Mes
 
     return msg
 
+  if disco.registrar.ads.contains(serviceId, ad):
+    debug "advertisement already in cache", serviceId, peerId = ad.data.peerId
+
+    cd_register_requests.inc(
```
richard-ramos: Wouldnt this end up rejecting self advertising refreshes since they'll have the same bytes? perhaps doing a duplicated registration should refresh instead the cached timestamp, instead of returning Reject 🤔

libp2p/protocols/service_discovery/advertisement_cache.nim
SionoiS: I think we are going in the direction of opaque ads. Ads are just bytes, the implementation doesn't know what's in them and doesn't care.

libp2p/protocols/service_discovery/registrar.nim
SionoiS: I was assuming the re-advertisement would get a wait time longer than the ad expiry but if not then yes your right that would not work.

Why is it so complicated... :(

jm-clius: > this end up rejecting self advertising refreshes

My understanding of the DISC-NG spec is that this is the expected behaviour, in other words, there is no "refresh" mechanism for advertisers like we have in RDV. An advertiser tries to maintain up to `K` registration attempts at random registrars from each bucket. Once it succeeds, that entry has an expiry `E` after which the advertiser's maintenance will select _another random_ registrar from the same bucket. I think the randomisation and decay is part of the algorithm here and therefore intentional. If there are fewer than `K` peers in the bucket, the same ones will constantly be selected, in which case the peer will be subjected to the same admission control rules than everyone else if it reattempts registration.

jm-clius: Ok, update: Indeed, after doing more thinking I think the expected behaviour is that there are no scheduled "keep alives" or "refreshes", but that an advertiser registers at random registrars for each bucket. However, this still means that probabilistically the same advertiser will attempt to register the same `ad` at the same registrar from time to time. This byte-equivalence check will prevent it from doing so, so I agree that we should remove it.

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,137 @@
...
+  for ip in ad.data.addresses.getIPs():
+    ipTree.removeIp(ip)
+
+proc adMaxScore(ipTree: IpTree, ad: Advertisement): float64 =
+  ## Max IP similarity score across the advertisement's addresses.
+  var maxScore = 0.0
+  for ip in ad.data.addresses.getIPs():
+    let score = ipTree.ipScore(ip)
+    if score > maxScore:
+      maxScore = score
```
gmelodie: could this be a heap queue?

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,137 @@
...
+  maxScore
+
+proc len*(c: AdvertisementCache): int =
+  var n = 0
+  for _, slots in c.byService:
+    n += slots.len
+  n
```
gmelodie: Could we keep this size up to date when inserting/removing and have `len()` just return a variable?

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,137 @@
...
+  ## True when an identical ad (same envelope signature) is already cached
+  ## under `serviceId`.
+  c.byService.withValue(serviceId, slots):
+    for slot in slots[]:
+      if slot.ad.envelope.signature.data == ad.envelope.signature.data:
+        return true
+  false
+
+proc adsForService*(c: AdvertisementCache, serviceId: ServiceId): seq[Advertisement] =
+  result = @[]
```
gmelodie: Don't use `result`

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,137 @@
...
+  ## Total IP multi-set size (IPv4 + IPv6 root counters).
+  c.ipTree.root.counter + c.ipTree.root6.counter
+
+proc ipScore*(c: AdvertisementCache, ip: IpAddress): float64 =
+  c.ipTree.ipScore(ip)
+
+proc removeSlot(c: AdvertisementCache, serviceId: ServiceId, index: int) =
+  c.byService.withValue(serviceId, slots):
+    if index < 0 or index >= slots[].len:
+      return
```
gmelodie: can we do this instead?
```suggestion
    if index not in 0..slots[].len:
      return
```

more importantly: should this be an error?

```suggestion
    if index not in 0..slots[].len:
      error "index out of bounds", start = 0, end = slots[].len
      return
```

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,137 @@
...
+        oldestTime = slot.timestamp
+        oldestEntree = Opt.some((serviceId, i))
+
+  return oldestEntree
+
+proc evictOldest(c: AdvertisementCache) =
+  let optIndex = c.findOldestIndex()
+  if optIndex.isSome():
+    let (oldestService, oldestIndex) = optIndex.get()
+    c.removeSlot(oldestService, oldestIndex)
```
gmelodie:
```suggestion
proc evictOldest(c: AdvertisementCache) =
  c.findOldestIndex().withValue((oldestService, oldestIndex)):
    c.removeSlot(oldestService, oldestIndex)
```

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,137 @@
...
+
+proc put*(c: AdvertisementCache, serviceId: ServiceId, ad: Advertisement, now: Moment) =
+  ## Append a new ad slot for `serviceId`. Evicts the oldest slot when full.
+  ## Callers must reject duplicates via `contains` before calling `put`.
```
gmelodie: This seems error prone. Let's add a guard

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,137 @@
...
+      else:
+        kept.add(slot)
+    slots = kept
+    if slots.len == 0:
+      emptyServices.add(serviceId)
+
+  for serviceId in emptyServices:
+    c.byService.del(serviceId)
+
+  removed
```
gmelodie, line 125: i asked claude to optimize this part, here's what it came up with:
```suggestion
proc pruneExpired*(c: AdvertisementCache, now: Moment, expiry: Duration): int =
  ## Remove slots whose timestamp is older than `expiry`. Returns how many
  ## slots were removed.
  let cutoff = now - expiry
  var
    removed = 0
    emptyServices: seq[ServiceId]

  for serviceId, slots in c.byService.mpairs:
    var kept = 0
    for i in 0 ..< slots.len:
      if slots[i].timestamp < cutoff:
        c.ipTree.removeAd(slots[i].ad)
      else:
        if kept != i:
          slots[kept] = move(slots[i])
        inc kept

    removed += slots.len - kept
    slots.setLen(kept)

    if kept == 0:
      emptyServices.add(serviceId)

  for serviceId in emptyServices:
    c.byService.del(serviceId)

  removed
```

gmelodie (approved):

libp2p/protocols/service_discovery/advertisement_cache.nim
richard-ramos: I'm surprised that copilot didnt catch this :thinking:

richard-ramos (approved):

libp2p/protocols/service_discovery/advertisement_cache.nim
SionoiS: What do you mean? The ip tree?

SionoiS:
I converted back to draft, waiting on more spec changes and discussions. Sorry for wasting your time guys.

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -475,14 +328,13 @@ proc registration*(disco: ServiceDiscovery, peerId: PeerId, inMsg: Message): Mes
...
-    disco.discoConfig, ad, disco.discoConfig.advertCacheCap, serviceId, now
-  )
+  let ips = disco.advertiserIps(peerId, connectionIps)
+  var tWait = disco.registrar.waitingTime(disco.discoConfig, serviceId, ips, now)
 
   disco.updateWaitAfterRetry(ticketOpt, now, tWait)
 
   if tWait <= ZeroDuration:
-    disco.acceptAdvertisement(now, serviceId, ad)
+    disco.acceptAdvertisement(now, serviceId, peerId, ad, ips)
```
jm-clius, line 330: I assume if we decide to index the `ad` against the encoded `peerId` rather than the advertiser's `peerId` is as simple as changing this argument?

jm-clius (approved):
Thanks for this simplification! ~I think the only open question is about whether we use the encoded `peerId` vs advertiser `peerId` for indexing in the registrar's advertisementCache. I am now leaning again to using encoded `peerId`, to allow the same advertiser to place multiple `ads` for the same service but for different modules. Either way, I think this is an improvement and mergeable.~

UPDATE: Actually, coming to think of it, we'll probably be safer to just use the advertiser's peerId for now - advertising multiple peerIds for the same service from the same advertiser is a bit of an exotic scenario. Furthermore, it will prevent the simple replacement attack, where an advertiser attempts to overwrite current ads for a peerId with old (but valid) information that it has cached for this purpose.

libp2p/protocols/service_discovery/registrar.nim
jm-clius: UPDATE: Actually, coming to think of it, we'll probably be safer to just use the advertiser's `peerId` for now - advertising multiple `peerId`s for the _same_ service from the same advertiser is a bit of an exotic scenario. Furthermore, it will prevent the simple replacement attack, where an advertiser attempts to overwrite current `ad`s for a `peerId` with old (but valid) information that it has cached for this purpose.

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,161 @@
...
+    ipTree.removeIp(ip)
+
+proc ipMaxScore(ipTree: IpTree, ips: seq[IpAddress]): float64 =
+  ## Max IP similarity score across the given addresses.
+  var maxScore = 0.0
+  for ip in ips:
+    let score = ipTree.ipScore(ip)
+    if score > maxScore:
+      maxScore = score
+  maxScore
```
jm-clius: It seems we'll return `0.0` for someone for whom we for some reason couldn't resolve a single `ip`. I don't think this should happen, but, for safety, an empty `ips` sequence should probably return a very, very high score?

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,161 @@
...
+  for (serviceId, advertiser) in toRemove:
+    c.removeSlot(serviceId, advertiser)
+    inc removed
+
+  for serviceId, peers in c.byService:
+    if peers.len == 0:
+      emptyServices.add(serviceId)
+
+  for serviceId in emptyServices:
+    c.byService.del(serviceId)
```
jm-clius: Don't we already remove the service when the last slot gets removed [here](https://github.com/vacp2p/nim-libp2p/pull/2849/changes#diff-701eecf561dfccf20596e5984412b02bca20b42f69b0b784b60994701cfafc32R82-R83)? In other words, these two loops are unreachable.

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,161 @@
...
+  0
+
+proc containsService*(c: AdvertisementCache, serviceId: ServiceId): bool =
+  serviceId in c.byService
+
+proc contains*(c: AdvertisementCache, serviceId: ServiceId, advertiser: PeerId): bool =
+  ## True when `advertiser` already has a cached ad under `serviceId`.
+  c.byService.withValue(serviceId, peers):
+    return advertiser in peers[]
+  false
```
jm-clius, line 29: I think since we don't deduplicate anymore we probably don't need this function anymore? Not sure.

jm-clius (approved):
Sorry for leaving two reviews - noticed some minor items when I started comparing to the ENR code.

libp2p/protocols/service_discovery/advertisement_cache.nim
SionoiS: This is plugged in the waiting time formula so to promote IP diversity we can't penalize IPs we never seen before.

jm-clius: Maybe I'm missing something, but I'm not referring to IPs we haven't seen before - I am referring to advertisers for whom we cannot resolve a single IP. I'm not sure under which circumstances this will happen, but the IP selection [heuristic here](https://github.com/vacp2p/nim-libp2p/pull/2849/changes#diff-af68724b10e129b71233b8432c71c63ab68f8afea5cccd122329f6ab2d2e9d64R67-R81) seems to suggest that it can happen (empty `ips` sequence). In that case, we'll not apply any waiting time based on IP looking at this `proc`.

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,150 @@
...
+
+export types
+
+proc insertIps(ipTree: IpTree, ips: seq[IpAddress]) =
+  for ip in ips:
+    ipTree.insertIp(ip)
+
+proc removeIps(ipTree: IpTree, ips: seq[IpAddress]) =
+  for ip in ips:
+    ipTree.removeIp(ip)
```
vladopajic: these two (insertIps, removeIps) would also make sense to be moved to `IpTree` type as well. they seem to be natural fit for that type.

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,150 @@
...
+
+proc ipMaxScore(ipTree: IpTree, ips: seq[IpAddress]): float64 =
```
vladopajic: could also be in `IpTree`

libp2p/protocols/service_discovery/advertisement_cache.nim
gmelodie: Could we keep a heap queue with the scores to avoid O(n) sweep every time we call ipMaxScore?

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,150 @@
...
+proc len*(c: AdvertisementCache): int =
+  c.count
+
+proc serviceCount*(c: AdvertisementCache): int =
+  c.byService.len
+
+proc serviceAdCount*(c: AdvertisementCache, serviceId: ServiceId): int =
+  c.byService.withValue(serviceId, peers):
+    return peers[].len
+  0
```
gmelodie: If serviceId is not in cache, do we want to return/log an error?

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,150 @@
...
+
+proc containsService*(c: AdvertisementCache, serviceId: ServiceId): bool =
+  serviceId in c.byService
```
gmelodie: nit: i think you can call this `contains` too

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,150 @@
...
+  c.byService.withValue(serviceId, peers):
+    peers[].withValue(advertiser, slot):
+      return Opt.some(slot[])
+  Opt.none(CachedAd)
+
+proc ipMaxScore*(c: AdvertisementCache, ips: seq[IpAddress]): float64 =
+  c.ipTree.ipMaxScore(ips)
+
+proc ipTotal*(c: AdvertisementCache): int =
+  c.ipTree.root.counter + c.ipTree.root6.counter
```
vladopajic:
```suggestion
  c.ipTree.total()
```

add total to `ipTree`

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,150 @@
...
+    return peers[].len
+  0
+
+proc containsService*(c: AdvertisementCache, serviceId: ServiceId): bool =
+  serviceId in c.byService
+
+proc contains*(c: AdvertisementCache, serviceId: ServiceId, advertiser: PeerId): bool =
+  c.byService.withValue(serviceId, peers):
+    return advertiser in peers[]
+  false
```
gmelodie, line 29: if we name the above `contains` too, this could be
```suggestion
  c.contains(serviceId) and advertiser in c[serviceId].peers[]
```
i *think*

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,150 @@
...
+
+proc adsForService*(c: AdvertisementCache, serviceId: ServiceId): seq[Advertisement] =
```
gmelodie: nit: this proc is very similar to the one below, so i would expect the name to be similar too
```suggestion
proc getCachedAds*(c: AdvertisementCache, serviceId: ServiceId): seq[Advertisement] =
```

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,150 @@
...
+): Opt[CachedAd] =
+  c.byService.withValue(serviceId, peers):
+    peers[].withValue(advertiser, slot):
+      return Opt.some(slot[])
+  Opt.none(CachedAd)
+
+proc ipMaxScore*(c: AdvertisementCache, ips: seq[IpAddress]): float64 =
+  c.ipTree.ipMaxScore(ips)
+
+proc ipTotal*(c: AdvertisementCache): int =
```
gmelodie: nit
```suggestion
proc lenIps*(c: AdvertisementCache): int =
```
or
```suggestion
proc ipsLen*(c: AdvertisementCache): int =
```

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,150 @@
...
+
+proc ipScore*(c: AdvertisementCache, ip: IpAddress): float64 =
+  c.ipTree.ipScore(ip)
+
+proc removeSlot(c: AdvertisementCache, serviceId: ServiceId, advertiser: PeerId) =
+  c.byService.withValue(serviceId, peers):
+    peers[].withValue(advertiser, slot):
+      c.ipTree.removeIps(slot[].ips)
+      peers[].del(advertiser)
+      dec c.count
```
gmelodie: Here `removeIps` passes alist of IPs, but `dec c.count` only decrements `count` by 1, should it be `c.count -= slot[].ips.len()`?

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -129,17 +61,35 @@ proc pruneExpiredAds*(registrar: Registrar, advertExpiry: Duration) =
...
+
+  var addrs = disco.switch.peerStore[AddressBook][advertiser]
+  if addrs.len == 0 and advertiser == disco.switch.peerInfo.peerId:
+    addrs = disco.switch.peerInfo.addrs
+
+  var ips = newSeqOfCap[IpAddress](addrs.len)
+  for ma in addrs:
+    ma.getIp().withValue(ip):
+      ips.add(ip)
+  ips
```
vladopajic: this can be some utility in multiaddress type.

then just call here:
```suggestion
  addrs.pluckIp()
```

seems like something that could be reused

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,150 @@
...
+proc ipMaxScore*(c: AdvertisementCache, ips: seq[IpAddress]): float64 =
+  c.ipTree.ipMaxScore(ips)
+
+proc ipTotal*(c: AdvertisementCache): int =
+  c.ipTree.root.counter + c.ipTree.root6.counter
+
+proc ipScore*(c: AdvertisementCache, ip: IpAddress): float64 =
+  c.ipTree.ipScore(ip)
+
+proc removeSlot(c: AdvertisementCache, serviceId: ServiceId, advertiser: PeerId) =
```
gmelodie: I found the `slot` wording confusing. What is a slot? This question would answer itself if instead we had a `type Slot` or `type IpTreeSlot` (then dev/user could just look at the fields it holds, or, if it's just a sequence of ip addresses, they'd see that)

libp2p/protocols/service_discovery/advertisement_cache.nim
SionoiS: > Maybe I'm missing something, but I'm not referring to IPs we haven't seen before - I am referring to advertisers for whom we cannot resolve a single IP. I'm not sure under which circumstances this will happen, but the IP selection [heuristic here](https://github.com/vacp2p/nim-libp2p/pull/2849/changes#diff-af68724b10e129b71233b8432c71c63ab68f8afea5cccd122329f6ab2d2e9d64R67-R81) seems to suggest that it can happen (empty `ips` sequence). In that case, we'll not apply any waiting time based on IP looking at this `proc`.

Ah yes that make sense. Good catch!

libp2p/protocols/service_discovery/advertisement_cache.nim
```diff
@@ -0,0 +1,150 @@
...
+proc containsService*(c: AdvertisementCache, serviceId: ServiceId): bool =
+  serviceId in c.byService
+
+proc contains*(c: AdvertisementCache, serviceId: ServiceId, advertiser: PeerId): bool =
+  c.byService.withValue(serviceId, peers):
+    return advertiser in peers[]
+  false
+
+proc adsForService*(c: AdvertisementCache, serviceId: ServiceId): seq[Advertisement] =
+  var ads: seq[Advertisement] = @[]
```
gmelodie: Can we pass a `limit` and use `newSeqOfCap` here?
```suggestion
  var ads = newSeqOfCap[Advertisement](limit)
```

gmelodie (commented):
Interesting finds by Claude (worth checking):
- The cache key is now the sender, not the record owner. registrar passes `peerId` (the stream sender) as `advertiser`. `isValidAdvertisement` never checks `ad.data.peerId == peerId`. Signed records are public, every getAds response hands them out. So N sybil peers can replay one victim's record and take N slots that all hold the same ad.
- The seqNo ordering check is gone. `put` (advertisement_cache.nim:98) always overwrites and always refreshes the timestamp. The old `updateExistingAd` dropped a lower `seqNo` as stale. A retried registration, a second connection, or a reconnect carrying an older cached record now replaces newer addresses with older ones and resets the expiry clock. The tests were rewritten to bless this ("lower seqNo also replaces"), so it reads as intentional — but nothing in the PR says why the ordering guarantee is safe to drop. Why is this the case?
- put unconditionally calls removeIps then insertIps — two tree walks of 32 or 128 levels per refresh. acceptAdvertisement calls updateRegistrarMetrics() unconditionally, recomputing ipTotal and setting three gauges. The old path skipped both when nothing changed.
- put costs up to five hash lookups of a 32-byte ServiceId. `withValue`, advertiser in peers[], inner withValue, serviceId notin c.byService, withValue again. `mgetOrPut` does it in one. (not: this seems to be true for most withValue blocks)

libp2p/protocols/service_discovery/advertisement_cache.nim
vladopajic: more idiomatic approach in nim is to overload `in`

```
proc `in`(value: ServiceId, c: AdvertisementCache): bool =
````

File: libp2p/protocols/service_discovery/advertisement_cache.nim (hunk 26 above)
Comment (vladopajic) [reply libp2p/protocols/service_discovery/advertisement_cache.nim:29]: even better
```
serviceId in c and advetiser in c[serviceId].peers[]
```

File: libp2p/protocols/service_discovery/registrar.nim
Comment (richard-ramos) [inline libp2p/protocols/service_discovery/registrar.nim:311]:
```suggestion


  if peerId != ad.data.peerId:
    error "advertisement peer mismatch",
      streamPeerId = peerId,
      advertisedPeerId = ad.data.peerId

    cd_register_requests.inc(
      labelValues = [$kademlia_protobuf.RegistrationStatus.Rejected]
    )

    return msg

```

Maybe something like this to make sure that the stream peerid matches the ad peerid?

richard-ramos (approved):

SionoiS:
> * The cache key is now the sender, not the record owner. registrar passes `peerId` (the stream sender) as `advertiser`. `isValidAdvertisement` never checks `ad.data.peerId == peerId`. Signed records are public, every getAds response hands them out. So N sybil peers can replay one victim's record and take N slots that all hold the same ad.

Yes that's fine the opposite is not. We changed the spec. The IP and peerID checked are now the advertiser's instead of what is included in the ad.

SionoiS:
> * The seqNo ordering check is gone. `put` (advertisement_cache.nim:98) always overwrites and always refreshes the timestamp. The old `updateExistingAd` dropped a lower `seqNo` as stale. A retried registration, a second connection, or a reconnect carrying an older cached record now replaces newer addresses with older ones and resets the expiry clock. The tests were rewritten to bless this ("lower seqNo also replaces"), so it reads as intentional — but nothing in the PR says why the ordering guarantee is safe to drop. Why is this the case?

Since there's now only one ad per advertiser per service. What an ad is and what it contains doesn't matter anymore.

SionoiS:
> * put unconditionally calls removeIps then insertIps — two tree walks of 32 or 128 levels per refresh. acceptAdvertisement calls updateRegistrarMetrics() unconditionally, recomputing ipTotal and setting three gauges. The old path skipped both when nothing changed.

True but inconsequential, the admission control mechanism make an ad replacement very unlikely anyway.

SionoiS:
> * put costs up to five hash lookups of a 32-byte ServiceId. `withValue`, advertiser in peers[], inner withValue, serviceId notin c.byService, withValue again. `mgetOrPut` does it in one. (not: this seems to be true for most withValue blocks)

I'll try to refactor this.

libp2p/protocols/service_discovery/advertisement_cache.nim
SionoiS: No the count is for the number of cached ad. Each of those can have multiple IPs.

libp2p/protocols/service_discovery/advertisement_cache.nim
SionoiS: True, the type is `CachedAd` and the wording should reflect that.

libp2p/protocols/service_discovery/advertisement_cache.nim
SionoiS: We are comparing the score of the IPs of an advertiser. We rarely have more than 2 to compare.

libp2p/protocols/service_discovery/advertisement_cache.nim
SionoiS: I can't see a reason to. 🤷

libp2p/protocols/service_discovery/registrar.nim
SionoiS: No we specifically don't want to do that because it would prevent the logos service disco module from advertising other logos module's ads.

libp2p/protocols/service_discovery/advertisement_cache.nim
SionoiS: I tried and you have to use `contains` otherwise you don't get `notin` for free.

gmelodie (approved):

PR #2850, opened 2026-07-17
fix(kad): exclude the requester from closerPeers

### LINKED ISSUE #2848: kad-dht: should closerPeers exclude the requester?
## Description

`findClosestPeers` builds the `closerPeers` list for `FIND_NODE` and `GET_PROVIDERS` replies. It filters out self but not the sender. So when the sender is in our routing table and lands among the `k` closest to the target key, we return the sender's own id to it.

Is that intended? It does no harm against a client that filters out its own id, but it caused a interop hang against py-libp2p (v0.6.0) ([unified-testing PR](https://github.com/libp2p/unified-testing/pull/119)). On receiving its own id, py-libp2p opened a stream back to itself, sent `FIND_NODE`, and blocked waiting for a reply that a node in client mode never sends. The whole lookup stalled until the test timeout. A more careful client would not hang, but the question stands: should nim avoid returning the sender at all?

## Spec Recommendation

The spec does not say. For `FIND_NODE` it states only that the reply's `closerPeers` is "the `k` closest `Peer`s", and the same for `GET_PROVIDERS`. There is no exception for the sender, nor for self.

[spec L490-L492](https://github.com/libp2p/specs/blob/6b6203ee6f62938ce67efdb33498173f475851c0/kad-dht/README.md?plain=1#L490-L492) (FIND_NODE)

[spec L503-L505](https://github.com/libp2p/specs/blob/6b6203ee6f62938ce67efdb33498173f475851c0/kad-dht/README.md?plain=1#L503-L505) (GET_PROVIDERS)

The only nearby guidance is on the querying side, where the lookup loop is expected to bound each query with a timeout and discard peers that time out. That is a client-side expectation, not a rule for the node building the reply.

[spec L164-L169](https://github.com/libp2p/specs/blob/6b6203ee6f62938ce67efdb33498173f475851c0/kad-dht/README.md?plain=1#L164-L169) (lookup loop)

## Implementation

`findClosestPeers` takes only `target` and filters only self, so it cannot drop the sender:

https://github.com/vacp2p/nim-libp2p/blob/0b262a3ae4c55aefe290fae33f8cc8d2d68bf6b5/libp2p/protocols/kademlia/find.nim#L507-L512

`handleFindNode` and `handleGetProviders` put its result straight into the reply. The sender is available as `stream.peerId` in both, but it is never threaded into `findClosestPeers`:

https://github.com/vacp2p/nim-libp2p/blob/0b262a3ae4c55aefe290fae33f8cc8d2d68bf6b5/libp2p/protocols/kademlia/find.nim#L514-L523

https://github.com/vacp2p/nim-libp2p/blob/0b262a3ae4c55aefe290fae33f8cc8d2d68bf6b5/libp2p/protocols/kademlia/provider.nim#L377-L396

## Reference: go-libp2p

go-libp2p-kad-dht excludes both self and the sender. `closestPeersToQuery` takes the sender as `from` and skips it. It asks the routing table for `count+1` peers so it can drop `from` and still return `count`:

[dht.go L751-L779](https://github.com/libp2p/go-libp2p-kad-dht/blob/cb9639642a8d54f22a49e7b45b66fd7615f2bbf5/dht.go#L751-L779)

The `FIND_NODE` handler passes the sender in as `from`:

[handlers.go L264-L272](https://github.com/libp2p/go-libp2p-kad-dht/blob/cb9639642a8d54f22a49e7b45b66fd7615f2bbf5/handlers.go#L264-L272)

go's comment calls this behavior "per spec", but the spec text above does not actually require it. So it is an established convention across implementations, not a written rule.

## Question

Should nim follow go and exclude the sender (`stream.peerId`) from `closerPeers` in `handleFindNode` and `handleGetProviders`, by threading it into `findClosestPeers`? It matches go and removes a hang for clients that do not filter out their own id. The cost is a little extra plumbing, and it arguably compensates for a client bug rather than a spec violation.

gmelodie:
## Summary

`findClosestPeers` built the `closerPeers` list for `FIND_NODE`, `GET_VALUE` and `GET_PROVIDERS` replies while filtering out only self, never the sender. When the sender was in our routing table and landed among the `k` closest to the target key, we handed it back its own peer id.

That is harmless against a client that filters its own id out, but it hung interop against py-libp2p (v0.6.0): on receiving its own id, py-libp2p opened a stream back to itself, sent `FIND_NODE`, and blocked waiting for a reply that a node in client mode never sends, stalling the whole lookup until the test timed out.

This PR excludes the requester as well as self, following go-libp2p's `closestPeersToQuery`. `findClosestPeers` takes the sender as a `requester` argument and the three handlers thread `stream.peerId` into it. It over-fetches from the routing table by the number of excluded keys and truncates back to `replication`, so dropping the sender does not shrink an otherwise full reply.

`FIND_NODE` additionally prepends the target itself when we know its addresses, which closes a real gap: a peer in client mode is in nobody's routing table, so `findPeer` could never resolve one. The target is looked up in the address book rather than the routing table, and is skipped when it is already the closest entry so it is not duplicated.

## Affected Areas

- [x] Peer Management / Discovery — which routing-table peers a node hands back in a reply, and whether a client-mode peer can be resolved at all.

- [x] Protocol Logic — `closerPeers` construction in `handleFindNode` (`find.nim`), `handleGetValue` (`get.nim`) and `handleGetProviders` (`provider.nim`).

The behavior change is confined to kad-dht reply construction and only removes an entry that a correct client already discards, so no downstream integration branches were needed:

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

`findClosestPeers*` is exported and gains a required `requester: PeerId` parameter — a source-breaking signature change for anyone calling it directly. Every in-tree caller is a DHT handler, and no tests referenced it. There is no wire-format change.

Behaviorally: a node no longer receives itself in `closerPeers`, and `FIND_NODE` for a client-mode peer whose addresses we know now resolves instead of returning nothing.

## Risk Assessment

- Backward compatibility: wire-compatible. Interop improves rather than regresses, since the removed entry is one every correct client already discards.
- Network behavior: replies still carry up to `replication` closest peers because the routing-table lookup over-fetches before filtering. `FIND_NODE` can return one extra entry when the target is prepended.
- Performance: one extra routing-table candidate sorted per reply, and one address-book lookup per `FIND_NODE`. The exclusion set is a stack array. Negligible.
- Security: none identified. The prepended target is only ever a peer we already hold addresses for.

## References

- Closes #2848
- go-libp2p `closestPeersToQuery`: https://github.com/libp2p/go-libp2p-kad-dht/blob/cb9639642a8d54f22a49e7b45b66fd7615f2bbf5/dht.go#L751-L779
- kad-dht spec, `FIND_NODE` / `GET_PROVIDERS` `closerPeers`: https://github.com/libp2p/specs/blob/6b6203ee6f62938ce67efdb33498173f475851c0/kad-dht/README.md?plain=1#L490-L505
- py-libp2p interop hang: https://github.com/libp2p/unified-testing/pull/119

## Additional Notes

The issue asked about `handleFindNode` and `handleGetProviders`. `handleGetValue` builds `closerPeers` the same way and go excludes the sender there too, so it is included for consistency. The target prepend is `FIND_NODE`-only, matching go.

**Deliberate divergence from go:** go prepends the target even when the target *is* the requester, and its comment justifies this as "per spec: FIND_PEER has a special exception where the target peer MUST be included". The spec says no such thing — it contains no `MUST` at all, and `FIND_NODE` states only that `closerPeers` is "the `k` closest `Peer`s". Adopting that half would re-arm the very hang this PR fixes: py-libp2p's `refresh_routing_table` performs a lookup for its own id, and v0.6.0 only filters self out of peerstore-seeded candidates, not out of peers received from a reply. It would dial itself and stall again. So the target is not prepended when it is the requester; a peer wanting its own addresses can use identify. "Find node for own PeerID excludes the requester" guards this.

nim is itself immune to that class of bug from either direction — `sortedShortlist` (`find.nim`) skips self when selecting peers to query, and `RoutingTable.insert` refuses self — but that is no reason to emit it to others.

### DIFF
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -515,6 +515,20 @@ proc findClosestPeers*(kad: KadDHT, target: Key, requester: PeerId): seq[Peer] =
     closestPeerKeys[0 ..< min(kad.config.replication, closestPeerKeys.len)]
   )
 
+proc findNodeCloserPeers(kad: KadDHT, target: Key, requester: PeerId): seq[Peer] =
+  ## Also returns the target itself, which keeps client-mode peers resolvable.
+  let closest = kad.findClosestPeers(target, requester)
+  if target == requester.toKey():
+    return closest
+
+  let targetPeer = target.toPeer(kad.switch).valueOr:
+    return closest
+
+  if closest.len > 0 and closest[0].id == targetPeer.id:
+    return closest
+
+  return @[targetPeer] & closest
+
 method handleFindNode*(
     kad: KadDHT, stream: Stream, msg: Message
 ) {.base, async: (raises: [CancelledError]).} =
@@ -524,7 +538,7 @@ method handleFindNode*(
 
   let response = Message(
     msgType: Opt.some(MessageType.findNode),
-    closerPeers: kad.findClosestPeers(msgKey, stream.peerId),
+    closerPeers: kad.findNodeCloserPeers(msgKey, stream.peerId),
   )
   let encoded = response.encode(kad.config.hideConnectionStatus)
   kad_message_bytes_sent.inc(encoded.len.int64, labelValues = [$MessageType.findNode])
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -504,12 +504,16 @@ proc findPeer*(
 
   return ok(PeerInfo(peerId: target, addrs: kad.switch.peerStore[AddressBook][target]))
 
-proc findClosestPeers*(kad: KadDHT, target: Key): seq[Peer] =
-  let closestPeerKeys = kad.rtable.findClosest(target, kad.config.replication).filterIt(
-      it != kad.switch.peerInfo.peerId.toKey()
-    )
+proc findClosestPeers*(kad: KadDHT, target: Key, requester: PeerId): seq[Peer] =
+  ## Over-fetches by `excluded.len` so dropping self and `requester` still fills the reply.
+  let excluded = [kad.switch.peerInfo.peerId.toKey(), requester.toKey()]
+  let closestPeerKeys = kad.rtable
+    .findClosest(target, kad.config.replication + excluded.len)
+    .filterIt(it notin excluded)
 
-  return kad.switch.toPeers(closestPeerKeys)
+  return kad.switch.toPeers(
+    closestPeerKeys[0 ..< min(kad.config.replication, closestPeerKeys.len)]
+  )
 
 method handleFindNode*(
     kad: KadDHT, stream: Stream, msg: Message
@@ -519,7 +523,8 @@ method handleFindNode*(
     return
 
   let response = Message(
-    msgType: Opt.some(MessageType.findNode), closerPeers: kad.findClosestPeers(msgKey)
+    msgType: Opt.some(MessageType.findNode),
+    closerPeers: kad.findClosestPeers(msgKey, stream.peerId),
   )
   let encoded = response.encode(kad.config.hideConnectionStatus)
   kad_message_bytes_sent.inc(encoded.len.int64, labelValues = [$MessageType.findNode])
--- a/libp2p/protocols/kademlia/get.nim
+++ b/libp2p/protocols/kademlia/get.nim
@@ -177,7 +177,7 @@ method handleGetValue*(
     let response = Message(
       msgType: Opt.some(MessageType.getValue),
       key: Opt.some(key),
-      closerPeers: kad.findClosestPeers(key),
+      closerPeers: kad.findClosestPeers(key, stream.peerId),
     )
     let encoded = response.encode(kad.config.hideConnectionStatus)
     kad_message_bytes_sent.inc(encoded.len.int64, labelValues = [$MessageType.getValue])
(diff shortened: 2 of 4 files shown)


rlve (approved):
🚀

libp2p/protocols/kademlia/find.nim
```diff
@@ -504,12 +504,30 @@ proc findPeer*(
...
+    return closest
+
+  let targetPeer = target.toPeer(kad.switch).valueOr:
+    return closest
+
+  if closest.len > 0 and closest[0].id == targetPeer.id:
+    return closest
 
-  return kad.switch.toPeers(closestPeerKeys)
+  return @[targetPeer] & closest
```
richard-ramos, line 530: Something to discuss tho:
- In go-libp2p you can find 'client-mode' targets
- In rust, you can find only 'servers'

In here, it seems we are trying both, trying to return clients like go, but then querying them,, however, this may not work for client-mode peers, because after receiving the target, we will send FIND_NODE to it, and client-mode peer will not reply, so we will remove it after retries, and findPeer will return "peer not found".  

We probably should create a unit test for this scenario to confirm tho, but would like to know your thoughts

also cc: @rlve

richard-ramos (commented):
While reviewing the diff between our impl and rust/go, i noticed a diff that is probably worth looking into later: we take the `closerPeers` that we receive and pass them to `updatePeers`, which insert them in the routing table, the problem is that these peers might not support kademlia at all. Go and Rust delay the insertion to the table until they confirm that they support kademlia.

I'll create a follow up issue

vladopajic (approved):

libp2p/protocols/kademlia/find.nim
gmelodie: In go it seems that an unreachable peer is still considered valid as it can be a client-only node.

By their own words, this is a bug https://github.com/libp2p/go-libp2p-kad-dht/blob/master/routing.go#L678-L687
> The fact that a peer that is returned can be a non-DHT server peer and is not identified as such is a bug.

richard-ramos (approved):

PR #2862, opened 2026-07-17
chore(kad): liveness-gated bucket eviction

### LINKED ISSUE #2852: kad: replace blind oldest-eviction with liveness/usefulness-gated bucket eviction
When a k-bucket is full, `routing_table.nim:75 replaceOldest` evicts the peer with the oldest `lastSeen` unconditionally. There is no liveness check, no usefulness signal, and no latency data anywhere in the stack (`ping()` in `ping.nim` exists but is never wired into eviction; peerstore has no latency EWMA).

This can evict a reliable, recently-useful close peer in favor of a fresh, slow, or already-dead one.

go-libp2p (`go-libp2p-kbucket/table.go`) instead:
- tracks `AddedAt`, `LastUsefulAt`, `LastSuccessfulOutboundQueryAt` per peer, with a `usefulnessGracePeriod` for new peers;
- only swaps peers currently marked `isReplaceable`; query-sourced peers refresh `LastUsefulAt` and are retained;
- rejects insertion when `LatencyEWMA(p) > maxLatency`.

Proposed work:
- Add `lastUsefulAt` (and optionally `addedAt`) to `NodeEntry` (`types.nim`), refreshed when a peer answers a query.
- On a full bucket, prefer evicting a peer past the usefulness grace period; optionally `ping()` the eviction candidate and only evict on failure.
- (Stretch) record RTT per RPC and gate insertion on a latency threshold.

Impact: routing-table quality and lookup reliability. High.

gmelodie:
## Summary

When a k-bucket is full, the routing table used to evict the peer with the oldest `lastSeen` unconditionally (`routing_table.nim` `replaceOldest`). There was no liveness or usefulness signal anywhere in the eviction path, so a reliable, recently-useful close peer could be dropped in favour of a fresh, slow, or already-dead one.

This PR replaces blind oldest-eviction with a usefulness/liveness-gated policy, mirroring go-libp2p's `isReplaceable` semantics: peers that recently answered a query are retained, and a full bucket of still-useful peers rejects the newcomer rather than evicting a good peer.

Closes #2852.

## Affected Areas

- [x] Protocol Logic
  

Specifics:
- `types.nim`: `NodeEntry` gains `addedAt: Moment` and `lastUsefulAt: Opt[Moment]`; new `DefaultUsefulnessGracePeriod` (1h) and a configurable `usefulnessGracePeriod` on `RoutingTableConfig`.
- `routing_table.nim`: new `isReplaceable` predicate (a peer is replaceable only once it has spent longer than the grace period in the table without answering a query); eviction (`replaceOldest` -> `replaceReplaceable`) evicts the least-recently-seen peer among replaceable ones, and rejects the newcomer when every peer is still fresh/useful; new `markUseful` refreshes `lastUsefulAt`/`lastSeen`.
- `find.nim`: every successful query reply in `lookOnce` calls `rtable.markUseful(peerId)`, wiring the usefulness signal into eviction.

N/A — internal Kademlia routing-table behavior change; no public API removed or renamed. New `RoutingTableConfig.usefulnessGracePeriod` defaults to preserve sensible behavior, and new `NodeEntry` fields default to zero/`Opt.none`.

## Impact on Library Users

- No breaking API changes. `RoutingTable.new`/`RoutingTableConfig.new` gain an optional `usefulnessGracePeriod` parameter with a default.
- Behavior change: a full bucket no longer always evicts on insert. A newcomer is rejected when all existing peers are within the usefulness grace period, so routing tables retain proven-useful close peers longer.

## Risk Assessment

- Backward compatibility: additive fields and one optional constructor parameter; existing call sites compile unchanged.
- Network behavior: routing-table churn decreases for healthy tables (fewer evictions of useful peers). During cold start, freshly inserted peers are protected for the grace period, so a full bucket can transiently reject new candidates — acceptable given the grace period is bounded and such peers are unproven. The grace value is a tuning knob; the 1h default is intentionally conservative.
- Performance: eviction is still a single linear scan of the bucket; no extra allocations or network calls on the insert path.
- Security: no new external input handling.

## References

- Parent: #1756
- Prior art: go-libp2p `go-libp2p-kbucket/table.go` (`isReplaceable`, `usefulnessGracePeriod`)

## Additional Notes

Scoped to the primary work in #2852. Deferred to follow-ups, matching the issue's own "optionally"/"stretch" framing:
- Network `ping()` of the eviction candidate before evicting — the sync `insert`/`updatePeers` path (also called from the sync `KadDHT.new`) cannot `await`, so wiring a ping requires an async refactor. The usefulness gate already prevents evicting a peer that was live within the grace period.
- RTT/latency-EWMA insertion gate (stretch goal).

### DIFF
--- a/libp2p/protocols/kademlia/routing_table.nim
+++ b/libp2p/protocols/kademlia/routing_table.nim
@@ -77,14 +77,13 @@ proc oldestPeer*(bucket: Bucket): (NodeEntry, int) =
   (oldest, oldestIdx)
 
 func isReplaceable*(entry: NodeEntry, gracePeriod: Duration, now: Moment): bool =
-  ## A peer is replaceable once it has spent longer than `gracePeriod` in the
-  ## table without proving useful. A peer that answered a query within the grace
-  ## period — or that was only just added — is retained.
+  ## Replaceable once past `gracePeriod` in the table without proving useful; a
+  ## peer that answered recently or was just added is retained.
   now - entry.lastUsefulAt.get(entry.addedAt) > gracePeriod
 
 proc replaceableCandidate(bucket: Bucket, gracePeriod: Duration): Opt[int] =
-  ## Least-recently-seen peer that is past the usefulness grace period, i.e. the
-  ## best candidate to evict. `Opt.none` when every peer is still useful/fresh.
+  ## Least-recently-seen peer past the grace period, i.e. the best eviction
+  ## candidate. `Opt.none` when every peer is still useful/fresh.
   let now = Moment.now()
   var candidateIdx = -1
   var oldestSeen: Moment
@@ -145,8 +144,7 @@ proc insert*(rtable: RoutingTable, nodeId: Key): bool =
     bucket.peers.add(NodeEntry(nodeId: nodeId, lastSeen: now, addedAt: now))
     kad_routing_table_insertions.inc()
   else:
-    # A full bucket with no replaceable peer rejects the newcomer rather than
-    # evicting a still-useful one.
+    # Full bucket with no replaceable peer: reject rather than evict a useful one.
     if not bucket.tryReplaceStalePeer(nodeId, rtable.config):
       debug "Cannot insert, no replaceable peer in bucket",
         bucket = idx, nodeId = nodeId
@@ -161,9 +159,8 @@ proc insert*(rtable: RoutingTable, peerId: PeerId): bool =
   insert(rtable, peerId.toKey())
 
 proc markUseful*(rtable: RoutingTable, nodeId: Key) =
-  ## Records that `nodeId` answered a query: refreshes its usefulness and
-  ## last-seen timestamps so it is retained through bucket eviction. No-op when
-  ## the peer is not in the table.
+  ## Records that `nodeId` answered a query, refreshing its usefulness so it
+  ## survives eviction. No-op when the peer is not in the table.
   let idx = rtable.bucketIndex(nodeId)
   if idx >= rtable.buckets.len:
     return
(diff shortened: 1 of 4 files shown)


vladopajic (approved):

SionoiS (approved):
LGTM Nice one!

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -113,12 +139,14 @@ proc insert*(rtable: RoutingTable, nodeId: Key): bool =
...
+    let now = Moment.now()
+    bucket.peers.add(NodeEntry(nodeId: nodeId, lastSeen: now, addedAt: now))
     kad_routing_table_insertions.inc()
   else:
-    # eviction policy: replace oldest key
-    if not bucket.replaceOldest(nodeId, rtable.config.replication):
-      debug "Cannot insert, failed to replace oldest key in bucket",
+    # A full bucket with no replaceable peer rejects the newcomer rather than
+    # evicting a still-useful one.
+    if not bucket.tryReplaceStalePeer(nodeId, rtable.config):
```
richard-ramos, line 148: Something curious that can happen here is that this can put a peer into the shortlist without saving its addresses. 

This is because `updateShortlist` adds `closerPeers` first, but `updatePeers` will only update the peerstore addresses when  `rtable.insert` returns true, so the peer will remain in shortlist without any address to dial, and with the grace period recejtion, we can end up rejecting the new peer when bucket is full yet the shortlist wont be affected.

Makes me wonder if we should either store addresses even if routing-table insert fails, or only add peers to the shortlist after their addresses are saved ...  or if this has really too much of an impact?

gmelodie: To me it's at the very least counter intuitive to have a peer on the shortlist that is not in the routing table

richard-ramos (approved):

Issue #2851, opened 2026-07-17
bug?: do not add unverified closer peers to the Kademlia routing table


richard-ramos:
During an iterative lookup, nim-libp2p adds every peer received in `closerPeers` directly to the routing table:

- [`lookOnce` processes received peers](https://github.com/vacp2p/nim-libp2p/blob/b4fc4de59c2b7625814ab058a89046e7eea684cf/libp2p/protocols/kademlia/find.nim#L416-L424)
- [`updatePeers` calls `rtable.insert`](https://github.com/vacp2p/nim-libp2p/blob/b4fc4de59c2b7625814ab058a89046e7eea684cf/libp2p/protocols/kademlia/find.nim#L323-L343)

A peer received from another node has not yet proved that it supports the Kademlia protocol. It could be:

- A client-mode peer
- A stale peer
- A peer with incorrect addresses
- A peer sent by an untrusted node

Adding it immediately can fill the routing table with peers that cannot answer Kademlia requests. These peers may later be returned to other nodes or cause lookup timeouts.

PR #2850 makes this easier to trigger because `FIND_NODE` can now return a target known only from the address book. That target may be a client and may not support Kademlia.

### Expected behavior

Received peers should first be stored as lookup candidates with temporary addresses. They should enter the routing table only after:

- A successful Kademlia request, or
- Confirmation that they support the Kademlia protocol

### Other implementations

go-libp2p keeps received peers in the peerstore/query state. It adds a peer to the routing table only after a successful Kademlia request:

- [[Process received peers as query candidates](https://github.com/libp2p/go-libp2p-kad-dht/blob/cb9639642a8d54f22a49e7b45b66fd7615f2bbf5/query.go#L465-L492)](https://github.com/libp2p/go-libp2p-kad-dht/blob/cb9639642a8d54f22a49e7b45b66fd7615f2bbf5/query.go#L465-L492)
- [[Remove failed peers and confirm successful peers](https://github.com/libp2p/go-libp2p-kad-dht/blob/cb9639642a8d54f22a49e7b45b66fd7615f2bbf5/query.go#L427-L451)](https://github.com/libp2p/go-libp2p-kad-dht/blob/cb9639642a8d54f22a49e7b45b66fd7615f2bbf5/query.go#L427-L451)

rust-libp2p also keeps received addresses in the active query instead of inserting them immediately:

[...]

PR #2863, opened 2026-07-17
chore(kad): probe peers before inserting them into the routing table

### LINKED ISSUE #2853: kad: probe peers (lookupCheck) before inserting them into the routing table
`find.nim:323 updatePeers` inserts any peer returned in a FIND_NODE reply directly into the routing table, gated only by the address policy and the IP-diversity filter. Peers that are unreachable or don't speak the DHT protocol still get admitted.

go-libp2p probes a candidate with a FIND_NODE (`lookupCheck`, bounded by a semaphore) and requires `validRTPeer` (peer speaks the DHT protocol) before insertion.

Proposed work:
- Before `rtable.insert` in `updatePeers`, run a bounded FIND_NODE/ping probe against the candidate; insert only on success.
- Bound concurrent probes with a dedicated semaphore so this doesn't amplify RPC load.

Impact: keeps the table populated with actually-usable peers. High.

Related to #2851

gmelodie:
## Summary

`find.nim`'s `updatePeers` used to insert any peer returned in a FIND_NODE reply straight into the routing table, gated only by the address policy and the IP-diversity filter. Peers that are unreachable or don't speak the DHT protocol still got admitted, polluting the table with unusable entries.

This PR gates routing-table admission of network-discovered peers behind a bounded FIND_NODE probe (`lookupCheck`), mirroring go-libp2p's `lookupCheck` + `validRTPeer`. A candidate is only inserted once the probe confirms it is reachable and returns a well-formed `findNode` reply.

To avoid slowing the lookup critical path or blocking request handlers, the probe runs in the background (as go-libp2p does in its `rtPeerLoop`), bounded by a dedicated semaphore. Only routing-table insertion is gated — addresses are still recorded in the AddressBook synchronously, so `findPeer` and in-flight lookups can still dial discovered peers even while a probe is pending or ultimately fails.

## Affected Areas

- [x] Protocol Logic
  

N/A — internal DHT admission behavior with no public API removed and no wire-format change (probes reuse the existing FIND_NODE RPC). No downstream integration branch prepared.

## Impact on Library Users

`KadDHT.updatePeers` is retained as the direct-insert path (used for trusted bootstrap seeds). Network-discovered peers now flow through the new `admitPeers`, which probes before insertion.

New configurable bound `KadDHTLimits.maxConcurrentProbes` (default `DefaultMaxConcurrentProbes = 20`) caps concurrent admission probes.

New public `AdmitHook` callback on `admitPeers(kad, rtable, peerInfos, onAdmit)`, so a caller that owns a routing table can observe successful insertions. Service discovery uses it to keep the `cd_service_table_insertions` counter accurate now that insertion is asynchronous.

Behavior change: a discovered peer now appears in the routing table asynchronously, once its background probe succeeds, rather than synchronously at the end of a lookup. The AddressBook is still populated synchronously (unconditionally, for any address-policy-approved peer), so peer-address resolution is unchanged.

Behavior change: admission is best-effort under load. Probes are never queued — a candidate that finds no free probe slot is dropped and re-considered the next time a reply names it. This keeps the in-flight probe table bounded by `maxConcurrentProbes` rather than growing with inbound request rate.

Performance: a newly-discovered candidate incurs at most one FIND_NODE probe per target routing table, deduped on `(routing table, peer)`. Service discovery admits into both the main and the per-service table, so a candidate discovered there can be probed once for each. Probing is off the lookup critical path. A new bucket-local `RoutingTable.contains` replaces a whole-table `allKeys()` scan on the admission hot path.

## Risk Assessment

Backward compatibility: routing-table population is now eventually-consistent. Any caller/test asserting routing-table membership immediately after a lookup must poll instead — the kad and service-discovery tests were updated to `checkUntilTimeout` accordingly. Bootstrap seeds are still inserted directly, so bootstrapping is unaffected.

New outbound-dial surface: because admission now probes candidates, the node actively dials the addresses advertised in FIND_NODE / GET / GET_PROVIDERS / ADD_PROVIDER replies — addresses that are attacker-controllable. This is inherent to `lookupCheck` (go-libp2p has the same property) and is bounded by `maxConcurrentProbes`, per-`(table, peer)` dedup, and drop-on-saturation, but operators concerned about connection amplification should deploy with a restrictive `addressPolicy` (e.g. `publicRoutableAddressPolicy`) so private/loopback targets are filtered before a probe is issued.

Memory: the in-flight probe table cannot outgrow `maxConcurrentProbes`, since a slot is acquired before a probe is spawned rather than awaited inside it. In-flight probes count toward the IP-diversity caps, so a single reply naming many peer ids on one address cannot fan out dials to it.

Shutdown: `stop` cancels and awaits in-flight probes, draining in a loop so a handler racing shutdown cannot orphan a probe.

Security: reduces routing-table pollution by unreachable or non-DHT peers; rejected probes are logged at `trace` for observability.

## References

- Closes #2853
- Parent #1756, related #2851

## Additional Notes

Verified in-sandbox: `test_find`, `test_bootstrap`, `test_get`, `test_get_providers`, `test_add_provider`, `test_put`, `test_limits`, `test_lookup`, `test_routing_table`, `test_private_addr_filter`, and the service-discovery suites `test_advertiser`, `test_discoverer`, `test_registrar`, `test_routing_table_manager` plus all seven `service_discovery/component` suites.

Two new tests cover the behaviour this PR exists for: `test_find` asserts a peer whose probe fails stays out of the routing table while remaining in the AddressBook, and `test_limits` asserts `admitPeers` spawns at most `maxConcurrentProbes` probes and drops the rest instead of queueing them.


libp2p/protocols/kademlia/types.nim
```diff
@@ -362,6 +364,11 @@ type KadDHTLimits* = object
   maxConcurrentRpcs*: int
     ## Maximum number of in-flight outbound RPCs (find/get/put/provider)
     ## across the whole node. Excess calls wait on a shared semaphore.
+  maxConcurrentProbes*: int
```
richard-ramos: Should be added to `KadDHTLimits` too

libp2p/protocols/kademlia.nim
```diff
@@ -151,6 +154,9 @@ method stop*(kad: KadDHT) {.async: (raises: []).} =
 
   kad.started = false
 
+  let admissionProbes = move kad.admissionProbes
+  await noCancel admissionProbes.values.toSeq().cancelAndWait()
```
richard-ramos: This should probably be moved to line 171, since the loops after this might still be 'producing' more admission probes. Also, maybe those cancelSoon should be changed to cancelAndWait?

libp2p/protocols/kademlia/find.nim
```diff
@@ -353,6 +392,85 @@ proc updatePeers*(kad: KadDHT, peers: seq[(PeerId, seq[MultiAddress])]) {.raises
...
+      continue
+    # Record addresses up front, independent of the probe outcome, so that an
+    # in-flight lookup can still dial the peer and ``findPeer`` can resolve it
+    # even while its admission probe is pending or ultimately fails.
+    addressBook.extend(p.peerId, addrs, AddressConfidence.Low)
+    if p.peerId.toKey() in rtable:
+      # Already admitted: refresh recency without re-probing.
+      discard rtable.insert(p.peerId)
+      continue
+    if kad.admissionProbes.hasKey(p.peerId):
```
richard-ramos: Should the admissionProbes be done per `peerId` and `rtable` too? because otherwise if you discover a peer for one rtable, and another for a service specific rtable, the second one will be skipped

libp2p/protocols/kademlia/find.nim
```diff
@@ -353,6 +392,85 @@ proc updatePeers*(kad: KadDHT, peers: seq[(PeerId, seq[MultiAddress])]) {.raises
...
+  let reply = await kad.dispatchFindNode(peerId, peerId.toKey(), Opt.some(addrs))
+  reply.isOk()
+
+proc admitPeer(
+    kad: KadDHT, rtable: RoutingTable, peerId: PeerId, addrs: seq[MultiAddress]
+) {.async: (raises: []).} =
+  ## Probe ``peerId`` under the dedicated probe semaphore and insert it into
+  ## ``rtable`` only if the probe succeeds.
+  defer:
+    kad.admissionProbes.del(peerId)
```
richard-ramos: Not sure about this. If you acquire probeSem, and for some reason the dial that happens inside lookupCheck fails immediatly,, admitPeer will might finish faster than its future being stored in line 469, thus ending up with a future stored in kad.admissionProbes permanently.


A possible suggestion to fix would be to remove this defer, and replacel ine 469 by something like this
```nim

let admitFut = kad.admitPeer(rtable, p.peerId, addrs)
kad.admissionProbes[peerId] = admitFut

if admitFut.finished():
  kad.admissionProbes.del(peerId)
else:
  admitFut.addCallback proc(udata: pointer) {.raises: [].} =
    if kad.admissionProbes.getOrDefault(peerId) == admitFut:
      kad.admissionProbes.del(peerId)
```

libp2p/protocols/kademlia/find.nim
```diff
@@ -353,6 +392,85 @@ proc updatePeers*(kad: KadDHT, peers: seq[(PeerId, seq[MultiAddress])]) {.raises
...
+    return
+  defer:
+    try:
+      kad.probeSem.release()
+    except AsyncSemaphoreError:
+      raiseAssert "probeSem released without acquire"
+
+  let reachable =
+    try:
+      await kad.lookupCheck(peerId, addrs)
```
richard-ramos, line 438: We should probably timeout this operation with kad.config.timeout

libp2p/protocols/kademlia/find.nim
```diff
@@ -353,6 +392,85 @@ proc updatePeers*(kad: KadDHT, peers: seq[(PeerId, seq[MultiAddress])]) {.raises
...
+    if p.peerId == selfPid:
+      continue
+    let addrs = kad.switch.admissibleAddrs(
+      kad.config.addressPolicy,
+      rtable,
+      p,
+      kad.config.limits.maxPeersPerIp,
+      kad.config.limits.maxPeersPerIpv4Subnet,
+      kad.config.limits.maxPeersPerIpv6Subnet,
+      kad.admissionProbes.keys.toSeq(),
```
richard-ramos: This seems a bit inefficient... Makes me wonder if we should use a hashset or store ` kad.admissionProbes.keys.toSeq()` once in a variable and maintain it inside this for loop.

.github/workflows/test_multiformat_exts.yml
```diff
@@ -57,6 +57,10 @@ jobs:
         shell: ${{ needs.pick-platform.outputs.shell }}
     steps:
       - name: selected platform
+        # The job-default shell is msys2 on Windows, but msys2 is only installed
+        # later by "Setup Nim"; use bash here so this pre-setup step never runs
+        # under an unavailable shell.
```
richard-ramos: wow that's a lotta words

libp2p/protocols/kademlia/find.nim
```diff
@@ -255,6 +255,23 @@ proc sharesSubnet64(addrs: seq[Ipv6Address], subnet: Ipv6Subnet64): bool =
...
 
+proc diversityPeers(
+    rtable: RoutingTable, pending: seq[PeerId]
+): seq[PeerId] {.raises: [].} =
+  ## Peers whose addresses count against the IP-diversity caps: the ones already
+  ## in the routing table plus the ones with an admission probe in flight. The
+  ## latter close a flooding hole — a single reply can list many distinct peer
+  ## ids all pointing at one victim address, and without counting in-flight
+  ## probes they would all pass the cap (none are in the table yet) and be
+  ## dialed at once.
```
richard-ramos: wow that's a lotta words

libp2p/protocols/kademlia/find.nim
```diff
@@ -320,6 +334,30 @@ proc hasIpDiversity(
...
+    p: PeerInfo,
+    maxPeersPerIp: int,
+    maxPeersPerIpv4Subnet: int,
+    maxPeersPerIpv6Subnet: int,
+    pending: seq[PeerId] = @[],
+): seq[MultiAddress] {.raises: [].} =
+  ## Addresses that survive the address policy and IP-diversity filter, or an
+  ## empty seq when the peer is not admissible to the routing table. ``pending``
+  ## lists peers with an admission probe in flight, counted toward the
+  ## diversity caps so a single reply cannot fan out many probes to one IP.
```
richard-ramos: wow that's a lotta words

libp2p/protocols/kademlia/find.nim
```diff
@@ -329,15 +367,16 @@ proc updatePeers*(
     maxPeersPerIpv4Subnet: int = DefaultMaxPeersPerSubnet,
     maxPeersPerIpv6Subnet: int = DefaultMaxPeersPerSubnet,
 ) {.raises: [].} =
+  ## Direct routing-table admission without probing. Used for trusted seed
+  ## peers (bootstrap). Network-discovered peers should go through
+  ## ``admitPeers`` so they are probed first.
```
richard-ramos: wow that's a lotta words

libp2p/protocols/kademlia/find.nim
```diff
@@ -353,6 +392,85 @@ proc updatePeers*(kad: KadDHT, peers: seq[(PeerId, seq[MultiAddress])]) {.raises
   let peerInfos = peers.mapIt(PeerInfo(peerId: it[0], addrs: it[1]))
   kad.updatePeers(peerInfos)
 
+proc lookupCheck(
+    kad: KadDHT, peerId: PeerId, addrs: seq[MultiAddress]
+): Future[bool] {.async: (raises: [CancelledError]).} =
+  ## Probe a candidate before routing-table admission. A FIND_NODE for the
+  ## peer's own key both confirms it is reachable and that it speaks the DHT
+  ## protocol; only a well-formed reply admits the peer.
```
richard-ramos: wow that's a lotta words

libp2p/protocols/kademlia/find.nim
```diff
@@ -353,6 +392,85 @@ proc updatePeers*(kad: KadDHT, peers: seq[(PeerId, seq[MultiAddress])]) {.raises
...
+  discard rtable.insert(peerId)
+
+proc admitPeers*(
+    kad: KadDHT, rtable: RoutingTable, peerInfos: seq[PeerInfo]
+) {.raises: [].} =
+  ## Admit network-discovered peers into ``rtable``. Addresses are recorded
+  ## immediately so lookups can still dial them, but insertion is gated behind
+  ## a bounded FIND_NODE probe (``lookupCheck``) that runs in the background.
+  ## In-flight probes count toward the IP-diversity caps, so a single reply
+  ## cannot make the node fan out many concurrent dials to one address.
```
richard-ramos: wow that's a lotta words

libp2p/protocols/kademlia/find.nim
```diff
@@ -353,6 +392,85 @@ proc updatePeers*(kad: KadDHT, peers: seq[(PeerId, seq[MultiAddress])]) {.raises
...
+      kad.config.limits.maxPeersPerIp,
+      kad.config.limits.maxPeersPerIpv4Subnet,
+      kad.config.limits.maxPeersPerIpv6Subnet,
+      kad.admissionProbes.keys.toSeq(),
+    )
+    if addrs.len == 0:
+      continue
+    # Record addresses up front, independent of the probe outcome, so that an
+    # in-flight lookup can still dial the peer and ``findPeer`` can resolve it
+    # even while its admission probe is pending or ultimately fails.
```
richard-ramos: wow that's a lotta words

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -90,7 +90,7 @@ proc processResponse(
     found: var HashSet[Advertisement],
     limit: int,
 ) =
-  disco.updatePeers(response.closerPeers)
+  disco.admitPeers(response.closerPeers)
```
richard-ramos: Ah this is a case where being specific with the routing table instead of using the main one makes sense. This admitPeers will use the main one, yet immediatly in line 95 it will insert it in the service specific rtable.

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -253,7 +253,7 @@ proc advertiseToRegistrar*(
       error "failed to register ad", serviceId, registrar, error
       return
 
-    disco.updatePeers(response.closerPeers)
+    disco.admitPeers(response.closerPeers)
 
     for p in response.closerPeers:
       disco.rtManager.insertPeer(serviceId, p.peerId.toKey())
```
richard-ramos: Similar to prev comment. AdmitPeers is using the main rtable, but in line 259 we insert the peer into a service specific rtable

libp2p/protocols/service_discovery/discoverer.nim
richard-ramos, line 99: should admitPeers be used inside this proc too?

richard-ramos (commented):
Image

https://www.youtube.com/shorts/skz9odeewpc

vladopajic (approved):

SionoiS (commented):
Make a lot of sense overall but I have one question. Why do a findnode call? Could we not call identify?

gmelodie:
> Why do a findnode call? Could we not call identify?

@SionoiS my answer would be "cuz the spec says so", but other than that, what would be the advantage of using identify?

SionoiS:
> "cuz the spec says so"

Say no more! I'm a big fan of THE SPEC. :rofl: In seriousness, where does it says that? I missed that part.

> what would be the advantage of using identify?

Identifying which protocol a peer support is part of what it does, I guess. :shrug:

richard-ramos:
I would no replace the FIND_NODE with Identify. The [spec](https://github.com/libp2p/specs/blob/master/kad-dht/README.md#client-and-server-mode) says the following

> Nodes, both those operating in client and server mode, add another node to their routing table if and only if that node operates in server mode. This distinction allows restricted nodes to utilize the DHT, i.e. query the DHT, without decreasing the quality of the distributed hash table, i.e. without polluting the routing tables.

> Nodes operating in server mode advertise the libp2p Kademlia protocol identifier via the [identify protocol](https://github.com/libp2p/specs/blob/master/identify/README.md). In addition server mode nodes accept incoming streams using the Kademlia protocol identifier.

TLDR: a server node can advertises the Kad protocol via Identify and they need to accept Kad streams. 
This means that Identify would only prove that the node is advertising they support kademlia, but does not prove that the peer can accept kademlia requests

Another consideration to take into account is that not necessarily Identify will be available always. It is optional, and rust-libp2p even has a comment on this regard, with them saying that it is [explicitly decoupled from Identify](https://docs.rs/libp2p/latest/libp2p/kad/index.html#important-discrepancies). The insertion in the routing tables there happens once kad protocol is confirmed

richard-ramos (approved):

PR #2864, opened 2026-07-17
chore(kad): replace round-synchronized lookup with a continuous alpha-concurrency pipeline

### LINKED ISSUE #2854: kad: replace round-synchronized lookup with a continuous alpha-concurrency pipeline
`find.nim:369 lookOnce` dispatches `alpha` peers per round, then blocks on `collectCompleted(timeout)` for the whole batch before starting the next round. A single slow peer stalls the entire round up to the timeout.

go-libp2p's `query.run()` keeps up to `alpha` queries continuously in flight over a channel, refilling a slot as soon as any query returns, so lookup wall-clock tracks throughput rather than the slowest peer per round.

Proposed work:
- Refactor `lookOnce`/`iterativeLookup` so that instead of awaiting a full batch, it maintains `alpha` in-flight dispatches and re-dispatches the next-closest unqueried peer whenever one completes.
- Preserve existing stop conditions and inflight-cancellation bookkeeping.

Impact: lookup latency, especially with straggler peers. High.

gmelodie:
## Summary

`lookOnce` dispatched `alpha` peers per round, then blocked on `collectCompleted(timeout)` for the whole batch before the next round — one slow peer stalled the round up to `timeout`. This reworks the lookup into a continuous pipeline (like go-libp2p's `query.run()`): keep up to `alpha` RPCs in flight and re-dispatch the next-closest unqueried peer the moment any slot frees, so wall-clock tracks throughput rather than the slowest peer per round.

Each dispatch gets its own `timeout` deadline instead of a per-round batch timeout; `awaitProgress` wakes on the first completion or the earliest slot's deadline. The loop is factored into `harvestInflight` / `applyReplies` / `dropDonePeers` / `fillSlots` / `awaitProgress`. Existing stop conditions and inflight-cancellation bookkeeping are preserved, and all in-flight/abandoned RPCs are cancelled via `defer` on every exit path including `CancelledError`.

`service_discovery/random_find.nim` used the internal `lookOnce` in a hand-rolled round loop to fire `getValue` per round; it now uses `iterativeLookup` with an `onReply` that issues `getValue` per response. The unused `inflight` field was removed from `LookupState`.

### Two non-obvious design points

- **Timed-out RPCs are abandoned, not cancelled.** A peer past its deadline frees its slot but its RPC keeps running (its late reply is ignored); it's retried by a fresh `fillSlots` dispatch until `attempts > retries` excludes it. Cancelling on timeout instead would kill cold-start dials that legitimately exceed `timeout` before they deliver, making the query reach the peer fewer than `retries + 1` times. This matches the old round model, which also never cancelled a timed-out RPC mid-flight.
- **`fillSlots` is gated on `not stopCond`.** Once the stop condition trips we stop dispatching new peers but keep looping to drain replies already in flight. Without this, `getValue` with `quorum = 1` would stop at the first reply and shrink its write-back set to one peer. Gating the refill (rather than breaking immediately) reproduces the same terminal peer set the round model produced; stragglers in the drain still time out normally.

## Affected Areas

- [x] Protocol Logic

Kademlia DHT iterative lookup (`libp2p/protocols/kademlia/find.nim`) and the service-discovery random-walk (`libp2p/protocols/service_discovery/random_find.nim`).

## Impact on Library Users

No API change — internal refactor. `iterativeLookup` and the `findNode` / `findPeer` / `getValue` / provider entry points keep their signatures; only the internal `lookOnce` primitive was removed. Behavior: lookups no longer wait for the slowest peer per round, cutting tail latency under straggler peers, while yielding the same terminal peer set.

## Risk Assessment

- No wire-protocol or API change.
- Concurrency is evenly held at up to `alpha` in flight; a slow peer stalls only its own slot for `timeout`. Abandoned-but-live RPCs are bounded by roughly `maxShortlistSize * retries` per lookup and all cancelled at completion.
- Cancellation verified leak-free by the suites' `checkTrackers()` teardown.

## References

Closes #2854
Parent: #1756

## Additional Notes

Verified with the 2.2.6 toolchain across `test_find`, `test_get`, `test_get_providers`, `test_add_provider`, `test_lookup`, `test_bootstrap`, `test_put`, the service-discovery component suites, `test_find_random`, `test_advertiser`, and `test_discoverer` — all green and stable across repeated runs.


vladopajic (approved):

SionoiS (approved):
LGTM except maybe the query/answers "plumbing" could be separated more from the kad focus find node logic.

libp2p/protocols/kademlia/find.nim
```diff
@@ -366,77 +365,146 @@ proc findNodeDispatch*(
...
+proc dropDonePeers(
+    state: LookupState, pending: var seq[Attempt]
+): seq[RpcFuture] {.raises: [].} =
+  ## Remove attempts whose peer is finished with — it responded (no duplicate
+  ## retry) or a closer peer evicted it from the shortlist — and return their
+  ## still-live RPCs so the caller can cancel them.
+  var keep: seq[Attempt]
+  var stale: seq[RpcFuture]
+  for a in pending:
+    if state.responded.hasKey(a.peer) or not state.shortlist.hasKey(a.peer):
```
richard-ramos: Should the futures abandoned not active that already exceeded the number of retries  be added to `stale` so they're canceled?

libp2p/protocols/kademlia/find.nim
```diff
@@ -446,17 +514,30 @@ proc iterativeLookup*(
     onReply: ReplyHandler,
     stopCond: StopCond,
 ): Future[LookupState] {.async: (raises: [CancelledError]).} =
+  ## Drive an iterative lookup as a continuous ``alpha``-concurrency pipeline:
+  ## keep up to ``alpha`` RPCs in flight and re-dispatch the next-closest
+  ## unqueried peer as soon as any slot frees, so wall-clock tracks throughput
+  ## rather than the slowest peer in a synchronized round. A slow peer only
+  ## stalls its own slot for ``timeout`` before it is abandoned and retried.
```
richard-ramos: Maybe make the comment smaller?
i.e.
```
  ## Drive lookup with continuous alpha concurrency instead of synchronized rounds.
  ## Timed-out RPCs free their slot and may be retried; late replies are ignored.
```

libp2p/protocols/kademlia/find.nim
```diff
@@ -445,17 +519,32 @@ proc iterativeLookup*(
...
+  # unwind, otherwise we return while their streams are still closing.
+  defer:
+    await noCancel pending.mapIt(it.fut).cancelAndWait()
+
+  while true:
+    let completed = pending.harvestInflight(Moment.now())
+    await kad.applyReplies(state, rtable, completed, onReply)
+    # `dropDonePeers` already removed these from `pending`, so the `defer` above
+    # no longer covers them: they must be awaited to completion here.
+    await noCancel state.dropDonePeers(pending).cancelAndWait()
```
richard-ramos: `cancelAndWait` is a template, thus using it like this will cause state.dropDonePeers(pending) to be called more than once.

Use this instead
```suggestion
    let dropDonePeersFut = state.dropDonePeers(pending)
    await noCancel dropDonePeersFut.cancelAndWait()
```


Here's a small example that demonstrates the strange behavior
https://play.nim-lang.org/#pasty=uLUxpxmL
```nim
var calls = 0
var pending = @[1, 2, 3]

proc dropDonePeers(): seq[int] =
  inc calls
  echo "dropDonePeers call ", calls, ": pending before = ", pending
  result = pending
  pending.setLen(0)
  echo "dropDonePeers call ", calls, ": pending after  = ", pending

template cancelAndWait(futs: seq[int]) =
  echo "capacity from len = ", futs.len
  for fut in futs:
    echo "canceling ", fut

   
dropDonePeers().cancelAndWait()

# Comment the previous line and uncomment the next two to see how output changes
#let stale = dropDonePeers()
#cancelAndWaitLike(stale)

echo "final pending = ", pending



```

richard-ramos (approved):

PR #2865, opened 2026-07-20
fix(kad): retry peers that respond with an error, not just silent ones

tinniaru3005:
### Problem

In Kademlia lookups, a peer that fully failed to respond (dial error, stream error, decode error) was treated identically to a peer that responded successfully for the purposes of retry eligibility: both got a single entry written into `state.responded`, and `sortedShortlist` excluded any peer with an entry there from being queried again.

Net effect: a peer that answered with an error got zero retries. Meanwhile, a peer that never answered at all (silent, still timing out) never got an entry in responded, so it stayed eligible and was retried up to the full retries budget (5 by default, 6 total attempts).

This is backwards. A peer that responded — even with an error — is proven reachable; the error could be transient. A peer that never answers at all might simply be dead. The peer more worth retrying was getting no chances, and the peer less worth retrying was getting several.

### Fix

A Failed response (dial-level or RPC-level) now falls through to the same attempts > retries check that already gates silent peers — so a peer that responds with an error gets the same retry budget as everyone else, instead of none.

## Affected Areas

- [ ] Gossipsub  
  

- [ ] Transports  
  

- [ ] Protocol Logic

- [ ] Build / Tooling

- [ ] Other  
  

- **Nimbus:**  
  

- **Waku:**  
  

- **Codex:**

### DIFF
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -100,8 +100,7 @@ proc sortedShortlist(
     if pid == selfPid:
       # do not return self
       continue
-    if excludeResponded and state.responded.contains(pid):
-      # already responded, do not query again
+    if excludeResponded and state.responded.getOrDefault(pid) == Success:
       continue
     if state.attempts.getOrDefault(pid, 0) > state.kad.config.retries:
       # depleted retries, do not query again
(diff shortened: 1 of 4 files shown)


SionoiS (approved):
LGTM Thanks!

gmelodie (approved):

richard-ramos (approved):

PR #2870, opened 2026-07-20
fix(service-disco): apply routing-table admission checks to per-service inserts

tinniaru3005:
### Problem
`ServiceRoutingTableManager.insertPeer` inserted any `PeerId`/`Key` into a per-service routing table with zero validation — unlike the main Kad-DHT table, which runs both via `updatePeers` before calling the same underlying `RoutingTable.insert`.

### Fix

`ServiceRoutingTableManager.insertPeer` used to insert any bare `PeerId` into a per-service routing table with no validation — no address-policy filter, no IP-diversity/Sybil check — unlike the main Kad-DHT table's `updatePeers`. Five call sites fed it unverified peers straight from network input (REGISTER/GET_ADS senders, gossiped `closerPeers`, even an ad's self-reported `peerId`).

Changed `insertPeer`'s signature from `(manager, serviceId, peerKey: Key)` to `(disco: ServiceDiscovery, serviceId, peerInfo: PeerInfo)`, and it now runs the same `addressPolicy.filterAddrs` + `hasIpDiversity` checks the main table already enforces (reused from `find.nim`, not reimplemented). Updated all five call sites to supply real `PeerInfo` instead of bare keys.

## Affected Areas

- [ ] Gossipsub  
  

- [ ] Transports  
  

- [ ] Protocol Logic

- [ ] Build / Tooling

- [ ] Other  
  

- **Nimbus:**  
  

- **Waku:**  
  

- **Codex:**

### DIFF
--- a/tests/libp2p/service_discovery/test_routing_table_manager.nim
+++ b/tests/libp2p/service_discovery/test_routing_table_manager.nim
@@ -5,10 +5,12 @@
 
 import chronos, results, sets, tables, sequtils
 import
+  ../../../libp2p/[peerinfo],
   ../../../libp2p/protocols/kademlia,
   ../../../libp2p/protocols/service_discovery/[types, routing_table_manager]
 import ../../tools/[lifecycle, unittest]
 import ../kademlia/[mock_kademlia, utils]
+import ./utils except randomPeerId
 
 proc makeKey(x: byte): Key =
   var buf: array[IdLength, byte]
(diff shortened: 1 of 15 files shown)


tests/libp2p/service_discovery/test_routing_table_manager.nim
```diff
@@ -223,30 +225,45 @@ suite "ServiceRoutingTableManager":
...
-    check manager.addService(
-      serviceId, mainRt, DefaultReplication, DefaultMaxBuckets, Interest
+    check disco.rtManager.addService(
+      serviceId, disco.rtable, DefaultReplication, DefaultMaxBuckets, Interest
     )
 
-    let peerKey = makeKey(2)
-    manager.insertPeer(serviceId, peerKey)
+    let peer = randomPeerId()
+    let peerInfo = PeerInfo(peerId: peer, addrs: @[makeMultiAddress("10.0.0.1")])
```
vladopajic: would be nicer with maker:
```
let peerInfo = makePeerInfo(addr = @[makeMultiAddress("10.0.0.1")])
```

if there is maker like this 
```
func makePeerInfo(peerId: PeerId = randomPeerId(), addrs: seq[MultiAddress] = @[]): PeerInfo =
```
peer id is always random value, addrs can be empty by default

... 

then always use `peerInfo` variable, it should be source of truth for peer id and addresses

```
check peerInfo.peerId.toKey() in table.allKeys()
```

tests/libp2p/service_discovery/test_routing_table_manager.nim
```diff
@@ -400,9 +417,9 @@ suite "ServiceRoutingTableManager - refreshAllTables":
       serviceId, mainRt, DefaultReplication, DefaultMaxBuckets, Interest
     )
 
-    manager.insertPeer(serviceId, selfId)
-
     let table = manager.getTable(serviceId).get()
+    discard table.insert(selfId)
```
vladopajic: `discard` can be easily replaced with `check` ->

```
check table.insert(selfId)
```

it should be much better to have this safety then to discard it.

tests/libp2p/service_discovery/test_routing_table_manager.nim
```diff
@@ -223,30 +225,45 @@ suite "ServiceRoutingTableManager":
...
+    let disco = setupServiceDiscoveryNode()
+    let serviceId = makeServiceId(1)
+
+    check disco.rtManager.addService(
+      serviceId, disco.rtable, DefaultReplication, DefaultMaxBuckets, Interest
+    )
 
-    check peerKey in table.allKeys()
+    let peer = randomPeerId()
+    disco.insertPeer(serviceId, PeerInfo(peerId: peer, addrs: @[]))
```
vladopajic: would be nice for `insertPeer` to return value `bool` as a result if peer is inserted or not.

tests/libp2p/service_discovery/test_routing_table_manager.nim
```diff
@@ -223,30 +225,45 @@ suite "ServiceRoutingTableManager":
...
+
+    let table = disco.rtManager.getTable(serviceId).get()
+
+    check peer.toKey() notin table.allKeys()
```
vladopajic: 1. there should be no empty lines here (same in other parts of pr), both instructs should be in the same logical group 

2. as a matter of fact, there should be helper proc:
   ```nim
     check not disco.hasPeerInServiceTable(serviceId, peerInfo.peerId)
   ```

vladopajic (approved):

gmelodie (approved):

SionoiS (approved):
LGTM

tests/libp2p/service_discovery/test_routing_table_manager.nim
tinniaru3005: Yes that would be better.

tests/libp2p/service_discovery/test_routing_table_manager.nim
tinniaru3005: Yup. Good point

tests/libp2p/service_discovery/test_routing_table_manager.nim
tinniaru3005: Yes that would be better

tests/libp2p/service_discovery/test_routing_table_manager.nim
tinniaru3005: Done

PR #2871, opened 2026-07-20
build(deps): replace custom pins with Nimble resolver setup

richard-ramos:
## Summary

Aligns repository dependency setup with Nimble's resolver-based workflow for a library.

This PR removes the custom `.pinned` / `tools/pinner.nim` flow and stops normal root/test CI from installing through committed root/test lockfiles. Library and test jobs now resolve dependencies per tested Nim version with local dependency mode, `--noLockfile`, and `--resolver:minver`; the scheduled latest-deps workflow uses `--resolver:maxver` to catch newest dependency breakage.

The setup path is now:

```sh
make        # same as make setup
make test
```

Lockfile-backed setup remains for app-like/build artifacts where it is still the representative workflow, notably `cbind` and Nix dependency generation.

## Affected Areas

- [x] Build / Tooling
  CI dependency resolution, Makefile setup/test targets, Nimble/Nix dependency generation, and C bindings setup.

- [x] Other
  Contributor docs, setup guidance, simulation/cbind setup notes, and interop Docker dependency setup.

## Impact on Library Users

No runtime API or protocol changes. Consumers still resolve `libp2p` through their own Nimble environment. Contributors should use `make` / `make setup` for local dependencies instead of the removed `.pinned` workflow.

## Risk Assessment

The main risk is dependency-resolution drift between PR CI and users. This PR intentionally tests declared lower bounds in PR CI with `--resolver:minver` and keeps a scheduled max-version workflow for newest dependency breakage. App-like artifacts that need reproducible builds keep explicit lockfile handling.


richard-ramos:
@gmelodie , @vladopajic 
if you got some time tomorrow or later in the week, pls take this PR for a test drive :)

nix/deps.nix
```diff
@@ -31,24 +38,24 @@
...
+    rev = "b66168735d6f3841c5239c3169d3fe5fe98b1257";
+    sha256 = "10n71vfa6klzd9dmal96jy0hiqk04gaj8wc9g91z6fclryf0yq92";
     fetchSubmodules = true;
   };
 
-  stew = pkgs.fetchgit {
-    url = "https://github.com/status-im/nim-stew";
-    rev = "4382b18f04b3c43c8409bfcd6b62063773b2bbaa";
-    sha256 = "0mx9g5m636h3sk5pllcpylk51brf7lx91izx3gc23k3ih3hrxyk2";
+  testutils = pkgs.fetchgit {
```
vladopajic, line 48: `testutils` is it really needed? i some previews commits i realize that this was in pinned file of nimlibp2p but it was not used.

vladopajic (approved):

vladopajic:
added `run-daily-ci` label

nix/deps.nix
richard-ramos: Will take a look!

richard-ramos: We probably didn't need to have it in the pinned file, but in nix file it appears due to being a transitive dep from chronicles https://github.com/status-im/nim-chronicles/blob/master/chronicles.nimble#L16

gmelodie (commented):
Looks good, a few places where `pinned` is still mentioned I'd like to see purged tho
```bash
tools/check_code_examples.nim
25:  # Pass the repo root (so examples can `import libp2p/...`) and each pinned

simulation/Dockerfile
8:# Install root libp2p deps, then the simulation-specific pinned deps.

Makefile
3:        install_pinned lock gen_multicodec format clean-nim
143:install_pinned:
```

gmelodie (approved):

.github/workflows/cbindings.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
     env:
-      NIMBLE_COMMIT: 9207e8b2bbdf66b5a4d1020214cff44d2d30df92 # v0.20.1
+      NIMBLE_COMMIT: 08be587ad79f1692d5adcb03ce0a13dfda5d8818 # v0.24.0
```
moigagoo: The latest Nimble is 0.24.1. It has important fixes compared to 0.24.0, better use it.

.github/workflows/cbindings.yml
```diff
@@ -44,34 +44,30 @@ jobs:
...
-          nimble setup
+          nimble --lockFile:nix/libp2p.lock -l setup -y
           cd cbind
-          nimble install_pinned
-          nimble setup
-          nimble buildffi
-          nimble genbindings_c
-          nimble genbindings_cddl
+          nimble -l setup -y
+          nimble -l buildffi
```
moigagoo: I'm not sure `-l` does anything for custom tasks. If you run the task after `nimble -l setup -y`, the deps are installed locally and because your have config.nims and nimble.paths, those local deps are used in tasks.

.github/workflows/cbindings.yml
```diff
@@ -44,34 +44,30 @@ jobs:
...
+          nimble -l genbindings_c
+          nimble -l genbindings_cddl
 
       - name: Build and run C bindings examples
         run: |
           cd cbind
-          nimble examples
+          nimble -l examples
```
moigagoo: Same as above. Is there any difference between `nimble examples` and `nimble -l examples`?

.gitignore
```diff
@@ -16,6 +16,8 @@ tests/pubsub/testgossipsub
 examples/*.md
 nimble.develop
 nimble.paths
+/nimble.lock
```
moigagoo, line 19: I thought .lock files are supposed to be tracked so that all developers could have identical environments locally.

moigagoo (approved):

.gitignore
richard-ramos: I was trying to follow @arnetheduck's suggestion from discord:
> don't use lock files  - instead, use nim --requires:"nim ~= 2.0 test in CI - this is more representative of how your users see libp2p - when you use a lock file in CI for a library, you don't really experience what your users experience (they don't use your lock file)

Maybe i missunderstood?

.github/workflows/cbindings.yml
richard-ramos: This was search and replace mistake.

.gitignore
moigagoo: This is how I see it.

If you're using lock files to pin the deps, they shouldn't be ignored as you want them shared and synced between the devs.

If you're not using lock files, you don't want them ignored either because that way you could have an invisible local lock file that would produce a different environment from what your teammate without it would get. I.e. you want to prevent the lock file from being used, not just ignore it.

So either way you don't want the lock file to be undetected.

arnetheduck: whether it's in gitignore or not wasn't part of my comment, I'm .. agnostic on this point as of today at least :)

.github/workflows/daily_test_all_latest_deps_matrix.yml
```diff
@@ -167,37 +131,33 @@ jobs:
...
-          nimble install -y $DEPS
-          mkdir tests/nimbledeps
-          echo "cd tests && nimble install -y $TEST_DEPS"
-          (cd tests && nimble install -y $TEST_DEPS)
-
-      - name: Build prerequisites
-        run: make nimble.paths tests/nimble.paths
+
+          echo "=== Setting up latest allowed dependencies ==="
+          rm -rf nimbledeps tests/nimbledeps nimble.paths tests/nimble.paths
```
arnetheduck, line 152: what's this rm doing here? and why is `--noLockfile` needed if there's no lock file to begin with?

arnetheduck: aaah, there is a lock file still 👍

nix/libp2p.lock
```diff
@@ -0,0 +1,326 @@
+{
```
arnetheduck, line 1: btw, nuance that I left out earlier: a shared library (.so/.dll) is de facto an "application" when it comes to "locking" logic since it no longer participates in version resolution but rather is shipped as a fixed / compiled black-box unit

Makefile
```diff
@@ -139,15 +102,11 @@ test_integration: nimble.paths tests/nimble.paths
...
-
-pin:
-	nimble pin
+setup:
+	$(MAKE) nimble.paths tests/nimble.paths
 
-unpin:
-	nimble unpin
+lock: nix/libp2p.lock
+	$(MAKE) -C cbind nimble.lock
```
arnetheduck, line 109: fwiw, when you're shipping c bindings as a shared library, there's no particular reason to support multiple _nim_ versions so a simplification of testing and shipping would be to test c bindings only with _one_ nim version, similar to how applications are tested .. others that want to use the C bindings can use whatever (nim) version they like since the shared library code is isolated as a unit from the user except the well-defined C abi.

Makefile
```diff
@@ -62,45 +56,14 @@ cbind:
...
-	      path="$$pkg"; \
-	    fi; \
-	    printf 'switch("path", "%s")\n' "$$path" >> $@; \
-	  done; \
-	done
+nimble.paths: libp2p.nimble
+	nimble --noLockfile --requires:"nim == $(NIM_VERSION)" --resolver:minver -l setup -y
+
+tests/nimble.paths: tests/tests.nimble libp2p.nimble
+	cd tests && nimble --noLockfile --requires:"nim == $(NIM_VERSION)" --resolver:minver -l setup -y
```
arnetheduck, line 66: is there still a lock file in tests? else `--noLockfile` should not be needed here..

PR #2872, opened 2026-07-21
chore(service-disco): add timeouts and refactor for concurency

SionoiS:
Small PR that add some timeouts and refactor concurrency when dealing with multiple async procs.


libp2p/protocols/kademlia/provider.nim
```diff
@@ -213,11 +213,21 @@ proc startProviding*(kad: KadDHT, c: Cid) {.async: (raises: [CancelledError]).}
 proc stopProviding*(kad: KadDHT, c: Cid) =
   kad.providerManager.providedKeys.del(c.toKey())
 
+proc republishProvidedKeys(kad: KadDHT) {.async: (raises: [CancelledError]).} =
+  let providedKeys = kad.providerManager.providedKeys.provided
+
+  var futs: seq[Future[void]]
+  for key in providedKeys.keys():
+    let fut = kad.addProvider(key)
+    futs.add(fut)
```
gmelodie:
```suggestion
  var futs = newSeqOfCap[Future[void]](providedKeys.len)
  for key in providedKeys.keys():
    futs.add(kad.addProvider(key))
```

gmelodie (approved):

libp2p/protocols/kademlia.nim
```diff
@@ -23,23 +23,27 @@ proc refreshTable*(
...
     let target = rtable.refreshTarget(i, kad.rng).valueOr:
       trace "No refresh target for bucket", bucket = i
       continue
-    discard await kad.findNode(target, rtable)
+
+    targets.add(target)
+
+  let futs = targets.mapIt(kad.findNode(it, rtable))
+
+  await allFutures(futs)
```
richard-ramos: In https://github.com/status-im/nim-chronos/blob/55adad164db35f7f6973177c5b5ff48fcd6b6450/chronos/internal/asyncfutures.nim#L969 it says that the awaited `futs` wont be cancelled. so even if using withTimeout, the children futures wont be cancelled once timeout happens, so maybe code should look like this:

```suggestion
  let futs = targets.mapIt(kad.findNode(it, rtable))
  try:
    await allFutures(futs)
  except CancelledError as exc:
    await noCancel futs.cancelAndWait()
    raise exc
```

This should be applied also in changes in republishProvidedKeys and refreshAllTables.  (might need to `import ../../utils/future` tho)

Here's a small vibecoded unittest that demonstrates the behavior:
```nim
import std/unittest
import chronos

test "cancelling allFutures does not cancel children":
  proc scenario(): Future[bool] {.async: (raises: [CancelledError]).} =
    let child = newFuture[void]("child")
    let aggregate = allFutures(child)

    echo "before timeout:"
    echo "  aggregate cancelled: ", aggregate.cancelled()
    echo "  child cancelled:     ", child.cancelled()
    echo "  child finished:      ", child.finished()

    discard await aggregate.withTimeout(10.millis)

    echo "after timeout:"
    echo "  aggregate cancelled: ", aggregate.cancelled()
    echo "  child cancelled:     ", child.cancelled()
    echo "  child finished:      ", child.finished()

    return aggregate.cancelled() and not child.cancelled() and not child.finished()

  check waitFor scenario()

```
Output:
```
before timeout:
  aggregate cancelled: false
  child cancelled:     false
  child finished:      false
after timeout:
  aggregate cancelled: true
  child cancelled:     false
  child finished:      false
[OK] cancelling allFutures does not cancel children
```

SionoiS: I'm baffled, why would a timeout not cancel the futures? It's good to know thanks!

richard-ramos (approved):

PR #2874, opened 2026-07-22
ci: fix dependnecies cache keys

vladopajic:
since `nim-libp2p` is monorepo of different subprojects, there needs to be different deps cache, storing everything under same key is not correct.

### DIFF
--- a/.github/workflows/cbindings.yml
+++ b/.github/workflows/cbindings.yml
@@ -47,7 +47,7 @@ jobs:
           path: nimbledeps
           # Keep this in sync with ci.yml's cache key so entries are
           # reusable. See ci.yml for the rationale on os/cpu/cc scoping.
-          key: nimbledeps-linux-amd64-gcc-${{ hashFiles('.pinned') }}
+          key: nimbledeps-root-linux-amd64-gcc-${{ hashFiles('.pinned') }}
 
       - name: Install deps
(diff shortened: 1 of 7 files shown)


.github/workflows/test_all.yml
```diff
@@ -15,7 +15,7 @@ concurrency:
 jobs:
   test:
     env:
-      NIMBLE_COMMIT: a399f502dec7ffcd905c1cf54b13274ad990bada # v0.24.1
+      NIMBLE_COMMIT: 42ef70c2102a942c46f13eb76872326edd525cec # v0.22.3
```
vladopajic, line 18: riveting nimble bump as new nimble has some errors

.github/workflows/cbindings.yml
```diff
@@ -107,7 +107,7 @@ jobs:
         uses: actions/cache@v5
         with:
           path: nimbledeps
```
richard-ramos, line 109: Not introduced in this PR, but... this should be cbind/nimbledeps right?

richard-ramos (approved):

gmelodie (approved):

.github/workflows/cbindings.yml
vladopajic: yes... will have that fixed in following, lets have this one go through because of nimble rivert

PR #2876, opened 2026-07-22
ci: composite actions for dependencies

vladopajic:
adding composite actions for getting dependencies:
- `install_root_deps/action.yml` gets from cache or installs dependencies for root project (nim-libp2p library)
- `install_tests_deps/action.yml` gets from cache or installs dependencies for tests
  -  two step deps `root` + `tests`
- `install_cbind_deps/action.yml` gets from cache or installs dependencies for cbind
  -  two step deps `root` + `cbind`
  
this reduces the strain on workflows were these actions are used. simplifies and localizes the code. 
also makes it easier for us to see and compare logic.
better reuse; as root dependence are reused everywhere, other sub-projects and only their deps.

also fixes issues in comment here: https://github.com/vacp2p/nim-libp2p/pull/2874#discussion_r3630264467

### DIFF
--- a/.github/actions/install_root_deps/action.yml
+++ b/.github/actions/install_root_deps/action.yml
@@ -11,6 +11,9 @@ inputs:
   cc:
     description: Platform cache compiler segment.
     required: true
+  shell:
+    description: Shell used to run Nimble commands.
+    default: bash
 
 runs:
   using: composite
@@ -24,7 +27,7 @@ runs:
 
     - name: Install deps
       if: ${{ steps.deps-cache.outputs.cache-hit != 'true' }}
-      shell: bash
+      shell: ${{ inputs.shell }}
       run: nimble install_pinned
 
     - name: Save dependencies cache
--- a/.github/actions/install_root_deps/action.yml
+++ b/.github/actions/install_root_deps/action.yml
@@ -1,0 +1,35 @@
+name: Install root dependencies
+description: Restore, install, and save root Nimble dependencies.
+
+inputs:
+  os:
+    description: Platform cache OS segment.
+    required: true
+  cpu:
+    description: Platform cache CPU segment.
+    required: true
+  cc:
+    description: Platform cache compiler segment.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Restore deps from cache
+      id: deps-cache
+      uses: actions/cache/restore@v5
+      with:
+        path: nimbledeps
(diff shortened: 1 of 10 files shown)


.github/actions/install_root_deps/action.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Restore deps from cache
+      id: deps-cache
+      uses: actions/cache/restore@v5
+      with:
+        path: nimbledeps
+        key: nimbledeps-root-${{ inputs.os }}-${{ inputs.cpu }}-${{ inputs.cc }}-${{ hashFiles('.pinned') }}
+
+    - name: Install deps
+      if: ${{ steps.deps-cache.outputs.cache-hit != 'true' }}
+      shell: bash
```
richard-ramos: This will force using git bash on windows instead of using msys {0} shell as it was before. Maybe it could be 
```
shell: ${{ inputs.shell }}  
```
(while adding a `shell` input to the action)

.github/workflows/test_all.yml
```diff
@@ -104,18 +104,12 @@ jobs:
...
-            nimbledeps
-            tests/nimbledeps
-          # Scope by os/cpu/cc so cache entries can't leak vendored/compiled
-          # artifacts across mismatched toolchains (linux<->windows,
-          # gcc<->clang). Keep this key in sync with the other workflows that
-          # cache nimbledeps so cache entries are reusable across them.
-          key: nimbledeps-with-tests-${{ matrix.platform.os }}-${{ matrix.platform.cpu }}-${{ matrix.platform.cc }}-${{ hashFiles('.pinned', 'tests/.pinned') }}
+          os: ${{ matrix.platform.os }}
+          cpu: ${{ matrix.platform.cpu }}
+          cc: ${{ matrix.platform.cc }}
```
richard-ramos: Since this action has a `nimble install`, shouldnt lines 107 to 112 be moved to line 155 to ensure that llvm-mingw setup and nasm install have both been done?

vladopajic: that code worked, but i have moved it anyway. not sure if `nimble install` really needs these, or it is compiling step that depends on these?

gmelodie (approved):

.github/workflows/test_all.yml
richard-ramos: Just did some review, and previously we did have some operations in nim-lsquic .nimble file that compiled boringssl assembly files but we moved away from that and use {.compile.} instead. ... i guess moving it would be a 'defensive' action in case any dependency does anything during installation.  :shrug: no strong feelings about it tho

richard-ramos (approved):

PR #2879, opened 2026-07-22
fix(kad): flaky Get providers returns at most k closest peers

### LINKED ISSUE #2877: bug:  Get providers returns at most k closest peers
https://github.com/vacp2p/nim-libp2p/actions/runs/29944367578/job/89005893332 
https://github.com/vacp2p/nim-libp2p/actions/runs/29944367578/job/89005893416

```
  [Test   ]           Get providers returns at most k closest peers
WRN 2026-07-22 18:05:17.234+00:00 Starting kad-dht twice                     topics="kad-dht" tid=8805
WRN 2026-07-22 18:05:17.234+00:00 Starting kad-dht twice                     topics="kad-dht" tid=8805
WRN 2026-07-22 18:05:17.234+00:00 Starting kad-dht twice                     topics="kad-dht" tid=8805
WRN 2026-07-22 18:05:17.234+00:00 Starting kad-dht twice                     topics="kad-dht" tid=8805
WRN 2026-07-22 18:05:17.234+00:00 Starting kad-dht twice                     topics="kad-dht" tid=8805
WRN 2026-07-22 18:05:17.234+00:00 Starting kad-dht twice                     topics="kad-dht" tid=8805
WRN 2026-07-22 18:05:17.234+00:00 Starting kad-dht twice                     topics="kad-dht" tid=8805
=============================================
  /home/runner/work/nim-libp2p/nim-libp2p/tests/test_all 'KadDHT - Get Providers::Get providers returns at most k closest peers'
---------------------------------------------
    /home/runner/work/nim-libp2p/nim-libp2p/tests/libp2p/kademlia/test_get_providers.nim(253, 39): Check failed: response.get().closerPeers.len() == 3

  [FAILED ] (  0.34s) Get providers returns at most k closest peers

```

gmelodie:
## Summary

Test-only fix for the flaky `KadDHT - Get Providers::Get providers returns at most k closest peers` (#2877).

The test built a 7-node hub with `replication = 3` and asserted the hub replies with exactly 3 `closerPeers`. Bucket capacity *is* `k = 3` and the bucket index is the leading-zero count of the XOR distance, so roughly 1 run in 64 all six spokes hash into the same bucket and the hub's routing table ends up holding only 3 peers.

That was harmless until #2862: a full bucket used to evict the least-recently-seen peer, so the table retained the *last* three spokes and never the requester (`kads[1]`, inserted first) — 3 eligible peers, assertion satisfied. With liveness-gated eviction a full bucket rejects the newcomer instead, so the table retains the *first* three, which always includes the requester. `findClosestPeers` filters the requester out, leaving 2 `closerPeers` and a failed check.

A Monte Carlo over uniform 256-bit keys with the same bucket/filter logic gives 0/200000 failures under the old eviction policy and 3043/200000 (1.5%) under the new one, which matches a test that only started flaking after #2862 and on any platform.

The fix removes the dependency on how random peer ids spread over buckets: the test uses two switches and adds `2 * k` peers to the hub with `addPeersWithAddrs`, a new helper in `tests/libp2p/kademlia/utils.nim` that retries ids a full bucket rejects (and registers an address so `toPeers` keeps them), so the routing table grows by exactly the requested count or fails with a `doAssert` naming how many it managed. The test also now asserts the reply carries the *k closest* known peers rather than only counting them.

`populateRoutingTable` is left alone: it deliberately adds peers with no addresses, which several bootstrap tests rely on.

## Affected Areas

- [x] Build / Tooling

  Tests only — `tests/libp2p/kademlia/test_get_providers.nim` and `tests/libp2p/kademlia/utils.nim`. No library code is touched.

N/A — test-only change, no library code affected.

## Impact on Library Users

None. No API or behavior change.

## Risk Assessment

Minimal. The rewritten test exercises the same `handleGetProviders` path and strengthens the assertion; it only replaces a probabilistic setup with a deterministic one. Peers are added after `start()`, and `bucketRefreshTime` defaults to 10 minutes, so nothing dials the synthetic addresses during the test.

The underlying routing-table behavior from #2862 is left as is: rejecting a newcomer when every peer in a full bucket is still within the usefulness grace period is the intended policy, and a table that legitimately holds only `k` peers returning fewer than `k` closer peers is correct.

## References

- Fixes #2877
- Regression surfaced by #2862 (liveness-gated bucket eviction)


tests/libp2p/kademlia/utils.nim
```diff
@@ -194,6 +194,26 @@ proc populateRoutingTable*(kad: KadDHT, count: int) =
   for i in 0 ..< count:
     discard kad.rtable.insert(randomPeerId())
 
+proc addPeersWithAddrs*(kad: KadDHT, count: int) =
+  ## Adds `count` random peers to the routing table, with an address each so they
+  ## survive `toPeers`. Unlike `populateRoutingTable`, rejected ids are retried.
+  const maxAttempts = 1000
+  var added = 0
+  for _ in 0 ..< maxAttempts:
```
vladopajic, line 203: `for _ in 0 ..< count + maxRetries:`
or 
`for _ in 0 ..< count * 2`

if utility is used for any count > 1000 it will fail in `doAssert` below. probably unlikely that this value will be used but still it's very rigid condition

vladopajic (approved):

richard-ramos (approved):

PR #2880, opened 2026-07-23
feat(quic): validate expected peer id during dial

richard-ramos:
## Summary

Updates QUIC transport to use the `nim-lsquic` v0.6.0 per-dial certificate validator when an expected peer ID is provided.

This validates the remote certificate against the expected peer ID during the TLS handshake instead of accepting the QUIC connection first and checking afterward. Dials without an expected peer ID continue using the default certificate verifier.

## Affected Areas

- [x] Transports
  - QUIC dial certificate validation
- [x] Build / Tooling
  - Bumps the `nim-lsquic` dependency requirement and pin to v0.6.0

## Impact on Library Users

The public QUIC transport API is unchanged. QUIC dials with an expected peer ID now fail during handshake when the certificate identity does not match.

## Risk Assessment

Low to medium. The change affects QUIC dial failure timing for peer ID mismatches and depends on the new per-dial validator support in `nim-lsquic` v0.6.0.

## References

- https://github.com/vacp2p/roadmap/blob/master/content/p2p/ift/2026q3-nim-lsquic-per-connection-cert-validators.md

### DIFF
--- a/libp2p.nimble
+++ b/libp2p.nimble
@@ -11,7 +11,7 @@ requires "nim >= 2.2.4",
   "nimcrypto >= 0.6.0", "bearssl >= 0.2.7",
   "https://github.com/vacp2p/nim-boringssl >= 0.0.8", "chronicles >= 0.12.3",
   "chronos >= 4.2.2", "metrics >= 0.2.2", "secp256k1", "stew >= 0.4.2", "results",
-  "serialization >= 0.5.0", "json_serialization >= 0.4.4", "lsquic >= 0.5.5",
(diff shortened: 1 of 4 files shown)


vladopajic (approved):
lgtm, but 0 coverage, or maybe there was from before?

richard-ramos:
There's coverage from before in https://github.com/vacp2p/nim-libp2p/blob/master/tests/libp2p/transports/test_quic.nim#L132-L144  but i'll look into making sure that there's no connection being made in that test unit

gmelodie (approved):

PR #2883, opened 2026-07-24
feat(cbind): dial with addrs

gmelodie:
## Summary

`libp2pDial` only wrapped nim-libp2p's `Switch.dial(peerId, protos)` — the form that
opens a stream on a peer the switch already knows. Callers holding multiaddrs had to
compose `libp2pConnect` + `libp2pDial` themselves, which loses upstream's
teardown-on-failure (`Dialer.dial`'s `cleanup()` closes the muxer if the stream open
fails after connecting), can't reach `forceDial`, and splits the caller's timeout
budget across two calls.

`DialRequest` gains `multiaddrs` and `forceDial`. Empty `multiaddrs` keeps the current
behaviour; populated, the call routes to `Switch.dial(peerId, addrs, protos, forceDial)`
under a single `req.timeoutMs`. `forceDial` without addrs is rejected rather than
silently ignored — the addr-less overload has no such parameter.

The issue proposes a separate `libp2p_dial_addrs` entry point; cbind's C API is
request-struct based (`libp2p_ctx_dial(ctx, const DialRequest*, cb, ud)`), so extending
`DialRequest` is the equivalent shape here and keeps one dial symbol.

`examples/echo.c` now runs its exchange twice — once dialing with the server's addrs,
once without over the connection that leaves behind — so `nimble examples` covers both
paths. The per-exchange steps moved into `echoRoundTrip()` to make that a two-line
change rather than a copy-paste.

## Affected Areas

- [x] Build / Tooling
  

No nim-libp2p core code is touched — the change is confined to `cbind/`, so Nimbus,
Waku and Codex are unaffected.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

Downstream consumer of the C bindings: logos-libp2p-module (see References).

## Impact on Library Users

Nim users: none.

C binding users: `DialRequest` grows `multiaddrs` and `forceDial` between `proto` and
`timeoutMs`. Positional initializers that passed `timeoutMs` as the third element must
be updated; regenerate `c_bindings/libp2p.h` and re-check call sites. Callers passing an
empty addr list keep the current semantics.

## Risk Assessment

- **Behaviour difference worth knowing before passing addrs unconditionally:** with
  addrs, `Dialer.dial` runs `cleanup()` on failure, which closes the muxer it obtained —
  including a *reused* one, since `internalConnect` returns the existing connection for
  an already-connected peer. So a failed stream negotiation (e.g. the remote doesn't
  support the proto) can tear down a connection other streams are using. The addr-less
  path does not do this. Callers that already have a connection should keep passing an
  empty list; noted in a comment at the call site.
- Struct layout change in the generated C header — a source break for positional
  initializers, caught at compile time rather than silently.
- Dial path is unchanged for empty `multiaddrs`; the populated path delegates to an
  existing `Switch` overload (the same one `libp2pDialCircuitRelay` already uses).
- `forceDial` bypasses dial-limit backoff by design; it is opt-in per call and defaults
  to false.

## References

- Addresses logos-co/logos-libp2p-module#93 (cross-repo, so no auto-close)
- Origin of the request: logos-co/logos-libp2p-module#82 (review comment on
  `protocolRequest`, which currently composes connect + dial)


richard-ramos (approved):

cbind/libp2p.nim
```diff
@@ -492,9 +494,21 @@ proc libp2pDial*(
   # Same 30s backstop as libp2pConnect (else a slow dial trips a spurious timeout and leaks the streamId); `req.timeoutMs` bounds the dial itself.
   let peerId = PeerId.init(req.peerId).valueOr:
     return err($error)
+  let multiaddresses = parseMultiaddrs(req.multiaddrs).valueOr:
+    return err(error)
+  if req.forceDial and multiaddresses.len == 0:
+    return err("forceDial requires multiaddrs")
```
vladopajic, line 527: doing this, code implies that `dial` below will crash? if this is so then this needs to be fixed in the library code.

this is also not ideal, because this is implementation detail of `dial`. if in future (which will not happen), dial changes to allow such dials, then all other code with checks will also have to be changed (like here) - double the work.

`cbind` should only map proc calls from one domain to another, it should not really deal with specific.

cbind/libp2p.nim
```diff
@@ -492,9 +494,21 @@ proc libp2pDial*(
...
+
+  # With addrs libp2p tears the connection down if the stream fails to open, so a
+  # caller that hands them over for an already-connected peer can lose that connection.
+  let dialing =
+    if multiaddresses.len == 0:
+      lib.switch.dial(peerId, req.proto)
+    else:
+      lib.switch.dial(peerId, multiaddresses, @[req.proto], req.forceDial)
```
vladopajic, line 537: this code should be:

```suggestion
  let dialing = lib.switch.dial(peerId, multiaddresses, @[req.proto], req.forceDial)
```

----

the original pattern from pr, can be seen like this:

```
    if multiaddresses.len == 0:
      dial-helper-method (peerId, req.proto)
    else:
      dial-main-method (peerId, multiaddresses, @[req.proto], req.forceDial)
```

^ that is because `dial-helper-method` is actually wrapper around `dial-main-method`  that does following:

```
 proc dial-helper-method(peerId, proto):
      dial-main-method (peerId, @[], @[proto], false)
```
it calls main method with zero multi address and false for forceDial.


having this in mind; using `if/else` to call main method or helper method is flawed design.

cbind/libp2p.nim
```diff
@@ -492,9 +494,21 @@ proc libp2pDial*(
   # Same 30s backstop as libp2pConnect (else a slow dial trips a spurious timeout and leaks the streamId); `req.timeoutMs` bounds the dial itself.
   let peerId = PeerId.init(req.peerId).valueOr:
     return err($error)
+  let multiaddresses = parseMultiaddrs(req.multiaddrs).valueOr:
+    return err(error)
+  if req.forceDial and multiaddresses.len == 0:
+    return err("forceDial requires multiaddrs")
+
+  # With addrs libp2p tears the connection down if the stream fails to open, so a
+  # caller that hands them over for an already-connected peer can lose that connection.
```
vladopajic: with what intention was this comment added?

cbind/examples/README.md
```diff
@@ -4,7 +4,7 @@ Worked examples of consuming nim-libp2p from C through the FFI bindings generate
...
 
-- `echo.c` — two TCP nodes; the server mounts a custom `/cbind/echo/1.0.0` protocol and echoes the bytes it reads, the client dials it and verifies the round-trip. The incoming-stream handler only hands the stream id to `main`, which serves it inline, since calling back into the library from the dispatch thread would deadlock.
+- `echo.c` — two TCP nodes; the server mounts a custom `/cbind/echo/1.0.0` protocol and echoes the bytes it reads, the client dials it and verifies the round-trip — once with the server's addrs on the `dial` (which connects and opens the stream in one call), then again without them over the connection that leaves behind. The incoming-stream handler only hands the stream id to `main`, which serves it inline, since calling back into the library from the dispatch thread would deadlock.
```
vladopajic: that first sentience is too big. also very hard to understand (not only because sentence is too big, but wording is vague). 

these are suggested improvements:


> the server mounts a custom protocol `/cbind/echo/1.0.0` and echoes the bytes back it reads.

"custom protocol" is one token (collocation), should not be broken with path (which probably should be left out - its is irrelevant code detail)

> the client dials server and verifies the round-trip two times.

use server instead of it. 
explains what server and client will do.

> first time client calls dial with server address passed to dial, which establishes new connection and opens the stream in one call.

explain how first dial happens.

> second time client dials without specify address, but only peer id, which opens new stream reusing connection that was left behind first dial.

explain how second dial happens.

vladopajic (approved):

cbind/libp2p.nim
gmelodie: But the two `dial`s have different behaviors, no?
- `switch.dial(peerId, proto)` reuses the existing connection and on stream-open failure it just raises (no connection teardown).
- `switch.dial(peerId, addrs, protos, forceDial)` has `cleanup()` that calls `await conn.close()` on any failure. Even with empty `addrs`, it reuses the existing muxer, so a failure tears down the connection (a connection the caller didn't create).

cbind/libp2p.nim
gmelodie: ah, i guess my comment above answers it. i guess this needs rewording?

vladopajic: as a user who read this code i was not sure what to do with this comment. 

- was this supposed to explain reasoning why there is `if/else` dial?
- was this supposed to be caution when using `libp2pDial`? - should have been docs, to explain that potentially dial will close connection if current dial fails.
- "With addrs libp2p" - is very vague description, even for this narrow context

PR #2884, opened 2026-07-24
feat(cbind): nat config

richard-ramos:
## Summary

Expose NAT runtime configuration through the C bindings so C callers can configure port mapping, reachability probing, and hole punching through `Libp2pConfig`.

The config parser now converts the flat C-facing fields into `NATConfig` during node creation, while preserving zero-initialized config behavior. Existing `autonatV2` remains supported as an AutoNAT v2 reachability alias, and `autonatV2Server` remains separate from `withNAT`.

The C surface intentionally exposes top-level runtime NAT features, not AutoNAT v2 service tuning knobs or Nim-only injection hooks.

## Affected Areas

- [x] Other  
  C bindings config, manual C config mirror, and C binding smoke coverage.
- [x] Protocol Logic  
  NAT config selection for port mapping, reachability, and hole punching.

## Impact on Library Users

C binding users can now configure:

- NAT port mapping via auto, UPnP, NAT-PMP, or explicit IP modes
- AutoNAT v1/v2 reachability probing
- DCUtR hole punching with relay count and scheduling options

Invalid combinations, such as multiple port mapping modes or reachability together with hole punching, now fail during config parsing.

### DIFF
--- a/cbind/libp2p/config.nim
+++ b/cbind/libp2p/config.nim
@@ -156,7 +156,101 @@ func optDurationMs(ms: int64): Opt[Duration] =
   else:
     Opt.none(Duration)
 
-proc parseNATConfig(config: Libp2pConfig): Result[Opt[NATConfig], string] =
+proc parsePortMappingConfig(
+    config: Libp2pConfig
+): Result[Opt[PortMappingConfig], string] =
+  if not (
+    config.natPortMappingAuto or config.natPortMappingUpnp or config.natPortMappingNatPmp or
+    config.natExplicitIp.len > 0
+  ):
+    return ok(Opt.none(PortMappingConfig))
+
+  let
+    discoveryTimeout =
+      if config.natDiscoveryTimeoutMs > 0:
+        chronos.milliseconds(config.natDiscoveryTimeoutMs)
+      else:
+        DefaultDiscoveryTimeout
+    mappingTimeout =
+      if config.natMappingTimeoutMs > 0:
+        chronos.milliseconds(config.natMappingTimeoutMs)
+      else:
+        DefaultMappingTimeout
+
+  if config.natExplicitIp.len > 0:
+    try:
+      return ok(
+        Opt.some(
+          PortMappingConfig(
+            mode: ExplicitIp, explicitIp: parseIpAddress(config.natExplicitIp)
+          )
+        )
+      )
+    except ValueError as e:
+      return err("invalid natExplicitIp: " & e.msg)
+  if config.natPortMappingUpnp:
+    return ok(
+      Opt.some(
+        PortMappingConfig(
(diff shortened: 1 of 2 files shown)


gmelodie (approved):

cbind/libp2p/config.nim
```diff
@@ -146,12 +166,144 @@ func parseTransportConfig(config: Libp2pConfig): Result[ParsedTransport, string]
...
+        onReservation: nil,
+        scheduleInterval: optDurationMs(config.natHolePunchingScheduleIntervalMs),
+      )
+    )
+
+  if nat.portMapping.isNone() and nat.reachability.isNone() and nat.holePunching.isNone():
+    ok(Opt.none(NATConfig))
+  else:
+    ok(Opt.some(nat))
+
```
vladopajic, line 299: whole function is very large, if possible it would be better to split it. here we see there are 3 sub properties that are set, so func call for each?

cbind/libp2p/config.nim
```diff
@@ -146,12 +166,144 @@ func parseTransportConfig(config: Libp2pConfig): Result[ParsedTransport, string]
...
+        config.natHolePunchingMaxNumRelays
+      else:
+        1
+    nat.holePunching = Opt.some(
+      HolePunchingConfig(
+        maxNumRelays: maxNumRelays,
+        onReservation: nil,
+        scheduleInterval: optDurationMs(config.natHolePunchingScheduleIntervalMs),
+      )
+    )
```
vladopajic:
```suggestion
    nat.holePunching = Opt.some(
      HolePunchingConfig(
        maxNumRelays: max(1, config.natHolePunchingMaxNumRelays),
        onReservation: nil,
        scheduleInterval: optDurationMs(config.natHolePunchingScheduleIntervalMs),
      )
    )
```

cbind/libp2p/config.nim
```diff
@@ -146,12 +166,144 @@ func parseTransportConfig(config: Libp2pConfig): Result[ParsedTransport, string]
...
+  else:
+    ok(Opt.some(nat))
+
 proc parse(config: Libp2pConfig): Result[ParsedConfig, string] =
   let transport = ?parseTransportConfig(config)
   let dnsServers = ?resolveDnsServers(config.dnsResolver)
   let addrs = ?parseMultiaddrs(config.addrs)
   let privKey = ?parsePrivateKey(config.privKey)
   let bootstrapNodes = ?parseBootstrapNodes(config.bootstrapNodes)
+  let nat = ?parseNATConfig(config)
```
vladopajic: `nat` variable seems unnecessary, directly assigning to `ParsedConfig` would simplify code here

vladopajic (approved):

cbind/libp2p/config.nim
richard-ramos: I'll do the same with variables from 301 to 305

PR #2885, opened 2026-07-24
chore(cbind): adopt nim-ffi 0.3.0

gmelodie:
## Summary

Moves `cbind/` from the pinned nim-ffi 0.2.0 commit to the 0.3.0-rc.0 release
candidate (`435eebf`) and reworks the annotations to use what 0.3.0 adds.

The bump is not optional-only: 0.3.0 replaced the per-handler timeout with the
non-terminal `RET_STALE_WARN` progress callback, so the `{.ffi: "timeout = 30000".}`
spec on `libp2pConnect` / `libp2pDial` / `libp2pDialCircuitRelay` is now a hard
compile error (`unknown FFI override 'timeout = 30000'`). nim-ffi never cancels a
handler now, so those procs just carry `{.ffi.}` and their own `timeoutMs` bounds
the dial.

On top of that, the new surface removes hand-maintained duplication:

- **Doc comments.** A `##` comment on an annotated proc is propagated into the
  generated bindings, so every exported proc is documented once in
  `cbind/libp2p.nim` and comes out as `/** ... */` on both the raw symbol and the
  `libp2p_ctx_*` wrapper in `c_bindings/libp2p.h` (and as `;` comments in the CDDL
  schema). They are written for the C reader, so they name C symbols
  (`libp2p_ctx_start`, `MAX_READ_BYTES`), not Nim ones.
- **`{.ffi.}` enums.** `Libp2pConfig.muxer`/`.transport` were `int` ordinals with a
  comment saying nim-ffi could not carry an enum, plus `parseMuxer` /
  `parseTransport` validating them by hand — and every C example redeclared
  `static const int64_t TransportTcp = 1;`. They are now `TransportType` /
  `MuxerType` `{.ffi.}` enums emitted as real C enums (`TRANSPORT_TYPE_TCP`,
  `MUXER_TYPE_MPLEX`), crossing the wire as text. The two ordinal parsers and all
  six example copies are gone, and `parseTransportConfig` no longer returns a
  `Result`. `libp2pConnectedPeers` takes a `PeerDirection` enum instead of an
  `int` with an "invalid direction" runtime check.
- **`{.ffiConst.}`.** `MaxReadBytes` is re-emitted as `MAX_READ_BYTES` so a host
  can check a read length against the same cap the library enforces. The two
  remaining `int` fields — `CreateCidRequest.version` and
  `NewPrivateKeyRequest.scheme` — get named `CID_VERSION_V*` / `KEY_SCHEME_*`
  constants for the same reason. They deliberately stay `int` rather than becoming
  `{.ffi.}` enums: Nim reads `CidV0` and libp2p's `CIDv0` as the *same*
  identifier (as `Rsa`/`RSA`, `Ecdsa`/`ECDSA`), so a mirror enum is a style-check
  error against the very type it mirrors. `KEY_SCHEME_*` is derived via
  `ord(PKScheme.X)` since the field is cast straight to `PKScheme`;
  `CID_VERSION_V*` keeps cbind's existing 0/1 numbering, which is deliberately
  *not* libp2p's `CidVersion` ordinals (`CIDvIncorrect` occupies 0 there).
- **`{.ffiStatic.}`.** `libp2pCreateCid` and `libp2pDecodeXpr` never touched `lib`;
  they now take no library param and bind as `libp2p_static_create_cid` /
  `libp2p_static_decode_xpr`, callable without constructing a node.
- **Implicit event names.** `{.ffiEvent.}` derives the wire name from the proc
  name, so the `"on_incoming_stream"` / `"on_pubsub_message"` literals are dropped
  (derivation produces the same names, so the ABI is unchanged).
- **`ctx_destroy` returns `int`.** `libp2p_mobile_check.c` now fails when teardown
  reports an error instead of discarding it.
- `genbindings_c`/`genbindings_cddl` use `--compileOnly` instead of
  `--app:lib -o:/dev/null`, matching the documented two-compile model.

## Affected Areas

- [x] Build / Tooling
  

- [x] Other
  

Nothing outside `cbind/` changes — no libp2p core, transport or protocol code is
touched, so Nimbus / Waku / Codex are unaffected.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

**The downstream that does break is `logos-co/logos-libp2p-module`**, which
consumes `c_bindings/libp2p.h`. It needs the migration in the table below, and
this PR should not leave draft until there is a companion branch there proving it
builds and passes. That is the open item on this PR.

Verified locally: `nim check` passes, both binding generators emit, the FFI shared
library builds, and the `echo`, `gossipsub`, `kad`, `relay`, `peerstore` and
`metrics` examples build and run green against it. `nph --check` is clean.

## Impact on Library Users

No impact on Nim users of nim-libp2p. Breaking for C consumers of
`c_bindings/libp2p.h`:

| Before | After |
| --- | --- |
| `cfg.transport = 1; cfg.muxer = 0;` | `cfg.transport = TRANSPORT_TYPE_TCP; cfg.muxer = MUXER_TYPE_MPLEX;` |
| `libp2p_ctx_connected_peers(ctx, 0, ...)` | `libp2p_ctx_connected_peers(ctx, PEER_DIRECTION_INBOUND, ...)` |
| `libp2p_ctx_create_cid(ctx, &req, ...)` | `libp2p_static_create_cid(&req, ...)` |
| `libp2p_ctx_decode_xpr(ctx, &req, ...)` | `libp2p_static_decode_xpr(&req, ...)` |
| `void libp2p_ctx_destroy(ctx)` | `int libp2p_ctx_destroy(ctx)` (existing statement calls still compile) |

`CreateCidRequest.version` and `NewPrivateKeyRequest.scheme` keep their existing
integer values, so those calls are source- and wire-compatible; the new
`CID_VERSION_V*` / `KEY_SCHEME_*` constants only replace the literals.

Runtime behaviour change inherited from nim-ffi: a slow request is no longer timed
out. It runs to its natural `RET_OK`/`RET_ERR`, and the result callback receives a
non-terminal `RET_STALE_WARN` (3) every 5s while it is still in flight. The
generated typed wrappers already filter it; only a caller sitting on the raw
result-callback boundary needs to skip it.

## Risk Assessment

- **Backward compatibility:** breaking at the C ABI (see the table above); every
  in-tree example and the mobile smoke check are updated here, but the external
  module consumer is not yet validated.
- **Test coverage:** cbind's only CI coverage is `nimble examples`, and those
  examples never call `connected_peers`, `decode_xpr`, `create_xpr` or
  `new_private_key`. So the `PeerDirection` enum and the `decode_xpr` static
  conversion are exercised by nothing in CI — reviewed by inspection and by the
  generated header, not by a test.
- **Network behavior:** unchanged — enum values cross the wire as text, and the
  dial timeouts are still the caller-supplied `timeoutMs` / libp2p's own default.
- **Performance:** unchanged for the CBOR wire this library uses.
- **Security:** `MaxReadBytes` still gates every stream read; publishing it as a
  constant only tells the host the limit it was already subject to.
- The dependency is a **release candidate**, not a final 0.3.0 tag.

## References

- nim-ffi 0.3.0 changelog: https://github.com/logos-messaging/nim-ffi/blob/v0.3.0-rc.0/CHANGELOG.md
- `{.ffiStatic.}`: logos-messaging/nim-ffi#134 · `{.ffiConst.}` + `{.ffi.}` enums: logos-messaging/nim-ffi#140
- Doc-comment propagation: logos-messaging/nim-ffi#127 · `RET_STALE_WARN`: logos-messaging/nim-ffi#126

## Additional Notes

Known follow-ups deliberately left out of this PR:

- `libp2pConnectedPeersAnyDirection` stays a separate exported proc rather than a
  third `PeerDirection` value, to avoid removing a symbol hosts already call.
- `CLibp2pConfig` / `libp2p_new_default_config` (the hand-written plain-C config
  mirror) is referenced by nothing in this repo and is not layout-compatible with
  the generated `Libp2pConfig`. Its comment is corrected here; deleting the
  exported symbols is a separate call.

`c_bindings/` and `cddl_bindings/` are generated, not checked in — CI regenerates
them with `nimble genbindings_c` / `genbindings_cddl`.

### DIFF
--- a/cbind/cbind.nimble
+++ b/cbind/cbind.nimble
@@ -10,7 +10,8 @@ import os, strutils
 
 requires "taskpools >= 0.1.0",
   "https://github.com/vacp2p/nim-cbor-serialization#1664160e04d153573373afddc552b9cbf6fbe4dc",
-  "https://github.com/logos-messaging/nim-ffi#7ef58f4bf5cf683cfdf2abbdb3d7688727763d65"
+  # nim-ffi v0.3.0-rc.0
+  "https://github.com/logos-messaging/nim-ffi#435eebf9e3896e800c1586058a85df120a7fe870"
 
 proc findInstalledPkgDir(prefix: string): string =
   ## Path of an installed dep dir matching `prefix` (e.g. "ffi-"). Lockfile
@@ -68,10 +69,12 @@ task buildffi, "Build the FFI shared library":
   buildFfiLib()
 
 proc genBindingsFor(lang, outDir: string) =
-  exec "nim c --threads:on --app:lib --noMain --mm:refc -d:metrics" &
+  # `--compileOnly`: the binding files are written during macro expansion, so
+  # codegen is enough — there is nothing to link.
+  exec "nim c --threads:on --noMain --mm:refc -d:metrics --compileOnly" &
     " --nimMainPrefix:liblibp2p -d:ffiGenBindings -d:targetLang=" & lang &
     " -d:ffiOutputDir=" & outDir & " -d:ffiSrcPath=libp2p.nim" & ffiDepPaths() &
-    " --nimcache:nimcache_" & lang & " -o:/dev/null libp2p.nim"
+    " --nimcache:nimcache_" & lang & " libp2p.nim"
 
 task genbindings_c, "Generate C bindings (cbind/c_bindings)":
   genBindingsFor("c", "c_bindings")
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -463,8 +463,9 @@ CBOR codec and event queue, and *generates* the C/CDDL bindings from the
 annotations in `libp2p.nim`:
 
 - `libp2p.nim` — the top-level FFI module: `declareLibrary` + `{.ffi.}` /
-  `{.ffiCtor.}` / `{.ffiDtor.}` / `{.ffiEvent.}` annotations and the ported
-  libp2p logic; `genBindings()` emits the bindings.
+  `{.ffiCtor.}` / `{.ffiDtor.}` / `{.ffiStatic.}` / `{.ffiEvent.}` /
+  `{.ffiConst.}` annotations and the ported libp2p logic; `genBindings()` emits
+  the bindings.
 - `libp2p/config.nim` — config types and parsing, `include`d into `libp2p.nim`.
 - `c_bindings/` — generated C header (`libp2p.h`), consumed by
   `logos-co/logos-libp2p-module`.
@@ -481,6 +482,12 @@ nimble genbindings_cddl # Generate the CDDL schema
 **cbind conventions**:
 - Requests/responses are CBOR; declare `{.ffi.}` object types for them — never
   raw `ptr`/`pointer` across the boundary.
+- A `##` doc comment on an annotated proc is propagated into the generated
+  bindings, so document the exported API there and nowhere else.
+- Closed value sets are `{.ffi.}` enums (they emit a native C enum and cross the
+  wire as text), not `int` ordinals the host has to hardcode.
+- A proc that needs no node takes no `lib` param and is `{.ffiStatic.}`; its
+  wrapper is `libp2p_static_<name>` and takes no ctx.
 - Stream handles are plain `uint64` ids tracked in `LibP2P` state (nim-ffi has
   no handle type for incoming streams).
 - Convert config-parsing failures to `return err(...)`; never `raiseAssert` on
--- a/cbind/examples/echo.c
+++ b/cbind/examples/echo.c
@@ -8,10 +8,6 @@
 // serves inline.
 #include "common.h"
 
-// TransportType / MuxerType ordinals, mirrored from libp2p/config.nim.
-static const int64_t TransportTcp = 1;
-static const int64_t MuxerMplex = 0;
-
 static const char *EchoProto = "/cbind/echo/1.0.0";
 static const int64_t EchoMaxSize = 4096;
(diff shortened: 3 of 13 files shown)


cbind/cbind.nimble
```diff
@@ -10,7 +10,8 @@ import os, strutils
 
 requires "taskpools >= 0.1.0",
   "https://github.com/vacp2p/nim-cbor-serialization#1664160e04d153573373afddc552b9cbf6fbe4dc",
-  "https://github.com/logos-messaging/nim-ffi#7ef58f4bf5cf683cfdf2abbdb3d7688727763d65"
+  # nim-ffi v0.3.0-rc.0
+  "https://github.com/logos-messaging/nim-ffi#435eebf9e3896e800c1586058a85df120a7fe870"
```
gmelodie, line 14: Will change this when we register 0.3.0 in nimble pkgs

vladopajic (approved):

richard-ramos (approved):

cbind/cbind.nimble
arnetheduck: these ffi requirements should be behind a nimble feature so that they don't get pulled in unless asked for / the dep can be avoided - pure nim consumers of libp2p have no benefit from the ffi interface (in fact, one could argue that it should live in a separate repository altogether)

richard-ramos: That .nimble is part of a subfolder within libp2p. Do projects downstream take those into account?

arnetheduck: ah, I missed that - no, then it's fine :+1:

PR #2886, opened 2026-07-27
test: improve checkTrackers

rlve:
## Summary

`checkTracker` now resets the chronos counter after it reports a leak.

The counters live on the thread dispatcher and are global, so a single leaked resource used to make every later `checkTrackers()` call in the same test binary fail as well.

`reconcileTracker` evens out `opened` and `closed` with `trackCounter` / `untrackCounter`, and it runs after `fail()` has already recorded the failure. The next test starts from a clean counter and only reports what it leaks itself.

The checkpoint output is now one line per tracker instead of two:
```
stream.transport: opened 3, closed 2 (delta 1)
```

### DIFF
--- a/tests/tools/trackers.nim
+++ b/tests/tools/trackers.nim
@@ -22,15 +22,26 @@ const AllTrackerNames* = [
   ChronosStreamTrackerName,
 ]
 
+proc reconcileTracker(name: string, opened: int, closed: int) =
+  if opened >= closed:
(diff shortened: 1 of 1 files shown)


vladopajic (approved):

gmelodie (approved):
> checkTracker now resets the chronos counter after it reports a leak.

YES!!!!!11ONE

PR #2893, opened 2026-07-28
refactor(pubsub): use implicit proto2 fields

richard-ramos:
## Summary

This PR keeps the protobuf-ser migration while preserving the public pubsub RPC field shapes used by downstream projects.

`Message` now exposes selected hot-path fields as plain values again, such as `data`, `seqno`, `fromPeer`, `signature`, and `key`. Control message topic and prune backoff fields also use plain values.

The wire format still uses proto2 semantics through protobuf-ser `implicit` fields. Absent wire fields decode to Nim default values, while `Message.topic` remains required.

## Affected Areas

- [x] Gossipsub
- [x] Protocol Logic
- [x] Build / Tooling

## Impact on Library Users

This restores direct field access for pubsub RPC users that read or construct these fields without `Opt`.

Examples include:

- `message.data`
- `message.fromPeer`
- `message.seqno`
- `graft.topicID`
- `prune.topicID`
- `prune.backoff`

`encodeMessage`, `encodeRpcMsg`, and `decodeRpcMsg` are not restored.

## Risk Assessment

The main risk is the temporary dependency on the protobuf-ser branch that adds `implicit` proto2 field support.

The decode behavior intentionally collapses absent fields to default values for the restored plain fields. `Message.topic` still fails decode when missing.

## References

- https://github.com/status-im/nim-protobuf-serialization/pull/108
- https://github.com/vacp2p/nim-libp2p/pull/2638#issuecomment-5102949032

### DIFF
--- a/tests/libp2p/pubsub/utils.nim
+++ b/tests/libp2p/pubsub/utils.nim
@@ -161,8 +161,8 @@ proc teardownGossipSub*(gossipSub: TestGossipSub, conns: seq[Stream]) {.async.} 
 
 func defaultMsgIdProvider*(m: Message): Result[MessageId, ValidationResult] =
   let mid =
-    if m.seqno.isSome and m.fromPeer.isSome:
-      byteutils.toHex(m.seqno.get()) & $m.fromPeer.get()
+    if m.seqno.len > 0 and m.fromPeer.data.len > 0:
+      byteutils.toHex(m.seqno) & $m.fromPeer
     else:
       # This part is irrelevant because it's not standard,
       # We use it exclusively for testing basically and users should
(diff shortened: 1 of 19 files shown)


tests/libp2p/pubsub/utils.nim
```diff
@@ -161,8 +161,8 @@ proc teardownGossipSub*(gossipSub: TestGossipSub, conns: seq[Stream]) {.async.}
 
 func defaultMsgIdProvider*(m: Message): Result[MessageId, ValidationResult] =
   let mid =
-    if m.seqno.isSome and m.fromPeer.isSome:
-      byteutils.toHex(m.seqno.get()) & $m.fromPeer.get()
+    if m.seqno.len > 0 and m.fromPeer.data.len > 0:
```
vladopajic, line 164: `fromPeer.data.len` -> `fromPeer.len` would be better to avoid access to `.data`

vladopajic (approved):

richard-ramos:
Closing in favor of https://github.com/vacp2p/nim-libp2p/pull/2900

PR #2896, opened 2026-07-29
feat(log): add runtime logs filtering support

vladopajic:
- adding support for runtime logs filtering
- building default nix target with runtime logs filtering enabled

### DIFF
--- a/libp2p/logging.nim
+++ b/libp2p/logging.nim
@@ -12,5 +12,8 @@ template setLogLevel*(level: LogLevel) =
   when defined(chronicles_runtime_filtering):
     chronicles.setLogLevel(level)
   else:
-    {.error: "Run-time Chronicles log filtering is disabled. " &
-      "Enable it with '-d:chronicles_runtime_filtering:on'.".}
+    {.
+      error:
+        "Run-time Chronicles log filtering is disabled. " &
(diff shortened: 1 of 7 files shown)


libp2p/logging.nim
```diff
@@ -0,0 +1,19 @@
...
+
+import pkg/chronicles
+
+export LogLevel
+
+template setLogLevel*(level: LogLevel) =
+  ## Set the runtime Chronicles log level for all configured sinks.
+  ##
+  ## This requires compiling with `-d:chronicles_runtime_filtering`.
+  when defined(chronicles_runtime_filtering):
```
richard-ramos:
```suggestion
  when chronicles.runtimeFilteringEnabled:
```
Like it's done in [here](https://github.com/logos-messaging/logos-delivery/blob/26e1d625240caa11b1973b3776ba011e6b899d40/logos_delivery/waku/rest_api/endpoint/admin/handlers.nim#L440)  


The reason is that this is a boolean compile time variable, so if someone does  `-d:chronicles_runtime_filtering:false`,  the `defined(chronicles_runtime_filtering)` would be triggered even if the logs are not meant to be active

gmelodie (approved):

richard-ramos (approved):

PR #2898, opened 2026-07-29
chore(kad): evict stale peer on refresh

### LINKED ISSUE #2890: kad: regularly ping and evict peers
### Problem

We currently do not regularly sweep our routing table(s) for peers that might have gone stale.
My suggestion is based on [a similar maintenance loop in go-libp2p](https://github.com/libp2p/go-libp2p-kad-dht/blob/fbe4251243c094fddbddc77af9f42eb4e1a1029d/rtrefresh/rt_refresh_manager.go#L134-L178), where peers that haven't been heard from in a while gets pinged and non-responders evicted. If the `lastUsefulAt`-timestamps work, we can skip peers that that were heard from within some grace period.

SionoiS:
Fixes #2890 by adding a new mechanism that track the "liveness" of peers in a table. When a table is refreshed this runs first, evict stale peers and then the usual `findnode` is called to fill each bucket.

### DIFF
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -34,9 +34,11 @@ proc livenessCandidates(
 
 proc checkAndEvictPeer(
     kad: KadDHT, rtable: RoutingTable, peerId: PeerId
-) {.async: (raises: []).} =
+) {.async: (raises: [CancelledError]).} =
   ## Probe one routing-table peer; remove it if unreachable or not speaking DHT.
   ## Takes ownership of one ``probeSem`` slot already acquired by the caller.
+  await kad.probeSem.acquire()
+
   defer:
     try:
       kad.probeSem.release()
@@ -77,13 +79,8 @@ proc pingAndEvictPeers*(
   if candidates.len == 0:
     return
 
-  var futs: seq[Future[void]]
+  var futs = newSeqOfCap[Future[void]](candidates.len)
   for peerId in candidates:
-    if kad.stopping:
-      break
-    if not kad.probeSem.tryAcquire():
-      # Remaining candidates are retried on the next maintenance pass.
-      break
     futs.add(kad.checkAndEvictPeer(rtable, peerId))
 
   if futs.len > 0:
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -18,11 +18,84 @@ logScope:
 
 const KadCodec* = "/ipfs/kad/1.0.0"
 
+proc livenessCandidates(
+    rtable: RoutingTable, gracePeriod: Duration
+): seq[PeerId] {.raises: [].} =
+  ## Peers past the liveness grace period that should be probed.
+  let now = Moment.now()
+  var peers: seq[PeerId]
+  for bucket in rtable.buckets:
+    for entry in bucket.peers:
+      if not entry.needsLivenessCheck(gracePeriod, now):
+        continue
+      entry.nodeId.toPeerId().withValue(pid):
+        peers.add(pid)
+  peers
+
+proc checkAndEvictPeer(
+    kad: KadDHT, rtable: RoutingTable, peerId: PeerId
+) {.async: (raises: []).} =
+  ## Probe one routing-table peer; remove it if unreachable or not speaking DHT.
+  ## Takes ownership of one ``probeSem`` slot already acquired by the caller.
+  defer:
+    try:
+      kad.probeSem.release()
+    except AsyncSemaphoreError:
+      raiseAssert "probeSem released without acquire"
+
+  let addrs = kad.switch.peerStore[AddressBook][peerId]
+  if addrs.len == 0:
+    trace "Evicting peer with no known addresses", peer = peerId.shortLog()
+    discard rtable.removePeer(peerId, reason = "liveness")
+    kad_routing_table_liveness_probes.inc(labelValues = ["no_addrs"])
+    return
+
+  let alive =
+    try:
+      await kad.lookupCheck(peerId, addrs)
+    except CancelledError:
+      return
+
+  if alive:
+    rtable.markUseful(peerId)
+    kad_routing_table_liveness_probes.inc(labelValues = ["ok"])
+    return
+
+  trace "Evicting unresponsive peer after liveness probe", peer = peerId.shortLog()
+  discard rtable.removePeer(peerId, reason = "liveness")
+  kad_routing_table_liveness_probes.inc(labelValues = ["fail"])
+
+proc pingAndEvictPeers*(
+    kad: KadDHT, rtable: RoutingTable
+) {.async: (raises: [CancelledError]).} =
+  ## Probes routing-table peers that have been quiet longer than the liveness
+  ## grace period and removes those that fail to answer a FIND_NODE.
+  if kad.stopping:
+    return
+
+  let candidates = rtable.livenessCandidates(kad.config.livenessGracePeriod)
+  if candidates.len == 0:
+    return
+
+  var futs: seq[Future[void]]
+  for peerId in candidates:
+    if kad.stopping:
+      break
+    if not kad.probeSem.tryAcquire():
+      # Remaining candidates are retried on the next maintenance pass.
+      break
+    futs.add(kad.checkAndEvictPeer(rtable, peerId))
+
+  if futs.len > 0:
+    await noCancel allFutures(futs)
+
 proc refreshTable*(
     kad: KadDHT, rtable: RoutingTable, forceRefresh = false
 ) {.async: (raises: [CancelledError]).} =
   ## Sends a findNode to find itself to keep nearby peers up to date
   ## Also sends a findNode to find a random key for each non-empty k-bucket
+
+  await kad.pingAndEvictPeers(rtable)
 
   discard await kad.findNode(rtable.selfId)
 
@@ -35,7 +108,7 @@ proc refreshTable*(
       continue
 
     # skip if refresh conditions not met (forceRefresh OR stale bucket)
-    if not (forceRefresh or bucket.isStale()):
+    if not (forceRefresh or bucket.isStale(rtable.config.bucketStaleTime)):
       continue
 
     let target = rtable.refreshTarget(i, kad.rng).valueOr:
@@ -77,7 +150,11 @@ proc new*(
 ): K {.raises: [].} =
   var rtable = RoutingTable.new(
     switch.peerInfo.peerId.toKey(),
-    config = RoutingTableConfig.new(replication = config.replication),
+    config = RoutingTableConfig.new(
+      replication = config.replication,
+      usefulnessGracePeriod = config.usefulnessGracePeriod,
+      bucketStaleTime = config.bucketStaleTime,
+    ),
   )
   let kad = K(
     rng: rng,
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -32,15 +32,6 @@ proc livenessCandidates(
         peers.add(pid)
   peers
 
-proc checkAndEvictPeer(
-    kad: KadDHT, rtable: RoutingTable, peerId: PeerId
-) {.async: (raises: [CancelledError]).} =
-  ## Probe one routing-table peer; remove it if unreachable or not speaking DHT.
-  await kad.probeSem.acquire()
-
-  defer:
-    try:
-      kad.probeSem.release()
 template withProbeSlotOrReturn*(kad: KadDHT) =
   ## Acquire one ``probeSem`` slot for the enclosing scope, or return without
   ## probing. Non-blocking: candidates that find no free slot are retried on a
@@ -56,9 +47,8 @@ template withProbeSlotOrReturn*(kad: KadDHT) =
 
 proc checkAndEvictPeer(
     kad: KadDHT, rtable: RoutingTable, peerId: PeerId
-) {.async: (raises: []).} =
+) {.async: (raises: [CancelledError]).} =
   withProbeSlotOrReturn(kad)
-      raiseAssert "probeSem released without acquire"
 
   # Candidate list is a snapshot; by the time a slot is free the peer may have
   # been refreshed (markUseful) or removed, and stop may have begun.
@@ -75,9 +65,7 @@ proc checkAndEvictPeer(
     kad_routing_table_liveness_probes.inc(labelValues = ["no_addrs"])
     return
 
-  let alive = await kad.lookupCheck(peerId, addrs)
-
-  if alive:
+  if (await kad.lookupCheck(peerId, addrs)):
     rtable.markUseful(peerId)
     kad_routing_table_liveness_probes.inc(labelValues = ["ok"])
     return
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -26,7 +26,7 @@ proc livenessCandidates(
   var peers: seq[PeerId]
   for bucket in rtable.buckets:
     for entry in bucket.peers:
-      if not entry.needsLivenessCheck(gracePeriod, now):
+      if not entry.isReplaceable(gracePeriod, now):
         continue
       entry.nodeId.toPeerId().withValue(pid):
         peers.add(pid)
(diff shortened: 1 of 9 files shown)


gmelodie (commented):
Please add a PR description with the problem, how this solves it, etc.

tests/libp2p/kademlia/test_bootstrap.nim
```diff
@@ -146,3 +146,89 @@ suite "KadDHT Bootstrap Component":
...
+    # Fresh insert times: within grace, so no probe and no eviction.
+    await hub.pingAndEvictPeers(hub.rtable)
+    check hub.hasKey(leafId.toKey())
+
+  asyncTest "pingAndEvictPeers removes peers with no known addresses":
+    let kad = setupMockKad()
+    startAndDeferStop(@[kad])
+
+    let orphan = randomPeerId()
+    discard kad.rtable.insert(orphan)
```
vladopajic: doing `discard` here is wasteful as `discard` can be easily converted to `check`
->
`check kad.rtable.insert(orphan)`.


... and in other places in this pr

tests/libp2p/kademlia/test_routing_table.nim
```diff
@@ -215,6 +215,38 @@ suite "KadDHT Routing Table":
...
+    let now = Moment.now()
+    var entry = NodeEntry(
+      nodeId: testKey(1),
+      lastSeen: now,
+      addedAt: now - 2.hours,
+      lastUsefulAt: Opt.none(Moment),
+    )
+    check entry.needsLivenessCheck(1.hours, now)
+    entry.lastUsefulAt = Opt.some(now - 30.minutes)
+    check not entry.needsLivenessCheck(1.hours, now)
```
vladopajic:
```
    let now = Moment.now()
    var entry = NodeEntry(
      addedAt: now - 2.hours,
     lastUsefulAt: Opt.none(Moment), 
    )
   check entry.needsLivenessCheck(1.hours, now)


    var entry = NodeEntry(
      addedAt: now - 2.hours,
      lastUsefulAt: Opt.some(now - 30.minutes),
    )
  check not entry.needsLivenessCheck(1.hours, now)
```

some, opinionated, cosmetics:
- better create entry each time then modifying property - this make it clearer to the reader to see at whole properties of object at once. 
- also separate, with line brake, independent  `check`s

tests/libp2p/kademlia/test_routing_table.nim
```diff
@@ -215,6 +215,38 @@ suite "KadDHT Routing Table":
...
+    let selfId = testKey(0)
+    var rt = RoutingTable.new(
+      selfId, config = RoutingTableConfig.new(hasher = Opt.some(noOpHasher))
+    )
+    let key = rt.keyInBucket(TargetBucket)
+    check rt.insert(key)
+    check key in rt
+    check rt.removePeer(key)
+    check key notin rt
+    check not rt.removePeer(key) # already gone
```
vladopajic:
```suggestion
    check:
       rt.insert(key)
       key in rt
    
    check: 
       rt.removePeer(key)
       key notin rt
    
    check not rt.removePeer(key) # already gone
```

opinionated, cosmetics following pattern for readability:

```
check:
  do_some_operation()                     # rt.removePeer(key)
  assert_same_operation                   # key notin rt
  assert_same_operation_different_assert  # key in rtt 
```

or when `do_some_operation` returns something other then bool

```
check:
  do_some_operation() == expected_value
  assert_same_operation
  assert_same_operation_different_assert
```

or when `do_some_operation` returns void:

```
do_some_operation()
check:
  assert_same_operation
  assert_same_operation_different_assert
```

tests/libp2p/kademlia/test_bootstrap.nim
```diff
@@ -146,3 +146,89 @@ suite "KadDHT Bootstrap Component":
...
+    await leaf.stop()
+    await leaf.switch.stop()
+
+    let past = Moment.now() - (DefaultLivenessGracePeriod + 1.minutes)
+    for b in hub.rtable.buckets.mitems:
+      for p in b.peers.mitems:
+        if p.nodeId == leafId.toKey():
+          p.addedAt = past
+          p.lastUsefulAt = Opt.none(Moment)
+          p.lastSeen = past
```
vladopajic: consider converting this to utility proc or template. it would ease readability of code.

consider:

```
var leafNode = hub.findByKey(leafId.toKey())
leafNode.addedAt = ...
leafNode.lastUsefulAt = ...
leafNode.lastSeen =...
```
^ introduces utility function to find leaf

or 

```
var leafNode = hub.findByKey(leafId.toKey())
leafNode.setPastGrace(DefaultLivenessGracePeriod) # note sure how this action should be called 
```
^ introduces action that changes leaf to be past grace period


or 

```
setNodePastGracePeriod(leafId, DefaultLivenessGracePeriod)
```
^ alternatively do everything as one utility. in which case it would still be nice to have  `findByKey` as same utility is likely to be reused.  

----

note there are similar patterns in other places of same pr

tests/libp2p/kademlia/test_bootstrap.nim
```diff
@@ -146,3 +146,89 @@ suite "KadDHT Bootstrap Component":
...
+    await hub.pingAndEvictPeers(hub.rtable)
+
+    check hub.hasKey(leafId.toKey())
+    # Successful probe marks the peer useful again.
+    var useful = false
+    for b in hub.rtable.buckets:
+      for p in b.peers:
+        if p.nodeId == leafId.toKey():
+          useful = p.lastUsefulAt.isSome()
+    check useful
```
vladopajic:
```
let leadNode = hub.findByKey(leafId)
check leadNode.lastUsefulAt.isSome
```

libp2p/protocols/kademlia.nim
```diff
@@ -18,12 +18,85 @@ logScope:
...
+    return
+
+  let candidates = rtable.livenessCandidates(kad.config.livenessGracePeriod)
+  if candidates.len == 0:
+    return
+
+  var futs: seq[Future[void]]
+  for peerId in candidates:
+    if kad.stopping:
+      break
```
vladopajic: doing `if kad.stopping:` is here unnecessary. it was checked right at the start of proc. since start there was no yielding, (there was no await) so kad could have not been stopped?

libp2p/protocols/kademlia.nim
```diff
@@ -18,12 +18,85 @@ logScope:
...
+    if not kad.probeSem.tryAcquire():
+      # Remaining candidates are retried on the next maintenance pass.
+      break
+    futs.add(kad.checkAndEvictPeer(rtable, peerId))
```
vladopajic: this seems like bad pattern.

- why acquiring multiple times in for?
- doing acquire and release in different procs is always a red flag. here it also seems unnecessary, code can do it in every future separately (in `checkAndEvictPeer` procs), then bellow you await on all futures. these futures will race, first one to start executing will get lock and progress until completion then release lock, this repeats until all futures are done.

vladopajic: there needs to be tests case which will make this `checkAndEvictPeer` to be triggered may times, as looking at this makes me think that this can deadlock. (probably fixed by acquire is moved to checkAndEvictPeer).

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -76,10 +78,19 @@ proc oldestPeer*(bucket: Bucket): (NodeEntry, int) =
...
 func isReplaceable*(entry: NodeEntry, gracePeriod: Duration, now: Moment): bool =
   ## Replaceable once past `gracePeriod` in the table without proving useful; a
   ## peer that answered recently or was just added is retained.
-  now - entry.lastUsefulAt.get(entry.addedAt) > gracePeriod
+  now - entry.lastActivity() > gracePeriod
+
+func needsLivenessCheck*(entry: NodeEntry, gracePeriod: Duration, now: Moment): bool =
+  ## True when the peer has had no successful DHT activity within `gracePeriod`
+  ## and should be probed (and possibly removed) during maintenance.
+  now - entry.lastActivity() > gracePeriod
```
gmelodie: These two are identical. Choose one

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -158,6 +169,28 @@ proc insert*(rtable: RoutingTable, nodeId: Key): bool =
...
+proc removePeer*(rtable: RoutingTable, nodeId: Key, reason = "remove"): bool =
+  ## Removes `nodeId` from the routing table. Returns true if it was present.
+  ## ``reason`` labels the ``kad_routing_table_evictions`` metric
+  ## (e.g. ``"liveness"``, ``"remove"``).
+  if nodeId == rtable.selfId or nodeId == rtable.localNodeId:
+    return false
+
+  let idx = rtable.bucketIndex(nodeId)
+  if idx >= rtable.buckets.len:
+    return false
```
gmelodie, line 189: Can this actually happen?

libp2p/protocols/kademlia.nim
```diff
@@ -18,12 +18,85 @@ logScope:
...
+
+  if futs.len > 0:
+    await noCancel allFutures(futs)
```
gmelodie: if we implement my suggestion from above this can be simplified to
 
```suggestion
var futs: seq[Future[void]]
for peerId in candidates:
  futs.add(kad.checkAndEvictPeer(rtable, peerId))

if futs.len > 0:
  await noCancel allFutures(futs)
```

libp2p/protocols/kademlia.nim
```diff
@@ -18,12 +18,85 @@ logScope:
...
+  if alive:
+    rtable.markUseful(peerId)
+    kad_routing_table_liveness_probes.inc(labelValues = ["ok"])
+    return
+
+  trace "Evicting unresponsive peer after liveness probe", peer = peerId.shortLog()
+  discard rtable.removePeer(peerId, reason = "liveness")
+  kad_routing_table_liveness_probes.inc(labelValues = ["fail"])
+
+proc pingAndEvictPeers*(
```
gmelodie:
```suggestion
proc probeAndEvictPeers*(
```
nit

libp2p/protocols/kademlia.nim
gmelodie: Interesting comment from claude:
> tryAcquire never blocks and every success is immediately followed by the spawn that arms the defer. The real hazard is subtler and worth saying explicitly — it's one edit away from leaking. Insert any continue or failable expression between tryAcquire() and futs.add(...) and you permanently burn a slot, shrinking probe capacity for the process lifetime with no error anywhere

gmelodie (commented):
Two comments from claude worth checking
- TOCTOU. livenessCandidates snapshots PeerIds, then each probe may run much later. By eviction time the peer could have been refreshed by unrelated traffic (markUseful) or already removed. Re-check needsLivenessCheck after the slot is acquired and before removePeer — cheap, and it matters more the longer probes are delayed.
- Shutdown asymmetry. Admission probes are tracked in kad.admissionProbes and explicitly cancelAndWait'd in stop (kademlia.nim:264-266). Liveness probes are tracked nowhere and awaited under noCancel, so they can't be cancelled at all. Either route them through trackProbe too, or justify why they're different.

libp2p/protocols/kademlia/routing_table.nim
SionoiS: If a fake key is passed in then sure but otherwise no.

gmelodie: I though any key in the key space should go inside one of the buckets, no?

libp2p/protocols/kademlia.nim
```diff
@@ -18,12 +18,82 @@ logScope:
...
+    trace "Evicting peer with no known addresses", peer = peerId.shortLog()
+    discard rtable.removePeer(peerId, reason = "liveness")
+    kad_routing_table_liveness_probes.inc(labelValues = ["no_addrs"])
+    return
+
+  let alive =
+    try:
+      await kad.lookupCheck(peerId, addrs)
+    except CancelledError:
+      return
```
vladopajic:
```
let alive = await kad.lookupCheck(peerId, addrs)
```

no special handling needed, cancelled error is propagated

libp2p/protocols/kademlia/routing_table.nim
SionoiS: I thought buckets were lazily created... Let me check.

edit: yeah they are lazy

libp2p/protocols/kademlia.nim
```diff
@@ -18,12 +18,93 @@ logScope:
...
+proc checkAndEvictPeer(
+    kad: KadDHT, rtable: RoutingTable, peerId: PeerId
+) {.async: (raises: [CancelledError]).} =
+  ## Probe one routing-table peer; remove it if unreachable or not speaking DHT.
+  await kad.probeSem.acquire()
+
+  defer:
+    try:
+      kad.probeSem.release()
+    except AsyncSemaphoreError:
```
gmelodie, line 46:
```suggestion
template withProbeSlotOrReturn*(kad: KadDHT) =
  ## Acquire one ``probeSem`` slot for the enclosing scope, or return without
  ## probing. Non-blocking: candidates that find no free slot are retried on a
  ## later pass rather than queued.
  if not kad.probeSem.tryAcquire():
    kad_routing_table_liveness_probes.inc(labelValues = ["skipped"])
    return
  defer:
    try:
      kad.probeSem.release()
    except AsyncSemaphoreError:
      raiseAssert "probeSem released without acquire"

proc checkAndEvictPeer(
    kad: KadDHT, rtable: RoutingTable, peerId: PeerId
) {.async: (raises: []).} =
  withProbeSlotOrReturn(kad)
```

gmelodie: Apologies, i forgot to click "comment" on this one while doing the review. The idea is to use this along with https://github.com/vacp2p/nim-libp2p/pull/2898#discussion_r3683428551 to avoid locking outside that function

libp2p/protocols/kademlia.nim
```diff
@@ -18,12 +18,93 @@ logScope:
...
+      await noCancel futs.cancelAndWait()
+      raise exc
+
 proc refreshTable*(
     kad: KadDHT, rtable: RoutingTable, forceRefresh = false
 ) {.async: (raises: [CancelledError]).} =
   ## Sends a findNode to find itself to keep nearby peers up to date
   ## Also sends a findNode to find a random key for each non-empty k-bucket
 
+  await kad.probeAndEvictPeers(rtable)
```
richard-ramos, line 111: Is there a risk for this line to affect the bucket refresh? `proveAndEvictPeers` can be 'slow' depending on the number of peers to probe, and if the peers timeout

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -20,6 +20,7 @@ proc new*(
     maxBuckets: int = DefaultMaxBuckets,
     selfIdPreHashed = false,
     usefulnessGracePeriod: Duration = DefaultUsefulnessGracePeriod,
+    bucketStaleTime: Duration = DefaultBucketStaleTime,
```
richard-ramos, line 23: Should these settings be passed also to the service routing tables in `addService` in routing_table_manager.nim?

richard-ramos (approved):

vladopajic (approved):

libp2p/protocols/kademlia.nim
```diff
@@ -18,12 +18,93 @@ logScope:
...
+  let addrs = kad.switch.peerStore[AddressBook][peerId]
+  if addrs.len == 0:
+    trace "Evicting peer with no known addresses", peer = peerId.shortLog()
+    discard rtable.removePeer(peerId, reason = "liveness")
+    kad_routing_table_liveness_probes.inc(labelValues = ["no_addrs"])
+    return
+
+  let alive = await kad.lookupCheck(peerId, addrs)
+
+  if alive:
```
vladopajic:
```suggestion
  let alive = await kad.lookupCheck(peerId, addrs)
  if alive:
```

does that line feels like having mouth open while not speaking anything, or it is just my ocd?

vladopajic (approved):

libp2p/protocols/kademlia.nim
SionoiS: That's true but we do have a timeout now for `refreshTable`. Do we need another mechanism like continuous eviction instead of periodic?

libp2p/protocols/kademlia/routing_table.nim
SionoiS: Using default is fine I think.

libp2p/protocols/kademlia.nim
SionoiS: Alternative -> https://github.com/vacp2p/nim-libp2p/pull/2907

tinniaru3005 (approved):
LGTM :)

PR #2901, opened 2026-07-30
chore(kad): reuse per-peer streams instead of dialing a fresh stream per RPC

### LINKED ISSUE #2856: kad: reuse per-peer streams instead of dialing a fresh stream per RPC
Every RPC dispatch opens and closes its own stream: `find.nim:171`, `get.nim:18`, `put.nim:54`, `provider.nim:111` each call `switch.dial(peer, addrs, codec)` and `stream.close()`. Stream setup cost is paid on every message.

go-libp2p (`internal/net/message_manager.go`) keeps one reused stream per peer (`peerMessageSender`, reused up to `streamReuseTries`), serialized by a ctx-aware mutex, with a pooled buffered protobuf writer and a single retry-on-reset.

Proposed work:
- Add a per-peer message-sender layer that lazily opens a stream, serializes requests on it, reuses it across RPCs, and resets+retries once on error.
- Route all four dispatch procs through it.

Impact: fewer stream negotiations, lower per-RPC overhead. High.

gmelodie:
## Summary

Every Kademlia RPC opened its own stream, sent one message, and closed it. Stream negotiation was paid once per message, so a single iterative lookup could open dozens of streams to the same set of peers.

This adds a per-peer message-sender layer (`libp2p/protocols/kademlia/message_sender.nim`) that lazily opens one stream per peer, serializes RPCs on it with a lock, reuses it across calls, and resets plus retries once when a reused stream turns out to be dead. It mirrors go-libp2p's `internal/net/message_manager.go` (`peerMessageSender`, `streamReuseTries`). All five dispatch procs — FIND_NODE, GET_VALUE, PUT_VALUE, ADD_PROVIDER, GET_PROVIDERS — now go through it.

The five dispatch bodies were nearly identical, so the shared path (take an RPC slot, encode, send, count metrics, decode) moved into `libp2p/protocols/kademlia/rpc.nim` as `dispatchRpc`. Each dispatcher is now two to six lines.

Closes #2856.

## Affected Areas

- [x] Protocol Logic
  Kademlia outbound RPC path only. Inbound handlers are unchanged — the existing handler loop already reads messages from one stream until EOF, so it serves reused streams without modification.

- [x] Other
  New Prometheus counters: `kad_streams_opened`, `kad_stream_reuses`, `kad_stream_reuse_failures`.

No wire-format change: the same messages are sent, on fewer streams. Reuse is transparent to remotes that already handle multiple messages per stream (go-libp2p, js-libp2p, and this implementation's own handler), and a remote that drops the stream between RPCs is handled by the reset-and-retry path.

- **Nimbus:** N/A — Kademlia DHT is not used.
- **Waku:** N/A — no integration branch needed; no API or wire change.
- **Codex:** N/A — no integration branch needed; no API or wire change.

## Impact on Library Users

No public API change to `KadDHT`. `KadDHT` gains a `msgSender` field, and `libp2p/protocols/kademlia/types.nim` re-exports the new `message_sender` module (`MessageSender`, `SendError`, `SendStage`).

Behaviour changes users may notice:

- Fewer outbound streams and fewer protocol negotiations per lookup, which is the point of the change.
- FIND_NODE, GET_VALUE, PUT_VALUE and GET_PROVIDERS now apply `config.timeout` to the reply read. Previously only ADD_PROVIDER did, and the other four relied on the caller's deadline; an unanswered RPC could hang until the lookup abandoned it.
- RPCs to the same peer serialize instead of running on parallel streams. In practice a lookup dispatches to distinct peers, so concurrency is unaffected; the wait for a peer's turn is bounded by the same `config.timeout`.
- PUT_VALUE now honours `config.hideConnectionStatus`. It called `Message.encode()` with the default while the other four passed the configured value.
- A reply whose message type differs from the request's is rejected with `"<type> reply type mismatch"`. One stream per peer carries no request ids, so a reply of the wrong type means the stream desynced and the answer belongs to another RPC. A reply that omits the type field is still accepted, which keeps remotes that leave it unset working.
- Decode-failure error strings are derived from the message type, so `"FindNode reply decode fail"` becomes `"findNode reply decode fail"`. RPC error strings now carry the stage that failed, for example `"dialStage: ..."`.
- `kad_message_duration_ms` now covers the wait for the peer's stream as well as the round trip, so it reports queueing behind a slow peer instead of hiding it.
- `kad_messages_sent` and `kad_message_bytes_sent` count only messages that got past the dial. A message to an unreachable peer is no longer counted as sent.
- The dial is bounded by what is left of `config.timeout`, so an RPC to a peer that accepts connections slowly can no longer outlive its own budget.

## Risk Assessment

**Backward compatibility:** no wire change. A remote that answers ADD_PROVIDER when this node has `providerRejection` disabled would leave its reply buffered on a shared stream, so a fire-and-forget send closes its stream and drops it from the pool instead of handing it to the next RPC.

**Correctness of stream state:** a stream is reset, never closed, on any failure or cancellation — a half-closed stream keeps its unread reply buffered and blocks the muxer for every other channel on the connection. A read timeout also resets, so a late reply cannot be picked up as the answer to the next RPC. A stream that is given up after a successful exchange is closed with `closeWithEOF`, bounded by the RPC's remaining budget, and reset if the remote never answers the close. A reply of an unexpected type is refused rather than returned to the caller.

**Dial lifetime:** `Dialer.dial` closes whatever it opened when it is cancelled, so the dial runs under the RPC's deadline instead of a cancellation shield. `stop` and `dropPeer` can still run while a dial is in flight, so the sender is checked again once the dial returns and the new stream is reset rather than used.

**Resource bounds:** the per-peer table is pruned two ways. A `Disconnected` conn-event handler drops the peer's entry, and an entry with no live stream and no in-flight RPC is deleted when its last user leaves — otherwise every peer id a lookup ever named would leave a permanent entry, including peers that were never reachable.

**Degradation:** a peer whose stream reuse keeps breaking falls back to one stream per RPC after `DefaultMaxStreamReuseFailures` (3), matching go-libp2p's `streamReuseTries`. The one retry after a broken reuse is skipped when the deadline is already spent, so a slow peer is not dialled again and sent a duplicate of the same RPC.

**Shutdown:** `MessageSender.stop()` refuses further sends, takes its `Disconnected` handler back off the switch and resets every open stream; `KadDHT.start` re-registers the handler, so restart still works.

**Head-of-line blocking:** `dispatchRpc` takes an `rpcSem` slot before it queues for the peer's stream, so RPCs waiting on one slow peer hold global slots for up to `config.timeout`. Master held a slot for an unbounded time instead, because four of the five RPC types read the reply with no timeout at all.

**Reference cycle:** the switch holds the conn-event closure, which holds the `MessageSender`, which holds the switch, and under `--mm:refc` that cycle is not collected. `stop` deregisters the handler, so a stopped DHT breaks the cycle. A running one keeps it, which matches the existing `KadDHT` and `Switch` relationship.

## References

- Closes #2856
- Parent issue #1756
- go-libp2p `internal/net/message_manager.go`

## Additional Notes

New test file `tests/libp2p/kademlia/test_message_sender.nim` covers stream reuse across RPCs, transparent reopen after the remote drops the stream, serialization of concurrent RPCs, stream retirement for reply-less sends, `dialStage` and `readStage` error reporting, `dropPeer`, and a stopped sender refusing to dial.

The fallback after `maxReuseFailures` is not covered by a test yet. It needs a stream that passes the liveness check and then breaks, which races the `atEof()` check.

### DIFF
--- a/libp2p/protocols/kademlia/message_sender.nim
+++ b/libp2p/protocols/kademlia/message_sender.nim
@@ -272,8 +272,19 @@ proc send(
   # deadline, and the retry the caller sends next would spend its whole budget
   # waiting for the stream, never reaching the peer.
   let deadline = Moment.now() + timeout
-  if not await ps.lock.acquire().withTimeout(timeout):
-    return err(SendError.init(dialStage, "timed out waiting for the peer's stream"))
+  # `withTimeout` on an `acquire` can drop the wait at the instant chronos hands
+  # the lock off, leaving it held with no owner. Release it if the acquire won
+  # that race, so a cancelled or timed-out wait never leaks the peer's lock.
+  let acquireFut = ps.lock.acquire()
+  try:
+    if not await acquireFut.withTimeout(timeout):
+      if acquireFut.completed():
+        ps.releaseLock()
+      return err(SendError.init(dialStage, "timed out waiting for the peer's stream"))
+  except CancelledError as e:
+    if acquireFut.completed():
+      ps.releaseLock()
+    raise e
   defer:
     ps.releaseLock()
 
--- a/libp2p/protocols/kademlia/message_sender.nim
+++ b/libp2p/protocols/kademlia/message_sender.nim
@@ -14,6 +14,7 @@ import std/tables
 import chronos, chronicles, results
 import ../../[peerid, switch]
 import ../../stream/connection
+import ../../utils/future
 import ./kademlia_metrics
 
 export results
@@ -55,6 +56,8 @@ type
     maxMsgSize: int
     maxReuseFailures: int
     senders: Table[PeerId, PeerMessageSender]
+    cleanups: seq[Future[void]]
+      ## Streams given up outside an RPC's own turn: `stop` waits for them.
     stopped: bool
 
   PreparedStream = tuple[stream: Stream, reused: bool]
@@ -81,17 +84,26 @@ proc discardStream(
     ps.stream = nil
   await stream.reset()
 
-proc retireStream(
-    ps: PeerMessageSender, stream: Stream, deadline: Moment
-) {.async: (raises: []), gcsafe.} =
+proc retireStream(stream: Stream, deadline: Moment) {.async: (raises: []), gcsafe.} =
   ## Close rather than reset, so a message written just before still reaches the
   ## remote, then wait for the remote's EOF so nothing stays buffered on the
   ## muxer. A remote that never answers the close gets a reset instead of an
   ## unbounded wait.
-  if ps.stream == stream:
-    ps.stream = nil
   if not await noCancel stream.closeWithEOF().withTimeout(deadline.timeLeft()):
     await stream.reset()
+
+proc trackCleanup(ms: MessageSender, fut: Future[void]) {.raises: [].} =
+  ms.cleanups.trackFut(fut)
+
+proc retire(
+    ms: MessageSender, ps: PeerMessageSender, stream: Stream, timeout: Duration
+) {.raises: [].} =
+  ## Give the stream up in the background. Retiring it in place would make the
+  ## next RPC to this peer wait for the remote's EOF before it even starts, and
+  ## a batch of one-way messages would then go out one round trip at a time.
+  if ps.stream == stream:
+    ps.stream = nil
+  ms.trackCleanup(retireStream(stream, Moment.now() + timeout))
 
 proc dropPeer*(ms: MessageSender, peerId: PeerId) {.async: (raises: []), gcsafe.} =
   ## Forget a peer's stream. The RPC holding it, if any, fails on its next
@@ -106,12 +118,17 @@ proc dropPeer*(ms: MessageSender, peerId: PeerId) {.async: (raises: []), gcsafe.
     await stream.reset()
 
 proc start*(ms: MessageSender) {.raises: [].} =
+  ## `new` already registered the handler, so only a restart re-registers it.
+  if not ms.stopped:
+    return
   ms.stopped = false
   ms.switch.addConnEventHandler(ms.onDisconnect, ConnEventKind.Disconnected)
 
 proc stop*(ms: MessageSender) {.async: (raises: []), gcsafe.} =
   ## Reject further sends, take the disconnect handler back off the switch and
   ## reset every open stream. `start` reopens it.
+  if ms.stopped:
+    return
   ms.stopped = true
   ms.switch.removeConnEventHandler(ms.onDisconnect, ConnEventKind.Disconnected)
   let senders = move ms.senders
@@ -120,6 +137,10 @@ proc stop*(ms: MessageSender) {.async: (raises: []), gcsafe.} =
     let stream = move ps.stream
     if not stream.isNil():
       await stream.reset()
+
+  # Streams given up outside an RPC's turn close on their own budget; leaving
+  # them behind would outlive the sender and trip the stream trackers.
+  await noCancel allFutures(move ms.cleanups)
 
 proc new*(
     T: typedesc[MessageSender],
@@ -135,10 +156,13 @@ proc new*(
     maxReuseFailures: maxReuseFailures,
   )
   # A dead connection takes its streams with it; drop the entry so the table
-  # stays bounded by the number of connected peers.
+  # stays bounded by the number of connected peers. A peer can hold more than
+  # one connection, and the stream on the surviving one is still good.
   ms.onDisconnect = proc(
       peerId: PeerId, event: ConnEvent
   ) {.async: (raises: [CancelledError]).} =
+    if ms.switch.isConnected(peerId):
+      return
     await ms.dropPeer(peerId)
 
   switch.addConnEventHandler(ms.onDisconnect, ConnEventKind.Disconnected)
@@ -161,6 +185,16 @@ proc releaseLock(ps: PeerMessageSender) {.raises: [].} =
     ps.lock.release()
   except AsyncLockError as e:
     raiseAssert "peer message sender lock released without acquire: " & e.msg
+
+proc dropLateDial(dialFut: Future[Stream]) {.async: (raises: []), gcsafe.} =
+  ## Wait out a dial we stopped waiting for and reset what it opened.
+  ## Cancelling it instead would close the connection the dialer reused, which
+  ## takes down every other stream the peer holds on it.
+  try:
+    let stream = await noCancel dialFut
+    await stream.reset()
+  except CatchableError:
+    discard
 
 proc prepStream(
     ms: MessageSender,
@@ -185,21 +219,21 @@ proc prepStream(
   if timeLeft.isZero():
     return err(SendError.init(dialStage, "timed out before dialing"))
 
-  # `Dialer.dial` closes whatever it opened when the dial is cancelled, so
-  # neither the timeout nor a cancellation leaves a half-opened channel behind.
   let dialFut = ms.switch.dial(peerId, addrs, ms.codec)
+  let shielded = noCancel dialFut
   let dialed =
     try:
-      await dialFut.withTimeout(timeLeft)
+      await shielded.withTimeout(timeLeft)
     except CancelledError as e:
-      await noCancel dialFut.cancelAndWait()
+      ms.trackCleanup(dropLateDial(dialFut))
       raise e
   if not dialed:
+    ms.trackCleanup(dropLateDial(dialFut))
     return err(SendError.init(dialStage, "timed out dialing the peer"))
 
   let stream =
     try:
-      await dialFut
+      await shielded
     except DialFailedError as e:
       return err(SendError.init(dialStage, e.msg))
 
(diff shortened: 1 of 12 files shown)


libp2p/protocols/kademlia/rpc.nim
```diff
@@ -0,0 +1,38 @@
...
+
+  withRpcSlot(kad)
+  let encoded = msg.encode(kad.config.hideConnectionStatus)
+
+  kad_messages_sent.inc(labelValues = [$msgType])
+  kad_message_bytes_sent.inc(encoded.len.int64, labelValues = [$msgType])
+
+  var sendRes: Result[seq[byte], SendError]
+  kad_message_duration_ms.time(labelValues = [$msgType]):
+    sendRes = await kad.msgSender.sendRequest(peer, addrs, encoded, kad.config.timeout)
```
vladopajic: `... ,move encoded, ....`

its safe to move since encoded bytes are created and owned by this proc.

libp2p/protocols/kademlia/message_sender.nim
```diff
@@ -0,0 +1,298 @@
...
+  try:
+    ok(await readFut)
+  except LPStreamError as e:
+    err(sendError(readStage, e.msg))
+
+proc send(
+    ms: MessageSender,
+    peerId: PeerId,
+    addrs: seq[MultiAddress],
+    payload: seq[byte],
```
vladopajic: `payload` can be `sink`

libp2p/protocols/kademlia/message_sender.nim
```diff
@@ -0,0 +1,298 @@
...
+    except CancelledError as e:
+      await noCancel readFut.cancelAndWait()
+      raise e
+  if not replied:
+    return err(sendError(readStage, "timed out waiting for reply"))
+
+  try:
+    ok(await readFut)
+  except LPStreamError as e:
+    err(sendError(readStage, e.msg))
```
vladopajic: `sendError(readStage, e.msg)` appears like error is going to be sent to somewhere. 

consider: `SendError.init(....)` to avoid misconception and as general rule from nim style guide.

libp2p/protocols/kademlia/message_sender.nim
```diff
@@ -0,0 +1,298 @@
...
+  let onDisconnect = proc(
+      peerId: PeerId, event: ConnEvent
+  ) {.async: (raises: [CancelledError]).} =
+    await ms.dropPeer(peerId)
+
+  switch.addConnEventHandler(onDisconnect, ConnEventKind.Disconnected)
+  ms
+
+proc senderFor(ms: MessageSender, peerId: PeerId): PeerMessageSender {.raises: [].} =
+  ms.senders.mgetOrPut(peerId, PeerMessageSender(peerId: peerId, lock: newAsyncLock()))
```
vladopajic: `peerId` seems duplicated information in table entry, as key and `peerId` prop

libp2p/protocols/kademlia/message_sender.nim
```diff
@@ -0,0 +1,298 @@
...
+  ms.stopped = false
+
+proc stop*(ms: MessageSender) {.async: (raises: []), gcsafe.} =
+  ## Reject further sends and reset every open stream. `start` reopens it.
+  ms.stopped = true
+  let senders = move ms.senders
+  for ps in senders.values():
+    ps.invalidated = true
+    let stream = ps.stream
+    ps.stream = nil
```
vladopajic:
```suggestion
    let stream = move ps.stream
```

this can be one liner

libp2p/protocols/kademlia/message_sender.nim
```diff
@@ -0,0 +1,298 @@
...
+    ## How far a send got before it gave up. Callers distinguish these: an RPC
+    ## that reached the remote but drew no reply differs from one that never
+    ## left the node.
+    dialStage
+    writeStage
+    readStage
+
+  SendError* = object
+    stage*: SendStage
+    msg*: string
```
vladopajic, line 41: what was reasoning to introduce complexity with custom `SendError`?
- it looks like it is only used in tests, in only two places, why not assert error using error message? this seems even harder assertation compared to checking same stage as error on same stage can happen due to many reasons.
- stringified value could be created by `errDial(msg)`, `errWrite(msg)`, `errRead(msg)` templates
   ```
   errWrite("timed out waiting for reply")
      vs
   sendError(readStage, "timed out waiting for reply")
   ```
- if `.stage` was needed by library code (not tests), usage of custom error would be easier to justify

vladopajic (approved):

libp2p/protocols/kademlia/message_sender.nim
```diff
@@ -0,0 +1,298 @@
...
+    if not (existing.closed() or existing.atEof()):
+      kad_stream_reuses.inc()
+      return ok((existing, true))
+    await ps.discardStream(existing)
+
+  # Shield the dial from cancellation: interrupting it mid-handshake leaks the
+  # half-opened channel, because nothing holds the stream yet to reset it.
+  let stream =
+    try:
+      await noCancel ms.switch.dial(ps.peerId, addrs, ms.codec)
```
richard-ramos: This `noCancel` is a bit strange, because assuming `dropPeer` (or `stop`) is executed while this `await` is awaited, `stream` will be still stored in ps.stream and used

libp2p/protocols/kademlia/message_sender.nim
```diff
@@ -0,0 +1,298 @@
...
+    ps.stream = nil
+  await stream.reset()
+
+proc retireStream(
+    ps: PeerMessageSender, stream: Stream
+) {.async: (raises: []), gcsafe.} =
+  ## Give up a stream whose reply was fully read, leaving nothing buffered.
+  if ps.stream == stream:
+    ps.stream = nil
+  await stream.close()
```
richard-ramos: Shouldnt this be a closeWithEOF?

libp2p/protocols/kademlia/message_sender.nim
```diff
@@ -0,0 +1,298 @@
...
+
+    if sendRes.isErr():
+      await ps.discardStream(stream)
+      if not reused:
+        return err(sendRes.error())
+      # A reused stream also fails when the remote quietly dropped it while it
+      # sat idle; one fresh attempt tells that apart from a real RPC failure.
+      ps.reuseFailures.inc()
+      kad_stream_reuse_failures.inc()
+      continue
```
richard-ramos, line 363: This is not ideal for read timeouts, as you'll do an extra dial and write a duplicate rpc to a peer that is slow. Probably makes sense to not retry if `deadline.timeLeft() == 0` or even better, only retry if the failure hapens on write stage or when receiving an EOF error.

libp2p/protocols/kademlia/message_sender.nim
```diff
@@ -0,0 +1,298 @@
...
+    maxReuseFailures: maxReuseFailures,
+  )
+  # A dead connection takes its streams with it; drop the entry so the table
+  # stays bounded by the number of connected peers.
+  let onDisconnect = proc(
+      peerId: PeerId, event: ConnEvent
+  ) {.async: (raises: [CancelledError]).} =
+    await ms.dropPeer(peerId)
+
+  switch.addConnEventHandler(onDisconnect, ConnEventKind.Disconnected)
```
richard-ramos: This should be removed on stop

libp2p/protocols/kademlia/message_sender.nim
richard-ramos: Shouldnt this dial use the `timeout`?

libp2p/protocols/kademlia/message_sender.nim
```diff
@@ -0,0 +1,343 @@
...
+
+proc exchange(
+    stream: Stream,
+    payload: seq[byte],
+    maxMsgSize: int,
+    deadline: Moment,
+    awaitReply: bool,
+): Future[Result[seq[byte], SendError]] {.async: (raises: [CancelledError]), gcsafe.} =
+  try:
+    await stream.writeLp(payload)
```
richard-ramos: This should be covered by `deadline` too.

libp2p/protocols/kademlia/message_sender.nim
```diff
@@ -0,0 +1,343 @@
...
+    codec: codec,
+    maxMsgSize: maxMsgSize,
+    maxReuseFailures: maxReuseFailures,
+  )
+  # A dead connection takes its streams with it; drop the entry so the table
+  # stays bounded by the number of connected peers.
+  ms.onDisconnect = proc(
+      peerId: PeerId, event: ConnEvent
+  ) {.async: (raises: [CancelledError]).} =
+    await ms.dropPeer(peerId)
```
richard-ramos, line 166: Not the usual, but, a peer can have multiple connections, so probably safer to do the following to not drop the streams of connection that is working fine:
```suggestion
    if not ms.switch.isConnected(peerId):
      await ms.dropPeer(peerId)
```

richard-ramos (approved):

richard-ramos:
It seems to me that this last commit did add much more complexity than required for the dial timeout.
But this is something that can be refactored in a subsequent PR

PR #2902, opened 2026-07-30
chore(kad): batch reprovide by keyspace region

### LINKED ISSUE #2857: kad: batch reprovide by keyspace region instead of one lookup per key
`provider.nim:216 manageRepublishProvidedKeys` iterates every provided key and calls `addProvider`, each of which runs a full `findNode` walk. A node providing N keys performs N independent lookups per republish cycle.

go-libp2p's `provider/provider.go` sweeps the keyspace into regions (each holding >= replication peers), batches all keys sharing a region into a single DHT walk, and schedules regions deterministically spread across the reprovide interval (rather than a burst).

Proposed work:
- Group provided keys by shared keyspace prefix/region.
- Perform one lookup per region and fan the ADD_PROVIDER RPCs for all keys in that region out from the discovered closest set.
- (Stretch) spread region reprovides across the interval instead of all-at-once.

Impact: republish cost scales with regions, not keys. Highest for nodes providing many CIDs.

gmelodie:
## Summary

`manageRepublishProvidedKeys` called `addProvider` for every provided key, and each call ran a full `findNode` walk. A node that provides N keys ran N independent lookups on every republish cycle.

This PR groups the provided keys into keyspace regions and runs one walk per region. A region is the set of keys whose hashes share a leading bit prefix. The prefix length comes from the routing table: bucket `d` covers a `2^-(d+1)` slice of the keyspace, so a deepest full bucket `d` puts the network at about `replication * 2^(d+1)` peers, and `d + 1` prefix bits cut that into regions of about `replication` peers each. A peer that shares the prefix is always closer to a key in the region than a peer that does not, so once a region holds at least `replication` peers, every key in it has all of its closest peers inside the region. One walk toward any member key therefore finds the peers of every other member.

Each key still gets its own target set: `closestFirst` re-sorts the peers that the region walk discovered by distance to that key, so no key is advertised to a peer that is farther than its own closest set. Republish cost now scales with regions instead of keys.

The regions also start evenly spread over the first half of the republish interval. This trickles the walks out instead of firing them all at once, and it leaves the second half of the interval as headroom for the last region to finish.

While no bucket is full, the routing table carries no evidence of the network size. The republish then falls back to one walk per key, which is the behaviour before this PR. A region guessed too wide would advertise keys to peers that are not their closest, so the fallback is the safe default.

## Affected Areas

- [x] Protocol Logic

New module `libp2p/protocols/kademlia/keyspace.nim` holds the region maths: `regionPrefix`, `keyspaceRegions`, `regionBits` and `closestFirst`.

`libp2p/protocols/kademlia/provider.nim` gains `providedKeyRegions`, `republishRegion` and the region schedule. The ADD_PROVIDER fan-out that `addProvider` shared with the spillover path moved into `storeProviderAt`, which both the single-key path and the region path now use. `addProvider` keeps its behaviour: with `providerRejection` it walks and spills over, without it it calls `findNode` and tells the `replication` closest peers.

`libp2p/protocols/kademlia/types.nim` gains the `republishRegionBits` config field and the `MaxRegionBits` constant.

No wire format change and no protocol change. The node sends the same ADD_PROVIDER messages to the same kind of peer set; only the number of lookups that finds that set changes. Downstream validation is not needed.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

No breaking API change. `KadDHTConfig` gains one optional field, `republishRegionBits`, which defaults to `Opt.none(int)`. `Opt.none` derives the prefix length from the routing table.

`KadDHT` gains `providedKeyRegions`, which returns the plan for the next republish: the provided keys grouped into the regions that the loop will walk.

Two metrics report the batching: `kad_provider_republish_regions` counts the walks, and `kad_provider_republish_keys` counts the keys those walks served. The ratio between them shows the saving.

The performance gain is largest for nodes that provide many CIDs.

## Risk Assessment

Backward compatibility: no wire change, so a batching node and a non-batching node interoperate.

Network behaviour: a republish cycle now spreads its walks over half the interval instead of starting them together. The heartbeat keeps a fixed period, so the cycle rate does not change.

Correctness: the region argument holds only while a region contains at least `replication` peers. `republishRegionBits` lets an operator override the derived prefix length, and too few bits advertise keys to peers that are not their closest. The field doc states this, and the constructor bounds the value to `0 .. MaxRegionBits`.

Security: a peer that poisons one walk now steers the advertisements of a whole region instead of one key. The per-key limits do not change: the shortlist stays bounded by `maxShortlistSize`, `closestFirst` re-sorts per key, and the fan-out stays bounded by `replication`.

`stopProviding` during a region's scheduled delay is handled: `republishRegion` re-checks the live provided set after the delay and drops the keys that are gone.

## References

Closes #2857. Sub-issue of #1756.

Design follows go-libp2p `provider/provider.go`, which sweeps the keyspace into regions and batches the keys of each region into a single DHT walk.

## Additional Notes

Tests: `tests/libp2p/kademlia/test_keyspace.nim` covers the region maths, including the prefix masking, the partition property and the bucket-derived prefix length. `tests/libp2p/kademlia/test_add_provider.nim` gains a suite that covers the grouping, the one-region-per-key fallback and the delivery of every key in a single region walk.

### DIFF
--- a/libp2p/protocols/kademlia/keyspace.nim
+++ b/libp2p/protocols/kademlia/keyspace.nim
@@ -1,0 +1,57 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Copyright (c) Status Research & Development GmbH
+
+## Keyspace regions: keys whose hashes share a leading bit prefix. Once a region
+## holds ``replication`` peers, every key in it has all of its closest peers
+## inside the region, so one walk toward any member key serves them all.
+
+import std/[algorithm, sequtils, tables]
+import results
+import ../../peerid
+import ./[types, routing_table]
+
+type RegionPrefix* = seq[byte]
+  ## Leading bits of a hashed key, trailing bits of the last byte cleared.
+
+func regionPrefix*(key: Key, bits: int, hasher: Opt[XorDHasher]): RegionPrefix =
+  let hashed = key.hashFor(hasher)
+  let width = clamp(bits, 0, hashed.len * 8)
+  var prefix = hashed[0 ..< width div 8]
+  let partial = width mod 8
+  if partial > 0:
+    prefix.add(hashed[width div 8] and (0xFF'u8 shl (8 - partial)))
+  prefix
+
+func keyspaceRegions*(
+    keys: seq[Key], bits: int, hasher: Opt[XorDHasher]
+): seq[seq[Key]] =
+  ## Partition `keys` by prefix. `bits == 0` yields a single region.
+  var regions: Table[RegionPrefix, seq[Key]]
+  for key in keys:
+    regions.mgetOrPut(key.regionPrefix(bits, hasher), @[]).add(key)
+  regions.values().toSeq()
+
+func regionBits*(rtable: RoutingTable): Opt[int] =
+  ## Prefix length that splits the keyspace into regions of at least
+  ## ``replication`` peers each. Bucket ``d`` covers a ``2^-(d+1)`` slice, so a
+  ## deepest full bucket ``d`` sizes the network at ``replication * 2^(d+1)``
+  ## peers and ``d + 1`` bits cut it into regions of ``replication`` peers.
+  ## ``Opt.none`` while no bucket is full: the table has no size estimate yet.
+  var deepestFull = -1
+  for i in 0 ..< rtable.buckets.len:
+    if rtable.buckets[i].peers.len >= rtable.config.replication:
+      deepestFull = i
+
+  if deepestFull < 0:
+    return Opt.none(int)
+  Opt.some(min(deepestFull + 1, MaxRegionBits))
+
+func closestFirst*(peers: seq[PeerId], key: Key, hasher: Opt[XorDHasher]): seq[PeerId] =
+  ## `peers` ordered by XOR distance to `key`, closest first.
+  peers
+    .mapIt((it, xorDistance(it, key, hasher)))
+    .sorted(
+      proc(a, b: (PeerId, XorDistance)): int =
+        cmp(a[1], b[1])
+    )
+    .mapIt(it[0])
--- a/libp2p/protocols/kademlia/keyspace.nim
+++ b/libp2p/protocols/kademlia/keyspace.nim
@@ -25,11 +25,17 @@ func regionPrefix*(key: Key, bits: int, hasher: Opt[XorDHasher]): RegionPrefix =
 func keyspaceRegions*(
     keys: seq[Key], bits: int, hasher: Opt[XorDHasher]
 ): seq[seq[Key]] =
-  ## Partition `keys` by prefix. `bits == 0` yields a single region.
-  var regions: Table[RegionPrefix, seq[Key]]
+  ## Partition `keys` by prefix, each region in the order its first key appears.
+  ## `bits == 0` yields a single region.
+  var
+    order: seq[RegionPrefix]
+    regions: Table[RegionPrefix, seq[Key]]
   for key in keys:
-    regions.mgetOrPut(key.regionPrefix(bits, hasher), @[]).add(key)
-  regions.values().toSeq()
+    let prefix = key.regionPrefix(bits, hasher)
+    if prefix notin regions:
+      order.add(prefix)
+    regions.mgetOrPut(prefix, @[]).add(key)
+  order.mapIt(regions.getOrDefault(it))
 
 func regionBits*(rtable: RoutingTable): Opt[int] =
   ## Prefix length that splits the keyspace into regions of at least
--- a/libp2p/protocols/kademlia/keyspace.nim
+++ b/libp2p/protocols/kademlia/keyspace.nim
@@ -13,7 +13,9 @@ import ./[types, routing_table]
 type RegionPrefix* = seq[byte]
   ## Leading bits of a hashed key, trailing bits of the last byte cleared.
 
-func regionPrefix*(key: Key, bits: int, hasher: Opt[XorDHasher]): RegionPrefix =
+func init*(
+    T: typedesc[RegionPrefix], key: Key, bits: int, hasher: Opt[XorDHasher]
+): RegionPrefix =
   let hashed = key.hashFor(hasher)
   let width = clamp(bits, 0, hashed.len * 8)
   var prefix = hashed[0 ..< width div 8]
@@ -31,7 +33,7 @@ func keyspaceRegions*(
     order: seq[RegionPrefix]
     regions: Table[RegionPrefix, seq[Key]]
   for key in keys:
-    let prefix = key.regionPrefix(bits, hasher)
+    let prefix = RegionPrefix.init(key, bits, hasher)
     if prefix notin regions:
       order.add(prefix)
     regions.mgetOrPut(prefix, @[]).add(key)
--- a/tests/libp2p/kademlia/test_add_provider.nim
+++ b/tests/libp2p/kademlia/test_add_provider.nim
@@ -472,6 +472,7 @@ suite "KadDHT - Republish By Keyspace Region":
 
     check:
       regions.len < keys.len
+      # 1 prefix bit cuts the keyspace in two, so no key falls outside 2 regions.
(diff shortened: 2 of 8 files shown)


tests/libp2p/kademlia/test_keyspace.nim
```diff
@@ -0,0 +1,122 @@
...
+    # A deeper bucket that is not full carries no evidence of network size.
+    rtable.fillBucket(7, 3)
+    check rtable.regionBits() == Opt.some(6)
+
+  test "closestFirst orders peers by distance to the key":
+    let hasher = Opt.none(XorDHasher)
+    let key = PeerId.random(rng()).get().toKey()
+    var peers: seq[PeerId]
+    for _ in 0 ..< 8:
+      peers.add(PeerId.random(rng()).get())
```
vladopajic:
```suggestion
    let peers = PeerId.random(8, rng())
```

tests/libp2p/kademlia/test_keyspace.nim
```diff
@@ -0,0 +1,122 @@
...
+import ../../tools/[unittest, crypto]
+
+let NoOp = Opt.some(noOpHasher)
+
+proc keyFrom(bytes: openArray[byte]): Key =
+  ## Key whose `noOpHasher` hash starts with `bytes`.
+  var buf: array[IdLength, byte]
+  for i, b in bytes:
+    buf[i] = b
+  @buf
```
vladopajic: ideally, `keyFrom` should be moved to location where `Key` type is defined, this way it could be resued everywhere.

consider following notation `Key.from(byes)` seems nicer then `keyFrom(bytes)`

tests/libp2p/kademlia/test_keyspace.nim
```diff
@@ -0,0 +1,122 @@
...
+
+proc keyInBucket(bucket: int, tag: byte): Key =
```
vladopajic: `keyInBucket` is maker for `Key`. usually code follows convention `makeXxxx` for any helper makes.

tests/libp2p/kademlia/test_keyspace.nim
```diff
@@ -0,0 +1,122 @@
...
+  ## Key landing in `bucket` of an all-zero selfId table. `tag` keeps keys apart.
+  var buf: array[IdLength, byte]
+  buf[bucket div 8] = 0x80'u8 shr (bucket mod 8)
+  buf[IdLength - 1] = tag
+  @buf
+
+proc emptyTable(replication: int): RoutingTable =
```
vladopajic: `emptyTable` -> `makeEmptyTable`

tests/libp2p/kademlia/test_keyspace.nim
```diff
@@ -0,0 +1,122 @@
...
+      keyFrom([0x80'u8]),
+      keyFrom([0xC0'u8]),
+      keyFrom([0xFF'u8]),
+    ]
+    let regions = keys.keyspaceRegions(1, NoOp)
+
+    check:
+      regions.len == 2
+      regions.mapIt(it.len).sorted() == @[2, 3]
+      regions.concat().toHashSet() == keys.toHashSet()
```
vladopajic: a lot of checks, but still doesn't beat this construct:

```
check regions == expectedRegions
```

tests/libp2p/kademlia/test_add_provider.nim
```diff
@@ -449,6 +449,68 @@ suite "KadDHT - Add Provider":
...
+    let kad = setupKad(testKadConfig(republishRegionBits = Opt.some(1)))
+    startAndDeferStop(@[kad])
+
+    let keys = kad.provideWithoutAnnouncing(8)
+    let regions = kad.providedKeyRegions()
+    let hasher = kad.rtable.config.hasher
+
+    check:
+      regions.len < keys.len
+      regions.len <= 2
```
vladopajic, line 474: where does 2 come from?

vladopajic (approved):

tests/libp2p/kademlia/test_add_provider.nim
gmelodie: One key/peer should not belong to more than 2 regions. Since the kad keyspace is linear, if a key belongs to 3 or more regions, it means at least one of those is entirely contained into another

libp2p/protocols/kademlia/keyspace.nim
```diff
@@ -0,0 +1,63 @@
...
+
+import std/[algorithm, sequtils, tables]
+import results
+import ../../peerid
+import ./[types, routing_table]
+
+type RegionPrefix* = seq[byte]
+  ## Leading bits of a hashed key, trailing bits of the last byte cleared.
+
+func regionPrefix*(key: Key, bits: int, hasher: Opt[XorDHasher]): RegionPrefix =
```
vladopajic: this looks like init func for RegionPrefix type

tests/libp2p/kademlia/test_add_provider.nim
```diff
@@ -449,6 +449,69 @@ suite "KadDHT - Add Provider":
...
+    checkTrackers()
+
+  proc provideWithoutAnnouncing(kad: KadDHT, count: int): seq[Key] =
+    ## Seed the store directly, so only the republish loop advertises the keys.
+    var keys: seq[Key]
+    for i in 0 ..< count:
+      let key = MultiHash.digest("sha2-256", @[i.byte]).get().toKey()
+      kad.providerManager.providedKeys.provided[key] = Moment.now()
+      keys.add(key)
+    keys
```
vladopajic:
```
let keys = Key.make(count) # or any utility that creates seq keys 
for key in keys:
  ....
keys
```

pattern for creating keys can repeat, or maybe already is but it is never reused

tests/libp2p/kademlia/test_keyspace.nim
```diff
@@ -0,0 +1,98 @@
...
+# Copyright (c) Status Research & Development GmbH
+
+{.used.}
+
+import std/sets
+import chronos, results
+import ../../../libp2p/[peerid, protocols/kademlia, crypto/crypto]
+import ../../tools/[unittest, crypto]
+
+let NoOp = Opt.some(noOpHasher)
```
vladopajic: NoOp seems like teastless name

vladopajic: what is difference between noop hasher and Opt.none hasher?

tests/libp2p/kademlia/test_keyspace.nim
```diff
@@ -0,0 +1,98 @@
...
+
+  test "region prefix is clamped to the hash length":
+    check Key.init([0xFF'u8]).regionPrefix(MaxRegionBits + 8, NoOp).len == IdLength
+
+  test "keys share a region exactly when their leading bits match":
+    let
+      a = Key.init([0b1010_0000'u8])
+      b = Key.init([0b1011_0000'u8])
+    check:
+      a.regionPrefix(3, NoOp) == b.regionPrefix(3, NoOp)
```
vladopajic: `a.regionPrefixNoHash` as local utility avoids second argument which is always the same

tests/libp2p/kademlia/test_keyspace.nim
```diff
@@ -0,0 +1,98 @@
...
+
+    rtable.fillBucket(5, 4)
+    check rtable.regionBits() == Opt.some(6)
+
+    # A deeper bucket that is not full carries no evidence of network size.
+    rtable.fillBucket(7, 3)
+    check rtable.regionBits() == Opt.some(6)
+
+  test "closestFirst orders peers by distance to the key":
+    let hasher = Opt.none(XorDHasher)
```
vladopajic: ideally Opt.none(T) shouldn't be explicitly passed down, as it is default value.

however, if intereface has default value, callsites may make mistake of not passing down hashers. in other words not have default value might be more verbouse but strict notation.

take it as inspo, for what it is worth.

richard-ramos (approved):

PR #2903, opened 2026-07-30
chore(kad): converge lookups on beta and confirm the k closest in a follow-up phase

### LINKED ISSUE #2855: kad: add beta resiliency + follow-up phase to lookup termination
`find.nim:129 hasResponsesFromClosestAvailable` terminates a lookup only once the closest `replication` (k=20) available peers have all responded. This is strictly more work than needed and has no cheap-converge/then-confirm split.

go-libp2p converges when the closest `beta` (default 3) peers have answered, then runs a separate follow-up phase that queries/connects the remaining top-K to confirm the true closest set (used for Provide/PutValue placement).

Proposed work:
- Introduce a `beta` config param (default ~3); terminate the core lookup when the closest `beta` peers have responded (plus starvation termination).
- Add a follow-up phase that queries the remaining top-`replication` heard peers before returning, so placement operations still target the real K-closest.

Impact: lookup cost and robustness against a single lying/dead closest peer. High.

gmelodie:
## Summary

`find.nim` ended a lookup only after the closest `replication` (k=20) available peers had all answered. That is more work than the lookup needs, and one slow, dead or lying closest peer held the whole lookup back.

go-libp2p splits this in two: the core lookup converges when the closest `beta` (default 3) peers answer, then a follow-up phase queries the rest of the k closest peers, so placement operations still target the real k-closest set. This PR does the same.

- New `beta` config parameter (`DefaultBeta = 3`). The core lookup now ends when the closest `beta` peers answered. `KadDHTConfig.new` asserts `1 <= beta <= replication`, since the follow-up phase never reaches past the k closest peers.
- New follow-up phase in `iterativeLookup`. When the core phase converges, the lookup sweeps the `replication` closest peers that did not answer yet. A sweep targets a fixed set, so a peer heard about mid-sweep waits for the next round and one sweep cannot grow without end.
- The two phases alternate until the k closest peers the lookup knows of have all been queried. A sweep that reveals a closer peer reopens the core phase, so a stateful caller keeps making progress. No peer is swept twice, which bounds the alternation.
- Convergence is now internal to `iterativeLookup`, so `getValue` and `getProviders` get it too. Their own stop conditions stay as external early exits: an external stop ends the lookup on the spot, follow-up included, because the caller already has what it asked for.
- The `StopCond` parameter is renamed to `EarlyExit` and made optional, because it changed role. It used to carry the termination rule itself (`closestAvailableStop`); now termination is internal, and the parameter is only an external early exit. The three call sites that want the k closest peers and nothing else pass nothing (`findNode`, `addProviderSpillover`, service discovery `randomRecords`).
- New `kad_lookup_followups` counter, incremented once per lookup that reaches the follow-up phase.

## Affected Areas

- [x] Protocol Logic
  

The change is internal to the Kademlia lookup loop. It changes no wire message and no public protocol behaviour, so no downstream integration branch is needed.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

- New optional `beta` parameter on `KadDHTConfig.new`, default 3. Existing code compiles unchanged. A `beta` outside `1 .. replication` now fails the constructor assert.
- `StopCond` becomes `EarlyExit`, `closestAvailableStop` is removed, and `noEarlyExit` is the default. Drop the argument: `iterativeLookup` applies beta convergence and the follow-up phase on its own, so a caller passes this only to stop sooner than that.
- `hasResponsesFromClosestAvailable(state)` becomes `hasResponsesFromClosest(state, amount)`, with `hasConverged(state)` for the `beta` case.
- Lookups cost fewer queries and finish sooner. `findNode` still returns the k closest peers it found.

## Risk Assessment

- Backward compatibility: source-level only, on the two lookup helpers listed above. No wire format change, so mixed-version networks are unaffected.
- Network behaviour: lookups send fewer RPCs and stop earlier. `getValue` and `getProviders` converge on `beta` first, then confirm the k closest peers. They still walk the shortlist until their quorum is met, because the phases alternate while the stop condition is unmet.
- Security: convergence at `beta` gives the closest few peers more say over when the core phase ends. The follow-up phase limits the effect, because placement still targets the k closest peers. This is the same trade-off go-libp2p makes.
- Termination: a sweep never targets a peer a previous sweep already targeted, so the number of follow-up rounds is bounded by the shortlist cap.
- Performance: no regression is expected. The follow-up phase queries a subset of the peers the old termination rule waited for, and it queries each of them once per retry budget instead of waiting for success.

## References

- Closes #2855
- Parent issue: #1756
- go-libp2p reference: `runLookupWithFollowup` and `isLookupTermination` in go-libp2p-kad-dht

## Additional Notes

Test coverage added in `tests/libp2p/kademlia/test_lookup.nim`:

- convergence on `beta`, including the clamp at `replication`
- the follow-up phase targets the k closest peers that did not answer
- the follow-up phase defers a peer it hears about to the next round
- a lookup keeps going for a stop condition that only the follow-up phase can unblock
- a lookup converges past a closest peer that does not answer
- an external stop condition skips the follow-up phase

The tests drive `iterativeLookup` with a recording dispatch proc, so they need no network.


libp2p/protocols/kademlia/find.nim
```diff
@@ -124,27 +124,38 @@ proc selectCloserPeers*(
...
+  # Both config fields are public and mutable, so pin the range at the point of use.
+  let beta = max(1, min(state.kad.config.beta, state.kad.config.replication))
+  state.hasResponsesFromClosest(beta)
 
-  return closetsRespondedCnt >= state.kad.config.replication
+proc followUpPeers*(state: LookupState): HashSet[PeerId] {.raises: [].} =
+  ## The `replication` closest peers heard about that never answered.
+  var peers = initHashSet[PeerId]()
+  for pid in state.selectCloserPeers(
+    state.kad.config.replication, excludeResponded = false
```
vladopajic: why name second argument but first is not named?

libp2p/protocols/kademlia/find.nim
```diff
@@ -498,8 +509,8 @@ proc noopReply*(
 ): Future[void] {.async: (raises: []), gcsafe.} =
   discard
 
-proc closestAvailableStop*(state: LookupState): bool {.raises: [], gcsafe.} =
-  state.hasResponsesFromClosestAvailable()
+proc noStop*(state: LookupState): bool {.raises: [], gcsafe.} =
+  false
```
vladopajic: why was this changed. this now indicates that this argument in iterative lookup is unnecessary, since it doesn't really do anything?

libp2p/protocols/kademlia/find.nim
```diff
@@ -681,14 +727,22 @@ proc iterativeLookup*(
...
+    kad.fillSlots(
+      state, pending, dispatch, state.nextQueryTargets(phase, followUp, stopCond)
+    )
+
+    if pending.activePeers().len > 0:
+      await awaitProgress(pending)
+      continue
 
-    if pending.activePeers().len == 0:
+    if phase == FollowUp or stopCond(state):
```
richard-ramos: This is gonna cause the iterativeLookup to exit after the follow up even if the stopCond(state) is false

For example with getValue and getProvider you can get in a reply closer peers, but those peers are not gonna be queried because FollowUp only queries the values returned by followUpPeers(), so you can end up returning before the quorum or provider is reached

richard-ramos: Here's a vibecoded test unit that demonstrates the behavior 

```nim
  asyncTest "Lookup keeps going when follow-up discovers a peer needed by stop condition":
    let kad = setupLookupKad()

    let targetKey = randomPeerId().toKey()
    # `closest` is closer to the target than every seeded peer, but it is only
    # learned from a follow-up reply, not from the initial routing table.
    let (closest, known) = kad.seedRoutingTableBelow(5, targetKey)

    let queried = new(seq[PeerId])
    # `known[2]` is queried during the follow-up phase and returns `closest`.
    # A lookup that still honors `stopCond` should continue and query `closest`.
    let dispatch = recordingDispatch(queried, {known[2]: @[closest]}.toTable())
    let stopAfterClosestReply = proc(state: LookupState): bool {.gcsafe.} =
      state.responded.hasKey(closest)

    let state = await kad.iterativeLookup(
      targetKey, dispatch, noopReply, stopAfterClosestReply
    )

    check:
      # The PR currently satisfies only this first assertion: `closest` was
      # discovered, but the follow-up phase exits before querying it.
      state.shortlist.hasKey(closest)
      closest in queried[]
      state.responded.hasKey(closest)
```

Output:
```
Check failed: closest in queried[]
Check failed: state.responded.hasKey(closest)
```

richard-ramos: And another one but this time using getValue:
```nim
  asyncTest "GetValue keeps querying closer peers discovered during follow-up":
    let
      requester = setupKad(testKadConfig(replication = 3))
      holder = setupKad()
      key = holder.rtable.selfId
      value = @[1.byte, 2, 3, 4, 5]

    requester.config.alpha = 1
    requester.config.beta = 1

    var hops = @[setupMockKad(), setupMockKad(), setupMockKad()]
    hops.sort(
      proc(a, b: MockKadDHT): int =
        cmp(
          xorDistance(a.switch.peerInfo.peerId, key, requester.rtable.config.hasher),
          xorDistance(b.switch.peerInfo.peerId, key, requester.rtable.config.hasher),
        )
    )

    for hop in hops:
      hop.dataTable.insert(key, value, Timestamp.now())
    holder.dataTable.insert(key, value, Timestamp.now())

    startAndDeferStop(
      @[requester, KadDHT(hops[0]), KadDHT(hops[1]), KadDHT(hops[2]), holder]
    )

    for hop in hops:
      await connect(requester, KadDHT(hop))

    # The first hop lets the core phase converge. The second hop is queried in
    # follow-up and reveals `holder`, which is closer than every initial hop.
    # Quorum 4 is only reachable if `getValue` continues and queries `holder`.
    hops[1].getValueResponse = Opt.some(
      Message(
        msgType: MessageType.getValue,
        key: key,
        record: protobuf.Record(
          key: key, value: value, timeReceived: Timestamp.now()
        ),
        closerPeers: @[holder.switch.peerInfo.toPeer()],
      )
    )

    let record = await requester.getValue(key, quorumOverride = Opt.some(4))

    check record.isOk()
    if record.isOk():
      check record.get().value == value
```

Output:
```
Check failed: record.isOk()
```
And a `Not enough valid records to achieve quorum, needed 4 got 3`

vladopajic (approved):

libp2p/protocols/kademlia/find.nim
```diff
@@ -681,14 +747,17 @@ proc iterativeLookup*(
...
+      kad.fillSlots(state, pending, dispatch, phases.targets(state))
+
+    # Dispatching nothing new only stops the lookup once the RPCs already in
+    # flight have drained, so the returned peer set stays complete.
+    if pending.activePeers().len > 0:
+      await awaitProgress(pending)
+      continue
 
-    if pending.activePeers().len == 0:
+    if earlyExit(state) or not phases.advance(state):
```
richard-ramos, line 759: > since the follow-up phase never reaches past the k closest peers.

Is this ideal tho. Wouldnt this condition stop the query from going further than k even if the quorum is not reached?

With this change it seems to me that getValue and getProviders will no longer query peers beyond closest k peers when quorum was not met, while this was possible previously

richard-ramos: Vibecoded unit test that demonstrates the issue:

```nim
  asyncTest "Lookup continues past k while early-exit remains false":
    let kad =
      setupKad(testKadConfig(replication = 3, alpha = 1, beta = 1))

    let target = randomPeerId().toKey()

    # Only the closest three peers seed the lookup. The fourth peer remains
    # outside k and can only become relevant after a response mentions it.
    let peers = kad.seedRoutingTable(4, target)
    let farther = peers[^1]

    let queried = new(seq[PeerId])

    # The first response introduces `farther`. It is retained in the shortlist,
    # but it is still outside the closest k.
    let dispatch = recordingDispatch(queried, {peers[0]: @[farther]}.toTable())

    # This represents an external condition such as a value quorum that can only
    # be satisfied by querying a peer beyond the closest k.
    let stopAfterFartherReply = proc(state: LookupState): bool {.gcsafe.} =
      state.responded.hasKey(farther)

    let state = await kad.iterativeLookup(
      target, dispatch, noopReply, stopAfterFartherReply
    )

    check:
      # Confirms that the lookup learned about the peer.
      state.shortlist.hasKey(farther)

      # Fails on this PR: after confirming the closest k, the lookup exits even
      # though its external early-exit condition remains false.
      farther in queried[]

```
Output
```
1 tests run: 0 OK, 1 FAILED
Check failed: farther in queried[]
```

richard-ramos: And this is a test that uses `getValue`:
```nim
  asyncTest "GetValue continues past k while quorum remains unmet":
    let
      requester = setupKad(testKadConfig(replication = 3, alpha = 1, beta = 1))
      key = randomPeerId().toKey()
      value = @[1.byte, 2, 3, 4, 5]

    # Sort four peers by distance to the record key. Only the closest three
    # initially seed the lookup; `holder` is deliberately outside k.
    var peers = @[setupMockKad(), setupMockKad(), setupMockKad(), setupMockKad()]
    peers.sort(
      proc(a, b: MockKadDHT): int =
        cmp(
          xorDistance(a.switch.peerInfo.peerId, key, requester.rtable.config.hasher),
          xorDistance(b.switch.peerInfo.peerId, key, requester.rtable.config.hasher),
        )
    )
    let
      closest = peers[0 .. 2]
      holder = peers[3]

    for peer in peers:
      peer.dataTable.insert(key, value, Timestamp.now())

    # The first response tells the requester about the farther holder. The first
    # three peers provide only three records, so quorum still requires it.
    closest[0].getValueResponse = Opt.some(
      Message(
        msgType: MessageType.getValue,
        key: key,
        record: Opt.some(
          protobuf.Record(key: key, value: value, timeReceived: Timestamp.now())
        ),
        closerPeers: @[holder.switch.peerInfo.toPeer()],
      )
    )

    startAndDeferStop(
      @[
        requester, KadDHT(peers[0]), KadDHT(peers[1]), KadDHT(peers[2]),
        KadDHT(holder),
      ]
    )
    for peer in closest:
      await connect(requester, KadDHT(peer))

    let record = await requester.getValue(key, quorumOverride = Opt.some(4))

    # Fails with only three records: the known holder is not queried
    # after the follow-up confirms the closest k peers.
    check record.isOk()
```

Output
```
1 tests run: 0 OK, 1 FAILED
Check failed: record.isOk()
```

PR #2904, opened 2026-07-30
fix(logs): adding logLevel to Libp2pConfig to avoid logs created in createLibp2pNode

vladopajic:
this is fix for case were `nim-ffi` initializes each context thread with `debug` text logging. some libp2p services can emit logs during `libp2p_ctx_create`:
> DBG ... Setting up WildcardAddressResolverService topics="libp2p wildcardresolverservice"

 before the host receives a context pointer and before it can reapply a desired log level.

to avoid this, this PR adds a `logLevel` field to the cbind `Libp2pConfig` and applies it at the start of `libp2pNew`, before `createLibp2pNode(config)` builds/mounts libp2p services.

this lets callers configure the chronicles runtime log level before node construction emits logs, instead of relying only on the existing post-construction `libp2pSetLogLevel` static call.

### DIFF
--- a/cbind/libp2p.nim
+++ b/cbind/libp2p.nim
@@ -382,20 +382,27 @@ proc createLibp2pNode(config: Libp2pConfig): Result[LibP2P, string] =
 
   ok(lib)
 
-proc libp2pSetLogLevel*(level: int): Future[Result[bool, string]] {.ffiStatic.} =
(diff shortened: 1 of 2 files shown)


gmelodie (approved):
a mazing!

richard-ramos (approved):

PR #2905, opened 2026-07-31
fix(test): guard the logging test on a dynamic chronicles sink

gmelodie:
## Summary

The daily job "Daily test all (latest dependencies)" fails to build `tests/test_all` on every matrix entry ([run 30612866362](https://github.com/vacp2p/nim-libp2p/actions/runs/30612866362)):

```
tests/libp2p/test_logging.nim(28, 37) Error: undeclared field: 'writer' for type log_output.StdOutOutput
```

`tests/libp2p/test_logging.nim` captures log records through the `writer` field of a chronicles sink. Only `DynamicOutput` has that field. `tests/config.nims` supplies one with `-d:chronicles_sinks=textlines[stdout],json[dynamic]`, but it applies that config only when the command line carries no chronicles define. The daily workflow passes `-d:chronicles_log_level=DEBUG` (`.github/workflows/daily_test_all_latest_deps_matrix.yml:120`), so the sink config falls back to the chronicles default of one `textlines[stdout]` sink. The test then assigns to `StdOutOutput.writer`, which does not exist, and the build stops.

The same define also turns off `chronicles_runtime_filtering`. That makes `libp2p_logging.setLogLevel` raise its own `{.error.}`, which is the next failure after the field error.

## Changes

The test now finds the dynamic sink with `compiles()` instead of a sink-count guess, and it stores the index in the `dynamicSink` const. It runs its body only when a dynamic sink and runtime filtering are both available; otherwise it calls `skip()`, so the report shows a skipped test instead of a broken build. The unused `std/typetraits` import is gone, because `UnusedImport` is a warning-as-error in `config.nims`.

Two further problems in the test are fixed here as well, because both make it unreliable inside the single `test_all` binary:

- The capture proc discarded the record and counted every DEBUG and INFO log on the thread. The writer stays installed for the whole binary, so `check deliveredInfo == 1` depended on no other suite emitting an INFO record. The proc now counts only records that carry the `logMarker` prefix.
- The test set the global chronicles level to INFO and never restored it, which suppressed DEBUG and TRACE output for every test file that ran after it. A `teardown` now restores `chronicles.enabledLogLevel`, the compile-time default.

The regular CI keeps full coverage. `test_all.yml` and `daily_test_all_no_flags.yml` pass no chronicles define, so `tests/config.nims` installs the dynamic sink and the test runs as before.

## Affected Areas

- [x] Build / Tooling
  Test only. The change touches `tests/libp2p/test_logging.nim` and nothing else.

N/A. The diff contains no production code.

## Impact on Library Users

No impact. The change is limited to the test suite.

## Risk Assessment

Low. The only risk is a silent loss of coverage if a build drops the dynamic sink without notice. The `skip()` call makes that visible in the test report.

## References

- Failing run: https://github.com/vacp2p/nim-libp2p/actions/runs/30612866362
- Test added in #2896

## Additional Notes

I did not run the test suite. I compile-checked `tests/libp2p/test_logging.nim` with Nim 2.2.6 in both configurations:

- default config: the test compiles and stays active
- `-d:chronicles_log_level=DEBUG` plus `--styleCheck:error`: the test compiles and skips

`nph` reports the file as formatted.

### DIFF
--- a/tests/libp2p/test_logging.nim
+++ b/tests/libp2p/test_logging.nim
@@ -3,38 +3,61 @@
 
 {.used.}
 
-import std/typetraits
 import chronicles
-import ../../libp2p/logging as libp2p_logging
-import ../tools/unittest
 
 {.push raises: [].}
 
-var
-  deliveredDebug {.threadvar.}: int
-  deliveredInfo {.threadvar.}: int
-
-let captureLog = proc(logLevel: LogLevel, msg: LogOutputStr) {.gcsafe, raises: [].} =
-  discard msg
-  case logLevel
-  of DEBUG:
-    inc deliveredDebug
-  of INFO:
-    inc deliveredInfo
+# Only a dynamic sink has a writer, and tests/config.nims drops that sink when the build sets its own chronicles defines.
+const dynamicSink =
+  when compiles(defaultChroniclesStream.outputs[0].writer):
+    0
+  elif compiles(defaultChroniclesStream.outputs[1].writer):
+    1
   else:
-    discard
+    -1
 
-when defaultChroniclesStream.outputs.type.arity == 1:
-  defaultChroniclesStream.outputs[0].writer = captureLog
-elif defaultChroniclesStream.outputs.type.arity == 2:
-  defaultChroniclesStream.outputs[1].writer = captureLog
+const canCaptureLogs = dynamicSink >= 0 and chronicles.runtimeFilteringEnabled
 
-suite "Logging":
-  test "setLogLevel changes Chronicles runtime threshold":
-    libp2p_logging.setLogLevel(INFO)
+when canCaptureLogs:
+  import std/strutils
+  import ../../libp2p/logging as libp2p_logging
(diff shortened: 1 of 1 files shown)


tests/libp2p/test_logging.nim
```diff
@@ -24,17 +28,35 @@ let captureLog = proc(logLevel: LogLevel, msg: LogOutputStr) {.gcsafe, raises: [
...
+    -1
+
+const canCaptureLogs = dynamicSink >= 0
+
+when canCaptureLogs:
+  defaultChroniclesStream.outputs[dynamicSink].writer = captureLog
 
 suite "Logging":
+  teardown:
+    when canCaptureLogs and chronicles.runtimeFilteringEnabled:
```
vladopajic: maybe do this once on whole suite. or even better do this as soon as possible, so nothing gets compiled not even variables if impossible to test - in this case add warning

vladopajic (approved):

richard-ramos (approved):

PR #2907, opened 2026-07-31
chore(kad): refactor probe & evict scheduling

SionoiS:
This is an alternative scheduling mechanism for probing and eviction of stale peers.

WDYT?

### DIFF
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -177,9 +177,11 @@ proc maintainLiveness(kad: KadDHT) {.async: (raises: [CancelledError]).} =
   ## bucket refresh so dead peers are not held until the next refresh tick.
   ## At most one probe is in flight per peer across all maintainable tables.
 
+  var idle = true
   while not kad.stopping:
-    # Initial idle: let start() finish bootstrap before the first scan.
-    await sleepAsync(kad.config.livenessIdleInterval)
+    if idle:
+      # Initial idle: let start() finish bootstrap before the first scan.
+      await sleepAsync(kad.config.livenessIdleInterval)
 
     let grace = kad.config.livenessGracePeriod
 
@@ -203,6 +205,9 @@ proc maintainLiveness(kad: KadDHT) {.async: (raises: [CancelledError]).} =
       except CancelledError as exc:
         await noCancel inFlight.cancelAndWait()
         raise exc
+      idle = false
+    else:
+      idle = true
 
 proc refreshTable*(
     kad: KadDHT, rtable: RoutingTable, forceRefresh = false
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -27,37 +27,22 @@ proc livenessCandidates(
   var peers: seq[PeerId]
   for bucket in rtable.buckets:
     for entry in bucket.peers:
-      if not entry.isReplaceable(gracePeriod, now):
+      if not entry.needsLivenessCheck(gracePeriod, now):
         continue
       entry.nodeId.toPeerId().withValue(pid):
         peers.add(pid)
   peers
 
-template withProbeSlotOrReturn*(kad: KadDHT) =
-  ## Acquire one ``probeSem`` slot for the enclosing scope, or return without
-  ## probing. Non-blocking: candidates that find no free slot are retried on a
-  ## later pass rather than queued.
-  if not kad.probeSem.tryAcquire():
-    kad_routing_table_liveness_probes.inc(labelValues = ["skipped"])
-    return
+proc checkAndEvictPeer(
+    kad: KadDHT, rtable: RoutingTable, peerId: PeerId
+) {.async: (raises: [CancelledError]).} =
+  ## Probe one routing-table peer; remove it if unreachable or not speaking DHT.
+  ## Takes ownership of one ``probeSem`` slot already acquired by the caller.
   defer:
     try:
       kad.probeSem.release()
     except AsyncSemaphoreError:
       raiseAssert "probeSem released without acquire"
-
-proc checkAndEvictPeer(
-    kad: KadDHT, rtable: RoutingTable, peerId: PeerId
-) {.async: (raises: [CancelledError]).} =
-  withProbeSlotOrReturn(kad)
-
-  # Candidate list is a snapshot; by the time a slot is free the peer may have
-  # been refreshed (markUseful) or removed, and stop may have begun.
-  if kad.stopping:
-    return
-  let grace = kad.config.livenessGracePeriod
-  if not rtable.isReplaceable(peerId, grace, Moment.now()):
-    return
 
   let addrs = kad.switch.peerStore[AddressBook][peerId]
   if addrs.len == 0:
@@ -71,15 +56,11 @@ proc checkAndEvictPeer(
     kad_routing_table_liveness_probes.inc(labelValues = ["ok"])
     return
 
-  # Probe can race with unrelated traffic that markUseful'd the peer mid-flight.
-  if not rtable.isReplaceable(peerId, grace, Moment.now()):
-    return
-
   trace "Evicting unresponsive peer after liveness probe", peer = peerId.shortLog()
   discard rtable.removePeer(peerId, reason = "liveness")
   kad_routing_table_liveness_probes.inc(labelValues = ["fail"])
 
-proc probeAndEvictPeers*(
+proc pingAndEvictPeers*(
     kad: KadDHT, rtable: RoutingTable
 ) {.async: (raises: [CancelledError]).} =
   ## Probes routing-table peers that have been quiet longer than the liveness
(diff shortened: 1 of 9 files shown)


libp2p/protocols/kademlia.nim
```diff
@@ -79,11 +91,29 @@ proc checkAndEvictPeer(
...
+  ## has no free slot — the caller should wait and retry.
+  let probeKey: ProbeKey = (rtable.selfId, peerId)
+  if kad.livenessProbes.hasKey(probeKey):
+    return true
+  if kad.stopping:
+    return true
+  if not kad.probeSem.tryAcquire():
+    kad_routing_table_liveness_probes.inc(labelValues = ["skipped"])
+    return false
+  kad.trackLivenessProbe(probeKey, kad.checkAndEvictPeer(rtable, peerId))
```
vladopajic: i would always avoid having lock and unlock in different places. it can lead to locking issues.

for example:
`kad.checkAndEvictPeer(rtable, peerId)` can return (bool, and future)
if bool is false then future was not started (lock was not acquired, future is nil?)
if bool is true then future was started

but it seems to me that you can just return future, if lock is not acquired then it early returns and future finishes fast. there is small overhead that creates this future but we should not care about that.

libp2p/protocols/kademlia.nim
```diff
@@ -102,13 +141,45 @@ proc probeAndEvictPeers*(
...
+      try:
+        discard await one(inFlight)
+      except ValueError:
+        # All futures already finished between the snapshot and the wait.
+        discard
+      except CancelledError as exc:
+        await noCancel inFlight.cancelAndWait()
+        raise exc
+    else:
+      await sleepAsync(kad.config.livenessIdleInterval)
```
vladopajic: just one away right below `while` should be enough ?

vladopajic:
@copilot+gpt-5.6-terra please review pr and in addition to your standard review:
- review how locking works; provide suggestion that will be future proof and still elegant (we want to avoid any locking issues)
- what will be new eviction, from this pr, impact compared to previews code

libp2p/protocols/kademlia.nim
```diff
@@ -93,7 +123,16 @@ proc probeAndEvictPeers*(
...
+    let probeKey: ProbeKey = (rtable.selfId, peerId)
+    kad.livenessProbes.withValue(probeKey, existing):
+      futs.add(existing[])
+      continue
+    if not kad.probeSem.tryAcquire():
+      kad_routing_table_liveness_probes.inc(labelValues = ["skipped"])
+      continue
+    let fut = kad.checkAndEvictPeer(rtable, peerId)
+    kad.trackLivenessProbe(probeKey, fut)
+    futs.add(fut)
```
vladopajic: seems like same code from `tryLaunchLivenessProbe`

vladopajic:
@SionoiS ai has pushed a commit instead of adding review :/

libp2p/protocols/kademlia.nim
vladopajic: now with ai change it doesn't seems that much reusable

libp2p/protocols/kademlia.nim
vladopajic: seems like ai has resolved this in [d555167](https://github.com/vacp2p/nim-libp2p/pull/2907/commits/d555167859d618e6bcb0f1ab982be1cc79a78945)

vladopajic (approved):
lgtm; but please review ai's change; smaller nitpicks left

libp2p/protocols/kademlia.nim
jm-clius: Nitpick, but I think code reads easier if the function name is more descriptive, e.g. `peersInGracePeriod()`

libp2p/protocols/kademlia.nim
```diff
@@ -79,11 +96,23 @@ proc checkAndEvictPeer(
...
+  kad.trackLivenessProbe(probeKey, kad.checkAndEvictPeer(rtable, peerId))
+
 proc probeAndEvictPeers*(
     kad: KadDHT, rtable: RoutingTable
 ) {.async: (raises: [CancelledError]).} =
-  ## Probes routing-table peers that have been quiet longer than the liveness
-  ## grace period and removes those that fail to answer a FIND_NODE.
+  ## One-shot batch: probes routing-table peers past the liveness grace period
+  ## and removes those that fail to answer a FIND_NODE. Used by tests and any
+  ## explicit maintenance call; production traffic uses ``maintainLiveness``.
```
jm-clius, line 145: Do we really need a function only exposed for tests? Or do we foresee explicit maintenance calls in some modules?

libp2p/protocols/kademlia.nim
```diff
@@ -79,11 +96,23 @@ proc checkAndEvictPeer(
...
+proc launchLivenessProbe(
+    kad: KadDHT, rtable: RoutingTable, peerId: PeerId
+) {.raises: [].} =
+  ## Starts a liveness probe unless it is already in flight.
+  let probeKey: ProbeKey = (rtable.selfId, peerId)
+  if kad.livenessProbes.hasKey(probeKey):
+    return
+  if kad.stopping:
+    return
+  kad.trackLivenessProbe(probeKey, kad.checkAndEvictPeer(rtable, peerId))
```
jm-clius: Mmm. Liveness is a property of the peer and is unrelated to routing table. In fact, I think here we'll spin up several liveness checks for the same peer across many service routing tables (clogging the semaphore). I think here the key should be just `peerId` and the function `checkPeerLiveness()`. The former deduplicates liveness checks by peer and the latter simply returns true or false to indicate if a peer should be removed. If yes, we perform `removePeer()` across all maintainable tables (even if the peer is not in that table, it should be a quick return).

libp2p/protocols/kademlia.nim
```diff
@@ -33,10 +33,32 @@ proc livenessCandidates(
...
+        kad.livenessProbes.del(probeKey)
+  )
+
+proc checkAndEvictPeer(
+    kad: KadDHT, rtable: RoutingTable, peerId: PeerId
+) {.async: (raises: [CancelledError]).} =
+  ## Returns immediately when all liveness probe slots are occupied.
+  if kad.stopping:
+    return
   if not kad.probeSem.tryAcquire():
```
jm-clius: IIUC this is the same semaphore that we use to limit lookup checks for admission to the routing table? I'm not convinced we should share a semaphore here. When we have a lot of stale peers is (i.e. multiple liveness checks, which will happen often), it's exactly the time where we don't want to drop new peers because we cannot perform the necessary lookup checks?

libp2p/protocols/kademlia.nim
```diff
@@ -102,13 +137,40 @@ proc probeAndEvictPeers*(
...
+  await sleepAsync(kad.config.livenessIdleInterval)
+
+  while not kad.stopping:
+    let grace = kad.config.livenessGracePeriod
+
+    for rtable in kad.maintainableTables():
+      if kad.stopping:
+        return
+      for peerId in rtable.livenessCandidates(grace):
+        kad.launchLivenessProbe(rtable, peerId)
```
jm-clius: I wonder if running this every ~30 seconds, bottle-necked by a semaphore and actual dials, would cause some issues? In either case, I think we need quite a bit of debug logging throughout so we can trace exactly how this behaves in practice.

jm-clius (changes requested):
Thanks!
Two blocking issues below:
- this is quite a significant change in how routing tables behave. we probably need some debugging logging throughout (and errors and warnings where appropriate)
- I think liveness should be checked (and deduplicated) per peer, regardless of routing table. If a peer is stale anywhere it is stale everywhere and can be removed from all maintainable tables.

SionoiS:
> * I think liveness should be checked (and deduplicated) per peer, regardless of routing table. If a peer is stale anywhere it is stale everywhere and can be removed from all maintainable tables.

I've wanted to refactor how tables work for some time. The time might be now.

jm-clius:
>I've wanted to refactor how tables work for some time. The time might be now.

Could be, but note that the suggestion I made can already be done with the existing routing tables (they already check if a peer is present before removing it, and even provide a `contains(nodeId)` check - it's not necessary to index liveness checks by routing table, in other words). Refactoring the tables may be necessary, but I'm trying to avoid piling refactoring upon refactoring :). Would suggest fixing this PR with the simpler approach that's already available and propose table refactoring in a follow-up.

libp2p/protocols/kademlia.nim
SionoiS: Yes we do need to test it.

libp2p/protocols/kademlia.nim
```diff
@@ -33,82 +33,182 @@ proc livenessCandidates(
...
+      if peers.len > 0:
+        debug "Liveness scan found replaceable peers", peers = peers.len
+      for peerId in peers:
+        kad.launchLivenessProbe(peerId)
+
+    if kad.livenessProbes.len > 0:
+      debug "Waiting for in-flight liveness probes", inFlight = kad.livenessProbes.len
+      let inFlight = kad.livenessProbes.values.toSeq()
+      try:
+        discard await one(inFlight)
```
richard-ramos: I think this might not be ideal for when you have a lot of peers
```nim
while not kad.stopping:
  await sleepAsync(livenessIdleInterval) 

  # LOGIC HERE

  await one(inFlightProbes)
```

'cause imagine you have 50 peers.  The first iteration will take peers 1-20 (and 21 to 50 are skipped due to no more slots in liveness semaphoere). Then, once one of the probes finished (in line 199), the iteration ends, and then you wait for 30s before launching more probes.


Shouldnt it instead keep probing as long as there is 'work' to do?, i.e. in pseudocode:
```nim
var idle = true
while not kad.stopping:
  if idle:
    await sleepAsync(livenessIdleInterval) 

  # LOGIC HERE

  if probes.len > 0:
      discard await one(kad.livenessProbes.values.toSeq())
      # semaphore released a slot, iterate again immedialty
      idle = false
  else:
      idle = true
```

SionoiS: Ah yes! I see your point. It make sense.

gmelodie (approved):

tinniaru3005 (approved):
LGTM

Issue #146, opened 2020-04-21
Research timeouts for different parts of libp2p


dryajov:
Some part of libp2p, noise for example, might require timeouts to prevent timing attacks, however chronos timers currently have a scaling issue that requires them being used sparingly.

PR #142, opened 2020-04-17
Fixed typo in link resulting in 404

NatalieChin80:
It's a minor issue - but just wanted to bring up that the following links in the root README lead to a 404. This was because the README was pointing to a `doc/` folder instead of `docs/`. The respective directory changes were made for the following links: 
- API.md
- GETTING_STARTED.md
- GO_DAEMON.md

PR #2909, opened 2026-07-31
chore(cbind): use nim-ffi 0.3.0

gmelodie:
## Summary

Repin cbind's `nim-ffi` dependency from the unmerged `fix/cbor-non-canonical` branch head (`b95e2b04`) to nim-ffi `v0.3.0-rc.2` (commit `53515de`).

The old pin was a workaround: the CBOR seq fix that cbind's bare `seq[byte]` returns rely on lived only on an unmerged branch, and `cbind.nimble` carried a "repin to a tag on merge" note. `v0.3.0-rc.2` now contains that fix (nim-ffi #141) plus the additive 0.3.0 feature integration (nim-ffi #142), so the workaround is no longer needed.

Upgrading also brings, for free:
- **Foreign-thread GC registration on method entry** — nim-ffi #146, the single commit between `v0.3.0-rc.1` and `v0.3.0-rc.2`. An `abi = c` method entry point now registers its calling thread with the GC, which only static entry points did before. cbind builds with `--mm:refc`, and the wrapper unpacks the request into GC memory on the calling thread, so a host that calls a `LibP2P` method from a thread-pool worker could die in the GC on that thread's first call.
- **Context recycling** — pooled slots reuse their worker thread, event thread and signal fds instead of churning them, which bounds `ThreadSignalPtr` fds so `select()` in `waitSync` stops failing with `EINVAL` under create/destroy load. This matters for cbind's per-context lifecycle.
- The merged canonical-seq CBOR generation fix that `libp2pPublicKey`, `libp2pCreateXpr` and `libp2pNewPrivateKey` (bare `seq[byte]` returns) depend on.
- The nil-library ctor guard and other #142 hardening.

The 0.3.0 feature PR is purely additive over master, so no cbind source changes are required. Every pragma cbind uses (`ffi`, `ffiCtor`, `ffiStatic`, `ffiDtor`, `ffiEvent`, `ffiConst`, `ffiHandle`, `declareLibrary`, `genBindings`) still exists at the tagged commit.

cbind keeps its explicit pragmas and its two single-field event payload objects, so it adopts none of the new spellings 0.3.0 offers. The `{.ffi.}` shape router only re-derives the pragma that the code already names, and a multi-parameter `{.ffiEvent.}` would replace `IncomingStreamEvent` and `PubsubMessageEvent` with generated envelope structs under new names, which breaks the C API that `examples/echo.c`, `examples/relay.c` and `examples/gossipsub.c` and every downstream consumer bind to.

## Affected Areas

- [x] Build / Tooling
  - `cbind/cbind.nimble`: `ffi` dep pinned by commit `53515de` (the `v0.3.0-rc.2` tag), with a comment naming the tag. Pins the commit rather than the movable tag name, matching the neighboring `nim-cbor-serialization` pin and the lock/nix entries.
  - `cbind/nimble.lock`: `ffi` entry updated to the tag's `vcsRevision` and nimble checksum.
  - `nix/cbind-deps.nix`: `ffi` `fetchgit` `rev` and `sha256` updated to match.

All three files pin the same commit `53515de`. The lock checksum and nix hash were computed with the exact algorithms nimble and `fetchgit` use, then cross-checked by regenerating the previous commit's values and confirming they match the entries they replace. `nimble -l setup` resolves the lock and validates the ffi checksum with no error.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

cbind is a self-contained C-bindings package; this dependency bump does not change nim-libp2p core behavior.

## Impact on Library Users

No impact on the nim-libp2p API. cbind's generated C/CDDL bindings are unchanged: no `{.ffi.}` type or proc signature changed, and no generated bindings are checked in, so nothing drifts.

## Risk Assessment

- Low. It is a dependency version bump with no source changes.
- `v0.3.0-rc.2` is a release candidate, not a final `v0.3.0` tag (no final tag exists yet). nim-ffi's `ffi.nimble` still declares version `0.3.0` at this commit, so `nimble.lock` records `"version": "0.3.0"`.
- The three dependency files (nimble requires, `nimble.lock`, nix deps) all pin the same commit, so `nimble setup` and the Nix build resolve the same revision.

## References

- nim-ffi tag `v0.3.0-rc.2`
- nim-ffi #141 (fix: non-canonical seq CBOR generation)
- nim-ffi #142 (integrate the 0.2 line: shape router, `{.ffiExport.}`, context recycling)
- nim-ffi #146 (fix: register the calling thread with the GC on `abi = c` method entry)

### DIFF
--- a/cbind/cbind.nimble
+++ b/cbind/cbind.nimble
@@ -8,9 +8,7 @@ license = "MIT"
 
 import os, strutils
 
-requires "taskpools >= 0.1.0",
-  "ffi >= 0.3.0",
-  "https://github.com/vacp2p/nim-cbor-serialization#1664160e04d153573373afddc552b9cbf6fbe4dc"
+requires "taskpools >= 0.1.0", "ffi >= 0.3.0", "cbor_serialization == 0.3.0"
 
 proc findInstalledPkgDir(prefix: string): string =
   ## Path of an installed dep dir matching `prefix` (e.g. "ffi-"). Lockfile
@@ -35,8 +33,8 @@ proc findInstalledPkgDir(prefix: string): string =
   )
 
 proc ffiDepPaths(): string =
-  # `setup` does not put direct Git URL deps on nimble.paths; point the compiler
-  # at the installed copies.
+  # A global install writes no nimble.paths; point the compiler at the installed
+  # copies.
   " --path:" & findInstalledPkgDir("ffi-") & " --path:" &
     findInstalledPkgDir("cbor_serialization-")
 
--- a/cbind/cbind.nimble
+++ b/cbind/cbind.nimble
@@ -8,10 +8,8 @@ license = "MIT"
 
 import os, strutils
 
-requires "taskpools >= 0.1.0",
-  "https://github.com/vacp2p/nim-cbor-serialization#1664160e04d153573373afddc552b9cbf6fbe4dc",
-  # nim-ffi v0.3.0-rc.2
-  "https://github.com/logos-messaging/nim-ffi#53515de17af0ef3e88b2aec9675b8163dddc14ae"
+requires "taskpools >= 0.1.0", "ffi >= 0.3.0",
+  "https://github.com/vacp2p/nim-cbor-serialization#1664160e04d153573373afddc552b9cbf6fbe4dc"
 
 proc findInstalledPkgDir(prefix: string): string =
   ## Path of an installed dep dir matching `prefix` (e.g. "ffi-"). Lockfile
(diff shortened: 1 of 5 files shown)


cbind/cbind.nimble
```diff
@@ -9,9 +9,8 @@ license = "MIT"
 import os, strutils
 
 requires "taskpools >= 0.1.0",
-  "https://github.com/vacp2p/nim-cbor-serialization#1664160e04d153573373afddc552b9cbf6fbe4dc",
-  # nim-ffi head of the unmerged `fix/cbor-non-canonical`; repin to a tag on merge
-  "https://github.com/logos-messaging/nim-ffi#b95e2b04a63fbd417938bf3ec0ac14be7935e21b"
+  "ffi >= 0.3.0",
+  "https://github.com/vacp2p/nim-cbor-serialization#1664160e04d153573373afddc552b9cbf6fbe4dc"
```
richard-ramos: Is it possible to use a version for cbor-serialization too?

richard-ramos (approved):

vladopajic (approved):

PR #2910, opened 2026-07-31
feat(kad): dynamic client/server mode via reachability

### LINKED ISSUE #2858: kad: dynamic client/server mode via reachability (autonat) instead of a static flag
Mode is fixed at construction: `kademlia.nim:62 new(..., client: bool)`. An undialable NAT'd node that starts in server mode keeps serving queries and polluting other nodes' routing tables; a node that later becomes reachable can't start serving.

go-libp2p's `ModeAuto` subscribes to autonat/reachability events and dynamically calls `moveToServerMode` / `moveToClientMode` (registering/deregistering stream handlers and resetting inbound server streams on downgrade).

Proposed work:
- Add a `ModeAuto` option that subscribes to reachability change events.
- On reachable → register the DHT stream handler (server); on unreachable → deregister and reset inbound streams (client).

Impact: network health; keeps undialable nodes out of others' tables. Medium.

gmelodie:
## Summary

Kad-DHT mode was fixed at construction (`client: bool`). An undialable NAT'd node started in server mode keeps answering queries and pollutes other nodes' routing tables; a node that later becomes reachable cannot start serving.

This PR makes the mode dynamic and drives it from autonat reachability:

- `KadDHT` keeps one `isServer` flag. The DHT stream handler is always mounted and gated on that flag. When the node does not serve, it resets inbound streams instead of answering.
- `changeMode(isServer)` flips the flag and reports whether the mode changed. The downgrade resets in-flight inbound server streams, so peers stop treating the node as a server.
- `KadMode` (`Client`, `Server`, `Auto`) is a configuration type for the switch builder only. `withKademlia(mode = ...)` selects the initial flag; `Auto` starts as a client.
- The builder owns the wiring between reachability and Kad-DHT: with `Auto` it subscribes `kadReachabilityHandler(kad)` to the active reachability service (AutoNAT v1/v2 or hole-punching). Kad-DHT itself knows nothing about autonat.
- `AutonatService` / `AutonatV2Service` gained an additive `addStatusAndConfidenceHandler`, so Kad-DHT can subscribe without displacing HPService's handler.

Closes #2858.

## Affected Areas

- [x] Protocol Logic
  

No downstream integration yet. The AutoNAT changes are additive (a second subscriber slot); existing single-handler behavior is untouched. The Kad-DHT constructor replaces `client: bool = false` with `isServer: bool = true`; no in-tree caller passed `client`, so the default preserves prior behavior.

- **Nimbus:** n/a
- **Waku:** n/a
- **Codex:** n/a

## Impact on Library Users

- API change: `KadDHT.new(..., client: bool)` becomes `KadDHT.new(..., isServer: bool = true)`. `withKademlia` gains a `mode: KadMode` parameter. The default keeps server behavior.
- New API: `KadDHT.changeMode`, `kadReachabilityHandler`, `AutonatService.addStatusAndConfidenceHandler`, `AutonatV2Service.addStatusAndConfidenceHandler`, `HPService.addStatusAndConfidenceHandler`, `NATService.addReachabilityHandler`.
- Behavior change: a client node (or an `Auto` node before it is proven reachable) resets inbound DHT streams instead of serving them.

## Risk Assessment

- Backward compatibility: AutoNAT changes are purely additive; the Kad-DHT constructor default still serves.
- Network behavior: `Auto` starts as a client until autonat proves the node reachable, which matches go-libp2p `ModeAuto`.
- `Auto` with no reachability service configured stays a client and logs a warning at build time.

## References

- Issue: #2858
- Parent: #1756
- go-libp2p `ModeAuto` (`moveToServerMode` / `moveToClientMode`)

## Additional Notes

There is no multistream unmount, so a client node keeps the codec mounted and advertised. The refusal to serve (and the stream reset) is what removes the node from peers' routing tables over time.

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -41,7 +41,8 @@ export
   natservice.NATService, natservice.natConfig, natservice.upnpConfig,
   natservice.natPmpConfig, natservice.explicitIpConfig, natservice.autonatConfig,
   natservice.holePunchingConfig, natservice.AutonatV2ServiceConfig,
-  natservice.AutonatV2Service, natservice.natService
+  natservice.AutonatV2Service, natservice.natService, natservice.ReachabilityHandler,
+  natservice.addReachabilityHandler
 
 const MemoryAutoAddress* = memorytransport.MemoryAutoAddress
 
@@ -534,6 +535,23 @@ proc setupServices(b: SwitchBuilder, switch: Switch) {.raises: [LPError].} =
   for service in switch.services:
     service.setup(switch)
 
+proc kadReachabilityHandler*(kad: KadDHT): ReachabilityHandler =
+  ## Handler for ``KadMode.Auto``: the node serves queries while it is
+  ## reachable, and stops serving when it is not. ``Unknown`` keeps the current
+  ## mode, since it means autonat has no verdict yet.
+  proc(
+      reachability: NetworkReachability,
+      confidence: Opt[float],
+      dialBackAddr: Opt[MultiAddress],
+  ) {.async: (raises: [CancelledError]).} =
+    case reachability
+    of NetworkReachability.Reachable:
+      discard await kad.changeMode(isServer = true)
+    of NetworkReachability.NotReachable:
+      discard await kad.changeMode(isServer = false)
+    of NetworkReachability.Unknown:
+      discard
+
 proc mountProtocols(b: SwitchBuilder, switch: Switch) {.raises: [LPError].} =
   if not switch.peerStore.identify.isNil:
     switch.mount(switch.peerStore.identify)
@@ -563,21 +581,15 @@ proc mountProtocols(b: SwitchBuilder, switch: Switch) {.raises: [LPError].} =
       bootstrapNodes = kadInfo.bootstrapNodes,
       config = config,
       rng = b.rng,
-      mode = kadInfo.mode,
+      # Auto starts as a client until autonat proves the node reachable.
+      isServer = kadInfo.mode == KadMode.Server,
     )
     switch.mount(kad)
 
     if kadInfo.mode == KadMode.Auto:
       var wired = false
       switch.natService().withValue(nat):
-        wired = nat.addReachabilityHandler(
-          proc(
-              reachability: NetworkReachability,
-              confidence: Opt[float],
-              dialBackAddr: Opt[MultiAddress],
-          ) {.async: (raises: [CancelledError]).} =
-            await kad.onReachabilityChanged(reachability)
-        )
+        wired = nat.addReachabilityHandler(kadReachabilityHandler(kad))
       if not wired:
         warn "Kad-DHT auto mode has no reachability service; it stays a client",
           hint = "configure withNAT reachability or hole-punching"
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -41,8 +41,7 @@ export
   natservice.NATService, natservice.natConfig, natservice.upnpConfig,
   natservice.natPmpConfig, natservice.explicitIpConfig, natservice.autonatConfig,
   natservice.holePunchingConfig, natservice.AutonatV2ServiceConfig,
-  natservice.AutonatV2Service, natservice.natService, natservice.ReachabilityHandler,
-  natservice.addReachabilityHandler
+  natservice.AutonatV2Service, natservice.natService
 
 const MemoryAutoAddress* = memorytransport.MemoryAutoAddress
 
@@ -60,6 +59,11 @@ type
 
   SecureProtocol* {.pure.} = enum
     Noise
+
+  KadMode* {.pure.} = enum
+    Client ## never serves queries; only issues them
+    Server ## always serves queries
+    Auto ## serves queries while autonat reports the node reachable
 
   KadInfo = object
     config*: KadDHTConfig
@@ -535,7 +539,7 @@ proc setupServices(b: SwitchBuilder, switch: Switch) {.raises: [LPError].} =
   for service in switch.services:
     service.setup(switch)
 
-proc kadReachabilityHandler*(kad: KadDHT): ReachabilityHandler =
+proc makeKadReachabilityHandler(kad: KadDHT): ReachabilityHandler =
   ## Handler for ``KadMode.Auto``: the node serves queries while it is
   ## reachable, and stops serving when it is not. ``Unknown`` keeps the current
   ## mode, since it means autonat has no verdict yet.
@@ -589,7 +593,7 @@ proc mountProtocols(b: SwitchBuilder, switch: Switch) {.raises: [LPError].} =
     if kadInfo.mode == KadMode.Auto:
       var wired = false
       switch.natService().withValue(nat):
-        wired = nat.addReachabilityHandler(kadReachabilityHandler(kad))
+        wired = nat.addReachabilityHandler(makeKadReachabilityHandler(kad))
       if not wired:
         warn "Kad-DHT auto mode has no reachability service; it stays a client",
           hint = "configure withNAT reachability or hole-punching"
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -63,6 +63,7 @@ type
   KadInfo = object
     config*: KadDHTConfig
     bootstrapNodes*: seq[(PeerId, seq[MultiAddress])]
+    mode*: KadMode
 
   SwitchBuilder* = ref object
     privKey: Opt[PrivateKey]
@@ -395,8 +396,9 @@ proc withKademlia*(
     b: SwitchBuilder,
     bootstrapNodes: seq[(PeerId, seq[MultiAddress])] = @[],
     config: KadDHTConfig = KadDHTConfig.new(),
+    mode: KadMode = KadMode.Server,
 ): SwitchBuilder =
-  b.kad = Opt.some(KadInfo(config: config, bootstrapNodes: bootstrapNodes))
+  b.kad = Opt.some(KadInfo(config: config, bootstrapNodes: bootstrapNodes, mode: mode))
   b
 
 proc withIdentifyPusher*(b: SwitchBuilder, enabled: bool = true): SwitchBuilder =
@@ -557,9 +559,28 @@ proc mountProtocols(b: SwitchBuilder, switch: Switch) {.raises: [LPError].} =
     var config = kadInfo.config
     config.addressPolicy = b.addressPolicy
     let kad = KadDHT.new(
-      switch, bootstrapNodes = kadInfo.bootstrapNodes, config = config, rng = b.rng
+      switch,
+      bootstrapNodes = kadInfo.bootstrapNodes,
+      config = config,
+      rng = b.rng,
+      mode = kadInfo.mode,
     )
     switch.mount(kad)
+
+    if kadInfo.mode == KadMode.Auto:
+      var wired = false
+      switch.natService().withValue(nat):
+        wired = nat.addReachabilityHandler(
+          proc(
+              reachability: NetworkReachability,
+              confidence: Opt[float],
+              dialBackAddr: Opt[MultiAddress],
+          ) {.async: (raises: [CancelledError]).} =
+            await kad.onReachabilityChanged(reachability)
+        )
+      if not wired:
+        warn "Kad-DHT auto mode has no reachability service; it stays a client",
+          hint = "configure withNAT reachability or hole-punching"
 
 proc build*(b: SwitchBuilder): Switch {.raises: [LPError].} =
   var switch = b.buildSwitch()
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -6,14 +6,13 @@ import chronos, chronicles, results
 import ../utils/[heartbeat, future]
 import ../[peerid, switch, multihash]
 import ./protocol
-import ./connectivity/autonat/types as autonat_types
 import
   ./kademlia/[routing_table, protobuf, types, find, get, put, keyspace, provider, ping]
 import ./kademlia/[kademlia_metrics, netsize]
 
 export
   chronicles, routing_table, protobuf, types, find, get, put, keyspace, provider, ping,
-  kademlia_metrics, netsize, NetworkReachability
+  kademlia_metrics, netsize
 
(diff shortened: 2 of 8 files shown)


libp2p/protocols/kademlia.nim
```diff
@@ -144,6 +153,46 @@ proc new*(
...
+  ## Stop serving inbound queries and reset any in-flight server streams. The
+  ## codec stays mounted (no unmount exists), so peers drop us once we stop serving.
+  if not kad.serving:
+    return
+  kad.serving = false
+
+  let streams = kad.serverStreams.toSeq()
+  kad.serverStreams.clear()
+  await noCancel allFutures(streams.mapIt(it.reset()))
+  debug "Kad DHT moved to client mode", resetStreams = streams.len
```
vladopajic:
```
proc changeMode*(kad: KadDHT, newMode: KadMode): bool
```

result can indicate if mode was changed

libp2p/protocols/kademlia.nim
```diff
@@ -144,6 +153,46 @@ proc new*(
...
+  ## Whether the node currently answers inbound queries.
+  kad.serving
+
+proc moveToServerMode*(kad: KadDHT) =
+  ## Start answering inbound queries. The handler is always mounted, so this
+  ## only flips the serving flag.
+  if kad.serving:
+    return
+  kad.serving = true
+  debug "Kad DHT moved to server mode"
```
vladopajic: `moveToServerMode`

will move start serving but mode will still remain `Client` if it was prior `mode`?

libp2p/protocols/kademlia.nim
```diff
@@ -144,6 +153,46 @@ proc new*(
...
+
+proc moveToClientMode*(kad: KadDHT) {.async: (raises: []).} =
+  ## Stop serving inbound queries and reset any in-flight server streams. The
+  ## codec stays mounted (no unmount exists), so peers drop us once we stop serving.
+  if not kad.serving:
+    return
+  kad.serving = false
+
+  let streams = kad.serverStreams.toSeq()
```
vladopajic: `HashSet` does not have `mapIt` ? :(

libp2p/protocols/kademlia.nim
```diff
@@ -144,6 +153,46 @@ proc new*(
 
   return kad
 
+func isServer*(kad: KadDHT): bool =
```
vladopajic: `isServer` is not intuitive name for it's function.

intuitively it means `kad.mode == Server`

vladopajic (commented):
`isServer`, `serving`, `mode == Server` is not ideal, could all reconciles to just case when `mode == Server` or something more intuitive

tests/libp2p/protocols/test_autonat_service.nim
```diff
@@ -129,6 +129,53 @@ suite "Autonat Service":
...
+    let switch2 = createSwitch()
+    let switch3 = createSwitch()
+    let switch4 = createSwitch()
+
+    let firstAwaiter = newFuture[void]()
+    let secondAwaiter = newFuture[void]()
+
+    proc isReachable(reachability: NetworkReachability, confidence: Opt[float]): bool =
+      reachability == NetworkReachability.Reachable and confidence.isSome() and
+        confidence.get() >= 0.3
```
vladopajic: as general tiding of this test file, his can be share utility. or `0.3` can be some const - with some note that `0.3` is empirically chosen const for tests - so that ai can know if/when it writes code.

libp2p/builders.nim
```diff
@@ -395,8 +396,9 @@ proc withKademlia*(
     b: SwitchBuilder,
     bootstrapNodes: seq[(PeerId, seq[MultiAddress])] = @[],
     config: KadDHTConfig = KadDHTConfig.new(),
+    mode: KadMode = KadMode.Server,
 ): SwitchBuilder =
-  b.kad = Opt.some(KadInfo(config: config, bootstrapNodes: bootstrapNodes))
+  b.kad = Opt.some(KadInfo(config: config, bootstrapNodes: bootstrapNodes, mode: mode))
```
vladopajic, line 406:
```suggestion
  b.kad = Opt.some(KadInfo(config: config, bootstrapNodes: bootstrapNodes, mode))
```

unnecessary naming

tests/libp2p/kademlia/utils.nim
```diff
@@ -89,9 +89,10 @@ proc testKadConfig*(
 proc setupKad*(
     config: KadDHTConfig = testKadConfig(),
     bootstrapNodes: seq[(PeerId, seq[MultiAddress])] = @[],
+    mode: KadMode = KadMode.Server,
 ): KadDHT =
   let switch = makeStandardSwitch(TcpAutoAddress)
-  let kad = KadDHT.new(switch, bootstrapNodes, config, rng = rng())
+  let kad = KadDHT.new(switch, bootstrapNodes, config, rng = rng(), mode = mode)
```
vladopajic:
```suggestion
  let kad = KadDHT.new(switch, bootstrapNodes, config, rng(), mode)
```

unnecessary naming

tests/libp2p/kademlia/test_mode.nim
```diff
@@ -0,0 +1,140 @@
...
+    # The downgrade must reset the parked stream: the querier's next read ends in
+    # EOF, not a timeout (a timeout would mean the reset never propagated).
+    var gotEof = false
+    try:
+      discard await stream.readLp(MaxMsgSize).wait(3.seconds)
+    except AsyncTimeoutError:
+      gotEof = false
+    except CatchableError:
+      gotEof = true
+    check gotEof
```
vladopajic:
```nim
expect SomeEOFError:
   discard await stream.readLp(MaxMsgSize).wait(3.seconds)
```

also catching `CatchableError` could leak anything else there

tests/libp2p/kademlia/test_mode.nim
```diff
@@ -0,0 +1,140 @@
...
+    waitFor auto.onReachabilityChanged(NetworkReachability.Unknown)
+    check auto.mode == KadMode.Client # Unknown leaves the current mode untouched
+
+    # Pinned modes ignore reachability changes.
+    let server = setupKad(mode = KadMode.Server)
+    waitFor server.onReachabilityChanged(NetworkReachability.NotReachable)
+    check server.mode == KadMode.Server
+
+    let client = setupKad(mode = KadMode.Client)
+    waitFor client.onReachabilityChanged(NetworkReachability.Reachable)
```
vladopajic: do not use `waitFor` use `asyncTest` instead.

`waitFor` is causing errors with latest deps https://github.com/vacp2p/nim-libp2p/pull/2867

libp2p/protocols/kademlia.nim
```diff
@@ -233,6 +242,43 @@ proc new*(
...
+  if not kad.autoMode:
+    return
+
+  case reachability
+  of NetworkReachability.Reachable:
+    discard await kad.changeMode(KadMode.Server)
+  of NetworkReachability.NotReachable:
+    discard await kad.changeMode(KadMode.Client)
+  of NetworkReachability.Unknown:
+    discard
```
vladopajic: `onReachabilityChanged` should not really be part of `Kad`'s interface.

- `Kad` has `changeMode`. all that `Kad` cares is to switch it's mode. it can be `autonat` it can be some other way manually triggered (from `logos/libp2p-module-ui`), but only way for `Kad` to change mode is via `changeMode`.

- `Kad` should not have dependency to autonat service. see import, nor it should know about `NetworkReachability`. the place were `Kad` and `NetworkReachability` meet should be thing that owns it both swtich / swithc_builder.

libp2p/protocols/kademlia.nim
```diff
@@ -233,6 +242,43 @@ proc new*(
...
+  kad.serverStreams.clear()
+  await noCancel allFutures(streams.mapIt(it.reset()))
+  debug "Reset inbound Kad DHT streams", streams = streams.len
+
+proc changeMode*(kad: KadDHT, newMode: KadMode): Future[bool] {.async: (raises: []).} =
+  ## Move the node to ``newMode`` and report whether the mode changed. ``Auto``
+  ## is a configuration value only, so it changes nothing here. Moving to
+  ## ``Client`` resets the in-flight inbound streams; the codec stays mounted
+  ## (no unmount exists), so peers drop us once the handler refuses to serve.
+  if newMode == KadMode.Auto or newMode == kad.mode:
```
vladopajic: since `newMode == KadMode.Auto` is invalid input it can be `doAssert`.
but that is if we do not want auto mode. 

but is it really necessary to prevent it here? it can have the same behavior as in `new`:
- change to client
- set autoMode: true

libp2p/protocols/kademlia.nim
vladopajic: note the implication:
- `kad.autoMode` will be unnecessary as property of `Kad`

libp2p/protocols/kademlia/types.nim
```diff
@@ -529,6 +529,13 @@ type ProbeKey* = tuple[tableId: Key, peerId: PeerId]
 
 type AdmitHook* = proc(peerId: PeerId) {.gcsafe, raises: [].}
 
+type KadMode* {.pure.} = enum
+  Client ## does not serve queries; only issues them
+  Server ## serves queries
+  Auto
+    ## configuration only: serves queries while the node is reachable (driven by
+    ## autonat). ``KadDHT.mode`` never holds this value.
```
vladopajic: `KadMode` now serves 2 purposes:
- as a user configuration 
- as a mode for `Kad`

but same type in both cases is not ideal, as configuration option `Auto` should never really be passed to constructor. there is `if/else` in constructor to fix this issues, but it would be better if there is no such `if/else` - because why have a dedicated type if you don't have type safety?

imho: 
- `KadMode` should be only type for configurations (for switch builder)
- `Kad` should keep boolean `isServer/isClient` flag - which is exclusive and a type safe way for choosing one mode or another. it can be `KadMode2` enum  (with only client/server values), but i dont see this as necessary nor better then bool.

libp2p/protocols/kademlia.nim
vladopajic: see bellow, anything related to auto should not really be part of kad

libp2p/protocols/kademlia/types.nim
```diff
@@ -529,6 +529,12 @@ type ProbeKey* = tuple[tableId: Key, peerId: PeerId]
 
 type AdmitHook* = proc(peerId: PeerId) {.gcsafe, raises: [].}
 
+type KadMode* {.pure.} = enum
```
vladopajic: `KadMode` should be part of `builders.nim`. it's not used anywhere else.

libp2p/builders.nim
```diff
@@ -532,6 +535,19 @@ proc setupServices(b: SwitchBuilder, switch: Switch) {.raises: [LPError].} =
   for service in switch.services:
     service.setup(switch)
 
+proc kadReachabilityHandler*(kad: KadDHT): ReachabilityHandler =
```
vladopajic: does not need to be exported, it is internal proc?

libp2p/builders.nim
```diff
@@ -41,7 +41,8 @@ export
   natservice.NATService, natservice.natConfig, natservice.upnpConfig,
   natservice.natPmpConfig, natservice.explicitIpConfig, natservice.autonatConfig,
   natservice.holePunchingConfig, natservice.AutonatV2ServiceConfig,
-  natservice.AutonatV2Service, natservice.natService
+  natservice.AutonatV2Service, natservice.natService, natservice.ReachabilityHandler,
+  natservice.addReachabilityHandler
```
vladopajic: why is `addReachabilityHandler` exported? also `natservice.ReachabilityHandler`

libp2p/builders.nim
```diff
@@ -557,10 +573,23 @@ proc mountProtocols(b: SwitchBuilder, switch: Switch) {.raises: [LPError].} =
...
+      rng = b.rng,
+      # Auto starts as a client until autonat proves the node reachable.
+      isServer = kadInfo.mode == KadMode.Server,
     )
     switch.mount(kad)
 
+    if kadInfo.mode == KadMode.Auto:
+      var wired = false
+      switch.natService().withValue(nat):
+        wired = nat.addReachabilityHandler(kadReachabilityHandler(kad))
```
vladopajic: `kadReachabilityHandler` -> `makeKadReachabilityHandler` or `makeReachabilityHandler`
- it is maker, it creates new handler each time

vladopajic (approved):
lgtm; with comments

libp2p/builders.nim
gmelodie: `KadInfo` is a constructor, so nim requires the named params

tests/libp2p/kademlia/test_builder.nim
```diff
@@ -9,10 +9,63 @@ import ../../../libp2p/protocols/kademlia
...
 import ./utils.nim
 
+proc kadBuilder(): SwitchBuilder =
+  SwitchBuilder
+    .new()
+    .withRng(rng())
+    .withAddresses(@[MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet()])
+    .withTcpTransport()
+    .withMplex()
+    .withNoise()
```
vladopajic: this seems copy of `makeStandardSwitchBuilder`

vladopajic: as general note `SwitchBuilder.new() ...` in unit tests is always a boilerplate wich we try to avoid with `makeStandardSwitchBuilder` or `makeStandardSwitch`

libp2p/builders.nim
```diff
@@ -557,10 +581,23 @@ proc mountProtocols(b: SwitchBuilder, switch: Switch) {.raises: [LPError].} =
...
     let kad = KadDHT.new(
-      switch, bootstrapNodes = kadInfo.bootstrapNodes, config = config, rng = b.rng
+      switch,
+      bootstrapNodes = kadInfo.bootstrapNodes,
+      config = config,
+      rng = b.rng,
+      # Auto starts as a client until autonat proves the node reachable.
+      isServer = kadInfo.mode == KadMode.Server,
     )
     switch.mount(kad)
```
richard-ramos, line 591:
```suggestion
    if kadInfo.mode != KadMode.Client:
      switch.mount(kad)
```

PR #2911, opened 2026-07-31
feat(kad): send optimistic ADD_PROVIDER during lookup

### LINKED ISSUE #2860: kad: optimistic provide — send ADD_PROVIDER during the lookup
`provider.nim` (`addProvider` / `addProviderSpillover`) sends ADD_PROVIDER only after the lookup fully converges, so Provide latency includes the whole walk.

go-libp2p's `lookup_optim.go` fires ADD_PROVIDER mid-lookup once a peer is statistically likely to be in the K-closest set (driven by a network-size estimator and gamma-function thresholds) and returns to the caller after ~75% of the RPCs complete, finishing the rest asynchronously.

Proposed work:
- Add a network-size estimator that tracks closest-peer distances across lookups.
- During a provide lookup, dispatch ADD_PROVIDER to peers that pass an "in K-closest" probability threshold; return after a configurable fraction complete, finishing the remainder in the background.

Impact: Provide latency. Medium (depends on provide workload).

gmelodie:
## Summary

Implements optimistic provide for the Kademlia DHT (issue #2860). Today `addProvider` sends `ADD_PROVIDER` only after the lookup fully converges, so provide latency includes the whole walk. This PR dispatches `ADD_PROVIDER` mid-lookup to peers that are statistically likely to be in the k-closest set. `addProvider` then returns once a fraction of the RPCs complete, and the remainder finish in the background. This cuts the long tail off provide latency.

The approach follows go-libp2p-kad-dht's `netsize` + `lookup_optim`:

- A network-size estimator (`netsize.nim`) tracks the normalized XOR distances of each converged lookup's k closest peers, grouped by closest-first index, and recovers the size by a weighted linear regression through the origin (`N = 1/slope - 1`).
- The estimate drives two distance thresholds through the inverse regularized lower incomplete gamma function. During a provide lookup, any newly found peer inside the individual threshold gets `ADD_PROVIDER` right away. The walk stops once k peers are covered, or once the k-closest average drops below the set threshold.
- `addProvider` returns after `ceil(k * 0.75)` RPCs complete. The rest run in the background, and `stop` cancels them.

The path is controlled by the new `optimisticProvide` config flag, which defaults to `true`. It needs a size estimate first, so early provides fall back to the classic path. Both paths feed the estimator, so it bootstraps without a chicken-and-egg dependency (a small, documented divergence from go-libp2p, which tracks only in the optimistic path).

## Affected Areas

- [x] Protocol Logic
  

Scope is confined to `libp2p/protocols/kademlia/`. No changes to transports, gossipsub, or the switch.

No downstream validation required. Kademlia is explicitly marked unstable / under active development. The flag is on by default, so an upgraded node changes its provide behavior as soon as its estimator has enough data; until then `addProvider` runs the classic path plus a passive network-size measurement. Set `optimisticProvide: false` to keep the classic path.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

- New config field `KadDHTConfig.optimisticProvide`, default `true`.
- `optimisticProvide` and `providerRejection` do not combine: once an estimate exists, the optimistic path runs and does not spill over to farther peers when a peer rejects the record. Turn `optimisticProvide` off if you rely on spillover.
- New public surface in the `kademlia` module: `NetworkSizeEstimator`, `NetSizeMeasurement`, `netsize.nim` (`track`, `networkSize`, `normedDistance`, `gammaIncRegInv`), a `kad_network_size_estimate` gauge, and `RoutingTable.commonPrefixLen` / `nPeersForCpl`.
- No migration required.

## Risk Assessment

- **Backward compatibility:** the config is additive, but the default is on, so provide behavior changes for every user once the estimator resolves.
- **Network behavior:** `ADD_PROVIDER` goes to more peers earlier, and `addProvider` returns before all RPCs settle. The background RPCs share the existing `maxConcurrentRpcs` semaphore, so total in-flight RPCs stay bounded, and `stop` cancels them.
- **Numerics:** the network-size estimate and the gamma thresholds are floating point. A bad estimate only shifts how eagerly records are stored; the correctness of a stored record is unaffected. `gammaIncRegInv` inverts the regularized lower incomplete gamma by bisection, and the tests pin it to `mpmath` reference values.
- **Security:** no new untrusted input paths; distances come from peer IDs already in the shortlist.

## References

- Closes #2860
- Parent: #1756
- go-libp2p-kad-dht `netsize/netsize.go` and `lookup_optim.go`

## Additional Notes

Optimistic mode treats a completed RPC as a success even if the peer replied `rejected`, which matches go-libp2p. Tests cover the gamma inverse (against `mpmath` references), `normedDistance`, estimator recovery of a seeded linear profile, and both the fallback and the optimistic provide path. The Kademlia suites pass locally, and CI runs the full suite.

### DIFF
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -8,11 +8,11 @@ import ../[peerid, switch, multihash]
 import ./protocol
 import
   ./kademlia/[routing_table, protobuf, types, find, get, put, keyspace, provider, ping]
-import ./kademlia/kademlia_metrics
+import ./kademlia/[kademlia_metrics, netsize]
 
 export
   chronicles, routing_table, protobuf, types, find, get, put, keyspace, provider, ping,
-  kademlia_metrics
+  kademlia_metrics, netsize
 
 logScope:
   topics = "kad-dht"
@@ -176,6 +176,7 @@ proc new*(
     config: config,
     providerManager:
       ProviderManager.new(config.providerRecordCapacity, config.providedKeyCapacity),
+    nsEstimator: NetworkSizeEstimator.new(config.replication),
     rpcSem: newAsyncSemaphore(config.limits.maxConcurrentRpcs),
     probeSem: newAsyncSemaphore(config.limits.maxConcurrentProbes),
   )
@@ -277,3 +278,7 @@ method stop*(kad: KadDHT) {.async: (raises: []).} =
   while kad.admissionProbes.len > 0:
     let admissionProbes = move kad.admissionProbes
     await noCancel admissionProbes.values.toSeq().cancelAndWait()
+
+  # Optimistic provide returns before its ADD_PROVIDER RPCs finish.
+  let provideTasks = move kad.provideTasks
+  await noCancel provideTasks.cancelAndWait()
--- a/libp2p/protocols/kademlia/kademlia_metrics.nim
+++ b/libp2p/protocols/kademlia/kademlia_metrics.nim
@@ -31,6 +31,8 @@ declarePublicCounter kad_provider_spillover_rounds,
 declarePublicCounter kad_provider_republish_regions,
   "keyspace regions republished, one DHT walk each"
 declarePublicCounter kad_provider_republish_keys, "provided keys republished"
+declarePublicGauge kad_network_size_estimate,
+  "estimated network size from lookup closest-peer distances"
 
 # Routing table metrics
 declarePublicGauge kad_routing_table_peers, "total peers in routing table"
--- a/libp2p/protocols/kademlia/netsize.nim
+++ b/libp2p/protocols/kademlia/netsize.nim
@@ -19,6 +19,7 @@ const
   MinMeasurementsThreshold = 5 ## per index, before an estimate is produced
   MaxMeasurementsThreshold = 150 ## newest kept per index
   TopBytesScale = 18446744073709551616.0 ## 2^64, the float64-usable distance part
+  MinStdDev = 1e-12 ## floor that keeps an inverse-variance weight finite
 
 func new*(T: typedesc[NetworkSizeEstimator], bucketSize: int): T =
   NetworkSizeEstimator(
@@ -83,7 +84,8 @@ func weightedStats(obs: seq[NetSizeMeasurement]): (float64, float64) =
 
 proc networkSize*(est: NetworkSizeEstimator): Result[int, string] =
   ## Current estimate, or an error while there is not enough data yet.
-  # Linear regression through the origin, standard deviations as fit weights.
+  # Linear regression through the origin, inverse variances as fit weights: a
+  # noisier index pulls the fit less than a stable one.
   let maxAgeTs = Moment.now() - MaxMeasurementAge
   var x2Sum, xySum = 0.0
   for i in 0 ..< est.bucketSize:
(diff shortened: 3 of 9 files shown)


vladopajic (approved):
lgtm; only comment is about decision when type def and type impl are together / separated.

for example `OptimisticState` has type def and impl in same file. which is okay. as general practice it is better for reading code as attention is on one file. 
  - note: could `OptimisticState` be in separate file and imported into `provider`?
  - would it make sense to split provider code to `provider_optimistic` new code from this pr, `provider_default` (naming?) the default implemtation and just `provider` file that will delegate proc call to on of these two files. not sure how code reuse wouldwork this is just a potential idea.

then `NetworkSizeEstimator` has type def and impl spited. i don't really see clear benefit for this decision in this case. then it's strange how importing / exporting works - because it still needs to be imported everwhere (impl file), so why not import everything to types and export it from types so there is no need for import/export?

gmelodie:
> then NetworkSizeEstimator has type def and impl spited. i don't really see clear benefit for this decision in this case.

`netsize.nim` needs `Key`, `XorDistance` and `RoutingTable` from `types.nim`. `KadDHT` in `types.nim` has the  `nsEstimator` field, so `types.nim` needs `NetworkSizeEstimator`. if we import `netsize` into `types.nim` and re-export, it would create a dep cycle.

> would it make sense to split provider code to provider_optimistic new code from this pr, provider_default 

hummmm i don't think it'd be a better organization, we'd also need to add a `provider_common` (naming?) to hold the common stuff to both of them. one interesting suggestion (for the future? but worth keeping in mind) from claude:

> If the file keeps growing, the better cut is by responsibility, publish and republish against serve and lookup, not by strategy.

vladopajic:
> and re-export, it would create a dep cycle.

i see.. all those entities `Key`, `RoutingTable`  could have been separate files?  different code organization in files might have been better approach? because right now there is not clear decision how code is organized, that is rule is: code must be in `types` when there is import cycle, otherwise it is free to live in separate file - seems like a smell to me

vladopajic:
pattern with `types` file is good for something else. for example:

rpc messages of some `module-a` can live `types`. then `module-b` does not really need to import anything else from `module-a` except `types`.  

`kad ` pacakge is something completely different. one needs to import _everything_ from kad package - or one file that exports everything. therefore motivation for `types` does not seem to make a lot of sense? or i am missing some benefit?

gmelodie:
> because right now there is not clear decision how code is organized, that is rule is: code must be in types when there is import cycle, otherwise it is free to live in separate file - seems like a smell to me

indeed... i think some standarization about this is in order (not only for kad, but other dirs as well). we could come up with a first version like in a pilot PR with kad (not this one tho), and if we like it try to extend that standard to other dirs/modules and document it. otherwise i see us going back and forth in bikeshedding over this and never really implementing it (cc @richard-ramos)

gmelodie:
> motivation for types does not seem to make a lot of sense? or i am missing some benefit?

The `types` split specifically is to make it possible to split other functionality inside `kad` into smaller files. So `find`, `get`, `put`, etc. can all import `types` instead of it all needing to be one big file

vladopajic:
> The `types` split specifically is to make it possible to split other functionality inside `kad` into smaller files. So `find`, `get`, `put`, etc. can all import `types` instead of it all needing to be one big file

there should be other options to achieve this (splitting kad type per function). 

only `KadDHT` needs to be in `types` (in this case better be in `kad.nim`).  then `kad_get.nim` will import `kad.nim` and `kad_provide.nim` will import `kad.nim`. and there is `module.nim` that imports and exports everything packaged for user (or any other file can serve this purpose, even `kad.nim`)

PR #2913, opened 2026-08-03
chore(cbind): change procs to static

### LINKED ISSUE #2906: cbind: creating a new private key should not require a context.
This is unnecessarily forcing the creation of a LibP2P node just to access the rng. 
https://github.com/vacp2p/nim-libp2p/blob/a6ab761c873b0720614a479debf2cc7f5a143ac8/cbind/libp2p.nim#L1146-L1148

Instead it should automatically create its own rng. This shouldnt be an expensive operation, compared to creating a node just to be able to generate a private key. `newRng()` is not really expensive compared to creating a node and destroying it, because what it does is:
- allocate a BearSSL HMAC-DRBG context
- initialize it with SHA-256
- seed it from the OS/system entropy source

And you would do that anyway when creating the node.

See https://github.com/logos-co/logos-libp2p-module/pull/96#discussion_r3690824819 for additional context

cc: @vladopajic / @gmelodie

vladopajic:
procs `libp2pNewPrivateKey` and `libp2pCollectMetrics` were depended on `LibP2P` although they didn't really used it. in this pr these procs have been changed to `ffiStatic` to avoid `LibP2P` argument.

closes:  #2906

### DIFF
--- a/cbind/examples/metrics.c
+++ b/cbind/examples/metrics.c
@@ -1,6 +1,6 @@
 // Metrics: a running node dumps the process-wide Prometheus registry as JSON
-// via libp2p_static_collect_metrics. The registry is global, so one started node
-// is enough to populate it. Calls are made blocking with the helpers in
+// via libp2p_static_collect_metrics. The registry is global, so one started 
+// node is enough to populate it. Calls are made blocking with the helpers in
 // common.h.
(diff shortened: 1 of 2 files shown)


gmelodie (approved):
YES

PR #2915, opened 2026-08-03
refactor(nat): a ReachabilityObservers store for reachability subscribers

gmelodie:
## Summary

This PR comes from #2910 (Kad-DHT dynamic client/server mode). That branch needs one thing from the NAT stack: a notice when reachability changes. No high-level hook exists for that, so in #2910 we needed to do:
```nim
 if self.reachability of AutonatService:
    AutonatService(self.reachability).statusAndConfidenceHandler(v1Handler())
  elif self.reachability of HPService:
    HPService(self.reachability).statusAndConfidenceHandler(v1Handler())
  elif self.reachability of AutonatV2Service:
  # ...
```

This PR adds the producer-side interface instead. Then #2910 rebases onto this PR and deletes the chain.

Three types produce reachability today. Each type has its own setter and its own handler signature.

| Type | Setter | Handler parameters |
| --- | --- | --- |
| `AutonatService` | `statusAndConfidenceHandler` | reachability, confidence |
| `AutonatV2Service` | `setStatusAndConfidenceHandler` | reachability, confidence, `dialBackAddr` |
| `HPService` | none; it owns an `AutonatService` | - |

`NATService` keeps the active backend in a field of type `Service`. Therefore it cannot ask the backend anything.

Both setters also replace the single stored handler. `HPService.setup` installs its own handler in that slot. That handler starts AutoRelay and writes the transport dial hint. A second subscriber removes it, and no warning appears. Therefore support for many subscribers is a prerequisite, not an extra.

This PR adds `ReachabilityObservers`, one store of subscribers with one handler signature. Every backend holds such a store, and `NATService` shares its own store with the backend that its config selects. `NATService` then gets the three calls that a consumer needs:

```nim
proc addReachabilityHandler*(self: NATService, handler: ReachabilityHandler): bool
proc removeReachabilityHandler*(self: NATService, handler: ReachabilityHandler): bool
proc networkReachability*(self: NATService): NetworkReachability
```

To add a fourth backend (AutoNAT v3, or a test stub), give it a `ReachabilityObservers` field and notify it. The call site needs no signature adapter.

## Affected Areas

- [x] Peer Management / Discovery — AutoNAT v1, AutoNAT v2, hole punching, `NATService`
- [x] Protocol Logic — the reachability notice path

Production changes:

- `libp2p/services/reachabilityobservers.nim` — a new file. It holds `ReachabilityHandler`, the one handler signature (reachability, confidence, `dialBackAddr`), and `ReachabilityObservers`, the one store. The store appends a handler with `add`, it drops one with `remove`, it reports the last value with `lastReachability`, and it dispatches with `notify`. Both `add` and `remove` return a bool, so the caller sees whether the call changed the store. AutoNAT v1 reports `dialBackAddr` as none, because it observes no dial-back address.
- `libp2p/utils/future.nim` — the file adds `allOrCancel`. It awaits every future in a seq, also a future that fails. If the caller cancels it, it cancels every future first, because `allFutures` leaves its children to run. `notify` uses it.
- `libp2p/utils/collections.nim` — the file adds `removeFirstIfIt`. It removes the first element that satisfies a condition and it stops there. `ReachabilityObservers.remove` uses it.
- `libp2p/protocols/connectivity/autonat/service.nim` and `.../autonatv2/service.nim` — the single handler field becomes a public `reachabilityObservers` field. `callHandler` becomes a `notify` call. The v1 store adapts the two-parameter handler that the deprecated setter takes.
- `libp2p/services/hpservice.nim` — `HPService` produces no reachability of its own, so `reachabilityObservers` hands out the store of the `AutonatService` that it drives. It also moves its own handler to that store.
- `libp2p/services/natservice.nim` — `NATService` builds one store in `new`, and `setup` hands that store to the backend that the config selects. The store is therefore never nil, and a consumer can subscribe before or after setup.

Test changes: concurrent delivery to three subscribers and a check that an unsubscribed handler gets nothing, delivery through both AutoNAT backends, a check that `HPService` shares the store of its `AutonatService`, per-backend coverage in the `NATService` tests that already exist, unit tests for the store and for `removeFirstIfIt`, and a nil-handler check. The rest is a mechanical move of the call sites to `reachabilityObservers.add`.

## Dispatch

`notify` starts every handler in subscription order, and then it awaits them together. A handler that blocks does not delay the handlers behind it.

Cancellation has two cases:

- The service cancels `notify`. It then cancels every handler future, it waits for all of them, and it re-raises `CancelledError`. No handler future stays behind.
- One handler raises `CancelledError`. The other handlers still run, and `notify` returns. One subscriber can no longer stop the AutoNAT loop.

`HPService` subscribes in `setup`, before a consumer can reach the service. Its handler therefore starts first, but it no longer completes first. That handler writes the transport dial hint after it starts or stops AutoRelay. A subscriber that reads `transport.networkReachability` in that window reads the previous value. No component reads that hint today, and the hint keeps its position in the handler, because a move would change which relay dials reuse the local port.

No downstream integration branch exists yet. The surface that matters for Nimbus is `AutonatService.statusAndConfidenceHandler`. It still compiles: it stays as a deprecated setter that appends to the store. `config.nims` does not list `Deprecated` in `warningAsError`. Therefore a downstream build does not break either. Every call site in this repo moves to the new API, so this repo emits no deprecation warning of its own.

- **Nimbus:** N/A — no source change is necessary
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

New API:

- `ReachabilityHandler` and `ReachabilityObservers` in `libp2p/services/reachabilityobservers`, re-exported by both AutoNAT services, by `hpservice` and by `natservice`
- the `reachabilityObservers` field on `AutonatService` and on `AutonatV2Service`, and the `reachabilityObservers` proc on `HPService`
- `NATService.addReachabilityHandler`, `NATService.removeReachabilityHandler`, and `NATService.networkReachability`
- `allOrCancel` in `utils/future`, and `removeFirstIfIt` in `utils/collections`

Deprecated, and still correct: `AutonatService.statusAndConfidenceHandler` and `AutonatV2Service.setStatusAndConfidenceHandler`.

One change of behaviour needs your attention. The two deprecated setters now append. They no longer replace. Code that calls a setter twice to install two handlers ends up with both handlers, which is the point. Code that calls a setter twice to swap a handler must call `reachabilityObservers.remove` on the first handler. The deprecation message states this. No design keeps replace-semantics and supports many subscribers at the same time, and replace-semantics is what removed the hole-punching handler without a warning.

`add` returns false when it changes nothing: the handler is nil, or the same handler is subscribed already. One handler therefore holds one subscription, and one `remove` call ends it. `remove` returns false when the handler was not subscribed. `NATService.addReachabilityHandler` also returns false when the config selects no reachability backend, because then no notice can ever arrive. The caller can log that the subscription had no effect.

A nil handler would crash the next dispatch, so it never reaches the store.

`NATService.networkReachability` also serves a late subscriber. The subscriber reads the current value, and it does not wait for the next change.

## Risk Assessment

- **Backward compatibility:** the change is additive. No type changes its base class, so downstream code that holds a backend as a `Service` still compiles. The append-instead-of-replace setter above is the one change of behaviour.
- **Network behaviour:** unchanged. The PR touches no wire format, no timing, and no schedule.
- **Performance:** the dispatch path builds one seq of futures, on a path that runs only when reachability or confidence changes. Concurrent handlers replace serial handlers, so the dispatch takes as long as the slowest subscriber, not as long as the sum.
- **Security:** none.

## Known limitations

`stop` does not clear the store. A caller that subscribes on each restart must unsubscribe on each stop, or the store grows.

## References

- #2910 — the PR that this work comes from, and the first consumer. It carries the `of` chain that this interface replaces.

### DIFF
--- a/libp2p/protocols/connectivity/autonatv2/service.nim
+++ b/libp2p/protocols/connectivity/autonatv2/service.nim
@@ -13,6 +13,7 @@ import
   ../../../peerid,
   ../../../wire,
   ../../../utils/heartbeat,
+  ../../../utils/future,
   ../../../crypto/crypto,
   ../autonat/types,
   ./types,
@@ -40,11 +41,11 @@ type
     enableAddressMapper: bool
     enableDialableCandidates: bool
 
-  AutonatV2Service* = ref object of Service
+  AutonatV2Service* = ref object of ReachabilityService
     config*: AutonatV2ServiceConfig
     confidence: Opt[float]
     newConnectedPeerHandler: PeerEventHandler
-    statusAndConfidenceHandler: StatusAndConfidenceHandler
+    subscribers: Subscribers[StatusAndConfidenceHandler]
     addressMapper: AddressMapper
     scheduleHandle: Future[void]
     networkReachability*: NetworkReachability
@@ -96,10 +97,13 @@ proc new*(
 proc callHandler(
     self: AutonatV2Service, dialBackAddr: Opt[MultiAddress]
 ) {.async: (raises: [CancelledError]).} =
-  if not isNil(self.statusAndConfidenceHandler):
-    await self.statusAndConfidenceHandler(
-      self.networkReachability, self.confidence, dialBackAddr
+  # Handlers start in subscription order, then run concurrently, so a subscriber
+  # that blocks does not delay the subscribers behind it.
+  await allOrCancel(
+    self.subscribers.handlers.mapIt(
+      it(self.networkReachability, self.confidence, dialBackAddr)
     )
+  )
 
 proc hasEnoughIncomingSlots(switch: Switch): bool =
   # we leave some margin instead of comparing to 0 as a peer could connect to us while we are asking for the dial back
@@ -287,7 +291,37 @@ method stop*(
     switch.peerInfo.addressMappers.keepItIf(it != self.addressMapper)
   await switch.peerInfo.update()
 
+proc addStatusAndConfidenceHandler*(
+    self: AutonatV2Service, handler: StatusAndConfidenceHandler
+): SubscriptionId {.discardable.} =
+  self.subscribers.subscribe(handler)
+
+proc removeStatusAndConfidenceHandler*(
+    self: AutonatV2Service, id: SubscriptionId
+): bool =
+  self.subscribers.unsubscribe(id)
+
 proc setStatusAndConfidenceHandler*(
-    self: AutonatV2Service, statusAndConfidenceHandler: StatusAndConfidenceHandler
-) =
-  self.statusAndConfidenceHandler = statusAndConfidenceHandler
+    self: AutonatV2Service, handler: StatusAndConfidenceHandler
+) {.deprecated: "use addStatusAndConfidenceHandler; it appends, it does not replace".} =
+  self.addStatusAndConfidenceHandler(handler)
+
+method addReachabilityHandler*(
+    self: AutonatV2Service, handler: ReachabilityHandler
+): SubscriptionId {.discardable.} =
+  if handler.isNil():
+    return NoSubscription
+  self.addStatusAndConfidenceHandler(
+    proc(
+        networkReachability: NetworkReachability,
+        confidence: Opt[float],
+        dialBackAddr: Opt[MultiAddress],
+    ) {.async: (raises: [CancelledError]).} =
+      await handler(networkReachability)
+  )
+
+method removeReachabilityHandler*(self: AutonatV2Service, id: SubscriptionId): bool =
+  self.removeStatusAndConfidenceHandler(id)
+
+method networkReachability*(self: AutonatV2Service): NetworkReachability =
+  self.networkReachability
--- a/libp2p/protocols/connectivity/autonatv2/service.nim
+++ b/libp2p/protocols/connectivity/autonatv2/service.nim
@@ -13,11 +13,13 @@ import
   ../../../peerid,
   ../../../wire,
   ../../../utils/heartbeat,
-  ../../../utils/future,
+  ../../../services/reachabilityservice,
   ../../../crypto/crypto,
   ../autonat/types,
   ./types,
   ./client
+
+export reachabilityservice
 
 declarePublicGauge(
   libp2p_autonat_v2_reachability_confidence,
@@ -45,7 +47,6 @@ type
     config*: AutonatV2ServiceConfig
     confidence: Opt[float]
     newConnectedPeerHandler: PeerEventHandler
-    subscribers: Subscribers[StatusAndConfidenceHandler]
     addressMapper: AddressMapper
     scheduleHandle: Future[void]
     networkReachability*: NetworkReachability
@@ -53,11 +54,9 @@ type
     client*: AutonatV2Client
     rng: Rng
 
-  StatusAndConfidenceHandler* = proc(
-    networkReachability: NetworkReachability,
-    confidence: Opt[float],
-    dialBackAddr: Opt[MultiAddress],
-  ): Future[void] {.gcsafe, async: (raises: [CancelledError]).}
+  StatusAndConfidenceHandler* = ReachabilityHandler
+    ## The handler of the replaced single-subscriber API. It has the same
+    ## signature as `ReachabilityHandler`; use that name instead.
 
 proc new*(
     T: typedesc[AutonatV2ServiceConfig],
@@ -92,17 +91,6 @@ proc new*(
     answers: initDeque[NetworkReachability](),
     client: client,
     rng: rng,
-  )
-
-proc callHandler(
-    self: AutonatV2Service, dialBackAddr: Opt[MultiAddress]
-) {.async: (raises: [CancelledError]).} =
-  # Handlers start in subscription order, then run concurrently, so a subscriber
-  # that blocks does not delay the subscribers behind it.
-  await allOrCancel(
-    self.subscribers.handlers.mapIt(
-      it(self.networkReachability, self.confidence, dialBackAddr)
-    )
   )
 
 proc hasEnoughIncomingSlots(switch: Switch): bool =
@@ -203,7 +191,7 @@ proc askPeer(
       Unknown
   let hasReachabilityOrConfidenceChanged = await self.handleAnswer(ans)
   if hasReachabilityOrConfidenceChanged:
-    await self.callHandler(dialBackAddr)
+    await self.callHandlers(self.confidence, dialBackAddr)
   await switch.peerInfo.update()
   return ans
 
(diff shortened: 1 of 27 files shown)


libp2p/protocols/connectivity/autonatv2/service.nim
```diff
@@ -287,7 +291,37 @@ method stop*(
...
-  self.statusAndConfidenceHandler = statusAndConfidenceHandler
+    self: AutonatV2Service, handler: StatusAndConfidenceHandler
+) {.deprecated: "use addStatusAndConfidenceHandler; it appends, it does not replace".} =
+  self.addStatusAndConfidenceHandler(handler)
+
+method addReachabilityHandler*(
+    self: AutonatV2Service, handler: ReachabilityHandler
+): SubscriptionId {.discardable.} =
+  if handler.isNil():
+    return NoSubscription
```
vladopajic: new line after early return improves code readability a lot. few other places in this pr as well

libp2p/protocols/connectivity/autonatv2/service.nim
```diff
@@ -287,7 +291,37 @@ method stop*(
...
+
 proc setStatusAndConfidenceHandler*(
-    self: AutonatV2Service, statusAndConfidenceHandler: StatusAndConfidenceHandler
-) =
-  self.statusAndConfidenceHandler = statusAndConfidenceHandler
+    self: AutonatV2Service, handler: StatusAndConfidenceHandler
+) {.deprecated: "use addStatusAndConfidenceHandler; it appends, it does not replace".} =
+  self.addStatusAndConfidenceHandler(handler)
+
+method addReachabilityHandler*(
```
vladopajic: seems strange:

`addReachabilityHandler` with argument `ReachabilityHandler`
  -> which calls `self.addStatusAndConfidenceHandler` and creates different handler that delegates to `ReachabilityHandler`.

`addReachabilityHandler` seems like a helper function that creates adapter from one interface to another, why is it a method? 

why is helper needed, why not just leave `confidence` and `dialBackAddr` arguments unused?

libp2p/protocols/connectivity/autonatv2/service.nim
```diff
@@ -287,7 +291,37 @@ method stop*(
     switch.peerInfo.addressMappers.keepItIf(it != self.addressMapper)
   await switch.peerInfo.update()
 
+proc addStatusAndConfidenceHandler*(
+    self: AutonatV2Service, handler: StatusAndConfidenceHandler
+): SubscriptionId {.discardable.} =
```
vladopajic: `SubscriptionId` seems unnecessary for this pattern. `handler` is id by itself.

libp2p/switch.nim
```diff
@@ -87,6 +88,62 @@ method stop*(
...
+  .}
+
+  SubscriptionId* = distinct int
+
+  Subscribers*[T] = object ## Handlers, each with a handle that unsubscribes it again.
+    entries: seq[tuple[id: SubscriptionId, handler: T]]
+    lastId: int
+
+  ReachabilityService* = ref object of Service
+    ## A Service that observes network reachability and notifies subscribers.
```
vladopajic: why `ReachabilityService` needs to be in `switch.nim` when Switch don't really care about any abstraction other then `Service`?

having it in `switch.nim` will unnecessary expose (export) this type to user as public interface. having more types in public interface make public interface hard to change without introducing braking changes.

libp2p/services/hpservice.nim
```diff
@@ -95,6 +97,17 @@ proc newConnectedPeerHandler(
   except CatchableError as err:
     debug "Hole punching failed during dcutr", description = err.msg
 
+method addReachabilityHandler*(
+    self: HPService, handler: ReachabilityHandler
+): bool {.discardable.} =
```
vladopajic: `discardable` not sure about use of that in general. it's painless in tests to add `check` before func call,
`check addReachabilityHandler()`

and other code should discard result or not.

libp2p/services/reachabilityobservers.nim
```diff
@@ -0,0 +1,63 @@
...
+    confidence: Opt[float],
+    dialBackAddr: Opt[MultiAddress],
+  ): Future[void] {.gcsafe, async: (raises: [CancelledError]).}
+    ## A subscriber of reachability changes. A subscriber that needs no
+    ## `confidence` or `dialBackAddr` leaves those parameters unused. AutoNAT v1
+    ## reports `dialBackAddr` as none, because it observes no dial-back address.
+
+  ReachabilityService* = ref object of Service
+    ## A Service that observes network reachability and notifies subscribers.
+    handlers: seq[ReachabilityHandler]
```
vladopajic, line 27: it seems to me that `ReachabilityService` as inheritance mechanism here is wrong, approach. 

`ReachabilityService` should be `ReachabilityObservers`

```
let ro: ReachabilityObservers()

ro.add(handler)
ro.remove(handler)
ro.notify(confidence, dialBackAddr).
```

those types that needs this behavior should just add this field:

```
type MyServce* = ref object of Service
   rechabilityObservers*: ReachabilityObservers
```

tests/libp2p/protocols/test_autonat_v2_service.nim
```diff
@@ -130,6 +130,48 @@ suite "AutonatV2 Service":
...
+    let
+      (v2Service, _) = newService(NetworkReachability.Reachable)
+      switch = createSwitch(Opt.some(v2Service))
+      observers = v2Service.reachabilityObservers
+    var switches = createSwitches(3)
+
+    let first = newFuture[NetworkReachability]()
+    let second = newFuture[NetworkReachability]()
+
+    discard observers.add(
```
vladopajic: `discard` -> `check`

tests/libp2p/protocols/test_autonat_service.nim
```diff
@@ -84,6 +84,70 @@ suite "Autonat Service":
...
+    var goneCalls = 0
+
+    let gone: ReachabilityHandler = proc(
+        reachability: NetworkReachability,
+        confidence: Opt[float],
+        dialBackAddr: Opt[MultiAddress],
+    ) {.async: (raises: [CancelledError]).} =
+      goneCalls.inc()
+
+    discard observers.add(
```
vladopajic: `discard` -> `check`

tests/libp2p/protocols/test_autonat_service.nim
```diff
@@ -84,6 +84,70 @@ suite "Autonat Service":
...
+    await switch2.start()
+    await switch1.connect(switch2.peerInfo.peerId, switch2.peerInfo.addrs)
+
+    let
+      firstSeen = await first.wait(5.seconds)
+      lastSeen = await last.wait(5.seconds)
+    check:
+      firstSeen == NetworkReachability.NotReachable
+      lastSeen == NetworkReachability.NotReachable
+      not firstDone # the last handler answered while the first one still blocks
```
vladopajic: ` not firstDone # the last handler answered while the first one still blocks`
impossible to happen becaue:

- gate.fire() is happening after that assertion
- gate.await() is happening before setting `firstDone = true`

tests/libp2p/protocols/test_autonat_service.nim
```diff
@@ -84,6 +84,70 @@ suite "Autonat Service":
...
+    await switch1.start()
+    await switch2.start()
+    await switch1.connect(switch2.peerInfo.peerId, switch2.peerInfo.addrs)
+
+    let
+      firstSeen = await first.wait(5.seconds)
+      lastSeen = await last.wait(5.seconds)
+    check:
+      firstSeen == NetworkReachability.NotReachable
+      lastSeen == NetworkReachability.NotReachable
```
vladopajic:
```
check:
  await first == NetworkReachability.NotReachable
  await last == NetworkReachability.NotReachable

```

variables seem unnecessary

tests/libp2p/protocols/test_autonat_service.nim
```diff
@@ -84,6 +84,70 @@ suite "Autonat Service":
...
+      not firstDone # the last handler answered while the first one still blocks
+      goneCalls == 0
+      autonatService.networkReachability == NetworkReachability.NotReachable
```
vladopajic: `NetworkReachability.NotReachable` here is hardodec but it should be linked to it's source of truth, that is this (few lines above):

```nim
autonatClientStub.answer = NotReachable
```
in other words all assertion here should do:

```nim
check:
  await first == autonatClientStub.answer
  await second == autonatClientStub.answer
  autonatService.networkReachability == autonatClientStub.answer

# or use some const for  autonatClientStub.answer
```

libp2p/services/natservice.nim
```diff
@@ -97,6 +98,31 @@ proc autonatV2Service*(self: NATService): Opt[AutonatV2Service] =
...
   Opt.none(AutonatV2Service)
 
+proc addReachabilityHandler*(self: NATService, handler: ReachabilityHandler): bool =
+  ## Subscribes `handler` to the active reachability service (AutoNAT v1/v2 or
+  ## hole-punching). False means no subscription, because no reachability service
+  ## is configured or because `handler` is nil. The caller can then log that the
+  ## subscription had no effect.
+  if self.observers.isNil():
+    return false
+
```
vladopajic, line 117: note sure about this, i would expect this not ever happen - code needs to make sure this never happens.
the best thing to do here is `doAssert ...`?

because after `NATService.setup` is finished `self.observers` will be set; and before that it doesn't make sense to call `addReachabilityHandler` or it will be invalid to call?

vladopajic (approved):
looks really nice; minor comments

libp2p/services/reachabilityobservers.nim
```diff
@@ -0,0 +1,62 @@
...
+    return false
+
+  self.handlers.add(handler)
+  true
+
+proc remove*(self: ReachabilityObservers, handler: ReachabilityHandler): bool =
+  ## Removes `handler`. False means it was not subscribed.
+  let before = self.handlers.len
+  self.handlers.keepItIf(it != handler)
+  self.handlers.len < before
```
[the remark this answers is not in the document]
vladopajic: i am fine with current code. but observer that forbid adding same handler twice seems like hardening.

then if that is implemented `keepItIf` seems inefficient then something that stops early. which is neglectable improvement for this case, but in general it would be good to have better pattern that will just remove single element and return: `someSeq.removeFirstIfIt(it == handler)` also seems more natural to read this code.

PR #2917, opened 2026-08-06
fix(service-disco): let a caller advertise a full XPR

gmelodie:
## Summary

`libp2pServiceDiscoStartAdvertising` could only advertise this node's own record. A caller that runs its service on a different switch had no way to publish that switch's peer id and addresses.

The core library already carried the plumbing. `startAdvertising(disco, service, advert: Opt[seq[byte]])` stores the caller's bytes in `advertiser.providedAdverts` and republishes them verbatim on every rotation. Only the FFI wrapper dropped the argument: `StartAdvertisingRequest` in `cbind/libp2p.nim` carried `serviceId` and `serviceData` alone, so there was no way to pass a signed extended peer record across the C boundary.

This PR adds `advertisement: seq[byte]` to that request and forwards it when it is non-empty. An empty field keeps the old behaviour.

It also closes the hole that feature opens. `addProvidedService` accepted its `advert` argument without checking it, so any caller could store arbitrary bytes in `providedAdverts` and have the node rebroadcast them to registrars on every rotation. `validateAdvert` now applies the same checks a registrar applies in `isValidAdvertisement`: the record must decode as a signed extended peer record, stay inside `MaxXPRSize` and `MaxServiceDataSize`, and list the service id among its services.

Reporting that rejection needs an error channel, so `addProvidedService` and `startAdvertising` now return `Result[void, string]`. That also surfaces three failures they previously swallowed: oversized service data, a missing routing table, and a record that cannot be built.

## Affected Areas

  `addProvidedService` and `startAdvertising` in `libp2p/protocols/service_discovery/advertiser.nim` gain validation and a return type.

- [x] Build / Tooling
  `StartAdvertisingRequest` gains the field; `libp2pServiceDiscoStartAdvertising` propagates the core error.

- **Nimbus:** N/A — does not consume service discovery.
- **Waku:** N/A — same reason.
- **Codex:** N/A — same reason.

Downstream consumer of the C bindings: https://github.com/logos-co/logos-libp2p-module/pull/99 needs this field. That PR adds a third `advertisement` argument (a base64 signed XPR) to its `discoStartAdvertising` API and does not build until this lands.

## Impact on Library Users

Nim users: `addProvidedService` and `startAdvertising` change from `void` to `Result[void, string]`. A caller that ignored the outcome now handles it or discards it explicitly. Both procs are new to the service-discovery protocol and have no downstream consumers in Nimbus, Waku, or Codex.

C binding users: additive. `StartAdvertisingRequest` gains one optional field. A caller that omits it, or sends an empty byte string, gets the previous behaviour. A caller that sets it gets an error string instead of a false success when the record is malformed, unsigned, oversized, or advertises a different service. The generated C header and CDDL schema change, so a consumer regenerates or updates its bindings.

## Risk Assessment

- Backward compatibility: the FFI field is additive and an empty value selects the old path. The Nim-side return-type change is source-breaking for any caller that ignored the result; all in-tree callers are updated.
- Network behaviour: unchanged when the field is empty. When it is set and valid, the node publishes the caller's record instead of its own, which is the intent. An invalid record is now rejected before it reaches `providedAdverts`, where it would otherwise have been rebroadcast on every rotation.
- Security: the supplied bytes cross a trust boundary and are checked at that boundary. `Advertisement.decode` verifies the envelope signature and runs `checkValid()`, `isValid()` bounds the sizes, and `advertisesService` confirms the service id.
- Performance: one decode per `startAdvertising` call that supplies a record. The bytes are stored once and reused on each rotation.

## References

- Issue: https://github.com/logos-co/logos-libp2p-module/issues/98
- Downstream PR: https://github.com/logos-co/logos-libp2p-module/pull/99

## Additional Notes

`tests/libp2p/service_discovery/test_advertiser.nim` gains a suite covering the four outcomes: a valid record is stored for reuse, and bytes that do not decode, that name another service, or that carry oversized service data are each rejected without registering the service. `makeOversizedAdvertisement` in the test utils builds the last case.

### DIFF
--- a/cbind/libp2p.nim
+++ b/cbind/libp2p.nim
@@ -171,6 +171,7 @@ type ExtendedRecordsResponse {.ffi.} = object
 type StartAdvertisingRequest {.ffi.} = object
   serviceId: string
   serviceData: seq[byte]
+  advertisement: seq[byte]
 
 type CreateXprRequest {.ffi.} = object
   addrs: seq[string]
@@ -914,12 +915,24 @@ proc libp2pServiceDiscoStop*(lib: LibP2P): Future[Result[bool, string]] {.ffi.} 
 proc libp2pServiceDiscoStartAdvertising*(
     lib: LibP2P, req: StartAdvertisingRequest
 ): Future[Result[bool, string]] {.ffi.} =
-  ## Advertises `serviceId` (with optional `serviceData`) in this node's record.
+  ## Advertises `serviceId` (with `serviceData`, which may be empty) in this
+  ## node's record. A non-empty `advertisement` is a signed extended peer record,
+  ## published verbatim instead of this node's own record; it fails when that
+  ## record does not decode, is oversized, or does not list `serviceId`.
   let disco = resolveServiceDiscovery(lib).valueOr:
     return err(error)
+
+  let advert =
+    if req.advertisement.len == 0:
+      Opt.none(seq[byte])
+    else:
+      Opt.some(req.advertisement)
+
   disco.startAdvertising(
-    ServiceInfo(id: req.serviceId, data: Opt.some(req.serviceData))
-  )
+    ServiceInfo(id: req.serviceId, data: Opt.some(req.serviceData)), advert
+  ).isOkOr:
+    return err(error)
+
   ok(true)
 
 proc libp2pServiceDiscoStopAdvertising*(
--- a/cbind/libp2p.nim
+++ b/cbind/libp2p.nim
@@ -922,14 +922,9 @@ proc libp2pServiceDiscoStartAdvertising*(
   let disco = resolveServiceDiscovery(lib).valueOr:
     return err(error)
 
-  let advert =
-    if req.advertisement.len == 0:
-      Opt.none(seq[byte])
-    else:
-      Opt.some(req.advertisement)
-
   disco.startAdvertising(
-    ServiceInfo(id: req.serviceId, data: Opt.some(req.serviceData)), advert
+    ServiceInfo(id: req.serviceId, data: Opt.some(req.serviceData)),
+    Opt.noneWhenEmpty(req.advertisement),
   ).isOkOr:
     return err(error)
 
--- a/libp2p/protocols/service_discovery/advertiser.nim
+++ b/libp2p/protocols/service_discovery/advertiser.nim
@@ -292,17 +292,35 @@ proc advertiseToRegistrar*(
       debug "registrar rejection, aborting", serviceId, registrar
       return
 
+proc validateAdvert(advert: seq[byte], service: ServiceInfo): Result[void, string] =
+  ## Applies the checks a registrar applies in `isValidAdvertisement`, so a bad
+  ## record fails here instead of being republished on every rotation.
+  let ad = Advertisement.decode(advert).valueOr:
+    return err("cannot decode advertisement: " & $error)
+
+  if not ad.isValid():
+    return err(
+      "oversized advertisement: the record must stay under " & $MaxXPRSize &
+        " bytes and each service data under " & $MaxServiceDataSize & " bytes"
+    )
+
+  if not ad.advertisesService(service.id.hashServiceId()):
+    return err("advertisement does not advertise service '" & service.id & "'")
+
+  ok()
+
 proc addProvidedService*(
     disco: ServiceDiscovery,
     service: ServiceInfo,
     advert: Opt[seq[byte]] = Opt.none(seq[byte]),
-) =
+): Result[void, string] =
   doAssert not disco.clientMode, "not supported in client mode"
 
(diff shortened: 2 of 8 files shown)


cbind/libp2p.nim
```diff
@@ -914,12 +915,24 @@ proc libp2pServiceDiscoStop*(lib: LibP2P): Future[Result[bool, string]] {.ffi.}
...
+  ## published verbatim instead of this node's own record; it fails when that
+  ## record does not decode, is oversized, or does not list `serviceId`.
   let disco = resolveServiceDiscovery(lib).valueOr:
     return err(error)
+
+  let advert =
+    if req.advertisement.len == 0:
+      Opt.none(seq[byte])
+    else:
+      Opt.some(req.advertisement)
```
vladopajic: utility `Opt.noneWhenEmpty(req.advertisement)`can work for any `seq[T]`

vladopajic (approved):

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -292,42 +294,60 @@ proc advertiseToRegistrar*(
       debug "registrar rejection, aborting", serviceId, registrar
       return
 
+proc validateAdvert(advert: seq[byte], service: ServiceInfo): Result[void, string] =
+  ## Applies the checks a registrar applies in `isValidAdvertisement`, so a bad
+  ## record fails here instead of being republished on every rotation.
+  let ad = Advertisement.decode(advert).valueOr:
```
richard-ramos, line 300: Length must be checked before decoding too because the protobuffer decoder can ignore unknown fields. 
 
```suggestion
  if advert.len > MaxXPRSize:
    return err(
      "oversized advertisement: the encoded record must stay under " &
        $MaxXPRSize & " bytes"
    )

  let ad = Advertisement.decode(advert).valueOr:
```

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -292,42 +294,60 @@ proc advertiseToRegistrar*(
...
+    ?validateAdvert(advert.get(), service)
 
   let serviceId = service.id.hashServiceId()
 
   if not disco.rtManager.addService(
     serviceId, disco.rtable, disco.config.replication, disco.discoConfig.bucketsCount,
     Provided,
   ):
-    return
+    return ok()
```
richard-ramos, line 334: A problem with returning ok here is that the `providedAdverts` are not updated with the new `advert`. I wonder if instead you should return an `err` indicating that the service is already provided. So you would then have to call `stopAdvertising` first, and then call `startAdvertising`.

In any case, we do need a test unit for this. case

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -233,7 +233,9 @@ proc advertiseToRegistrar*(
     ticket: Opt[Ticket],
     advert: seq[byte],
 ) {.async: (raises: [CancelledError]).} =
-  doAssert not disco.clientMode, "not supported in client mode"
+  if not disco.isServer:
+    debug "not advertising while in client mode", serviceId, registrar
+    return
```
richard-ramos, line 238:
```suggestion
```

libp2p/protocols/service_discovery/advertiser.nim
richard-ramos, line 252: Since you can use `changeMode`, the check needs to be done continuosly
```suggestion
  while true:
    if not disco.isServer:
      trace "not advertising while in client mode", serviceId, registrar
      return
```

richard-ramos: Probably same type of check needs to be done in `maintainRegistrations` and in `startLocalRegistration`

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -292,42 +294,60 @@ proc advertiseToRegistrar*(
...
+    return err("routing table missing for service '" & service.id & "'")
 
   # When a caller supplied an explicit advert we store it so that future
   # rotations / replacements (maintenance) reuse exactly the same bytes.
   if advert.isSome():
     disco.advertiser.providedAdverts[serviceId] = advert.get()
 
   let advertBytes = disco.getAdvertBytes(advert).valueOr:
-    return
+    return err("cannot build the extended peer record to advertise")
```
richard-ramos, line 350: @SionoiS 
while reviewing this PR, i noticed that we return errors in diff points of the code, but keep the state 'partially changed', like when we call `rtManager.addService`, modify `disco.services` or increase metrics. Do you think this is something that should be fixed or it's not really a problem / wont happen IRL?

SionoiS: The error paths are cases that normally won't happen except if the code is wrong.

I think it should be fine.

PR #2918, opened 2026-08-06
refactor(kad): tables into virtual system

SionoiS:
Instead of each table owning peers, a new registry own peer information and tables are just referencing this.

This avoid copying peers to multiple tables and simplify peer state management.

### DIFF
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -212,11 +212,9 @@ proc maintainLiveness(kad: KadDHT) {.async: (raises: [CancelledError]).} =
   ## bucket refresh so dead peers are not held until the next refresh tick.
   ## At most one probe is in flight per peer across all maintainable tables.
 
-  var idle = true
   while not kad.stopping:
-    if idle:
-      # Initial idle: let start() finish bootstrap before the first scan.
-      await sleepAsync(kad.config.livenessIdleInterval)
+    # Initial idle: let start() finish bootstrap before the first scan.
+    await sleepAsync(kad.config.livenessIdleInterval)
 
     let grace = kad.config.livenessGracePeriod
 
@@ -240,9 +238,6 @@ proc maintainLiveness(kad: KadDHT) {.async: (raises: [CancelledError]).} =
       except CancelledError as exc:
         await noCancel inFlight.cancelAndWait()
         raise exc
-      idle = false
-    else:
-      idle = true
 
 proc maintainLiveness(kad: KadDHT) {.async: (raises: [CancelledError]).} =
   ## Continuous background task: drain replaceable peers via liveness probes
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -264,8 +264,7 @@ proc maintainLiveness(kad: KadDHT) {.async: (raises: [CancelledError]).} =
         kad.launchLivenessProbe(peerId)
 
     if kad.livenessProbes.len > 0:
-      debug "Waiting for in-flight liveness probes",
-        inFlight = kad.livenessProbes.len
+      debug "Waiting for in-flight liveness probes", inFlight = kad.livenessProbes.len
       let inFlight = kad.livenessProbes.values.toSeq()
       try:
         discard await one(inFlight)
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -65,6 +65,17 @@ proc checkAndEvictPeer(
     debug "Liveness probe skipped: no free slot", peer = peerId.shortLog()
     kad_routing_table_liveness_probes.inc(labelValues = ["skipped"])
     return
+  kad.livenessProbes[probeKey] = probe
+  probe.addCallback(
+    proc(udata: pointer) {.gcsafe, raises: [].} =
+      if kad.livenessProbes.getOrDefault(probeKey) == probe:
+        kad.livenessProbes.del(probeKey)
+  )
+
+proc checkAndEvictPeer(
+    kad: KadDHT, rtable: RoutingTable, peerId: PeerId
+) {.async: (raises: [CancelledError]).} =
+  ## Takes ownership of one ``probeSem`` slot already acquired by the caller.
   defer:
     try:
       kad.livenessSem.release()
@@ -83,11 +94,6 @@ proc checkAndEvictPeer(
   if dueTables.len == 0:
     debug "Liveness probe skipped: peer no longer replaceable", peer = peerId.shortLog()
     return
-
-proc checkAndEvictPeer(
-    kad: KadDHT, rtable: RoutingTable, peerId: PeerId
-) {.async: (raises: [CancelledError]).} =
-  withProbeSlotOrReturn(kad)
 
   # Candidate list is a snapshot; by the time a slot is free the peer may have
   # been refreshed (markUseful) or removed, and stop may have begun.
@@ -221,6 +227,40 @@ proc maintainLiveness(kad: KadDHT) {.async: (raises: [CancelledError]).} =
       idle = false
     else:
       idle = true
+
+proc maintainLiveness(kad: KadDHT) {.async: (raises: [CancelledError]).} =
+  ## Continuous background task: drain replaceable peers via liveness probes
+  ## as soon as they become due, bounded by ``probeSem``. Independent of
+  ## bucket refresh so dead peers are not held until the next refresh tick.
+  # Yield once so start() can finish bootstrap before the first scan.
+  await sleepAsync(kad.config.livenessIdleInterval)
+
+  while not kad.stopping:
+    let grace = kad.config.livenessGracePeriod
+    var saturated = false
+
+    for rtable in kad.maintainableTables():
+      if kad.stopping:
+        return
+      for peerId in rtable.livenessCandidates(grace):
+        if not kad.tryLaunchLivenessProbe(rtable, peerId):
+          saturated = true
+          break
+      if saturated:
+        break
+
+    if kad.livenessProbes.len > 0:
+      let inFlight = kad.livenessProbes.values.toSeq()
+      try:
+        discard await one(inFlight)
+      except ValueError:
+        # All futures already finished between the snapshot and the wait.
+        discard
+      except CancelledError as exc:
+        await noCancel inFlight.cancelAndWait()
+        raise exc
+    else:
+      await sleepAsync(kad.config.livenessIdleInterval)
 
 proc refreshTable*(
     kad: KadDHT, rtable: RoutingTable, forceRefresh = false
--- a/libp2p/protocols/kademlia.nim
+++ b/libp2p/protocols/kademlia.nim
@@ -68,17 +68,20 @@ proc checkAndEvictPeer(
   kad.livenessProbes[probeKey] = probe
   probe.addCallback(
     proc(udata: pointer) {.gcsafe, raises: [].} =
-      if kad.livenessProbes.getOrDefault(probeKey) == probe:
-        kad.livenessProbes.del(probeKey)
+      if kad.livenessProbes.getOrDefault(peerId) == probe:
+        kad.livenessProbes.del(peerId)
   )
 
 proc checkAndEvictPeer(
-    kad: KadDHT, rtable: RoutingTable, peerId: PeerId
+    kad: KadDHT, peerId: PeerId
 ) {.async: (raises: [CancelledError]).} =
-  ## Returns immediately when all liveness probe slots are occupied.
+  ## Probes a peer once and applies the result to every maintainable table
+  ## where that peer is past the liveness grace period. Returns immediately
+  ## when all liveness probe slots are occupied.
   if kad.stopping:
     return
-  if not kad.probeSem.tryAcquire():
+  if not kad.livenessSem.tryAcquire():
+    debug "Liveness probe skipped: no free slot", peer = peerId.shortLog()
     kad_routing_table_liveness_probes.inc(labelValues = ["skipped"])
     return
   defer:
@@ -105,7 +108,13 @@ proc checkAndEvictPeer(
   if kad.stopping:
     return
   let grace = kad.config.livenessGracePeriod
-  if not rtable.isReplaceable(peerId, grace, Moment.now()):
+  var dueTables: seq[RoutingTable]
+  for rtable in kad.maintainableTables():
+    if rtable.isReplaceable(peerId, grace, Moment.now()):
+      dueTables.add(rtable)
+  if dueTables.len == 0:
+    debug "Liveness probe skipped: peer no longer replaceable",
+      peer = peerId.shortLog()
     return
 
   let addrs = kad.switch.peerStore[AddressBook][peerId]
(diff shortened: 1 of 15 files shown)


SionoiS:
cc @jm-clius @tinniaru3005

libp2p/protocols/kademlia/peer_registry.nim
```diff
@@ -0,0 +1,102 @@
...
+  let record = PeerRecord(
+    nodeId: nodeId, lastSeen: now, addedAt: now, lastUsefulAt: Opt.none(Moment)
+  )
+  registry.peers[nodeId] = record
+  record
+
+proc touch*(registry: PeerRegistry, nodeId: Key, now = Moment.now()) =
+  ## Bump ``lastSeen`` when the peer is observed again. No-op if missing.
+  registry.peers.withValue(nodeId, record):
+    record[].lastSeen = now
```
gmelodie: I think these could be colapsed as:

```suggestion
proc upsert*(registry: PeerRegistry, nodeId: Key, now = Moment.now()) =
    ## Create the row, or refresh ``lastSeen`` when it already exists.
    registry.peers.withValue(nodeId, record):
      record[].lastSeen = now
      return

    registry.peers[nodeId] = PeerRecord(
      nodeId: nodeId, lastSeen: now, addedAt: now, lastUsefulAt:
  Opt.none(Moment)
    )
```

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -71,76 +83,85 @@ func nPeersForCpl*(rtable: RoutingTable, cpl: int): int =
...
-    count += bucket.peers.countIt(rtable.commonPrefixLen(it.nodeId) == cpl)
+    count += bucket.peers.countIt(rtable.commonPrefixLen(it) == cpl)
   count
 
 proc peerIndexInBucket(bucket: Bucket, nodeId: Key): Opt[int] =
   for i, p in bucket.peers:
-    if p.nodeId == nodeId:
+    if p == nodeId:
       return Opt.some(i)
   return Opt.none(int)
```
gmelodie: Now we can do this instead:
```nim
func peerIndexInBucket(bucket: Bucket, nodeId: Key): Opt[int] =
  let i = bucket.peers.find(nodeId)
  if i < 0:
    return Opt.none(int)
  Opt.some(i)
```

libp2p/protocols/kademlia/types.nim
```diff
@@ -217,16 +217,27 @@ proc xorDistance*(a, b: PeerId, hasher: Opt[XorDHasher]): XorDistance =
...
+  PeerRegistry* = ref object
+    peers*: Table[Key, PeerRecord]
+    tablesByPeer*: Table[Key, HashSet[Key]]
+      ## peer key → set of routing-table ``selfId``s that reference the peer.
+
+  ## A k-bucket is an index partition: keys only, ordered for eviction by
+  ## looking up ``PeerRecord`` timestamps in the shared registry.
   Bucket* = object
-    peers*: seq[NodeEntry]
+    peers*: seq[Key]
```
gmelodie, line 245: Not necessarily related to this pr, but either now or as a follow-up, we could get rid of the public peers field
```suggestion
    peers: seq[Key]
```
and implement these instead:
```nim
func len*(bucket: Bucket): int =
  bucket.peers.len

func contains*(bucket: Bucket, nodeId: Key): bool =
  nodeId in bucket.peers

iterator items*(bucket: Bucket): Key =
  for nodeId in bucket.peers:
    yield nodeId
```

libp2p/protocols/kademlia/types.nim
```diff
@@ -217,16 +217,27 @@ proc xorDistance*(a, b: PeerId, hasher: Opt[XorDHasher]): XorDistance =
...
     lastUsefulAt*: Opt[Moment]
       ## Set when the peer answers a query; `Opt.none` until then, which makes it
-      ## replaceable once past the grace period since it was added.
+      ## replaceable once past the grace period since it was first recorded.
 
+  ## Shared store of peer rows plus reverse membership (which table selfIds
+  ## index each peer). Owned by ``KadDHT``; every ``RoutingTable`` shares it.
+  PeerRegistry* = ref object
+    peers*: Table[Key, PeerRecord]
+    tablesByPeer*: Table[Key, HashSet[Key]]
```
gmelodie: We could add two native operators and get rid of these public fields instead:

```suggestion
    peers: Table[Key, PeerRecord]
    tablesByPeer: Table[Key, HashSet[Key]]
```
adding
```nim
iterator items*(registry: PeerRegistry): PeerRecord =
  for record in registry.peers.values:
    yield record

iterator pairs*(registry: PeerRegistry): (Key, PeerRecord) =
  for nodeId, record in registry.peers:
    yield (nodeId, record)
```
And a helper
```nim
template withRecord*(registry: PeerRegistry, nodeId: Key, record, body: untyped) =
  ## Mutable access to one row. ``record`` is a ``ptr PeerRecord``; body uses ``record[]``.
  registry.peers.withValue(nodeId, record):
    body
```
Call sites would then be like (e.g. on tests)
```nim
proc makeBucketStale*(rtable: RoutingTable, bucketIdx: int) =
  let past = Moment.now() - (DefaultBucketStaleTime + 1.minutes)
  for nodeId in rtable.buckets[bucketIdx].peers:
    rtable.registry.withRecord(nodeId, record):
      record[].lastSeen = past
```

libp2p/protocols/kademlia/routing_table.nim
gmelodie: While we're at it, we might want to do this instead
```nim
proc findClosestPeerIds*(rtable: RoutingTable, targetId: Key, count: int): seq[PeerId] =
  findClosest(rtable, targetId, count).toPeerIds()
```

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -301,21 +323,25 @@ proc randomPeersClosestFirstPeerIds*(
     .filterIt(it.isOk)
     .mapIt(it.value())
```
gmelodie: While we're at it, we might want to do this instead
```nim
proc randomPeersClosestFirstPeerIds*(rtable: RoutingTable, rng: Rng, count: int, maxPerBucket = high(int)): seq[PeerId] =
  randomPeersClosestFirst(rtable, rng, count, maxPerBucket).toPeerIds()
```

libp2p/protocols/kademlia/peer_registry.nim
```diff
@@ -0,0 +1,102 @@
...
+  registry.peers.len
+
+func lastActivity*(record: PeerRecord): Moment =
+  ## Most recent time the peer answered a DHT query, or when it was first recorded.
+  record.lastUsefulAt.get(record.addedAt)
+
+func isReplaceable*(record: PeerRecord, gracePeriod: Duration, now: Moment): bool =
+  ## Replaceable once past ``gracePeriod`` without proving useful. Grace is global
+  ## (from first registry sighting), not per routing-table membership.
+  now - record.lastActivity() > gracePeriod
```
gmelodie: Because now grace periods are inherited, the following can occur (claude-created example):

> Inserting P into table T stamped addedAt = now for that table, so P got a fresh hour in > every table it joined. Grace was per membership.
>
> Now there is one row, stamped at first sighting anywhere:
>
> - 09:00 — P joins the main table. Row created, addedAt = 09:00.
> - 10:00 — P still hasn't answered anything, so it becomes replaceable in main. Correct.
> - 11:00 — you register interest in service S. addService seeds every main-table key into the new S table, P included. insert calls ensure(P), which finds the row and leaves addedAt at 09:00.
> - 11:00 and a millisecond — S needs to evict. now - lastActivity() is 2 hours, over the 1-hour grace. P is replaceable in a table it joined a millisecond ago.

Fix is the same as for the `markUseful` problem: `addedAt` belongs on the membership record, not shared.

libp2p/protocols/kademlia/peer_registry.nim
```diff
@@ -0,0 +1,102 @@
...
+    if tables[].len > 0:
+      return false
+    registry.tablesByPeer.del(nodeId)
+    registry.peers.del(nodeId)
+    return true
+  # Membership missing but a row may still exist (failed insert path).
+  if nodeId in registry.peers and nodeId notin registry.tablesByPeer:
+    registry.peers.del(nodeId)
+    return true
+  false
```
gmelodie: By changing `tablesByPeer` to `Table[Key, Table[Key, Membership]]` we could simplify this:

```suggestion
 proc removeMembership*(registry: PeerRegistry, nodeId: Key, tableId: Key) =
  ## Drop one table's reference, and the peer row along with the last one.
  registry.tablesByPeer.withValue(nodeId, tables):
    tables[].excl(tableId)
    if tables[].len > 0:
      return

  registry.tablesByPeer.del(nodeId)
  registry.peers.del(nodeId)
```

note: no need to return `bool` since no call site is currently using it.

gmelodie (commented):
`markUseful` being shared across tables means that one answered main-table `FIND_NODE` refreshes `lastUsefulAt` for every service table indexing that peer.


Interesting finds from Claude:

>  `libp2p/protocols/service_discovery/routing_table_manager.nim:77` `manager.tables.del(serviceId)` drops the table object and never drops that
 table's `selfId` from `registry.tablesByPeer`.
> 
> Fix: add `proc dropTable*(registry: PeerRegistry, tableId: Key)` to `peer_registry.nim` that walks `tablesByPeer`, excludes `tableId`, and deletes rows whose set empties. Call it from `removeService` before `manager.tables.del`, using the removed table's `selfId`.

and

> Four functions give four different answers to **“the bucket holds a key with no registry row”**:
>
> * **`oldestPeer`** (`routing_table.nim:99, 103`): treats it as `Moment.now()`, so it is **newest** and is never evicted.
> * **`replaceableCandidate`** (`routing_table.nim:137`): `continue`, so it is **never replaceable**.
> * **`isStale`** (`routing_table.nim:335`): returns `true`, so it is **always stale**.
> * **`isReplaceable`** (`peer_registry.nim:95`): returns `false`.

libp2p/protocols/service_discovery/routing_table_manager.nim
```diff
@@ -45,19 +45,19 @@ proc addService*(
...
+    registry = mainRoutingTable.registry,
   )
 
-  # Seed from main table
-  for bucket in mainRoutingTable.buckets:
-    for peer in bucket.peers:
-      discard rtable.insert(peer.nodeId)
+  # Seed by referencing peers already in the main index (no peer-row copies).
+  for nodeId in mainRoutingTable.allKeys():
+    discard rtable.insert(nodeId)
```
richard-ramos, line 60: Probably makes sense to have a helper function that does:
```
for nodeId in table.allKeys():
  discard table.registry.removeMembership(nodeId, table.selfId)
manager.tables.del(serviceId)
```

And have this function be called both by `removeService()` and by `ServiceRoutingTableManager.clear()`

tests/libp2p/kademlia/test_bootstrap.nim
```diff
@@ -236,3 +236,88 @@ suite "KadDHT Bootstrap Component":
...
+    ## An in-flight entry must be awaited by a concurrent batch instead of
+    ## launching a second probe (and acquiring another livenessSem slot).
+    let kad = setupMockKad()
+    startAndDeferStop(@[kad])
+
+    let peer = randomPeerId()
+    check kad.rtable.insert(peer)
+    agePeerPastLivenessGrace(kad.rtable, peer.toKey())
+
+    let hang = newFuture[void]()
```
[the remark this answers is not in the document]
richard-ramos: Good catch

richard-ramos (approved):

libp2p/protocols/kademlia.nim
```diff
@@ -7,29 +7,32 @@ import ../utils/[heartbeat, future]
...
     rtable: RoutingTable, gracePeriod: Duration
 ): seq[PeerId] {.raises: [].} =
   ## Peers past the liveness grace period that should be probed.
   let now = Moment.now()
   var peers: seq[PeerId]
   for bucket in rtable.buckets:
-    for entry in bucket.peers:
-      if not entry.isReplaceable(gracePeriod, now):
+    for nodeId in bucket.peers:
+      if not rtable.isReplaceable(nodeId, gracePeriod, now):
```
gmelodie: No need to have isRepleceable recompute `bucketIndex(nodeId)` and re-run `peerIndexInBucket` over the bucket we're already looking at
```suggestion
      if not rtable.registry.isReplaceable(nodeId, rtable.selfId, gracePeriod, now):
```

libp2p/protocols/kademlia.nim
gmelodie: we can use `dueTables` here
```suggestion
  for rtable in dueTables:
```

libp2p/protocols/kademlia/peer_registry.nim
```diff
@@ -0,0 +1,140 @@
...
+    registry.peers.del(nodeId)
+
+proc dropPeer*(registry: PeerRegistry, nodeId: Key) =
+  ## Remove the peer row and all membership tracking. Callers must also drop
+  ## the key from every routing-table bucket that still holds it.
+  registry.peers.del(nodeId)
+  registry.tablesByPeer.del(nodeId)
+
+func tableIds*(registry: PeerRegistry, nodeId: Key): HashSet[Key] =
+  result = initHashSet[Key]()
```
gmelodie: Don't use `result`

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -218,16 +268,10 @@ proc contains*(rtable: RoutingTable, nodeId: Key): bool =
...
-    return
-  let pos = peerIndexInBucket(rtable.buckets[idx], nodeId).valueOr:
-    return
-  let now = Moment.now()
-  rtable.buckets[idx].peers[pos].lastUsefulAt = Opt.some(now)
-  rtable.buckets[idx].peers[pos].lastSeen = now
+  ## Records that ``nodeId`` answered a query. Updates the shared registry row
+  ## so every index that references the peer sees the new usefulness.
+  ## No-op when the peer has no registry row (not in any table).
+  rtable.registry.markUseful(nodeId)
```
gmelodie, line 276:
```suggestion
    ## Records that ``nodeId`` answered a query. Updates the shared registry row
    ## so every index that references the peer sees the new usefulness. No-op when
    ## the table is detached or the peer has no registry row (not in any table).
    if rtable.detached:
      return
    rtable.registry.markUseful(nodeId)
```

gmelodie (approved):

libp2p/protocols/kademlia/peer_registry.nim
SionoiS: My assistant like using result and I always miss those...

libp2p/protocols/kademlia/peer_registry.nim
```diff
@@ -0,0 +1,140 @@
...
+## Usefulness grace is membership-local via ``Membership.addedAt``.
+
+import std/[tables, sets]
+import chronos, results
+import ./types
+import ../../peerid
+
+{.push raises: [].}
+
+proc new*(T: typedesc[PeerRegistry]): PeerRegistry =
```
vladopajic, line 15: would be nice to add test coverage for `PeerRegistry` type. it can be easily tested

libp2p/protocols/service_discovery/routing_table_manager.nim
```diff
@@ -164,6 +167,8 @@ proc serviceIds*(manager: ServiceRoutingTableManager): seq[ServiceId] =
   return manager.tables.keys.toSeq()
 
 proc clear*(manager: ServiceRoutingTableManager) =
+  for table in manager.tables.values:
+    table.detachAll()
```
vladopajic, line 171: `manager.tabels.values.alliIt(it.detachAll())`

one liner nit

vladopajic (approved):
nice effort!

libp2p/protocols/kademlia/peer_registry.nim
SionoiS: I'm not following. What do you want tests about? The procs in `kademlia/types.nim`?

vladopajic: `PeerRegistry` has added a lot of new functions in this pr, upser, markUseful, etc... non of this code is tested in isolation as unit tests for this type. ideally there should be `test_peer_registry` with test coverage for this type.

SionoiS: Ah! I thought we had those. My bad.

PR #2919, opened 2026-08-07
feat(cbind): expose the GossipSub ingress limits in Libp2pConfig

gmelodie:
## Summary

`cbind` mounted GossipSub with the defaults, so a C host could not bound what an untrusted peer sends to it. `overheadRateLimit` stayed `Opt.none`, `disconnectPeerAboveRateLimit` stayed `false`, and `maxMessageSize` stayed at the 1 MiB pubsub default.

Part 1 bounds the per-topic queue of the module. This PR is Part 2. It bounds the ingress rate on the same path.

`Libp2pConfig` now holds one `gossipsub` sub-object instead of flat `gossipsub*` fields.

| field | meaning |
| --- | --- |
| `gossipsub.mount` | Mount GossipSub. Was `mountGossipsub`. |
| `gossipsub.triggerSelf` | Deliver locally published messages to self. Was `gossipsubTriggerSelf`. |
| `gossipsub.maxMessageSize` | Largest message accepted or sent, in bytes. 0 uses the core default. The ceiling is `MAX_GOSSIPSUB_MESSAGE_SIZE` (64 MiB). |
| `gossipsub.overheadRateLimit.bytes` | Per-peer budget of protocol-overhead bytes per interval. 0 disables the limit. |
| `gossipsub.overheadRateLimit.intervalMs` | Refill interval of that budget, up to `MAX_OVERHEAD_RATE_LIMIT_INTERVAL_MS` (1 hour). |
| `gossipsub.disconnectPeerAboveRateLimit` | Disconnect a peer that spends its budget. |

Read the last row with care. An empty bucket only increments `libp2p_gossipsub_peers_rate_limit_hits` and logs at `debug` (`gossipsub.nim:617-627`, `scoring.nim:346-357`). It does not lower the score of the peer. The budget stays a measurement until `disconnectPeerAboveRateLimit` makes it an enforcement. The field docs say so.

`maxMessageSize` needs no such caveat. It bounds `readLp` on the receive path (`pubsubpeer.nim:315`) and the send path (`pubsubpeer.nim:585`).

`parse` validates every value once, and a limit fails closed.

- A negative value on any numeric field is an error, not a silent default. A host with an arithmetic underflow must not ship a node with the defense off. This differs from the `maxConnections` convention nearby, on purpose.
- The bytes and the interval go together, or neither.
- `disconnectPeerAboveRateLimit` without a budget is an error, because it never fires.
- Any limit without `gossipsub.mount` is an error, because nothing applies it.
- The interval stops at one hour. `chronos.milliseconds` multiplies by a million with no overflow guard (`chronos/timer.nim:273-275`), and `buildffi` compiles with overflow checks on. An unbounded interval raises an `OverflowDefect` that the `except LPError` of `libp2pNew` cannot catch.
- Both ceilings are `{.ffiConst.}`. A host reads `MAX_GOSSIPSUB_MESSAGE_SIZE` and `MAX_OVERHEAD_RATE_LIMIT_INTERVAL_MS` from `libp2p.h` instead of an error string.

## Affected Areas

- [x] Gossipsub
  `GossipSubParams.overheadRateLimit` changes from `Opt[tuple[bytes: int, interval: Duration]]` to `Opt[RateLimit]`, where `RateLimit` is a new exported object with the same two fields. The named type stops a silent field swap. No scoring, validation or peer-handling logic changes. `pubsub.nim` gains one exported const, `DefaultPubSubMaxMessageSize`, which replaces the `1024 * 1024` literal in `PubSub.init` so cbind stops the copy.

- [ ] Transports

- [ ] Protocol Logic

- [ ] Build / Tooling

- [x] Other
  C bindings (`cbind/`): the new `GossipsubConfig` and `RateLimitConfig` FFI types, the `ParsedGossipsub` parse result, the hand-maintained `libp2p_config` C mirror, and the `gossipsub.c` example.

- **Nimbus:** check needed. The project breaks only if it sets `overheadRateLimit`.
- **Waku:** check needed. Same condition.
- **Codex:** check needed. Same condition.

The `cbind/` changes reach none of these projects.

## Impact on Library Users

Nim API, breaking: a caller that writes `overheadRateLimit = Opt.some((bytes, interval))` must write `overheadRateLimit = Opt.some(RateLimit(bytes: bytes, interval: interval))`. Field access and behaviour stay the same. `DefaultPubSubMaxMessageSize` is new and exported. No default moves.

C hosts, breaking: the layout of `Libp2pConfig` changes, and `mountGossipsub` and `gossipsubTriggerSelf` move to `gossipsub.mount` and `gossipsub.triggerSelf`. A host must regenerate `c_bindings/libp2p.h` and rebuild. A zero-filled config still behaves as it does today, with no rate limit and 1 MiB messages.

## Risk Assessment

Backward compatibility: two breaking renames, one in the Nim API and one in the C config layout. Both fail at compile time, so neither fails silently.

Network behaviour: unchanged until a host opts in. With `disconnectPeerAboveRateLimit` set, the node disconnects a peer over budget, so a budget that is too tight drops honest peers. That flag needs an explicit budget, which keeps the choice deliberate.

Performance: one token bucket per peer, and only when the host sets a budget.

Security: a host can now bound the size of a single inbound message, and it can measure or enforce the per-peer protocol overhead. Two judgment calls are mine and need your review. The 64 MiB message ceiling matches the `MaxReadBytes` cap on stream reads. The one-hour interval ceiling keeps `chronos.milliseconds` from an overflow.

## References

Part 1 of this work bounds the per-topic queue of the module.

## Additional Notes

Verification: `nimble genbindings_c` and `genbindings_cddl` compile the whole binding surface and carry the nested types and the two constants. `nimble examples` built and ran all six examples, and `gossipsub` delivered its message with the limits active. A throwaway C probe confirmed nine cases: six rejected combinations return their intended error, and two valid configs start. `test_gossipsub` (31 tests) and `test_gossipsub_scoring` (11 tests) pass with the `RateLimit` object.

Known gap: no automated test covers the validation rules, because `cbind/` has no test target. `cbindings.yml` builds and runs the examples only. A probe outside the repo covers the six error paths. A cbind test target is worth the work, and it stays out of this PR on purpose.

The `gossipsub.c` example sets a 1 MiB message cap and 64 KiB of overhead per second per peer. It leaves enforcement off so a slow CI runner cannot drop its own peer.

### DIFF
--- a/cbind/libp2p/config.nim
+++ b/cbind/libp2p/config.nim
@@ -172,7 +172,7 @@ const
     ## unbounded interval overflows `Duration` and raises a Defect no caller
     ## catches. An hour is already a quota rather than a rate limit.
 
-func requireGossipsubForLimits(config: Libp2pConfig): Result[void, string] =
+func requireGossipsubMount(config: Libp2pConfig): Result[void, string] =
   if config.mountGossipsub:
     return ok()
   if config.gossipsubMaxMessageSize != 0 or config.gossipsubOverheadRateLimitBytes != 0 or
@@ -181,7 +181,7 @@ func requireGossipsubForLimits(config: Libp2pConfig): Result[void, string] =
     return err("gossipsub limits require mountGossipsub")
   ok()
 
-func parseGossipsubMaxMessageSize(config: Libp2pConfig): Result[int, string] =
+func parseMaxMessageSize(config: Libp2pConfig): Result[int, string] =
   if config.gossipsubMaxMessageSize < 0:
     return err("gossipsubMaxMessageSize must be 0 or greater")
   if config.gossipsubMaxMessageSize == 0:
@@ -192,7 +192,7 @@ func parseGossipsubMaxMessageSize(config: Libp2pConfig): Result[int, string] =
     )
   ok(config.gossipsubMaxMessageSize)
 
-func validateOverheadRateLimitRange(config: Libp2pConfig): Result[void, string] =
+func checkRateLimitRange(config: Libp2pConfig): Result[void, string] =
   if config.gossipsubOverheadRateLimitBytes < 0:
     return err("gossipsubOverheadRateLimitBytes must be 0 or greater")
   if config.gossipsubOverheadRateLimitIntervalMs < 0:
@@ -204,12 +204,12 @@ func validateOverheadRateLimitRange(config: Libp2pConfig): Result[void, string] 
     )
   ok()
 
-func parseGossipsubOverheadRateLimit(
+func parseOverheadRateLimit(
     config: Libp2pConfig
 ): Result[Opt[OverheadRateLimit], string] =
   ## The per-peer token bucket bounding protocol overhead an untrusted peer can
   ## push at us: `bytes` per `interval`. Both halves are needed, or neither.
-  ?validateOverheadRateLimitRange(config)
+  ?checkRateLimitRange(config)
   if config.gossipsubOverheadRateLimitBytes == 0:
     if config.gossipsubOverheadRateLimitIntervalMs > 0:
       return err(
@@ -233,13 +233,13 @@ func parseGossipsubOverheadRateLimit(
     )
   )
 
-func parseGossipsubConfig(config: Libp2pConfig): Result[ParsedGossipsub, string] =
-  ?requireGossipsubForLimits(config)
+func parseGossipsub(config: Libp2pConfig): Result[ParsedGossipsub, string] =
+  ?requireGossipsubMount(config)
   ok(
     ParsedGossipsub(
       triggerSelf: config.gossipsubTriggerSelf,
-      maxMessageSize: ?parseGossipsubMaxMessageSize(config),
-      overheadRateLimit: ?parseGossipsubOverheadRateLimit(config),
+      maxMessageSize: ?parseMaxMessageSize(config),
+      overheadRateLimit: ?parseOverheadRateLimit(config),
       disconnectPeerAboveRateLimit: config.gossipsubDisconnectPeerAboveRateLimit,
     )
   )
@@ -422,7 +422,7 @@ proc parse(config: Libp2pConfig): Result[ParsedConfig, string] =
       nat: ?parseNATConfig(config),
       autonatV2Server: config.autonatV2Server,
       mountGossipsub: config.mountGossipsub,
-      gossipsub: ?parseGossipsubConfig(config),
+      gossipsub: ?parseGossipsub(config),
       mountKad: config.mountKad,
       mountServiceDiscovery: config.mountServiceDiscovery,
     )
--- a/cbind/libp2p/config.nim
+++ b/cbind/libp2p/config.nim
@@ -28,25 +28,22 @@ type BootstrapNode {.ffi.} = object
   peerId: string
   multiaddrs: seq[string]
 
+type RateLimitConfig {.ffi.} = object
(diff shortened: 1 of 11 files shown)


cbind/libp2p/config.nim
```diff
@@ -138,6 +155,69 @@ proc parsePrivateKey(raw: seq[byte]): Result[Opt[PrivateKey], string] =
...
+  let limit = cfg.overheadRateLimit
+  ?checkRange(limit)
+  if limit.bytes == 0:
+    if limit.intervalMs > 0:
+      return err("overheadRateLimit.intervalMs requires overheadRateLimit.bytes")
+    if cfg.disconnectPeerAboveRateLimit:
+      return err("disconnectPeerAboveRateLimit requires overheadRateLimit.bytes")
+    return ok(Opt.none(RateLimit))
+  if limit.intervalMs == 0:
+    return err("overheadRateLimit.bytes requires overheadRateLimit.intervalMs")
```
vladopajic: should be one `?validateLimit(imit)` func call, because now sanity checks are done in two places

vladopajic (approved):

libp2p/protocols/pubsub/gossipsub/types.nim
```diff
@@ -153,7 +157,7 @@ type
     bandwidthEstimatebps*: int
       # This is currently used only for limting flood publishing. 0 disables flood-limiting completely
 
-    overheadRateLimit*: Opt[tuple[bytes: int, interval: Duration]]
+    overheadRateLimit*: Opt[RateLimit]
```
richard-ramos, line 160: This is a breaking change. Is this something used downstream?

cbind/libp2p/config.nim
```diff
@@ -138,6 +155,72 @@ proc parsePrivateKey(raw: seq[byte]): Result[Opt[PrivateKey], string] =
...
+func parseMaxMessageSize(size: int): Result[int, string] =
+  if size < 0:
+    return err("maxMessageSize must be 0 or greater")
+  if size == 0:
+    return ok(DefaultPubSubMaxMessageSize)
+  if size > MaxGossipsubMessageSize:
+    return err("maxMessageSize must be at most " & $MaxGossipsubMessageSize & " bytes")
+  ok(size)
+
+func validateLimit(cfg: GossipsubConfig): Result[void, string] =
```
richard-ramos: Shouldnt these validations go in https://github.com/vacp2p/nim-libp2p/blob/e9dcbf6d07d35c1ecafc95a7bc3f61b65a828a86/libp2p/protocols/pubsub/gossipsub.nim#L165 ?

cbind/libp2p/config.nim
```diff
@@ -138,6 +155,72 @@ proc parsePrivateKey(raw: seq[byte]): Result[Opt[PrivateKey], string] =
     return err("invalid private key: " & $error)
   ok(Opt.some(key))
 
+const
+  MaxGossipsubMessageSize {.ffiConst.} = 64 * 1024 * 1024
+    ## A peer can make us preallocate this much before a byte is validated.
+  MaxOverheadRateLimitIntervalMs {.ffiConst.} = 60 * 60 * 1000
+    ## Above an hour `chronos.milliseconds` overflows `Duration` and raises a Defect.
```
richard-ramos: What defect are you getting here? This does not seem correct, as a Duration uses nanoseconds, so the max value you can put here is `high(int64) div 1_000_000'i64`

richard-ramos (approved):

PR #2922, opened 2026-08-10
fix(service-disco): validate advert length and re-check client mode

gmelodie:
## Summary

Applies the review comments left on https://github.com/vacp2p/nim-libp2p/pull/2917 after it merged.

`validateAdvert` decoded the caller's bytes before it measured them. The protobuf decoder skips unknown fields, and `isValid()` measures the record after a re-encode, which drops those fields again. A caller could therefore pass the FFI a valid signed record plus one large unknown padding field, clear validation, and have the node store it in `providedAdverts` and republish it to every registrar on every rotation. The length is now checked on the bytes as they arrive, before the decode, as `isValidAdvertisement` already does in `registrar.nim`.

`addProvidedService` returned `ok()` when the service was already provided, while `providedAdverts` kept the previous bytes. A caller that wanted to publish a fresh record silently kept advertising the old one. It now returns an error, so the caller stops the service and starts it again.

`advertiseToRegistrar` read `isServer` once, before its retry loop, so a task already inside the loop kept registering after `changeMode` downgraded the node. The mode is now read on every iteration. `maintainRegistrations` and `startLocalRegistration` get the same guard, and both resume when the node returns to server mode.

The last review comment, about partially changed state on an error return, does need a change after all. `addProvidedService` writes to `disco.services` and to the routing table manager, and the two error paths that follow those writes used to leave both in place. The new "already advertised" error turns that into a dead end, because the retry then fails for the wrong reason. `undoProvidedService` now reverts both writes on each of those paths, so a caller retries after it fixes the cause. `cd_advertiser_services_added` moved to the success path, so it counts only the services that the node really added.

Three cleanups came out of the same procs: `advertFor` replaces the "stored advert, else this node's record" expression that `startLocalRegistration` and `maintainRegistrations` each carried, `scheduleRegistrations` holds the per-bucket fan-out that `addProvidedService` inlined, and `trackAdvertiseTask` holds the future, the task record, the set insert and the metric that `scheduleRegistrations` and `maintainRegistrations` each repeated.

## Affected Areas

  `validateAdvert`, `addProvidedService`, `advertiseToRegistrar`, `maintainRegistrations` and `startLocalRegistration` in `libp2p/protocols/service_discovery/advertiser.nim`, plus the four new private procs.

- [x] Build / Tooling
  A doc comment on `libp2pServiceDiscoStartAdvertising` in `cbind/libp2p.nim`. No generated header or CDDL schema carries this text, so nothing needs a regeneration.

- **Nimbus:** N/A — does not consume service discovery.
- **Waku:** N/A — same reason.
- **Codex:** N/A — same reason.

## Impact on Library Users

`addProvidedService` and `startAdvertising` now fail on a service that is already advertised, where they used to report success. A caller that refreshes its record calls `stopAdvertising` first. C binding users meet the same rule through the error string of `libp2pServiceDiscoStartAdvertising`; the request type is unchanged.

An advertisement larger than `MaxXPRSize` on the wire is rejected where it used to be accepted. A record that fits the limit is unaffected, because the registrar applied the same bound on receipt already.

The rejection strings changed. An oversized advertisement now reports its own size next to the limit, and a record that fails `isValid()` after the decode reads "invalid advertisement" instead of the same "oversized" text. No caller matches on these strings in this repository.

## Risk Assessment

- Backward compatibility: the FFI surface is unchanged. The new "already advertised" error is the one behaviour change a caller observes.
- Network behaviour: a node that flips to client mode stops advertising within one `advertExpiry`, and resumes at the next maintenance tick after it flips back. Server-mode behaviour is unchanged.
- Security: the length check closes the path where one FFI call made the node rebroadcast unbounded caller-supplied bytes on every rotation.
- Error paths: a failed `addProvidedService` now leaves the node in the state it held before the call, so a retry behaves like a first call.
- Performance: one length compare per `startAdvertising`, and one boolean read per registration loop iteration.

## References

- Review comments: https://github.com/vacp2p/nim-libp2p/pull/2917#discussion_r3750728187, https://github.com/vacp2p/nim-libp2p/pull/2917#discussion_r3750773019, https://github.com/vacp2p/nim-libp2p/pull/2917#discussion_r3750798765, https://github.com/vacp2p/nim-libp2p/pull/2917#discussion_r3750806744
- Prior PR: https://github.com/vacp2p/nim-libp2p/pull/2917
- Issue: https://github.com/logos-co/logos-libp2p-module/issues/98

## Additional Notes

New tests: a record padded past `MaxXPRSize` with an unknown field, which still decodes, so it proves why the raw length check is needed; a second advertisement that needs a stop before a restart; a failed record build that leaves no half-added service behind; `maintainRegistrations` in both mode directions; and an end-to-end downgrade that ends the local loop and the running tasks.

The `MaxXPRSize` bound now sits in two validators, `registrar.isValidAdvertisement` and `advertiser.validateAdvert`. Merging them is worth doing, but the two run different follow-up checks, so a shared validator would change how strict the receive path is. That belongs in its own PR.

Neither validator checks that the advertised record belongs to this node. A caller can hand the FFI any validly signed extended peer record and the node registers it at every registrar, and the receive path accepts a record from a peer that does not own it. The rule has to be decided for both sides at once, so it is not in this PR.

### DIFF
--- a/libp2p/protocols/service_discovery/advertiser.nim
+++ b/libp2p/protocols/service_discovery/advertiser.nim
@@ -49,6 +49,13 @@ proc getAdvertBytes(disco: ServiceDiscovery, explicit: Opt[seq[byte]]): Opt[seq[
     return Opt.none(seq[byte])
   Opt.some(extRecord.encode())
 
+proc advertFor(disco: ServiceDiscovery, serviceId: ServiceId): seq[byte] =
+  ## The caller's own bytes when it supplied them, this node's record otherwise.
+  if serviceId in disco.advertiser.providedAdverts:
+    return disco.advertiser.providedAdverts.getOrDefault(serviceId)
+
+  disco.getAdvertBytes(Opt.none(seq[byte])).get(@[])
+
 proc advertiseToRegistrar*(
   disco: ServiceDiscovery,
   serviceId: ServiceId,
@@ -60,6 +67,10 @@ proc advertiseToRegistrar*(
 proc startLocalRegistration(disco: ServiceDiscovery) =
   ## Starts (or restarts) the single long-lived local self-registration task.
 
+  if not disco.isServer:
+    trace "not registering locally while in client mode"
+    return
+
   if not disco.localRegistrationLoop.isNil and not disco.localRegistrationLoop.finished:
     return
   if disco.services.len == 0:
@@ -68,12 +79,7 @@ proc startLocalRegistration(disco: ServiceDiscovery) =
   let someService = disco.services.toSeq()[0]
   let sid = someService.id.hashServiceId()
 
-  let advertBytes = (
-    if sid in disco.advertiser.providedAdverts:
-      disco.advertiser.providedAdverts.getOrDefault(sid)
-    else:
-      disco.getAdvertBytes(Opt.none(seq[byte])).get(@[])
-  )
+  let advertBytes = disco.advertFor(sid)
   if advertBytes.len == 0:
     return
 
@@ -101,6 +107,10 @@ proc maintainRegistrations*(
   ## maintenance loop can rotate the slot to another peer in the bucket.
 
   cleanupFinishedTasks(disco.advertiser)
+
+  if not disco.isServer:
+    trace "no registration maintenance while in client mode"
+    return
 
   let selfPeer = disco.switch.peerInfo.peerId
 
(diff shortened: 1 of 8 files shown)


vladopajic (approved):

libp2p/protocols/service_discovery/advertiser.nim
```diff
@@ -297,20 +313,57 @@ proc advertiseToRegistrar*(
...
+      continue
+
+    for peer in peers:
+      let registrar = peer.nodeId.toPeerId().valueOr:
+        error "cannot convert key to peer id", error
+        continue
+
+      disco.trackAdvertiseTask(serviceId, registrar, bucketIdx, advertBytes)
+
+proc undoProvidedService(
```
richard-ramos, line 358: Should this also cancel and remove the bootstrap future stored in `serviceBootstrapFuts`?

richard-ramos (approved):

PR #2924, opened 2026-08-10
feat(addr-mgr): the switch owns the ObservedAddrManager

gmelodie:
## Summary

First task of the [2026Q3 address manager roadmap](https://roadmap.vac.dev/p2p/ift/2026q3-nimlibp2p-addr-manager): decouple the `ObservedAddrManager`.

The manager belonged to Identify, and each consumer reached it through `switch.peerStore`. The switch now owns it as `switch.observedAddrManager`. Identify keeps a reference, and it only feeds the manager. The manager gets `start`, `stop` and `isStarted`. The switch starts it first and stops it last.

The manager also owns every rule about what a valid observation is. It takes an address which starts with ip4 or ip6, which carries a transport, and which is not relayed. Identify no longer holds a copy of that rule.

## Affected Areas

  Observed addresses move from `identify.nim` and `peerstore.nim` to `switch.nim`.

- [x] Protocol Logic
  `hpservice`, the dcutr server, and both autonat services read the manager from the switch. `identify.nim` drops its own relay filter, because `addObservation` applies it.

- [x] Other
  `withObservedAddrManager` takes an `ObservedAddrManagerConfig`. `TorSwitch` copies the new field. `multiaddress.nim` gets two helpers, `hasIp` and `hasTransport`, which the manager uses to validate an observation.

No downstream branch is necessary. Each old route stays as a deprecated shim. Nimbus, Waku and Codex: N/A.

## Impact on Library Users

Deprecated, but still functional: `withObservedAddrManager(instance)`, `PeerStore.guessDialableAddr` and `PeerStore.getMostObservedProtosAndPorts`.

Three behavior changes. `switch.stop()` drops the observations. `addObservation` rejects an address which does not start with ip4 or ip6, which has no transport part, or which is relayed. The manager takes an observation only while it runs, so a manager which never started, or which stopped, rejects every address.

A `Switch` built field by field must set `observedAddrManager`, and it must call `switch.start()` before the manager takes anything.

`ObservedAddrManagerConfig` raises a `maxSize` or a `minCount` below one to one. A `maxSize` of zero used to evict from an empty window on the first observation.

## Risk Assessment

The getters keep the same algorithm and the same defaults. The validation drops only an address which the getters can never return, so the announced addresses stay the same. The relay rule moves from Identify to the manager without a change of outcome, and the manager now also rejects a relayed address which carries an IP prefix.

## References

- https://roadmap.vac.dev/p2p/ift/2026q3-nimlibp2p-addr-manager (`obsaddr-decouple`)

## Additional Notes

`tests/libp2p/test_observed_addr_manager.nim` covers an end-to-end case over TCP, the start and stop order, the rejection rules, the threshold clamp, and the builder hooks.

`tests/libp2p/transports/stream_tests.nim` carries an unrelated fix for two flaky cases: `runSingleStreamScenario` stops the server as soon as the client handler returns, which cancels the server read before it can see the EOF that the test asserts. The client now waits for the server handler.

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -420,6 +420,7 @@ proc withObservedAddrManager*(
     b: SwitchBuilder, config: ObservedAddrManagerConfig
 ): SwitchBuilder =
   ## Set the thresholds of the observed address manager.
+  b.observedAddrManager = nil
   b.observedAddrConfig = Opt.some(config)
   b
 
@@ -428,6 +429,7 @@ proc withObservedAddrManager*(
 ): SwitchBuilder {.
     deprecated: "the switch owns the manager; pass an ObservedAddrManagerConfig"
 .} =
+  b.observedAddrConfig = Opt.none(ObservedAddrManagerConfig)
   b.observedAddrManager = observedAddrManager
   b
 
@@ -451,9 +453,6 @@ proc buildObservedAddrManager(b: SwitchBuilder): ObservedAddrManager =
   if b.observedAddrManager.isNil():
     return
       ObservedAddrManager.new(b.observedAddrConfig.get(ObservedAddrManagerConfig()))
-
-  if b.observedAddrConfig.isSome():
-    warn "the deprecated withObservedAddrManager(instance) drops the given config"
   b.observedAddrManager
 
 proc buildSwitch(b: SwitchBuilder): Switch {.raises: [LPError].} =
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -98,6 +98,7 @@ type
     kad: Opt[KadInfo]
     identifyPusherEnabled: bool
     observedAddrManager: ObservedAddrManager
+    observedAddrConfig: Opt[ObservedAddrManagerConfig]
     enableWildcardResolver: bool
     addressPolicy: PeerAddressPolicy
 
@@ -124,6 +125,7 @@ proc new*(T: type[SwitchBuilder]): T =
     enableWildcardResolver: true,
     addressPolicy: defaultAddressPolicy,
     addressTtls: AddressConfidenceTtls(),
+    observedAddrConfig: Opt.none(ObservedAddrManagerConfig),
   )
 
 proc withPrivateKey*(
@@ -415,8 +417,17 @@ proc withIdentifyPusher*(b: SwitchBuilder, enabled: bool = true): SwitchBuilder 
   b
 
 proc withObservedAddrManager*(
-    b: SwitchBuilder, observedAddrManager: ObservedAddrManager
+    b: SwitchBuilder, config: ObservedAddrManagerConfig
 ): SwitchBuilder =
+  ## Set the thresholds of the observed address manager.
+  b.observedAddrConfig = Opt.some(config)
+  b
+
+proc withObservedAddrManager*(
+    b: SwitchBuilder, observedAddrManager: ObservedAddrManager
+): SwitchBuilder {.
+    deprecated: "the switch owns the manager; pass an ObservedAddrManagerConfig"
+.} =
   b.observedAddrManager = observedAddrManager
   b
 
@@ -435,6 +446,15 @@ proc withPrivateAddressFilter*(b: SwitchBuilder): SwitchBuilder =
   ## - Private addresses received from other peers are discarded
   ## Circuit relay and DNS addresses are never filtered.
   b.withAddressPolicy(publicRoutableAddressPolicy)
+
+proc buildObservedAddrManager(b: SwitchBuilder): ObservedAddrManager =
+  if b.observedAddrManager.isNil():
+    return
+      ObservedAddrManager.new(b.observedAddrConfig.get(ObservedAddrManagerConfig()))
+
+  if b.observedAddrConfig.isSome():
+    warn "the deprecated withObservedAddrManager(instance) drops the given config"
+  b.observedAddrManager
 
 proc buildSwitch(b: SwitchBuilder): Switch {.raises: [LPError].} =
   if isNil(b.rng):
@@ -463,11 +483,8 @@ proc buildSwitch(b: SwitchBuilder): Switch {.raises: [LPError].} =
     announcedAddrs = b.announcedAddrs,
   )
 
-  let identify =
-    if b.observedAddrManager != nil:
-      Identify.new(peerInfo, b.sendSignedPeerRecord, b.observedAddrManager)
-    else:
-      Identify.new(peerInfo, b.sendSignedPeerRecord)
+  let observedAddrManager = b.buildObservedAddrManager()
+  let identify = Identify.new(peerInfo, b.sendSignedPeerRecord, observedAddrManager)
 
   var peerStore = block:
     b.peerStoreCapacity.withValue(capacity):
@@ -522,6 +539,7 @@ proc buildSwitch(b: SwitchBuilder): Switch {.raises: [LPError].} =
     rng: b.rng,
     muxedUpgrade: muxedUpgrade,
     services: services,
+    observedAddrManager: observedAddrManager,
   )
 
   return switch
--- a/libp2p/observedaddrmanager.nim
+++ b/libp2p/observedaddrmanager.nim
@@ -7,19 +7,46 @@ import std/[sequtils, tables, sugar]
 import chronos
 import multiaddress, multicodec
 
+const
+  DefaultObservedAddrMaxSize* = 10
+  DefaultObservedAddrMinCount* = 3
+
 type
-  ## Manages observed MultiAddresses by reomte peers. It keeps track of the most observed IP and IP/Port.
+  ObservedAddrManagerConfig* = object
+    maxSize*: int = DefaultObservedAddrMaxSize
+    minCount*: int = DefaultObservedAddrMinCount
+
+  ObservedAddrManagerState = enum
+    Fresh ## accepts observations: a manager seeded before the start is valid
+    Started
+    Stopped
+
   ObservedAddrManager* = ref object of RootObj
     observedIPsAndPorts: seq[MultiAddress]
     maxSize: int
     minCount: int
+    state: ObservedAddrManagerState
+
+func namesDialableAddr(observedAddr: MultiAddress): bool =
+  # a remote peer picks what it reports, and the window holds maxSize entries,
+  # so an address which no getter can return would only evict a useful one
+  let code = (observedAddr[0].flatMap(protoCode)).valueOr:
+    return false
+  if code != multiCodec("ip4") and code != multiCodec("ip6"):
+    return false
+  observedAddr.len().get(0) > 1
 
 proc addObservation*(self: ObservedAddrManager, observedAddr: MultiAddress): bool =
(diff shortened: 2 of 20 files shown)


libp2p/observedaddrmanager.nim
```diff
@@ -7,19 +7,46 @@ import std/[sequtils, tables, sugar]
...
+    state: ObservedAddrManagerState
+
+func namesDialableAddr(observedAddr: MultiAddress): bool =
+  # a remote peer picks what it reports, and the window holds maxSize entries,
+  # so an address which no getter can return would only evict a useful one
+  let code = (observedAddr[0].flatMap(protoCode)).valueOr:
+    return false
+  if code != multiCodec("ip4") and code != multiCodec("ip6"):
+    return false
+  observedAddr.len().get(0) > 1
```
vladopajic: i have no clue what it going here, func naming, comment, code..

libp2p/observedaddrmanager.nim
```diff
@@ -76,11 +103,32 @@ proc guessDialableAddr*(self: ObservedAddrManager, ma: MultiAddress): MultiAddre
   return concat(observedIP, maRest).valueOr:
     ma
 
+func isStarted*(self: ObservedAddrManager): bool =
+  self.state == Started
+
+proc start*(self: ObservedAddrManager) {.async: (raises: [CancelledError]).} =
+  if self.state == Started:
```
vladopajic:
```
if self.isStarted:
  return
```

libp2p/observedaddrmanager.nim
```diff
@@ -76,11 +103,32 @@ proc guessDialableAddr*(self: ObservedAddrManager, ma: MultiAddress): MultiAddre
...
 proc `$`*(self: ObservedAddrManager): string =
   ## Returns a string representation of the ObservedAddrManager.
   return "IPs and Ports: " & $self.observedIPsAndPorts
 
-proc new*(T: typedesc[ObservedAddrManager], maxSize = 10, minCount = 3): T =
-  ## Creates a new ObservedAddrManager.
+proc new*(
+    T: typedesc[ObservedAddrManager],
+    maxSize = DefaultObservedAddrMaxSize,
+    minCount = DefaultObservedAddrMinCount,
```
vladopajic: why have this is there is constructor with config ?

libp2p/observedaddrmanager.nim
```diff
@@ -76,11 +103,32 @@ proc guessDialableAddr*(self: ObservedAddrManager, ma: MultiAddress): MultiAddre
...
+): T =
   return
     T(observedIPsAndPorts: newSeq[MultiAddress](), maxSize: maxSize, minCount: minCount)
+
+proc new*(T: typedesc[ObservedAddrManager], config: ObservedAddrManagerConfig): T =
+  ## Creates a new ObservedAddrManager from a switch-level config.
```
vladopajic: comment why ` from a switch-level config` is this necessary to ephasize? it's just constructor.

libp2p/switch.nim
```diff
@@ -354,6 +356,10 @@ proc stop*(s: Switch) {.async: (raises: [CancelledError]).} =
 
   await s.ms.stop()
 
+  # stopped last, after every component which can still observe an address
+  if not s.observedAddrManager.isNil():
+    await s.observedAddrManager.stop()
```
vladopajic: why nil check if it is always created in start, see code bellow

libp2p/switch.nim
```diff
@@ -366,6 +372,12 @@ proc start*(s: Switch) {.async: (raises: [CancelledError, LPError]).} =
 
   debug "starting switch for peer", peerInfo = s.peerInfo
 
+  if s.observedAddrManager.isNil():
+    s.observedAddrManager = ObservedAddrManager.new()
```
vladopajic: starting should not really do any initialization. it should have happened when switch was created.
start should only start.

tests/libp2p/test_observed_addr_manager.nim
```diff
@@ -68,3 +68,193 @@ suite "ObservedAddrManager":
...
+    switch.services.add(probe)
+
+    await switch.start()
+    check switch.observedAddrManager.isStarted()
+
+    await switch.stop()
+
+    check:
+      probe.managerStartedOnServiceStart
+      probe.managerStartedOnServiceStop
```
vladopajic: why is probe necessary? why is it different then doing `check switch.observedAddrManager.isStarted()` before/after start/stop?

libp2p/observedaddrmanager.nim
```diff
@@ -76,11 +103,32 @@ proc guessDialableAddr*(self: ObservedAddrManager, ma: MultiAddress): MultiAddre
...
+  self.state = Stopped
+  self.observedIPsAndPorts.setLen(0)
+
 proc `$`*(self: ObservedAddrManager): string =
   ## Returns a string representation of the ObservedAddrManager.
   return "IPs and Ports: " & $self.observedIPsAndPorts
 
-proc new*(T: typedesc[ObservedAddrManager], maxSize = 10, minCount = 3): T =
-  ## Creates a new ObservedAddrManager.
+proc new*(
```
vladopajic, line 111: since this variant is only used in tests, it should be test local utility.

libp2p/observedaddrmanager.nim
```diff
@@ -7,19 +7,50 @@ import std/[sequtils, tables, sugar]
...
+    state: ObservedAddrManagerState
+
+func isIpWithTransport(observedAddr: MultiAddress): bool =
+  ## Is the address an IP4/IP6 address followed by a transport, as in
+  ## `/ip4/1.2.3.4/tcp/1`? The getters below return nothing else.
+  let ipCode = observedAddr[0].flatMap(protoCode).valueOr:
+    return false
+  if ipCode != multiCodec("ip4") and ipCode != multiCodec("ip6"):
+    return false
+  observedAddr.len().get(0) > 1
```
vladopajic: `return observedAddr.hasIp() and observedAddr.hasTransport()`

because MultiAddress utilities tend to repeat also having utilities improves code readability

libp2p/observedaddrmanager.nim
```diff
@@ -7,19 +7,50 @@ import std/[sequtils, tables, sugar]
...
 type
-  ## Manages observed MultiAddresses by reomte peers. It keeps track of the most observed IP and IP/Port.
+  ObservedAddrManagerConfig* = object
+    maxSize*: int = DefaultObservedAddrMaxSize
+    minCount*: int = DefaultObservedAddrMinCount
+
+  ObservedAddrManagerState = enum
+    Fresh ## accepts observations: a manager seeded before the start is valid
+    Started
+    Stopped
```
vladopajic: where is `Fresh` used?
why this cant be simple `bool` property `ObservedAddrManager.isStarted` ?

tests/libp2p/transports/stream_tests.nim
```diff
@@ -399,20 +399,25 @@ template streamTransportTest*(
     )
 
   asyncTest "multiple empty writes before closeWrite":
+    var serverHandlerDone = newFuture[void]()
```
vladopajic: why ?

vladopajic: this is still in code, and it's not clear why it was added here, do you mind explaining ?

vladopajic (approved):

libp2p/observedaddrmanager.nim
```diff
@@ -76,11 +93,25 @@ proc guessDialableAddr*(self: ObservedAddrManager, ma: MultiAddress): MultiAddre
   return concat(observedIP, maRest).valueOr:
     ma
 
+func isStarted*(self: ObservedAddrManager): bool =
+  self.started
```
richard-ramos: Is this getter really needed?

tests/libp2p/transports/stream_tests.nim
gmelodie: Ah yes, I added a comment just above:
```
# the scenario stops the server once the client returns, cancelling the read
```
But that's kinda cryptic now looking at it. The reasoning is that since the client is the one to close/tear the connection down, it might close it too early, this future syncs client/server (client only tears the connection down once the server is done with the assertions).

Below is an ai-generated flow explanation:

The client does a half-close of the stream with `closeWrite()`. The server must then do three assertions about the EOF condition. The instruction `await serverHandlerDone` keeps the client in its handler until the server completes these assertions.

If you remove this instruction, two failures occur:

- The client does a full close. The full close resets the channel. The server then gets an incorrect error type, and the `expect` blocks fail.
- The muxer close cancels the server handler. The code discards the `CancelledError`. The test then shows a pass, but it does not do the checks.

The same condition applies to the five other uses of this pattern.

vladopajic: aha, i see now, this was already done in other tests, but not all. so this pr just adds this in those that didn't have it.

libp2p/builders.nim
```diff
@@ -415,8 +417,17 @@ proc withIdentifyPusher*(b: SwitchBuilder, enabled: bool = true): SwitchBuilder
   b
 
 proc withObservedAddrManager*(
-    b: SwitchBuilder, observedAddrManager: ObservedAddrManager
+    b: SwitchBuilder, config: ObservedAddrManagerConfig
 ): SwitchBuilder =
+  ## Set the thresholds of the observed address manager.
+  b.observedAddrConfig = Opt.some(config)
```
richard-ramos, line 424:
```suggestion
  b.observedAddrManager = nil
  b.observedAddrConfig = Opt.some(config)
```

libp2p/builders.nim
```diff
@@ -415,8 +417,17 @@ proc withIdentifyPusher*(b: SwitchBuilder, enabled: bool = true): SwitchBuilder
...
+  b
+
+proc withObservedAddrManager*(
+    b: SwitchBuilder, observedAddrManager: ObservedAddrManager
+): SwitchBuilder {.
+    deprecated: "the switch owns the manager; pass an ObservedAddrManagerConfig"
+.} =
   b.observedAddrManager = observedAddrManager
```
richard-ramos, line 433:
```suggestion
  b.observedAddrConfig = Opt.none(ObservedAddrManagerConfig)
  b.observedAddrManager = observedAddrManager
```

richard-ramos (approved):

libp2p/observedaddrmanager.nim
vladopajic: it's like read only for `started` property. but since it is used only in tests it can be conditional exported only for `libp2p_testing`

libp2p/builders.nim
vladopajic: or doAssert if config is some?

libp2p/builders.nim
vladopajic: or doAssert if observedAddrManager is not nil

libp2p/builders.nim
vladopajic: this also has implication on `buildObservedAddrManager` - `warn` log is not needed then since builder methods ensure correct behaviour

richard-ramos (approved):

PR #2926, opened 2026-08-11
ci: change log lvl

vladopajic:
- changed log lvl for `test_all` binaries to `INFO`
- changed log lvl for interop binaries to `DEBUG` since they generate much shorter log due to smaller usage


arnetheduck:
not sure if it's covered elsewhere, but it's usually good to compile at least once somewhere with trace enabled at compile time .. ie it might make more sense to compile with TRACE and then reduce the actual log level when running the test at runtime

vladopajic:
daily tests run with debug.

alternativly it can also be helpful not showing logs , at least in ci, if test have passed. then we can show very detailed logs for failed tests and avoid noise in stdout

richard-ramos (approved):

gmelodie (approved):

PR #2927, opened 2026-08-11
chore: adding more metrics

vladopajic:
adding various metrics to improve node insight.

these new metrics will be utilized in logos-libp2p-ui

### DIFF
--- a/libp2p/protocols/connectivity/relay/relay.nim
+++ b/libp2p/protocols/connectivity/relay/relay.nim
@@ -11,6 +11,7 @@ import
   ./messages,
   ./rconn,
   ./utils,
+  ./relay_metrics,
   ../../../peerinfo,
   ../../../switch,
   ../../../multiaddress,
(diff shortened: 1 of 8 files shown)


libp2p/stream/lpstream.nim
```diff
@@ -296,6 +298,7 @@ proc resetStream(s: LPStream): Future[void] {.async: (raises: [], raw: true).} =
 
   s.isClosed = true
   s.isResetLocally = true
+  libp2p_stream_resets.inc(labelValues = [s.objName, $s.dir])
```
richard-ramos: In other metrics we do use `in` and `out`. Can we make sure the capitalization of $s.dir is also lowercase?

libp2p/protocols/connectivity/relay/relay.nim
richard-ramos, line 177: Should this case have the metric too?

gmelodie (approved):
Dunno if this was in the plans but... will be really cool to see the stats on the ui!

libp2p/protocols/connectivity/relay/relay.nim
vladopajic: yes

richard-ramos (approved):

PR #2928, opened 2026-08-11
feat(addr-mgr): the AddressManager owns every announced address

gmelodie:
## Summary

Six components decide what a node announces today, and each one appends its own mapper to `PeerInfo.addressMappers`. Nothing records where an address came from, or drops it when its producer stops offering it.

`ObservedAddrManager` becomes `AddressManager` and takes that job: it installs the one mapper `PeerInfo` runs, and every producer registers with it.

```
before: PeerInfo.addressMappers = [wildcard, upnp, autonat, autorelay]
after:  PeerInfo.addressMappers = [AddressManager]
                                   └─ wildcard → upnp → autonat → autorelay
```

It keeps one candidate `(address, source, state)` per address. `AddrSource` names the producer: `Autonat`, `Upnp`, `NatPmp`, `PortMapped`, `ExplicitIp`, `Circuit`, `Listen`, `Announced`. A libplum mapper in `Auto` mode reports `PortMapped`, because libplum does not say which protocol the gateway answered with. `AddrState` is what a verifier found. The verifier arrives in the next roadmap task, so every candidate stays `Unverified` and every candidate is announced, which is what a node does today.

The ownership buys one behavior now: a mapper which stops producing an address withdraws it, so an expired UPnP mapping or a lost relay reservation leaves the announce set.

Three pieces of the roadmap task are left out, to keep this PR reviewable. They land with the verifier: the candidate TTL and its pruning, the per-listen-address observation windows, and a debounce on the `PeerInfo` observer notifications. The observation list stays flat, and it stays the size it is on master.

Read `add`, `update`, `remove`, `candidates` and `mapperSources` as an interface, not as live code: they have no caller outside the tests yet. They are the store the verifier drives in the next task.

## Affected Areas

  New `libp2p/addressmanager.nim`, which also absorbs the wildcard expansion from `wildcardresolverservice`; `switch.observedAddrManager` becomes `switch.addressManager`.

- [x] Protocol Logic
  The NAT service, both AutoNAT services and AutoRelay call `addressManager.addMapper(mapper, source)`.

The `observedaddrmanager` module, the `switch.observedAddrManager` accessor, the `withObservedAddrManager` hook and `AddressManager.guessDialableAddr` all keep working, and all are deprecated. Nimbus, Waku and Codex: N/A.

## Impact on Library Users

- `addObservation` drops an unspecified, multicast or reserved IP, so a peer can no longer push `/ip4/0.0.0.0/tcp/1` into the observation list. A loopback or a private address is still accepted, because a peer on the same host or the same LAN does reach the node there, and `PeerAddressPolicy` stays the gate on what is announced.
- The oldest observation is now evicted in order. Master used `seq.del`, which is an unordered swap-remove: it moved the newest entry into the freed slot instead of dropping the oldest one.
- The mapper chain also runs when `announcedAddrs` is set, so the NAT service maps the bound ports in that case too. The announced addresses still win, and no mapper rewrites them.
- A `Switch` built field by field must set `addressManager`.

## Risk Assessment

The observation list, the two thresholds and the two legacy getters keep their algorithm, so a node without an announce list announces the same addresses as before. The chain runs the mappers in registration order, which is the service start order, as before.

One known weakness is unchanged from master, and it is worth naming: `addObservation` does no per-peer accounting, so one peer which connects `minCount` times decides the most observed address. The fix is the per-listen-address window and the verifier, both in the next task.

## References

- https://roadmap.vac.dev/p2p/ift/2026q3-nimlibp2p-addr-manager (`obsaddr-revamp`)
- #2924 (`obsaddr-decouple`, the previous task)

## Additional Notes

Most of the twenty-six files carry the rename only. The behavior lives in `libp2p/addressmanager.nim` (`resolve` runs the chain, folds the outcome into the candidates, and returns the announce list when the operator set one) and in `libp2p/peerinfo.nim` (the mapper chain runs before the announce list wins, instead of being skipped).

The state is `AddrState.Unverified`, not the roadmap's `unknown`: `Unknown` is ambiguous with `NetworkReachability.Unknown` in every module which imports the switch.


libp2p/address_manager.nim
```diff
@@ -0,0 +1,366 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
```
vladopajic, line 1: file name can be with `_`

for new files we are doing it user friendly way

vladopajic (approved):

libp2p/address_manager.nim
```diff
@@ -0,0 +1,366 @@
...
+  toSeq(self.candidates.values)
+
+func notFromMappers(self: AddressManager, address: MultiAddress): bool =
+  address notin self.chainAddrs
+
+func isAnnounceable(self: AddressManager, address: MultiAddress): bool =
+  # announced until a verifier proves it unreachable
+  self.candidates.getOrDefault(address).state != AddrState.Unreachable
+
+proc addMapper*(self: AddressManager, mapper: AddressMapper, source: AddrSource) =
```
richard-ramos, line 208: Question:
Mappers added via this function will get removed on `stop`. Should they be persisted?

libp2p/addressmanager.nim
```diff
@@ -0,0 +1,366 @@
...
+    source: AddrSource,
+    state = AddrState.Unverified,
+): bool {.discardable.} =
+  ## Adds a candidate, or refreshes the source of the one already under
+  ## `address`, keeping the state a verifier assigned. True for a new candidate.
+  let
+    isNew = address notin self.candidates
+    fresh = AddrCandidate(address: address, state: state)
+  self.candidates.mgetOrPut(address, fresh).source = source
+  isNew
```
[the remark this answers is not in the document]
richard-ramos: Looks valid. Not sure about how common it is tho. It seems the solution should be to have `track` track independent sources per address and have `withdraw` be aware of the source :thinking: and remove the address only after last source withdraws :thinking:

richard-ramos (approved):

libp2p/address_manager.nim
gmelodie: I don't see why, mappers are registered on `start`

PR #2929, opened 2026-08-12
fix(dialer): bound outbound dials so one stuck peer cannot stall the node

gmelodie:
## Summary

An outbound dial has no deadline. `Dialer.internalConnect` takes the per-peer dial lock. It then waits for `transport.dial`, for `transport.upgrade` and for `peerStore.identify`. None of the three has a bound. The inbound path caps the same work with `UpgradeTimeout` (`switch.nim:252`). A remote that completes the transport handshake and then goes quiet parks the dial forever, and holds the peer's dial lock for the same time.

A 1000-node run showed the effect. About 400 nodes never returned from `KadDHT.bootstrap(forceRefresh = true)`. There was no error, no timeout and no retry. `-d:chronosFutureTracking` showed the queue: dials parked in `internalConnect`, more dials waited at `AsyncLock.acquire`, and `cancelAndWait` never completed, because that build dialed under `noCancel`. `lookupCheck` waits for its probe, `iterativeLookup` waits for `lookupCheck`, `refreshTable` waits for the lookup, and `bootstrap` waits for `refreshTable`.

#2901 removed the `noCancel` dial from the Kademlia RPC path. This PR fixes the dialer.

Dialer:

- `DefaultDialerTimeout` (30s) bounds one attempt to reach a peer. All addresses of that peer share the deadline, because a per-address budget multiplies by a count that the remote picks.
- `identify` gets the same budget again after the connection stands. It runs under the same lock.
- A dial without a peer id takes no lock. Such dials shared one `default(PeerId)` entry, so one dead address blocked all of them.
- `acquireDialLock` handles a cancel at the instant that chronos hands the lock over. That race left the lock held with no owner.
- Dial lock entries count their users, and the dialer drops each entry when it goes idle. The table grew with every peer id.
- A failed protocol dial no longer closes `conn`. The connection manager owns that muxer, and other protocols hold streams on it. The dialer resets the stream instead.

Kademlia:

- `start` and `maintainBuckets` cancel the refresh future when `withTimeout` fires. Both used `discard await ...withTimeout(...)`, which left the lookup alive and kept it sending RPCs.

Tests:

- Two tests in `test_dialer.nim` dial a TCP server that accepts the connection and stays silent. The first test checks that the dial gives up at the timeout, twice, which proves that it frees the lock. The second test checks that a dial without a peer id does not block a second one.

`timeLeft(deadline)` moves from `message_sender.nim` to `utils/future.nim`, because the dialer needs it too.

## Affected Areas

- [x] Protocol Logic
  

Every downstream user shares this dial path. A dial that used to hang now fails with `DialFailedError` after 30 seconds.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

No API is removed. `Dialer.new` gains an optional `dialTimeout`. Both `dialAndUpgrade` overloads gain an optional `deadline`.

- One attempt to reach a peer costs at most `DefaultDialerTimeout` for all of its addresses, plus the same again for identify. Before, each address got an unbounded attempt.
- A failed or cancelled protocol dial keeps the connection open. Streams that other protocols hold on it stay alive.
- `TorTransport` users with slow circuit setup can need a larger `dialTimeout`. `TorSwitch` builds its own `Dialer` and can pass one.

## Risk Assessment

- Backward compatibility: no signature is removed, and every new parameter has a default.
- Network behaviour: the node can now drop a peer that is slow but alive at 30 seconds. `UpgradeTimeout` already applies that value to the inbound side of the same handshake.
- Performance: one more timer per dial attempt.
- Security: a remote that does not answer can no longer hold a dial slot, a dial lock or a bootstrap for an unbounded time. It also cannot extend the hold with more addresses.

## References

- #2901 removed the `noCancel` dial from the Kademlia RPC path, and added the per-RPC deadline in `message_sender`.

## Additional Notes

`kademlia/ping.nim:14` and `service_discovery/connection.nim:30` still dial under `noCancel`. They can no longer hang forever, but a cancel there waits up to `DefaultDialerTimeout`. Take care with the service-discovery one: the `catch:` around it swallows a `CancelledError`.

### DIFF
--- a/libp2p/dialer.nim
+++ b/libp2p/dialer.nim
@@ -76,7 +76,7 @@ proc dialAndUpgrade*(
       let dialed =
         try:
           libp2p_total_dial_attempts.inc()
-          deadline.awaitWithDeadline(transport.dial(hostname, addrs, peerId, dir))
+          transport.dial(hostname, addrs, peerId, dir).awaitWithDeadline(deadline)
         except CancelledError as e:
           trace "Dialing canceled", description = e.msg, peerId
           raise e
@@ -98,7 +98,7 @@ proc dialAndUpgrade*(
           # The if below is more general and might handle other use cases in the future.
           if dialed.dir != dir:
             dialed.dir = dir
-          deadline.awaitWithDeadline(transport.upgrade(dialed, peerId))
+          transport.upgrade(dialed, peerId).awaitWithDeadline(deadline)
         except CancelledError as e:
           await dialed.close()
           raise e
@@ -131,7 +131,7 @@ proc dialAndUpgrade*(
   return nil
 
 proc expandDnsAddr(
-    self: Dialer, peerId: Opt[PeerId], address: MultiAddress
+    self: Dialer, peerId: Opt[PeerId], address: MultiAddress, deadline: Moment
 ): Future[seq[(MultiAddress, Opt[PeerId])]] {.
     async: (raises: [CancelledError, MaError, TransportAddressError, LPError])
 .} =
@@ -142,16 +142,22 @@ proc expandDnsAddr(
     return @[]
 
   trace "Start trying to resolve addresses"
-  let
-    toResolve =
-      if peerId.isSome:
-        try:
-          address & MultiAddress.init(multiCodec("p2p"), peerId.tryGet()).tryGet()
-        except ResultError[void]:
-          raiseAssert "checked with if"
-      else:
-        address
-    resolved = await self.nameResolver.resolveDnsAddr(toResolve)
+  let toResolve =
+    if peerId.isSome:
+      try:
+        address & MultiAddress.init(multiCodec("p2p"), peerId.tryGet()).tryGet()
+      except ResultError[void]:
+        raiseAssert "checked with if"
+    else:
+      address
+  # A dnsaddr record points at more dnsaddr records, and each lookup can take
+  # seconds, so the chain answers to the dial deadline like the dial itself.
+  let resolved =
+    try:
+      self.nameResolver.resolveDnsAddr(toResolve).awaitWithDeadline(deadline)
+    except AsyncTimeoutError:
+      debug "Out of time resolving dnsaddr", ma = toResolve
+      return @[]
 
   debug "resolved addresses",
     originalAddresses = toResolve, resolvedAddresses = resolved
@@ -171,6 +177,20 @@ proc expandDnsAddr(
     else:
       addrs.add((resolvedAddress, peerId))
   addrs
+
+proc resolveWithDeadline(
+    self: Dialer, address: MultiAddress, deadline: Moment
+): Future[seq[MultiAddress]] {.
+    async: (raises: [CancelledError, MaError, TransportAddressError])
+.} =
+  ## Empty when the resolver has no answer left in the dial's budget.
+  if isNil(self.nameResolver):
+    return @[address]
+  try:
+    self.nameResolver.resolveMAddress(address).awaitWithDeadline(deadline)
+  except AsyncTimeoutError:
+    debug "Out of time resolving address", ma = address
+    @[]
 
 proc normalizedDialAddrs(
     peerId: Opt[PeerId], addrs: seq[MultiAddress]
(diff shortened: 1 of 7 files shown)


libp2p/utils/future.nim
```diff
@@ -34,6 +34,11 @@ proc anyCompleted*[T](
     except CatchableError:
       continue
 
+proc timeLeft*(deadline: Moment): Duration =
```
richard-ramos: This can be a template

libp2p/dialer.nim
```diff
@@ -62,13 +76,13 @@ proc dialAndUpgrade*(
       let dialed =
         try:
           libp2p_total_dial_attempts.inc()
-          await transport.dial(hostname, addrs, peerId, dir)
-        except CancelledError as exc:
-          trace "Dialing canceled", description = exc.msg, peerId
-          raise exc
-        except CatchableError as exc:
+          await transport.dial(hostname, addrs, peerId, dir).wait(deadline.timeLeft())
```
richard-ramos: The `wait` does not cancel the future if its duration is 0, so you can end up with a dial still happening even if it's meant to not happen. For this to be 'safe' you should do something +- like this (which maybe can become a template) so you use it in line 101 too

```
let remaining = deadline.timeLeft()
if remaining.isZero:
  return nil

let dialFut = transport.dial(hostname, addrs, peerId, dir)
await dialFut.wait(remaining)
```

richard-ramos: Example:
```nim
  asyncTest "A zero-duration wait does not cancel the underlying future":
    let pending = sleepAsync(1.hours)
    expect AsyncTimeoutError:
      await pending.wait(ZeroDuration)

    check not pending.finished()
```

richard-ramos (approved):
LGTM.


> Additional Notes
> kademlia/ping.nim:14 and service_discovery/connection.nim:30 still dial under noCancel. They can no longer hang forever, but a cancel there waits up to DefaultDialerTimeout. Take care with the service-discovery one: the catch: around it swallows a CancelledError.

Is this to be done on a separate PR or it is not really needed

libp2p/utils/future.nim
```diff
@@ -34,6 +34,19 @@ proc anyCompleted*[T](
     except CatchableError:
       continue
 
+template timeLeft*(deadline: Moment): Duration =
+  ## Zero once the deadline passed: chronos clamps a negative `Duration`.
+  deadline - Moment.now()
+
+template awaitWithDeadline*(deadline: Moment, fut: untyped): untyped =
```
vladopajic: - if fut is completed does it make sense to throw error?
- how about different argument order: fut, deadline

richard-ramos:
Some of copilot suppresed comments look interesting

libp2p/dialer.nim
```diff
@@ -62,13 +76,13 @@ proc dialAndUpgrade*(
       let dialed =
         try:
           libp2p_total_dial_attempts.inc()
-          await transport.dial(hostname, addrs, peerId, dir)
-        except CancelledError as exc:
-          trace "Dialing canceled", description = exc.msg, peerId
-          raise exc
-        except CatchableError as exc:
+          deadline.awaitWithDeadline(transport.dial(hostname, addrs, peerId, dir))
```
vladopajic: `deadline.awaitWithDedline`

```
let dialFut = ...
dialFut.awaitWithDeadline(deadline)
```

libp2p/dialer.nim
```diff
@@ -208,26 +233,68 @@ proc tryReusingConnection(self: Dialer, peerId: PeerId): Opt[Muxer] =
...
+): Future[DialLock] {.async: (raises: [CancelledError]).} =
+  let dialLock = self.dialLocks.mgetOrPut(peerId, DialLock(lock: newAsyncLock()))
+  dialLock.users.inc()
+  # Cancellation can land at the instant chronos hands the lock over, which
+  # leaves it held with no owner and stalls every later dial to this peer.
+  let acquireFut = dialLock.lock.acquire()
+  try:
+    await acquireFut
+  except CancelledError as e:
+    if acquireFut.completed():
```
vladopajic: strange if this can be true

libp2p/dialer.nim
```diff
@@ -208,26 +233,68 @@ proc tryReusingConnection(self: Dialer, peerId: PeerId): Opt[Muxer] =
   trace "Reusing existing connection", muxer, direction = $muxer.connection.dir
   return Opt.some(muxer)
 
-proc internalConnect(
+proc release(dialLock: DialLock) {.raises: [].} =
```
vladopajic: does it make sense for release to dropUser?

or dropUser to do a release as well.

seems to me that these two operations should be merged???

vladopajic (approved):

PR #2931, opened 2026-08-13
test(autotls): broker

rlve:
## Summary

Adds coverage for `AutotlsBroker`. Four cases driving `sendChallenge` directly: 
- an empty address set raises before any request goes out, 
- the payload carries the key authorization and every address in order, 
- the configured broker URL is what gets contacted, 
- and a non-200 response raises instead of being treated as success.

Also replaces `MockPeerIDAuthClient` with `PeerIDAuthClientStub`, injected through the `peerIdAuthClient` constructor parameter. The mock served fixed headers, and no fixed header carries a valid signature over a challenge that changes on every call, so it could not complete the handshake. 

The stub signs with its own key and records the URIs and payloads sent. Either response header can still be overridden to serve a malformed or absent one, which is what the existing `test_peer_id_auth` cases use, so `libp2p/peeridauth/mockclient.nim` is deleted and the test double no longer ships in the library.


tests/stubs/peer_id_auth_client_stub.nim
```diff
@@ -0,0 +1,100 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
```
vladopajic, line 1: `peer_id_auth_client_stub` 🙏

gmelodie (approved):

tests/libp2p/autotls/test_broker.nim
```diff
@@ -0,0 +1,70 @@
...
+    check client.requestedUris.len == 0
+
+  asyncTest "the payload carries the key authorization and every address in order":
+    await broker.sendChallenge(peerInfo, addrs, KeyAuth)
+
+    check client.payloads.len == 1
+    check parseJson(client.payloads[0]) ==
+      %*{
+        "value": KeyAuth,
+        "addresses": ["/ip4/1.2.3.4/tcp/4001", "/ip4/1.2.3.4/tcp/4002/ws"],
```
vladopajic: nit:

`["/ip4/1.2.3.4/tcp/4001", "/ip4/1.2.3.4/tcp/4002/ws"]` reuse as variable

vladopajic (approved):

PR #2934, opened 2026-08-13
feat(addr-mgr): a Verifier confirms candidates on a heartbeat

gmelodie:
## Summary

The `AddressManager` calls a `Verifier` on a heartbeat to confirm each candidate address, and this PR adds an AutoNATv2-backed verifier.

Until now a candidate carried a state (`Unverified`, `Confirmed`, `Unreachable`) which nothing set. The manager now runs a verification heartbeat, asks the registered verifier about every candidate, and writes the verdicts back.

The verifier is optional. Without one, every candidate stays `Unverified` and nothing changes.

A verdict sets the state only. The announce set no longer drops an `Unreachable` candidate, so a peer which lies about a dial cannot silence a node. `confirmedAddrs()` exposes the confirmed subset for the consumers which come with the wiring milestone.

`AutonatV2Verifier` sends one AutoNATv2 dial request per address, each to a peer of its own, and it maps the answer: `Reachable` to `Confirmed`, `NotReachable` to `Unreachable`, `Unknown` to no verdict. It skips a peer which already holds an inbound connection to us, and it stops when the node runs out of incoming slots for the dial back. Each request times out after twice the time a server has to dial back.

One verification run fits in one `verifyInterval`. The manager cancels a run which needs more, so a slow verifier delays no other run and starts no second one.

This is milestone 3 of the address-manager roadmap. Nothing wires the verifier into the switch yet, which is milestone 4.

## Affected Areas

- [x] Protocol Logic

No downstream change is needed. The new API is additive and the verification heartbeat idles while no verifier is registered.

## Impact on Library Users

New API: `Verifier`, `AddrVerdict`, `AddressManager.verifier=`, `AddressManager.confirmedAddrs`, `AddressManagerConfig.verifyInterval` (5 minutes by default) and `AutonatV2Verifier`.

Behaviour change: the manager announces an `Unreachable` candidate again. Only a caller which set that state by hand saw the old behaviour, because no verifier existed to set it.

`hasEnoughIncomingSlots` and `hasIncomingConn` moved from `autonatv2/service.nim` to `autonatv2/utils.nim`, where the verifier also uses them. Both were private, and `hasIncomingConn` carried the name `doesPeerHaveIncomingConn`. The copies in `autonat/service.nim` stay as they are.

## Risk Assessment

One peer answers for one address per run, and it can claim a dial failure it never made. The cost of that lie is the annotation: the announce set ignores the state, and a fresh peer answers on the next run. A verdict of `Confirmed` needs a real dial back which carries the nonce, so a peer cannot forge it alone. Agreement across several peers belongs to the consumers, in the wiring milestone.

The heartbeat holds no lock and it drops a verdict on an address which went away while the verifier ran.

`AddressManager.stop` is synchronous, so it cancels the heartbeat with `cancelSoon` and returns. A request in flight unwinds after the switch stops, as the AutoNATv2 service schedule does today.

## References

Roadmap: https://roadmap.vac.dev/p2p/ift/2026q3-nimlibp2p-addr-manager, milestone "Verifier Interface & AutoNATv2 Implementation".

Follows #2924 and #2928.

## Additional Notes

Tests: `tests/libp2p/test_address_manager.nim` covers the heartbeat, the optional verifier, a verdict on an unknown address, a verifier which runs out of time and the stop path. `tests/libp2p/protocols/test_autonat_v2_verifier.nim` covers the four answers and the peer selection with the AutoNATv2 client mock.


tests/libp2p/test_address_manager.nim
```diff
@@ -23,6 +23,13 @@ proc newStartedManager(
   manager.start()
   manager
 
+proc newVerifyingManager(verifier: Verifier = nil): AddressManager =
```
vladopajic: currently tests uses `makeXXX` as convention for all functions that are making, creating, instantiating, constructing any types. and other places

tests/libp2p/test_address_manager.nim
```diff
@@ -267,6 +293,80 @@ suite "AddressManager candidates":
...
+  teardown:
+    checkTrackers()
+
+  asyncTest "a candidate stays unverified while no verifier runs":
+    let
+      manager = newVerifyingManager()
+      address = ma("/ip4/1.2.3.4/tcp/1")
+
+    manager.add(address, AddrSource.Listen)
+    await sleepAsync(50.milliseconds)
```
vladopajic: why is sleepAsync needed? why 50ms?

tests/libp2p/test_address_manager.nim
```diff
@@ -267,6 +293,80 @@ suite "AddressManager candidates":
...
+      verifier = newStubVerifier()
+      manager = newVerifyingManager(verifier)
+
+    manager.add(address, AddrSource.Listen)
+    await verifier.ran.wait()
+
+    manager.stop()
+    let runs = verifier.asked.len
+    manager.add(address, AddrSource.Listen)
+    await sleepAsync(50.milliseconds)
```
vladopajic: how to we know that heartbeat is not 100ms in this test? 
time is `sleepAsync` needs to come from some source of truth value.

`await sleepAsync(cfg.verifyInterval * 2)` or something like that

vladopajic (approved):

richard-ramos (approved):

PR #2936, opened 2026-08-14
fix(kad): an admission probe no longer waits out the dialer timeout

gmelodie:
## Summary

Around 100 nodes in a simulation run never joined the DHT. Each one connected to the bootstrap node, kept that single connection and one routing-table entry, and reported no error. Every admission probe on such a node failed, and each failure took about 30 seconds. The same probes finish in 20-50 ms on a healthy node.

`prepStream` waited on the dial through `noCancel dialFut`. A `noCancel` future clears its cancel callback and carries `OwnCancelSchedule`, so `tryCancel` does nothing on it and `cancelSoon` retries until the wrapped future ends on its own. Both ways out of that wait therefore ignored the RPC deadline:

- The `withTimeout(timeLeft)` timer cancelled the shielded future, which ignored the cancel, so the wait ran until the dial ended.
- `lookupCheck` bounds a probe with `kad.config.timeout` (5s) and settles it with `cancelAndWait`. That cancel landed on the same shielded wait.

Both paths ended only when the dial hit `DefaultDialerTimeout` (30s), which is the number in the report. `admitPeer` holds one of the 20 `admissionSem` slots for the whole probe, so 20 candidates on blackholed addresses parked the whole semaphore for 30 seconds at a time. `admitPeers` then hits `tryAcquire` failure and breaks out, and a cold node makes no progress.

The fix waits through `join()`. It completes when the dial completes, and cancelling it only removes the callback, so a cancelled or timed-out wait drops the dial instead of waiting it out. The dial itself keeps running under `dropLateDial`, which is what keeps a reused connection alive for the peer's other streams. A failed probe now costs `kad.config.timeout` instead of `DefaultDialerTimeout`, so the 20 probe slots turn over six times faster.

`lookupCheck` also logs why a probe failed, at debug level: the timeout, or the send error with its stage. That information was missing, which is why the failure needed a custom build to diagnose.

## Affected Areas

  Kademlia admission and liveness probes: a probe that cannot reach its peer now gives up at the RPC deadline.

- [x] Protocol Logic
  `MessageSender.prepStream`, shared by every outbound Kademlia RPC.

No downstream integration branches. The change is internal to the Kademlia message sender, and the wire format, the API and the defaults are untouched.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

No API change. One behavior change: an outbound Kademlia RPC to an unreachable peer now returns at its own `timeout` instead of at `DefaultDialerTimeout`. The error is unchanged (`dialStage`, "timed out dialing the peer"). Kademlia on a node that reaches many dead addresses admits peers again instead of stalling with the probe semaphore full.

## Risk Assessment

- Backward compatibility: none affected. No wire, API or config change.
- Network behavior: a node abandons a slow dial sooner, so it re-probes a candidate sooner. The dial is not cancelled, so the reused connection and its other streams survive, which the existing test "a cancelled RPC leaves the peer's connection up" covers.
- Dial amplification, the cost of this change: a peer that blackholes its addresses is now re-probed every `timeout` instead of every `DefaultDialerTimeout`, which is up to six times as many dial attempts on a dead peer. The extra dials queue on the peer's dial lock in `internalConnect`, so they take no outgoing connection slot, but each one does hold a probe slot until it times out. A negative cache for recently failed candidates is the real answer, and it is not in this PR.
- Performance: strictly less waiting on the probe path. The abandoned dial still runs to completion under `dropLateDial` and `MessageSender.stop` still waits for it, as before.
- Security: none.

## References

Reported from a simulation run in which about 100 nodes stayed on their bootstrap connection.

## Additional Notes

Three regression tests. Two are in `tests/libp2p/kademlia/test_message_sender.nim` and use a server that accepts a connection and never speaks, so the dial stalls in the handshake: one covers the RPC deadline, the other an external cancel. Both fail at exactly 30.0 s on the old code and pass in about 0.2 s with the fix. The third is in `tests/libp2p/kademlia/test_limits.nim` and asserts the reported symptom directly: an admission probe against such a peer frees its `admissionSem` slot at the probe timeout. It fails on the old code with `AsyncTimeoutError` after 5 s.

`StallServer` moves out of `tests/libp2p/test_dialer.nim` into `tests/tools/stall_server.nim`, unchanged apart from its now private `accepted` field, so all three suites use it.

Suites run locally: message_sender 11 OK, limits 11 OK, dialer 5 OK, find 20 OK, bootstrap 19 OK, lookup 17 OK, ping 1 OK.

One caveat for the reported run: a 30-second probe means the dial reached the dialer timeout, so those addresses drop the SYN or stall in the handshake. This change makes a node give up faster and keeps its probe slots turning. It does not make an unreachable address reachable. The new debug lines name the failing stage.

A related instance of the same defect stays untouched, on purpose: `peerStore.identify` closes its stream in a `finally` with `await noCancel stream.closeWithEOF()` and no timeout (`libp2p/peerstore.nim:569`), while `Dialer.finishUpgrade` calls that identify under `wait(self.dialTimeout)`. A timeout there cancels identify, and the cancellation then parks in the unbounded close. `MessageSender.retireStream` already wraps the same call in `withTimeout`. That fix belongs in a dialer PR, not in this one.


libp2p/protocols/kademlia/find.nim
```diff
@@ -381,8 +381,10 @@ proc lookupCheck*(
     await noCancel probe.cancelAndWait()
   discard await probe.withTimeout(kad.config.timeout)
   if not probe.completed():
+    debug "Kad probe timed out", peer = peerId.shortLog(), timeout = kad.config.timeout
```
richard-ramos: Wouldnt this be too noisy? (same on line 387)

vladopajic (approved):

libp2p/protocols/kademlia/message_sender.nim
```diff
@@ -216,10 +216,10 @@ proc prepStream(
     return err(SendError.init(dialStage, "timed out before dialing"))
 
   let dialFut = ms.switch.dial(peerId, addrs, ms.codec)
-  let shielded = noCancel dialFut
+  # use `join` so a cancelled wait drops the dial instead of waiting it out.
   let dialed =
     try:
-      await shielded.withTimeout(timeLeft)
+      await dialFut.join().withTimeout(timeLeft)
```
richard-ramos, line 222: I'm not sure of this solution. 
The `join` will still leave the dialing ongoing, so, the code will return once the 5s happens, but the dial will still continue in the background (for the 30s a dial can take before timeing out)

richard-ramos: Here's a vibecoded example that demonstrates what happens. Let me know your thoughts.


```nim

# SPDX-License-Identifier: Apache-2.0 OR MIT
# Copyright (c) Status Research & Development GmbH

{.used.}

import chronos
import ../../tools/unittest

type SerializedWork = ref object
  lock: AsyncLock ## Models the dialer's per-peer lock.
  release: Future[void].Raising([])
    ## Models a remote that stalls the active dial. Completing this future lets
    ## the queued operations acquire the lock in turn.
  started: int ## Operations that acquired the lock, not merely callers created.

proc newSerializedWork(): SerializedWork =
  SerializedWork(
    lock: newAsyncLock(),
    release: Future[void].Raising([]).init(
      "serialized work release", {FutureFlag.OwnCancelSchedule}
    ),
  )

proc run(work: SerializedWork) {.async: (raises: [CancelledError]).} =
  # The dial timeout would normally start after this acquisition, so time spent
  # queued behind another dial is outside that budget.
  await work.lock.acquire()
  defer:
    try:
      work.lock.release()
    except AsyncLockError as exc:
      raiseAssert "serialized work lock released without acquire: " & exc.msg

  work.started.inc()
  await work.release

suite "join with timeout":
  teardown:
    checkTrackers()

  asyncTest "timed-out joins leave serialized work queued":
    let work = newSerializedWork()
    var underlying: seq[Future[void]]

    # `withTimeout` cancels each temporary `join` future, but `join` deliberately
    # does not forward cancellation to `fut`. All three `run` futures stay live.
    for _ in 0 ..< 3:
      let fut = work.run()
      underlying.add(fut)
      check not await fut.join().withTimeout(10.milliseconds)

    # The first operation is stalled while the other two wait for its lock.
    check work.started == 1

    # Once the stall ends, the abandoned operations drain from the lock queue
    # even though none of their original callers is waiting anymore.
    work.release.complete()
    await allFutures(underlying)

    # The two callers that timed out still execute after the lock is released.
    check work.started == 3

```

richard-ramos (approved):

libp2p/protocols/kademlia/message_sender.nim
richard-ramos: Approving in the meantime, but i think the logs should be changed to trace, and respecting my comment on the join() evaluate how it goes during dst tests

gmelodie: huh....

gmelodie: I don't think we want to cancel the dial though, just stop waiting on it. Cancelling the dial would kill the connection that other streams depend on, no? The real fix is a negative cache for recently failed candidates (so that we don't dial them again), which is precisely https://github.com/vacp2p/nim-libp2p/pull/2939

libp2p/protocols/kademlia/find.nim
gmelodie: changed to `trace`

PR #2937, opened 2026-08-14
feat(kad): re-seed a routing table that fell below a minimum size

### LINKED ISSUE #2859: kad: add a fixLowPeers loop to auto-heal a shrinking/partitioned routing table
Bootstrap runs once at `kademlia.nim:131 start`; afterward only the periodic 10-minute bucket refresh (`maintainBuckets`) runs. There is no recovery path when the table shrinks below a healthy size or the node is partitioned.

go-libp2p's `runFixLowPeersLoop` triggers when the RT drops below `minRTRefreshThreshold`: it re-adds currently-connected peers, and if still too small re-dials bootstrap peers, then kicks a refresh.

Proposed work:
- Add a background loop that, when the routing table falls below a threshold, re-inserts connected peers and re-dials configured bootstrap nodes, then forces a refresh.

Impact: partition/failure recovery. Medium.

gmelodie:
## Summary

Bootstrap runs one time, in `KadDHT.start`. After that only the 10-minute bucket refresh (`maintainBuckets`) runs, and it can only refresh peers that the table already holds. A node that loses its peers to an eviction storm, or that comes back from a network partition, has no path back into the DHT: an empty or near-empty table gives the refresh nothing to walk from.

This PR adds `fixLowPeers`, the equivalent of the go-libp2p `runFixLowPeersLoop`. A background loop compares the routing table against `config.minRoutingTableSize` every `config.fixLowPeersInterval`. When the table is short, the loop re-seeds it from two sources and then forces a refresh:

- the currently connected peers, through the normal admission probe (`admitPeers`), because a live connection does not prove that the peer speaks the DHT protocol.
- the configured bootstrap nodes, through the trusted path (`updatePeers`), because the operator already trusted them at construction time, and because the admission probes only land after the current pass.

The forced refresh (`refreshTable(forceRefresh = true)`) then walks every bucket, and not only the stale ones, so the re-seeded peers give closer peers immediately.

To make the bootstrap nodes available after construction, `KadDHT` now keeps them in a `bootstrapNodes` field. `initKadBase` takes them and does the initial `updatePeers`, so `KadDHT.new` and `ServiceDiscovery.new` share one seeding path instead of two copies.

## Affected Areas

- [x] Protocol Logic

Kademlia only. `libp2p/protocols/kademlia.nim` gets `connectedPeerInfos`, `fixLowPeers` and the `maintainMinPeers` loop, started in `start` and cancelled in `stop`. `routing_table.nim` gets `peerCount`, which `updateRoutingTableMetrics` now reuses. `find.nim` gets `toPeerInfos`, so the tuple to `PeerInfo` conversion lives in one place. `service_discovery.nim` passes its bootstrap nodes through `initKadBase`.

No API break, so no downstream branch is needed.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

No API change. `KadDHTConfig.new` gets two new optional parameters, `fixLowPeersInterval` (default 1 minute) and `minRoutingTableSize` (default 10 peers). Existing call sites keep compiling and get the loop with the defaults.

Behaviour change: a node with fewer than 10 peers in its routing table now re-dials its bootstrap nodes and its connected peers one time per minute until it recovers. Set `minRoutingTableSize = 0` to turn the re-seed off.

New metric: `kad_routing_table_reseeds`, a counter of the re-seed attempts.

## Risk Assessment

- Backward compatibility: the defaults preserve the current behaviour for a healthy table, because the loop returns early while `peerCount() >= minRoutingTableSize`.
- Network behaviour: a node that stays below the minimum, for example a fresh node on a small network, re-seeds every minute. The dial set is bounded by the number of configured bootstrap nodes plus the connected peers, and the forced refresh is the same work that the bucket refresh already does.
- Trust: the bootstrap nodes take the unprobed path on every re-seed, and not only at construction. A dead or repointed seed can therefore re-enter the table until the liveness loop evicts it again. The peers are operator-configured, so this re-applies a trust that the operator already granted, but it is a real change from a one-time decision to a recurring one.
- Attack surface: the connected peers come from `switch.connectedPeers()`, so an inbound connection can offer itself. It still has to answer a FIND_NODE probe and pass the per-IP and per-subnet caps in `admissibleAddrs`, and the probes run behind `admissionSem`.
- Performance: the size check is a bucket walk, and it runs one time per interval.

## Tests

`tests/libp2p/kademlia/test_fix_low_peers.nim` covers five cases: a healthy table triggers no lookup, a short table forces a refresh of every non-empty bucket, an evicted bootstrap node returns to the table, a connected peer is admitted into an empty table, and the loop itself re-seeds with no explicit call.

## References

Closes #2859


libp2p/protocols/kademlia.nim
```diff
@@ -257,15 +257,53 @@ proc maintainBuckets(kad: KadDHT) {.async: (raises: [CancelledError]).} =
...
+    peers = count, minimum = kad.config.minRoutingTableSize
+  kad_routing_table_reseeds.inc()
+
+  kad.admitPeers(kad.rtable, kad.connectedPeerInfos())
+  # Seed the trusted nodes too: the probes above only land after this pass.
+  kad.updatePeers(kad.bootstrapNodes)
+
+  await kad.refreshTable(kad.rtable, forceRefresh = true)
+
+proc maintainMinPeers(kad: KadDHT) {.async: (raises: [CancelledError]).} =
```
vladopajic: it's unnecessary to raise here

vladopajic (approved):

libp2p/protocols/kademlia.nim
gmelodie: You mean the `{.async: (raises: [CancelledError]).}` part? Or the `doAssert` below it? I removed the `doAssert`

richard-ramos (approved):

libp2p/protocols/kademlia.nim
vladopajic: `raises: [CancelledError]`, because no future is waiting on that future. this just minor detail

PR #2938, opened 2026-08-14
chore(kad): per-bucket IP diversity caps

### LINKED ISSUE #2861: kad: strengthen the IP-diversity filter (ASN grouping + per-bucket caps)
`find.nim:258 hasIpDiversity` caps peers per exact IP (4) and per subnet (/24 IPv4, /64 IPv6, 10) globally across the whole table. It groups only by fixed subnet masks and enforces only table-wide caps.

go-libp2p's `peerdiversity/filter.go` groups by **ASN** for IPv6 (with a subnet fallback), and enforces caps both per-bucket (`maxPerCpl = 2`) and per-table (`maxForTable = 3`), giving stronger eclipse/Sybil resistance at bucket granularity.

Proposed work:
- Add per-bucket (per-CPL) diversity caps in addition to the existing table-wide caps.
- (Stretch) group IPv6 peers by ASN instead of a fixed /64 where ASN data is available.

Impact: eclipse/Sybil resistance. Medium (core mechanism already present, this refines it).

gmelodie:
## Summary

`hasIpDiversity` capped peers per exact IP and per subnet across the whole routing table. A table-wide cap alone lets one operator fill a single bucket: 10 peers of one /24 all land in the bucket that matters for a given key range, and the lookups over that range go to one network.

This PR adds a second scope. Every candidate is now counted twice, table-wide as before, and inside the bucket it would join (`RoutingTable.bucketIndex`). The peer is admitted when at least one of its public addresses stays under both caps. This follows go-libp2p `peerdiversity/filter.go`, which enforces `maxPerCpl` next to `maxForTable`.

New limits, next to the existing table-wide caps of 4 per IP and 10 per subnet:

- `maxPeersPerIpPerBucket`, default 2
- `maxPeersPerIpv4SubnetPerBucket`, default 3
- `maxPeersPerIpv6SubnetPerBucket`, default 3

A per-bucket field set to 0 or less means the table-wide cap is the only limit. This keeps a `KadDHTLimits` object that is built as a literal, and therefore misses the new fields, on the old behavior instead of a filter that rejects every peer with a public IP.

In-flight admission probes still count against the caps, so a burst of pending dials cannot pass the bucket cap.

The filter moved out of `find.nim` into `libp2p/protocols/kademlia/ip_diversity.nim`. `find.nim` drives the iterative lookup; the admission policy is a routing-table concern that service discovery uses as well. The IPv4 and IPv6 count loops are now one generic `admitsIp`, each peer parses its addresses one time per call instead of one time per candidate address, and the bucket index of a table peer comes from the bucket that holds it rather than from a fresh sha256 of its key.

## Affected Areas

  New `libp2p/protocols/kademlia/ip_diversity.nim`, and the admission call sites in `find.nim` (`admissibleAddrs`, `updatePeers`, `admitPeers`).

- [x] Protocol Logic
  Service-discovery tables use the same filter through `routing_table_manager.insertPeer`.

Not run. The change is limited to the Kademlia and service-discovery routing-table admission path, and it adds no wire-format or protocol change.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

API change, source level:

- `hasIpDiversity` and the switch-level `updatePeers` take one `DiversityCaps` object in place of the three `maxPeersPer*` integer parameters. Build it with `limits.diversityCaps()`, or take `defaultDiversityCaps()`.
- `hasIpDiversity` now lives in `kademlia/ip_diversity`. `import protocols/kademlia` still reaches it, since `kademlia.nim` exports the new module.
- `KadDHTLimits` gains the three per-bucket fields above. Code that builds the object with `KadDHTLimits.new` gets the defaults, and code that builds it as a literal keeps the table-wide caps alone.

Behavior change: admission is stricter. A node that ran with many peers of one subnet in one bucket now keeps fewer of them. Raise the per-bucket fields to widen this, or set them to 0 to disable the bucket scope.

Performance: the filter parses each peer's addresses one time instead of one time per candidate address, and it no longer hashes a key to find a bucket index. It is cheaper than the version it replaces.

## Risk Assessment

- Backward compatibility: source-breaking for direct callers of `hasIpDiversity` or of the switch-level `updatePeers` with positional limits. Both in-repo callers are updated. No wire or storage format changes.
- Network behavior: a legitimate cluster of nodes behind one NAT or one hosting /24 fills fewer slots of one bucket. The defaults (2 per IP, 3 per subnet) still admit small clusters.
- Security: an attacker who holds one /24 no longer fills a bucket that serves a key range.
- Known limit, unchanged by this PR: a peer is admitted when any one of its addresses clears the caps, and the address book then keeps every address it advertised. An operator who adds one address in a fresh /24 per identity still clears the filter. Closing that means admitting only the addresses that clear the caps, which changes what the address book stores for every peer. It wants its own PR.
- Performance: bounded by the table size, as before, with less work per candidate.

## References

- Closes #2861
- Parent: #1756
- go-libp2p `p2p/net/peerdiversity/filter.go`, which groups by ASN or subnet and enforces `maxPerCpl` and `maxForTable`.

## Additional Notes

The stretch item of #2861, ASN grouping for IPv6, is out of scope here. It needs an embedded IP-to-ASN dataset, which go-libp2p ships as a compiled-in table in `asnutil`. That is a data dependency for the repo to accept, so it wants its own PR.

Tests: `tests/libp2p/kademlia/test_ip_diversity.nim` covers the bucket cap through `updatePeers`, the pending-probe accounting, and the fallback when a per-bucket cap is unset. The first two fail when the bucket cap is raised, so they exercise the new path.


tests/libp2p/services/test_private_addr_filter.nim
```diff
@@ -18,6 +18,14 @@ import
   ]
 import ../../tools/[unittest, crypto, switch_builder, multiaddress]
 
+proc peerIdInBucket(kad: KadDHT, bucket: int): PeerId =
+  ## A draw lands in `bucket` with probability ~2^-(bucket+1), so keep drawing.
+  for _ in 0 ..< 1000:
+    let peerId = PeerId.random(rng()).tryGet()
+    if kad.rtable.bucketIndex(peerId.toKey()) == bucket:
+      return peerId
+  raiseAssert "no random peer id landed in bucket " & $bucket
```
vladopajic: there is same proc in this pr

vladopajic (approved):

libp2p/protocols/kademlia/find.nim
richard-ramos: should this filter the pending probes to only include those whose `tableId` matches the `rtable.selfId`?

richard-ramos (approved):

PR #2939, opened 2026-08-14
chore(kad): stop re-probing a peer whose admission probe just failed

gmelodie:
## Summary

An admission probe frees its `admissionSem` slot after `config.timeout` (5s), but the dial under it keeps running until the dialer's own timeout (30s, #2936). The candidate is therefore re-probed while the first dial is still in flight, the new probe queues on the per-peer dial lock, and it burns a probe slot without doing any work. One dead peer named by several replies can cost up to 6 dial attempts.

This PR adds a failure cache in the DHT, in the new module `libp2p/protocols/kademlia/probe_backoff.nim`:

- `KadDHT.probeFailures: Table[PeerId, ProbeFailure]` holds the consecutive failure count, the instant before which no re-probe happens, and a digest of the addresses the failed probe dialed.
- `admitPeer` records a failure when `lookupCheck` reports the peer unreachable, and clears the entry when a probe succeeds.
- `admitPeers` skips a backed-off candidate after the in-flight `admissionProbes` check and before `admissionSem.tryAcquire()`, so the slot stays free for a candidate that can use it.
- The backoff starts at `config.timeout`, doubles per consecutive failure, and stops at `config.probeBackoffMax` (default 5 minutes). A cap below the base still bounds the backoff, so no configuration combination is rejected.
- The entry suppresses a re-probe only for the address set that already failed. A reply that names the same peer with other addresses earns a probe, so a bogus address from one peer cannot keep a victim's working address out of the routing table.
- Suppression is per peer, not per routing table, so one dead peer is skipped for the main table and for every service-discovery table at once.
- `limits.maxProbeFailures` (default 1024) caps the table. Every insert first drops the entries whose backoff elapsed, then the entries closest to elapsing, which keeps the repeat offenders and discards the one-shot failures.
- The new counter `kad_admission_probes_backed_off` makes the suppression visible to operators.

`updatePeers` (the unprobed seed-peer path) and the liveness/eviction path are untouched.

## Affected Areas

- [x] Protocol Logic

Kademlia routing-table admission (`find.nim`, new `probe_backoff.nim`), plus the service-discovery tables that share the same admission path.

The change is internal to the Kademlia admission path and adds no wire, API, or configuration requirement, so no downstream branch is needed.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

No API change. `KadDHTConfig.new` gets one more optional parameter, `probeBackoffMax` (default 5 minutes), and `KadDHTLimits` gets one more field, `maxProbeFailures` (default 1024). Users of the `KadDHTLimits.new` constructor get the defaults; a user who builds the object literal field by field must add the field.

Behaviour change: a peer whose admission probe fails is not probed again on the same addresses until its backoff elapses. Fewer dials leave the node, and a dead peer no longer holds admission-probe slots that a reachable candidate can use.

## Risk Assessment

- Backward compatibility: no wire format change and no persisted state.
- Network behaviour: a peer that recovers within its backoff window waits up to 5 minutes before the next probe on the same addresses. The backoff resets on the first successful probe, and new addresses bypass it. Addresses are still recorded in the `AddressBook` before the check, so lookups and `findPeer` can dial the peer directly.
- Performance: one table lookup per candidate, and an allocation-free address digest for candidates that have a failure entry. The prune is O(n log n) and runs only when the table is at its cap.
- Security: the keys are remote-supplied peer ids, so the table is capped at `maxProbeFailures`. Tests cover the cap, the eviction order under a flood of throwaway ids, and the address-set rule that stops a remote peer from suppressing admission of an honest peer.

## References

- #2936 (probe slot freed at 5s while the dial runs to 30s)
- go-libp2p `swarm_dial.go`: per-peer dial backoff, cleared when the peerstore learns new addresses

## Additional Notes

The alternative is a dial backoff in the dialer itself: the peerstore already has the success half (`AddressBook.markConnected`) and no failure half, so a `markDialFailed` from `dialer.nim` would match go-libp2p. That changes dial behaviour for every consumer, so it belongs in its own PR.


libp2p/protocols/kademlia/probe_backoff.nim
```diff
@@ -0,0 +1,71 @@
...
+    proc(a, b: (PeerId, ProbeFailure)): int =
+      cmp(a[1].until, b[1].until)
+  )
+  for i in 0 ..< excess:
+    kad.probeFailures.del(byExpiry[i][0])
+
+proc probeRecordFailure*(kad: KadDHT, peerId: PeerId, addrs: seq[MultiAddress]) =
+  let now = Moment.now()
+  kad.probePruneFailures(now)
+  let count = kad.probeFailures.getOrDefault(peerId).count + 1
```
richard-ramos: Shouldnt the `probeFailures` be read BEFORE calling `probePruneFailures()`?

richard-ramos (approved):

vladopajic (approved):

PR #2942, opened 2026-08-14
fix(autotls): retry a failed issuance round and fix the issueRetries count

rlve:
## Summary

Two fixes in `AutotlsService`, with the tests that pin them.

**A failed issuance round ended certificate management for good.** `manageCert` read the certificate with `valueOr: return`, so a round that finished without one returned out of the `heartbeat` loop and the service went quiet until the process restarted. It now uses `withValue`, so the round falls through to the heartbeat's sleep and the next one tries again. 

**The retry loop counted wrong and slept at the wrong time.** `for _ in 0 ..< issueRetries` made `issueRetries` the total attempt count, and the `sleepAsync` sat after the final failed attempt where nothing waited on it. It now runs `for attempt in 0 .. issueRetries` with the sleep at the top of each retry.

The same loop fix clears a potential config bug. `issueRetries = 0` used to disable issuance entirely, because `0 ..< 0` never runs the body. Now 0 means one attempt and no retries, which is what the name says.

Three new cases in `tests/libp2p/autotls/test_service.nim` cover the attempt count, a failed round being retried on the next heartbeat, and a service stopped mid-issuance making no further attempt. `ACMEApiStub` gains a `stalls` flag that parks a request on a never-completing future instead of raising, which is what makes the cancellation case reachable.

## Affected Areas

- [x] Other
    AutoTLS certificate issuance and renewal (`libp2p/autotls/service.nim`).

## Impact on Library Users

`issueRetries` changes meaning. It is now the number of retries after the first attempt, not the total. At the default of 3 a failed round makes 4 ACME attempts instead of 3, and 0 now means one attempt instead of none. No API change, `AutotlsConfig.new` and its defaults are untouched.

A service that fails its first issuance now keeps trying once per `renewCheckTime` (default 1 hour) instead of stopping forever.

## Risk Assessment

One extra attempt per round, and rounds now repeat instead of stopping after the first failure.

### DIFF
--- a/libp2p/autotls/service.nim
+++ b/libp2p/autotls/service.nim
@@ -219,7 +219,9 @@ proc hasTcpStarted(switch: Switch): bool =
   switch.transports.filterIt(it of TcpTransport and it.running).len == 0
 
 proc tryIssueCertificate(self: AutotlsService) {.async: (raises: [CancelledError]).} =
-  for _ in 0 ..< self.config.issueRetries:
+  for attempt in 0 .. self.config.issueRetries:
+    if attempt > 0:
+      await sleepAsync(self.config.issueRetryTime)
     try:
       await self.issueCertificate()
       return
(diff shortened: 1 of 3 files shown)


gmelodie (approved):
Good findings!

richard-ramos (approved):

PR #2943, opened 2026-08-14
fix(kad): findNode no longer double-hashes a service table's target

### LINKED ISSUE #2846: bug: findNode double-hashes the target, so service-table refresh never targets the table's center
Split out from https://github.com/vacp2p/nim-libp2p/pull/2842#discussion_r3598456429, which fixed bucket selection

`refreshTable` opens with `discard await kad.findNode(rtable.selfId, rtable)` to keep nearby peers fresh, but for service-discovery tables that lookup targets the wrong point in the ID space. Those tables are built with `selfIdPreHashed = true` and a `serviceId` that is already a hash (`routing_table_manager.nim:47-53`), so the table's center is `serviceIdHash`. The lookup ignores that flag: `LookupState.init` (`find.nim:155-158`) ranks candidates with `xorDistance(pid, target, hasher)`, which hashes the target unconditionally (`types.nim:180-181`), so the search converges on `hash(serviceIdHash)` (an unrelated region) and the peers it was meant to refresh never get touched. `RoutingTable.findClosest` *does* honor `selfIdPreHashed` (`routing_table.nim:148-153`), so the two distance paths disagree on what a target key means, and that inconsistency is the root of the bug. Only the pre-hashed `selfId` target is affected; random targets from `randomKeyInBucket` and the bucket-peer fallback from `refreshTarget` are raw keys that the wire hashes exactly as the table does.

Two directions: keep the raw pre-image service id alongside the hashed one and use it as the lookup target (uniform wire contract), or teach the lookup about pre-hashed targets the way `findClosest` already does (removes the two-meanings-of-a-target-key split). Either way `xorDistance` and `bucketIndexFor` should stop disagreeing.

One thing to settle while fixing this: `LookupState.init` seeds its shortlist from `kad.rtable` and its hasher, ignoring the `rtable` argument threaded through `iterativeLookup` (the passed table is only used by `lookOnce` to insert discovered peers). Seeding service lookups from main-table peers may be deliberate (a fresh service table has nobody to ask) but it means the service table's own hasher and `selfIdPreHashed` never reach the ranking, so that should be an explicit decision rather than a side effect.

To repro: build a service table via `ServiceRoutingTableManager.addService`, run `refreshTable` against it, and assert the lookup ranks peers by distance to `serviceId` rather than `hash(serviceId)`.

gmelodie:
## Summary

A key had two meanings in the DHT, and a service-table refresh walked to the wrong place because of it.

`bucketIndexFor` always hashes the key it is given. `findClosest` hashed the target only when `selfIdPreHashed` was false. A service routing table sets that flag and centers on a `serviceId` that is already `sha256(serviceName)` (`routing_table_manager.nim:52`), so the two disagreed on what a target key means.

`refreshTable` opened with `discard await kad.findNode(rtable.selfId)`. For a service table that ranked candidates by distance to `hash(serviceId)`, an unrelated region, so the peers the refresh existed to keep fresh were never touched. The same call passed no `rtable`, so the walk ran on the main table and no peer it discovered reached the service table.

This PR settles on one meaning: a target key is raw and every distance path hashes it, exactly as a remote peer does when it answers the findNode. Only `selfId` may come pre-hashed.

- `findClosest` hashes the target unconditionally, so it agrees with `bucketIndexFor`.
- `RoutingTable.refreshSelfTarget` supplies the center target for one table. A raw `selfId` is its own target. A pre-hashed `selfId` has no pre-image a remote peer could hash back to it, so the known key that lands nearest the center stands in, picked by the new `nearestToCenter`.
- `KadDHT.centerTarget` holds the cross-table policy: a service table with no peers of its own borrows the main table's keys to start the walk. `refreshCenter` then runs that one findNode against `rtable`, so discovered peers land in the table being refreshed.
- `LookupState` now holds the table it drives. It seeds and ranks with that table's peers and hasher, and falls back to the main table only when the given table has nobody to ask. The issue asked for this to be an explicit decision rather than a side effect of reading `kad.rtable`.

Two directions were considered and dropped. Sending the raw pre-image needs the service *name*, which the public `lookup(disco, serviceId)` never has. Teaching the lookup about pre-hashed targets fixes only the local ranking: a remote peer answers `findNode(serviceId)` from its own main table and returns peers near `hash(serviceId)` regardless, so the walk still explores the wrong region. Per-bucket refresh targets from `randomKeyInBucket` already work because they are raw keys, and the center target now follows the same contract.

## Affected Areas

- [x] Protocol Logic

Kad-DHT routing table (`findClosest`, new `nearestToCenter` and `refreshSelfTarget`), the iterative lookup state, `refreshTable`, and the per-service routing tables that service discovery builds on them.

No wire format, message type, or codec changes. A findNode target key stays a raw key, which is what every remote peer already assumes when it hashes the key to answer.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

No public API breakage. `LookupState.init` gains a three-argument form that takes the routing table to drive; the two-argument form keeps its behavior and passes `kad.rtable`. `RoutingTable.findClosest` changes behavior for a `selfIdPreHashed` table only: it now hashes the target. Nothing in the library passed a pre-hashed target to it.

Service discovery gets the behavior fix: a service table refresh reaches the peers around the service id, and admits what it finds into that service table.

## Risk Assessment

- Backward compatibility: no protocol change, so a patched node and an unpatched node interoperate.
- Network behavior: a service table refresh now walks toward the known key nearest the service center instead of a fixed unrelated region. It reaches the center approximately, not exactly, because no random key can hash near a given center: that is a preimage search, which is why `randomKeyInBucket` caps at `MaxRefreshLeadingZeros`.
- A service table with no peers and a main table with no peers issues no self-lookup, since no raw key stands for the center. Before this PR that call ran, but it walked the wrong region on the wrong table.
- Performance: `findClosest` on a pre-hashed table now hashes the target once per call, and a pre-hashed table's center target costs one pass over its keys per refresh. The main table's keys are copied only when the service table is empty.
- Known pre-existing trap, untouched here: `ServiceId` is an unvalidated `seq[byte]` on the public `lookup`, and `xorDistance` asserts a 32-byte key, so a short service id crashes `addService` before any of this code runs. Worth its own issue.

## References

- Fixes #2846
- Split out from the bucket-selection fix in #2842 (https://github.com/vacp2p/nim-libp2p/pull/2842#discussion_r3598456429)

## Additional Notes

All 25 kademlia and service-discovery test binaries pass locally (398 tests, 0 failures). New coverage: `refreshSelfTarget` for a raw `selfId` and for a pre-hashed one, `nearestToCenter` over another table's keys, `findClosest` on a `selfIdPreHashed` table, and lookup seeding from the given table with its fallback to the main table. The mock records which routing table each findNode ran against, so the refresh tests assert the service table is the one threaded through.


libp2p/protocols/kademlia.nim
```diff
@@ -211,13 +211,26 @@ proc maintainLiveness(kad: KadDHT) {.async: (raises: [CancelledError]).} =
     else:
       idle = true
 
+proc centerTarget(kad: KadDHT, rtable: RoutingTable): Opt[Key] =
+  ## A table with no peers of its own borrows the main table's to start the walk.
```
```suggestion
  ## A table with no peers of its own borrows the main table's peers to start the walk.
```

richard-ramos (approved):
Consider adding `{.raises: [].}` to the new procs

vladopajic (approved):

PR #2947, opened 2026-08-14
feat(quic): add stream idle timeouts

richard-ramos:
## Summary

- Add configurable inbound and outbound inactivity timeouts for QUIC streams.
- Reset an expired stream without closing its containing QUIC connection.
- Refresh stream activity after successful QUIC writes.

## Affected Areas

- [x] Transports — QUIC stream lifecycle and builder configuration

## Impact on Library Users

QUIC streams now use the same five-minute default inactivity timeout as mplex and yamux. Applications can customize the values through `withQuicTransport(inTimeout, outTimeout)` or the `QuicTransport` constructors.

## Risk Assessment

Long-lived QUIC streams with no reads or writes will now be reset after the configured timeout. Active sibling streams and the underlying QUIC connection remain open.

### DIFF
--- a/libp2p/builders.nim
+++ b/libp2p/builders.nim
@@ -248,10 +248,17 @@ proc withWsTransport*(
       )
   )
 
-proc withQuicTransport*(b: SwitchBuilder): SwitchBuilder =
+proc withQuicTransport*(
+    b: SwitchBuilder,
+    inTimeout: Duration = DefaultChanTimeout,
+    outTimeout: Duration = DefaultChanTimeout,
+): SwitchBuilder =
   b.withTransport(
     proc(config: TransportConfig): Transport =
-      QuicTransport.new(config.upgr, config.privateKey, config.rng, config.connManager)
+      QuicTransport.new(
+        config.upgr, config.privateKey, config.rng, config.connManager, inTimeout,
+        outTimeout,
+      )
   )
 
 proc withMemoryTransport*(b: SwitchBuilder): SwitchBuilder =
--- a/libp2p/transports/quictransport.nim
+++ b/libp2p/transports/quictransport.nim
@@ -43,6 +43,8 @@ type
   QuicSession* = ref object of P2PConnection
     connection: QuicConnection
     streams: HashSet[QuicStream]
+    inTimeout: Duration
+    outTimeout: Duration
     when defined(libp2p_agents_metrics):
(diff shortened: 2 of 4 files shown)


gmelodie (approved):
Cool! I just wonder why different timeouts for inbound and outbound streams, and why is there need for a timeout handler (what else would someone want to do other than reset the stream?)

richard-ramos:
> why different timeouts for inbound and outbound stream

Good question. I copied the style for handling timeouts from withYamux and withMplex , but do not know the reasoning behind having separate parameters. Downstream i checked and both nimbus and delivery set it to 5m. .. i could set it as a single value but then lose consistency against other muxers, or... change all muxers so they have a single parameter, but then we end up with a breaking change :(

> why is there need for a timeout handler (what else would someone want to do other than reset the stream?)

The timeout handler is called by the connection activity monitor once it detects that the stream got idle https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/stream/connection.nim#L118-L122

gmelodie:
Claude's reasoning for the first question seems to make sense: 

> Trust asymmetry / DoS surface. Inbound streams are remote-initiated: a peer can open thousands and sit on them, consuming your stream IDs, flow control credit, and per-stream state. You want an aggressive "sent nothing useful within N seconds → reset" timeout there. Outbound streams are ones you opened with known intent — you know what you're waiting for (protocol negotiation response, DHT reply), so the timeout is really an application request timeout and often longer or governed by the caller's deadline instead.
Who's responsible for the first byte. On an outbound stream, you write first; the timeout measures peer responsiveness after your request. On an inbound stream, you're waiting for the opener to say something (e.g. multistream-select header in libp2p). A peer that opens a stream and stays silent is either buggy or hostile — no reason to be as patient as you'd be for a response to your own request that may legitimately take a while to compute.
Backpressure semantics differ. Timing out an outbound stream means failing your own operation — the caller usually wants to control that (context/deadline). Timing out an inbound stream is resource garbage collection — it's the muxer/transport's job, not the application's. Separate knobs because separate owners.

> The timeout handler is called by the connection activity monitor once it detects that the stream got idle

I meant why does this need to be a configurable handler instead of a "static" `proc`, but I guess more flexibility is better if it's all the same?

vladopajic (approved):

PR #2951, opened 2026-08-17
fix(switch): prevent accept-loop resource exhaustion

### LINKED ISSUE #2335: Bug: 100% CPU Starvation Loop on File Descriptor Exhaustion (Linux)
## Description

When `nim-libp2p` reaches the operating system's file descriptor limit (e.g., `ulimit -n`), the `accept` loop in the TCP transport layer falls into an infinite, non-yielding busy loop. This causes the node to instantly peg the CPU at 100%, starving the `chronos` async event loop and effectively deadlocking the entire application.

This is a critical stability issue for production nodes (which primarily run on Linux) under heavy connection load or targeted connection-spam attacks.

## Root Cause & Platform Differences

The bug stems from how the async engine (`chronos`) interacts with the underlying OS event polling mechanisms (`epoll` on Linux vs. `kqueue` on macOS) when the `accept()` system call fails with `EMFILE` (Too many open files).

1. **The TCP Transport `accept` Loop (`tcptransport.nim`)**
   When `tcptransport.accept()` encounters `TransportTooManyError` from Chronos, it catches the exception, logs a `debug` message, and returns `nil`.
   
2. **The Switch `accept` Loop (`switch.nim`)**
   When the `switch` receives a `nil` connection from the transport layer, it simply calls `continue` to instantly retry the `accept()` call without yielding or backing off.

3. **The Platform Difference (Why it hides on macOS but kills Linux nodes)**
   - **On Linux (`epoll`):** Chronos uses level-triggered `epoll` (`EPOLLIN`). Because there are still pending TCP connections in the kernel's listen backlog that we couldn't accept, `epoll` instantly wakes up the event loop again. The loop calls `accept()`, hits `EMFILE`, returns `nil`, and loops again immediately. This loops thousands of times a second, consuming 100% CPU.
   - **On macOS (`kqueue`):** Chronos uses `kqueue`. When `accept()` fails, Chronos removes and re-adds the socket reader. Because no *new* state change has occurred on the listen socket since it was re-added, `kqueue` does not wake the event loop. The loop naturally pauses until a new connection arrives, masking the 100% CPU bug on local Mac development machines.

## Steps to Reproduce (on Linux)

1. Checkout to this commit: https://github.com/vacp2p/nim-libp2p/tree/2a1411ddc06ccabf33b93a914ddd09a43953a7f5
2. Run `docker build -t reproduce-emfile -f Dockerfile.emfile . && docker run --rm reproduce-emfile`
3. You will see errors keep popping up: Server accept error (x1830000): [EMFILE] Too many open files in the process

## Proposed Solution

Introduce an explicit async backoff mechanism in `tcptransport.nim` when `TransportTooManyError` is caught. This yields control back to the event loop, allowing the application to process existing connections, close old ones, and eventually free up file descriptors.

**File:** `libp2p/transports/tcptransport.nim`
```diff
    except TransportTooManyError as exc:
      debug "Too many files opened", description = exc.msg
+     await sleepAsync(100.milliseconds)
      return nil
```

By adding a 100ms `sleepAsync`, we guarantee that the event loop can breathe and process other events, completely mitigating the CPU starvation deadlock.

richard-ramos:
## Summary

- acquire inbound slots after transport accept and reject excess connections without blocking the accept loop
- bound rejected-connection close concurrency and drain cleanup safely during shutdown
- back off transient nil accepts to prevent Linux file-descriptor exhaustion from causing a CPU spin
- retry watermark trimming when inbound capacity remains exhausted

Also fixes #2335.

## Affected Areas

- [x] Transports — incoming accept, rejection cleanup, and retry backoff
- [x] Peer Management / Discovery — connection limits and watermark trimming

## Impact on Library Users

Nodes at their inbound limit keep accepting without serializing on graceful closes. Repeated transient accept failures retry every 100 ms, and rejected closes use bounded backpressure.

## Risk Assessment

The accept loop waits once 32 rejected closes are in flight, bounding retained sessions while avoiding the previous per-rejection stall. Shutdown waits for those closes to finish so transport cleanup bookkeeping is not abandoned.

### DIFF
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -535,6 +535,11 @@ proc getIncomingSlot*(
     await c.inSema.acquire()
   return ConnectionSlot(connManager: c, direction: In)
 
+proc tryGetIncomingSlot*(c: ConnManager): Opt[ConnectionSlot] =
+  if c.inSema != nil and not c.inSema.tryAcquire():
+    return Opt.none(ConnectionSlot)
+  Opt.some(ConnectionSlot(connManager: c, direction: In))
+
 proc getOutgoingSlot*(
     c: ConnManager, forceDial = false
 ): ConnectionSlot {.raises: [TooManyConnectionsError].} =
@@ -783,6 +788,12 @@ proc triggerTrim(c: ConnManager) {.gcsafe, raises: [].} =
       if Moment.now() - lastTrim < wm.silencePeriod:
         return
     c.trimFut = c.trimConnections()
+
+proc triggerTrimIfNeeded*(c: ConnManager) {.gcsafe, raises: [].} =
+  ## Retries watermark trimming while the peer count remains above the high watermark.
+  c.watermark.withValue(wm):
+    if c.muxerStore.countPeers() > wm.highWater:
+      c.triggerTrim()
 
 proc drainOnCloseTasks(c: ConnManager) {.async: (raises: []).} =
   ## Drains the per-muxer onClose tasks once muxers are closed, so they finish
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -445,7 +445,7 @@ proc selectMuxer*(c: ConnManager, peerId: PeerId): Muxer =
     trace "connection not found", peerId
   return mux
 
-proc triggerTrim(c: ConnManager) {.gcsafe, raises: [].}
+proc triggerTrim*(c: ConnManager) {.gcsafe, raises: [].}
 
 proc triggerTrimAfter(
     c: ConnManager, fut: Future[void].Raising([CancelledError])
@@ -455,8 +455,7 @@ proc triggerTrimAfter(
       await fut
   except CancelledError:
     return
-  if c.watermark.isSome and c.muxerStore.countPeers() > c.watermark.get().highWater:
-    c.triggerTrim()
+  c.triggerTrim()
 
 proc storeMuxer*(
     c: ConnManager, muxer: Muxer
@@ -777,23 +776,19 @@ proc trimConnections(c: ConnManager) {.async: (raises: []).} =
 
   c.lastTrim = Opt.some(Moment.now())
 
-proc triggerTrim(c: ConnManager) {.gcsafe, raises: [].} =
+proc triggerTrim*(c: ConnManager) {.gcsafe, raises: [].} =
   ## Schedules a trim cycle if none is running and the silence period has elapsed.
   if not c.trimFut.isNil and not c.trimFut.finished:
     # trim is ongoing
     return
 
   c.watermark.withValue(wm):
+    if c.muxerStore.countPeers() <= wm.highWater:
+      return
     c.lastTrim.withValue(lastTrim):
       if Moment.now() - lastTrim < wm.silencePeriod:
         return
     c.trimFut = c.trimConnections()
-
-proc triggerTrimIfNeeded*(c: ConnManager) {.gcsafe, raises: [].} =
-  ## Retries watermark trimming while the peer count remains above the high watermark.
-  c.watermark.withValue(wm):
-    if c.muxerStore.countPeers() > wm.highWater:
-      c.triggerTrim()
 
 proc drainOnCloseTasks(c: ConnManager) {.async: (raises: []).} =
   ## Drains the per-muxer onClose tasks once muxers are closed, so they finish
--- a/libp2p/switch.nim
+++ b/libp2p/switch.nim
@@ -41,8 +41,11 @@ logScope:
 # and only if the channel has been secured (i.e. if a secure manager has been
 # previously provided)
 
-const ConcurrentUpgrades* = 32
-const UpgradeTimeout* = 30.seconds
+const
+  ConcurrentUpgrades* = 32
+  MaxRejectedConnectionCloses = ConcurrentUpgrades
+  UpgradeTimeout* = 30.seconds
+  AcceptRetryDelay = 100.millis
 
 type
   Switch* = ref object of Dial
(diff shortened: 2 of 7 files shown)


libp2p/connmanager.nim
```diff
@@ -784,6 +789,12 @@ proc triggerTrim(c: ConnManager) {.gcsafe, raises: [].} =
         return
     c.trimFut = c.trimConnections()
 
+proc triggerTrimIfNeeded*(c: ConnManager) {.gcsafe, raises: [].} =
+  ## Retries watermark trimming while the peer count remains above the high watermark.
+  c.watermark.withValue(wm):
+    if c.muxerStore.countPeers() > wm.highWater:
+      c.triggerTrim()
```
vladopajic: it might be better if `triggerTrimIfNeeded` and `triggerTrim` are the same proc. that is `triggerTrim` should be the only proc to call that will perform trim if all conditions are meet. note that triggerTrim currently is not really trimming on each call it is "if needed" trimming.

tests/libp2p/test_switch_accept.nim
```diff
@@ -57,13 +63,15 @@ suite "Switch accept-loop failure handling":
 
     check server.connManager.availableSlots(Direction.In) == 1
 
-  asyncTest "accept returning nil is non-fatal and the loop keeps retrying":
+  asyncTest "accept returning nil retries with backoff":
     let (server, transport) = newStubAcceptSwitch(NilAlways)
     startAndDeferStop(@[server])
 
-    # nil is treated as a transient miss, so the loop keeps calling accept indefinitely
+    await sleepAsync(250.millis)
```
vladopajic: why sleep?

vladopajic (approved):

libp2p/switch.nim
```diff
@@ -278,27 +282,32 @@ proc accept(s: Switch, transport: Transport) {.async: (raises: []).} =
...
+      if slot.isNone:
+        debug "Incoming connection limit reached", conn
+        s.rejectedConnCloseFuts.trackFut(conn.close())
+        s.connManager.triggerTrim()
+        if s.rejectedConnCloseFuts.len >= MaxRejectedConnectionCloses:
+          discard await one(s.rejectedConnCloseFuts)
         continue
 
-      slot.trackConnection(conn)
+      slot.get().trackConnection(conn)
```
gmelodie:
```suggestion
      s.connManager.tryGetIncomingSlot().withValue(slo):
        slo.trackConnection(conn)
      else:
        debug "Incoming connection limit reached", conn
        s.rejectedConnCloseFuts.trackFut(conn.close())
        s.connManager.triggerTrim()
        if s.rejectedConnCloseFuts.len >= MaxRejectedConnectionCloses:
          discard await one(s.rejectedConnCloseFuts)
        continue
```
either that or `valueOr` maybe

gmelodie (approved):

tests/libp2p/test_switch_accept.nim
richard-ramos: This is because if we dont wait, `acceptCalls < 10` passes imediatly.  I wanted this to ensure there are some accept calls enqueued.
Since the accept loops sleeps for 100ms, this 250ms sleep lets 2-3 calls occur, then afterwards  the next checkUntilTimeout verifies if retries continue.

 I'll change `AcceptRetryDelay` to be public and use that instead of hardcoding the 250. Also change `acceptCalls` to `<= 3`

tests/libp2p/test_switch_accept.nim
```diff
@@ -57,13 +63,15 @@ suite "Switch accept-loop failure handling":
...
     check server.connManager.availableSlots(Direction.In) == 1
 
-  asyncTest "accept returning nil is non-fatal and the loop keeps retrying":
+  asyncTest "accept returning nil retries with backoff":
     let (server, transport) = newStubAcceptSwitch(NilAlways)
     startAndDeferStop(@[server])
 
-    # nil is treated as a transient miss, so the loop keeps calling accept indefinitely
+    await sleepAsync(2 * AcceptRetryDelay + 50.millis)
+    check transport.acceptCalls <= 3
```
vladopajic, line 72:
```suggestion
    let delay = 2 * AcceptRetryDelay + 50.milliseconds
    let retries = (delay.inMilliseconds / AcceptRetryDelay.inMilliseconds).ceil
    await sleepAsync(delay)
    check transport.acceptCalls <= retries
```

libp2p/switch.nim
```diff
@@ -41,8 +41,11 @@ logScope:
 # and only if the channel has been secured (i.e. if a secure manager has been
 # previously provided)
 
-const ConcurrentUpgrades* = 32
-const UpgradeTimeout* = 30.seconds
+const
+  ConcurrentUpgrades* = 32
+  MaxRejectedConnectionCloses = ConcurrentUpgrades
+  UpgradeTimeout* = 30.seconds
+  AcceptRetryDelay* = 100.millis
```
vladopajic, line 48: optionally `AcceptRetryDelay` can be exported only with `libp2p_testing`

tests/libp2p/test_switch_accept.nim
```diff
@@ -57,13 +63,18 @@ suite "Switch accept-loop failure handling":
 
     check server.connManager.availableSlots(Direction.In) == 1
 
-  asyncTest "accept returning nil is non-fatal and the loop keeps retrying":
+  asyncTest "accept returning nil retries with backoff":
     let (server, transport) = newStubAcceptSwitch(NilAlways)
     startAndDeferStop(@[server])
 
-    # nil is treated as a transient miss, so the loop keeps calling accept indefinitely
+    await sleepAsync(2 * AcceptRetryDelay + 50.millis)
```
vladopajic:
```suggestion
```

vladopajic: @richard-ramos last suggestion was bad. it should have included this line as well

PR #2952, opened 2026-08-17
fix(TcpTransport): dual-stack reuse bug

### LINKED ISSUE #2703: transports: tcp port reuse binds the wrong address family on dual-stack nodes
### Description

A dual-stack TCP node that is `NotReachable` cannot dial peers over the address family that is not first in its listen set.

When `networkReachability` is `NotReachable`, `dial` reuses `self.addrs[0]` as the local address so the connection reuses the listen port. It uses `addrs[0]` regardless of the dial target's family. On a node listening on both families, `addrs[0]` is the IPv4 address. Dialing an IPv6 peer then binds an IPv4 local address for an IPv6 remote, and the connect fails.

This breaks TCP hole punching toward IPv6 peers on dual-stack nodes.

### Implementation

The reuse branch takes `self.addrs[0]` with no family check:

https://github.com/vacp2p/nim-libp2p/blob/32270773f35c588f0fa7e3d6b3cb281dd8471493/libp2p/transports/tcptransport.nim#L301-L305

QUIC already handles this. It selects the listener endpoint whose family matches the dial target, then falls back to a per-family dial-only endpoint:

https://github.com/vacp2p/nim-libp2p/blob/32270773f35c588f0fa7e3d6b3cb281dd8471493/libp2p/transports/quictransport.nim#L538-L548

https://github.com/vacp2p/nim-libp2p/blob/32270773f35c588f0fa7e3d6b3cb281dd8471493/libp2p/transports/quictransport.nim#L565-L572

TCP should do the same. Pick a listen address whose family matches the dial target instead of always `addrs[0]`.

vladopajic:
fixed the dual-stack TCP reuse bug:
 - TCP now selects a listening address with the same family as the dial target before reusing the listener port
 - if no matching listener exists, it dials normally

Closes: #2703

### DIFF
--- a/libp2p/transports/tcptransport.nim
+++ b/libp2p/transports/tcptransport.nim
@@ -304,14 +304,17 @@ method dial*(
 
   let ta = initTAddress(address).valueOr:
     raise (ref TcpTransportError)(msg: "Unsupported address: " & $address)
-  let local = findAddressByFamily(self.addrs, ta.family)
+  let local =
+    if self.networkReachability == NetworkReachability.NotReachable:
+      findAddressByFamily(self.addrs, ta.family)
+    else:
+      Opt.none(TransportAddress)
 
   trace "Dialing remote peer", address = $address
   let transp =
     try:
       await(
-        if self.networkReachability == NetworkReachability.NotReachable and
-            local.isSome():
+        if local.isSome():
           self.clientFlags.incl(SocketFlags.ReusePort)
           connect(ta, flags = self.clientFlags, localAddress = local.get())
         else:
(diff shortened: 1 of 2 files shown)


libp2p/transports/tcptransport.nim
```diff
@@ -294,16 +304,19 @@ method dial*(
...
   let transp =
     try:
       await(
-        if self.networkReachability == NetworkReachability.NotReachable and
-            self.addrs.len > 0:
-          let local = initTAddress(self.addrs[0]).expect("self address is valid")
+        if local.isSome():
           self.clientFlags.incl(SocketFlags.ReusePort)
-          connect(ta, flags = self.clientFlags, localAddress = local)
+          connect(ta, flags = self.clientFlags, localAddress = local.get())
```
gmelodie:
```suggestion
        local.withValue(localAddr):
self.clientFlags.incl(SocketFlags.ReusePort)
          connect(ta, flags = self.clientFlags, localAddress = localAddr)
```

gmelodie (approved):

richard-ramos (approved):

PR #2953, opened 2026-08-17
feat(addr-mgr): register the verifier and migrate the consumers

gmelodie:
## Summary

Registers the AutoNATv2 verifier with the address manager, derives a node-level reachability summary for back-compat, and migrates the existing consumers off the old paths. This is milestone 4 (last) of the [2026Q3 address-manager roadmap](https://roadmap.vac.dev/p2p/ift/2026q3-nimlibp2p-addr-manager), on top of #2924, #2928, and #2934.

What changes:

- `AutonatV2Service` registers an `AutonatV2Verifier` with `switch.addressManager` on start and clears it on stop. The manager heartbeat now drives verification; `scheduleInterval` overrides the manager's `verifyInterval`. A change to the verifier or the interval restarts the heartbeat, so a service which starts after the manager takes effect immediately instead of after one stale default interval.
- The v1-style confidence `Deque`, the `askPeer`/`askConnectedPeers` probe loop, and the `libp2p_autonat_v2_reachability_confidence` metric are retired. `AutonatV2Service.networkReachability` is now a read-only view of the manager summary.
- A `Verifier` concludes on one address per call: `verify(address): Opt[AddrState]`. The manager owns the loop, applies each verdict as it lands, and stops the run at `verifyTimeout`. A run which runs out of time keeps the verdicts it already collected, and the addresses it did not reach wait for the next run.
- `verifyInterval` is the cadence only. The new `AddressManagerConfig.verifyTimeout` (default 2 minutes) caps one run. Before this, one run had to fit in one interval, so a short schedule such as the interop harness's one second cancelled valid dial requests.
- `AddressManager.reachability()` derives `NetworkReachability` from the per-address states: any confirmed candidate means `Reachable`, else any unreachable one, or any refuted guess, means `NotReachable`. An `onReachabilityChange` hook (one slot) fires after a verify run changes the summary and feeds the reachability observers. A cancelled handler call is retried on the next run instead of being lost.
- Announce rule: only confirmed or unverified candidates announce. A verified-unreachable candidate stays out when it is publicly routable; a LAN or loopback address keeps announcing, because a remote peer proves nothing about an address it was never going to reach. The explicit announce list still wins. The verifier skips relayed candidates, so a server which refuses circuit dials cannot kill the one relay address a NAT'd node has.
- The manager also drops a relay address once a direct address of the same IP family is confirmed. The rule runs after the mapper chain, so a mapping which expires during a pass restores the relay address in that same pass.
- `enableDialableCandidates` moves into the manager: each verify run derives `AddrSource.Identify` candidates from the identify observations. An `Identify`-only candidate announces only after confirmation (it is remote input). A refuted guess is removed and remembered in a ring capped by `maxSize`, so it is neither re-derived nor re-dialed on the next runs. The entry expires after 12 runs, so a wrong refutation heals.
- `hpservice` feeds dcutr from `confirmedAddrs()` first (per-family detail), then the observation quorum, then the announce set.
- The deprecated `withObservedAddrManager` builder hooks are removed.

## Affected Areas

- [ ] Gossipsub
- [ ] Transports
- [x] Protocol Logic
- [ ] Build / Tooling
- [ ] Other

Address manager, AutoNAT v2 service, AutoRelay, hole punching, switch builder, AutoNATv2 interop harness.

- **Nimbus:** not run yet
- **Waku:** not run yet
- **Codex:** not run yet

## Impact on Library Users

- `SwitchBuilder.withObservedAddrManager` is removed; use `withAddressManager(AddressManagerConfig(...))`. Injecting a manager instance is no longer possible.
- `Verifier.verify` takes one address and returns `Opt[AddrState]`; the `AddrVerdict` type is removed. A custom verifier must be updated.
- `AddressManagerConfig` gains `verifyTimeout`. A short `verifyInterval` no longer shortens a run.
- `AddressManager.add` loses its `state` parameter. A verifier sets the state.
- `AutonatV2ServiceConfig.new` loses `askNewConnectedPeers`, `numPeersToAsk`, `maxQueueSize`, and `minConfidence`. `scheduleInterval`, `enableAddressMapper`, and `enableDialableCandidates` stay.
- `AutonatV2Service.networkReachability` changes from a field to a derived proc; the call syntax is unchanged.
- Reachability handlers still subscribe through `reachabilityObservers.add`. Under AutoNAT v2 the `confidence` argument is now always `Opt.none(float)` and `dialBackAddr` is no longer reported; read `switch.addressManager.confirmedAddrs()` instead. A handler which requires `confidence.isSome()` never fires under v2 (the interop harness had exactly this and is updated here).
- The `libp2p_autonat_v2_reachability_confidence` metric is gone.

## Risk Assessment

- A verdict comes from one random peer per address, so a single wrong answer flips a state until a later run heals it. The retired confidence deque smoothed this; the per-address model re-verifies on every heartbeat instead. There is no quorum: a hostile server can hold one public address out of the announce set for one interval, and one refuted guess pins the summary to `NotReachable` while nothing else is confirmed. The failure direction is safe, because the node then reaches for relays and hole punching.
- The observation ring counts addresses, not peers. A peer which reconnects `minCount` times owns the derived guess, which `enableDialableCandidates` turns into a dial request. The address policy gates the target, and the option is off by default.
- AutoNAT v1 and hole punching keep their existing confidence path; this PR does not touch them beyond the dcutr candidate preference.

## References

- Roadmap: https://roadmap.vac.dev/p2p/ift/2026q3-nimlibp2p-addr-manager
- Prior milestones: #2924, #2928, #2934

## Additional Notes

The integration test for the announce pipeline (identify + upnp + circuit sources, verify, announce) lives in `tests/libp2p/test_address_manager.nim` under "AddressManager verify and announce".


tests/libp2p/services/test_autorelay.nim
```diff
@@ -90,6 +91,49 @@ suite "Autorelay":
 
     await allFutures(switchClient.stop(), switchRelay.stop())
 
+  asyncTest "a confirmed direct address withdraws only the same-family relay address":
+    switchRelay = createSwitch(Relay.new())
+    relayClient = RelayClient.new()
+    let fut = newFuture[void]()
+    proc checkMA(addresses: seq[MultiAddress]) =
+      if not fut.finished:
+        fut.complete()
```
vladopajic: `fut.completeOnce()`

tests/libp2p/services/test_autorelay.nim
```diff
@@ -90,6 +91,49 @@ suite "Autorelay":
...
+
+    autorelay = AutoRelayService.new(3, relayClient, checkMA, rng())
+    switchClient = createSwitch(relayClient, autorelay)
+    await allFutures(switchClient.start(), switchRelay.start())
```
vladopajic: `startAndDeferStop` form /tests/tools/lifecycle.nim

libp2p/protocols/connectivity/autonatv2/service.nim
```diff
@@ -86,197 +58,74 @@ proc new*(
...
+    return await addressMapper(self, listenAddrs)
 
 method start*(
     self: AutonatV2Service, switch: Switch
 ) {.async: (raises: [CancelledError]).} =
   trace "Running AutonatV2Service"
 
-  switch.connManager.addPeerEventHandler(
-    self.newConnectedPeerHandler, PeerEventKind.Joined
-  )
```
richard-ramos, line 256: Since this part of the code was removed,  wouldnt this mean that determining reachability would be `delayed` for the duration of the heartbeat?

Since the heartbeat initializes in `start()`, then that means it will immediatly run, and it does at a moment no peer is connected.

Is this intentional? because the effect is for the node to remain `Unknown` for a while.

tests/libp2p/services/test_autorelay.nim
```diff
@@ -90,6 +91,49 @@ suite "Autorelay":
...
+    await switchClient.connect(switchRelay.peerInfo.peerId, switchRelay.peerInfo.addrs)
+    await fut.wait(1.seconds)
```
vladopajic: why is there timeout?

tests/libp2p/test_address_manager.nim
```diff
@@ -369,6 +378,195 @@ suite "AddressManager verification":
...
 
+  asyncTest "an interval and a verifier set after start take effect without a stale wait":
+    let
+      address = ma("/ip4/1.2.3.4/tcp/1")
+      verifier =
+        makeStubVerifier(@[AddrVerdict(address: address, state: AddrState.Confirmed)])
+      # the default interval is minutes; a stale first sleep would outlive the test
+      manager = AddressManager.new()
+
+    manager.start()
```
vladopajic: it has started to become very smelly how address manager is created. 
- sometimes it is created without starting, but usually it is stated?
- there are two makers, that do the same thing, `makeVerifyingManager` is essentially `makeStartedManager` with new arg.
- see lifecycle from before, this can be concept for anything that can start() and stop().

vladopajic (approved):

libp2p/services/autorelayservice.nim
```diff
@@ -28,10 +29,29 @@ type
...
   return self.running
 
+func confirmedIpFamilies(manager: AddressManager): set[IpAddressFamily] =
+  var families: set[IpAddressFamily]
+  for address in manager.confirmedAddrs():
+    if address.contains(multiCodec("p2p-circuit")).get(false):
+      continue
+    let ip = address.getIp().valueOr:
+      continue
+    families.incl(ip.family)
```
richard-ramos: I wonder if we should check `isPublicMA()` here before adding to `families`. Because as it is right now, if you connected to a node on a private address, you'd remove all other relay addresses, i.e.

Imagine you have a confirmed address: 
`/ip4/192.168.1.20/tcp/4001`

And a relay address:
`/ip4/1.2.3.4/tcp/4001/p2p/relayPeerId/p2p-circuit`

If a peer on the same LAN dials your node on `192.168.1.20`, this address would get confirmed,  thus the `confirmedIpFamilies` would contain `IPv4`

Then, for the relay address, in `stillNeeded()` you only check for the ip family with `ip.family notin confirmed`.

But since the relay also uses `IPv4`, the condition would be `false`, so the relay address would get removed

libp2p/address_manager.nim
```diff
@@ -389,26 +435,104 @@ proc resolveMapper(self: AddressManager): AddressMapper =
...
-      await self.verifier.verify(toSeq(self.candidates.keys)).wait(self.verifyInterval)
+      await self.verifier.verify(addresses).wait(self.verifyInterval)
     except AsyncTimeoutError:
       debug "Verifier ran out of time", interval = self.verifyInterval
       return
 
-  # update drops a verdict on an address which went away while the verifier ran
-  for verdict in verdicts:
-    self.update(verdict.address, verdict.state)
+  if not self.applyVerdicts(verdicts):
```
richard-ramos: Seems to me that this will cause observers to not know about reachability's changes, like for example,

if your node was reachable, but at some point the nat mapper stops returning an address, the logic says that the address manager `withdraw()` will delete the address, and `reachability()` will return `Unknown` from that point onwards... however, since in here we do not let `notifyReachability()` be called, the observers will still think the node is `Reachable`

tests/tools/lifecycle.nim
```diff
@@ -28,3 +28,10 @@ template startAndDeferStop*[T](nodes: seq[T]): untyped =
   await startNodes(nodes)
   defer:
     await stopNodes(nodes)
+
+template startAndDeferStop*(
+    manager: AddressManager, peerInfo: PeerInfo = nil
+): untyped =
+  manager.start(peerInfo)
```
vladopajic: starting address menager with optional peer info is bad design. it should have been:
`.setPeerInfo(..)` followed by `.start()` in cases when peer info should be set. the advantage is that start and stop interface will match other logic

richard-ramos (approved):

PR #2960, opened 2026-08-18
chore: add linter for nix dependencies

richard-ramos:
Adds a linter that determines whether the nix lock file dependencies drifted away from the versions specified in libp2p.nimble


vladopajic (approved):

gmelodie (approved):
a mazing

Issue #950, opened 2023-09-18
Improve GossipSub publish API


Menduist:
Currently, the GS.publish returns the amount of peers the data was sent to.

This data is actually not useful, and when we return 0, the caller can't really know why

We should change this API to enable us to return different errors for:
- message too big
- no peers on topic
- message already seen
- Other cases?

However, `publish` is a stable API that we can't break, so we'll need to either add a `tryPublish`, or something smart like that

arnetheduck:
> message too big

This is essentially a precondition that the caller knows about - I don't think `publish` is the right spot for this kind of error checking.

> message already seen

From the point of view of `publish`, a semantically sound option is to return the number of peers in the mesh and/or fanout here - conceptually, this reframes the return value as "number of peers that reasonably can be expected to have seen the message, already or soon" which is probably more useful to the caller, as an indication of "quality of publishing". This can be done without changing the API though it's a small semantic adjustment to the function itself.

PR #2965, opened 2026-08-18
refactor(pubsub): separate app's path from protocol messages send/brodcast path 

### LINKED ISSUE #1674: pubsub: improve splitting messages
- we should remove splitting messages when message is published by application
  - as per comment https://github.com/vacp2p/nim-libp2p/pull/1665#issuecomment-3266425444; application is in charge of sending messages so it will know how to handle if message is larger then configured
  -  publishing this kind of message can throw error 
- however, messages sent as protocol responses should be split so that they do not exceed `maxMessageSize`
  -  naive approach is to send each message individually as it will defiantly be less the configured 

note should refactor internals of sending messages to clearly separate flow when message is sent from application and when message is sent as response (to protocol).

vladopajic:
- separated app messages path for broadcast/send message from path that sends protocol messages (aka responses)
- app messages are not being spitted anymore (dropping message if size exceeds)
- protocol messages are being splinted and batched up to max message size

closes: #1674

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -881,6 +881,15 @@ method publish*(
     debug "Empty topic, skipping publish"
     return 0
 
+  let msg =
+    if g.anonymize:
+      Message.init(Opt.none(PeerInfo), data, topic, Opt.none(uint64), false)
+    else:
+      inc g.msgSeqno
+      Message.init(Opt.some(g.peerInfo), data, topic, Opt.some(g.msgSeqno), g.sign)
+
+  g.handleSelfPublishing(topic, data)
+
   trace "Publishing message on topic", data = data.shortLog
 
   let pubParams = publishParams.get(PublishParams())
@@ -906,17 +915,10 @@ method publish*(
     libp2p_gossipsub_failed_publish.inc()
     return 0
 
-  let
-    msg =
-      if g.anonymize:
-        Message.init(Opt.none(PeerInfo), data, topic, Opt.none(uint64), false)
-      else:
-        inc g.msgSeqno
-        Message.init(Opt.some(g.peerInfo), data, topic, Opt.some(g.msgSeqno), g.sign)
-    msgId = g.msgIdProvider(msg).valueOr:
-      trace "Error generating message id, skipping publish", error = error
-      libp2p_gossipsub_failed_publish.inc()
-      return 0
+  let msgId = g.msgIdProvider(msg).valueOr:
+    trace "Error generating message id, skipping publish", error = error
+    libp2p_gossipsub_failed_publish.inc()
+    return 0
 
   logScope:
     msgId = shortLog(msgId)
@@ -950,8 +952,6 @@ method publish*(
       g.extensionsState.preambleBroadcast(
         RPCMsg.withPreamble(topic, msgId, msg.data.len), peers.mapIt(it.peerId)
       )
-
-  g.handleSelfPublishing(topic, data)
 
   g.broadcast(
     peers,
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -926,11 +926,7 @@ method publish*(
   # effects occur.
   let encodedSize = encode(RPCMsg.withMessages(msg), g.anonymize).len
   if encodedSize > g.maxMessageSize:
-    raise newException(
-      MessageTooLargeError,
-      "message of size " & $encodedSize & " exceeds maxMessageSize " &
-        $g.maxMessageSize,
-    )
+    raise MessageTooLargeError.new(encodedSize, g.maxMessageSize)
 
   if g.addSeen(g.salt(msgId)):
     # If the message was received or published recently, don't re-publish it -
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -881,6 +881,15 @@ method publish*(
     debug "Empty topic, skipping publish"
     return 0
 
+  # Application-published messages are never split - reject oversized messages
+  # up front so the caller can handle the error before any dedup/cache side
+  # effects occur.
+  let messageSize = RPCMsg.withMessages(msg).byteSize()
+  if messageSize > g.maxMessageSize:
+    warn "message exceeds maximum message size",
+      messageSize, maxMessageSize = g.maxMessageSize
+    return 0
+
   handleSelfPublishing(g, topic, data)
 
   trace "Publishing message on topic", data = data.shortLog
@@ -924,15 +933,6 @@ method publish*(
     msgId = shortLog(msgId)
 
   trace "Created new message", message = shortLog(msg), peers = peers.len
-
-  # Application-published messages are never split - reject oversized messages
-  # up front so the caller can handle the error before any dedup/cache side
-  # effects occur.
-  let encodedSize = encode(RPCMsg.withMessages(msg), g.anonymize).len
-  if encodedSize > g.maxMessageSize:
-    warn "message exceeds maximum message size",
-      encodedSize, maxMessageSize = g.maxMessageSize
-    return 0
 
   if g.addSeen(g.salt(msgId)):
     # If the message was received or published recently, don't re-publish it -
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -16,6 +16,7 @@ import
   ./peertable,
   ./mcache,
   ./timedcache,
+  ./rpc_send,
   ./rpc/[messages, message, protobuf],
   ../protocol,
   ../../stream/connection,
@@ -29,6 +30,9 @@ import
   ../../utils/future
 
 export types, scoring, behavior, pubsub, results
+
+when defined(libp2p_testing):
+  export sendResponse, broadcastResponse
 
 logScope:
   topics = "libp2p gossipsub"
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -881,17 +881,6 @@ method publish*(
     debug "Empty topic, skipping publish"
     return 0
 
-  # Application-published messages are never split - reject oversized messages
-  # up front so the caller can handle the error before any dedup/cache side
-  # effects occur.
-  let messageSize = RPCMsg.withMessages(msg).byteSize()
-  if messageSize > g.maxMessageSize:
-    warn "message exceeds maximum message size",
-      messageSize, maxMessageSize = g.maxMessageSize
-    return 0
-
-  handleSelfPublishing(g, topic, data)
-
   trace "Publishing message on topic", data = data.shortLog
 
   let pubParams = publishParams.get(PublishParams())
@@ -941,6 +930,15 @@ method publish*(
     trace "Dropping already-seen message"
     return 0
 
+  # Application-published messages are never split - reject oversized messages
+  # up front so the caller can handle the error before any dedup/cache side
+  # effects occur.
+  let messageSize = RPCMsg.withMessages(msg).byteSize()
+  if messageSize > g.maxMessageSize:
+    warn "message exceeds maximum message size",
+      messageSize, maxMessageSize = g.maxMessageSize
+    return 0
+
   if not pubParams.skipMCache:
     g.mcache.put(msgId, msg)
 
@@ -952,6 +950,8 @@ method publish*(
       g.extensionsState.preambleBroadcast(
         RPCMsg.withPreamble(topic, msgId, msg.data.len), peers.mapIt(it.peerId)
       )
+
+  g.handleSelfPublishing(topic, data)
 
   g.broadcast(
     peers,
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -888,6 +888,15 @@ method publish*(
       inc g.msgSeqno
       Message.init(Opt.some(g.peerInfo), data, topic, Opt.some(g.msgSeqno), g.sign)
 
+  # Application-published messages are never split - reject oversized messages
+  # up front so the caller can handle the error before any dedup/cache side
+  # effects occur.
+  let messageSize = RPCMsg.withMessages(msg).byteSize()
+  if messageSize > g.maxMessageSize:
+    warn "message exceeds maximum message size; message will not be published",
+      messageSize, maxMessageSize = g.maxMessageSize
+    return 0
+
   g.handleSelfPublishing(topic, data)
 
   trace "Publishing message on topic", data = data.shortLog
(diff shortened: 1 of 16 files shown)


libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -294,6 +301,30 @@ proc broadcast*(
...
+    sendPeers: openArray[PubSubPeer],
+    msg: RPCMsg,
+    priority: MessagePriority,
+    useCustomStream: bool = false,
+) {.raises: [MessageTooLargeError].} =
+  ## Sends an application-published `msg` (of type `RPCMsg`) to a specified
+  ## group of peers in the PubSub network.
+  ##
+  ## The message is sent as-is, without splitting - the application is
+  ## responsible for ensuring the message fits within `maxMessageSize`. Raises
```
gmelodie:
```suggestion
```

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -302,10 +333,64 @@ proc broadcast*(
...
+    msg: RPCMsg,
+    priority: MessagePriority,
+    useCustomStream: bool = false,
+) {.raises: [].} =
+  ## Sends a protocol response `msg` (of type `RPCMsg`) to a specified group of
+  ## peers. This is an internal, protocol-facing broadcast path - messages that
+  ## together exceed `maxMessageSize` are split and sent individually.
+  ##
+  ## This should not be used for application-published messages; use `broadcast`
+  ## instead.
```
gmelodie:
```suggestion
  ## peers.
```

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -614,47 +615,35 @@ proc sendEncoded*(
...
     anonymize: bool,
     priority: MessagePriority,
     useCustomStream: bool = false,
-) {.raises: [].} =
-  ## Asynchronously sends an `RPCMsg` to a specified `PubSubPeer` with an option for anonymization.
+) {.raises: [MessageTooLargeError].} =
+  ## Sends an application-published `RPCMsg` to this peer, without splitting.
+  ##
+  ## The message is sent as-is - the application is responsible for ensuring it
+  ## fits within `maxMessageSize`. Raises `MessageTooLargeError` if the encoded
```
gmelodie:
```suggestion
  ## Raises `MessageTooLargeError` if the encoded
```

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -663,34 +652,37 @@ proc send*(
...
+    msg: RPCMsg,
+    anonymize: bool,
+    priority: MessagePriority,
+    useCustomStream: bool = false,
+) {.raises: [].} =
+  ## Sends a protocol response `RPCMsg` to this peer, splitting multi-message
+  ## responses into individual messages so that none exceed `maxMessageSize`.
+  ##
+  ## This is an internal, protocol-facing send path - individual messages that
+  ## still exceed `maxMessageSize` are silently dropped.
```
gmelodie, line 673: > individual messages that still exceed `maxMessageSize` are silently dropped

Why is that? Can we not assume that code calling this proc should know what it's doing?

tests/libp2p/pubsub/component/test_floodsub.nim
```diff
@@ -256,7 +256,12 @@ suite "FloodSub Component":
     checkUntilTimeout:
       messageReceived == 2
 
-    check (await smallNode.publish(topic, bigMessage)) > 0
+    # Publishing a message larger than the configured maxMessageSize raises.
```
gmelodie: no shit
```suggestion
```

gmelodie (approved):
This makes me wonder if we should spend more time in refactoring to have better error reporting and handling. Maybe some interesing self-healing cases appear...

libp2p/protocols/pubsub/pubsubpeer.nim
vladopajic: > Why is that

messages are sent individually, but individual message can still theoretically exceed message size.
these messages are protocol driven messages, which are less likely to exceed messages size but if they do message will be dropped.

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -663,34 +652,37 @@ proc send*(
...
+
+  if msg.messages.len > 1:
+    for m in msg.messages:
+      var single = RPCMsg(messages: @[m])
```
richard-ramos: What happens with other fields the `msg` might have, like subscriptions, extensions, etc?

libp2p/protocols/pubsub/gossipsub.nim
richard-ramos: The validation for msg size happens 'too late'. If you have `triggeSelf` enabled, local handlers will be triggered first before the msg size validation even happens

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -921,6 +921,13 @@ method publish*(
 
   trace "Created new message", message = shortLog(msg), peers = peers.len
 
+  # Application-published messages are never split - reject oversized messages
+  # up front so the caller can handle the error before any dedup/cache side
+  # effects occur.
+  let encodedSize = encode(RPCMsg.withMessages(msg), g.anonymize).len
```
richard-ramos: You're doing encoding twice. Once here and once in broadcast. Couldn't `msg.byteSize()` be used instead? Because this introduces a performance regression and `publish*` is a hot path

richard-ramos: this is happening in floodsub too

libp2p/protocols/pubsub/gossipsub.nim
richard-ramos: this is happening in floodsub too

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -302,10 +333,64 @@ proc broadcast*(
...
+    p: PubSub,
+    sendPeers: HashSet[PubSubPeer],
+    msg: RPCMsg,
+    priority: MessagePriority,
+    useCustomStream: bool = false,
+) {.raises: [MessageTooLargeError].} =
+  ## Overload for `HashSet[PubSubPeer]`: order is irrelevant for a broadcast,
+  ## so the peers are materialized into a sequence and delegated to the
+  ## `openArray` overload.
+  p.broadcast(sendPeers.toSeq, msg, priority, useCustomStream)
```
richard-ramos: using this toSeq introduces an extra allocation. Couldnt this use an iterator like we had before?

richard-ramos: I see this happening in `broadcastResponse` too

libp2p/protocols/pubsub/pubsubpeer.nim
richard-ramos: This has an effect over IWANT responses, because previously in a single response you would contain multiple pubsub messages, but now, with this code you are sending 1 message per IWANT.

And it is a bad bug, because you can ~~receive~~ send an IHAVE that contains 5k message ids  (there's a const `IHaveMaxLength = 5000`), thus,  we could potentially attempt to enqueue 5k messages now, and from msg 1024 onwards, those would get dropped, since low priority queue does not allow more than those in queue

richard-ramos (changes requested):
Some additional points:

1. Comments from @arnetheduck that should be taken into account with regards `publish` raising exceptions:

 https://github.com/vacp2p/nim-libp2p/pull/1243:
> the original sin here is that publish allows raising errors at all - publish is a "broadcast" operation, ie "fire-and-forget" - its nature is such that it cannot fail by definition and thus shouldn't be raising anything at all

https://github.com/vacp2p/nim-libp2p/issues/950: 
> > message too big
> This is essentially a precondition that the caller knows about - I don't think publish is the right spot for this kind of error checking.

----

2. This PR should be tested in the nimbus fleet for libp2p by updating this branch https://github.com/status-im/nimbus-eth2/tree/nim-libp2p-auto-bump-unstable so vendor/nim-libp2p points to this commit, as it is introducing behavioral changes in pubsub.

libp2p/protocols/pubsub/gossipsub.nim
vladopajic: this was true for other edge cases (early returns). trigger self was first thing that was happening. there were other cases that could return early, triggering self while not publishing. 

i have now changed behavior so that trigger self is happening after all checks are done. but note this is different behavior then before - i think it makes sense to be this way (new way).

libp2p/protocols/pubsub/pubsub.nim
arnetheduck: +1 - this is important, broadcasting is a hotspot and auto was the chosen solution to avoid pointless allocations - ditto all other hot path mechanisms like the avoidance of future allocations, using raw async for the common case etc

libp2p/protocols/pubsub/pubsubpeer.nim
vladopajic: i was thinking to add this in next pr, but could be here as well https://github.com/vacp2p/nim-libp2p/pull/2965/commits/af516e082e8e36509847015394a3c8f2686ccedb

vladopajic: ^ that commit only does batching of messages. IHAVE, IWANT etc, are sent as part of `RPCMsg.control` which was never touched.

vladopajic: now it's obsolete, but was good point

libp2p/protocols/pubsub/pubsub.nim
vladopajic: changed argument to union type
https://github.com/vacp2p/nim-libp2p/pull/2965/commits/859e6df7de8e55359951994ec5490c0aff4cbbea

vladopajic:
nimbus pipeline passed https://ci.status.im/blue/organizations/jenkins/nimbus-eth2%2Fplatforms%2Flinux%2Fx86_64%2Fmain/detail/PR-7014/10/pipeline/

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -663,34 +647,56 @@ proc send*(
...
+      if batch.messages.len > 0 and batch.byteSize() + m.byteSize() > maxBatchSize:
+        # The batch is full - send it and start a new one.
+        var encoded = encodeRpcMsg(p, batch, anonymize)
+        trace "sending response msg to peer", peer = p, rpcMsg = shortLog(batch)
+        p.trackSend(p.sendEncoded(move(encoded), priority, useCustomStream))
+
+        batch = RPCMsg()
+
+      batch.messages.add(m)
+
```
richard-ramos, line 702:
```suggestion
    var batch = RPCMsg()
    var batchSize = 0

    for m in msg.messages:
      let msgSize = m.byteSize()
      if batch.messages.len > 0 and batchSize + msgSize > maxBatchSize:
        # The batch is full - send it and start a new one.
        var encoded = encodeRpcMsg(p, batch, anonymize)
        trace "sending response msg to peer", peer = p, rpcMsg = shortLog(batch)
        p.trackSend(p.sendEncoded(move(encoded), priority, useCustomStream))

        batch = RPCMsg()
        batchSize = 0

      batch.messages.add(m)
      batchSize += msgSize

```

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -940,6 +951,8 @@ method publish*(
         RPCMsg.withPreamble(topic, msgId, msg.data.len), peers.mapIt(it.peerId)
       )
 
+  g.handleSelfPublishing(topic, data)
```
richard-ramos: `handleSelfPublishing` should be executed even if there are no peers and moving this line here can cause that the return on line 907 stops it from running. `publish` should work even if there are no peers. Basically code should go like this:

1. Validate the topic
2. Validate the encoded size
3. Self publish
4. Construct the signed/anonymized msg
5. Select remote peers
6. Generate msgid, update seen cache, and broadcast

This also means that moving the encoded size validation from line 933 should probably be moved before the `addSeen` to not pollute the seen cache with a msg that we already know we wont publish.

(somethign similar needs to be done in floodsub)

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -663,34 +647,56 @@ proc send*(
...
+  ## still exceed `maxMessageSize` are silently dropped.
+
+  if msg.messages.len > 1:
+    # Batch as many messages as possible into a single encoded RPC instead of
+    # encoding one RPC per message. 
+    # Batches are sized with the cheap `byteSize` estimate, leaving 10% headroom 
+    # for protobuf encoding overhead on top of the raw payload sizes.
+    let maxBatchSize = (p.maxMessageSize * 90) div 100
+
+    var batch = RPCMsg()
```
richard-ramos, line 683: The problem i mentioned before of other fields being discarded still happens, although re-reviewing the code part of it did happen before this PR but in a different shape:

Previously:
- If the encoded rpc was < max msg size and contained only 1msg, it would be sent as is, thus preserving any other field the rpc could have.
- But if it was oversized and had multi msg, then the RPC was rebuilt in `splitRPCMsg` and rebuilt and it was then when the fields got lost

Currently: 
- All RPCs regardless of their size, as long as they have more than one msg, they will lose the other fields.

So. This is in practice not a problem we are suffering from since at the moment I dont see cases in which we send multiple messages that also contain extension/control messages but, it is still a deficiency we have on the code that can bite us in the future. 

Since this is preexisting (altho its effect increased with this PR) we should probably add a comment like this so we don't forget (and maybe an issue to investigate)

```suggestion
    var batch = RPCMsg()
    
    # Multi-message batching currently supports message-only RPCs.
    # If an extension is associated to multiple messages, it would require
    # splitting associated to the extension itself
```

libp2p/protocols/pubsub/pubsubpeer.nim
vladopajic: unless, there is hard opinion, i would rather keep the original code as it is simpler.
the performance difference is neglectable.

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -663,34 +647,57 @@ proc send*(
...
+) {.raises: [].} =
+  ## Sends a protocol response `RPCMsg` to this peer, batching multi-message
+  ## responses together so that none exceed `maxMessageSize`.
+  ## Batching is only performed for message-only RPCs - if the RPC carries other
+  ## data (subscriptions, control or extensions), it is sent as-is.
+  ##
+  ## This is an internal, protocol-facing send path - individual messages that
+  ## still exceed `maxMessageSize` are silently dropped.
+
+  if msg.messages.len > 1 and msg.hasOnlyMessages():
```
richard-ramos, line 680: Amazing!

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -663,34 +647,57 @@ proc send*(
...
+    for m in msg.messages:
+      batch.messages.add(m)
+
+      if batch.messages.len > 0 and batch.encodedSize() > p.maxMessageSize:
+        discard batch.messages.pop() # Remove last element
+
+        # The batch is full - send it and start a new one.
+        var encoded = encodeRpcMsg(p, batch, anonymize)
+        trace "sending response msg to peer", peer = p, rpcMsg = shortLog(batch)
+        p.trackSend(p.sendEncoded(move(encoded), priority, useCustomStream))
```
richard-ramos:
```suggestion
        if batch.messages.len > 0:
          var encoded = encodeRpcMsg(p, batch, anonymize)
          trace "sending response msg to peer", peer = p, rpcMsg = shortLog(batch)
          p.trackSend(p.sendEncoded(move(encoded), priority, useCustomStream))
```

richard-ramos (approved):
Let's see it in action during next week in the libp2p nodes on nimbus hoodi fleet 👍

Issue #2957, opened 2026-08-18
autotls: no certificate is issued, and a node listening on wss hangs in switch.start


rlve:
## Description

Two separate defects, on either side of the same handoff. Together they mean autotls issues nothing and a node configured for wss never finishes starting. Fixing one leaves the other, so both need to be addressed before autotls works.

## 1. The service returns before it schedules any issuance

`AutotlsService.start` checks for a `TcpTransport` that is already running:

https://github.com/vacp2p/nim-libp2p/blob/1c6dcfdf03c362b2c62a4b1c15881a58e590dae0/libp2p/autotls/service.nim#L215-L216

https://github.com/vacp2p/nim-libp2p/blob/1c6dcfdf03c362b2c62a4b1c15881a58e590dae0/libp2p/autotls/service.nim#L240-L242

`Switch.start` collects every service future before any transport future, and a chronos body runs synchronously up to its first await, so no transport is running when that check runs:

https://github.com/vacp2p/nim-libp2p/blob/1c6dcfdf03c362b2c62a4b1c15881a58e590dae0/libp2p/switch.nim#L403-L411

Every switch built with `.withAutotls()` logs `Could not find a running TcpTransport in switch` and stops there. No ACME request is made and no certificate is ever issued.

The check was correct when it was written in #1535. Back then `Switch.start` started the transports first and awaited the TCP one before running any service:

```nim
      if t of TcpTransport:
        await fut
        ...
  # some transports require some services to be running
  # in order to finish their startup process
  for service in s.services:
    discard await service.setup(s)
```

#2462 (`81ac39fd6`, 2026-05-14) reversed that order and carried the check over unchanged, so it has measured nothing since.

## 2. The wait for the certificate is unbounded

`WsTransport.start` bounds the wait for the service to be running with `DefaultAutotlsWaitTimeout` (3 seconds) and then waits for the certificate with no bound at all:

https://github.com/vacp2p/nim-libp2p/blob/1c6dcfdf03c362b2c62a4b1c15881a58e590dae0/libp2p/transports/wstransport.nim#L330-L345

`getCertWhenReady` waits on `certReady`, and the service fires that event in exactly one place, right after installing a certificate. An issuance that never succeeds leaves the wait pending forever:

https://github.com/vacp2p/nim-libp2p/blob/1c6dcfdf03c362b2c62a4b1c15881a58e590dae0/libp2p/autotls/service.nim#L82-L86

[...]

PR #2966, opened 2026-08-19
test(autotls): certificate issuance

rlve:
## Summary

Adds test coverage for AutoTLS certificate issuance in `autotls/test_service`:
- `a certificate is issued, installed, and not issued again`. Full issuance through to a downloaded certificate. Asserts the challenge domain, the key authorization sent to the broker, and that a successful round does not start again.
- `the certificate is not requested until the DNS records are published`. The A record resolves but the TXT record is missing, so the round stops before finalizing an order.
- `no certificate is issued without a TcpTransport, and the service still runs`. On a memory-only switch, no issuance runs and `running` still fires.
- `issuance aborts when no IP address is configured`. Currently ends in a `ResultDefect` and carries a TODO for #2957.

New fixtures: `tests/tools/cert_server.nim` serves one certificate over plain HTTP, since `downloadCertificate` fetches the order's certificate URL through the session rather than the overridable `get`. `tlsCertPemGenerator` in `tests/tools/crypto.nim` returns the certificate as a PEM string, before `TLSCertificate.init` parses it. `scriptCertificate` on `ACMEApiStub` queues the five responses `getCertificate` consumes.

## References

  #2957. The `no IP address` test documents that nothing calls `AutotlsService.setup` in the builder path, so `config.ipAddress` stays `Opt.none`. The `Opt.get` in `issueCertificate` is then a Defect, which `tryIssueCertificate` does not catch. The TcpTransport guard returns first, so nothing reaches it today.

### DIFF
--- a/tests/tools/cert_server.nim
+++ b/tests/tools/cert_server.nim
@@ -1,0 +1,28 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Copyright (c) Status Research & Development GmbH
+
+{.used.}
+
+import chronicles, chronos
+import websock/http/[common, server]
+
+type CertServer* = ref object
+  ## Serves one certificate over plain HTTP.
+  ## `downloadCertificate` fetches the order's certificate URL through the session,
+  ## so it is the one ACME request a stub cannot answer.
+  server: HttpServer
+  url*: string
+
+proc startCertServer*(certificate: string): CertServer =
+  proc serve(request: HttpRequest) {.async.} =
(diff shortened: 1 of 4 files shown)


tests/tools/cert_server.nim
```diff
@@ -0,0 +1,28 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
```
gmelodie, line 1: Cool addition!

gmelodie (approved):

vladopajic (approved):

PR #2967, opened 2026-08-19
fix(kad): keep peer addresses when the admission probe cap is full

gmelodie:
## Summary

A lookup dialed peers with an empty address list and failed with `Unable to establish outgoing link in establishConnection: peer_id=... addrs=@[]`. The peers were reachable. The node had discarded their addresses.

`admitPeers` recorded a candidate's addresses and then started an admission probe, both in one loop, and the probe-slot check was a `break`. Every candidate after the first full slot skipped `addressBook.extend`. Their ids were already on the shortlist, because `applyReplies` calls `updateShortlist` before `admitPeers`, so `fillSlots` later dispatched to an id whose addresses were nowhere. The `break` gained nothing either: the loop is synchronous, so no slot can free during it, and `continue` starts the same number of probes.

With one bootstrap node and 500 joining nodes this is the normal case rather than an edge case. The first reply takes every probe slot and every reply after it stores nothing.

## What changed

1. **The probe cap no longer aborts address recording.** `scheduleAdmissionProbe` (`find.nim:334`) returns `false` instead of leaving the loop. Probes are still never queued: a candidate with no slot is retried when a later reply names it.
2. **An unreachable id can no longer evict a dialable peer.** `updateShortlist` (`find.nim:96`) let an id the node could not dial displace a closer one through `tryEvictFarthest`. `canBecomeDialable` now requires the local address policy to keep at least one offered address, or the peer to be reachable already. Under the permissive default policy this is a correctness gate; under `publicRoutableAddressPolicy` it is a real filter.
3. **The lookup drops candidates it cannot dial.** `dropUndialable` (`find.nim:431`) runs before `fillSlots` and removes shortlist entries with no stored address and no live connection. This covers candidates the address policy or the IP-diversity caps rejected during admission.
4. **The dialability decision is shared.** `dialAddrs` (`rpc.nim:21`) is now used by `dispatchRpc`, `isDialable` and `checkAndEvictPeer` (`kademlia.nim:89`), so the lookup guard and the liveness eviction cannot disagree about whether a peer is usable.
5. **`DefaultMaxConcurrentProbes` goes from 20 to 50,** so the routing table fills faster under a join storm. See Risk Assessment.
6. **`admitPeers` reads as its gate list.** Each step is a named call in order (`recordAddrs`, `tryRefreshAdmitted`, `scheduleAdmissionProbe`), which makes the two exits visible: a candidate is recorded once its addresses pass, and admitted once the probe answers. `updatePeers` is the same list minus the probe and shares `recordAddrs`. No behavior change.
7. **A routing-table seat now implies addresses that outlive it.** A seat resists liveness eviction for `DefaultLivenessGracePeriod` (1 hour) but `updatePeers` recorded its seeds at `AddressConfidence.Low` (15 minutes). A seed is never dialed and never connected, so neither `markConnected` nor identify lifts it, and its addresses expired under a standing seat. `updatePeers` and `admitPeer` now record at Medium.

## Affected Areas

- [x] Peer Management / Discovery: routing-table admission, the address book records it writes, and the lookup shortlist.
- [x] Protocol Logic: `kad-dht` `FIND_NODE` lookup dispatch and the shared outbound RPC path.
- [ ] Gossipsub
- [ ] Transports
- [ ] Build / Tooling
- [ ] Other

No wire-format change and no public signature change, so no downstream integration branch is needed. Nimbus, Waku and Codex: N/A.

## Impact on Library Users

- A peer named in a reply keeps its addresses even when the probe cap is full, so lookups can dial it. This is the fix.
- A `closerPeers` entry whose addresses the local policy rejects, and that is unknown locally, no longer joins the shortlist. `toPeer` refuses to serialize an address-free `Peer`, so a conforming implementation never sends one.
- `findPeer` returns `err("peer not found")` instead of a `PeerInfo` with `addrs: @[]`. The old value was undialable.
- `dispatchPeer` logs a lookup RPC failure at `debug` instead of `error`. At 500 nodes the `error` line flooded the log.
- New metrics: `kad_admission_probes_dropped` and `kad_lookup_undialable_peers`.
- `DefaultMaxConcurrentProbes` rises to 50. A node pinning its own `KadDHTLimits` is unaffected.

## Risk Assessment

- **Probe fan-out.** This is the change worth arguing about. A cap of 50 means one reply can put 50 concurrent dials in flight instead of 20, and remote input drives that number. The IP-diversity caps still apply, so reaching 50 needs 50 distinct attacker-controlled addresses, the same requirement as reaching 20 before. The cost is higher peak socket and file-descriptor use. Reviewers who want the address fix without this can drop the `types.nim` hunk on its own.
- **Probe concurrency against the RPC cap.** 50 probes take up to half of `DefaultMaxConcurrentRpcs` (100). `rpcSem` waits rather than dropping, so the worst case is added lookup latency during a join storm. `DefaultMaxConcurrentRpcs` is left alone; raise it in a follow-up if measurement warrants.
- **Lookup completeness.** `dropUndialable` shrinks the shortlist, so a remote naming unresolvable ids reduces that lookup's candidate set. Those entries were undialable, so it replaces a certain dial failure with no dial, and change 2 stops such ids from displacing good peers first.

## References

None.

## Additional Notes

Test fixtures now give every seeded peer an address (`insertRandomPeers`, `populateRoutingTable`, `addRandomPeers`, plus new `closerPeer` and `peersWithAddrs` helpers). The old fixtures modelled states production never produces, since both admission paths write the address book before inserting, and `toPeer` refuses an address-free peer. `addRandomPeers` takes the `KadDHT` rather than the hasher, because `LookupState.kad` is not exported and the helper needs the peer store.

Each new guard was checked against a build with the guard removed, and fails there. Worth noting for a follow-up: the pre-existing `updateShortlist prefers closer peers when over cap` test builds fixture ids from raw 32-byte `Key` values that `PeerId.init` rejects, so its far peer never enters the shortlist and its assertion is met by the close peers alone. Left as found.

`test_limits` (16), `test_lookup` (20), `test_find` (20), `test_bootstrap` (19), `test_probe_backoff` (8), `test_ip_diversity` (4) and `test_fix_low_peers` (5) pass locally. The remaining kademlia suites were not run locally; CI is the gate for those.

### DIFF
--- a/libp2p/protocols/kademlia/find.nim
+++ b/libp2p/protocols/kademlia/find.nim
@@ -55,6 +55,20 @@ proc getFarthest(
   else:
     Opt.none((PeerId, XorDistance))
 
+proc dropPeer(state: LookupState, pid: PeerId) {.raises: [].} =
+  state.shortlist.del(pid)
+  state.attempts.del(pid)
+  state.responded.del(pid)
+
+proc isDialable(kad: KadDHT, peerId: PeerId): bool {.raises: [].} =
+  kad.dialAddrs(peerId).len > 0 or kad.switch.isConnected(peerId)
+
+proc canBecomeDialable(
+    kad: KadDHT, peerId: PeerId, addrs: seq[MultiAddress]
+): bool {.raises: [].} =
+  ## Whether admission could keep one of `addrs`, or the peer is reachable already.
+  addrs.anyIt(kad.config.addressPolicy.accepts(it)) or kad.isDialable(peerId)
+
 proc tryEvictFarthest(state: LookupState, newDist: XorDistance): bool {.raises: [].} =
   ## Drop the worst (farthest) peer from the shortlist if it is farther than
   ## ``newDist``. Considers all peers — including ones that already responded —
@@ -65,9 +79,7 @@ proc tryEvictFarthest(state: LookupState, newDist: XorDistance): bool {.raises: 
     return false
   if newDist >= dist:
     return false
-  state.shortlist.del(pid)
-  state.attempts.del(pid)
-  state.responded.del(pid)
+  state.dropPeer(pid)
   return true
 
 proc updateShortlist*(state: LookupState, msg: Message): seq[PeerInfo] {.raises: [].} =
@@ -80,6 +92,8 @@ proc updateShortlist*(state: LookupState, msg: Message): seq[PeerInfo] {.raises:
     let pid = PeerId.init(raw).valueOr:
       continue
     if state.shortlist.contains(pid):
+      continue
+    if not state.kad.canBecomeDialable(pid, newPeer.addrs):
       continue
 
     let dist = xorDistance(pid, state.target, state.rtable.config.hasher)
@@ -216,6 +230,14 @@ proc admissibleAddrs(
     kad.config.addressPolicy, rtable, p, kad.config.limits.diversityCaps(), pending
   )
 
+proc recordAddrs(
+    switch: Switch,
+    peerId: PeerId,
+    addrs: seq[MultiAddress],
+    confidence = AddressConfidence.Low,
+) {.raises: [].} =
+  switch.peerStore[AddressBook].extend(peerId, addrs, confidence)
+
 proc updatePeers*(
     switch: Switch,
     addressPolicy: PeerAddressPolicy,
@@ -224,14 +246,11 @@ proc updatePeers*(
     caps: DiversityCaps = defaultDiversityCaps(),
 ) {.raises: [].} =
   ## Unprobed admission, for trusted seed peers only; see ``admitPeers``.
-  let addressBook = switch.peerStore[AddressBook]
   for p in peerInfos:
     let addrs = switch.admissibleAddrs(addressPolicy, rtable, p, caps)
     if addrs.len == 0:
       continue
-    # Store before insert: a peer rejected for lack of bucket space still reaches
-    # the lookup shortlist, and would be undialable without its addresses.
-    addressBook.extend(p.peerId, addrs, AddressConfidence.Low)
+    switch.recordAddrs(p.peerId, addrs, AddressConfidence.Medium)
     discard rtable.insert(p.peerId)
 
 proc updatePeers*(kad: KadDHT, peerInfos: seq[PeerInfo]) {.raises: [].} =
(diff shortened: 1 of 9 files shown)


libp2p/protocols/kademlia/find.nim
```diff
@@ -336,22 +383,12 @@ proc admitPeers*(
...
-    if kad.admissionProbes.hasKey(probeKey):
+    kad.switch.recordAddrs(p.peerId, addrs)
+    if rtable.tryRefreshAdmitted(p.peerId):
       continue
-    if kad.probeBackedOff(p.peerId, addrs):
-      trace "Kad admission probe backed off", peer = p.peerId.shortLog()
-      kad_admission_probes_backed_off.inc()
+    if not kad.scheduleAdmissionProbe(rtable, p.peerId, addrs, onAdmit):
       continue
-    if not kad.admissionSem.tryAcquire():
```
gmelodie, line 351: This was the problematic line

libp2p/protocols/kademlia/find.nim
```diff
@@ -81,6 +93,8 @@ proc updateShortlist*(state: LookupState, msg: Message): seq[PeerInfo] {.raises:
       continue
     if state.shortlist.contains(pid):
       continue
+    if not state.kad.canBecomeDialable(pid, newPeer.addrs):
```
richard-ramos: Should this also check the ip diversity cap? (basicall all the conditions from admitPeers)

richard-ramos (approved):

libp2p/protocols/kademlia/find.nim
```diff
@@ -309,25 +338,46 @@ proc trackProbe(kad: KadDHT, probeKey: ProbeKey, probe: Future[void]) {.raises:
...
+  ## False when no probe starts; the candidate is retried on a later reply.
+  let probeKey: ProbeKey = (rtable.selfId, peerId)
+  if kad.admissionProbes.hasKey(probeKey):
+    return false
+  if kad.probeBackedOff(peerId, addrs):
+    trace "Kad admission probe backed off", peer = peerId.shortLog()
+    kad_admission_probes_backed_off.inc()
+    return false
+  if not kad.admissionSem.tryAcquire():
+    trace "Kad admission probe dropped: no free slot", peer = peerId.shortLog()
```
vladopajic, line 359: there is no need for `.admissionSem.tryAcquire()` to happen in different proc where release is happening. 

having it in different procs is always a bad pattern, as there are higher changes of making issues down the line - overlocking, underlocking, deadlocking etc

vladopajic (approved):

PR #2989, opened 2026-08-25
fix(bufferstream): always wake the reader when the stream closes

gmelodie:
## Summary

`BufferStream.closeImpl` skips its whole wakeup block when a read and a push are in flight at the same time (`s.reading and s.pushing`). The state table above the code called that combination "Na", but no type and no assertion enforces it. When it happens with an empty `readQueue`, `close()` returns without a wakeup: the parked reader gets neither an Eof nor an item, and the close path is finished with it.

This PR removes only the outer guard, and keeps the branches as they are:

```nim
if s.reading:
  if s.readQueue.empty():
    s.readQueue.addLastNoWait(Eof)
elif s.pushing:
  if not s.readQueue.empty():
    discard s.readQueue.popFirstNoWait()
```

The reader has priority. A queued item belongs to the parked reader, so the pop branch runs only when no read is in flight. This keeps the case that a flat `elif s.pushing and not s.readQueue.empty()` breaks: a putter can add an item and wake the reader, and `close()` can then run before the reader resumes. A pop at that moment steals the item, and the reader waits again on an empty queue after the close is over.

`addLastNoWait` stays safe: the branch runs only on an empty queue of capacity 1, so it cannot raise `AsyncQueueFullError`. The existing `try`/`except` around the block is unchanged.

## Tests

Three tests in `tests/libp2p/stream/test_bufferstream.nim`:

- `close wakes a reader parked on an empty queue`: the plain case.
- `close wakes both a parked reader and a parked pusher`: the state that master skips. It hangs on master and passes here.
- `close does not remove data reserved for an active reader`: the reader is awake but not resumed, and the queue holds the item of the putter. It guards the pop branch against the reader.

## Affected Areas

- [x] Other: `libp2p/stream/bufferstream.nim`, the close path shared by every muxed channel.

N/A. One line removed in one close path, no API and no wire-format change, so no downstream build exercises anything this PR touches.

## Impact on Library Users

No API change. One observable behavior change: when `close()` races a putter that is parked on a full queue while a reader is parked on an empty one, the reader now gets an Eof (`readOnce` returns 0) instead of a hang. The push future completes after the reader consumes the Eof.

## Risk Assessment

- Backward compatibility: no API and no protocol change.
- Network behavior: unchanged. The edit is local to one stream's close path.
- Performance: unchanged. The same branches run, with one condition less.
- Security: no new input, no new state, no new allocation.


libp2p/stream/bufferstream.nim
```diff
@@ -148,31 +148,12 @@ method closeImpl*(s: BufferStream): Future[void] {.async: (raises: [], raw: true
...
-        if s.readQueue.empty():
-          # There is an active reader
-          s.readQueue.addLastNoWait(Eof)
-      elif s.pushing:
-        if not s.readQueue.empty():
-          discard s.readQueue.popFirstNoWait()
+    if s.reading and s.readQueue.empty():
+      s.readQueue.addLastNoWait(Eof)
+    elif s.pushing and not s.readQueue.empty():
+      discard s.readQueue.popFirstNoWait()
```
richard-ramos: This is introducing a race when reading and pushing are both true and the queue is not empty. 

A waiting pusher can add an item and wake the reader, but the reader might not have resumed yet.... if `closeImpl` runs at this point, that `elif` will remove the item from queue, and then reader finds the empty queue and hangs because the close has already finished 

```suggestion
    if s.reading:
      if s.readQueue.empty():
        s.readQueue.addLastNoWait(Eof)
    elif s.pushing:
      if not s.readQueue.empty():
        discard s.readQueue.popFirstNoWait()
```

richard-ramos:
```nim
asyncTest "close does not remove data reserved for an active reader":
  let stream = BufferStream.new()
  await stream.pushData("123".toBytes())

  let pushFut = stream.pushData("456".toBytes())
  discard stream.readQueue.popFirstNoWait()

  var data: array[3, byte]
  let readFut = stream.readOnce(addr data[0], data.len)
  check:
    stream.reading
    stream.pushing
    stream.readQueue.empty()

  proc closeStream(udata: pointer) {.gcsafe, raises: [].} =
    discard cast[BufferStream](udata).close()

  callSoon(closeStream, cast[pointer](stream))

  let readCompleted = await readFut.withTimeout(100.milliseconds)
  if not readCompleted:
    await readFut.cancelAndWait()

  check readCompleted
  check await pushFut.withTimeout(100.milliseconds)
```

Output:
```
    test_bufferstream.nim(315, 10): Check failed: readCompleted

  [FAILED ] (  0.10s) close does not remove data reserved for an active reader
```

libp2p/stream/bufferstream.nim
```diff
@@ -148,31 +148,12 @@ method closeImpl*(s: BufferStream): Future[void] {.async: (raises: [], raw: true
...
-  # notify the reader that the channel closed
-  #
-  # In all other cases, there should be a data to complete
-  # a read or enough room in the queue/buffer to complete a
-  # push.
-  #
-  # State       | Q Empty  | Q Full
-  # ------------|----------|-------
-  # Reading     | Push Eof | Na
-  # Pushing     | Na       | Pop
```
richard-ramos, line 168: Why remove this comment?  This is one of the rare cases that looks useful

richard-ramos (approved):

PR #2992, opened 2026-08-25
feat(dialer): collect dial candidates before dialing, behind a flag

gmelodie:
## Summary

`Dialer.dialAndUpgrade` held three nested loops: raw addresses, dnsaddr expansion, DNS resolution, with the dial at the bottom. This PR splits that walk into named steps and puts a second dial path behind an opt-in flag.

`expandCandidate` expands one dnsaddr record. `resolveCandidate` resolves one name and drops the addresses no transport handles. Both dial paths call the same two procs, so the expansion and the resolution have one implementation.

`withDialRanking` on `SwitchBuilder` selects the new path, and it is off by default. The default path, `dialInOrder`, keeps the old behavior: resolve one address, dial it, move to the next.

With the flag on, `dialRanked` calls `collectCandidates` first. That proc starts every lookup at once and gathers the results in the order the peer advertised them, so a dial waits for the slowest name instead of for the sum of the names. A ranker needs the whole candidate set before it opens the first socket, and this is the step that hands it that set.

There is no ranking yet. The candidates keep the advertised order, and the dials stay sequential. The ranker and the racer come next.

## Affected Areas

  `libp2p/dialer.nim`, the new `libp2p/dialcandidate.nim`, and the `withDialRanking` builder in `libp2p/builders.nim`.

- [x] Transports
  Tests only. A new cancellation test runs against TCP, WS, WSS and QUIC, and its body lives in one proc so the four call sites do not each compile a copy.

`dialAndUpgrade` now declares `raises: [CancelledError]`. It used to declare `[CancelledError, MaError, TransportAddressError, LPError]`. The expansion and the resolution steps catch their own errors and skip the address that failed, so those three effects no longer escape. A caller that wraps `dialAndUpgrade` in `except MaError` gets a branch the compiler can prove is dead.

Everything else keeps its signature, and the default dial path keeps its order and its results.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

New public names: `SwitchBuilder.withDialRanking`, the `dialRanking` parameter on `Dialer.new`, and `MaxDialCandidates`.

A switch that does not call `withDialRanking` dials exactly as before.

A switch that calls it gets two changes. Every dnsaddr and DNS name resolves before the first socket opens. A peer advertising more than 32 dialable candidates gets the first 32 dialed.

## Risk Assessment

The flag is the safety net: the default path is the old path, so a regression in the collection step cannot reach a user who did not ask for it.

The part to review is the fan-out. A remote peer picks its own address list through identify, each entry can be a dnsaddr, and `resolveDnsAddr` caps its recursion at depth 6 but never caps its breadth. `limitFanOut` therefore caps two things: the number of names one stage turns into lookups, and the number of addresses held in flight before the transport filter runs. Both caps log what they drop.

The candidate cap of 32 is a judgment call. It is high enough that no honest peer reaches it, and low enough that a racer cannot be told to open hundreds of sockets. A peer whose only reachable address sits past position 32 becomes unreachable, which takes a longer address list than anything seen in practice.

Front-loading the resolution costs the slowest lookup before the first dial. While the dials stay sequential, that can lose to the default path when one name is slow and an early address would have connected. Concurrent lookups keep the cost to one lookup instead of a chain of them, and the racer is what turns the collected set into a latency win.


libp2p/dialer.nim
```diff
@@ -200,6 +208,47 @@ proc normalizedDialAddrs(
...
+          expandedAddress = expandedAddress,
+          hostname = hostname,
+          resolvedAddresses = resolvedAddresses
+
+        for resolvedAddress in resolvedAddresses:
+          candidates.add(
+            DialCandidate(
+              address: resolvedAddress, hostname: hostname, peerId: addrPeerId
+            )
+          )
```
richard-ramos: probably makes sense to filter by supported transports, i.e. if you only support TCP, why add a QUIC address (or any other transport address) to candidates

libp2p/dialcandidate.nim
```diff
@@ -0,0 +1,15 @@
...
+import multiaddress, peerid
+
+export multiaddress, peerid
+
+type DialCandidate* = object
+  ## One address a dial can attempt, with dnsaddr and DNS already resolved.
+  address*: MultiAddress
+  hostname*: string
+    ## Name the address was resolved from, empty when it is a wire address.
+  peerId*: Opt[PeerId] ## Pinned by a dnsaddr record, otherwise the dialed peer's.
```
richard-ramos, line 13: Do all of these fields need to be public?

libp2p/dialcandidate.nim
```diff
@@ -0,0 +1,15 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Copyright (c) Status Research & Development GmbH
+
+{.push raises: [].}
+
+import multiaddress, peerid
+
+export multiaddress, peerid
```
richard-ramos: Is it necessary to export this?

tests/libp2p/transports/cancellation_tests.nim
```diff
@@ -0,0 +1,58 @@
...
+      var accepted: seq[RawConn]
+
+      proc acceptLoop() {.async: (raises: []).} =
+        while true:
+          try:
+            let conn = await server.accept()
+            if not conn.isNil():
+              accepted.add(conn)
+          except CatchableError:
+            return
```
richard-ramos: Could it be more specific?

libp2p/dialer.nim
```diff
@@ -200,6 +208,47 @@ proc normalizedDialAddrs(
...
+
+          if candidates.len >= MaxDialCandidates:
+            debug "Dropping the addresses over the candidate limit",
+              peerId, limit = MaxDialCandidates
+            break collect
+
+  candidates
+
 proc dialAndUpgrade*(
```
richard-ramos: Something we discussed is the importance of making this opt-in. The changes in this code are very much not opt-in, as the candidate selection is executed regardless of your intention of having ranking or not.

With this PR you have
- Resolution of addresses before each dial
- Cutoff of addresses to 32.
- Failure to resolve an address on time can affect the dialing

libp2p/dialer.nim
```diff
@@ -200,6 +208,47 @@ proc normalizedDialAddrs(
...
+        debug "Out of time expanding the remaining addresses", peerId, addrs
+        break collect
+
+      for (expandedAddress, addrPeerId) in await self.expandDnsAddr(
+        peerId, rawAddress, deadline
+      ):
+        let
+          hostname = expandedAddress.getHostname()
+          resolvedAddresses = await self.resolveWithDeadline(expandedAddress, deadline)
+
```
richard-ramos: We gotta make sure that a failure resolving an address does not cause the whole thing to fail. If an address cannot be resolved, then we should log and skip it  (it seems we return a `MaError` here).  

It is worth adding a test unit for this

richard-ramos (changes requested):

richard-ramos:
While this PR looks reasonable, I'm thinking this might not be the best final design for the dial ranking commitment, because now you do the following:

1. Resolve everything
2. Rank everything
3. Dial anything

Which it is simple, it kinda conflicts with the goal of reducing connection latency.
Probably having a producer/consumer design fit the goal better.
Direct addresses could be submitted to the dialer immediately while DNS addresses could be submitted as they resolved. The dialer would then maintain a priority queue, and dial sequentially or concurrently depending on the switch configuration.

We can discuss this idea tomorrow

libp2p/dialcandidate.nim
gmelodie: Yes, `dialer.nim` uses this in another module

gmelodie:
@richard-ramos from your point I think the real bottleneck is only 
> 1. Dial everything

No? I made the resolutions concurrent, so we only really wait for the slowest one. I assume ranking (next) would be *trivial* (?)

richard-ramos:
concurrent resolution indeed helps, but we still wait for the slowest dns lookup before dialing a peer.
for example,you might have a direct IP ready to be dialed, but a slow dns address can use the whole timeout. 
then we never try the direct IP.

The ranking is fast, but it is the wait for the candidates the part that is concerning,
Creating a test unit with one direct ip address and one slow dns address would show clearly the problem

richard-ramos (approved):

richard-ramos:
Discussed plans via DM, so, approving :)

libp2p/dialer.nim
```diff
@@ -200,47 +237,165 @@ proc normalizedDialAddrs(
...
+    peerId: Opt[PeerId],
+    addrs: seq[MultiAddress],
+    dir = Direction.Out,
+    deadline = Moment.now() + self.dialTimeout,
+): Future[Muxer] {.async: (raises: [CancelledError]).} =
+  ## Dial the addresses, sharing one `deadline`. Nil when all of them fail.
+
+  let dialAddrs = normalizedDialAddrs(peerId, addrs)
+  debug "Dialing peer", peerId = peerId, addrs = dialAddrs
+
```
richard-ramos: Probably makes sense to defensively deduplicate the `dialAddrs` at this point

PR #2999, opened 2026-08-28
test(autotls): full integration flow in docker

rlve:
## Summary

Adds an integration test that drives `AutotlsService` all the way to a real certificate, against Let's Encrypt's Pebble and a real p2p-forge running beside it in Docker.

A hosted runner cannot get one from the public brokers. p2p-forge publishes the DNS-01 TXT record only after dialing the peer's advertised multiaddrs back over libp2p, and GitHub runners have no inbound connectivity, which is why `AutotlsService correctly downloads challenges` and `switch successfully dials over wss using autotls certificate` call `skip()`. With both servers on loopback the dial-back lands and the flow completes.

`make test_autotls_docker_integration` builds the test image, runs the test inside the containers' network namespace and tears everything down, so it behaves the same on Linux and macOS.

`tests/integration/autotls_docker/README.md` covers what each container is, where every config file comes from and why it differs from upstream.

Screenshot 2026-08-28 at 13 08 48


gmelodie (approved):

vladopajic (approved):

gmelodie:
@rlve for future reference, you can label your pr with `run-daily-ci` to add daily ci as a passing requirement for the pr to be accepted

rlve:
> @rlve for future reference, you can label your pr with `run-daily-ci` to add daily ci as a passing requirement for the pr to be accepted

@gmelodie thanks for the heads up! This particular one didn't need it as the docker tests are set up to run on PRs.

PR #3001, opened 2026-08-28
feat(dialer): overlap ranked dialing with name resolution

gmelodie:
## Summary

Follow-up to #2992. `dialRanked` could not dial anything before every name resolved.

`collectCandidates` calls `expandAll` and then `resolveAll`, and each one is a
barrier over `collectCompleted`. A peer that advertises `/ip4/1.2.3.4/tcp/4001`
and `/dnsaddr/bootstrap.example.org` could not touch the wire address until the
dnsaddr chain answered or the dial deadline killed it. The chain recurses up to
depth 6 and every hop is a TXT lookup, so the cost is worst where ranking helps
least: one good direct address and one dead name pay the full DNS timeout before
the first packet.

`dialRanked` now splits the advertised addresses on `DNS.matchPartial`. It starts
one lookup per name without awaiting them, dials the wire addresses while they
run, then dials each name's addresses the moment that name answers. Nothing waits
for the slowest name: `dialAsResolved` walks the lookups with `one()`, takes
whichever finished, and dials its batch before returning to wait for the next.

A wire address costs no round trip to try, so it goes first whatever its rank: a
dnsaddr that ranks above a direct address loses its place. Among the names, the
dial order is completion order, not the advertised order. Both are deliberate.

A dial that connects cancels the pending lookups in a `defer`, which also covers
the failure path and cancellation.

New procs in `libp2p/dialer.nim`:

- `directCandidates` keeps the advertised addresses that fail `DNS.matchPartial`
  and that a transport handles.
- `nameCandidates` keeps the rest, capped at `MaxDialCandidates` lookups.
- `resolveName` expands and resolves one name, and is the producer.
- `dialAsResolved` is the consumer.
- `dialEach` holds the dial loop that both halves need.
- `capCandidates` holds the cap that `collectCandidates` applied inline.

`collectCandidates`, `expandAll`, `resolveAll` and `limitFanOut` are gone: each
was a barrier over every address, which is the thing this PR removes.

All three caps survive the split. Lookups stay at `MaxDialCandidates`, applied to
the name list. Addresses resolved at once stay at `MaxExpandedAddresses`, now a
`FanOut` counter the lookups share instead of a per-batch filter. Addresses
dialed stay at `MaxDialCandidates` in total, the resolved side getting
`MaxDialCandidates - direct.len`.

`directCandidates` sets `hostname` from `getHostname()`, like `expandCandidate`
does. Without it `wstransport` falls back to `$initAddress` and sends
`1.2.3.4:443` as the Host header instead of `1.2.3.4`. The `DialCandidate.hostname`
doc comment claimed the field is "empty when it is a wire address", which was
already wrong before this PR. The comment is corrected, the code is not.

## Affected Areas

The path is behind the `dialRanking` flag, which is off by default and has no
downstream user yet. No integration branch needed.

- **Nimbus:** N/A
- **Waku:** N/A
- **Codex:** N/A

## Impact on Library Users

No API change. Behaviour changes only for a switch built with
`withDialRanking(true)`: a peer's wire addresses are dialed before its names,
instead of after every lookup finishes.

## Risk Assessment

- Backward compatibility: the default dial path (`dialInOrder`) is untouched.
- Ordering: the advertised rank no longer decides between a name and a wire
  address, nor between two names. A deployment that relies on a dnsaddr being
  tried first would see the direct address, or the faster name, win.
- Resource use: the lookups now run while a dial is in flight, so a dial holds
  both at once. The `defer` cancels them on every exit path, and the total
  candidate cap is unchanged at `MaxDialCandidates`.

## References

- #2992 — the ranking PR this builds on.

## Additional Notes

Tests in `tests/libp2p/test_dialer.nim`:

- A wire address connects while a name still stalls, and the stalled lookup is
  cancelled. This one fails without the change.
- A name that answers is dialed while another name still stalls. This one fails
  without the producer/consumer loop.
- A peer that advertises names only still connects.
- The resolved addresses are still reached when every wire address fails.
- A `/ip4/1.2.3.4/tcp/443/wss` address carries `hostname == "1.2.3.4"` into
  `Transport.dial`.

`tests/tools/resolver.nim` gains `StallingResolver`, which answers the names in
its `txtResponses` table, never answers any other, and records whether the caller
cancelled it. `FailingDialTransport` gains
`dialedHosts` and a `handlesAny` flag.

The tests are not run yet: the container's nimble packages are stale against
`libp2p.nimble` and `make setup` does not complete here.

### DIFF
--- a/libp2p/dialer.nim
+++ b/libp2p/dialer.nim
@@ -20,6 +20,7 @@ import
   transports/transport,
   nameresolving/nameresolver,
   upgrademngrs/upgrade,
+  utils/collections,
   utils/future,
   errors
 
@@ -50,6 +51,8 @@ const DefaultDialerTimeout* = 30.seconds
   ## that goes quiet mid-handshake holds the peer's dial lock forever.
 
 type
+  DialAttempt = Future[Muxer].Raising([CancelledError])
+
   DialLock = ref object
     lock: AsyncLock
     users: int ## dials holding or waiting for `lock`
@@ -63,7 +66,7 @@ type
     peerStore: PeerStore
     nameResolver: NameResolver
     ms: MultistreamSelect
-    dialRanking: bool ## resolve every address before the first dial
+    dialRanking: bool ## overlap the name lookups with the dials
     ongoingReleaseOnClose: seq[Future[void].Raising([])]
 
 proc transportFor(self: Dialer, address: MultiAddress): Opt[Transport] =
@@ -237,27 +240,20 @@ proc normalizedDialAddrs(
   else:
     addrs
 
-proc limitFanOut(candidates: seq[DialCandidate]): seq[DialCandidate] =
-  ## The peer picks the list, so cap the lookups and the addresses it can spawn.
-  var
-    kept: seq[DialCandidate]
-    lookups = 0
+proc remaining(limit: int): ref int =
+  ## One count shared by every holder, so concurrent lookups keep to one cap.
+  let left = new(int)
+  left[] = max(limit, 0)
+  left
 
-  for candidate in candidates:
-    if kept.len >= MaxExpandedAddresses:
-      debug "Dropping the addresses over the fan-out limit",
-        limit = MaxExpandedAddresses
-      break
-    if DNS.matchPartial(candidate.address):
-      lookups.inc()
-      if lookups > MaxDialCandidates:
-        continue
-    kept.add(candidate)
+proc takeCandidates(left: ref int, candidates: seq[DialCandidate]): seq[DialCandidate] =
+  let limit = left[]
+  if candidates.len > limit:
+    debug "Dropping the addresses over the candidate limit", limit
 
-  if lookups > MaxDialCandidates:
-    debug "Dropping the names over the lookup limit", limit = MaxDialCandidates
-
-  kept
+  let taken = candidates.take(limit)
+  left[] -= taken.len
+  taken
 
 proc expandCandidate(
     self: Dialer, candidate: DialCandidate, deadline: Moment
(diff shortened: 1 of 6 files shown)


libp2p/dialer.nim
```diff
@@ -359,25 +355,84 @@ proc dialInOrder(
...
+
+    let mux = done.value()
+    if not isNil(mux):
+      return mux
+
+proc dialAll(
+    self: Dialer, candidates: seq[DialCandidate], dir: Direction, deadline: Moment
+): Future[Muxer] {.async: (raises: [CancelledError]).} =
+  ## Dial every candidate at once. Nil when none of them connects.
+
```
richard-ramos, line 424: Probably makes sense to defensively deduplicate the dial candidates at this point (or before they're fed to this proc)

libp2p/dialer.nim
```diff
@@ -359,25 +355,84 @@ proc dialInOrder(
...
-      debug "Out of time for the remaining addresses", peerId, addrs
-      return nil
+  let
+    dialable = remaining(MaxDialCandidates)
+    names = remaining(MaxDialCandidates)
+    unresolved = remaining(MaxExpandedAddresses)
+    direct = dialable.takeCandidates(self.directCandidates(peerId, addrs))
+    lookups = names.takeCandidates(dnsCandidates(peerId, addrs)).mapIt(
+        self.resolveName(it, unresolved, deadline)
+      )
```
richard-ramos: Should we stop or cancel the dns lookups when `dialable == 0`?  because a slow lookup that returns something that cannot be dialed just keeps the whole thing waiting until the deadline

richard-ramos: Should be pretty rare for this to happen tho, but if claude can come up with a solution that is not complex, then probably makes sense to do it. Vibecoded unit test that demonstrates the issue:

```
  asyncTest "Ranked dialing stops when the candidate limit is reached":
    let src = makeStandardSwitch()
    await src.start()
    defer:
      await src.stop()

    let
      resolver = StallingResolver.new()
      transport = FailingDialTransport.new(Upgrade(), rng())
      dialer = Dialer.new(
        src.peerInfo.peerId,
        src.connManager,
        src.peerStore,
        @[Transport(transport)],
        src.ms,
        resolver,
        dialRanking = true,
      )

    var addrs: seq[MultiAddress]
    for i in 0 ..< MaxDialCandidates:
      addrs.add(MultiAddress.init("/memorytransport/addr-" & $i).tryGet())
    addrs.add(MultiAddress.init("/dnsaddr/stalls.example").tryGet())

    let dial = dialer.connect(PeerId.random(rng()).tryGet(), addrs)
    defer:
      await noCancel dial.cancelAndWait()

    let finished = await dial.join().withTimeout(1.seconds)
    check finished
    if finished:
      expect DialFailedError:
        await dial
```

richard-ramos (commented):
Pretty cool!
Some observations before this PR's good to go 🧑‍🍳

richard-ramos (approved):

PR #3006, opened 2026-08-31
fix(cbind): guard service discovery entry points on a started switch

### LINKED ISSUE #3003: cbind: ServiceDiscovery start/stop ordering is unguarded — service_disco_start before ctx_start silently latches a broken node
## Summary

The cbind FFI exposes `libp2p_ctx_service_disco_start` / `libp2p_ctx_service_disco_stop`, but `ServiceDiscovery` is *already* started and stopped by the switch lifecycle. Nothing guards the ordering, so a host that calls `service_disco_start` before `libp2p_ctx_start` gets a silently half-functional node that cannot be repaired by starting the switch afterwards.

## Current wiring

`mountServiceDiscovery` (`cbind/libp2p.nim:303`) mounts the protocol during `libp2pNew` and stores it upcast in `lib.kad`. Every disco entry point re-derives it via `resolveServiceDiscovery` (`cbind/libp2p.nim:915`), which checks only that `lib.kad` is set and is a `ServiceDiscovery` — it never checks `lib.running`.

Because `switch.start()` ends with `await s.ms.start()` (`libp2p/switch.nim:426`), which calls `protocol.start()` on every mounted handler (`libp2p/multistream.nim:250`), `libp2p_ctx_start` alone already starts `ServiceDiscovery` (`libp2p/protocols/service_discovery.nim:164`) — after transports are listening and `peerInfo.update()` has signed the record. Teardown mirrors this via `ms.stop()`.

So in the intended flow the two FFI procs are redundant.

## The problem

`libp2p_ctx_create` → `libp2p_ctx_service_disco_start` (no `libp2p_ctx_start`) succeeds and returns `ok(true)`. At that point:

| Piece | State | Consequence |
|---|---|---|
| transports | `running = false`, `stopping = false` | outbound dials still work — `TcpTransport.dial` guards on `stopping`, not `running` (`libp2p/transports/tcptransport.nim:303`) |
| accept loops | absent | no inbound streams; SD serves no `register`/`getAds`, Kad answers no queries |
| `peerInfo.addrs` | empty (`libp2p/peerinfo.nim:188`) | `update()` has not run |
| `peerInfo.signedPeerRecord` | unset (`libp2p/peerinfo.nim:111`) | `maintainSelfSignedPeerRecord` publishes an XPR with no addresses |

Kad's bootstrap (`libp2p/protocols/kademlia.nim:436-441`) dials and fills the routing table, and the advertiser starts registering — all for an address-less, undialable node.

**The latch:** `KadDHT.start()` sets `kad.started = true` (`libp2p/protocols/kademlia.nim:452`). A later `libp2p_ctx_start` → `ms.start()` → `ServiceDiscovery.start()` hits the guard at `libp2p/protocols/service_discovery.nim:165`, logs `warn "Starting kad-disco twice"` and returns. The switch comes up listening, but discovery is never re-bootstrapped with the now-real addresses. Recovery requires an explicit `service_disco_stop` + `service_disco_start`.

## Reproduction

1. `libp2p_ctx_create` with `mountServiceDiscovery = true` and bootstrap nodes.
2. `libp2p_ctx_service_disco_start` — returns `ok(true)`.
3. `libp2p_ctx_start` — succeeds; log shows `Starting kad-disco twice`.
4. Node advertises an extended peer record with no addresses and never serves inbound discovery traffic.

## Suggested fix

Guard the FFI entry point rather than changing the core lifecycle:

```nim
proc libp2pServiceDiscoStart*(lib: LibP2P): Future[Result[bool, string]] {.ffi.} =
  if not lib.running:
    return err("switch not started; call libp2p_ctx_start first")
  ...
```

Worth considering alongside it:

- Document that `service_disco_start` is **resume-after-stop**, not init — its only legitimate use is restarting discovery after `service_disco_stop`, which does clear `started` via `procCall stop(KadDHT(disco))` (`libp2p/protocols/service_discovery.nim:215`).
- Same guard on the other `libp2p_ctx_service_disco_*` entry points, which are equally reachable pre-start.
- There is no C example exercising service discovery under `cbind/examples/`; one would have made this ordering obvious.

## Notes

- `libp2pKadRandomRecords` (`cbind/libp2p.nim:922`) and `libp2pServiceDiscoRandomLookup` (`cbind/libp2p.nim:1004`) have identical bodies — unrelated, but noticed while tracing this.

gmelodie:
## Summary

The switch lifecycle already starts and stops `ServiceDiscovery`, but nothing guarded the ordering: `libp2p_ctx_service_disco_start` before `libp2p_ctx_start` returned `ok(true)` and latched a node that advertised no addresses and served no inbound queries. Every `libp2p_ctx_service_disco_*` entry point now resolves through `runningServiceDiscovery`, which requires `lib.running` and otherwise fails with `switch not started; call libp2p_ctx_start first`.

Adds `cbind/examples/service_disco.c`, the first C example for this family: it asserts the guard, then advertises and looks up a service across two nodes.

Fixes #3003.

## Affected Areas

- [x] Build / Tooling

## Impact on Library Users

C hosts only, on a sequence that was already broken: a `service_disco_*` call before `libp2p_ctx_start` now errors instead of half-starting the node. The intended order is unchanged.

## Risk Assessment

Low. Nothing that worked before stops working. Two known gaps left open: the `lib.running` read is a snapshot, so a concurrent `libp2p_ctx_stop` still races a call in flight (pre-existing, shared by every entry point in the file), and the `libp2p_ctx_kad_*` procs stay unguarded since they have no FFI start/stop and cannot latch.

## References

- Issue: #3003
- Verified in-sandbox: all 7 examples build and run to exit 0.

### DIFF
--- a/cbind/cbind.nimble
+++ b/cbind/cbind.nimble
@@ -106,7 +106,9 @@ task examples, "Build and run the C bindings examples":
     cborObjs.add obj
   let cborObjsStr = cborObjs.join(" ")
 
-  for example in ["echo", "gossipsub", "kad", "relay", "peerstore", "metrics"]:
+  for example in [
+    "echo", "gossipsub", "kad", "service_disco", "relay", "peerstore", "metrics"
+  ]:
     let outBin = "../build/" & example
     exec "gcc -std=c11 -O2 -I c_bindings -I " & vendor & " examples/" & example & ".c " &
       cborObjsStr & " " & lib & " -pthread -Wl,-rpath,'$ORIGIN' -o " & outBin
--- a/cbind/examples/CMakeLists.txt
+++ b/cbind/examples/CMakeLists.txt
@@ -62,7 +62,7 @@ set_property(TARGET tinycbor PROPERTY POSITION_INDEPENDENT_CODE ON)
 
 find_package(Threads REQUIRED)
 
-foreach(example echo gossipsub kad relay peerstore metrics)
+foreach(example echo gossipsub kad service_disco relay peerstore metrics)
   add_executable(${example} "${example}.c")
   target_include_directories(${example} PRIVATE "${GEN_DIR}")
   target_link_libraries(${example} PRIVATE "${LIB_FILE}" tinycbor Threads::Threads)
--- a/cbind/examples/README.md
+++ b/cbind/examples/README.md
@@ -7,6 +7,7 @@ The bindings are a header-only C API over a C ABI. Every generated method is asy
 - `echo.c` — two TCP nodes. The server mounts a custom protocol (`/cbind/echo/1.0.0`) and echoes the bytes it reads. The client dials the server and verifies the round-trip twice: the first dial passes the server's address, so it opens the connection and the stream in one call; the second passes only the peer id, opening a fresh stream over the connection the first dial left behind. The incoming-stream handler only hands the stream id to `main`, which serves it inline, since calling back into the library from the dispatch thread would deadlock.
 - `gossipsub.c` — two TCP nodes join a shared topic and exchange one message, delivered to the subscriber through its pub/sub-message listener.
 - `kad.c` — two TCP nodes form a Kademlia DHT (a server, and a client that lists the server as its bootstrap node), then round-trip a value (`kad_put_value` / `kad_get_value`) and a provider record (`create_cid`, `kad_start_providing` / `kad_get_providers`) over the wire.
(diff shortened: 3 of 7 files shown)


gmelodie:
Tagging @NagyZoltanPeter as I can't tag you as a reviewer directly

richard-ramos (approved):
lgtm

PR #3011, opened 2026-09-01
feat(dialer): back off the addresses and the peers that keep failing

gmelodie:
## Summary

Last part of the [dial ranking and backoffs](https://roadmap.vac.dev/p2p/ift/2026q3-nimlibp2p-dial-ranking-backoffs) roadmap item, after #2992 and #3001. An address that refuses a dial, and a peer whose addresses all fail, cost the full connect timeout again on every retry.

New `libp2p/dialbackoff.nim` counts consecutive failures per address and per peer. Past `tolerance` failures the key waits `base * factor^n`, capped at `maxDelay`. A dial that lands deletes the entry, and an entry that blocks nothing for one `maxDelay` decays. Both tables hold `MaxBackoffEntries` keys, and a full table drops the entry whose wait ends first.

Opt-in through `withDialBackoff()`. Without it the dialer holds a nil `DialBackoff` on which every call is a no-op, so dialing behaves as it does today.

`dialAndUpgrade` skips a backed-off address, and records the transport failure, the failed upgrade and the deadline. `establishConnection` refuses a backed-off peer before it takes a connection slot. `forceDial` goes through the wait, the way it already goes through the connection limits. Metrics `libp2p_dial_backoffs` and `libp2p_dial_backoff_skips` are labelled `scope` = `peer` / `address`.

## Affected Areas

N/A. The dialer keeps its current behavior unless the switch calls `withDialBackoff()`.

## Impact on Library Users

New `withDialBackoff(config = DefaultDialBackoff)` on `SwitchBuilder` and a `dialBackoff` argument on `Dialer.new`, both off by default. With it on, `connect` raises `DialFailedError` at once for a peer on backoff.

## Risk Assessment

A long `maxDelay` over a low `tolerance` holds a peer back longer than the caller may want, and `forceDial` is the way through. Nothing changes for a switch that does not opt in.

## References

- https://roadmap.vac.dev/p2p/ift/2026q3-nimlibp2p-dial-ranking-backoffs
- #2992, #3001


libp2p/dialer.nim
```diff
@@ -88,6 +90,9 @@ proc dialAndUpgrade*(
   let transport = self.transportFor(addrs).valueOr:
     return nil
 
+  if self.dialBackoff.blocked(addrs):
```
richard-ramos: Should this check `forcedDial` too?

libp2p/dialer.nim
```diff
@@ -128,6 +134,7 @@ proc dialAndUpgrade*(
       libp2p_dial_duration_ms.observe(
         (Moment.now() - dialStarted).milliseconds, labelValues = ["upgrade_failed"]
       )
+      self.dialBackoff.recordFailure(addrs)
```
richard-ramos: Question: How does this work with circuit relay addresses? since you can have multiple peers reserving a circuit relay address in the same node. Does a failure dialing one of the peerIds should affect all of the other peerIds that reserved in the same node?

libp2p/dialer.nim
```diff
@@ -569,12 +585,14 @@ proc establishConnection(
...
   if isNil(muxed): # None of the addresses connected
     slot.release()
+    self.dialBackoff.recordFailure(peerId)
     raise newException(
       DialFailedError,
       "Unable to establish outgoing link in establishConnection: peer_id=" &
         shortLog(peerId) & " addrs=" & $dialAddrs,
     )
 
+  self.dialBackoff.recordSuccess(peerId)
```
richard-ramos: This should probably be moved after line 597

libp2p/dialbackoff.nim
```diff
@@ -0,0 +1,178 @@
...
+    failures: int
+    until: Moment
+
+  DialBackoff* = ref object
+    config: DialBackoffConfig
+    peers: Table[PeerId, BackoffEntry]
+    addrs: Table[MultiAddress, BackoffEntry]
+
+const DefaultDialBackoff* =
+  DialBackoffConfig(tolerance: 2, base: 5.seconds, factor: 2, maxDelay: 5.minutes)
```
richard-ramos: Consider starting the default backoff after the first failure with a `tolerance` of 0, because with `tolerance` of 2, the first backoff will start only after the third failed dial.

libp2p/dial_backoff.nim
```diff
@@ -0,0 +1,178 @@
...
+
+const DefaultDialBackoff* =
+  DialBackoffConfig(tolerance: 2, base: 5.seconds, factor: 2, maxDelay: 5.minutes)
+
+func delay(config: DialBackoffConfig, failures: int): Duration =
+  ## The wait after `failures` consecutive failures, zero under the tolerance.
+
+  if failures <= config.tolerance:
+    return ZeroDuration
+
```
richard-ramos, line 64: just to avoid looping thru history if factor is <= 1

```suggestion
  
  if config.base <= ZeroDuration or config.maxDelay <= ZeroDuration:
    return ZeroDuration

  if config.factor == 1:
    return min(config.base, config.maxDelay)

```

libp2p/dialbackoff.nim
```diff
@@ -0,0 +1,178 @@
...
+    entries.prune(self.config, now)
+    if entries.len >= MaxBackoffEntries:
+      entries.evictSoonest()
+
+  var entry = entries.getOrDefault(key, BackoffEntry(until: now))
+  if self.config.decayed(entry, now):
+    entry = BackoffEntry(until: now)
+
+  entry.failures.inc()
+  entry.until = now + self.config.delay(entry.failures)
```
richard-ramos: Wdyt of adding some jittering here?

libp2p/builders.nim
```diff
@@ -520,6 +526,10 @@ proc buildSwitch(b: SwitchBuilder): Switch {.raises: [LPError].} =
       )
     )
 
+  var dialBackoff: DialBackoff
+  b.dialBackoff.withValue(config):
+    dialBackoff = DialBackoff.new(config)
```
vladopajic: `DialBackoff` better be implementation detail of `Dialer`.


`DialBackoffConfig` was already exposed, so this type can be passed to Dialer to configure. having `DialBackoff` here looks like second type for same purpose

libp2p/dial_backoff.nim
```diff
@@ -0,0 +1,178 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
```
vladopajic, line 1: snake case for file names are more readable

libp2p/dialbackoff.nim
```diff
@@ -0,0 +1,178 @@
...
+
+proc new*(T: type DialBackoff, config = DefaultDialBackoff): T =
+  T(
+    config: DialBackoffConfig(
+      tolerance: max(config.tolerance, 0),
+      base: config.base,
+      factor: max(config.factor, 1),
+      maxDelay: config.maxDelay,
+    )
+  )
```
vladopajic: seems a bit strange to have this logic.
`new` already gets `config` so it should just set it.
validation should be done at point where config is created by user and passed to lib.

libp2p/dialbackoff.nim
```diff
@@ -0,0 +1,178 @@
...
+proc recordSuccess*(self: DialBackoff, address: MultiAddress) =
+  if self.isNil():
+    return
+  self.addrs.del(address)
+
+proc recordSuccess*(self: DialBackoff, peerId: Opt[PeerId]) =
+  peerId.withValue(pid):
+    self.recordSuccess(pid)
+
+proc new*(T: type DialBackoff, config = DefaultDialBackoff): T =
```
vladopajic: not sure if default value is necessary - design vise

libp2p/dialbackoff.nim
```diff
@@ -0,0 +1,178 @@
...
+  self.addrs.blockedIn(address, "address", now)
+
+proc blocked*(self: DialBackoff, peerId: Opt[PeerId], now = Moment.now()): bool =
+  peerId.withValue(pid):
+    return self.blocked(pid, now)
+  false
+
+proc recordFailure*(self: DialBackoff, peerId: PeerId, now = Moment.now()) =
+  if self.isNil():
+    return
```
vladopajic: `if self.isNil()` 

`DialBackoff` was ref objects just so that nil value could be treated as noop? 
why not use `Opt[DialBackoff]` in dialer - seems like approach we use more often?

tests/libp2p/test_dialbackoff.nim
```diff
@@ -0,0 +1,149 @@
...
+      )
+      start = Moment.now()
+
+    backoff.recordFailure(Opt.none(PeerId), start)
+    check not backoff.blocked(Opt.none(PeerId), start)
+
+  test "A nil backoff blocks nothing":
+    let
+      backoff: DialBackoff = nil
+      peerId = PeerId.random(rng()).tryGet()
```
vladopajic: `peerId` can be suite var like `config`

vladopajic (approved):

libp2p/dialer.nim
```diff
@@ -563,12 +597,16 @@ proc establishConnection(
...
-      await self.dialAndUpgrade(peerId, dialAddrs, dir, Moment.now() + self.dialTimeout)
+      await self.dialAndUpgrade(
+        peerId, dialAddrs, dir, Moment.now() + self.dialTimeout, forceDial
+      )
     except CancelledError as e:
       slot.release()
       raise e
   if isNil(muxed): # None of the addresses connected
     slot.release()
+    self.dialBackoff.withValue(backoff):
```
richard-ramos: dialAndUpgrade can return a `nil` if the address was skipped due to being already backed off, thus, you could be recording a failure for a peer you havent made a dial...

For example,
imagine a malicious peerA that advertices an address via kademlia, or in their own peer record

/ip4/1.2.3.4/tcp/9000

and this address does not belong to  peerA, but to a diff legit peerB. What would happen then if you wanted to dial that address but this time using peerB peerId. Would peerB get backed off?

richard-ramos: I think part of the fix should probably involve to have `backoffKey` use the peerId as part of the key 🤔

libp2p/dialer.nim
```diff
@@ -577,6 +615,9 @@ proc establishConnection(
 
   slot.trackMuxer(muxed)
   await self.finishUpgrade(muxed, dir)
```
richard-ramos: we should probably do something like:
```
try:
   await self.finishUpgrade(muxed, dir)
except DialFailedError as exc
   backouff.recordFailure(peerId)
   raise exc

```

richard-ramos (approved):

PR #3012, opened 2026-09-01
chore(autotls): rename dnsServerURL to domainSuffix

rlve:
## Summary

`AutotlsConfig.dnsServerURL` is the domain every peer's certificate name sits under. `issueCertificate` joins it as `encodePeerId(peerId) & "." & dnsServerURL`, names the result `baseDomain`, and asks the ACME server for a certificate covering `*.<peerID>.libp2p.direct`.

The name is wrong in two ways. The value is a bare domain, `libp2p.direct`, with no scheme and no path. And nothing ever connects to it, resolution goes through the separate `nameResolver` field.

## Affected Areas

- [x] Other
  AutoTLS service configuration.

## Impact on Library Users

Breaking change to `AutotlsConfig`, which is the type `withAutotls` takes.

| Old | New |
| --- | --- |
| `AutotlsConfig.dnsServerURL` | `AutotlsConfig.domainSuffix` |
| `AutotlsConfig.new(dnsServerURL = ...)` | `AutotlsConfig.new(domainSuffix = ...)` |
| `AutoTLSDNSServer` | `DefaultDomainSuffix` |

Values and behaviour are unchanged. Only code that names the field or the constant needs updating, so anyone on the default is unaffected.

## Risk Assessment

Nothing changes at runtime. The default value, the `baseDomain` built from it, and the code paths around them are untouched. The only risk is that callers naming the old field or constant stop compiling.

## References

- [p2p-forge](https://github.com/ipshipyard/p2p-forge) documents `FORGE_DOMAIN` as "the domain suffix of the forge"
- Kubo, the Go IPFS implementation, exposes the client-side setting as [`AutoTLS.DomainSuffix`](https://github.com/ipfs/kubo/blob/master/docs/config.md#autotlsdomainsuffix)
- #2997, the previous rename of `acmeServerURL` and `brokerURL` on `AutotlsConfig`

## Additional Notes

The defaults test in `test_autotls_config.nim` also gains an assertion on `nameResolver`, so it now covers every field of the config. Unrelated to the rename, included here rather than as a separate change.

### DIFF
--- a/libp2p/autotls/service.nim
+++ b/libp2p/autotls/service.nim
@@ -38,7 +38,7 @@ const
   DefaultIssueRetries = 3
   DefaultIssueRetryTime = 1.seconds
 
-  AutoTLSDNSServer* = "libp2p.direct"
+  DefaultDomainSuffix* = "libp2p.direct"
 
 type AutotlsCert* = ref object
   cert*: TLSCertificate
@@ -54,7 +54,7 @@ type AutotlsConfig* = object
   issueRetries*: int
   issueRetryTime*: Duration
   registrationURL*: Uri
-  dnsServerURL*: string
+  domainSuffix*: string
   dnsRetries*: int
   dnsRetryTime*: Duration
   acmeRetries*: int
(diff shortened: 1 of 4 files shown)


gmelodie (approved):

vladopajic (approved):

richard-ramos (approved):

vladopajic:
imho, this should be `chore`.
`refactor` should be when code structure/design is changed.

Issue #2474, opened 2026-05-15
Replace BearSSL in phases by BoringSSL


richard-ramos:
> [!important]
> This is iceboxed and not to be worked on unless priorities change.

## Key Changes
- Remove public `HmacDrbgContext` usage and introduce a libp2p-owned `Rng` type.
- Use BoringSSL `RAND_bytes` as the final production randomness source.
- Add missing FFI in `nim-boringssl` for the headers/functions needed by libp2p.
- Replace BearSSL-backed primitives in libp2p: RNG, HKDF-SHA256, ChaCha20-Poly1305, X25519, RSA, ECNIST, and PEM encoding.
- Remove `bearssl` and `bearssl_pkey_decoder` dependencies once all first-party imports are gone.

### 1. Own the RNG API in `nim-libp2p` `[Completed]`

- Use `nim-boringssl`
- Add `libp2p/crypto/rng.nim` with `Rng`, `newRng`, `generate`,
  `generateBytes`, `shuffle`, `pick`, and `pickOne`.
- Replace all public `HmacDrbgContext` signatures with `Rng` across libp2p,
  tests, examples, and cbind.
- Keep BearSSL as the temporary backend, including a short-lived
  `newBearSslRng(ctx: ref HmacDrbgContext)` compatibility wrapper for callers
  that still own BearSSL contexts.
- Do not expose a BoringSSL-backed `Rng` constructor yet, because some
  first-party primitives still require BearSSL PRNG handles internally.

### 2. Move Noise Symmetric Primitives
- Replace libp2p HKDF, ChaCha20-Poly1305, and X25519/Curve25519 code with
  `nim-boringssl` APIs.
- Keep Noise wire format unchanged.
- Remove BearSSL usage from `hkdf.nim`, `chacha20poly1305.nim`,
  `curve25519.nim`, and Noise.

### 3. Move RSA
- Rework `crypto/rsa.nim` to use BoringSSL `RSA_*`, `BN_*`, `SHA256`,
  `RSA_sign`, and `RSA_verify`.
- Preserve existing serialized key/signature formats and PKCS#1 v1.5 SHA-256
  behavior.

### 4. Move ECNIST
- Rework `crypto/ecnist.nim` to use BoringSSL `EC_KEY_*`, `ECDSA_*`, and `ECDH_compute_key`.
- Preserve P-256/P-384/P-521 support, raw key formats, DER ECDSA signatures, and shared-secret behavior.

[...]

PR #3014, opened 2026-09-01
chore(perf): improve ChaChaPoly and remove byte-at-a-time frame copies

### LINKED ISSUE #1484: issue(perf): protocol is vary slaw comapred to other implementations

gmelodie:
## Summary

The perf protocol ran 12-17x slower than go-libp2p (#1484). On loopback, TCP + noise, 2 GiB in each direction, throughput goes from 292 MiB/s to 1053 MiB/s.

Both endpoints were profiled with gprof (`nim c --debugger:native --passC:-pg --passL:-pg`). The suspects named in the issue thread are not on the throughput path: `readLine` runs once per stream during multistream negotiation, and ref10 ed25519 plus BearSSL RSA/EC run only during the handshake. Two other things account for almost all of the time.

**1. The AEAD used BearSSL's portable C code.** `chacha20poly1305.nim` called `poly1305CtmulRun` and `chacha20CtRun` directly, which measure 615 MiB/s here. The file carried a comment claiming the optimized versions were selected at runtime; no such code existed. BoringSSL's `EVP_aead_chacha20_poly1305` measures 2960 MiB/s, a factor of 4.8. The case for that backend is in the next section.

`decrypt` now takes the expected tag and returns whether it authenticates, instead of writing out a computed tag for the caller to compare. BoringSSL verifies it in constant time inside `EVP_AEAD_CTX_open_gather`; the old call site compared two arrays with `!=`.

**2. Nim compiles `array[a..b]` and `openArray[a..b]` into an element-wise loop.** Slicing a 64 KB buffer copied it one byte at a time, on every frame, in two places:

- `noise.decryptWithAd` did `buf = data[0 .. data.high - ChaChaPolyTag.len]`. gprof attributes 52% of receive-side CPU to that expression. `readMessage` now decrypts in place through `decryptInPlaceWithAd`, which shrinks the frame with `setLen`. The handshake keeps the copying variant, because `decryptAndHash` mixes the original ciphertext into the hash; it runs four times per connection on tens of bytes.
- `perf/client.nim` and `perf/server.nim` did `buf[0 ..< toWrite]` on an `array[PerfSize, byte]` that was never written to. gprof attributes 52% of send-side CPU to that expression. Both now use `newSeq[byte](toWrite)`: same allocation the slice already made, without the byte loop.

`yamux.readHeader` rides along and was **not** measured. It allocated three seqs to parse 12 header bytes and now uses `toOpenArray`. One header per frame, frames up to 256 KB, so it never showed up in the profile and earns no part of the numbers below.

## Why BoringSSL and not BearSSL for this primitive

**It is not a new dependency.** `libp2p.nimble` already declares `nim-boringssl >= 0.0.8`, and `transports/tls/certificate_ffi.nim` already imports it. Tag v0.0.8, the declared minimum, already exports `EVP_aead_chacha20_poly1305`, `EVP_AEAD_CTX_seal_scatter` and `EVP_AEAD_CTX_open_gather`, and already compiles `chacha-x86_64-linux.S` and `chacha20_poly1305_x86_64-linux.S`. The assembly this PR reaches is object code every current build links and never calls. No constraint changes.

**It is the direction the repo already chose.** Issue #2474, "Replace BearSSL in phases by BoringSSL", names ChaCha20-Poly1305 in Phase 2. Phase 1 is complete. The issue is iceboxed "unless priorities change"; a 12-17x deficit against go-libp2p is that change. This PR does the ChaCha20-Poly1305 third of Phase 2 and leaves HKDF and X25519, which are handshake-only and absent from the profile.

**BearSSL cannot reach these numbers at its best.** The vendored csources ship two ChaCha20 implementations, `chacha20_ct.c` and `chacha20_sse2.c`, and four Poly1305 implementations, all C. No AVX2, no AVX-512, no NEON, and no Poly1305 assembly on any target. Switching to BearSSL's best x86-64 pair, `chacha20Sse2Get()` plus `poly1305CtmulqGet()`, is worth about 7%. BoringSSL ships a fused kernel that interleaves both halves in one pass, with AVX2 on x86-64 and NEON on armv8.

aarch64 is worse: BearSSL has no ARM code for either half, so its best there is portable C twice over. nim-boringssl already compiles `chacha-armv8-linux.S` and `chacha20_poly1305_armv8-linux.S`. ARM nodes gain more from this than the benchmark machine did.

`tests/libp2p/crypto/test_chacha20poly1305_speed.nim` turns that into a check. It encrypts 64 MiB through `ChaChaPoly.encrypt` and 64 MiB through the fastest ChaChaPoly BearSSL offers on the same machine in the same process, selected at runtime with a fall back to the portable pair, and requires the first to be at least twice as fast. A ratio, not a threshold, so it needs no tuning on a slow CI runner or an ARM builder. It fails on every BearSSL backend, passes on BoringSSL, and prints both figures.

**BearSSL has no releases, and that costs something visible here.** No tagged release since 0.6 on 2018-08-14; fixes land in the git tree and stay there. Its vendored `README.txt` still says "BearSSL is considered beta-level software" and "Using BearSSL for production purposes would be a relatively bold but not utterly crazy move."

Follow one fix through the chain. Upstream commit `7bea48e5` of 2026-04-06: "Fixed bug in handling incoming records with invalid length (impacted CBC encryption with 3DES or with the aes_small or aes_big AES implementations)", credited to Thai Duong at Calif.io. No release carries it, no CVE, no advisory. nim-bearssl v0.2.7 and v0.2.8 pin the parent tree `79c060ee` of 2023-02-21; v0.2.9 and later pin `7bea48e5`. `libp2p.nimble` says `bearssl >= 0.2.7`, so both pre-fix tags satisfy it, and a lock file resolved to either ships pre-fix code while reading as compliant. The constraint cannot express the difference, because the pinned thing has no version number. That bug is **not** reachable from nim-libp2p: it is in BearSSL's TLS record layer, and this repo uses BearSSL only for RNG, HKDF, EC, RSA and PEM. The point is the reporting chain, not this repo's exposure.

BoringSSL sits on the Chromium security process and carries Project Wycheproof vectors for this exact primitive in tree (`third_party/wycheproof_testvectors/chacha20_poly1305_test.json`). Both libraries are in OSS-Fuzz, so fuzzing is not a difference between them. Release engineering is.

**Constant-time behavior is preserved.** ChaCha20 and Poly1305 are add-rotate-xor and a field multiply, with no table lookup or secret-dependent branch in any implementation of either library. The tag comparison is the part that must be constant time, and it improves here: `EVP_AEAD_CTX_open_gather` compares inside BoringSSL, where the removed Nim code compared two `array[16, byte]` with `!=`.

**It reduces backend mixing.** One fewer BearSSL module links. BearSSL stays for RNG, HKDF, X25519, RSA, ECNIST and PEM, exactly as Phases 1 and 3 through 5 of #2474 describe.

The cost on the other side of the ledger is in Risk Assessment: a wrong-parameter path in BoringSSL now aborts, where the old code could not fail at all.

## Affected Areas

- [x] Transports
  

- [x] Protocol Logic
  

The wire format does not change. Noise frames, yamux headers and the perf protocol are byte for byte what they were, so a patched node interoperates with an unpatched one and with go-libp2p.

- **Nimbus:** not run
- **Waku:** not run
- **Codex:** not run

## Impact on Library Users

`ChaChaPoly.decrypt` changes signature. It took `tag: var ChaChaPolyTag` and filled it with the computed tag; it now takes `tag: ChaChaPolyTag` and returns a `bool` saying whether that tag authenticates the data. External callers check the return value instead of comparing tags. The result is not `discardable`, so an ignored return is a compile error. `encrypt` is unchanged. Everything else is internal.

Builds that reach `chacha20poly1305.nim` now link BoringSSL. Every build already did, through `builders.nim` and the TLS transport.

Throughput on loopback, TCP + noise, 2 GiB per direction, MiB/s:

| muxer | direction | before | after |
| --- | --- | --- | --- |
| yamux | upload | 292 | 1053 |
| yamux | download | 292 | 988 |
| mplex | upload | 344 | 930 |
| mplex | download | 342 | 906 |

The go+go figure in #1484 is 900719925 bytes in 0.8 s, which is 1074 MiB/s.

## Risk Assessment

The backend swap is the real risk, so it was checked against the RFC 7539 vector directly: same ciphertext, same tag, a flipped tag bit rejected, empty and 2-byte plaintexts round-tripping. Full 4 GiB transfers complete over both yamux and mplex.

`test_crypto.nim` is updated for the new signature and gains a forgery case and a check that a failed decrypt zeroes its output. The short-input cases now assert that the plaintext round-trips, where before they only compared tags. `test_chacha20poly1305_speed.nim` is new; it keeps a `bearssl/blockx` import alive in the test tree on purpose, and Phase 5 of #2474 removes it along with BearSSL.

On a failed tag check, `EVP_AEAD_CTX_open_gather` fills the output with zeros (`aead.h:392`), so `decryptInPlaceWithAd` leaves a zeroed, shortened buffer rather than unauthenticated plaintext. Both call sites raise `NoiseDecryptTagError` and never read it. The failure log carries the tag, the nonce counter and the frame length, and no longer dumps the frame body.

`encrypt` has no error channel, so a non-zero return from `EVP_AEAD_CTX_init` or `EVP_AEAD_CTX_seal_scatter` is a `raiseAssert`. Both require a wrong key or tag length, both compile-time constants here, so neither is reachable from network input. The old code could not abort at all, so this is a new, narrow crash surface.

No network behavior changes. No wire compatibility break.

## References

- Fixes #1484
- Implements the ChaCha20-Poly1305 part of Phase 2 of #2474
- RFC 7539

### DIFF
--- a/libp2p/crypto/chacha20poly1305.nim
+++ b/libp2p/crypto/chacha20poly1305.nim
@@ -1,17 +1,13 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 # Copyright (c) Status Research & Development GmbH
 
-## This module integrates BearSSL ChaCha20+Poly1305
-##
-## This module uses unmodified parts of code from
-## BearSSL library <https://bearssl.org/>
-## Copyright(C) 2018 Thomas Pornin <pornin@bolet.org>.
+## This module integrates BoringSSL ChaCha20+Poly1305
 
 # RFC @ https://tools.ietf.org/html/rfc7539
 
 {.push raises: [].}
 
-import bearssl/blockx
+import boringssl
 from stew/assign2 import assign
 from stew/ptrops import baseAddr
 
@@ -44,9 +40,23 @@ proc intoChaChaPolyTag*(s: openArray[byte]): ChaChaPolyTag =
   assign(tag, s)
   tag
 
-# bearssl allows us to use optimized versions
-# this is reconciled at runtime
-# we do this in the global scope / module init
+template withAeadCtx(key: ChaChaPolyKey, body: untyped) =
+  # This AEAD keeps no heap state, so a per-call context is cheaper than caching one.
+  block:
+    var ctx {.inject.}: EVP_AEAD_CTX
+    if EVP_AEAD_CTX_init(
+      addr ctx,
+      EVP_aead_chacha20_poly1305(),
+      addr key[0],
+      csize_t(ChaChaPolyKeySize),
+      csize_t(ChaChaPolyTagSize),
+      nil,
+    ) != 1:
+      raiseAssert "EVP_AEAD_CTX_init failed for ChaChaPoly"
+    defer:
+      EVP_AEAD_CTX_cleanup(addr ctx)
+
+    body
 
 proc encrypt*(
     _: type[ChaChaPoly],
@@ -56,48 +66,62 @@ proc encrypt*(
     data: var openArray[byte],
     aad: openArray[byte],
 ) =
+  ## Encrypts `data` in place, writing its tag to `tag`, which must not alias `data`.
   let ad =
     if aad.len > 0:
       addr aad[0]
     else:
       nil
 
-  poly1305CtmulRun(
-    addr key[0],
-    addr nonce[0],
-    baseAddr(data),
-    uint(data.len),
-    ad,
-    uint(aad.len),
-    baseAddr(tag),
-    # cast is required to workaround https://github.com/nim-lang/Nim/issues/13905
-    cast[Chacha20Run](chacha20CtRun), #[encrypt]#
-    1.cint,
-  )
+  var tagLen: csize_t
+  withAeadCtx(key):
+    if EVP_AEAD_CTX_seal_scatter(
+      addr ctx,
+      baseAddr(data),
+      addr tag[0],
+      addr tagLen,
+      csize_t(ChaChaPolyTagSize),
+      addr nonce[0],
+      csize_t(ChaChaPolyNonceSize),
+      baseAddr(data),
+      csize_t(data.len),
+      nil,
+      0,
+      ad,
+      csize_t(aad.len),
+    ) != 1:
+      raiseAssert "EVP_AEAD_CTX_seal_scatter failed for ChaChaPoly"
+    doAssert tagLen == csize_t(ChaChaPolyTagSize)
 
 proc decrypt*(
     _: type[ChaChaPoly],
     key: ChaChaPolyKey,
     nonce: ChaChaPolyNonce,
-    tag: var ChaChaPolyTag,
+    tag: ChaChaPolyTag,
     data: var openArray[byte],
     aad: openArray[byte],
-) =
+): bool =
+  ## Decrypts `data` in place, returning false and zeroing it when `tag` fails.
   let ad =
     if aad.len > 0:
       addr aad[0]
     else:
       nil
 
-  poly1305CtmulRun(
-    addr key[0],
-    addr nonce[0],
-    baseAddr(data),
-    uint(data.len),
-    ad,
-    uint(aad.len),
-    baseAddr(tag),
-    # cast is required to workaround https://github.com/nim-lang/Nim/issues/13905
-    cast[Chacha20Run](chacha20CtRun), #[decrypt]#
-    0.cint,
-  )
+  var authenticated = false
+  withAeadCtx(key):
+    authenticated =
+      EVP_AEAD_CTX_open_gather(
+        addr ctx,
+        baseAddr(data),
+        addr nonce[0],
+        csize_t(ChaChaPolyNonceSize),
+        baseAddr(data),
+        csize_t(data.len),
+        addr tag[0],
+        csize_t(ChaChaPolyTagSize),
+        ad,
+        csize_t(aad.len),
+      ) == 1
+
+  authenticated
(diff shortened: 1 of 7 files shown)


vladopajic (approved):
lgtm; tho before closing original issue we should manually run perf for go and nim implementations and compare.

richard-ramos (commented):
As described in https://github.com/vacp2p/nim-libp2p/issues/2474 and in DM discussions, changes related to moving from BearSSL to BoringSSL are not desirable for the time being even if this clearly introduces improvements and cost to do so is small:

- It addresses a concrete production-relevant deficit. The reported TCP + Noise throughput rises from about 292 MiB/s to 988–1053 MiB/s, nearly reaching the go-libp2p result from [#1484](https://github.com/vacp2p/nim-libp2p/issues/1484).
- ChaCha20-Poly1305 is on the data path; HKDF and X25519 are handshake-only and were absent from the profile. Migrating only the AEAD is therefore proportionate.
- BoringSSL is already a dependency, so this adds little marginal dependency cost.
- Authentication handling improves: the previous API produced a tag for Nim code to compare with ordinary !=; BoringSSL authenticates internally using its constant-time machinery and zeroes the plaintext on failure.
- The wire format is unchanged.

So... not approving this PR for now unless #2474 is un-iceboxed

vladopajic:
this pr can still merge other changes with boringSSL / bearSSL

Issue #1081, opened 2024-04-08
GossipSub messages relayed back to source on 3 node network


cskiraly:
Setting up a 3 node network (nodes A,B,C) and a single GS topic. When A is sending a message to B and C, these are then relayed between them, finally sent back to A. That last step is not supposed to happen, since on that link the same message already travelled and was received.

Message sequence
A->B , A->C
B->C, C->B
C->A, B->A

cskiraly:
was identified by @ufarooqstatus as potentially caused by https://github.com/vacp2p/nim-libp2p/blob/unstable/libp2p/protocols/pubsub/gossipsub.nim#L327

diegomrsantos:
Could you please point out what part of the spec does it violate?

cskiraly:
Ok, so the correct permanent link to the offending line is https://github.com/vacp2p/nim-libp2p/blob/467b5b4f0cbfca31ececce9dfd1e649df87067e8/libp2p/protocols/pubsub/gossipsub.nim#L327

Regarding what it violates:
- first of all, it is simply something better not done, just simply not reasonable to send back a message on a link where it was received.
- second, it violates https://github.com/libp2p/specs/blob/b9efe152c29f93f7a87931c14d78ae11e7924d5a/pubsub/gossipsub/gossipsub-v1.0.md#L268-L270
```
 In all pubsub implementations, we can first check the seen cache before forwarding messages to avoid wastefully republishing the same message multiple times.
```
I would qualify this as wasteful republishing.

cskiraly:
@richard-ramos #2376 is not really a fix to this issue.

While in a small testcase A might be the original sender, that's not the main point here. Also, we might not know who is the original sender.
The issue instead is that:
- C->A should not happen because A->C already happened,
- similarly, B->A should not happen, since A->B already happened.

PR #3016, opened 2026-09-02
fix(gossipsub): cancel a queued relay when the peer already has the message

gmelodie:
## Summary

GossipSub sends a message on a link where that same message already travelled. Issue #1081 reports the 3-node case: A publishes, B and C relay to each other, and the copies come back to A.

PR #2376 excluded `msg.fromPeer` from the relay set. That closes one topology, and the issue author rejected it as the fix: a node often does not know the original sender, and `msg.fromPeer` is a field of the message, not a property of the link that delivered it.

This PR adds the link-local rule instead. A relay that sits in the low-priority queue is dropped when the node learns that the target peer already holds the message.

`PubSubPeer` tags each queued relay with the `SaltedId` of the message it carries. `cancelQueuedRelays` makes one pass over the low-priority queue and drops every entry whose id the peer already holds. A peer holds a message in two cases, and both are facts about our own connections, so no remote field can forge them:

- the peer sent us that message. The duplicate branch of `rpcHandler` collects the id.
- the peer sent an IDONTWANT for it. `holdsIDontWant` reads the same per-peer history that the broadcast filter reads, so a relay queued before the IDONTWANT arrived is dropped too.

`rpcHandler` gathers the duplicate ids of the whole RPC, calls `handleControl` so the IDONTWANTs of that RPC are recorded first, and then scans the queue at most once. An RPC that carries no duplicate and no IDONTWANT never touches the queue and allocates nothing.

`sendNonHighPriorityTask` now waits for the pending high-priority writes before it pops the next queued message, instead of after. A message that waits stays in the queue, so it stays cancellable. It also stays counted against `maxLowPriorityQueueLen` for that time.

`handleIDontWant` keeps the behavior it has on master. The `IDontWantMaxCount` cap per heartbeat still limits how many ids one peer can make us salt.

`isMsgInIdontWant` in `validateAndRelay` now calls `holdsIDontWant` as well, so the history walk has one implementation. That instantiation move makes the `nimcrypto/hash` import of `floodsub.nim` unused, which `UnusedImport` rejects as an error, so the import is dropped.

The cancelled bytes land on `libp2p_gossipsub_saved_bytes` under the label value `cancelled_relay`.

The `msg.fromPeer` exclusion from #2376 stays in place.

## Affected Areas

- [x] Gossipsub
  Relay path in `validateAndRelay`, the duplicate branch of `rpcHandler`, and the IDONTWANT history lookup. `PubSubPeer` gains one field on the queued message and two procs.

- [ ] Transports

- [x] Protocol Logic
  `sendEncoded`, `sendResponse` and `broadcastResponse` take an optional `relayedSaltedId`. The parameter defaults to `Opt.none`, so every other caller is unchanged. `sendResponse` tags the RPC only when it carries exactly one message, so a cancel can never drop a message the peer still needs.

- [ ] Build / Tooling

- [ ] Other

- **Nimbus:**
- **Waku:**
- **Codex:**

A fleet run is still open. Gossipsub carries the whole message load, so the change wants fleet numbers before it merges.

## Impact on Library Users

No API break. The three send procs gain a trailing optional parameter, and the existing call sites compile unchanged.

Behavior change: a node no longer sends a relay that is still queued for a peer once that peer proves it holds the message. This saves bandwidth on the sender and duplicate processing on the receiver.

The window this closes is the queued one. `determineQueueAction` promotes a low-priority send to `High` when all queues are empty, so an idle peer gets the relay written at once and there is nothing left to cancel. The gain shows up on a backlogged peer, which is the case that costs bandwidth.

## Risk Assessment

Low. The cancel touches the low-priority queue only, so published messages (medium) and control traffic (high) keep their current path.

The scan is one pass over a queue that `maxLowPriorityQueueLen` caps at 1024 by default, with a hash lookup per queued relay, and it runs at most once per RPC. A peer that sends a duplicate or an IDONTWANT in every RPC pays for one scan per RPC, and it cannot make the node hash more ids than `IDontWantMaxCount` per heartbeat.

The messages that the cancel drops are the ones the peer already has, so the drop cannot lose a delivery. No wire format changes, so old and new nodes interoperate.

## References

- Issue: https://github.com/vacp2p/nim-libp2p/issues/1081
- Prior PR: #2376
- Spec: https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md, "check the seen cache before forwarding messages to avoid wastefully republishing the same message multiple times"

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -321,25 +321,12 @@ proc handleIHave*(
     g.rng.shuffle(res.messageIDs)
     return res
 
-proc handleIDontWant*(
-    g: GossipSub, peer: PubSubPeer, iDontWants: seq[ControlIWant]
-): int =
-  ## Records what the peer does not want, and counts the bytes saved by the relays it cancels.
-  var unwanted = initHashSet[SaltedId]()
-
+proc handleIDontWant*(g: GossipSub, peer: PubSubPeer, iDontWants: seq[ControlIWant]) =
   for dontWant in iDontWants:
     for messageId in dontWant.messageIDs:
       if peer.iDontWants[0].len >= IDontWantMaxCount:
         break
-
-      let saltedId = g.salt(messageId)
-      peer.iDontWants[0].incl(saltedId)
-      unwanted.incl(saltedId)
-
-  peer.cancelQueuedRelays(
-    proc(saltedId: SaltedId): bool {.gcsafe, raises: [].} =
-      saltedId in unwanted
-  )
+      peer.iDontWants[0].incl(g.salt(messageId))
 
 proc handleIWant*(
     g: GossipSub, peer: PubSubPeer, iwants: seq[ControlIWant]
--- a/libp2p/protocols/pubsub/gossipsub/behavior.nim
+++ b/libp2p/protocols/pubsub/gossipsub/behavior.nim
@@ -321,12 +321,25 @@ proc handleIHave*(
     g.rng.shuffle(res.messageIDs)
     return res
 
-proc handleIDontWant*(g: GossipSub, peer: PubSubPeer, iDontWants: seq[ControlIWant]) =
+proc handleIDontWant*(
+    g: GossipSub, peer: PubSubPeer, iDontWants: seq[ControlIWant]
+): int =
+  ## Records what the peer does not want, and counts the bytes saved by the relays it cancels.
+  var unwanted = initHashSet[SaltedId]()
+
   for dontWant in iDontWants:
     for messageId in dontWant.messageIDs:
       if peer.iDontWants[0].len >= IDontWantMaxCount:
         break
-      peer.iDontWants[0].incl(g.salt(messageId))
+
+      let saltedId = g.salt(messageId)
+      peer.iDontWants[0].incl(saltedId)
+      unwanted.incl(saltedId)
+
+  peer.cancelQueuedRelays(
+    proc(saltedId: SaltedId): bool {.gcsafe, raises: [].} =
+      saltedId in unwanted
+  )
 
 proc handleIWant*(
     g: GossipSub, peer: PubSubPeer, iwants: seq[ControlIWant]
--- a/libp2p/protocols/pubsub/pubsubpeer.nim
+++ b/libp2p/protocols/pubsub/pubsubpeer.nim
@@ -121,7 +121,7 @@ type
     data: seq[byte]
     useCustomStream: bool
     relayedSaltedId: Opt[SaltedId]
-      ## Set when this RPC carries one relayed message and was not split, so the send stays cancellable.
+      ## Id of the one relayed message this RPC carries, when it was not split.
 
   RpcMessageQueue* = ref object
     # Tracks async tasks for sending high-priority peer-published messages.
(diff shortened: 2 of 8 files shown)


libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -752,6 +790,14 @@ proc sendNonHighPriorityTask(p: PubSubPeer) {.async: (raises: [CancelledError]).
       await p.rpcmessagequeue.dataAvailableEvent.wait()
       p.rpcmessagequeue.dataAvailableEvent.clear()
 
+    while p.rpcmessagequeue.sendPriorityQueue.len > 0:
+      p.clearSendPriorityQueue()
+      # One wait on the last future costs less than one wait per future.
+      if p.rpcmessagequeue.sendPriorityQueue.len > 0:
+        # `race` keeps the awaited future alive when this task is cancelled.
+        discard await race(p.rpcmessagequeue.sendPriorityQueue[^1])
+
```
richard-ramos, line 836: Looping the whole low priority queue seems not very efficient.

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -568,11 +576,38 @@ proc dropNonHighPriorityMessage(
...
+    queued.relayedSaltedId.withValue(id):
+      cancel = isUnwanted(id)
+
+    if not cancel:
+      p.rpcmessagequeue.lowPriorityQueue.addLast(queued)
+      continue
+
+    savedBytes += queued.data.len
+    when defined(pubsubpeer_queue_metrics):
+      libp2p_gossipsub_low_priority_queue_size.dec(labelValues = [$p.peerId])
```
richard-ramos, line 629: Same here 🤔

libp2p/protocols/pubsub/gossipsub/behavior.nim
```diff
@@ -321,12 +321,25 @@ proc handleIHave*(
     g.rng.shuffle(res.messageIDs)
     return res
 
-proc handleIDontWant*(g: GossipSub, peer: PubSubPeer, iDontWants: seq[ControlIWant]) =
+proc handleIDontWant*(
+    g: GossipSub, peer: PubSubPeer, iDontWants: seq[ControlIWant]
+): int =
+  ## Records what the peer does not want, and counts the bytes saved by the relays it cancels.
+  var unwanted = initHashSet[SaltedId]()
```
richard-ramos: this is doing an allocation for every RPC received. It is part of the hot path

richard-ramos: I also wonder if we could avoid this temporary set? because the accpetedIDs are stored both in `peer.iDontWants[0]` and in `unwanted`, and the queue scanning is hashing each entry again to check membership in `unwanted`,  so perhaps we can use `peer.iDontWants[0]` instead

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -120,6 +120,8 @@ type
     # separate routine.
     data: seq[byte]
     useCustomStream: bool
+    relayedSaltedId: Opt[SaltedId]
```
richard-ramos, line 123: Adding this here means that every message in the medium priority queue will be doing an allocation that wont be used. This should probably not go here

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -568,11 +576,38 @@ proc dropNonHighPriorityMessage(
     raiseAssert "high-priority messages are not dropped via queue overflow scoring"
   return newFutureCompleted[void]()
 
+proc cancelQueuedRelays*(
+    p: PubSubPeer, isUnwanted: proc(saltedId: SaltedId): bool {.gcsafe, raises: [].}
+): int =
+  ## Drops every queued relay that `isUnwanted` accepts, in one pass, and counts the saved bytes.
+  if p.rpcmessagequeue.lowPriorityQueue.len == 0:
+    return 0
```
vladopajic: this early return seems unnecessary becasue `for` loop below, without this return have same outcome

vladopajic (approved):

richard-ramos (changes requested):
Some general observations:
-----

Overall i think it should be reestructured to:
- not allocate a set unlest you have already an id to cancel
-  try to accumulate duplicates and idont wants when processing an RPC protobuf
- try to scan the low priority queue as little as possible

And probably a good idea to also pass this PR thru the eyes of claude to review it.

----

Is this really solving the issue? it seems to maybe solve it partially because 
- when the queues are empty, low priority messages are immediatly sent and never reach the lowPriorityQueue, so there's no cancel to happen
-  once the sender task has removed a queued message, cancellation is too late.
- the changes around `msg.fromPeer` only work if the message is not anonymous (as described in https://github.com/vacp2p/nim-libp2p/issues/1081#issuecomment-4363252886 )
- relying on IDONTWANTs means that this code only applies to large messages

And reading the test case, are we sure they prove this fixes the issue? because you are using a delayed high priority write, to force the relayed message to remain in queue, so we are only provinng cancellations that happen ONLY while the message remains in queue.

I am thinking that the test should have three nodes as described in the issue, with normal streams (that are not stalled), and somehow observe the writes that happen on the gossip streams; to also use an anonymous message, and assert the scenario described by the issue: A -> B and A -> C,  with B -> A and C -> A not writing the message.

----

For gossipsub due to the impact this has, we need to test this PR in the fleet to make sure we dont introduce regressions

Issue #692, opened 2022-02-02
feat: descore peers for which message id generation fails


emizzle:
In `gossipsub` and `floodsub` `rpcHandler` procs, descore peers for which `ValidationResult.Reject` is returned. See https://github.com/status-im/nim-libp2p/pull/688#discussion_r794334077 for more info.

gmelodie:
Fixed gossipsub in 0911cb20f.

PR #3017, opened 2026-09-02
fix(floodsub): punish nodes sending malformed messages

gmelodie:
## Summary

Closes the FloodSub half of #692.

`3b718baa9` ("feat: allow msgIdProvider to fail (#688)") made message id generation fallible and left the same TODO in both protocols: `# TODO: descore peers due to error during message validation (malicious?)`. `0911cb20f` ("chore(gossipsub): cleanups (#1096)") satisfied the GossipSub half with `punishInvalidMessage`. FloodSub kept the TODO because `punishInvalidMessage` is a GossipSub proc: it needs `peerStats`, `TopicInfo` and the v1.1 score, none of which FloodSub has.

A score does not fit FloodSub. Four of the seven v1.1 terms (P1 time in mesh, P3 mesh delivery rate, P3b mesh failure penalty, P7 behaviour penalty) are mesh-relative, and FloodSub has no mesh. Every score consumer in GossipSub is a mesh or gossip decision (`behavior.nim:198`, `:285`, `:296`, `:336`, `:542`, `:688`, `scoring.nim:140`), so the only reachable consumer in FloodSub is disconnect. That is a rate limiter, not a score.

So this PR gives FloodSub the overhead token bucket instead. A peer that sends messages we decode but cannot hash now burns its byte budget, and gets disconnected when `disconnectPeerAboveRateLimit` is set.

The shared machinery moves to `PubSub`, where it removes three existing duplications:

- `disconnectPeer` was defined on `GossipSub` in `gossipsub/scoring.nim`; one copy on `PubSub` now serves both.
- The bucket check and the punishment were written twice, in `gossipsub.nim:rateLimit` and in `scoring.nim:punishInvalidMessage`. Both now call `overBudget` and `punishOverBudget`.
- The `Opt[RateLimit]` to `Opt[TokenBucket]` conversion was inline in GossipSub's `getOrCreatePeer`; it is now `newOverheadBucket`.

`RateLimit` moves from `gossipsub/types.nim` to `pubsub.nim` so both protocols can name it.

## Affected Areas

- [x] Gossipsub
  Pure refactor. `rateLimit` and `punishInvalidMessage` call the shared procs. Order of the metric, the log and the disconnect is preserved, and `libp2p_gossipsub_peers_rate_limit_hits` still fires before the disconnect raises.

- [ ] Transports

- [x] Protocol Logic
  FloodSub gains `overheadRateLimit` and `disconnectPeerAboveRateLimit`, a `getOrCreatePeer` override that builds the peer's token bucket, and validation of both fields in `initPubSub`.

- [ ] Build / Tooling

- [ ] Other

- **Nimbus:**
- **Waku:**
- **Codex:**

## Impact on Library Users

GossipSub behaviour is unchanged. `GossipSubParams` keeps both fields and stays the only place a GossipSub user configures them, including mutation after `initPubSub`, which several tests rely on.

FloodSub users get two new public fields on the `FloodSub` object, both off by default. Set them before `start` and `initPubSub` validates them: a limit with `bytes <= 0` or a non-positive `interval` raises `InitializationError`, as does `disconnectPeerAboveRateLimit` without a limit. This mirrors `validateOverheadRateLimit` in `gossipsub.nim`.

New public API on `PubSub`: `disconnectPeer`, `newOverheadBucket`, `overBudget`, `punishOverBudget`. `RateLimit` keeps its old import path for anyone importing `gossipsub`, because `gossipsub.nim` re-exports `pubsub`.

## Risk Assessment

Low for GossipSub, which keeps reading `g.parameters` at the point of use rather than a cached copy. An earlier version of this change mirrored the two fields into `PubSub` at init time; that silently breaks post-init mutation of `parameters`, which seven in-tree tests do (`test_gossipsub.nim:597`, `:630`, `:659` and `test_gossipsub_scoring.nim:83`, `:122`, `:168`, `:222`). `punishOverBudget` therefore takes the disconnect flag as an argument instead of reading a field.

New behaviour for FloodSub only, and only when a user opts in by setting `overheadRateLimit`. With the field unset, `newOverheadBucket` returns `Opt.none`, `overBudget` is a no-op and the code path is identical to today, minus the TODO.

The FloodSub `getOrCreatePeer` override is not reached from GossipSub: `GossipSub.getOrCreatePeer` calls `procCall PubSub(g).getOrCreatePeer`, which skips it. No peer gets two buckets.

## References

- Issue: #692, "feat: descore peers for which message id generation fails"
- Origin of the TODO: `3b718baa9`, "feat: allow msgIdProvider to fail (#688)"
- GossipSub half: `0911cb20f`, "chore(gossipsub): cleanups (#1096)"

### DIFF
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -99,11 +99,10 @@ method unsubscribePeer*(f: FloodSub, peer: PeerId) =
 
   procCall PubSub(f).unsubscribePeer(peer)
 
-proc punishIfOverBudget(
-    f: FloodSub, peer: PubSubPeer, overhead: int
-) {.async: (raises: [PeerRateLimitError]).} =
+template chargeOverhead(f: FloodSub, peer: PubSubPeer, overhead: int) =
+  # a template, so that a peer within its budget allocates no future
   if not peer.tryCharge(overhead):
-    await f.punishOverBudget(peer, overhead, f.disconnectPeerAboveRateLimit)
+    f.punishOverBudget(peer, overhead, f.disconnectPeerAboveRateLimit)
 
 method rpcHandler*(
     f: FloodSub, peer: PubSubPeer, data: sink seq[byte]
@@ -111,7 +110,7 @@ method rpcHandler*(
   let msgSize = data.len
   var rpcMsg = RPCMsg.decode(move(data)).valueOr:
     debug "failed to decode msg from peer", peer, err = error
-    await f.punishIfOverBudget(peer, msgSize)
+    f.chargeOverhead(peer, msgSize)
     raise newException(PeerMessageDecodeError, "Peer msg couldn't be decoded")
 
   trace "decoded msg from peer", peer, rpcMsg = rpcMsg.shortLog
@@ -124,12 +123,13 @@ method rpcHandler*(
 
     f.handleSubscribe(peer, sub.topic.get(), sub.isSubscribe)
 
+  var invalidBytesSent = 0
   for msg in rpcMsg.messages: # for every message
     let msgIdResult = f.msgIdProvider(msg)
     if msgIdResult.isErr:
       debug "Dropping message due to failed message id generation",
         error = msgIdResult.error
-      await f.punishIfOverBudget(peer, msg.data.len)
+      invalidBytesSent += msg.data.len
       continue
 
     let
@@ -144,13 +144,13 @@ method rpcHandler*(
     if (msg.signature.len > 0 or f.verifySignature) and not msg.verify():
       # always validate if signature is present or required
       debug "Dropping message due to failed signature verification", msgId, peer
-      await f.punishIfOverBudget(peer, msg.data.len)
+      invalidBytesSent += msg.data.len
       continue
 
     if msg.seqno.len > 0 and msg.seqno.len != 8:
       # if we have seqno should be 8 bytes long
       debug "Dropping message due to invalid seqno length", msgId, peer
-      await f.punishIfOverBudget(peer, msg.data.len)
+      invalidBytesSent += msg.data.len
       continue
 
     if f.addSeen(saltedId):
@@ -184,6 +184,7 @@ method rpcHandler*(
     trace "Forwared message to peers", peers = toSendPeers.len
 
   f.updateMetrics(rpcMsg)
+  f.chargeOverhead(peer, invalidBytesSent)
 
 method init*(f: FloodSub) =
   proc handler(stream: Stream, proto: string) {.async: (raises: [CancelledError]).} =
--- a/libp2p/protocols/pubsub/floodsub.nim
+++ b/libp2p/protocols/pubsub/floodsub.nim
@@ -29,11 +29,11 @@ const FloodSubCodec* = "/floodsub/1.0.0"
 type FloodSub* = ref object of PubSub
   floodsub*: PeerTable # topic to remote peer map
   seen*: TimedCache[SaltedId]
-    # Early filter for messages recently observed on the network
-    # We use a salted id because the messages in this cache have not yet
-    # been validated meaning that an attacker has greater control over the
-    # hash key and therefore could poison the table
+    # salted: these ids are unvalidated, so a plain key lets an attacker poison the table
   seenSalt: array[32, byte] # random data used as salt
+  # gossipsub rejects these two at init and reads GossipSubParams instead
+  overheadRateLimit*: Opt[RateLimit]
+  disconnectPeerAboveRateLimit*: bool
 
 proc salt*(f: FloodSub, msgId: MessageId): SaltedId =
   var hash: sha256
@@ -99,11 +99,19 @@ method unsubscribePeer*(f: FloodSub, peer: PeerId) =
 
   procCall PubSub(f).unsubscribePeer(peer)
 
+proc punishIfOverBudget(
+    f: FloodSub, peer: PubSubPeer, overhead: int
+) {.async: (raises: [PeerRateLimitError]).} =
+  if not peer.tryCharge(overhead):
+    await f.punishOverBudget(peer, overhead, f.disconnectPeerAboveRateLimit)
+
 method rpcHandler*(
     f: FloodSub, peer: PubSubPeer, data: sink seq[byte]
 ) {.async: (raises: [CancelledError, PeerMessageDecodeError, PeerRateLimitError]).} =
+  let msgSize = data.len
   var rpcMsg = RPCMsg.decode(move(data)).valueOr:
     debug "failed to decode msg from peer", peer, err = error
+    await f.punishIfOverBudget(peer, msgSize)
     raise newException(PeerMessageDecodeError, "Peer msg couldn't be decoded")
 
   trace "decoded msg from peer", peer, rpcMsg = rpcMsg.shortLog
@@ -121,7 +129,7 @@ method rpcHandler*(
     if msgIdResult.isErr:
       debug "Dropping message due to failed message id generation",
         error = msgIdResult.error
-      # TODO: descore peers due to error during message validation (malicious?)
+      await f.punishIfOverBudget(peer, msg.data.len)
       continue
 
     let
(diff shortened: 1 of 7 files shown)


libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -452,6 +452,24 @@ method getOrCreatePeer*(
 
   return pubSubPeer
 
+proc disconnectPeer*(p: PubSub, peer: PubSubPeer) {.async: (raises: []).} =
+  try:
+    await p.switch.disconnect(peer.peerId)
+  except CatchableError as e:
+    trace "Failed to close connection", peer, errName = e.name, description = e.msg
+
```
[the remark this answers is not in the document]
vladopajic: this is valid

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -99,11 +99,19 @@ method unsubscribePeer*(f: FloodSub, peer: PeerId) =
 
   procCall PubSub(f).unsubscribePeer(peer)
 
+proc punishIfOverBudget(
+    f: FloodSub, peer: PubSubPeer, overhead: int
+) {.async: (raises: [PeerRateLimitError]).} =
+  if not peer.tryCharge(overhead):
+    await f.punishOverBudget(peer, overhead, f.disconnectPeerAboveRateLimit)
```
vladopajic: naming-vise this code seems bad:

```
proc punishIfOverBudget
  if not tryCharge
    await f.punishOverBudget
```
 it's same name in different type

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -99,11 +99,19 @@ method unsubscribePeer*(f: FloodSub, peer: PeerId) =
...
+
 method rpcHandler*(
     f: FloodSub, peer: PubSubPeer, data: sink seq[byte]
 ) {.async: (raises: [CancelledError, PeerMessageDecodeError, PeerRateLimitError]).} =
+  let msgSize = data.len
   var rpcMsg = RPCMsg.decode(move(data)).valueOr:
```
vladopajic, line 111: why don't do this budget check here? why is it not only once checked ?

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -136,11 +144,13 @@ method rpcHandler*(
     if (msg.signature.len > 0 or f.verifySignature) and not msg.verify():
       # always validate if signature is present or required
       debug "Dropping message due to failed signature verification", msgId, peer
+      await f.punishIfOverBudget(peer, msg.data.len)
```
vladopajic: `punishIfOverBudget` will unnecessary do `await` in most cases - await is only needed if peer is disconnected. it would be more suitable to have template that checks and only awaits if peer is disconnected.

tests/libp2p/pubsub/component/test_floodsub.nim
```diff
@@ -346,3 +365,46 @@ suite "FloodSub Component":
 
     node.unsubscribePeer(peerId)
     check node.floodsub.len == 0
+
+  asyncTest "FloodSub charges the overhead budget when message id generation fails":
+    const bytes = 10
```
vladopajic: `bytes`  ?

vladopajic (approved):

libp2p/protocols/pubsub/floodsub.nim
vladopajic: `punishIfOverBudget` and `punishOverBudget` is the same name

libp2p/protocols/pubsub/floodsub.nim
gmelodie: `punishIfOverBudget` means "i want to punish in case it's over budget", so if the msg is valid (i.e. outside of the valueor block) i do not want to punish even if it's over budget, since the budget is for invalid messages. So we only call that if message is invalid

vladopajic: Image

this comment was about that ^, `punishIfOverBudget` was called many times in this procs, so i was curious if it was really necessary to call it always

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -124,12 +123,13 @@ method rpcHandler*(
 
     f.handleSubscribe(peer, sub.topic.get(), sub.isSubscribe)
 
+  var uselessBytes = 0
```
vladopajic: `uselessBytes` -> `invalidBytesSent`

libp2p/protocols/pubsub/floodsub.nim
gmelodie: Ah!

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -247,8 +258,33 @@ method publish*(
...
+    return err("floodsub: overheadRateLimit.bytes parameter error, Must be > 0")
+  if limit.interval <= ZeroDuration:
+    return err("floodsub: overheadRateLimit.interval parameter error, Must be > 0")
+  ok()
+
+method getOrCreatePeer*(
+    f: FloodSub, peerId: PeerId, protosToDial: seq[string], protoNegotiated: string = ""
+): PubSubPeer =
+  let peer = procCall PubSub(f).getOrCreatePeer(peerId, protosToDial, protoNegotiated)
+  peer.overheadRateLimitOpt = newOverheadBucket(f.overheadRateLimit)
```
richard-ramos: It seems we're calling newOverheadBucket even if getOrCreatePeer returns a peer....  so this means that a peer could just reopen a stream to have their overhead bucket reset (also... this ends up unnecesarily allocating a `TokenBucket` and an `AsyncEvent`)

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -116,12 +123,13 @@ method rpcHandler*(
...
     f.handleSubscribe(peer, sub.topic.get(), sub.isSubscribe)
 
+  var invalidBytesSent = 0
   for msg in rpcMsg.messages: # for every message
     let msgIdResult = f.msgIdProvider(msg)
     if msgIdResult.isErr:
       debug "Dropping message due to failed message id generation",
         error = msgIdResult.error
-      # TODO: descore peers due to error during message validation (malicious?)
+      invalidBytesSent += msg.data.len
```
richard-ramos: I think we gotta check byteSize(msg) instead of msg.data.len, because otherwise i could send a message with no data, (that will still contain other protobuf fields),  and not use tokens from the bucket, right?

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -29,11 +29,11 @@ const FloodSubCodec* = "/floodsub/1.0.0"
...
   seen*: TimedCache[SaltedId]
-    # Early filter for messages recently observed on the network
-    # We use a salted id because the messages in this cache have not yet
-    # been validated meaning that an attacker has greater control over the
-    # hash key and therefore could poison the table
+    # salted: these ids are unvalidated, so a plain key lets an attacker poison the table
   seenSalt: array[32, byte] # random data used as salt
+  # gossipsub rejects these two at init and reads GossipSubParams instead
+  overheadRateLimit*: Opt[RateLimit]
+  disconnectPeerAboveRateLimit*: bool
```
richard-ramos, line 36: Do you think these fields should be added to the `init` method of `FloodSub`?

libp2p/protocols/pubsub/pubsubpeer.nim
```diff
@@ -202,6 +206,24 @@ proc getAgent*(peer: PubSubPeer): string =
...
+
+  Opt.none(TokenBucket)
+
+proc tryCharge*(peer: PubSubPeer, overhead: int): bool =
+  ## Charges `overhead` bytes of useless data to the peer, false when it cannot pay.
+  peer.overheadRateLimitOpt.withValue(overheadRateLimit):
+    return overheadRateLimit.tryConsume(overhead)
+
+  true
+
```
richard-ramos, line 229: maybe this?

```suggestion
proc tryCharge*(peer: PubSubPeer, overhead: int): bool =
  ## Charges `overhead` bytes of useless data to the peer, false when it cannot pay.
  
   if overhead <= 0:
     return true
  
  peer.overheadRateLimitOpt.withValue(overheadRateLimit):
    return overheadRateLimit.tryConsume(overhead)

  true

```

libp2p/protocols/pubsub/floodsub.nim
```diff
@@ -174,6 +184,7 @@ method rpcHandler*(
     trace "Forwared message to peers", peers = toSendPeers.len
 
   f.updateMetrics(rpcMsg)
+  f.chargeOverhead(peer, invalidBytesSent)
```
richard-ramos: I wonder if we should instead of callin this here, and also deal with `invalidBytes` in 147 and 153, if we should instead do a f.chargeOverhead(peer, msg.byteSize()) instead?  that way we dont ave to verify  signatures of all messages before punishing the peer

richard-ramos (commented):
Sorry for the delay

richard-ramos (approved):

PR #3024, opened 2026-09-03
chore: native hkdf and autotls pem encoding

gmelodie:
## Summary

Drops `bearssl/[kdf, hash, pem]` by replacing thin wrappers with the equivalent code over dependencies already in the tree.

`hkdf.nim` becomes RFC 5869 over `nimcrypto` HMAC-SHA256, with the `T(i)` chain written out. `outputs` is one continuous OKM stream, which `noise.nim` relies on and the BearSSL stream API left implicit. A new test locks it. Key material is now zeroed on exit.

`pemEncode` moves to `autotls/utils.nim` over `stew/base64`. The BearSSL wrapper took `unsafeAddr data[0]` with no length check, so an empty key crashed.

## Affected Areas

- [x] Protocol Logic (Noise key derivation; derived keys unchanged)
- [x] Other (`crypto/hkdf.nim`, `crypto/rng.nim`, `autotls/{service,utils}.nim`)

## Impact on Library Users

`pemEncode` wraps at 64 columns per RFC 7468. BearSSL wrapped at 76, because `BR_PEM_LINE64` was never passed. Same key, different line layout; every PEM reader takes both.

`newBearSslRng` is removed. It had no in-tree caller; use `newRng`.

`hkdf.nim` no longer re-exports all of `nimcrypto`.

## Risk Assessment

HKDF returns the same bytes. Covered by the two existing RFC 5869 vectors plus a 3 x 32 expansion. `pemEncode` is covered at two lines, at the exact column boundary, and empty.

## References

RFC 5869, RFC 7468.

### DIFF
--- a/libp2p/autotls/service.nim
+++ b/libp2p/autotls/service.nim
@@ -4,7 +4,6 @@
 {.push raises: [].}
 
 import sequtils
-import bearssl/pem
 import chronos, chronicles, net, results, uri
 import chronos/streams/tlsstream
 from times import DateTime, now, toTime, toUnix
--- a/libp2p/autotls/utils.nim
+++ b/libp2p/autotls/utils.nim
@@ -3,7 +3,7 @@
 {.push raises: [].}
 
 import chronos, chronicles, strutils
-import stew/base36
+import stew/[base36, base64]
 import
   ../errors,
   ../peerid,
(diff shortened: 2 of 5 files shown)


libp2p/crypto/rng.nim
```diff
@@ -24,15 +24,6 @@ proc newRng*(): Rng =
     return nil
   rng
 
-proc newBearSslRng*(drbg: ref HmacDrbgContext): Rng =
```
richard-ramos, line 27: This is used downstream in nimbus-eth2

richard-ramos (changes requested):

richard-ramos (approved):
It seems you're missing an `import bearssl/[hash, rand]` in failing tests

vladopajic (approved):

PR #3025, opened 2026-09-03
fix(rendezvous): discover no longer allocates one int per registration

gmelodie:
## Summary

A DISCOVER with no namespace materialised a `seq[int]` of every registration index before it read one entry (the `toSeq` at `rendezvous.nim:264`); a small message drove a large allocation per stream. An inline iterator now yields that range. The loop also copied a whole registration per skipped entry, and read the clock twice, which wrapped a just-expired TTL into a huge `uint64`.

## Affected Areas

## Impact on Library Users

No API or wire change.

## Risk Assessment

Same records and cookies. CI must run `tests/libp2p/discovery/test_rendezvous.nim`.


tests/libp2p/test_dialer.nim
```diff
@@ -84,9 +83,12 @@ suite "Dialer":
     let src = makeStandardSwitch()
     switches.add(src)
     await src.start()
-    check not await src.connect(dst.peerInfo.peerId, dst.peerInfo.addrs).withTimeout(
-      1000.millis
-    )
+
+    # lsquic can hide dst's close from the dialer (vacp2p/nim-lsquic#162)
```
richard-ramos: This should no longer be the case:
https://github.com/vacp2p/nim-lsquic/issues/162

richard-ramos (approved):

vladopajic (approved):

PR #3026, opened 2026-09-03
chore(wstransport): bump websock to 0.4.1

gmelodie:
## Summary

`WsTransport` passed its BearSSL DRBG to websock through `bearSslDrbgRef`, the overload websock deprecates. It now passes a closure over `Rng.generate`. That drops the last caller of `bearSslDrbg` and `bearSslDrbgRef`, so both templates go too.

The v0.4.1 bump lifts websock's chronos ceiling from `< 4.4.0` to `< 4.6.0`. libp2p asks for `chronos >= 4.2.2` with no ceiling, so chronos 4.4 conflicts against v0.4.0.

## Affected Areas

- [x] Transports
- [x] Build / Tooling

## Impact on Library Users

Breaking for direct importers of `libp2p/crypto/rng` that call `bearSslDrbg` or `bearSslDrbgRef`. `crypto.nim` never re-exported either one. `bearSslPrng` stays.

## Risk Assessment

Low. Same DRBG, same frames, wire format unchanged. `tests/libp2p/transports/test_ws.nim` covers both handshakes.


libp2p/crypto/rng.nim
```diff
@@ -33,12 +33,6 @@ proc newBearSslRng*(drbg: ref HmacDrbgContext): Rng =
     return nil
   Rng(drbg: drbg)
 
-template bearSslDrbg*(rng: Rng): untyped =
-  rng.drbg[]
-
-template bearSslDrbgRef*(rng: Rng): untyped =
-  rng.drbg
-
```
richard-ramos, line 41: This is a breaking change for logos-delivery:

- https://github.com/logos-messaging/logos-delivery/blob/676459e67101d66cf03e4634ee2144fc0c037306/logos_delivery/waku/discovery/waku_discv5.nim#L96
 
- https://github.com/logos-messaging/logos-delivery/blob/676459e67101d66cf03e4634ee2144fc0c037306/tests/testlib/common.nim#L36  - I think this one is easier to fix since `generate` from `libp2p/crypto/rng.nim` could be used instead (cc: @Ivansete-status) .

gmelodie: Does this require some action on my side rn? I think it's just going to take some tweaks when we release next time, no?

richard-ramos: Yes. 
Mark these as deprecated and not remove them to not introduce breaking changes, specially since the first item there requires access to a `ref  HmacDrbgContext`, which logos-delivery does not have access to, since they create the rng via `crypto.newRng()` and with these removed, they will no longer have access to the underlying bearssl hmacdrbg reference

richard-ramos (approved):

vladopajic (approved):

PR #3027, opened 2026-09-03
fix(utils): make the offsettedseq apply and flushIf procs compile

gmelodie:
## Summary

The three `apply` overloads in `libp2p/utils/offsettedseq.nim` passed an undeclared `pred` instead of their `op`, and the mutating ones plus `flushIf` took a non-var receiver that `sequtils.apply` cannot write to. Nim checks a generic body only at instantiation, and nothing called them, so this never failed a build.

Fixed, `flushIf` now delegates to `flushIfIt`, and `tests/libp2p/utils/test_offsettedseq.nim` is the caller that keeps them compiled.

## Impact and Risk

None: no code can call a proc that does not compile. Low risk, one leaf module, `flushIfIt` unchanged.


richard-ramos (approved):
Nice catch! I'm surprised this was never captured before

vladopajic (approved):

PR #3028, opened 2026-09-03
fix(rendezvous): cap the namespace table and prune empty namespaces

gmelodie:
## Summary

`RegistrationLimitPerPeer` caps one peer id, and peer ids are free, so the `namespaces` table had no limit. This adds `RendezVousConfig.maxNamespaces` (default 1000) and refuses a new namespace once the table is full; an existing one still accepts registrations.

The table also never shrank: `deletesRegister` left an empty `seq[int]` under a dead key, which would make the cap permanent. `pruneNamespaces` deletes it.

## Affected Areas

- [x] Protocol Logic

No wire change; a refusal reuses `NotAuthorized`. Waku should confirm 1000 fits a fleet node. Nimbus, Codex: N/A.

## Impact on Library Users

`RendezVousConfig.new` gets an optional parameter, so call sites keep working. `Discover` on a fully expired namespace now answers `InvalidNamespace` instead of an empty `Ok`; `request` maps both to an empty list.

## Risk Assessment

One peer id can still squat 1000 namespaces until they expire. A per-peer cap is the follow-up.


libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -208,6 +216,9 @@ proc register*[E](
     return stream.sendRegisterResponseError(InvalidSignedPeerRecord, pr.error())
   if rdv.countRegister(stream.peerId) >= RegistrationLimitPerPeer:
     return stream.sendRegisterResponseError(NotAuthorized, "Registration limit reached")
+  if rdv.namespaces.len >= rdv.config.maxNamespaces and
```
richard-ramos: This line should probably be put instead inside `save`, and make save return a `Result` depending on success or not. Because `rdv.namespaces` is populated on other places, besides `register`.

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -598,17 +609,23 @@ proc new*(
   rdv.setup(switch)
   return rdv
 
+proc pruneNamespaces[E](rdv: GenericRendezVous[E]) =
+  var emptyNamespaces: seq[string]
+  for ns, indexes in rdv.namespaces.mpairs():
+    indexes.keepItIf(it >= rdv.registered.offset)
```
richard-ramos: Can we also remove expired registrations here? because `flushItIf` iiuc, removes expired registrations starting from the beginning of the sequence, but there's no guarantee that the sequence is ordered.

libp2p/protocols/rendezvous/rendezvous.nim
```diff
@@ -50,14 +51,17 @@ type RendezVousConfig* = object
   maxDuration*: Duration
   minTTL*: uint64
   maxTTL*: uint64
+  maxNamespaces*: int
```
richard-ramos, line 54: Just noticed that this is a breaking change for logos-delivery that they would not realize immediatly, because they do something like `RendezvousConfig(minTTL: ... , maxTTL: ...), so once they bump to new version, `maxNamespaces` would automatically be set to 0 since they wouldnt know about it....

not sure how to fix it? perhaps by assuming that when `maxNamespaces == 0`, `MaximumNamespaces` should be used instead? wdyt?

richard-ramos (approved):

vladopajic (approved):

PR #3033, opened 2026-09-04
fix(service-disco): cache the advertisement bytes so rotations keep the seqNo

gmelodie:
## Summary

`addProvidedService` stored the advert bytes only for an explicit advert, so a node without one rebuilt its record on every rotation, each with a later `seqNo`. Rotation leaves the old record on the previous registrar, and `lookup` merges registrar results into a `HashSet[Advertisement]` that dedupes by value, so the node shows up twice. Storing the encoded bytes for every service pins one record and one `seqNo`.

## Affected Areas

- [x] Peer Management / Discovery: `libp2p/protocols/service_discovery/advertiser.nim`.

N/A. No downstream project consumes service discovery yet.

## Impact on Library Users

`startAdvertising(service)` with no advert now republishes the same record instead of a fresh one. An explicit advert behaves as before.

## Risk Assessment

A pinned record misses a later address change until the service restarts. Re-encoding only on a content change belongs in its own PR.

## References

Split out of #3029.

### DIFF
--- a/libp2p/protocols/service_discovery/advertiser.nim
+++ b/libp2p/protocols/service_discovery/advertiser.nim
@@ -50,7 +50,7 @@ proc getAdvertBytes(disco: ServiceDiscovery, explicit: Opt[seq[byte]]): Opt[seq[
   Opt.some(extRecord.encode())
 
 proc advertFor(disco: ServiceDiscovery, serviceId: ServiceId): seq[byte] =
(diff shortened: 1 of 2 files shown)


vladopajic (approved):

richard-ramos (approved):
> A pinned record misses a later address change until the service restarts. Re-encoding only on a content change belongs in its own PR.

Pls create an issue for this

PR #3037, opened 2026-09-04
fix(addr-mgr): count the peers which observe an address, not the reports

gmelodie:
`AddressManager` picks the external address by a count over the observation window. The window stored bare `MultiAddress` values, so `minCount` counted identify messages and not the peers behind them: one peer which reported the same address three times reached the threshold alone.

The window now stores one entry per observer peer id. A new report from a peer replaces the previous report of that peer, so a peer holds one slot of `maxSize` and contributes one vote. `mostObservedProtosAndPorts` and `externalAddrFor` are unchanged, and the count they run is now a count of peers.

### DIFF
--- a/tests/libp2p/test_address_manager.nim
+++ b/tests/libp2p/test_address_manager.nim
@@ -13,6 +13,9 @@ import ../tools/[unittest, crypto, switch_builder, multiaddress, lifecycle]
 
 const VerifyInterval = 10.milliseconds
 
+proc randomPeerId(): PeerId =
+  PeerId.random(rng()).expect("the rng produces a valid peer id")
+
 proc makeManager(
     maxSize = DefaultObservedAddrMaxSize,
     minCount = DefaultObservedAddrMinCount,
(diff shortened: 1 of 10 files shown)


tests/libp2p/test_address_manager.nim
```diff
@@ -13,6 +13,9 @@ import ../tools/[unittest, crypto, switch_builder, multiaddress, lifecycle]
 
 const VerifyInterval = 10.milliseconds
 
+proc randomPeerId(): PeerId =
+  PeerId.random(rng()).expect("the rng produces a valid peer id")
```
vladopajic: this should already exists; and should be reused in this pr (there is same utility proc in other file).

this can maybe be done in separate effort - to standardize on template that creates random peer id (there are few different ways in tests).

ideally this should be `PeerId.random()` template in tests, that has `rng` argument with default value `rng()`.

vladopajic (approved):

richard-ramos (approved):

PR #3040, opened 2026-09-07
chore(perf): rename uploadSize to responseSize

gmelodie:
The `/perf/1.0.0` server handler reads an 8-byte big-endian size field and sends that many bytes back. The variable held that count but was named `uploadSize`, which names the wrong direction: it is the size of the server's response, not of the client's upload.

This renames it to `responseSize`, and renames `uploadSizeBuffer` to `sizeBuffer`. No behaviour changes.

An earlier version of this PR also capped the size to a configurable maximum. That is wrong: the spec gives the client the size, up to max uint64, and a server-side cap makes the server send a short response with no error, which produces a wrong throughput number for the client. Dropped.

Perf spec: https://github.com/libp2p/specs/blob/master/perf/perf.md

### DIFF
--- a/libp2p/protocols/perf/server.nim
+++ b/libp2p/protocols/perf/server.nim
@@ -23,9 +23,9 @@ proc new*(T: typedesc[Perf]): T =
     try:
       trace "Received benchmark performance check", stream
 
-      var uploadSizeBuffer: array[8, byte]
-      await stream.readExactly(addr uploadSizeBuffer[0], 8)
-      var uploadSize = uint64.fromBytesBE(uploadSizeBuffer)
+      var sizeBuffer: array[8, byte]
+      await stream.readExactly(addr sizeBuffer[0], 8)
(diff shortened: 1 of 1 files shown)


vladopajic (commented):
not sure if it makes sense to do it

spec:
> Tell the server how many bytes we want the server to send us as a single big-endian uint64 number. Zero is a valid number, so is the max uint64 value.

client needs to tell what maximum is; server cant cap to anything that is not by the protocol as using nim server with go client (that uses value larger that this const) will give invalid result.


description:
> That value was used directly, so the response size a single request produces was fixed by the peer alone and could not be tuned by the operator.

^ that was by the spec

gmelodie:
@vladopajic indeed. Changed this to be only the variable name change.

richard-ramos (approved):

vladopajic (approved):
`refactor` is inappropriate label when pr does a rename; changed to `chore`

PR #3042, opened 2026-09-07
fix(gossipsub): ignore RPCs from graylisted peers

### LINKED ISSUE #1141: When a peer's score drops below this threshold, the peer is graylisted and its RPCs are ignored
Implement this part of the GossipSub spec: https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#score-thresholds

> GraylistThreshold: when a peer's score drops below this threshold, the peer is graylisted and its RPCs are ignored. This threshold must be negative, and less than the gossip/publish threshold.

gmelodie:
## Summary

`graylistThreshold` had one consumer, `disconnectIfBadScorePeer`, and that one sits behind `disconnectBadPeers`, which defaults to `false`. A graylisted peer therefore kept full access to `rpcHandler`. This adds the gate the spec asks for: `rpcHandler` returns early when `g.isGraylisted(peer, peer.score)`, placed after the decode and after `rateLimit` so the sender still pays the decode penalty and still pays for its bytes.

An ignored RPC includes its subscriptions. go-libp2p still processes those; this does not.

## Affected Areas

- [x] Gossipsub — the `rpcHandler` gate, `isGraylisted` extracted in `scoring.nim` so the predicate has one definition, and the counter `libp2p_gossipsub_graylisted_rpcs{agent}`.

N/A. The default `graylistThreshold` is `-10000.0` and neither Nimbus nor Waku raise it, so the gate is a no-op for the shipped parameters.

## Impact on Library Users

No API change. A peer below `graylistThreshold` stops having its RPCs handled. The gate reads `peer.score`, which `updateScores` refreshes once per `decayInterval`, so it is not instant.

## Risk Assessment

Low. One float compare per RPC on the receive path. Recv observers no longer see traffic from a graylisted peer.

## References

Closes #1141. [Spec: score thresholds](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#score-thresholds).

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -698,6 +698,14 @@ method rpcHandler*(
     g: GossipSub, peer: PubSubPeer, data: sink seq[byte]
 ) {.async: (raises: [CancelledError, PeerMessageDecodeError, PeerRateLimitError]).} =
   let msgSize = data.len
+
+  if g.isGraylisted(peer, peer.score):
+    await rateLimit(g, peer, msgSize)
+    trace "PubSub RPC ignored",
+      peerId = peer.peerId, reason = "graylisted", score = peer.score
+    libp2p_gossipsub_graylisted_rpcs.inc(labelValues = [peer.getAgent()])
+    return
+
   var rpcMsg = RPCMsg.decode(move(data)).valueOr:
     trace "PubSub RPC decode failed",
       err = error, peerId = peer.peerId, messageType = "rpc", messageSize = msgSize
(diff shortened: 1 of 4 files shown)


vladopajic (approved):

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -717,6 +717,12 @@ method rpcHandler*(
     peerId = peer.peerId, messageType = "rpc", messageSize = msgSize
   await rateLimit(g, peer, g.messageOverhead(rpcMsg, msgSize))
 
+  if g.isGraylisted(peer, peer.score):
+    trace "PubSub RPC ignored",
+      peerId = peer.peerId, reason = "graylisted", score = peer.score
+    libp2p_gossipsub_graylisted_rpcs.inc(labelValues = [peer.getAgent()])
+    return
+
```
richard-ramos: This should be moved to line 701 (right after the `let msgSize`, so you avoid doing protobuf decoding and calculating msgOverhead etc to end up ignoring the RPC.

richard-ramos (commented):
Only a single observation and this PR is good to go!

libp2p/protocols/pubsub/gossipsub.nim
richard-ramos: ah! probably a good idea to rate limit too, so it would be something like this:

```nim
  let msgSize = data.len

  if g.isGraylisted(peer, peer.score):
    await rateLimit(g, peer, msgSize)
    trace "PubSub RPC ignored",
      peerId = peer.peerId, reason = "graylisted", score = peer.score
    libp2p_gossipsub_graylisted_rpcs.inc(labelValues = [peer.getAgent()])
    return

```

richard-ramos (approved):

Issue #1318, opened 2025-04-09
Limit duplicate IWANT requests


ufarooqstatus:
On receiving an IHAVE announcement for an unseen message, a peer immediately sends an IWANT request, [even if it has already made IWANT request(s) for the same message.](https://github.com/vacp2p/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub/behavior.nim#L292-#L303) 

IWANT requests add more burden on early message relayers, which increases message dissemination time (and duplicates).

We can fix the maximum number of outstanding IWANT requests for a single message to a reasonable default (maybe 1/heartbeat).

ufarooqstatus:
[rust-libp2p](https://github.com/libp2p/rust-libp2p/blob/master/protocols/gossipsub/src/behaviour.rs#L1252-#L1281)

[go-libp2p](https://github.com/libp2p/go-libp2p-pubsub/issues/606)

Issue #1136, opened 2024-06-24
Keep track of in-flight IWANT requests


diegomrsantos:
Related to https://github.com/vacp2p/nim-libp2p/pull/943

PR #943, opened 2023-08-18
Prevent concurrent IWANT of the same message

diegomrsantos:
closes https://github.com/status-im/nim-libp2p/issues/868

PR #3043, opened 2026-09-07
fix(gossipsub): track in-flight IWANT requests

gmelodie:
A node that receives an IHAVE for a message it lacks sends an IWANT, and nothing records the request. Every peer gossiping the same id gets its own IWANT and the payload arrives from all of them (#1101).

`GossipSub.requestedIWants` now counts outstanding requests and `handleIHave` skips a duplicate while one is in flight. Per-peer state ages in the heartbeat next to `sentIHaves`, so an unanswered request expires after `IWantHistoryLen = 3` heartbeats. `rpcHandler` releases it on arrival, before validation, so a bad answer cannot block the retry.

#943 did this and was reverted by #977. Its three constraints hold: no penalty on an unanswered IWANT (#976), per-peer expiring state, and clearing on delivery from any source.

### References

#1136, #1318, #1101, #943, #977, #976

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -403,7 +403,7 @@ method unsubscribePeer*(g: GossipSub, peer: PeerId) =
 
   pubSubPeer.stopTasks()
 
-  g.releasePeerIWantRequests(pubSubPeer)
+  g.releasePeerIWantRequests(peer)
 
   g.extensionsState.removePeer(peer)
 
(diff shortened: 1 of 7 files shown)


richard-ramos (changes requested):
The idea of the PR makes sense but the code can be simpler and use less memory, because in this PR you're 'duplicating' state by storing data in a global counter and also in the per peer history, so every update must maintain both. Probably makes sense to have a single counter (IMO the global counter, but with some modifications)

With the PR as it is right now:

- If you set `maxIWantsPerMessage`  to be `>1`, you can send the same IHAVE several times. The counter increases but the hashset stores the message id once, so when the request expires, the counter decreases only once, meaning you'll have an invalid counter and future requests will be blocked

- `forgetIWantRequests` removes only the global counter, and the message id stays in the peer history.... so later that entry can expire and decrement the global counter 
- 
Also, using integers in the global counter is not ideal and can quickly become inconsistent. Better to track the PeerIDs instead

- Peer A advertises message X, so we request X. Global state becomes X = 1.
- before peer answers, peer A advertices X again in another RPC
- We request X from A again, so X = 2
- Peer A's per peer HashSet only contains X because it's a set so no duplicates
- Peer A disconnects
- The cleanup removes X, so the global count becomes X = 1
- Now there is no request... but we still belive there's one  :RIP: 

---

besides this, since this PR introduces changes in gossipsub, pls test it on nimbus on the libp2p nodes for couple of days to ensure we're not introducing regressions.

PR #3045, opened 2026-09-07
chore(logs): add logs with background operation

vladopajic:
some background operations (futures) could fail (terminal failure) and would produce no logs to indicate failure. this pr adds log for these cases


gmelodie (approved):
cool!

richard-ramos (approved):

PR #3057, opened 2026-09-08
fix(service-discovery): use Unix seconds for ticket timestamps

### LINKED ISSUE #2622: Do not use `Moment` in protobuffers schemas
It does not makes sense to have an protobuf extension for Moment, as it is a monotonic clock, and shouldn't be used across peers as a timestamp.

This happens in `kademlia/protobuf` in the `tInit` and `tMod` fields of `Ticket` with them being `Moment`. This was the wrong type to use for those fields and it should have been a `Duration` or `DateTime` or `Time`  or some `Timestamp = int64` type instead even though they are documented and encoded as Unix time in seconds. If the value is meant to represent elapsed time or a wait period, use Duration; if it is meant to represent an absolute point in time, use a timestamp/wall-clock type.

richard-ramos:
## Summary

Service discovery tickets used monotonic `Moment` values for fields documented as Unix timestamps. Store `tInit` and `tMod` as `UnixTimestamp` values (an `int64` alias for Unix seconds), use wall-clock time when issuing and validating tickets, and remove the protobuf `Moment` extension. Local cache timers remain monotonic.

The protobuf field numbers, integer encoding, and canonical signing layout are preserved: identical numeric values produce identical bytes.

## Impact on Library Users

Ticket timestamp fields now accept `Opt[UnixTimestamp]` instead of `Opt[Moment]`. Previously issued tickets containing monotonic timestamps will fail the Unix-time window checks.

## References

Closes #2622.


richard-ramos:
cc: @jm-clius @chaitanyaprem

libp2p/protocols/kademlia/protobuf.nim
```diff
@@ -59,9 +59,9 @@ type
   Ticket* {.proto2.} = object
     advertisement* {.fieldNumber: 1.}: Opt[seq[byte]]
       # field 1 - Copy of the original advertisement
-    tInit* {.fieldNumber: 2, ext.}: Opt[Moment]
+    tInit* {.fieldNumber: 2, pint.}: Opt[int64]
```
vladopajic: instead of `int64` it would be better to have some type - type alias like "Timestamp" or any name that will indicate what kind of information is stored in this type.

vladopajic (approved):

gmelodie (approved):
As long as it works... I still don't understand why not use `Moment` here though, is it just to make the code a bit shorter?

jm-clius (approved):
Right, this makes more sense. I think we got away with it, as the Registrar (that created the `Moment`s) is also the only consumer of those `Moment`s. However, any restart at the Registrar would have made all previously issued tickets illegible. It makes more sense to use wall clock timestamps here.

jm-clius:
>  I still don't understand why not use Moment here though, is it just to make the code a bit shorter?

A `Moment` is only meaningful within a single process to measure relative durations. You can't compare the `Moment` generated by one process to another `Moment` in another process - it has no objective meaning.

richard-ramos:
Yeah, this PR is not so much about code length, but about having a portable timestamp.

PR #3060, opened 2026-09-08
chore(logs): use shortlog for large fields

vladopajic:
- using shotlog for some large fields 
- added lint to ensure shortlog is used (on some fields)

### DIFF
--- a/libp2p/dialer.nim
+++ b/libp2p/dialer.nim
@@ -384,7 +384,7 @@ proc dialInOrder(
 
   for rawAddress in addrs:
     if deadline.timeLeft().isZero():
-      debug "Peer dial timed out", peerId, addresses = addrs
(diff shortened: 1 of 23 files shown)


gmelodie (approved):

richard-ramos (commented):
For seq[Multiaddress], I believe you could avoid doing `.shortLog` for all sequences of addresses, by adding something like this in multiaddress.nim

```
import pathToShortLogUtils.nim
chronicles.formatIt(seq[MultiAddress]):
  shortLog(it)
```

richard-ramos (approved):

vladopajic:
> For seq[Multiaddress], I believe you could avoid doing `.shortLog` 

aha, i will investigate, but if this is the case then it will have do be done for all fields. i'll do this in next pr (if it is improvement)

PR #3063, opened 2026-09-08
feat(service-discovery): measure provider discovery latency

gmelodie:
Nothing in service discovery prefetches provider advertisements: the per-service routing table holds Kademlia peers, and a provider of S reaches a node only through `lookup(S)`, or passively when that provider registers with the node. This PR times the second event, from `registerInterest(S)` or the first `lookup(S)` until the node first sees each provider.

Every first sight produces one row: provider, rank, elapsed time, and source (`FromLookup` or `FromRegistration`). Read them with `disco.discoveries(service)` or `disco.tracker.toCsv()`; each row is also logged as `info "Provider found"`, so the simulation collects the raw data through Loki. `cd_provider_discovery_seconds` and `cd_first_provider_discovery_seconds` cover the aggregate view. The tracker ignores the local node and caps at 1024 providers per service.


libp2p/utils/collections.nim
```diff
@@ -3,7 +3,15 @@
 
 {.push raises: [].}
 
-import std/sets
+import std/[sets, tables]
+import results
+
+func getOpt*[K, V](t: Table[K, V], key: K): Opt[V] =
```
vladopajic: hmm not sure what is more suitable, but `nim-libp2p/libp2p/utils/opt.nim` file seems a bit better

libp2p/utils/collections.nim
```diff
@@ -3,7 +3,15 @@
...
+  ## `Opt.some` of the value at `key`, `Opt.none` when `t` has no such key.
+  if t.hasKey(key):
+    Opt.some(t.getOrDefault(key))
```
vladopajic: since it is checked with `hasKey` use `get`

libp2p/protocols/service_discovery/discovery_tracker.nim
```diff
@@ -0,0 +1,120 @@
...
+      source: source,
+    )
+    interest[].found.add(discovery)
+
+    let elapsedSeconds = elapsed.nanoseconds.float64 / 1_000_000_000.0
+    cd_provider_discovery_seconds.observe(elapsedSeconds)
+    if discovery.rank == 1:
+      cd_first_provider_discovery_seconds.observe(elapsedSeconds)
+
+    info "Provider found",
```
vladopajic: why `info` ?

vladopajic (approved):

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -251,6 +255,8 @@ proc acceptAdvertisement*(
   disco.registrar.ads.put(serviceId, advertiser, ad, advertiserIps, now)
   disco.registrar.updateRegistrarMetrics()
 
+  disco.tracker.recordProvider(serviceId, ad.data.peerId, FromRegistration)
```
richard-ramos: Shouldnt this go before line 325? 

Cause otherwise, if you get a Wait response, the discovery time will be longer than necessary (waiting until retry)

libp2p/utils/collections.nim
```diff
@@ -3,7 +3,15 @@
...
+  else:
+    Opt.none(V)
```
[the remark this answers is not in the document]
vladopajic: `getOrDefault` is unnecessary, just get - access by key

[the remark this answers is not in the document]
vladopajic: better not use `withValue` here since it doesn't really improve if raw code is used

richard-ramos (approved):

libp2p/protocols/service_discovery/discovery_tracker.nim
```diff
@@ -0,0 +1,119 @@
...
+    tracker: DiscoveryTracker,
+    serviceId: ServiceId,
+    provider: PeerId,
+    source: DiscoverySource,
+) {.raises: [].} =
+  ## Records the first sight of `provider`, timed from the start of the interest.
+  if provider == tracker.selfId:
+    return
+
+  tracker.interests.withValue(serviceId, interest):
```
vladopajic: this `withValue` creates indentation for almost entire code in this proc; would be better to invers with early return when value is not found in `interests`

libp2p/protocols/service_discovery/discovery_tracker.nim
```diff
@@ -0,0 +1,119 @@
...
+    let elapsed = Moment.now() - interest[].startedAt
+    let discovery = ProviderDiscovery(
+      provider: provider,
+      rank: interest[].found.len + 1,
+      elapsed: elapsed,
+      source: source,
+    )
+    interest[].found.add(discovery)
+
+    let elapsedSeconds = elapsed.nanoseconds.float64 / 1_000_000_000.0
```
vladopajic: would be better to avoid doing conversion in place and insted use some utility

```
let elapsedSeconds = elapsed.secondsFloat64
```

^or some other name

vladopajic (approved):

libp2p/protocols/service_discovery/discovery_tracker.nim
vladopajic: thinking more about it why is it necessary to have precise second? just round it or use whatever `.seconds` gives

vladopajic: or use milliseconds if that precision is needed

PR #3068, opened 2026-09-08
fix(connmanager): restore readiness waits and tag decay on restart

richard-ramos:
## Summary

After a switch is stopped and started again, its connection manager remains marked closed and retained peer tags stop decaying. Restore readiness waits and restart tag decay when the switch starts after shutdown.

Add regression tests for connection-manager restart and for accepting identified connections after a switch restart.

### DIFF
--- a/libp2p/connmanager.nim
+++ b/libp2p/connmanager.nim
@@ -104,7 +104,7 @@ type
   .}
 
   ConnManager* = ref object of RootObj
-    closed: bool
+    started: bool
     muxerStore: MuxerStore
     maxConnsPerPeer: int
     maxConnectionsIn: int
@@ -171,7 +171,7 @@ proc new*(
     watermark: Opt[WatermarkPolicy] = Opt.none(WatermarkPolicy),
     scoring: PeerScoring = PeerScoring(),
 ): ConnManager =
-  ## Creates a `ConnManager`.
+  ## Creates a stopped `ConnManager`. Call `start` before using it.
   ##
   ## `maxConnsPerPeer` accepts values ≤0 to mean "use the default value".
   ##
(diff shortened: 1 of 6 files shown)


libp2p/connmanager.nim
```diff
@@ -699,6 +699,14 @@ proc runDecayLoop(c: ConnManager) {.async: (raises: [CancelledError]).} =
     c.applyDecay()
   c.decayLoopFut = nil
 
+proc start*(c: ConnManager) =
+  ## Resume readiness waits and tag decay after a completed shutdown.
+  if not c.closed:
```
vladopajic: some components have `warn` log when starting already started component

libp2p/connmanager.nim
```diff
@@ -699,6 +699,14 @@ proc runDecayLoop(c: ConnManager) {.async: (raises: [CancelledError]).} =
...
+    return
+  c.closed = false
+  if c.decayingTags.len > 0:
+    c.decayLoopFut = c.runDecayLoop()
```
vladopajic: just run it, leave the out check, there check inside `runDecayLoop`

vladopajic (approved):
lgtm, but it is different behavior then other components.

it might make sense to standardize it similar to others:
- `start` is not obligatory; components with start should be started before doing anything on them
- `close` should be `stop`  (if there is `start`)
 - `c.closed` should be `running / started` (because of above)

tests/libp2p/test_conn_manager.nim
```diff
@@ -20,13 +20,13 @@ method newStream*(
   Connection.new(m.peerId, Direction.Out)
 
 proc newMaxTotal(maxConnections = 10, maxConnsPerPeer = 1): ConnManager =
-  ConnManager.new(
+  result = ConnManager.new(
```
gmelodie: `result`????????

Image

gmelodie (approved):

tests/libp2p/test_conn_manager.nim
richard-ramos: Oh no!

PR #3074, opened 2026-09-08
fix(service_discovery): propagate RPC cancellation and check local ads first

richard-ramos:
## Summary

Allow service-discovery RPC cancellation to interrupt dialing and propagate through reads and writes. Return immediately after a write failure instead of attempting to read a reply.

Check local advertisements before querying routing buckets so lookup can return cached results even when the routing table has no peers. Skip remote queries when local results already satisfy the lookup limit.

Add regression coverage for RPC cancellation and cached advertisements with an empty routing table.


gmelodie (approved):

libp2p/protocols/service_discovery/connection.nim
```diff
@@ -45,18 +46,20 @@ proc send*(
...
       await stream.writeLp(encodedMsg)
-    readRes = catch:
-      await stream.readLp(ServiceDiscoveryMaxMsgSize)
-
-  if writeRes.isErr:
-    return err("connection writing failed: " & writeRes.error.msg)
-  let replyBuf = readRes.valueOr:
-    return err("connection reading failed: " & readRes.error.msg)
+    except CancelledError as e:
+      raise e
```
vladopajic:
```suggestion
```

libp2p/protocols/service_discovery/connection.nim
```diff
@@ -45,18 +46,20 @@ proc send*(
...
+    except LPStreamError as e:
+      return err("connection writing failed: " & e.msg)
+    try:
+      replyBuf = await stream.readLp(ServiceDiscoveryMaxMsgSize)
+    except CancelledError as e:
+      raise e
```
vladopajic:
```suggestion
```

vladopajic (approved):

PR #3075, opened 2026-09-09
chore(logs): add `formatIt` for all types

vladopajic:
this pr add `chronicles.formatIt` for all types that have `shortLog` to avoid calling `shortLog` when logging this type.

improved logging lint that will check if type has `formatIt`.


richard-ramos (approved):

arnetheduck:
keep in mind that `formatIt` doesn't apply to json logs

gmelodie (approved):

PR #3079, opened 2026-09-09
fix(nat): enforce dial-back validation and clean up relay and hole-punching lifecycles

richard-ramos:
## Summary

Reject AutoNAT v2 dial-backs when the anti-amplification challenge is malformed or times out. Previously, completing the challenge with a false result could be mistaken for successful validation. Close newly created dial-back connections when stream creation or negotiation fails, and simplify response conversion.

Wait for autorelay reservation and backoff tasks to stop, clear retained reservations, and withdraw advertised addresses when relays are lost. Register hole-punching handlers on start and remove them on stop so restarting restores their behavior.

Include regression coverage for malformed challenges, failed dial-back negotiation, reservation shutdown, stale relay addresses, and hole-punching service restart.


libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -198,17 +191,24 @@ proc forceNewConnection(
...
+    except CancelledError as exc:
+      await mux.close()
+      raise exc
+    except LPError as exc:
+      await mux.close()
+      raise exc
   except CancelledError as exc:
     raise exc
-  except CatchableError:
+  except LPError:
```
gmelodie, line 207: nice 💅

gmelodie (approved):

libp2p/protocols/connectivity/autonatv2/server.nim
vladopajic: should avoid `try` inside `try`

libp2p/protocols/connectivity/autonatv2/server.nim
```diff
@@ -255,12 +255,12 @@ proc handleDialRequest(
...
-      self.config.amplificationAttackTimeout
-    ):
-      debug "Amplification attack prevention timeout",
-        timeout = self.config.amplificationAttackTimeout, peer = stream.peerId
+    try:
+      await self.amplificationAttackPrevention(stream, addrIdx).wait(
+        self.config.amplificationAttackTimeout
+      )
+    except AutonatV2Error, AsyncTimeoutError:
+      debug "Amplification attack prevention failed", peer = stream.peerId
```
vladopajic: this needs separate `except` blocks
- if it is timeout - then it is requires user action to configure better timeout - needs `warn` log 
- if is any other error - okay to `debug` log

vladopajic (approved):

libp2p/protocols/connectivity/autonatv2/server.nim
vladopajic: log message is to generic - how would user know what has happend error or timeout?

vladopajic: in case of error (not timeout), it would be helpful to add `err` filed carrying errorr message

PR #3080, opened 2026-09-09
fix(service-discovery): stop publishing and dialing undialable addresses

gmelodie:
A node floods its logs with `Dialing failed` on `/logos/service-discovery/1.0.0` against addresses nobody can reach, `/ip4/0.0.0.0/tcp/60000` and `/ip4/127.0.0.1/tcp/0`, and repeats it every discovery cycle. `ServiceDiscovery.record()` published `switch.peerInfo.addrs`, which holds the raw bind address until `switch.start()` resolves it, and nothing rejected such an address on the way in either, since `KadDHTConfig.addressPolicy` defaults to accepting everything.

New `isDialableMA` in `wire.nim` rejects a wildcard host and an unresolved port `0`; `peeraddrpolicy.dialableAddrs` pairs it with the configured policy and replaces `filterAddrs` wherever an address is stored to dial later. `record()` now fails rather than publish nothing dialable, an address change republishes at once instead of waiting for `bucketRefreshTime`, and `send` backs off a peer that will not answer, dropping it from the service tables after `maxDialFailures`. The policy itself stays the caller's choice, so a private testnet keeps its LAN peers.

### DIFF
--- a/libp2p/protocols/service_discovery.nim
+++ b/libp2p/protocols/service_discovery.nim
@@ -41,26 +41,30 @@ proc refreshSelfSignedPeerRecord(
   (await disco.putValue(key, Value.fromBytes(encodedSR))).isOkOr:
     debug "Failed to put signed peer record", err = error
 
-proc republishInFlight(disco: ServiceDiscovery): bool =
-  not disco.selfRecordRepublish.isNil() and not disco.selfRecordRepublish.finished()
-
 proc republishOnAddressChange(disco: ServiceDiscovery): PeerInfoObserver =
   ## Without this, a moved address stays stale in the DHT for a `bucketRefreshTime`.
   proc(p: PeerInfo) {.gcsafe, raises: [].} =
-    if disco.stopping or disco.republishInFlight():
-      return
+    disco.addressChanged.fire()
 
-    disco.selfRecordRepublish = disco.refreshSelfSignedPeerRecord()
-
-proc maintainSelfSignedPeerRecord(
+proc maintainSelfPublications(
     disco: ServiceDiscovery
 ) {.async: (raises: [CancelledError]).} =
-  heartbeat "refresh self signed peer record", disco.config.bucketRefreshTime:
-    if not await disco.refreshSelfSignedPeerRecord().withTimeout(
-      disco.config.bucketRefreshTime
-    ):
-      warn "Signed peer record refresh timed out",
-        timeout = disco.config.bucketRefreshTime
+  ## One publisher for both triggers: an address change never races the heartbeat.
+  var addressMoved = false
+  while true:
+    disco.addressChanged.clear()
+
+    let refresh = disco.config.bucketRefreshTime
+
+    if disco.xprPublishing:
+      if not await disco.refreshSelfSignedPeerRecord().withTimeout(refresh):
+        warn "Signed peer record refresh timed out", timeout = refresh
+
+    if addressMoved:
+      if not await disco.republishProvidedAdverts().withTimeout(refresh):
+        warn "Provided advert republish timed out", timeout = refresh
+
+    addressMoved = await disco.addressChanged.wait().withTimeout(refresh)
 
 proc maintainRegistrar(disco: ServiceDiscovery) {.async: (raises: [CancelledError]).} =
   heartbeat "prune expired advertisements",
@@ -107,6 +111,7 @@ proc new*(
     services: toHashSet(services),
     discoConfig: discoConfig,
     xprPublishing: xprPublishing,
+    addressChanged: newAsyncEvent(),
   )
   disco.initKadBase(
     switch,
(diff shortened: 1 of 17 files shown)


richard-ramos (approved):

tests/libp2p/service_discovery/utils.nim
```diff
@@ -145,6 +139,10 @@ proc setupServiceDiscoveryNode*(
     mount: bool = true,
 ): ServiceDiscovery =
   let switch = createSwitch(privateKey, addresses)
+  # `peerInfo.addrs` only fills in on `switch.start()`, which most tests skip.
+  if switch.peerInfo.addrs.len == 0:
+    switch.peerInfo.addrs = @[makeMultiAddress("127.0.0.1")]
```
vladopajic, line 144: i don't understand this, there is `addresses: seq[MultiAddress] = @[TcpAutoAddress()],` aregument in this function

after switch is create and started `switch.peerInfo.addrs` will be set.

 why was this added here?

tests/libp2p/service_discovery/test_routing_table_manager.nim
```diff
@@ -305,6 +305,20 @@ suite "ServiceRoutingTableManager":
...
+    check disco.rtManager.addService(
+      serviceId, disco.rtable, DefaultReplication, DefaultMaxBuckets, Interest
+    )
+
+    let peerInfo = makePeerInfo(addrs = @[ma("/ip4/0.0.0.0/tcp/60000")])
+    check not disco.insertPeer(serviceId, peerInfo)
+
+    disco.switch.peerStore.allowUndialableAddrs = true
+    check disco.insertPeer(serviceId, peerInfo)
+    check disco.hasPeerInServiceTable(serviceId, peerInfo.peerId)
```
vladopajic, line 320: should have this ins test:

```
    disco.switch.peerStore.allowUndialableAddrs = false
    check not disco.insertPeer(serviceId, peerInfo)
    check not disco.hasPeerInServiceTable(serviceId, peerInfo.peerId)
```

?

tests/libp2p/service_discovery/component/test_disco_dial_backoff.nim
```diff
@@ -0,0 +1,63 @@
...
+    check table.insert(dead)
+
+    let msg = kad_protobuf.Message(msgType: kad_protobuf.MessageType.ping)
+
+    check (await disco.send(dead, msg)).isErr()
+
+    let backedOff = await disco.send(dead, msg)
+    check:
+      backedOff.isErr()
+      strutils.contains(backedOff.error, "backoff")
```
vladopajic, line 49: `strutils.contains(backedOff.error, "backoff")` 

this pattern is not really exactly matching error, but in this case (state of code) it probably is. in general there should be more precise pattern here:

```
check:
  backedOff == makeDialBackOffError(dead)
```

`makeDialBackOffError` should be template (or whatever) used to create this error in library source and exported (for tests) to be used here.

tests/libp2p/service_discovery/component/test_disco_dial_backoff.nim
```diff
@@ -0,0 +1,63 @@
...
+      dialBackoffBase = 1.hours, dialBackoffMax = 1.hours, maxDialFailures = 3
+    )
+    let disco = setupServiceDiscoveryNode(discoConfig = discoConfig)
+    startAndDeferStop(@[disco])
+
+    let service = makeServiceInfo()
+    check disco.addProvidedService(service).isOk()
+    let table = disco.rtManager.getTable(service.id.hashServiceId()).get()
+
+    let dead = randomPeerId()
```
vladopajic, line 37: `dead` -> `deadPeerId`

libp2p/wire.nim
```diff
@@ -187,3 +187,35 @@ proc isPrivateMA*(ma: MultiAddress): bool =
...
   hostIP.isPrivate() or hostIP.isShared() or hostIP.isLinkLocal()
+
+const
+  AnyAddressV4 = [0'u8, 0, 0, 0]
+  AnyAddressV6 = [0'u8, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
+  AnyAddressV4Mapped = [0'u8, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff, 0, 0, 0, 0]
+    ## ``::ffff:0.0.0.0``, the IPv4-mapped spelling of the wildcard host.
+
+proc portOf(ma: MultiAddress): Opt[Port] =
+  for codec in [multiCodec("tcp"), multiCodec("udp")]:
```
vladopajic, line 198: `multiCodec("tcp"), multiCodec("udp")` should come form some consts, then:

```
for codec in  [tcpMC, udpMC]
```

these are likley reused everywhere

libp2p/protocols/service_discovery.nim
```diff
@@ -41,15 +41,30 @@ proc refreshSelfSignedPeerRecord(
...
-        timeout = disco.config.bucketRefreshTime
+  ## One publisher for both triggers: an address change never races the heartbeat.
+  var addressMoved = false
+  while true:
+    disco.addressChanged.clear()
+
+    let refresh = disco.config.bucketRefreshTime
+
+    if disco.xprPublishing:
+      if not await disco.refreshSelfSignedPeerRecord().withTimeout(refresh):
```
vladopajic, line 60: in many places there are things that are done with this timeout, this could could be improved with template

```
await disco.refreshSelfSignedPeerRecord().withTimeout(disco.config.bucketRefreshTime)
```
->
```
await disco.refreshSelfSignedPeerRecord().withBucketRefreshTimeout()
```

add template `withBucketRefreshTimeout`

libp2p/protocols/service_discovery.nim
```diff
@@ -41,15 +41,30 @@ proc refreshSelfSignedPeerRecord(
...
+        warn "Signed peer record refresh timed out", timeout = refresh
+
+    if addressMoved:
+      if not await disco.republishProvidedAdverts().withTimeout(refresh):
+        warn "Provided advert republish timed out", timeout = refresh
+
+    addressMoved = await disco.addressChanged.wait().withTimeout(refresh)
```
vladopajic, line 67: i think this logic `maintainSelfPublications` has become more confusing since it does two things at once. it would be better to split it.


If `disco.xprPublishing  == false` then:
-  `addressChanged.wait()` should not have timeout braking this loop in this case would not make sense
- also first part of heartbeat should never be called 

---

insted of having `addressChanged` event; do do not scadule `republishProvidedAdverts()` in background on each change? loops in not really needed for this case?

PR #3084, opened 2026-09-09
chore(errors): valueOrRaise and onErrorRaise helpers

gmelodie:
`errors.nim` gains a generic `toException` overload and two templates that replace the `valueOr: raise newException(...)` and `isOkOr: raise newException(...)` blocks. `valueOrRaise(X)` returns the value of a `Result[T, E]` or raises `X` with the error message. `onErrorRaise(X)` raises `X` when a `Result[void, E]` holds an error. Each template accepts only its kind of result, so a wrong use fails at compile time.

The call sites in autonat, autonatv2, dcutr, identify, noise, floodsub, gossipsub and the tcp, ws and quic transports now use the templates. `tests/libp2p/test_errors.nim` covers the ok path, the raised exception and its message, the single evaluation of the argument, and the rejection of the wrong kind of result.

### DIFF
--- a/libp2p/errors.nim
+++ b/libp2p/errors.nim
@@ -23,13 +23,15 @@ func toException*(e: string): ref LPError =
 func toException*[E](e: E, X: typedesc): ref X =
   (ref X)(msg: $e)
 
-template raiseOr*[T, E](r: Result[T, E], X: typedesc): T =
+template valueOrRaise*[T: not void, E](r: Result[T, E], X: typedesc): T =
   ## Unwrap `r`, or raise `X` carrying the error message.
-  let res = r
-  if res.isErr():
-    raise res.error().toException(X)
-  when T isnot void:
-    res.unsafeGet()
+  r.valueOr:
+    raise error.toException(X)
+
+template onErrorRaise*[E](r: Result[void, E], X: typedesc) =
+  ## Raise `X` carrying the error message when `r` is an error.
+  r.isOkOr:
+    raise error.toException(X)
 
 # TODO: could not figure how to make it with a simple template
 # sadly nim needs more love for hygienic templates
--- a/libp2p/errors.nim
+++ b/libp2p/errors.nim
@@ -1,22 +1,51 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 # Copyright (c) Status Research & Development GmbH
 
-# this module will be further extended in PR
-# https://github.com/status-im/nim-libp2p/pull/107/
-
 import chronos
 import chronicles
 import macros
+import results
+
+export results
+
+{.push raises: [].}
 
 type
   # Base exception type for libp2p
   LPError* = object of CatchableError
+
+  LPErrorObj*[K: enum] = object
+    ## Error value of a `Result`: `kind` to branch on, `msg` for logs.
+    kind*: K
+    msg*: string
+
+  LPResult*[T; K: enum] = Result[T, LPErrorObj[K]]
+
+func lpError*[K: enum](kind: K, msg = ""): LPErrorObj[K] =
+  LPErrorObj[K](kind: kind, msg: msg)
+
+func `$`*[K: enum](e: LPErrorObj[K]): string =
+  if e.msg.len == 0:
+    $e.kind
+  else:
+    $e.kind & ": " & e.msg
 
 func toException*(e: cstring): ref LPError =
   (ref LPError)(msg: $e)
 
 func toException*(e: string): ref LPError =
   (ref LPError)(msg: e)
+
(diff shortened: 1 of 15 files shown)


richard-ramos (approved):

vladopajic (commented):
i would like to see usage of this (in next pr or whatever) to be convinced

vladopajic:
other way around is to add utility with usage in same pr. for example:

- add `raiseOr` template 
- updated code usage to use this template
- ^ all in same pr

then repeat for each pattern - code improvement

libp2p/errors.nim
```diff
@@ -18,6 +20,17 @@ func toException*(e: cstring): ref LPError =
...
 
+func toException*[E](e: E, X: typedesc): ref X =
+  (ref X)(msg: $e)
+
+template raiseOr*[T, E](r: Result[T, E], X: typedesc): T =
+  ## Unwrap `r`, or raise `X` carrying the error message.
+  let res = r
+  if res.isErr():
+    raise res.error().toException(X)
+  when T isnot void:
```
vladopajic: `when T isnot void:` -> should never be

libp2p/protocols/connectivity/dcutr/server.nim
```diff
@@ -29,8 +29,7 @@ proc new*(
   ) {.async: (raises: [CancelledError]).} =
     var peerDialableAddrs: seq[MultiAddress]
     try:
-      let connectMsg = DcutrMsg.decode(await stream.readLp(1024)).valueOr:
-        raise newException(DcutrError, error)
+      let connectMsg = DcutrMsg.decode(await stream.readLp(1024)).raiseOr(DcutrError)
```
vladopajic: how about:

```nim
let msg = DcutrMsg.decode(bytes).valueOrRaise(DcutrError)
```

this template should always be called for results with value. `void` results should use different template:

```nim
f.validateOverheadRateLimit().onErrorRaise(InitializationError)
```

so this pr should have 2 templates:
- `valueOrRaise` / `valOrRaise` 
  - i vote for `valueOrRaise`
- `onErrRaise` / `ifErrorRaise` / `ifErrRaise` 

^ we should decided if we should use shorter or full word (err vs error) here

vladopajic (commented):
thanks for expanding this with usage now intention and usage is clear

libp2p/protocols/connectivity/dcutr/server.nim
vladopajic: aha so "orignal" pattern used `.valueOr`  so this patter should specialized pattern of same kind, so `.valueOrRaise` seems really reasonable and also looks natural

vladopajic: @richard-ramos @gmelodie

vladopajic: > `void` results should use different template

not necessarily, results with value can also use that template, but not really sure if that patter will ever be needed:

```nim
let res = DcutrMsg.decode(bytes)
res.onErrorRaise(InitializationError)

nextProc(res.get())

```

^ could be used like that as well, but it probably should not

tests/libp2p/test_errors.nim
```diff
@@ -0,0 +1,69 @@
...
+
+  test "valueOrRaise returns the value":
+    let r = Result[int, string].ok(7)
+    check r.valueOrRaise(DemoError) == 7
+
+  test "valueOrRaise raises the requested exception with the message":
+    let r = Result[int, string].err("bad address")
+    try:
+      discard r.valueOrRaise(DemoError)
+      check false
```
vladopajic: raiseAasert "should not get here" 

there was effort some time ago to remove all check false from code. raiseAssert with message is more descriptive

vladopajic (approved):

libp2p/errors.nim
```diff
@@ -18,6 +20,22 @@ func toException*(e: cstring): ref LPError =
...
+template valueOrRaise*[T: not void, E](r: Result[T, E], X: typedesc): T =
+  ## Unwrap `r`, or raise `X` carrying the error message.
+  let res = r
+  if res.isErr():
+    raise res.error().toException(X)
+  res.unsafeGet()
+
+template onErrorRaise*[E](r: Result[void, E], X: typedesc) =
+  ## Raise `X` carrying the error message when `r` is an error.
+  let res = r
```
vladopajic: since this template creates `res` variable. if temolate is called two times innsame scope, would it work - since theres going to be 2 variables with same name? would it be possible to avoid variable

vladopajic: same for other templates

PR #3085, opened 2026-09-09
chore(kad): add `Value` type

vladopajic:
adds `Value` type alias similar to `Key` type.
adds `chronicles.formatIt` for `Value` type.

`Value` type enables future were `Key` and `Value` types could be made `distinct` to avoid interchangeable use (to avoid issue from comment below).

### DIFF
--- a/libp2p/protocols/kademlia/put.nim
+++ b/libp2p/protocols/kademlia/put.nim
@@ -48,7 +48,7 @@ proc manageExpiredRecords*(kad: KadDHT) {.async: (raises: [CancelledError]).} =
       trace "Expired record removed", key
 
 proc dispatchPutVal*(
-    kad: KadDHT, peer: PeerId, key: Key, value: seq[byte]
+    kad: KadDHT, peer: PeerId, key: Key, value: Value
 ): Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
   let msg = Message(
     msgType: Opt.some(MessageType.putValue),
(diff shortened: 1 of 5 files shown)


libp2p/protocols/kademlia/types.nim
```diff
@@ -369,11 +371,11 @@ proc nowUnixSeconds*(): int64 {.gcsafe, raises: [].} =
   now().utc.toTime().toUnix()
 
 type EntryRecord* = object
-  value*: seq[byte]
+  value*: Value
   time*: Timestamp
 
 proc init*(
-    T: typedesc[EntryRecord], value: Key, time: Opt[Timestamp]
+    T: typedesc[EntryRecord], value: Value, time: Opt[Timestamp]
```
vladopajic, line 381: started this pr because of this _issue_

will do next pr with distinct type to avoid interchangeable types

gmelodie (approved):

libp2p/protocols/kademlia/put.nim
```diff
@@ -121,7 +121,7 @@ proc handlePutValue*(
       reason = "keyMismatch", messageType = "putValue", stream
     return
 
-  let value = record.value.valueOr:
+  let value: Value = record.value.valueOr:
```
richard-ramos, line 124: Interesting! the compiler was not able to determine the type itself?

richard-ramos (approved):

libp2p/protocols/kademlia/put.nim
vladopajic: it was here just to enforce this type from start. otherwise it would be seq[byte].

in following pr this will be unecessery

PR #3086, opened 2026-09-09
chore(kad): make `Key` & `Value` distinc types

vladopajic:
this pr has changed  `Key` and `Value` types form `seq[byte]` type alias to `distinct seq[byte]` types:
```nim
type
  Key* = distinct seq[byte]
  Value* = distinct seq[byte]
```

this ensures stricter type safety in code, and prevents any future issues where for example `Key` values are passed and used as `Value` values, and vice versa.

---

**braking change**

downstream deps will have to explicitly create these types instead of using `seq[byte]` in places of `Value`, for example:

previously `Value` was alias for `seq[byte] so following code is okay:
``` nim
    let value: Value = record.value.valueOr:
```
with this change in order to create `Value` form same `seq[byte]`:

```nim
    let valueBytes = record.value.valueOr:
    let value = Value.fromBytes(valueBytes)
```


gmelodie (approved):

libp2p/protocols/kademlia/types.nim
```diff
@@ -87,25 +87,80 @@ const
   MaxProviderKeyLen* = 80 ## Upper bound (bytes) on an ADD_PROVIDER key
 
 type
-  Key* = seq[byte]
-  Value* = seq[byte]
+  Key* = distinct seq[byte]
+    ## A Kademlia routing key. Construct with ``Key.init`` or ``Key.fromBytes``.
+  Value* = distinct seq[byte]
+    ## A Kademlia record value. Construct with ``Value.init`` or ``Value.fromBytes``.
```
richard-ramos, line 93: in service_discovery.nim when calling `putValue` it still passes a seq[byte] and in discoverer.nim There's a Opt[seq[byte]]] being assigned instead of a Opt[Key] in the Message.key. This might be why mobile is failing to compile

libp2p/protocols/kademlia/get.nim
```diff
@@ -141,19 +144,31 @@ method handleGetValue*(
...
+
+  let response = Message(
     msgType: Opt.some(MessageType.getValue),
-    key: Opt.some(key),
-    closerPeers: kad.findClosestPeers(key, stream.peerId),
-  )
-  entryRecordOpt.withValue(entryRecord):
-    response.record = Opt.some(
+    key: Opt.some(key.toBytes()),
+    record: Opt.some(
```
richard-ramos, line 165: Line 147 (the `closerPeers`) was deleted by mistake, i think

richard-ramos (changes requested):

libp2p/protocols/kademlia/get.nim
vladopajic: line 147 is now 151; all is good (also tests `test_get` are passing)

libp2p/protocols/kademlia/types.nim
vladopajic: yes, there are few places that needs to update type conversion. i expected it to be fast, but it is actually slow since i have to 1) run compile 2) address compile issues, and 1) is slow... will get there soon

libp2p/protocols/kademlia/get.nim
richard-ramos: Hm but this is not correct, because the `response` is being built from scratch in line 161, and the closerPeers from 151 is not being used at all in lines 161:169

richard-ramos: (i.e.,  `closerPeers` must be returned always https://github.com/libp2p/specs/blob/master/kad-dht/README.md#value-retrieval `with or without value`)

vladopajic: ah, you are righ! i now see the code and issue

vladopajic: fixed

richard-ramos (approved):

PR #3087, opened 2026-09-09
fix(service-disco): admit registrar senders through the shared Kademlia path

gmelodie:
Both registrar handlers inserted the stream peer into the main Kademlia table through the low-level primitive, skipping the address policy, diversity and liveness checks that `admitPeers` applies. `admitSender` now routes them through the same path, after the request validates.

Service-table admission keeps `insertPeer`, which stays synchronous and already applies policy and diversity; `test_registrar_closer_peers` pins that. Four new tests in `test_registrar` cover both handlers, a rejected request and the closer-peer list.

### DIFF
--- a/tests/libp2p/service_discovery/test_registrar.nim
+++ b/tests/libp2p/service_discovery/test_registrar.nim
@@ -1602,7 +1602,7 @@ suite "Service Discovery Registrar - sender admission":
 
     let inMsg = kadprotobuf.Message(
       msgType: kadprotobuf.MessageType.register,
-      key: Opt.some(serviceId.toBytes()),
+      key: serviceId,
       register: Opt.some(
         kadprotobuf.RegisterMessage(
           advertisement: ad.encode().get(),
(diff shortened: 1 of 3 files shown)


richard-ramos (approved):

vladopajic (approved):

tests/libp2p/service_discovery/test_registrar.nim
```diff
@@ -1588,3 +1588,55 @@ suite "Service Discovery Registrar - connection IPs":
...
+    )
+    let serviceName = "service"
+    let serviceId = serviceName.hashServiceId()
+    let ad = makeAdvertisement(serviceName)
+    let senderId = ad.data.peerId
+    disco.switch.peerStore[AddressBook][senderId] = @[makeMultiAddress("198.51.100.7")]
+
+    let inMsg = kadprotobuf.Message(
+      msgType: kadprotobuf.MessageType.register,
+      key: Opt.some(serviceId.toBytes()),
```
vladopajic: heads-up: since #3099 is going to get merged before this pr, folowing code:

```
key: Opt.some(serviceId.toBytes()),
```
should be
```
key: serviceId,
```

after that pr is merged

PR #3088, opened 2026-09-09
fix(autonat): classify addresses per family

### LINKED ISSUE #2710: ipaddr: isPrivate classifies every IPv6 address as public, so autonat canDial dials private IPv6 targets

gmelodie:
AutoNAT v2 classified addresses with a string prefix check that knew IPv4 only, so an IPv6 target read as public. `isPrivate`/`isPublic` are gone; `isGlobalIP` and the new `isGlobalMA` are the family-aware replacements, and the dial-back gate reads the wire address.

The peerstore address policy now also covers addresses from a name lookup and DCUtR candidates. The default policy accepts everything, so behaviour is unchanged unless a caller sets one. Closes #2710.


libp2p/dialer.nim
```diff
@@ -302,18 +302,27 @@ proc awaitLookup(
...
   for (address, addrPeerId) in await self.tryExpandDnsAddr(
     candidate.peerId, candidate.address, deadline
   ):
     expanded.add(
       DialCandidate(
-        address: address, hostname: address.getHostname(), peerId: addrPeerId
+        address: address,
+        hostname: address.getHostname(),
+        peerId: addrPeerId,
+        fromName: fromName,
```
vladopajic, line 325:
```suggestion
        fromName: candidate.fromNameLookup(),
```

unnecessary `fromName` variable

vladopajic (approved):

libp2p/dialer.nim
vladopajic: @gmelodie this is not correct suggestion, please revert, sorry

richard-ramos (approved):

PR #3090, opened 2026-09-09
fix(muxers): preserve yamux framing when discarding stream data and remove redundant guard

richard-ramos:
## Summary

Consume Yamux data payloads for unknown streams and rejected stream-opening frames before reading the next header. Previously, those payload bytes could be interpreted as subsequent frame headers.

Reject excess streams before constructing channels, account for the rejected opening payload against the retained receive window, and drain discarded data through an 8 KiB buffer instead of allocating a frame-sized sequence. Remove the redundant mplex size check already enforced by its message decoder.

Add regression coverage for unknown-stream framing, rejected SYN payloads, multi-chunk draining, and receive-window enforcement.

## Risk Assessment

The discard buffer bounds temporary payload memory. Draining still waits for the declared payload length; completely unknown streams have no retained receive-window check, and the helper adds no timeout of its own.


libp2p/muxers/mplex/mplex.nim
```diff
@@ -184,11 +184,6 @@ method handle*(m: Mplex) {.async: (raises: []).} =
           # All the errors are handled inside `handleStream()` procedure.
           channel.handlerFut = m.handleStream(channel)
       of MessageType.MsgIn, MessageType.MsgOut:
-        if data.len > MaxMsgSize:
```
richard-ramos, line 187: This check is redundant: the receive loop obtains data through `readMsg()`, which already calls `readLp(MaxMsgSize)` and rejects oversized payloads before returning. Removing this branch preserves the existing message-size limit; it simply avoids checking the same bound again after decoding. The separate buffered-byte limits below remain in place.

libp2p/muxers/yamux/yamux.nim
```diff
@@ -606,6 +606,18 @@ proc handleStream(m: Yamux, channel: YamuxChannel) {.async: (raises: []).} =
   trace "Yamux stream handler completed", channel = $channel
   doAssert(channel.isClosed, "connection not closed by handler!")
 
+proc discardData(
+    conn: RawConn, length: uint32
+) {.async: (raises: [CancelledError, LPStreamError]).} =
+  # Unknown or reset streams still carry framed payloads; never interpret those
+  # bytes as headers or allocate an untrusted frame-sized discard buffer.
+  var buffer: array[8192, byte]
```
vladopajic: `8192` where does this value comes from?

vladopajic (approved):

libp2p/muxers/yamux/yamux.nim
richard-ramos: Nowhere. I used 8kb as the size of the buffer to discard payloads. But this is not a value from a spec or anything.

vladopajic:
```
# empirically choose buffer size used here to discard payloads
const bufferSize = 8192
var buffer: array[bufferSize, byte]
```

gmelodie (approved):

PR #3100, opened 2026-09-10
fix(pubsub): stop peer tasks before connection teardown

### LINKED ISSUE #1926: pubsubpeer: keeps estabiishing connection even when node is shutting down
`pubsubpeer.connectImpl` keeps running and attempting to reestablish connection even when node is shutting down.

expected behavior:
all logic related to pubsub protocol should be stoppend when switch is stopped.

how to reproduce:

- run `Flood publish to all peers with score above threshold, regardless of subscription` test it `trace` debug lvl
- observer logs at the end of `stdout`:
  -  `Could not establish send connection`
  - `connection not found`
  - `Error dialing`
  - `etc`...

note: since nodes are gracefully stopped there should be no logs of this kind

richard-ramos:
## Summary

Stop and await pubsub peer tasks before closing established switch connections, preventing closed send streams from triggering reconnect attempts during shutdown. Apply peer cleanup to both GossipSub and FloodSub, reject late peer events and incoming streams after stop, and treat connector cancellation as normal shutdown.

Use one awaitable peer teardown operation and retain cleanup futures even after peers leave the peer table. Add regression coverage for shutdown ordering with in-flight upgrades and slow protocol teardown, reconnect prevention, unfinished cleanup of removed peers, and restart. Update the GRAFT test to pause heartbeats directly instead of stopping the protocol.

## Impact on Library Users

Pubsub stop now tears down peers and waits for their tasks. The switch drains accepts, upgrades, and services before stopping mounted protocols, while established connections remain available for cleanup. Pubsub refuses new dials as soon as switch shutdown begins.

`PubSubPeer.stopTasks()` now returns a future. Callers must await it or explicitly retain/discard it; await it when teardown must finish before proceeding.

## References

Fixes #1926.

### DIFF
--- a/libp2p/protocols/pubsub/gossipsub.nim
+++ b/libp2p/protocols/pubsub/gossipsub.nim
@@ -397,8 +397,6 @@ method unsubscribePeer*(g: GossipSub, peer: PeerId) =
     for topic, info in stats[].topicInfos.mpairs:
       info.firstMessageDeliveries = 0
 
-  pubSubPeer.stopTasks()
-
   g.extensionsState.removePeer(peer)
 
   procCall FloodSub(g).unsubscribePeer(peer)
@@ -1165,13 +1163,12 @@ proc createExtensionsState(g: GossipSub): ExtensionsState =
     g.parameters.preambleExtensionConfig,
   )
 
-method start*(
-    g: GossipSub
-): Future[void] {.async: (raises: [CancelledError], raw: true).} =
+method start*(g: GossipSub): Future[void] {.async: (raises: [CancelledError]).} =
   if g.started:
     warn "Starting gossipsub twice"
-    return newFutureCompleted[void]()
+    return
 
+  await procCall PubSub(g).start()
   info "gossipsub start"
 
   g.heartbeatFut = g.heartbeat()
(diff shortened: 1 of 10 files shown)


gmelodie (approved):

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -1180,23 +1180,21 @@ method start*(
...
   reportBackgroundFailure(g.directPeersLoop, "gossipsub direct peer maintenance")
-  g.started = true
-  newFutureCompleted[void]()
 
-method stop*(g: GossipSub): Future[void] {.async: (raises: [], raw: true).} =
+method stop*(g: GossipSub) {.async: (raises: []).} =
   info "gossipsub stop"
 
-  if not g.started:
-    warn "Stopping gossipsub without starting it"
```
vladopajic, line 1190: it would be nice to keep this warning and early reutn here

libp2p/protocols/pubsub/gossipsub.nim
```diff
@@ -1180,23 +1180,21 @@ method start*(
...
+  for fut in [g.directPeersLoop, g.scoringHeartbeatFut, g.heartbeatFut]:
+    if not fut.isNil:
+      pending.add(fut)
+  await noCancel chronos.cancelAndWait(pending)
   g.pendingTasks = @[]
-  newFutureCompleted[void]()
+  g.directPeersLoop = nil
+  g.scoringHeartbeatFut = nil
+  g.heartbeatFut = nil
+  await peersStopped
```
vladopajic:
```
var loops = [move g.directPeersLoop, move g.scoringHeartbeatFut, move g.heartbeatFut]
await noCancel loops.cancelAndWait()

var pending = move g.pendingTasks
await noCancel pending.cancelAndWait()
```

something like this would be cleaner

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -507,6 +518,18 @@ template handleSelfPublishing*(p: PubSub, topic: string, data: seq[byte]) =
   if p.triggerSelf:
     await handleData(p, topic, data)
 
+method start*(p: PubSub) {.async: (raises: [CancelledError]).} =
+  p.stopping = false
+  await procCall LPProtocol(p).start()
+
+method stop*(p: PubSub) {.async: (raises: []).} =
+  p.stopping = true
+  p.started = false
```
vladopajic, line 527: why are both flags needed seems like both are changed at the same time

vladopajic (approved):

libp2p/protocols/pubsub/pubsub.nim
richard-ramos: Currently we allow peer activity before `start()` (i.e. the `init()` registering the peer event handler, and the `Joined` event calling `subscribePeer`), but this gotta stop once `stop()` is called. I use two flags to differentiate whether we havent started yet, or if we have started the stopping.
 
Before the first `start()`, both are `false`, after `stop()`, `started` is `false`, and `stopping` is `true`. I could see this as being ugly tho. Will try on a subsequent PR to avoid this to only allow peer activity before start() and stop() and that way we could use a single `started` flag.

PR #3103, opened 2026-09-11
fix(service-discovery): pick closer peers by distance to the service

chaitanyaprem:
## Summary

When a node answers `REGISTER` or `GET_ADS` for a service it has no table for, `getCloserPeers` falls back to the main Kad table. That table is bucketed by distance to the node itself, so the returned `closerPeers` are near the responder, not the service.

A client discoverer whose service table was seeded only with far peers then never gets closer. Every node that doesn't hold the ad replies with more peers from the same far bucket, and lookups return nothing until one happens to reach a holder.

This PR buckets the main-table peers by distance to the service ID and samples that view the same way the service-table path does. This is the spec's `GETPEERS`, which initializes `RegT(service_id_hash)` from `KadDHT(peerID)`. It builds the view without `insert`, so it doesn't touch the shared registry or the Kad metrics.

## Affected Areas

- [x] Peer Management / Discovery: `closerPeers` selection in service discovery

No API change. Downstream builds not run.

## Impact on Library Users

Behaviour change only. A node with no table for a service now returns peers near that service in `closerPeers`. A request whose key is not 32 bytes gets no `closerPeers`; before, it got peers near the responder.

## Risk Assessment

Each such request re-buckets the main table, costing one hash per main-table peer. The path used when a service table exists is unchanged.

## References

- Spec: [LOGOS-SERVICE-DISCOVERY, Recommended Peer Selection Algorithm](https://github.com/logos-co/logos-lips/blob/master/docs/anoncomms/raw/logos-service-discovery.md#recommended-peer-selection-algorithm) (`GETPEERS`)

## Additional Notes

Tested with a new component test, which fails on master and passes with this change. In a local 222-node run, fresh client discoverers that previously stayed at zero results found the ad by their second lookup.

**Alternatives considered**

- **Closest-N to the service.** Return the `fReturn` main-table peers nearest the service ID. Simpler, but it drops the per-bucket spread the spec asks for, and `findClosest` hashes its target, so it would need a pre-hashed variant.
- **Create the service table on first request.** Call `addService` on an unknown `GET_ADS`/`REGISTER` and keep the table. Its bootstrap walks toward the service, so over time it could hold closer peers than the main table has. But creating it starts that bootstrap lookup, every service table is refreshed each `bucketRefreshTime`, and nothing removes or caps these tables. Any peer could then trigger outbound lookups and permanent refresh work by querying random service IDs.


jm-clius (approved):
LGTM!

libp2p/protocols/service_discovery/registrar.nim
gmelodie, line 310:
```suggestion
let maxPerBucket = disco.discoConfig.kRegister
  let table = disco.rtManager.getTable(serviceId)
  let keys =
    if table.isSome():
      table.get().randomPeersClosestFirst(disco.rng, count, maxPerBucket)
    else:
      # No table for this service: view the main table by distance to the
      # service (the spec's GETPEERS), not by distance to this node.
      disco.rtable.randomPeersClosestFirst(
        serviceId, disco.rng, count, maxPerBucket, disco.discoConfig.bucketsCount
      )

return disco.switch.toPeers(keys)
```

gmelodie, line 321:
```suggestion

if serviceId.len != IdLength:
    trace "Key is not a service id: registration", msg = inMsg
    return

```

gmelodie, line 424:
```suggestion

if serviceId.len != IdLength:
    trace "Key is not a service id: getAdvertisements", msg
    return

```

tests/libp2p/service_discovery/component/test_registrar_closer_peers.nim
```diff
@@ -107,6 +109,44 @@ suite "Service Discovery Component - Registrar Closer Peers":
...
+    let targetIdx = buckets.minIndex()
+    check buckets[targetIdx] < buckets.max()
+
+    let serviceId = Key.init(keys[targetIdx].hashFor(rtable.config.hasher))
+    check registrarNode.rtManager.getTable(serviceId).isNone()
+
+    let response = registrarNode.getAdvertisements(
+      others[0].switch.peerInfo.peerId,
+      Message(msgType: Opt.some(MessageType.getAds), key: Opt.some(serviceId)),
+    )
```
gmelodie, line 144: When `targetIdx` == 0 asker and target are the same.
```suggestion
    let requester = others[(targetIdx + 1) mod others.len]
    let response = registrarNode.getAdvertisements(
      requester.switch.peerInfo.peerId,
      Message(msgType: Opt.some(MessageType.getAds), key: Opt.some(serviceId)),
    )
```

gmelodie (commented):
If we had this
```nim
proc pickClosestFirst(
    buckets: openArray[seq[Key]], rng: Rng, count: int, maxPerBucket: int
): seq[Key] =
  ## Up to `count` keys sampled randomly per bucket, closest bucket (highest
  ## index) first, at most `maxPerBucket` from each.
  var selected: seq[Key] = @[]
  var remaining = count
  for i in countdown(buckets.high, 0):
    if remaining <= 0:
      break
    let take = min(remaining, min(maxPerBucket, buckets[i].len))
    for nodeId in rng.pick(buckets[i], take).valueOr(@[]):
      selected.add(nodeId)
      remaining.dec
  selected
```
Then we could have two impls of `randomPeersClosestFirst`:
```nim
proc randomPeersClosestFirst*(
     rtable: RoutingTable, rng: Rng, count: int, maxPerBucket = high(int)
 ): seq[Key] =
   ## Returns up to `count` peers sampled randomly from the routing table's
   ## buckets, starting from the closest buckets (highest indices) and moving
   ## to farther buckets (lower indices).

   if count <= 0:
     return @[]
  pickClosestFirst(rtable.buckets.mapIt(it.peers), rng, count, maxPerBucket)

proc randomPeersClosestFirst*(
    rtable: RoutingTable,
    target: Key,
    rng: Rng,
    count: int,
    maxPerBucket = high(int),
    maxBuckets = rtable.config.maxBuckets,
): seq[Key] =
  ## Same sampling, but with the table's peers viewed by distance to the
  ## pre-hashed ``target`` (which must be ``IdLength`` bytes) instead of to
  ## ``selfId``. Read-only: nothing is inserted and neither the registry nor
  ## the metrics are touched.
  if count <= 0:
    return @[]

  var view = newSeq[seq[Key]](bucketCount(maxBuckets))
  for bucket in rtable.buckets:
    for nodeId in bucket.peers:
      let lz = xorDistance(target, Key.fromBytes(nodeId.hashFor(rtable.config.hasher)))
        .leadingZeros()
      view[min(lz, view.high)].add(nodeId)
  pickClosestFirst(view, rng, count, maxPerBucket)
```

With this we could get rid of `serviceView`


Also, `RegT` should also use the routing table's hasher (following patch from claude):
```patch
@@ -49,7 +49,10 @@ proc addService*(
   var rtable = RoutingTable.new(
     serviceId,
     config = RoutingTableConfig.new(
-      replication = replication, maxBuckets = bucketsCount, selfIdPreHashed = true
+      replication = replication,
+      hasher = mainRoutingTable.config.hasher,
+      maxBuckets = bucketsCount,
+      selfIdPreHashed = true,
     ),
     localNodeId = Opt.some(mainRoutingTable.localNodeId),
     registry = mainRoutingTable.registry,
```

richard-ramos (approved):

libp2p/protocols/kademlia/routing_table.nim
```diff
@@ -310,6 +310,25 @@ proc findClosest*(rtable: RoutingTable, targetId: Key, count: int): seq[Key] =
 proc findClosestPeerIds*(rtable: RoutingTable, targetId: Key, count: int): seq[PeerId] =
   findClosest(rtable, targetId, count).toPeerIds()
 
+proc pickClosestFirst(
+    buckets: openArray[seq[Key]], rng: Rng, count: int, maxPerBucket: int
+): seq[Key] =
+  ## Up to `count` keys sampled randomly per bucket, closest bucket (highest
+  ## index) first, at most `maxPerBucket` from each.
+  var selected: seq[Key] = @[]
```
vladopajic:
```suggestion
  var selected: seq[Key] = newSeqOfCap[Key](count)
```

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -291,6 +297,24 @@ proc registration*(
     trace "Key not set: registration", msg = inMsg
     return
 
+  if serviceId.len != IdLength:
+    trace "Key is not a service id: registration", msg = inMsg
```
vladopajic: `Key is not a service id` -> `Key does not have service id length`

same bellow

vladopajic (approved):

gmelodie (approved):

PR #3104, opened 2026-09-11
chore(opt): improvements

vladopajic:
- ~`getOpt` had unnecessary 2 lookups when value is in Table, since this pr it will always have 1 lookup~
  - reverted: description in comments
- avoid use of `toOpt()` inside utilities here. it crates unnecessary intermediate value 
- restrict some type parameters where type must not be void

### DIFF
--- a/libp2p/utils/opt.nim
+++ b/libp2p/utils/opt.nim
@@ -8,13 +8,10 @@ export results
 
 func getOpt*[K, V](t: Table[K, V], key: K): Opt[V] =
   ## `Opt.some` of the value at `key`, `Opt.none` when `t` has no such key.
-  if key notin t:
-    return Opt.none(V)
-
   try:
     Opt.some(t[key])
   except KeyError:
-    raiseAssert "key presence checked above"
+    Opt.none(V)
 
 func toOpt*[T](v: Opt[T] | T): Opt[T] =
   when v is T:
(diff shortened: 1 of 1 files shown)


gmelodie (approved):

libp2p/utils/opt.nim
```diff
@@ -8,13 +8,10 @@ export results
...
 func getOpt*[K, V](t: Table[K, V], key: K): Opt[V] =
   ## `Opt.some` of the value at `key`, `Opt.none` when `t` has no such key.
-  if key notin t:
-    return Opt.none(V)
-
   try:
     Opt.some(t[key])
   except KeyError:
-    raiseAssert "key presence checked above"
+    Opt.none(V)
```
richard-ramos: How do we know if this is really an improvement?

There are two scenarios:

With the `key` present:
- Before the PR, lookup for `not in` then `try` that will not raise, and another lookup to actually get the `key`
- With this PR: one lookup + `try` that wil not raise
- Likely faster in new PR

With the `key` absent:
- Before the PR: one lookup + if branch,
- With this PR: one lookup, and raise a KeyError and catch it
- This is almost certainly much slower

richard-ramos: ccing: @tersec for his thoughts

vladopajic: > This is almost certainly much slower

it's not intuitive to me that it is, but i will make some code to benchmark it.
i assume neglectavle difference in that case. goal was mainly to optimize 2 lookups to 1 lookup when there is value.

vladopajic: what a stupid language decision

at 100,000 iterations on Nim 2.2.10 and `-d:release`
```
guarded / hit          12.14 ns/op
try/except / hit        8.60 ns/op
guarded / miss          9.00 ns/op
try/except / miss      79.28 ns/op
```

exceptions make misses roughly 9x slower.

--- 

nuff said; reverted

richard-ramos (approved):

PR #3105, opened 2026-09-11
chore(opt): rename `withValue` to `ifValue`

vladopajic:
```nim
foo.ifValue(v):
  echo $v
else:
  echo "no value"
```

objectively should be better name because of `if - else` statement chain. wdyt?

### DIFF
--- a/cbind/libp2p.nim
+++ b/cbind/libp2p.nim
@@ -379,7 +379,7 @@ proc createLibp2pNode(config: Libp2pConfig): Result[LibP2P, string] =
   if cfg.autonat:
     switchBuilder = switchBuilder.withAutonat()
 
-  cfg.nat.withValue(natCfg):
+  cfg.nat.ifValue(natCfg):
     switchBuilder = switchBuilder.withNAT(natCfg)
 
   if cfg.autonatV2Server:
--- a/interop/gossipsub/src/node.nim
+++ b/interop/gossipsub/src/node.nim
@@ -75,7 +75,7 @@ proc createNode*(
   params.behaviourPenaltyDecay = 0.0
   params.disconnectBadPeers = false
   params.disconnectPeerAboveRateLimit = false
-  partialMessageConfig.withValue(pmConfig):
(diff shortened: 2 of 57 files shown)


gmelodie (approved):

richard-ramos:
It's a breaking change since it's used downstream tho:
- https://github.com/logos-messaging/logos-delivery/blob/49f55fe4aa8d56fb7ecbf7daa56134672d24f0e2/logos_delivery/waku/node/waku_switch.nim#L100
- https://github.com/logos-messaging/logos-delivery/blob/49f55fe4aa8d56fb7ecbf7daa56134672d24f0e2/logos_delivery/waku/common/rate_limit/request_limiter.nim#L141

Otoh. TIL this is part of nim-libp2p. This is a generic functionality and useful enough to be used across projects, which should make nim-results is a better place for this to live than in libp2p.

vladopajic:
> It's a breaking change

internal utilitieas should not be treated as part of library api. 

downstram deps using this these helpers is a hack

richard-ramos:
cc: @Ivansete-status . 
Is this something you could fix in logos-delivery? This function's being renamed so when 2.5 is eventually released, bumping libp2p will no longer be straightforward

richard-ramos (approved):

richard-ramos:
> internal utilitieas should not be treated as part of library api.

sucks that we cannot control the visibility like in Rust or even Go with `internal/` folder

vladopajic:
> nim-results is a better place for this to live than in libp2p.

that's not goal of `nim-results`, it just wants to offer basic functionality and dialect is left to downstream deps

richard-ramos:
> that's not goal of nim-results, it just wants to offer basic functionality and dialect is left to downstream deps

Fair enough, thought one can argue that nim-results goes beyond basic functionality with things like `valueOr`, `isOkOr`, `isErrOr` which are already a results "dialect", with `ifValue` being part of the same type of syntax sugar templates.

That said, I dont have strong feelings about getting this in nim-results or not hence prev approval

PR #3108, opened 2026-09-11
fix(service-discovery): query closer peers within the same lookup

chaitanyaprem:
## Summary

The spec describes the lookup as iterative: each `GET_ADS` response's `closerPeers` are added to `DiscT(service_id_hash)`, and the lookup continues over that growing table. `lookup` instead walks a copy of the service table's buckets taken at the start, and closer peers only reach the table after a background admission probe. So they help the next lookup, not the current one. A discoverer that starts with no registrar for the service returns nothing on its first lookup, even when the first reply names one.

This PR keeps each reply's closer peers in the lookup, bucketed by distance to the service, and walks the live table by index so later buckets include them. A peer is kept only if it isn't this node, has an address to dial (admission records addresses only for peers it would seat), and falls in a bucket the walk hasn't reached yet. Background admission into the table is unchanged. Kademlia's iterative lookup does the same with its shortlist.

## Affected Areas

- [x] Peer Management / Discovery: service discovery `lookup`

No API change. Downstream builds not run.

## Impact on Library Users

Behaviour change only. A lookup can now query registrars it learns about during that lookup, so a discoverer that starts far from the service can find ads on its first lookup instead of the next one.

## Risk Assessment

A lookup may send `GET_ADS` to peers that haven't passed an admission probe. The `GET_ADS` itself shows whether they serve the protocol, and a failed one only uses its slot until the bucket's timeout. A responder can steer later buckets toward peers it names; the spec's LOOKUP allows the same, since it adds `closerPeers` to `DiscT` right away. The walk still covers at most `bucketsCount` buckets with `K_lookup` queries each.

## References

- Spec: [LOGOS-SERVICE-DISCOVERY, Example LOOKUP algorithm](https://github.com/logos-co/logos-lips/blob/master/docs/anoncomms/raw/logos-service-discovery.md#example-lookup-algorithm)
- Stacked on #3103.

## Additional Notes

Tested with a new component test where the discoverer knows only a relay and the relay knows the node holding the ad. It fails without this change and passes with it.

Within a bucket the spec's example queries `K_lookup` peers one at a time and adds closer peers between them. This PR keeps the parallel queries per bucket, so closer peers apply from the next bucket on.

**Alternatives considered**

- **Wait for admission probes.** Walk the live table and, after each bucket, wait for the table's pending admission probes before moving on. That keeps "only probed peers are queried", but adds up to one probe round-trip per bucket, and probes can be dropped (no free admission slot, or backoff), so closer peers can still be missed.


libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -201,18 +207,36 @@ proc lookup*(
   local.ifValue(response):
     disco.processResponse(serviceId, response, found, disco.discoConfig.fLookup)
 
-  let buckets = searchTable.buckets
-  for bucket in buckets:
+  # Admission into the table waits on a background probe, so closer peers from a
+  # reply are also kept here, by bucket, and queried when the walk reaches them.
+  # The walk covers every bucket the table can have, as it allocates them lazily.
```
gmelodie:
```suggestion
```

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -201,18 +207,36 @@ proc lookup*(
...
-
-    found =
+    let peers = disco.peersToQuery(candidates)
+    let bucketAds =
       await disco.collectBucketAds(serviceId, peers, found, disco.discoConfig.fLookup)
+    found = bucketAds.found
+
+    for peer in bucketAds.closerPeers:
+      # Admission records addresses only for peers it would seat, so a peer
+      # without one could not be dialed.
```
gmelodie:
```suggestion
```

vladopajic (approved):

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -201,18 +207,36 @@ proc lookup*(
...
+      if peer.peerId == disco.switch.peerInfo.peerId or
+          disco.switch.peerStore[AddressBook][peer.peerId].len == 0:
+        continue
+      let key = peer.peerId.toKey()
+      let peerIdx = searchTable.bucketIndex(key)
+      if peerIdx > bucketIdx:
+        learned[peerIdx].add(key)
```
gmelodie: maybe extract this
 ```suggestion
    found = bucketAds.found
    disco.recordCloserPeers(searchTable, bucketAds.closerPeers, bucketIdx, learned)
```

into this

```nim
proc recordCloserPeers(
    disco: ServiceDiscovery,
    searchTable: RoutingTable,
    closerPeers: openArray[PeerInfo],
    afterBucket: int,
    learned: var seq[seq[Key]],
) =
  ## File closer peers into buckets the walk has not reached yet. Peers
  ## without a recorded address are skipped since they cannot be dialed.
  for peer in closerPeers:
    if peer.peerId == disco.switch.peerInfo.peerId or
        disco.switch.peerStore[AddressBook][peer.peerId].len == 0:
      continue
    let key = peer.peerId.toKey()
    let peerIdx = searchTable.bucketIndex(key)
    if peerIdx > afterBucket:
      learned[peerIdx].add(key)
```

gmelodie (approved):
nice!

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -201,18 +207,36 @@ proc lookup*(
...
 
-    let peers = disco.peersToQuery(bucket)
-
-    found =
+    let peers = disco.peersToQuery(candidates)
+    let bucketAds =
       await disco.collectBucketAds(serviceId, peers, found, disco.discoConfig.fLookup)
+    found = bucketAds.found
+
+    for peer in bucketAds.closerPeers:
```
jm-clius: Not sure if we have properly implemented this in the registrar, but by spec a registrar should _at most_ return one closer peer per bucket. This is a Sybil protection mechanism - otherwise we can easily receive a response with a very large amount of (fake) closer peers that fills our routing table or hijacks the discovery process. Can we at least add a sanity check here that closerPeers must be <= number of remaining closer buckets?

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -201,18 +207,36 @@ proc lookup*(
   local.ifValue(response):
     disco.processResponse(serviceId, response, found, disco.discoConfig.fLookup)
 
-  let buckets = searchTable.buckets
-  for bucket in buckets:
+  # Admission into the table waits on a background probe, so closer peers from a
+  # reply are also kept here, by bucket, and queried when the walk reaches them.
+  # The walk covers every bucket the table can have, as it allocates them lazily.
+  var learned = newSeq[seq[Key]](bucketCount(searchTable.config.maxBuckets))
```
jm-clius, line 253: See my comment below on Sybil protection: `closerPeers` hints should _at most_ contain a single peer per closer bucket to prevent a malicious Sybil response from dominating the lookup. Fixing when parsing `closerPeers` should be enough, but we might consider bounding the inner `seq[Key]` to `bucketCount` as well?

libp2p/protocols/service_discovery/discoverer.nim
```diff
@@ -201,18 +207,36 @@ proc lookup*(
...
+  for bucketIdx in 0 ..< learned.len:
     if found.len >= disco.discoConfig.fLookup:
       break
 
-    if bucket.peers.len == 0:
+    var candidates = learned[bucketIdx]
+    if bucketIdx < searchTable.buckets.len:
+      candidates.add(searchTable.buckets[bucketIdx].peers)
+    candidates = candidates.deduplicate()
```
jm-clius, line 261: To prevent the impact of Sybil responses, perhaps better to fill the `candidates` from the `searchTable` first and then augment with learned candidates? The routing table has a strong admissions gate, so should in theory have better candidates.

jm-clius (approved):
Thanks! Non-blocking comments below - mostly to prevent impact of possible Sybil attacks/malicious `closerPeers` response.

libp2p/protocols/service_discovery/discoverer.nim
chaitanyaprem: good idea, irrespective of whether it is implemented in registrar or not it is a good idea to add this defensive check so that discoverers can be protected here.

chaitanyaprem: > but by spec a registrar should _at most_ return one closer peer per bucket

looks like it is capped at `kRegister(3)`. 
On thinking about it, maybe it makes sense to leave it at `3` because otherwise an advertiser will not be able to fill slots faster in AdvT which may slow down placing of ads on cold start. 
Correct me if i am missing something here.

libp2p/protocols/service_discovery/discoverer.nim
chaitanyaprem: Sure, i wanted to do this anyways.

libp2p/protocols/service_discovery/discoverer.nim
chaitanyaprem: Being addressed [here](https://github.com/vacp2p/nim-libp2p/pull/3127)

libp2p/protocols/service_discovery/discoverer.nim
chaitanyaprem: looked at this again while doing #3127, and i don't think `learned[i]` needs a bound of its own.

the cap is applied at parse time now, in `atMostOnePerBucket`, per reply, before anything consumes it. on top of that, `recordCloserPeers` files into `learned[bucketIndex(key)]` and `insert` places peers at `buckets[bucketIndex(key)]`, so the candidates for step `j` are exactly the peers whose bucket index is `j`. each peer is a candidate in only one step, which means a responder is queried at most once per lookup. one query per responder, one peer per bucket per query, so it works out to one peer per responder per bucket for the whole walk.

`learned[j]` can still grow to `kLookup * j` over the earlier steps, since bucket `j` collects from every step before it. but each entry there costs a distinct identity that had to win a draw in its own bucket first. and i don't think a bound helps in the buckets that matter: the ones nearest the service key are mostly empty in our own table, so hints are the only candidates we have there. capping how many hints there are does not change that, all `kLookup` queries still go to hints either way.

correct me if i am missing something here.

libp2p/protocols/service_discovery/discoverer.nim
chaitanyaprem: looked into this and the ordering does not actually matter. `peersToQuery` goes through `rng.pick`, which is a partial Fisher-Yates over the whole list (`rng.nim:143-152`), so an element's chance of being picked depends only on the list length and `kLookup`, not on where it sits. filling from `searchTable` first gives the same multiset and the same distribution over picks. only the ratio matters, `L/(L+T)` per draw.

biasing the selection instead of the order would do something, but i think it helps in the wrong place. bucket `0` covers half the keyspace and the last buckets almost none, so our table is dense at the start of the walk and empty by the end. preferring table peers would protect the buckets where the risk is already lowest and do nothing near the service key, which is where the providers are and where hints are all we have.

PR #3110, opened 2026-09-11
feat(dialer): rank dials by relay and transport

gmelodie:
The ranked dial path from #2992 and #3001 races every candidate with no order or limit. `RankedDial` in `libp2p/dialer.nim` now queues the candidates of a peer dial by `DialRank`: direct QUIC, other direct, relay QUIC, other relay. A free slot dials the best queued candidate, with 8 slots at most, and a connection cancels the rest.
Tests check each rank, the dial order, the slot limit, the reuse of a failed attempt's slot, and the cancel of a stalled attempt. Check `run` and `nextEvent` for lost wake-ups, because the loop has no timer and trusts the dial deadline.

### DIFF
--- a/libp2p/dialer.nim
+++ b/libp2p/dialer.nim
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 # Copyright (c) Status Research & Development GmbH
 
-import std/[sequtils, sets, tables]
+import std/[algorithm, sequtils, sets, tables]
 
 import pkg/[chronos, chronicles, metrics, results]
 
@@ -25,7 +25,7 @@ import
   utils/future,
   errors
 
-export dial, dial_backoff, errors, results
+export dial, dial_backoff, dialcandidate, errors, results
 
 logScope:
   topics = "libp2p dialer"
@@ -71,7 +71,8 @@ type
     peerStore: PeerStore
     nameResolver: NameResolver
     ms: MultistreamSelect
-    dialRanking: bool ## overlap the name lookups with the dials
+    dialRanking: bool ## overlap the name lookups with the dials, and rank the dials
+    dialRankingConfig: DialRankingConfig
     dialBackoff: Opt[DialBackoff] ## none unless the switch opted in
     ongoingReleaseOnClose: seq[Future[void].Raising([])]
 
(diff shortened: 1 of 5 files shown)


richard-ramos (changes requested):
This PR adds an unnecessary complexity and latency. The requirement for ranking is priority based:

1. Direct addresses before relay
2. WIthin each category, QUIC before TCP

In this PR (correct me if i am wrong) you are introducing time based 'head starts', which introduce complexity and a behavior that is not really required

So, dialing already is following a producer/consumer pattern. The producer resolves and emit addresses, and the consumer select the highest priority candidate available when a dial slot opens. This can be represented with priority ordering, without needing delays

Interestingly enough, it seems that all relay candidates are equal without prefering quic over tcp.

libp2p/dialer.nim
```diff
@@ -430,83 +432,112 @@ proc dropLosers(attempts: seq[DialAttempt]) {.async: (raises: []).} =
...
+
+proc wake(dial: RankedDial) =
+  if not dial.changed.finished():
+    dial.changed.complete()
+
+func byRank(a, b: DialCandidate): int =
+  cmp(a.dialRank(), b.dialRank())
+
+proc queue(dial: RankedDial, candidates: seq[DialCandidate]) =
+  for candidate in dial.dialable.take(candidates):
```
richard-ramos, line 455:
```suggestion
  for candidate in dial.dialable.take(candidates.sorted(byRank)):
```

libp2p/dialer.nim
```diff
@@ -430,83 +432,112 @@ proc dropLosers(attempts: seq[DialAttempt]) {.async: (raises: []).} =
...
+
+    self.openSlots(dial)
+    if dial.pending.len == 0 and dial.lookups == 0:
+      return nil
+
+    await dial.nextEvent()
 
-proc dialName(
+proc queueResolved(
     self: Dialer,
```
richard-ramos, line 523:
```suggestion
```

libp2p/dialer.nim
```diff
@@ -430,83 +432,112 @@ proc dropLosers(attempts: seq[DialAttempt]) {.async: (raises: []).} =
...
+    dial.lookups.dec()
+    dial.wake()
 
-  await firstConnected(
-    lookups.mapIt(self.dialResolved(it, dialBudget, dir, deadline, forceDial, reach))
+  let expanded = dial.unresolved.take(
+    await dial.dialable.awaitLookup(self.expandCandidate(candidate, dial.deadline))
   )
+  let lookups = expanded.mapIt(self.resolveCandidate(it, dial.deadline))
+  await allOrCancel(lookups.mapIt(self.queueResolved(dial, it)))
```
richard-ramos, line 540:
```suggestion
  await allOrCancel(lookups.mapIt(dial.queueResolved(it)))
```

tests/libp2p/test_dialer.nim
```diff
@@ -454,6 +474,96 @@ suite "Dialer":
...
+    defer:
+      await src.stop()
+
+    let
+      transport = FailingDialTransport.new(Upgrade(), rng(), handlesAny = true)
+      dialer = src.rankedDialer(@[Transport(transport)])
+      relayTcp = relayAddr("/ip4/5.6.7.8/tcp/4001")
+      tcp = ma("/ip4/1.2.3.4/tcp/4001")
+      relayQuic = relayAddr("/ip4/5.6.7.8/udp/4001/quic-v1")
+      quic = ma("/ip4/1.2.3.4/udp/4001/quic-v1")
```
richard-ramos, line 497:
```suggestion
      dialer = src.rankedDialer(@[Transport(transport)])
      relayTcp = relayAddr("/ip4/5.6.7.8/tcp/4001")
      tcp = ma("/ip4/1.2.3.4/tcp/4001")
      ws = ma("/ip4/1.2.3.4/tcp/4002/ws")
      relayQuic = relayAddr("/ip4/5.6.7.8/udp/4001/quic-v1")
      quic = ma("/ip4/1.2.3.4/udp/4001/quic-v1")
```

tests/libp2p/test_dialer.nim
```diff
@@ -454,6 +474,96 @@ suite "Dialer":
...
+    expect DialFailedError:
+      await dialer.connect(
+        PeerId.random(rng()).tryGet(), @[relayTcp, tcp, relayQuic, quic]
```
richard-ramos, line 500:
```suggestion
        PeerId.random(rng()).tryGet(), @[relayTcp, tcp, ws, relayQuic, quic]
```

richard-ramos (approved):

tests/libp2p/test_dialer.nim
vladopajic: `PeerId.random(rng()).tryGet()` should be utility from before; and in ohter places

libp2p/dialer.nim
```diff
@@ -430,83 +432,112 @@ proc dropLosers(attempts: seq[DialAttempt]) {.async: (raises: []).} =
       if not isNil(mux):
         await mux.close()
 
-proc firstConnected(
-    attempts: seq[DialAttempt]
-): Future[Muxer] {.async: (raises: [CancelledError]).} =
-  ## The first attempt that connects. Nil when none of them does.
+type RankedDial = ref object
```
vladopajic, line 435: if `RankedDial` type and implementation could be in individual file, then better be so

libp2p/dialer.nim
```diff
@@ -430,83 +432,112 @@ proc dropLosers(attempts: seq[DialAttempt]) {.async: (raises: []).} =
...
-  var pending = attempts
-  defer:
-    await dropLosers(pending)
+  while dial.queued.len > 0 and dial.pending.len < MaxParallelDials:
+    let candidate = dial.queued[0]
+    dial.queued.delete(0)
+    trace "Ranked dial attempt opened",
+      peerId = candidate.peerId,
+      address = candidate.address,
+      rank = candidate.dialRank()
```
vladopajic, line 469: here `trace` is logging candidate. it is better to create `shortLog` + `chronicles.formatIt` and log that struct. in general that should be practice for any type that is being logged as it will reduce code if there is more then one logging (even if there is just a single place logging this struct it would be better to have those functions).

note that `shortLog` can only log same properties as here, it doesn't have to all all properties of struct.

```nim
trace "Ranked dial attempt opened", candidate
```

libp2p/dialer.nim
```diff
@@ -430,83 +432,112 @@ proc dropLosers(attempts: seq[DialAttempt]) {.async: (raises: []).} =
...
-        it.peerId, it.hostname, it.address, dir, deadline, forceDial, reach
-      )
-    )
-  )
+proc nextEvent(dial: RankedDial) {.async: (raises: [CancelledError]).} =
+  ## Every attempt and lookup ends by the dial deadline, so no timer wakes the loop.
+  try:
+    discard await race(dial.pending.mapIt(FutureBase(it)) & FutureBase(dial.changed))
+  except ValueError as e:
+    raiseAssert "race() over a non-empty seq: " & e.msg
```
vladopajic, line 498: in my personal taste, this `race` would be better without this proc. it is used once and from this code here it is not obvious that  `raiseAssert` will never happen. this code teased me 

comment is about loop so having this code in loop (and comment) still +1 for above.

tests/libp2p/test_dialer.nim
```diff
@@ -44,6 +45,25 @@ proc stallAfterIdentify(sw: Switch) =
 
   sw.replaceIdentifyHandler(hold)
 
+proc relayAddr(relay: string): MultiAddress =
+  ma(relay & "/p2p/" & $PeerId.random(rng()).tryGet() & "/p2p-circuit")
+
+proc rankOf(address: MultiAddress): DialRank =
+  DialCandidate(address: address).dialRank()
+
+proc tcpAddrs(count: int): seq[MultiAddress] =
```
vladopajic, line 54: makeTcpAddrs

tests/libp2p/test_dialer.nim
```diff
@@ -44,6 +45,25 @@ proc stallAfterIdentify(sw: Switch) =
...
+  (0 ..< count).mapIt(ma("/ip4/1.2.3.4/tcp/" & $(4001 + it)))
+
+proc rankedDialer(src: Switch, transports: seq[Transport]): Dialer =
```
vladopajic, line 57: makeRankedDialer

tests/libp2p/test_dialer.nim
```diff
@@ -44,6 +45,25 @@ proc stallAfterIdentify(sw: Switch) =
 
   sw.replaceIdentifyHandler(hold)
 
+proc relayAddr(relay: string): MultiAddress =
+  ma(relay & "/p2p/" & $PeerId.random(rng()).tryGet() & "/p2p-circuit")
```
vladopajic, line 49: `PeerId.random(rng()).tryGet` should be utility that was created earlier
(and in ohter places..)

vladopajic (approved):

PR #3111, opened 2026-09-11
fix(quic): avoid caching partially constructed dial endpoints

richard-ramos:
## Summary

An exception while constructing a dial-only QUIC endpoint can partially write the cached `Opt`, leaving `some(nil)` and causing a subsequent dial to segfault. Construct the endpoint before assigning it to the IPv4 or IPv6 cache so failed initialization leaves the cache empty and retryable.

Add coverage that injects a certificate-generation failure, retries successfully against a local peer, and shuts down cleanly for both address families.

## References

- https://github.com/nim-lang/Nim/issues/26208 (missing warnings for partial writes during construction)


gmelodie (approved):

vladopajic (approved):
:O

PR #3112, opened 2026-09-11
chore(coverage): filter gcov noise from the report

gmelodie:
The coverage report counts lines with no code as untested: blank lines, comments, proc headers, closing brackets, and lines past the end of a file. About half of the misses on master are such lines. This PR moves the coverage job to Ubuntu 24.04, whose lcov can filter them out, with gcc 14 like the other 24.04 jobs. It also removes the job's build cache, which restored coverage data from old runs.
With these filters, master's Codecov data goes from 83.7% to 89.4%, and every real missed line stays. Please check the coverage job log for lcov errors.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -42,6 +42,8 @@ jobs:
         run: |
           sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 14
           sudo update-alternatives --set gcc /usr/bin/gcc-14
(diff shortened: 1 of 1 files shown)


richard-ramos:
Will review once :green_circle:

gmelodie:
@richard-ramos ready!

richard-ramos (approved):
a-mazing!

vladopajic (approved):
awesome! i did notice that coverage metrics was strange and not correct

vladopajic:
image


diff is still not correct here ^ ?

PR #3113, opened 2026-09-11
test(dialer): cover ranked dial timing, limits and cancellation

gmelodie:
PR #3110 checks the ranking rules as a pure `DialPlan` and with two real switches, but no test checks the order, delay, limit or cancellation of the attempts that `RankedDial` starts. This PR adds `ScriptedDialTransport` to `tests/stubs/transportstub.nim`, which fails or stalls each dial as the test lists and records each dial, its time and each cancellation, plus seven tests in `tests/libp2p/test_dialer.nim`.
Tests only, stacked on #3110; `nim check tests/libp2p/test_dialer.nim` passes and CI runs the tests. Check that each delay test measures from a moment taken before the dial starts, so a slow scheduler only makes the measured delay longer.

### DIFF
--- a/tests/libp2p/test_dialer.nim
+++ b/tests/libp2p/test_dialer.nim
@@ -9,6 +9,7 @@ import
     builders,
     muxers/muxer,
     nameresolving/mockresolver,
+    nameresolving/nameresolver,
     peerstore,
     stream/bridgestream,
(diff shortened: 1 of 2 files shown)


gmelodie:
This will be one smaller pr only

PR #3117, opened 2026-09-11
fix(service-disco): admit the registering peer instead of the advertised one

jm-clius:
Noticed while investigating the module integration in Delivery. The `peerId` and `addresses` encoded in the advertisement does not necessarily match those of the advertiser (i.e. the actual module participating in the KadDHT). In fact, when using the Service Discovery module as a proxy (as Delivery does), it never will. Fix ensures its the advertiser's information that is admitted to the routing table and not the unrelated info in the advertisement.

cc @NagyZoltanPeter

### DIFF
--- a/libp2p/protocols/service_discovery/registrar.nim
+++ b/libp2p/protocols/service_discovery/registrar.nim
@@ -251,11 +251,16 @@ proc acceptAdvertisement*(
     serviceId, disco.rtable, disco.config.replication, disco.discoConfig.bucketsCount,
     Interest,
   )
-  disco.rtManager.admitPeers(
-    disco,
-    serviceId,
-    @[PeerInfo(peerId: ad.data.peerId, addrs: ad.data.addresses.mapIt(it.address))],
-  )
+  # Admit the registering DHT peer, not the advertised subject. Under proxy
+  # advertising ad.data.peerId need not participate in KadDHT; the DHT peer is
+  # `advertiser`. Mirror seatSender and use its peerstore addresses.
+  let advertiserAddrs = disco.switch.peerStore[AddressBook][advertiser]
+  if advertiserAddrs.len > 0:
+    disco.rtManager.admitPeers(
+      disco,
+      serviceId,
+      @[PeerInfo(peerId: advertiser, addrs: advertiserAddrs)],
+    )
 
   disco.registrar.ads.put(serviceId, advertiser, ad, advertiserIps, now)
   disco.registrar.updateRegistrarMetrics()


richard-ramos:
I got this observation from codex, but not sure if valid:


- Medium — [registrar.nim:257](https://github.com/vacp2p/nim-libp2p/blob/bc03162c74e92b77f792b77c0aaa7d4fc157db74/libp2p/protocols/service_discovery/registrar.nim#L257): preserve advertisement-address fallback for non-proxy registrations. The new peerstore-only lookup skips admission when Identify has not yet populated addresses. When advertiser == ad.data.peerId, the signed advertisement still contains valid addresses belonging to that advertiser, so previously working direct registrations can now be cached without ever becoming service-table candidates. This also statically breaks the existing admission-probe test at [test_registrar.nim (line 1085)](/home/r/vacp2p/nim-libp2p/tests/libp2p/service_discovery/test_registrar.nim:1085), which supplies the address only through the advertisement. Suggested fix: prefer peerstore addresses, but fall back to ad.data.addresses only when advertiser == ad.data.peerId. Never use that fallback for proxy advertisements.

richard-ramos (approved):

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -251,11 +251,16 @@ proc acceptAdvertisement*(
...
-  disco.rtManager.admitPeers(
-    disco,
-    serviceId,
-    @[PeerInfo(peerId: ad.data.peerId, addrs: ad.data.addresses.mapIt(it.address))],
-  )
+  # Admit the registering DHT peer, not the advertised subject. Under proxy
+  # advertising ad.data.peerId need not participate in KadDHT; the DHT peer is
+  # `advertiser`. Mirror seatSender and use its peerstore addresses.
+  let advertiserAddrs = disco.switch.peerStore[AddressBook][advertiser]
+  if advertiserAddrs.len > 0:
```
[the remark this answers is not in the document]
gmelodie: seems legit

libp2p/protocols/service_discovery/registrar.nim
```diff
@@ -251,11 +251,16 @@ proc acceptAdvertisement*(
...
     Interest,
   )
-  disco.rtManager.admitPeers(
-    disco,
-    serviceId,
-    @[PeerInfo(peerId: ad.data.peerId, addrs: ad.data.addresses.mapIt(it.address))],
-  )
+  # Admit the registering DHT peer, not the advertised subject. Under proxy
+  # advertising ad.data.peerId need not participate in KadDHT; the DHT peer is
+  # `advertiser`. Mirror seatSender and use its peerstore addresses.
```
gmelodie: I don't think we need this
```suggestion
```

libp2p/protocols/service_discovery/registrar.nim
gmelodie, line 260: Why would advertiser not be on the peerstore? If this is supposed to always be true, some error handling/logging should happen here I think

libp2p/protocols/service_discovery/registrar.nim
jm-clius: Applied in [a5a2753](https://github.com/vacp2p/nim-libp2p/pull/3117/commits/a5a2753990fe7b83f62488b8a163df31a02c0411)

libp2p/protocols/service_discovery/registrar.nim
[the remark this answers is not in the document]
jm-clius: Yeah, tests was failing. Fixed in [a5a2753](https://github.com/vacp2p/nim-libp2p/pull/3117/commits/a5a2753990fe7b83f62488b8a163df31a02c0411)

jm-clius: Mmm, well, to me it seems the peer store management is sort of orthogonal? It seems likely that an advertiser would be in the peer store, but it might not have had acceptable announced addresses in `identify`. In my mind, nothing is specifically "expected" here - we are simply opportunistically adding possibly interesting peers. Without peer store entry, we can just assume they are not of interest.

gmelodie (approved):

PR #3119, opened 2026-09-14
chore(logs): standardize topics

vladopajic:
previously log topics where defined ad-hoc. in this pr log topics are unified so that:

- all files use following topics pattern`libp2p <component>`
   - `libp2p` is general topic for this library, downstream deps can silence all libp2p logs using this topic
   - `<component>` is topic that represents component for logs. this must be single word (kebab-case for readability). all utilities sub components and everything related must use only this topic (component topic). 
     - having too many topics will not help downstream deps, most likely they will not care about `utils` topic of `pubsub` component. 
- improved topic component names


libp2p/protocols/connectivity/autonatv2/verifier.nim
```diff
@@ -21,7 +21,7 @@ import
   ./utils
 
 logScope:
-  topics = "libp2p autonatv2 verifier"
+  topics = "libp2p autonat"
```
vladopajic, line 24: when user wants to debug `autonat` feature they should only care about `"autonat"` topic in logs. autonat could have been configured using `.withAutonatV2` or any `.withAutonnatV15` switch builder method, when debugging to them is should not matter. 
similar user should not care about sub components like "autonat-verifier" or "autonat-utils" as even using "autonat" as topic is advanced, and also "autonat-utils" requires knowledge of source code which should not be expected.

similary, all logs related to "gossipsub" should be under this topic, no matter what version of gossipsub is currently implemented.

richard-ramos (approved):

gmelodie (approved):

PR #3121, opened 2026-09-15
chore(logs): improve pubsub.Message shortLog

vladopajic:
- `seqno` logged as `uin64` or `"<unset>"` if not specified
  - `data` logged only with `data.len`, as data can potentially sensitive values
  - `signaturePresent` is not really necessary to logged as full information


libp2p/protocols/pubsub/rpc/messages.nim
```diff
@@ -210,12 +210,16 @@ chronicles.formatIt(ControlMessage):
 
 func shortLog*(msg: Message): auto =
   (
+    topic: msg.topic.shortLog,
+    seqno:
+      if msg.seqno.len > 0:
+        $fromBytesBE(uint64, msg.seqno)
+      else:
+        "<unset>",
```
richard-ramos:
```suggestion
      if msg.seqno.len == 8:
        $fromBytesBE(uint64, msg.seqno)
      elif msg.seqno.len == 0:
        "<unset>"
      else:
        "<invalid>",
```
To prevent a crash if the seqno is 1 < x < 8 (and assume those > 8 bytes to be invalid as well)

richard-ramos (approved):

gmelodie (approved):

PR #3122, opened 2026-09-15
chore(logs): add log summary for batched operation

vladopajic:
this pr aims to improve logging in case of batched operations. the pr change is that:
- all smaller bits of information (log events) inside main batched operations are logged as `trace`
- there is final log event, servers as action summary, which is logged as `debug`

this should result in bit lower noise if users are using `debug` log lvl - since those sub operations of batch will not be included in these logs. there will be only single `debug` log per batch / action. 

---

easiest example of a batch of single isolated change is in this commit: https://github.com/vacp2p/nim-libp2p/pull/3122/changes/47e0435961e8b97dd49c8624941822747e7088c1

----

but in nutshell:
```nim

proc processSubAction1(item: Item):
  # ... do something with item
  # but should never use any log lvl here except TRACE

proc processSubAction2(item: Item):
  # ... do something with item
  # but should never use any log lvl here except TRACE
  
proc someAction(items: seq[Item]):
   for item in items:
       processSubAction1(item)
       processSubAction2(item)
    
   debug "someAction completed", itemsCount: items.len, succeeded: ...., failed: ...
```

there is single `debug` log event in this action explaining outcome of this operation and providing information about what has happened. all additionally explaining log events are `trace`.


gmelodie (approved):

libp2p/autotls/service.nim
```diff
@@ -219,29 +219,35 @@ proc hasTcpStarted(switch: Switch): bool =
...
   for attempt in 0 .. self.config.issueRetries:
     if attempt > 0:
       await sleepAsync(self.config.issueRetryTime)
+    attempts.inc()
     try:
+      outcome = "issued"
       await self.issueCertificate()
       return
     except CancelledError as exc:
       raise exc
```
richard-ramos, line 237: Should probably change the outcome to "cancelled" here

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -487,32 +487,49 @@ proc handleData*(
...
+
+      debug "Topic handlers finished",
+        topic,
+        handlers,
+        succeeded = handlers - failed - cancelled - pending,
+        failed,
+        cancelled,
+        pending
+
+    return waiter()
```
richard-ramos:
```suggestion
    return waiter()
    
  newFutureCompleted[void]()
```

Otherwise, `handleData`  returns `nil`  if code does not enter the `withValue` at line 488, and having `{.async: (raises: [], raw: true).}` means that you gotta return a valid future (so u can `await` it)

libp2p/connmanager.nim
```diff
@@ -319,17 +319,36 @@ proc triggerConnEvent*(
 
   trace "Connection event callbacks started", peerId, event = $event.kind
 
+  var connEvents: seq[Future[void]]
```
richard-ramos: Curious about why replace the `newSeqOfCap` by `seq`

libp2p/protocols/pubsub/pubsub.nim
```diff
@@ -487,32 +487,49 @@ proc handleData*(
...
-            let err = fut.error()
-            warn "Error in topic handler", err = err.msg
-
-      return waiter()
-
-  # Fast path - futures finished synchronously or nobody cared about data
-  newFutureCompleted[void]()
+    if futs.len() == 0:
+      # Fast path - futures finished synchronously or nobody cared about data
+      debug "Topic handlers finished",
```
richard-ramos: This is likely  too noisy. `handleData` is called for each msg. Same for line 524

libp2p/autotls/service.nim
vladopajic: in general, the the best approach for complected summaries, like here with many exit points, is to have log  event in `defer`. in this case there is default `var outcome = "cancelled"` value which will be used if execution exits due to cancellation - then we do not have to think about cancellation as it is default case.

libp2p/connmanager.nim
vladopajic: it was unintentional

libp2p/protocols/pubsub/pubsub.nim
vladopajic: could be, but changed to trace

richard-ramos (approved):

PR #3123, opened 2026-09-15
chore(result): multistream negotiation and PeerInfo.init return Result

gmelodie:
This is step 2 of the raises-to-Result migration after #3084. `trySelect`, `tryHandle` and `PeerInfo.tryNew` return `Result`, so internal callers handle a malformed message or a bad handshake without an exception. `select`, `handle` and `PeerInfo.new` keep their signatures and raise the same error text through `valueOrRaise`.


libp2p/peerinfo.nim
```diff
@@ -169,7 +169,7 @@ proc toFullAddress*(peerId: PeerId, ma: MultiAddress): MaResult[MultiAddress] =
   let peerIdPart = ?MultiAddress.init(p2pMultiCodec, peerId.data)
   concat(ma, peerIdPart)
 
-proc new*(
+proc init*(
```
richard-ramos: Wrt lines 172 and 201, 
https://status-im.github.io/nim-style-guide/language.objconstr.html

the `new` procs are used for returning `ref` objects, and `init` for non ref... makes me wonder if the `init` should be called `tryNew`.

libp2p/peerinfo.nim
```diff
@@ -200,3 +197,21 @@ proc new*(
     addressMappers: addressMappers,
     addressPolicy: addressPolicy,
   )
+
+proc new*(
```
richard-ramos, line 201: Is the `new` intended to be function marked with deprecated?

libp2p/multistream.nim
```diff
@@ -34,58 +41,62 @@ type
...
+  if proto.len() == 0:
+    return ok(Codec)
+
+  let response = ?(await stream.readMessage())
+  trace "Protocol negotiation response received", stream, protocol = response
+  if response == proto[0]:
+    trace "Protocol negotiation completed", stream, protocol = proto[0]
+    stream.protocol = proto[0]
+    return ok(proto[0])
+
```
richard-ramos, line 80: Could we keep the trace log for 
```
      trace "Protocol alternatives available", stream, protocols = protos
``` ?

File: libp2p/peerinfo.nim (hunk 2 above)
Comment (gmelodie) [reply libp2p/peerinfo.nim:201]: No need, new is supposed to use the new `tryNew`

**Review by richard-ramos — APPROVED** [review]

File: libp2p/multistream.nim [hunk 4]
```diff
@@ -34,58 +41,64 @@ type
...
-  else:
-    raise (ref MultiStreamError)(msg: "MultistreamSelect failed, malformed message")
+func decodeMessage(data: openArray[byte]): MultiStreamResult[string] =
+  if data.len == 0 or data[^1] != byte('\n'):
+    return err(MultiStreamFailure.MalformedMessage)
 
-proc select*(
+  ok(string.fromBytes(data.toOpenArray(0, data.high - 1)))
+
+proc readMessage(
```
vladopajic: `readMessage` here servers as pattern for "extract method". it is a case when a bit code is extracted to function (or proc, whatever methodology) so this code is named (describes what it does) so it servers as building block in code and also improves redability, etc..

in cases where extracted code is really small, one liner, i think i would be better to do it as template, but not all cases should be a template. with template we still keep same readability / maintainability benefits and remain same performance (before refactor).

some cases when to use template:
  - as a hack to avoid `inline` pragma
  -  for one liners that have `async`  pragma 
    - this will improve (remain old) performance a bit  
  - for one liners in general
     - not all one liners need to be template
 ^ one liners above can be two-three liners as well

we should only care this in library code due to performance imacpt


---

this case is nice pattern for other location to do extract method simillary, to have `readMessage` template that does `await stream.readLp(MsgSize)` or decode it as well. 

in other locations it does not have to be `readMessage` it can be `readFooMessage`, `readBarMessage` if there are many types of messages in file (or package)

vladopajic: tldr: use template

libp2p/peerinfo.nim
```diff
@@ -179,16 +179,13 @@ proc new*(
...
+): Result[PeerInfo, string] =
   let pubkey = key.getPublicKey().valueOr:
-    raise
-      newException(PeerInfoError, "invalid private key creating PeerInfo: " & $error)
+    return err("invalid private key creating PeerInfo: " & $error)
   let peerId = PeerId.init(pubkey).valueOr:
-    raise newException(
-      PeerInfoError, "invalid public key creating PeerInfo peer id: " & $error
-    )
+    return err("invalid public key creating PeerInfo peer id: " & $error)
```
vladopajic: this massage is not correct
`invalid public key creating PeerInfo`
what is ` peer id` part was meant to be here for? is it `peer_id=$peerId`, or ?

libp2p/peerinfo.nim
```diff
@@ -179,16 +179,13 @@ proc new*(
     addressMappers = newSeq[AddressMapper](),
     addressPolicy: PeerAddressPolicy = defaultAddressPolicy,
     announcedAddrs: openArray[MultiAddress] = [],
-): PeerInfo {.raises: [LPError].} =
+): Result[PeerInfo, string] =
   let pubkey = key.getPublicKey().valueOr:
-    raise
-      newException(PeerInfoError, "invalid private key creating PeerInfo: " & $error)
+    return err("invalid private key creating PeerInfo: " & $error)
```
vladopajic: it would be nice if we can come up with some standardized way for error messages.

`"invalid private key creating PeerInfo"`
-> 
`"PeerInfo.tryNew called with invalid private key. " & $error`

- start with loc identifier
- do not name what this location does in the case avoid `creating` - sometimes it will be hard to explain, `creating` here is easy 
- i would prefer to end message with `". "` incited of `": "` if there is error following it this principle creates following pattern:
   -   "PeerInfo.tryNew called with invalid private key. PrivateKey.getPublicKey is nil" (for example)
   - this message on upper layer will be:
    "MyService.validation called with invalid config. PeerInfo.tryNew called with invalid private key. PrivateKey.getPublicKey is nil"

vladopajic: ^ do not follow this blindly, take it as idea and improve upon it or discard it

libp2p/upgrademngrs/upgrade.nim
```diff
@@ -4,7 +4,7 @@
 {.push gcsafe.}
```
vladopajic, line 4: adding comment here: pr title is labeled as "feat". 

`feat` should mean new feature (added functionality) to downstream users. 

some prs can be labled as `refactor` or `chore` where those changes could be used as new feature (by downstream users). for example after refactoring logs where `error` log really means error, it can be used as feature by user as user can react on that pr and crate some kind of monitoring. so sometimes is is hard to cut it (without clear systematisation what it really means in nim-libp2p).
i would keep this pr as a `chore`. i would use `refactor` when there is significant change is logic, `feat` when user can do something new with `Swithch/SwitchBuilder`. `refactor` for all changes that also depracte old and introduce new things. ... we could work on this

tests/libp2p/test_multistream.nim
```diff
@@ -675,3 +675,56 @@ suite "Multistream :: stream limits":
...
+    let handling = MultistreamSelect.handle(server, @[codecs])
+    defer:
+      await client.close()
+      await server.close()
+    await client.writeLp(codecs & "\n")
+    try:
+      discard await handling.wait(1.seconds)
+      raiseAssert "handle must raise"
+    except MultiStreamError as e:
+      check e.msg == $MultiStreamFailure.InvalidFirstMessage
```
vladopajic: if you can make this code to be like this 

```
e == MultiStreamFailure.InvalidFirstMessage
```

it would be awesome

idealy only change in `==` operator or 

`MultiStreamFailure* {.pure.} = enum` -> `MultiStreamFailure* = errorEnum` ????? not sure if this is possible but create some type of enum and inherit that type so `==` does not have to be added always. or create macro to add `==` operator for this type

this is just ideal

vladopajic (approved):

PR #3124, opened 2026-09-15
fix(service-disco): bind stored and fetched XPRs to their key

gmelodie:
`ServiceDiscovery.new` kept the validator of the passed `KadDHTConfig` (cbind passes `DefaultEntryValidator`), and `lookupRandom` decoded each `getValue` reply without a check against the queried peer. Now `ServiceDiscovery.new` always installs `ExtEntryValidator` and `ExtEntrySelector`, and both paths call `boundXpr`, which accepts an XPR only when its signer is the peer that the key names and its service data passes `isValid`.
New tests in `test_types.nim`, `test_error.nim` and `test_find_random.nim` cover each rejected case; `nim check` passes on them. Check that no caller needs a custom validator on service discovery, because the config value is now replaced.


tests/libp2p/service_discovery/test_types.nim
```diff
@@ -49,3 +49,19 @@ suite "seq[Advertisement] encode":
       check:
         decoded.isOk()
         decoded.get() == origAds[i]
+
+suite "ExtEntryValidator":
+  test "accepts a record signed by the peer its key names":
+    let ad = makeAdvertisement()
+    let record = EntryRecord(value: Value.fromBytes(ad.encode()), time: Timestamp.now())
```
vladopajic: `value: Value.fromBytes(ad.encode()` 
or 
`value: ad.encode()`

there should be converter from `seq[bytes]` to `Value`, so we don't have to do it manually in tests.
all converters are in `nim-libp2p/tests/libp2p/kademlia/utils.nim` file

libp2p/protocols/service_discovery/random_find.nim
```diff
@@ -11,6 +11,23 @@ import ./[types]
...
+type PendingGetVal = object
+  key: Key
+  fut: Future[Result[Message, string]].Raising([CancelledError])
+
+proc replyXpr*(key: Key, reply: Message): Opt[SignedExtendedPeerRecord] =
+  let record = reply.record.valueOr:
+    return Opt.none(SignedExtendedPeerRecord)
+
+  if record.key != Opt.some(key):
+    debug "Get-value reply names another key", key
```
vladopajic: `debug` -> `trace`, since `trace` is used for log events of higher frequency. here `replyXpr` will be called many times for each `PendingGetVal`.

vladopajic: i get it now, this was copy paste and this was changed in pr that is about to get merged

vladopajic (approved):

cbind/tsan.supp
```diff
@@ -1,3 +1,6 @@
 # nimcrypto picks its SHA-2 compress function on first use and caches it in a
 # global, so every thread writes the same pointer at once.
 race:default_sha*_compress_func
+
+# localtime_r guards its TZ cache with a glibc-internal lock that TSAN cannot see.
+race:getLocalOffsetAndDst
```
richard-ramos: Was this commited by error?

richard-ramos (approved):

